PR #4601, opened 2015-11-11
Update DraftReg

samchrisinger:
# Purpose

There are a couple conditions where while a user is editing a draft they will receive errors from the server: 
  -- the user is no longer an admin on the project (permissions changed, removed from project, etc)
  -- the draft is registered by another admin
  in these cases the user's requests will error can a growlbox error should appear. No description was being passed to the GrowlBox constructor, and consequently the growl was appearing without a background. 

This required two changes:
  - give a title/description to the growlbox constructor
  - check that the draft is not registered on `edit_draft_registration_page` and `update_draft_registration`

Also, remove AddonModelMixin from DraftRegistration as it is unused

Additionally, the Draft Reg views never respected the HF in https://github.com/CenterForOpenScience/osf.io/pull/4571. See https://github.com/CenterForOpenScience/osf.io/commit/158b120eceb3b89d366cb31d04a5a67dc13e9fb9 for details.

### DIFF
--- a/api_tests/nodes/views/test_node_registrations_list.py
+++ b/api_tests/nodes/views/test_node_registrations_list.py
@@ -19,12 +19,12 @@ class TestNodeRegistrationList(ApiTestCase):
         self.user = AuthUserFactory()
 
         self.project = ProjectFactory(is_public=False, creator=self.user)
-        self.registration_project = RegistrationFactory(creator=self.user, project=self.project)
+        self.registration_project = RegistrationFactory(creator=self.user, project=self.project, is_public=True)
         self.project.save()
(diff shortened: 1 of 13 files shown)


samchrisinger:
@felliott I think in this case, since the user still has admin access to the resource, but it is no longer editable, a 400 may be more appropriate.

felliott:
@samchrisinger, 400s tend to be more for malformed requests or requests that don't make any sense at all.  In this case, the request is reasonable, we just don't permit it.  A 400 also implies that the client can adjust the request to get it working.  A 403 says "Your request is fine, but no, just no.".  A 403 also means that authentication will not make a difference, which is the case here.

PR #4604, opened 2015-11-11
[Feature Fix] GitHub show wrong branch file

haoyuchen1992:
## Purpose
When clicking view file on github, it will always lead user to the master file link. However, github has different branches, and therefore the file linked is not the one rendering in MFR.

In this PR, it will create the correct link url for file (with current branch).  

## Changes
Firstly, get the url for the branch. And then use ajax and get file object from waterbulter. The ajax ur is something like -- `http://localhost:7777/data?path=%2Fmain.py&provider=github&nid=9j5hu&branch=redesign` 
screenshot 2015-11-11 11 02 24

In this object, we can get the webView url from file object.  Then assign it to  `window.contextVars.file.urls.external`

## Side Effect
None. This is limited to github.

### DIFF
--- a/website/static/js/filepage/index.js
+++ b/website/static/js/filepage/index.js
@@ -134,6 +134,18 @@ var FileViewPage = {
         });
 
         if ($osf.urlParams().branch) {
+            var newFileBranchUrl = waterbutler.buildMetadataUrl(self.file.path, self.file.provider, self.node.id, {branch : $osf.urlParams().branch});
+            $.ajax({
+                dataType: 'json',
+                async: true,
+                url: newFileBranchUrl,
+                beforeSend: $osf.setXHRAuthorization,
+                success:function(response) {
+                    window.contextVars.file.urls.external = response.data.extra.webView;
+                    console.log(response);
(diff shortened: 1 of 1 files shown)


mfraezz:
There may be another way of constructing the `external` url, but given that it's the definitive link according to github, the extra request is probably merited. 

Looks good to me. Pass complete :shipit:

website/static/js/filepage/index.js
```diff
@@ -134,6 +134,16 @@ var FileViewPage = {
         });
 
         if ($osf.urlParams().branch) {
+            var fileWebViewUrl = waterbutler.buildMetadataUrl(self.file.path, self.file.provider, self.node.id, {branch : $osf.urlParams().branch});
+            $.ajax({
+                dataType: 'json',
+                async: true,
+                url: fileWebViewUrl,
+                beforeSend: $osf.setXHRAuthorization,
+                success:function(response) {
```
sloria: As per COSDev, prefer the promise interface (`.done() or .then()`) to the `success` callback.

I will fix this when I merge.

PR #4492, opened 2015-10-21
Disconnect signal for adding contributors email in merge_users function

hmoco:

PR #4607, opened 2015-11-12
Disconnect signal through context_manager

hmoco:
https://github.com/CenterForOpenScience/osf.io/pull/4492

### DIFF
--- a/framework/auth/core.py
+++ b/framework/auth/core.py
@@ -30,6 +30,7 @@ from framework.sessions.model import Session
 from framework.sessions.utils import remove_sessions_for_user
 
 from website import mails, settings, filters, security
+from website.util import disconnected_signal
 
 name_formatters = {
     'long': lambda user: user.fullname,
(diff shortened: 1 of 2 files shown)


website/util/__init__.py
```diff
@@ -180,3 +181,9 @@ def waterbutler_api_url_for(node_id, provider, path='/', **kwargs):
     url.path.segments.extend([urllib.quote(x.encode('utf-8')) for x in segments])
     url.args.update(kwargs)
     return url.url
+
+@contextmanager
+def disconnected_to(signal, listener):
```
sloria: Should be `disconnected_from`. Will change this when I merge.

framework/auth/core.py
```diff
@@ -1231,45 +1231,46 @@ def merge_user(self, user):
         # be sure to reconnect it at the end of this code block. Import done here to prevent circular import error.
         from website.project.signals import contributor_added
         from website.project.views.contributor import notify_added_contributor
-        contributor_added.disconnect(notify_added_contributor)
+        from website.util import disconnected_to
```
sloria: This shouldn't need to be imported here.

sloria: Disregard; moving this leads to a circular import. No change needed.

PR #4609, opened 2015-11-12
Added Academic.edu and researchgate.net to social tab on profile page OSF-4995, OSF-2427

sbt9uc:
Purpose
------------

To disable Ace Auto Complete so that it does not pop up on the first character of every word that is typed.

Changes
-----------------

Disabled Ace Live Auto complete. Ace Basic Auto Complete is still enabled, which means the feature is still there, but must be prompted each time by the user.

Side Effects
----------------

Auto Complete no longer pops up for users that need it. It can still be accessed with 'control' + 'space', but for users that rely heavily on this feature, that can be tedious.


samanehsan:
Can you look into the failing tests please? :octocat:

sbt9uc:
Yeah, I'm trying to fix them. I think I took the tickets off of CR1.

samanehsan:
While you're looking into the failing tests, can you also look into why the settings page doesn't allow you to save an Academia.edu or ResearchGate link? I think they are related errors.

sloria:
@sbt9uc Please resend this to `develop` if this PR is still relevant.

GaryKriebel:
This has been picked up by another intern and will probably see a PR then.

PR #4613, opened 2015-11-12
[hotfix] otp

icereval:


api/base/authentication/drf.py
```diff
@@ -67,12 +73,25 @@ def authenticate_credentials(self, userid, password):
...
 
+    def authenticate_twofactor_credentials(self, user, request):
+        """
+        Authenticate the user's two-factor one time password code.
+        """
+        user_addon = user.get_addon('twofactor')
+        if user_addon and user_addon.is_confirmed:
+            otp = request.META.get('HTTP_X_OSF_OTP')
+            if otp is None:
+                raise TwoFactorRequiredError(_('Must specify two-factor authentication OTP code.'))
```
sloria: Rather than pass the `detail` argument here, better to override `default_detail` for `TwoFactorRequiredError` for consistent error messages.

sloria: I'll make this change when I merge.

PR #4614, opened 2015-11-12
Add Society for Marine Mammalogy 2015 to OSF4M

jolene-esposito:


scripts/populate_conferences.py
```diff
@@ -540,9 +540,18 @@ def main():
         'poster': True,
         'talk': True,
     },
+        'SMM2015': {
```
sloria: Indent by 4--not 8-- spaces here (indentation is significant in Python).

I will fix this when I merge.

PR #4594, opened 2015-11-10
[WIP] Draft Registrations

samchrisinger:
Continuation of: https://github.com/CenterForOpenScience/osf.io/pull/4428

## Migration:
`python -m scripts.migrations/migrate_registered_meta`

PR #4615, opened 2015-11-12
[WIP] Update Draft Reg. Migration

samchrisinger:
# Purpose

The core Draft Reg PR assumes that Registrations have only one registered_schema. As it turns out there are 12 Nodes on production with two. How this is possible is TDB, but we need to accommodate this.

# Changes

- Update Node model to make Node.registered_schema a list field
- Update migration to account for registrations with multiple schemas
- Update new metaschema json files for a 1-to-1 mapping of question ids from old to new schemas 
- Update views to support new model behavior
- Update frontend support new model behavior
- Fix failing tests

# TODO
- [x] update migration tests
- [x] update draft registration editor frontend and views (and probably broken tests)

### DIFF
--- a/scripts/migration/migrate_registered_meta.py
+++ b/scripts/migration/migrate_registered_meta.py
@@ -2,7 +2,6 @@
 Changes existing registered_meta on a node to new schema layout
 required for the prereg-prize
 """
-from nose.tools import *  # noqa PEP8 asserts
 import json
 import sys
 import logging
@@ -75,13 +74,12 @@ def main(dev=False, _db=None):
                     Q('name', 'eq', name)
                 ).sort('-schema_version')[0]
             except IndexError as e:
-                logger.error('No MetaSchema matching name: {0} found'.format(name))
+                logger.error('No MetaSchema matching name: {0} found for node: {1}.'.format(name, node['_id']))
                 # Skip over missing schemas
                 skipped += 1
                 if dev:
                     continue
                 else:
-                    import ipdb; ipdb.set_trace()
                     raise e
             else:
                 registered_meta[meta_schema._id] = {
--- a/scripts/migration/migrate_registered_meta.py
+++ b/scripts/migration/migrate_registered_meta.py
@@ -10,13 +10,15 @@ import logging
 from modularodm import Q
 from modularodm.exceptions import NoResultsFound
 
+from framework.mongo import database as db
+nodes = db['node']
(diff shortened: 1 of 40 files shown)


scripts/migration/migrate_registered_meta.py
```diff
@@ -71,14 +86,10 @@ def main(dry_run, dev=False):
...
-                'embargoEndDate': schema.get('embargoEndDate', ''),
-                'registrationChoice': schema.get('registrationChoice', ''),
-            }
-            schema_data.update(schema)
+            # append matching schema to node.registered_schema
             try:
                 meta_schema = MetaSchema.find_one(
-                    Q('name', 'eq', _name_to_id(name)) &
+                    Q('name', 'eq', name) &
                     Q('schema_version', 'eq', 2)
```
sloria: Why `2`?

scripts/migration/migrate_registered_meta.py
```diff
@@ -71,14 +86,10 @@ def main(dry_run, dev=False):
...
                 )
             except NoResultsFound:
```
sloria: Again, why is `2` hardcoded in the line below?

scripts/migration/migrate_registered_meta.py
```diff
@@ -8,109 +8,103 @@
...
                 skipped += 1
-                continue
-            node.registered_schema = meta_schema
-            node.registered_meta = {
-                key: {
-                    'value': value
+                if dev:
+                    continue
+                else:
+                    import ipdb; ipdb.set_trace()
```
sloria: Remove this.

scripts/migration/migrate_registered_meta.py
```diff
@@ -8,109 +8,103 @@
...
-                    Q('name', 'eq', _name_to_id(name)) &
-                    Q('schema_version', 'eq', 2)
-                )
-            except NoResultsFound:
-                logger.error('No MetaSchema matching name: {0}, version: {1} found'.format(name, 2))
+                meta_schema = MetaSchema.find(
+                    Q('name', 'eq', name)
+                ).sort('-schema_version')[0]
+            except IndexError as e:
+                logger.error('No MetaSchema matching name: {0} found'.format(name))
```
sloria: It would be helpful to log the node ID here.

sloria:
Closing in favor of #4594

PR #4616, opened 2015-11-13
[WIP]Permissions and User Invites

zkraime:
<b>Purpose: </b> add user invitation functionality to django-admin

<b> To do 
- [x] Add base code for functionality, including views and admin.py
- [ ] Add templates for emails
- [x] Basic code clean up - including merging updates from user_login branch

Create users/set permissions from django-admin:

screen shot 2015-08-21 at 6 27 57 am

Select + send email invitations:

screen shot 2015-08-21 at 6 29 35 am

Email: 
screen shot 2015-08-24 at 5 38 08 pm

Then:
screen shot 2015-08-24 at 5 41 10 pm

Then redirect:
screen shot 2015-08-24 at 5 41 19 pm

Notes:
User created inactive and is activated after reseting their password

### DIFF
--- a/admin/base/settings/defaults.py
+++ b/admin/base/settings/defaults.py
@@ -95,20 +95,7 @@ TEMPLATES = [
         }
     }]
 
-# Database
-# Postgres:
-# DATABASES = {
-#     'default': {
-#         'ENGINE': 'django.db.backends.postgresql_psycopg2',   # Add 'postgresql_psycopg2', 'mysql', 'sqlite3' or 'oracle'.
-#         'NAME': 'mydb',                                       # Or path to database file if using sqlite3.
-#         # The following settings are not used with sqlite3:
-#         'USER': 'myuser',
-#         'PASSWORD': 'password',
-#         'HOST': 'localhost',                                  # Empty for localhost through domain sockets or           '127.0.0.1' for localhost through TCP.
-#         'PORT': '',                                           # Set to empty string for default.
-#     }
-# }
-# https://docs.djangoproject.com/en/1.8/ref/settings/#databases
+# Postgres settings in local.py
 
 DATABASES = {
     'default': {
--- a/admin/base/settings/defaults.py
+++ b/admin/base/settings/defaults.py
@@ -26,11 +26,7 @@ ALLOWED_HOSTS = [
 ]
 
 # Email settings. Account created for testing. Password shouldn't be hardcoded
-EMAIL_HOST = 'smtp.gmail.com'
-EMAIL_HOST_USER = 'openscienceframeworktest@gmail.com'
-EMAIL_HOST_PASSWORD = ''
-EMAIL_PORT = 587
-EMAIL_USE_TLS = True
+EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 
 # Application definition
 
--- a/admin/base/settings/defaults.py
+++ b/admin/base/settings/defaults.py
@@ -25,9 +25,6 @@ ALLOWED_HOSTS = [
     '.osf.io'
 ]
 
-# Custom user model (extends AbstractBaseUser)
-AUTH_USER_MODEL = 'common_auth.MyUser'
-
 # Email settings. Account created for testing. Password shouldn't be hardcoded
 EMAIL_HOST = 'smtp.gmail.com'
 EMAIL_HOST_USER = 'openscienceframeworktest@gmail.com'
@@ -52,6 +49,9 @@ INSTALLED_APPS = (
     'raven.contrib.django.raven_compat',
     'webpack_loader',
 )
+
+# Custom user model (extends AbstractBaseUser)
+AUTH_USER_MODEL = 'common_auth.MyUser'
 
 # TODO: Are there more granular ways to configure reporting specifically related to the API?
 RAVEN_CONFIG = {
--- a/admin/base/settings/defaults.py
+++ b/admin/base/settings/defaults.py
@@ -25,6 +25,9 @@ ALLOWED_HOSTS = [
     '.osf.io'
 ]
 
+# Custom user model (extends AbstractBaseUser)
+AUTH_USER_MODEL = 'common_auth.MyUser'
+
 # Email settings. Account created for testing. Password shouldn't be hardcoded
 EMAIL_HOST = 'smtp.gmail.com'
 EMAIL_HOST_USER = 'openscienceframeworktest@gmail.com'
@@ -45,6 +48,7 @@ INSTALLED_APPS = (
     'admin.spam',
 
     # 3rd party
+    'django_extensions',
     'raven.contrib.django.raven_compat',
     'webpack_loader',
 )
--- a/admin/base/settings/defaults.py
+++ b/admin/base/settings/defaults.py
@@ -41,11 +41,10 @@ INSTALLED_APPS = (
     'django.contrib.messages',
     'django.contrib.sessions',
     'django.contrib.staticfiles',
-    'admin.base',
+    'admin.common_auth',
     'admin.spam',
 
     # 3rd party
-    'admin.common_auth',
     'raven.contrib.django.raven_compat',
     'webpack_loader',
 )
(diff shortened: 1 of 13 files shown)


zkraime:
@cwisecarver  Please review this; thank you!

zkraime:
@cwisecarver Please review this; thank you!

admin/base/settings/defaults.py
```diff
@@ -25,18 +25,34 @@
     '.osf.io'
 ]
 
+# Email settings. Account created for testing. Password shouldn't be hardcoded
```
cwisecarver: Might be worth looking into this https://docs.djangoproject.com/en/dev/topics/email/#console-backend OSF uses a similar concept for development.

zkraime: @cwisecarver I removed the hardcoded password - I figured I should just use a local.py constant

cwisecarver: I'm sure that's fine too. I was just suggesting that maybe you don't want dev to send emails at all. That link should let you add a line to the settings file and any requested emails will just print out in the server console instead of actually sending an email. That's what OSF does when you sign up for an account. The confirmation email just prints to the server console. That line could be overwritten in the local.py for the staging and live servers.

cwisecarver:
There aren't any migrations for common_auth. I ran makemigrations and migrated common_auth and everything worked.

cwisecarver:
Right now, I'm getting `KeyError at /admin/auth/login/ u'app.min.js'` on login.html around `{% render_bundle 'app.min' %}` I'm trying to figure out why it's erring but my webpack-fu is not that strong.

cwisecarver:
I changed `{% render_bundle 'app.min' %}` to `{% render_bundle 'admin-base-page' %}` in both login.html and base.html and it started working.

admin/common_auth/admin.py
```diff
@@ -0,0 +1,63 @@
...
+from django.contrib.auth.admin import UserAdmin
+from django.contrib.auth.models import Permission
+from .models import MyUser
+
+from django.contrib.auth.forms import PasswordResetForm, UserCreationForm
+
+class PermissionAdmin(admin.ModelAdmin):
+    search_fields = ['name', 'codename']
+
+class CustomUserRegistrationForm(UserCreationForm):
```
cwisecarver: Form classes should be in forms.py. I know the custom admin example on the django site says to do this. It's a bad example :penguin:

admin/common_auth/admin.py
```diff
@@ -0,0 +1,63 @@
...
+        (None, {'fields':
+                ('email', 'first_name', 'last_name', 'password1', 'password2'),
+                }),)
+    search_fields = ('email', 'first_name', 'last_name',)
+    ordering = ('last_name', 'first_name',)
+    actions = ['send_email_invitation']
+
+    def send_email_invitation(self, request, queryset):
+        for user in queryset:
+            if user.is_active is False:  # email doesn't send unless user is active
```
cwisecarver: Loop through all the selected users, If the user is not active make them active and save them. Create a reset form for their user, save the form, deactivate them, and save them again. By setting them to is_active = False regardless of what they used to be you're disabling everyone you send invitations to. Is that intended functionality?

admin/common_auth/forms.py
```diff
@@ -0,0 +1,5 @@
+from django import forms
+
+class LoginForm(forms.Form):
+    username = forms.CharField(label=(u'User Name'), required=True)
```
cwisecarver: It may or may not become confusing later that the form refers to this field as username and the models and manager all refer to it as email.

admin/templates/base.html
```diff
@@ -0,0 +1,110 @@
...
+      </footer>
+
+    </div><!-- ./wrapper -->
+
+    <!-- REQUIRED JS SCRIPTS -->
+
+    <!-- Bootstrap 3.3.2 JS -->
+    {% render_bundle 'vendor' %}
+    <!-- AdminLTE App -->
+    {% render_bundle 'app.min' %}
```
cwisecarver: I couldn't get this line to not 500. I ended up changing it to `{% render_bundle 'admin-base-page' %}` both here and in login.html.

admin/base/views.py
```diff
@@ -1,5 +1,29 @@
...
+    messages.success(request, 'You have successfully reset your password and activated your admin account. Thank you')
+    return redirect('auth:login')
+
+def password_reset_confirm_custom(request, **kwargs):
+    response = views.password_reset_confirm(request, **kwargs)
+    # i.e. if the user successfully resets their password
+    if response.status_code == 302:
+        try:
+            uid = urlsafe_base64_decode(kwargs['uidb64'])
+            user = User.objects.get(pk=uid)
```
cwisecarver: I think this should reference MyUser not User. After clicking the link included in the reset email I got a 500. `Manager isn't available; User has been swapped for 'common_auth.MyUser'`

admin/base/views.py
```diff
@@ -1,5 +1,29 @@
...
+        except (TypeError, ValueError, OverflowError, User.DoesNotExist):
```
cwisecarver: Probably the same thing with this `User.DoesNotExist`.

cwisecarver:
I'm done @zkraime. Back to you. Let me know if you have questions about anything.

cwisecarver:
Remembered this morning that django-extensions is in your installed apps but not in the requirements file.

admin/base/views.py
zkraime: @cwisecarver Ah you're right, thank you

admin/common_auth/admin.py
zkraime: @cwisecarver Noted - thanks!

admin/common_auth/admin.py
zkraime: @cwisecarver Basically I wanted to make sure that any accounts we sent invitations to would be inactive until a password reset. 
But certainly this does not cover the case where somebody had already reset their password but wanted to do so again. If they received the email and don't change their password, their account would be inactive even though it should be active. 
Perhaps we can circumvent this by adding a new confirmed flag to determine if a user has had an initial password reset?

cwisecarver: The confirmed flag makes sense to me and I believe it would solve this.

admin/common_auth/forms.py
zkraime: @cwisecarver I had noticed this and should have looked into it deeper - I can just change the naming here. But after I get the post variable, and call authenticate, I still need to say 
authenticate(username=email, password=password)  
because MyUser uses email as username_field, right? 
i.e. it wouldn't be authenticate(email=email, password=password)

cwisecarver: Correct. The kwarg name has to be username. so `authenticate(username=email, password=password)` will work. 

I'm 99% sure you could do `authenticate(email, password)`  and everything would be fine. The first example is more explicit that you're shoving email into the username field.

IIRC keyword args in python are just like regular arguments except you have the option to ignore position. So if you wanted to be mean you could do `authenticate(password=password, username=email)` but you couldn't do `authenticate(password, email)`.

admin/base/settings/defaults.py
zkraime: @cwisecarver Ah I understand. 
Would I be able to test the password reset link from the console? Like just copy and paste it?

cwisecarver: Exactly. It does all the template processing for the email and then instead of sending it via SMTP it prints the rendered email template to the console.

zkraime:
@cwisecarver Thank you very much for the review,
everything should be addressed. 
Let's maybe chat briefly Friday afternoon for a generate update

cwisecarver:
@zkraime You don't have any migrations for common_auth. Or at least you haven't committed any. I fixed this in order to test on the first pass but forgot to put a note in here. If you do have migrations on your local I would suggest blowing them all away and creating one from scratch just to make it simpler.

cwisecarver:
@zkraime Added a couple of more comments some things I missed on the first pass. Let me know if you have questions.

PR #4600, opened 2015-11-10
Add embed= support to API endpoints

pattisdr:
# Purpose

I am finishing up @samchrisinger's PR https://github.com/CenterForOpenScience/osf.io/pull/4260

> JSONAPI defines a way to include related resources in a request (http://jsonapi.org/format/#fetching-includes). For us this means we could make a request like /api/v2/nodes/abcdef/?include=children to fetch a node and its children in a single requests.

> Unfortunately the JSON-API spec has some drawbacks, so we're implementing a similar feature: embed. This directly embeds the view response for the related resource in the API response.

# Changes

> Implement JSONAPIBaseView to generically handle
injection of the request into the serializer context
injection of a dictionary of zero or more partial functions ( object -> dict) into the serializer context. These functions are called in the JSONAPISerializer's to_representation method when appropriate.
Add JSONAPIBaseView to existing view classes

Rewrite JSONAPIPagination (my changes)
By default, always embed node links and contributors

- Make sure that embedded pagination links are for embedded resource, not parent resource: 
url was http://localhost:8000/v2/users/?embed=nodes.  URL's link to user nodes list: Embeds parameter is not included for embedded resource.
screen shot 2015-11-11 at 1 40 03 pm

- When visiting second page of results, http://localhost:8000/v2/nodes/?embed=contributors&page=2, the page=2 filter does not extend to embedded resources:

- When filtering, filtering not attempted on embedded resources.

# Side effects

PR #4619, opened 2015-11-16
[WIP] Feature/remove contributors multiple components

billyhunt:
### Purpose

Give users the choice to remove a user from either the current node or from the current node and all its children.

#### Current:
screen shot 2015-11-16 at 10 26 31 am

#### Proposed:

screen shot 2015-11-16 at 10 53 50 am

screen shot 2015-11-16 at 10 54 25 am copy

### Changes:

* Delete the user from the given node as a default with a modal.
* If the Admin wants to remove the Contributor from the node AND its children, Open a second modal view, and use the API V2 to list of every component associated with the Contributor to remove.
*  If the Admin clicks "Remove" on the second view, use the API v2 to delete those contributors from all child nodes previously listed.

### TODO
*  Use the new Embed feature of the API to get the list of nodes so we can delete the Contributor from all the relevant children, then delete those dogs.


billyhunt:
Currently blocked by #4600

PR #4219, opened 2015-08-31
[RfR] API logs endpoint

samchrisinger:
# Purpose

Add API endpoints for listing NodeLogs

# Changes

- Add /logs/ endpoint for listing NodeLogs
- Add /logs/{log_id}/nodes endpoint for listing Nodes related to a given NodeLog
- Add /nodes/{node_id}/logs for listing NodeLogs related to a given Node

# Side Effects

None

**Sample API Response**
```json
{
  "data": [
     {
            "id": "55df519d5ac80d15b5392234",
            "date": "2015-08-27T18:06:21.204000",
            "action": "retraction_approved",
            "params": {
                "node": "9yrst",
                "retraction_id": "55df517b5ac80d15b5392231"
            },
            "links": {
                "nodes": {
                    "related": "http://localhost:8000/v2/logs/55df519d5ac80d15b5392234/nodes/"
                },
                "user": {
                    "related": "http://localhost:8000/v2/users/j84ad/"
                }
            },
            "type": "logs"
        },
    ]
}
```

screen shot 2015-08-31 at 3 59 32 pm

```json
{
  "data": [
    {
      "id": "23ehy",
      "type": "nodes",
      "attributes": {
        "title": "test",
        "description": "",
        "category": "project",
        "date_created": "2015-08-27T18:04:25.097000",
        "date_modified": "2015-08-31T18:24:22.065000",
        "tags": [],
        "registration": false,
        "collection": false,
        "dashboard": false,
        "public": false
      },
      "relationships": {
        "children": {
          "links": {
            "related": {
              "href": "http://localhost:8000/v2/nodes/23ehy/children/",
              "meta": {
                "count": 1
              }
            }
          }
        },
        "contributors": {
          "links": {
            "related": {
              "href": "http://localhost:8000/v2/nodes/23ehy/contributors/",
              "meta": {
                "count": 1
              }
            }
          }
        },
        "files": {
          "links": {
            "related": "http://localhost:8000/v2/nodes/23ehy/files/"
          }
        },
        "node_links": {
          "links": {
            "related": {
              "href": "http://localhost:8000/v2/nodes/23ehy/node_links/",
              "meta": {
                "count": 0
              }
            }
          }
        },
        "parent": {
          "links": {
            "self": null
          }
        },
        "registrations": {
          "links": {
            "related": {
              "href": "http://localhost:8000/v2/nodes/23ehy/registrations/",
              "meta": {
                "count": 4
              }
            }
          }
        }
      },
      "links": {
[...]

PR #4622, opened 2015-11-16
Feature/api/logs list

brianjgeiger:
Replaces #4219 Issue https://openscience.atlassian.net/browse/OSF-3089

# Purpose

Add API endpoints for listing NodeLogs

# Changes

- Add /logs/ endpoint for listing NodeLogs
- Add /logs/{log_id}/nodes endpoint for listing Nodes related to a given NodeLog
- Add /nodes/{node_id}/logs for listing NodeLogs related to a given Node

# Side Effects

None

**Sample API Response**
```json
{
  "data": [
     {
            "id": "55df519d5ac80d15b5392234",
            "date": "2015-08-27T18:06:21.204000",
            "action": "retraction_approved",
            "params": {
                "node": "9yrst",
                "retraction_id": "55df517b5ac80d15b5392231"
            },
            "links": {
                "nodes": {
                    "related": "http://localhost:8000/v2/logs/55df519d5ac80d15b5392234/nodes/"
                },
                "user": {
                    "related": "http://localhost:8000/v2/users/j84ad/"
                }
            },
            "type": "logs"
        },
    ]
}
```

screen shot 2015-08-31 at 3 59 32 pm

```json
{
  "data": [
    {
      "id": "23ehy",
      "type": "nodes",
      "attributes": {
        "title": "test",
        "description": "",
        "category": "project",
        "date_created": "2015-08-27T18:04:25.097000",
        "date_modified": "2015-08-31T18:24:22.065000",
        "tags": [],
        "registration": false,
        "collection": false,
        "dashboard": false,
        "public": false
      },
      "relationships": {
        "children": {
          "links": {
            "related": {
              "href": "http://localhost:8000/v2/nodes/23ehy/children/",
              "meta": {
                "count": 1
              }
            }
          }
        },
        "contributors": {
          "links": {
            "related": {
              "href": "http://localhost:8000/v2/nodes/23ehy/contributors/",
              "meta": {
                "count": 1
              }
            }
          }
        },
        "files": {
          "links": {
            "related": "http://localhost:8000/v2/nodes/23ehy/files/"
          }
        },
        "node_links": {
          "links": {
            "related": {
              "href": "http://localhost:8000/v2/nodes/23ehy/node_links/",
              "meta": {
                "count": 0
              }
            }
          }
        },
        "parent": {
          "links": {
            "self": null
          }
        },
        "registrations": {
          "links": {
            "related": {
              "href": "http://localhost:8000/v2/nodes/23ehy/registrations/",
              "meta": {
                "count": 4
              }
            }
          }
        }
      },
      "links": {
        "self": "http://localhost:8000/v2/nodes/23ehy/",
        "html": "http://localhost:5000/23ehy/"
      }
    },
    {
      "id": "zdr43",
      "type": "nodes",
      "attributes": {
        "title": "Simple",
        "description": "",
        "category": "project",
        "date_created": "2015-08-27T18:04:25.097000",
        "date_modified": "2015-08-27T20:29:22.308000",
        "tags": [],
        "registration": true,
        "collection": false,
        "dashboard": false,
        "public": true
      },
      "relationships": {
        "children": {
          "links": {
            "related": {
              "href": "http://localhost:8000/v2/nodes/zdr43/children/",
              "meta": {
                "count": 1
              }
            }
          }
        },
        "contributors": {
          "links": {
            "related": {
              "href": "http://localhost:8000/v2/nodes/zdr43/contributors/",
              "meta": {
                "count": 1
              }
            }
          }
        },
        "files": {
          "links": {
            "related": "http://localhost:8000/v2/nodes/zdr43/files/"
          }
        },
        "node_links": {
          "links": {
            "related": {
              "href": "http://localhost:8000/v2/nodes/zdr43/node_links/",
              "meta": {
                "count": 0
              }
            }
          }
        },
        "parent": {
          "links": {
            "self": null
          }
        },
        "registrations": {
          "links": {
            "related": {
              "href": "http://localhost:8000/v2/nodes/zdr43/registrations/",
              "meta": {
                "count": 4
              }
            }
          }
        }
      },
      "links": {
        "self": "http://localhost:8000/v2/nodes/zdr43/",
        "html": "http://localhost:5000/zdr43/"
      }
    }
  ]
}
```

screen shot 2015-08-31 at 4 00 45 pm

```
{
    "data": [
        {
            "id": "55cb6ce25ac80daa6ef2a8c4",
            "type": "logs",
            "attributes": {
                "date": "2015-08-12T15:57:22.618000",
                "action": "project_created",
                "version": 1,
                "name": "nodelog"
            },
            "links": {
                "self": "http://localhost:8000/v2/nodes/a3t57/",
                "user": {
                    "self": null
                }
            }
        },
        {
            "id": "55cb6cf85ac80daa6ef2a8c8",
            "type": "logs",
            "attributes": {
                "date": "2015-08-12T15:57:44.400000",
                "action": "project_created",
                "version": 1,
                "name": "nodelog"
            },
            "links": {
                "self": "http://localhost:8000/v2/nodes/t6wav/",
                "user": {
                    "self": null
                }
            }
        }
]
}
```

image


website/util/__init__.py
```diff
@@ -180,3 +180,32 @@ def waterbutler_api_url_for(node_id, provider, path='/', **kwargs):
...
+    both collections and classes.
+    """
+    if not something:
+        return default
+    elif isinstance(something, collections.Mapping):
+        return something.get(key, default)
+    else:
+        return getattr(something, key, default)
+
+def deep_get(obj, key):
```
sloria: `DRF` already has a function for this, `get_attribute`.

website/project/model.py
```diff
@@ -397,10 +397,19 @@ class NodeLog(StoredObject):
     REGISTRATION_APPROVAL_INITIATED = 'registration_initiated'
     REGISTRATION_APPROVAL_APPROVED = 'registration_approved'
 
+    @classmethod
+    def actions(cls):
```
sloria: Minor: Is there any reason this shouldn't be a class variable? Also, better to use a set for constant-time membership lookup.

```python
class NodeLog(StoredObject):
    #...
    ALL_ACTIONS = set([
        CREATED_FROM, PROJECT_CREATED, ...
    ])
```

api/logs/views.py
```diff
@@ -0,0 +1,216 @@
...
+from framework.auth.oauth_scopes import CoreScopes
+
+from api.base.filters import ODMFilterMixin
+from api.base.utils import get_user_auth
+from api.base import permissions as base_permissions
+from api.logs.serializers import NodeLogSerializer
+from api.nodes.serializers import NodeSerializer
+from api.nodes.utils import get_visible_nodes_for_user
+
+class LogList(generics.ListAPIView, ODMFilterMixin):
```
sloria: I'm not sure if it makes sense to include this in the API at this point.

1. It's a very expensive endpoint. On production data, this endpoint took 7.5 seconds to return a response.
2. We don't have a use case for this.

brianjgeiger: If we don't need a global watch list, and if you are correct that we don't have a use case, I'm happy to remove that view.

api/base/filters.py
```diff
@@ -234,6 +245,13 @@ def __init__(self, *args, **kwargs):
         if not self.serializer_class:
             raise NotImplementedError()
 
+    def get_comparison_operator(self, key):
```
sloria: I don't think this method is used anywhere.

sloria:
I get a 500 when visiting `/v2/nodes/<private_node_id>/logs/`:

```
Environment:


Request Method: GET
Request URL: http://localhost:8000/v2/nodes/vrzmw/logs/

Django Version: 1.8
Python Version: 2.7.10
Installed Applications:
('django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.staticfiles',
 'rest_framework',
 'rest_framework_swagger',
 'corsheaders',
 'raven.contrib.django.raven_compat',
 'django_extensions')
Installed Middleware:
('api.base.middleware.DjangoGlobalMiddleware',
 'api.base.middleware.TokuTransactionsMiddleware',
 'corsheaders.middleware.CorsMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'django.middleware.security.SecurityMiddleware')


Traceback:
File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/django/core/handlers/base.py" in get_response
  132.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/django/views/decorators/csrf.py" in wrapped_view
  58.         return view_func(*args, **kwargs)
File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/django/views/generic/base.py" in view
  71.             return self.dispatch(request, *args, **kwargs)
File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/rest_framework/views.py" in dispatch
  452.             response = self.handle_exception(exc)
File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/rest_framework/views.py" in dispatch
  449.             response = handler(request, *args, **kwargs)
File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/rest_framework/generics.py" in get
  199.         return self.list(request, *args, **kwargs)
File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/rest_framework/mixins.py" in list
  39.         queryset = self.filter_queryset(self.get_queryset())
File "/Users/sloria/projects/cos-projects/osf/api/nodes/views.py" in get_queryset
  1622.         return self.get_queryset_from_request()
File "/Users/sloria/projects/cos-projects/osf/api/base/filters.py" in get_queryset_from_request
  319.         default_queryset = self.get_default_queryset()
File "/Users/sloria/projects/cos-projects/osf/api/nodes/views.py" in get_default_queryset
  1619.         return self.get_node().get_aggregate_logs_queryset(auth)
File "/Users/sloria/projects/cos-projects/osf/website/project/model.py" in get_aggregate_logs_queryset
  1627.                             if n.can_view(auth)]
File "/Users/sloria/projects/cos-projects/osf/website/project/model.py" in can_view
  944.             (auth.user and self.has_permission(auth.user, 'read')) or
File "/Users/sloria/projects/cos-projects/osf/website/project/model.py" in has_permission
  1054.         if permission in self.permissions.get(user._id, []):

Exception Type: AttributeError at /v2/nodes/vrzmw/logs/
Exception Value: 'AnonymousUser' object has no attribute '_id'
```

sloria:
Won't we need to serialize the the `NodeLog#params` and `user` for this feature to be useful?

sloria:
Doesn't necessarily need to be addressed before this is merged, but the `/v2/nodes/<nid>/logs/` endpoint is rather slow. For RPP (5024 logs), the response time is 4.6 seconds. Even a smaller project (310 logs) took 3.7 seconds.

I'm unsure where the bottleneck is; it's entirely possible that it isn't specific to this feature. =/

sloria:
@brianjgeiger Is there a ticket for this?

sloria:
Pass finished.

api/logs/views.py
brianjgeiger: So, I chatted with Sam. On the OSF right now, the JS that shows you a node's logs also shows you the logs for all of the children of that node that you have access to. That's what this endpoint is for.

website/project/model.py
brianjgeiger: Talked to Sam. Not really. It's meant to be immutable, and that may have been what he was thinking. He does not have strong feelings on the subject.

brianjgeiger:
If params are an unstructured dictionary, then we can't just serialize that information. So I will need to determine what all the possible entries are and serialize those. Hopefully @sloria is correct and it's been mostly standardized.

sloria:
@brianjgeiger Only node update logs fields are standardized; otherwise, each log type will have different params. Can't we use a `DictField` for unstructured data?

brianjgeiger:
Andy is against it in principle, and it goes somewhat against the purpose of the API having a predictable structure. It's why we're not just throwing in the extras field in the file endpoints.

api/logs/views.py
brianjgeiger: As I dig in more, I see what you're saying about using the get_aggregate_logs_queryset. That being said, a quick refactor will allow us to grab the query rather than the queryset, thus allowing us to filter with mongo rather than the list filter. So probably worth doing.

brianjgeiger:
Failing check is claiming base serializers is importing rapply but not using it, but it's being used all over the place.

api/base/serializers.py
```diff
@@ -282,7 +281,15 @@ def lookup_attribute(self, obj, lookup_field):
         bracket_check = _tpl(lookup_field)
         if bracket_check:
             source_attrs = bracket_check.split('.')
-            return get_nested_attributes(obj, source_attrs)
+            # If you are using a nested attribute for lookup, and you get the attribute wrong, you will not get an
+            # error message, you will just not see that field. This allows us to have slightly more dynamic use of
+            # nested attributes in relationship fields.
+            try:
+                return_val = get_nested_attributes(obj, source_attrs)
+            except KeyError:
```
sloria: Couldn't an `AttributeError` get raised here as well?

api_tests/nodes/views/test_node_logs.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def test_project_created(self):
+        res = self.app.get(self.public_url)
+        assert_equal(res.status_code, 200)
+        assert_equal(len(res.json['data']), len(self.public_project.logs))
+        assert_equal(self.public_project.logs[OSF_FIRST].action, "project_created")
+        assert_equal(self.public_project.logs[OSF_FIRST].action,res.json['data'][API_LATEST]['attributes']['action'])
+
+    def test_log_create_on_public_project(self):
+        datetime.datetime.strptime('2015-07-28 21:06:34.965114', '%Y-%m-%d %H:%M:%S.%f')
```
sloria: Remove this?

api/nodes/utils.py
```diff
@@ -0,0 +1,23 @@
+import itertools
+
+from modularodm import Q
+
+from website.util import permissions as osf_permissions
+from website.models import Node
+
+def get_visible_nodes_for_user(user, include_public=True):
```
sloria: This doesn't appear to be used anywhere. Can it be removed?

sloria:
The `/nodes/<nid>/logs/` endpoint has slow responses: loading just 50 logs with default pagination params takes 4.5-5 seconds to load.

I haven't looked into what the bottleneck might be. It doesn't necessarily need to be addressed before this is merged, but it's something we'll need to look into.

tests/test_utils.py
```diff
@@ -398,6 +398,27 @@ def test_get_recent_public_registrations(self):
         regs = [r for r in project_utils.recent_public_registrations(7)]
         assert_equal(len(regs), 7)
 
+class DummyAttrAttr(object):
```
sloria: Remove these?

api/base/serializers.py
brianjgeiger: Maybe, but not in the situation that I was having trouble with it. What are you thinking of?

sloria: I just checked the source for `get_nested_attributes` and saw that it could return an AttributeError. It seems like it should be handled.

brianjgeiger: Mmmmaybe. I think if we're passing an object in, it's going to have a predictable structure, so it should raise an error. But let's say that we're passing in a pointer node rather than a dictionary called pointer, so the case where `params.pointer.id` wouldn't return a key error. If there's no pointer in that case, the type would be none and it would return none at the top of the `get_nested_attributes`. This problem comes in when `params.pointer` exists but `params.pointer` equals `{}`. It's not none, but it doesn't have the key. But if it is an object rather than a dictionary, then the only reason the Attribute wouldn't exist is if you don't know what you're passing in, right? That seems like something that won't happen dynamically, and handling the error would make it harder to debug. 

I could be thinking about it wrong, but I think we should wait until we cause the error before we try to fix this one, unless you see the flaw in my thinking.

api/nodes/utils.py
brianjgeiger: Looks like it can.

api_tests/nodes/views/test_node_logs.py
brianjgeiger: Absolutely.

tests/test_utils.py
brianjgeiger: Sure. Sounds like a good idea.

api/base/serializers.py
sloria: Good point. I think this is fine as is.

brianjgeiger:
Definitely concerning about the load time. Was that on a multi-level project or a node with no children?

sloria:
Multi-level; it had 3 immediate children and 1 deeply nested child.

brianjgeiger:
My guess is it's related to that. We're *really* not supposed to be doing recursive things in the API, but this is a hard one to get around while maintaining functionality. But we should make an issue to investigate it if we don't fix now.

brianjgeiger:
My guess is we need to refactor NodeLogs so they're in their own collection.

PR #4624, opened 2015-11-17
[WIP]add download of osfoffline into osf

chennan47:
Add download of osfoffline into osf's project settings page.


chennan47:
Not needed for alpha version.

PR #4631, opened 2015-11-17
[Ready to Review]Feature/bulk skip permission failure nodes

pattisdr:
# Purpose 

Needed for https://openscience.atlassian.net/browse/OSF-5207

Currently, bulk functionality is all-or-nothing. If you try to update ten nodes and you don't have edit permissions for the tenth, the entire request will fail with a 403, and none of the nodes will be updated.

This PR permits you to proceed with the bulk request if you don't have permissions for all the nodes.  Only the nodes you have edit permissions for will be updated, and excluded validated_data for the nodes in which you didn't have edit permissions, will be returned under `/errors`.

# Changes

Modifies `bulk update nodes` and `bulk delete nodes` to skip nodes for which you don't have permissions for and modify the rest, if `skip_uneditable=True`
# Example

`PATCH /v2/nodes/?skip_uneditable=True`
I did not have edit access for node `39y2v`.  However, I did have access for `9tas3` and the changes to this node proceeded.

screen shot 2015-12-23 at 11 47 44 am

`DELETE /v2/nodes/?skip_uneditable=True`
I don't have access for `39y2v` but have it for `9tas3`.  `9tas3` is deleted, and a 200 Response is returned with `3qy2v` under errors.

screen shot 2015-12-23 at 11 53 47 am


lbanner:
@pattisdr Do we have a JIRA ticket on this one?

pattisdr:
@lbanner this was mostly exploratory but contributes to https://openscience.atlassian.net/browse/OSF-5055

The way I understand it, there is a lot of other v2 functionality needed to support this issue, but the changes here are in support of it.

api/base/generic_bulk_views.py
```diff
@@ -115,6 +125,18 @@ def bulk_destroy(self, request, *args, **kwargs):
...
+        if self.request.query_params.get('skip_uneditable', 'False') == 'True':
+            allowed = []
+            skipped = []
+            for resource in resource_object_list:
+                if resource.has_permission(user, ADMIN):
+                    allowed.append(resource)
+                else:
+                    skipped.append({'id': resource._id, 'type': object_type})
+            self.perform_bulk_destroy(allowed)
+            return Response(status=status.HTTP_200_OK, data={'meta': {'errors': skipped}})
```
mfraezz: This will be returning a `200 OK` rather than a `204 NO CONTENT` if `skip_uneditable==True` and `skipped==[]` (i.e. if a user has admin access to everything in the bulk request). Please refactor this method to return a `204` if all items are deleted.

api/base/serializers.py
```diff
@@ -609,25 +609,42 @@ class JSONAPIListSerializer(ser.ListSerializer):
...
+            ret.append({'errors': errors})
+
+        return ret
+
     # Overrides ListSerializer which doesn't support multiple update by default
     def update(self, instance, validated_data):
-        if len(instance) != len(validated_data):
-            raise exceptions.ValidationError({'non_field_errors': 'Could not find all objects to update.'})
+        bulk_skip_uneditable = self.context['request'].query_params.get('skip_uneditable', 'False')
+        if bulk_skip_uneditable == 'False':
```
mfraezz: It would be better to rewrite this with 
```python
from api.base.utils import is_truthy
[...]

if not is_truthy(bulk_skip_uneditable):
```
 here, as things like `?filter_uneditable=9001` would cause this `if` to evaluate to `False`. You want to limit the ways this `if` can be `True`

api/base/generic_bulk_views.py
```diff
@@ -115,6 +125,18 @@ def bulk_destroy(self, request, *args, **kwargs):
         if not self.allow_bulk_destroy_resources(user, resource_object_list):
             raise PermissionDenied
 
-        self.perform_bulk_destroy(resource_object_list)
+        # If skip_uneditable=True in query_params, skip resources for which the user does not have admin permissions
+        # and delete the remaining resources
+        if self.request.query_params.get('skip_uneditable', 'False') == 'True':
```
mfraezz: See other comment about using `is_truthy()`. There are a couple other of these instances in this PR, but I'm not going to flood your inbox with notifications for them.

api/nodes/views.py
```diff
@@ -293,6 +305,8 @@ def perform_create(self, serializer):
     # overrides BulkDestroyJSONAPIView
     def allow_bulk_destroy_resources(self, user, resource_list):
         """User must have admin permissions to delete nodes."""
+        if self.request.query_params.get('skip_uneditable', 'False') == 'True':
+            return True
```
mfraezz: This will return `True` if a user requests a set in which they have no write access for any object. Consider rewrite:
```python
if utils.is_truthy(self.request.query_params.get('skip_uneditable', 'False')):
    return any([node.has_permission(user, ADMIN) for node in resource_list])

File: api/base/generic_bulk_views.py (hunk 1 above)
Comment (pattisdr) [inline api/base/generic_bulk_views.py]: Nice - I didn't think think of that scenario.

Comment (mfraezz) [inline api/base/generic_bulk_views.py]: Please also add tests to confirm that the proper method is returned here, after the above change goes in.

File: api/nodes/views.py [hunk 5]
```diff
@@ -1,9 +1,9 @@
 import requests
 
 from modularodm import Q
+from rest_framework import status
 from rest_framework import generics, permissions as drf_permissions
 from rest_framework.exceptions import PermissionDenied, ValidationError, NotFound
-from rest_framework.status import is_server_error
```
mfraezz: Why was this removed in favor of import the whole `status` module? It's not of great importance, but would likely incur a small performance hit.

api/nodes/views.py
```diff
@@ -293,11 +305,30 @@ def perform_create(self, serializer):
...
+        """
+        allowed = []
+        skipped = []
+        for resource in resource_object_list:
+            if resource.has_permission(user, ADMIN):
+                allowed.append(resource)
+            else:
+                skipped.append({'id': resource._id, 'type': object_type})
+
+        if is_truthy(self.request.query_params.get('skip_uneditable', False)):
```
mfraezz: You should probably do this check before the costly resource permission calculation.
```python
if not is_truthy(self.request.query_params.get('skip_uneditable', False)):
    return None

# build skipped / allowed dict```

File: api/nodes/views.py [hunk 7]
```diff
@@ -293,11 +305,30 @@ def perform_create(self, serializer):
     # overrides BulkDestroyJSONAPIView
     def allow_bulk_destroy_resources(self, user, resource_list):
         """User must have admin permissions to delete nodes."""
+        if is_truthy(self.request.query_params.get('skip_uneditable', False)):
+            return any([node.has_permission(user, ADMIN) for node in resource_list])
         for node in resource_list:
```
mfraezz: Unimportant, but in keeping with the `bulk_boolean_fn([list_comprehension])` above, the rest of this method could be simplified to 
```python
return all([node.has_permission(user, ADMIN) for node in resource_list])
```

api_tests/nodes/views/test_node_list.py
```diff
@@ -1312,3 +1480,106 @@ def test_bulk_delete_invalid_payload_one_not_found(self):
...
+    def test_skip_uneditable_bulk_delete_query_param_required(self):
+        url = '/{}nodes/'.format(API_BASE)
+        res = self.app.delete_json_api(url, self.payload, auth=self.user_one.auth, expect_errors=True, bulk=True)
+        assert_equal(res.status_code, 403)
+
+        res = self.app.get('/{}nodes/'.format(API_BASE), auth=self.user_one.auth)
+        assert_equal(res.status_code, 200)
+        assert_equal(len(res.json['data']), 4)
+
+    def test_skip_uneditable_has_admin_permission_for_all_nodes(self):
```
mfraezz: :+1:

mfraezz:
Looking good @pattisdr  :moyai: 

A couple minor performance improvements and this should be `Ready to Merge`

sloria:
From the spec:

> Error objects MUST be returned as an array keyed by errors in the top level of a JSON API document.

So I think the errors should be placed in `/errors` rather than`/meta/errors`.

sloria:
^ source: http://jsonapi.org/format/#error-objects

sloria:
> A server MUST return a 200 OK status code if a deletion request is successful **and the server responds with only top-level meta data.**

http://jsonapi.org/format/#crud-deleting-responses-200

Since we may be returning errors in the response, we may be breaking spec here. However, given that the Bulk extension is experimental and the response format for partially-valid requests is underspecified, I think returning a `200` is fine for now, as long as we include `/errors` when appropriate.

sloria:
Pass finished. Nice work. Once https://github.com/CenterForOpenScience/osf.io/pull/4631#issuecomment-166929191 is addressed, this should be good to merge.

PR #4602, opened 2015-11-11
[WIP] Prereg Prize

samchrisinger:
This PR is mostly open to track changes made beyond what is included in feature/draft-reg

Some parts of this PR are isolated in https://github.com/CenterForOpenScience/osf.io/pull/4603

In particular it:
- Adds a generalized way for customizing login flows 
- Updates the RegistrationEditor to support commenting and importing contributor names

PR #4632, opened 2015-11-18
[WIP] Prereg Prize

samchrisinger:
# Post Draft-Reg additions
- Add generic custom login/registration flows with 'campaigns' 
- Add Prereg splash page: . It's my understanding that much of this code is borrowed from the onboarder, but this is almost all work done by Lyndsy. QA has fairly thoroughly tested this page, so unless their are major errors or blockers in the code we can give this a lighter touch in CR. 
  - views in `website.prereg` to support this
  - JS added:
    -  `...js/pages/prereg-landing-page'
    - `../js/components/autocomplete' for autocomplete widgets on the prereg landing page
- Add the 'Prereg Challenge' registration schema
- Updates the registration editor (QA has thoroughly tested this):
  - Bugfixes and improvements to file picker
  - Add a contributor importer
  - Many small improvements/bugfixes
- Add special `PreregCallbackMixin` class to handle sending a custom-branded email to Prereg applicants when their registration or embargo is made active
- Update `DraftRegistrationApproval` class to actually implement on_complete and on_reject callbacks-- This code is unreachable until the admin interface goes live
- Refactor `registrationEmbargo.js` into componentized `registrationModal.js` and add tests 
- Add scripts to auto-approve/reject draft registration approvals (testing only)

Replacing: https://github.com/CenterForOpenScience/osf.io/pull/4602


samchrisinger:
@felliott, replacing https://github.com/CenterForOpenScience/osf.io/pull/4602. I will still ping you when this PR is ready for review.

framework/auth/campaigns.py
```diff
@@ -0,0 +1,46 @@
+import httplib as http
```
samchrisinger: @abought, bearing in mind that this is used only for prereg, a closer look at the 'campaign' code would be great.

Mostly we want to be sure the default case (all existing users') experience is not disrupted.

website/project/model.py
```diff
@@ -3933,11 +3967,70 @@ class DraftRegistrationApproval(Sanction):
...
+            sanction = functools.partial(
+                registration.embargo_registration,
+                draft.initiator,
+                parse_date(self.meta.get('embargo_end_date'), ignoretz=True)
+            )
+        else:
+            raise ValueError("'registration_choice' must be either 'embargo' or 'immediate'")
+        try:
+            sanction(notify_initiator_on_complete=True)
+        except NodeStateError as e:
```
samchrisinger: @abought since the Prereg admin app is not yet deployed this code is unreachable.

framework/auth/campaigns.py
```diff
@@ -0,0 +1,46 @@
...
+    campaigns = [tag for tag in user.system_tags if tag in VALID_CAMPAIGNS]
+    if campaigns:
+        # TODO: This is a bit of a one-off to support the Prereg Challenge.
+        # We should think more about the campaigns architecture and in
+        # particular define the behavior if the user has more than one
+        # campagin tag in their system_tags.
+        return campaigns[0]
+
+def campaign_url_for(campaign):
+    # Defined inside this function to ensure a request context
```
abought: Although this references a request context, the actual URL in question is hardcoded. Should it be?

(as far as I can tell, this function is only used from within the flask application, which means that `web_url_for` should be available)

website/routes.py
```diff
@@ -227,6 +228,20 @@ def make_url_map(app):
         ),
 
         Rule('/news/', 'get', {}, OsfWebRenderer('public/pages/news.mako')),
+
+        Rule(
+            '/prereg/',
+            'get',
+            prereg.prereg_landing_page,
+            OsfWebRenderer('prereg_landing_page.mako')
```
abought: Please add the `trust=False` flag to mako routes, to ensure that markupsafe rendering is turned on.

framework/auth/core.py
```diff
@@ -445,13 +445,16 @@ def create(cls, username, password, fullname):
...
+    def create_unconfirmed(cls, username, password, fullname, do_confirm=True,
+                           campaign=None):
         """Create a new user who has begun registration but needs to verify
         their primary email address (username).
         """
         user = cls.create(username, password, fullname)
         user.add_unconfirmed_email(username)
         user.is_registered = False
+        if campaign:
+            user.system_tags.append(campaign)
```
abought: Is `system_tags` the best place to store campaign information? Since our mechanism of identifying campaign membership is based on just looking for a generic word in a catchall field, I'm worried that any other tag using the word "prereg" elsewhere later could accidentally trigger this functionality.

(it's a very common word- used 158 times in our codebase)

If we do continue using this field, here are two suggestions that could help:

1. In `VALID_CAMPAIGNS`, change `prereg` to `prereg_challenge_campaign` (make the special word more unique)
2. Building on the idea of `VALID_CAMPAIGNS`: Create a single central place where we can look up valid user system tags. Add a validator on the model field to ensure that only recognized tags can be used.

From recent experience with draft registrations, I can confirm that field validators help to avoid accidental mistakes or deletions of code, by helping other developers know what values can / can't be used in a field.

framework/auth/campaigns.py
```diff
@@ -0,0 +1,46 @@
...
+from framework.exceptions import HTTPError
+
+from website import mails
+
+VALID_CAMPAIGNS = (
+    'prereg',
+)
+
+EMAIL_TEMPLATE_MAP = {
+    'prereg': mails.CONFIRM_EMAIL_PREREG
```
abought: Given my suggestion of changing the name of the campaign identifier:

Have you considered merging `EMAIL_TEMPLATE_MAP` with `VALID_CAMPAIGNS` above? Specifying the campaign name key only once might help to keep the code maintainable.

eg 
```
VALID_CAMPAIGNS = {
    'prereg': {
        'email_template': mails.CONFIRM_EMAIL_PREREG,
        'redirect_url': 'http://www.huffingtonpost.com/2015/02/11/australias-oldest-man-sweaters-penguins-oil_n_6660962.html'
    }
}

File: framework/auth/campaigns.py [hunk 7]
```diff
@@ -0,0 +1,47 @@
...
+}
+
+def email_template_for_campaign(campaign, default=None):
+    if campaign in VALID_CAMPAIGNS:
+        try:
+            return EMAIL_TEMPLATE_MAP[campaign]
```
abought: Aside from other refactoring options mentioned, this could be shortened to
```
template = EMAIL_TEMPLATE_MAP.get(campaign, default)
if template is None:
    raise DeveloperError
return template
```

(Unless there's a reason why VALID_CAMPAIGNS and EMAIL_TEMPLATE_MAP would be decoupled; I just assume that there wouldn't be a template defined if the campaign wasn't valid)

framework/auth/campaigns.py
```diff
@@ -0,0 +1,47 @@
...
+    return default
+
+def campaign_for_user(user):
+    campaigns = [tag for tag in user.system_tags if tag in VALID_CAMPAIGNS]
+    if campaigns:
+        # TODO: This is a bit of a one-off to support the Prereg Challenge.
+        # We should think more about the campaigns architecture and in
+        # particular define the behavior if the user has more than one
+        # campagin tag in their system_tags.
+        return campaigns[0]
```
abought: I agree.

(also, line 33 typo; should be "campaign")

framework/auth/views.py
```diff
@@ -108,14 +108,16 @@ def forgot_password_get(auth, *args, **kwargs):
...
 ###############################################################################
 
-# TODO: Rewrite async
 @collect_auth
 def auth_login(auth, **kwargs):
     """If GET request, show login page. If POST, attempt to log user in if
     login form passsed; else send forgot password email.
 
     """
+    campaign = request.args.get('c')
```
abought: For readability (and to avoid collision with other options added in the future), consider changing the name of this URL argument from `c` to `campaign`.

ref: http://cosdev.readthedocs.org/en/latest/style_guides/python.html#general-naming-guidelines

framework/auth/views.py
```diff
@@ -130,17 +132,18 @@ def auth_login(auth, **kwargs):
...
 
-    code = http.OK
     if next_url:
         status.push_status_message(language.MUST_LOGIN)
-        # Don't raise error if user is being logged out
-        if not request.args.get('logout'):
-            code = http.UNAUTHORIZED
     # set login_url to form action, upon successful authentication specifically w/o logout=True,
-    # allows for next to be followed or a redirect to the dashboard.
+    # allows for next to be followed or a redirect to the dashboard.X
```
abought: Extraneous letter at end of line.

(new "most trivial comment ever")

scripts/approve_draft_registrations.py
```diff
@@ -0,0 +1,33 @@
...
+logging.basicConfig(level=logging.WARN)
+logging.disable(level=logging.INFO)
+
+
+def main(dry_run=True):
+    if dry_run:
+        logger.warn('DRY RUN mode')
+    pending_approval_drafts = DraftRegistration.find()
+    need_approval_drafts = [draft for draft in pending_approval_drafts
+                            if draft.requires_approval and draft.approval and draft.approval.state == Sanction.UNAPPROVED]
```
abought: It seems like this could be done more efficiently in a mongo query object,  rather than filtering the list of all results.

Eg (untested and guessing at default values)

`Q('requires_approval', 'eq', True) & Q('approval', 'ne', None)`

(not sure if approval.state can be directly queried this way; I'd need to dig through odm docs a bit more. References get weird in mongo)

scripts/approve_draft_registrations.py
```diff
@@ -0,0 +1,33 @@
+import sys
```
abought: Thanks for the notes on this PR, which helped me realize that this script (and certain others) were for testing/ local development use only.

Can you add a docstring to the top of this script, so that future developers can figure out what it is for? Ditto for other developer-only scripts in this PR.

(the scripts folder is getting very cluttered, and it's sometimes difficult to wade through and find a particular thing)

tests/factories.py
```diff
@@ -46,7 +46,8 @@
 from website.addons.wiki.model import NodeWikiPage
 from tests.base import fake
 
-from tests.base import DEFAULT_METASCHEMA
+from tests.base import get_default_metaschema
+DEFAULT_METASCHEMA = get_default_metaschema()
```
abought: This method is called at import time, possibly before the test database is set up. As discussed in person, that may not be the intended result.

tests/test_models.py
```diff
@@ -58,8 +58,7 @@
 from tests.test_features import requires_piwik
 from tests.utils import mock_archive
 
-
-from tests.base import DEFAULT_METASCHEMA
+DEFAULT_METASCHEMA = get_default_metaschema()
```
abought: This method is called at import time, possibly before the test database is set up. As discussed in person, that may not be the intended behavior.

tests/test_registrations/test_models.py
```diff
@@ -0,0 +1,187 @@
...
+    ProjectFactory, NodeLogFactory, WatchConfigFactory,
+    NodeWikiFactory, RegistrationFactory, UnregUserFactory,
+    ProjectWithAddonFactory, UnconfirmedUserFactory, CommentFactory, PrivateLinkFactory,
+    AuthUserFactory, DashboardFactory, FolderFactory,
+    NodeLicenseRecordFactory, DraftRegistrationFactory
+)
+from tests.test_registrations.base import RegistrationsTestBase
+
+
+class TestDraftRegistrations(RegistrationsTestBase):
```
abought: Thanks for keeping the tests organized!

framework/auth/views.py
```diff
@@ -130,17 +132,18 @@ def auth_login(auth, **kwargs):
     if status_message == 'expired':
         status.push_status_message('The private link you used is expired.')
 
-    code = http.OK
     if next_url:
         status.push_status_message(language.MUST_LOGIN)
-        # Don't raise error if user is being logged out
-        if not request.args.get('logout'):
```
abought: I admit it's a bit odd to call the login url with a `logout` parameter- but was this deletion intentional? (was this meant to cover some sort of edge case?)

framework/auth/views.py
```diff
@@ -174,6 +177,12 @@ def confirm_email_get(token, auth=None, **kwargs):
 
     if auth and auth.user and auth.user in (user, user.merged_by):
         if not is_merge:
+            # determine if the user registered through a campaign
```
abought: Does this imply that *every* login attempt will send the user to the campaign page now and forever?

Is this the intended behavior if, eg, the user continues to use the OSF after the campaign is officially over?

framework/auth/views.py
```diff
@@ -230,9 +239,16 @@ def send_confirm_email(user, email):
     except NoResultsFound:
         merge_target = None
 
+    campaign = campaigns.campaign_for_user(user)
+    # Choose the appropriate email template to use
+    if merge_target:
+        mail_template = mails.CONFIRM_MERGE
+    else:
```
abought: This should be `elif campaign:` (don't take for granted that `campaign` is truthy is `merge_target` isn't)

framework/auth/views.py
```diff
@@ -263,10 +280,15 @@ def register_user(**kwargs):
         full_name = request.json['fullName']
         full_name = strip_html(full_name)
 
+        campaign = json_data.get('campaign')
+        if campaign and campaign not in campaigns.VALID_CAMPAIGNS:
+            campaign = None
```
abought: +1 on validation for nonsense inputs. :octopus:

tests/test_models.py
```diff
@@ -4242,118 +4241,6 @@ def test_node_scale_with_deleted_parent(self):
         link.save()
         assert_equal(link.node_scale(node), -40)
 
-class TestDraftRegistration(OsfTestCase):
```
abought: Usually I check this for accidentally deleted code, but if you don't mind, I'm going to just make a blanket assumption that the entire test got correctly moved over in this case. :)

tests/test_registrations/base.py
```diff
@@ -0,0 +1,68 @@
...
+        super(RegistrationsTestBase, self).setUp()
+
+        self.user = AuthUserFactory()
+        self.auth = Auth(self.user)
+        self.node = ProjectFactory(creator=self.user)
+        self.non_admin = AuthUserFactory()
+        self.node.add_contributor(
+            self.non_admin,
+            permissions.DEFAULT_CONTRIBUTOR_PERMISSIONS,
+            auth=Auth(self.user),
```
abought: You can reuse the value of `self.auth` here.

tests/test_registrations/test_models.py
```diff
@@ -0,0 +1,187 @@
+# -*- coding: utf-8 -*-
+'''Unit tests for models and their factories.'''
```
abought: Very minor, but per https://www.python.org/dev/peps/pep-0257/ :
> For consistency, always use """triple double quotes""" around docstrings.

tests/test_registrations/test_models.py
```diff
@@ -0,0 +1,187 @@
...
+        user = AuthUserFactory()
+        draft = DraftRegistrationFactory(initiator=user)
+        assert_equal(draft.initiator, user)
+
+        node = ProjectFactory()
+        draft = DraftRegistrationFactory(branched_from=node)
+        assert_equal(draft.branched_from, node)
+        assert_equal(draft.initiator, node.creator)
+
+        schema = MetaSchema.find()[1]
```
abought: Is there any reason to expect that the content and number of metaschemas will be consistent? (eg, is `Metaschema.find()[1]` always going to get the schema you intend?)

tests/test_registrations/test_models.py
```diff
@@ -0,0 +1,187 @@
...
+            },
+            'c': {
+                'value': 2,
+            },
+        })
+        self.draft.save()
+        for key in ['foo', 'c']:
+            assert_in(key, changes)
+
+    @unittest.skip("TODO(samchrisinger): update this")
```
abought: TODO noted. If I understand this correctly, it applies to a feature that won't be usable until the admin tool is ready?

If that's the case, I'm ok with us just making a record, and fixing this in a followup PR.

tests/test_registrations/test_models.py
```diff
@@ -0,0 +1,187 @@
...
+        self.authorizer2 = AuthUserFactory()
+        for authorizer in [self.authorizer1, self.authorizer2]:
+            self.approval.add_authorizer(authorizer)
+        self.approval.save()
+
+    def test_on_complete(self):
+        pass
+
+    def test_send_rejection_email(self):
+        pass
```
abought: This isn't skipped- is this still a WIP section?

tests/test_registrations/test_views.py
```diff
@@ -0,0 +1,476 @@
...
+from website.util import permissions, api_url_for
+from website.project.views import drafts as draft_views
+
+from tests.factories import (
+    NodeFactory, AuthUserFactory, DraftRegistrationFactory, RegistrationFactory
+)
+from tests.test_registrations.base import RegistrationsTestBase
+
+from tests.base import get_default_metaschema
+DEFAULT_METASCHEMA = get_default_metaschema()
```
abought: See note above regarding `get_default_metaschema()` call. Hopefully it's not a problem; just a bit surprised it works.

tests/test_registrations/test_views.py
```diff
@@ -0,0 +1,476 @@
...
+
+class TestRegistrationViews(RegistrationsTestBase):
+
+    def test_node_register_page_not_registration_redirects(self):
+        url = self.node.web_url_for('node_register_page')
+        res = self.app.get(url, auth=self.user.auth)
+        assert_equal(res.status_code, http.FOUND)
+
+    @mock.patch('website.archiver.tasks.archive')
```
abought: Your patching of unnecessary tasks is appreciated!

But just in case some tests (like the above, L33) don't do what is expected: have you considered patching the archiver tasks there as well? (just in case they would otherwise have run, unintentionally)

tests/test_registrations/test_views.py
```diff
@@ -0,0 +1,476 @@
...
+        url = reg.web_url_for('node_register_page')
+        res = self.app.get(url, auth=self.user.auth)
+        assert_equal(res.status_code, http.OK)
+
+    def test_non_admin_can_view_node_register_page(self):
+        non_admin = AuthUserFactory()
+        self.node.add_contributor(
+            non_admin,
+            permissions.DEFAULT_CONTRIBUTOR_PERMISSIONS,
+            auth=Auth(self.user),
```
abought: From the base class, it seems you can just reuse the predefined `self.auth` here.

abought: Ditto for other places in this file- I'll make a single comment encouraging you to do a search, rather than spamming your inbox with redundant emails.

framework/auth/views.py
samchrisinger: Agreed. Given our time constraints, I'm going to leave this as a TODO for now

samchrisinger: Actually, I'm just gonna do this.

abought: Thanks- doing it now makes the most sense. We can change a lot of things after release, but we should try not to break anything user-facing (like URLs) once they're pushed out to the world.

scripts/approve_draft_registrations.py
samchrisinger: `requires_approval` is a property method

tests/test_registrations/test_views.py
```diff
@@ -0,0 +1,476 @@
...
+        assert_equal(data['status'], 'initiated')
+
+        self.draft.reload()
+        assert_is_not_none(self.draft.approval)
+        assert_equal(self.draft.approval.meta, {
+            u'registration_choice': unicode(self.embargo_payload['registrationChoice']),
+            u'embargo_end_date': unicode(self.embargo_payload['embargoEndDate'])
+        })
+
+    def test_draft_before_register_page(self):
```
abought: The test name makes it a bit unclear about what this test is supposed to do. A more descriptive name might be helpful.

Sample clarification questions: Are we just verifying the page exists? Or do we also (as part of the test) need to check that the page contains particular expected text (like node title)?

framework/auth/views.py
samchrisinger: `campaigns.email_template_for_campaign(campaign, default=mails.CONFIRM_EMAIL)` 
takes a default fallback template to handle falsy values of campaign.

tests/test_registrations/test_views.py
```diff
@@ -0,0 +1,476 @@
...
+        assert_false(registration.is_public)
+        assert_true(registration.is_pending_embargo)
+        assert_is_not_none(registration.embargo)
+
+    @mock.patch('framework.tasks.handlers.enqueue_task')
+    def test_register_draft_registration_invalid_embargo_end_date_raises_HTTPError(self, mock_enqueue):
+        res = self.app.post_json(
+            self.node.api_url_for('register_draft_registration', draft_id=self.draft._id),
+            self.invalid_embargo_date_payload,
+            content_type='application/json',
```
abought: No need to specify `content_type` when using the `post_json` method; it's done automatically for you.

framework/auth/views.py
samchrisinger: I inherited this code with the L142-43 removals already made... which meant these lines did nothing (`code` is unused). So I'm assuming the developer before me made good assumptions : ). 

My manual tests confirm that at least no unexpected exceptions are raised.

tests/test_registrations/test_views.py
```diff
@@ -0,0 +1,476 @@
...
+        res = self.app.delete(url, auth=non_admin.auth, expect_errors=True)
+        assert_equal(res.status_code, http.FORBIDDEN)
+        assert_equal(1, DraftRegistration.find().count())
+
+    def test_get_metaschemas(self):
+        url = api_url_for('get_metaschemas')
+        res = self.app.get(url).json
+        assert_equal(len(res['meta_schemas']), len(ACTIVE_META_SCHEMAS))
+
+        url = '/api/v1/project/drafts/schemas/?include=all'
```
abought: Other than the addition of `include=all`, this URL seems to be identical to the above. It's a good idea to avoid hard-coded URLs in tests, but doubly so if the hard-coded URL appears to be a duplicate request.

tests/test_registrations/test_views.py
samchrisinger: Yes, this is just a tests that the page exists and gives an 2XX response.

framework/auth/views.py
abought: That's a good point. A change is less pressing in that case.

Speaking generally, though, I'm still a bit leery of the generic else condition here, particularly since we're still evolving the campaigns feature. 

The behavior of that other function is the only thing keeping this view from doing awkward things, and we're not 100% committed to the design of the other function. By making this function responsible for checking its own assumptions, you can avoid coupling it to pieces whose behavior may depend later.

tests/test_views.py
```diff
@@ -64,7 +65,7 @@
 )
 from website.settings import ALL_MY_REGISTRATIONS_ID, ALL_MY_PROJECTS_ID
 
-from tests.base import DEFAULT_METASCHEMA
+DEFAULT_METASCHEMA = get_default_metaschema()
```
abought: Ok, you get the idea. 

If it's broken I'll stop asking about this line and just assume you'll do a search for all the usages. If it works as intended, great.

website/project/views/drafts.py
```diff
@@ -31,6 +32,49 @@
...
+    elif not node._is_embargo_date_valid(end_date):
+        max_end_date = today + settings.DRAFT_REGISTRATION_APPROVAL_PERIOD
+        raise HTTPError(http.BAD_REQUEST, data={
+            'message_short': 'Invalid embargo end date',
+            'message_long': 'Embargo end date must on or before {0}.'.format(max_end_date.isoformat())
+        })
+
+def check_draft_state(draft):
+    if draft.registered_node:
+        raise HTTPError(http.FORBIDDEN, data={
```
abought: Is 403 the best return code? This seems like a bad/nonsense request (400).

This isn't a permissions error, insofar as there's no combination of permissions under which the user should ever actually be allowed to access the resource in this case.

samchrisinger: @felliott convinced me to make this a 403. Maybe he can convince you too.

felliott: That's what 403 Forbidden is:  "You can't access this, don't even try".   400 would be "I don't understand what you want".   http://tools.ietf.org/html/rfc7231#section-6.5.3

felliott: 403 can be "you don't have enough permission", but also "you will never have enough permissions, no one does".

website/routes.py
```diff
@@ -227,6 +228,20 @@ def make_url_map(app):
...
+
+        Rule(
+            '/prereg/',
+            'get',
+            prereg.prereg_landing_page,
+            OsfWebRenderer('prereg_landing_page.mako', trust=False)
+        ),
+
+        Rule(
+            '/api/v1/prereg/draft_registrations/',
```
abought: Standard confirmation question: have these new OSF routes been approved by routing Tsars Jeff or BrianG?

website/project/views/drafts.py
abought: Talked it over a bit more.

Basically, it boils down to the following design question: do we want the user to know that the old draft data exists, after the draft is complete? Or do we want to pretend that a completed draft ceases to exist? (similar to how deletion / vanishing of other content types is handled)

If we want to acknowledge the data exists, 403 makes sense. If not, 404. And if this discussion seems pedantic, there's always 418.

samchrisinger: Pedantic implies you might have something to teach :grinning:.

We have discussed allowed DraftRegistrations to become editable again under certain circumstances. In particular, for Prereg, a DraftRegistration who's DraftRegistrationApproval is rejected becomes editable again. We may also extend this behavior to DraftRegistration's who's RegistrationApproval or Embargo gets rejected.

I'm inclined to leave this as a 403 for the time being.

website/static/js/components/autocomplete.js
```diff
@@ -0,0 +1,171 @@
...
+    },
+    onSubmit: function () {
+        // Redirect the user to continue the draft registration
+        window.location = this.itemSelected().url;
+    },
+    suggestionTemplate: function(item) {
+        var dateUpdated = new $osf.FormattableDate(item.value.dateUpdated);
+        var dateCreated = new $osf.FormattableDate(item.value.dateCreated);
+        return '<p>' + item.value.node.title + '</p>' +
+            '<p><small class="m-l-md text-muted">' + 'Initiated by: ' + item.value.initiator.name + '</small></p>' +
```
abought: In general, we very strongly discourage generating raw HTML in knockout- all display logic/ markup should be handled by the view layer. (in a mako file directly, or using knockout templates similarly to how `log_templates` are handled)

In particular, as written, this trusts the values passed in and renders them to the template as raw markup. If you absolutely must write markup in knockout, use $osf.htmlEscape` on that user-entered values (title, initiator.name, etc).

website/static/js/components/autocomplete.js
```diff
@@ -0,0 +1,171 @@
...
+ * that can be subclassed to create components to search through lists of
+ * objects other than nodes.
+**/
+'use strict';
+var $ = require('jquery');
+var ko = require('knockout');
+var $osf = require('js/osfHelpers');
+require('typeahead.js');
+
+var MAX_RESULTS = 14;
```
abought: As discussed in person, this seems like a very arbitrary limit, and can be revisited.

website/static/js/components/autocomplete.js
```diff
@@ -0,0 +1,171 @@
...
+            var matches = [];
+            var substrRegex = new RegExp(q, 'i');
+            var count = 0;
+
+            $.each(strs, function(i, str) {
+                if (substrRegex.test(str.node.title)) {
+                    count += 1;
+                    matches.push({ value: str });
+
+                    //alex's hack to limit number of results
```
abought: Is this hack necessary? Are other options (such as scrollbar technology) available?

website/static/js/osfHelpers.js
```diff
@@ -734,6 +734,23 @@ function indexOf(array, searchFn) {
     return -1;
 }
 
+/**
+ * Check if any of the values in an array are truthy
+ *
+ * @param {Array[Any]} listOfBools
+ * @returns {Boolean}
+ **/
+var any = function(listOfBools) {
```
abought: Depending on how general we want this function to be, underscore.js provides a possibly more generic implementation.
http://underscorejs.org/docs/underscore.html#section-31

website/static/js/pages/login-page.js
```diff
@@ -1,6 +1,25 @@
 /**
  * Login page
  */
-var LogInForm = require('js/signIn');
+var $ = require('jquery');
 
+var LogInForm = require('js/signIn');
 new LogInForm.SignIn('#logInForm');
+
+var SignUp = require('../signUp.js');
```
abought: Please move all `require` statements to the top of the module (before other statements, such as L7).

website/static/js/pages/login-page.js
```diff
@@ -1,6 +1,25 @@
...
+new SignUp('#signUpScope', '/api/v1/register/', $('#campaign').val());
```
abought: In general we discourage hard-coding URLs inside JS files. It limits both reusability and maintainability of the code.

website/static/js/pages/prereg-landing-page.js
```diff
@@ -0,0 +1,63 @@
...
+    $('#newProject, #newProjectXS').click( function() {
+        var title = $(this).parent().find('.new-project-title').val();
+        if (!title) {
+            return;
+        }
+        $osf.postJSON('/api/v1/project/new/', {
+            title: title,
+            campaign: 'prereg'
+        }).done(function(response) {
+            window.location = response.projectUrl + 'registrations/?c=prereg';
```
abought: This link (and possibly others in the diff) may have been broken by the change to "campaign" as the name of the URL parameter.

website/static/js/registrationEditorExtensions.js
```diff
@@ -131,19 +148,110 @@ ko.bindingHandlers.osfUploader = {
...
     self.filePicker = null;
 
+    self.preview = function() {
+        var value = question.value();
+        if (!value || value === 'no file selected') {
+            return 'no file selected';
+        }
+        else {
+            var extra = question.extra();
+            return $('<a target="_blank" href="' + extra.viewUrl + '">' + extra.selectedFileName + '</a>');
```
abought: Avoid writing HTML (with user-entered values) in KO. If doing so is the only way to avoid being pecked to death by parakeets, `$osf.htmlEscape` is a minimum precaution for `filename`, etc.

(For the URL it gets a bit trickier; you don't want to HTML escape ampersand url parameters, for example)

tests/test_registrations/test_views.py
samchrisinger: L33 will not touch any archiver code, but I will review this module for unpatched archiver tasks.

website/routes.py
samchrisinger: I'm actually not sure about this one. @brianjgeiger can you comment here?

website/static/js/pages/login-page.js
samchrisinger: Agreed, but this code appears shamelessly copy/pasted from https://github.com/CenterForOpenScience/osf.io/blob/6661fdc11aef164c6bce25a34ac9a83a0bb1a16c/website/static/js/pages/home-page.js. If you feel strongly about this I can change it.

website/routes.py
brianjgeiger: A long time ago, yeah, at least for the API route. And that should be more ephemeral than many, hopefully.

The `/prereg/` route is new. I think that ones fine, insofar as, since it's marketing related. The one thing that will need to be done is that we'll need to blacklist the term for our guid picker, since we will eventually have 7 character guids.

website/static/js/pages/login-page.js
abought: This is a "nice to have"- it won't break things for the release timeline, but it's not a good habit to get into, and "we copy-pasted it" isn't a defense. It's actually *less* reassuring than if it had only happened once.

samchrisinger: Are you willing to offer an alternative approach?

website/static/js/registrationUtils.js
```diff
@@ -15,51 +15,214 @@ var language = osfLanguage.registrations;
...
+     **/
+    self.isOwner = ko.pureComputed(function() {
+        return self.user.id === $osf.currentUser().id;
+    });
+
+    /**
+     * Returns the author as the actual user, not 'You'
+     **/
+    self.author = ko.pureComputed(function() {
+        return self.user.fullname;
```
abought: This (and probably others in this file)  does not need to be a `computed`.

Use a computed observable when the value depends on something that changes (another observable). This just depends on a static value, so there's no reason for knockout to keep watching it.

http://knockoutjs.com/documentation/computedObservables.html
> That’s where computed observables come in - these are functions that are dependent on one or more other observables, and will automatically update whenever any of these dependencies change.

website/static/js/registrationUtils.js
```diff
@@ -15,51 +15,214 @@ var language = osfLanguage.registrations;
...
+     **/
+    self.canEdit = ko.pureComputed(function() {
+        return !self.isDeleted() && self.saved() && self.isOwner();
+    });
+}
+/** Toggle the comment's save state **/
+Comment.prototype.toggleSaved = function(save) {
+    var self = this;
+
+    self.saved(!self.saved());
```
abought: If I'm reading this correctly, the comment is marked as saved *before* the operation has completed, and `save` appears to send an http request.

As a general rule, this flag should only toggle after successful completion of the operation.  If your save method returns a promise, this is an easy change- you could write:

```javascript
if (!self.saved()){
    result = save();
    result.done(self.saved(true));
};
```

website/static/js/registrationUtils.js
```diff
@@ -117,19 +297,50 @@ Question.prototype.toggleUploader = function() {
  * @property {ko.observableArray[Question]} questions
  * @property {String} title
  * @property {String} id
+ * @property {Question[]} questions
+ * @property {Question} current Question
  **/
-var Page = function(schemaPage, data) {
+var Page = function(schemaPage, schemaData) {
```
abought: Params not mentioned in docstring.

website/static/js/registrationUtils.js
```diff
@@ -365,8 +650,7 @@ Draft.prototype.register = function(data) {
  * @property {ko.observable[Question]} currentQuestion
  * @property {Object} extensions: mapping of extenstion names to their view models
  **/
-var RegistrationEditor = function(urls, editorId) {
-
+var RegistrationEditor = function(urls, editorId, preview) {
```
abought: Update docstring to reflect all parameters

website/static/js/registrationUtils.js
```diff
@@ -419,24 +713,82 @@ var RegistrationEditor = function(urls, editorId) {
...
+                    $elem.append(
+                        $.map(question.properties, function(subQuestion) {
+                            subQuestion = self.context(subQuestion);
+                            var value;
+                            if (self.extensions[subQuestion.type] ) {
+                                value = subQuestion.preview();
+                            } else {
+                                value = subQuestion.value();
+                            }
+                            return $('<p>').append(value);
```
abought: This appears to be a case of KO constructing raw HTML.

Am I correct in assuming that this text contains no user user-entered input of any kind, and sources its value entirely from the premade JSON prereg-schema files?

If this field can contain user input, use `$osf.htmlEscape` or similar.

website/static/js/registrationUtils.js
```diff
@@ -674,12 +1087,33 @@ RegistrationEditor.prototype.save = function() {
...
+/**
+ * Returns the `user_fullname` of each contributor attached to a node.
+ **/
+RegistrationEditor.prototype.getContributors = function() {
+    var self = this;
+    return self.makeContributorsRequest()
+        .then(function(data) {
+            return $.map(data.contributors, function(c) { return c.fullname; });
+        }).fail(function() {
+            $osf.growl('Could not retrieve contributors.', 'Please refresh the page or ' +
```
abought: Consider moving this text to `osfLanguage.js`, or reusing the predefined text snippets already available in that module.

website/static/js/registrationUtils.js
```diff
@@ -779,9 +1211,24 @@ RegistrationManager.prototype.init = function() {
         );
     });
 
-    $.when(getSchemas, getDraftRegistrations).done(function() {
+    var ready = $.when(getSchemas, getDraftRegistrations).done(function() {
         self.loading(false);
     });
+
+    var urlParams = $osf.urlParams();
+    if (urlParams.c && urlParams.c === 'prereg') {
```
abought: This may be broken by the recent renaming of the url parameter `c` to `campaign`.

website/templates/prereg_landing_page.mako
```diff
@@ -0,0 +1,141 @@
...
+</%def>
+
+<%def name="existingPrereg(size=None)">
+<% size = size or '' %>
+<div id="existingPrereg${size}" class="prereg-exisitng-prereg p-md osf-box box-round clearfix m-b-lg" style="display:none">
+  <p>Go to an existing preregistration:</p>
+  <form>
+    <osf-draft-registrations-search
+       params="placeholder: 'Type to search for a project',
+               data: '/api/v1/prereg/draft_registrations/',
```
abought: The `web_url_for` helper is directly available in the template context; there's no need to hard-code URLs here.

website/templates/project/edit_draft_registration.mako
```diff
@@ -40,29 +39,41 @@
...
-                    <!-- ko ifnot: onLastPage -->
-                      <a data-bind="click: nextPage" class="btn btn-primary pull-right">Next Page</a>
-                    <!-- /ko -->
               </div>
             </div>
           </div>
         </div>
       </div>
     </div>
+  </div>
```
abought: This closing tag may be unnecessary (does not appear to have a matching opening tag)

website/templates/project/register.mako
```diff
@@ -6,7 +6,7 @@
 <%def name="title()">Registration of ${node['title']}</%def>
   <div id="registrationMetaDataScope" class="container scripted">
     <div class="row">
-      <div class="span8 col-md-2 columns eight large-8">        
+      <div class="span8 col-md-2 columns eight large-8">
```
abought: Several of these styles do not appear to exist in bootstrap 3 (or elsewhere in our CSS), and should be removed.

website/templates/project/registration_utils.mako
```diff
@@ -0,0 +1,53 @@
...
+      <button type="submit" class="btn btn-default pull-right"
+              style="margin-right: 5px"
+              data-bind="click: cancel">Cancel</button>
+    </div>
+  </div>
+</script>
+
+<script type="text/html" id="preRegistrationTemplate">
+  <ul data-bind="foreach: preRegisterPrompts">
+    <li data-bind="html: $data"></li>
```
abought: Where does the value of `preRegisterPrompts` come from? Avoid using `data-bind:html` for any field that could ever conceivably be derived (even partially) from user-entered text.

website/templates/project/registrations.mako
```diff
@@ -81,27 +80,41 @@
...
+                      <i style="margin-right: 5px;" class="fa fa-pencil"></i>Preview
+                    </a>
                     <button class="btn btn-danger"
-                            data-bind="click: $root.deleteDraft"><i style="margin-right: 5px;" class="fa fa-times"></i>Delete</button>
+                            data-bind="click: $root.deleteDraft">
+                      <i style="margin-right: 5px;" class="fa fa-times"></i>Delete
+                    </button>
                   </div>
                   <div class="col-md-1">
+                    <!-- TODO(samchrisinger): pin down behavior here
```
abought: Is this TODO still relevant? I'm happy to be a sounding board if you need someone to discuss ideas with on this part.

abought:
First-readthrough complete. I'll pause while a few remaining comments are addressed; passing back to @samchrisinger .

I'll pick this back up once responses are done, and once all unit tests are passing.

website/templates/project/registration_utils.mako
samchrisinger: It's generated server side and contains no user input.

website/templates/project/registrations.mako
samchrisinger: I'm waiting on community response for this.

website/templates/project/register.mako
samchrisinger: It seems like you identified some specific classnames here that can be removed. It would save me time if you could indicated which ones explicitly.

abought: Ah, right- the original list of styles got buried on a different PR. It's hard to keep track of who saw what comment- sorry.

Basically, this templatate references a mix of bootstrap 2 and 3 styles (`span8` is deprecated`), plus some that appear to be derived from other libraries entirely and aren't used in the page (`eight`, `large8`, `columns`).

There may be some trial and error to clean up the templates here, but that should give a start.

website/static/js/registrationUtils.js
samchrisinger: Consider that this is likely only very marginally less efficient that a pure function (see http://knockoutjs.com/documentation/computed-dependency-tracking.html), and that it protects future developers from implementation changes down the line.

framework/auth/campaigns.py
```diff
@@ -0,0 +1,39 @@
...
+
+from werkzeug.datastructures import ImmutableDict
+from framework.exceptions import HTTPError
+
+from website import mails
+from website.util import web_url_for
+
+CAMPAIGNS = ImmutableDict({
+    'prereg': {
+        'system_tag': 'prereg_challenge_campaign',
```
abought: Thanks for making the system tag less ambiguous.

But building on that- do we really need a special `system_tag` key (and a function to convert old->new name) at all?

It seems that `campaign` is passed in from the front end, in which case this function could be removed if the front-end (and unit tests) simply passed in the correct name originally. I'm in favor of removing unnecessary aliasing code where possible; it reduces boilerplate and makes following the flow of the program easier.

See front-end code here:
https://github.com/CenterForOpenScience/osf.io/pull/4632/files#diff-96b2a4b9793e1418b2ff319410c1be9dR24

This may also affect other areas like URLs; not sure. Thoughts?

website/project/model.py
```diff
@@ -3933,11 +3967,61 @@ class DraftRegistrationApproval(Sanction):
...
+        draft = DraftRegistration.find_one(
+            Q('approval', 'eq', self)
+        )
+        auth = Auth(draft.initiator)
+        registration = draft.register(
+            auth=auth,
+            save=True
+        )
+        registration_choice = self.meta['registration_choice']
+        sanction = None
```
abought: Unused assignment (if value can't be set, then an exception is raised, so no default needed)

website/routes.py
abought: Thanks, @brianjgeiger . Out of curiosity: where does the GUID blacklist live? (who do we ping to make sure this gets done?)

brianjgeiger: I don't know. I *think* @samanehsan might have worked on that, but it was a long time ago and I never did anything with it directly. That could also have been a project that was only done to prototype stage, though I think it made it in.

website/static/js/components/autocomplete.js
```diff
@@ -0,0 +1,189 @@
...
+        // jQuery implicity escapes HTML
+        return $('<div>').append(
+            $('<p>', {
+                text: item.value.node.title
+            }).append(
+                [
+                    $('<p>').append(
+                        $('<small>', {
+                            className: 'm-l-md text-muted',
+                            text: 'Initiated by: ' + item.value.initiator.name
```
abought: Confirmed that this escapes user-entered values.

Thanks. :)

website/static/js/components/autocomplete.js
```diff
@@ -0,0 +1,189 @@
...
+ * ViewModel for a generic typeahead search component.
+ *
+ * Template: osf-search in components/autocomplete.js
+ *
+ * Params:
+ *  data: the URL from which to fetch the items to be searched
+ *  inputElement: the element upon which to call .typeahead(). Injected during
+ *      component registration.
+ */
+var baseSearchViewModel = function (params) {
```
abought: Use `UpperCamelCase` for class names.

https://github.com/felixge/node-style-guide#use-uppercamelcase-for-class-names

website/static/js/components/autocomplete.js
```diff
@@ -0,0 +1,189 @@
...
+        this.itemSelected(null);
+    },
+    /** Set the item as selected */
+    onSelected: function (item) {
+        this.itemSelected(item);
+    }
+});
+
+
+var draftRegistrationsSearchViewModel = function (params) {
```
abought: Use `UpperCamelCase` for class names.

https://github.com/felixge/node-style-guide#use-uppercamelcase-for-class-names

website/static/js/pages/login-page.js
abought: As discussed via flowdock, we normally use `window.contextVars` to pass variables (like URLs) to JS files. 

(This comment accidentally got lost on an "outdated" diff but probably still applies)

framework/auth/campaigns.py
samchrisinger: My motivation was to keep `campaign=prereg_challenge_campaign` out of URLs.

website/routes.py
samanehsan: @icereval can add it as a `BlacklistGuid` object so that it doesn't get used.

abought:
Functional/ style review:
The instructions portion of a prereg question should stand out more distinctly. Possibly remove the well and blockquote styles so that `data.description` text simple appears to be part of an unstyled white-background area at the top of the page.

screen shot 2015-11-30 at 3 01 33 pm

abought:
A question is required, even when not relevant.

I'm not sure if this can be changed, but it's a bit disconcerting to the user. Thoughts?

screen shot 2015-11-30 at 3 12 43 pm

abought:
The tab order goes `Last question box` --> `Save as draft`. I can't seem to tab to the "next page" button at all, which is awkward for users who interact with the form via keyboard controls.

(Eg, an accessibility concern)

abought:
I uploaded a file for an optional question, but did not click on it. 

It showed up as selected in the UI anyway (which is good), but the "preview for submission" page said no file was selected (which is confusing)

Also, there should be no period in the "preview for submission" question; `.:` looks odd. (see second screenshot)

screen shot 2015-11-30 at 3 21 56 pm

screen shot 2015-11-30 at 3 22 09 pm

abought:
The purpose of the comments box is unclear on this page- some explanatory text might be helpful.

Also, the comments only appear on the first page of the registration preview screen. It's not very intuitive where to find these. (I'm not sure whether to suggest that this be its own item on the sidebar nav, or what to recommend- depends on the use case)

screen shot 2015-11-30 at 3 33 32 pm

abought:
I went back and selected the optional analysis script. This representation is not a lot more helpful:

screen shot 2015-11-30 at 3 47 17 pm

Showing the filename would be preferred.

abought:
Light functional/style pass complete.

abought:
> I uploaded a file for an optional question, but did not click on it.
> 
> It showed up as selected in the UI anyway (which is good), but the "preview for submission" page said no file was selected (which is confusing)

I am still seeing this issue with the newest branch.

abought:
Style guide comments. I'm ok with resolving these in a followup (provided issue is tracked):

When deleting a draft from the registrations page,
- The button should be read
- The text should reference an action, not the word "ok"

https://centerforopenscience.github.io/osf-style/style.html#buttons

screen shot 2015-12-01 at 10 16 17 am

website/static/js/registrationUtils.js
```diff
@@ -791,31 +1242,36 @@ RegistrationManager.prototype.init = function() {
 RegistrationManager.prototype.deleteDraft = function(draft) {
     var self = this;
 
-    bootbox.confirm({
+    bootbox.dialog({
+        title: 'Please confim',
```
abought: Typo, should be "confirm"

abought:
Confirming fix for filepicker bug works for me. 

Tested on Firefox and Chrome, Mac OS.

PR #4635, opened 2015-11-18
[Ready for Review] Fix faulty error message logic in _fangornDropzoneError

TomBaxter:
# Purpose

Resolves: CenterForOpenScience/waterbutler#33 and perhaps others

# Changes

All DropZone errors passed to _fangornDropzoneError were getting "Unable to reach the provider..."
Fixed so that logic for alternate messages, including for 507, are not overwritten.

# Side Effects

None. However there were two different "default" messages. This might be a good opportunity to choose which one to use, or perhaps update it to something else.


website/static/js/fangorn.js
```diff
@@ -754,6 +754,10 @@ function _fangornDropzoneSuccess(treebeard, file, response) {
  */
 var DEFAULT_ERROR_MESSAGE = 'Could not upload file. The file may be invalid ' +
     'or the file folder has been deleted.';
+// Which DEFAULT_ERROR_MESSAGE is best?
+// var DEFAULT_ERROR_MESSAGE = 'Unable to reach the provider, please try ' +
```
sloria: Since this was the message that we were using before, let's go with this one as the default. You can remove the other one above.

sloria:
Logic looks correct. As mentioned above, just change the default message.

There's a lot of noise in the history for this PR. @TomBaxter would you mind sending a new PR branched of the latest `develop`, containing a commit with only your changes?

Thanks for your help with this!

TomBaxter:
Thanks @sloria . I created a new PR #4850 . However this file was updated quite a bit since the original PR and it looks like the logic was fixed in the mean time. The new PR just contains an update of the error message, no logic changes.

PR #4636, opened 2015-11-18
[WIP] NodeList is sorting by id and it thinks it's supposed to be sorting by date modified

pattisdr:
# Purpose

Issue https://openscience.atlassian.net/browse/OSF-5173

NodeList is sorting by id and it thinks it's supposed to be sorting by date modified.  This is because `date_modified` is a property on the Node model, not an actual field.  You can't filter on properties because they are python code and the filtering takes place at the database level.

~~You can get around it by turning the node queryset into a list, and then sorting that, but I think this is too slow.~~

# Changes

~~Sorts by title instead of date_modified.~~
Decision made to make date_modified a field on Node.  NodeLogs should update date_modified when something is changed.

### DIFF
--- a/api/nodes/views.py
+++ b/api/nodes/views.py
@@ -222,7 +222,7 @@ class NodeList(bulk_views.BulkUpdateJSONAPIView, bulk_views.BulkDestroyJSONAPIVi
 
     serializer_class = NodeSerializer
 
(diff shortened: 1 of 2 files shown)


pattisdr:
Date modified will be a field on the Node.

PR #4637, opened 2015-11-18
Admin Spam

Ghalko:
## Proposal
Comment Spam basic utility.

For reviewers, main directories modified:
- admin/spam
- admin/templates/spam


admin/spam/serializers.py
```diff
@@ -0,0 +1,54 @@
...
+
+def retrieve_comment(id, full_user=False):
+    query = Q("_id", 'eq', id)
+    comment = Comment.find_one(query)
+    return serialize_comment(comment, full=full_user)
+
+
+def serialize_comments():
+    query = Q('reports', 'ne', {})
+    return [
```
cwisecarver: Not sure which is clearer but you could also do:
```python
return map(serialize_comment, Comment.find(query)
```

admin/spam/views.py
```diff
@@ -1,9 +1,37 @@
-from django.http import HttpResponse
+from django.shortcuts import render, render_to_response
+from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
+
+from .serializers import serialize_comments, retrieve_comment
```
cwisecarver: Just like we talked with @leb2dg about there are DRF serializers for comments that you could import and use/subclass.

admin/templates/spam/comment_list.html
```diff
@@ -0,0 +1,39 @@
+{% extends "base.html" %}
```
cwisecarver: I don't think this file is being used by anything.

admin/templates/spam/email.html
```diff
@@ -0,0 +1,78 @@
...
+{% block content %}
+    {{ block.super }}
+
+    <div class="container-fluid">
+        <div class="row">
+            <div class="btn-group col-sm-12">
+                <a href="{% url 'spam:spam' %}" class="btn btn-primary">
+                    Back to list
+                </a>
+                <a href="{% url 'spam:detail' comment.id %}"
```
cwisecarver: It would be nice if this took you back to the page in the list you were previously.

admin/spam/views.py
```diff
@@ -1,9 +1,37 @@
...
 
 
 def spam_list(request):
```
cwisecarver: Not sure if you've investigated the class-based views in django but they might lend some benefit to some of these views that end up being similar. These aren't very long or complicated at this point but as you add more apps to admin I would say CBV would allow you to more easily share code. Probably a judgement call at this point but worth at least investigating.

admin/static/js/spam/spamList.js
```diff
@@ -0,0 +1,30 @@
+/** Facilitate actions in the list view of spam.
```
cwisecarver: I will admit to not being up on my knockout but I don't think this does anything. :)

admin/templates/spam/comment_list.html
Ghalko: Removed.

admin/templates/spam/email.html
Ghalko: @cwisecarver suggestions for how to do that? Any good examples?

cwisecarver:
```html
<a href="javascript:window.history.back();return true;">Back to list</a>
```
will work and it's not *too* awful. The only other thing I could think of is to pass the current page number to all the detail links in a query string and then build the back to list links using that. But the javascript way is quicker and less fragile. This was more a usability thing. I created 3500 comments and 4500 spam reports and then started poking around in the admin. It was inconvenient to go from page 35 to detail to page 1.

admin/static/js/spam/spamList.js
Ghalko: Removed. It seems I will be able to what I need in Django.

admin/templates/spam/email.html
Ghalko: Would you also suggest more navigation on the pagination itself? Right now I only have basic pagination nav, not so good for at least getting back to the first page. Also thanks for the snippet!

cwisecarver: I'll add a ```simple_tag``` that I wrote a few years ago for google-search-style pagination to the JIRA story. Github won't let me attach py or html files.

google search style: < prev 7 8 9 10 11 **12** 13 14 15 16 17 next>

admin/spam/serializers.py
Ghalko: Changed, definitely a simpler way to do this.

admin/templates/spam/email.html
Ghalko: I think we can leave that for a future improvement. But I changed the code for the original comment. Now sends the user back to the paginated page.

cwisecarver: Awesome, thanks.

admin/spam/serializers.py
```diff
@@ -0,0 +1,48 @@
...
+        'node': comment.node,
+    }
+    comment.update(
+        category=comment['reports'][0]['category'],
+    )
+    return comment
+
+
+def retrieve_comment(id, full_user=False):
+    query = Q("_id", 'eq', id)
```
cwisecarver: You don't need the intermediate variable here or on line 32. It appears to be common practice to just do:
```python
comment = Comment.find_one(Q('_id', 'eq', id)
```
Since you're doing a primary key lookup you could even just do:
```python
comment = Comment.load(id)
```
Not sure of the differences between the error states of those two methods if an item with that id isn't found.

cwisecarver:
@Ghalko Looks good. I think you'd talked about revisiting the CBV stuff in a future iteration, I think that's still worth it. This is a nice small app to get some experience with them. The last comment I made isn't a show stopper by any means. Just readability.

Ghalko:
@cwisecarver Thank you for the code review. I will simplify the pointed out issue for this PR and move it on to ready-to-merge unless there are reservations.

I think that CBV is definitely worthwhile, but I need to explore that in a different PR as I expect it will take a little while to understand the implications.

cwisecarver:
:+1: Good to go.

admin/spam/serializers.py
```diff
@@ -0,0 +1,45 @@
...
+    )
+    return comment
+
+
+def retrieve_comment(id, full_user=False):
+    return serialize_comment(Comment.load(id), full=full_user)
+
+
+def serialize_comments():
+    query = Q('reports', 'ne', {})
```
sloria: Is it possible for `reports` to be `None`?

Ghalko: I would guess not, but this is a question for someone with production database access.

PR #4644, opened 2015-11-20
Large number of files upload 

hmoco:
https://openscience.atlassian.net/browse/OSF-4829

## Purpose
Provide a report of how many files were successfully uploaded, and why others weren't .
For this first iteration, the user will receive a modal showing which files uploaded successfully and which didn't if the total files uploaded exceeds 3. The modal is shown below.

1. The behavior is the same for all providers and will show the name of the file as it was uploaded, rather than potential rename (such as file.txt (1)). 
2. Files that fail will have an 'X' next to their names and a little 'i' that should show what caused the file to fail on tooltip when the icon is hovered. 
3. Files that succeed will have a check mark next to their names, and the name should be a hyperlink (blue) to the file detail page.

QA, I was unable to test for case when 'Replace one, keep both' modal is used, so please pay attention to that potential use case.
screen shot 2015-11-23 at 10 20 59 am

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "raw-loader": "^0.5.1",
     "share": "0.7.27",
     "style-loader": "^0.8.3",
-    "treebeard": "git://github.com/caneruguz/treebeard.git#02e5b770640b5e7410cab9685ae7536fac04c3de",
+    "treebeard": "git://github.com/caneruguz/treebeard.git#c62685d66f62e7ed893eea32e77094b19a438885",
     "typeahead.js": "^0.10.5",
     "url-loader": "^0.5.5",
     "webpack": "^1.7.2",
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "raw-loader": "^0.5.1",
     "share": "0.7.27",
     "style-loader": "^0.8.3",
-    "treebeard": "git://github.com/caneruguz/treebeard.git#50a71133305496a440586ed880ae63f0ac5edda8",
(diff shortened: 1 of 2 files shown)


jinluyuan:
Could you change the "Done" button to "Got it"?

jinluyuan:
It's confusing that the "x" and check icons are merely symbols while the "i" icon has an action associated with it (hover and see information). It's also confusing that we use an "x" icon here when the same "x" icon is used throughout the site to indicate some kind of close or cancel action (to close a modal or cancel a file upload).

Could the "x" icon be changed to an exclamation point? Like [this](http://fortawesome.github.io/Font-Awesome/3.2.1/icon/exclamation-sign/)

Then make the check icon green and the exclamation point/x red. That helps the user quickly understand the meaning of the symbols and also differentiates the "i" icon.

@saradbowman Thoughts?

erinmayhood:
@jinluyuan For the UI: I agree with green for success and red for failure but the checkmark and X are pretty standard for these messages so let's keep it this way. Thanks.

felliott:
Hey @hmoco,

I'm sending this back to In Progress.  I'd like to make another pass after you've implemented the changes above.  When you're done, put it back in CR and assign me.

felliott:
@hmoco, this is looking awesome!  When the modal is displayed and the error message is the "Unable to reach the provider" message, the link html is being shown:

screen shot 2015-12-02 at 2 50 04 pm

Please scrub that out for the tooltip display.  Once that's fixed, I'm happy for this to go to RTM.

Cheers,
Fitz

felliott:
WFM. Sending this on!

sloria:
Looks good.

PR #4629, opened 2015-11-17
[Ready for Review] Make sure URI's are always JSON-API valid

pattisdr:
# Purpose

Resolves: https://openscience.atlassian.net/browse/OSF-4643

This replaces @samchrisinger's PR https://github.com/CenterForOpenScience/osf.io/pull/4356 and brings it up to date. Primarily, it reconciles the new RelationshipsField for serializing relationships with his changes.

# Changes

make sure 'self' and 'href' links are always URI strings
for relationships: /relationships//links/related/href is a string or /relationships/ == {}
add tests
If a related resource is null, the serialization of /data/relationships/<relation> is {}

# Side Effects

None

PR #4646, opened 2015-11-23
JSON API valid URI's fix 

pattisdr:
# Purpose

Solves issue https://openscience.atlassian.net/browse/OSF-5249

Corrects bug caused by https://github.com/CenterForOpenScience/osf.io/pull/4629

Above PR raises SkipField in multiple places when URL is none.  However, this causes comments relationships link to disappear.  When going to fetch `get_unread_comments`, if there are 0 comments on a node, URL is none and SkipField is raised.  The comments relationships link disappears.

# Changes

Small change to `def _url_val`. 

`if not url and url != 0:`  ==>  comment link is restored.

screen shot 2015-11-23 at 10 38 43 am

### DIFF
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -84,7 +84,7 @@ def _url_val(val, obj, serializer, **kwargs):
     else:
         url = val
 
-    if not url and url !=0:
+    if not url and url != 0:
         raise SkipField
     else:
         return url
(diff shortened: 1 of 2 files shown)


api/base/serializers.py
```diff
@@ -84,7 +84,7 @@ def _url_val(val, obj, serializer, **kwargs):
     else:
         url = val
 
-    if not url:
+    if not url and not url == 0:
```
sloria: This is fine, but `url != 0` is preferred for readability. Not a blocker for merge.

sloria:
@pattisdr Can you look into the failing build?

PR #4648, opened 2015-11-23
[Ready for Review] Refactor API v2 comment routes

samanehsan:
Purpose
------------
With the addition of commenting on files, users must be able to see all of the comments on their project (whether they are on the project overview page or individual files), and comment on either the project or a specific file. Instead of adding more routes, change the `create` endpoint to require users to specify the comment target and type. 

Changes
------------
Make the user pass in the comment's target_id and type when creating a comment instead of getting the target from the url. 

Side Effects
----------------
Since a user can specify the comment target, users can reply to comments through the node/comments endpoint. So the comment/replies endpoint is now a `ListView` instead of a `ListCreate` view.


api/comments/serializers.py
```diff
@@ -72,10 +60,56 @@ def update(self, comment, validated_data):
...
+        else:
             raise InvalidModelValueError('Invalid comment target.')
-        return object_type
+
+
+class CommentCreateSerializer(CommentSerializer):
+
+    target_type = ser.SerializerMethodField(method_name='check_target_type')
+
+    def check_target_type(self, obj):
```
sloria: This is a bit  of a misnomer. `check_target_type` implies that a boolean is being returned. How about `get_validated_target_type`?

felliott:
I can't create a deleted comment. The CommentCreateSerializer seems to ignore that attribute.

felliott:
Hey @samanehsan,

This is looking pretty good!  I'm sending this back to In Progress.  Once you've made the above changes (including Steve's suggestion), please reassign me and I'll take another look.  I promise I'll get to it faster this time!  Sorry for the delay!

Cheers,
Fitz

samanehsan:
> I can't create a deleted comment. The CommentCreateSerializer seems to ignore that attribute.

I think that's because the `deleted` field is `read_only`...but it still works for updating so that's weird. I'll remove the `read_only` requirement since it should be a writeable field. Should users be able to create deleted comments though? That seems like something we shouldn't allow.

felliott:
> I can't create a deleted comment. The CommentCreateSerializer seems to ignore that attribute.
> 
> I think that's because the deleted field is read_only...but it still works for updating so that's weird. I'll remove the read_only requirement since it should be a writeable field. Should users be able to create deleted comments though? That seems like something we shouldn't allow.
> 
Yeah, on second thought that does seem weird.  I guess `deleted` should be an update-only field.

samanehsan:
> I guess `deleted` should be an update-only field.

How can I make `deleted

felliott:
Hey @samanehsan,

Looks fantastic to me!  Once you've added the JSON-API error pointers, this is Ready-to-Merge.

Cheers,
Fitz

samanehsan:
Made those changes @felliott! The two failing tests have to do with HTTP connection errors...should I try restarting the build?

felliott:
> Made those changes @felliott <https://github.com/felliott>! The two failing tests have to do with HTTP connection errors...should I try restarting the build?
> 
> 

Perfect!  Thanks, @samanehsan!  Restarting the build isn’t necessary.  This is can go to Ready to Merge.

Cheers,
Fitz

api/comments/serializers.py
```diff
@@ -72,10 +60,62 @@ def update(self, comment, validated_data):
...
+class CommentCreateSerializer(CommentSerializer):
+
+    target_type = ser.SerializerMethodField(method_name='get_validated_target_type')
+
+    def get_validated_target_type(self, obj):
+        target = obj.target
+        target_type = self.context['request'].data.get('target_type')
+        expected_target_type = self.get_target_type(target)
+        if target_type != expected_target_type:
+            raise Conflict('Invalid target type. Expected \"{0}\", got \"{1}.\"'.format(expected_target_type, target_type))
```
sloria: Minor: The escaping here is unnecessary, but it's not a big deal.

api/comments/serializers.py
```diff
@@ -72,10 +60,62 @@ def update(self, comment, validated_data):
...
+
+    def create(self, validated_data):
+        user = validated_data['user']
+        auth = Auth(user)
+        node = validated_data['node']
+        target_id = self.context['request'].data.get('id')
+
+        try:
+            target = self.get_target(node._id, target_id)
+        except ValueError:
```
sloria: Good use of exceptions for flow control here :+1:

api/comments/serializers.py
```diff
@@ -72,10 +60,62 @@ def update(self, comment, validated_data):
...
+            raise InvalidModelValueError(
+                source={'pointer': '/data/relationships/target/data/id'},
+                detail='Invalid comment target \'{}\'.'.format(target_id)
+            )
+        validated_data['target'] = target
+        validated_data['content'] = validated_data.pop('get_content')
+        if node and node.can_comment(auth):
+            comment = Comment.create(auth=auth, **validated_data)
```
sloria: You might need to handle a ValidationError here, but I'm not sure.

sloria:
Looks good. Comments above are not blockers.

sloria:
I've restarted the build. This is good to merge when the build passes.

sloria:
Merged into `release/0.60.0` and `develop`.

PR #4477, opened 2015-10-19
[Bug Fix]Zotero/Mendeley revoked credentials do not show errors when used

GageGaskins:
Fixes [OSF-3964] and [OSF-2214].

Mendeley and Zotero now check if they have access to their API's and raise any exceptions found. Zotero checks just for not Authorized due to pyZotero having custom error handling. Addon settings panel now shows error similar to project dashboard error to indicate error to user.

PR #4649, opened 2015-11-24
[Bug Fix]Zotero/Mendeley fix rework

GageGaskins:
# Purpose
Redo Zotero/Mendeley API fix so that project settings nodes cannot be used if credentials are wrong. User must now to go User Settings and fix before they can be used again. Took out growl box from project page.

# Side Effects
Disables node use in project settings until error is fixed.

### DIFF
--- a/website/addons/citations/provider.py
+++ b/website/addons/citations/provider.py
@@ -20,9 +20,12 @@ class CitationsProvider(object):
     def check_credentials(self, node_addon):
         valid = True
         try:
-            node_addon.api.client()
-        except HTTPError:
-            valid = False
+            node_addon.api.client
+        except HTTPError as err:
+            if err.code == 403:
+                valid = False
+            else:
+                raise err
         return valid
 
     def user_accounts(self, user):
--- a/website/addons/citations/provider.py
+++ b/website/addons/citations/provider.py
@@ -17,6 +17,14 @@ class CitationsProvider(object):
     def serializer(self):
         pass
 
+    def check_credentials(self, node_addon):
+        valid = True
+        try:
+            node_addon.api.client()
+        except HTTPError:
+            valid = False
+        return valid
+
     def user_accounts(self, user):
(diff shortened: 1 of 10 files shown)


website/addons/citations/provider.py
```diff
@@ -17,6 +17,14 @@ def __init__(self, provider_name):
     def serializer(self):
         pass
 
+    def check_credentials(self, node_addon):
+        valid = True
+        try:
+            node_addon.api.client()
+        except HTTPError:
```
sloria: This error handling can be improved; only certain status codes should be allowed to pass here (probably 403).  Other status codes should still raise an error.

website/addons/citations/provider.py
```diff
@@ -17,6 +17,14 @@ def __init__(self, provider_name):
     def serializer(self):
         pass
 
+    def check_credentials(self, node_addon):
+        valid = True
+        try:
+            node_addon.api.client()
```
sloria: Is this correct? `client` appears to be a property of the `Mendeley` provider class, not a method.

website/addons/citations/provider.py
```diff
@@ -17,6 +17,14 @@ def __init__(self, provider_name):
     def serializer(self):
         pass
 
+    def check_credentials(self, node_addon):
```
sloria: Please write a unit test for this method.

sloria:
Pass finished.

sloria:
@GageGaskins Is this a fix for https://github.com/CenterForOpenScience/osf.io/pull/4477/files ?  If so, please resend this to `release/0.59.0`. Also be sure to remove any redundant code. For example, these lines: https://github.com/CenterForOpenScience/osf.io/blob/ae2b5d8bf49b2871a4f4b37344b0b3350e467e21/website/addons/mendeley/model.py#L76-L88 seem to do what you are doing `check_credentials`.

GageGaskins:
Correct branch/pull request at #4657

PR #4511, opened 2015-10-23
[ADM-16] [Feature] Admin webpack

leb2dg:
## Purpose
Add webpack settings specifically for the admin module. The `webpack.admin.config.js` can be run with by running `webpack --config webpack.admin.config.js` from the admin directory. Must run `npm install` within the admin directory when first created.

## Status
* Currently runs with no errors.
* Not sure that this is the ideal due to the large amount of files but it works

PR #4652, opened 2015-11-24
Feature/admin webpack

Ghalko:
Rebased on current admin-app based on PR #4511.

I can't get webpack to build the necessary `webpack-stats.json` from a clean project.


sloria:
Looks good. A few improvements can be made:

- Separate admin invoke tasks into its own module, e.g. `admin/tasks.py`. invoke has an API for namespacing: http://docs.pyinvoke.org/en/0.11.1/concepts/namespaces.html . This will allow us to do `inv admin.server`, `inv admin.assets`, etc. I may just do this when I merge.
- Remove dependency on bower. This can happen in a followup PR.

sloria:
Moved admin tasks to admin/tasks.py when I merged.

PR #4490, opened 2015-10-20
[WIP] Update Dropbox to ExternalAccounts

samchrisinger:
# Purpose

To support coming APIv2 addon management, we're updating all of our addon models to utilize the ExternalAccount/Provider architecture. 

# Changes

- Update Dropbox to use ExternalProvider
- Generalize Box *Settings model test cases into generic tests suites
   - implement test suites for Box and Dropbox

# Side Effects
TODO

TODO 
=====
After merge (of parent PR):
* Update GoogleDrive tests to use mixins
* Update Dataverse tests to use mixins
* Create mixins for citation tests
* (Consider) Create/use generic views for citation addons
* (Consider) Create/use generic views for non-standard oauth addons (Dataverse/S3)
* (Consider) Use `pytest.fixture` over mixins for tests.

PR #4654, opened 2015-11-29
[Feature] Generalize OAuth2 token refreshing logic

mfraezz:
Purpose
======
Only two addons currently support token refreshing as defined in the [OAuth2 RFC](http://tools.ietf.org/html/rfc6749#section-6): Box and GoogleDrive. However, with more addons in development, it would be an improvement to the architecture that new addons are to be built on if token refreshing could be done nearly-implicitly.

Changes
=======
* Generalized token refreshing code added to `ExternalProvider`, dead code removed
* Tests added/updated
* Scripts updated

Side Effects
=========
Conflict with #4490 / #4658 anticipated

Testing Notes (CR, not QA)
--------------------------------------
Manually set `force = True` in [this function](https://github.com/CenterForOpenScience/osf.io/pull/4654/files#diff-6391da5157423e52547c9e0c599e76d9R371) to make Box and GD refresh the token each time the node settings page is viewed.

[#OSF-4779]


felliott:
Hey @mfraezz,

This is looking pretty good!  I'm sending it back to In Progress.  Please reassign me after you've fixed mentioned issues.  Thank you!

Cheers,
Fitz

felliott:
In test_oauth.py, `test_refresh_oauth_key` needs an update.  The new computed `expires_at` is Dec. 31, 1969, I think because the mock response body is returning an interval for `expires_at`.  Please fix that and update the third assert to compare that the new expiry ts is greater than the old expiry ts.  Also please add another test that provides a `resp_expiry_fn`.

mfraezz:
I think the Zulu time issue is something with your local machine. It's the same evaluation function we've been using in prod, and I can't reproduce it locally. However, given that both affected addons return an `expires_in` key, I've set the default `resp_expiry_fn` to calculate it based on a delta of `time.time()` and that, avoiding any timezone-related issues.

Tests updated.

felliott:
WFM!  However, the tests need a few more adjustments.

1.)  The `expires_at` values provided to ExternalAccountFactory are floating seconds since epoch, but `expires_at` is expected to be a datetime.  Please cast these, so the tests match the expected data.

2.)  Please also update the third assert in `test_refresh_oauth_key` to compare that `external_account.expires_at > old_expiry`.  The new expiration should be farther in the future than the old date.

3.) Please add another test to exercise the _needs_refresh() logic.  The current test relies on `force=True` so the expiration calculation is never done.  You should be able to just copy-and-paste `test_refresh_oauth_key`, remove the `force=True`, and provide a `refresh_time`.

mfraezz:
Default refresh time for `MockOAuth2Provider` is 300. Test case for other branch added.

felliott:
@mfraezz, everything looks divine!  I'm sending this to RTM.

mfraezz:
:+1:

felliott:
Took a second look after @mfraezz worked out the merge conflicts.  Everything looks good to me.

PR #4656, opened 2015-11-30
Change confirm merge accounts modal text

kch8qx:
to be the same for all users, either associated or unassociated with another OSF account

[OSF-4883]

### DIFF
--- a/website/static/js/accountSettings.js
+++ b/website/static/js/accountSettings.js
@@ -208,7 +208,7 @@ var UserProfileViewModel = oop.extend(ChangeMessageMixin, {
                         this.emailInput('');
                         var addrText = $osf.htmlEscape(email.address());
                             bootbox.confirm({
-                                title: 'Merge Accounts?',
+                                title: 'Confirm Change?',
                                 message: 'Please check your email for confirmation of this change. ' + 
                                 'If there is another OSF account associated with ' + '<em>' + addrText + '</em>, ' +
                                 'you will have the ability to confirm an account merge.',
--- a/website/static/js/accountSettings.js
+++ b/website/static/js/accountSettings.js
@@ -201,30 +201,22 @@ var UserProfileViewModel = oop.extend(ChangeMessageMixin, {
 
             this.client.update(this.profile()).done(function (profile) {
                 this.profile(profile);
-
                 var emails = profile.emails();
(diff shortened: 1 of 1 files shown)


website/static/js/accountSettings.js
```diff
@@ -207,7 +207,22 @@ var UserProfileViewModel = oop.extend(ChangeMessageMixin, {
                     if (emails[i].address() === email.address()) {
                         this.emailInput('');
                         var addrText = $osf.htmlEscape(email.address());
-                        $osf.growl('<em>' + addrText  + '</em> added to your account.','You will receive a confirmation email at <em>' + addrText  + '</em>. Please check your email and confirm.', 'success');
+                            bootbox.confirm({
```
mfraezz: Currently, if a user adds an email but presses `Cancel`, the address is still added to their account and we send them an email. This method needs to be refactored if we want to allow a user to actually cancel after reading this but before the new address is added to their account. If not, you could use `bootbox.alert` here.

website/static/js/accountSettings.js
```diff
@@ -207,7 +207,22 @@ var UserProfileViewModel = oop.extend(ChangeMessageMixin, {
...
+                                title: 'Confirm Change?',
+                                message: 'Please check your email for confirmation of this change. ' + 
+                                'If there is another OSF account associated with ' + '<em>' + addrText + '</em>, ' +
+                                'you will have the ability to confirm an account merge.',
+                                callback: function (confirmed) {
```
mfraezz: Tests are failing because this function is defined within a loop. It's checking to make sure `email` was actually added to the users `emails` list, but there are other ways of doing that check.

mfraezz:
Pinging @erinmayhood for confirmation that this language is what we want for both the title and description.
screen shot 2015-12-07 at 17 11 57

Pending that, this looks good to merge :hatching_chick:

sloria:
@kch8qx Please resend this to `develop` if this PR is still relevant.

kch8qx:
@sloria I wasn't sure how to resend this PR, but I created a new PR with this feature to develop 

https://github.com/CenterForOpenScience/osf.io/pull/4741

sloria:
@kch8qx That's the correct way to do it =)

PR #4657, opened 2015-11-30
[Bug Fix] Zotero/Mendeley credentials error redux

GageGaskins:
# Purpose
Old fix had a bug during importing from account in project settings page. This fix checks for valid credentials upon creating of node settings and tells user to fix in settings if something is wrong while disabling importing/deactivating account from project settings page.

### DIFF
--- a/website/addons/citations/provider.py
+++ b/website/addons/citations/provider.py
@@ -20,9 +20,12 @@ class CitationsProvider(object):
     def check_credentials(self, node_addon):
         valid = True
         try:
-            node_addon.api.client()
-        except HTTPError:
-            valid = False
+            node_addon.api.client
+        except HTTPError as err:
+            if err.code == 403:
+                valid = False
+            else:
+                raise err
         return valid
 
     def user_accounts(self, user):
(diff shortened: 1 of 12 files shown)


website/addons/citations/provider.py
```diff
@@ -17,6 +17,17 @@ def __init__(self, provider_name):
     def serializer(self):
         pass
 
+    def check_credentials(self, node_addon):
+        valid = True
+        try:
+            node_addon.api.client
```
sloria: I think I commented on this in the other PR, but this appears incorrect. Why would this raise an `HTTPError`?

sloria: Ah, I see now; accessing client for the first time will authenticate with Mendeley. Disregard the previous comment.

sloria: However, client is a cached property, i.e. it will try to connect to Mendeley the first time and will error if it can't connect. The second time `client` is accessed, a request will *not* be sent to Mendeley, and an error will not be raised.

This is a bug: the first call to `check_credentials` with invalid creds will return `False`. The second call, however, will always return `True`. It's probably best to instantiate the Mendeley client here rather than rely on the `client` property.

sloria:
Pass finished. Once https://github.com/CenterForOpenScience/osf.io/pull/4657#discussion_r46328617 is addressed, this should be good to merge.

PR #4659, opened 2015-11-30
[Ready to Review] Restore self link using get_absolute_url method.

pattisdr:
# Purpose
Issue https://openscience.atlassian.net/browse/OSF-5284
The self link had been removed in NodeContributorSerializer, so it used the default, obj.get_absolute_url() to formulate the self link. When embedding, this caused the embedded user and the contributor self links to be the same.

# Changes
Restoring `'self': 'get_absolute_url'` to node contributors links will allow the self-link to be generated properly.


brianjgeiger:
Yup, this looks good. :tada: :penguin: :tada:

PR #4660, opened 2015-11-30
[Ready to Review] Feature/sort nodes by date modified

pattisdr:
# Purpose 

Issue: https://openscience.atlassian.net/browse/OSF-5173
NodeList is sorting by id and it thinks it's supposed to be sorting by date modified

# Changes

Makes date_modified a field on the node.  When there is a change to the node log, date_modified is updated.  Since date_modified is now a field on the node rather than a property, node list can be sorted by date_modified.

# Side Effects
- Serialized node list sorted by date_modified rather than id
- Field added on model

### DIFF
--- a/scripts/migration/migrate_date_modified_for_existing_nodes.py
+++ b/scripts/migration/migrate_date_modified_for_existing_nodes.py
@@ -9,6 +9,15 @@ from website.app import init_app
 
 logger = logging.getLogger(__name__)
 
+def date_updated(self):
+        """
+        The most recent datetime when this node was modified, based on
+        the logs.
+        """
+        try:
+            return self.logs[-1].date
+        except IndexError:
+            return self.date_created
 
 def main():
     init_app(routes=False)
@@ -19,8 +28,10 @@ def main():
     for node in models.Node.find():
         logger.info('Node {0} "date_modified" added'.format(node._id))
         if not dry_run:
-            node.date_modified = node.date_updated()
+            node.date_modified = date_updated(node)
             node.save()
 
 if __name__ == '__main__':
     main()
+
+
--- a/scripts/migration/migrate_date_modified_for_existing_nodes.py
+++ b/scripts/migration/migrate_date_modified_for_existing_nodes.py
@@ -1,5 +1,5 @@
 """
-This will add a date_modified field to all nodes.  Date_modified will be set equal to the date of the last log.
+This will add a date_modified field to all nodes.  Date_modified will be equivalent to the date of the last log.
 """
 import sys
 import logging
--- a/scripts/migration/migrate_date_modified_for_existing_nodes.py
+++ b/scripts/migration/migrate_date_modified_for_existing_nodes.py
@@ -3,35 +3,45 @@ This will add a date_modified field to all nodes.  Date_modified will be equival
 """
 import sys
 import logging
-
-from website import models
 from website.app import init_app
+from website import models
+from scripts import utils as script_utils
+from framework.transactions.context import TokuTransaction
 
 logger = logging.getLogger(__name__)
 
-def date_updated(self):
-        """
-        The most recent datetime when this node was modified, based on
-        the logs.
-        """
-        try:
-            return self.logs[-1].date
-        except IndexError:
-            return self.date_created
+def date_updated(node):
+    """
+    The most recent datetime when this node was modified, based on
+    the logs.
+    """
+    try:
+        return node.logs[-1].date
+    except IndexError:
+        return node.date_created
 
 def main():
     init_app(routes=False)
-    dry_run = 'dry' in sys.argv
+
     logger.warn('Date_modified field will be added to all nodes.')
-    if dry_run:
-        logger.warn('Dry_run mode')
+
     for node in models.Node.find():
+        node.date_modified = date_updated(node)
+        node.save()
         logger.info('Node {0} "date_modified" added'.format(node._id))
-        if not dry_run:
-            node.date_modified = date_updated(node)
-            node.save()
+
 
 if __name__ == '__main__':
-    main()
+
+    dry_run = 'dry' in sys.argv
+    if not dry_run:
+        script_utils.add_file_logger(logger, __file__)
+
+    with TokuTransaction():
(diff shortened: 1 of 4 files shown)


website/project/model.py
```diff
@@ -2158,8 +2159,13 @@ def add_log(self, action, params, auth, foreign_user=None, log_date=None, save=T
             foreign_user=foreign_user,
             params=params,
         )
+
         if log_date:
             log.date = log_date
+
+        self.date_modified = log.date.replace(tzinfo=None)
```
mfraezz: Is there any reason to do this here, rather than in `.save`? No specific examples come to mind (maybe user merging), but it's possible that there's an action that changes and saves a node without adding a log.

website/project/model.py
```diff
@@ -1776,11 +1777,11 @@ def get_recent_logs(self, n=10):
         """
         return list(reversed(self.logs)[:n])
 
-    @property
-    def date_modified(self):
-        '''The most recent datetime when this node was modified, based on
+    def date_updated(self):
```
mfraezz: This method would be unused after migration, please remove it and add equivalent logic to the migration.

mfraezz:

Tests look good, and the migration worked with staging data without any issues:
```python
In [1]: all([(e.date_modified == e.logs[-1].date) for e in Node.find()])
Out[1]: True
```
After one small change and a question, this should be good to merge. :whale:

website/project/model.py
pattisdr: When you add a component to a node, a log isn't added to the node.  

If I create a project `abcde` and add component `12345`, both nodes have one log each, with action 'project_created' even though `12345` is added to node.nodes.

mfraezz: That isn't really affecting the node in a way that should change its modified date, so I guess it makes the most sense to do that update here.

However, considering this functions `save` kwarg defaults to `True`, should it be calling `.save()` outside of the `if save:` block?

pattisdr: I guess I wanted to make sure that date_modified was definitely set, but I guess the if save: block is saving the new log list on the node only then should date_modified be updated. 

Removed .save() that was outside this block.

pattisdr: Removing the .save() causes some tests to fail - looking into it.

pattisdr: Made a change to `def update_node` in `/website/project/views/node.py`. Pulled date_modified from the updated_fields_dict was being returned.     There were two tests in test_views that were failing before this change.

I'm not sure that update_node is even being used?  When you change the title, it is edit_node that is called, not update_node.

mfraezz: Consulted with @brianjgeiger to confirm that it makes sense to not include it in the dict. The `date_modified` field is implicitly updated, so, like `logs`, it doesn't need to be serialized.

mfraezz:
LGTM. Passing it off to Steve. :water_buffalo: 

One minor note: as per [`/scripts/README.md/`](https://github.com/CenterForOpenScience/osf.io/blob/develop/scripts/README.md), please use a `TokuTransaction` for future migrations. This simplifies logic for `dry` runs and adds the benefit of rolling back the migration if any errors are raised, which prevents possibly invalid database states.

scripts/migration/migrate_date_modified_for_existing_nodes.py
```diff
@@ -0,0 +1,37 @@
...
+import sys
+import logging
+
+from website import models
+from website.app import init_app
+
+logger = logging.getLogger(__name__)
+
+def date_updated(self):
+        """
```
sloria: This block appears to be over-indented (should be 4 spaces), but Python doesn't seem to mind..?

scripts/migration/migrate_date_modified_for_existing_nodes.py
```diff
@@ -0,0 +1,37 @@
...
+"""
+import sys
+import logging
+
+from website import models
+from website.app import init_app
+
+logger = logging.getLogger(__name__)
+
+def date_updated(self):
```
sloria: `self` is a bit of a strange param name (this isn't a method). How about `node`?

scripts/migration/migrate_date_modified_for_existing_nodes.py
```diff
@@ -0,0 +1,37 @@
...
+        """
+        try:
+            return self.logs[-1].date
+        except IndexError:
+            return self.date_created
+
+def main():
+    init_app(routes=False)
+    dry_run = 'dry' in sys.argv
+    logger.warn('Date_modified field will be added to all nodes.')
```
sloria: As per COSDev, add a file logger to the logger when not in dry mode.

scripts/migration/migrate_date_modified_for_existing_nodes.py
```diff
@@ -0,0 +1,37 @@
...
+    if dry_run:
+        logger.warn('Dry_run mode')
+    for node in models.Node.find():
+        logger.info('Node {0} "date_modified" added'.format(node._id))
```
sloria: Might make sense to put this logger statement after the node is saved so that it is only emitted after a successful migration of a single node.

sloria:
Pass finished. Once the changes suggested above are made to the migration, this should be good to merge.

scripts/migration/migrate_date_modified_for_existing_nodes.py
```diff
@@ -0,0 +1,46 @@
...
+        return node.logs[-1].date
+    except IndexError:
+        return node.date_created
+
+def main():
+    init_app(routes=False)
+
+    logger.warn('Date_modified field will be added to all nodes.')
+
+    for node in models.Node.find():
```
sloria: Hm, I think we can make this resume-able by adding `Q('date_modified', 'eq', None)`. @pattisdr Can you please make this change in a follow-up PR?

pattisdr: Yes.

pattisdr: https://github.com/CenterForOpenScience/osf.io/pull/4798

PR #4540, opened 2015-10-30
[Ready to Review] Feature/restore registrations to nodes list, limit retractions, add fields to reg serializer

pattisdr:
# Purpose

Issue https://openscience.atlassian.net/browse/OSF-5036

Registrations are currently excluded from the nodes endpoint but we have decided we will make them available. Retractions currently show all fields, when many should be hidden.  Registrations serializer is not complete.

# Changes
Makes numerous changes to registration serializer among others:
- restores registrations to /v2/nodes/
- removes retractions from /v2/nodes/
- restores filtering on `registration` to /v2/nodes/
- creates /v2/users/me/registrations/
- only serializes limited fields when returning a retraction with /v2/registrations/, others are None
- changes registration relationship URL's from node endpoints to registration endpoints
- throws 403 if you try to access retraction detail information through /v2/nodes/<retraction_id>
- throws 403 if you try to access retraction relationship information aside from contributors
- Adds additional registration serializer fields
Added Fields:
 -  pending_retraction
 -  pending_registration_approval
 - pending_embargo
 - registered_meta - contains name of registration schema, embargo end date, and answers to supplemental registration questions
 - registration_justification
 - registration supplement

**Retraction response**
screen shot 2015-11-05 at 6 42 20 pm

**New fields for registration**
screen shot 2015-11-20 at 12 17 54 pm

PR #4661, opened 2015-11-30
Institutions API

hmoco:
Preliminary institutions work. Includes model and API work. Adds several endpoints:
 - ` /institutions/ `, with list of all institutions. (methods: ` GET `)
 - ` /institutions/<id> ` detail of institutions. (methods: ` GET `)
 - ` /institutions/<id>/nodes` and ` /institutions/<id>/users ` an endpoint for its nodes and users. (methods: ` GET `)
 - ` /nodes/<nid>/institution ` a node's primary institution, if any. (methods: ` GET `)
 - `/nodes/<nid>/relationships/institution `, endpoint to add an institution to a node. (methods: ` GET `, ` PUT `, ` PATCH `)
 - `/users/<uid>/institutions ` all of a user's affiliated institutions. (methods: ` GET `)

Also adds relationship serializer and parser for the api. A relationship request would look like:
` {"data": {"type": <types>, "id": <id> } } ` or ` {"data": null} `for ` PUT, PATCH ` update. For institutions, docs are: 

### Create
```
Method:        PUT || PATCH
URL:           /links/self
Query Params:  <none>
Body (JSON):   {
                 "data": {
                   "type": "institutions",   # required
                   "id": <institution_id>   # required
                 }
               }
Success:       200
```
This requires both admin permission on the node, and for the user that is making the request to be affiliated with the institution (User.affiliated_institutions)

### Destroy

```
Method:        PUT || PATCH
URL:           /links/self
Query Params:  <none>
Body (JSON):   {
                 "data": {
                   null
                 }
               }
Success:       204
```
This requires admin permission on the node.


api/institutions/serializers.py
```diff
@@ -0,0 +1,34 @@
...
+
+    nodes = RelationshipField(
+        related_view='institutions:institution-nodes',
+        related_view_kwargs={'institution_id': '<pk>'},
+    )
+
+    users = RelationshipField(
+        related_view='institutions:institution-users',
+        related_view_kwargs={'institution_id': '<pk>'}
+    )
```
pattisdr: You need `self` and `html` links for Institutions.  You can use `LinksField` for this.

framework/auth/core.py
```diff
@@ -1313,8 +1314,32 @@ def n_projects_in_common(self, other_user):
...
+
+    @property
+    def pk(self):
+        return self._id
+
+    def auth(self, user):
+        return user.has_inst_auth(self)
+
+    def view(self):
+        return 'Static paths for custom pages'
```
pattisdr: Institutions will need a `url` property and an `absolute_url` property.

api/institutions/serializers.py
pattisdr: Something like:

screen shot 2015-12-04 at 1 28 54 pm

api/nodes/views.py
```diff
@@ -1726,3 +1727,31 @@ def perform_create(self, serializer):
...
+    serializer_class = InstitutionDetailSerializer
+
+    model = Institution
+    view_category = 'nodes'
+    view_name = 'node-institution-detail'
+
+    def get_object(self):
+        node = self.get_node()
+        if self.request.method in drf_permissions.SAFE_METHODS:
+            return node.primary_institution
```
pattisdr: If there is no primary_institution associated with the Node, the response will be incorrect.  I would raise NotFound.

api/nodes/views.py
```diff
@@ -1726,3 +1727,31 @@ def perform_create(self, serializer):
...
+    )
+
+    required_read_scopes = [CoreScopes.NODE_BASE_READ, CoreScopes.INSTITUTION_READ]
+    required_write_scopes = [CoreScopes.NODE_BASE_WRITE]
+    serializer_class = InstitutionDetailSerializer
+
+    model = Institution
+    view_category = 'nodes'
+    view_name = 'node-institution-detail'
+
```
pattisdr: If the intent is to mirror node_links for adding an institution (and relationships will be in the request body), you will need to override `def get_parser_context` on this view to add `is_relationship` is True to the parser context.
 
See NodeLinksList view for reference.

pattisdr:
The parser in api/base/parsers would probably need some changes.  Currently, the parser expects POST requests when creating relationships, like node links and node contributors.  Since the POST part won't won't be true, it will assume you aren't creating a relationship and thus expect `attributes` to be in the request body and raise an exception.
 https://github.com/CenterForOpenScience/osf.io/blob/develop/api/base/parsers.py#L58

pattisdr:
Regarding adding a primary_institution to a node, I think this is a good case for using relationship self links, similar to the proposed changes in https://openscience.atlassian.net/browse/OSF-5174 since the relationship between a node and its primary institution won't have other metadata (that I know of?).  This is how modifying node links will work in the future.

This would be a simpler request/response than creating a node contributor, for example, because the contributor relationship has additional attributes, permission & bibliographic, while you want to merely create a relationship between a node and an institution. 

See Updating To-One Relationships http://jsonapi.org/format/#crud-updating-to-one-relationships

*Self link*: /nodes/{node_id}/relationships/institution/
*Related link*: /nodes/{node_id}/institution/

So to add an institution: PATCH /nodes/{node_id}/relationships/institution/  
```
{
    "data":  {"type": institution", "id": "UVA"}
}
```

What do you think @brianjgeiger?

brianjgeiger:
Sounds right. I'm all for having the self links put in.

pattisdr:
Stopping my review: I've only looked at part of the InstitutionSerializer and class NodeInstitutionDetail.  Using relationship self links for adding an institution will require several changes.

api/institutions/serializers.py
```diff
@@ -0,0 +1,34 @@
+from rest_framework import serializers as ser
+
+from api.base.serializers import JSONAPISerializer, RelationshipField
+
+class InstitutionSerializer(JSONAPISerializer):
+    name = ser.CharField(required=True)
+    id = ser.CharField(required=True, source='_id')
+
+    nodes = RelationshipField(
+        related_view='institutions:institution-nodes',
```
pattisdr: If using relationship self links to add an institution for a node, will also need a `self_view` and `self_view_kwargs`.

pattisdr:
Stopping my review: I've only looked at part of the InstitutionSerializer and class NodeInstitutionDetail. Using relationship self links for adding an institution will require several changes.

api/base/parsers.py
```diff
@@ -119,3 +119,6 @@ class JSONAPIParserForRegularJSON(JSONAPIParser):
     Allows same processing as JSONAPIParser to occur for requests with application/json media type.
     """
     media_type = 'application/json'
+
+class JSONAPIRelationshipParser(JSONParser):
```
pattisdr: Class should have docstring.

pattisdr: I think the JSONAPIRelationshipParser needs its own parse method, even though the data is simple.  We use the JSONAPIParser parse method to do many things.  It checks the formatting to ensure it matches the jsonapi spec and then only returns the important properties.  Relationship data is going to come in looking like this:
```
{
    "data": {
          "id": "UVA",
          "type": "institution"
     }
}
```
The parser can strip off the "data" key; it isn't needed.  It should also validate that "id" and "type" are present.

api/nodes/serializers.py
```diff
@@ -373,3 +379,29 @@ class Meta:
     @staticmethod
     def get_id(obj):
         return '{}:{}'.format(obj.node._id, obj.provider)
+
+class NodeInstitutionRelationshipSerializer(JSONAPIRelationshipSerializer):
```
pattisdr: I would add id and type fields as required fields to this serializer to take advantage of some of the code we have.  `id` should be a CharField since the IDField would try to match node id to institution id.  `type` should be a TypeField.   This TypeField will take care of the 'institution' check.

api/base/parsers.py
pattisdr: I'm thinking something like this: 
```
    def parse(self, stream, media_type=None, parser_context=None):
        result = super(JSONAPIRelationshipParser, self).parse(stream, media_type=media_type, parser_context=parser_context)

        if not isinstance(result, dict):
            raise ParseError()
        data = result.get('data', {})

        if data:
            object_id = data.get('id')
            object_type = data.get('type')

            if object_id is None:
                raise JSONAPIException(source={'pointer': '/data/id'}, detail=NO_ID_ERROR)

            if object_type is None:
                raise JSONAPIException(source={'pointer': '/data/type'}, detail=NO_TYPE_ERROR)

            return data

        else:
            return {'_id': None, 'type': None}
```

api/nodes/serializers.py
pattisdr: With the changes on the parser, then `id` and `type` can be fields on the serializer.  `id` would be a CharField and `type` a TypeField.  The fields would have required=False, and maybe allow_null=True?

The TypeField would take care of ensuring the institution is the correct type, and throw a 409 error.

These changes on the parser and this serializer, to allow id and type to be fields on the serializer are really matters of opinion, but I think help keep things consistent with how our current parser works.

api/nodes/serializers.py
```diff
@@ -373,3 +379,29 @@ class Meta:
...
+
+    def relationship(self, obj):
+        return obj.primary_institution
+
+    def update(self, instance, validated_data):
+        node = instance
+        user = self.context['request'].user
+        data = self.context['request'].data['data']
+        if data:
+            assert data['type'] == 'institution', 'Not right endpoint for this type, expected \'institution\' '
```
pattisdr: This line can be removed if you decide to make type=TypeField(...) a field on the serializer.

api/nodes/serializers.py
```diff
@@ -373,3 +379,29 @@ class Meta:
...
+    class Meta:
+        type_ = 'institution'
+
+    def relationship(self, obj):
+        return obj.primary_institution
+
+    def update(self, instance, validated_data):
+        node = instance
+        user = self.context['request'].user
+        data = self.context['request'].data['data']
```
pattisdr: You can use validated_data instead of pulling data out of the request context.

api/nodes/serializers.py
```diff
@@ -373,3 +379,29 @@ class Meta:
...
+        if data:
```
pattisdr: If you add a parse_method to JSONAPIRelationshipParser, you could do a check like, if validated_data.get('_id).

api/institutions/views.py
```diff
@@ -0,0 +1,164 @@
...
+from api.base.utils import get_object_or_error
+from api.nodes.views import NodeMixin
+from api.nodes.serializers import NodeSerializer, NodeDetailSerializer
+from api.nodes.permissions import ContributorOrPublic
+from api.users.serializers import UserSerializer, UserDetailSerializer
+
+from .serializers import InstitutionSerializer, InstitutionDetailSerializer
+
+
+class InstitutionList(JSONAPIBaseView, generics.ListAPIView):
```
pattisdr: Normally these views would have documentation for how the endpoints work.  Are we going to be advertising this capability right now?

api/base/parsers.py
pattisdr: I'm thinking something like this: 
```
    def parse(self, stream, media_type=None, parser_context=None):
        result = super(JSONAPIRelationshipParser, self).parse(stream, media_type=media_type, parser_context=parser_context)

        if not isinstance(result, dict):
            raise ParseError()
        data = result.get('data', {})

        if data:
            object_id = data.get('id')
            object_type = data.get('type')

            if object_id is None:
                raise JSONAPIException(source={'pointer': '/data/id'}, detail=NO_ID_ERROR)

            if object_type is None:
                raise JSONAPIException(source={'pointer': '/data/type'}, detail=NO_TYPE_ERROR)

            return data

        else:
            return {'_id': None, 'type': None}
```

api/institutions/views.py
```diff
@@ -0,0 +1,164 @@
...
+    required_read_scopes = [CoreScopes.INSTITUTION_READ, CoreScopes.NODE_BASE_READ]
+    required_write_scopes = [CoreScopes.NULL]
+    model_class = Node
+
+    serializer_class = NodeSerializer
+    view_category = 'institutions'
+    view_name = 'institution-nodes'
+
+    def get_default_odm_query(self):
+        inst = Institution.load(self.kwargs['institution_id'])
```
pattisdr: This line is repeated many times. Is it worth moving this into a method similar to get_node()?  `def get_institution`?

api/institutions/views.py
```diff
@@ -0,0 +1,164 @@
...
+        inst_query = Q('primary_institution', 'eq', inst)
+        base_query = (
+            Q('is_deleted', 'ne', True) &
+            Q('is_folder', 'ne', True) &
+            Q('is_registration', 'eq', False)
```
pattisdr: FYI there's a PR in code review https://github.com/CenterForOpenScience/osf.io/pull/4540 that will allow registrations to show up under the `/v2/nodes/` endpoint.  If it gets merged before this, registrations should show up under institution nodes endpoints.

api/institutions/views.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def get_object(self):
+        inst = get_object_or_error(
+            Institution,
+            self.kwargs['institution_id'],
+            display_name='institution'
+        )
+        return inst
+
+class InstitutionNodeList(JSONAPIBaseView, ODMFilterMixin, generics.ListAPIView):
```
pattisdr: Class definitions should be surrounded by two blank lines https://www.python.org/dev/peps/pep-0008/#blank-lines

api/institutions/serializers.py
```diff
@@ -0,0 +1,32 @@
+from rest_framework import serializers as ser
+
+from api.base.serializers import JSONAPISerializer, RelationshipField, LinksField
+
+class InstitutionSerializer(JSONAPISerializer):
+    name = ser.CharField(required=True)
+    id = ser.CharField(required=True, source='_id')
```
pattisdr: Since you're not allowing users to create/modify institutions, I don't know if these need to be required fields.  There's no validation on this data coming in, so it's not being used.

api/nodes/serializers.py
```diff
@@ -115,6 +116,11 @@ class NodeSerializer(JSONAPISerializer):
         related_meta={'count': 'get_registration_count'}
     ))
 
+    primary_institution = RelationshipField(
+        related_view='nodes:node-institution-detail',
+        related_view_kwargs={'node_id': '<pk>'}
+    )
```
pattisdr: You have the functionality for adding a node institution but you need to give the user the link to do so.
```
primary_institution = RelationshipField(
        related_view='nodes:node-institution-detail',
        related_view_kwargs={'node_id': '<pk>'},
        self_view='nodes:node-relationships-institution',
        self_view_kwargs={'node_id': '<pk>'}
    )
```

api/base/serializers.py
```diff
@@ -701,6 +701,25 @@ def is_valid(self, clean_html=True, **kwargs):
 
         return ret
 
+class JSONAPIRelationshipSerializer(ser.Serializer):
+    def relationship(self, obj):
```
pattisdr: Thinking about this more, do you need this?  The relationship id can be found in `self.validated_data.get('_id')`.  If the original request had `{'data': null}`, then this would be `None`, and you could return None.  If there was some other problem before we got here, I would think you would have raised an exception.

api/base/serializers.py
```diff
@@ -701,6 +701,25 @@ def is_valid(self, clean_html=True, **kwargs):
...
+        relationship = self.relationship(obj)
+        if relationship:
+            data = collections.OrderedDict({'type': type_, 'id': relationship._id})
+        else:
+            data = None
+        if envelope:
+            ret[envelope] = data
+        else:
+            ret = data
+        return ret
```
pattisdr: L718-722 are the same as JSONAPISerializer - this is just the final touchup on the response data to see if it needs to be nested inside `data` or not.  Could this be consolidated?

api/base/serializers.py
```diff
@@ -701,6 +701,25 @@ def is_valid(self, clean_html=True, **kwargs):
...
+class JSONAPIRelationshipSerializer(ser.Serializer):
+    def relationship(self, obj):
+        raise NotImplementedError
+
+    def to_representation(self, obj, envelope='data'):
+        ret = {}
+        meta = getattr(self, 'Meta', None)
+        type_ = getattr(meta, 'type_', None)
+        assert type_ is not None, 'Must define Meta.type_'
+        relationship = self.relationship(obj)
```
pattisdr: `relationship` = self.validated_data.get('_id')?

api/institutions/serializers.py
```diff
@@ -0,0 +1,32 @@
...
+        return obj.get_api_url()
+
+    def get_absolute_url(self, obj):
+        return obj.get_absolute_url()
+
+    class Meta:
+        type_ = 'institutions'
+
+class InstitutionDetailSerializer(InstitutionSerializer):
+    pass
```
pattisdr: Is this serializer necessary?

api/institutions/views.py
```diff
@@ -0,0 +1,164 @@
...
+    required_write_scopes = [CoreScopes.NULL]
+    model_class = Institution
+
+    serializer_class = InstitutionDetailSerializer
+    view_category = 'institutions'
+    view_name = 'institution-detail'
+
+    def get_object(self):
+        inst = get_object_or_error(
+            Institution,
```
pattisdr: Recommend making a consolidated `get_institution` method.

api/nodes/serializers.py
pattisdr: Thinking about the json-api spec for relationships links.  I'm wondering if we should include links in addition to the data in the response. See http://jsonapi.org/format/#fetching-relationships-responses-200
```
{
  "links": {
    "self": "/nodes/abcde/relationships/institution",
    "related": "/nodes/abcde/institution"
  },
  "data": {
    "type": "institution",
    "id": "UVA"
  }
}
```

api/nodes/views.py
```diff
@@ -1726,3 +1730,42 @@ def perform_create(self, serializer):
...
+    def get_object(self):
+        node = self.get_node()
+        if not node.primary_institution:
+            raise NotFound
+        return node.primary_institution
+
+
+class NodeInstitutionRelationship(JSONAPIBaseView, generics.UpdateAPIView, NodeMixin):
```
pattisdr: You should be able to do a GET request to `/v2/nodes/<node_id>/relationships/institution/` as well.  The response would be the same simple response you get when adding an institution to a node.

api_tests/institutions/views/test_institution_detail.py
```diff
@@ -0,0 +1,24 @@
...
+from tests.base import ApiTestCase
+from tests.factories import InstitutionFactory
+
+from api.base.settings.defaults import API_BASE
+
+class TestInstitutionDetail(ApiTestCase):
+    def setUp(self):
+        super(TestInstitutionDetail, self).setUp()
+        self.institution = InstitutionFactory()
+        self.institution.save()
```
pattisdr: This save isn't needed.

api_tests/institutions/views/test_institution_nodes_detail.py
```diff
@@ -0,0 +1,48 @@
+from nose.tools import *  # flake8: noqa
+
+from tests.base import ApiTestCase
+from tests.factories import InstitutionFactory, AuthUserFactory, NodeFactory
+
+from api.base.settings.defaults import API_BASE
+
+class TestInstitutionNodeDetail(ApiTestCase):
```
pattisdr: Add test_return_private_node_non_contrib_auth.

api_tests/institutions/views/test_institution_users_detail.py
```diff
@@ -0,0 +1,32 @@
+from nose.tools import *  # flake8: noqa
+
+from tests.base import ApiTestCase
+from tests.factories import InstitutionFactory, AuthUserFactory
+
+from api.base.settings.defaults import API_BASE
+
+class TestInstitutionUsersList(ApiTestCase):
```
pattisdr: test name should be `TestInstitutionUserDetail`

api_tests/nodes/views/test_node_relationship_institution.py
```diff
@@ -0,0 +1,159 @@
...
+        self.user.save()
+        node = NodeFactory(creator=self.user)
+        res = self.app.put_json_api(
+            '/{0}nodes/{1}/relationships/institution/'.format(API_BASE, node._id),
+            {'data': {'type': 'not_institution', 'id': self.institution._id}},
+            expect_errors=True,
+            auth=self.user.auth
+        )
+
+        assert_equal(res.status_code, 500)
```
pattisdr: This should be a 409 Conflict.  Putting `type=TypeField(...)` on the NodeInstitutionRelationshipSerializer will take care of this.

api_tests/nodes/views/test_node_relationship_institution.py
```diff
@@ -0,0 +1,159 @@
+from nose.tools import *
+
+from framework.auth.core import Auth
+
+from tests.base import ApiTestCase
+from tests.factories import InstitutionFactory, AuthUserFactory, NodeFactory
+
+from api.base.settings.defaults import API_BASE
+
+class TestNodeRelationshipInstitution(ApiTestCase):
```
pattisdr: I would add tests for checking if your request format is incorrect: if type is missing, and your just have id, or vice versa.

pattisdr: Thorough tests :+1:

pattisdr:
Excited about this PR - especially your work on the relationship self links.  This is great functionality for the API - laying groundwork for to-many relationships like node_links!

API pass finished :sunny:

api/nodes/serializers.py
brianjgeiger: Yes, it looks like this would be appropriate.

api/institutions/urls.py
```diff
@@ -0,0 +1,12 @@
...
+
+from api.institutions import views
+
+urlpatterns = [
+    url(r'^$', views.InstitutionList.as_view(), name=views.InstitutionList.view_name),
+    url(r'^(?P<institution_id>\w+)/$', views.InstitutionDetail.as_view(), name=views.InstitutionDetail.view_name),
+    url(r'^(?P<institution_id>\w+)/nodes/$', views.InstitutionNodeList.as_view(), name=views.InstitutionNodeList.view_name),
+    url(r'^(?P<institution_id>\w+)/nodes/(?P<node_id>\w+)/$', views.InstitutionNodeDetail.as_view(), name=views.InstitutionNodeDetail.view_name),
+    url(r'^(?P<institution_id>\w+)/users/$', views.InstitutionUserList.as_view(), name=views.InstitutionUserList.view_name),
+    url(r'^(?P<institution_id>\w+)/users/(?P<user_id>\w+)/$', views.InstitutionUserDetail.as_view(), name=views.InstitutionUserDetail.view_name),
```
pattisdr: Should `institutions/{instution_id}/users/{user_id}/` and `institutions/{institution_id}/nodes/{node_id}/` be a thing?  You can have multiple URL's pointing to the same resource, but you don't actually give users these links.  If you navigate to `/institution/institution_id/users/`, this endpoint is using the UserSerializer, so each of the self links in the list point to the UserDetail view (`users/user_id`), not `institutions/{instution_id}/users/{user_id}/`. 

I don't these two links are needed since they don't reveal anything that `/users/user_id` and `nodes/node_id` don't already reveal.

pattisdr: You can think of them like node children:  `/v2/nodes/{node_id}/children/` The self links refer you to the child node: `/v2/nodes/{node_id}`, not `/v2/nodes/{node_id}/children/{child_id}/`

api/base/serializers.py
```diff
@@ -758,6 +758,27 @@ def is_valid(self, clean_html=True, **kwargs):
...
+        type_ = getattr(meta, 'type_', None)
+        assert type_ is not None, 'Must define Meta.type_'
+
+        data = collections.OrderedDict([
+            ('links', collections.OrderedDict()),
+            ('data', collections.OrderedDict()),
+        ])
+        relation_id_field = self.fields['id']
+
+        relationship = relation_id_field.to_representation(relation_id_field.get_attribute(obj))
```
pattisdr: I like this a lot because it's how the Django Rest Framework serializes the fields.

For readability, split it into two lines:
`attribute = relation_id_field.get_attribute(obj)`
`relationship = relation_id_field.to_representation(attribute)`

api/base/serializers.py
```diff
@@ -758,6 +758,27 @@ def is_valid(self, clean_html=True, **kwargs):
 
         return ret
 
+class JSONAPIRelationshipSerializer(ser.Serializer):
+
```
pattisdr: Class needs docstring.

api/institutions/views.py
```diff
@@ -0,0 +1,247 @@
...
+from api.base.views import JSONAPIBaseView
+from api.base.utils import get_object_or_error
+from api.nodes.views import NodeMixin
+from api.nodes.serializers import NodeSerializer, NodeDetailSerializer
+from api.nodes.permissions import ContributorOrPublic
+from api.users.serializers import UserSerializer, UserDetailSerializer
+
+from .serializers import InstitutionSerializer
+
+class InstitutionMixin(object):
```
pattisdr: :+1:

api/nodes/views.py
```diff
@@ -1745,3 +1748,115 @@ def perform_create(self, serializer):
...
+        node = self.get_node()
+        if not node.primary_institution:
+            raise NotFound
+        return node.primary_institution
+
+
+class NodeInstitutionRelationship(JSONAPIBaseView, generics.RetrieveUpdateAPIView, NodeMixin):
+    """ Relationship Endpoint for Node -> Instituion Relationship
+
+    Used to set the primary_institution of a node to an institution
```
pattisdr: This is a bit confusing because in InstitutionsList for example, type is "institutions" and for this endpoint, type is "institution".  This either needs to clearly be spelled out or we should just pick one. My feelings are that we shouldn't be worrying about single versus plural in the "type", this is really to differentiate is this a "users" type, or a "nodes" type, for example.   I feel the institutions should have "institutions" type across the board, plural.

api/institutions/views.py
```diff
@@ -0,0 +1,247 @@
...
+    ###Nodes
+    List of nodes that have this institution as its primary institution.
+
+    ###Users
+    List of users that are affiliated with this institution.
+
+    ##Links
+
+        self:  the canonical api endpoint of this node
+        html:  this node's page on the OSF website
```
pattisdr: this institution's page

api/institutions/views.py
```diff
@@ -0,0 +1,247 @@
...
+
+    ###Nodes
+    List of nodes that have this institution as its primary institution.
+
+    ###Users
+    List of users that are affiliated with this institution.
+
+    ##Links
+
+        self:  the canonical api endpoint of this node
```
pattisdr: of this institution

api/institutions/views.py
```diff
@@ -0,0 +1,247 @@
...
+        query = Q('affiliated_institutions', 'eq', inst)
+        return query
+
+    # overrides RetrieveAPIView
+    def get_queryset(self):
+        query = self.get_query_from_request()
+        return User.find(query)
+
+
+class InstitutionUserDetail(JSONAPIBaseView, generics.RetrieveAPIView, InstitutionMixin):
```
pattisdr: Related to this comment: https://github.com/CenterForOpenScience/osf.io/pull/4661/files#r47368430

Is this a needed view?

api/nodes/serializers.py
```diff
@@ -373,3 +381,37 @@ class Meta:
...
+    id = ser.CharField(source='institution_id', required=False, allow_null=True)
+    type = TypeField(required=False, allow_null=True)
+
+    links = LinksField({
+        'self': 'get_self_link',
+        'related': 'get_related_link',
+    })
+
+    class Meta:
+        type_ = 'institution'
```
pattisdr: My opinion is that this should be changed to 'institutions'.  I don't know if we should distinguish between single and plural.   When we create a node, type is "nodes" even though we're creating one node.

api/nodes/views.py
```diff
@@ -1745,3 +1748,115 @@ def perform_create(self, serializer):
...
+
+    ###Create
+
+        Method:        GET
+        URL:           /links/self
+        Query Params:  <none>
+        Success:       200,
+                       {
+                         "data": {                  # "data": null, if no institution
+                           "type": "institution",
```
pattisdr: institutions?

api/nodes/views.py
```diff
@@ -1745,3 +1748,115 @@ def perform_create(self, serializer):
...
+                       }
+
+    ###Create
+
+        Method:        PUT || PATCH
+        URL:           /links/self
+        Query Params:  <none>
+        Body (JSON):   {
+                         "data": {
+                           "type": "institution",   # required
```
pattisdr: institutions?

api/nodes/views.py
```diff
@@ -1745,3 +1748,115 @@ def perform_create(self, serializer):
...
+class NodeInstitutionRelationship(JSONAPIBaseView, generics.RetrieveUpdateAPIView, NodeMixin):
+    """ Relationship Endpoint for Node -> Instituion Relationship
+
+    Used to set the primary_institution of a node to an institution
+
+    ##Actions
+
+    ###Create
+
+        Method:        GET
```
pattisdr: You don't need to show the GET action.

api/institutions/views.py
```diff
@@ -0,0 +1,247 @@
...
+            Institution,
+            self.kwargs[self.institution_lookup_url_kwarg],
+            display_name='institution'
+        )
+        return inst
+
+
+class InstitutionList(JSONAPIBaseView, generics.ListAPIView):
+    """
+    Verified Institutions affiliated with COS
```
pattisdr: A paginated list of verified institutions affiliated with COS.

api/institutions/views.py
```diff
@@ -0,0 +1,247 @@
...
+        query = base_query & permission_query & inst_query
+        return query
+
+    # overrides RetrieveAPIView
+    def get_queryset(self):
+        query = self.get_query_from_request()
+        return Node.find(query)
+
+
+class InstitutionNodeDetail(JSONAPIBaseView, generics.RetrieveAPIView, NodeMixin, InstitutionMixin):
```
pattisdr: Related to this comment: https://github.com/CenterForOpenScience/osf.io/pull/4661/files#r47368430

Is this a needed view?

pattisdr:
Pass complete.

Almost there - most of the comments above relate to two points:

1) should `institutions/institution_id/nodes/node_id/` and `institutions/institution_id/users/user_id/` be endpoints? @brianjgeiger might have input.  

If so, you need to provide these links in the InstitutionNodeDetail view and InstitutionUserDetail view.  If not, you can remove these URL's and their associated views.  I feel they are not necessary as they don't add any new information.  You can just give references to the respective NodeDetail view or UserDetail view (as you are currently doing). 

2) change all institution types to "institutions" not "institution" to avoid confusion.

When you're done with your docs I think Fitz is a good person for review.

brianjgeiger:
1) Follow Dawn's suggestions; they do not need their own endpoints.

api/base/serializers.py
```diff
@@ -758,6 +758,34 @@ def is_valid(self, clean_html=True, **kwargs):
...
+        relation_id_field = self.fields['id']
+
+        attribute = relation_id_field.get_attribute(obj)
+        relationship = relation_id_field.to_representation(attribute)
+
+        data['data'] = {'type': type_, 'id': relationship} if relationship else None
+        data['links'] = {key: val for key, val in self.fields.get('links').to_representation(obj).iteritems()}
+
+        ret.update({envelope: data} if envelope else data)
+        return ret
```
pattisdr: Return `data` instead of `ret`.  Data and links should not need to be nested inside of `data`.  They should both be top-level keys.  http://jsonapi.org/format/#fetching-relationships

screen shot 2015-12-14 at 11 32 44 am

api/base/serializers.py
```diff
@@ -758,6 +758,34 @@ def is_valid(self, clean_html=True, **kwargs):
...
+        ])
+        relation_id_field = self.fields['id']
+
+        attribute = relation_id_field.get_attribute(obj)
+        relationship = relation_id_field.to_representation(attribute)
+
+        data['data'] = {'type': type_, 'id': relationship} if relationship else None
+        data['links'] = {key: val for key, val in self.fields.get('links').to_representation(obj).iteritems()}
+
+        ret.update({envelope: data} if envelope else data)
```
pattisdr: Remove line.

api/base/serializers.py
```diff
@@ -758,6 +758,34 @@ def is_valid(self, clean_html=True, **kwargs):
...
 
+class JSONAPIRelationshipSerializer(ser.Serializer):
+    """Base Relationship serializer. Requires that a `type_` option is set on `class Meta`.
+    Provides a simplified serialization of the relationship, allowing for simple update request
+    bodies. Also provides links to itself and to the endpoint providing information about the related
+    object.
+    """
+
+    def to_representation(self, obj, envelope='data'):
+        ret = {}
```
pattisdr: Remove line.

api/base/serializers.py
```diff
@@ -758,6 +758,34 @@ def is_valid(self, clean_html=True, **kwargs):
...
+        meta = getattr(self, 'Meta', None)
+        type_ = getattr(meta, 'type_', None)
+        assert type_ is not None, 'Must define Meta.type_'
+
+        data = collections.OrderedDict([
+            ('links', collections.OrderedDict()),
+            ('data', collections.OrderedDict()),
```
pattisdr: Can you switch the order of data and links to match the order of nodes?

pattisdr:
Looks good Henrique.  Only a couple of changes to make the relationships/institution response look like this:
screen shot 2015-12-14 at 11 58 48 am

api/base/parsers.py
```diff
@@ -119,3 +119,37 @@ class JSONAPIParserForRegularJSON(JSONAPIParser):
...
+
+            if id_ is None:
+                raise JSONAPIException(source={'pointer': '/data/id'}, detail=NO_ID_ERROR)
+
+            if type_ is None:
+                raise JSONAPIException(source={'pointer': '/data/type'}, detail=NO_TYPE_ERROR)
+
+            return data
+
+        return {'type': None, 'id': None}
```
brianjgeiger: This seems like it should raise an exception. Under what circumstance would it hit here? Or are you following a pattern we used from somewhere else?

brianjgeiger:
Since this is for QA (and reviewers) it would be good to detail how one would use these things in the PR description or the Jira ticket description. Give an example especially of how you'd use the relationship self link: what you start with, what you send, what gets changed when you send it). Also what information is serialized by the various endpoints and how they relate to other things.

It would be good to link back to the Jira ticket from here as well, especially if all that is documented there.

api/institutions/views.py
```diff
@@ -0,0 +1,188 @@
...
+    def get_institution(self):
+        inst = get_object_or_error(
+            Institution,
+            self.kwargs[self.institution_lookup_url_kwarg],
+            display_name='institution'
+        )
+        return inst
+
+
+class InstitutionList(JSONAPIBaseView, generics.ListAPIView):
```
brianjgeiger: This should be filterable, so see about adding the ODM filter mixin from api/base/filters.py

api/base/parsers.py
pattisdr: I recommended this - this is because with the relationships self link you can send 
``
{
  "data": null
}
` to clear the institution.

api/institutions/views.py
```diff
@@ -0,0 +1,188 @@
...
+    def get_queryset(self):
+        return Institution.find()
+
+
+class InstitutionDetail(JSONAPIBaseView, generics.RetrieveAPIView, InstitutionMixin):
+    """ Details about a given institution.
+
+    ###Permissions
+
+    All institutions are available to be read by everyone. However, no one has write
```
brianjgeiger: I don't think you'll need the bit about write permission. The auto-generated docs will understand this from the RetrieveAPIView class.

api/base/parsers.py
brianjgeiger: :+1:

api/nodes/views.py
brianjgeiger: Agreed. Plural everywhere.

api/institutions/serializers.py
```diff
@@ -0,0 +1,29 @@
+from rest_framework import serializers as ser
+
+from api.base.serializers import JSONAPISerializer, RelationshipField, LinksField
+
+class InstitutionSerializer(JSONAPISerializer):
+    name = ser.CharField(required=False)
+    id = ser.CharField(required=False, source='_id')
+    logopath = ser.CharField(source='logo_path')
```
brianjgeiger: Should be logo_path

framework/auth/core.py
```diff
@@ -1331,8 +1334,63 @@ def n_projects_in_common(self, other_user):
...
     '''Merge multiple sorted inputs into a single output in reverse order.
     '''
     return sorted(itertools.chain(*iterables), reverse=True)
+
+
+class Institution(StoredObject):
+
+    _id = fields.StringField(index=True, unique=True, primary=True)
+    name = fields.StringField(required=True)
+    logo_name = fields.StringField()
```
brianjgeiger: Is this different from `logo_path`?

brianjgeiger: Ah, I found it. nvm.

brianjgeiger: Should this be required? It looks like you're counting on the existence of a logo in the front end.

brianjgeiger:
Okay, that's what I found through inspection. After merge conflicts are resolved, I'll check out the branch locally and try it out, and I will also look more closely at the tests. Pass done. :penguin:

api/base/serializers.py
```diff
@@ -835,14 +835,13 @@ class JSONAPIRelationshipSerializer(ser.Serializer):
     Provides a simplified serialization of the relationship, allowing for simple update request
     bodies.
     """
-    id = ser.CharField(source='node._id', required=False, allow_null=True)
+    id = ser.CharField()
```
mfraezz: It makes sense to not define `source` in the base class, but I'm not sure that `required` and `allow_null` should have been removed here.

api/nodes/permissions.py
```diff
@@ -30,7 +30,18 @@ def has_object_permission(self, request, view, obj):
         else:
             return node.has_permission(auth.user, osf_permissions.ADMIN)
 
+class AdminOrPublicRelationship(AdminOrPublic):
```
mfraezz: This seems superfluous. See parent class above.

api/nodes/serializers.py
```diff
@@ -429,6 +436,54 @@ class Meta:
...
+        return obj.institution_relationship_url()
+
+    def get_related_link(self, obj):
+        return obj.institution_url()
+
+    def update(self, instance, validated_data):
+        node = instance
+        user = self.context['request'].user
+
+        inst = validated_data.get('institution_id')
```
mfraezz: You may want to make this explicitly default to `None`

api/nodes/views.py
```diff
@@ -1909,3 +1913,98 @@ def perform_create(self, serializer):
...
+
+        Method:        PUT || PATCH
+        URL:           /links/self
+        Query Params:  <none>
+        Body (JSON):   {
+                         "data": {
+                           null
+                         }
+                       }
+        Success:       200
```
mfraezz: I think this should be a `204`:

>A server **MUST** return a `204 No Content` status code if an update is successful and the representation of the resource in the request matches the result.
-http://jsonapi.org/format/#crud-updating-relationship-responses

api_tests/nodes/views/test_node_institutions_detail.py
```diff
@@ -0,0 +1,29 @@
...
+
+        assert_equal(res.status_code, 200)
+        assert_equal(res.json['data']['attributes']['name'], self.institution.name)
+        assert_equal(res.json['data']['id'], self.institution._id)
+
+    def test_return_no_institution(self):
+        url = '/{0}nodes/{1}/institution/'.format(API_BASE, self.node2._id)
+        res = self.app.get(url, expect_errors=True)
+
+        assert_equal(res.status_code, 401)
```
mfraezz: Shouldn't this be a `404`?

>Detail of the one primary_institution a node has, if any. Returns NotFound if the node does not have a primary_instituion.
-https://github.com/CenterForOpenScience/osf.io/pull/4661/files#diff-51eef6ea2a6c35e2b080c9a6fe591e38R1918

hmoco: Yep, 401 is for not authorized, which this is returning because no auth is being passed to a private node. Will fix test to pass auth to a node with no institution.

website/templates/project/project.mako
```diff
@@ -21,7 +21,7 @@
                     % endif
                 % endif
                 <h2 class="node-title">
-                    <span id="nodeTitleEditable" class="overflow">${node['title']}</span>
+                    <span id="nodeTitleEditable" class="overflow"> ${node['title']}</span>
```
mfraezz: This slightly affects UI by prepending a space to the node's title. Please remove.

website/project/model.py
```diff
@@ -790,6 +790,31 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin):
 
     piwik_site_id = fields.StringField()
 
+    # Primary institution node is attached to
+    primary_institution = fields.ForeignField('institution', backref='node')
```
mfraezz: Please remove this backref. `Node.find(Q('primary_institution', 'eq', institution._id))` would work as well when you need to find the associated nodes.

mfraezz:
Pass complete, will move on to functional testing when above comments have been addressed. :spaghetti:

hmoco:
@mfraezz back at ya. (Also check out that sweet commit message!)

framework/auth/core.py
```diff
@@ -1345,8 +1348,63 @@ def n_projects_in_common(self, other_user):
...
 
 def _merge_into_reversed(*iterables):
     '''Merge multiple sorted inputs into a single output in reverse order.
     '''
     return sorted(itertools.chain(*iterables), reverse=True)
+
+
+class Institution(StoredObject):
+
+    _id = fields.StringField(index=True, unique=True, primary=True)
```
mfraezz: You want to set `default=lambda: str(ObjectId())`, or an `_id` won't get assigned on save. `ObjectId` can be imported from either `framework.mongo` or `bson`.

hmoco: Would that still occur if a custom id is passed?

mfraezz: No, if one is explicitly set at institution creation, that also works. I suppose if we only have people creating them through the admin interface, it should be fine as is.

mfraezz:
LGTM. :battery: 

Sending it on.

website/templates/institution.mako
```diff
@@ -0,0 +1,27 @@
+<%inherit file="base.mako"/>
+
```
sloria: This template doesn't appear to be used anywhere. Can it be removed?

tests/factories.py
```diff
@@ -523,6 +524,24 @@ def _create(cls, target_class, *args, **kwargs):
...
+    def _create(cls, target_class, _id, name):
+        instance = target_class(
+            _id=_id,
+            name=name,
+            logo_name='logo.img',
+        )
+        instance.save()
+        return instance
+
+    _build = _create
```
sloria: `_build` should not save the object. Only `_create` should save.

sloria:
Looks good! Just a couple minor things to clean up. Otherwise good to merge. :octopus:

hmoco:
@sloria, just responded!

PR #4663, opened 2015-11-30
View-only links for API v2

brianjgeiger:
Purpose
-----------
Enable use of view-only (a.k.a private) links, and prevent anonymous data from leaking in the case of using the links.  Closes https://openscience.atlassian.net/browse/OSF-4501

Changes
------------
1. Pass along the view_only key to user auth when accessing the safe functions of node access.
2. Create an optional field for serializers to implement if their data is necessarily unsafe (user and contributor are the only things that need this.
3. Modify the JSONAPISerializer to strip anonymized data from the serialization.

Details on expected output to follow soon

Side-effects
----------------
If you are using an anonymous link in your request, you will not be able to serialize any user data at all. Make the request again without the link if you think you should have access to the data without the permissions that the link grants you.

Samples
-----------
Normal nodes request

screen shot 2015-12-01 at 9 15 14 am

Private node with private (non-anonymous) key

screen shot 2015-12-01 at 9 15 37 am

Private node with anonymous key

screen shot 2015-12-01 at 9 15 58 am

Normal user request

screen shot 2015-12-01 at 9 16 19 am

User request with non-anonymous key

screen shot 2015-12-01 at 9 16 54 am

User request with anonymous key

screen shot 2015-12-01 at 9 17 14 am


lbanner:
OSF-4501 Thanks @brianjgeiger !

api/base/serializers.py
```diff
@@ -653,7 +676,8 @@ def to_representation(self, obj, envelope='data'):
                         data['embeds'][field.field_name] = result
                 else:
                     try:
-                        data['relationships'][field.field_name] = field.to_representation(attribute)
+                        if not (is_anonymous and hasattr(field, 'view_name') and field.view_name.find('users:') != -1):
```
cwisecarver: The string matching on a view name terrifies me a little bit. Could you make a base serializer class for all "user serializers" and then do an `not isinstance(field.root, BaseUserSerializer)` Just thinking of the inevitable refactor when someone changes every instance of user to member or something.

brianjgeiger: Whhaaaa? Why don't you like… no, that sounds like a good plan. I lean towards doing a mixin instead. Does that suit you?

cwisecarver: For sure, a mixin would work. Just some thing that a pycharm refactor by an intern probably won't break.

If you're not worried about it, I'm not. You'll probably be the one refactoring it anyway whenever it changes.

brianjgeiger: No, I like the isinstance idea much better.

api/base/serializers.py
```diff
@@ -44,6 +48,16 @@ def format_relationship_links(related_link=None, self_link=None, rel_meta=None,
     return ret
 
 
+def is_anonymized(request):
+    is_anonymous = False
+    private_key = request.query_params.get('view_only', None)
+    if private_key is not None:
+        link = PrivateLink.find_one(Q('key', 'eq', private_key))
```
cwisecarver: PrivateLink.find_one throws a NoResultsFound exception so `/v2/nodes/3wkv5/?view_only=woot_woot&format=jsonapi` crashes to a django stack trace, assuming of course that you do not have a PrivateLink with a key of `woot_woot` in your database.

api/base/serializers.py
```diff
@@ -653,7 +684,9 @@ def to_representation(self, obj, envelope='data'):
                         data['embeds'][field.field_name] = result
                 else:
                     try:
-                        data['relationships'][field.field_name] = field.to_representation(attribute)
+                        if not (is_anonymous and hasattr(field, 'view_name') and
+                                        isinstance(field.root, DoNotRelateWhenAnonymous) != -1):
```
sloria: This appears incorrect; `isinstance` returns a boolean, not an integer.

brianjgeiger: Yeah. Maybe I was checking a list at some point? No idea why I did that.

api/nodes/serializers.py
```diff
@@ -219,12 +220,7 @@ class NodeContributorsSerializer(JSONAPISerializer):
                                  default=osf_permissions.reduce_permissions(osf_permissions.DEFAULT_CONTRIBUTOR_PERMISSIONS),
                                  help_text='User permission level. Must be "read", "write", or "admin". Defaults to "write".')
 
-    links = LinksField(add_dev_only_items({
-        'html': 'absolute_url',
-
-    }, {
-        'profile_image': 'profile_image_url',
```
sloria: Why was `profile_image` removed?

brianjgeiger: Contributors do not have profile_images. Users do.

sloria: OK, thanks

sloria:
Looks good. Once the minor issues above addressed and this is updated with develop, this should be good to merge.

brianjgeiger:
Merged and fixed.

PR #4669, opened 2015-12-01
[feature] Add Edge-Side-Include functionality to API

cwisecarver:
## Description

- **Purpose**
    - [Edge-Side-Includes](https://en.wikipedia.org/wiki/Edge_Side_Includes) allow Varnish to replace specific xml tags with the contents of the url specified at the href property.
        - Example: `<esi:include href="http://www.google.com"/>` would be replaced with the contents of the google homepage.
    - The embed functionality in the API makes this very useful since a user embedded into a node is the same as a user requested individually. So varnish can cache `/v2/users/1234z/` and anytime that user is embedded in a request it can use the cached version replacing the `<esi:include href="/v2/users/1234z/"/>` with the full json representation of the user.

- **Changes**
    - There is a `to_esi_representation` method that was added on certain serializers. This method takes the place of the `to_representation` when the `ENABLE_ESI` setting is `True` and a resource is embedded.
    - Instead of python embedding the full content of the user as it does with to_representation it responds with `<esi:include href="{{ self.get_absolute_url() }}" />`. When the response is processed by Varnish it will do a request for the specified href and replace the xml tag with the response from that request and caching the response. The next time that url is included in an esi tag it will be pulled from cache.
    - Blinker events were added to GuidStoredObject that fire upon model save. When these events fire they're caught by a listener that fires off a celery task to send a `BAN` HTTP request to the varnish server in the pattern of `BAN {{ self.get_absolute_url }}.*`. This is processed as a regex and should account for all possible urls that represent that object. Since objects are stored by varnish individually, even when they're embedded in another object, this pattern should remove all instances, even those that are embedded, from cache.
        - `BAN /v2/users/1234z/.*`, to varnish, means add this pattern to the list of invalid url patterns. When someone requests a URL that matches an invalid url pattern make sure to request it from the backend server instead of serving it from cache. BANning instead of PURGEing means that we can avoid the thundering herd problem that commonly occurs with cache invalidation and it also means that we're not actually emptying the cache until we need to.

- **Side Effects**
	- URLs are supposed to represent unique data. Since we combine URLs and user authentication information to represent unique data this complicates caching. Varnish's configuration is VARYing on osf cookie, Authorization HTTP header, and URL. This **should** account for different users hitting the same url and getting a different result based on their individual permissions. There is always a chance that I've missed something. My hope is that the only side-effect of this is a low cache hit-rate. Meaning that since we're caching things per user our cache won't be utilized as well as it could for non-anonymous users.
	- Embed code is somewhat in flux right now. I've been basing my changes off of the two PRs that Dawn has been working on related to embeds but they have been changing while I've been working. I've tried to keep this branch as up to date with her stuff as I possibly could but there is a chance that a merge has gone sideways or I've missed a PR that was related to the embed work.
	- Since everything is behind a setting, unless you've switched ENABLE_ESI to true there should be no change from this PR.
    - The JSONAPIRenderer has a bit of code in it that I would really like to refactor but I couldn't find a better way to do it. Right now, because ESI is XML and all our stuff is JSON, DRF is quoting the value of the `to_esi_representation` method. Varnish doesn't expect ESI tags to be in quotes so `"<esi:include href="something"/>"` is ignored. The JSONAPIRenderer.render method has a regex that removes the double-quotes from the outside of the xml tag. This causes invalid JSON to be passed to varnish and varnish replaces the ESI tag prior to the client seeing it making it valid JSON again. While I love regex it can be brittle. Advice and alternate approaches are welcomed.

	
## Test Setup

For testing/CR, you'll need to install and setup varnish. A sample VCL file is attached to this PR. VCL is a domain-specific language designed to be used to define request handling and document caching policies for Varnish Cache. It is compiled when varnish starts up. The final VCL, very similar to the attached one, is included in the private docker repo.

- **Assumptions**
	- Your API server is running on port 8000
	- You want Varnish to listen on port 8193
	- You've got 2 gigs of free ram for varnish to use for caching (change `-s malloc,2G` in the varnish startup command to something else, `-s malloc,256M` for instance)
- **Steps**
	- `brew update; brew install varnish`
	- Download the attached VCL file and copy it to `/usr/local/etc/varnish/default.vcl` or wherever your varnish installation is. [default.vcl.txt](https://github.com/CenterForOpenScience/osf.io/files/48741/default.vcl.txt)
	- Startup varnish with the following command `sudo /usr/local/opt/varnish/sbin/varnishd -n /usr/local/var/varnish -f /usr/local/etc/varnish/default.vcl -s malloc,2G -T 127.0.0.1:2000 -a 127.0.0.1:8193 -p feature=+esi_disable_xml_check -p workspace_backend=262144 -p workspace_client=262144`
	- Change the `ENABLE_ESI` setting in your `api/base/settings/local.py` to `True`
	- Navigate to `http://localhost:8193/v2/nodes/?format=jsonapi&embed=contributors&page[size]=100`. In your `inv apiserver` console you should see the initial request followed by several other requests for individual contributors. This is varnish replacing the ESI tags with the individual responses. If you refresh this page you should see your pageload time decrease significantly and you should see no requests in your `inv apiserver` console.
	- Now update a contributor or node that was included in your previous request. If celery is running in your development environment the next time you request the list of nodes you should see the node and or the contributor you updated be requested from the API server but the other components should be pulled from the varnish cache. This means that the signal fired upon save, the celery task banned that content from varnish, and varnish re-requested the banned content the next time you needed it.
	- Making the same list-of-nodes request with either a non-logged-in user or logged in as a different user should cause requests to the python apiserver. But repeated requests as a non-logged in user or repeated requests as the same logged in user should be drawn from cache instead of making a request to the python apiserver.

## Random informations

- Varnish is **ONLY** configured to cache GET and HEAD requests where the URL starts with /v2/. Anything that does not meet that criteria will be passed directly to the python api server.
- If you open the developer console in your browser and have a look at the response headers you should see `X-Cache: HIT` (it was a cache hit) and `X-Cache-Hits: #` (there have been # hits on this version of the cached object) and `Age: ##` (the cached object is ## seconds old) and `X-beresp-ttl: 120.000` (the time-to-live of this cached object is 120.000 seconds). When `Age` == `X-beresp-ttl` the object is removed from the cache.
- We can send headers back from DRF to control the TTL based on whatever we'd like. Individual objects could be cached for days or seconds. This is not implemented yet.


api_tests/caching/test_caching.py
```diff
@@ -0,0 +1,239 @@
...
+        if not settings.RUN_VARNISH_IN_DEV:
+            return
+        self.authorization = HTTPBasicAuth(u'mocha@osf.io', u'password')
+
+    def test_compare_python_responses_to_varnish_responses(self):
+        if not settings.RUN_VARNISH_IN_DEV:
+            return
+        querystrings = dict(
+            nodes=[
+                u'comments',
```
chrisseto: `from __future__ import unicode_literals` will make all strings in the file unicode by default.

cwisecarver: You're right of course. I'll change that. I was getting keys that weren't matching and I put the `u` all over the place while debugging and never went back and cleaned it up.

sloria:
@cwisecarver Please resend this to `develop` if this PR is still relevant

Issue #249, opened 2014-02-21
Crowded links in add contributor modal


sloria:
It's difficult to see that there are two links
- Import contributors from..
- Get list of recently added ...

![open science framework project](https://f.cloud.github.com/assets/2379650/2235699/a5638ef0-9b4c-11e3-9e37-149109c00400.png)

Issue #206, opened 2013-12-18
Allow reverting to previous versions of wiki pages


jmcarp:
Not a top priority; just noting this for now.

PR #4670, opened 2015-12-01
[WIP] Evernote addon:  skeleton with OAuth authorization and notebook selection

rdhyee:
Per suggestion from @lbanner and @samchrisinger, I'm issuing this pull request to begin process of eventually getting the first version of an Evernote addon merged into `develop`.  This is clearly a WIP.  

The basic approach I've taken is to build from the ground up an addon modeled mostly on the Box addon.  A few things to know:
- Evernote uses OAuth 1.0a.  I had to make a small fix [outside of the Evernote addon directory](https://github.com/CenterForOpenScience/osf.io/compare/develop...rdhyee:evernote-oauth?expand=1#diff-6391da5157423e52547c9e0c599e76d9R168) to account for OAuth 1.0a.
- Evernote has a production vs sandbox.
- I mimicked Box in treating Evernote notebooks like Box folders. There are definite limits to this model:  notebooks contain notes but not other folders.  There are stacks, which can contain folders.  (I don't implement stacks yet.)
- There is obviously extraneous stuff included to scaffold my learning process.  Also I'm sure there mistakes reflecting limited understanding of osf.io.  Please point them out.


tasks.py
```diff
@@ -57,6 +57,22 @@ def server(host=None, port=5000, debug=True, live=False):
 
 
 @task
+def debug_server(host=None, port=5000, debug=True, live=False):
```
samchrisinger: Can you please explain the use case for this task? Typically we just add:
`import ipdb; ipdb.set_trace()` to trigger a breakpoint in the code.

tasks.py
```diff
@@ -57,6 +57,22 @@ def server(host=None, port=5000, debug=True, live=False):
...
+    """Run the app server with debugging"""
+
+    def run_app():
+
+        from website.app import init_app
+        app = init_app(set_backends=True, routes=True)
+        settings.API_SERVER_PORT = port
```
samchrisinger: Can you please explain why you're changing the API server port? This is a different app than the main flask app.

rdhyee: That's definitely legacy stuff I need to get rid of.  I wrote that code **before** I figured out how to use `import pdb; ipdb.set_trace()`.

tasks.py
rdhyee: That's definitely legacy stuff I need to get rid of. I wrote that code before I figured out how to use import pdb; ipdb.set_trace().

website/addons/evernote/serializer.py
```diff
@@ -0,0 +1,77 @@
...
+        EvernoteNodeSettings record. Provides the return value for the
+        evernote config endpoints.
+        """
+        valid_credentials = True
+        user_settings = node_settings.user_settings
+        self.node_settings = node_settings
+        current_user_settings = current_user.get_addon('evernote')
+        user_is_owner = user_settings is not None and user_settings.owner == current_user
+
+        # if user_settings:
```
samchrisinger: For the Box addon, these lines perform a simple (and arbitrary) API call to see if the credentials are valid. I'm not sure this is totally necessary, but might be simple to implement.

samchrisinger:
Without digging in very deeply, this code looks good so far. 

@lbanner I do think we need some discussion about the unimplemented aspects of this work, in particular what the real UI for this addon should look like and do. We have examples of addons that deviate from the norm set by 'storage' addons-- in particular Mendeley and Zotero. One path Raymond might take would be to develop a widget (similar to citations), that allows users to add/edit/delete Evernote notes.

An incremental addition to this would be wiki integration, but this seems like a much more daunting task.

tasks.py
samchrisinger: :+1: I was guessing so, not a problem

rdhyee:
@samchrisinger The next thing I was thinking of doing was seeing whether I could get Evernote notes to show up as a file in the Files part of the a project.  I have a some preliminary code on rendering Evernote notes from ENML (the Evernote Markup Language) into HTML.  Is that a reasonable next step?

@lbanner @samchrisinger I'm definitely open to talking further with you about the UI and looking at models other than the storage addons.  At the moment, using Box still seems like a good starting point, but maybe not for long.

samchrisinger:
@rdhyee I think whatever path you end up taking rendering ENML as HTML, markdown, or plain text is a must. You mentioned when we talked about ENML to markdown conversion-- we're already using markdown-it (https://github.com/markdown-it/markdown-it) for client-side markdown rendering. So yes, this seems like a good first step. 

After this is in place, I think it's reasonable to look at how Mendeley and Zotero integrate with the citations widget (try enabling one, then visiting the project dashboard)
screen shot 2015-12-01 at 12 09 43 pm. 
I'm imagining an Evernote widget that shows a list of notes from the linked notebook. It allows users to add, update, and delete notes. 

If you do follow this path, two things to note:
- No need to develop a waterbulter provider
- There would be benefit to creating wireframes or mockups of the widget before actually creating UI. These should outline what the different widget states will look like, where buttons go, etc. These can be very rough, but enough to convey the general structure. A developer here can then review the widget plans and can make suggestions as needed.

samchrisinger:
Also, we use flake8 to lint and check .py files. This can be useful for identifying unimported modules and bugs in general. There is an invoke task to run this: `inv flake`. All travis builds with flake8 errors will fail.

rdhyee:
Thanks, @samchrisinger, for your feedback!  I'll have now think a bit about next steps and then write them up.

rdhyee:
A next step for me is to figure out why nothing from the evernote addon shows up in a project page (e.g., http://localhost:5000/7pbqz/) as shown in the following screenshot:

image

I thought that my implementation so far would get something to show up. There could many reasons, including not having waterbutler provider for evernote.

rdhyee:
To answer [my question from yesterday](https://github.com/CenterForOpenScience/osf.io/pull/4670#issuecomment-161127825) about how to get the Evernote addon to show up on the project page like the Zotero addon does:   add `widget` to the `VIEWS` setting for the addon (`HAS_HGRID_FILES` controls placement into the Files "widget"):

```
VIEWS = ['widget']
HAS_HGRID_FILES = False  # set to True to put into Files widget
```

rdhyee:
@samchrisinger I've had a chance to get to know widgets a bit more. (I'm a much stronger at Python  than I am at JavaScript.  So on a tight timeline, I'm inclined to favor a Python-heavier solution, such as WaterButler (I don't see much JS at all in the box addon, for example.)  With some time to learn, I'd definitely game to pursue a JS heavy approach, which I think is the case for making a widget.

You [wrote yesterday](https://github.com/CenterForOpenScience/osf.io/pull/4670#issuecomment-161036908) "I'm imagining an Evernote widget that shows a list of notes from the linked notebook. It allows users to add, update, and delete notes."  My first pass is would be make use of the existing boxes and layout used in the citation addons and make them do things for Evernote.  Instead of searching for a citation style, the user could be searching and selecting a notebook -- or a note. The grid holding the citations would be retrofitted to hold the titles of notes and I could have an action to copy to the clipboard a URL for the note.  Or in moving further away from the citation model, maybe I'd have an edit window holding the note itself.

Tomorrow, I think I'll take a look at what it'd take to do a WaterButler implementation for Evernote. 

Definitely interested in your thoughts.

samchrisinger:
Whether or not you end up writing a WaterButler provider is in my mind fairly independent of how much javascript you'll have to write to get this working. WaterButler's function is as a file server between third party API's and clients, and to me seems like the last potential feature you should consider implementing.

Your acknowledgement of time constraints is definitely appreciated, but I'll reiterate that we probably won't merge semi-functional code or a tool below some minimum feature set. No matter how you proceed, there would be some benefit in trying to clearly articulate the goals and minimum viable product here. 

I do think there's some value in looking more at the citations widget, but am not convinced it the best or easiest model for what you'll want to accomplish. The citations widget uses our hierarchical data renderer Treebeard (not the most beginner friendly), whereas you might get away with something far far simpler. Off the top of my head, a skeleton (using Knockout.js) for this might look like:

```javascript
var EvernoteWidget = function($element, urls) {
    var self = this;

    self.urls = urls;

    self.notes = ko.observableArray();
    self.fetchNotes = $.getJSON.bind(null, urls.list, function(notes) {
        self.notes(notes);
    });
};

EvernoteWidget.prototype.init = function() {
    this.fetchNotes();
};

EvernoteWidget.prototype.addNote = function(note) {
    return $.post(this.urls.create, note).done(addNoteToNotesArray).fail(handleError);
};

EvernoteWidget.prototype.editNote = function(note) {
    return $.ajax({
        url: this.urls.update + note.id + '/',
        method: 'PUT',
        data: JSON.stringify(note)
    }).done(updateNoteInArray).fail(handleError);
};

EvernoteWidget.prototype.deleteNote = function(note) {
    return $.ajax({
        url: this.urls.delete + note.id + '/',
        method: 'DELETE'
    }).done(removeNoteFromNotesArray).fail(handleError);
};
```

```html
<div id="evernoteWidget">
  <div class="col-md-12">
    <div class="row">
      <button class="btn btn-success" data-bind="click: openAddDialog"></button>
    </div>
    <div class="row">
      <div data-bind="foreach: notes">
        <p>
          <button class="btn btn-default" data-bind="click: openEditDialog"></button>
          <button class="btn btn-danger" data-bind="click: openDeleteNoteDialog"></button>
        </p>
        <p>          
          <span data-bind="text: note.title"></span>
          <p data-bind="text: note.body"></p>
        </p>
      </div>
    </div>                
  </div>
</div>
```
* note some methods are deliberately unimplemented


Later steps here could be:
-  pagination (both in fetching notes from the server and the UI)
-  a search bar to quickly find a note by title/text

rdhyee:
@samchrisinger I appreciate your writing that "we probably won't merge semi-functional code or a tool below some minimum feature set."  I would expect nothing less.  I'm hoping to work out with @AndrewSallans a mutually agreeable timeline that enables me to work we're all happy with.

I also agree that I should "clearly articulate the goals and minimum viable product here."  You are providing me exactly the type of help I need to get to point of articulating what a MVP for Evernote would be.  From this conversation, I'm coming to the conclusion that 

* expressing Evernote functionality as a widget is a promising approach and
* I should not try to reuse too much UI/code from the citation widgets

Thank you so much for the [skeleton code for an Evernote widget](https://github.com/CenterForOpenScience/osf.io/pull/4670#issuecomment-161625047).  That "some methods are deliberately unimplemented" makes it even more helpful, I suspect. :-)  As I get a simple widget working, I will be able to articulate more  concretely what I believe a MVP for an Evernote widget actually is.

samchrisinger:
Great!. I'm glad this helps to get you on your way. 

I don't mean to scare you away from reusing code (please do when you can!), but not all of the code in the OSF codebase is what I'd call exemplary :).

rdhyee:
@samchrisinger I understand your point about reusing code -- I'd be very glad to do so.  (I've been an avid reader of the existing code, which I've been working hard to understand and reuse.) It's very helpful to find the easier-to-understand parts first -- and your skeleton is very helpful for someone new to osf.io!

rdhyee:
@samchrisinger re https://github.com/rdhyee/osf.io/commit/df6e3e642c43c46af3764d99ee7d43af146d075b: It took a bit of (clarifying) work to figure out how to get the tiniest skeleton of a widget for Evernote (which I can now use to plug in your more fulsome skeleton) -- I can show that the Knockout binding is working in my skinny frame.  I'm happy to have this to build out now.

samchrisinger:
This looks like a good step forward!

rdhyee:
@samchrisinger I have a question about how use attach an `init` function to the `EvernoteWidget` in [your outline](https://github.com/CenterForOpenScience/osf.io/pull/4670#issuecomment-161625047):

```JavaScript
EvernoteWidget.prototype.init = function() {
    this.fetchNotes();
};
```

When I first saw `init`, I assumed that the function would be either automatically called when the object is instantiated -- or by some other code in the OSF framework.  But in my experience, I've had to explicitly call `init` for it to run -- for example, in [my most recent commit](https://github.com/CenterForOpenScience/osf.io/commit/cbdbc9ebc4e686c2cc69faa977cefeffac3ae704#diff-20a73ce740934732aa4dc96cff68a7e4R18).   Am I missing something about how `init` is supposed to work?

samchrisinger:
@rdhyee, no this is expected and is one of the minor annoyances of Knockout.js. We generally add an `init` function to view models that make AJAX calls when being instantiated-- this is a primarily a stylistic preference that asserts constructor methods should not make AJAX calls. 

One workaround is to make a KO component: http://knockoutjs.com/documentation/component-overview.html, but otherwise calling `init` explicitly is the only way for this to work.

rdhyee:
With https://github.com/rdhyee/osf.io/commit/9f42d21a39f086466633216ef24f49b864da72d0, we're now rendering the HTML of a selected note.  e.g., 

image

lbanner:
Nice look'n!

rdhyee:
@samchrisinger I'm trying to style my widget by including [evernote.css](https://github.com/CenterForOpenScience/osf.io/pull/4670/files#diff-0add96cda57e0352cfc0a46901481effR1) and [including require('./evernote.css');](https://github.com/CenterForOpenScience/osf.io/pull/4670/files#diff-ef573e4afca950a037bce81df9719d48R3) -- but I can't get `evernote.css` to load into the page. How am I supposed to properly create and load a CSS file properly for my addon?  (Thanks!)

website/addons/evernote/model.py
```diff
@@ -0,0 +1,127 @@
...
+class EvernoteNodeSettings(StorageAddonBase, AddonOAuthNodeSettingsBase):
+    oauth_provider = Evernote
+    serializer = EvernoteSerializer
+
+    folder_id = fields.StringField(default=None)
+    folder_name = fields.StringField()
+    folder_path = fields.StringField()
+
+    # to hold data about current Notebook
+    _folder_data = None
```
samchrisinger: Not sure if you need this or not, but this field will not persist in the database as is. Changing it to:
```_folder_data = fields.DictionaryField(default=None)```
would fix this.

File: website/addons/evernote/model.py [hunk 5]
```diff
@@ -0,0 +1,127 @@
...
+                return
+
+            self.folder_name = self._folder_data['name']
+            self.folder_path = self._folder_data['name']
+
+            self.save()
+
+    # boilerplate from https://github.com/CenterForOpenScience/osf.io/blob/e4e1bd951b6e79d3bb2335d326bb44d9939bffb8/website/addons/box/model.py#L94-L99
+    @property
+    def complete(self):
```
samchrisinger: The superclass defines this property. Unless you need to override the base case behavior this can be removed.

website/addons/evernote/settings/local-dist.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
```
samchrisinger: There should be default values for these defined in defaults.py to prevent AttributeErrors for developers who have not configured Evernote.

samchrisinger:
@rdhyee with regards to difficulties finding `evernote.css`-- where are you expecting these styles to get loaded? Running the code locally I see the file is indeed loaded on the project settings page -- but the styles aren't used anywhere (Evernote is using the default template for project configuration).

rdhyee:
@samchrisinger About apparently not having `evernote.css` load: I'm hoping to have the `evernote.css` load in the *main project page*. e.g., <http://localhost:5000/t6nus/>.  I made a [simple commit](https://github.com/rdhyee/osf.io/commit/cb952a45f6072ac7833f3cb8611a1973492258eb) to mark the notebook name in red but don't see its effect:

image

In [your note](https://github.com/CenterForOpenScience/osf.io/pull/4670#issuecomment-170577189), you say that `evernote.css` loads in the **project settings page**.    Does `evernote.css` load in the *main project page* for you, in addition to the **project settings page**?  

At the moment, I'm not able to see whether `evernote.css` loads in the project settings page. When I now go to the project settings page (http://localhost:5000/t6nus/settings/), I get an error I had never gotten before: 

image.

The error stems from a [recent update](https://github.com/CenterForOpenScience/osf.io/blame/75af0c21abb9c809b3436fdd388c0c38daf2dae8/website/templates/project/settings.mako#L417) in `osf.io/website/templates/project/settings.mako `:

```JavaScript
window.contextVars.wiki.isEnabled = ${wiki.short_name in addons_enabled | sjson, n };
```

Do you have any hints on how to correct the missing `wiki.short_name` problem with my add-on?  (I may just need to run some migrations or I will clean up my database to start anew.  I'm also wondering whether I need to modify my Evernote add-on code to handle the new updates.

samchrisinger:
@rdhyee working on a fix for the new exception. Not sure what's going on here.

You want to apply styles to the Evernote widget, but are not doing anything to require that css file in the context of that widget. You might trying moving the `require('../evernote.css')` to widget-cfg.js instead.

rdhyee:
@samchrisinger I cleared by database and don't explicitly authorize any addon but still see the same problem with missing `wiki.short_name`.  Maybe I'm still enabling the evernote addon in some local setting somewhere...let me check.

samchrisinger:
Can you please trying clearing your mako template cache: `rm -r /tmp/mako`?

Also restart `invoke assets` and restart `invoke server`

rdhyee:
@samchrisinger  Still same problem.  Here's what I did:

1. I updated to the latest code on `CenterForOpenScience/osf.io`.

1. I removed `evernote` from `addons.json` and `website/settings/local.py` so that I no longer see evernote in the list of available addons

2. I purged the directories and database:
   ```
   find . -name "*.pyc" -delete
   find . -type d -empty -delete
   mongo osf20130903 --eval "db.dropDatabase()"
```

3. I deleted the mako modules dir:  `rm -rf /tmp/mako_modules/`  (I didn't see any `/tmp/mako`)

4. I restarted `invoke assets` and restarted `invoke server`

5. I created my account and a new project in the UI -- but still got the same error.

Anything else to try?  You don't get this error, I presume?

**Comment by samchrisinger** [comment]
It seems you are somehow creating a project without the wiki and that some code that expects the wiki to be there is breaking.

Can you think of any settings that may have accidentally been changed?

A quick fix would be to modify the statement to: 
`wiki and wiki.short_name in addons_enabled`

**Comment by rdhyee** [comment]
@samchrisinger Weird...I'm now running on the `develop` branch and still see the same problem.  Hmm...I see what you mean about my project's not having a wiki.  (I get a page not found error in <http://localhost:5000/9hxvz/wiki>, for example).  I've now gotten rid of all the addons but the problem persists.  I have no idea what setting could be causing this problem.  What file could be relevant? I know of `website/settings/local.py`, which I paste here:

```
# -*- coding: utf-8 -*-
'''Example settings/local.py file.
These settings override what's in website/settings/defaults.py

NOTE: local.py will not be added to source control.
'''

from . import defaults

DEV_MODE = True
DEBUG_MODE = True  # Sets app to debug mode, turns off template caching, etc.

SEARCH_ENGINE = 'elastic'
ELASTIC_TIMEOUT = 10

# Comment out to use SHARE in development
USE_SHARE = False

# Comment out to use celery in development
USE_CELERY = False

# Comment out to use GnuPG in development
USE_GNUPG = False  # Changing this may require you to re-enter encrypted fields

# Email
USE_EMAIL = False
MAIL_SERVER = 'localhost:1025'  # For local testing
MAIL_USERNAME = 'osf-smtp'
MAIL_PASSWORD = 'CHANGEME'

# Mailchimp email subscriptions
ENABLE_EMAIL_SUBSCRIPTIONS = False

# Session
OSF_COOKIE_DOMAIN = None
COOKIE_NAME = 'osf'
SECRET_KEY = "CHANGEME"

# Uncomment if GPG was installed with homebrew
# GNUPG_BINARY = '/usr/local/bin/gpg'

##### Celery #####
## Default RabbitMQ broker
BROKER_URL = 'amqp://'

# Default RabbitMQ backend
CELERY_RESULT_BACKEND = 'amqp://'

USE_CDN_FOR_CLIENT_LIBS = False

# Example of extending default settings
# defaults.IMG_FMTS += ["pdf"]

# Add-ons in local dev

ADDONS_REQUESTED = []
#ADDONS_REQUESTED = ['box', 'dropbox', 'github', 'zotero', 'mendeley']
#ADDONS_REQUESTED += ['evernote']
```

**Comment by rdhyee** [comment]
@samchrisinger Maybe I need to create a new directory in which I start fresh to see whether I still get the same problem.

**Comment by rdhyee** [comment]
@samchrisinger Thanks for helping me with getting `evernote.css` to load now. I haven't been able to figure out why my projects aren't creating a wiki -- so I'm working around that problem for now.

**Comment by rdhyee** [comment]
@samchrisinger I've been making progress on prototyping Evernote functionality using Jupyter notebooks and [ipython/ipywidgets - JavaScript](https://github.com/ipython/ipywidgets).  I'm almost ready to transport that functionality into OSF.  An immediate question: Do you have a JavaScript widget/datagrid  for holding lists of Evernote titles that a user can sort and select on?  Something that can handle potentially thousands of items.  I have some experience using [mleibman/SlickGrid: A lightning fast JavaScript grid/spreadsheet](https://github.com/mleibman/SlickGrid).

**Comment by rdhyee** [comment]
@samchrisinger (Don't know whether you're still the right person to ask the following question) To pick up on what I wrote in https://github.com/CenterForOpenScience/osf.io/pull/4670#issuecomment-176360878: unless there are objections, I'm planning to use datatables.net to hold a scrollable/pagable list of Evernote notes.  My immediate question is how to load the datatables package.  Should I add the dependencies to https://github.com/CenterForOpenScience/osf.io/blob/develop/package.json?  A bigger question: do you see any problem with using datatables?

**Comment by samchrisinger** [comment]
@rdhyee sorry for my slowness in responding. 

I think datatables.net is a fine choice. To add the dependency use `npm install --save datatables.net` (which automatically adds an entry to package.json). You can then `require('datatables.net')` in your JS files.

**Comment by samchrisinger** [comment]
Actually, I just took a closer look at datatables.net. Seems to be a paid product, which is a big red flag for us. 

Can you please look for an open-source unpaid alternative?

@caneruguz do you have thoughts on a good library here?

**Comment by rdhyee** [comment]
@samchrisinger I believe that datatables.net is licensed under an MIT license https://www.datatables.net/faqs/index#Licensing:

> DataTables is free, open source software that you can download and use for whatever purpose you wish, on any and as many sites you want. It is free for you to use! DataTables is available under the MIT, with which you must comply (to do this, basically keep the copyright notices in the software).

**Comment by rdhyee** [comment]
According to https://www.datatables.net/purchase/index, one can support the DataTables project financially by "by purchasing one of the supporter packages"

**Comment by samchrisinger** [comment]
I see, sorry I didn't realized the paid tier was just support access.

In that case, I say go ahead and use the library

**Comment by rdhyee** [comment]
@samchrisinger Thanks for your help -- I now have DataTables partly working now. What I'm now struggling with is how to get the datatables CSS to load. According to https://datatables.net/manual/installation#NPM, I need to do:

```
npm install datatables.net-dt # Styling
```

I then thought I'd be able to include

```
require('datatables.net-dt');
```

in [website/addons/evernote/static/widget-cfg.js](https://github.com/rdhyee/osf.io/blob/evernote-oauth/website/addons/evernote/static/widget-cfg.js) but I got the error

> Module not found: Error: Cannot resolve module 'datatables.net-dt' in /Users/raymondyee/C/src/osf.io/website/addons/evernote/static

According to https://github.com/DataTables/DataTables/issues/434#issuecomment-161278064:  

> The reason for this is that the `datatables.net-dt` package does not contain a Javascript file - it doesn't need one - it contains only CSS (it should actually contain a couple of images as well, that will be corrected in 1.10.11).

I can confirm that after installing  the npm package `datatables.net-dt`, I now have the following CSS file:

osf.io/node_modules/datatables.net-dt/css/jquery.dataTables.css

My question is:  how do I get `jquery.dataTables.css` to load in the context of the evernote `widget-cfg.js`.  (I've been looking for a hint in lines like [osf.io/base-page.js](https://github.com/CenterForOpenScience/osf.io/blob/d334898e9cc3d05f2e544517e1aec865dcea4dfe/website/static/js/pages/base-page.js#L7-L11) -- but I'm unclear, for example, about how to specify the right path for `jquery.dataTables.css`.   Thanks!

**Comment by samchrisinger** [comment]
@rdhyee I had success doing:
`require('datatables.net-dt/css/jquery.dataTables.css')();`

Looking at the package.json for the datatables.net-dt module it's totally unclear to me how that require command could work. Not sure if the library maintainers have any way of tracking issues, but you may consider posting this question there.

**Comment by rdhyee** [comment]
@samchrisinger Thanks for getting me unstuck.  Obviously, I didn't know that `require` takes a path to a specific file -- I thought it could take only the name of an entire package.  BTW, for me, 

```
require('datatables.net-dt/css/jquery.dataTables.css')
```
worked but not

```
require('datatables.net-dt/css/jquery.dataTables.css')();
```

Given [DataTables doesn't work well with Browserify · Issue #434 · DataTables/DataTables](https://github.com/DataTables/DataTables/issues/434#issuecomment-161278064), I suspect that this use (misuse?) of npm packages to hold only a CSS file is **intentional**.

**Comment by samchrisinger** [comment]
@rdhyee yes it seems intentional. It may be worth looking deeper at the disparity between the maintainer's docs and the solution we've come across.

**Comment by rdhyee** [comment]
I posted a question to the `datatables` repo about the best way to load the CSS file: https://github.com/DataTables/DataTables/issues/434#issuecomment-197013164

**Comment by rdhyee** [comment]
@samchrisinger I got a response to my question from the `datatables` folks:  https://github.com/DataTables/DataTables/issues/434#issuecomment-197440810:

> Are you using [Browserify CSS](https://www.npmjs.com/package/browserify-css)?

> The CSS file is in the files array used by the datatables.net-dt package.json file so it can be referenced by just including the package, if the processor can use that, but it doesn't use the main option the the package file to reference the CSS since it isn't an entry point that most would expect (i.e. Javascript).

Should I try

```
npm install --save browserify-css
```
?

**Comment by samchrisinger** [comment]
@rdhyee you can certainly try it, but I'm not sure what the implications will be.

We use webpack and the https://github.com/webpack/style-loader to bundle our CSS. 

One thing you might try is to unininstall the npm version on datatables and use the bower installed version instead (`bower install --save datatables.net`)

**Comment by rdhyee** [comment]
@samchrisinger Any advice on the display and sorting of datetime in osf.io?  My current prototype now loads up created/updated timestamps (in seconds since 1970) -- obviously, not the optimal choice:

osf_datetime_display

I see on osf.io in my logging display that 

> All times displayed at -0700 UTC offset.

e.g.,

> 2016-03-16 11:39 AM

Is that a format I should try? Do you have utilities for the proper handling, formatting, and localization of time for osf.io?  (Thanks!)

**Comment by samchrisinger** [comment]
@rdhyee the logFeed has some examples of datetime formatting:
https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/logFeed.js

**Comment by rdhyee** [comment]
Conceptually, is the general approach for handling datetimes to send to the browser the dates in UTC and then render them in the browser's local timezone?

File: website/addons/evernote/settings/local-dist.py (hunk 6 above)
Comment (rdhyee) [inline website/addons/evernote/settings/local-dist.py]: I think adding 

```
EVERNOTE_CLIENT_ID = None
EVERNOTE_CLIENT_SECRET = None
EVERNOTE_SANDBOX = True
```
to  `website/addons/evernote/settings/defaults.py` satisfies this issue.

**Comment by rdhyee** [comment]
# Functionality for Evernote add-on MVP (round 1)

This PR represents a MVP (minimal viable product) for the evernote add-on.  It allows a user to:

* pick an Evernote notebook
* view notes in a notebook
* filter, sort by title, created, updated and see a selected note rendered in HTML

See evernote add-on

This first pass does not include any functionality for the creation, editing/updating, or deletion of notes -- functionality we can discuss after this PR.

## ENML and CSS

Notes in Evernote are represented in the Evernote Markup Language ([ENML](https://dev.evernote.com/doc/articles/enml.php)).  ENML, specified by a DTD, is a superset of XHTML in which some XHTML elements are prohibited. 

In the MVP, evernote is converted from ENML to XHTML using `ENML_PY`.  Embedded images are rendered with the [Data URI scheme](https://en.wikipedia.org/wiki/Data_URI_scheme): <https://github.com/CenterForOpenScience/osf.io/pull/4670/files#diff-d73811cbfc589b33d077f159111202bcR135>

Note that any CSS in ENML must be inline. (There is no support for internal or external CSS style sheets in ENML). Sometimes the display of the rendered notes pours out beyond the bounding box --> will need help on how to "santize the CSS" -- not easy --- and a problem I see in evernote.com and the desktop client on the Mac itself. 

## writing/editing notes

We need to talk about how to approach the writing and editing of notes.  My ideal path would be to leverage [bordaigorl/sublime-evernote: Open and Save Evernote notes from Sublime Text 3 using Markdown](https://github.com/bordaigorl/sublime-evernote), a Sublime 3 plugin for the editing of Evernote notes.  In other word, write code to inter-converting ENML and markdown, and have users edit Evernote notes as markdown.

I have prototypes of how to edit notes in markdown, using the following libraries:

   * [html2text.py](https://github.com/bordaigorl/sublime-evernote/blob/master/lib/html2text.py), a derivative of the GNU GPL 3 License library originally written by Aaron Schwartz ([aaronsw/html2text: Convert HTML to Markdown-formatted text.](https://github.com/aaronsw/html2text))
   * [markdown2.py](https://github.com/bordaigorl/sublime-evernote/blob/master/lib/markdown2.py) (MIT License)
   * [pygments at master](https://github.com/bordaigorl/sublime-evernote/tree/master/lib/pygments), (BSD License)

A major challenge is how to work with the GPL license of [html2text.py].  Can we incorporate the code into the osf.io code base? If not, are you open to running `html2text` as a separate process to be called from osf.io?   

I've been looking for non-GPL libraries that convert HTML to markdown but haven't found a suitable one yet.  Writing one from scratch is a task I'd like to avoid if possible. (There may be some resuable code in [cs150bf/ever-notedown: Atom Editor Plugin](https://github.com/cs150bf/ever-notedown)). Another approach would be to allow users to edit ENML directly (with some HTML WYSIWYG editor configured to deal with ENML).  **Lots of work to think through here.**





# Notes on the code 

## External libraries in evernote add-on

In osf.io/website/addons/evernote/requirements.txt -> [osf.io/requirements.txt at evernote-oauth · rdhyee/osf.io](https://github.com/rdhyee/osf.io/blob/evernote-oauth/website/addons/evernote/requirements.txt), we have:

* `evernote==1.25.1`:  [evernote 1.25.1 : Python Package Index](https://pypi.python.org/pypi/evernote) and [evernote/evernote-sdk-python: Evernote SDK for Python](https://github.com/evernote/evernote-sdk-python).
* `git+https://github.com/rdhyee/ENML_PY.git@0.0.1`: for converning ENML to HTML -->  [rdhyee/ENML_PY: This is a python library for converting ENML (Evernote Markup Language, http://dev.evernote.com/start/core/enml.php) to/from plain text and HTML.](https://github.com/rdhyee/ENML_PY) is basically an identical fork (with the adding of [0.0.1 tag](https://github.com/rdhyee/ENML_PY/tree/0.0.1)) of [CarlLee/ENML_PY: This is a python library for converting ENML (Evernote Markup Language, http://dev.evernote.com/start/core/enml.php) to/from plain text and HTML.](https://github.com/CarlLee/ENML_PY) license: [MIT License](https://github.com/rdhyee/ENML_PY/blob/master/LICENSE.md)


## Evernote keys

API keys from Evernote are required to run the add-on.  [Evernote Cloud API — Python Quick-start Guide - Evernote Developers](https://dev.evernote.com/doc/start/python.php) is a useful place for Python programmers new to the Evernote API to start with.

To request a key, go to <https://dev.evernote.com/doc/> and click on "GET AN API KEY" button on the upper left.  For the answer to the question "Select the access level that your app will need:", select `Full Access`. Don't check "This integration will sync all user data.".

See [API Key Permissions - Evernote Developers](https://dev.evernote.com/doc/articles/permissions.php) for difference between Full Access vs Basic Access. We need Full Access because the add-on requires `Read existing notes` permission, which is not part of the `Basic Access` regime. 

Set 3 parameters in `website/addons/evernote/settings/local.py`:

* `EVERNOTE_CLIENT_ID` :  Consumer Key
* `EVERNOTE_CLIENT_SECRET`:  Consumer Secret
* `EVERNOTE_SANDBOX`:  `True` to access the Evernote Sandbox; `False` for accessing the production server


We may want to re-examine the request for `Full Access` and instead use [App Notebooks - Evernote Developers](https://dev.evernote.com/doc/articles/app_notebook.php), which is about "Connecting your application to a single notebook.  The latter is not as flexible but requires less trust from users.

## Sandbox vs Production access

Developers can make use of the Evernote Sandbox:  [Testing in the Sandbox - Evernote Developers](https://dev.evernote.com/doc/articles/testing.php).

By default when you request an evernote key, the key is tied to the Evernote sandbox.  You have to request it to be put on production. 
Key I've been using to test my application is

```
EVERNOTE_CLIENT_ID = 'rdhyee-1811'
```

How to get production access for a key? Fill out form at <https://dev.evernote.com/support/glossary.php#k>

```
Name:
Email:
API Consumer Key:
Application Name:
Permissions:
Please explain how your application
uses create, read, and update permissions.

Additional Info:
Cloud Service:
```


# Will do after this PR

- [ ] create a MVP for dmptool add-on based on what I have here -- that will be a chance to make explicit the model upon which I base the add-on --> continue on in a `dmptool-evernote` branch
- [ ] add better error checking and relevant tests for Evernote add-on
- [ ] draft end-user documentation for add-on based on [OSF | Getting Started](https://osf.io/getting-started/#addons)
- [ ] handle project replication for evernote add-on
- [ ] describe more fully how I modeled this add-on on other add-ons (specifically Box) -- set access to a given "folder" -- or in Evernote parlance, Notebook. [I'll be better able to do so when I translate [cos-integration/dmptool.ipynb at master · rdhyee/cos-integration](https://github.com/rdhyee/cos-integration/blob/master/dmptool.ipynb) to a dmptool widget]
- [ ] fix how the MediaStore handles non-image attachments.

## next steps for dmptool add-on

* figure out how to allow user to enter a dmptool key to configure key in settings
* the add-on will list all the plans owned by the user including private ones.  Add links to PDF, docx (or put in iframe)?
* look into how to render a plan simply based on JSON
* write add-on as a waterbutler plugin? (easy way to drop into files?)

**Comment by nicipfeiffer** [comment]
@rdhyee Could we line up a demo with you to see your progress and provide feedback?

**Comment by rdhyee** [comment]
@nicipfeiffer  -- certainly, let's arrange a time for a demo.  I'm on vacation until June 14 -- so let's look at a time that week.

File: package.json [hunk 7]
```diff
@@ -14,6 +14,10 @@
     "bower": "^1.3.12",
     "c3": "^0.4.10",
     "css-loader": "^0.9.1",
+    "datatables-responsive": "^1.0.7",
```
mfraezz: Refactoring this to use a read-only WaterButler provider will likely obviate these packages.

website/addons/evernote/__init__.py
```diff
@@ -0,0 +1,49 @@
+from website.addons.evernote import model, routes
+
+# MUST
+
+SHORT_NAME = 'evernote'
+FULL_NAME = 'Evernote'
+
+ROUTES = [routes.api_routes]
+
+MODELS = []
```
mfraezz: Superfluous definition; you're redefining `MODELS` to what it should be below.

website/addons/evernote/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+
+ADDED_DEFAULT = []
+ADDED_MANDATORY = []
+
+# make the evernote addon show up as a widget like Zotero, Mendeley
+VIEWS = ['widget']
+
+# does this setting make it show up in the Files section?
```
mfraezz: Yes. `HAS_HGRID_FILES` will need to be set to `True` after adding appropriate providers to WB/MFR

mfraezz:
Please add an `/evernote/README.md` file explaining how to get this addon up and running for local development.

website/addons/evernote/model.py
```diff
@@ -0,0 +1,128 @@
...
+           else "https://sandbox.evernote.com"
+
+    auth_url_base = '{}/OAuth.action'.format(BASE_URL)
+    request_token_url = '{}/oauth'.format(BASE_URL)
+    callback_url = '{}/oauth'.format(BASE_URL)
+
+    _oauth_version = OAUTH1
+
+    def handle_callback(self, response):
+        """View called when the Oauth flow is completed. Adds a new BoxUserSettings
```
mfraezz: Please update or remove this docstring

website/addons/evernote/model.py
```diff
@@ -0,0 +1,128 @@
...
+                self._folder_data = utils.get_notebook(client, self.folder_id)
+            except Exception:
+                return
+
+            self.folder_name = self._folder_data['name']
+            self.folder_path = self._folder_data['name']
+
+            self.save()
+
+    # Sam C says following can be removed https://github.com/CenterForOpenScience/osf.io/pull/4670/files#r49327525
```
mfraezz: Sam C is correct

website/addons/evernote/model.py
```diff
@@ -0,0 +1,128 @@
...
+    def _update_folder_data(self):
+        """
+        given the folder_id, compute folder_name, folder_path
+        """
+        if self.folder_id is None:
+            return None
+
+        if not self._folder_data:
+            try:
+                # don't know whether we'll need to do something like this
```
mfraezz: According to the [EverNote docs](https://dev.evernote.com/doc/articles/authentication.php), it looks like you won't be able to:

>Each access token has an expiration date. By default, the duration of access token validity is 1 year from the date of issue. The user can alter this duration to 1 day, 1 week or 1 month. In all these cases (including tokens valid for 1 year), the expiration date will be included as the parameter edam_expires. The value of this parameter will be a standard UNIX timestamp in units of milliseconds indicating the date and time of expiration.

The don't return a refresh token for API applications, and [`UserStore.refreshAuthentication` is only available for internal applications](https://dev.evernote.com/doc/reference/UserStore.html#Fn_UserStore_refreshAuthentication)

website/addons/evernote/model.py
```diff
@@ -0,0 +1,128 @@
...
+        self.user_settings = user_settings
+        #nodelogger = BoxNodeLogger(node=self.owner, auth=Auth(user_settings.owner))
+        #nodelogger.log(action="node_authorized", save=True)
+
+    def set_folder(self, folder_id, auth):
+        self.folder_id = str(folder_id)
+        self._update_folder_data()
+        self.save()
+
+        if not self.complete:
```
mfraezz: This conditional doesn't need to exist. It's copied from an addon that uses folder-level OAuth grants, which EverNote doesn't need. When you're using the `generic_views`, `set_auth` will be called to give a node-level grant.

website/addons/evernote/model.py
```diff
@@ -0,0 +1,128 @@
...
+            self.user_settings.grant_oauth_access(
+                node=self.owner,
+                external_account=self.external_account,
+                metadata={'folder': self.folder_id}
+            )
+            self.user_settings.save()
+
+        # Add log to node
+        #nodelogger = BoxNodeLogger(node=self.owner, auth=auth)
```
mfraezz: Consider using `self.nodelogger.log` -- anything that inherits from `AddonOAuthNodeSettingsBase` now has that (you may need to pull recent changes, it looks like this hasn't been updated in a while).

website/addons/evernote/model.py
```diff
@@ -0,0 +1,128 @@
...
+        if not self._folder_data:
+            try:
+                # don't know whether we'll need to do something like this
+                #refresh_oauth_key(self.external_account)
+                client = utils.get_evernote_client(self.external_account.oauth_key)
+                self._folder_data = utils.get_notebook(client, self.folder_id)
+            except Exception:
+                return
+
+            self.folder_name = self._folder_data['name']
```
mfraezz: You can move this and the calls that hit EverNote's API into `set_folder`, and remove this method.

Also, are you not already getting this information to populate the list of folders (NoteBooks) in the first place? Can you avoid hitting their API multiple times for such similar requests?

website/addons/evernote/model.py
```diff
@@ -0,0 +1,128 @@
...
+    # @property
+    # def complete(self):
+    #     return bool(self.has_auth and self.user_settings.verify_oauth_access(
+    #         node=self.owner,
+    #         external_account=self.external_account,
+    #     ))
+
+    # based on https://github.com/CenterForOpenScience/osf.io/blob/4a5d4e5a887c944174694300c42b399638184722/website/addons/box/model.py#L105-L107
+    def fetch_full_folder_path(self):
+        # don't know why this would be needed for Evernote
```
mfraezz: Given that EverNote won't let you refresh OAuth tokens, you don't need this.

website/addons/evernote/serializer.py
```diff
@@ -0,0 +1,72 @@
+from website.addons.base.serializer import OAuthAddonSerializer
+from website.util import api_url_for, web_url_for
+
+class EvernoteSerializer(OAuthAddonSerializer):
```
mfraezz: While not *technically* a "storage addon," this should probably be inheriting from `StorageAddonSerializer`, as `serialize_settings` is doing the same thing in both. You'll need to implement a `credentials_are_valid(self, user_settings, client)` method that checks to see if API calls will still work, and a `serialized_folder(self, node_settings)` that returns a JSON-formatted representation of the selected "NoteBook", but it should end up being simpler than what you're currently doing.

website/addons/evernote/serializer.py
```diff
@@ -0,0 +1,72 @@
...
+
+        return {
+            'auth': api_url_for('oauth_connect',
+                                service_name='evernote'),
+            'importAuth': node.api_url_for('evernote_add_user_auth'),
+            'folders': node.api_url_for('evernote_folder_list'),
+            'config': node.api_url_for('evernote_set_config'),
+            'notes': node.api_url_for('evernote_notes'),
+            'note': node.api_url_for('evernote_note'),
+            # 'files': node.web_url_for('collect_file_trees'),
```
mfraezz: I'm not yet sure what the distinction between `note` and `notes` is, but I have a feeling one of them will need to become `files`. This is used to build the UI

website/addons/evernote/serializer.py
```diff
@@ -0,0 +1,72 @@
...
+            #'share': 'https://app.box.com/files/0/f/{0}'.format(self.node_settings.folder_id),
```
mfraezz: `share` can be removed

website/addons/evernote/requirements.txt
```diff
@@ -0,0 +1,2 @@
+evernote==1.25.1
+git+https://github.com/rdhyee/ENML_PY.git@0.0.1
```
mfraezz: You shouldn't need this library -- the MFR will be doing this work.

Also, while I know that [ENML prohibits many "dangerous" tags](https://dev.evernote.com/doc/articles/enml.php#prohibited) by design, this still seems like an attack vector, and I'd prefer not using this library in the MFR either. @felliott might have opinions about this too.

website/addons/evernote/static/evernoteLogActionList.json
```diff
@@ -0,0 +1,6 @@
+{
```
mfraezz: Please also add an `evernoteAnonymousLogActionList.json` that contains anonymized language for these logs, e.g. `"A user linked an Evernote NoteBook to a node"`. This hasn't gone in yet, but it will soon. See https://github.com/CenterForOpenScience/osf.io/pull/5825/ for reference

website/addons/evernote/static/evernoteLogActionList.json
```diff
@@ -0,0 +1,6 @@
...
+  "evernote_folder_selected" : "${user} linked evernote folder ${folder} in ${node}",
```
mfraezz: Please use `ADDON_FULL_NAME` casing, i.e. `Evernote`.

Also, please use addon-specific language for things that will be displayed to the user --  in this case, 'folders' should be called 'notebooks'

website/addons/evernote/static/widget-cfg.js
```diff
@@ -0,0 +1,117 @@
+'use strict';
```
mfraezz: With the MFR/WB refactor, this file will likely be removed

mfraezz: ...and likely replaced with an `evernoteFangornConfig.js` file that does a few overrides to make this behave as read-only.

mfraezz:
Please remove the blank file `website/addons/evernote/templates/evernote_user_settings.mako`

website/addons/evernote/templates/evernote_widget.mako
```diff
@@ -0,0 +1,12 @@
+<%inherit file="project/addon/widget.mako"/>
```
mfraezz: This will likely be removed with the WB/MFR refactor

website/addons/evernote/templates/log_templates.mako
```diff
@@ -0,0 +1,25 @@
+<script type="text/html" id="evernote_folder_selected">
```
mfraezz: Depending on timing with #5825, this file might not need to exist. We're moving to a more generic system to display NodeLogs, and a clean-up PR will be going in after that to remove files like these. This isn't a blocker, though

mfraezz:
Switching to use `generic_views` will give you a lot of generic tests for free. You may need to do some overrides because this is read-only, but it should still be significantly easier than writing tests for all this. For reference, see `website/addons/base/testing/*.py`, and how they're used by other addons.

website/addons/evernote/views.py
```diff
@@ -0,0 +1,215 @@
...
+
+    token = node_addon.external_account.oauth_key
+    client = utils.get_evernote_client(token)
+    note = utils.get_note(client, noteid,
+        withContent=True,
+        withResourcesData=False)
+
+    # convert to HTML
+    mediaStore = utils.MyMediaStore(client.get_note_store(), note.guid)
+    html = ENML2HTML.ENMLToHTML(note.content, pretty=True, header=False,
```
mfraezz: The MFR should be doing this conversion, this method can probably go away

website/addons/evernote/views.py
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+
+evernote_deauthorize_node = generic_views.deauthorize_node(
+    SHORT_NAME
+)
+
+@must_have_addon('evernote', 'node')
+@must_be_addon_authorizer('evernote')
+def evernote_notes(node_addon, **kwargs):
```
mfraezz: This functionality will likely be moved into WaterButler -- it appears to be retrieving metadata about notes, preparing them for display.

website/addons/evernote/views.py
```diff
@@ -0,0 +1,215 @@
...
+    html = ENML2HTML.ENMLToHTML(note.content, pretty=True, header=False,
+          media_store=mediaStore)
+
+    return {'noteid': noteid,
+            'content': note.content,
+            'html': html}
+
+@must_have_addon('evernote', 'node')
+@must_be_addon_authorizer('evernote')
+def evernote_folder_list(node_addon, **kwargs):
```
mfraezz: Given that Evernote doesn't have allow nesting with their Notebooks, you can probably remove the `None` case, and do a few overrides in `node-cfg.js` to hide the `+` icons for folder expanding. Alternatively, you could do something like what Dataverse does to select which Dataset should be connected.

mfraezz: GitHub also avoids using FolderPicker, and might be another good reference

website/addons/evernote/views.py
```diff
@@ -0,0 +1,215 @@
...
+import ENML2HTML
+
+import logging
+logger = logging.getLogger(__name__)
+
+SHORT_NAME = 'evernote'
+FULL_NAME = 'Evernote'
+
+@must_be_logged_in
+def evernote_get_user_settings(auth):
```
mfraezz: Many of these view functions can be replaced with `generic_views`. This should be replaced with `generic_views.account_list`

website/addons/evernote/views.py
```diff
@@ -0,0 +1,215 @@
...
+    """ Returns the list of all of the current user's authorized Evernote accounts """
+    serializer = EvernoteSerializer(user_settings=auth.user.get_addon('evernote'))
+    return serializer.serialized_user_settings
+
+
+@must_have_addon('evernote', 'node')
+@must_have_permission(permissions.WRITE)
+def evernote_get_config(node_addon, auth, **kwargs):
```
mfraezz: This can be replaced with `generic_views.get_config`

website/addons/evernote/views.py
```diff
@@ -0,0 +1,215 @@
...
+    return {
+        'result': EvernoteSerializer().serialize_settings(node_addon, auth.user),
+    }
+
+@must_not_be_registration
+@must_have_addon('evernote', 'user')
+@must_have_addon('evernote', 'node')
+@must_be_addon_authorizer('evernote')
+@must_have_permission(permissions.WRITE)
+def evernote_set_config(node_addon, user_addon, auth, **kwargs):
```
mfraezz: This should be replaced with `generic_views.set_config`

website/addons/evernote/views.py
```diff
@@ -0,0 +1,215 @@
...
+    """View for changing a node's linked Evernote folder."""
+    folder = request.json.get('selected')
+    serializer = EvernoteSerializer(node_settings=node_addon)
+
+    uid = folder['id']
+    path = folder['path']
+
+    node_addon.set_folder(uid, auth=auth)
```
mfraezz: `node_addon.set_folder` should be modified to accept the JSON dict containing the `id` and `path`, which should allow you to remove the extra API calls I referenced in [another comment](https://github.com/CenterForOpenScience/osf.io/pull/4670/files#r67699265)

website/addons/evernote/views.py
```diff
@@ -0,0 +1,215 @@
...
+            'urls': serializer.addon_serialized_urls,
+        },
+        'message': 'Successfully updated settings.',
+    }
+
+
+@must_have_addon('evernote', 'user')
+@must_have_addon('evernote', 'node')
+@must_have_permission(permissions.WRITE)
+def evernote_add_user_auth(auth, node_addon, user_addon, **kwargs):
```
mfraezz: This can be replaced with `generic_views.import_auth`

website/addons/evernote/views.py
```diff
@@ -0,0 +1,215 @@
...
+                'folders': node.api_url_for('evernote_folder_list', notebookId=notebook['guid']),
+            }
+
+        } for notebook in notebooks]
+    else:
+        return []
+
+@must_be_contributor_or_public
+@must_have_addon('evernote', 'node')
+def evernote_widget(node_addon, **kwargs):
```
mfraezz: This widget will likely be going away, but possibly replaced with a settings widget to select the Notebook. It should be possible to use FolderPicker for that, but it might be simpler to use a widget.

website/oauth/models/__init__.py
```diff
@@ -170,6 +170,9 @@ def auth_url(self):
             oauth = OAuth1Session(
                 client_key=self.client_id,
                 client_secret=self.client_secret,
+                callback_uri=web_url_for('oauth_callback',
```
mfraezz: Can you verify 

1. that Evernote doesn't work without this (which seems probable, given how OAuth is supposed to work) 
2. that this doesn't break Zotero, the other OAuth1 provider?

website/templates/project/settings.mako
```diff
@@ -469,7 +469,8 @@
       window.contextVars.node.nodeType = ${ node['node_type'] | sjson, n };
       window.contextVars.nodeCategories = ${ categories | sjson, n };
       window.contextVars.wiki = window.contextVars.wiki || {};
-      window.contextVars.wiki.isEnabled = ${wiki.short_name in addons_enabled | sjson, n };
+      ##window.contextVars.wiki.isEnabled = ${wiki and wiki.short_name in addons_enabled | sjson, n };
```
mfraezz: Why was this changed?

website/addons/evernote/model.py
```diff
@@ -0,0 +1,128 @@
...
+            'display_name': user.name,
+            'profile_url': ''
+        }
+
+class EvernoteUserSettings(AddonOAuthUserSettingsBase):
+    oauth_provider = Evernote
+    serializer = EvernoteSerializer
+
+
+class EvernoteNodeSettings(StorageAddonBase, AddonOAuthNodeSettingsBase):
```
mfraezz: This class will need 

* an `api` property that returns a possibly authenticated `Evernote(ExternalProvider)` session
* a `deauthorize` method that logs the action and calls `clear_settings` (also required addition) and `clear_auth` (`super` version is fine)
* a `clear_settings` method that unsets 'folder' related fields
* a `serialize_waterbutler_credentials` method that prepares auth credentials to be passed to WB
* a `serialize_waterbutler_settings` method that tells WB what folder (Notebook) to look for files (Notes) in

It may also require a `create_waterbutler_log` eventually -- when write functionality is added -- but it shouldn't need it now.

website/addons/evernote/model.py
mfraezz: I think this field is unnecessary.

website/addons/evernote/utils.py
```diff
@@ -0,0 +1,136 @@
+
```
mfraezz: Much of this file will likely be removed after refactor, I'll review anything that remains after.

Also, this file should either start with an `import`, by specifying the coding (i.e. `# -*- coding: utf-8 -*-
`), or with module-level comments/docstrings, as per PEP8

mfraezz:
When switching to use the `generic_views`, please also make testing classes that inherit from the mixins found in `/website/addons/base/testing/` -- these should give you most of the tests you need for free, though you may have to override a few of them.

Please pull from `develop`, this has merge conflicts now. It's not unlikely that some of the seemingly unrelated issues (like the weird wiki behavior noted in comments from several months ago) will be resolved by this.

I'll try to be on Slack to answer any questions, but feel free to `@tag` me here either.

Pass complete 🐧

website/templates/project/settings.mako
rdhyee: For whatever reason, without this change, I've never been able to get my local setup working. I don't know why the `wiki` is missing for me for any given project.  

```
AttributeError: 'Undefined' object has no attribute 'short_name'
```

Any suggestions on what I might be configuring incorrectly on my machine?

website/addons/evernote/model.py
rdhyee: Thanks for outlining how I should rewrite `EvernoteNodeSettings`. I'm guessing [`BoxNodeSettings`](https://github.com/CenterForOpenScience/osf.io/blob/48b4e86c41dbfbd0751407a120b0a4e49158d0a7/website/addons/box/model.py#L62-L209) is a good to study.

website/addons/evernote/model.py
rdhyee: Are you sure this line is unnecessary? Looking at the [box add-on code](https://github.com/CenterForOpenScience/osf.io/blob/095b16bccf4d929903555a5234bc1321d85621ba/website/addons/box/model.py#L89), I see:

```Python
_folder_data = None
```

I was mimicking this code and thinking that if this line is need for `box`, it would be needed for `evernote`.

website/addons/evernote/model.py
```diff
@@ -0,0 +1,128 @@
...
+from website.oauth.models import (ExternalProvider, OAUTH1)
+
+from modularodm import fields
+
+import logging
+logger = logging.getLogger(__name__)
+
+class Evernote(ExternalProvider):
+    """
+    First cut at the Evernote provider
```
rdhyee: Done

website/addons/evernote/model.py
```diff
@@ -0,0 +1,128 @@
...
+    short_name = 'evernote'
+
+    # Evernote has a sandbox and a production service
+    # with different endpoints
+
+    client_id = settings.EVERNOTE_CLIENT_ID
+    client_secret = settings.EVERNOTE_CLIENT_SECRET
+    _sandbox = settings.EVERNOTE_SANDBOX
+
+    BASE_URL = "https://www.evernote.com" if not _sandbox  \
```
rdhyee: Done

website/addons/evernote/model.py
rdhyee: Done

website/addons/evernote/model.py
rdhyee: I've removed this section.

website/addons/evernote/model.py
rdhyee: I've removed `_update_folder_data`.

website/addons/evernote/model.py
rdhyee: Addressed.

website/addons/evernote/model.py
rdhyee: Done

website/addons/evernote/model.py
rdhyee: I'll keep in mind this issue of minimizing hits to the Evernote API.

website/addons/evernote/model.py
rdhyee: Done

website/addons/evernote/requirements.txt
rdhyee: `ENML_PY` is used to convert ENML to HTML. So I'm confused by your comment:

> You shouldn't need this library -- the MFR will be doing this work.

website/addons/evernote/serializer.py
rdhyee: I've moved to using `StorageAddonSerializer`.

website/addons/evernote/serializer.py
rdhyee: done

website/addons/evernote/static/evernoteLogActionList.json
rdhyee: Added the file https://github.com/rdhyee/osf.io/commit/48eb9f336042bbb41124827d87aa4f8bcd138996

website/addons/evernote/static/evernoteLogActionList.json
rdhyee: done

rdhyee:
I'm trying to figure out the minimal code to get something to show up in the File page for the Evernote add-on. Right now, in the files page for a project (e.g., <http://localhost:5000/nedj5/files/>), "Evernote is currently unavailable".  Here are I currently have:

[[WIP] Evernote addon: skeleton with OAuth authorization and notebook selection by rdhyee · Pull Request #4670 · CenterForOpenScience/osf.io](https://github.com/CenterForOpenScience/osf.io/pull/4670/files#diff-c2795ade1614fa639fc1dd81d381b686R17):

> `HAS_HGRID_FILES = True`

[[WIP] Evernote addon: skeleton with OAuth authorization and notebook selection by rdhyee · Pull Request #4670 · CenterForOpenScience/osf.io](https://github.com/CenterForOpenScience/osf.io/pull/4670/files#diff-c2795ade1614fa639fc1dd81d381b686R18):

> `GET_HGRID_DATA = views.evernote_root_folder`

[[WIP] Evernote addon: skeleton with OAuth authorization and notebook selection by rdhyee · Pull Request #4670 · CenterForOpenScience/osf.io](https://github.com/CenterForOpenScience/osf.io/pull/4670/files#diff-fcb727fc5a1fbaaae766737e87af7be1R217):

```
 evernote_root_folder = generic_views.root_folder(
     SHORT_NAME
)
```

What else do I need?  Do I actually need a skeletal `WB/MFR` provider for Evernote to be in place for the "Evernote is currently unavailable" message to disappear?

rdhyee:
This PR closed, to be replaced by the PRs for `feature/evernote`:

* [Feature/evernote add-on by rdhyee · Pull Request #7043 · CenterForOpenScience/osf.io](https://github.com/CenterForOpenScience/osf.io/pull/7043)
* [Feature/evernote WB provider by rdhyee · Pull Request #206 · CenterForOpenScience/waterbutler](https://github.com/CenterForOpenScience/waterbutler/pull/206)
* [Feature/evernote (ENML) by rdhyee · Pull Request #249 · CenterForOpenScience/modular-file-renderer](https://github.com/CenterForOpenScience/modular-file-renderer/pull/249)

PR #4501, opened 2015-10-22
OSF-3425 Prevent line breaks in file extension

ag4ml:
# Purpose

Currently, if the window is resized on a file page, there is a possibility of an "awkward" line break with the file name. This means that part of the file extension could be split and a .docx could be interpreted as a .doc.

b92f05f0-1e76-11e5-8495-087cf0d98fb0

# Changes

* Add a new CSS rule to the file-view-page.css file using the "file-ext" id
* Modify the view_file.mako file so that it splits the file name into two parts: with and without extension
* Apply the CSS id to the extension part

PR #4671, opened 2015-12-01
[IE Fix] OSF-3425 Prevent line breaks in file extension

ag4ml:
# Fix  

The previous pull request (https://github.com/CenterForOpenScience/osf.io/pull/4501) was working on all browsers except for IE. This pull request is an attempt to fix the incompatibility issue with IE. The issue occurs because two CSS rules are conflicting in IE. To fix this, I changed the value of one of the rules. 
http://stackoverflow.com/questions/18129005/white-space-nowrap-is-not-working-in-ie-in-a-horizontally-scrolling-box

----------------------------------------------------------------------------------------------------------------

# Purpose

Currently, if the window is resized on a file page, there is a possibility of an "awkward" line break with the file name. This means that part of the file extension could be split and a .docx could be interpreted as a .doc.

b92f05f0-1e76-11e5-8495-087cf0d98fb0

# Changes

* Add a new CSS rule to the file-view-page.css file using the "file-ext" id
* Modify the view_file.mako file so that it splits the file name into two parts: with and without extension
* Apply the CSS id to the extension part


sloria:
I'm not sure this is the right approach. If you only want to apply `nowrap` to the file extension and `.break-word` to the file name, why not surround those parts in a span and apply the proper classes rather than adding new CSS rules?

Something like:

```
    <h2>
      ## Split file name into two parts: with and without extension
        <span class="break-word">
            ${file_name_title | h}
        </span>
        <span id="file-ext">${file_name_ext | h}</span>
    </h2>
```

sloria:
Pass finished.

sloria:
Merged into release/0.61.0

PR #4676, opened 2015-12-02
Changes made in Css style to prevent overflow of lengthy tags

arpitar:
Added break-all property to enable lengthy tags to break in order to fit the div and not overflow


mfraezz:
Looks good to me. No bugs found. :star2: 

Tested and confirmed that it works on Firefox and IE:

screen shot 2015-12-04 at 13 18 46
screen shot 2015-12-04 at 13 19 18

PR #4677, opened 2015-12-02
[Ready to Review] Bulk GET special case filtering on ID

pattisdr:
# Purpose

Issue https://openscience.atlassian.net/browse/OSF-5041
Allows for the equivalent of a bulk GET request through filtering (for example, if I want to request two nodes, specifically).  The reason we are not doing bulk GET requests is that putting JSON in the request body of a GET request is discouraged.  So for bulk GET capacity, we are implementing filtering on multiple ids.

# Changes
Allows return of multiple resources via a comma-separated list of values in the query_param.

For example, the following two node id's are requested in the following URL:
`/v2/nodes/?filter[id]=3z4tf,vatpw`

This turns into the query, `'_id', 'in', ['3z4tf, 'vatpw']`

The response will contain representations of the two nodes, `3z4tf` and `vatpw`.

**Example:**

url = `v2/nodes/?filter[id]=69wkq, y3mxq, aegu6,12345`

The first three are nodes, the fourth node DNE.

Response (most serializer fields removed for conciseness):

The three nodes that exist (and the user has permissions for) are returned.

screen shot 2015-12-02 at 3 09 36 pm

:star: Also, this filtering on multiple resources extends to other fields and endpoints.  

so for example, you could filter on multiple user fullnames
   `v2/users/?filter[family_name]=Pattison,Jackson`
Will return all users with either a surname of Pattison or Jackson

Or, you could filter of multiple log actions:
   `/v2/nodes/x2qrm/logs/?filter[action]=comment_added,project_created`
Will return all logs on node `x2qrm` with action of `comment_added` or `project_created`.

### DIFF
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -155,8 +155,12 @@ class FilterMixin(object):
         If url contained `/nodes/?filter[id]=12345, abcde`, the returned values would be:
         [u'12345', u'abcde']
         """
+        if value[0] == '[' and value[-1] == ']':
+            value = value.replace('[', '').replace(']', '')
         separated_values = value.split(',')
-        cleaned_values = [re.sub(r"[^\w\s]+]", '', val).strip() for val in separated_values]
+        cleaned_values = []
+        for val in separated_values:
+            cleaned_values.append(val.strip())
         values = [self.convert_value(val, field) for val in cleaned_values]
         return values
 
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -188,7 +188,7 @@ class FilterMixin(object):
                 # Special case date(time)s to allow for ambiguous date matches
                 if isinstance(field, self.DATE_FIELDS):
                     query[field_name].extend(self._parse_date_param(field, field_name, op, value))
-                elif not isinstance(value, int) and len(value.split(',')) > 1:
+                elif not isinstance(value, int) and field_name == '_id':
                     query[field_name].append({
                         'op': 'in',
                         'value': self.bulk_get_values(value, field)
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -148,6 +148,18 @@ class FilterMixin(object):
                 'value': stop
             }]
 
+    def bulk_get_values(self, value, field):
+        """
+        Returns list of values from query_param for IN query
+
+        If url contained `/nodes/?filter[id]=12345, abcde`, the returned values would be:
+        [u'12345', u'abcde']
+        """
+        separated_values = value.split(',')
+        cleaned_values = [re.sub(re.compile(r'\W'), '', val) for val in separated_values]
+        values = [self.convert_value(val, field) for val in cleaned_values]
+        return values
+
     def parse_query_params(self, query_params):
         """Maps query params to a dict useable for filtering
         :param dict query_params:
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -158,10 +158,7 @@ class FilterMixin(object):
         if value[0] == '[' and value[-1] == ']':
             value = value.replace('[', '').replace(']', '')
         separated_values = value.split(',')
-        cleaned_values = []
-        for val in separated_values:
-            cleaned_values.append(val.strip())
-        values = [self.convert_value(val, field) for val in cleaned_values]
+        values = [self.convert_value(val.strip(), field) for val in separated_values]
         return values
 
     def parse_query_params(self, query_params):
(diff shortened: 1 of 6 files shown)


api/base/filters.py
```diff
@@ -177,10 +190,16 @@ def parse_query_params(self, query_params):
                 if isinstance(field, self.DATE_FIELDS):
                     query[field_name].extend(self._parse_date_param(field, field_name, op, value))
                 else:
```
mfraezz: You can probably remove this `else` and turn the next `if` into an `elif`

api/base/filters.py
```diff
@@ -148,6 +148,19 @@ def _parse_date_param(self, field, field_name, op, value):
...
 
+    def bulk_get_values(self, value, field):
+        """
+        Returns list of values from query_param for IN query
+
+        If url contained `/nodes/?filter[id]=12345, abcde`, the returned values would be:
+        [u'12345', u'abcde']
+        """
+        if value[0] == '[' and value[-1] == ']':
+            value = value.replace('[', '').replace(']', '')
```
mfraezz: If `value` will never contain additional brackets (and this might already break things if it does), these two lines could be replaced with `value = value.lstrip('[').rstrip(']')`, which would have no effect if the conditions in the above `if` statement weren't met.

api/base/serializers.py
```diff
@@ -99,11 +99,12 @@ def __init__(self, **kwargs):
 
     # Overrides CharField
     def to_internal_value(self, data):
-        request = self.context['request']
-        if request.method in utils.UPDATE_METHODS and not utils.is_bulk_request(request):
-            id_field = getattr(self.root.instance, self.source, '_id')
-            if id_field != data:
-                raise Conflict()
+        request = self.context.get('request')
```
mfraezz: Under what conditions would `context` not have a `request` here?

mfraezz: NVM, figured it out.

mfraezz:

Tests are great, and manual testing worked perfectly. A few small suggestions, but this looks really good.

api/base/filters.py
pattisdr: This is better.  I don't want the user to put brackets around the ids and the documentation says to put `filter[id]=12345,abcde`, but I think it's likely someone tries it with brackets.   I wanted to handle this case, but I also don't want brackets to be stripped from a title with brackets, for example.

api/base/filters.py
```diff
@@ -148,6 +148,19 @@ def _parse_date_param(self, field, field_name, op, value):
...
             }]
 
+    def bulk_get_values(self, value, field):
+        """
+        Returns list of values from query_param for IN query
+
+        If url contained `/nodes/?filter[id]=12345, abcde`, the returned values would be:
+        [u'12345', u'abcde']
+        """
+        if value[0] == '[' and value[-1] == ']':
```
mfraezz: Superfluous `if` statement, `lstrip` and `rstrip` will do nothing if the characters don't match.

```python
In [7]: s
Out[7]: '[abcde,fghij,klmno]'

In [8]: s = s.lstrip('[').rstrip(']')

In [9]: s
Out[9]: 'abcde,fghij,klmno'

In [10]: s = s.lstrip('[').rstrip(']')

In [11]: s
Out[11]: 'abcde,fghij,klmno'
```

api/base/filters.py
```diff
@@ -176,6 +189,11 @@ def parse_query_params(self, query_params):
                 # Special case date(time)s to allow for ambiguous date matches
                 if isinstance(field, self.DATE_FIELDS):
                     query[field_name].extend(self._parse_date_param(field, field_name, op, value))
+                elif not isinstance(value, int) and len(value.split(',')) > 1:
```
mfraezz: Should this be checking `isinstance(field, IDField)` instead? If we only want to special case bulk `GET`s on IDs, that seems like a cleaner way of doing it.

pattisdr: Well my original intent for this PR was allowing you to bulk `GET` any filterable field, including title, description, etc.  But after just discussing with @brianjgeiger, I will restrict this functionality to just the id's.  He is right that more thought will be needed to filter on some of these other fields.

pattisdr: Since not all `id` fields are instances of IDField, I will am looking for the field_name to equal '_id'

api/base/serializers.py
```diff
@@ -99,11 +99,12 @@ def __init__(self, **kwargs):
...
+        if request:
+            if request.method in utils.UPDATE_METHODS and not utils.is_bulk_request(request):
+                id_field = getattr(self.root.instance, self.source, '_id')
```
sloria: I don't understand why the default is `_id` here. Isn't `data` going to be an ID, not a field name? If that's the case, we're comparing a GUID to `_id`, which doesn't seem right.

That said, I see that this isn't new code, so we can leave it alone. I would still like to understand this logic, though.

sloria:
Looks good!

PR #4681, opened 2015-12-02
[CR] OSF addon for Microsoft OneDrive

caseyrygt:
# Initial Work In Progress PR for Microsoft OneDrive OSF addon
## Notes

We copied Box as the starting point for this addon. There are debug and info logging statements that we'll remove before the final PR.

Also see the corresponding [PR for WaterButler.](https://github.com/CenterForOpenScience/waterbutler/pull/102)
## Limitations
- This only works with OneDrive for personal use. OneDrive for business has a different authentication endpoints and the API is slightlightly different at this time. Additionally, Microsoft has released graph which, I believe, should (when it is complete) unify OneDrive for Personal and OneDrive for Business under one API. The OneDrive API used here should be compatible with Graph based on the documentation I've seen.
- OneDrive's API only returns the most recent revision for an item (file or folder). I don't believe there is a way around this now.
## Issues under development
- Delete Folder
- Download Zip
- Rename File
- Rename Folder
- Copy Folder
- Copy File
## Functionality in Place (i.e. "dev complete")
- Upload files
- Create folders
- Download Files
- Delete files
- View Revisions
- List Files/Folders in a folder
## Screenshots

file-detail
file-list
file-revisions

Note: I don't have MFR running.  I'm having issues getting it running in ubuntu.


lbanner:
@samchrisinger ... when you get a window of time ... @caseyrygt is ready for some feedback on this add-on PR. Thanks!

website/addons/onedrive/client.py
```diff
@@ -0,0 +1,95 @@
...
+from framework.exceptions import HTTPError
+
+from website.util.client import BaseClient
+#from website.addons.googledrive import settings
+from website.addons.googledrive import exceptions
+from website.addons.onedrive import settings
+
+logger = logging.getLogger(__name__)
+
+logging.getLogger('onedrive1').setLevel(logging.WARNING)
```
samchrisinger: Is there a reason not to use `logger` as defined above?

website/addons/onedrive/settings/local-dist.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
+"""Example OneDrive local settings file. Copy this file to local.py and change
+these settings.
+"""
+# Get an app key and secret at https://account.live.com/developers/applications
+ONEDRIVE_KEY = '000000004416C3D3'
```
samchrisinger: Please remove these

felliott:
I'm not an addon expert by any means, so I don't know if this is the right way to solve this, but I couldn't connect my account until I uncommented the `auth` attribute in the return value of `addon_serialized_urls()` in serializer.py.

website/addons/onedrive/README.md
```diff
@@ -0,0 +1,8 @@
+# OSF OneDrive Addon
+
+Enabling the addon for development
+
+1. In `website/settings/local.py` add, `"onedrive"` to the `ADDONS_REQUESTED` list.
```
samchrisinger: Looks like you've added `onedrive` to addons.json, which should take care of this step automatically.

website/addons/onedrive/serializer.py
```diff
@@ -0,0 +1,105 @@
...
+        ret.update({'settings': web_url_for('user_addons')})
+        return ret
+
+    @property
+    def addon_serialized_urls(self):
+        logger.error('addon_serialized_urls-1')
+        node = self.node_settings.owner
+        
+        return {
+            #'auth': api_url_for('oauth_connect',
```
samchrisinger: This should probably be uncommented.

website/addons/onedrive/serializer.py
```diff
@@ -0,0 +1,105 @@
...
+             #                   service_name='onedrive'),
+            'importAuth': node.api_url_for('onedrive_add_user_auth'),
+            'files': node.web_url_for('collect_file_trees'),
+            'folders': node.api_url_for('onedrive_folder_list'),
+            'config': node.api_url_for('onedrive_set_config'),
+            #'emails': node.api_url_for('onedrive_get_share_emails'),
```
samchrisinger: You can delete these two commented-out urls. These refer to old features that have been phased on in recent iterations.

website/addons/onedrive/serializer.py
```diff
@@ -0,0 +1,105 @@
...
+        
+        logger.error('addon_serialized_settings-1')
+        #TODO: review onedrive serilaized settings to determine if we need to add more
+        valid_credentials = True
+        user_settings = node_settings.user_settings
+        self.node_settings = node_settings
+        current_user_settings = current_user.get_addon('onedrive')
+        user_is_owner = user_settings is not None and user_settings.owner == current_user
+
+#        if user_settings:
```
samchrisinger: This block is used to check if the current credentials are still valid, which can be useful for error handling if the user's credentials have expired or have been invalidated for some reason.

The idea is to make an arbitrary trivial request to the third party API, using a try/except to see if there are errors (thus indicating the credentials are invalid). Not a perfect pattern by any means, but something worth considering.

website/addons/onedrive/serializer.py
```diff
@@ -0,0 +1,105 @@
...
+from website.addons.base.serializer import OAuthAddonSerializer
+
+from website.util import api_url_for, web_url_for
+
+logger = logging.getLogger(__name__)
+
+logging.getLogger('onedrive1').setLevel(logging.WARNING)
+
+
+class OneDriveSerializer(OAuthAddonSerializer):
```
samchrisinger: I think you will need to add some more fields to this class in order for it to be instantiated. See the `abc.abstractproperty`s defined in https://github.com/CenterForOpenScience/osf.io/blob/develop/website/addons/base/serializer.py#L15. 

E.g.
`addon_short_name` = 'onedrive'

website/addons/onedrive/model.py
```diff
@@ -0,0 +1,258 @@
...
+
+    _auth_client = OneDriveAuthClient()
+    _drive_client = OneDriveClient()
+
+    def handle_callback(self, response):        
+        """View called when the Oauth flow is completed. Adds a new OneDriveUserSettings
+        record to the user and saves the user's access token and account info.
+        """
+        
+        userInfoRequest = requests.get(("{}me?access_token={}").format(settings.MSLIVE_API_URL, response['access_token']))
```
samchrisinger: This confused me at first, but I actually really like having this request made here. :+1: 

I see a method in OneDriveClient that looks like it could handle this call. Is there any reason not to use the client here?

website/addons/onedrive/model.py
```diff
@@ -0,0 +1,258 @@
...
+    serializer = OneDriveSerializer
+#     myBase = AddonOAuthUserSettingsBase
+
+
+class OneDriveNodeSettings(StorageAddonBase, AddonOAuthNodeSettingsBase):
+
+    oauth_provider = OneDrive
+    serializer = OneDriveSerializer
+
+    foreign_user_settings = fields.ForeignField(
```
samchrisinger: In other addons this field is a pointer to a legacy database collection and can be removed here.

website/addons/onedrive/model.py
```diff
@@ -0,0 +1,258 @@
...
+    def display_name(self):
+        return '{0}: {1}'.format(self.config.full_name, self.folder_id)
+
+    @property
+    def has_auth(self):
+        """Whether an access token is associated with this node."""
+        return bool(self.user_settings and self.user_settings.has_auth)
+
+    @property
+    def complete(self):
```
samchrisinger: This looks identical to the superclass's definition and can be removed.

website/addons/onedrive/model.py
```diff
@@ -0,0 +1,258 @@
...
+            self.path = self.folder_name
+#             self.folder_path = '/'.join(
+#                 [x['name'] for x in self._folder_data['path_collection']['entries']]
+#                 + [self._folder_data['name']]
+#             )
+            self.save()
+
+    def set_folder(self, folder, auth):
+        self.onedrive_id = folder['id']
+        self.folder_id = folder['name'] #folder['id']
```
samchrisinger: Can you please explain the meanings behind 'id' and 'name' in this context? It seems to me it might make sense to either:
 
- redefine `folder_id` as a property method proxying `onedrive_id`
- get rid of `onedrive_id` in favor of `folder_id`

website/addons/onedrive/model.py
```diff
@@ -0,0 +1,258 @@
...
+        nodelogger = OneDriveNodeLogger(node=self.owner, auth=Auth(user_settings.owner))
+        nodelogger.log(action="node_authorized", save=True)
+
+    def deauthorize(self, auth=None, add_log=True):
+        """Remove user authorization from this node and log the event."""
+        node = self.owner
+
+        if add_log:
+            extra = {'folder_id': self.folder_id}
+            nodelogger = OneDriveNodeLogger(node=node, auth=auth)
```
samchrisinger: You can probably use the superclass's nodelogger property here instead.

website/addons/onedrive/utils.py
```diff
@@ -0,0 +1,93 @@
...
+
+from website.util import rubeus
+
+from website.addons.onedrive import settings
+
+
+logger = logging.getLogger(__name__)
+
+
+class OneDriveNodeLogger(object):
```
samchrisinger: Please see the generic NodeLogger defined in `website.addons.base.logger` and see if it will suit your purposes.

There's a corresponding property method in in the AddonOAuthNodeSettingsBase model that dynamically generates NodeLogger subclasses https://github.com/CenterForOpenScience/osf.io/blob/develop/website/addons/base/__init__.py#L803

website/addons/onedrive/views.py
```diff
@@ -0,0 +1,191 @@
+"""Views for the node settings page."""
```
samchrisinger: The code here in general looks very solid. 

We did just release some new addon code using generic view method factories: https://github.com/CenterForOpenScience/osf.io/blob/develop/website/addons/base/generic_views.py (with an example usage: https://github.com/CenterForOpenScience/osf.io/blob/develop/website/addons/dropbox/views.py)

Certainly not a blocker for this PR, but you might consider using the generic views whenever possible. One advantage to this is you get lots of free tests ([see dropbox](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/addons/dropbox/tests/test_views.py)), and you help us improve uniformity across the addon code. 

But again, since you've already taken the time to write these views and corresponding tests, this is optional.

website/files/models/onedrive.py
```diff
@@ -0,0 +1,16 @@
+from website.files.models.base import File, Folder, FileNode
+
+
+__all__ = ('OneDriveFile', 'OneDriveFolder', 'OneDriveFileNode')
+
+
+class OneDriveFileNode(FileNode):
```
samchrisinger: @chrisseto Do these models look ok to you?

website/addons/onedrive/client.py
```diff
@@ -0,0 +1,95 @@
...
+#import requests #TODO: remove this after determining onedrive connection issues w/make_request
+
+from requests_oauthlib import OAuth2Session
+from oauthlib.oauth2 import InvalidGrantError
+
+from framework.exceptions import HTTPError
+
+from website.util.client import BaseClient
+#from website.addons.googledrive import settings
+from website.addons.googledrive import exceptions
```
samchrisinger: Maybe (some of) these exceptions could be generalized and added to `website.addons.base.execptions`?

samchrisinger:
@caseyrygt I'm done with a first pass here. Please take a look at the comments. 

It would be good to have the Travis build passing (looks like both flake8 errors and test failures) before the next pass. 

Probably the next step is to finish addressing comments on the WaterButler PR, then I'll do a full functional review and another pass of CR.

caseyrygt:
@samchrisinger, thanks for the feedback.  I'll get things cleaned up and submit another PR along with the cleanup of the WB code.

Happy holidays!

website/addons/onedrive/model.py
caseyrygt: Just to confirm, the "foreign_user_settings" line can be removed.  Is that correct?

website/addons/onedrive/model.py
caseyrygt: Can you confirm which definitions are identical to the superclass?  I'm comparing OneDriveNodeSettings against StorageAddonBase and AddonOAuthNodeSettingsBase.

website/addons/onedrive/model.py
caseyrygt: @samchrisinger, the reason I have both is that if I assign the onedrive id (1234!88) to folder_id, then when the user loads the project settings page, the user sees 1234!88 instead of the folder name (my documents).  I looked but couldn't quite figure out where that was getting displayed/pulled from.  Do you know of the top of your head?

Thanks.

website/addons/onedrive/model.py
samchrisinger: Yes

website/addons/onedrive/model.py
samchrisinger: Sorry, I see now. 

It looks to me that calls to `self.user_settings.verify_oauth_access` when `external_account=None` will raise an exception. `AddonOAuthNodeSettingsBase#complete` adds a check for this in it's chain of conditions. 

So can you please confirm that the omission of `and self.external_account` is intentional here?

felliott:
Hey @caseyrygt,

I'm just getting back to review for the WB-side of this PR.  The OSF-side addon is working for me, but I get a lot of flake8 errors for this branch.  Could you fix those up?  Travis requires a passing flake check.

Also, I'm assuming https://github.com/caseyrygt/waterbutler/tree/feature/onedrive-provider-35 is the correct WB-side branch to review?  If so, could you re-open a PR for me to leave comments on?  Thank you!

Cheers,
@felliott

caseyrygt:
Hi Fitz,

You're correct, onedrive-provider-35 is the correct (python 3.5) branch to
review.

I'll submit a new PR and fix the flake8 errors in OSF.

Thanks!

On Wed, Apr 27, 2016 at 2:43 PM, Fitz Elliott <notifications@github.com>
wrote:

> Hey @caseyrygt <https://github.com/caseyrygt>,
>
> I'm just getting back to review for the WB-side of this PR. The OSF-side
> addon is working for me, but I get a lot of flake8 errors for this branch.
> Could you fix those up? Travis requires a passing flake check.
>
> Also, I'm assuming
> https://github.com/caseyrygt/waterbutler/tree/feature/onedrive-provider-35
> is the correct WB-side branch to review? If so, could you re-open a PR for
> me to leave comments on? Thank you!
>
> Cheers,
> @felliott <https://github.com/felliott>
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/CenterForOpenScience/osf.io/pull/4681#issuecomment-215188828>
>



-- 
Ryan Casey
CEO



<http://www.growingtechnologies.com/>
growingtechnologies.com
rcasey@growingtechnologies.com
Direct: (434) 326-0938
For Technical Support: 434-296-1500 option 2 /
techsupport@growingtechnologies.com

caseyrygt:
@samchrisinger & @felliott:  Sorry for the delay on this.  I think I have resolved all of the issues / items you had.  Could you please review this?

Thanks!

caseyrygt:
@samchrisinger: I meat to ask you about an error I'm getting in the test_views tests:
BuildError: ('JSONRenderer__onedrive_folder_list', {}, None)
It happens whenever I call api_url_for (e.g. url = api_url_for('onedrive_folder_list')).

In looking at the dropbox test_views, it calls: self.project.api_url_for.  Unfortunately, when I call that method I get: AttributeError: 'TestAuthViews' object has no attribute 'project'

I'm guessing that between when I'd copied the tests from dropbox to now, some other changes were made to OSF's testing or "base" framework.

Could you give me any points for resolving this?

Thanks.

samchrisinger:
@caseyrygt some of these are happening because the route just doesn't exist: https://github.com/CenterForOpenScience/osf.io/pull/4681/files#diff-0ee0e60ebe1815cd3f65f833c2e4976aR35
(here you need `api_url_for('oauth_connect', service_name='onedrive')`-- see https://github.com/CenterForOpenScience/osf.io/blob/develop/website/routes.py#L354)

I can't find an example test case using 'onedrive_folder_list', but all of the urls you try to look up need to exist in your onedrive/routes.py.

website/addons/onedrive/model.py
```diff
@@ -0,0 +1,240 @@
...
+        #  userInfo = userInfoRequest.json()
+        logger.debug("userInfo:: %s", repr(userInfo))
+
+        return {
+            'provider_id': userInfo['id'],
+            'display_name': userInfo['name'],
+            'profile_url': userInfo['link']
+        }
+
+    def _refresh_token(self, access_token, refresh_token):
```
mfraezz: This and the two other `refresh`-related functions in this class are likely superfluous, with `auto_refresh_url` defined. See `website/oauth/models/__init__.py:ExternalProvider.refresh_oauth_key` -- it does all of this work. This would also obviate `OneDriveAuthClient.refresh`; `.user_info` could be moved into `OneDriveClient`.

website/addons/onedrive/model.py
mfraezz: This can be removed.

website/addons/onedrive/model.py
mfraezz: I agree with @samchrisinger, unless there's a need to override a base implementation, please don't.

website/addons/onedrive/model.py
```diff
@@ -0,0 +1,240 @@
...
+        if self._api is None:
+            self._api = OneDrive(self.external_account)
+        return self._api
+
+    @property
+    def display_name(self):
+        return '{0}: {1}'.format(self.config.full_name, self.folder_name)
+
+    @property
+    def has_auth(self):
```
mfraezz: This should also probably use the base implementation -- a quick search through this PR indicated that the differences between this and the base could break things.

website/addons/onedrive/model.py
```diff
@@ -0,0 +1,240 @@
...
+    @property
+    def complete(self):
+        return bool(self.has_auth and self.user_settings.verify_oauth_access(
+            node=self.owner,
+            external_account=self.external_account,
+        ))
+
+    def fetch_folder_name(self):
+        self._update_folder_data()
+        return self.folder_name.replace('All Files', '/ (Full OneDrive)')
```
mfraezz: [OneDrive docs](https://dev.onedrive.com/items/get.htm#http-request-1) appear to indicate the string for the default root folder (erm... 'drive') is `'OneDrive'`. Can you confirm that linking the root causes to be shown on the files page?

website/addons/onedrive/model.py
```diff
@@ -0,0 +1,240 @@
...
+class OneDriveNodeSettings(StorageAddonBase, AddonOAuthNodeSettingsBase):
+
+    oauth_provider = OneDrive
+    serializer = OneDriveSerializer
+
+    foreign_user_settings = fields.ForeignField(
+        'onedriveusersettings', backref='authorized'
+    )
+    folder_id = fields.StringField(default=None)
+    onedrive_id = fields.StringField(default=None)
```
mfraezz: This should be removed in favor of using `folder_id` for this purpose.

website/addons/onedrive/model.py
```diff
@@ -0,0 +1,240 @@
...
+    folder_name = fields.StringField()
+    folder_path = fields.StringField()
```
mfraezz: This appears unused / not set by your code. Consider replacing with a `@property` that just returns `folder_id`

website/addons/onedrive/model.py
```diff
@@ -0,0 +1,240 @@
...
+
+    @property
+    def api(self):
+        """authenticated ExternalProvider instance"""
+        if self._api is None:
+            self._api = OneDrive(self.external_account)
+        return self._api
+
+    @property
+    def display_name(self):
```
mfraezz: I think this is unused -- two other references to `display_name` in this PR: one is in a test file, the other is when a field of the same name gets set on an `ExternalAccount` object.

website/addons/onedrive/utils.py
mfraezz: Ditto this^

website/addons/onedrive/views.py
```diff
@@ -0,0 +1,183 @@
...
+            ],
+        }
+    }
+
+
+@must_have_addon('onedrive', 'node')
+@must_be_addon_authorizer('onedrive')
+def onedrive_folder_list(node_addon, **kwargs):
+    """Returns a list of folders in OneDrive"""
+    if not node_addon.has_auth:
```
mfraezz: You'd have no way of knowing this, but we're planning on doing a small refactor to how folder-type addons list their folders soon -- we want this to be something that can happen outside of a flask context.

Please move the bulk of this to a function on the `OneDriveNodeSettings` class, and have this view 
`return node_addon.folder_list(<folder_id extracted from flask's request.args>)`

website/files/models/onedrive.py
mfraezz: Assuming that OneDrive is fairly typical in how their files work (and from a quick glance over their API, that seems to be a valid assumption), these should be fine.

website/addons/onedrive/model.py
mfraezz: Disregard, saw that it was being set as such in the view (presently, at least). I think this function, as well as `fetch_full_folder_path` and `_update_folder_data`, can be removed. Any references to them should be using `self.folder_name`

website/addons/onedrive/model.py
```diff
@@ -0,0 +1,240 @@
...
+        node = self.owner
+
+        if add_log:
+            extra = {'folder_id': self.folder_id}
+            nodelogger = OneDriveNodeLogger(node=node, auth=auth)
+            nodelogger.log(action="node_deauthorized", extra=extra, save=True)
+
+        self.folder_id = None
+        self._update_folder_data()
+        self.user_settings = None
```
mfraezz: This is done by `self.clear_auth()`

website/addons/onedrive/model.py
```diff
@@ -0,0 +1,240 @@
...
+    def deauthorize(self, auth=None, add_log=True):
+        """Remove user authorization from this node and log the event."""
+        node = self.owner
+
+        if add_log:
+            extra = {'folder_id': self.folder_id}
+            nodelogger = OneDriveNodeLogger(node=node, auth=auth)
+            nodelogger.log(action="node_deauthorized", extra=extra, save=True)
+
+        self.folder_id = None
```
mfraezz: Please write a `def clear_settings(self)` that sets this and `folder_name` to `None`. It's expected to exist by the base classes.

website/addons/onedrive/model.py
```diff
@@ -0,0 +1,240 @@
...
+                external_account=self.external_account,
+                metadata={'folder': self.folder_id}
+            )
+            self.user_settings.save()
+
+        # Add log to node
+        nodelogger = OneDriveNodeLogger(node=self.owner, auth=auth)  # AddonOAuthNodeSettingsBase.nodelogger(self)
+        nodelogger.log(action="folder_selected", save=True)
+
+    def set_user_auth(self, user_settings):
```
mfraezz: This appears to duplicate part of `<NodeSettings>.set_auth`, consider removing.

mfraezz:
Pausing review here for the night, picking up at the serializer in the AM.

website/addons/onedrive/serializer.py
```diff
@@ -0,0 +1,97 @@
...
+
+from website.addons.base.serializer import OAuthAddonSerializer
+
+from website.util import api_url_for, web_url_for
+
+logger = logging.getLogger(__name__)
+
+
+class OneDriveSerializer(OAuthAddonSerializer):
+    def credentials_owner(self, user_settings=None):
```
mfraezz: This should probably be using the base implementation. Currently it would throw an `AttributeError` if `user_settings` isn't passed in.

website/addons/onedrive/serializer.py
```diff
@@ -0,0 +1,97 @@
...
+        return user_settings.owner or self.user_settings.owner
+
+    @property
+    def addon_short_name(self):
```
mfraezz: This doesn't need to be a property; a class variable would be more consistent.

website/addons/onedrive/serializer.py
```diff
@@ -0,0 +1,97 @@
...
+        return 'onedrive'
+
+    @property
+    def user_is_owner(self):
```
mfraezz: This should probably instead be the base implementation, the only difference between the two (`or len(user_accounts)`) was removed from the base version because it could cause bugs.

website/addons/onedrive/serializer.py
```diff
@@ -0,0 +1,97 @@
...
+        user_accounts = self.user_settings.external_accounts
+        return bool(
+            (
+                self.node_settings.has_auth and
+                (self.node_settings.external_account in user_accounts)
+            ) or len(user_accounts)
+        )
+
+    @property
+    def serialized_urls(self):
```
mfraezz: This should probably be the base version, the only difference is the existence of error checking.

website/addons/onedrive/serializer.py
```diff
@@ -0,0 +1,97 @@
...
+                               service_name='onedrive'),
+            'importAuth': node.api_url_for('onedrive_add_user_auth'),
+            'files': node.web_url_for('collect_file_trees'),
+            'folders': node.api_url_for('onedrive_folder_list'),
+            'config': node.api_url_for('onedrive_set_config'),
+            'deauthorize': node.api_url_for('onedrive_remove_user_auth'),
+            'accounts': node.api_url_for('onedrive_get_user_settings'),
+        }
+
+    def serialize_settings(self, node_settings, current_user, client=None):
```
mfraezz: Instead of duplicating most of the base behavior, please implement a `def credentials_are_valid(sellf, user_settings, client)` that does the commented-out credential checking below.

mfraezz: Please also implement a `def serialized_folder(self, node_settings)` that returns a `dict` of the form `{'name': '<name>', 'path': '<path>'}`

website/addons/onedrive/static/onedrive.css
```diff
@@ -0,0 +1,39 @@
+.onedrive-confirm-selection {
```
mfraezz: I'm not certain this file is necessary -- styling is generally handled higher up the chain. Will investigate further later.

website/addons/onedrive/views.py
mfraezz: Might not have been a blocker in December, but now that every addon that can use `generic_views` does, it is. Sorry that this PR went without much attention for so long.

website/addons/onedrive/tests/test_models.py
```diff
@@ -0,0 +1,106 @@
+# -*- coding: utf-8 -*-
```
mfraezz: Not going to go through every test yet, but please update this file, as well as `test_serializers` and `test_views` to use the mixins that provide automated tests found in `website/addons/base/testing/`. See aforementioned test files for Box, Dropbox, or GoogleDrive for closest approximations of expected testing behaviors.

website/addons/onedrive/tests/utils.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def init_storage():
+    set_up_storage(MODELS, storage_class=storage.MongoStorage)
+
+
+class OneDriveAddonTestCase(AddonTestCase):
+
+    ADDON_SHORT_NAME = 'onedrive'
+
```
mfraezz: This class will need to be updated to use the generic tests -- it will need:
- to inherit from `OAuthAddonTestCaseMixin`
- `ExternalAccountFactory` and `Provider` defined,
- `set_user_settings` should probably go away in favor of using  `OAuthAddonTestCaseMixin.set_user_settings`, 
- `set_node_settings` will need a `super` call, and a `settings.save()` call.

website/addons/onedrive/utils.py
```diff
@@ -0,0 +1,88 @@
...
+        # Prefix the action with onedrive_
+        self.node.add_log(
+            action="onedrive_{0}".format(action),
+            params=params,
+            auth=self.auth
+        )
+        if save:
+            self.node.save()
+
+def onedrive_addon_folder(node_settings, auth, **kwargs):
```
mfraezz: This can go away when `generic_views` are implemented.

mfraezz:
Trivial, but all the console logging will need to go before this gets merged.

mfraezz:
Somewhat light first pass complete. 🐧 

I'll review in more depth and do functional testing when the above comments have been addressed. I know updating the views and tests to the new paradigm seems like a lot of work, but there's a good chance it'll be simpler for both of us than fixing and verifying the correctness of all the commented-out tests that presumably need some amount of maintenance.

If you have any questions or need any clarifications, feel free to @ mention me, or if you're on Slack I can hop on and discuss things with you there.

Back to you, @caseyrygt

PR #4682, opened 2015-12-02
OSF-5265 Opens the link in a new tab

adlius:
<h3>Purpose</h3>
Changes the template file so that new links open in new tabs.

<h3>Changes</h3>
Add "target= "_blank"" to the url.

### DIFF
--- a/website/static/css/pages/meeting-landing-page.css
+++ b/website/static/css/pages/meeting-landing-page.css
@@ -57,6 +57,15 @@
     margin-bottom: -15px;
 }
 
+.org-logo {
+    display: flex;
+    justify-content: center;
+}
+
+.org-logo div{
+    align-self: center;
+}
+
 .org-logo img{
     padding: 20px 20% 0 20%;
 }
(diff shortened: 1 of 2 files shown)


mfraezz:
Confirmed that the links now open in a new tab in FF/Chrome/IE.

The images should also be center aligned, as detailed in the [Jira ticket](https://openscience.atlassian.net/browse/OSF-5265). When that is done this will be ready to merge. :octopus:

adlius:
Center align images done.

whit537:
Looks like flexbox is [IE 10+](http://caniuse.com/#feat=flexbox), and buggy at that.

@adlius What's our browser support policy? Does this at least work in IE 10 and 11?

mfraezz:
@whit537 we don't have a formal browser support policy yet, but a proposal is in the works and should be through the review process within a few days.

adlius:
@mfraezz @whit537 I use the flexbox feature as it seems to be a more "modern" way of aligning responsive elements in CSS. However that does depend on what browsers and which versions we eventually decide to support.
Tell me when we have a standard on browser compatibility. I can make changes that supports older browsers. The current fix works on both the newest Safari and the newest Chrome. I do not have IE so I can't speak to how my changes render in IEs.

whit537:
> a proposal is in the works and should be through the review process within a few days.

Bump.

PR #4686, opened 2015-12-03
[RfR] Component logs reflect inherited actions regarding license

barbour-em:
## Purpose 
Since the license is inherited to all the components and nested components, the logs on those components should record the license actions.

## Changes

`add_log` called for each primary descendant of the updated node; trickle down license-omics. 

screen shot 2015-12-03 at 10 02 10 am

## Side effects
None apparent 

[OSF-5083]

### DIFF
--- a/website/project/model.py
+++ b/website/project/model.py
@@ -1239,18 +1239,20 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin):
         record.copyright_holders = copyright_holders or []
         record.save()
         self.node_license = record
-        self.add_log(
-            action=NodeLog.CHANGED_LICENSE,
-            params={
-                'parent_node': self.parent_id,
-                'node': self._primary_key,
-                'new_license': node_license.name
-            },
-            auth=auth,
-            save=False,
-        )
-        if save:
-            self.save()
+
+        for node in self.node_and_primary_descendants():
+            node.add_log(
+                action=NodeLog.CHANGED_LICENSE,
+                params={
+                    'parent_node': node.parent_id,
+                    'node': node._primary_key,
+                    'new_license': node_license.name
+                },
+                auth=auth,
+                save=False,
+            )
+            if save:
+                node.save()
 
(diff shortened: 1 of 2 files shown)


samanehsan:
Hey @barbour-em, I just tested this out and noticed that the logs aren't saving for the components. Mind looking into that?

sloria:
@barbour-em Can you please add regression test(s) for this change?

website/project/model.py
```diff
@@ -1239,18 +1239,20 @@ def set_node_license(self, license_id, year, copyright_holders, auth, save=True)
...
+                action=NodeLog.CHANGED_LICENSE,
+                params={
+                    'parent_node': node.parent_id,
+                    'node': node._primary_key,
+                    'new_license': node_license.name
+                },
+                auth=auth,
+                save=False,
+            )
+            if save:
```
sloria: I'm not sure this check is desired. `save=False` is useful for preventing a save of the primary object. Any secondary objects (e.g. descendant nodes) should be saved automatically.

This is to prevent a situation where not all nodes will get saved.

```python
node.set_node_license(... save=False)
# ...
node.save()
```

In this case, the descendant nodes would not be saved.

One workaround would be to skip saving `self` in the loop:

```python
for node in self.node_and_primary_descendants():
    # add log....
    if node._id != self._id:
        node.save()

if save:
    self.save()
```

sloria:
Pass finished.

website/project/model.py
```diff
@@ -1285,7 +1290,7 @@ def update(self, fields, auth=None, save=True):
                     value.get('year'),
                     value.get('copyright_holders'),
                     auth,
-                    save=False
+                    save=True
```
sloria: It probably makes sense to pass `save=save` here, so we don't have any unwanted implicit saves. @barbour-em Can you make that change and make sure this still works as expected?

sloria:
Pass finished.

barbour-em:
@sloria looks like it still works with the change :+1:

PR #4687, opened 2015-12-03
[Hotfix] Fix Project Settings Layout

ticklemepierce:
Purpose
-
As noted on the [Jira Ticket][1], the layout on the project settings page was messed up and one panel was wrapping all of the other panels.  This PR fixes this layout issue.
projectsettingslayout

Changes
-
Added a missing </div> tag and slightly restructured the panel (it had two panel-body elements which created a bunch of dead white space) to make the top panel be separate from the others and to be aesthetically pleasing
screen shot 2015-12-03 at 10 36 17 am

Side Effects
-
Since this PR is so small there should be nothing.  The only one I can think of would be messing up the knockout binding but this should not be an issue as it just adds in two unbound elements to the existing binding.  These elements should not be messed up as a result of being bound however.

[1]: https://openscience.atlassian.net/browse/OSF-5192


sloria:
@ticklemepierce Since this is a hotfix, can you please resend this PR to master?

ticklemepierce:
@sloria my mistake.  Had to restart the PR and forgot to change that the second time.  Here is the new one: https://github.com/CenterForOpenScience/osf.io/pull/4692

sloria:
Thanks

PR #4688, opened 2015-12-03
Change commenting defaults to any OSF user

barbour-em:
## Purpose
Change the default configuration to "any OSF user can comment." 

## Changes
Default `comment_level` field `'private'` => `'public'`

## Side effects
Any new, public project is commentable for any OSF user, not just contributors

[OSF-5002]

### DIFF
--- a/api_tests/comments/views/test_comment_report_list.py
+++ b/api_tests/comments/views/test_comment_report_list.py
@@ -86,11 +86,10 @@ class TestCommentReportsView(ApiTestCase):
         assert_equal(len(report_json), 0)
         assert_not_in(self.contributor._id, report_ids)
 
-    def test_public_node_non_contributor_does_not_see_report(self):
+    def test_public_node_non_contributor_does_see_report(self):
         self._set_up_public_project_comment_reports()
         res = self.app.get(self.public_url, auth=self.non_contributor.auth, expect_errors=True)
-        assert_equal(res.status_code, 403)
-        assert_equal(res.json['errors'][0]['detail'], 'You do not have permission to perform this action.')
+        assert_equal(res.status_code, 200)
 
     def test_public_node_logged_out_user_cannot_view_reports(self):
         self._set_up_public_project_comment_reports()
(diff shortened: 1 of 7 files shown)


samanehsan:
@barbour-em, please look into the failing tests. The comment API tests include default public and private projects, and public and private projects with `comment_level` set to `public`. Now that public and private projects have `comment_level = public` by default, please change the tests that were explicitly setting the `comment_level = public` to `comment_level = private`.

api_tests/nodes/views/test_node_comments_list.py
```diff
@@ -123,7 +123,6 @@ def _set_up_public_project(self):
     def _set_up_public_project_with_public_comment_level(self):
         """ Public project configured so that any logged-in user can comment."""
         self.project_with_public_comment_level = ProjectFactory(is_public=True, creator=self.user)
-        self.project_with_public_comment_level.comment_level = 'public'
```
samanehsan: This is currently the same as `_set_up_public_project` now that `comment_level = public` by default. Instead, it should be `_set_up_public_project_with_private_comment_level`, with `comment_level = private`.  

Then you can add `test_public_node_non_contributor_cannot_comment` back in using this set-up method, since we still want to be sure non-contributors can't comment on projects when `comment_level = private` 

And you can replace any usages of `_set_up_public_project_with_public_comment_level` with `_set_up_public_project`

api_tests/comments/views/test_comment_report_list.py
```diff
@@ -237,10 +236,10 @@ def test_public_node_non_contributor_can_report_comment(self):
             comment (comment_level == 'public), non-contributors
             can also report comments.
         """
-        project = ProjectFactory(is_public=True, comment_level='public')
```
samanehsan: This is now testing the same thing as `test_public_node_logged_in_non_contributor_can_report_comment`. In this test, can you instead make `comment_level = private` and test that non-contributors cannot report comments on public projects with a private comment level? Basically what `test_public_node_logged_in_non_contributor_cannot_report_comment` was doing.

api_tests/comments/views/test_comment_report_list.py
```diff
@@ -86,11 +86,10 @@ def test_public_node_reported_user_does_not_see_report(self):
         assert_equal(len(report_json), 0)
         assert_not_in(self.contributor._id, report_ids)
 
-    def test_public_node_non_contributor_does_not_see_report(self):
+    def test_public_node_non_contributor_does_see_report(self):
         self._set_up_public_project_comment_reports()
         res = self.app.get(self.public_url, auth=self.non_contributor.auth, expect_errors=True)
-        assert_equal(res.status_code, 403)
-        assert_equal(res.json['errors'][0]['detail'], 'You do not have permission to perform this action.')
+        assert_equal(res.status_code, 200)
```
samanehsan: Non-contributors should be able to access this endpoint, but they should not be able to see any reports that are not theirs. Can you assert that no report info is returned? In other tests, I've been doing this with
``` 
report_json = res.json['data']
report_ids = [report['id'] for report in report_json]  
assert_equal(len(report_json), 0)
```

samanehsan:
Actually, I have an open PR #4648 that makes a lot of changes to the commenting tests, and adds in a bunch of new tests that would probably need to be changed too. Should we hold off on this until #4648 is merged? @erinmayhood

sloria:
Closing this for now, given the significant changes made to commenting.

PR #4689, opened 2015-12-03
[Ready for Review] Add queryable root and parent fields to Node

erinspace:
# Purpose
Relates to https://openscience.atlassian.net/browse/OSF-5207
Allows a quick way to do a modular ODM query for all nodes descending from a top level node.
No user-facing changes!
Any node will now have a root field and a parent that contains the top level node._id and it's immediate parent's _id. If the current node is the top level node, the root is that node's ID. 
API can also now filter on that root field, and the parent field can be filtered on being None. 

# Changes
- Add root field to node
- Save root when saving a node
- Tests that forks, registrations, and template projects have their own roots.
- modify existing root property to _root
- Add parent_node field to node
- Save parent_node when saving a node
- Tests that forks, registrations, and template projects have their own parent_node.
- modify existing parent_node property to _parent_node
- migration script to include parent_node and root on existing nodes.
- modification to create_fakes that lets you add project structure and node depth
- add root to filterable_fields and as a separate field in the Node serializer
- add parent to filterable_fields and as a separate field in the Node serializer with source parent_node
- add ability to filter on null parent
- API test that filters on root
- API test that filters on parent eq None

# Side effects
- create_fakes.py component creation takes a string, that can be a list describing component structure and depth
- This assumes the HideIfRetraction wrapper around these fields really does go away on parent_node and root node fields, otherwise will expose a field in retracted registrations. https://openscience.atlassian.net/browse/OSF-5359

- **note!** You might possibly maybe need to run ```scripts/consistency/fix_bad_children.py``` before running the migration if there's a problem with some data


api/base/filters.py
```diff
@@ -215,6 +218,8 @@ def convert_value(self, value, field):
                 )
         elif isinstance(field, self.LIST_FIELDS):
             return value
+        elif isinstance(field, RelationshipField):
```
mfraezz: This will conflict with a [PR](https://github.com/CenterForOpenScience/osf.io/pull/4693/files#diff-8a17f6dba621ae1374cefd4514bfc739R219) that @samanehsan has open. No action required at this time, just a heads up.

scripts/migration/migrate_ultimate_parent_node.py
```diff
@@ -0,0 +1,58 @@
...
+                            child.parent_id
+                        )
+                    )
+                else:
+                    child.save()
+    assert (all_nodes_count == touched_counter)
+
+
+if __name__ == '__main__':
+    main()
```
mfraezz: Please update this script to use the newer way of handling dry runs. See one of [these](https://github.com/CenterForOpenScience/osf.io/blob/develop/scripts/googledrive/migrate_to_external_account.py#L83-L95) [scripts](https://github.com/CenterForOpenScience/osf.io/blob/develop/scripts/migration/migrate_registered_meta.py#L103-L108) for example. It just puts the migration in a `TokuTransaction` and throws an error to roll it back afterwards if `dry`

scripts/migration/migrate_ultimate_parent_node.py
```diff
@@ -0,0 +1,58 @@
...
+                        '{}/{}: touched node {} with parent {}'.format(
+                            touched_counter,
+                            all_nodes_count,
+                            child._id,
+                            child.parent_id
+                        )
+                    )
+                else:
+                    child.save()
+    assert (all_nodes_count == touched_counter)
```
mfraezz: This is throwing an error with a recent staging DB — a single node isn’t being hit because its parent doesn’t think it’s its child:
``` python
In [44]: problem
Out[44]: <Node(title=u'Title for a test Node.', category=u'project') with _id u'kszcd'>

In [45]: problem.parents
Out[45]: [<Node(title=u'Private Node', category=u'project') with _id u'5xe8z'>]

In [46]: ppar = Node.load('5xe8z')

In [47]: pci = [e._id for e in ppar.get_descendants_recursive(include=lambda n: n.primary)]

In [48]: problem._id in pci
Out[48]: False

In [49]: problem.creator
Out[49]: <User(u'cwisecarver+ultimate-user@cos.io') with id u'y352a'>
```
Consult @sloria to see if this is a problem on prod, and blame @cwisecarver for causing trouble.

To get around this, you could keep track of all the `_id`s for `touched_nodes`, find the set difference above this assert (e.g. `untouched_nodes = list(set([e._id for e in all_nodes])-set(touched_nodes))`), and then `save` them, but this might not be necessary.

website/project/model.py
```diff
@@ -1364,6 +1365,14 @@ def save(self, *args, **kwargs):
 
         saved_fields = super(Node, self).save(*args, **kwargs)
 
+        ultimate_parent = self._primary_key
+        if getattr(self, 'parent', None):
+            # TODO Remove if statement after ultimate_parent migration
```
mfraezz: It's possible to avoid a follow-up PR by interacting with the database directly, but if you would rather `DO` this afterwards to avoid the headache, I have no objections.

scripts/migration/migrate_ultimate_parent_node.py
cwisecarver: I think this happened while I was testing our transaction stuff on staging. Reverse relationships weren't being saved correctly. I blame our transaction code.

mfraezz:
First pass complete. 

Tests are very thorough, code is straightforward. However, while the migration script works on a fairly trivial local database, it fails on staging (with 0 of ~12000 nodes having a `.ultimate_parent`), even with the proposed fix for the problem node. As discovered in person, approximately 1 in 15 failed when testing with your local db of ~15000 nodes. 

It's worth noting that newly nodes created nodes and their subcomponents do have a correct `ultimate_parent`, and that the field is `modularodm.Q`ueryable. :octocat: 

Once this migration issue has been addressed, this should be good to merge.

scripts/migration/migrate_ultimate_parent_node.py
brianjgeiger: What's weird is the node creation date seems to be november 11, 2015, and all those problems were pre-october, so I'm not sure what's happening there. Maybe someone killed the db in the middle of a runscope test?

erinspace:
Thanks so much @mfraezz! Changes made.

website/project/model.py
```diff
@@ -755,6 +755,7 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin):
     nodes = fields.AbstractForeignField(list=True, backref='parent')
     forked_from = fields.ForeignField('node', backref='forked', index=True)
     registered_from = fields.ForeignField('node', backref='registrations', index=True)
+    ultimate_parent = fields.StringField(index=True)
```
mfraezz: After an architectural consult with Steve, please change this to be a `ForeignField` named `root`, and the `root` property to be named `_root`, as it's the cached field that we want to expose.

scripts/migration/migrate_ultimate_parent_node.py
```diff
@@ -0,0 +1,73 @@
...
+from framework.transactions.context import TokuTransaction
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def do_migration():
+    init_app(routes=False)
+    logger.warn('ultimate_parent field will be added to all nodes.')
+    all_nodes = models.Node.find()
```
mfraezz: You should only be migrating non-deleted nodes.

api/nodes/serializers.py
```diff
@@ -107,6 +108,11 @@ class NodeSerializer(JSONAPISerializer):
         related_view_kwargs={'node_id': '<parent_id>'}
     )
 
+    ultimate_parent = RelationshipField(
+        related_view='nodes:node-detail',
+        related_view_kwargs={'node_id': '<ultimate_parent>'}
```
mfraezz: This will need to change to get the `._id` from the newly renamed `root`. L41 above will also need to be renamed.

mfraezz:
Please add tests covering the case where a node with children is deleted.

mfraezz:
CR complete. :pear: 

This is looking good, just a few more changes and it should be ready for merge.

erinspace:
OK! @mfraezz ready for another lookover!

api/nodes/serializers.py
```diff
@@ -102,10 +104,11 @@ class NodeSerializer(JSONAPISerializer):
         related_meta={'count': 'get_pointers_count'},
     )))
 
-    parent = HideIfRetraction(RelationshipField(
```
mfraezz: I know these are due to an underlying bug in the api. I'm okay with this as long as @sloria is, but please make a note of the side effects in the PR description.

brianjgeiger: I added a note.

erinspace: oh noo @brianjgeiger I am sorry I was adding a note at the same time as you added your note and I think I deleted your note :(

brianjgeiger: You did. It's okay, I've added the link to the relevant issue back.

scripts/migration/migrate_ultimate_parent_node.py
```diff
@@ -0,0 +1,73 @@
...
+from framework.transactions.context import TokuTransaction
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def do_migration():
+    init_app(routes=False)
+    logger.warn('ultimate_parent field will be added to all nodes.')
+    all_undeleted_nodes = models.Node.find(Q('is_deleted', 'eq', False))
```
mfraezz: Do we want to ignore registrations here as well, or just deleted nodes?

scripts/migration/migrate_ultimate_parent_node.py
```diff
@@ -0,0 +1,73 @@
...
+    all_undeleted_nodes_count = all_undeleted_nodes.count()
+    touched_counter = 0
+    logger.info('There are {} total nodes'.format(all_undeleted_nodes_count))
+    for node in all_undeleted_nodes:
+        if not getattr(node, '_parent_node', None):
+            touched_counter += 1
+            node.save()
+            children = [child for child in node.get_descendants_recursive(include=lambda n: n.primary)]
```
mfraezz: Either here or [here](https://github.com/CenterForOpenScience/osf.io/pull/4689/files#diff-2ba596be2052918b9e0ac1caeba22d7dR43), you probably want to do a check to make sure none of the children are nodes that should be excluded either (`is_deleted`, etc).

scripts/migration/migrate_ultimate_parent_node.py
```diff
@@ -0,0 +1,73 @@
...
+                logger.info(
+                    '{}/{}: touched node {} with parent {}'.format(
+                        touched_counter,
+                        all_undeleted_nodes_count,
+                        child._id,
+                        child.parent_id
+                    )
+                )
+
+                child.save()
```
mfraezz: When I run this locally, for some reason the `.save()` call seems to not actually save `child.root` the first time, but calling it again does. May be worth investigating.

scripts/migration/migrate_ultimate_parent_node.py
```diff
@@ -0,0 +1,73 @@
...
+                assert child.root._id == node._id
+                logger.info('Saved Node {} with root {}'.format(child._id, child.ultimate_parent))
```
mfraezz: With the rename, the last format arg should be changed to `child.root`.

scripts/migration/migrate_ultimate_parent_node.py
erinspace: Just deleted nodes! Registrations need both parents and roots.

scripts/migration/migrate_ultimate_parent_node.py
mfraezz: Recent model changes fixed this :+1:

mfraezz:
LGTM. Ensured that new functionality still worked as expected and that the migration ran successfully with staging data. :sunrise: 
```python
In [1]: all_undeleted_nodes = models.Node.find(Q('is_deleted', 'eq', False))

In [2]: all([(n.root != None and n.root == n._root) for n in all_undeleted_nodes])
Out[2]: True
```

api/base/serializers.py
```diff
@@ -266,6 +266,7 @@ def __init__(self, related_view=None, related_view_kwargs=None, self_view=None,
         self.related_meta = related_meta
         self.self_meta = self_meta
         self.always_embed = always_embed
+        self.source_field = source_field
```
sloria: Why is `source_field` necessary when fields already have `source`?

sloria: Could you instead override the behavior of `HyperlinkedIdentityField` and set `self.source` after the `super()` call?

api_tests/nodes/views/test_node_list.py
```diff
@@ -96,6 +97,35 @@ def test_omit_retracted_registration(self):
...
+        res = self.app.get(self.url+'?embed=root&embed=parent', auth=self.user.auth)
+
+        projects_with_correct_root = 0
+        for project in res.json['data']:
+            root_id = project['embeds']['root']['data']['id']
+
+            if project['embeds']['parent'].get('data'):
+                parent_id = project['embeds']['parent']['data']['id']
+                if parent_id == project_one._id:
+                    if root_id == project_one._id:
```
sloria: Rather than having just one assert at the end of this test, why not make this comparison an assertion:

```python
assert_equal(root_id, project_one._id)
```

That way, the test will fail early, and the output will tell you *which* root ID is incorrect.

api/base/serializers.py
erinspace: It was indeed because the HyperlinkedIdentityField was setting ```kwargs['source'] = *``` no matter what

I tried setting it after the ```super()``` call instead but something went wrong with that. I can try again and better document the issue, because I don't remember exactly what it was.

api_tests/nodes/views/test_node_list.py
```diff
@@ -96,6 +97,35 @@ def test_omit_retracted_registration(self):
...
+        project_one = ProjectFactory(title="Project One", is_public=True)
+        ProjectFactory(parent=project_one, is_public=True)
+
+        res = self.app.get(self.url+'?embed=root&embed=parent', auth=self.user.auth)
+
+        projects_with_correct_root = 0
+        for project in res.json['data']:
+            root_id = project['embeds']['root']['data']['id']
+
+            if project['embeds']['parent'].get('data'):
```
sloria: There's a lot of logic here, which makes this test hard to follow. If you're just trying to check that the root IDs are correct, then you can simplify. How about something along the lines of:

```python
for project_json in res.json['data']:
     # Load the project from the DB
    project = Node.load(project_json['id'])
    # The root ID in the JSON should be the correct root ID
    assert_equal(project_json['embeds']['root']['data']['id'], project.root._id)
```

api_tests/nodes/views/test_node_list.py
```diff
@@ -346,6 +376,31 @@ def test_incorrect_filtering_field_not_logged_in(self):
         assert_equal(len(errors), 1)
         assert_equal(errors[0]['detail'], "'notafield' is not a valid field for this endpoint.")
 
+    def test_filtering_on_root(self):
+        # Build up a family of nodes
+        node_structure = [5, 2, [1, 2], 3, [10, 2], 2, [4, [2, 2], 6]]
+        render_generations_from_node_structure_list(self.project_one, node_structure)
```
sloria: Is it necessary to create such a complex node structure? Not only will creating 42 nodes slow down the test, but it will also make it harder to debug the test if it fails.

sloria: Better to keep it simple, e.g. create a single root project that has a child, then some other unrelated nodes. Filtering on the root project should return the child only.

api_tests/nodes/views/test_node_list.py
```diff
@@ -346,6 +376,31 @@ def test_incorrect_filtering_field_not_logged_in(self):
...
+    def test_filtering_on_null_parent(self):
+        # Build up a family of nodes not to be included
+        node_structure = [2, [1, 2]]
+        render_generations_from_node_structure_list(self.project_one, node_structure)
+
+        url = '/{}nodes/?filter[parent]=null'.format(API_BASE)
+
+        res = self.app.get(url, auth=self.user_one.auth)
+        assert_equal(res.status_code, 200)
+        assert_equal(res.json['links']['meta']['total'], 4)
```
sloria: If I understand this test correctly, `filter[parent]=null` results in no filtering at all. Is this the desired behavior, @brianjgeiger @erinspace ? 

As a user, I think I'd expect `filter[parent]=null` to mean "filter on nodes with no parents", i.e. root nodes only.

api_tests/nodes/views/test_node_list.py
brianjgeiger: Why do you hate awesomeness and thoroughness and awesome thoroughness, Steve?

api_tests/nodes/views/test_node_list.py
brianjgeiger: Yeah, `filter[parent]=null` should return root nodes.

api_tests/nodes/views/test_node_list.py
```diff
@@ -346,6 +376,31 @@ def test_incorrect_filtering_field_not_logged_in(self):
...
+        render_generations_from_node_structure_list(self.project_one, node_structure)
+        # create some unrelated projects
+        ProjectFactory(title="Road Dogg Jesse James", is_public=True)
+        ProjectFactory(title="Badd *** Billy Gunn", is_public=True)
+
+        url = '/{}nodes/?filter[root]={}'.format(API_BASE, self.project_one._id)
+
+        res = self.app.get(url, auth=self.user_one.auth)
+        assert_equal(res.status_code, 200)
+        assert_equal(res.json['links']['meta']['total'], 42)
```
sloria: When possible, try not to rely on counts in tests, as they are very brittle (if someone adds `ProjectFactory()` in the wrong place, the test will fail in a sad way).

Better to make assertions against values that are more stable, such as primary keys.

```python
# make root project...
# make child of root project...
# make filtered request
# child should be included in the result
result_ids = [each['id'] for each in ....]
assert_in(child._id, result_ids)

File: api_tests/nodes/views/test_node_list.py (hunk 16 above)
Comment (brianjgeiger) [inline api_tests/nodes/views/test_node_list.py]: On a slightly more serious note, my only real concern is if someone refactors the function that gathers all of the children into a flat list in such a way that, on some structure of node, will return a child node earlier in a list than the parent node. This test is, perhaps, a bit more aggressive than it needs to be on that front, but if that happens, it would be pretty terrible. Not disastrous, though; once the migration is done, the only likely scenario will be if someone is moving a tree structure from one node to another, and that *could* be fixed by hand if need be.

File: api_tests/nodes/views/test_node_list.py (hunk 17 above)
Comment (sloria) [inline api_tests/nodes/views/test_node_list.py]: Or... I misunderstood the test, and `4` is the number of public root nodes. Is this correct, @erinspace ?

Again, try to avoid counts--they're brittle and can be hard to track. Relying on `/links/meta/total` is brittle--what if we change the format of `meta`, or remove counts, or there's a bug in the formatting of links?

You can get the count more directly:

```python
assert_equal(len(res.json['data'], 4)
```

Or even:

```python
# so robust
public_root_nodes = Node.find(Q('is_public', 'eq', True) & Q('root', 'eq', None))
assert_equal(len(res.json['data']), public_root_nodes.count())
```

api_tests/nodes/views/test_node_list.py
sloria: I'm a fan of awesomethoroughnomics so long as the 'awesome' is simple to understand. If we need to test deeply nested relationships, you can create those easily:

```python
root = ProjectFactory()
child = ProjectFactory(parent=root)
sibling = ProjectFactory(parent=root)
grandchild = ProjectFactory(parent=child)
```

Now, if something breaks in the test, you can throw a breakpoint in and inspect the above fixtures if need be.

api/base/serializers.py
sloria: I see. The `source` attribute defines how values are pulled off of the objects being serialized. It'd be best not to mess with that.

However, "source_field" is a bit of a misnomer; it implies that it, too, has something to do with pulling values from the objects to serialize. In reality, it defines the key used for filtering the field. Maybe rename to "filter_key"?

sloria: Also, it would be useful to document what that does in the docstring, as it isn't obvious how it's used..

website/project/model.py
```diff
@@ -1362,6 +1364,10 @@ def save(self, *args, **kwargs):
         else:
             suppress_log = False
 
+        self.root = self._root._id
```
sloria: Shouldn't this be just `self._root`?

scripts/migration/migrate_ultimate_parent_node.py
```diff
@@ -0,0 +1,76 @@
+"""
+This will add an ultimate_parent field to all nodes.
+Ultimate_parent will be the primary key of the originating parent node
```
sloria: Please update the name of this script and update all references to `ultimate_parent`.

tests/test_models.py
```diff
@@ -3055,6 +3054,183 @@ def test_permission_override_on_readded_contributor(self):
...
+        found_ids = []
+        for node in family_nodes:
+            assert_in(node._id, family_ids)
+            found_ids.append(node._id)
+        for node_id in family_ids:
+            assert_in(node_id, found_ids)
+
+    def test_get_descendants_recursive_returns_in_depth_order(self):
+        # Build up a family of nodes
+        node_structure = [5, 2, [1, 2], 3, [10, 2, 1], 2, [4, [2, 2], 6]]
```
sloria: Again, this complex structure makes this test slow and hard to debug. See previous comments on simplifying.

website/project/model.py
erinspace: It seems like this is correct - changing this to ```self._root```  it breaks almost all existing tests with the error ```DatabaseError: Record must be loaded.```

api/nodes/serializers.py
```diff
@@ -26,9 +26,12 @@ def to_internal_value(self, data):
 
 
 class NodeSerializer(JSONAPISerializer):
-    # TODO: If we have to redo this implementation in any of the other serializers, subclass ChoiceField and make it
-    # handle blank choices properly. Currently DRF ChoiceFields ignore blank options, which is incorrect in this
-    # instance
+    """TODO: If we have to redo this implementation in any of the other serializers, subclass ChoiceField and make it
```
sloria: Will making this a docstring have any effect on the autogenerated swagger docs or BAPI?

brianjgeiger: I don't *think* so, because I think it's just the views that get made into docs, but I'd lean towards keeping it the # comments.

api/nodes/serializers.py
```diff
@@ -32,6 +32,8 @@ class NodeSerializer(JSONAPISerializer):
     # TODO: If we have to redo this implementation in any of the other serializers, subclass ChoiceField and make it
     # handle blank choices properly. Currently DRF ChoiceFields ignore blank options, which is incorrect in this
     # instance
+    # filter_key is passed in the RelationshipField as to not interfere with source which operates in a specialized way
+    # for HyperlinkedFields!
```
sloria: It would make more sense to document `filter_key` in the docstring of `RelationshipField`.

api_tests/nodes/views/test_node_list.py
```diff
@@ -96,6 +98,25 @@ def test_omit_retracted_registration(self):
         res = self.app.get(self.url, auth=self.user.auth)
         assert_equal(len(res.json['data']), 2)
 
+    def test_node_list_has_root(self):
+        res = self.app.get(self.url, auth=self.user.auth)
+        projects_with_root = 0
+        for project in res.json['data']:
+            if project['relationships'].get('root', None):
+                projects_with_root += 1
+        assert_equal(projects_with_root, 2)
```
sloria: Shouldn't all projects have a root? If so, then it might be more useful to assert that

```python
assert_true(all([each['relationships'].get('root') is not None for each in res.json['data']]))
```

That would conveniently get rid of the need to rely on a count (`projects_with_root`), which is brittle.

api_tests/nodes/views/test_node_list.py
```diff
@@ -385,6 +406,39 @@ def test_incorrect_filtering_field_not_logged_in(self):
...
+        ProjectFactory(parent=child, is_public=True)
+        # create some unrelated projects
+        ProjectFactory(title="Road Dogg Jesse James", is_public=True)
+        ProjectFactory(title="Badd *** Billy Gunn", is_public=True)
+
+        url = '/{}nodes/?filter[root]={}'.format(API_BASE, root._id)
+
+        res = self.app.get(url, auth=self.user_one.auth)
+        assert_equal(res.status_code, 200)
+        assert_equal(res.json['links']['meta']['total'], 4)
```
sloria: As suggested in https://github.com/CenterForOpenScience/osf.io/pull/4689#discussion_r47702972, don't rely on `/links/meta/total`; you can get the length of the data directly.

api_tests/nodes/views/test_node_list.py
brianjgeiger: We'd also need to make sure that it runs a non-zero number of times, but otherwise that would be fine.

api_tests/nodes/views/test_node_list.py
```diff
@@ -385,6 +406,39 @@ def test_incorrect_filtering_field_not_logged_in(self):
...
+
+    def test_filtering_on_null_parent(self):
```
sloria: Is there a test for filtering on a valid parent ID?

website/project/model.py
sloria: Oh, I see; at this point, the root node may not have been loaded. Thanks for the clarification.

api_tests/nodes/views/test_node_list.py
brianjgeiger: I wouldn't be surprised if there isn't. We aren't testing everything filtering related, just some general example and new or exceptional cases. Testing filtering on root covers the "relationship link" filtering, and null parent tests the filtering on a null relationship link.

sloria: Sounds good. Was just curious; this isn't a blocker.

api/base/filters.py
```diff
@@ -296,7 +299,10 @@ def query_params_to_odm_query(self, query_params):
             query_parts = []
             for field_name, params in filters.iteritems():
                 for group in params:
-                    query = Q(field_name, group['op'], group['value'])
+                    value = group['value']
+                    if value == 'null':
```
sloria: Converting input values from the client to their corresponding Python values is the responsibility of `convert_value`. Please move this.

sloria:
Pass finished. Just minor changes to make, and this is good to merge. :full_moon_with_face:

scripts/migration/migrate_root_and_parent_on_node.py
```diff
@@ -0,0 +1,79 @@
...
+from scripts import utils as script_utils
+from framework.transactions.context import TokuTransaction
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def do_migration():
+    init_app(routes=False)
+    logger.warn('ultimate_parent field will be added to all nodes.')
```
sloria: Should be "root"

scripts/migration/migrate_root_and_parent_on_node.py
```diff
@@ -0,0 +1,79 @@
...
+    all_undeleted_nodes = models.Node.find(Q('is_deleted', 'eq', False))
```
sloria: [nitpick] Or just "all_nodes"`.

sloria: Also, you can probably make this migration resume-able if you also query by `root_node`.

```python
Q('root_node', 'eq', None)
```

It's worth trying.

tests/utils.py
```diff
@@ -136,6 +137,24 @@ def make_drf_request(*args, **kwargs):
     return Request(http_request, *args, **kwargs)
 
 
+def render_generations_from_parent(parent, num_generations):
```
sloria: Are these used any more? If not, they should be removed.

sloria:
Looks good.

I will make the above minor changes when I merge. :octopus:

sloria:
Running migration locally on prod data. Will merge if it succeeds.

scripts/migration/migrate_root_and_parent_on_node.py
```diff
@@ -0,0 +1,79 @@
...
+                logger.info('Child Node saved: Verifying that save Node {} with root {}'.format(child._id, child.root))
+                assert child.parent_node._id == child.parent_id
+                assert child.root._id == node._id
+
+    assert all_nodes_count == touched_counter
+
+
+def main(dry=True):
+    init_app(set_backends=True, routes=False)  # Sets the storage backends on all models
+    with TokuTransaction():
```
sloria: Since the whole migration is wrapped in a transaction, it won't actually be resume-able. We can fix this by only transacting updates to individual nodes.

scripts/migration/migrate_root_and_parent_on_node.py
```diff
@@ -0,0 +1,79 @@
...
+                    '{}/{}: touched node {} with parent {}'.format(
+                        touched_counter,
+                        all_nodes_count,
+                        child._id,
+                        child.parent_id
+                    )
+                )
+
+                child.save()
+                logger.info('The child root id is {} and the node id is {}'.format(child.root._id, node._id))
```
sloria: This logger statement is redundant; I'll remove it.

sloria:
Even though all nodes will have `root` set after the migration finishes, it's possible (quite likely, actually--the migration has been running for ~5 minutes and is only 20% finished) that a user visits a node *during* the migration. In this case, they will get a 500 error because of this line: https://github.com/erinspace/osf.io/blob/c1d97915d68708d79646322f0a0cad64cdd23a45/website/project/views/node.py#L751 .

We can fix this by handling when `root` is `None`:

```python
      'root_id': node.root._id if node.root else None,
```

sloria:
Got a traceback:

```
/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/modularodm/storedobject.py:940: UserWarning: Setting an attribute that is neither a field nor a protected value.
  warnings.warn('Setting an attribute that is neither a field nor a protected value.')
    exec code in run_globals
  File "/Users/sloria/projects/cos-projects/osf/scripts/migration/migrate_root_and_parent_on_node.py", line 78, in <module>
    main(dry=dry)
  File "/Users/sloria/projects/cos-projects/osf/scripts/migration/migrate_root_and_parent_on_node.py", line 71, in main
    do_migration(dry=dry)
  File "/Users/sloria/projects/cos-projects/osf/scripts/migration/migrate_root_and_parent_on_node.py", line 65, in do_migration
    assert child.root._id == node._id
  File "/Users/sloria/projects/cos-projects/osf/scripts/migration/migrate_root_and_parent_on_node.py", line 35, in do_migration
    node.save()
  File "website/project/model.py", line 1388, in save
    saved_fields = super(Node, self).save(*args, **kwargs)
  File "framework/guid/model.py", line 94, in save
    self._ensure_guid()
  File "framework/guid/model.py", line 73, in _ensure_guid
    guid = Guid.load(self._primary_key)
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/modularodm/storedobject.py", line 110, in wrapped
    return func(*args, **kwargs)
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/modularodm/storedobject.py", line 82, in wrapped
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/modularodm/storedobject.py", line 567, in load
    data = cls.from_storage(data)
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/modularodm/storedobject.py", line 110, in wrapped
    return func(*args, **kwargs)
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/modularodm/storedobject.py", line 347, in from_storage
    value = field_object.from_storage(data_value, translator)
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/modularodm/fields/abstractforeignfield.py", line 47, in from_storage
    self.get_primary_field(value[1])\
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/modularodm/fields/abstractforeignfield.py", line 23, in get_primary_field
    schema_class = self.get_schema_class(schema)
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/modularodm/fields/abstractforeignfield.py", line 20, in get_schema_class
    return self._schema_class.get_collection(schema)
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/modularodm/storedobject.py", line 287, in get_collection
    return cls._collections[name.lower()]
KeyError: u'nodefile'
```

I'm guessing this is a database inconsistency and doesn't have to do with the code in this patch. I'll investigate.

sloria:
Yep, the above traceback was due to a node whose files were unable to be migrated in the NodeFile migration; unrelated to this migration.

sloria:
Another traceback:

```
    exec code in run_globals
  File "/Users/sloria/projects/cos-projects/osf/scripts/migration/migrate_root_and_parent_on_node.py", line 92, in <module>
    main(dry=dry)
  File "/Users/sloria/projects/cos-projects/osf/scripts/migration/migrate_root_and_parent_on_node.py", line 85, in main
    do_migration(dry=dry)
  File "/Users/sloria/projects/cos-projects/osf/scripts/migration/migrate_root_and_parent_on_node.py", line 76, in do_migration
    assert child.root._id == node._id
  File "/Users/sloria/projects/cos-projects/osf/scripts/migration/migrate_root_and_parent_on_node.py", line 75, in do_migration
    assert child.parent_node._id == child.parent_id
AttributeError: 'NoneType' object has no attribute '_id'
```

Investigating it now.

sloria:
The above was caused by a strange inconsistency whereby a node has a deleted parent, but the node itself was not deleted. Will address this.

PR #4690, opened 2015-12-04
this adds a clean task to remove *.pyc files

danielneis:
Hello,

sometimes during development we had to clean *.pyc files to get things working.
So I am adding this as a task =)

Hope you like.

Kind regards,
Daniel

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -907,3 +907,10 @@ def update_citation_styles():
     from scripts import parse_citation_styles
     total = parse_citation_styles.main()
     print("Parsed {} styles".format(total))
+
+@task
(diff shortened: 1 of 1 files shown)


lbanner:
You are right @danielneis -- appreciate it. Pinging @sloria ... when you get a chance to review. Thanks!

sloria:
`find . -name "*.pyc" -delete`  is bit more straightforward.

Also, we typically don't add a `verbose` flag to our tasks. Instead, we pass `echo=True` to `run()`.

These are minor nitpicks; I can make those change those when I merge. 

Thanks!

PR #4691, opened 2015-12-04
Add to admin tasks

Ghalko:
## Proposal
Add manage task to admin tasks to include env variable so that testing/deploying is much simpler.


admin/tasks.py
```diff
@@ -5,6 +5,21 @@
...
 
+
+@task()
+def manage(*args):
+    """Take arguments and run python manage
+
+    :param args: ex. runserver, migrate
+    """
+    if os.getcwd() != HERE:
+        os.chdir(HERE)
```
sloria: Why is this necessary?

Ghalko: Is that taken care of by the namespace?

sloria: No, but why is it necessary to change the working directory in the first place?

Ghalko: Because we need to know where manage.py is. I guess I could do a `os.path.join(HERE, '..', 'manage.py')`?

Ghalko:
@sloria Improved, but couldn't find anything on how to handle a list of args since invoke can handle running commands sequentially on the same line.

tasks/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+"""Consolidates invoke tasks and utilities
+
+"""
+from .tasks import *  # noqa
```
sloria: I'm not sure of the side effects of having both `tasks.py` and `tasks/__init__.py`. It might make more sense to just move all the code from tasks.py into `tasks/__init__.py` to avoid potential clashes.

Ghalko: Is that a good approach? Is there anyplace that we import tasks.py? The only place I see a potential is in archiver/listeners.py

Ghalko: And is it good design practice to put code in init?

sloria: We wouldn't import tasks in the application code, but `invoke` must be importing the tasks code; having two conflicting modules to import `tasks.py` and `tasks/__init__.py` is hard to reason about.

It's not necessarily a bad thing to have code in `__init__.py`. For readability, we might break up `tasks/__init__.py`  into multiple files (and then import those modules in `__init__.py`), but that doesn't need to happen now.

Ghalko: Then I will update this shortly.

Ghalko:
@sloria Moved, good?

sloria:
Looks good.

Re: the `manage` command, I think what you had before--a single, required argument for the command string--was preferred. You would just need to make sure you use quotes around the command.

```python
inv manage 'runserver -p 9876'
```

That removes the need for the `target` argument, and allows any number of additional arguments to be passed to the management. commands.

sloria:
I'll make the above change when I merge.

Ghalko:
Sounds good. Thanks. (It will be inv admin.manage though ;)

PR #4693, opened 2015-12-04
[Feature][Ready for Review] Filter by a relationship field's object id

samanehsan:
Purpose
-----------
Addresses JIRA [OSF-5330]

Add ability to filter by a relationship field's object id. This is required for incorporating the API v2 comments routes into the commenting feature in development. 

E.g. `/v2/nodes/node_id/comments/?filter[target]=target_id`

Changes
------------
- Update `filters.py` to handle `RelationshipField` and `TargetField` 
- Make comment `target` a filterable field
- Test that filtering by `target` works

I included comment filtering in this PR so that there could be something to test this filtering with. 

Note to QA: Right now, you can only filter by target=node_id, since the node comments endpoint only lists top-level comments on the node. With the commenting feature, this endpoint will return all comments and users will also be able to filter by target comment and file.

### DIFF
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -19,6 +19,7 @@ from api.base.exceptions import (
     InvalidFilterFieldError
 )
 from api.base import utils
+from api.base.serializers import RelationshipField, TargetField
 
 def sort_multiple(fields):
     fields = list(fields)
@@ -214,6 +215,8 @@ class FilterMixin(object):
                     field_type='date'
                 )
         elif isinstance(field, self.LIST_FIELDS):
+            return value
+        elif isinstance(field, (RelationshipField, TargetField)):
             return value
         else:
             try:
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -73,6 +73,8 @@ class FilterMixin(object):
 
     LIST_FIELDS = (ser.ListField, )
 
+    RELATIONSHIP_FIELDS = (RelationshipField, TargetField)
+
     def __init__(self, *args, **kwargs):
         super(FilterMixin, self).__init__(*args, **kwargs)
         if not self.serializer_class:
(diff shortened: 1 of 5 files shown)


lbanner:
Hi @samanehsan ... JIRA ticket # for this? Thanks -

samanehsan:
Writing it up now @lbanner!

api/base/filters.py
```diff
@@ -215,6 +216,8 @@ def convert_value(self, value, field):
                 )
         elif isinstance(field, self.LIST_FIELDS):
             return value
+        elif isinstance(field, (RelationshipField, TargetField)):
```
mfraezz: `self.LIST_FIELDS` is a tuple of one that doesn't appear to be used anywhere outside of this file (aside from, perhaps, somewhere inside of Django). Is there any reason not to put `RelationshipField, TargetField` into `LIST_FIELDS`, or to do the comparison together (as they all return `value`)?

mfraezz:
Please update documentation to include this new query param ([here](https://github.com/samanehsan/osf.io/blob/55fe79668aa9b8d4eae828cd82a1c2d57a5ce4bd/api/nodes/views.py#L1685) and [maybe here](https://github.com/samanehsan/osf.io/blob/55fe79668aa9b8d4eae828cd82a1c2d57a5ce4bd/api/comments/views.py#L99) )

Pass complete, no bugs found :100: 

For QA purposes, it may be worth noting in the PR that you can currently only filter by `node_id` due to the way comment lists are handled by the node endpoint.

api/base/filters.py
samanehsan: I don't think `RelationshipField` and `TargetField` belong in `self.LIST_FIELDS`. But I do agree the comparison can be combined since they all return `value`. I can add `self.RELATIONSHIP_FIELDS = (RelationshipField, TargetField)` to follow the patten in this mixin.

mfraezz:
Thanks for updating the description and docs.

CR complete, passing it off to @sloria.

sloria:
Looks good!

PR #4695, opened 2015-12-04
[WIP] Prereg Admin

samchrisinger:


samchrisinger:
@Ghalko, @cwisecarver this is a WIP, but I'm hoping this is all of the major changes that need to be made. A couple notes:
- Adds OsfUser model, and reference from Django User to OsfUser
- Adds inline form component to User for OsfUser.osf_id (Django Admin)
- Remove dead view and JS code
- Some template refactor

admin/pre-reg/urls.py
```diff
@@ -0,0 +1,12 @@
+from django.conf.urls import url
+
+from . import views
+
+urlpatterns = [
+    url(r'^$', views.prereg, name='prereg'),
```
cwisecarver: Wouldn't these be more RESTful if they were 
```python
    url(r'^drafts/(?P<draft_pk>[0-9a-z]+)/form/$', views.prereg_form, name='prereg_form'),
    url(r'^drafts/(?P<draft_pk>[0-9a-z]+)/approve/$', views.approve_draft, name='approve_draft'),
    url(r'^drafts/(?P<draft_pk>[0-9a-z]+)/reject/$', views.reject_draft, name='reject_draft'),
    url(r'^drafts/(?P<draft_pk>[0-9a-z]+)/update/$', views.update_draft, name='update_draft'),
    url(r'^drafts/get/$', views.get_drafts, name='get_drafts')
```

tasks.py
```diff
@@ -90,6 +90,12 @@ def adminserver(port=8001):
     cmd = '{} python manage.py runserver {} --nothreading'.format(env, port)
     run(cmd, echo=True, pty=True)
 
+# TODO - create task to handle manage.py tasks after setting the settings_module
+@task
+def admin_task(args):
+    """Run Admin manage.py command"""
+    cmd = 'DJANGO_SETTINGS_MODULE="api.base.settings" python manage.py ' + args
```
cwisecarver: Wrong setting module :shipit:

admin/common_auth/admin.py
```diff
@@ -0,0 +1,54 @@
+from django.contrib import admin
+from django.contrib.auth.admin import UserAdmin
+from django.contrib.auth.models import Permission
+from django.contrib.auth.forms import PasswordResetForm
+from .models import MyUser
+from forms import CustomUserRegistrationForm
+
+from models import OsfUser
```
cwisecarver: This import is coming from the same module as .models.MyUser on line 5. These could be combined just for ease of merging later.

admin/common_auth/models.py
```diff
@@ -0,0 +1,72 @@
...
+        ordering = ['email']
+
+    # Todo: implement this if needed
+    # @property
+    # def is_staff(self):
+    #     "Is the user a member of staff?"
+    #     return self.is_admin
+
+
+class OsfUser(models.Model):
```
cwisecarver: There's no migration for this model. Running `DJANGO_SETTINGS_MODULE="admin.base.settings" python manage.py makemigrations` will make a 0003_osfuser.py. Then running `DJANGO_SETTINGS_MODULE="admin.base.settings" python manage.py migrate` will make the table for this user in your db.

admin/common_auth/models.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def get_full_name(self):
+        # The user is identified by their email address
+        return self.email
+
+    def get_short_name(self):
+        # The user is identified by their email address
+        return self.email
+
+    def __str__(self):
```
cwisecarver: We should define __unicode__ in py2 and __str__ in py3.

admin/templates/login.html
```diff
@@ -0,0 +1,70 @@
...
+
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>OSFAdmin | Log in</title>
+    <!-- Tell the browser to be responsive to screen width -->
+    <meta content='width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no' name='viewport'>
+    <!-- Bootstrap 3.3.4 -->
+    <link href="{% static 'css/bootstrap.min.css' %}" rel="stylesheet" type="text/css" />
```
cwisecarver: I think @zkraime is fixing these static references but they don't work. Referring back to `base.html`, these should point to `vendor/bower/something`, not sure how well that will work once we're in prod and nginx is serving static files. This page is also missing jQuery right now and throws an error regarding that.

admin/pre-reg/utils.py
```diff
@@ -0,0 +1,19 @@
+from framework.utils import iso8601format
+from website.project.metadata.utils import serialize_meta_schema
+
+# TODO: Write and use APIv2 serializer for this
+def serialize_draft_registration(draft, auth=None):
```
cwisecarver: The auth kwarg is never used. This also might be better in a serializers.py unless the utils.py filename is already a pattern that exists in osf for old-style serializers.

admin/pre-reg/views.py
```diff
@@ -0,0 +1,171 @@
...
+
+from framework.auth.core import User as OsfUser
+from framework.mongo.utils import get_or_http_error
+from framework.exceptions import HTTPError, PermissionsError
+from website.project.model import MetaSchema, DraftRegistration
+from website.exceptions import NodeStateError
+
+get_draft_or_error = functools.partial(get_or_http_error, DraftRegistration)
+
+def load_osf_user(django_user):
```
cwisecarver: You could move this to a property on `MyUser` just for discoverability.

admin/pre-reg/views.py
```diff
@@ -0,0 +1,171 @@
...
+
+    if not django_user.osf_user:
+        raise RuntimeError()
+    else:
+        return OsfUser.load(django_user.osf_user.osf_id)
+
+def get_prereg_users():
```
cwisecarver: Django's preferred way of doing this type of thing is with a custom manager and a custom queryset: https://docs.djangoproject.com/en/1.9/topics/db/managers/#calling-custom-queryset-methods-from-the-manager

This does create quite a lot of boilerplate code so the django_model_utils package includes a PassThroughManager that encapsulates most of the boilerplate http://django-model-utils.readthedocs.org/en/latest/managers.html#passthroughmanager

The for loop with the filter in it will run one query for each user in users, plus the initial query to get users. Using the manager/queryset option should do a `MyUser.objects.filter(groups__name='prereg_group')` which will make one query and give you a list of users that are in the prereg group.

Alternatively, you could do `Group.objects.filter(name='prereg_group').user_set.all()` and that would give you all the users in the prereg_group.

admin/pre-reg/views.py
```diff
@@ -0,0 +1,171 @@
...
+    """Retrieves users on the admin site who are in the prereg_group
+    :return: List of usernames of those who are in the prereg_group
+    """
+    reviewers = []
+    users = User.objects.all()
```
cwisecarver: This should be `common_auth.models.MyUser` as well as the import on line 5.

admin/pre-reg/views.py
```diff
@@ -0,0 +1,171 @@
...
+    :param draft_pk: Unique id for current draft
+    :return: DraftRegistrationApproval obj
+    """
+    draft = get_draft_or_error(draft_pk)
+
+    user = load_osf_user(request.user)
+    try:
+        draft.approve(user)
+    except PermissionsError:
+        draft.approval.add_authorizer(user, save=True)
```
cwisecarver: Maybe this is okay... it just reads like an anti-pattern to me.

```
1. Try to do a thing that needs permissions
2. If you don't have permission add permissions for the user
3. ???
4. Profit.
```

If we know what we're doing this is probably fine it just reads like it's going to be exploited one day.

admin/pre-reg/views.py
```diff
@@ -0,0 +1,171 @@
...
+@login_required
+# @user_passes_test(is_in_prereg_group)
+@csrf_exempt
+def update_draft(request, draft_pk):
+    """Updates current draft to save admin comments
+
+    :param draft_pk: Unique id for current draft
+    :return: DraftRegistration obj
+    """
+    data = json.load(request)
```
cwisecarver: It's working but this should probably be `json.loads(request.body)`. I believe it's working on an edge case around the request.\__repr__ method returning the request.body.

admin/static/js/pages/admin-registration-edit-page.js
```diff
@@ -0,0 +1,21 @@
+var $ = require('jQuery');
```
cwisecarver: no `'use strict';`?

admin/static/js/pages/prereg-admin-page.js
```diff
@@ -0,0 +1,254 @@
...
+ProofOfPub.prototype.editItem = function() {
+    var self = this;
+    self.proofOfPub.edit(!self.editing);
+    self.editing(!self.editing);
+};
+
+/**
+* Payment sent column.
+*/
+var PaymentSent = function() {
```
cwisecarver: For some reason, in my test pre-reg, in the admin side my dropdown was defaulted to Payment Sent === 'Yes'. I have yet to receive this payment.

admin/pre-reg/views.py
samchrisinger: I should have given this a comment at least, and this will likely need to be tweaked. 

Since Prereg admins can feasibly added at any time, they won't necessarily be "authorizers" for each pending DraftRegistrationApproval. The Sanction architecture was not designed to support this, so this is a little hack to make Prereg approvals works. I see two options for dealing with this: 

1) Change Sanction inheritance structure to allow for non-token based approvals/rejections. Probably this should have happened initially, but we didn't have the use case until now. (1-2 hrs)
2) Add a comment here, and perhaps also add: 
````
if not user in draft.get_authorizers():
    raise PermissionsError()
else: 
    draft.approval.add_authorizer(user, save=True)
```

I'd like to see 1) pursued eventually, but this should not be a blocker to the first release.

admin/templates/login.html
```diff
@@ -0,0 +1,70 @@
...
+            <p>{{ form.password }}</p>
+            {% if form.password.errors %}<span class="text-danger"> {{ form.password.errors }}</span>{% endif %}
+          </div>
+          <div class="row">
+            <div class="col-xs-4">
+              <button type="submit" class="btn btn-primary btn-block btn-flat">Sign In</button>
+            </div><!-- /.col -->
+          </div>
+          <div>
+            {% if messages %} <!-- TODO: refactor message code to display proper class type -->
```
cwisecarver: List items should have either a OL or a UL around them.

admin/templates/pre-reg/edit_draft_registration.html
```diff
@@ -0,0 +1,98 @@
+{% extends 'base.html' %}
+{% load render_bundle from webpack_loader %}
+{% load staticfiles %}
+{% load filters %}
+
+{% block content %}
+{% render_bundle 'vendor' %}
+<link href="{% static 'css/registrations.css' %}" rel="stylesheet" type="text/css" />
+<script>
```
cwisecarver: `type="application/javascript"` attribute just for fun?

admin/templates/pre-reg/edit_draft_registration.html
```diff
@@ -0,0 +1,98 @@
...
+                </div>
+              </div>
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
+{% render_bundle 'admin-registration-edit-page' %}
```
cwisecarver: There's a `bottom_js` block defined in base.html that this could go in.

admin/templates/pre-reg/prereg.html
```diff
@@ -0,0 +1,70 @@
+{% extends 'base.html' %}
+{% load render_bundle from webpack_loader %}
+
+{% block content %}
+{% render_bundle 'vendor' %}
+<script>
```
cwisecarver: `type="application/javascript"` attribute just for fun?

admin/templates/pre-reg/prereg.html
```diff
@@ -0,0 +1,70 @@
+{% extends 'base.html' %}
+{% load render_bundle from webpack_loader %}
+
+{% block content %}
+{% render_bundle 'vendor' %}
```
cwisecarver: There's a block called `top_includes` that appears right before the closing head tag. All pre-content javascript and styles should probably go in there. Also, render_bundle 'vendor' is already being called in base.html at the bottom of the page.

admin/common_auth/models.py
Ghalko: Done, also cleaned up defaults to use a local.py file in settings and ignore the db file that is created.

admin/templates/pre-reg/prereg.html
```diff
@@ -0,0 +1,70 @@
...
+          <td class="text-center" id="assignee">
+            <select data-bind="visible: adminPermission() ,value: assignee.assignee, options: assignee.reviewers, click: assignee.editItem, clickBubble: false"></select>
+            <span data-bind="visible: !adminPermission(), text: assignee.assignee, click: assignee.editItem, clickBubble: false"></span>
+          </td>
+        </tr>
+    </tbody>
+  </table>
+</div>
+
+{% render_bundle 'prereg-admin-page' %}
```
cwisecarver: There's a bottom_js block that this could be inside of. The benefit of sticking js inside of blocks instead of inside the main content block is you might want to turn off just the js one day without duplicating the entire body of the page.

admin/pre-reg/views.py
cwisecarver: Makes sense. I was just pointing a light at it.

website/static/js/registrationUtils.js
```diff
@@ -646,9 +648,55 @@ Draft.prototype.submitForReview = function() {
...
+            approve: {
+                label: 'Approve',
+                class: 'btn-danger',
+                callback: ret.resolve
+            }
+        }                
+    });
+    return ret.promise();
+};
+Draft.prototype.reject = function(confirmed) {
```
cwisecarver: Why does reject take a `confirmed` that it never uses and approve doesn't?

website/static/js/registrationUtils.js
```diff
@@ -646,9 +648,55 @@ Draft.prototype.submitForReview = function() {
...
+                }
+            },
+            approve: {
+                label: 'Approve',
+                class: 'btn-danger',
+                callback: ret.resolve       
+            }
+        }                
+    });
+    return ret.promise();
```
cwisecarver: Couldn't the object passed to the `bootbox.dialog` be constructed once, send to approve and then cloned/updated when it's sent to the reject method? Cuts about 20 lines out of the file.

cwisecarver:
If I am on a detail page for a draft registration and I take an action (add a comment, edit a comment) I am no longer able to take another action until I reload the page. The first (successful) time that I make a request I do a PUT to `{id}/update/`. The subsequent times, until a refresh, I end up doing a POST to the current url which fails with a 403. I couldn't figure out what part of the JS was becoming confused but I believe that it's somewhere in the JS.

cwisecarver:
I know this is a first draft for internal only. Just pointing out that there are some UI oddities: 
- The cursor doesn't change to a pointer over rows in the table that are clickable. 
- Hitting the edit comment button does focus the text box but doesn't indicate that it is focused, no cursor or highlight happens. 
- The dropdowns in the table on the listing page don't seem to do anything.  
- On the draft detail page the sections of  the submission don't use the correct mouse cursor. 
- The approve button changes size on hover. 
- The confirm on the reject button says Approve.
- The last saved date doesn't seem to work unless it's a fresh page view.
- Approved registrations continue to appear in the list and trying to reject an approved registration throws an unhandled exception around creating a url.

cwisecarver:
Back to @samchrisinger, @Ghalko, et al. I'm happy to talk through any of my comments. Let me know if I can help.

admin/pre-reg/utils.py
Ghalko: Done. removed auth, renamed the file to serializers.

admin/pre-reg/views.py
Ghalko: Should we add it to the MyUser as a field?

cwisecarver: Doesn't need to be a database field but if you added:
```python
@property
def osf_user_instance():
    if not self.osf_user:
        raise RuntimeError()
    else:
        return OsfUser.load(self.osf_user.osf_id)
```

Then you could do:
```python
my_user_instance = MyUser.objects.all()[0]
osf_user_instance = my_user_instance.osf_user_instance
```

Django prefers fat-models over helper methods.

admin/pre-reg/views.py
Ghalko: Done. Removed get_pre_reg users replaced with `MyUser.objects.prereg_users()`

admin/common_auth/admin.py
samchrisinger: Done

admin/templates/login.html
samchrisinger: Done

admin/pre-reg/views.py
samchrisinger: Done

admin/pre-reg/views.py
samchrisinger: Done

admin/static/js/pages/admin-registration-edit-page.js
samchrisinger: Done

admin/pre-reg/views.py
samchrisinger: @cwisecarver if https://github.com/samchrisinger/osf.io/commit/11146cc8806750841ae4bcda3a146b3b08429a20 goes beyond the scope of this PR I'll pull it out.

admin/pre-reg/views.py
samchrisinger: TODO: Actually remove the helper and usages

admin/static/js/pages/prereg-admin-page.js
```diff
@@ -0,0 +1,254 @@
...
+};
+
+Notes.prototype.enlargeIcon = function(data, event) {
+    var icon = event.currentTarget;
+    $(icon).addClass("fa-2x");
+};
+
+Notes.prototype.shrinkIcon = function(data, event) {
+    var icon = event.currentTarget;
+    $(icon).removeClass("fa-2x");
```
samchrisinger: @leb2dg the Knockout way to do this would be:
```
<... data-bind="css: {hover: 'fa-2x'}, event: {hover: toggleHover}" >
```
where toggle hover toggles the state of a boolean observable on the view model.

And if you want to re-use this pattern of UI elements that change on hover, you might consider using the [oop module](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/oop.js) to create a mixin class to contain the duplicated logic (e.g. [ChangeMessageMixin] (https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/changeMessage.js))

admin/static/js/pages/prereg-admin-page.js
```diff
@@ -0,0 +1,254 @@
...
+        permission = true;
+    }
+    self.adminPermission = ko.observable(permission);
+
+    self.editing = ko.observable(false);
+
+    self.title = params.registration_metadata.q1.value || 'no title';
+    self.fullname = params.initiator.fullname;
+    self.username = params.initiator.emails[0].address;
+    self.initiated = self.formatTime(params.initiated);
```
samchrisinger: It would be better to create `Date` instances and use the formatting method only in the template i.e. 
```self.initiated = new Date(params.initiated)``` and 
```<... data-bind="text: $root.formatDate(initiated)">```

File: admin/static/js/pages/prereg-admin-page.js [hunk 25]
```diff
@@ -0,0 +1,254 @@
...
+
+Row.prototype.formatTime = function(time) {
+    var parsedTime = time.split(".");
+    return parsedTime[0]; 
+};
+
+Row.prototype.goToDraft = function(data, event) {
+    var self = this;
+    if (self.editing() === false) {
+        self.viewingDraft(true);
```
samchrisinger: Is `viewingDraft` still used?

admin/static/js/pages/prereg-admin-page.js
```diff
@@ -0,0 +1,254 @@
...
+    self.drafts = ko.observable([]);
+    self.loading = ko.observable(true);
+
+    self.sortBy = ko.observable('title');
+    self.sortedDrafts = ko.computed(function() {
+        var row = self.sortBy();
+        return self.drafts().sort(function (left, right) { 
+            var a = deep_value(left, row).toLowerCase();
+            var b = deep_value(right, row).toLowerCase();
+            return a == b ? 0 :
```
samchrisinger: This could be `return a - b;`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

admin/static/js/pages/prereg-admin-page.js
```diff
@@ -0,0 +1,254 @@
...
+    // create new view model for each row
+    getDrafts.then(function(response) {
+        self.drafts(
+            $.map(response.drafts, function(draft){
+                return new Row(draft, self.user.admin, self.reviewers);
+            })
+        );
+    });
+
+    $.when(getDrafts).then(function() {
```
samchrisinger: It would be better not to use `$.when` for a single promise, as this is typically used when you need to run a callback after multiple promises are resolved. L213  could just go in the done callback to `getDrafts`. 

You might consider adding a `fail()` callback here as well so the user gets feedback if the request fails.

admin/static/js/pages/prereg-admin-page.js
```diff
@@ -0,0 +1,254 @@
...
+    var reviewers = prereg_reviewers;
+    var adminView = new AdminView('#prereg-row', user, reviewers);
+    adminView.init();
+    ko.applyBindings(adminView, $('#prereg-row')[0]);
+});
+
+/**
+* Gets value from nested properties.
+*
+* @param  {Dict}    obj    The object to find value from
```
samchrisinger: `@param {Object} obj`

admin/pre-reg/urls.py
sloria: `drafts/<pk>/update` should be unnecessary. How about implementing `PUT` and/or `PATCH` for `/drafts/<pk>/`?

Also, `/drafts/get` can just be `/drafts/`.

tasks.py
```diff
@@ -90,6 +90,12 @@ def adminserver(port=8001):
     cmd = '{} python manage.py runserver {} --nothreading'.format(env, port)
     run(cmd, echo=True, pty=True)
 
+# TODO - refactor? It is functional. e.g. 'invoke admin_task createsuperuser'
+@task
+def admin_task(args):
```
sloria: Does `args` work? I wouldn't expect it to; invoke creates CLI options for each function argument, so I you would have to do

```python
inv admin_task -a createsuperuser
```

You might experiment with doing `*args` instead.

sloria: Ha, I just read the TODO above this that says this works.

sloria: Might make sense to rename this task to `manage`.

sloria: Also, any admin tasks should be moved to `admin/tasks.py`.

Ghalko: That is in my https://github.com/CenterForOpenScience/osf.io/pull/4691 it only works for one argument, otherwise the invoke expects to move on to a new invoke task.

admin/static/js/pages/admin-registration-edit-page.js
sloria: Use `require('jquery')` (case matters in non-OSX environments).

admin/static/js/pages/prereg-admin-page.js
samchrisinger: TODO(@leb2dg) this can be addressed post-merge

admin/pre-reg/urls.py
sloria: Also, using a class-based view (the simple View class, not the generic CRUD ones) may help to group views for RESTful resources.

admin/static/js/pages/prereg-admin-page.js
```diff
@@ -0,0 +1,254 @@
...
+*/
+var Notes = function() {
+    var self = this;
+    self.edit = ko.observable(false);
+    self.notes = ko.observable('none');
+};
+
+Notes.prototype.enlargeIcon = function(data, event) {
+    var icon = event.currentTarget;
+    $(icon).addClass("fa-2x");
```
sloria: See Sam's suggestion re: Knockout usage. Prefer declarative bindings to manual DOM updates like this.

admin/pre-reg/urls.py
Ghalko: @sloria That is the way we expect to go, more CBV. :+1:

requirements.txt
```diff
@@ -52,3 +52,6 @@ django-rest-swagger==0.3.2
 djangorestframework-bulk==0.2.1
 pyjwt==1.4.0
 pyjwe==0.1.6
+
+# Admin requirements
+django-extensions==1.5.9
```
sloria: If this is an admin requirement, then it should be in `requirements/admin.txt`.

That said, are you sure this is a production requirement? IIRC, the extensions are for development environments only, in which case this shouldn't be installed on production.

cwisecarver: I think this was used for debugging urls. I think it could be strictly dev or removed for now. It does give you some model mixing but I don't think we're using any of those in this PR.

tasks.py
sloria: Yeah, it makes sense now. Since `args` is a required argument (i.e. there is no default), you can pass it positionally on the command line. Subsequent tokens will be treated as separate arguments. You might be able to workaround it by using quotes, like:

```python
inv admin.manage 'createsuperuser foo@bar.com'
```

sloria:
Pass finished. 

The only blocker is https://github.com/CenterForOpenScience/osf.io/pull/4695/files#r47130423 ; the other comments should be resolved in a followup PR.

@samchrisinger Can you please get this up to date with feature/admin-app one last time and fix the above bug?

requirements.txt
samchrisinger: I uninstalled the module and the app still runs... so unless @zik3qe can tell use a good reason to keep this I'm removing this dependency.

admin/base/settings/defaults.py
```diff
@@ -26,14 +23,21 @@
...
 INSTALLED_APPS = (
     'django.contrib.admin',
     'django.contrib.auth',
     'django.contrib.contenttypes',
+    'django.contrib.messages',
+    'django.contrib.sessions',
     'django.contrib.staticfiles',
+    'admin.common_auth',
     'admin.base',
+    'admin.pre-reg',
```
sloria: Doesn't this need to be `admin.pre_reg` to work?

admin/base/settings/defaults.py
```diff
@@ -41,6 +45,9 @@
     'webpack_loader',
 )
 
+# Custom user model (extends AbstractBaseUser)
+AUTH_USER_MODEL = 'common_auth.MyUser'
```
sloria: Is this correct? Or does this need to be `common_auth.models.MyUser`? Or does Django do some magic to make the `models` unnecessary?

cwisecarver: I thought the same thing. It does some magic. It's app_name.ModelName...

Ghalko: Django does the magic.

cwisecarver: https://docs.djangoproject.com/en/1.9/topics/auth/customizing/#substituting-a-custom-user-model

admin/base/settings/defaults.py
Ghalko: At the moment it does work, but it should probably be the underscore.

admin/base/settings/defaults.py
sloria: OK, thanks

admin/base/settings/defaults.py
sloria: OK, I'll change this when I merge.

admin/templates/login.html
zkraime: @cwisecarver Ah yes I will add the OL tag when I fix the message class type

admin/static/js/pages/prereg-admin-page.js
samchrisinger: @leb2dg, really though a CSS class with the `:hover` selector would be the best way to do this.

PR #4701, opened 2015-12-07
Add missing node logs parameters

brianjgeiger:
Purpose
-----------
When /v2/nodes/logs was added, we didn't go too crazy with the params serialization. But in that I missed the addons params. This adds those params in. Closes https://openscience.atlassian.net/browse/OSF-5347

Changes
-----------
New fields mostly. A couple parameters are accessed differently than before, such as dataset becoming data_set, github.user becoming github_user, and so on. Templated node is a relationship link rather than having serialized data.

screen shot 2015-12-07 at 12 44 04 pm

Side effects
-------------
None

### DIFF
--- a/api/logs/serializers.py
+++ b/api/logs/serializers.py
@@ -25,14 +25,29 @@ class NodeLogFileParamsSerializer(RestrictedDictSerializer):
 
 class NodeLogParamsSerializer(RestrictedDictSerializer):
 
-    tags = ser.CharField(read_only=True)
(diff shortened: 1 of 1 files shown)


sloria:
For readability, it'd be nice to group the addon fields together, but this is fine to merge as is.

brianjgeiger:
Yeah, when I was adding them, it was a lot easier to do alphabetically to ensure that everything was in.

PR #4702, opened 2015-12-07
[Ready for Review] Feature: Replace v1 comment routes with v2 routes

samanehsan:
Purpose
------------
Replace existing API v1 routes with API v2 routes for comments. 

Changes
------------
In addition to changing the routes...

CommentSerialzer:
- Add `can_edit`, `has_children` and `is_abuse` fields to the comment serializer
- Add max_length to comment content field 
- Do not allow whitespace comments

Comment model:
- `create` method sends a signal to send the notification event, since that logic was happening in the v1 create view function which is no longer being used.

Side Effects
----------------
Instead of escaping html in comments and making it a string, it gets removed entirely by DRF. E.g. the comment `"<em>hi</em>"` would be `"hi"` instead.

**Note: This pull request cannot be merged until the v2 API supports anonymous view only links**


mfraezz:
After a few offline q's, pass complete. One bug found: replying to comments no longer works. They nest in the UI, but upon reloading that page they no longer display as replies.

api/comments/serializers.py
```diff
@@ -47,12 +51,31 @@ class CommentSerializer(JSONAPISerializer):
     class Meta:
         type_ = 'comments'
 
+    def get_is_abuse(self, obj):
+        auth = Auth(self.context['request'].user)
```
sloria: Is it necessary to construct an `Auth` object here? This method only seems to need a `user`.

api/comments/serializers.py
```diff
@@ -47,12 +51,31 @@ class CommentSerializer(JSONAPISerializer):
...
+        if not auth or auth.user.is_anonymous():
```
sloria: I don't think it's possible for `auth` to be falsy here.

api/comments/serializers.py
```diff
@@ -40,6 +41,9 @@ class CommentSerializer(JSONAPISerializer):
     date_modified = ser.DateTimeField(read_only=True)
     modified = ser.BooleanField(read_only=True, default=False)
     deleted = ser.BooleanField(read_only=True, source='is_deleted', default=False)
+    is_abuse = ser.SerializerMethodField()
```
sloria: Minor: It would be good to add `help_text` for these fields, but that doesn't need to block this PR.

api/comments/serializers.py
```diff
@@ -47,12 +51,31 @@ class CommentSerializer(JSONAPISerializer):
...
         type_ = 'comments'
 
+    def get_is_abuse(self, obj):
+        auth = Auth(self.context['request'].user)
+        if not auth or auth.user.is_anonymous():
+            return False
+        return auth.user and auth.user._id in obj.reports
+
+    def get_can_edit(self, obj):
+        auth = Auth(self.context['request'].user)
```
sloria: See previous comment--I think `Auth` is unnecessary here.

api_tests/comments/views/test_comment_detail.py
```diff
@@ -193,6 +194,44 @@ def test_public_node_non_contributor_commenter_can_update_comment(self):
...
         assert_equal(payload['data']['attributes']['content'], res.json['data']['attributes']['content'])
 
+    def test_update_comment_cannot_exceed_max_length(self):
+        self._set_up_private_project_with_comment()
+        payload = {
+            'data': {
+                'id': self.comment._id,
+                'type': 'comments',
+                'attributes': {
+                    'content': ('contentcontentcontentcontentcontentcontentcontentcontentcontentcontentcontentcontent'
```
sloria:
```python
''.join(['c' for c in range(max_length + 1)])
```

^ might make it easier to change this test if the max length ever changes.

api_tests/comments/views/test_comment_detail.py
```diff
@@ -193,6 +194,44 @@ def test_public_node_non_contributor_commenter_can_update_comment(self):
...
+                    + 'ntentcontentcontentcontentcontentcontentcontentcontentcontentcontentcontentcontentcontentconte'
+                    + 'ntcontentcontentcontentcontentcontentcontent'),
+                    'deleted': False
+                }
+            }
+        }
+        res = self.app.put_json_api(self.private_url, payload, auth=self.user.auth, expect_errors=True)
+        assert_equal(res.status_code, 400)
+        assert_equal(res.json['errors'][0]['detail'],
+                     'Ensure this field has no more than ' + str(osf_settings.COMMENT_MAXLENGTH) + ' characters.')
```
sloria: As per COSDev, prefer `str.format` to concatenation.

tests/test_models.py
```diff
@@ -4103,6 +4103,17 @@ def test_create(self):
         assert_equal(len(comment.node.logs), 2)
         assert_equal(comment.node.logs[-1].action, NodeLog.COMMENT_ADDED)
 
+    def test_create_sends_comment_added_signal(self):
```
sloria: :+1:

website/project/model.py
```diff
@@ -219,11 +219,15 @@ def find_unread(cls, user, node):
...
         if node.is_contributor(user):
+            if user.comments_viewed_timestamp is None:
+                user.comments_viewed_timestamp = {}
+                user.save()
             view_timestamp = user.comments_viewed_timestamp.get(node._id, default_timestamp)
             n_unread = Comment.find(Q('node', 'eq', node) &
                                     Q('user', 'ne', user) &
-                                    Q('date_created', 'gt', view_timestamp) &
-                                    Q('date_modified', 'gt', view_timestamp)).count()
+                                    Q('is_deleted', 'eq', False) &
```
sloria: Is it possible for `is_deleted` to be `None`? It might be safest to do `Q('is_deleted', 'ne', True)`.

website/static/js/comment.js
```diff
@@ -132,29 +132,61 @@ BaseComment.prototype.setupToolTips = function(elm) {
...
+        {'isCors': true});
+    request.done(function(response) {
+        self.author = {
+            'id': response.data.id,
+            'url': response.data.links.html,
+            'name': response.data.attributes.full_name,
+            'gravatarUrl': response.data.links.profile_image
+        };
+        return self.author;
+    });
```
sloria: As per COSDev, also add proper error handling.

website/static/js/comment.js
```diff
@@ -132,29 +132,61 @@ BaseComment.prototype.setupToolTips = function(elm) {
...
+            })
+        );
+        setUnreadCommentCount(self);
+        deferred.resolve(self.comments());
+        self._loaded = true;
+    });
     return deferred;
 };
 
+var setUnreadCommentCount = function(self) {
```
sloria: You shouldn't need to pass in `self` to a function. Instead, make this a method.

website/static/js/comment.js
```diff
@@ -132,29 +132,61 @@ BaseComment.prototype.setupToolTips = function(elm) {
...
+        self.comments(
+            ko.utils.arrayMap(response.data, function(comment) {
+                return new CommentModel(comment, self, self.$root);
+            })
+        );
+        setUnreadCommentCount(self);
+        deferred.resolve(self.comments());
+        self._loaded = true;
+    });
     return deferred;
```
sloria: This should return `deferred.promise()` so that callers can call `.then`, `.done`, etc.

website/static/js/comment.js
```diff
@@ -132,29 +132,61 @@ BaseComment.prototype.setupToolTips = function(elm) {
...
 };
 
+var setUnreadCommentCount = function(self) {
+    var request = osfHelpers.ajaxJSON(
+        'GET',
+        osfHelpers.apiV2Url('nodes/' + window.contextVars.node.id + '/', {query: 'related_counts=True'}),
+        {'isCors': true});
+    request.done(function(response) {
+        self.unreadComments(response.data.relationships.comments.links.related.meta.unread);
+    });
```
sloria: Methods that do async work should return a promise. In this case, you can just return `request` (jQuery XHRs implement the promise interface).

website/static/js/comment.js
```diff
@@ -187,7 +241,8 @@ BaseComment.prototype.submitReply = function() {
             self.$root.commented(true);
         }
         self.onSubmitSuccess(response);
-    }).fail(function() {
+    });
+    request.fail(function() {
         self.cancelReply();
         self.errorMessage('Could not submit comment');
```
sloria: Unexpected errors should be logged to Sentry with Raven.

website/static/js/comment.js
```diff
@@ -289,19 +360,33 @@ CommentModel.prototype.submitEdit = function(data, event) {
...
         self.editErrorMessage('');
         self.$root.editors -= 1;
         // Refresh tooltip on date modified, if present
         $tips.tooltip('destroy').tooltip();
-    }).fail(function() {
+    });
+    request.fail(function() {
         self.cancelEdit();
         self.errorMessage('Could not submit comment');
     });
```
sloria: See previous comment--return `request` here.

website/static/js/comment.js
```diff
@@ -338,13 +433,26 @@ CommentModel.prototype.startDelete = function() {
...
+                    }
+                }
+            }
+        });
+    request.done(function() {
         self.isDeleted(true);
         self.deleting(false);
-    }).fail(function() {
+    });
+    request.fail(function() {
```
sloria: Log unexpected errors to Sentry. Also, return `request`.

website/static/js/comment.js
```diff
@@ -359,12 +467,25 @@ CommentModel.prototype.startUndelete = function() {
...
+                        'deleted': false
+                    }
+                }
+            }
+        });
+    request.done(function() {
         self.isDeleted(false);
-    }).fail(function() {
+    });
+    request.fail(function() {
```
sloria: Log unexpected errors to Sentry. Return `request`.

website/static/js/comment.js
```diff
@@ -132,29 +132,61 @@ BaseComment.prototype.setupToolTips = function(elm) {
...
 
+BaseComment.prototype.fetchUserData = function() {
+    var self = this;
+    var request = osfHelpers.ajaxJSON(
+        'GET',
+        osfHelpers.apiV2Url('users/me/', {}),
+        {'isCors': true});
+    request.done(function(response) {
+        self.author = {
+            'id': response.data.id,
```
sloria: We can probably eliminate this request by using the data in `contextVars`. You could pass in a user to the constructor of `BaseComment`.

```js
new BaseComment({user: {id: .., url: .., name: .., gravatarURL: ...}});
```

sloria:
Pass finished.

sloria:
Also, as a heads up, support for view-only links has been added on `develop`.

samanehsan:
Done making those changes @sloria! :sunglasses:

api_tests/comments/views/test_comment_detail.py
```diff
@@ -193,6 +194,39 @@ def test_public_node_non_contributor_commenter_can_update_comment(self):
...
+                     'Ensure this field has no more than {} characters.'.format(str(osf_settings.COMMENT_MAXLENGTH)))
+
+    def test_update_comment_cannot_be_empty(self):
+        self._set_up_private_project_with_comment()
+        payload = {
+            'data': {
+                'id': self.comment._id,
+                'type': 'comments',
+                'attributes': {
+                    'content': '',
```
sloria: It would also be useful to test content that only has whitespace characters in it.

sloria: Ah, but I see you've done that in `TestNodeCommentsList#test_create_comment_trims_whitespace`.

api/comments/serializers.py
```diff
@@ -48,12 +52,31 @@ class CommentSerializer(JSONAPISerializer):
...
+
     def update(self, comment, validated_data):
         assert isinstance(comment, Comment), 'comment must be a Comment'
         auth = Auth(self.context['request'].user)
         if validated_data:
             if 'get_content' in validated_data:
-                comment.edit(validated_data['get_content'], auth=auth, save=True)
+                content = validated_data.pop('get_content')
+                content = content.strip()
+                if not content:
```
sloria: Can you set `required=True` instead of validating this manually? Or perhaps `allow_blank=False` and `allow_null=False`?

sloria: We should also have validation at the model layer. If `content` is not allowed to be blank, then a validator should be added to the `content` field.

samanehsan: I remember setting `required=True` and `allow_blank=False` but it still considered whitespace characters to be valid input.

api/comments/serializers.py
```diff
@@ -111,7 +134,10 @@ def create(self, validated_data):
                 detail='Invalid comment target \'{}\'.'.format(target_id)
             )
         validated_data['target'] = target
-        validated_data['content'] = validated_data.pop('get_content')
+        content = validated_data.pop('get_content')
+        validated_data['content'] = content.strip()
+        if not validated_data['content']:
```
sloria: See https://github.com/CenterForOpenScience/osf.io/pull/4702/files#r48058588

api/comments/serializers.py
```diff
@@ -48,12 +52,31 @@ class CommentSerializer(JSONAPISerializer):
...
         assert isinstance(comment, Comment), 'comment must be a Comment'
         auth = Auth(self.context['request'].user)
         if validated_data:
             if 'get_content' in validated_data:
-                comment.edit(validated_data['get_content'], auth=auth, save=True)
+                content = validated_data.pop('get_content')
+                content = content.strip()
+                if not content:
+                    raise ValidationError('Comment cannot be empty.')
+                comment.edit(content, auth=auth, save=True)
```
sloria: For consistency with the reset of the codebase, `comment.edit`, `comment.delete` etc. should do permissions checks, and `ValidationError` should be handled here.

api/comments/serializers.py
```diff
@@ -29,7 +30,7 @@ class CommentSerializer(JSONAPISerializer):
 
     id = IDField(source='_id', read_only=True)
     type = TypeField()
-    content = AuthorizedCharField(source='get_content')
+    content = AuthorizedCharField(source='get_content', max_length=osf_settings.COMMENT_MAXLENGTH)
```
sloria: As suggested in my other comments, this validation should also happen at the model layer.

sloria:
Pass finished.

samanehsan:
> We should also have validation at the model layer. If content is not allowed to be blank, then a validator should be added to the content field.

I added an ODM validator for checking that a field is not empty: https://github.com/CenterForOpenScience/modular-odm/pull/126

But then I found a `string_required` validator in `validators.py`: https://github.com/CenterForOpenScience/osf.io/blob/develop/framework/mongo/validators.py#L8

Should I use the `string_required` validator since it's already there?

samanehsan:
Do you mean that these methods should check that the user is able to edit, delete, etc.? 

>For consistency with the reset of the codebase, comment.edit, comment.delete etc. should do permissions checks, and ValidationError should be handled here.

sloria:
> Should I use the string_required validator since it's already there?

Yes.

> Do you mean that these methods should check that the user is able to edit, delete, etc.?

Yes, although I now see that my original suggestion may be incorrect; a PermissionsError--not a ValidationError--should be raised.

website/static/js/comment.js
```diff
@@ -166,16 +182,33 @@ BaseComment.prototype.submitReply = function() {
...
+            'data': {
+                'data': {
+                    'type': 'comments',
+                    'attributes': {
+                        'content': self.replyContent()
+                    },
+                    'relationships': {
+                        'target': {
+                            'data': {
+                                'type': self.id() === undefined ? 'nodes' : 'comments',
```
sloria: Minor: It's safer to check `self.id() ? 'comments' : 'nodes'`; that will ensure you get the correct value if `self.id()` is either `undefined` or `null`.

samanehsan: These lines get changed when I add in file comments: https://github.com/samanehsan/osf.io/blob/feature/file-comments-v2-merge/website/static/js/comment.js#L231-L232

samanehsan:
Done making those changes @sloria! :star:

sloria:
Closing in favor of https://github.com/CenterForOpenScience/osf.io/pull/4804

PR #4703, opened 2015-12-08
Remove dashboard node creation from depth_use count

brianjgeiger:
Purpose
-----------
Make depth_use statistics more accurate.
Closes https://openscience.atlassian.net/browse/OSF-5032

Changes
------------
In the analytics script, no longer count the creation of a dashboard collection towards the depth_use statistic if the dashboard creation is the user's first NodeLog, since all it really means is that the user logged in after some date in the Autumn of 2014. 

Running the script locally, before the change I had
```
fullname         email                logs
---------------  -----------------  ------
Brian J. Geiger  bgeiger@pobox.com      48
```

after the change
```
fullname         email                logs
---------------  -----------------  ------
Brian J. Geiger  bgeiger@pobox.com      47
```

Side effects
----------------
The user's activity points will be different from the analytics script that is run for depth use, but those numbers are never compared, so it's unlikely that it will be noticed by anyone who isn't explicitly looking for it. If we ever want to use activity points for anything meaningful, then a more robust change will have to be made.

The downside of this approach is that there are some people who will be counted incorrectly as depth users. Those would be easy to figure out. Also, there will be a little variation between the number of log entries in some cases. However, nobody cares about any of these cases. This should take a little longer to run than the version currently in use, but it shouldn't be a significant problem. If need be, we could either make this more accurate or a little faster, but this is probably a fine set of compromises for our current needs.

### DIFF
--- a/scripts/analytics/depth_users.py
+++ b/scripts/analytics/depth_users.py
@@ -28,8 +28,13 @@ def count_user_logs(user, query=None):
         query &= Q('user', 'eq', user._id)
     else:
         query = Q('user', 'eq', user._id)
-    items = [item for item in NodeLog.find(query) if item.action != 'project_created' or not item.node.is_dashboard]
-    return len(items)
+    logs = NodeLog.find(query)
+    length = logs.count()
+    if length > 0:
+        item = logs[0]
+        if item.action == 'project_created' and item.node.is_dashboard:
+            length -= 1
+    return length
 
 
 def get_depth_users(users):
--- a/scripts/analytics/depth_users.py
+++ b/scripts/analytics/depth_users.py
@@ -28,7 +28,8 @@ def count_user_logs(user, query=None):
         query &= Q('user', 'eq', user._id)
     else:
         query = Q('user', 'eq', user._id)
-    return NodeLog.find(query).count()
+    items = [item for item in NodeLog.find(query) if item.action != 'project_created' or not item.node.is_dashboard]
+    return len(items)
 
 
 def get_depth_users(users):


mfraezz:
The decrease in performance that this change incurs is prohibitive, please consider an alternative approach. :snail:
```bash
(remotes/upstream/develop) time python -m scripts.analytics.depth_users
[script output omitted]
python -m scripts.analytics.depth_users  1.98s user 0.41s system 83% cpu 2.869 total

(remotes/upstream/pr/4703) time python -m scripts.analytics.depth_users
[sciprt output omitted]
python -m scripts.analytics.depth_users  25.56s user 1.44s system 94% cpu 28.660 total
```

You can probably prevent the creation of these logs by adding logic to enable `suppress_log` around [here](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/project/model.py#L1362). Please also add a migration script to set the `should_hide` flag for each of these `NodeLog`s, which would allow this filtering to be done in the query.

CR complete :partly_sunny:

brianjgeiger:
Okay, optimized. I'll be interested to see the time difference in execution on the staging dataset.

mfraezz:
```bash
python -m scripts.analytics.depth_users  2.58s user 0.33s system 89% cpu 3.251 total
```

Significant improvement. From the output of the script, it looks like all but the least active users saw a decrement of 1. I still think a proper fix would be better, but I'm fine with this as is. Pass complete, moving to RfM.

brianjgeiger:
Well, by definition a proper fix would be *better*. But I think this is more appropriate to the use case, and the ticket's been sitting around for months.

PR #2534, opened 2015-04-20
[feature] comment pane

samanehsan:
Continuation of PR #1762 

Fixes #1085

PR #4707, opened 2015-12-09
[WIP] Feature: Allow users to comment on files

samanehsan:

Changes
------------
<b>OSF Changes:</b>

<b>API Changes</b>: 
- File serializer:


api/base/serializers.py
```diff
@@ -385,12 +394,23 @@ def get_url(self, obj, view_name, request, format):
...
-
+                    url = self.reverse(view, kwargs=kwargs, request=request, format=format)
+                    if self.filter:
+                        url += self.format_filter(obj)
+                    urls[view_name] = url
         if not urls['self'] and not urls['related']:
             urls = None
         return urls
 
+    def format_filter(self, obj):
```
cwisecarver: Just in the spirit of not reinventing the wheel, this might be a easier route to the same outcome. Plus, someone already wrote tests for it so you don't have to.

https://docs.djangoproject.com/en/1.9/ref/request-response/#django.http.QueryDict.update

QueryDicts are immutable by default but you can pass a mutable kwarg to them and then `qd.update(regular_dictionary)` will work.

Then `qd.urlencode()` will get you a properly formatted querystring that's urlencoded. You don't have to worry about the last & or any of that stuff just `"{}?{}".format(the_url, qd.urlencode())`

api/base/serializers.py
```diff
@@ -406,6 +426,16 @@ def to_representation(self, value):
         return ret
 
 
+class FileCommentRelationshipField(RelationshipField):
+    def __init_(self, **kwargs):
```
cwisecarver: If you're just calling super do you need to define this? Also, I don't think it's working because it's `__init_` instead of `__init__`.

api/comments/serializers.py
```diff
@@ -48,12 +56,31 @@ class CommentSerializer(JSONAPISerializer):
...
+        return user._id in obj.reports
+
+    def get_can_edit(self, obj):
+        user = self.context['request'].user
+        if user.is_anonymous():
+            return False
+        return obj.user._id == user._id
+
+    def get_has_children(self, obj):
+        return bool(Comment.find(Q('target', 'eq', obj)).count())
```
cwisecarver: I would change this to`return Comment.find(Q('target', 'eq', obj)).count() > 0` just to be more explicit. Also, what does count() do in the case of an exception? Is the exception going to bubble and kill the whole response or will it incorrectly return false?

samanehsan: If `obj` does not have any comments, `Comment.find(Q('target', 'eq', obj))` is an empty queryset and  `Comment.find(Q('target', 'eq', obj)) > 0` will return `False`.

website/project/model.py
```diff
@@ -215,84 +236,157 @@ def get_content(self, auth):
         return self.content
 
     @classmethod
-    def find_unread(cls, user, node):
+    def find_unread(cls, user, node, page=None, root_id=None, check=False):
```
cwisecarver: This might be better if it had count or something in the name. It's not obvious that it's finding the number of comments or the actual unread comments.

website/project/model.py
```diff
@@ -215,84 +236,157 @@ def get_content(self, auth):
...
+                            (Q('date_created', 'gt', view_timestamp) |
+                            Q('date_modified', 'gt', view_timestamp)) &
+                            Q('is_deleted', 'eq', False) &
+                            Q('root_target', 'eq', node)).count()
+
+    @classmethod
+    def n_unread_file_comments(cls, user, node):
+        n_unread = 0
+        for file_id in node.commented_files:
+            file_obj = File.load(file_id)
```
cwisecarver: Would it be better to do a `File.find(Q('_id', 'in', node.commented_files))` and then loop through that so that you're limiting the mongo queries?

framework/auth/core.py
```diff
@@ -1331,6 +1336,29 @@ def n_projects_in_common(self, other_user):
         """Returns number of "shared projects" (projects that both users are contributors for)"""
         return len(self.get_projects_in_common(other_user, primary_keys=True))
 
+    def get_node_comment_timestamps(self, node, page):
```
cwisecarver: This implementation seems overly confusing to me. I think I understand what it's doing but it seems to be getting there in a non-obvious way. Also, you define default_timestamp and then don't use it inside the page == 'files' block.

website/files/models/base.py
```diff
@@ -167,6 +168,10 @@ def get_guid(self, create=False):
                 return None
         return Guid.generate(self)
 
+    # used by django and DRF
+    def get_absolute_url(self):
```
cwisecarver: We've been using this pattern other places. Not saying it's the best pattern but it's a decent idea to keep it consistent.

```python
    @property
    def absolute_api_v2_url(self):
        return absolute_reverse('nodes:node-detail', kwargs={'node_id': self._id})

    # used by django and DRF
    def get_absolute_url(self):
```

I believe there are some things that call the absolute_api_v2_url directly.

website/project/model.py
```diff
@@ -215,84 +236,157 @@ def get_content(self, auth):
...
-                'project': self.node.parent_id,
-                'node': self.node._id,
-                'user': self.user._id,
-                'comment': self._id,
-            },
+            log_dict,
             auth=auth,
             save=False,
         )
+        self.node.save()
```
mfraezz: Should this be being called outside of the `if save:` block? You probably don't want a log to be added and saved if the comment edit won't be saved. Applies to `.delete` and `.undelete` too

website/project/views/node.py
```diff
@@ -716,7 +717,9 @@ def _view_project(node, auth, primary=False):
         for addon in node.get_addons():
             messages = addon.before_page_load(node, user) or []
             for message in messages:
-                status.push_status_message(message, kind='info', dismissible=False, trust=True)
+                status.push_status_message(message, 'info', dismissible=False, trust=True)
```
mfraezz: `kind` is a kwarg, please include the keyword in this call.

scripts/update_comments.py
```diff
@@ -0,0 +1,40 @@
...
+    comments = Comment.find()
+    for comment in comments:
+        comment.root_target = comment.node
+        comment.page = Comment.OVERVIEW
+        comment.save()
+
+
+if __name__ == '__main__':
+    script_utils.add_file_logger(logger, __file__)
+    main()
```
mfraezz: Please add a newline to the end of this file. Also, it didn't fail with staging data, but it may be good to allow a `--dry` flag to be passed in, and wrap the migration in a `TokuTransaction`:
```python
# Other imports
from framework.transactions.context import TokuTransaction

def main():
    update_comments_viewed_timestamp()
    update_comments()

# Methods

if __name__ == '__main__':
    dry = '--dry' in sys.argv
    if not dry:
        script_utils.add_file_logger(logger, __file__)
    init_app(routes=False, set_backends=True)
    with TokuTransaction:
        main()
        if dry:
            raise Exception('Dry Run -- Aborting Transaction')

```

mfraezz:
UI Bug: When there are enough comments to overflow the comment pane, the size calculation is off in such a way that the last comment cannot be seen:
screen shot 2015-12-18 at 16 55 51

Similar bug exists on prod, but only the reply/report/delete buttons are cut off:
screen shot 2015-12-18 at 16 58 14

website/project/model.py
samanehsan: You're totally right...Do you think we even need the `save` flag for comments? The `create` method doesn't have one.

framework/auth/core.py
cwisecarver: Would this work?
```python
    def get_node_comment_timestamps(self, node, page):
        """ Returns the timestamp for when comments were last viewed on a node or
            a dictionary of timestamps for when comments were last viewed on files.
        """
        default_timestamp = dt.datetime(1970, 1, 1, 12, 0, 0)

        timestamps = self.comments_viewed_timestamp.get(node._id,  {})
        if page == 'node':
            page_timestamps = timestamps.get(page, default_timestamp)
        else if page == 'file':
            page_timestamps = timestamps.get(page, {})
            for file_id in node.commented_files:
                page_timestamps[file_id] = default_timestamp
        self.save()
        return page_timestamps
```

samanehsan: I think that should do it! I think I can also remove setting the timestamps for individual files here.

samanehsan:
Done responding to code review! @mfraezz @cwisecarver :octocat: :snowflake:

mfraezz:
Previous UI bug uncovered an issue with API pagination as well -- only the more recent 10 comments will load. Otherwise, changes to what I covered look good :corn:

samanehsan:
I was not supposed to merge develop into this branch! :scream: I opened a new PR to the `release/0.61.0` branch without the merge commit: https://github.com/CenterForOpenScience/osf.io/pull/4806

PR #4708, opened 2015-12-09
[Hotfix] Registration Language

samchrisinger:
# Purpose
see: https://openscience.atlassian.net/browse/OSF-5348

# Changes
- only show "You can start a new registration..." text if user is a project admin


samanehsan:
Looks good to me @samchrisinger :octocat: I think this is good to merge.

PR #4711, opened 2015-12-10
[Fix] Prevent `addon_added` NodeLogs during migration

mfraezz:
Purpose
=======
When overriding the creation of an addon for a node, a `NodeLog` with no auth (and consequently, no template) is generated. 

Changes
=======
* prevent log creation

Side Effects
=========
None

### DIFF
--- a/scripts/dropbox/migrate_to_external_accounts.py
+++ b/scripts/dropbox/migrate_to_external_accounts.py
@@ -126,6 +126,7 @@ def make_new_node_settings(node, node_settings_document, external_account=None, 
     node_settings_instance = node.get_or_add_addon('dropbox', auth=None, override=True)
     node_settings_instance.folder = node_settings_document['folder']
     node_settings_instance.save()
+    suppress_log(node)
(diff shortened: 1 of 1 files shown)


sloria:
Rather than hiding the created log, can you modify `add_addon`/`get_or_add_addon` to prevent the log from being created in the first place?

samchrisinger:
:+1:, see `Node#set_visible`

sloria:
@samchrisinger That's for bibliographic contributors; here we're trying to prevent a log from being created for adding an addon.

samchrisinger:
@sloria I mean the `log=True` kwarg pattern for optionally disabling logs.

sloria:
Ah, ok. Good call.

mfraezz:
Yeah, I realized it was using the `add_addon` *with* a `log` kwarg a little while ago. Testing has shown it to work locally for me

sloria:
Cherry-picked this change into `release/0.60.0`.

PR #4718, opened 2015-12-11
[WIP] Updates to User Login and Invites

zkraime:
<h1>Purpose: </h1> Miscellaneous fixes and additions

<b> To do 
- [x] Custom email templates
- [x] Password length restriction
- [x] Fix UI errors
     - [x] Login view cutoff
     - [x] Restrict home page by authenticated (fix "Logged in as Sign Out" bug) 
     - [x] Add password change page for users
     - [x] Add forgot password link
     - [x] Display proper Django message class type (current 'patch' handles only success and error)
     - [ ] Handle warning/failure cases for send_email_invitation in Django-admin
- [x] Set up OSF email server for admin app
         - Needs to be tested on staging


admin/common_auth/forms.py
```diff
@@ -15,3 +16,11 @@ def __init__(self, *args, **kwargs):
         super(CustomUserRegistrationForm, self).__init__(*args, **kwargs)
         self.fields['first_name'].required = True
         self.fields['last_name'].required = True
+
+    def clean_password1(self):
```
Ghalko: While it is certainly good to enforce password length, doesn't Django already do this somewhere?

admin/templates/base.html
```diff
@@ -55,6 +55,9 @@
                 </a>
               </li>
               <li>
+                <a href="{% url 'password_change' %}" class="btn btn-flat">Change password</a>
```
Ghalko: I know this is probably not a good thing to skip, but we should probably have a different base page and user page where admin users can update their information.

So for now a button on the base page will be okay, but we should add a ticket for user settings.

admin/templates/common_auth/registration/invitation_email.html
```diff
@@ -0,0 +1 @@
+{% extends 'registration/password_reset_email.html' %}
```
Ghalko: EOF.

Since some of these emails will be seen by more than just COS we should consider following the OSF in language used within the email.

admin/templates/registration/password_reset_done.html
```diff
@@ -0,0 +1,20 @@
...
+{% block title %}{{ title }}{% endblock %}
+{% block content_title %}<h1>{{ title }}</h1>{% endblock %}
+{% block content %}
+<div>
+<p>{% trans 'Your password was changed.' %}</p>
+</div>
+<div>
+<a href="{% url 'auth:login' %}">Return to home</a>
+</div>
+{% endblock %}
```
Ghalko: EOF

admin/templates/registration/password_reset_email.html
```diff
@@ -0,0 +1,15 @@
...
+{{ protocol }}://{{ domain }}{% url 'password_reset_confirm' uidb64=uid token=token %}
+{% endblock %}
+
+{% trans "Your account email is:" %} {{ user.get_username }}
+
+{% trans "Thank you." %}
+
+{% blocktrans %}-OSF Admin{% endblocktrans %}
+
+{% endautoescape %}
```
Ghalko: EOF

admin/templates/base.html
zkraime: @Ghalko I chatted with Chris about this briefly - that's exactly what I was thinking, having a user-settings page for updating account information. Same conclusion - leave it as is for now but keep it noted

admin/common_auth/forms.py
zkraime: @Ghalko Chris enlightened me on model-level character limits, so I will use that for password restriction instead of a form-level validation

admin/templates/common_auth/registration/invitation_email.html
zkraime: @Ghalko I would absolutely agree. The current override is arbitrary (but not totally generic). I need to find an OSF template to model after

sloria:
Is there a JIRA ticket associated with this?

zkraime:
@sloria Actually no, I hadn't created one yet, I will do that

zkraime:
@cwisecarver Give this a quick pass? 
I have yet to test the emailing, but I asked Michael for the sendgrid password so hopefully I'll get that soon. 
The other thing is handling unsuccessful email invites on admin; I wanted to ask you about this real quick in person if you get the chance.
Thanks!

admin/base/settings/defaults.py
```diff
@@ -28,6 +29,14 @@
 # Email settings. Account created for testing. Password shouldn't be hardcoded
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
```
cwisecarver: Just a note for devops, this should be set to `'django.core.mail.backends.smtp.EmailBackend'` in the > dev local.py.

admin/base/urls.py
```diff
@@ -16,8 +16,14 @@
                          url(r'^auth/', include('admin.common_auth.urls', namespace='auth')),
                          url(r'^reset/(?P<uidb64>[0-9A-Za-z_\-]+)/(?P<token>[0-9A-Za-z]{1,13}-[0-9A-Za-z]{1,20})/$',
                             views.password_reset_confirm_custom, name='password_reset_confirm'),
-                         url(r'^reset/done/$', views.password_reset_done,
-                            name='password_reset_complete'),
+                         url(r'^resetpassword/$', 'django.contrib.auth.views.password_reset', name='reset_password'),
+                         url(r'^password_reset_done/$', 'django.contrib.auth.views.password_reset_done', name='password_reset_done'),
+                         url(r'^accounts/password_change/$',
```
cwisecarver: In the spirit of not hardcoding urls anywhere you don't have to I changed these lines to 
```python
                         url(r'^accounts/password_change/$',
                            'django.contrib.auth.views.password_change',
                            {'post_change_redirect': reverse_lazy('password_change_done')},
                            name="password_change"),
                         url(r'^accounts/password_change/done/$',
                            'django.contrib.auth.views.password_change_done', {'template_name': 'registration/password_change_done.html'}, name='password_change_done'),
                         )
```
And then imported `from django.core.urlresolvers import reverse_lazy` and everything still worked.

admin/common_auth/admin.py
```diff
@@ -39,10 +39,6 @@ def send_email_invitation(self, request, queryset):
     send_email_invitation.short_description = 'Send email invitation to selected users'
 
     def save_model(self, request, obj, form, change):
```
cwisecarver: You can just remove this method. You're not doing anything the grandfather class method isn't doing.

admin/base/views.py
```diff
@@ -5,14 +5,13 @@
 from django.utils.http import urlsafe_base64_decode
 from admin.common_auth.models import MyUser
 
-
 @login_required
 def home(request):
     context = {'user': request.user}
     return render(request, 'home.html', context)
 
 def password_reset_done(request, **kwargs):
```
cwisecarver: I don't think you're using this anywhere. At least not that pycharm find usages or the silver searcher could find.

admin/base/views.py
```diff
@@ -5,14 +5,13 @@
...
-    messages.success(request, 'You have successfully reset your password and activated your admin account. Thank you')
+    messages.success(request, 'You have successfully reset your password and activated your admin account.')
     return redirect('auth:login')
 
 def password_reset_confirm_custom(request, **kwargs):
```
cwisecarver: I know this method isn't part of this PR but I can't figure out what it does differently than [the regular one](https://docs.djangoproject.com/en/1.9/topics/auth/default/#django.contrib.auth.views.password_reset_confirm)

cwisecarver:
Done with this pass @Ghalko, @zkraime. Whoever picks this up let me know if there are questions.

admin/base/urls.py
zkraime: Thanks

admin/base/views.py
zkraime: The login_required decorator? Or the whole 'home' method?

admin/base/views.py
zkraime: The only difference is that it is setting the confirmed flag to true

admin/common_auth/admin.py
zkraime: The save_model method? 
Yeah you're right

Ghalko:
@zkraime What is the status of this?

admin/base/views.py
cwisecarver: It's been a while but I think I was talking about the password_reset_done method. Interestingly, it does seem to be used in the urls file. Not sure why the silver searcher didn't find it...

zkraime:
@cwisecarver Please do a really quick pass over this to make sure I addressed your comments? 

Also, not sure how to go about handling warning and failure cases for email invites so I will leave that be for now since it's lower priority than other work (it's the one unchecked box)

cwisecarver:
Looks good to me @zkraime. Make sure you get with @mattclark when this goes to this being ready-to-merge about the email settings that will need to be put in local.py.

sloria:
As mentioned on the JIRA ticket: please update this branch with develop and resend the PR to develop. Thanks!

PR #4698, opened 2015-12-04
Prereg Followup

samchrisinger:
# Purpose
- Update RegistrationEditor.authorImport to use ContribAdder. This creates a full coupling between the 'authorship' field of the Prereg Challenge schema and the Node's bibliographic contributors. 
- Add an Archiver hook to run once files are finished copying. The RegistrationEditor.FilePicker makes references to files on the unregistered Node, and these must be updated to reference the registration Node's copies.

# Tickets
- https://openscience.atlassian.net/browse/OSF-4975
- https://openscience.atlassian.net/browse/OSF-5319

PR #4719, opened 2015-12-11
Feature: More Prereg Uploaders

samchrisinger:
# Depends on:
https://github.com/CenterForOpenScience/osf.io/pull/4698

# Purpose 
see https://openscience.atlassian.net/browse/OSF-5335

# Changes
The most significant change here is modifying Question properties to be an Array (ordered) rather than an Object (unordered). 

Additionally some logic was added to ensure when multiple uploaders are included on a single page that their corresponding DOM containers have unique ids. 

# Side Effects
This will break some questions for existing DraftRegistrations. No exception will be raised, but previously answered questions may appear empty. The only affected submissions will be for the Prereg Challenge and since we're in a beta period, this is acceptable (and lets us avoid running a migration).


website/static/js/registrationEditorExtensions.js
```diff
@@ -159,17 +160,23 @@ ko.bindingHandlers.osfUploader = {
     init: osfUploader
 };
 
+
+var UPLOADER_COUNT = 0;
```
sloria: This isn't a constant; it gets updated on ~L171. Make this `lowerCamelCase`.

PR #4514, opened 2015-10-26
[feature] alternative citations

ticklemepierce:
This feature allows for admins on a project to create alternative citations that display underneath of the MLA/APA/Chicago citations that are provided by default.  This feature covers the following use cases:

 * I made an OSF project to include materials as a supplement to my journal article and would like any citations of those materials to contribute to citations of my paper
 * I made an OSF project and then was published, giving me a DOI that I want to use as a replacement for the citation provided by the OSF
 * I made an OSF project that has components that each represent a study in a paper, and I would like each of those components to have the same citation
 * I haven't figured out how to change my name in my profile and want to edit the citation to reflect the name I usually use to publish with. 
 * I have a publication that I would like visitors on the OSF to be able to find, and correcting the citation for my OSF project to match the citation would increase its relevance in search.

PR #4721, opened 2015-12-11
[Feature] Alternative Citations Logs

ticklemepierce:
Purpose
-
Add Create/Edit/Remove logs for alternative citations

Changes
-
Moved some code for this into model.py and then added the necessary calls to create logs.
Added appropriate templates to log_templates.mako

Side Effects
-
Should not be any

Notes
-
Depends on https://github.com/CenterForOpenScience/osf.io/pull/4514

[#OSF-5332]


sloria:
Users should not need to be introduced to the concept of an "alternative citation". In the user's mind, they are just "citations".

The logs should read: 

> Rosa updated a citation name from *My citation* to *My Citation*.

Erin M has approved this change.

ticklemepierce:
> Users should not need to be introduced to the concept of an "alternative citation". In the user's mind, they are just "citations".

Should I also update the button on the main PR to say "Add Citation" instead of "Add Alternative Citation" (and anywhere else this language may occur)?

sloria:
@ticklemepierce Yes.

website/project/model.py
```diff
@@ -2162,6 +2166,64 @@ def add_tag(self, tag, auth, save=True, log=True):
             if save:
                 self.save()
 
+    def add_citation(self, auth, save=True, log=True, **kwargs):
```
sloria: Minor: in general, we try to avoid implicit saves; I suggest changing the default for `save` to False.

Note: Saving the `AlternativeCitation` implicitly is OK because it is a necessary side effect. The implicit `self.save` is what I'm referring to.

website/project/model.py
```diff
@@ -2162,6 +2166,64 @@ def add_tag(self, tag, auth, save=True, log=True):
...
+                    'citation': citation.name
+                },
+                auth=auth,
+                save=False
+            )
+            if save:
+                self.save()
+        return citation
+
+    def edit_citation(self, auth, instance, save=True, log=True, **kwargs):
```
sloria: See previous comment about implicit saves.

website/project/model.py
```diff
@@ -2162,6 +2166,64 @@ def add_tag(self, tag, auth, save=True, log=True):
...
+    def add_citation(self, auth, save=True, log=True, **kwargs):
+        citation = AlternativeCitation(**kwargs)
+        citation.save()
+        self.alternative_citations.append(citation)
+        if log:
+            self.add_log(
+                action=NodeLog.CITATION_ADDED,
+                params={
+                    'node': self._primary_key,
+                    'citation': citation.name
```
sloria: We should probably store both the name and the text of the citation, just in case we ever decide to add citation text to logs.

```python
'citation': {'name': ..., 'text': ...}
```

Same goes for the CITATION_REMOVED logs.

website/project/model.py
```diff
@@ -2162,6 +2166,64 @@ def add_tag(self, tag, auth, save=True, log=True):
...
+                },
+                auth=auth,
+                save=False
+            )
+            if save:
+                self.save()
+        return citation
+
+    def edit_citation(self, auth, instance, save=True, log=True, **kwargs):
+        citation = {'name': instance.name}
```
sloria: Why not include the `text` here as well?

sloria:
Pass finished.

sloria:
I had added to my comment in https://github.com/CenterForOpenScience/osf.io/pull/4721#issuecomment-164490417 with the text for "edit" and "remove" logs.

I can make those changes when I merge.

website/templates/project/project.mako
```diff
@@ -309,7 +309,7 @@
                     </div>
                     % if 'admin' in user['permissions'] and not node['is_registration']:
                         <!-- ko ifnot: editing() -->
-                            <button data-bind="ifnot: editing(), click: addAlternative" class="btn btn-default m-t-md"><i class="fa fa-plus"></i> Add Alternative Citation</button>
+                            <button data-bind="ifnot: editing(), click: addAlternative" class="btn btn-default m-t-md"><i class="fa fa-plus"></i> Add Citation</button>
```
sloria: As per the style guide, use sentence case. I will change this when I merge.

sloria:
The color of the cancel button is inconsistent with the rest of the site. It should be light grey.

osf___alternative_citations

PR #4727, opened 2015-12-11
Make Prereg Admin readonly (temporarily)

samchrisinger:
We've decided to do an incremental release of a read-only version of the Prereg admin app. This eliminates any danger via writes to the prod database.


sloria:
@samchrisinger Looks like flake checks are failing. Can you please resolve those? Once that's done, this should be good to merge.

samchrisinger:
@sloria we're not going to need this given that we're waiting for the 60 release.

PR #4728, opened 2015-12-11
Remove ability to serialize registrations through /nodes/

brianjgeiger:
Purpose
-----------
Closes https://openscience.atlassian.net/browse/OSF-5359

Changes
------------
Removes ability to access registrations through a /nodes/ endpoint. Updates a bunch of tests (many of which changing an expected 403 to a 404)

Side effects
----------------
Node lists will be *much* faster


brianjgeiger:
I am reasonably certain the failing JS test is not because of my code.

sloria:
I'll restart the build.

brianjgeiger:
Those failing tests definitely look like something that could be mine, but I'd run API tests before submitting and I didn't see those errors on my setup. Trying again, and I'll look closer even if they do pass locally.

api/nodes/views.py
```diff
@@ -245,11 +246,7 @@ def get_default_odm_query(self):
 
     def filter_non_retracted_nodes(self, query):
```
mfraezz: This method is superfluous now that we're no longer filtering out retractions here.

brianjgeiger:
Failing tests are due to Alternative Citations not implementing Registrations endpoints and only testing against nodes.

sloria:
Please resend to `release/0.60.0`.

brianjgeiger:
Closing in favor of #4755

PR #4730, opened 2015-12-14
[Feature] Add fork ability to alternative citations

ticklemepierce:
Purpose
-
With my last improvement (adding log statements to alternative citations) I discovered that when a fork was created, it would show all relevant alternative citations logs but would not copy the citations.  This seems to be contradictory so I decided the best fix would be to copy all citations from the original to the fork.

Changes
-
In model.py I added a loop to iterate through each citation from the original and clone it and add it to the forked project.  This also required a small change to my node.add_citation function from the [logs PR][1].  

Side Effects
-
This is a small change so there should not be any, however the ones that could occur would be a shallow copy in which the forked alternativeCitations list contains references to the same citations as the original rather than copies.  This would mean that editing/removing a citation in one would affect the corresponding citation in the other, which should not be the case.

The other potential side effect would be that the node.add_citation method does not work properly, however this should be caught by unit tests and does not appear to be a problem.

Notes
-
As semi-referenced earlier, this depends on my previous PR for [alternative citations logs][1]

[1]: https://github.com/CenterForOpenScience/osf.io/pull/4721


sloria:
Would be good to have a models test for this change, but this can happen in a follow up PR.

PR #4731, opened 2015-12-14
Feature/modify registration serializer fields

pattisdr:
# Purpose

Issue: https://openscience.atlassian.net/browse/OSF-5398
Makes a few small changes to RegistrationSerializer.  One is an error found by Nici in https://openscience.atlassian.net/browse/OSF-5036.

# Changes

1) Adds field embargo_end_date.  If no embargo_end_date, returns null
2) Changes source of pending_registration_approval to match new registration changes and name to pending_registration.
3) Updates help text.

Embargo, awaiting approval by project admins:
screen shot 2015-12-15 at 4 00 11 pm

Embargo, approved by project admins:
screen shot 2015-12-15 at 4 01 31 pm

### DIFF
--- a/api/registrations/serializers.py
+++ b/api/registrations/serializers.py
@@ -9,19 +9,22 @@ from api.base.serializers import IDField, RelationshipField, LinksField, HideIfR
 
 class RegistrationSerializer(NodeSerializer):
 
-    retracted = ser.BooleanField(source='is_retracted', read_only=True,
-                                 help_text='Whether this registration has been retracted.')
-    date_registered = ser.DateTimeField(source='registered_date', read_only=True, help_text='Date time of registration.')
-    embargo_end_date = HideIfRetraction(ser.SerializerMethodField(help_text='When will embargo be lifted?'))
-    retraction_justification = ser.CharField(source='retraction.justification', read_only=True)
+    pending_embargo_approval = HideIfRetraction(ser.BooleanField(read_only=True, source='is_pending_embargo',
+        help_text='The associated Embargo is awaiting approval by project admins.'))
+    pending_registration_approval = HideIfRetraction(ser.BooleanField(source='is_pending_registration', read_only=True,
+        help_text='The associated RegistrationApproval is awaiting approval by project admins.'))
     pending_retraction = HideIfRetraction(ser.BooleanField(source='is_pending_retraction', read_only=True,
-        help_text='Is this registration pending retraction?'))
-    pending_registration_approval = HideIfRetraction(ser.BooleanField(source='sanction.is_pending_approval', read_only=True,
-        help_text='Does this registration have a sanction pending approval?'))
-    registered_meta = HideIfRetraction(ser.SerializerMethodField(
-        help_text='A dictionary with embargo end date, whether registration choice was immediate or embargoed,'
-                  ' and answers to supplemental registration questions'))
+        help_text='The registration is awaiting retraction approval by project admins.'))
+    retracted = ser.BooleanField(source='is_retracted', read_only=True,
+                                 help_text='The registration has been retracted.')
+
+    date_registered = ser.DateTimeField(source='registered_date', read_only=True, help_text='Date time of registration.')
+    embargo_end_date = HideIfRetraction(ser.SerializerMethodField(help_text='When the embargo on this registration will be lifted.'))
+
+    retraction_justification = ser.CharField(source='retraction.justification', read_only=True)
     registration_supplement = ser.SerializerMethodField()
+    registered_meta = HideIfRetraction(ser.SerializerMethodField(
+        help_text='A dictionary with supplemental registration questions and responses.'))
 
     registered_by = HideIfRetraction(RelationshipField(
         related_view='users:user-detail',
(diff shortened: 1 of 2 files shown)


samchrisinger:
@pattisdr with regard to 3) above, there is a subtle difference between `is_pending_embargo` and `is_pending_registration`. 

`is_pending_embargo` implies that the registration has an associated Embargo that has not yet been approved by project admins
`is_pending_registration` implies that the registration has an associated RegistrationApproval that has not yet been approved ... 

Does this make sense? I'm happy  to talk more if it is unclear.

h/t @mfraezz for point this out

api/registrations/serializers.py
```diff
@@ -12,13 +12,12 @@ class RegistrationSerializer(NodeSerializer):
     retracted = ser.BooleanField(source='is_retracted', read_only=True,
                                  help_text='Whether this registration has been retracted.')
     date_registered = ser.DateTimeField(source='registered_date', read_only=True, help_text='Date time of registration.')
+    embargo_end_date = HideIfRetraction(ser.SerializerMethodField(help_text='When will embargo be lifted?'))
```
mfraezz: Please change `help_text` to be more descriptive. @brianjgeiger suggested `When the embargo on this registration will be lifted.`

Also, to keep `help_text` consistent across all serializers in the API, could you please change the three with questions to have statements (`pending_retraction`, `pending_registration_approval`, and `pending_embargo`, which should be re-added.)

pattisdr: @samchrisinger - I understand that they have different meanings in the background but since pending_registration_approval and pending_embargo give the same value, I didn't think that pending_embargo was necessary:

If I create a registration and put an embargo on it, but haven't clicked the link in the email approving the embargo, pending_embargo=True and pending_registration_approval=True.  Once I click the link in the email approving the embargo, pending_embargo=False and pending_registration_approval=False.   So I thought pending_registration_approval could handle everything.

samchrisinger: To be clear, we're talking specifically about `Node#is_pending_embargo` and `Node#is_pending_registration`? I may missing it, but don't see `pending_registration_approval` anywhere. 

It would be pretty concerning to me if these values are consistently the same (although it could just be a bug in the factory or test case). Do you think you could write a test to document this behavior?

mfraezz: One of the serializer fields in question (`pending_embargo`) comes from the property `Node.is_pending_embargo`, which does a check on the nodes/parents associated `Embargo` object. 

The field `pending_registration_approval` should perhaps come from `Node.is_pending_registration`, which does a check on the nodes/parents associated `RegistrationApproval` object, but it actually comes from `Node.sanction.is_pending_approval`, based on the `Sanction` associated with a given `Node`.

pattisdr: Ah - I see.  `pending_registration_approval` is what I was calling the field in the serializer.  My source was wrong.  It was `sanction.pending_approval`, not `is_pending_registration`.  Apparently I don't understand sanctions.

If the source is corrected, then the values are different and everything works great. I will correct the source and restore the pending_embargo field.  Thanks @samchrisinger

samchrisinger: +1. Thank you both.

mfraezz:
Thanks working with me on this one, @samchrisinger 

Pass complete :dragon:. This is looking good. After few minor language changes, adding back the removed functionality, and correcting the source `pending_registration_approval`, this should be ready to merge. 

Aside: I don't think it should block this PR, as this is a fix, but after reviewing another registrations-touching API PR from @brianjgeiger, we determined that the `/registrations/` endpoint is under-tested (created [follow-up ticket](https://openscience.atlassian.net/browse/OSF-5413) for testing improvements). Much of that work will be refactoring code that tests registrations out of the `/node/` tests, but there are also some that need to be written. If you compelled to, adding a few tests to ensure the registrations serializer returns what it should when it should (not only that it doesn't return what it shouldn't when it shouldn't, as `test_retractions_display_limited_fields` does) would be beneficial.

api/registrations/serializers.py
```diff
@@ -9,20 +9,22 @@
 
 class RegistrationSerializer(NodeSerializer):
 
+    pending_embargo_approval = HideIfRetraction(ser.BooleanField(read_only=True, source='is_pending_embargo',
+        help_text='The associated Embargo is awaiting approval by project admins.'))
+    pending_registration_approval = HideIfRetraction(ser.BooleanField(source='is_pending_registration', read_only=True,
+        help_text='The associated RegistrationApproval is awaiting approval by project admins.'))
+    pending_retraction = HideIfRetraction(ser.BooleanField(source='is_pending_retraction', read_only=True,
+        help_text='The registration is awaiting retraction approval by project admins.'))
```
mfraezz: I think this should read `The retraction [...]`, not `The registration [...]`.

mfraezz:
LGTM. Confirmed with @brianjgeiger that language changes were acceptable.

api/registrations/serializers.py
```diff
@@ -95,6 +97,11 @@ def get_registered_meta(self, obj):
                 return meta_values
         return None
 
+    def get_embargo_end_date(self, obj):
+        if obj.embargo_end_date:
```
sloria: Is there any reason you couldn't do:

```python
embargo_end_date = ser.DateTimeField(allow_null=True, read_only=True)
```

I imagine this would improve the auto-documentation of the serializer.

sloria:
Just one question above. Once that is addressed, this is good to merge.

api/registrations/serializers.py
pattisdr: if there is no embargo_end_date, False is returned.  False can't appear in a date-time field.

sloria: In that case, you could do 

```python
obj.embargo_end_date or False
```

But what you have here is fine.

PR #4734, opened 2015-12-14
Disable user function

hmoco:


framework/auth/core.py
```diff
@@ -977,6 +977,22 @@ def get_activity_points(self, db=None):
         db = db or framework.mongo.database
         return analytics.get_total_activity_count(self._primary_key, db=db)
 
+    def disable_user(self):
+        import mailchimp
```
sloria: You should be able to import this at the top of the file. Follow the COSDev guidelines for import order.

framework/auth/core.py
```diff
@@ -977,6 +977,22 @@ def get_activity_points(self, db=None):
         db = db or framework.mongo.database
         return analytics.get_total_activity_count(self._primary_key, db=db)
 
+    def disable_user(self):
```
sloria: Minor: `user.disable_user()` is redundant, since this is an instance method; this can just be `disable()` or `disable_account()`.

framework/auth/core.py
```diff
@@ -977,6 +977,22 @@ def get_activity_points(self, db=None):
...
+            list_id = mailing_list['data'][0]['id']
+            m.lists.unsubscribe(id=list_id, email={'email': self.username}, send_goodbye=True)
+
+        # Update mailing_list user field
+        if self.mailchimp_mailing_lists is None:
+            self.mailchimp_mailing_lists = {}
+
+        self.mailchimp_mailing_lists[settings.MAILCHIMP_GENERAL_LIST] = False
+        self.is_disabled = True
+        self.save()
```
sloria: In general, we try to avoid implicit saves. To be consistent with other methods, you could add a `save` parameter that defaults to `False`.

sloria:
There's a lot of duplication with the code here: https://github.com/CenterForOpenScience/osf.io/blob/56edc41ab9ad7550f65520c497bebf9c09fba331/website/mailchimp_utils.py#L79-L89 . Can you separate that code out into a utility function (`unsubscribe_from_mailchimp_list`) so that it can be reused?

sloria:
Pass finished.

PR #4735, opened 2015-12-14
Feature/api/do not clear node tags on patch

brianjgeiger:
Purpose
-----------
Keep tags from clearing out when someone does a PATCH on /v2/nodes/{{node_id}}/ when no tags are included in the patch. Closes https://openscience.atlassian.net/browse/OSF-5403

Changes
------------
 - Check for the partial update condition in the update method and, in that case, make the tags equal to the original tags
 - Add a test for this situation

Side effects
-----------------
None


sloria:

PR #4739, opened 2015-12-15
Home route

hmoco:


website/templates/home.mako
```diff
@@ -0,0 +1,5 @@
+<%inherit file="base.mako"/>
+<%def name="title()">Home</%def>
+<%def name="content()">
+    <h1>Placeholder</h1>
+</%def>
```
caneruguz: This placeholder section will change obviously; further work will be done here but for this PR this is accurate.

caneruguz:
Looks good to me. @sloria can you do a CR2 on this?

website/routes.py
```diff
@@ -173,6 +173,8 @@ def make_url_map(app):
 
     process_rules(app, [
 
+        Rule('/home/', 'get', website_views.home, OsfWebRenderer('home.mako')),
```
sloria: Ping @JeffSpies for approval of this new route.

sloria:
Failing test is unrelated to this change; I've restarted the build. The new route will need Jeff's review before this is merged.

@caneruguz If this is a blocker, go ahead and work off of this branch for now.

website/routes.py
JeffSpies: What is this route for? How is different than what we have?

Sent and possibly dictated from my iPhone

hmoco: This is a placeholder for the new "dashboard" page, for Caner's work, while the current dashboard is going to be replaced with "My Projects". When a user logs in or goes to osf.io/, this page will be seen instead of dashboard, with a link to "My Projects". Pinging @caneruguz for a better explanation.

sloria: What will be on `/dashboard/`?

hmoco: "My Projects", maybe the route will be renamed.

sloria: We need to maintain backwards compat for all web routes, so we can't rename.

How about 

`/dashboard/` -> New dashboard
`/dashboard/projects/` -> My projects

?

hmoco: That would be something that needs to be discussed with Caner

JeffSpies: @caneruguz

Sent and possibly dictated from my iPhone

caneruguz: When we say we need to maintain old routes does that mean "osf.io/" has to go to dashboard?

How about this:
"osf.io/" goes to what we call here "home.mako"  when logged in and index when not
"osf.io/myprojects" and "osf.io/dashboard"  goes to the "dashboard.mako"

If we just use "osf.io" for home we move away from calling it something like dashboard or home, which gives flexibility for future too. I don't know if there are technical issues about it.

caneruguz:
Hey @JeffSpies , we need a final solution to this to go forward. I talked with Steve, here's my suggestion for new and old routes. 

"osf.io/" goes to -> "home.mako" when logged in and -> index.mako when not
"osf.io/myprojects" ->  goes to the "dashboard.mako"
for old links 
"osf.io/dashboard" goes to ->  "osf.io/"

JeffSpies:
Can you describe these routes in english rather than mako? :) (We’ll need that for the community team as well.)

We’ll need buy-in from @sloria1 and @brianjgeiger as well.

J.

caneruguz:
Good point @JeffSpies 

I'm not fully confident about terminology. It would be good to discuss. The question is whether we keep using the word "Dashboard" to mean the home page. 
Arguments for Dashboard:
- It's the page after login where the view changes and they have an initial look at the inside
- The page before login will be the home page for most people, they don't think of it as a landing page so home page will refer to that page. 

Arguments Against Dashboard:
- Dashboard is used to mean a place where quick views on activity, monitoring, progress etc. type of information is displayed
- If not already familiar with the above image, many users may not associate dashboard with anything. 
- Although there is a login page, for most logged in users this is the main page and hence the "home" page that they see
- We are limiting ourselves to what we can use this entry page for if we call it a dashboard.

My inclination is to not use the word dashboard but also not label things as home but we may refer to this page as "Home page" in our documentation for how to material.

sloria:
OK, I propose  the following routes and terminology for the Phase 0 release:

`/` is the *home* route and is accessed by clicking the navbar logo in the top-left corner. When logged out, it goes to the current landing page. When logged in, it goes to the user's *home page*, which has a Quick Search widget and "New and Noteworthy" projects. 

`/home/` should be removed. Ticket: https://openscience.atlassian.net/browse/OSF-5605

`/dashboard/` is the *dashboard* route and can be accessed by clicking "Dashboard" in the navbar. It contains the *user dashboard*, which is for browsing and organizing the users' projects.

brianjgeiger:
Works for me

JeffSpies:
Does this account for the new dashboard? What will the homepage be when the new dashboard arrives?

sloria:
The homepage will be the user home page, containing a Quick Search widget and "New and Noteworthy" projects.

brianjgeiger:
I know at one point the project organizer was going to go into "myprojects" and the dashboard would be at home. What Steve is describing as the home page is what the dashboard will be as it grows up. I think removing the `/home/` route is definitely a thing to do. Whether we call the project organizer page `dashboard` or not may still need to be discussed.

JeffSpies:
Can home, rather than being deleted, just forward to “/“?

brianjgeiger:
Does it need to? `/home/`on production right now 404s.

sloria:
It can. To be clear: `/home/` doesn't exist on prod now. Should it be added?

JeffSpies:
Oh no. I thought you said it should be removed. Maybe that’s from a dev branch. If it has never appeared on production, you can ignore my comment about forwarding.

PR #4740, opened 2015-12-15
Add categories to engagement emails for analytics; use Sendgrid API for sending mail

sloria:
Replaces #4606

### Changes

Emails are now sent via the SendGrid API (`settings.SENDGRID_API_KEY` must be set).

### Purpose

This allows us to easily add categories using Sendgrid's
SMTPAPI. We will use categories to calculate open rates
of engagement emails and other emails.

Categories have been added to all engagement and notifications
emails.

### Notes

Should be backwards-compatible for local dev environments. If `settings.SENDGRID_API_KEY` is not set, emails will be sent via SMTP.

### Ticket

https://openscience.atlassian.net/browse/OSF-4806

### Deployment considerations

- Must update Python requirements (to install `sendgrid`)
- Must set `settings.SENDGRID_API_KEY` in the `local.py` for all deployments


whit537:
So, @sloria ... the reviewer has become ... the reviewed! :sunglasses:

website/settings/defaults.py
```diff
@@ -88,10 +88,16 @@ def parent_dir(path):
...
 MAIL_SERVER = 'smtp.sendgrid.net'
 MAIL_USERNAME = 'osf-smtp'
 MAIL_PASSWORD = ''  # Set this in local.py
 
+# OR, if using Sendgrid's API
+SENDGRID_API_KEY = None
+
+
 # Mandrill
 MANDRILL_USERNAME = None
```
whit537: Still need these, if we're on Sendgrid?

website/mails/queued_mails.py
```diff
@@ -95,6 +96,7 @@ def queue_mail(to_addr, mail, send_at, user, **context):
 #EMAIL_TYPE = {
 #    'template': the mako template used for email_type,
 #    'subject': subject used for the actual email,
+#    'categories': categories to attach to the email using Sendgrid's SMTPAPI.
 #    'presend': function undes presends that can modify mail.data and decides whether the email should be sent,
```
whit537: "undes"? Maybe fix that while we're here?

framework/email/tasks.py
```diff
@@ -4,50 +4,84 @@
...
+    if settings.SENDGRID_API_KEY:
+        return send_with_sendgrid(
+            from_addr=from_addr,
+            to_addr=to_addr,
+            subject=subject,
+            message=message,
+            mimetype=mimetype,
+            categories=categories
+        )
+    else:
```
whit537: I'd prefer to see this branch factored out into a function as well. Maybe `send_with_smtp`?

framework/email/tasks.py
```diff
@@ -4,50 +4,84 @@
...
+            s.login(username, password)
+        s.sendmail(
+            from_addr=from_addr,
+            to_addrs=[to_addr],
+            msg=msg.as_string()
+        )
+        s.quit()
+        return True
+
+def send_with_sendgrid(from_addr, to_addr, subject, message, mimetype='html', categories=None, client=None):
```
whit537: Consider prefixing with `_` to soft-indicate private API (this function ain't a task).

website/mails/mails.py
```diff
@@ -107,7 +112,8 @@ def send_mail(to_addr, mail, mimetype='plain', from_addr=None, mailer=None,
         login=login,
         username=username,
         password=password,
-        mail_server=mail_server
+        mail_server=mail_server,
```
whit537: It's weird to pass `mail_server` down through `send_mail` without parallel-ly passing down `sendgrid_api_key`. Seems like we should either stop passing the one, or start passing the other.

website/settings/defaults.py
whit537: Looks like they're just used in `scripts/admin_permission_email.py`. Makes me think that script is obsolete and should be removed.

whit537: [It's](https://github.com/CenterForOpenScience/osf.io/blob/25a97810cc557d8dc1792279a29d5b5c59656a68/scripts/admin_permission_email.py) a year old.

whit537:
##### Punchlist

 - [ ] fix adjacent docstring typo—https://github.com/CenterForOpenScience/osf.io/pull/4740#discussion_r50793323
 - [ ] factor out a `send_with_smtp` function—https://github.com/CenterForOpenScience/osf.io/pull/4740#discussion_r50793926
 - [ ] consider prefixing helper functions with `_`—https://github.com/CenterForOpenScience/osf.io/pull/4740#discussion_r50794008
 - [ ] prune `MANDRILL_*` settings—https://github.com/CenterForOpenScience/osf.io/pull/4740#discussion_r50793219
 - [ ] prune or update `scripts/admin_permission_email.py`—https://github.com/CenterForOpenScience/osf.io/pull/4740#discussion_r50794496
 - [ ] accept `sendgrid_api_key` alongside `mail_server`—https://github.com/CenterForOpenScience/osf.io/pull/4740#discussion_r50794072

website/mails/mails.py
whit537: Looks like the `mail_server` key is used only in [`scripts/admin_permission_email.py`](https://github.com/CenterForOpenScience/osf.io/blob/25a97810cc557d8dc1792279a29d5b5c59656a68/scripts/admin_permission_email.py) (cf. https://github.com/CenterForOpenScience/osf.io/pull/4740#discussion_r50793219). If we can get rid of that script then it seems that we can get rid of this kwarg as well.

whit537:
Alright, @sloria, made a pass through and left a punchlist. Over to you! :-)

website/settings/defaults.py
sloria: We don't typically remove migration scripts, even if they're one-off. I'll remove the MANDRILL settings though.

PR #4742, opened 2015-12-15
[WIP] Feature/admin app forms

samchrisinger:


admin/base/filters.py
```diff
@@ -0,0 +1,9 @@
+from django import template
+from django.utils.safestring import mark_safe
+from django.utils import simplejson
+
+register = template.Library()
+
+@register.filter('jsonify')
```
cwisecarver: You don't technically need to pass jsonify to filter. It'll use the function name by default.

admin/base/views.py
```diff
@@ -8,7 +8,10 @@
 
 @login_required
 def home(request):
-    context = {'user': request.user}
+    context = {
+        'user': request.user,
+        'user_groups': map(lambda g: g.name, request.user.groups.all())
```
cwisecarver: You should be able to do `request.user.groups.all().values_list('name', flat=True)` and it will do the same as your map but do it in sql.

admin/common_auth/models.py
```diff
@@ -73,7 +73,14 @@ class Meta:
     #     return self.is_admin
 
     @property
+    def full_name(self):
```
cwisecarver: There is traditionally a method on the django user model called `get_full_name` it might be worth using that instead of having this property or having the property call that method and changing what that method returns. Most django admin UI things use `get_full_name` or `get_short_name` to display info on a user.

samchrisinger: I'm seeing:
```
In [5]: u.get_full_name()
Out[5]: u'foo@bar.com'

In [6]: u.get_short_name()
Out[6]: u'foo@bar.com'
```
in my django shell. 

Do you think it's a good idea to override these methods?

admin/pre_reg/urls.py
```diff
@@ -3,10 +3,10 @@
 from . import views
 
 urlpatterns = [
-    url(r'^$', views.prereg, name='prereg'),
-    url(r'^drafts/(?P<draft_pk>[0-9a-z]+)/$', views.prereg_form, name='prereg_form'),
+    url(r'^(?:\?page=(?P<page_number>\d{1,5}))?$', views.prereg, name='prereg'),
```
cwisecarver: I'm pretty sure this isn't working. You just need to use `request.GET.get('page_number')`. I believe this is working because of the question mark at the end letting the `?page=00000` be optional. You can just switch this back to the previous line 6. `url(r'^$', views.prereg, name='prereg'),`

admin/pre_reg/utils.py
```diff
@@ -0,0 +1,12 @@
+import itertools
+
+from django.contrib.auth import models
+
+def get_prereg_reviewers():
+    return itertools.chain(
+        ((None, 'None'), ),
```
cwisecarver: I would move the prepending (None, 'None') into the place(s) this method is being called just for readability. 

An alternative to the comprehension would be 
```python
from django.db.models.functions import Concat
from django.db.models import Value

MyUser.objects
    .filter(groups__name='prereg_group')
    .annotate(fuller_name=Concat('first_name', Value(' '), 'last_name'))
    .values_list('email', 'fuller_name')
``` 
which looks up all the users with a group of prereg_group and returns a list of tuples of their `('email', 'full_name')`. I called it fuller_name because django doesn't allow you to annotate onto an object with a property of the same name.

Ends up making this sql:

```sql
SELECT "common_auth_myuser"."email", COALESCE("common_auth_myuser"."first_name", %s) || COALESCE(COALESCE(%s, %s) || COALESCE("common_auth_myuser"."last_name", %s), %s) AS "fuller_name" FROM "common_auth_myuser" INNER JOIN "common_auth_myuser_groups" ON ( "common_auth_myuser"."id" = "common_auth_myuser_groups"."myuser_id" ) INNER JOIN "auth_group" ON ( "common_auth_myuser_groups"."group_id" = "auth_group"."id" ) WHERE "auth_group"."name" = %s ORDER BY "common_auth_myuser"."email" ASC LIMIT 21' - PARAMS = (u'', u' ', u'', u'', u'', u'prereg_group'); args=('', ' ', '', '', '', 'prereg_group')
```

samchrisinger: Nice!

admin/common_auth/models.py
cwisecarver: They're defined in `common_auth.models.MyUser`. I'd talk it over with @Ghalko or @zkraime but I'd wager that they're used for displaying things to the user and it's probably fine to change them. Especially since if they need the email for something programmatic or for display in a template u.email will still work.

admin/pre_reg/views.py
```diff
@@ -20,45 +21,63 @@
...
+        page_number = 1
+
+    page = paginator.page(page_number)
+
+    try:
+        drafts = [serializers.serialize_draft_registration(d, json_safe=False) for d in page]
+    except EmptyPage:
+        drafts = []
+
+    for draft in drafts:
```
cwisecarver: An alternative to this would be django's forests.

You should be able to do:
```python
from django.forms import formset_factory
from somewhere import DraftRegistrationForm

DraftRegistrationFormset = formset_factory(DraftRegistrationForm, initial=drafts, extra=0) # extra=0 tells it not to put a blank on in
context = {
    formset=DraftRegistrationFormset,
    ...
    ...
}
```

Then you should be able to loop through formset like a list in the template, each iteration will give you a form.

https://docs.djangoproject.com/en/1.9/topics/forms/formsets/#using-initial-data-with-a-formset

samchrisinger: I tried formsets, but since the forms aren't rendered in a tabular fashion it seemed like more trouble than it was worth.

admin/common_auth/models.py
zkraime: @samchrisinger There's no particular reason for overriding those methods to return email. Simply, I overrode them b/c our model doesn't use username. I can remove the override

samchrisinger: Thanks for the clarification. I will just change it in this PR

admin/pre_reg/views.py
cwisecarver: Ah, okay. I haven't gotten to the template part yet. I just thought the factory seemed clearer than the same variable appearing on both sides of the equals sign. That always makes me double-take. I'll keep looking.

admin/base/views.py
```diff
@@ -8,7 +8,9 @@
 
 @login_required
 def home(request):
-    context = {'user': request.user}
+    context = {
+        'user': request.user,
+    }
     return render(request, 'home.html', context)
```
cwisecarver: Because `django.template.context_processors.request` is included in your context processors request is always in your context so you can remove 11-13 and do
```python
return render(request, 'home.html')
```
If that doesn't work send context as an empty object.

`{{ request.user }}` should work in the template.

admin/pre_reg/forms.py
```diff
@@ -0,0 +1,38 @@
...
+    assignee = forms.ChoiceField(
+        label="Assignee",
+        choices=PREREG_REVIEWERS
+    )
+
+    notes = forms.CharField(
+        label="Notes",
+        widget=forms.Textarea(
+            attrs={
+                'style': 'height: 50px;'
```
cwisecarver: This would be better if it were in css. At least when someone tries to figure out how to change it.

admin/pre_reg/views.py
cwisecarver: This might help you with the forms rendering as a table. If not, feel free to keep it the other way.
`{{ form.as_table }}` will render them as table cells wrapped in `<tr>` tags
https://docs.djangoproject.com/en/1.9/topics/forms/#form-rendering-options

admin/pre_reg/views.py
```diff
@@ -122,12 +116,10 @@ def reject_draft(request, draft_pk):
 
     user = request.user.osf_user
     draft.reject(user)
-    return JsonResponse({})
-
+    return redirect('pre_reg:prereg', permanent=True, page_number=int(request.POST.get('page', 1)))
```
cwisecarver: Is there anyway something can be un-rejected? If so you might want a 302 not a 301.

admin/pre_reg/views.py
```diff
@@ -137,10 +129,20 @@ def update_draft(request, draft_pk):
     data = json.loads(request.body)
```
cwisecarver: Somewhere in here you should be passing request.POST to the form and calling .is_valid(). https://docs.djangoproject.com/en/1.8/topics/forms/#the-view

cwisecarver:
#### Functional notes:
- The save button on http://localhost:8000/admin/prereg/ doesn't appear to do anything. No errors, no form submit, no network activity in the browser console.
- View Draft opens to the "Other" page
- My user is in the prereg_group but Approve and Reject have a "no cursor" and don't do anything.
- The action for the add comment button from the view draft page appears to be pointing to /admin/pre-reg/drafts/<id>/update/ but that url is invalid. It should be /admin/prereg/drafts/<id>/update/ so comments don't save.

admin/pre_reg/forms.py
```diff
@@ -0,0 +1,38 @@
...
+            }
+        )
+    )
+
+
+DraftRegistrationFormset = forms.formset_factory(DraftRegistrationForm)
```
cwisecarver: Probably don't need this if you're not using formsets.

cwisecarver:
Running the tests, both of them fail for me. 
```
ERROR: Decorator for views that checks that the user passes the given test,
  File "/Users/chriswisecarver/.venvs/osf/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/chriswisecarver/.venvs/osf/lib/python2.7/site-packages/nose/util.py", line 620, in newfunc
    return func(*arg, **kw)
TypeError: user_passes_test() takes at least 1 argument (0 given)
```

```
ERROR: test_prereg (admin_tests.pre_reg.test_views.PreregViewsTests)
  File "/Users/chriswisecarver/Dropbox/cos-dev/osf.io/admin_tests/pre_reg/test_views.py", line 107, in test_prereg
    assert_in('drafts', res.context)
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/case.py", line 802, in assertIn
    if member not in container:
TypeError: argument of type 'NoneType' is not utterable
```

cwisecarver:
@samchrisinger I think I've noted everything I've seen. Let me know if you want to walk through anything. I'm over by the couches.

admin/pre_reg/views.py
samchrisinger: For now, I'd rather leave it as is. We're mixing non-form draft-specific data and form data. I would rather this read-only draft data not be form fields. So it seems challenging to use the tabular formset rendering in parallel with this approach.

admin/pre_reg/views.py
samchrisinger: Not currently.

PR #4743, opened 2015-12-15
Dashboard redesign component 3

caneruguz:
## Purpose
This PR adds new functionality and makes improvements on prior items based on earlier code reviews for the Jira tickets listed at the bottom

## Changes
- Added collections end points
- Improved collection UI
- Added add project plugin
- Added new functionality for activity logs
- UI improvements overall

## Side effects
- Will break prereg challenge due to the use of a deleted onboarder feature being used for registration pages. This will be fixed before merge to develop. 

JIRA Tickets:
https://openscience.atlassian.net/browse/OSF-4799
https://openscience.atlassian.net/browse/OSF-4800
https://openscience.atlassian.net/browse/OSF-5155
https://openscience.atlassian.net/browse/OSF-5263
https://openscience.atlassian.net/browse/OSF-5408


website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,799 @@
...
+                args.collections.map(function(item, index){
+                    if (item.id === args.activeFilter()) {
+                        selectedCSS = 'active';
+                    } else if (item.id === args.collectionMenuObject().item.id) {
+                        selectedCSS = 'bg-color-hover';
+                    } else {
+                        selectedCSS = '';
+                    }
+                    if (!item.data.systemCollection) {
+                        submenuTemplate = m('i.fa.fa-ellipsis-v.pull-right.text-muted.p-xs', {
```
hmoco: The cursor does not change for this, making it not clear that it is clickable. Maybe switch it to ` a.fa.fa-ellipsis-v... `

website/static/js/addProjectPlugin.js
```diff
@@ -0,0 +1,178 @@
...
+                self.goToProjectLink(result.data.links.html);
+            };
+            var error = function _error (result) {
+                self.viewState('error');
+                console.log('error', result);
+            };
+            m.request({method : 'POST', url : url, data : data, config : xhrconfig})
+                .then(success, error);
+            self.newProjectName('');
+            m.redraw(true);
```
sloria: Why is this redraw necessary? IIRC, mithril will redraw when `m.request` resolves.

website/static/js/addProjectPlugin.js
```diff
@@ -0,0 +1,178 @@
...
+
```
sloria: Methods that do async work should return a promise. This allows for chaining

```js
AddProject.add()
    .then(....)
```

 Since mithril xhr's are promises, you can just return the output of `m.request`.

website/static/js/addProjectPlugin.js
```diff
@@ -0,0 +1,178 @@
...
+                            m('textarea.form-control#projectDesc', {
+                                onkeyup: function(ev){
+                                    ctrl.newProjectDesc($(this).val());
+                                },
+                                value : ctrl.newProjectDesc()
+                            })
+                        ]),
+                        m('.f-w-lg.text-bigger','Category'),
+                        m('.category-radio.p-h-md', [
+                            m('.radio', m('label', [m('input[type="radio"][name="projectCategory][id="catProject][value="project][checked="checked"]',{ onchange : ctrl.chooseCategory }), 'Project'])),
```
sloria: Is the `checked` attribute being set correctly here? Shouldn't it be based on the `newCategory` prop rather than being hardcoded?

website/static/js/addProjectPlugin.js
```diff
@@ -0,0 +1,178 @@
...
+                            m('.radio', m('label', [m('input[type="radio"][name="projectCategory][id="catHypothesis][value="hypothesis"]',{ onchange : ctrl.chooseCategory }), 'Hypothesis'])),
```
sloria: Are the CSS IDs necessary here? They seem to only add pollution.

website/static/js/addProjectPlugin.js
```diff
@@ -0,0 +1,178 @@
...
+                                         ctrl.add();
+                                    }
+                                    ctrl.newProjectName($(this).val());
+                                },
+                                value : ctrl.newProjectName()
+                            })
+                        ]),
+                        m('.form-group.m-v-sm', [
+                            m('label[for="projectDesc].f-w-lg.text-bigger', 'Project Description'),
+                            m('textarea.form-control#projectDesc', {
```
sloria: Remove unnecessary CSS IDs.

website/static/js/addProjectPlugin.js
```diff
@@ -0,0 +1,178 @@
...
+                                onkeyup: function(ev){
```
sloria: mithril has a useful helper for handlers like this:

```js
onkeyup: m.withAttr('value', ctrl.newProjectDesc)
```

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,799 @@
...
+            return $osf.apiV2Url('nodes/' + self.data.uid + '/children/', { query : { 'related_counts' : true, 'embed' : 'contributors' }});
+        }
+        // If nothing
+        throw new Error('Link could not be generated from linkObject data');
+    };
+    self.link = self.generateLink();
+};
+
+if (!window.fileBrowserCounter) {
+    window.fileBrowserCounter = 0;
```
sloria: Why is it necessary to set `fileBrowserCounter` on `window`?

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,799 @@
...
+
+        // Default system collections
+        self.collections = [
+            new LinkObject('collection', { path : 'users/me/nodes/', query : { 'related_counts' : true, 'page[size]'  : 100, 'embed' : 'contributors' }, systemCollection : true}, 'All My Projects'),
+            new LinkObject('collection', { path : 'registrations/', query : { 'related_counts' : true, 'page[size]'  : 100, 'embed' : 'contributors'}, systemCollection : true}, 'All My Registrations'),
+        ];
+
+        var collectionsUrl = $osf.apiV2Url('collections/', { query : {'related_counts' : true, 'sort' : 'date_created'}});
+        m.request({method : 'GET', url : collectionsUrl, config : xhrconfig}).then(function(result){
+            console.log(result);
```
sloria: :dollar:

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,799 @@
...
+            new LinkObject('name', { id : '8q36f', query : { 'related_counts' : true }}, 'Caner Uguz'),
+        ];
+        self.tagFilters = [
+            new LinkObject('tag', { tag : 'something', query : { 'related_counts' : true }}, 'Something Else'),
+        ];
+        self.data = m.prop([]);
+        self.activityLogs = m.prop();
+        self.getLogs = function _getLogs (nodeId) {
+            var url = $osf.apiV2Url('nodes/' + nodeId + '/logs/', { query : { 'embed' : ['nodes', 'user', 'linked_node', 'template_node']}});
+            m.request({method : 'GET', url : url, config : xhrconfig}).then(function(result){
```
sloria: See previous comments--return promises from async methods.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,799 @@
...
+
+        // USER FILTER
+        self.activeFilter = m.prop(1);
+        self.updateFilter = function(filter) {
+            self.activeFilter(filter.data.id);
+            self.updateFilesData(filter);
+        };
+
+        self.updateListSuccess = function(value) {
+            console.log(value);
```
sloria: :dollar:

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,799 @@
...
+            if (success === undefined){
+                success = self.updateListSuccess;
+            }
+            if (error === undefined){
+                error = self.updateListError;
+            }
+            if (typeof url !== 'string'){
+                throw new Error('Url argument for updateList needs to be string');
+            }
+            m.request({method : 'GET', url : url, config : xhrconfig})
```
sloria: Return this.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,799 @@
...
+                    }
+                }
+            };
+            m.request({method : 'POST', url : url, config : xhrconfig, data : data}).then(function(result){
+                console.log(result);
+                var node = result.data;
+                args.collections.push(new LinkObject('collection', { path : 'collections/' + node.id + '/linked_nodes/', query : { 'related_counts' : true }, systemCollection : false, node : node }, node.attributes.title));
+            });
+            self.newCollectionName('');
+            m.redraw(true);
```
sloria: Again, I'm not sure this redraw is necessary.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,799 @@
...
+            var data = {
+                'data': {
+                    'type': 'collections',
+                    'attributes': {
+                        'title': self.newCollectionName(),
+                    }
+                }
+            };
+            m.request({method : 'POST', url : url, config : xhrconfig, data : data}).then(function(result){
+                console.log(result);
```
sloria: :dollar:

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,799 @@
...
+                                m('p.m-t-sm', 'After you create your collection drag and drop projects to the collection. ')
+                            ]),
+                            m('.modal-footer', [
+                                m('button[type="button"].btn.btn-default[data-dismiss="modal"]', 'Cancel'),
+                                m('button[type="button"].btn.btn-success', { onclick : ctrl.addCollection },'Add')
+                            ])
+                        ])
+                    )
+                ),
+                m('#renameColl.modal.fade[tabindex=-1][role="dialog"][aria-labelledby="renameCollLabel"][aria-hidden="true"]',
```
sloria: Again, remove any unnecesssary CSS IDs.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,799 @@
...
+                        }, [
+                            m('i.fa.fa-trash'),
+                            ' Delete'
+                        ])
+                    ])
+                ]) : ''
+            ]),
+            m('.fb-collections-modals', [
+                m('#addColl.modal.fade[tabindex=-1][role="dialog"][aria-labelledby="addCollLabel"][aria-hidden="true"]',
+                    m('.modal-dialog',
```
sloria: There's a lot of repetition in the view for this component. Can you factor some of the common code into smaller, stateless components? These modal dialogs seem like a good candidate.

website/static/js/logActionsList.js
```diff
@@ -0,0 +1,82 @@
...
+    'box_file_updated' : '${user} updated file ${path} in Box in ${node}',
+    'box_file_removed' : '${user} removed ${path} from Box in ${node}',
+    'box_folder_selected': '${user} linked Box folder ${folder} to ${node}',
+    'box_node_deauthorized' : '${user} deauthorized the Box addon for ${node}',
+    'box_node_authorized' : '${user} authorized the Box addon for ${node}',
+    'dataverse_file_added' : '${user} added file ${filename} to Dataverse dataset {$dataset} in ${node}',
+    'dataverse_file_removed' : '${user} removed file ${filename} from Dataverse dataset ${dataset} in ${node}',
+    'dataverse_dataset_linked' : '${user} linked Dataverse dataset ${dataset} to ${node}',
+    'dataverse_study_linked' : '${user} linked Dataverse dataset ${study} to ${node}',
+    'dataverse_dataset_published' : '${user} published a new version of Dataverse dataset ${dataset} to ${node}',
```
hmoco: ` ${dataset} ` not being properly populated

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,799 @@
...
+    view : function (ctrl, args) {
+        var selectedCSS;
+        return m('.fb-filters.m-t-lg',
+            [
+                m('h5', 'Contributors'),
+                m('ul', [
+                    args.nameFilters.map(function(item, index){
+                        selectedCSS = item.id === args.activeFilter() ? '.active' : '';
+                        return m('li' + selectedCSS,
+                            m('a', { href : '#', onclick : args.updateFilter.bind(null, item)}, item.label)
```
sloria: Might be useful to document required arguments (e.g. `updateFilter`) in the docstring for components.

sloria: Why is `updateFilter` bound to `null`? Why not `this`?

sloria:
Just a few drive-by comments; haven't fully reviewed.

website/static/js/logTextParser.js
```diff
@@ -0,0 +1,232 @@
...
+            }
+        }
+    },
+    // Contrubutor list of added, updated etc.
+    contributors: {
+        view: function (ctrl, logObject) {
+            var contributors = logObject.embeds.contributors;
+            if(paramIsReturned(contributors, logObject)) {
+                return contributors.map(function(item){
+                    return m('a', {href: '#'}, 'Person');
```
hmoco: I assume this is waiting actual interaction with the contributors being returned

hmoco:
Making the screen small defaults to having the menu open, darkening the other parts of the screen. It is almost modal-like, which implies clicking in the darkened area would dismiss the menu. However (unless I missed something), the menu icon needs to be clicked. Not sure what is the wanted behavior.

website/static/js/fileBrowser.js
caneruguz: the a tag adds views and is not really meaningful here since these are not links. It;s best to add pointer class to these.

website/static/js/addProjectPlugin.js
sloria: Also, I think `onchange` might be preferred here, so that this handles pasted text.

website/static/js/addProjectPlugin.js
```diff
@@ -0,0 +1,178 @@
...
+                m('.modal-body', [
+                    m('', [
+                        m('.form-group.m-v-sm', [
+                            m('label[for="projectName].f-w-lg.text-bigger', 'Project Name'),
+                            m('input[type="text"].form-control#projectName', {
+                                onkeyup: function(ev){
+                                    if (ev.which === 13) {
+                                         ctrl.add();
+                                    }
+                                    ctrl.newProjectName($(this).val());
```
sloria: What is `this` in this context? It seems most straightforwared to do `ev.value`.

hmoco:
Minimal clipping in very small windows. 
screen shot 2015-12-15 at 11 41 17 am

website/static/js/logTextParser.js
```diff
@@ -0,0 +1,232 @@
...
+        }
+        return m('span', '"' + source + '"');
+    }
+    return m('span', text);
+};
+
+var LogText = {
+    view : function(ctrl, logObject) {
+        var text = logActions[logObject.attributes.action];
+        var list = text.split(/(\${.*?})/);
```
hmoco: After looking at alternative in other commits, I really like this.

website/static/js/addProjectPlugin.js
caneruguz: Good point, this was mainly for refreshing the newProjectName but it doesn't seem to be necessary. Changed as suggested.

website/static/js/addProjectPlugin.js
caneruguz: so something like this:
```
self.add = function() {
    ....  
    return m.request(...)
}

self.add().then(success, error)
```
I don't see the benefit of doing this in this case though, and it seems to be decoupling success and error responses which are still part of adding. Can you write out an example of the refactoring?

website/static/js/addProjectPlugin.js
caneruguz: Good catch, I was trying to avoid using m.withatt but with a bit more work I got that to work and removed repitition in the code with this:
```js
m('.category-radio.p-h-md', [
    ctrl.categoryList.map(function(cat){
        return m('.radio', m('label', [  m('input', {
            type: 'radio',
            name: 'projectCategory',
            id: cat.id,
            value: cat.value,
            checked: ctrl.newProjectCategory() === cat.value,
            onchange : m.withAttr('value', ctrl.newProjectCategory)
        }), cat.label ]));
    })
])
```

sloria: Much better. Do you need the `ids`?

website/static/js/addProjectPlugin.js
caneruguz: Hmm, I see your point. I have been thinking that using ID's with form elements is good practice but I can't remember exactly why I have that in my head. I'll remove to ids.

website/static/js/addProjectPlugin.js
caneruguz: Changed as suggested. :+1:

website/static/js/addProjectPlugin.js
caneruguz: Changed as suggested. :+1:

website/static/js/fileBrowser.js
caneruguz: This library may be invoked several times from different parts of the code (i.e. multiple browsers in single page, this is unlikely but the architecture is to allow it) and ID's need to still be unique across the page. I thought window is the safest bet for these scenarios.

website/static/js/fileBrowser.js
caneruguz: Hmm, nothing like console logging things... Okay I will remove these before sending out to final review.

website/static/js/fileBrowser.js
caneruguz: Changes as suggested.

website/static/js/logActionsList.js
```diff
@@ -0,0 +1,82 @@
+/**
+ * Created by cos-caner on 12/4/15.
+ */
+
+var logActions = {
```
brianjgeiger: The only real downsides to this method are:
* This requires a centralized place that will have to be updated with addons, while most of the rest of the addon architecture is modular;
* If we get around to localizing strings, these will not be where many of the other strings are.

The second item is less important than the first, as we've not done amazingly well at ensuring even our Python strings will be ready for localization. However, if we can at least have a method of gathering log actions from individual addons rather than consolidating them here, that should make life a little easier for release management of addons.

website/static/js/logTextParser.js
brianjgeiger: You can assume that this will be a list of users if you'd like. That is almost certainly how it will be implemented.

website/static/js/fileBrowser.js
caneruguz: I was trying to bind a function (i.e. not call) without binding this to it, but since the function uses self for its main context it doesn't really matter. I can pass this but it won't be used necessarily. What's the best thing to do here?

website/static/js/logTextParser.js
caneruguz: Yes this is waiting for interaction but it is already implemented as expecting a list of users as contributors, the view is going through the list and returning a person name for each.

website/static/js/fileBrowser.js
caneruguz: The repitition is mainly on the structure of the modal and all modals have that kind of repitition. One solution could be to make a mithril modal component.

However my experience is that these kind of don't repeat yourself methods end up making convoluted plugins that have a thick layer of options and thin layer of repeated code being saved. For instance not all modals have footers, not all have headers, and almost everything inside body section will be different.  Also components will split the code to pieces which may cause difficulties in debugging. For large components this makes sense but are we not getting too modular where each component is a tiny section of the UI? 

I'm not necessarily convinced that this shouldn't be refactored but I'd like to talk more about it.

caneruguz:
Closing this until ready for code review

PR #4744, opened 2015-12-15
Add 'Software' as a component option

Marcus-Rosti:
# Purpose

Adds software to the component lists and uses the font awesome laptop icon.  

# Changes

The two files affected are $OSFHOME/website/static/js/iconmap.js which adds the icon and $OSFHOME/website/project/model.py which adds software as an option in components

# Side Effects
NA

### DIFF
--- a/website/static/js/components/legend.js
+++ b/website/static/js/components/legend.js
@@ -25,7 +25,7 @@ var find_index = function (array, str) {
 var remove_index = function (array, index) {
     var new_array = [];
     for (var i = 0; i < array.length; i++) {
-        if (i != index)
+        if (i !== index)
             new_array.push(array[i]);
     }
     return new_array;
@@ -33,29 +33,20 @@ var remove_index = function (array, index) {
 
 module.exports = {
     view: function (data, repr, opts) {
-
-        var clean_array = data.splice();
-
-        console.log("removing other");
-        var other_index = find_index(data, "Other");
+        var other_index = find_index(data, 'Other');
         var other = data[other_index];
         data = remove_index(data, other_index);
 
-
-        console.log("removing Uncat");
-        var uncat_index = find_index(data, "Uncategorized");
+        var uncat_index = find_index(data, 'Uncategorized');
         var uncat = data[uncat_index];
         data = remove_index(data, uncat_index);
 
-        console.log("Sorting");
         if (data[0].label) {
             data.sort(function (a, b) {
                 return a.label.localeCompare(b.label);
             });
         }
 
-        console.log("Pushing objects back");
-        console.log(other, uncat);
         data.push(other);
         data.push(uncat);
 
--- a/website/static/js/components/legend.js
+++ b/website/static/js/components/legend.js
@@ -1,17 +1,68 @@
 require('css/legend.css');
 var m = require('mithril');
 
+/** Finds the index of a string in an array
+ *
+ * @param {Array} array The array to search through
+ * @param {String} str The string to find
+ * @returns {*}
+ */
+var find_index = function (array, str) {
+    for (var i = 0; i < array.length; i++) {
+        console.log(array[i].label);
+        if (array[i].label === str)
+            return i;
+    }
+    return -1;
+};
+
+/** Removes an index from an array immutably
+ *
+ * @param {Array} array the array to modify
+ * @param {Int} index the index to remove
+ * @returns {*} a new array missing the index value
+ */
+var remove_index = function (array, index) {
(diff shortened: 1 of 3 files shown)


website/static/js/components/legend.js
```diff
@@ -1,17 +1,56 @@
...
+    }
+    return new_array;
+};
+
 module.exports = {
-    view: function(data, repr, opts) {
-        if(data[0].label){
-            data.sort(function(a, b) {
+    view: function (data, repr, opts) {
+        var other_index = find_index(data, 'Other');
```
mfraezz: If it's coming directly from the (now properly ordered) `OrderedDict`, `data` doesn't need to be modified.

website/static/js/components/legend.js
```diff
@@ -1,17 +1,56 @@
 require('css/legend.css');
 var m = require('mithril');
 
+/** Finds the index of a string in an array
```
mfraezz: I really appreciate the quality of documentation, but these methods are superfluous.

website/static/js/components/legend.js
Marcus-Rosti: The issue is that "Other" And "Uncategorized" need to come last, which would be out of order.

website/static/js/components/legend.js
Marcus-Rosti: Find index probably is, since I think JS provides .indexOf, however remove_index was the only way I could figure out a static way of removing an index without leaving a misindexed arrays or missing values.

website/static/js/components/legend.js
mfraezz: Oh, okay... They already come last, but `projectOrganizerCategories` extends the sorted `CATEGORY_MAP` in an unsorted way, necessitating a sort. This is the passed-in ordering:

screen shot 2015-12-18 at 10 29 53

`Collections`, `Smart Collections`, and `Link` aren't in `CATEGORY_MAP`. `Project` is but is also included in the `.extend` call, bumping it up.

Marcus-Rosti: Ah okay, I thought the issue called for the ordering to be as such:

screen shot 2015-12-18 at 10 58 10 am 2

GaryKriebel: Let's double check with Erin on what the ordering should be.

website/static/js/components/legend.js
mfraezz: Actually neither are, I was working under the assumption that everything in the legend was also in `CATEGORY_MAP`, my mistake.

`find_index` isn't because it's an array of objects, so you can't easily use something like `.indexOf` to query on an attribute of each object. You *could* do some magic with `jQuery.grep`, or, more simply modify the sorting function we're using to something that special cases what we want to push to the back, like 
```js
data.sort(function (a, b) {
    if (a.label === 'Other' && b.label !== 'Uncategorized') {
        return 1;
    }
    if (b.label === 'Other' && a.label !== 'Uncategorized') {
        return -1;
    }
    return a.label.localeCompare(b.label);
});
```

website/static/js/components/legend.js
mfraezz: It does, I was attempting to explain a misunderstanding I had about where the list was coming from.

Marcus-Rosti: Ah I see what you mean, I guess then I'm not sure where the list is generated and provided to the client but that was the point where it was being sorted and displayed

mfraezz: Most of it is coming from the `OrderedDict` in `website/project/model.py` that you modified, but it's being extended in `website/static/js/projectorganizer.js`, which breaks ordering. My bad.

This file could be simplified by modifying the sort function instead of the data, which would also give a small performance boost.

website/static/js/components/legend.js
```diff
@@ -2,16 +2,39 @@ require('css/legend.css');
...
+                sort a b
+                    a/b == Uncat     = -1
+                    a/b == Other
+                        b/a == Uncat =  1
+                        otherswise   = -1
+                    otherwise        = compare a b
+             */
+            data.sort(function (a, b) {
+                // Uncategorized should come last
+                if(a.label === 'Uncategorized'|| b.label === 'Uncategorized') return -1;
```
mfraezz: This will work, but only because `a.label` won't ever be `'Uncategorized'` with current initial ordering. 

Returning `-1` in the function that `sort` takes signifies the two items should not swap places, so if it was (somehow) `a.label`, `Uncategorized` wouldn't be last in the final result.

PR #4509, opened 2015-10-23
Anonymized view only link public view changes

alexschiller:
## Purpose
Previously: Public projects in anonymized view only links showed all information
Now: Only projects selected when creating a view only link will be visible.

## Changes
Adds to the node.can_view logic to see if a project is being viewed through an anonymous private link. If it is, only projects selected when the private link was created will be viewable

Adds private_link to auth which gives access to the current private links details from auth without needing to explicitly look it up each time 

## Side Effects
1. Because this changes what can be viewed there are many places that side effects will occur in both hiding content that should not be seen, and showing content that should not be seen. I believe it is working properly, but there are lots of combos that may need to be tested to ensure they all work properly. 

2. Anonymous view only links anonymize indiscriminately. Navigating to another project with the view only link will anonymize that project, the dashboard, everything. It shouldn't be too hard to fix this, but wanted it noted that this occurs.

## QA Considerations
For each auth situation, the files, components, and recent activity should show only the proper information, and be properly anonymized. 

Combinations of these variables should be tested, some may be excluded where overlap occurs:
1. Logged in (any read/write access), Not logged in
2. Anonymized private link, Not anonymized private link, Non existent private link, Private link from another project 
3. Public Project, Private Project
In the project:
4. Public Node (checked), Public Node (unchecked), Private Node (checked), Private Node (unchecked)

When there is an anonymous private link, this overrides all other situations. 
1. Nodes included in the anonymous private link should be visible
2. Nodes not included in the anonymous private link should not be visible

Auth should work as expected in all other situations.

PR #4745, opened 2015-12-15
[Ready to merge] [feature] Add Edge-Side-Include functionality to API #4669

cwisecarver:
## Description

- **Purpose**
    - [Edge-Side-Includes](https://en.wikipedia.org/wiki/Edge_Side_Includes) allow Varnish to replace specific xml tags with the contents of the url specified at the href property.
        - Example: `<esi:include href="http://www.google.com"/>` would be replaced with the contents of the google homepage.
    - The embed functionality in the API makes this very useful since a user embedded into a node is the same as a user requested individually. So varnish can cache `/v2/users/1234z/` and anytime that user is embedded in a request it can use the cached version replacing the `<esi:include href="/v2/users/1234z/"/>` with the full json representation of the user.

- **Changes**
    - There is a `to_esi_representation` method that was added on certain serializers. This method takes the place of the `to_representation` when the `ENABLE_ESI` setting is `True` and a resource is embedded.
    - Instead of python embedding the full content of the user as it does with to_representation it responds with `<esi:include href="{{ self.get_absolute_url() }}" />`. When the response is processed by Varnish it will do a request for the specified href and replace the xml tag with the response from that request and caching the response. The next time that url is included in an esi tag it will be pulled from cache.
    - Blinker events were added to GuidStoredObject that fire upon model save. When these events fire they're caught by a listener that fires off a celery task to send a `BAN` HTTP request to the varnish server in the pattern of `BAN {{ self.get_absolute_url }}.*`. This is processed as a regex and should account for all possible urls that represent that object. Since objects are stored by varnish individually, even when they're embedded in another object, this pattern should remove all instances, even those that are embedded, from cache.
        - `BAN /v2/users/1234z/.*`, to varnish, means add this pattern to the list of invalid url patterns. When someone requests a URL that matches an invalid url pattern make sure to request it from the backend server instead of serving it from cache. BANning instead of PURGEing means that we can avoid the thundering herd problem that commonly occurs with cache invalidation and it also means that we're not actually emptying the cache until we need to.

- **Side Effects**
	- URLs are supposed to represent unique data. Since we combine URLs and user authentication information to represent unique data this complicates caching. Varnish's configuration is VARYing on osf cookie, Authorization HTTP header, and URL. This **should** account for different users hitting the same url and getting a different result based on their individual permissions. There is always a chance that I've missed something. My hope is that the only side-effect of this is a low cache hit-rate. Meaning that since we're caching things per user our cache won't be utilized as well as it could for non-anonymous users.
	- Embed code is somewhat in flux right now. I've been basing my changes off of the two PRs that Dawn has been working on related to embeds but they have been changing while I've been working. I've tried to keep this branch as up to date with her stuff as I possibly could but there is a chance that a merge has gone sideways or I've missed a PR that was related to the embed work.
	- Since everything is behind a setting, unless you've switched ENABLE_ESI to true there should be no change from this PR.
    - The JSONAPIRenderer has a bit of code in it that I would really like to refactor but I couldn't find a better way to do it. Right now, because ESI is XML and all our stuff is JSON, DRF is quoting the value of the `to_esi_representation` method. Varnish doesn't expect ESI tags to be in quotes so `"<esi:include href="something"/>"` is ignored. The JSONAPIRenderer.render method has a regex that removes the double-quotes from the outside of the xml tag. This causes invalid JSON to be passed to varnish and varnish replaces the ESI tag prior to the client seeing it making it valid JSON again. While I love regex it can be brittle. Advice and alternate approaches are welcomed.

	
## Test Setup

For testing/CR, you'll need to install and setup varnish. A sample VCL file is attached to this PR. VCL is a domain-specific language designed to be used to define request handling and document caching policies for Varnish Cache. It is compiled when varnish starts up. The final VCL, very similar to the attached one, is included in the private docker repo.

- **Assumptions**
	- Your API server is running on port 8000
	- You want Varnish to listen on port 8193
	- You've got 2 gigs of free ram for varnish to use for caching (change `-s malloc,2G` in the varnish startup command to something else, `-s malloc,256M` for instance)
- **Steps**
	- `brew update; brew install varnish`
	- Download the attached VCL file and copy it to `/usr/local/etc/varnish/default.vcl` or wherever your varnish installation is. [default.vcl.txt](https://github.com/CenterForOpenScience/osf.io/files/48741/default.vcl.txt)
	- Startup varnish with the following command `sudo /usr/local/opt/varnish/sbin/varnishd -n /usr/local/var/varnish -f /usr/local/etc/varnish/default.vcl -s malloc,2G -T 127.0.0.1:2000 -a 127.0.0.1:8193 -p feature=+esi_disable_xml_check -p workspace_backend=262144 -p workspace_client=262144`
	- Change the `ENABLE_ESI` setting in your `api/base/settings/local.py` to `True`
	- Change the `ENABLE_VARNISH` setting in your `api/base/settings/local.py` to `True`
	- Set the `VARNISH_SERVERS` setting in your `api/base/settings/local.py` to `['http://127.0.0.1:8193']`
	- Navigate to `http://localhost:8193/v2/nodes/?format=jsonapi&embed=contributors&page[size]=100`. In your `inv apiserver` console you should see the initial request followed by several other requests for individual contributors. This is varnish replacing the ESI tags with the individual responses. If you refresh this page you should see your pageload time decrease significantly and you should see no requests in your `inv apiserver` console.
	- Now update a contributor or node that was included in your previous request. If celery is running in your development environment the next time you request the list of nodes you should see the node and or the contributor you updated be requested from the API server but the other components should be pulled from the varnish cache. This means that the signal fired upon save, the celery task banned that content from varnish, and varnish re-requested the banned content the next time you needed it.
	- Making the same list-of-nodes request with either a non-logged-in user or logged in as a different user should cause requests to the python apiserver. But repeated requests as a non-logged in user or repeated requests as the same logged in user should be drawn from cache instead of making a request to the python apiserver.

## Random informations

- Varnish is **ONLY** configured to cache GET and HEAD requests where the URL starts with /v2/. Anything that does not meet that criteria will be passed directly to the python api server.
- If you open the developer console in your browser and have a look at the response headers you should see `X-Cache: HIT` (it was a cache hit) and `X-Cache-Hits: #` (there have been # hits on this version of the cached object) and `Age: ##` (the cached object is ## seconds old) and `X-beresp-ttl: 120.000` (the time-to-live of this cached object is 120.000 seconds). When `Age` == `X-beresp-ttl` the object is removed from the cache.
- We can send headers back from DRF to control the TTL based on whatever we'd like. Individual objects could be cached for days or seconds. This is not implemented yet.


whit537:
Can I get a status update on this PR, @cwisecarver? It's in the Code Review queue in Jira, but Travis ain't too happy. ;-)

Should we drop this back to `In Progress`?

whit537:
[OSF-5081](https://openscience.atlassian.net/browse/OSF-5081)

cwisecarver:
It should probably go back to in progress until I can get the tests passing
again.

whit537:
I've made a first pass through reading the description.

whit537:
I've installed Varnish and started the API server with ENABLE_ESI turned on. Appears to be working so far. I need to a) set up celery, and b) figure out how to create and modify contributors and/or nodes in order to test further ...

cwisecarver:
@whit537 I think I forgot to mention in the docs there's an ENABLE_VARNISH setting that enables the celery tasks for cache invalidation.

whit537:
@cwisecarver Yeah, I see that along with [`VARNISH_SERVERS`](https://github.com/cwisecarver/osf.io/blob/41139b20b78550982ee8242c18e24ad81e2d13ae/api/base/settings/defaults.py#L172). I'm currently mired in user registration, however. I've got `inv server` running on port 5000, but the login form posts to port 8080. I'm missing a setting, I suppose?

whit537:
Maybe [`CAS_SERVER_URL`](https://github.com/cwisecarver/osf.io/blob/41139b20b78550982ee8242c18e24ad81e2d13ae/website/settings/defaults.py#L264)? (What's CAS? What's MFR?)

cwisecarver:
@whit537 There's a fakecas server, which pretends to be an actual CAS server, that you have to be running on that port IIRC. https://github.com/CenterForOpenScience/fakecas

whit537:
Ah, right! The authentication server. Thanks.

cwisecarver:
@whit537 I think CAS is centralized authentication (maybe authorization) service. MFR is the modular file renderer. You shouldn't need that one for this specific test but there's a repo under CenterForOpenScience.

whit537:
Thanks. I've got `fakecas` running and am proceeding apace ...

whit537:
I now get an ESI string as expected on port 8000:

screen shot 2015-12-29 at 11 51 28 am

cwisecarver:
Excellent. 

I feel bad about the invalid JSON but varnish's ESI processor won't parse the tag if it's inside quotes.

whit537:
And I get the replaced value on port 8193, with caching per `inv apiserver` console and the response headers:

screen shot 2015-12-29 at 11 53 15 am

whit537:
> I feel bad about the invalid JSON but varnish's ESI processor won't parse the tag if it's inside quotes.

Did you look into [`esi_disable_xml_check`](http://stackoverflow.com/questions/25304475/esi-includes-in-json-response)?

whit537:
I'm going to get celery going so I can test out cache invalidation.

cwisecarver:
Yep, the startup command you ran for varnish is enabling that flag. It appears that just convinces varnish to process invalid XML at all. Varnish, like the old apache ESI, just expects you to be using x(ht)ml.

The JSONRenderer which you will see later, unwraps the quotes around the ESI tag and also unescapes it's quotes, causing the json returned by the app server to be invalid. You could try commenting the render method and it should just call it's parent. I'm 99% sure I did that after I enabled the esi_disable_xml_check but it's probably worth testing again.

whit537:
> the startup command you ran for varnish is enabling that flag.

Ah, okay. I see it now. I had looked in `default.vcl`.

> I'm 99% sure I did that after I enabled the esi_disable_xml_check but it's probably worth testing again.
 Show all checks

Okay, will do. Currently brew installing rabbitmq ...

whit537:
@cwisecarver Do you have a link to the discussion that led to the decision to adopt ESI? In particular, what is the performance target we're trying to achieve with this optimization?

whit537:
I have rabbitmq installed and `inv {rabbitmq,celery_worker}` running. I'm attempting to reproduce the `BAN` case.

whit537:
Do we have any kind of seed data or fake data facility? If so it's not jumping out at me on the README.

cwisecarver:
@whit537 I believe most of the discussions were in person. I've spoken to @brianjgeiger and @JeffSpies about the decision. They might have some insights on targets. 

After looking at modular odm and what optimizations I thought we could make there I thought that reverse-proxy caching would achieve a quicker (less coding, testing time) and potentially faster (more throughput, less cpu time) result than tweaking modm. I was also much more familiar with this stack and felt more comfortable tweaking DRF and varnish than I did tweaking our in-house ORM that nearly everything depends on. :smile:

cwisecarver:
@whit537 You okay with me emailing a dropbox link for a mongo dump from the staging server to your github-public email address?

whit537:
@cwisecarver Is it test data or production data?

cwisecarver:
Test data, but it's 137mb and I'm a nice github citizen.

whit537:
> Test data, but it's 137mb and I'm a nice github citizen.

:-)

Sure, chad@zetaweb.com.

whit537:
I'm trying to get a contributor change to propagate through celery. I'm updating my name at `/settings/`, and I do see a `POST` in the `inv server` console along with a logline from elasticsearch:

```
[website.search.elastic_search]  INFO: Updating node 89xbr
```

However, I don't see anything in the consoles for `inv celery_worker` or `inv rabbitmq`.

whit537:
I do have `USE_CELERY = True` in `website/settings/local.py`.

cwisecarver:
I disabled logging of tasks because it was breaking travisci. https://github.com/cwisecarver/osf.io/commit/0fc74d878a15f98ca3d01ce47fb848ec6a0091d0

So many things are changed in our unit tests (and on occasion we save early and often) that the logs get > 4MB and travis kills the build.

You can enable it by reversing that commit or you can run `varnishlog` in another terminal session and you should see the BAN request appear with all of it's particulars.

Any suggestions on leaving logging on without breaking travis or our centralized logging system when it's in production are more than welcome.

whit537:
> I disabled logging of tasks because it was breaking travisci.

Gotcha. I'm having trouble consistently reproducing the cache invalidation behavior, so I think I do need to watch more granular logs ...

cwisecarver:
Because it's a BAN instead of a PURGE there is occasionally a delay between when the BAN is issued and when the stale content will no longer be served. That's not what the docs say but it's been my experience within the first few requests after a BAN the content is refetched from the origin but PURGE requests happen immediately. https://www.varnish-cache.org/docs/3.0/tutorial/purging.html#purging-and-banning

whit537:
Breaking for lunch, bbiab ...

whit537:
I neglected to actually set `ENABLE_VARNISH` and `VARNISH_SERVERS` ...

whit537:
Alright, they're set, but I think it's time for me to understand what this code is actually doing in order to debug further. :)

cwisecarver:
Did I have those two settings documented in the PR? If not I should update it.

cwisecarver:
Updated first comment with examples for VARNISH_SERVERS and ENABLE_VARNISH settings.

whit537:
Thanks. I've got those. I'm unraveling how the signaling works between `website` and `api`. When I `POST` the profile edit form to port 5000, what's the flow out from there to celery and varnish? /me reading code ...

cwisecarver:
Anything that inherits from GuidStoredObject, when saved, fires a signal that is caught by blinker. That kicks off the BAN request in a celery task. It uses the method on the inheriting model that knows the absolute url for the v2 api. So, if you change a node with an ID of abcxy then the celery task should issue a ban to /v2/nodes/abcxy/. Because of the ESI anything that embeds that node in an Api response should be requesting that same url within varnish to embed it so every instance of that node should be banned. The cache key in varnish is determined based on your osf cookie value, any Authorization header you send, and the url. That means it should lower the hit rate but it should also only BAN *your* version of that url.

whit537:
Alright, I've got a `set_trace` in `api.caching.tasks.ban_url` and I'm trying to trigger it.

whit537:
I see that `api.caching.listeners.log_object_saved` enqueues `ban_url`.

whit537:
And `log_object_saved` is connected via a decorator to the `guid_stored_object_saved` signal.

whit537:
But isn't that all still within the API server process (`inv apiserver`)? What about the `inv server` process to which I'm `POST`ing a contributor update?

cwisecarver:
Are you not seeing it fire when you update things in the OSF? My understanding was that since they were both in the same `PYTHON_PATH` and guid_stored_object_saved **should** be fired in either context, that it would work for both the API and the OSF instance of the `model.save()`. I'm 99% sure I tested that case but this PR has been sitting for a while. It's possible I'm misremembering.

whit537:
Ah! [Here](https://github.com/cwisecarver/osf.io/blob/41139b20b78550982ee8242c18e24ad81e2d13ae/website/app.py#L28)'s an import of `api.caching.listeners` into `website.app`. Seems promising ...

whit537:
I can get inside of `log_object_saved` in the `inv server` process. However, `abs_url` is `None`. Investigating ...

```
(Pdb) l
 10         if hasattr(guid_stored_object, 'absolute_api_v2_url'):
 11             abs_url = guid_stored_object.absolute_api_v2_url
 12         else:
 13             #  I don't think this should ever happen, but ... just in case.
 14             typedModel = Guid.load(guid_stored_object._id).referent
 15  ->         if hasattr(typedModel, 'absolute_api_v2_url'):
 16                 abs_url = typedModel.absolute_api_v2_url
 17     
 18         if abs_url is not None:
 19             enqueue_task(ban_url.s(abs_url))
[EOF]
(Pdb) 
[EOF]
(Pdb) ban_url
<@task: caching_tasks.ban_url of __main__:0x1040c5150>
(Pdb) abs_url
(Pdb) n
> /Users/whit537/workbench/cos/osf.io/api/caching/listeners.py(18)log_object_saved()
-> if abs_url is not None:
(Pdb) n
--Return--
> /Users/whit537/workbench/cos/osf.io/api/caching/listeners.py(18)log_object_saved()->None
-> if abs_url is not None:
(Pdb) abs_url
(Pdb) 
```

brianjgeiger:
So, fun fact: celery tasks run synchronously through the API. The fix is most succinctly shown in https://github.com/CenterForOpenScience/osf.io/pull/4791/files (though the code that will get merged in is located in feature/admin-app-prod rather than this PR). Just in case that becomes important.

Also, if you're doing inspection of things that happen via the API, you will not have access to the Flask context from within an API request, so reverse lookups from Flask will not work. I haven't read closely enough to see if you're doing that, but it's a thing.

cwisecarver:
@brianjgeiger The signal that fires off the celery task is called here: https://github.com/CenterForOpenScience/osf.io/pull/4745/files#diff-146d9ba388e69688255458cc0948519dR98

The signal itself is here, only because it logically made sense to me to put it here, https://github.com/CenterForOpenScience/osf.io/pull/4745/files#diff-bfb338bacba071d5c112bd8842d24ab8R6

And the celery task itself is here: https://github.com/CenterForOpenScience/osf.io/pull/4745/files#diff-a74ffc85f1b853e07331e93ebf2864e3R20

Since both the API and the flask app use the models that inherit from GuidStoredObject I assume this would have to be able to run successfully in both contexts.

Maybe I need to move my `ENABLE_VARNISH` setting into `website.settings.local` instead of `api.base.settings.local` since the latter imports the former?

whit537:
Thanks @brianjgeiger. I will keep that in mind.

cwisecarver:
Since the models have `absolute_api_v2_url` methods on them I can't imagine they wouldn't work in both the flask and api contexts.

whit537:
> Since the models have `absolute_api_v2_url` methods on them

Ah, those are methods? The code in `log_object_saved` indicated to me that it was a property, so I've been looking for it in a contributor document in Mongo. I don't find a `contributors` collection or similar, however.

whit537:
I find a few instances, but it still looks weird ... method or attribute?

```
api/caching/listeners.py
10:    if hasattr(guid_stored_object, 'absolute_api_v2_url'):
11:        abs_url = guid_stored_object.absolute_api_v2_url
15:        if hasattr(typedModel, 'absolute_api_v2_url'):
16:            abs_url = typedModel.absolute_api_v2_url

framework/auth/core.py
416:    def absolute_api_v2_url(self):
422:        return self.absolute_api_v2_url

website/oauth/models/__init__.py
532:    def absolute_api_v2_url(self):
537:        return self.absolute_api_v2_url
598:    def absolute_api_v2_url(self):
603:        return self.absolute_api_v2_url

website/project/model.py
198:    def absolute_api_v2_url(self):
203:        return self.absolute_api_v2_url
2297:    def absolute_api_v2_url(self):
2306:        return self.absolute_api_v2_url
```

cwisecarver:
They're decorated properties, not actually database fields. https://github.com/CenterForOpenScience/osf.io/blob/develop/website/project/model.py#L198

I misspoke when I called them methods.

whit537:
Gotcha. I'm seeing that now. Looks like [`get_absolute_url`](https://github.com/cwisecarver/osf.io/blob/41139b20b78550982ee8242c18e24ad81e2d13ae/framework/auth/core.py#L421-L422) is the method that dereferences it.

cwisecarver:
Yes, I specifically avoided calling `get_absolute_url` because that has special meaning to django and I assumed that one day we might not want it pointing to the api url but to ... literally anything else. :wink:

whit537:
A clue!

```
(Pdb) l
  9         abs_url = None
 10         if hasattr(guid_stored_object, 'absolute_api_v2_url'):
 11             abs_url = guid_stored_object.absolute_api_v2_url
 12         else:
 13             #  I don't think this should ever happen, but ... just in case.
 14  ->         typedModel = Guid.load(guid_stored_object._id).referent
 15             if hasattr(typedModel, 'absolute_api_v2_url'):
 16                 abs_url = typedModel.absolute_api_v2_url
 17     
 18         if abs_url is not None:
 19             enqueue_task(ban_url.s(abs_url))
(Pdb) guid_stored_object
<User(u'chad@zetaweb.com') with id u'ty4ef'>
(Pdb) guid_stored_object.absolute_api_v2_url
*** ImproperlyConfigured: Requested setting ROOT_URLCONF, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.
(Pdb)
```

cwisecarver:
That makes me think that absolute_api_v2_url doesn't work outside of django context...

whit537:
I'm trying to think if I made a change to my local environment under #4811 that would trigger this bug for me but not others.

cwisecarver:
I don't believe this, https://github.com/CenterForOpenScience/osf.io/blob/develop/api/base/utils.py#L49, will work in the flask context.

cwisecarver:
I may have to consult with @brianjgeiger about the best way to get an v2 api url inside of flask context and then adjust that method.

whit537:
I found the `user` collection in Mongo, btw. Contributors are users.

brianjgeiger:
@cwisecarver Generically and absolute or by model and/or relative? As a general rule, I recommend, essentially, hard-coding the path, because that's the thing that's guaranteed not to change (well, you know, once we finally announce that we're stable), but I'm not sure if that's what you need for this.

cwisecarver:
@brianjgeiger I need a method that would be on each model that inherits from GuidStoredObject that would deliver a relative url. Absolute would be better but I can sort out the host per environment. 

I could hard-code the url but I have several public comments on github admonishing people for that in PRs. :disappointed: 

It's basically the URL that the celery task will send a BAN request to. So if node `abcxy` is changed then an http request needs to be made like `BAN https://staging-api.osf.io/v2/nodes/abcxy`

whit537:
> I could hard-code the url but I have several public comments on github admonishing people for that in PRs. :disappointed:

GitHub comments are editable? :-P

brianjgeiger:
Heh. For Flask and v1 routes, we definitely shouldn't be hand-coding routes. For v2 routes, frankly, the more hand coding the better, because it will discourage people from wanting to change them.

cwisecarver:
@brianjgeiger They would all be v2 urls. I will still feel bad about it though. 

@whit537 GitHub comments are editable but that almost means I'd have to stop telling people not to do it in the future or go back and change my comments after the PR is closed.

brianjgeiger:
Yeah, it's weird, and the big downside is that you can't use your IDE to refactor. But it is the canonical reference, so… All that being said, in two years or fewer I may be saying what a terrible mistake I made recommending that. We'll see.

cwisecarver:
[This](https://github.com/CenterForOpenScience/osf.io/blob/04056a550b2585b01c01183f473f0c166c3b8a15/website/util/__init__.py#L93) would at least handle environmentalization for me. Maybe ... I could put a property on all the models with the proper url path pattern for the api_v2 url?

whit537:
@cwisecarver Yeah, I was kidding, sorry. :mouse:

brianjgeiger:
Oh, right, back to the absolute thing. Sorry, I forgot when we starting talking about feelings. Yes, you should use that to generate your absolute URLs. And yeah, a property would be fine, I *guess*.

cwisecarver:
@whit537 I got it :smile: I just chose to use that opportunity to make @brianjgeiger feel bad to suggesting I hardcode things.

whit537:
:stuck_out_tongue_winking_eye:

brianjgeiger:
As well you should have.

whit537:
@cwisecarver Why hadn't you hit the django configuration issue already?

cwisecarver:
@whit537 I must have not tested the flask side. I was very focused on getting the api to be snappy. 

Good catch. That would have been tricky to find when it was in production.

whit537:
Also: does it help at all if you limit the hard-coding to only take place beneath `absolute_api_v2_url`? Then it's in one place and you can change it if need be (two years from now as @brianjgeiger says).

cwisecarver:
Okay, I'll add the properties and adjust [this function](https://github.com/CenterForOpenScience/osf.io/pull/4745/files#diff-bfb338bacba071d5c112bd8842d24ab8R10) to use it. I also just realized that the celery task is using one url and the to_esi_representation is using another one. They **should** be the same but if anyone changed get_absolute_url to not reference that property then it all would have broken.

cwisecarver:
That might be a better option @whit537. What do you think @brianjgeiger?

whit537:
Okay. I'm gonna see about this data dump you sent and some performance benchmarks.

cwisecarver:
When I was testing I used [siege](https://www.joedog.org/siege-home/) installable via homebrew.

Here's a urls file that's based off that dataset I sent you.
[urls.txt](https://github.com/CenterForOpenScience/osf.io/files/74361/urls.txt)

whit537:
Okay. I've used [wrk](https://github.com/wg/wrk) successfully in the past. I'll check out Siege since you've already got the URLs file for it and it'll give us more comparable results across our environments.

cwisecarver:
Definitely feel free to use your tools. I just thought, regardless, having a list of urls that mostly shouldn't 500 would be useful. :+1:

cwisecarver:
@whit537 Tomorrow morning I'm going to work through updating the `absolute_api_v2_url` properties to work regardless of django or flask context. I can create a new PR or I can just warn you before I push things. Whatever is easier for you.

brianjgeiger:
I'm good with whichever works best for you. I'm only marginally fussy about using api urls canonically.

whit537:
No need to create a new PR. I'll watch for more commits here.

I'm seeing a lot of 500s with siege. Example below. Probably something obvious in my environment, might get it squared before EOD here ...

screen shot 2015-12-29 at 5 03 21 pm

```
done.
siege aborted due to excessive socket failure; you
can change the failure threshold in $HOME/.siegerc

Transactions:                    683 hits
Availability:                  39.83 %
Elapsed time:                 217.52 secs
Data transferred:             123.67 MB
Response time:                  3.47 secs
Transaction rate:               3.14 trans/sec
Throughput:                     0.57 MB/sec
Concurrency:                   10.88
Successful transactions:          10
Failed transactions:            1032
Longest transaction:           10.03
Shortest transaction:           0.00
 
FILE: /usr/local/var/siege.log
You can disable this annoying message by editing
the .siegerc file in your home directory; change
the directive 'show-logfile' to false.
```

cwisecarver:
I've honestly never seen a HideIfRetraction object. My gut tells me it was conceived in between me working on this and it actually getting to CR.

whit537:
@cwisecarver Browsing through [`api/base/serializers.py`](https://github.com/cwisecarver/osf.io/blob/41139b20b78550982ee8242c18e24ad81e2d13ae/api/base/serializers.py) ... could we implement `to_esi_representation` generically? It seems like unneeded friction to have to remember to implement it for each new serializer field separately.

cwisecarver:
@whit537 I would really like to do that. I could create a mixin that has the method that both TargetField and RelationshipField inherit from.

whit537:
> both TargetField and RelationshipField inherit from

And now `HideIfRetraction`, too, ya?

whit537:
I stubbed out `HideIfRetraction.to_esi_representation` and siege appears to find many more 200s than previously. There's also a healthy number of 400s and 410s: is this expected?

cwisecarver:
The 400's, not sure. The 410s should be for objects that have been marked as deleted.

whit537:
Okay. There are still a few 500s as well, e.g.:

screen shot 2015-12-30 at 10 05 11 am

cwisecarver:
There's a HideIfRegistration too. It might be time for a mixin.

whit537:
> There's a HideIfRegistration too.

My bad. That's what I meant. `HideIfRetraction` already subclasses `HideIfRegistration`.

cwisecarver:
Ah, okay.

whit537:
Siege is still chuggin' along ...

cwisecarver:
I'm working through adding to_esi... to HideIfRegistration. Just trying to figure out what that actually **does** first.

brianjgeiger:
I think it was used for when we had registrations in the node list. It may be vestigial at this point.

api/base/serializers.py
```diff
@@ -716,7 +731,7 @@ def to_representation(self, obj, envelope='data'):
         ])
 
         embeds = self.context.get('embed', {})
-
+        esi = self.context.get('esi', {})
```
whit537: This makes me think that some value in `esi` will be dereferenced below, when in fact it's just a boolean. I think it'd be more readable in the future if it were `enable_esi` and the default were `False` instead of `{}`.

api/caching/tasks.py
```diff
@@ -0,0 +1,40 @@
...
+class VarnishTask(celery.Task):
+    abstract = True
+    max_retries = 5
+
+def get_varnish_servers():
+    #  TODO: this should get the varnish servers from HAProxy or a setting
+    return settings.VARNISH_SERVERS
+
+@celery_app.task(base=VarnishTask, name='caching_tasks.ban_url')
+# @logged('ban_url')
```
whit537: I generally think of commented-out code as a bad smell. Is it common practice for OSF?

cwisecarver: No, I left it in when debugging. I will remove.

api/base/serializers.py
cwisecarver: Will change.

api_tests/caching/test_caching.py
```diff
@@ -0,0 +1,236 @@
...
+                    self.validate_keys(varnish_authed_resp.json(), original_embed_values)
+                except AssertionError as ex:
+                    print 'Validation failed varnish_url authed: {} - {}\n\n\n'.format(varnish_url, ex.message)
+                except Exception as ex:
+                    varnish_authed_data[key]['_'.join(embed_values)] = dict(error=ex.message)
+                    print 'Failed to get and validate varnish_url authed: {} - {}\n\n\n'.format(varnish_url, ex.message)
+
+                embed_values.pop()
+
+        # Uncomment these lines to write out files containing responses for comparison
```
whit537: More commented out code. Consider moving to a function if you think it's valuable to keep around for future testing.

cwisecarver: rgr, I'll remove it. I didn't put datadiff in the requirements because that was it's only usage so uncommenting it would fail anyway.

api/base/serializers.py
whit537: thx

api/caching/tasks.py
whit537: cool

api_tests/caching/test_caching.py
whit537: :+1:

whit537:
Okay! I think you're topped up on additional changes here, ya @cwisecarver? I'm gonna turn my gaze elsewhere, ping me when you're ready for more review here.

whit537:
Siege found it to be much more available with the stubbed-out `to_esi_representation`:

```
done.
siege aborted due to excessive socket failure; you
can change the failure threshold in $HOME/.siegerc

Transactions:                 211833 hits
Availability:                  99.52 %
Elapsed time:                8734.67 secs
Data transferred:             666.77 MB
Response time:                  0.06 secs
Transaction rate:              24.25 trans/sec
Throughput:                     0.08 MB/sec
Concurrency:                    1.36
Successful transactions:      122297
Failed transactions:            1024
Longest transaction:         1119.66
Shortest transaction:           0.00

FILE: /usr/local/var/siege.log
You can disable this annoying message by editing
the .siegerc file in your home directory; change
the directive 'show-logfile' to false.
```

cwisecarver:
@brianjgeiger @whit537 In playing around with the reverse url stuff I realized that if I start flask with the environment variable DJANGO_SETTINGS_MODULE="api.base.settings" then everything works. I can't think of any reason why that would interfere with flask. Thoughts?

cwisecarver:
@whit537 I think I've corrected all the comments above. I also moved the save signal up a class from `GuidStoredObject` to `StoredObject` because Files weren't being `BAN`ned. 

Regarding my comment above about the `DJANGO_SETTINGS_MODULE` env variable, I'm wandering the office in my spare time gathering opinions on whether this could create some sort of time paradox where we're all turned into ... something else. I can't think of anything that would make this bad unless somewhere flask is expecting this variable to be something else which would be almost pure evil.

I also removed the code that would take a second try if an object scheduled for `BAN`ning didn't have a `absolute_api_v2_url` property in https://github.com/CenterForOpenScience/osf.io/commit/71f8ed10bc044566d0762c0be40f137c757bc537. Now it just doesn't try if there's no `absolute_api_v2_url` on the object. I would have it log but many things inherit from StoredObject that are not in the API.

whit537:
@cwisecarver Cool, thanks. My head is in #4509 right now, I'll circle back here in a bit ...

cwisecarver:
Sounds good @whit537. Thanks for taking the time. I checked with a few of the village elders about the env variable. Seems like nobody can think of a reason we **shouldn't** do it so I'm going to talk to devops about getting it put in our deployment config, at least for staging. It should be safe to go out before this code as long as it goes out before this code.

Right now, I'm using the following commands with great success.
- Run tests: `DJANGO_SETTINGS_MODULE="api.base.settings" inv test`
- Run flask app: `DJANGO_SETTINGS_MODULE="api.base.settings" inv server`

whit537:
I'm planning to work on this tomorrow (Tuesday).

whit537:
@cwisecarver Needs a rebase.

whit537:
Alright, took a pass through the latest commits. Gonna get back to `BAN`ning ...

whit537:
I get a 500 when I try to modify a node. I dropped and recreated the database.

screen shot 2016-01-05 at 10 41 55 am

cwisecarver:
Do you have the following setting in your local.py?
`VARNISH_SERVERS = ['http://localhost:8193']`

whit537:
Yeah, in my `api/base/settings/local.py`, right?

cwisecarver:
Correct.

On second though, I feel like that error is unrelated to my changes. It looks like the url generation is failing on the client side.

whit537:
@cwisecarver I'm way before needing that, though.

> On second thought ...

Okay ...

whit537:
Gosh. Now I get a 400(!).

I dropped and recreated the database, and bounced `inv {server,apiserver,celery}`.

whit537:
Then I registered, created a project, and tried to edit the title.

whit537:
A clue!

> errors: [{detail: "Please confirm your account before using the API."}]

whit537:
Interesting. Usually when I copy and paste the registration URL from the `inv server` console, I get an "Invalid token" error. But I did it just now and it worked. Maybe a celery task needs to have run?

screen shot 2016-01-05 at 10 51 01 am

cwisecarver:
Not sure... @icereval, any idea on why his registration token wouldn't ever ... work?

whit537:
Interesting that I can *create* a project without verifying my email, but I can't modify one. Seems like a bug. Also, the error message is unhelpful. Reticketed as [OSF-5487](https://openscience.atlassian.net/browse/OSF-5487) and [OSF-5489](https://openscience.atlassian.net/browse/OSF-5489).

whit537:
Despite having seen the "Welcome to the OSF!" message, I still get the "Please confirm your account before using the API." message when trying to change the title of a project. I don't see any indication on my profile page whether my account is verified or not.

screen shot 2016-01-05 at 11 05 09 am

whit537:
Ah! Here it is:

screen shot 2016-01-05 at 11 05 57 am

whit537:
Alright, I'm going to debug this verification [issue](https://github.com/CenterForOpenScience/osf.io/pull/4509#issuecomment-168057300) I'm [having](https://github.com/CenterForOpenScience/osf.io/pull/4745#issuecomment-169041442) ...

whit537:
Blorg. I registered a new account, copied and pasted the confirmation URL, and ... it worked. :flushed: 

Okay!

whit537:
Now I'm back to "Invalid Schema" (https://github.com/CenterForOpenScience/osf.io/pull/4745#issuecomment-169038417).

whit537:
Hah! That's *URL* schema, not database schema. The error is inside the requests library. Here's the end of the traceback:

```
File "/Users/whit537/workbench/cos/osf.io/api/caching/tasks.py" in ban_url
  33.                 Host=parsed_url.hostname
File "/Users/whit537/workbench/cos/osf.io/env/lib/python2.7/site-packages/requests/api.py" in request
  49.     response = session.request(method=method, url=url, **kwargs)
File "/Users/whit537/workbench/cos/osf.io/env/lib/python2.7/site-packages/requests/sessions.py" in request
  461.         resp = self.send(prep, **send_kwargs)
File "/Users/whit537/workbench/cos/osf.io/env/lib/python2.7/site-packages/requests/sessions.py" in send
  567.         adapter = self.get_adapter(url=request.url)
File "/Users/whit537/workbench/cos/osf.io/env/lib/python2.7/site-packages/requests/sessions.py" in get_adapter
  646.         raise InvalidSchema("No connection adapters were found for '%s'" % url)
```

Probably a misconfiguration in my environment? ...

cwisecarver:
@whit537 I think it must be. Since there's no schema in front of the colon requests doesn't know how to make the request.

whit537:
Right. Time for me to [grok URL construction](https://github.com/cwisecarver/osf.io/commit/71f8ed10bc044566d0762c0be40f137c757bc537#commitcomment-15267045). :-)

cwisecarver:
That's why I was pointing you to that VARNISH_SERVERS setting. That's where I thought the error was. You could put a breakpoint in the api.caching.tasks file and see what's getting passed into it. It should be pulling from that setting.

whit537:
Ah, gotcha. Yeah, getting in there now ...

whit537:
> Time for me to [grok URL construction](https://github.com/cwisecarver/osf.io/commit/71f8ed10bc044566d0762c0be40f137c757bc537#commitcomment-15267045). :-)

Or not. You were right: it was `VARNISH_SERVERS`. I had set it without the scheme. Hey! Why does Requests say "InvalidSchema" instead of "InvalidScheme"?

cwisecarver:
I think, by schema, it means your url didn't match their regex pattern. I think it was a coincidence that scheme was involved in the failure. Also, I think my rebase/merge did something horrible so I'd suggest not pulling right now.

whit537:
@cwisecarver https://github.com/kennethreitz/requests/pull/512 suggests otherwise. We shall see! https://github.com/kennethreitz/requests/pull/2960 :-)

whit537:
> Also, I think my rebase/merge did something horrible so I'd suggest not pulling right now.

Heh. Roger that! :-)

whit537:
Blam. :-)

screen shot 2016-01-05 at 11 34 56 am

whit537:
P.S. @cwisecarver When we [added](https://github.com/CenterForOpenScience/osf.io/pull/4745#issuecomment-167858209) `VARNISH_SERVERS` to the PR description, we didn't include the scheme:

> Set the `VARNISH_SERVERS` setting in your `api/base/settings/local.py` to `['127.0.0.1:8193']`

cwisecarver:
@whit537 I'm basically the worst. I'll update it. Sorry for the trouble.

whit537:
No trouble! Nice to have the chance to submit a PR over on Requests. :smiley: :bow:

whit537:
And now I get the `inv apiserver` logging as described in steps 7 and 8!

@cwisecarver++

whit537:
Step 9 confirmed! Okay! That completes my first pass through this PR. :smile_cat: 

From here I want to explore the `*absolute*_url*` scenario, and also dig into the `BAN` specifics a bit. I should probably also re-run siege. I'd also like to work through the blinker logic a bit.

whit537:
I'm reading https://www.varnish-cache.org/docs/4.1/users-guide/purging.html (after confirming that I'm running `varnishd` version 4.1 locally).

whit537:
> BANning instead of PURGEing means that we can avoid the thundering herd problem that commonly occurs with cache invalidation and it also means that we're not actually emptying the cache until we need to.

I'm wrapping my head around this. ^^^

whit537:
Actually ...

/me starts up `siege` first ...

whit537:
... or not. I deleted da db dump like a doofus, and dropped da database deleteriously. :disappointed:

whit537:
All my siege are 404.

cwisecarver:
Still have the email with the dropbox link to the dump? I can resend.

whit537:
@cwisecarver I do have the email, but the dropbox link is coming up 404. I guess I assumed that you had taken it down. No?

cwisecarver:
Sent you a new email. Dropbox auto-expires links so I don't have to.

whit537:
Ah, word. Thanks! Got it again.

cwisecarver:
@whit537 As I understand it, the ban-purge business boils down to: Purge actively and immediately removes an item from cache. Ban lazily marks the cached object as old-and-busted and when that object is next requested the previous version is removed and replaced with a new version which is also delivered to the user. There is also a ban-lurker process that wanders the cache looking for things marked as "banned" when it finds them it will preemptively remove them from the cache prior to them being requested.

whit537:
> There is also a ban-lurker process that wanders the cache looking for things marked as "banned" when it finds them it will preemptively remove them from the cache prior to them being requested.

And this is what makes bans less susceptible to the thundering herd?

cwisecarver:
I believe it's less susceptible to that because varnish will serve the banned version until it gets the new one. So if index.html is marked as banned and 600,000 requests come in for it at the same time varnish will start fetching it from the origin and deliver the old one to everyone else while it's waiting instead of sending 600,000 requests to the origin to update the cache for that url.

cwisecarver:
If it's not in the cache at all, because it's been purged, it has no way to know so it will just proxy all those requests to the origin.

cwisecarver:
Then :sob:

whit537:
Gooottttccha. Okay, cool. Yay, `BAN`! :-)

cwisecarver:
@whit537 This is old, but a pretty good overview.

http://itsallabtamil.blogspot.com/2012/07/varnish-cache-purge-ban-and-ban-lurker.html

whit537:
Alright, sorry, got [distracted](https://github.com/CenterForOpenScience/centerforopenscience.org/pull/579
) [with](https://github.com/CenterForOpenScience/centerforopenscience.org/issues/580) [the](https://openscience.atlassian.net/browse/NCP-357) [prereg](https://openscience.atlassian.net/browse/NCP-358) [challenge](https://openscience.atlassian.net/browse/NCP-359)! So exciting! :dancer:

Back here now ...

whit537:
If `BAN` and `PURGE` work as you describe, then `BAN` sounds like the right choice.

whit537:
> I believe it's less susceptible to that because varnish will serve the banned version until it gets the new one.

I'm not seeing this mentioned in [the documentation](https://www.varnish-cache.org/docs/4.1/users-guide/purging.html). Let's check the source! Looks like [BAN_CheckObject](https://github.com/varnish/Varnish-Cache/blob/varnish-4.1.0/bin/varnishd/cache/cache_ban.c#L894-L972) is the starting point ...

whit537:
I'm having trouble lining up the code with [this SVG](https://github.com/varnish/Varnish-Cache/blob/varnish-4.1.0/doc/sphinx/phk/brinch_hansens_arrows_1.svg), but in the code I see this call chain:

```
cnt_lookup
cnt_purge
  HSH_Lookup
    BAN_CheckObject
```

These are starting to line up with the [doc on states](https://www.varnish-cache.org/docs/4.1/reference/states.html).

whit537:
[](https://www.varnish-cache.org/docs/4.1/_images/cache_req_fsm.svg)

cwisecarver:
I am also a little distracted by the prereg challenge. I could be confused about how the BAN command works regarding it's laziness. I thought I read that somewhere.

This is an html version of the pdf manual. This seems better than the RTD and more readable than the source (to me at least).

https://www.varnish-software.com/book/4.0/chapters/Cache_Invalidation.html

whit537:
> This is an html version of the pdf manual. This seems better than the RTD and more readable than the source (to me at least).

 Ah, very cool. Thanks! I've jumped through their email hoop to download [the latest PDF](https://github.com/CenterForOpenScience/osf.io/files/79109/varnish-book.pdf). Let's see what it says ...

cwisecarver:
I should have warned you. They're going to call and email you now to see if
you'd be interested in buying their newest pro version.

whit537:
That's called takin' one for the team, I believe. ;-)

whit537:
Upon skimming I don't find a mention of `BAN`'s willingess to keep serving stale content. I do find this table:

screen shot 2016-01-05 at 5 40 08 pm

cwisecarver:
It appears that all signs point to me making it up. The documentation does
say that it helps with the thundering herd problem, right? Maybe my brain
just decided that's how it helped.

whit537:
Aaaand with that, I'm out of time for today. Looking back at [the list of things I wanted to look at](https://github.com/CenterForOpenScience/osf.io/pull/4745#issuecomment-169057609), I'm not seeing any as blockers:

 - I think `BAN`s are probably fine. Sounds like you've done your homework, anyway. We can adjust down the line once we have some real-world experience.
 - I saw your [comment](https://github.com/cwisecarver/osf.io/commit/71f8ed10bc044566d0762c0be40f137c757bc537#commitcomment-15269697) about URL construction and indeed it looks like a big ol' chunk of tech debt that's well out of scope for this PR. Let's make a ticket about it though if we don't already have one.
 - I lost my `siege` run when my battery died. I'm satisfied with the run I [did](https://github.com/CenterForOpenScience/osf.io/pull/4745#issuecomment-168043536) with a stubbed-out `to_esi_representation` earlier.
 - I don't need to fully audit our blinker usage here. That area seems low-risk to me.

My plan was to not work again for COS until Friday, so if @sloria et al. want to keep this moving then consider it :+1: from me! Good work, @cwisecarver! :dancer:

cwisecarver:
I mostly chose ban over purge because of the regex syntax. If a node is
invalidated in the cache the things that belong to the node should to.

whit537:
> It appears that all signs point to me making it up. The documentation does say that it helps with the thundering herd problem, right? Maybe my brain just decided that's how it helped.

I'm happy to pick up with this on Friday if the question is still hanging out there, but per my other comment I don't see it as a blocker.

whit537:
> I mostly chose ban over purge because of the regex syntax. If a node is invalidated in the cache the things that belong to the node should to.

Yeah, worst-case is that it behaves the same as a PURGE (immediately clearing the cache, not serving stale content ever), in which case a BAN is not *worse* than a PURGE here.

whit537:
Okay! I'm out. :-)

@cwisecarver++
prereg++
COS++
science++

cwisecarver:
Thanks @whit537!

cwisecarver:
I'm 99% sure travis is failing because of something in develop when I merged last. I'll loop back and re-merge develop this afternoon.

sloria:
@cwisecarver Yep, the failing test has since been fixed on develop.

api_tests/caching/test_caching.py
```diff
@@ -0,0 +1,190 @@
+from __future__ import unicode_literals
+import copy
+
+import requests
+from django.conf import settings
+from requests.auth import HTTPBasicAuth
+
+
+class TestVarnish(object):
```
sloria: Will nose find this if it's not a subclass of `unittest.TestCase`?

api_tests/caching/test_caching.py
```diff
@@ -0,0 +1,190 @@
...
+    local_varnish_base_url = 'http://localhost:8193/v2/'
+    local_python_base_url = 'http://localhost:8000/v2/'
+
+    def setUp(self):
+        if not settings.ENABLE_VARNISH:
+            return
+        self.authorization = HTTPBasicAuth('mocha@osf.io', 'password')
+
+    def test_compare_python_responses_to_varnish_responses(self):
+        if not settings.ENABLE_VARNISH:
```
sloria: Minor: You can use `unittest.skipif(not settings.ENABLE_VARNISH, '...')` so that the test output will show that this test was explicitly skipped.

api_tests/caching/test_caching.py
```diff
@@ -0,0 +1,190 @@
...
+                python_authed_resp = requests.get(python_url, auth=self.authorization, timeout=120)
+
+                varnish_authed_resp = requests.get(varnish_url, auth=self.authorization, timeout=120)
+
+
+                try:
+                    python_data[key]['_'.join(embed_values)] = python_resp.json()
+                    self.validate_keys(python_resp.json(), original_embed_values)
+                except AssertionError as ex:
+                    print 'Validation failed python_url unauthed: {} - {}\n\n\n'.format(python_url, ex.message)
```
sloria: Should the test fail at this point?

framework/tasks/handlers.py
```diff
@@ -43,7 +43,7 @@ def enqueue_task(signature):
     try:
         if signature not in queue:
             queue.append(signature)
-    except RuntimeError:
+    except (RuntimeError, AttributeError):
```
sloria: Can you explain why an `AttributeError` needs to be caught here? It looks like `queue` should always be a list at this point.

sloria:
IIRC, the different URL methods and properties were supposed to serve distinct purposes:

- `absolute_url`: The URL to the html representation of the object - `http://osf.io/abc12/`
- `absolute_v2_url`: The APIv2 URL - `http://api.osf.io/v2/nodes/abc12/`
- `get_absolute_url`: Same as absolute_v2_url, but included for DRF compatibility (?). I believe DRF's serializers use this method internally.

sloria:
Pass finished. Looks good from a code standpoint. I didn't do a functional review since Chad has done a fine job at that. Just had a few questions for clarification above. Once those are addressed and the build is fixed, this should be good to merge.

api_tests/caching/test_caching.py
cwisecarver: This is a great idea. I changed it. Thanks.

whit537:
> Let's make a ticket about it though if we don't already have one.

https://openscience.atlassian.net/browse/OSF-5519

framework/tasks/handlers.py
cwisecarver: I just took this out. I figured out why it was there and it's no longer needed.

tasks/__init__.py
```diff
@@ -504,6 +519,10 @@ def test_travis_else():
     karma(single=True, browsers='PhantomJS')
 
 @task
+def test_travis_varnish():
```
sloria: This probably isn't necessary.

tasks/__init__.py
```diff
@@ -459,6 +464,16 @@ def test_api():
     test_module(module="api_tests/")
 
 @task
+def test_varnish():
+    """Run the Varnish test suite."""
+    """Run the OSF test suite."""
```
sloria: Remove stray docstring.

sloria:
I see that `ENABLE_VARNISH` has not been added to `website/settings/defaults.py`, which requires that the setting is configured in `local.py`. Is this intentional?

tasks/__init__.py
cwisecarver: @sloria This one is here because it matches the pattern in travis. There's a test_travis_osf  and a test_travis_else too. I can remove test_varnish if you think that's worth it.

cwisecarver:
@sloria It is, most developers won't have varnish installed or enabled.

api/caching/tests/test_caching.py
```diff
@@ -0,0 +1,242 @@
...
+
+                try:
+                    python_data[key]['_'.join(
+                        embed_values)] = python_resp.json()
+                    self.validate_keys(python_resp.json(),
+                                       original_embed_values)
+                except Exception as ex:
+                    python_data[key]['_'.join(embed_values)] = dict(
+                        error=ex.message)
+                    print 'Failed to get python_url unauthed: {} - {}\n\n\n'.format(
```
sloria: Should the test fail if we get here?

sloria:
@cwisecarver Ah woops, disregard previous comment--just saw it was an in the `api.settings` module, not `website.settings`.

tasks/__init__.py
sloria: `test_travis_osf` and `test_travis_else` only exist to make it easier to split Travis runs into multiple, parallel builds. `test_travis_varnish` doesn't really contribute to that, so it can probably be removed.

api/caching/tests/test_caching.py
cwisecarver: @sloria It probably should since the assertions wouldn't actually be running. I wasn't sure what to do in this situation. It would indicate a misconfiguration or a problem with travis rather than an **actual** test failure.

tasks/__init__.py
sloria: Oh, but I just noticed that the varnish tests run as a separate parallel build. I see what you mean now. OK, I think this fine either way.

cwisecarver: Yeah, I was just trying to match the existing standard. It looks like test_travis_else and test_travis_osf end up calling other methods and just acting as a wrapper. test_travis_varnish only has one call but keeping test_varnish would allow people to call it from their development machines and not have to think about what the differences between test_travis_varnish and test_varnish are.

sloria:
The Travis logs for the Varnish tests get clipped due to this unhelpful warning:

job__12324_3_-_centerforopenscience_osf_io_-_travis_ci

Would be worth silencing that log so that we can get useful output from the tests.

cwisecarver:
@sloria Is there a way to do this selectively for this test run or something? That warning doesn't belong to me, I didn't put it in or know it's import.

api/caching/tests/test_caching.py
sloria: I see. So maybe it makes sense to remove the exception handling then?

sloria:
@cwisecarver It comes from here: https://github.com/CenterForOpenScience/osf.io/blob/60a3b26fe9dc49cf68b8edf6f646136b8fd447fa/website/project/model.py#L1121 

You can either

1. Add `website.project.model` to the list of silent loggers [here](https://github.com/CenterForOpenScience/osf.io/blob/60a3b26fe9dc49cf68b8edf6f646136b8fd447fa/tests/base.py#L68-L76)
2. Remove the logging statement

I am fine with either.

sloria:
@cwisecarver Can you get this up to date with develop one last time? Once that's done, this should be good to merge. @mattclark Please prep deployment of this to staging.

cwisecarver:
Yep, working on that now @sloria just running tests to make sure before I burden travis.

sloria:
Great!

cwisecarver:
Assuming the tests pass, which they just did on local, should be GTG @sloria. Thanks for helping me polish it.

cwisecarver:
:shipit: :+1:

icereval:
:+1:

whit537:
!m @cwisecarver

PR #4748, opened 2015-12-15
Updates for Wiki Drag and Drop

ticklemepierce:
Purpose
-
For the sake of future readability some variables were renamed and then a bunch of code was moved to a new file.

Changes
-
See above

Side effects
-
Should not be any if done correctly.


sloria:
Much better. Thanks

PR #4750, opened 2015-12-15
[Feature Fix] Fix ContribAdder double-bind

samchrisinger:
# Purpose

Since the ContribAdder relies on a single static modal for rendering and the Prereg authorImport extension potentially instantiates the Adder multiple times, we were getting double-binding errors from KO.

# Changes

Clean the node before instantiating the ContribAdder in authorImport extension.


sloria:
Cherry-picked into `release/0.60.0`. Please send future fixes to the release branch.

PR #4751, opened 2015-12-15
[Feature] Added tests for alternative citation forks and fixed a related bug

ticklemepierce:
Purpose
-
Added a few tests to ensure that forks work properly with alternative citations

Changes
-
Added the appropriate tests.  As a result I found that forks were failing so I added a check in the add_citation to fix this.

Side Effects
-
Working forks and tests to make sure they continue working.


sloria:
Cherry-picked into `release/0.60.0`. Please send future fixes to the release branch.

PR #4752, opened 2015-12-15
Updated CSS selector to fix menus not working

ticklemepierce:
Purpose
-
Menus were broken on the contributors page as a result of my PR.  This PR fixes that.

Changes
-
This was due to a bad CSS selector which effectively disabled buttons.  I fixed this CSS selector to only select the ones that were necessary.

Side Effects
-
Should be none


sloria:
Cherry-picked the commit into `release/0.60.0`. Please send future feature fixes to the release branch.

PR #4753, opened 2015-12-15
Changing a node's privacy setting removes all tags on the node

pattisdr:
# Purpose

Issue https://openscience.atlassian.net/browse/OSF-5403

When doing a bulk PATCH request, self.partial, a DRF thing, was set to False.  When the request hit the update method in the node serializer, self.partial, being set to False, acted as if this was a PUT request and deleted all the tags.

# Changes

In the JSONAPIListSerializer, if this is a bulk PATCH request, set the child serializer self.partial equal to True.

**Basically, the request is now correctly treated as a PATCH request instead of a PUT request when it comes to tags.**  PATCH was only broken for tags, because the tags logic was specifically looking for the self.partial flag.

### DIFF
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -615,6 +615,11 @@ class JSONAPIListSerializer(ser.ListSerializer):
 
     # Overrides ListSerializer which doesn't support multiple update by default
     def update(self, instance, validated_data):
+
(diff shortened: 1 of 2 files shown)


mfraezz:
Pass complete, no bugs found :dizzy:

sloria:
Merged into release/0.60.0

PR #4755, opened 2015-12-15
Remove ability to serialize registrations through /nodes/

brianjgeiger:
Purpose
-----------
Closes https://openscience.atlassian.net/browse/OSF-5359

Changes
------------
Removes ability to access registrations through a /nodes/ endpoint. Updates a bunch of tests (many of which changing an expected 403 to a 404)

Side effects
----------------
Node lists will be *much* faster


api_tests/nodes/views/test_node_alternative_citation_list.py
```diff
@@ -748,16 +748,21 @@ def test_add_citation_logged_out_private_reg(self):
         assert_equal(res.json['errors'][0]['detail'], 'Authentication credentials were not provided.')
         assert_equal(len(registration.alternative_citations), 0)
 
+
 class TestGetAlternativeCitations(ApiTestCase):
     def request(self, errors=False, is_admin=False, is_contrib=True, logged_out=False, **kwargs):
         admin = AuthUserFactory()
+        registration = kwargs.get('registration', None)
```
sloria: Better to add registration to the function signature rather than pull of `kwargs`. Otherwise, the `registration` parameter gets passed to `create_project`, which doesn't really make sense. That said, it doesn't appear to break anything, so \shrug/

brianjgeiger: I think it is supposed to be sending it to create_project explicitly. That was the original test, passing the kwargs along to the create project, I just needed the extra knowledge of if it was a registration to change the URL.

sloria: Ah ok; thanks for the clarification.

sloria:
Good to merge when the tests pass.

PR #4756, opened 2015-12-15
Two feature fixes

samchrisinger:
# Purpose/Changes
1) Using Node.url_for in the `archvie_success` task was raising an exception when outside of a request context
2) We should only clean the AuthorImport DOM node if it exists (otherwise a JS exception is raised)


sloria:

PR #4757, opened 2015-12-16
Admin App Forms: The Good Parts

samchrisinger:
# Purpose
There are some important changes in samchrisinger:feature/admin-app-forms that should be merged into the release branch. This PR aggregates those changes. 

# Changes
- let $osf.block apply to a specific DOM element
- add $osf.dialog
- change flags to snake case in Prereg Challenge schema
- don't delete DraftReg approval reference upon rejection
- let DraftReg flags get updated (but not overwritten) if the schema
- changes
- add setter for DraftReg flags
- add is_rejected property to DraftReg
- update DraftRegApproval to use a URL template rather than url_for

### DIFF
--- a/website/static/js/osfHelpers.js
+++ b/website/static/js/osfHelpers.js
@@ -218,23 +218,31 @@ var handleEditableError = function(response) {
     return 'Error: ' + response.responseJSON.message_long;
 };
 
-var block = function(message) {
-    $.blockUI({
-        css: {
-            border: 'none',
-            padding: '15px',
-            backgroundColor: '#000',
-            '-webkit-border-radius': '10px',
-            '-moz-border-radius': '10px',
-            opacity: 0.5,
-            color: '#fff'
-        },
-        message: message || 'Please wait'
-    });
+var block = function(message, $element) {
+    ($element ? $element.block : $.blockUI).call(
+        $element || window,
+        {
+            css: {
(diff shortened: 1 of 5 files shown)


website/static/js/osfHelpers.js
```diff
@@ -744,6 +752,50 @@ var any = function(listOfBools) {
...
+ * @param {String} actionButtonLabel:
+ * @param {Object} options: optional options
+ * @param {String} options.actionButtonClass: CSS class for action button, default 'btn-success'
+ * @param {String} options.cancelButtonLabel: label for cancel button, default 'Cancel'
+ * @param {String} options.cancelButtonClass: CSS class for cancel button, default 'btn-default'
+ *
+ * @example
+ * dialog('Hello', 'Just saying hello', 'Say hi').done(successCallback).fail(doNothing);
+ **/
+var dialog = function(title, message, actionButtonLabel, options) {
```
sloria: :+1:

sloria: This is nice, but is this used anywhere?

samchrisinger: It is in the admin app. If need be I can remove it, but it'll just get added later.

sloria: It's fine for now.

PR #4759, opened 2015-12-16
[RfR] add fields to V2 file serializer, improve non-osfstorage support

felliott:
Purpose
----
This PR fixes four minor closely-related tickets for changes to the V2 FileSerializer.  Two new attributes, `materialized_path` and `date_created`, were added.  Two existing attributes, `date_modified` and `extra.hashes`, were being pulled from the file versions, which non-osfstorage providers lack.  For these, we can use the metadata history to get this information.

*If desired I can split these up into separate PRs, but they're all pretty small, so I figured they'd be better together.*

Changes
----
* added `materialized_path` attribute to show unix-style path for entity relative to provider root.
* added `date_created` attribute. For osfstorage, `date_created` is the timestamp when the file was first uploaded to osfstorage.  For other providers, `date_created` is the `modified` field of the first entry in the file's metadata history.
* updated `date_modified` to work for non-osfstorage.  Returns the `modified` field of the most recent entry in the file's metadata history.
* updated `extras.hashes` to work with non-osfstorage providers, which have histories, rather than versions.  Only s3 currently provides md5.

Side Effects
----
Should be minimal, most changes are additions.  `date_modified` was updated to include a tz indicator, but datetime parsers should handle this, as we're just making explicit what was implicit before.

----------

**Fixes:** [#OSF-5417] - parent ticket, has testing information
**Sub-tasks:**
* [#OSF-5289] - add `date_created` to file serializer
* [#OSF-5346] - add `materialized_path` to file serializer
* [#OSF-5418] - update file serializer's `date_modified` to work for non-osfstorage
* [#OSF-5419] - update file serializer's `extra.hashes` to work for non-osfstorage

### DIFF
--- a/api/files/serializers.py
+++ b/api/files/serializers.py
@@ -140,13 +140,17 @@ class FileSerializer(JSONAPISerializer):
         return mod_dt and mod_dt.replace(tzinfo=pytz.utc)
 
     def get_extra(self, obj):
-        extras = {}
-        if obj.versions:
+        metadata = {}
+        if obj.provider == 'osfstorage' and obj.versions:
             metadata = obj.versions[-1].metadata
-            extras['hashes'] = {  # mimic waterbutler response
-                'md5': metadata.get('md5', None),
-                'sha256': metadata.get('sha256', None),
-            }
+        elif obj.provider != 'osfstorage' and obj.history:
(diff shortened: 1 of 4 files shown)


mfraezz:
Pass complete, no bugs found :shipit:

PR #4765, opened 2015-12-16
Add script to migrate metadata for Prereg submissions

samchrisinger:
# Purpose
Recent changes with the impl. of file uploaders in the Prereg registration schema require existing draft metadata to be supplemented. 

Additionally, some questions that did not previously have uploaders now have them, and consequently saved data for those questions needs to be updated.


sloria:
And `__init__`  file will need to be added to make the script runnable with

```
python -m scripts.prereg.migrate_metadata_for_uploaders
```

sloria:
Please add logging and use `scripts.utils.add_file_logger` to log to a file in the correct place.

sloria:
Ran on a current copy of prod with no errors.

PR #4767, opened 2015-12-16
Add ASCE Rocky Mountain

AndrewSallans:


sloria:
Ah, woops. The new conference was added to the wrong file. It should be added to `populate_conferences.py`, not `dev_populate_conferences.py`.

sloria:
I will fix this.

AndrewSallans:
Oops, sorry about that.  Not sure why I got the wrong one this time.  Thanks for fixing.

PR #4773, opened 2015-12-17
Handle renamed, moved, and deleted files in archive_success task

samchrisinger:
# Purpose
There is now an Archiver task to update references to files picked in the Prereg Challenge schema to the correct path on the registered Node. Unfortunately I assumed when the task encountered renamed, moved, or deleted files that the resulting link would simply be broken. Instead an exception was raised and the task failed. Compounding the unfortunateness this particular task did not subclass ArchiverTask (which contains built-in error handling), which in turn caused the failing task to fail silently. The resulting behavior is that file paths were not updated on registrations with renamed, moved, or deleted files.

# Changes
Raise a custom exception when the selected file cannot be found. This gets captured in the ArchiveTask error handler, and ultimately the initiating user gets an email informing them that we could not archive their project because the registration template referenced one or more files that have been renamed, moved, or deleted in between the time of their selection and archival.


sloria:
Can you please look into the failing test?

samchrisinger:
@sloria I feel like this PR may need community review.

sloria:
@erinmayhood Can you please take a look at this with Sam?

website/archiver/utils.py
```diff
@@ -25,6 +26,7 @@ def send_archiver_size_exceeded_mails(src, user, stat_result):
         mail=mails.ARCHIVE_SIZE_EXCEEDED_USER,
         user=user,
         src=src,
+        can_change_preferences=False,
```
samchrisinger: At some point extra text like "You can change your email preferences... " was added to the nofify_base email template. This text is not relevant to any archiver emails.

PR #4774, opened 2015-12-17
Feature/institutions auth

hmoco:


api/base/settings/defaults.py
```diff
@@ -167,6 +167,9 @@
 
 DEBUG_TRANSACTIONS = DEBUG
 
+JWT_SECRET = 'osf_api_cas_login_jwt_secret_32b'
```
sloria: Isn't there already a JWT_SECRET setting?

api/base/settings/defaults.py
```diff
@@ -167,6 +167,9 @@
...
+JWE_SECRET = 'osf_api_cas_login_jwe_secret_32b'
```
sloria: These settings should not be allowed to be set to the default value. You should add validation like we have in `website/settings/__init__.py` to ensure this are changed in `local.py`.

api/institutions/parsers.py
```diff
@@ -0,0 +1,29 @@
...
+import httplib
+
+import jwe
+import jwt
+from rest_framework.parsers import BaseParser
+
+from api.base import settings
+from framework.exceptions import HTTPError
+
+class InstitutionAuthParser(BaseParser):
```
sloria: I'm not sure that it makes sense to use a `Parser` to do this work. I think it makes sense to implement a drf `Authentication` class instead. See http://www.django-rest-framework.org/api-guide/authentication/

api/institutions/serializers.py
```diff
@@ -1,7 +1,38 @@
...
+from website import security
+from website.models import Institution, User
+from framework.auth import signals
+from framework.auth.core import get_user
 from api.base.serializers import JSONAPISerializer, RelationshipField, LinksField
 
+def find_institution_by_domain(username):
+    domain = username.split('@')[1]
+    try:
+        inst = Institution.find_one(Q('domains', 'eq', domain))
```
sloria: Will case matter? Do domains get stored in mixed-case?

sloria:
Please add a views tests for your new view.

api/institutions/serializers.py
```diff
@@ -1,7 +1,38 @@
...
+    except NoResultsFound:
+        raise NotFound
+    return inst
+
+def get_or_create_user(fullname, address):
```
sloria: Rather than copying this, better to move `get_or_create_user` and import it in two places. I think it'd make sense to move this function to `framework.auth.core`.

api/institutions/serializers.py
```diff
@@ -33,3 +64,27 @@ def get_absolute_url(self, obj):
 
     class Meta:
         type_ = 'institutions'
+
+
+class InstitutionAuthSerializer(ser.Serializer):
```
sloria: This serializer has no fields. Therefore, it doesn't really do any (de)serialization. You can probably remove this. Instead, you could move this logic to the new `Authentication` policy, as suggested in https://github.com/CenterForOpenScience/osf.io/pull/4774/files#r51460341

api/institutions/views.py
```diff
@@ -184,3 +185,15 @@ def get_default_odm_query(self):
...
         query = self.get_query_from_request()
         return User.find(query)
+
+class InstitutionAuth(JSONAPIBaseView, generics.CreateAPIView):
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        base_permissions.TokenHasScope,
+    )
+    required_read_scopes = [CoreScopes.NULL]
+    required_write_scopes = [CoreScopes.NULL]
```
sloria: @icereval Are these scopes correct?

framework/auth/core.py
```diff
@@ -1359,6 +1359,7 @@ def has_inst_auth(self, inst):
         return False
 
     affiliated_institutions = fields.ForeignField('institution', list=True)
+    institutions_metadata = fields.DictionaryField()
```
sloria: Please document the expected format of this field.

framework/auth/core.py
```diff
@@ -1385,7 +1386,10 @@ class Institution(StoredObject):
 
     _id = fields.StringField(index=True, unique=True, primary=True)
     name = fields.StringField(required=True)
+
     logo_name = fields.StringField(required=True)
+    domains = fields.StringField(list=True)
+    metadata_request_fields = fields.DictionaryField()
```
sloria: What is this field for?

api/institutions/parsers.py
```diff
@@ -0,0 +1,29 @@
...
+        value = stream.read()
+        try:
+            data = jwt.decode(
+                jwe.decrypt(value, settings.JWE_SECRET),
+                settings.JWT_SECRET,
+                options={'verify_exp': False},
+                algorithm='HS256'
+            )
+        except (jwt.InvalidTokenError, KeyError):
+            raise HTTPError(httplib.FORBIDDEN)
```
sloria: API code should not raise `framework.exceptions.HTTPError`; DRF has its own HTTP error classes you should raise.

api/base/settings/defaults.py
sloria: Oh, I see now that there is a JWT_SECRET in `website.settings` but not in `api.base.settings`. @icereval What do you think? Should we add this here?

sloria:
Pass finished. The main refactor I'm suggesting is to move the work of decrypting the JWT payload and get-or-creating a user to an `Authentication` class.

icereval:
@hmoco, if you have thoughts please note them but don't worry about making the updates yet. I'll be updating the code in this PR w/ the integration testing occurring so I'll try to take care of what I can.

framework/auth/core.py
hmoco: This field is to store where in the request body are we expecting the metadata to be stored, since that will vary depending on the institution. It might not end up being a dictionary field.

PR #4775, opened 2015-12-17
Feature/webinar workshop statistics

GageGaskins:
# Purpose
Script to find user activity after webinars and workshops.

Note, input csv file headers must be Workshop_Date, Name, and Email Address or script will fail to find the columns.


GageGaskins:
@icereval Script is ready for review or use

PR #4776, opened 2015-12-17
Feature/card ui fix

ticklemepierce:
Purpose
-
Fixes [OSF-5416][1].  Also, it looked odd to have a max width of 75% for non-expandable card titles so this was fixed as well

Changes
-
Removed `word-break: break-all` from the card headers.  Removed `class="header-content"` from unexpandable cards.

Side Effects
-
Its possible that this could cause slight UI issues but I highly doubt it and have not seen any.

[1]: https://openscience.atlassian.net/browse/OSF-5416


sloria:
Discussed with @ticklemepierce -- a new PR will be created with just the relevant changes.

PR #4780, opened 2015-12-17
[Feature] Move wiki enable to settings

billyhunt:
Purpose
-------
Remove the wiki enable/disable checkbox from the add-ons section and add it to the Wiki Config section.

Changes
------------

1.  Delete Wiki as an checkbox in the addons section of the project settings page.
2.  Enable Configure Wiki all the time in both the sidebar and settings page, not only when you can change permissions.
3.  Add enable wiki checkbox in the Configure Wiki module.
4.  Change explanatory language for both enabling the wiki and configuring it.
5.  Check to see if node is public or private, and change language for configuration accordingly.  Basically, if the project is already public, do not tell the user they need to make the project public to allow all users to contribute.  
6.  To not say Configure 400 times on the page, I removed it from all titles except Configure Add-Ons, which is necessary to differentiate it from Select Add-Ons.  

Before
---------

screen shot 2015-12-17 at 1 12 23 pm

screen shot 2015-12-17 at 1 09 33 pm

screen shot 2015-12-17 at 1 56 19 pm

After
------

screen shot 2015-12-17 at 1 53 02 pm

screen shot 2015-12-17 at 1 11 59 pm

screen shot 2015-12-17 at 1 57 04 pm


website/static/js/pages/project-settings-page.js
```diff
@@ -252,5 +252,57 @@ $(document).ready(function() {
             }
         }
     });
+    $('.wiki-select').on('change', function() {
+        var that = this,
```
sloria: > Use self to save a reference to this.

https://cosdev.readthedocs.org/en/latest/style_guides/javascript.html#style

> Declare one variable per var statement, it makes it easier to re-order the lines.

https://github.com/felixge/node-style-guide#declare-one-variable-per-var-statement via COSDev

website/templates/project/settings.mako
```diff
@@ -33,15 +33,13 @@
                             <li><a href="#configureAddonsAnchor">Configure Add-ons</a></li>
                         % endif
 
-                        % if include_wiki_settings:
-                            <li><a href="#configureWikiAnchor">Configure Wiki</a></li>
-                        % endif
+                        <li><a href="#configureWikiAnchor">Wiki</a></li>
 
                         % if 'admin' in user['permissions']:
-                            <li><a href="#configureCommentingAnchor">Configure Commenting</a></li>
```
sloria: While I think this is a good change, the documentation on the Getting Started page will need to be updated to stay consistent with these headers.

website/templates/project/settings.mako
```diff
@@ -203,35 +200,62 @@
...
+                                    <div class="logo-spin logo-lg"></div>
+                                    <p class="m-t-sm fg-load-message"> Loading wiki settings...  </p>
+                                </div>
+                            </div>
+                            <div class="help-block" style="padding-left: 15px">
+                                <p id="configureWikiMessage"></p>
+                            </div>
+                        </form>
+                    % endif
+                    % if not include_wiki_settings:
```
sloria: Can this just be an `else` block?

website/templates/project/settings.mako
```diff
@@ -132,15 +130,14 @@
...
 
                             % for category in addon_categories:
-
                                 <%
-                                    addons = [
-                                        addon
-                                        for addon in addons_available
-                                        if category in addon.categories
-                                    ]
+                                    addons = []
```
sloria: If we are treating wiki as a core OSF feature, then this change should be happening on the backend, i.e. at the view layer around here: https://github.com/CenterForOpenScience/osf.io/blob/91078b2875b38f5f3f6df0698540e5084fd06be2/website/project/views/node.py#L281 , rather than doing all this logic here.

sloria:
Pass finished.

website/templates/project/settings.mako
billyhunt: Would treating the wiki as core vs addon on the backend effect the rest of the codebase?  Are you suggesting a refactor everywhere addons_available is used?

sloria: Reusing Python code, i.e. using view helper functions, will be *much* simpler than trying to reuse mako code.

billyhunt: I removed the wiki code from the mako template and put it in the python nodes view, and left the old python addons code in the mako template.  Let me know if this works for you.

sloria:
Can we remove the modal for enabling/disabling the wiki? As osf-style suggests, modals should only be used for critical, irreversible actions. Enabling/disabling the wiki is easily reversible. It would be more user-friendly to automatically save the user's preference.

If the modal wasn't part of the spec for this feature, I suggest removing it.

website/static/js/pages/project-settings-page.js
```diff
@@ -252,5 +252,57 @@ $(document).ready(function() {
             }
         }
     });
+    $('.wiki-select').on('change', function() {
+        var self = this;
+        var $self = $(self);
+        var wikiEnabled = $self.is(':checked');
```
sloria: Could you simplify this and prevent storing state in the DOM by using a simple ViewModel? You could even use a POJO:

```js
var WikiSettingsViewModel = {
     enabled: ko.observable(ctx.wiki.isEnabled), // <- this would get set in the mako template, as usual
     onChange: function(e) { 
         // send request then reload page
      }
}
```

```html
...
<input type="checkbox" data-bind="event: onChange, checked: enabled" />
```

website/project/views/node.py
```diff
@@ -302,9 +302,13 @@ def node_setting(auth, node, **kwargs):
         addon
         for addon in settings.ADDONS_AVAILABLE
         if 'node' in addon.owners
-        and addon.short_name not in settings.SYSTEM_ADDED_ADDONS['node']
+        and addon.short_name not in settings.SYSTEM_ADDED_ADDONS['node'] and addon.short_name != 'wiki'
     ], key=lambda addon: addon.full_name.lower())
 
+    for addon in settings.ADDONS_AVAILABLE:
+        if 'node' in addon.owners and addon.short_name not in settings.SYSTEM_ADDED_ADDONS['node'] and addon.short_name == 'wiki':
+            ret['wiki'] = addon
```
sloria: Once you get here, you should be able to `break` out of this loop.

sloria:
Pass finished.

website/templates/project/settings.mako
billyhunt: Done.  The only change I found that was needed was for Configure Comments being changed to Comments.

website/templates/project/settings.mako
```diff
@@ -387,6 +411,8 @@
       window.contextVars.node.description = ${node['description'] | sjson, n };
       window.contextVars.node.nodeType = ${ node['node_type'] | sjson, n };
       window.contextVars.nodeCategories = ${ categories | sjson, n };
+      window.contextVars.wiki = window.contextVars.wiki || {};
```
sloria: Why is this line necessary?

billyhunt: It was complaining when I did not create the wiki object first.  I created a wiki object first thinking there might be other wiki variables that are needed when I do the big overhaul, but I could make a simple boolean like window.contextVars.wikiEnabled if you want me to keep it simple.

sloria: Or shouldn't it be 

```
window.contextVars.wiki = ${ wiki | sjson, n };
```

or am I missing something?

website/templates/project/settings.mako
```diff
@@ -203,35 +201,64 @@
...
+            <div class="panel-body">
+                %if wiki:
+                    <form id="selectWikiForm">
+                        <div>
+                            <label>
+                                <input
+                                        type="checkbox"
+                                        name="${wiki.short_name}"
+                                        class="wiki-select"
+                                        data-bind="checked: enabled, click: onChange"
```
sloria: Using `click` is OK, but I think using a subscriber function like in https://jsfiddle.net/rniemeyer/cnkVA/2/ might be better for a11y. There may be alternative ways to toggle the checkbox, such as tabbing and hitting Enter.

That said, this isn't a blocker. Let's go with the "5-minute rule"--if the change takes <= 5 mins to fix, go ahead and make it. Otherwise, defer.

sloria:
Pass finished. Very close to merging!

website/templates/project/settings.mako
billyhunt: If I add that, I get the following when I load the page.

Error !

TypeError: <website.addons.base.AddonConfig object at 0x1141a83d0> is not JSON serializable

sloria: Oh ok; so where is `window.contextVars.wiki` assigned?

billyhunt: There.  I just need the enabled or disabled information.  I created the object to mimic the pseudocode you created and also thinking the object could come in handy later if and when I want to add additional attributes when you can change all the wiki permissions of the children.

website/static/js/pages/project-settings-page.js
```diff
@@ -252,5 +252,37 @@ $(document).ready(function() {
...
     });
-
 });
+
+var WikiSettingsViewModel = {
+    enabled: ko.observable(ctx.wiki.isEnabled), // <- this would get set in the mako template, as usual
+    wikiMessage: ko.observable('')
+};
+
+var self = this;
```
sloria: No need to attach to `this` here; it does not serve a purpose. You should just be able to do

```js
WikiSettingsViewModel.enabled.subscribe(...)
```

website/templates/project/settings.mako
sloria: Oh, OK, I missed the line under this one. I see now. Carry on

PR #4782, opened 2015-12-17
Add academia edu field

kch8qx:


website/static/js/profile.js
```diff
@@ -536,7 +537,15 @@ var SocialViewModel = function(urls, modes) {
         ko.observable().extend({trimmed: true, cleanup: cleanByRule(socialRules.github)}),
         self, 'github', 'https://github.com/'
     );
-
+    self.academia1 = extendLink(
+        ko.observable().extend({trimmed: true, cleanup: cleanByRule(socialRules.academia)}),
+        self, 'academia1', 'https://'
+    );
+    self.academia2 = extendLink(
```
mfraezz: It would be better to give these two more descriptive names, such as `academiaInstitution` and `academiaProfileID`

mfraezz:
Looking good, one small change regarding variable naming and it should be ready to merge. :boat:

sloria:
It would be great if the `institution` input were smaller in width, since it will almost always be shorter than the profileID.

osf___settings

But it's not a blocker. @kch8qx , perhaps you could look at this when you get back and send a follow-up PR.

sloria:
Good to merge. I'll take care of conflicts.

kch8qx:
@sloria I have to go back and work on the academia.edu field anyway to make it show up when you search a user, so I can try and fix the institution field length at the same time.

PR #4788, opened 2015-12-18
Give the v2 API it's own _celery_tasks

samchrisinger:
The v2 API couldn't use Flask's thread locals, and consenquently calls
to framework.tasks.enqueue_task made in API threads were likely running
syncrhonously. This add a _celery_tasks to the API's thread locals and
then dynamically import the configured THREAD_LOCALS module in
framework.tasks.handlers.

### DIFF
--- a/framework/tasks/handlers.py
+++ b/framework/tasks/handlers.py
@@ -3,24 +3,34 @@
 import logging
 import functools
 
-from flask import g
 from celery import group
 
 from website import settings
 
-
 logger = logging.getLogger(__name__)
+
+def dynamic_import(module):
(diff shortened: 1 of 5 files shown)


framework/tasks/handlers.py
```diff
@@ -3,24 +3,34 @@
...
 
-from flask import g
 from celery import group
 
 from website import settings
 
-
 logger = logging.getLogger(__name__)
 
+def dynamic_import(module):
```
sloria: You can probably use `werkzeug.utils.import_string` instead of reimplementing.

PR #4790, opened 2015-12-18
[Ready to Review] Log API endpoint contributor_added is not returning required parameters

pattisdr:
# Purpose

Issue https://openscience.atlassian.net/browse/OSF-5354

The API for logs has a log action called "contributor_added". This log action should be returning the names of the contributors added so the logs on the front end can be generated.

# Changes

Adds a `added_contributors` relationship link to node logs.  If the log action was not 'contributor_added', this link will return an empty list.  Otherwise, it will return a list of serialized user representations associated with that log.  To get the names of the associated users in one request, the added_contributors can be embedded.

screen shot 2015-12-18 at 10 18 01 am


samchrisinger:
This feels like a pretty heavyweight fix for such a small problem. I'm wondering if we had something like:

```python
class NodeLogContributorsField(ser.Field):
    def to_internal_value(self, users):
        return [u._id for u in users]

   def to_representation(self, user_ids):
         return [User.load(uid) for uid in user_ids]
```

pattisdr:
`to_representation` needs to be called with a UserSerializer instance and this is a User instance.  I'm not sure how to turn this into a list of serialized users without making a view.  Will keep thinking.

samchrisinger:
Can you instantiate it like `UserSerializer`

whit537:
Popped this one from the Jira [review queue](https://openscience.atlassian.net/issues/?filter=11402), taking a look ...

whit537:
What's a "log" here? We show logs somewhere in the app, it sounds like?

whit537:
Grepping around for `logs`, `website/static/js/logFeed.js` looks promising ...

whit537:
Also `project.js` ...

whit537:
`dashboard-page.js` instantiates a `LogFeed` ...

pattisdr:
node.logs

whit537:
@pattisdr Where does that show up in the UI? Does it yet?

whit537:
A [clue](https://github.com/CenterForOpenScience/osf.io/blob/0640d6d95e583de753a20f0157ae521c5605c8c5/website/static/js/project.js#L213)! :-)

> Display recent logs for for a node on the project view page.

pattisdr:
Logs show up in UI under 'Recent Activity' for a node.

whit537:
Yeah, seeing it now ...

screen shot 2016-01-22 at 2 05 28 pm

whit537:
Alright, so it sounds like the point of this PR (~= [OSF-5354](https://openscience.atlassian.net/browse/OSF-5354)) is to pave the way for adding additional event types to the activity feed (~= `node.logs`). Yes?

screen shot 2016-01-22 at 2 06 45 pm

pattisdr:
Also, logs pertaining to a particular node can be accessed via the API with `v2/nodes/{node_id}/logs/`.  Or detail about a specific log: `/v2/logs/{log_id}/`

pattisdr:
There are already numerous event types - `contributor_added` is already an event type.  However, I think a specific UI case wanted more than just the action "contributor_added", it needed the contributor's name.

whit537:
> There are already numerous event types - `contributor_added` is already an event type. However, I think a specific UI case wanted more than just the action "contributor_added", it needed the contributor's name.

Gotcha. Okay! So ...

whit537:
Sorry, had to find better wifi, back now ...

whit537:
I'm skimming [`RelationshipField`](https://github.com/pattisdr/osf.io/blob/51b0b02dbcf0d07bcd8d387468b63f8f504c7321/api/base/serializers.py#L244) to start wrapping my head around our DSF-based serialization framework ...

whit537:
> DSF

Sorry, D*R*F.

whit537:
I'm going through the [DRF tutorial](http://www.django-rest-framework.org/tutorial/1-serialization/).

whit537:
> Dealing with relationships between entities is one of the more challenging aspects of Web API design. There are a number of different ways that we might choose to represent a relationship:
>
[...]
>
REST framework supports all of these styles, and can apply them across forward or reverse relationships, or apply them across custom managers such as generic foreign keys.

http://www.django-rest-framework.org/tutorial/5-relationships-and-hyperlinked-apis/#hyperlinking-our-api

Okay, so it looks like that is what the `HyperlinkedIdentityField` base class for `RelationshipField` is about.

whit537:
https://github.com/tomchristie/rest-framework-tutorial/pull/26 :cat:

whit537:
I'm wondering whether a one-off `added_contributors` attribute fits with our API architecture, or whether there isn't a generic `payload`-type attribute that should carry the contributors list for the case where `action` is `contributors_added` . Maybe [`params`](https://github.com/pattisdr/osf.io/blob/51b0b02dbcf0d07bcd8d387468b63f8f504c7321/api/logs/serializers.py#L60)? Investigating ...

whit537:
I don't understand the syntax of the `related_view` kwarg to `RelationshipField`.

pattisdr:
It is similar to the HyperlinkedIdentityField, the view_name.  Except we are adhering to the jsonapi spec, `http://jsonapi.org/format/`, which allows you to have self and related links.  So the related_view kwarg allows you to create a related link versus a self link.

whit537:
> Except we are adhering to the jsonapi spec, `http://jsonapi.org/format/`, which allows you to have self and related links.

Ah, right. That rings a bell, now that you mention it. :)

> So the related_view kwarg allows you to create a related link versus a self link.

Okay, cool. So `related_view` has the same syntax as `view_name`, which I guess is a DRF thing.

http://www.django-rest-framework.org/api-guide/relations/#hyperlinkedidentityfield

whit537:
> If you're using [the standard router classes](http://www.django-rest-framework.org/api-guide/relations/#hyperlinkedidentityfield) this will be a string with the format [<model_name>-detail`

whit537:
I don't see any colons in http://www.django-rest-framework.org/api-guide/routers/#defaultrouter.

pattisdr:
'view_category: view_name'

whit537:
> 'view_category: view_name'

Sorry ... what does that mean? I don't find that string on http://www.django-rest-framework.org/api-guide/routers/#defaultrouter, for example. A grep of our codebase for `view_category` is also inconclusive.

whit537:
It [looks like](https://github.com/tomchristie/django-rest-framework/blob/3.1.1/rest_framework/relations.py#L204) `view_name` may eventually make it's way to `resolve`, either [DRF's version or Django's version](https://github.com/tomchristie/django-rest-framework/blob/3.1.1/rest_framework/reverse.py).

pattisdr:
from api/nodes/serializers:
```
 children = RelationshipField(
        related_view='nodes:node-children',
        related_view_kwargs={'node_id': '<pk>'},
        related_meta={'count': 'get_node_count'},
    )
```
related view is `nodes:node-children`.  

Now going to api/nodes/views, the node children list view has 
```
 view_category = 'nodes'
 view_name = 'node-children'
```
This related view will `resolve` to /v2/nodes/<node_id>/children.  The view_category and view_name are defined on the view.

whit537:
> either [DRF's version or Django's version](https://github.com/tomchristie/django-rest-framework/blob/3.1.1/rest_framework/reverse.py)

Sorry, I had searched for `reverse` instead of `resolve`. Afaict, DRF doesn't have a `resolve`; it just uses [Django's](https://github.com/django/django/blob/1.8/django/core/urlresolvers.py#L519-L522).

whit537:
Waaaaiiiiitttt, but I *meant* `reverse` at https://github.com/CenterForOpenScience/osf.io/pull/4790#issuecomment-174036532, because I was looking at [this code](https://github.com/tomchristie/django-rest-framework/blob/3.1.1/rest_framework/relations.py#L204).

whit537:
> 'view_category: view_name'
> Now going to api/nodes/views, the node children list view has

Okay, so the syntax for `related_view` ~= `view_name` is `view_category:view_name` ... which instructs DRF (not Django?) to load a view class by matching against the `view_category` and `view_name` attributes of ... whatever view classes are ... registered with the framework? More or less?

whit537:
Is that a Django thing? A DRF thing? Where is the syntax of `related_view`/`view_name` documented?

pattisdr:
The `related_view = view_category`: view_name is how we tell which view we want to reverse for the related url.

whit537:
Okay, but I don't see any indication on, e.g., [the HyperlinkedIdentityField docs](http://www.django-rest-framework.org/api-guide/relations/#hyperlinkedidentityfield) (per https://github.com/CenterForOpenScience/osf.io/pull/4790#issuecomment-174033406) that the syntax of `view_name` is `view_category:view_name`. Oh well. Now I know. :-)

Okay! Onward!

pattisdr:
I don't know where this originated - @brianjgeiger might know more - why do we do view_category: view_name to refer to the view?  For example, in the `RelationshipsField`  `related_view='nodes:node-children`', and earlier, in the LinksField, `'related': Link('nodes:node-children', kwargs={'node_id': '<pk>'})`

whit537:
@pattisdr BTW, point of order: you don't have to respond to my comments if you don't want to at this point. I'm sure you've got other work to do, and I'm happy to do my own research. :-)

After I'm done digesting the code, I'll make up a punchlist of items for you to specifically address (either with explanations or commits). See https://github.com/CenterForOpenScience/osf.io/pull/4799#issuecomment-170572129 and https://github.com/CenterForOpenScience/osf.io/pull/4509#issuecomment-168799268 for examples. Until then you can safely assume I'm just thinking out loud. :-)

Does that work for you?

pattisdr:
great, thanks

whit537:
@pattisdr++

whit537:
Alright, I am seeing the `added_contributors` embed as intended (https://github.com/CenterForOpenScience/osf.io/pull/4790#issue-122965156).

screen shot 2016-01-25 at 9 51 29 am

whit537:
And if I remove the [`added_contributors`](https://github.com/CenterForOpenScience/osf.io/blob/51b0b02dbcf0d07bcd8d387468b63f8f504c7321/api/logs/serializers.py#L75-L78) field, I get an error:

screen shot 2016-01-25 at 9 58 05 am

whit537:
Alright, so we've got a concept of "embeddable fields." What's that about?

whit537:
What's the purpose of embeddable fields? Is that a DRF concept or an OSF concept?

whit537:
Googling "django rest framework embed" doesn't turn up much of anything. On the other hand, grepping for "embed" yields https://api.osf.io/v2/#embedding.

pattisdr:
Embed is an OSF thing - similar to jsonapi includes, but 'includes' had some drawbacks, so we implemented embeds instead.

whit537:
Interesting. So linked entities are in *either* `relationships` *or* `embeds`:

screen shot 2016-01-25 at 10 15 25 am
screen shot 2016-01-25 at 10 15 40 am

whit537:
Interesting that the [`linked_node`](https://github.com/CenterForOpenScience/osf.io/blob/51b0b02dbcf0d07bcd8d387468b63f8f504c7321/api/logs/serializers.py#L81-L84) and [`template_node`](https://github.com/CenterForOpenScience/osf.io/blob/51b0b02dbcf0d07bcd8d387468b63f8f504c7321/api/logs/serializers.py#L85-L88) `RelationshipField`s don't show up in `relationships`/`embeds`.

pattisdr:
Relationships only show up if they exist.  If there is no linked_node or template_node for a log, we're not going to provide a link to it.

whit537:
Alright, so how do serializations and views relate? Investigating ...

whit537:
I see the standalone `added_contributors` endpoint:

screen shot 2016-01-25 at 10 40 40 am

whit537:
Looks like the view [specifies](https://github.com/CenterForOpenScience/osf.io/blob/51b0b02dbcf0d07bcd8d387468b63f8f504c7321/api/logs/views.py#L242) a serializer class. Presumably there is some DRF (or OSF?) wiring to make use of the class.

api/logs/views.py
```diff
@@ -9,10 +11,25 @@
 from api.base.utils import get_user_auth, get_object_or_error
 from api.base import permissions as base_permissions
 from api.nodes.serializers import NodeSerializer
+from api.users.serializers import UserSerializer
 from api.logs.serializers import NodeLogSerializer
 from api.base.views import JSONAPIBaseView
 
 
+class LogMixin(object):
```
whit537: Good refactor.

api/logs/views.py
```diff
@@ -171,3 +183,72 @@ def get_object(self):
...
+
+    serializer_class = UserSerializer
+
+    view_category = 'logs'
+    view_name = 'log-added_contributors'
+
+    # overrides ListAPIView
+    def get_queryset(self):
+        log = self.get_log()
+        added_contrib_ids = log.params.get('contributors')
```
whit537: Does `log.params.get('contributors')` return all contributors currently on the node, or only those contributors added during the event logged?

whit537: I suspect @caneruguz needs the delta, not the current set of contributors.

@caneruguz Can you say more about the UI requirement on #4790 ([OSF-5354](https://openscience.atlassian.net/browse/OSF-5354))?

whit537: My first guess would be that `log.params` is an [instance](https://github.com/CenterForOpenScience/osf.io/blob/51b0b02dbcf0d07bcd8d387468b63f8f504c7321/api/logs/serializers.py#L60) of `NodeLogParamsSerializer`, but I don't find a `contributors` attribute [there](https://github.com/CenterForOpenScience/osf.io/blob/51b0b02dbcf0d07bcd8d387468b63f8f504c7321/api/logs/serializers.py#L26-L50). Hmm ...

whit537: Looks like `log` is actually a `website.project.model.NodeLog`, and `log.params` is a `dict`.

```
(Pdb) l
246     
247         # overrides ListAPIView
248         def get_queryset(self):
249             log = self.get_log()
250             import pdb; pdb.set_trace()
251  ->         added_contrib_ids = log.params.get('contributors')
252             if added_contrib_ids is None:
253                 return []
254             added_users = User.find(Q('_id', 'in', added_contrib_ids))
255             return added_users
[EOF]
(Pdb) log
<NodeLog(u'contributor_added', params={u'project': u'ba9zp', u'node': u'498yc', u'contributors': [u'nfqxe']}) with id u'561fbd6d029bdb16f3f32dac'>
(Pdb) log.__class__
<class 'website.project.model.NodeLog'>
(Pdb) log.params
{u'project': u'ba9zp', u'node': u'498yc', u'contributors': [u'nfqxe']}
(Pdb) log.params.__class__
<type 'dict'>
(Pdb)
```

whit537: Here's what `NodeLog` looks like in the db:

screen shot 2016-01-25 at 3 19 53 pm

whit537: Looks like there's a [couple](https://github.com/CenterForOpenScience/osf.io/blob/51b0b02dbcf0d07bcd8d387468b63f8f504c7321/website/project/model.py#L2808-L2817) [places](https://github.com/CenterForOpenScience/osf.io/blob/51b0b02dbcf0d07bcd8d387468b63f8f504c7321/website/project/model.py#L2854-L2866) where we log `contributor_added`.

whit537: The second is the plural case of the former: we only want one log entry whether adding a single contributor or multiple contributors.

whit537: In both cases, it looks like the node log carries only the contributors *added*—which is good.

whit537:
Alright! This is lookin' pretty good so far. Three more questions I'd like to answer:

 - [ ] What was @samchrisinger on about at https://github.com/CenterForOpenScience/osf.io/pull/4790#issuecomment-166973417 (what exactly was he suggesting?)?
 - [ ] How does solving this with an `added_contributor` embed compare with similar *current* cases?
 - [ ] What are the implications of the `added_contributor` solution for similar *future* cases?

whit537:
I *think* what @samchrisinger is suggesting is that this be implemented with an always-on `contributors` attribute on the `NodeLog` JSON object, vs. an `added_contributors` `relationship` that one must explicitly request as an `embed`. It seems like this could make more sense, because with the `embed` solution, the client must either:

 1. Always request `added_contributors` as an `embed`, regardless of the log action.
 1. Make multiple HTTP requests: an initial one to sniff the `contributor_added` log action, and a second to request the `added_contributors` `embed`.

The second option would be suboptimal from a performance point of view. But, insofar as multiple payload types are possible (likely? already actual?), the former option seems suboptimal from an API design point of view: you wouldn't want to have to enumerate every possible `embed` for every node log. Right?

whit537:
Here's the [list](https://github.com/CenterForOpenScience/osf.io/blob/51b0b02dbcf0d07bcd8d387468b63f8f504c7321/website/project/model.py#L425) of log actions; I count 54.

caneruguz:
@whit537 
In response to this: https://github.com/CenterForOpenScience/osf.io/pull/4790/files#r50714009

User log is saying that a user added contributors X, Y, Z; which are newly added contributor names. The names should ideally appear in the log (instead of "User added some contributors" only). Does this answer the question?

whit537:
> Does this answer the question?

@caneruguz Yes, thanks.

> User log is saying that a user added contributors X, Y, Z; which are newly added contributor names. The names should ideally appear in the log (instead of "User added some contributors" only).

I believe this PR delivers on that particular need. However, it doesn't seem right to me to address this in an ad-hoc fashion. Looks like `params` already exists [on `nodelogs`](https://github.com/CenterForOpenScience/osf.io/pull/4790#discussion_r50749871) in the db as a container for arbitrary data that varies with log `action`—that's where this PR [pulls](https://github.com/CenterForOpenScience/osf.io/blob/51b0b02dbcf0d07bcd8d387468b63f8f504c7321/api/logs/views.py#L250) `added_contributors` from!—*and* it looks like we've already got [plumbing](https://github.com/CenterForOpenScience/osf.io/blob/51b0b02dbcf0d07bcd8d387468b63f8f504c7321/api/logs/serializers.py#L60) to expose `params` via JSON. 

@pattisdr Is there a reason to prefer the one-off `embed` approach over adding `contributors` to `NodeLogSerializer.params`?

pattisdr:
If a contributor was added, in the NodeLog for that action, a contributor_id would be placed into a list in the `params` dictionary: 
screen shot 2016-01-25 at 5 57 13 pm

There are just user_ids in this contributors list.  They way I understand it, If I want to return information about the user, I have to return a User object - I have to use the UserSerializer - I shouldn't just return a dictionary with their names listed.  I don't know how to return a list of serialized users without making a view.  This is what I was telling Sam above. Because I need to return a list of serialized users, it made sense to me to make this is a relationship that could be embedded.

samchrisinger:
@pattisdr as proof of concept I tried:
```python
class NodeLogContributorsField(ser.Field):
    def to_internal_value(self, users):
        return [u._id for u in users]

    def to_representation(self, user_ids):
        from website.models import User
        from api.users.serializers import UserSerializer  # not sure if needed, being lazy
        return [
            UserSerializer(context=self.context).to_representation(
                User.load(uid)
            )
            for uid in user_ids
        ]
```
and:

```python
class NodeLogParamsSerializer(RestrictedDictSerializer):
    ...
    contributors = NodeLogContributorsField(read_only=True)
```

And found the contributors appear to be getting serialized correctly. Since this is read-only, a SerializerMethodField may work as well.

Please don't let this be a blocker for this PR, just wanted to toss the idea out there.

whit537:
> There are just user_ids in this contributors list.

Okay, fair enough: `params` contains ids, and the API should return hydrated objects in some fashion. Therefore, it seems that `params` is not the place for this.

> Because I need to return a list of serialized users, it made sense to me to make this a relationship that could be embedded.

@pattisdr Do you anticipate extending this pattern to other log `action`s in the future?

In general, I don't feel that I have enough context to know how logs are currently used or what the general direction is in terms of surfacing them in the UI. This PR ably solves the immediate need on [OSF-5354](https://openscience.atlassian.net/browse/OSF-5354), so, like @samchrisinger, I'm disinclined to block it or ask that it be (more or less) completely rewritten. I think we should land this to unblock @caneruguz, and reticket a refactor to generalize the pattern of including `action`-specific data in log nodes (something I don't think the `log.contributors` proposal adequately addresses, either, btw).

:+1: from me—with the one last-minute caveat that this will create a backwards-compatibility burden if/when we decide to change how we implement this pattern to better-accommodate the general case.

Over to you, @sloria? :-)

whit537:
> with the one last-minute caveat that this will create a backwards-compatibility burden 

@pattisdr reports at https://github.com/caneruguz/osf.io/pull/17#discussion_r51287541 that we're not caring about backwards-compatibility in the API.

sloria:
Agreed with @whit537 --we'll want to revisit to see if we can generalize how log params are represented. There are many potential business objects that may be represented within a log. But in the interest of unblocking Caner, we can merge this as is.

whit537:
> we'll want to revisit to see if we can generalize how log params are represented. There are many potential business objects that may be represented within a log.

Reticketed as [OSF-5667](https://openscience.atlassian.net/browse/OSF-5667).

PR #4791, opened 2015-12-18
Use a common thread local queue for enqueued celery tasks

samchrisinger:


brianjgeiger:
What ever happened with this? It seemed like it was very important to what you were doing.

samchrisinger:
@brianjgeiger it is merged/deployed in feature/admin-app-prod. But yeah I totally agree that this should be included in the next release.

sloria:
This code was added in https://github.com/CenterForOpenScience/osf.io/commit/9c2729cf6d48667b3585cf05b4d7b332baf648e9

PR #4792, opened 2015-12-18
WIP Log analytics for Recent Activity widget 

hmoco:


scripts/create_fakes.py
```diff
@@ -310,4 +310,6 @@ def main():
 
 if __name__ == '__main__':
     init_app(set_backends=True, routes=False)
+    for i in range(0,35):
```
sloria: This should be added to the `main()` function and parametrized rather than hardcoding here.

hmoco: Sorry this shouldnt be here

scripts/analytics/logs_analytics.py
```diff
@@ -0,0 +1,103 @@
...
+def recent_time_frame(days=NUMBER_OF_DAY_DATA_POINTS):
+    now = dt.utcnow()
+    return [now - timedelta(days=i) for i in range(0, days)]
+
+def build_time_query(end):
+    return Q('date', 'gt', end - timedelta(days=1)) & Q('date', 'lt', end)
+
+def order_users_get(sample_size=NUMBER_OF_USERS_TO_SAMPLE):
+    us = User.find()
+    points = {u.get_activity_points: u for u in us}
```
sloria: Does this work? It seems like `u.get_activity_points` should be called.

scripts/analytics/logs_analytics.py
```diff
@@ -0,0 +1,120 @@
...
+    print('Found users.')
+    l = len(ordered)
+    return ordered[int(l/50): int(l/50) + sample_size], ordered[int(l/40): int(l/40) + sample_size], ordered[int(l/30): int(l/30) + sample_size], ordered[int(l/20): int(l/20) + sample_size]
+
+def get_agg_for_user(user, date):
+    ret = {}
+    nodes = list(Node.find(Q('contributors', 'contains', user._id)))
+    for node in nodes:
+        query_aggs = {
+            'comments': len(list(NodeLog.find(Q('was_connected_to', 'eq', node) & Q('action', 'eq', NodeLog.COMMENT_ADDED) & build_time_query(date)))),
```
sloria: What are you trying to get here? Are you sure `was_connected_to` is correct? IIRC, this field was added before a log migration just to maintain provenance--it doesn't actually get used on the site.

sloria: Also, you shouldn't need to cast to a list; you should be able to use `.count()`. This will prevent loading up the entire result sets in memory.

scripts/analytics/logs_analytics.py
```diff
@@ -0,0 +1,120 @@
...
+        if how_many < 1:
+            users.remove(user)
+    ordered = sorted(users, key=lambda x: x.get_activity_points(), reverse=True)
+    print('Found users.')
+    l = len(ordered)
+    return ordered[int(l/50): int(l/50) + sample_size], ordered[int(l/40): int(l/40) + sample_size], ordered[int(l/30): int(l/30) + sample_size], ordered[int(l/20): int(l/20) + sample_size]
+
+def get_agg_for_user(user, date):
+    ret = {}
+    nodes = list(Node.find(Q('contributors', 'contains', user._id)))
```
sloria: Same here; no need to cast to a list.

PR #4793, opened 2015-12-18
Prereg validation feedback

samchrisinger:
# Purpose
Give feedback to users as to why a draft cannot be submitted:
https://openscience.atlassian.net/browse/OSF-5336

When a user visits the last page of the schema, and has validation errors: 
screen shot 2015-12-09 at 4 36 43 pm

And sees question specific validation errors:
screen shot 2015-12-09 at 4 38 56 pm

# Changes
- Update templates to bind validation messages
- Update RegistrationEditor and related models to aggregate validation info


website/static/js/osfHelpers.js
```diff
@@ -741,10 +741,15 @@ function indexOf(array, searchFn) {
  * @param {Array[Any]} listOfBools
  * @returns {Boolean}
  **/
-var any = function(listOfBools) {
+var any = function(listOfBools, check) {
```
sloria: Couldn't you achieve the same with

```js
stuff.filter(check).length
```

PR #4798, opened 2015-12-18
[Ready] Make migration resumable

pattisdr:
# Purpose

Makes /migration/migrate_date_modified_for_existing_nodes.py resumable.  

# Changes

From @slora, adds `Q('date_modified', 'eq', None)` to node query

### DIFF
--- a/scripts/migration/migrate_date_modified_for_existing_nodes.py
+++ b/scripts/migration/migrate_date_modified_for_existing_nodes.py
@@ -3,6 +3,7 @@ This will add a date_modified field to all nodes.  Date_modified will be equival
 """
 import sys
 import logging
+from modularodm import Q
 from website.app import init_app
 from website import models
 from scripts import utils as script_utils
@@ -25,7 +26,7 @@ def main():
 
     logger.warn('Date_modified field will be added to all nodes.')
 
-    for node in models.Node.find():
+    for node in models.Node.find(Q('date_modified', 'eq', None)):
         node.date_modified = date_updated(node)
         node.save()
         logger.info('Node {0} "date_modified" added'.format(node._id))


scripts/migration/migrate_date_modified_for_existing_nodes.py
```diff
@@ -25,7 +26,7 @@ def main():
 
     logger.warn('Date_modified field will be added to all nodes.')
 
-    for node in models.Node.find():
+    for node in models.Node.find(Q('date_modified', 'eq', None)):
```
sloria: You may also need to also query `Q('date_modified', 'exists', False)`, but I'm not sure. Can you verify?

pattisdr: I don't understand what this query adds.  I have four nodes, one has date_modified as a field, the others do not.  If I do `Node.find(Q('date_modified', 'eq', None))`, three nodes are returned, and if I do `Node.find(Q('date_modified', 'exists', False))`, three nodes are also returned.

sloria: But did you create those nodes after the `date_modified` field was added to the codebase? What about nodes that were created before the field was added?

pattisdr: I dropped my database, switched to an older branch, and created four nodes.  None of them had date_modified as a field.  Then I made one node have date_modified as a field.

pattisdr: Can you explain what that query means?

sloria: It should query for nodes that don't have the `date_modified` at all.

```
{_id: '123abc'}

// vs. 

{_id: '123abc', date_modified: '2015-.....'}
```

But it sounds like it might be unnecessary?

pattisdr: It seems to me like `Node.find(Q('date_modified', 'eq', None))` returns nodes that don't have date_modified at all.

sloria: OK, thanks for verifying.

PR #4800, opened 2015-12-21
Use APIv2 to fetch current list of contributors in contribAdder

samchrisinger:
# Purpose

The v1 call to get_contributors was omitting non-bibliographic contribs,
and consequently those users were appearing as addable in the UI.

see: https://openscience.atlassian.net/browse/OSF-5461

# Changes

Use the v2 endpoint for fetching Node contributors.

### DIFF
--- a/website/static/js/contribAdder.js
+++ b/website/static/js/contribAdder.js
@@ -158,15 +158,22 @@ var AddContributorViewModel = oop.extend(Paginator, {
     getContributors: function() {
         var self = this;
         self.notification(false);
-        return $.getJSON(
-            self.nodeApiUrl + 'get_contributors/', {},
-            function(result) {
-                var contributors = result.contributors.map(function(userData) {
-                    return userData.id;
-                });
-                self.contributors(contributors);
-            }
-        );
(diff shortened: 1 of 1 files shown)


website/static/js/contribAdder.js
```diff
@@ -158,15 +158,22 @@ var AddContributorViewModel = oop.extend(Paginator, {
...
-            function(result) {
-                var contributors = result.contributors.map(function(userData) {
-                    return userData.id;
-                });
-                self.contributors(contributors);
-            }
-        );
+        var url = window.contextVars.apiV2Prefix + 'nodes/' + window.contextVars.node.id + '/contributors/';
+
+        return $.ajax({
```
mfraezz: There's an [osfHelper function](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/osfHelpers.js#L65) that would simplify this request, but it's not strictly necessary

mfraezz:
Pass complete, no bugs found. :elephant: 

Verified that it works as intended; non-bibliographic contributors are now displayed properly:
screen shot 2015-12-23 at 11 34 01

PR #4804, opened 2015-12-21
Feature: Replace v1 comment routes with v2 routes

samanehsan:
Purpose
------------
Replace existing API v1 routes with API v2 routes for comments. 

Changes
------------
In addition to changing the routes...

CommentSerialzer:
- Add `can_edit`, `has_children` and `is_abuse` fields to the comment serializer
- Add max_length to comment content field 
- Do not allow whitespace comments

Comment model:
- `create` method sends a signal to send the notification event, since that logic was happening in the v1 create view function which is no longer being used.

Side Effects
----------------
Instead of escaping html in comments and making it a string, it gets removed entirely by DRF. E.g. the comment `"<em>hi</em>"` would be `"hi"` instead.


sloria:
Reviewed in #4702 -- Looks good to merge.

PR #4806, opened 2015-12-22
Feature: Allow users to comment on files

samanehsan:

Changes
------------
<b>OSF Changes:</b>

<b>API Changes</b>: 
- File serializer:

Post-Merge
---------------
Run `scripts/update_comments.py` to update format of `User.comments_viewed_timestamp`


api/base/serializers.py
```diff
@@ -274,11 +275,19 @@ class RelationshipField(ser.HyperlinkedIdentityField):
         related_view_kwargs={'node_id': '<_id>', 'wiki_id': '<wiki_pages_current.home>'}
     )
 
+    Field can include optional filters:
```
sloria: Can you explain further what this does? `develop` currently has a `filter_key` param; how is this different?

sloria:
I'm seeing a traceback when viewing a project Overview page:

```
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/flask/app.py", line 1836, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/flask/app.py", line 1820, in wsgi_app
    response = self.make_response(self.handle_exception(e))
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/flask/app.py", line 1403, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/flask/app.py", line 1381, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/sloria/projects/cos-projects/osf/framework/routing/__init__.py", line 111, in wrapped
    data = fn(*args, **kwargs)
  File "/Users/sloria/projects/cos-projects/osf/website/views.py", line 361, in resolve_guid
    return proxy_url(url)
  File "/Users/sloria/projects/cos-projects/osf/framework/routing/__init__.py", line 267, in proxy_url
    response = app.view_functions[match[0]](**match[1])
  File "/Users/sloria/projects/cos-projects/osf/framework/routing/__init__.py", line 111, in wrapped
    data = fn(*args, **kwargs)
  File "/Users/sloria/projects/cos-projects/osf/website/project/decorators.py", line 95, in wrapped
    return func(*args, **kwargs)
  File "/Users/sloria/projects/cos-projects/osf/website/project/decorators.py", line 238, in wrapped
    return response or func(*args, **kwargs)
  File "/Users/sloria/projects/cos-projects/osf/website/tokens/__init__.py", line 80, in wrapper
    return func(*args, **kwargs)
  File "/Users/sloria/projects/cos-projects/osf/website/project/views/node.py", line 376, in view_project
    ret = _view_project(node, auth, primary=primary)
  File "/Users/sloria/projects/cos-projects/osf/website/project/views/node.py", line 720, in _view_project
    n_unread_node = Comment.find_n_unread(user, node, page='node')
  File "/Users/sloria/projects/cos-projects/osf/website/project/model.py", line 245, in find_n_unread
    return cls.n_unread_node_comments(user, node)
  File "/Users/sloria/projects/cos-projects/osf/website/project/model.py", line 267, in n_unread_node_comments
    view_timestamp = user.get_node_comment_timestamps(node, 'node')
  File "/Users/sloria/projects/cos-projects/osf/framework/auth/core.py", line 1346, in get_node_comment_timestamps
    page_timestamps = timestamps.get(page, default_timestamp)
AttributeError: 'datetime.datetime' object has no attribute 'get'
```

By the way, I am looking at a local branch that has this PR merged with develop; not sure if that makes a difference.

sloria:
FWIW, I haven't been able to reproduce the above traceback on new projects.

sloria:
The tooltip says "Discussion Pane"

osf___alt_snapshot_gif

Was this in the spec? Why not the more conventional "Comments"?

sloria:
It isn't clear how to close the comment pane.

Some suggestions:

- Clicking outside of the comment pane hides it.
- Add an "x" in the top-right corner of the comment pane for closing it
- Indicate the toggle state in the speech bubble icon. Maybe the speech bubble should be filled with blue when the pane is active.

These can be added in a follow-up PR. Ping @erinmayhood for feedback ticket creation.

sloria:
Did we ever come to a consensus on whether to issue GUIDs for comments?

sloria:
The cursor for this button should be a pointer because it is clickable:

osf_____foo_txt

sloria:
As a user, it is difficult to find replies. It isn't obvious that only comments with the "+" icon have replies. It would be user-hostile to give the impression that replies got lost.

I think adding a tooltip to the "+" button that shows the number of replies will help to some extent. This can happen in a follow-up PR, but go ahead and add it if it's a trivial change.

@erinmayhood You may want to have a look at this. I think there is room for improvement in a follow-up PR/release.

sloria:
Is the backref defined here ever used? https://github.com/samanehsan/osf.io/blob/9b4c5d9afacf9a8a997b62fae4c1fda62e8acb65/website/project/model.py#L184

If it's not used, please remove it. If it is used, can you use a query instead of using a backref? We have been trying to avoid backrefs (see #4810 and #4813) because they result in document locking. In this case, a comment's target will be locked while a comment is being written to.

I realize this is part of a previous comments PR, but it'd be good to address this issue before we release.

api/comments/serializers.py
```diff
@@ -26,17 +28,19 @@ class CommentSerializer(JSONAPISerializer):
...
         'date_created',
         'date_modified',
+        'page',
         'target'
     ])
 
     id = IDField(source='_id', read_only=True)
     type = TypeField()
     content = AuthorizedCharField(source='get_content', required=True, max_length=osf_settings.COMMENT_MAXLENGTH)
+    page = ser.CharField(read_only=True)
```
sloria: Can you explain why this field and its corresponding modm field are necessary given that we already have `root_target`?

api/comments/urls.py
```diff
@@ -3,7 +3,6 @@
 
 urlpatterns = [
     url(r'^(?P<comment_id>\w+)/$', views.CommentDetail.as_view(), name=views.CommentDetail.view_name),
-    url(r'^(?P<comment_id>\w+)/replies/$', views.CommentRepliesList.as_view(), name=views.CommentRepliesList.view_name),
```
sloria: :+1:

website/static/js/comment.js
```diff
@@ -571,7 +605,7 @@ CommentModel.prototype.onSubmitSuccess = function() {
 /*
     *
     */
-var CommentListModel = function(canComment, hasChildren, currentUser) {
+var CommentListModel = function(options) {
```
sloria: :+1:

website/static/js/pages/project-base-page.js
```diff
@@ -26,34 +27,34 @@ if (node.isPublic && node.piwikSiteID) {
...
-    };
-    var ADJUST_PANEL_DEBOUNCE = 10;
-    var RESIZE_DEBOUNCE  = 50;
-
-    self.adjustPanelPosition(); /* Init when refreshing the page*/
-    $(window).scroll($osf.debounce(self.adjustPanelPosition, ADJUST_PANEL_DEBOUNCE));
-    $(window).resize($osf.debounce(self.adjustPanelPosition, RESIZE_DEBOUNCE));
+$(window).scroll(function() {
+    var scrollTop = $(this).scrollTop();
+    var offset = OFFSET;
```
sloria: This variable seems unnecessary.

website/static/js/pages/project-base-page.js
```diff
@@ -26,34 +27,34 @@ if (node.isPublic && node.piwikSiteID) {
...
-            $('.cp-handle').css('margin-top', NAV_MAX_TOP_MARGIN);
-        }
-    };
-    var ADJUST_PANEL_DEBOUNCE = 10;
-    var RESIZE_DEBOUNCE  = 50;
-
-    self.adjustPanelPosition(); /* Init when refreshing the page*/
-    $(window).scroll($osf.debounce(self.adjustPanelPosition, ADJUST_PANEL_DEBOUNCE));
-    $(window).resize($osf.debounce(self.adjustPanelPosition, RESIZE_DEBOUNCE));
+$(window).scroll(function() {
```
sloria: Can you please explain the purpose of this code?

website/templates/include/comment_pane_template.mako
```diff
@@ -0,0 +1,49 @@
...
+    <div class="cp-bar"></div>
+
+
+    <div class="comments cp-sidebar">
+        <div class="cp-sidebar-content">
+            <button type="button" class="close visible-xs text-smaller" data-bind="click: togglePane">
+                <i class="fa fa-times"></i>
+            </button>
+            <h4>
+                <span data-bind="if: page() == 'node' ">${node['title']} Discussion</span>
```
sloria: Is there a reason to use mako here instead of Knockout? Does using mako prevent a FOUC?

sloria:
Pass finished. Looking good. I mostly had questions for clarification.

samanehsan:
`AttributeError: 'datetime.datetime' object has no attribute 'get'` occurs on old projects because a migration needs to be run

sloria:
@samanehsan Ah ok. Thanks. Can you please list that migration script and any other deployment steps in the PR description?

samanehsan:
I just updated the description @sloria!

sloria:
Thanks

samanehsan:
>The tooltip says "Discussion Pane"

>osf___alt_snapshot_gif

>Was this in the spec? Why not the more conventional "Comments"?

I do not think is covered by the spec. @erinmayhood, is it okay to leave this tooltip here and change it to say "Comments"?

samanehsan:
> Did we ever come to a consensus on whether to issue GUIDs for comments? 

We did not really discuss it. Current behavior is that comments have GUIDs.

api/base/serializers.py
samanehsan: `RelationshipField` formats a link with the view and view_kwargs that get passed to it. The `filter` param will append a filter query to that link. 

The `filter_key` param in develop is essentially the `source_field`. It looks like when you try to filter on a `RelationshipField` the source is `*` so it has to be defined: https://github.com/CenterForOpenScience/osf.io/commit/64fbe9d73d1fb1c7347b58ab29af4f0760a798ba

website/templates/include/comment_pane_template.mako
samanehsan: I think it's because this information is coming from `_view_project` and not from the comment view model.

website/static/js/pages/project-base-page.js
samanehsan: When the comment pane is open on small screen sizes (when the project navigation menu is collapsed) and you scroll down the page there is a gap where the nav bar used to be. This code moves the comment panel to be in line with the nav bar. 

screen shot 2016-01-04 at 10 57 39 am

screen shot 2016-01-04 at 10 52 24 am

api/comments/serializers.py
samanehsan: The `page` field is primarily used to make filtering easier. In the discussions page (which is not a part of this PR), users have to option to view all comments on the project overview page, all comments on files, etc.

samanehsan:
Done responding to your comments @sloria! :octocat: I also fixed an issue I found with fetching comments.

sloria:
Thanks. Can you please look into the failing test? It appears to be related to comments on deleted files.

api/comments/serializers.py
sloria: OK, I see. It might be useful to document this on the model field. Also, It would be good to add validation here (you could use a `ChoiceField` here); this will have the ancillary benefit of self-documentation.

erinmayhood:
@samanehsan a few responses:

1. x should close the comment pane
2. Clicking outside the comment pane should close it
3. Tooltip should say Comments
4. I'm checking with @JeffSpies about GUIDs

Thanks!

scripts/update_comments.py
```diff
@@ -0,0 +1,50 @@
...
+        comment.page = Comment.OVERVIEW
+        comment.save()
+
+
+if __name__ == '__main__':
+    dry = '--dry' in sys.argv
+    if not dry:
+        script_utils.add_file_logger(logger, __file__)
+    init_app(routes=False, set_backends=True)
+    with TokuTransaction:
```
sloria: I'm getting an error:

```
    exec code in run_globals
  File "/Users/sloria/projects/cos-projects/osf/scripts/update_comments.py", line 47, in <module>
    with TokuTransaction:
AttributeError: __exit__
```

Looks like you need to do:

```python
with TokuTransaction():
```

JeffSpies:
Should get guids; should be longer though. Uuids or longer guids would be fine. Probably won't be referenced too much in print. What is the precedent for longer guids?

Sent and possibly dictated from my iPhone

scripts/update_comments.py
```diff
@@ -0,0 +1,50 @@
...
+from modularodm import Q
+
+from framework.auth.core import User
+from framework.transactions.context import TokuTransaction
+
+from website.models import Comment
+from website.app import init_app
+from scripts import utils as script_utils
+
+logger = logging.getLogger(__name__)
```
sloria: There isn't any logging in the script. It would be useful to log--at the very least--the IDs of users who were migrated.

sloria:
> 1. x should close the comment pane
2. Clicking outside the comment pane should close it
3. Tooltip should say Comments
4. I'm checking with @JeffSpies about GUIDs

@erinmayhood Which, if any of these, should be included in this PR/release? I suggest adding the "x" and changing the tooltip for the MVS. Number 2 can be added in a follow-up PR.

sloria:
> Should get guids; should be longer though. Uuids or longer guids would be fine. Probably won't be referenced too much in print. What is the precedent for longer guids?

We do not currently have a system for longer GUIDs. 

Alternatively, we could do UIDs namespaced by project GUIDs. 

GitHub uses the following scheme for issue comments:

```
https://github.com/CenterForOpenScience/osf.io/pull/4806#issuecomment-168767261
```

Reddit uses the following:

```
https://www.reddit.com/r/science/comments/3ze9ym/promising_progress_for_new_treatment_of_type_1/cylk4v2
```

Where the first ID (3ze9ym) is the article ID and the second ID (cylk4v2) is the comment ID.

sloria:
@erinmayhood Also, any suggestions for making replies more obvious (https://github.com/CenterForOpenScience/osf.io/pull/4806#issuecomment-167630961)? Quite easy for them to get lost ATM. Let me or Saman know if you'd like to see this in person.

sloria:
Unless this confirmation step when restoring a comment was documented in the spec, I suggest removing it.

osf_____foo_txt

Deleting/restoring a comment is an easily reversible action; no need to interrupt the flow with a confirmation.

sloria:
Minor: I suggest using muted color and italics to distinguish deleted comments from regular comments.

Before:

osf_____foo_txt

After (using `.text-muted` and `<em>`):

osf_____foo_txt

Certainly not a blocker, but it's an easy change.

website/project/model.py
```diff
@@ -216,19 +236,52 @@ def get_content(self, auth):
...
-                user.save()
-            view_timestamp = user.comments_viewed_timestamp.get(node._id, default_timestamp)
-            n_unread = Comment.find(Q('node', 'eq', node) &
-                                    Q('user', 'ne', user) &
-                                    Q('is_deleted', 'ne', True) &
-                                    (Q('date_created', 'gt', view_timestamp) |
-                                    Q('date_modified', 'gt', view_timestamp))).count()
+            if not page:
+                return cls.n_unread_node_comments(user, node) + cls.n_unread_file_comments(user, node)
+            elif page == 'node':
```
sloria: You can use the named constants (e.g. `Comment.OVERVIEW`) here.

website/project/model.py
```diff
@@ -216,19 +236,52 @@ def get_content(self, auth):
         return self.content
 
     @classmethod
-    def find_unread(cls, user, node):
-        default_timestamp = datetime.datetime(1970, 1, 1, 12, 0, 0)
+    def find_n_unread(cls, user, node, page=None, root_id=None, check=False):
         n_unread = 0
```
sloria: This variable is now unnecessary; it is never reassigned, so you can just `return 0` at the end of the method.

api/base/serializers.py
sloria: Makes sense. Thanks

api/comments/serializers.py
```diff
@@ -71,7 +75,7 @@ def get_can_edit(self, obj):
         return obj.user._id == user._id
 
     def get_has_children(self, obj):
-        return bool(getattr(obj, 'commented', []))
+        return Comment.find(Q('target', 'eq', obj)).count() > 0
```
sloria: Would be good to add an index to `target` to optimize this query.

api/nodes/serializers.py
```diff
@@ -154,9 +155,28 @@ def get_pointers_count(self, obj):
         return len(obj.nodes_pointer)
 
     def get_unread_comments_count(self, obj):
```
sloria: Does this get computed for every node that gets serialized, or is this opt-in (e.g. via a query parameter, as with `?related_count=true`). Does this significantly impact response times? Would be good to benchmark this with projects with many comments.

sloria:
Every click to expand or collapse a comment's replies results in 4 requests:

comment_requests

- Is this the desired behavior? I can see you might want this if we want the user to be able to "refresh" a comment's replies by collapsing than re-expanding a node, but that's not an intuitive way to refresh (a user is more likely to just hit "refresh" in their browser). Probably better to prevent the extra requests.
- Why is the request to `/v2/nodes/<nid>/?related_counts=true` necessary for every expand/collapse?

api/comments/serializers.py
samanehsan: Isn't the `ChoiceField` only for writable values? I made `page` read-only so that whenever a comment is created the page is determined from the root target.

website/templates/project/project_header.mako
```diff
@@ -68,8 +81,18 @@
                         % if user['has_read_permissions'] and not node['is_registration'] or (node['is_registration'] and 'admin' in user['permissions']):
                             <li><a href="${node['url']}settings/">Settings</a></li>
                         % endif
-
                     % endif
+                    % if user['can_comment'] or node['has_comments']:
```
sloria: Is this UI supposed to be visible? I'm not seeing a "Comments" button on a project that I am an admin on.

sloria:
Pass finished.

samanehsan:
> Unless this confirmation step when restoring a comment was documented in the spec, I suggest removing it.

This is the current behavior on production -- should I still make this change?

sloria:
> This is the current behavior on production -- should I still make this change?

Ah ok. Don't worry too much about it then. If it's an easy change, go ahead and make it. If it takes > 5 min, let's defer.

website/templates/project/project_header.mako
samanehsan: It refers to the "Comments" link in the project nav bar on small screen sizes:  
screen shot 2016-01-04 at 3 25 22 pm

sloria:
^ In fact, you can apply the "5-minute rule" to most of my comments above, as most of them aren't blockers. Though it'd be worth looking into preventing extra requests (https://github.com/CenterForOpenScience/osf.io/pull/4806#issuecomment-168789900), and also doing a quick check to make sure that getting unread counts isn't too expensive (https://github.com/CenterForOpenScience/osf.io/pull/4806#discussion_r48772359).

api/nodes/serializers.py
samanehsan: I think I made it opt-in, but I don't think it is opt-in any more. I will change it back to opt-in, since getting the number of unread comments is non-trivial.

api/comments/serializers.py
samanehsan: What do you mean by adding an index?

samanehsan:
> Can you please look into the failing test? It appears to be related to comments on deleted files.

@sloria, this test is failing because the node comments list filters out comments where `root_target` is `None`, but now that the `root_target` backref is removed the `root_target` does not get set to `None` when a file is deleted. 

Should I add the backref back in? Another option is to check if the `root_target` of the comment still exists but that could make the request take longer.

sloria:
From @erinmayhood above, for visibility:

> @sloria @samanehsan agreed to the first three in this release. @JeffSpies hasn't been able to respond on GUIDs so let's keep things as they are for this release and put this on the agenda to discuss with @sloria , @icereval and @JeffSpies for a final decision.

api/comments/serializers.py
sloria: Oh, that's right. Disregard my comment.

sloria: It would still be good to have a validator on the modm field.

api/comments/serializers.py
sloria: A MongoDB index. It's a good idea to add `index=True` to fields that you will query often.

sloria:
> @sloria, this test is failing because the node comments list filters out comments where root_target is None, but now that the root_target backref is removed the root_target does not get set to None when a file is deleted.

> Should I add the backref back in? Another option is to check if the root_target of the comment still exists but that could make the request take longer.

I think you can add the backref back in for now. @brianjgeiger Do you see any problem with this for the short-term? We can explore a better solution as a follow-up to this PR.

sloria:
@brianjgeiger For context, we are talking about adding back the `comment_target` backref to this field: https://github.com/CenterForOpenScience/osf.io/blob/9d1bb2e512f93129ad6e8e12fffd93662eda8d72/website/project/model.py#L184

brianjgeiger:
So are we talking one MODM query per item, worst case, or is it more complicated than that?

JeffSpies:
I think I’ve responded twice about this. We need longer GUIDs. Is this not going through?

samanehsan:
@brianjgeiger, I'm not sure if this answers your question but in this case I cannot replace the backref with an ODM query. Whenever a `root_target` is deleted, the `root_target` field of all of its comments are set to `None`

@JeffSpies, do you mean we should give comments longer GUIDs? Or keep them as is and use longer ones when we run out of the 5-character ones?

JeffSpies:
We should give comments 12 character GUIDs and save our 5 character space for users/projects. Make sense?

sloria:
@JeffSpies : @icereval and I discussed using the normal Mongo ObjectIDs for comments. We thought it would be as simple as changing the `Comment` to a `StoredObject` instead of a `GuidStoredObject`. However, I believe @samanehsan tried this and it wasn't as trivial as we thought (correct me if I'm wrong, Saman). Therefore, @erinmayhood and I decided it best to defer the change for a follow-up PR in the interest of getting this to staging ASAP.

JeffSpies:
We should probably be consistent. Is it problematic to just change the length of the GUID? There used to be an argument for that.

sloria:
We haven't implemented a way to issue longer **G**UIDs; right now, we have either 5-char GUIDs or 24-char UIDs.

JeffSpies:
By my look at the code, the longer GUIDs is a 3 line change across 3 files. 

And I thought you said we didn’t have a history of using UIDs?

Unfortunately, I’d suggest that this a blocker and will need to be on staging to make sure we have the appropriate respect for provenance and URIs. UIDs might be a fine temporary patch if my above estimate is off (it could be).

Jeff.

sloria:
To be clear: GUIDs for Comments is the status quo; this PR doesn't change anything about that.

We have always used ObjectIds for non-GUID models (with the exception of `Tag` which uses the tag text as its primary key).

@samanehsan Can you look into giving Comments longer GUIDs?

samanehsan:
Yes, I can look into it. I think this can go in a follow-up PR though, since as Steve mentioned comments are already using the 5-character GUIDs on production.

JeffSpies:
A quick fix; if you add length as a parameter to save, ensure_guid, and generate, you can pass through length in the save function of the comment.

https://github.com/CenterForOpenScience/osf.io/blob/04056a550b2585b01c01183f473f0c166c3b8a15/framework/guid/model.py <https://github.com/CenterForOpenScience/osf.io/blob/04056a550b2585b01c01183f473f0c166c3b8a15/framework/guid/model.py>

@sloria, the early spec was to be able to cite comments as first-class objects—that means using guids. Where do we use osf.io/objectid or osf.io/uuid?

Thanks,

J.

sloria:
> @sloria, the early spec was to be able to cite comments as first-class objects—that means using guids. Where do we use osf.io/objectid or osf.io/uuid?

@JeffSpies Ah ok, I was unaware of that requirement. Thanks for clarifying. We do not use `osf.io/<objectid>` anywhere.

I suggest a slight modification to the quick fix: make the length configurable at the class level.

```python
class GuidStoredObject(...):
    __guid_min_length__ = 5

    def _ensure_guid(self):
        ...
        # need to add length param
        Guid.generate(min_length=self.__guid_min_length__)

class Comment(GuidStoredObject):
    __guid_min_length__ = 12
```

That way, you don't have to change the signature of `StoredObject#save`, and you don't have to remember to pass `length` for every `save` call or override `Comment#save`.

JeffSpies:
I’m not sure if it’s in the current spec; it may very well have been lost in the current spec.

Much, much better solution. I forgot that Comments inherited GuidStoredObjects.

J.

sloria:
@samanehsan Please resend this to `release/0.62.0` when this is ready for review.

PR #4809, opened 2015-12-22
Add node_logs endpoint to user

hmoco:


api/users/views.py
```diff
@@ -403,3 +405,23 @@ def get_default_odm_query(self):
...
+
+    def get_queryset(self):
+        current_user = self.request.user
+        if current_user.is_anonymous():
+            auth = Auth(None)
+        else:
+            auth = Auth(current_user)
+        query = self.get_query_from_request()
+        raw_nodes = Node.find(self.get_default_odm_query() & query)
+        nodes = [each for each in raw_nodes if each.is_public or each.can_view(auth)]
```
samchrisinger: You can add this into the `.find`:

```
Node.find(self.get_default_odm_query() & query & (
    Q('is_public', 'eq', True) | Q('permissions.{0}', 'icontains', osf_permissions.READ)
)
```

api/users/views.py
```diff
@@ -403,3 +405,23 @@ def get_default_odm_query(self):
...
+        if not nodes:
+            raise NotFound
+        query = [Q('params.node', 'eq', node._id) for node in nodes]
```
samchrisinger: Could be `Q('params.node', 'in', [n._id for n in nodes])

PR #4810, opened 2015-12-22
[Ready for Review] Remove user nodes backref

erinspace:
Relates to https://openscience.atlassian.net/browse/OSF-5406

# Purpose
The backref on the Node's contributor field to a user was causing the user to be locked down when trying to update a node. This PR removes the backref ```contributed``` on the ```contributors``` field of the Node model, and instead adds two things elsewhere:
- a ```contributed``` property on the User model that does a query for all of the nodes that user has contributed to
- a ```find_for_user``` classmethod on the Node that will find all of the nodes that a user contributes to, with an optional subquery that when used can narrow down results even more.

# Changes
- Add ```contributed``` property to the User Model, that returns a queryset with the user's nodes
- Add a ```find_for_user``` classmethod to the Node that accepts optional subquery to further narow results
- Change every instance of ```user.node__contributed``` to just use the property ```user.contributed```
- When there was a subsequent call to ```find()``` on the former ```user.node__contributed```, replace that with the ```find_for_users``` classmethod, passing in the user and any other subquery to narrow the results
- Update merging user tests to reload before checking for the merged user properties

# Side effects
- Can no longer use the ```node.user__contributed``` backref


website/views.py
```diff
@@ -191,23 +192,27 @@ def get_dashboard_nodes(auth):
...
-
-    nodes = contributed.find(
-        Q('category', 'eq', 'project') &
-        Q('is_deleted', 'eq', False) &
-        Q('is_registration', 'eq', False) &
-        Q('is_folder', 'eq', False)
+    nodes = Node.find_for_user(
+        user,
+        subquery=(
+            Q('category', 'eq', 'project') &
```
brianjgeiger: I know this is traditionally correct, and I'm not sure what's relying on this, but this is technically not true any more. Category can be project for subcomponents. I would recommend switching this to checking for a null parent instead. @sloria, so you agree?

website/views.py
```diff
@@ -191,23 +192,27 @@ def get_dashboard_nodes(auth):
...
-            Q('category', 'ne', 'project') &
-            # exclude deleted nodes
-            Q('is_deleted', 'eq', False) &
-            # exclude registrations
-            Q('is_registration', 'eq', False)
+        comps = Node.find_for_user(  # NOTE - this used to be a find on nodes above. Does this mess it up?
+            user,
+            (
+                # components only
+                Q('category', 'ne', 'project') &
```
brianjgeiger: See above about the difference between projects and components in the modern world.

website/views.py
brianjgeiger: And yes, I recognize that this is entirely out of scope for this PR. But if we agree that it should be done but believe it shouldn't be done here, it would be good to have an issue made for it.

sloria: Good point. Since this is such a small change, I think we can go ahead and make it in this PR. @erinspace Can you try out Brian's suggestion and make sure the Dashboard still has the correct nodes?

erinspace: @brianjgeiger @sloria made the change to query for ```Q('parent', 'eq', None)``` instead of ```Q('category', 'eq', 'project)``` and the two dashboard node tests ```test_get_dashboard_nodes``` and ```test_get_dashboard_nodes_shows_components_if_user_is_not_contrib_on_project```  **did** indeed fail, both of them finding too many nodes.

Happy to investigate further if I should continue down this path, just let me know.

sloria: OK, let's hold off on this change for now. @erinspace Can you please create a follow-up ticket to investigate this further?

erinspace: Yep! Created a jira issue here: https://openscience.atlassian.net/browse/OSF-5472

Labeled it as "improvement" -- tho is it a bug?

brianjgeiger: It is, but I changed it.

PR #4813, opened 2015-12-22
Remove node created backref

erinspace:
Similar to #4810 
Relates to https://openscience.atlassian.net/browse/OSF-5406

# Purpose
Remove the ```created``` backref on the ```creator``` field of the Node, to prevent unecessary locking when updating nodes or users caused by the backref.

# Changes
- Remove ```created``` backref on the ```creator``` field of the Node
- Make new ```created``` property on User that queries for nodes that user has created
- Update merge_user method of User model to use new ```created``` property
- Remove test made to test the removed backref
- Reload ingeretance project test to update test context

# Side effects
- ```created``` backref can no longer be used


website/project/model.py
```diff
@@ -747,7 +747,7 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin):
     wiki_private_uuids = fields.DictionaryField()
     file_guid_to_share_uuids = fields.DictionaryField()
 
-    creator = fields.ForeignField('user', backref='created')
+    creator = fields.ForeignField('user')
```
sloria: It would be good to add an index to this field, since we will often query `Node` on this field alone (as you've implemented in `User.created`).

framework/auth/core.py
```diff
@@ -649,6 +649,11 @@ def csl_name(self):
             'given': self.csl_given_name,
         }
 
+    @property
+    def created(self):
```
sloria: Please add a unit test for this new property.

sloria:
Pass finished.

website/project/model.py
```diff
@@ -747,7 +747,7 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin):
     wiki_private_uuids = fields.DictionaryField()
     file_guid_to_share_uuids = fields.DictionaryField()
 
-    creator = fields.ForeignField('user', backref='created')
+    creator = fields.ForeignField('user', index=True)
     contributors = fields.ForeignField('user', list=True, backref='contributed')
```
brianjgeiger: What's happening with contributors?

brianjgeiger: Oh, nvm. #4810

framework/auth/core.py
```diff
@@ -1292,7 +1297,7 @@ def merge_user(self, user):
                 node.save()
 
         # - projects where the user was the creator
-        for node in user.node__created:
+        for node in user.created:
```
brianjgeiger: As discussed in person: so this doesn't actually appear to be *used* anywhere. Can it be removed entirely?

erinspace: I removed the created property and all tests passed just fine! Happy to do it. 

Would this be just removing the created property on the user, or the creator field from the node itself?

brianjgeiger: Ideally the creator filed from the Node, but yeah, hold off on that. Removing the backref is the important thing. It does get used a *tiny* bit at least, so no sense causing trouble in this ticket.

erinspace: ok cool. Leaving it as is for now.

PR #4821, opened 2015-12-23
Only append comments if actually saved

samchrisinger:
# Purpose

There was a general lack of user feedback when adding/saving comments.

# Changes

- only add a comment to the UI once saved sucessfully
- block the commenting UI while a comment is saving

### DIFF
--- a/website/static/js/registrationUtils.js
+++ b/website/static/js/registrationUtils.js
@@ -291,16 +291,22 @@ var Question = function(questionSchema, data) {
  *
  * @param {function}: save: save function for the current registrationDraft
  **/
-Question.prototype.addComment = function(save) {
+Question.prototype.addComment = function(save, page, event) {
     var self = this;
 
     var comment = new Comment({
         value: self.nextComment()
     });
     comment.seenBy.push($osf.currentUser().id);
-    self.comments.push(comment);
-    self.nextComment('');
-    save();
+
+    var $comments = $(event.target).closest('.registration-editor-comments');
+    $osf.block('Saving...', $comments);
+    save()
+        .always($osf.unblock.bind(null, $comments))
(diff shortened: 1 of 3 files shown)


sloria:
@samchrisinger Can you please look into the failing JS test?

website/static/js/registrationUtils.js
```diff
@@ -291,16 +291,22 @@ var Question = function(questionSchema, data) {
...
-    self.comments.push(comment);
-    self.nextComment('');
-    save();
+
+    var $comments = $(event.target).closest('.registration-editor-comments');
+    $osf.block('Saving...', $comments);
+    return save()
+        .always($osf.unblock.bind(null, $comments))
+        .done(function () {
+            self.comments.push(comment);
```
sloria: Why was this moved to the `done` callback? Optimistic updates--e.g. adding a comment without having to wait for a HTTP request to finish--is better UX, so long as errors are handled gracefully.

samchrisinger: @sloria this came about based on feedback from QA. There was a general lack of feedback when commenting, and blocking the UI while saving was an easy addition.

Error handling for the comment UI could get awkward-- type a comment, see it get added, then a red alert appears and the comment is deleted? Or moved back to a pending state?

sloria: > Error handling for the comment UI could get awkward-- type a comment, see it get added, then a red alert appears and the comment is deleted? Or moved back to a pending state?

I'd think it would be moved back to a pending state. Not that it matters so much. The overwhelmingly common case will be that the request succeeds, so better to optimize the UI for that case.

All that said, this PR is an improvement to the status quo, so I'll stop picking nits. =)

PR #4487, opened 2015-10-20
[Feature] Multiple public projects dialog

billyhunt:
Jira Issue OSF-4457

Purpose
-----------
When making a project public or private, there is no option to make the children of the project (components or projects under the project) public or private at the same time.  

Changes
------------
Here are the three views that will guide the user through the permissions process.
  
View 1.  warn users of the implications of making their work public or private.  
View 2.  A tree with checkboxes to allow users to change the privacy settings for some or all of the child nodes under their project.
View 3.  A final warning saying which addons will be effected by changing the privacy settings of the projects or components selected.

Side Effects
---------------
1.  Making a single node public or private will be more complicated (four buttons to click vs three).
2.  Both Public and Private buttons will link to the same privacy settings dialog.  This could add confusion.  It is currently acceptable as a project organizer redesign is in the works, and the new UI should more accurately reflect the idea of privacy settings. 

#### Before:  
screen shot 2015-10-20 at 10 13 48 am

#### Now:

screen shot 2015-11-21 at 8 01 36 pm
screen shot 2015-11-21 at 8 01 46 pm
screen shot 2015-11-21 at 8 01 52 pm

PR #4829, opened 2015-12-28
Feature/multiple public projects dialog

billyhunt:
Jira Issue OSF-4457

Resubmission of #4487

Purpose
-----------
When making a project public or private, there is no option to make the children of the project (components or projects under the project) public or private at the same time.  

Changes
------------
Here are the three views that will guide the user through the permissions process.
  
View 1.  warn users of the implications of making their work public or private.  
View 2.  A tree with checkboxes to allow users to change the privacy settings for some or all of the child nodes under their project.
View 3.  A final warning saying which addons will be effected by changing the privacy settings of the projects or components selected.

Side Effects
---------------
1.  Making a single node public or private will be more complicated (four buttons to click vs three).
2.  Both Public and Private buttons will link to the same privacy settings dialog.  This could add confusion.  It is currently acceptable as a project organizer redesign is in the works, and the new UI should more accurately reflect the idea of privacy settings. 

#### Before:  
screen shot 2015-10-20 at 10 13 48 am

#### Now:

screen shot 2015-11-21 at 8 01 36 pm
screen shot 2015-11-21 at 8 01 46 pm
screen shot 2015-11-21 at 8 01 52 pm


billyhunt:
Note:  The issue we thought was an API issue (return request before done) was an error in my code.

billyhunt:
Note:  this went through QA before under the PR #4487 

The primary differences is a the logic for updating nodes via the APIv2 was simplified, the upper limit of the bulk api was changed from 10 to 100 (and corresponding tests), and a newer version of jQuery was added by @brianjgeiger to support Internet Explorer PATCH requests by the API v2.

whit537:
### Punchlist (for @billyhunt)

### Still Investigating (for @whit537) ...

 - [ ] Does this PR address all of the issues raised on #4487?
 - [ ] Stand it up locally and see what we see.
 - [ ] Grok the code and see what we see.
 - [ ] Take a pass through for little things.

whit537:
> Does this PR address all of the issues raised on #4487?

Actually, #4487 is merged (not closed), so I guess this is a non-issue. :space_invader:

sloria:
@whit537 Some context: this feature was merged in #4487 but reverted in https://github.com/CenterForOpenScience/osf.io/commit/8f621fd72a2e8c53873a93f7a865a219704251e2 due to limitations of APIv2 that were causing database lock errors. The API issues have since been addressed, so this PR is a resubmission of that feature with a few modifications to take advantage of APIv2's increased bulk upload limit.

sloria:
...so the majority of the code has been reviewed and QA'd.  @whit537 If you still feel it worthwhile to do further review, by all means go for it. In the interest of getting this to QA, however, I'm going to merge this in now and put it on staging.

whit537:
@sloria Okay, cool. Onward! :-)

PR #4830, opened 2015-12-29
Feature/checkout logging

Marcus-Rosti:
# Implement Checkout/Checkin logging in recent activity
## Purpose
Checking out a file was not logged in recent activity, even though it's a blocking activity. This feature adds that functionality.
## Changes
Several files were amended to add this feature. Primarily, the ./website/files/models/osfstorage.py is where most of the action happened. Here I refactored the save method to log a checkin or checkout activity.  To update the view, I added template for the activity along with signals and function parameters to look for the new activity. No code in these sections was edited; rather, I just added the new parameters.
## Issues
The tests passed locally and the view threw no errors due to the new logs.


website/files/models/osfstorage.py
```diff
@@ -81,10 +84,42 @@ def move_under(self, destination_parent, name=None):
             raise exceptions.FileNodeCheckedOutError()
         return super(OsfStorageFileNode, self).move_under(destination_parent, name)
 
-    def save(self):
+    def save(self):  # TODO write log
```
mfraezz: This `TODO` appears to be done

website/files/models/osfstorage.py
```diff
@@ -81,10 +84,42 @@ def move_under(self, destination_parent, name=None):
...
+        """
+        Uses OsfSFN save to save the node,
```
mfraezz: Docstring is wordy and uses a non-immediately-apparent abbreviation. Consider something like 
>Saves node, logs any check-out or check-in actions

website/files/models/osfstorage.py
```diff
@@ -81,10 +84,42 @@ def move_under(self, destination_parent, name=None):
...
         self.materialized_path = ''
-        return super(OsfStorageFileNode, self).save()
+        ret = super(OsfStorageFileNode, self).save()
+
+        if 'checkout' in ret and len(ret) == 1:
+            # Set parameters
+            auth = Auth(user=self.node.creator)
+
+            # is checkout vs is checkin
+            action = [lambda: NodeLog.CHECKED_IN, lambda: NodeLog.CHECKED_OUT][self.is_checked_out]()
```
mfraezz: I don't think these need to be lamdas

website/files/models/osfstorage.py
```diff
@@ -81,10 +84,42 @@ def move_under(self, destination_parent, name=None):
...
+                    'node': self.node._id,
+                    'urls': {
+                        'download': "/project/" + self.node._id + "/files/osfstorage/" + self._id + "/?action=download",
+                        'view': "/project/" + self.node._id + "/files/osfstorage/" + self._id},
+                    'path': self.materialized_path
+                },
+                auth=auth,
+            )
+
+            file_checkout_toggle.send(self.node, user=self.node.creator)
```
mfraezz: If you're doing the log action here, you don't need to use signals. Emitting a signal to log the action would keep `.save()` cleaner, but signal use is also kind of ugly. I'd say to just remove the signal.

website/files/models/osfstorage.py
Marcus-Rosti: Yeah I agree here.  The idea is that if you're two sides of the if/else are computationally intensive then you would only calculate the one that is chosen.

website/files/models/osfstorage.py
Marcus-Rosti: Got it. Someone walked me through the call stack and said it was necessary but I could never see when it was really called or used.

website/files/models/osfstorage.py
mfraezz: This isn't computationally intensive, the string is in memory because the `NodeLog` model has been imported into memory.

Marcus-Rosti: already swapped it out

website/files/models/osfstorage.py
mfraezz: Yeah, that was me trying to keep `.save()` cleaner because I was told that was best, even though signals are ugly.

mfraezz:
Pass complete, I'll do functional testing after the above comments are resolved. :flags: 

Travis is failing, but it's not obviously your fault, it might be a weird nondeterministic failure. However, please add tests to confirm proper logs are added when files are checked in/out.

mfraezz:
This seems to have developed both a merge conflict and a circular import:

```  
# Circular import
File "/Users/matt/Devel/osf.io/website/project/model.py", line 57, in <module>
    from website.files.models.base import File, StoredFileNode
  File "/Users/matt/Devel/osf.io/website/files/__init__.py", line 1, in <module>
    from website.files import models  # noqa
  File "/Users/matt/Devel/osf.io/website/files/models/__init__.py", line 9, in <module>
    from website.files.models.osfstorage import *  # noqa
  File "/Users/matt/Devel/osf.io/website/files/models/osfstorage.py", line 10, in <module>
    from website.project.model import NodeLog
ImportError: cannot import name NodeLog
```
This can be fixed by using the `NodeLog.ACTIONS` string (`'checked_in'` or `'checked_out'`) when determining the action to log in `osfstorage.py`.

```
# Conflict in api_tests/files/views/test_file_detail.py
<<<<<<< HEAD
=======
from framework.auth.core import Auth
from website.addons.osfstorage import settings as osfstorage_settings

>>>>>>> 62a9ea83e799cf9bdecef06ce94f58df63c81b69
```
`Auth` needs to stay, `osfstorage_settings` is imported a few lines down.


Confirmed that logs were added as expected after resolving the above issues. :space_invader: 
I'll branch off of this and submit a new PR, thanks for your hard work @Marcus-Rosti

PR #4831, opened 2015-12-29
Add an index for the backref connecting NodeLogs to Nodes

brianjgeiger:
Purpose
-----------
NodeLogs is slower than it should be. One reason is because, whenever you get the aggregate logs for a node, it's doing an 'in' query on an unindexed field. This should fix that.  Functionality should be the same, but it should be faster. 

Changes
------------
Modified the NodeLogs model to include an extra index.

Side effects
---------------
If anything, it's *possible* that forking or registering large/complex projects will become slower.


sloria:
Build failure is unrelated to this change. This is good to merge.

PR #4832, opened 2015-12-29
[WIP] Migrate GoogleDrive files/folders to use unique IDs

felliott:
## Purpose

The Google Drive storage addon currently uses the path of the files and folders to identify the stored entities.  When a file is moved or renamed, we don't have a way of tracking that.  GDrive provides unique ids in their metadata which would allow us to do this.  @TomBaxter has already written the WaterButler-side code to allow us to do this, and now we need to update the stored metadata in the OSF with these new IDs.  Since we do not currently track these IDs in the metadata at all, we have to use the users' refresh tokens to query GDrive and get the IDs.
## Changes

All StoredFileNodes with `"provider": "googledrive"` will be updated so that their `path` is set to their ID.  StoredFileNodes that cannot be matched to something currently stored in GDrive will be removed.  If a file is matched and has a new revision, the ID will be updated, and a new entry will be added to its history.  New files or folders will not be created, as that will happen on the next NodeFilesList access.
## Side Effects

After this migration is performed, WB needs to be updated with [this PR](https://github.com/CenterForOpenScience/waterbutler/pull/103).

Fixes: [#OSF-5205]

### DIFF
--- a/scripts/migrate_googledrive_to_uid.py
+++ b/scripts/migrate_googledrive_to_uid.py
@@ -4,6 +4,7 @@ import logging
 import argparse
 import requests
 import itertools
+import collections
 
 from framework.transactions.context import TokuTransaction
 from modularodm.query.querydialect import DefaultQueryDialect as Q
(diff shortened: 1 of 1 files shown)


felliott:
This is crazy stale and needs a rethink b/c of how GDrive models files.  Closing.

PR #4835, opened 2016-01-04
Rename dashboard to bookmarks

brianjgeiger:
Purpose
------------
The new project organizer will no longer have a dashboard collection, but it will have a bookmarks collection that serves much of the same purpose. 

Changes
------------
* Rename `is_folder` in the model to `is_collection` and `is_dashboard` in the model to `is_bookmark_collection`
* Rename `Dashboard` collection to `Bookmarks`
* Add `bookmarks` property to the Collection serializer
* Remove `dashboard` property from Node serializer
* Prevent `is_bookmark_collection` nodes from being renameable
* Migrate rename of 'Dashboard' to 'Bookmarks' (or have it happen automatically)
* Remove v1 views for the project organizer, tests, routes, and related rubeus code
* Remove legacy expand/collapse functionality, views, routes, and tests

screen shot 2016-01-06 at 11 14 02 am

Side effects
----------------
The old javascript relying upon dashboard and folder routes won't work. These changes will only work with the new Project Organizer and similar changes. I'm not modifying v1 views except for the simplest cases.

Migration
------------
It would probably easiest to take down the OSF while doing the migration, since it's a quick migration.  To migrate, 

`python -m scripts.migration.migrate_dashboard_collection_to_bookmarks`

If you need to reverse the migration,

`python -m scripts.migration.migrate_dashboard_collection_to_bookmarks reverse`

If you didn't take down the OSF, there's a tiny, tiny chance that someone could end up with two bookmark collections in the same account, which could be annoying.


sloria:
Can you please document any migration steps in the PR description?

sloria:
Also please document whether the migration will require downtime or not.

PR #4836, opened 2016-01-04
[WIP] Make sure username is always added to email list

erinspace:
Relates to https://openscience.atlassian.net/browse/OSF-5462

# Purpose
There were 50 users on the OSF whose usernames weren't added to their list of emails
Hard to find a pattern, so just added some extra checks.

Stats:
- There are 50 registered
- There are 50 claimed
- There are 0 invited
- There are 0 merged
- There are 0 disabled accounts
- There are 10 with osf4m system tags, coming from OSF for meetings

# Changes
- Add tests that check when an email is confirmed
- Add sentry log message to gather when a username isn't added when an email is confirmed

# Side effects
- Shouldn't be any, as this is just testing and a bit of extra logging and tests

### DIFF
--- a/framework/auth/core.py
+++ b/framework/auth/core.py
@@ -861,7 +861,7 @@ class User(GuidStoredObject, AddonModelMixin):
 
         # log to sentry if somehow the username isn't in the email list
         # This is for data gathering purposes, so we can find when this is happening
-        if username.lower() not in self.emails:
+        if self.username.lower() not in self.emails:
             log_message('Username not added to list of emails.')
 
         return True
--- a/framework/auth/core.py
+++ b/framework/auth/core.py
@@ -844,6 +844,9 @@ class User(GuidStoredObject, AddonModelMixin):
         if email not in self.emails:
             self.emails.append(email)
 
+        if self.username not in self.emails:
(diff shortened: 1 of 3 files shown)


framework/auth/core.py
```diff
@@ -859,6 +862,11 @@ def confirm_email(self, token, merge=False):
 
         self.update_search_nodes()
 
+        # log to sentry if somehow the username isn't in the email list
+        # This is for data gathering purposes, so we can find when this is happening
+        if self.username.lower() not in self.emails:
```
sloria: This doesn't seem necessary, since you're already ensuring that this will be `False` ~L848.

framework/auth/core.py
```diff
@@ -844,6 +844,9 @@ def confirm_email(self, token, merge=False):
         if email not in self.emails:
             self.emails.append(email)
 
+        if self.username not in self.emails:
```
sloria: Would it make sense to do this in `User#save` instead of here?

sloria:
Pass finished.

sloria:
Closing this, as the issue has been resolved (see comments in https://openscience.atlassian.net/browse/OSF-5462).

PR #4841, opened 2016-01-04
Feature/remove forks and registrations from anonymized nodes

brianjgeiger:
Purpose
-----------
For https://openscience.atlassian.net/browse/OSF-4501

Registrations and forks are leaky for anonymized data, and were recently removed from the project view on the OSF. This matches that removal. Also, node_logs params were super-leaky, so we're applying strict anonymization to them as well while we're at it.

Note: This will not prevent people from going to the node's registration list URL. We can talk about if we really want to go that far, but as of the current implementation, it's just making it harder to do. But there are more ways for people to find anonymous data if they *really* want to, such as searching for registrations with the same name or similar, so if we want to get serious about this, we may have to re-think our approach.

Changes
------------
* Restrict fields in the node serializer so that registration links aren't serialized and, consequently, you can't embed registrations. When forks are added to the node serializer, we'll want to avoid whitelisting the forks field;
* Embed meta information to the API when data is anonymized (there is a meta dictionary at the same level as the top-level data dictionary. The meta dictionary will say `anonymous: true` if data is anonymized, otherwise it will not appear);
* Update logs so that nobody can accidentally add leaky information. This is done by restricting log information to id, date, and the type of action; no params or relationship links will show.

Side effects
----------------
The upcoming log renderer will need to handle the case of not having much information, and Henrique has been notified.

### DIFF
--- a/api/base/pagination.py
+++ b/api/base/pagination.py
@@ -9,6 +9,7 @@ from rest_framework.response import Response
 from rest_framework.utils.urls import (
     replace_query_param, remove_query_param
 )
+from api.base.serializers import is_anonymized
 
 class JSONAPIPagination(pagination.PageNumberPagination):
     """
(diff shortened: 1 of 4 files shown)


sloria:
:+1:

PR #4844, opened 2016-01-05
Fix anonymized comments

brianjgeiger:
@sloria this fixes the broken test on staging. https://openscience.atlassian.net/browse/OSF-5488

Purpose
-----------
Ensure all use of a relationship to user:user-detail and nodes:node-registrations are prevented from being exposed when anonymized.

Changes
------------
Replaces broken mixin with a blacklist feature.

Side effects
---------------
Shouldn't be any. There's a test to ensure when forks are added, if they are called 'forks', will need to be added to the blacklist.

### DIFF
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -21,6 +21,15 @@ from modularodm import Q
 from modularodm.exceptions import NoResultsFound
 
 
+def hide_view_when_anonymous(view_name):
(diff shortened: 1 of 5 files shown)


nicipfeiffer:
@sloria Should this PR be in the 0.63 milestone and go through QA testing?

PR #4845, opened 2016-01-05
Keep nodes from showing in any of the registration views

brianjgeiger:
Purpose
-----------
Fixes [https://openscience.atlassian.net/browse/OSF-5491]. Ensures /registrations/ views require the node to be a registration rather than just catching it elsewhere

Changes
-------------
* In get_node() for the registration mixin, verify that the node is a registration and raise a NotFound error otherwise.
* Fix the test that was checking for that to the new way of saying the node is not available here
* Add another test for a subview

Side effects
---------------
One (unrelated) test will still fail due to failing test in develop - fixed in #4844.

### DIFF
--- a/api/registrations/views.py
+++ b/api/registrations/views.py
@@ -43,7 +43,7 @@ class RegistrationMixin(NodeMixin):
         )
         # Nodes that are folders/collections are treated as a separate resource, so if the client
         # requests a collection through a node endpoint, we return a 404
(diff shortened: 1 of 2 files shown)


sloria:
Build failure is unrelated to this PR and has since been fixed on develop. I've cherry-picked eb47501 into `release/0.61.0`. :octopus:

PR #4847, opened 2016-01-05
Dashboard redesign component 3

caneruguz:
Revert dashboard name change to avoid test and backwards compatibility errors


sloria:
@caneruguz Please look into the failing tests. This will not autodeploy unless the build succeeds.

PR #4848, opened 2016-01-06
Feature/s3 calling format update

billyhunt:
Jira Issue OSF-5442

Purpose
-----------
Periods in buckets are currently not allowed in Amazon s3 buckets on the OSF, although they are allowed by Amazon s3, because of restrictions in boto.  (for example "mybucket.with.periods") The new version of boto, currently in staging, will allow buckets with periods to be created, but not linked.  This will create a 500 error on the server and disrupts our error messaging.  The workaround is to set calling_format=OrdinaryCallingFormat() in S3Connection.  We already do this in bucket_exists.  (Thanks to @felliott for exploring this and being amazing).

Changes
------------

1.  set calling_format=OrdinaryCallingFormat() in S3Connection
2.  set that periods are allowed in the exception handling.

Things to check
---------------------
1.  Buckets with and without periods should be added and linked.
2.  Check case sensitivity for buckets, which should already be supported but just make sure.

### DIFF
--- a/website/addons/s3/static/s3NodeConfig.js
+++ b/website/addons/s3/static/s3NodeConfig.js
@@ -108,7 +108,7 @@ ViewModel.prototype.selectBucket = function() {
     self.loading(true);
 
     var ret = $.Deferred();
-    if (!isValidBucketName(self.selectedBucket(), false)) {
+    if (!isValidBucketName(self.selectedBucket(), true)) {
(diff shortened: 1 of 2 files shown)


sloria:
@billyhunt Can you please redo this fix, branched off of `release/0.61.0` so that we can get it into today's release?

PR #4850, opened 2016-01-06
Update default error message in function _fangornDropzoneError

TomBaxter:
Title says it all.

# Side Effects

None.

### DIFF
--- a/website/static/js/fangorn.js
+++ b/website/static/js/fangorn.js
@@ -757,8 +757,8 @@ function _fangornDropzoneRemovedFile(treebeard, file, message, xhr) {
  * @param message Error message returned
  * @private
  */
-var DEFAULT_ERROR_MESSAGE = 'Could not upload file. The file may be invalid ' +
-    'or the file folder has been deleted.';
+var DEFAULT_ERROR_MESSAGE = 'Unable to reach the provider, please try again ' +
+    'later. If the problem persists, please contact support@osf.io.';
 function _fangornDropzoneError(treebeard, file, message, xhr) {
     var tb = treebeard;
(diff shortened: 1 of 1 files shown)


felliott:
Hey @TomBaxter,

The distinction between the two messages was intentional.  If WB doesn't respond (`xhr.status === 0`) then that's a problem on our end, and we want the user to notify us.  If WB does respond, but we don't recognize the response or handle it, then it's probably a user error and we want them to think first before retrying.  That said, the original text ('Could not upload file. The file may be invalid or the file folder has been deleted.') could probably be improved.  It might also make sense to distinguish between  50x (our bad) and 40x (your bad) error codes.  50x could get the 'Unable to reach the provider' message, 40x could get 'Could not upload file'.  What do you think?

Cheers,
@felliott

TomBaxter:
@felliott 

  Yes, that makes sense. I didn't realize that this function only get's touched during uploads, I thought the original message might be too specific.

  I like to avoid specific messages if I can't be sure that it's correct. If we aren't sure that the error is "The file may be invalid or the file folder has been deleted." , I wonder if it isn't counter productive to return it.

  Perhaps the default message should be just "Failed to upload file" if we don't get a "message" from the provider append "provider did not send an error message" Something like that.

  Different messages for 4XX and 5XX sounds like a good plan as well, although in the case of 507 it would be user error more then our error. Out of space.

  It might be fine to just leave it as is as well. The reason I opened a PR originally was due to a logic issue where this function was only ever returning the default message. But someone one fixed it in the mean time. At the time I was trouble shooting an issue posted to the waterbutler repo.

Thanks @TomBaxter

sloria:
Closing this for now. Feel free to reopen if it is still relevant.

PR #4851, opened 2016-01-06
[wip] [Feature] s3 bucket exception handling

billyhunt:
Jira Issue OSF-5442

Purpose
-----------
Periods in buckets are currently not allowed in Amazon s3 buckets on the OSF, although they are allowed by Amazon s3, because of restrictions in boto.  (for example "mybucket.with.periods") The new version of boto, currently in staging, will allow buckets with periods to be created, but not linked.  This will create a 500 error on the server and disrupts our error messaging.  The workaround is to set calling_format=OrdinaryCallingFormat() in S3Connection.  We already do this in bucket_exists.  (Thanks to @felliott for exploring this and being amazing).

Changes
------------

1.  set calling_format=OrdinaryCallingFormat() in S3Connection
2.  set that periods are allowed in the exception handling.

Things to check
---------------------
1.  Buckets with and without periods should be added and linked.
2.  Check case sensitivity for buckets, which should already be supported but just make sure.


billyhunt:
Thanks to @felliott for his massive help with this.

felliott:
Why is the revert patch in here?

felliott:
Please remove the `OrdinaryCallingFormat` code from `utils.bucket_exists`, too.

felliott:
Also, either in the commit message or the commit, can you document why we have lax vs. strict checking in the JS, and under what conditions we can get rid of lax checking?  It'll help future code divers.

felliott:
There's a test failure in s3NodeConfig.test.js: 'alerts the user that the S3 addon does not support bucket names containing periods' is failing (expectedly) and needs to be updated.

sloria:
Once @felliott 's review is address, this should be good to merge.

felliott:
Hey Steve,

This isn't going to make the release.  It's gonna require some non-trivial fixes on the WB side in tandem.

sloria:
Ah ok, thanks for the update.

sloria:
@billyhunt @felliott Please resend this fix as a hotfix (or `release/0.62.0` if impacts the codebase in a non-trivial way) when it is ready.

PR #4853, opened 2016-01-06
[RTM][Feature] Update GitHub addon to ExternalAccounts

mfraezz:
Purpose
=======
To support coming APIv2 addon management, we're updating all of our addon models to utilize the ExternalAccount/Provider architecture. 

Changes
=======
* Update GitHub to use `ExternalAccount`
* Update tests/etc accordingly

Side Effects
==========
None

Migration Steps
---------------------
`python -m  scripts.github.migrate_to_external_accounts`

[#OSF-4934]


felliott:
I can't seem to create new repos from the node settings page.  I keep getting 404s when trying to POST to `/api/v1/github/repo/create/`.

felliott:
Please update the callback url in `README.md`.

mfraezz:
Above issues have been addressed. I'm not sure how repo creation was ever working, because the `nid` used to get the `GitHubNodeSettings` object wasn't being passed in. ¯\\\_(ツ)_/¯ 

I applied more restrictive permissions to repo creation, in the process of evaluating other functions to see if they need perm changes too:

>non-authorizers shouldn't be able to change anything about the authorizer's account
>other than CRUD on the repo that the authorizer chose
>-Steve

felliott:
In the migration script, the log messages for `broken_user_or_oauth_settings` and `no_oauth_creds` are reporting the wrong variables.

scripts/github/migrate_to_external_accounts.py
```diff
@@ -0,0 +1,377 @@
...
+                    os.path.join(
+                        Node.load(document['owner']).api_url, 'github', 'hook/'
+                    )
+                ),
+                'content_type': github_settings.HOOK_CONTENT_TYPE,
+                'secret': secret,
+            }
+        )
+    except (GitHubError, ApiError):
+        raise
```
sloria: This doesn't appear to do anything. Remove?

scripts/github/migrate_to_external_accounts.py
```diff
@@ -0,0 +1,377 @@
...
+    assert_in('user', document)
+    assert_in('registration_data', document)
+    assert_in('owner', document)
+    assert_is_not_none(document['owner'])
+    assert_in('user_settings', document)
+    try:
+        assert_in('hook_secret', document)
+    except AssertionError:
+        try:
+            add_hook_to_old_node_settings(document, account)
```
sloria: Could use some logging here to indicate that a request to the GH API is being sent

scripts/github/migrate_to_external_accounts.py
```diff
@@ -0,0 +1,377 @@
...
+        except (GitHubError, ApiError):
+            return False
```
sloria: Could also use logging here.

scripts/github/migrate_to_external_accounts.py
```diff
@@ -0,0 +1,377 @@
...
+            )
+        )
+    if deleted_node_settings:
+        logger.info(
+            "Failed to migrate {0} deleted node settings:\n{1}".format(
+                len(deleted_node_settings), [e for e in deleted_node_settings]
+            )
+        )
+    if nodeless_node_settings:
+        logger.info(
```
sloria: Use `logger.error` for error messages.

scripts/github/migrate_to_external_accounts.py
mfraezz: If it fails here it gets put into `unverifiable_node_settings`, which I've updated to include the `external_account._id` as well.

felliott:
`website.addons.github.tests.test_models.TestUserSettings` doesn't appear to have any tests in it.

mfraezz:
Appearances can be deceiving. It inherits from a [mixin](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/addons/base/testing/models.py#L31) that runs a number of tests that are generalized for `<Addon>UserSettings` objects that subclass `AddonOAuthUserSettingsBase`.

The only thing that might need a test there is the `public_id` function defined in `website.addons.github.model.GitHubUserSettings`, so I'll add a test for that.

felliott:
Derp. Of course.  Sorry for the noise.

Re: the `set_privacy` method that wasn't used anywhere.  There are tests for it in `website.addons.github.tests.test_models.TestCallbacks` and a conf key for it in `settings.defaults.py`.  I guess this was intended to be a feature, but never got enabled.  Do you know if anyone is intending to use this? If not, I think we should delete the tests and the settings key, too.

felliott:
:+1:

felliott:
`website.addons.github.tests.test_models`: the comment at the end can be deleted, but it would be good to have tests for `GitHubNodeSettings.to_json(user)` where `user == owner` and `user != owner`.  The `test_to_json` in the mixin isn't very thorough.

felliott:
@mfraezz, LGTM!  Please ping me when you've removed the `set_privacy` tests, and added the node settings tests.

felliott:
Love it!  I'm sending it on.

sloria:
Verified the migration works on both staging and prod data. :ship: :it:

PR #4854, opened 2016-01-07
[Feature] Check for write access to allow wiki enable/disable

billyhunt:
Fix bug where users with read access see wiki enable button.


billyhunt:
Thanks to @jinluyuan for finding this.

PR #4856, opened 2016-01-07
Feature: Allow users to comment on files

samanehsan:

Changes
------------
<b>OSF Changes:</b>

<b>API Changes</b>: 
- File serializer:

Post-Merge
---------------
Run `scripts/update_comments.py` to update format of `User.comments_viewed_timestamp`


sloria:
There are still multiple requests being sent when re-expanding comments:

comment_requests2

sloria:
The "Delete" tooltip does not word-break correctly:

osf___comments

Unless the tooltips were in the spec or @erinmayhood objects, I think we can get rid of the tooltips. The functions of the buttons are already clear.

samanehsan:
@sloria, it looks like one is an `OPTIONS` request and the second is the `GET` request. Is there a way to prevent the `OPTIONS` request from getting sent?

sloria:
The preflight OPTIONS request is necessary. However, once the user has requested replies for a comment, they shouldn't need to request them again.

samanehsan:
Ok, I see what you mean

sloria:
That said, the above change is subject to the "5-minute" rule. I see that the undesired behavior is also on production.

sloria:
osf_____foo_txt

1. The "+" on the deleted comment is a different color from a non-deleted comment. They should both be black.
2. The cursor for the "+" on a deleted comment should be a pointer.
3. The position for "+" buttons should be the same for both deleted and non-deleted comments, i.e. both on the right side.

These are minor and can go in a follow-up PR.

website/addons/base/views.py
```diff
@@ -568,6 +568,9 @@ def addon_view_file(auth, node, file_node, version):
         'size': version.size if version.size is not None else 9966699,
         'private': getattr(node.get_addon(file_node.provider), 'is_private', False),
         'file_tags': [tag._id for tag in file_node.tags],
+        'file_id': file_node._id,
+        # Note: Comments are allowed for all addons that support waterbutler
+        'allow_comment': True
```
sloria: Will this ever be `False`?

samanehsan: It used to be `False` for dataverse, but then it got switched over to use waterbutler. If all future storage addons will use waterbutler, then I think we can get rid of this.

sloria: OK nice.

sloria:
I'm finishing my pass here. No blockers above. I'll merge and deploy to staging later this eve or first thing in the morning. If you can address the comments above, great; otherwise, they can be in a follow-up PR.

samanehsan:
I've addressed everything except for removing the tooltips, removing the `allow_comment` flag and UI fixes to the toggle replies button. @sloria, I can make these improvements in a follow-up PR or in this one. :tada:

sloria:
Merged! :octopus:  Will deploy to staging in the morning.

PR #4858, opened 2016-01-07
[Hotfix] Add File Download functionality for Prereg-admins

cwisecarver:
This PR modifies the `website.addons.views.check_access` method allowing users with the `settings.PREREG_ADMIN_TAG` system_tag to **download** files associated with DraftRegistrations of the Prereg Challenge Metaschema that they otherwise would not have permission to download.

It also creates a route and view that redirects to the proper waterbutler URL for a referenced file.

### DIFF
--- a/admin/base/urls.py
+++ b/admin/base/urls.py
@@ -10,6 +10,7 @@ base_pattern = '^{}'.format(ADMIN_BASE)
 
 urlpatterns = [
     ### ADMIN ###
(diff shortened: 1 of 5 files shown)


sloria:
Merged into `master`. `feature/admin-app-prod` is no longer used.

cwisecarver:
Belongs to this JIRA issue: https://openscience.atlassian.net/browse/ADM-50

PR #4859, opened 2016-01-07
[Feature] Admin Spam List

Ghalko:
Originally #4637 

Status: Code reviewed.


Ghalko:
@sloria I replied to your comment in #4637 about the possibility of reports being `None`.

I don't think this is a possibility, but I do not have access to the prod db to really check this out. If there are instances, then we should consider a simple migration.

sloria:
If you're uncertain if `reports` can be `None`, better to code defensively and handle that case.

admin/spam/serializers.py
```diff
@@ -0,0 +1,45 @@
...
+        category=comment['reports'][0]['category'],
+    )
+    return comment
+
+
+def retrieve_comment(id, full_user=False):
+    return serialize_comment(Comment.load(id), full=full_user)
+
+
+def serialize_comments():
```
sloria: This is inconsistent with the other serialization functions because it queries for the objects it is serializing (rather than taking the objects as input). Better to make this consistent with the other functions.

admin/spam/views.py
```diff
@@ -1,9 +1,42 @@
...
+from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
+from django.contrib.auth.decorators import login_required
 
+from .serializers import serialize_comments, retrieve_comment
 
+
+@login_required
 def spam_list(request):
-    return HttpResponse('This is a list of spam')
+    comment_list = serialize_comments()
```
sloria: You are loading all comments that have a report here, even if the client only requests a single page. You should only need to load the records that the clients asks for. You should be able to use the `limit` and `offset` functionality of modm to achieve this.

admin/spam/views.py
```diff
@@ -1,9 +1,42 @@
...
-    return HttpResponse('Looking at spam {}'.format(spam_id))
+    comment = retrieve_comment(spam_id)
+    context = {'comment': comment}
+    return render(request, 'spam/comment.html', context)
+
+
+@login_required
+def spam_sub_list(request, spam_ids):
+    comments = None
+    context = {'comments': comments}
```
sloria: This doesn't appear correct; `comments` is always `None` here. Is this a stub?

admin/templates/spam/email.html
```diff
@@ -0,0 +1,76 @@
+{% extends "base.html" %}
+{% load static %}
+{% block title %}
+    <title>Email</title>
+{% endblock title %}
+{% block content %}
+    <div class="container-fluid">
+        <div class="row">
+            <div class="btn-group col-sm-12">
+                <a href="javascript:window.history.go(-2);"
```
sloria: Will this always work? Is it possible for the list page to be more than two items back?

Ghalko: It is possible. I've experimented with a few different options but I couldn't do it without using the url. Any good resources?

admin/spam/serializers.py
```diff
@@ -0,0 +1,45 @@
...
+        'date_created': comment.date_created,
+        'date_modified': comment.date_modified,
+        'content': comment.content,
+        'has_children': bool(getattr(comment, 'commented', [])),
+        'modified': comment.modified,
+        'is_deleted': comment.is_deleted,
+        'reports': serialize_reports(comment.reports),
+        'node': comment.node,
+    }
+    comment.update(
```
sloria: This `.update` is unnecessary; just add `category` to the dict.

sloria:
Pass finished.

Ghalko:
Thanks! If you know some resources for passing variables in django without using the URL I could really use that info, or discuss why it would be okay, (like staff are the only ones seeing this.)

Otherwise I will get right to this.

sloria:
I'm not familiar with how you're doing routing in the admin app; are you're doing client-side routing or are you doing normal multi-page app/Django routing? It appears that you are doing the latter. If that's the case, then just use normal `a` tags with the links to the right pages. You can use `reverse` to avoid hard-coding the URLs.

admin/spam/serializers.py
Ghalko: Changed.

admin/spam/views.py
Ghalko: @sloria I switched this to resemble more of the prereg with only serializing the current items on the page. Would you like this to go further and actually implement `limit` and `offset` on the database? The pagination on the Django side will not be able to use the available solutions for Django pagination.

admin/spam/views.py
Ghalko: This is a stub, removing.

admin/spam/serializers.py
Ghalko: Done.

Ghalko:
Submitting for the day. Most of the first pass issues are taken care of. Going to work on URL stuff again tomorrow to get functioning buttons.

sloria:
Looks good!

PR #4860, opened 2016-01-07
[Feature] Admin Spam Email

Ghalko:
Originally #4733 

## Set-up
See #4992, hopefully merged before this.


admin/spam/forms.py
```diff
@@ -0,0 +1,15 @@
...
+class EmailForm(forms.Form):
+    author = forms.CharField(label='Author', max_length=100)
+    email = forms.ChoiceField(label='Email')
+    subject = forms.CharField(label='Subject', required=True)
+    message = forms.CharField(label='Message', required=True,
+                              widget=CKEditorWidget())
+
+    def __init__(self, *args, **kwargs):
+        choices = kwargs.get('initial', {}).get('email', [])
+        self.base_fields['email'] = forms.ChoiceField(choices=choices)
```
idwtsiwgh: You can set the choices kwarg to a callable so you could do:
```python
class EmailForm(forms.Form):
    def _get_email_choices(self):
        return self.initial.get('email', [])
    email = forms.ChoiceField(label='Email', choices=self._get_email_choices)
```
https://github.com/django/django/blob/master/django/forms/forms.py#L81 says \__init__ sets self.initial to kwargs[initial].

Ghalko: `self` in the last line makes no sense since self is not defined outside of functions.

cwisecarver: You're right. I can't find another way to do this.

admin/spam/views.py
```diff
@@ -1,9 +1,117 @@
...
+            return self.form_valid(form)
+        else:
+            return self.form_invalid(form)
+
+    def set_initial(self, spam):
+        self.initial = {
+            'author': spam['author'].fullname,
+            'email': [(r, r) for r in spam['author'].emails],
+            'subject': 'Reports of spam',
+            'message': render_to_response(
```
cwisecarver: render_to_response is likely to be deprecated: https://docs.djangoproject.com/en/1.9/topics/http/shortcuts/#render-to-response

admin/spam/views.py
```diff
@@ -1,9 +1,117 @@
...
+                'spam/email_template.html',
+                {'item': spam}
+            ).content,
+        }
+
+    def form_valid(self, form):
+        send_mail(
+            subject=form.cleaned_data.get('subject').strip(),
+            message=form.cleaned_data.get('message'),
+            from_email='support@osf.io',
```
cwisecarver: there's a SUPPORT_EMAIL setting in the osf settings that contains this. Maybe use that instead of hard-coding?

admin/spam/views.py
```diff
@@ -1,9 +1,117 @@
...
+    def success_url(self):
+        return reverse('spam:detail', kwargs={'spam_id': self.spam_id}) + '?page={}'.format(self.page)
+
+
+def get_spam_list():
+    query = (
+        Q('reports', 'ne', {}) &
+        Q('reports', 'ne', None)
+    )
+    return sorted(
```
cwisecarver: Since `date_created` is a field in mongo you should be able to get mongo to sort the data instead of relying on python.
**Descending**
```python
return Comment.find(query).sort('-date_created')
```
**Ascending**
```python
return Comment.find(query).sort('date_created')
```

admin/spam/views.py
```diff
@@ -1,9 +1,117 @@
...
+        Comment.find(query),
+        key=operator.attrgetter('date_created')
+    )
+
+
+@login_required
 def spam_list(request):
-    return HttpResponse('This is a list of spam')
+    paginator = Paginator(get_spam_list(), 10)
```
cwisecarver: Seems weird to jump back and forth between CBV and FBV. Nothing technically wrong, just a style thing.

admin/spam/views.py
```diff
@@ -1,9 +1,117 @@
...
+
+    page_number = request.GET.get('page', 1)
+    try:
+        page = paginator.page(page_number)
+    except PageNotAnInteger:
+        page = paginator.page(1)
+    except EmptyPage:
+        page = paginator.page(paginator.num_pages)
+    context = {
```
cwisecarver: You could change this to

```python
page.object_list = map(serialize_comment, page.object_list)
context = {
    'page': page,
    'page_number': page_number,
}
```
And then loop through page.object_list in your template.

cwisecarver:
Nothing major @Ghalko. I'm done.

admin/spam/views.py
Ghalko: #4859

admin/spam/views.py
Ghalko: #4859

admin/spam/views.py
Ghalko: Replaced with `render(None, ...`

admin/spam/views.py
Ghalko: Done

admin/spam/views.py
Ghalko: Basically the experiment came after #4859 which means that the FBV can certainly be upgraded to CBV in a subsequent merge.

admin/spam/views.py
Ghalko: There is a tradeoff between overwriting the `page.object_list` to conserve the number of bytes sent and writing to a new variable which can be less confusing to read later. @sloria opinion?

admin/spam/views.py
idwtsiwgh: Sounds good.

admin/spam/views.py
sloria: It feels a little strange to overwrite `object_list`; seems like it'd violate the "least surprise" principle. That said, I have no strong preference.

Ghalko: I am going to leave this one alone.

Ghalko:
@cwisecarver Finished with response to CR. Double check?

admin/spam/views.py
cwisecarver: Sounds good.

cwisecarver:
@Ghalko Double-checked. Good to go.

admin/spam/views.py
```diff
@@ -1,9 +1,119 @@
...
+        }
+        return render(request, self.template_name, context)
+
+    def post(self, request, *args, **kwargs):
+        spam_id = kwargs.get('spam_id', None)
+        self.spam_id = spam_id
+        self.page = request.GET.get('page', 1)
+        try:
+            spam = serialize_comment(Comment.load(spam_id))
+        except:
```
sloria: Avoid naked `excepts`. What error is raised here and how should it be handled?

admin/spam/views.py
```diff
@@ -1,9 +1,119 @@
...
+    spam_id = None
+    page = 1
+
+    def get(self, request, *args, **kwargs):
+        spam_id = kwargs.get('spam_id', None)
+        self.spam_id = spam_id
+        self.page = request.GET.get('page', 1)
+        try:
+            spam = serialize_comment(Comment.load(spam_id))
+        except:
```
sloria: See https://github.com/CenterForOpenScience/osf.io/pull/4860/files#r50708474

admin/spam/views.py
```diff
@@ -1,9 +1,119 @@
...
+            pass
+        self.set_initial(spam)
+        form = self.get_form()
+        if form.is_valid():
+            return self.form_valid(form)
+        else:
+            return self.form_invalid(form)
+
+    def set_initial(self, spam):
```
sloria: Would it make more sense to override `FormMixin#get_initial` instead of using this setter?

admin/spam/views.py
```diff
@@ -1,9 +1,119 @@
...
+            'subject': 'Reports of spam',
+            'message': render(
+                None,
+                'spam/email_template.html',
+                {'item': spam}
+            ).content,
+        }
+
+    def form_valid(self, form):
+        send_mail(
```
sloria: What SMTP server is used to send mail? What configuration is necessary to get this to work with our existing Sendgrid setup?

admin/spam/views.py
```diff
@@ -1,9 +1,119 @@
...
+    def post(self, request, *args, **kwargs):
+        spam_id = kwargs.get('spam_id', None)
+        self.spam_id = spam_id
+        self.page = request.GET.get('page', 1)
+        try:
+            spam = serialize_comment(Comment.load(spam_id))
+        except:
+            pass
+        self.set_initial(spam)
+        form = self.get_form()
```
sloria: Can you call `super(..)` instead here?

sloria:
Pass finished. Biggest concern is getting admin's emails working on production. Please include any necessary configuration and deployment steps in the PR description.

admin/spam/views.py
Ghalko: Zak was working on this before break. Not 100% sure where he is on the Emails but we can certainly check tomorrow. Ref: #4718

admin/spam/views.py
Ghalko: Added `AttributeError` when MODM loads `None` if it can't find it.

admin/spam/views.py
Ghalko: Added same as above reference.

admin/spam/views.py
Ghalko: Overwrote `get_initial` had to move spam in as a class variable to get access to it when `get_initial` is called by `get_form_kwargs` in normal order of operations.

admin/spam/views.py
Ghalko: Yep, done.

Ghalko:
@sloria #4992 *should* take care of the set-up concerns.

Ghalko:
@sloria merge conflicts removed.

PR #4864, opened 2016-01-08
Collection Linked Nodes Relationship endpoint and Relationship API groundwork

hmoco:
## Purpose
Lay down the groundwork for relationship endpoints as per JSON API specs and, more specifically, add the ` /collections/<collection_id>/relationships/linked_nodes/ ` endpoint to allow the creation, retrieval, update and destruction if node links in a collection. This is done in this manner:

### Create
```
        Method:        POST
        URL:           /links/self
        Query Params:  <none>
        Body (JSON):   {
                         "data": [{
                           "type": "linked_nodes",   # required
                           "id": <node_id>   # required
                         }]
                       }
        Success:       201
```
This requires both admin permission on the collection, and for the user that is
making the request to be able to read the nodes requested. Data can be contain any number of
node identifiers. This will create a node_link for all node_ids in the request that
do not currently have a corresponding node_link in this collection.

### Update
```
        Method:        PUT || PATCH
        URL:           /links/self
        Query Params:  <none>
        Body (JSON):   {
                         "data": [{
                           "type": "linked_nodes",   # required
                           "id": <node_id>   # required
                         }]
                       }
        Success:       200
```
This requires both admin permission on the collection, and for the user that is
making the request to be able to read the nodes requested. Data can be contain any number of
node identifiers. This will replace the contents of the node_links for this collection with
the contents of the request. It will delete all node links that don't have a node_id in the data
array, create node links for the node_ids that don't currently have a node id, and do nothing
for node_ids that already have a corresponding node_link. This means a update request with
{"data": []} will remove all node_links in this collection

### Destroy
```
        Method:        DELETE
        URL:           /links/self
        Query Params:  <none>
        Body (JSON):   {
                         "data": [{
                           "type": "linked_nodes",   # required
                           "id": <node_id>   # required
                         }]
                       }
        Success:       204
```
This requires admin permission on the node. This will delete any node_links that have a
corresponding node_id in the request.

This PR also adds tests for the intended behavior of this relationship endpoint.

## Changes
The addition of the endpoint, and base classes for one-to-many relationships

## Side Effects
None? (easier to interact with node_links via api)

### DIFF
--- a/api/base/parsers.py
+++ b/api/base/parsers.py
@@ -131,19 +131,18 @@ class JSONAPIRelationshipParser(JSONParser):
 
         if not isinstance(res, dict):
             raise ParseError('Request body must be dictionary')
-        datas = res.get('data')
+        data = res.get('data')
 
-        if datas:
-            for data in datas:
-                id_ = data.get('id')
-                type_ = data.get('type')
+        if data:
+            for value in data:
 
-                if id_ is None:
+                if value.get('id') is None:
                     raise JSONAPIException(source={'pointer': '/data/id'}, detail=NO_ID_ERROR)
 
-                if type_ is None:
+                if value.get('type') is None:
                     raise JSONAPIException(source={'pointer': '/data/type'}, detail=NO_TYPE_ERROR)
-            return {'data': datas}
+
+            return {'data': data}
 
         return {'data': []}
 
--- a/api/base/parsers.py
+++ b/api/base/parsers.py
@@ -119,3 +119,36 @@ class JSONAPIParserForRegularJSON(JSONAPIParser):
     Allows same processing as JSONAPIParser to occur for requests with application/json media type.
     """
     media_type = 'application/json'
+
+class JSONAPIRelationshipParser(JSONParser):
+    """
+    Parses JSON-serialized data for relationship endpoints. Overrides media_type.
+    """
+    media_type = 'application/vnd.api+json'
+
+    def parse(self, stream, media_type=None, parser_context=None):
+        res = super(JSONAPIRelationshipParser, self).parse(stream, media_type, parser_context)
+
+        if not isinstance(res, dict):
+            raise ParseError('Request body must be dictionary')
+        data = res.get('data')
+
+        if data:
+            id_ = data.get('id')
+            type_ = data.get('type')
+
+            if id_ is None:
+                raise JSONAPIException(source={'pointer': '/data/id'}, detail=NO_ID_ERROR)
+
+            if type_ is None:
+                raise JSONAPIException(source={'pointer': '/data/type'}, detail=NO_TYPE_ERROR)
+
+            return data
+
+        return {'type': None, 'id': None}
+
+class JSONAPIRelationshipParserForRegularJSON(JSONAPIRelationshipParser):
+    """
+    Allows same processing as JSONAPIRelationshipParser to occur for requests with application/json media type.
+    """
+    media_type = 'application/json'
--- a/api/base/parsers.py
+++ b/api/base/parsers.py
@@ -131,19 +131,19 @@ class JSONAPIRelationshipParser(JSONParser):
 
         if not isinstance(res, dict):
             raise ParseError('Request body must be dictionary')
-        data = res.get('data')
+        datas = res.get('data')
 
-        if data:
-            id_ = data.get('id')
-            type_ = data.get('type')
+        if datas:
+            for data in datas:
+                id_ = data.get('id')
+                type_ = data.get('type')
 
-            if id_ is None:
-                raise JSONAPIException(source={'pointer': '/data/id'}, detail=NO_ID_ERROR)
+                if id_ is None:
+                    raise JSONAPIException(source={'pointer': '/data/id'}, detail=NO_ID_ERROR)
 
-            if type_ is None:
-                raise JSONAPIException(source={'pointer': '/data/type'}, detail=NO_TYPE_ERROR)
-
-            return data
+                if type_ is None:
+                    raise JSONAPIException(source={'pointer': '/data/type'}, detail=NO_TYPE_ERROR)
+            return {'data': datas}
 
         return {'type': None, 'id': None}
 
--- a/api/base/parsers.py
+++ b/api/base/parsers.py
@@ -145,7 +145,7 @@ class JSONAPIRelationshipParser(JSONParser):
                     raise JSONAPIException(source={'pointer': '/data/type'}, detail=NO_TYPE_ERROR)
             return {'data': datas}
 
-        return {'type': None, 'id': None}
+        return {'data': []}
 
 class JSONAPIRelationshipParserForRegularJSON(JSONAPIRelationshipParser):
     """
--- a/api/base/parsers.py
+++ b/api/base/parsers.py
@@ -136,13 +136,13 @@ class JSONAPIRelationshipParser(JSONParser):
         if data:
             if not isinstance(data, list):
                 raise ParseError('Data must be an array')
-            for i in range(0, len(data)):
+            for i, datum in enumerate(data):
 
-                if data[i].get('id') is None:
-                    raise JSONAPIException(source={'pointer': '/data/' + str(i) + '/id'}, detail=NO_ID_ERROR)
+                if datum.get('id') is None:
+                    raise JSONAPIException(source={'pointer': '/data/{}/id'.format(str(i))}, detail=NO_ID_ERROR)
 
-                if data[i].get('type') is None:
-                    raise JSONAPIException(source={'pointer': '/data/' + str(i) + '/type'}, detail=NO_TYPE_ERROR)
+                if datum.get('type') is None:
+                    raise JSONAPIException(source={'pointer': '/data/{}/type'.format(str(i))}, detail=NO_TYPE_ERROR)
 
             return {'data': data}
 
--- a/api/base/parsers.py
+++ b/api/base/parsers.py
@@ -134,13 +134,15 @@ class JSONAPIRelationshipParser(JSONParser):
         data = res.get('data')
 
         if data:
-            for value in data:
+            if not isinstance(data, list):
+                raise ParseError('Data must be an array')
+            for i in range(0, len(data)):
 
-                if value.get('id') is None:
-                    raise JSONAPIException(source={'pointer': '/data/id'}, detail=NO_ID_ERROR)
+                if data[i].get('id') is None:
+                    raise JSONAPIException(source={'pointer': '/data/' + str(i) + '/id'}, detail=NO_ID_ERROR)
 
-                if value.get('type') is None:
-                    raise JSONAPIException(source={'pointer': '/data/type'}, detail=NO_TYPE_ERROR)
+                if data[i].get('type') is None:
+                    raise JSONAPIException(source={'pointer': '/data/' + str(i) + '/type'}, detail=NO_TYPE_ERROR)
 
             return {'data': data}
 
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -816,7 +816,7 @@ class JSONAPIRelationshipsSerializer(ser.Serializer):
         assert type_ is not None, 'Must define Meta.type_'
 
         relation_id_field = self.fields['id']
-
+        import ipdb; ipdb.set_trace()
         attribute = relation_id_field.get_attribute(obj)
         relationship = relation_id_field.to_representation(attribute)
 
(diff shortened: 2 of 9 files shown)


brianjgeiger:
It looks like you haven't added the relationship self link to the collections serializer yet.

brianjgeiger:
I tried to post to the relationship link to add a link to a collection to the list. This failed in the wrong way. It gave a 500 and said that "Pointer to folder 8pnwb already exists. Only one pointer to any given folder allowed". I *believe* collections won't allow a link to another collection by way of giving a 404 (because a collection can't be acquired through a /nodes/ endpoint), which is the desired behavior here.

brianjgeiger:
Don't forget to update the description of the PR to match the COSDev standards and link to the related issue.

hmoco:
@brianjgeiger updated PR, but not sure I was ever assigned the ticket for this task, could you point me to it?

brianjgeiger:
Awesome, thanks. Okay, passing this to @sloria. :tada: :penguin: :tada:

api/base/serializers.py
```diff
@@ -803,6 +803,25 @@ def is_valid(self, clean_html=True, **kwargs):
...
+    Provides a simplified serialization of the relationship, allowing for simple update request
+    bodies.
+    """
+
+    def to_representation(self, obj):
+        meta = getattr(self, 'Meta', None)
+        type_ = getattr(meta, 'type_', None)
+        assert type_ is not None, 'Must define Meta.type_'
+
+        relation_id_field = self.fields['id']
```
sloria: This assumes that the serializer has an `id` field. Would it make sense to put the `id` and `type` fields on this class and get rid of `RelationshipResourceIdObjectSerializer`?

website/project/model.py
```diff
@@ -2317,6 +2317,14 @@ def deep_url(self):
         return '/project/{}/'.format(self._primary_key)
 
     @property
+    def linked_nodes_self_url(self):
```
sloria: These properties are probably unnecessary, but I think they are fine for now.

sloria:
> This requires both admin permission on the collection, and for the user that is
making the request to be able to read the nodes requested. 

Are there tests for creating/updating a collection that the user does not have admin permissions on?

hmoco:
@sloria yes

api/base/serializers.py
hmoco: Yes, I believe that would make sense, will do!

hmoco:
@sloria, well I believe so for creating a collection. I know so for updating a collection via this endpoint, as it cannot create collections.

api/base/parsers.py
```diff
@@ -119,3 +119,35 @@ class JSONAPIParserForRegularJSON(JSONAPIParser):
...
+
+        if not isinstance(res, dict):
+            raise ParseError('Request body must be dictionary')
+        data = res.get('data')
+
+        if data:
+            for value in data:
+
+                if value.get('id') is None:
+                    raise JSONAPIException(source={'pointer': '/data/id'}, detail=NO_ID_ERROR)
```
sloria: The JSON pointers should contain the index of the invalid datum, e.g. `/data/3/id`).

api/base/parsers.py
```diff
@@ -119,3 +119,35 @@ class JSONAPIParserForRegularJSON(JSONAPIParser):
...
+
+    def parse(self, stream, media_type=None, parser_context=None):
+        res = super(JSONAPIRelationshipParser, self).parse(stream, media_type, parser_context)
+
+        if not isinstance(res, dict):
+            raise ParseError('Request body must be dictionary')
+        data = res.get('data')
+
+        if data:
+            for value in data:
```
sloria: What happens if the client passes an invalid type, e.g. `{data: {foo: 'bar'}}?

sloria:
Stopping my review here because I need to debug and issue on prod. Go ahead and resolve my most recent comments, and I will continue when I am done with the issue I'm working on.

api/base/parsers.py
```diff
@@ -119,3 +119,37 @@ class JSONAPIParserForRegularJSON(JSONAPIParser):
...
+        if not isinstance(res, dict):
+            raise ParseError('Request body must be dictionary')
+        data = res.get('data')
+
+        if data:
+            if not isinstance(data, list):
+                raise ParseError('Data must be an array')
+            for i in range(0, len(data)):
+
+                if data[i].get('id') is None:
```
sloria: Better to use `enumerate` here.

api/collections/serializers.py
```diff
@@ -94,3 +96,72 @@ def get_absolute_url(self, obj):
...
+        type_ = 'linked_nodes'
+
+    def get_pointers_to_add_remove(self, pointers, new_pointers):
+        diff = relationship_diff(
+            current_items={pointer.node._id: pointer for pointer in pointers},
+            new_items={val['node']['_id']: val for val in new_pointers}
+        )
+
+        nodes_to_add = []
+        for node_id in diff['add'].keys():
```
sloria: `.keys()` is unnecessary.

api/collections/serializers.py
```diff
@@ -94,3 +96,72 @@ def get_absolute_url(self, obj):
...
+        ], 'self': obj}
+
+    def update(self, instance, validated_data):
+        collection = instance['self']
+        auth = get_user_auth(self.context['request'])
+
+        add, remove = self.get_pointers_to_add_remove(pointers=instance['data'], new_pointers=validated_data['data'])
+
+        for pointer in remove:
+            collection.rm_pointer(pointer, auth)
```
sloria: Do you expect an exception to be raised here (if the client sends invalid data)? Is it possible that the pointer isn't in the collection?

api/collections/serializers.py
```diff
@@ -94,3 +96,72 @@ def get_absolute_url(self, obj):
...
+        for node in add:
+            collection.add_pointer(node, auth)
```
sloria: Same here: could an exception be raised here? What if the node is already in the collection?

api/collections/serializers.py
```diff
@@ -94,3 +96,72 @@ def get_absolute_url(self, obj):
...
+
+    def create(self, validated_data):
+        instance = self.context['view'].get_object()
+        auth = get_user_auth(self.context['request'])
+        collection = instance['self']
+
+        add, remove = self.get_pointers_to_add_remove(pointers=instance['data'], new_pointers=validated_data['data'])
+
+        for node in add:
+            collection.add_pointer(node, auth)
```
sloria: See https://github.com/CenterForOpenScience/osf.io/pull/4864/files#r50126289

api/collections/views.py
```diff
@@ -533,3 +536,102 @@ def perform_destroy(self, instance):
...
+
+    def perform_destroy(self, instance):
+        data = self.request.data['data']
+        user = self.request.user
+        auth = Auth(user)
+        current_pointers = {pointer.node._id: pointer for pointer in instance['data']}
+        collection = instance['self']
+        for val in data:
+            if val['id'] in current_pointers.keys():
+                collection.rm_pointer(current_pointers[val['id']], auth)
```
sloria: See previous comment: could an exception be raised here?

api/collections/views.py
```diff
@@ -533,3 +536,102 @@ def perform_destroy(self, instance):
...
+        return obj
+
+    def perform_destroy(self, instance):
+        data = self.request.data['data']
+        user = self.request.user
+        auth = Auth(user)
+        current_pointers = {pointer.node._id: pointer for pointer in instance['data']}
+        collection = instance['self']
+        for val in data:
+            if val['id'] in current_pointers.keys():
```
sloria: `.keys()` is unnecessary.

api/nodes/permissions.py
```diff
@@ -74,6 +75,32 @@ def has_object_permission(self, request, view, obj):
...
+
+    def has_object_permission(self, request, view, obj):
+        assert isinstance(obj, dict)
+        auth = get_user_auth(request)
+        parent_node = obj['self']
+
+        if request.method in permissions.SAFE_METHODS:
+            return parent_node.can_view(auth)
+        elif request.method == 'DELETE':
+            return parent_node.can_edit(auth)
```
sloria: The docstring says that this requires admin permissions, but you are only checking read+write here; can you please verify this?

api/nodes/permissions.py
```diff
@@ -74,6 +75,32 @@ def has_object_permission(self, request, view, obj):
...
+        else:
+            pointer_nodes = []
+            for pointer in request.data.get('data', []):
+                node = Node.load(pointer['id'])
+                if not node or node.is_folder:
+                    raise exceptions.NotFound
+                pointer_nodes.append(node)
+            has_parent_auth = parent_node.can_edit(auth)
```
sloria: You should be able to bail out early (i.e. return `False`) if this is `False`.

But again, is this supposed to check admin permissions rather than r+w only?

sloria:
Pass finished.

api/collections/serializers.py
hmoco: I don;t belive so. The pointers to remove are pointers currently in the collection that are not in the request body. Also the function ` get_pointers_to_add_remove ` checks to make sure the client has sent proper Node ids.

api/collections/serializers.py
hmoco: ` relationship_diff ` should take of this. It only returns items that are not already in the collection.

api/collections/views.py
hmoco: I don't believe so, ` rm_pointer ` only raises an error if the pointer being passed is not one of the current pointers in the collection, which is checked before calling ` rm_pointer `.

api/nodes/permissions.py
hmoco: The docstrings are incorrect, will update.

hmoco:
Responded, back to @sloria

api/collections/serializers.py
```diff
@@ -94,3 +96,72 @@ def get_absolute_url(self, obj):
...
+        diff = relationship_diff(
+            current_items={pointer.node._id: pointer for pointer in pointers},
+            new_items={val['node']['_id']: val for val in new_pointers}
+        )
+
+        nodes_to_add = []
+        for node_id in diff['add']:
+            node = Node.load(node_id)
+            if not node:
+                raise exceptions.NotFound
```
sloria: This should have a descriptive error message so that the client knows which node can't be found.

api/collections/serializers.py
sloria: OK, it would be good to have tests verifying this. I don't, for example, see a test for the client passing a node ID that is already part of the collection (should return a 400).

api/nodes/permissions.py
```diff
@@ -74,6 +75,34 @@ def has_object_permission(self, request, view, obj):
...
+        elif request.method == 'DELETE':
+            return parent_node.can_edit(auth)
+        else:
+            has_parent_auth = parent_node.can_edit(auth)
+            if not has_parent_auth:
+                return False
+            pointer_nodes = []
+            for pointer in request.data.get('data', []):
+                node = Node.load(pointer['id'])
+                if not node or node.is_folder:
```
sloria: These 2 cases might need to be handled separately; 404 if the node can't be found and 400 if the node isn't a folder. Yes, @brianjgeiger ?

api/nodes/permissions.py
```diff
@@ -74,6 +75,34 @@ def has_object_permission(self, request, view, obj):
...
+                    raise exceptions.NotFound
```
sloria: Add a useful error message so the client knows what can't be found.

sloria:
Pass finished.

api/nodes/permissions.py
brianjgeiger: No, if you try to load a non-folder node through the collections endpoint, it should 404. Like loading a user through the nodes endpoint, the fact that a collection is a node is an implementation detail that api users don't need to know about.

brianjgeiger: Oh, speaking of, when this is merged with dashboard_autodeploy, is_folder is changing to is_collection.

hmoco:
@sloria should be good now!

sloria:
Restarted the build; Travis looks to be having a period of slow performance, so the build may take longer than usual.

api/nodes/permissions.py
sloria: OK, thanks for the heads up Brian

sloria:
Bleh, travis failed again on installing requirements. Restarted again. 

In the meantime, can you address https://github.com/CenterForOpenScience/osf.io/pull/4864/files#r50182798 ? I don't see it as a blocker--it can go in a follow-up PR if Travis starts cooperating and allows this to get merged--but it would be good to have those tests before this feature is released in 0.63.0.

sloria:
Travis is green! In the interest of moving this forward for dashboard work, I'm going to go ahead and merge this. I've created a ticket for the additional tests here: https://openscience.atlassian.net/browse/OSF-5579 .

PR #4799, opened 2015-12-21
[Feature] Remove Contributors Multiple Components

billyhunt:
### Purpose

Give users the choice to remove a user from either the current node or from the current node and all its children.

#### Current:
screen shot 2015-11-16 at 10 26 31 am

#### Proposed:

screen shot 2015-11-16 at 10 53 50 am

screen shot 2015-11-16 at 10 54 25 am copy

### Changes:

* Delete the user from the given node as a default with a modal.
* If the Admin wants to remove the Contributor from the node AND its children, Open a second modal view, and use the API V2 to list of every component associated with the Contributor to remove.
*  If the Admin clicks "Remove" on the second view, use the API v2 to delete those contributors from all child nodes previously listed.

PR #4869, opened 2016-01-08
fix maintenance-mode favicon

whit537:
Iiiiiiiit's currently [an HTML file](https://github.com/CenterForOpenScience/osf.io/blob/0d5d5ef459a12c197d8040df03790237f8ab71f2/website/static/maintenance/favicon.ico). :flushed:


whit537:
Discovered via `ack css` from https://github.com/CenterForOpenScience/osf.io/pull/4799/files#r49213086.

sloria:
Hm, interesting. Good catch and thanks for the PR!

PR #4871, opened 2016-01-11
Feature/api/projects from templates

brianjgeiger:
Purpose
-----------
Closes https://openscience.atlassian.net/browse/OSF-5510

Allow API v2 to create projects from templates. This is not the best documented feature on the OSF, but I think I have it working at least well enough to match what we're doing on the onboarder within the API.

Changes
-------------
* Add API parameter to choose a node to template from on POST
* Unit tests to ensure that basic functionality works
* Documentation for the new parameter

Side effects and errata
----------------
Technically the templating allows for complex modification of the templated project during the creation, but that doesn't really work well with how the API is set up. Also, I'm not convinced anyone uses all the power available. So for now it's a minimal change to the API. That being said, it shouldn't break anything in the OSF nor on the API.

### DIFF
--- a/api/nodes/serializers.py
+++ b/api/nodes/serializers.py
@@ -200,7 +200,7 @@ class NodeSerializer(JSONAPISerializer):
             template_node = Node.load(key=template_from)
             validated_data.pop('creator')
             changed_data = {template_from: validated_data}
-            node=template_node.use_as_template(auth=self.get_user_auth(self.context['request']), changes=changed_data)
+            node = template_node.use_as_template(auth=self.get_user_auth(self.context['request']), changes=changed_data)
         else:
             node = Node(**validated_data)
         try:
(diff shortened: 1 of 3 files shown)


api/nodes/serializers.py
```diff
@@ -83,6 +83,13 @@ class NodeSerializer(JSONAPISerializer):
     collection = DevOnly(ser.BooleanField(read_only=True, source='is_folder'))
     dashboard = ser.BooleanField(read_only=True, source='is_dashboard')
     tags = JSONAPIListField(child=NodeTagField(), required=False)
+    template_from = ser.CharField(required=False, allow_blank=False, allow_null=False,
+                                  help_text='Specify a node id for a node you would like to use as a template for the '
+                                            'new node. Templating is like forking, except that you do not copy the '
```
sloria: :+1:

PR #4875, opened 2016-01-11
[hotfix] OSF4M: Remove "All Submissions" for now

sloria:
The endpoint is too inefficient to be used in production at the
moment; requests for All Submissions are timing out. This removes
the UI and the route for viewing All Submissions

see OSF-4756


sloria:
I reverted this in c93b20f . Note to self: the "all submissions" endpoint is necessary for updating the submission counts.

PR #4886, opened 2016-01-12
Feature/api/current user node permissions

brianjgeiger:
Purpose
-----------
Create a node serializer field that tells you what the current user's permissions are for that node. If the user is not logged in, the permissions are ["read"]. If they don't have read permissions, they won't be able to serialize the node.

Closes https://openscience.atlassian.net/browse/OSF-5546

Changes
------------
* New serializer field current_user_permissions
* Tests
* Test utility to compare two lists and ensure they are the same. There was an assertion that seemed similar, but the docstring suggested it only did length comparison.

Side effects
---------------
None

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -5,7 +5,7 @@ import datetime
 
 from django.http import HttpRequest
 from nose import SkipTest
-from nose.tools import assert_equal, assert_not_equal
+from nose.tools import assert_equal, assert_not_equal, assert_in
 
 from framework.auth import Auth
 from website.archiver import ARCHIVER_SUCCESS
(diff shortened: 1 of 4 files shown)


tests/utils.py
```diff
@@ -178,3 +178,15 @@ def wrapper():
         used.append(item)
         return item
     return wrapper
+
+
+def assert_mutual_in(list_one, list_two):
```
sloria: There is already `assert_items_equal` for this: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertItemsEqual .

I can change this when I merge.

brianjgeiger: Cool, thanks. I looked through, but had a hard time finding something that was quite right.

PR #4887, opened 2016-01-13
Project Organizer redesign

caneruguz:
## Purpose
This PR is changing the functionality of the project organizer on the dashboard page by changing the way project organizer functions work and creating a new paradigm for handling project related tasks including collections.

## Changes
- All project related aspects are in fileBrowser.js file 
- A new method of adding projects is used in addProjectPlugin.js
- The way activity logs are presented are changed into a js plugin with logTextParser.js 
- API adjustments have been made for end points to provide the data the new dashboard needs 

## Side Effects
This is a first step PR to get the majority of the items through code review. Some elements are missing
- Removing project from collection is waiting on another API PR
- Prereg adding existing project feature is broken, will be fixed soon
- Some logs do not have text parsing yet
- Dashboard name will be changed to My Project in routes, url, mako and tests

Because this PR is large and the current PR contains 98% of the work it should be merged in to dashboard while other minor work continues. 

This PR relates to these Jira tickets:
https://openscience.atlassian.net/browse/OSF-4797 [OSF-4797]
https://openscience.atlassian.net/browse/OSF-4799 [OSF-4799]
https://openscience.atlassian.net/browse/OSF-4800 [OSF-4800]
https://openscience.atlassian.net/browse/OSF-5155 [OSF-5155]
https://openscience.atlassian.net/browse/OSF-5262 [OSF-5262]
https://openscience.atlassian.net/browse/OSF-5263 [OSF-5263]
https://openscience.atlassian.net/browse/OSF-5408 [OSF-5408]

### Screenshot:
screen shot 2016-01-13 at 9 59 19 am


felliott:
In `website/static/js/project-organizer.js` the `_poContributors` function should have fallback display names.  My main user account doesn't have a family name, so my contributors list looks like:

screen shot 2016-01-14 at 5 42 02 pm

In this example, "Coward" is the family_name of a second contributor.

felliott:
I've encountered several deeply-nested hash references that are causing JS errors:

* `fileBrowser.js` in `ActivityLogs.view`:  `item.embeds.user.data.links.profile_image` is failing because some of my logs do not have a user associated with them.  Instead the `embeds.user` field is:

screen shot 2016-01-15 at 10 03 08 am

* `project-organizer.js` in `_poContributors`:  `item.data.embeds.contributors.data` can sometimes be null because of a bug in the API, where `/users/me/nodes/` is including registrations.  I've confirmed with @brianjgeiger and reported this as [#OSF-5555].

* `project-organizer.js` in `_poResolveToggle`: `item.data.relationships.children.links.related.meta.count` causes the retractions list to fail because retracted registrations do not have a `children` relationship.  I've checked with @pattisdr to confirm that this is the expected behavior.

felliott:
Registration activity logs are not available because of an API bug, reported as [#OSF-5556].

felliott:
It might be helpful to have a spinner or some sort of "in progress" indicator.  The API calls are very slow on my machine, so when I click on a new collection, nothing appears to happen for 5-10 seconds.

felliott:
I get a bunch of console warnings for "There is no text entry in dictionary for the action :`$foo`".  Most of them seem to be related to external provider actions.  I compare the actions in my nodelogs against `_allLogTexts.js` and found the following ones missing.  This is by no means exhaustive, it's just the ones I've happened to log.

````
dropbox_file_added
dropbox_file_removed
dropbox_file_updated
dropbox_folder_created
dropbox_folder_selected
dropbox_node_authorized
figshare_content_linked
figshare_node_authorized
figshare_node_deauthorized
github_file_added
github_file_removed
github_file_updated
github_folder_created
github_node_authorized
github_repo_linked
googledrive_file_added
googledrive_file_removed
googledrive_folder_created
googledrive_folder_selected
googledrive_node_authorized
s3_bucket_linked
s3_file_added
s3_file_removed
s3_file_updated
s3_folder_created
s3_node_authorized
````

**Comment by felliott** [comment]
Unrecognized log actions can cause the page to crash.  `LogText.view` in `logTextParser.js` returns undefined if it doesn't recognize the `logObject.attribute.action`, causing mithril to die.  It should return some sort of `m()` object instead.

When I look at the activity for a node with unrecognized log actions, I see:

screen shot 2016-01-15 at 12 21 53 pm

When I modify `LogText.view` to return a dummy `m()` object, I see:

screen shot 2016-01-15 at 12 17 24 pm

**Comment by felliott** [comment]
Minor: paging for Tags & Contributors initializes to `1/0` instead of `1/1` like Collections:

screen shot 2016-01-21 at 1 40 54 pm

**Comment by felliott** [comment]
Two comments on [d3ab32e](https://github.com/caneruguz/osf.io/commit/d3ab32e6440ed9c385205b7ea3feb2bf09ce56b1):

* `name = 'A Contributor'` is missing a semicolon

* Just below the patch is the following line. I think the first conditional should be returning an empty `m()` to match the returns of the others.

````
        if (index > 2) {
            return;
        }
        if (index === 2) {
            return m('span', ' + ' + (arr.length - 2));
        }
````

**Comment by felliott** [comment]
This is just me being annoying:  `__proto__` is a valid tag, but won't show up because the tags object has a built-in `__proto__` key.

**Comment by felliott** [comment]
`fileBrowser.js:_formatDataforPO`:  with the display name change, it's possible to search for a name that is visible on the screen and get no results. 

Unfiltered results:
screen shot 2016-01-21 at 2 39 22 pm

Searching by "Fotz":
screen shot 2016-01-21 at 2 39 38 pm

This is because Contributors displays the `family_name` or `given_name` of the user, while the filter searches on `full_name`.

**Comment by felliott** [comment]
Also, please ignore the "Invalid date" in my screenshots.  I had to ditch my Node.date_modified fields to CR comments.

**Comment by felliott** [comment]
Not sure if this is intentional: the collections tooltip looks like title text.  I think it used to be bigger?

screen shot 2016-01-21 at 2 46 11 pm

**Comment by felliott** [comment]
Creating a new collection results in the count being '(undefined)':

screen shot 2016-01-21 at 2 49 02 pm

**Comment by felliott** [comment]
Minor: click on the edit collection modal, then click on the collection name.  The collection loads, but the modal doesn't disappear:
screen shot 2016-01-21 at 2 51 00 pm

**Comment by felliott** [comment]
Dragging and dropping a project onto a new collection should (maybe) invoke the spinner.  After the project is added, the collection's project count becomes `NaN`:

screen shot 2016-01-21 at 3 01 15 pm

**Comment by felliott** [comment]
In `fileBrowser.js:_makeTree` it would be good to have a comment on this line explaining that children must register themselves into the parent's child list.  I'm presuming this is because the parents may have entries in the children list that have been filtered out. Maybe something like `// only save children present in flatData`

````
        } else { // If this node is in object list it's likely because it was created as a parent so fill out the rest of the information
            n.children = node_list[n.id].children;
            node_list[n.id] = n;
        }
````

**Comment by felliott** [comment]
When creating a new project, the Project Name input field often drops input characters.  Here is a video of me trying to type the phrase 'EmptyCategory' a bunch of times.  It succeeds on the second time, but no other.  I'm pretty sure I typed it correctly each time.

slow typing

**Comment by felliott** [comment]
Minor interface issue:  After creating a new project, I'm given the option to 'Keep Working Here' or 'Go to new project'.  'Keep Working Here' forces a whole page reload, which is quite slow locally.  I understand why this might need to be done, but there is also a close button in the top-right of the modal that dismisses the modal without a page reload.  If a reload is necessary, should the close button force a reload, too?  I can see arguments for both ways.

**Comment by felliott** [comment]
When I have the 'Activity' tab open and switch between projects, there is a long delay while the new project logs are loaded.  In the meantime, the previous project's activity is shown.  I think either the old activities should be blanked, a spinner displayed, or both. 

slow-activity

The gif is sped up.  The original movie is about 20 seconds.  It actually takes about 8 seconds after clicking for Shub-hub's activities to be displayed.

**Comment by felliott** [comment]
Expanding a project to show its children works in initial display, but stops working when you apply a filter to it.  The console logs an error in `treebeard:rowFilterResult` that this line:

`if (o.filter && item.data[o.data].toString().toLowerCase().indexOf(filter) !== -1) {`

..fails because `item.data[o.data]` is undefined.  `o.data` in this case is `"name"`. It looks like that when a filter is applied, the children are not being passed through `_formatDataforPO`, as they lack the `name`, `uid`, `tags`, etc. attributes.

missing-children

**Comment by felliott** [comment]
`FileBrowser.controller.addTbProperties` and `_formatDataforPO` have a lot in common.  Can addTbProperties call `_formatDataforPO`?

**Comment by felliott** [comment]
Typo in `FileBrowser.controller.updateListSuccess`:  "You have notcreated any projects yet." needs space between "not" and "created".

**Comment by felliott** [comment]
Minor: many of the functions created in `FileBrowser.controller` are anonymous.  Please give them names.

**Comment by felliott** [comment]
In `FileBrowser.controller` there are two variables that are confusing: `loadingNodes` and `loadingPages`.  Please initialize and comment these or rename them.  Perhaps `loadingPages` could be renamed to `loadingNodePages` or something to indicate it runs within `loadingNodes`.

**Comment by felliott** [comment]
If I try to create a new subcomponent, there is a noticeable pause while the subcomponent is being created.  If I click on the Add button during this pause, a second request will be sent with an empty title.  The first request will succeed, but the second request, which causes a 400 error, will override it and the modal will display a "Something went wrong" error.  Is there a way to make the Add button non-interactive while the request is still in progress?

**Comment by felliott** [comment]
Minor: when a component only has 1 subcomponent, the browser shows "1 Projects" instead of "1 Project".
screen shot 2016-01-22 at 5 39 07 pm

**Comment by felliott** [comment]
A collection cannot be created with the name 'Bookmarks', but an existing collection can be renamed to 'Bookmarks'.

screen shot 2016-01-22 at 6 05 39 pm

**Comment by felliott** [comment]
This might be completely valid, but the Add Collection modal does allow me to name my new collection 'All My Projects'.

**Comment by felliott** [comment]
In `fileBrowser.js:Filters.view`, `selectedCSS` is declared outside the two functions where it is used, but I can't see any reason why it shouldn't be declared inside the `for` loops of `returnNameFilters` and `returnTagFilters`.

**Comment by felliott** [comment]
In `addProjectPlugin.js` I never see the `processing` template, even between when I add a project and the success callback is called.

**Comment by felliott** [comment]
In `logTextParser.js:LogText.view` inside the construction of `span.osf-log-item`, the map function could return undef if `pieces` is undefined.  Should it be returning an empty `m()` instead?

**Comment by felliott** [comment]
Trivial: in `logTextParser.js:LogPieces.tag` the `if` statement is only three spaces indented.  What are you trying to do us?  COS has a reputation to maintain! :smile:

**Comment by felliott** [comment]
Also trivial, but possibly serious for some people: `logTextParser.js:returnTextParams` omits the Oxford comma for arrays.

**Comment by felliott** [comment]
Minor: in `logTextParser.js:LogPieces` the keys from `source` on should have their comments filled in.

**Comment by felliott** [comment]
Hey @caneruguz,

I'm going to pause here. I still need to CR `project-organizer.js` and test responsive views, but everything else has been reviewed.  If its okay with you, I'll skip reassigning the JIRA tickets, since there's more review to be done.  Take care and enjoy the snow!

Cheers,
Fitz

**Comment by felliott** [comment]
`project-organizer.js` requires `iconmap`, but its usage seems to have been removed in commit `d65ae1d295`.

**Comment by felliott** [comment]
Ditto `Raven`, which was removed in a726f1f7e.

**Comment by felliott** [comment]
`project-organizer.js:_poTitleColumn`: The docstring has text from 2014 saying "Used here to make smart folder italicized" which I think is out of date.  Please update the comment with current behavior.

**Comment by felliott** [comment]
`project-organizer.js:_poModified:` docstring claims to return user name, but only returns date.

**Comment by felliott** [comment]
`project-organizer.js:_poResolveToggle`: last conditional returns empty string instead of empty `m()`.  Is that okay?

**Comment by felliott** [comment]
`project-organizer.js:_poResolveLazyLoad`: JS style question, is it okay to have a function return two different types? This function returns a boolean or a string, which strikes me as odd.  Would it make sense to return the empty string instead of false?

**Comment by felliott** [comment]
`project-organizer.js:filterRowsNotInParent:` Can the css `background-color` property be moved into a class?  If not, could you add a quick comment to explain what `#D18C93` represents? (e.g. "warning", "error", "gratitude").

**Comment by felliott** [comment]
`project-organizer.js:_poIconView:`  Can you inline this in `tbOptions` and add a comment. `// project-organizer doesn't use icons`.  Ignore that if we're planning on adding icons later.

**Comment by felliott** [comment]
If we're not running any of the tests in `website/static/js/tests/projectOrganizer.test.js`, can we delete the commented out code or the whole file?  Keeping it around doesn't seem to serve much purpose.

**Comment by felliott** [comment]
In `_allLogTexts.js`, `tag_deleted` and `dataverse_file_added` both have syntax errors in the strings where the `$` and opening curly brace are swapped.

**Comment by felliott** [comment]
In Firefox responsive view, 360x640, there's a lot of whitespace between the search bar and the "Select a list below to see the project" text.  Also, when I first saw the text, I thought the close icon above and to the right was just to dismiss the text, not the collections list.

screen shot 2016-01-25 at 9 40 43 am

**Comment by felliott** [comment]
In Firefox responsive view, 360x640, I don't a way to create subcomponents unless the current project has no subcomponents.

screen shot 2016-01-25 at 9 45 00 am

**Comment by felliott** [comment]
In Firefox responsive view, 360x640, clicking on an empty project displays the "This project has no subcomponents" message, but the "Add new Subcomponent" link isn't displayed until all the aysnc calls finish.  On my test machine this gap is about 7 seconds, but the gif below speeds it up.

slow-add-component

**Comment by felliott** [comment]
In Firefox responsive view, 768x1024, the Collections header doesn't quite fit, causing the pager and add collection buttons to be partially hidden:
screen shot 2016-01-25 at 10 09 46 am

**Comment by felliott** [comment]
Minor: In Firefox responsive view, 768x1024, the Information and Activity tabs stack on each other.  When I mouseover the Activity tab, it overlaps the divider underneath of information by ~1px, but not for the entire length:

screen shot 2016-01-25 at 10 17 29 am

**Comment by felliott** [comment]
Not a responsiveness issue:  when I navigate into a subcomponent then search, I'm searching the entire collection list.  The breadcrumbs imply that I'm searching just within that project.  In addition, the project count above the search bar still gives the number of projects found within the subcomponent.

screen shot 2016-01-25 at 10 19 59 am

In this screenshot, none of the three list projects are subcomponents of `DBOv3 - EmptyCaregory`.  Also, the active information panel still applies to a project that is not present in the search results.

**Comment by felliott** [comment]
Hey @caneruguz,

I'm all done with this pass.  It's looking fantastic!  Please ping me with any clarifications I can provide.

Cheers,
Fitz

**Comment by sloria** [comment]
Please remove `console.log` statements before committing.

**Comment by sloria** [comment]
There doesn't appear to be proper error handling when a server error occurs; I see the loading indicator indefinitely after getting a 404 response (the 404 might be a local problem, but the client should still handle these errors gracefully).

osf___dashboard

File: website/static/js/_allLogTexts.js [hunk 1]
```diff
@@ -0,0 +1,76 @@
+var logActions = {
+    'project_created' : '${user} created ${node}',
```
sloria: Minor: Trim trailing whitespace. Some editors and linters consider this a linting violation.

website/static/js/_allLogTexts.js
```diff
@@ -0,0 +1,76 @@
...
+    'project_registered' : '${user} registered ${node}', 
+    'project_deleted' : '${user} deleted ${node}', 
+    'created_from' : '${user} created ${node} based on ${template}', 
+    'pointer_created' : '${user} created a link to ${pointer}', 
+    'pointer_forked' : '${user} forked a link to ${pointer}', 
+    'pointer_removed' : '${user} removed a link to ${pointer}', 
+    'made_public' : '${user} made ${node} public', 
+    'made_private' : '${user} made ${node} private', 
+    'tag_added' : '${user} tagged ${node} as ${tag}', 
+    'tag_removed' : '${user} removed {$tag} from ${node}',
```
sloria: Typo: should be `${tag}`. h/t @felliott

website/static/js/_allLogTexts.js
```diff
@@ -0,0 +1,76 @@
...
+    'osf_storage_file_updated' : '${user} updated file ${path} in OSF Storage in ${node}', 
+    'osf_storage_file_removed' : '${user} removed ${path} from OSF Storage in ${node}', 
+    'box_file_added' : '${user} added a ${file} to Box in ${node}', 
+    'box_folder_created' : '${user} created folder ${path} in Box in ${node}', 
+    'box_file_updated' : '${user} updated file ${path} in Box in ${node}', 
+    'box_file_removed' : '${user} removed ${path} from Box in ${node}', 
+    'box_folder_selected' : '${user} linked Box folder ${folder} to ${node}', 
+    'box_node_deauthorized' : '${user} deauthorized the Box addon for ${node}', 
+    'box_node_authorized' : '${user} authorized the Box addon for ${node}', 
+    'dataverse_file_added' : '${user} added file ${filename} to Dataverse dataset {$dataset} in ${node}',
```
sloria: Typo: should be `${dataset}`; h/t @felliott

sloria:
The cursor should be a pointer when hovering over the expand/collapse bar:

osf___dashboard

sloria:
I'm seeing very slow load times when loading my projects in dashboard. Loading my 48 projects to 33 seconds.

This will need to be addressed before this is merged.

sloria:
Stopping review here for now. Going to dig into `filebrowser.js` next.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+            }
+        }
+    };
+}
+
+/**
+ * Initialize File Browser. Prepeares an option object within FileBrowser
+ * @constructor
+ */
+var FileBrowser = {
```
sloria: Why "FileBrowser"? The implementation of this seems tied to ProjectOrganizer, so I would expect this to be named something like "Dashboard".

website/static/js/project-organizer.js
```diff
@@ -400,37 +334,75 @@ var tbOptions = {
...
+        };
+        allProjectsCache = args.allProjects;
+        self.tb = self.updateTB();
     },
     view : function (ctrl, args) {
-        return m('.fb-project-organizer#projectOrganizer', ctrl.tb);
+        var tb = ctrl.tb;
+        if (args.reload()) {
+            tb = ctrl.updateTB();
+            args.reload(false);
```
sloria: This is probably fine for now, but this can be improved by moving this update to the higher-level component (`FileBrowser`) which will make `ProjectOrganizer` mostly stateless.

Something along the lines of 

```js
var FileBrowser = {
    controller() {
        self.reload(false);
        self.handlePOUpdate = function() { self.reload(false); }
        self.shouldPOUpdate = function() {self.reload() === true; };
    }

    view(ctrl, ...) {
        
        m.component(ProjectOrganizer, {
            shouldUpdate: ctrl.shouldPOUpdate,
            onUpdate: ctrl.handlePOUpdate
        })
    }
}
```

```js
var ProjectOrganizer = {
    controller(args) {
        self.updateTB = function() {
            //...
            args.onUpdate.call(self);
        }
    },

    view(ctrl, args) {
        if (args.shouldUpdate()) {
            ctrl.updateTb()
        }
    }
}
```

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+                    'id':  nodeId,
+                    'attributes': {
+                        'title': title
+                    }
+                }
+            };
+            var promise = m.request({method : 'PATCH', url : url, config : xhrconfig, data : data});
+            promise.then(function(result){
+                self.collectionMenuObject().item.label = title;
+                m.redraw(true);
```
sloria: Is this redraw necessary? IIRC, `m.request` will automatically call `redraw` when the promise is resolved.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+                m.component(Filters, {
+                    activeFilter : ctrl.activeFilter,
+                    updateFilter : ctrl.updateFilter,
+                    nameFilters : ctrl.nameFilters,
+                    tagFilters : ctrl.tagFilters
+                })
+            ]) : '',
+            mobile && ctrl.showSidebar() ? '' : m('.fb-main', { style : poStyle },[
+                ctrl.refreshView() ? m('.spinner-div', m('i.fa.fa-refresh.fa-spin')) : '',
+                ctrl.data().data.length === 0 ? ctrl.nonLoadTemplate() : m('#poOrganizer',  m.component( ProjectOrganizer, {
```
sloria: Remove unnecessary usage of CSS IDs.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+        self.breadcrumbs = m.prop([
+            new LinkObject('collection', { path : 'users/me/nodes/', query : { 'related_counts' : true, 'embed' : 'contributors' }, systemCollection : 'nodes'}, 'All My Projects')
+        ]);
+        // Calculate name filters
+        self.nameFilters = [];
+        // Calculate tag filters
+        self.tagFilters = [];
+
+        // Placeholder for node data
+        self.data = m.prop([]);
```
sloria: `self.data` is initialized as an Array prop, but it is treated elsewhere in the code as an object, which is confusing. For example, `self.data().data` is assigned in `updateListError`. Better to ensure that `self.data` is always an Array, unwrapping the `data` envelope when necessary.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+                    console.log(result);
+                }, function _removeProjectFromCollectionsFail(result){
+                    console.log(result);
+                });
+            });
+        };
+        // GETTING THE NODES
+        self.updateList = function _updateList (linkObject, success, error){
+            if(linkObject.data.systemCollection === 'nodes' && self.allProjectsLoaded()){
+                self.data(self.allProjects());
```
sloria: `self.data` is assigned to `self.allProjects` here, but then `self.allProjects` is assigned to `self.data` in `updateListSuccess`. This redundancy suggests that `self.allProjects` is unnecessary. Can you remove it?

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+                if(!self.allProjectsLoaded()) {
+                    collData = { systemCollection : 'nodes' };
+                }
+                self.updateList({link : value.links.next, data : collData });
+                return; // stop here so the reloads below don't run
+            } else {
+                self.loadingNodePages = false;
+            }
+            if(self.loadingAllNodes) {
+                self.data().data = _makeTree(self.data().data);
```
sloria: I'm not sure these direct modifications of `self.data().data` will behave as expected. Mithril props are meant to be called when assigning, so that controllers will properly re-instantiate. As suggested in https://github.com/CenterForOpenScience/osf.io/pull/4887/files#r51177967, better to have `self.data()` be the unwrapped data, so that you can do

```js
self.data(_makeTree(self.data()));
```

which will behave more predictably.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+        item.embeds.contributors.data.forEach(function(c){
+            var attr = c.embeds.users.data.attributes;
+            item.contributors += attr.full_name + ' ' + attr.middle_names + ' ' + attr.given_name + ' ' + attr.family_name + ' ' ;
+        });
+    }
+    item.date = new $osf.FormattableDate(item.attributes.date_modified);
+    return item;
+}
+
+var LinkObject = function _LinkObject (type, data, label, index) {
```
sloria: Is the `index` parameter ever used? If not, please remove it.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+                    } else {
+                        self.nonLoadTemplate(m('.fb-non-load-template.m-md.p-md.osf-box',
+                            'This collection has no projects. To add projects go to "All My Projects" collection; drag and drop projects into the collection link'));
+                    }
+                } else {
+                    self.nonLoadTemplate(m('.fb-non-load-template.m-md.p-md.osf-box.text-center', [
+                        'This project has no subcomponents.',
+                        m.component(AddProject, {
+                            buttonTemplate : m('.btn.btn-link[data-toggle="modal"][data-target="#addSubcomponent"]', 'Add new Subcomponent'),
+                            parentID : self.breadcrumbs()[self.breadcrumbs().length-1].data.id,
```
sloria: Minor: you can use `lastcrumb` here and ~L299.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+        self.getLogs = function _getLogs (nodeId, link, addToExistingList) {
+            var url = link || $osf.apiV2Url('nodes/' + nodeId + '/logs/', { query : { 'page[size]' : 6, 'embed' : ['nodes', 'user', 'linked_node', 'template_node']}});
+            var promise = m.request({method : 'GET', url : url, config : xhrconfig});
+            self.activityLogs([]); // Empty old logs first;
+            self.showMoreActivityLogs(null);
+            promise.then(function _getLogsThen (result){
+                result.data.map(function(log){
+                    log.attributes.formattableDate = new $osf.FormattableDate(log.attributes.date);
+                    if(addToExistingList){
+                        self.activityLogs().push(log);
```
sloria: As mentioned in a previous comment, mithril props are supposed to be assigned by calling them with a value. So, rather than treating `map` like `forEach` here, better to create a new array of formatted logs, then assign `self.activityLogs`.

```js
var activityLogs = result.data.map(function(log){
    var formattedLog = $.extend(true, {}, log, {
        attributes: { formattableDate: new $osf.FormattableDate(log.attributes.date); }
    });
    return formattedLog;
});
self.activityLogs(activityLogs);
```

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+                }
+                self.showMoreActivityLogs(result.links.next); // Set view for show more button
+            });
+            return promise;
+        };
+        // separate concerns, wrap getlogs here to get logs for the selected item
+        self.getCurrentLogs = function _getCurrentLogs ( ){
+            if(self.selected().length === 1){
+                var id = self.selected()[0].data.id;
+                var promise = self.getLogs(id);
```
sloria: Minor: `promise` is unused. Might as well return it.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+        return m('.fb-collections', [
+            collectionListTemplate,
+            m('.fb-collections-modals', [
+                m.component(mC.modal, {
+                    id: 'addColl',
+                    header : m('.modal-header', [
+                        m('button.close[data-dismiss="modal"][aria-label="Close"]', [
+                            m('span[aria-hidden="true"]','×')
+                        ]),
+                        m('h3.modal-title#addCollLabel', 'Add New Collection')
```
sloria: Remove unnecessary CSS IDs

sloria:
Minor: This input could use a placeholder to clarify it's intended use.
osf___dashboard

Something like

osf___dashboard

sloria:
I am unable to click "Remove from collection" here:

osf___dashboard

sloria:
I am getting an error when trying to load a collection:

osf___dashboard

sloria:
I am seeing a strange bug when loading a collection while "All projects" is still loading.

Steps to reproduce:

1. Create a new collection. Put a single project in the collection.
2. Refresh the page.
3. While "All projects" is loading, click on the new collection on the left to view it.

Expected:

The collection loads with the single project in it.

Actual:

A random sampling (?) of all my projects is loaded instead of the collection


dashboard_bug

sloria:
The "Activity" tab in the info panel takes very long to load. There is also no loading indicator.

dashboard_bug

Also, the activity log should be cached so that revisiting the "Activity" tab multiple times doesn't result in multiple requests. It would be nice to have "Reload" button if the user wants the most up-to-date data, but that can probably be deferred until phase 1.

sloria:
This button should change the cursor to a pointer:

osf___dashboard

sloria:
When adding or removing projects from a collection, can you update the counts optimistically rather than waiting for requests to finish?

Right now there is a delay:

count_delay

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+        // GETTING THE NODES
+        self.updateList = function _updateList (linkObject, success, error){
+            if(linkObject.data.systemCollection === 'nodes' && self.allProjectsLoaded()){
+                self.data(self.allProjects());
+                self.reload(true);
+                self.refreshView(false);
+                return;
+            }
+            self.refreshView(true);
+            m.redraw();
```
sloria: Is this redraw necessary? Won't a redraw happen when the request on ~L267 finishes anyway?

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+                    config : xhrconfig
+                }).then(function _removeProjectFromCollectionsSuccess(result){
+                    console.log(result);
+                }, function _removeProjectFromCollectionsFail(result){
+                    console.log(result);
+                });
+            });
+        };
+        // GETTING THE NODES
+        self.updateList = function _updateList (linkObject, success, error){
```
sloria: In general, promises are preferred to `success` and `error` parameters. That said, `success` and `error` don't appear to be used anywhere in the code. If you can confirm that, please remove them.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+            var collection = self.activeFilter().data.node;
+            self.selected().map(function _removeProjectFromCollectionsMap (item){
+                m.request({
+                    method : 'DELETE',
+                    url : collection.links.self + 'node_links/' + item.data.id + '/',
+                    config : xhrconfig
+                }).then(function _removeProjectFromCollectionsSuccess(result){
+                    console.log(result);
+                }, function _removeProjectFromCollectionsFail(result){
+                    console.log(result);
```
sloria: Add proper handling here.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+        self.activeFilter = m.prop({});
+        self.updateFilter = function _updateFilter (filter) {
+            self.activeFilter(filter);
+            self.updateFilesData(filter);
+        };
+
+        self.removeProjectFromCollections = function _removeProjectFromCollection () {
+            // Removes selected items from collect
+            var collection = self.activeFilter().data.node;
+            self.selected().map(function _removeProjectFromCollectionsMap (item){
```
sloria: @brianjgeiger Can we avoid sending requests in a loop? Does APIv2 have a way to remove multiple projects from a collection?

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+
+var xhrconfig = function (xhr) {
+    xhr.withCredentials = true;
+    xhr.setRequestHeader('Content-Type', 'application/vnd.api+json');
+    xhr.setRequestHeader('Accept', 'application/vnd.api+json; ext=bulk');
+};
+
+// Refactor the information needed for filtering rows
+function _formatDataforPO(item) {
+    item.kind = 'folder';
```
sloria: Avoid mutation of arguments when possible:

```js
function _formatDataforPO(item) {
    return $.extend({}, item, {
        kind: 'folder',
        uid: item.id,
        //...
    })
}
```

sloria:
The "Add project" modal doesn't seem to be working: the "Add" and "More" buttons aren't clickable, even when I've input a title.


osf___dashboard

sloria:
Also, I believe "Add new project" should be sentence-case for consistency with the style guide and the rest of the site.

website/static/js/fileBrowser.js
brianjgeiger: Bulk requests should handle this as long as they're from the same collection.

sloria:
Some of the slowness is probably due to the number of requests that are sent. Loading 43 projects resulted in 20 requests (incl. OPTIONS requests).

osf___dashboard

Should we increase the page size for these requests?

api/collections/serializers.py
```diff
@@ -24,6 +25,7 @@ class CollectionSerializer(JSONAPISerializer):
     title = ser.CharField(required=True)
     date_created = ser.DateTimeField(read_only=True)
     date_modified = ser.DateTimeField(read_only=True)
+    bookmarks = ser.BooleanField(read_only=False, default=False, source='is_bookmark_collection')
```
sloria: "bookmarks" implies that the field is a list. Might make sense to rename this to `bookmark`.

website/project/model.py
```diff
@@ -3436,13 +3415,6 @@ class Sanction(StoredObject):
     # Sanction subclasses must have an initiated_by field
     # initiated_by = fields.ForeignField('user', backref='initiated')
 
-    # Expanded: Dictionary field mapping admin IDs their approval status and relevant tokens:
```
sloria: What was this removed?

website/static/js/addProjectPlugin.js
```diff
@@ -0,0 +1,224 @@
...
+    controller : function (options) {
+        var self = this;
+        self.defaults = {
+            buttonTemplate : m('.btn.btn-primary[data-toggle="modal"][data-target="#addProjectModal"]', 'Add new Project'),
+            parentID : null,
+            modalID : 'addProjectModal',
+            stayCallback :null // Function to call when user decides to stay after project creation
+        };
+        self.viewState = m.prop('form'); // 'processing', 'success', 'error';
+        self.options = $.extend(self.defaults, options);
```
sloria: This will mutate `self.defaults`; this should be `$.extend({}, self.defaults, options);`

website/static/js/addProjectPlugin.js
```diff
@@ -0,0 +1,224 @@
...
+        self.goToProjectLink = m.prop('');
+        self.saveResult = m.prop({});
+        self.errorMessageType = m.prop('unknown');
+        self.errorMessage = {
+            'unknown' : 'There was an unknown error. Please try again later.'
+        };
+        // Load Category list from API
+        self.categoryList = [];
+        self.loadCategories = function _loadCategories () {
+            m.request({method : 'OPTIONS', url : $osf.apiV2Url('nodes/', { query : {}}), config : xhrconfig}).then(function _success(results){
```
sloria: I'm not sure that this request should happen in this component. This request will happen whenever the controller gets instantiated, which may happen multiple times (e.g. when props are updated). 

Better to make the request in the parent component, and pass the sorted categories to this component as an 'arg'.

website/static/js/addProjectPlugin.js
```diff
@@ -0,0 +1,224 @@
...
+                }
+            }, function _error(results){
+                console.error('Error loading category names:', results);
+            });
+        };
+        self.loadCategories();
+        // Validation
+        self.isValid = m.prop(false);
+        self.checkValid = function _checkValid() {
+            var projectNameNotEmpty = self.newProjectName().trim().length > 0 ? true : false;
```
sloria: This method can be simplified to 

```js
self.isValid(self.newProjectName().trim().length > 0);
```

sloria: You should be able to remove `isValid`, or at least turn it into a method rather than a prop, since it's derived data.

```js
self.isValid = function() {
    self.newProjectName().trim().length > 0;
}
```

website/static/js/addProjectPlugin.js
```diff
@@ -0,0 +1,224 @@
...
+                url = $osf.apiV2Url('nodes/' + self.options.parentID + '/children/', { query : {}});
+            } else {
+                url = $osf.apiV2Url('nodes/', { query : {}});
+            }
+            data = {
+                    'data' : {
+                        'type': 'nodes',
+                        'attributes': {
+                            'title': self.newProjectName(),
+                            'category': self.newProjectCategory,
```
sloria: Doesn't this need to be called?

website/static/js/addProjectPlugin.js
```diff
@@ -0,0 +1,224 @@
...
+                self.goToProjectLink(result.data.links.html);
+                self.saveResult(result);
+            };
+            var error = function _error (result) {
+                self.viewState('error');
+                console.log('error', result);
+            };
+            m.request({method : 'POST', url : url, data : data, config : xhrconfig})
+                .then(success, error);
+            self.newProjectName('');
```
sloria: Would it make sense to call `self.reset()` here?

website/static/js/addProjectPlugin.js
```diff
@@ -0,0 +1,224 @@
...
+                            m('label[for="projectName].f-w-lg.text-bigger', 'Project Name'),
+                            m('input[type="text"].form-control', {
+                                onkeyup: function(ev){
+                                    if (ev.which === 13) {
+                                         ctrl.add();
+                                    }
+                                    ctrl.newProjectName($(this).val());
+                                    ctrl.checkValid();
+                                },
+                                value : ctrl.newProjectName()
```
sloria: You should be able to do

```js
onchange: m.withAttr('value', ctrl.newProjectName)
```

then remove L125.

website/static/js/addProjectPlugin.js
```diff
@@ -0,0 +1,224 @@
...
+                        ]
+                    ),
+                    m('.modal-footer', [
+                        m('button[type="button"].btn.btn-default[data-dismiss="modal"]', {
+                            onclick : function() {
+                                ctrl.reset();
+                                ctrl.options.stayCallback.call(ctrl); // results are at ctrl.saveResult
+                            }
+                        },  'Keep Working Here'),
+                        m('a.btn.btn-success', { href : ctrl.goToProjectLink() },'Go to New Project')
```
sloria: Should be sentence-case according to osf-style, unless there is good reason to make this title-case.

website/static/js/addProjectPlugin.js
```diff
@@ -0,0 +1,224 @@
...
+                            m('h4.p-md.add-project-success.text-success', 'Project created successfully!')
+                        ]
+                    ),
+                    m('.modal-footer', [
+                        m('button[type="button"].btn.btn-default[data-dismiss="modal"]', {
+                            onclick : function() {
+                                ctrl.reset();
+                                ctrl.options.stayCallback.call(ctrl); // results are at ctrl.saveResult
+                            }
+                        },  'Keep Working Here'),
```
sloria: Is this the correct text, or is this placeholder?

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+            ]));
+            self.data().data = [];
+            self.refreshView(false);
+            throw new Error('Receiving initial data for File Browser failed. Please check your url');
+        };
+        self.generateFiltersList = function _generateFilterList () {
+            self.users = {};
+            self.tags = {};
+            self.data().data.map(function _generateFiltersListMap(item){
+                var contributors = item.embeds.contributors.data ? item.embeds.contributors.data : [];
```
sloria: Minor: this can be simplified to

```js
var contributors = item.embeds.contributors.data || [];
```

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+                        self.users[u.id] = {
+                            data : u,
+                            count: 1
+                        };
+                    } else {
+                        self.users[u.id].count++;
+                    }
+                }
+
+                var tags = item.attributes.tags ? item.attributes.tags : [];
```
sloria: Same here: can be simplified

```js
var tags = item.attributes.tags || [];
```

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+                for(var j = 0; j < tags.length; j++) {
```
sloria: There's duplication here with L344-354 which can be refactored, but it's not absolutely necessary.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+                return;
+            }
+            if(linkObject.ancestors.length > 0){
+                linkObject.ancestors.forEach(function(item){
+                    var ancestorLink = new LinkObject('node', item.data, item.data.name);
+                    self.breadcrumbs().push(ancestorLink);
+                });
+            }
+            self.breadcrumbs().push(linkObject);
+        }.bind(self);
```
sloria: Minor: the `bind` probably isn't necessary, since you use `self` in the method anyway.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+                        'title': self.newCollectionName(),
+                    }
+                }
+            };
+            var promise = m.request({method : 'POST', url : url, config : xhrconfig, data : data});
+            promise.then(function(result){
+                var node = result.data;
+                var count = node.relationships.linked_nodes.links.related.meta.count || 0;
+                self.collections().push(new LinkObject('collection', { path : 'collections/' + node.id + '/linked_nodes/', query : { 'related_counts' : true }, systemCollection : false, node : node, count : m.prop(count) }, node.attributes.title));
+                args.sidebarInit();
```
sloria: I don't see `sidebarInit` being passed as an arg anywhere; I would expect this to throw an error.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+            });
+            self.newCollectionName('');
+            self.dismissModal();
+            self.calculateTotalPages();
+            return promise;
+        };
+        self.deleteCollection = function _deleteCollection(){
+            var url = self.collectionMenuObject().item.data.node.links.self;
+            var promise = m.request({method : 'DELETE', url : url, config : xhrconfig});
+            promise.then(function(result){
```
sloria: Please add error handling to this request.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+            var data = {
+                'data': {
+                    'type': 'collections',
+                    'id':  nodeId,
+                    'attributes': {
+                        'title': title
+                    }
+                }
+            };
+            var promise = m.request({method : 'PATCH', url : url, config : xhrconfig, data : data});
```
sloria: Please add error handling to this request.

website/static/js/fileBrowser.js
sloria: If you are trying to get the title to update, I think this will do the trick:

```js
var updatedCollMenuObj = $.extend(true, {}, self.collectionMenuObject(), {item: {label: title}});
self.collectionMenuObject(updatedCollMenuObj);
```

The invocation of `collectionMenuObject` should result in a redraw.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+                                saveNodetoCollection(index+1);
+                            }
+                            collection.data.count(collection.data.count()+1);
+                        }
+                        m.request({
+                            method : 'POST',
+                            url : collection.data.node.links.self + 'node_links/', //collection.data.node.relationships.linked_nodes.links.related.href,
+                            config : xhrconfig,
+                            data : dataArray[index]
+                        }).then(doNext, doNext); // In case of success or error. It doesn't look like mithril has a general .done method
```
sloria: It doesn't seem correct to do the same thing on both success and failure. Shouldn't an error be shown if an error occurs?

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+                }
+            });
+        };
+        self.validateName = function _validateName (val){
+            if (val === 'Bookmarks') {
+                self.validation(false);
+                self.validationError('Your collection name can\'t be "Bookmarks", because bookmarks feature requires this name. Please select any other name. ');
+            } else {
+                self.validationError('');
+                if(val.length > 0) {
```
sloria: In general, avoid code that looks like `if(someBool) then assign otherBool to true else assign otherBool to false`. Simplify to

```js
self.validation(val.length > 0);
```

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+                    if(dataArray.length > 0){
+                        saveNodetoCollection(0);
+                    }
+                }
+            });
+        };
+        self.validateName = function _validateName (val){
+            if (val === 'Bookmarks') {
+                self.validation(false);
+                self.validationError('Your collection name can\'t be "Bookmarks", because bookmarks feature requires this name. Please select any other name. ');
```
sloria: Possible revision:

> "Bookmarks" is a reserved collection name. Please use another name.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+        self.totalPages = m.prop(1);
+        self.calculateTotalPages = function _calculateTotalPages(result){
+            if(result){ // If this calculation comes after GET call to collections
+                self.totalPages(Math.ceil((result.links.meta.total + args.systemCollections.length)/self.pageSize()));
+            } else {
+                self.totalPages(Math.ceil((self.collections().length)/self.pageSize()));
+            }
+        };
+        self.pageSize = m.prop(5);
+        self.validation = m.prop(false);
```
sloria: Better to call this `isValid` for consistency with other parts of the code.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+        };
+        // Default system collections
+        self.collections = m.prop([
+            new LinkObject('collection', { path : 'users/me/nodes/', query : { 'related_counts' : true, 'page[size]'  : 12, 'embed' : 'contributors' }, systemCollection : 'nodes'}, 'All My Projects'),
+            new LinkObject('collection', { path : 'users/me/registrations/', query : { 'related_counts' : true, 'page[size]'  : 12, 'embed' : 'contributors'}, systemCollection : 'registrations'}, 'All My Registrations')
+        ]);
+        // Load collection list
+        var loadCollections = function _loadCollections (url){
+            var promise = m.request({method : 'GET', url : url, config : xhrconfig});
+            promise.then(function(result){
```
sloria: Please add proper error handling to this request.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+                result.data.forEach(function(node){
+                    var count = node.relationships.linked_nodes.links.related.meta.count;
+                    self.collections().push(new LinkObject('collection', { path : 'collections/' + node.id + '/linked_nodes/', query : { 'related_counts' : true, 'embed' : 'contributors' }, systemCollection : false, node : node, count : m.prop(count) }, node.attributes.title));
```
sloria: As suggested in previous comments, mithril expects props to be updated by invoking them.

```js
var links = result.data.map(...);
self.collections(links);
```

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+            var url = $osf.apiV2Url('collections/', {});
+            var data = {
+                'data': {
+                    'type': 'collections',
+                    'attributes': {
+                        'title': self.newCollectionName(),
+                    }
+                }
+            };
+            var promise = m.request({method : 'POST', url : url, config : xhrconfig, data : data});
```
sloria: Please add proper error handling to this request.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+                        m('button.close[data-dismiss="modal"][aria-label="Close"]', [
+                            m('span[aria-hidden="true"]','×')
+                        ]),
+                        m('h3.modal-title#renameCollLabel', 'Rename Collection')
+                    ]),
+                    body: m('.modal-body', [
+                        m('.form-inline', [
+                            m('.form-group', [
+                                m('label[for="addCollInput]', 'Rename to: '),
+                                m('input[type="text"].form-control.m-l-sm#renameCollInput',{
```
sloria: Remove unnecessary CSS IDs.

website/static/js/fileBrowser.js
```diff
@@ -0,0 +1,1175 @@
...
+            $('[data-toggle="tooltip"]').tooltip();
+        };
+
+        self.init = function _init_fileBrowser() {
+            self.updateList(self.systemCollections[0]);
+        };
+        self.init();
+    },
+    view : function (ctrl, args) {
+        var mobile = window.innerWidth < 767; // true if mobile view
```
sloria: Avoid magic numbers. 767 is used in multiple places; better to declare it as a named constant at the top of the module.

webpack.common.config.js
```diff
@@ -98,7 +102,24 @@ addons.addons.forEach(function(addonName) {
...
+function objToString (obj) {
+    var str = '';
+    for (var p in obj) {
+        if (obj.hasOwnProperty(p)) {
+            str += '    \'' + p + '\' : ' + '\'' + obj[p] + '\', \n';
+        }
+    }
+    return str;
+}
+fs.writeFileSync(staticPath('js/_allLogTexts.js'), 'var logActions = {\n' + objToString(mainLogs) + '}; \n module.exports = logActions;');
```
sloria: Why not just write to a JSON file?

sloria: Also, since you're building the file, it should be gitignored.

website/static/js/logTextParser.js
```diff
@@ -0,0 +1,239 @@
...
+ * @param logObject {Object} the entire log object returned from the api
+ * @returns {boolean}
+ */
+var paramIsReturned = function(param, logObject){
+    if(!param){
+        var message = 'Expected parameter for Log action ' + logObject.attributes.action + ' was not returned from log api.';
+        Raven.captureMessage(message, {logObject: logObject});
+        return false;
+    }
+    return true;
```
sloria: You can simplify to 

```js
return !!param;
```

or 

```js
return Boolean(param);
```

website/static/js/logTextParser.js
```diff
@@ -0,0 +1,239 @@
...
+        view: function (ctrl, logObject) {
+            var template_node = logObject.embeds.template_node;
+            if(paramIsReturned(template_node, logObject)){
+                return m('a', {href: template_node.data.links.html}, template_node.data.attributes.title);
+            }
+            return m('span','a project' );
+        }
+    },
+    // The original title of node involved
+    title_original: {
```
sloria: As per COSDev, use lowerCamelCase for object key names.

website/static/js/onboarder.js
```diff
@@ -0,0 +1,6 @@
+/*
+
+This file is required for prregistration but it needs to change to not use v1 api.
+This placeholder file needs to be deleted before merge to develop
```
sloria: Can this file be removed now?

website/static/js/project-organizer.js
```diff
@@ -129,13 +46,23 @@ function _poTitleColumn(item) {
...
+    return contributorList.map(function (person, index, arr) {
+        var name;
+        var familyName = person.embeds.users.data.attributes.family_name;
+        var givenName = person.embeds.users.data.attributes.given_name;
+        if (familyName) {
+            name = familyName;
+        } else if(givenName){
+            name = givenName;
+        } else {
+            name = 'A Contributor';
```
sloria: "contributor" should be lowercased.

website/static/js/fileBrowser.js
caneruguz: self.allProjects should be acting like a cache of loaded projects so they don't load again. Initial success should add the projects into self.allProjects and after that self.data should be pulling from allProjects. Is this implementation making a mistake there?

website/static/js/project-organizer.js
```diff
@@ -172,30 +98,46 @@ function _poModified(item) {
...
-    var css = '',
-        default_columns;
     if(this.isMultiselected(item.id)){
         item.css = 'fangorn-selected';
     } else {
         item.css = '';
     }
 
-     default_columns = [{
+    default_columns.push({
```
sloria: As per COSDev, use lowerCamelCase for variable names.

sloria:
Pass finished.

website/static/js/fileBrowser.js
sloria: So how does `self.data` differ from `self.allProjects`?

caneruguz: AllProjects is the cache of all projects. Data can be any list of projects loaded into the project organizer, like collections, user's projects, tag projects etc.

website/static/js/fileBrowser.js
caneruguz: Yes, this makes the who self.data more intuitive, it should have been like you said from the beginning.

website/static/js/fileBrowser.js
caneruguz: I'm not sure that this is necessarily true of mithril props. Mithril itself has examples of:
``` 
    proppedArray().push(something)
```
Is that not correct? There is an example of this on the mithril front page: http://mithril.js.org/

website/static/js/fileBrowser.js
caneruguz: @brianjgeiger Can you point me in the right direction about doing bulk request on this?

website/static/js/fileBrowser.js
caneruguz: I see why argument mutation would be a problem but item in this case should have the .kind attribute throughout its existence in this dashboard application. Isn't this a bit overkill for this purpose?

website/static/js/addProjectPlugin.js
caneruguz: Ok in this case, this makes a lot of sense, and your method is the common practice.

website/static/js/fileBrowser.js
brianjgeiger: http://jsonapi.org/extensions/bulk/ <http://jsonapi.org/extensions/bulk/>

api/collections/serializers.py
sloria: ^ @brianjgeiger Are you ok with this change?

website/project/model.py
sloria: @brianjgeiger

website/static/js/dashboard.js
```diff
@@ -0,0 +1,1187 @@
...
+                        dataArray.push(buildCollectionNodeData(ui.draggable.find('.title-text>a').attr('data-nodeID'))); // data-nodeID attribute needs to be set in project organizer building title column
+                    }
+                    function saveNodetoCollection (index) {
+                        function doNext (){
+                            if(dataArray[index+1]){
+                                saveNodetoCollection(index+1);
+                            }
+                            collection.data.count(collection.data.count()+1);
+                        }
+                        m.request({
```
sloria: @pattisdr Is there a way we can use the Bulk API to add multiple nodes to a collection in a single request, rather than sending multiple requests as we're doing here?

api/collections/serializers.py
brianjgeiger: Yeah, `bookmark` makes sense, or possibly `bookmark_collection`.

website/project/model.py
brianjgeiger: It was in error. I thought that was the expanded property from the node that the old project organizer used. My guess is they copied the comment from said property and didn't rename it, because it's probably supposed to be called `approval_state`.

website/static/js/dashboard.js
brianjgeiger: http://jsonapi.org/extensions/bulk/ first example. Also, I think for node_links, @hmoco has made a super-cool relationship self link for doing that in an easier way.

api/collections/serializers.py
sloria: +1 to `bookmark`; @caneruguz can you please do this rename in this PR?

website/static/js/dashboard.js
brianjgeiger: https://github.com/CenterForOpenScience/osf.io/pull/4864

website/project/model.py
sloria: Ah ok; @caneruguz please add this comment back.

website/static/js/dashboard.js
sloria: Awesome. @caneruguz Can you see if you can tap in to this super-cool awesomeness?

pattisdr: Yes, I think Brian implemented this, if my collection id is `xfhge` a bulk POST request to `/v2/collections/xfhge/node_links`/ will allow you to add multiple nodes as node_links.

pattisdr: Neat - didn't know this could be done with self links.

website/static/js/logTextParser.js
caneruguz: I think this is trying to match the key names in the returned field from the api for what they are supposed to be. Otherwise we would need to have an object that maps these naming differences. Does this make sense?

website/static/js/dashboard.js
caneruguz: @sloria can we leave this to the next PR into the branch, which will need to revisit adding and removing items from collections anyway. It looks like bulk operations should best be merged in with those.

samchrisinger:
Doing some investigation of slowness on the page load. It seems the bottleneck is embedded resources. 

I've dug into this some, but don't have a good solution yet. There are two slow queries (the latter being much slower):

`http://localhost:8000/v2/collections/?related_counts=true&page%5Bsize%5D=5&sort=date_created&embed=node_links`  

and 

`http://localhost:8000/v2/users/me/nodes/?related_counts=true&page%5Bsize%5D=60&embed=contributors`


@caneruguz can you please walk me through the need for each embedded field so we can think through potential optimizations?

sloria:
I'm currently getting a 500 error in the API which is preventing me from seeing the Dashboard:

```

Traceback:
File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/django/core/handlers/base.py" in get_response
  132.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/django/views/decorators/csrf.py" in wrapped_view
  58.         return view_func(*args, **kwargs)
File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/django/views/generic/base.py" in view
  71.             return self.dispatch(request, *args, **kwargs)
File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/rest_framework/views.py" in dispatch
  452.             response = self.handle_exception(exc)
File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/rest_framework/views.py" in dispatch
  449.             response = handler(request, *args, **kwargs)
File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/rest_framework/generics.py" in get
  199.         return self.list(request, *args, **kwargs)
File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/rest_framework/mixins.py" in list
  44.             return self.get_paginated_response(serializer.data)
File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/rest_framework/serializers.py" in data
  618.         ret = super(ListSerializer, self).data
File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/rest_framework/serializers.py" in data
  213.                 self._data = self.to_representation(self.instance)
File "/Users/sloria/projects/cos-projects/osf/api/base/serializers.py" in to_representation
  678.             self.child.to_representation(item, envelope=None) for item in data
File "/Users/sloria/projects/cos-projects/osf/api/base/serializers.py" in to_representation
  839.                             data['relationships'][field.field_name] = field.to_representation(attribute)
File "/Users/sloria/projects/cos-projects/osf/api/base/serializers.py" in to_representation
  444.             related_meta = self.get_meta_information(self.related_meta, value)
File "/Users/sloria/projects/cos-projects/osf/api/base/serializers.py" in get_meta_information
  355.                     meta[key] = website_utils.rapply(meta_data[key], _url_val, obj=value, serializer=self.parent)
File "/Users/sloria/projects/cos-projects/osf/website/util/__init__.py" in rapply
  53.         return func(data, *args, **kwargs)
File "/Users/sloria/projects/cos-projects/osf/api/base/serializers.py" in _url_val
  152.         url = getattr(serializer, val)(obj) if obj is not None else None
File "/Users/sloria/projects/cos-projects/osf/api/nodes/serializers.py" in get_unread_comments_count
  211.         file_comments = self.get_unread_file_comments(obj)
File "/Users/sloria/projects/cos-projects/osf/api/nodes/serializers.py" in get_unread_file_comments
  226.                     self.context['view'].get_file_object(obj, file_obj.path, file_obj.provider, check_object_permissions=False)

Exception Type: AttributeError at /v2/users/me/nodes/
Exception Value: 'UserNodes' object has no attribute 'get_file_object'
```

sloria:
Looks like the above error is happening on develop. I get the same 500 when i request `http://localhost:8000/v2/users/me/nodes/?related_counts=true&page[size]=60`.

sloria:
The error is happening when calculating comment counts. @caneruguz Do you need comment counts on the dashboard?

sloria:
Discussed on Flowdock with Caner: only children counts for collections are needed. We should not be calculating file comment counts (though the bug above should still be addressed; I'll look into it so that we can get the dashboard working on staging2 ASAP).

I've added a ticket (https://openscience.atlassian.net/browse/OSF-5725) for requesting specific relationship counts. That will need to be resolved before merging this redesign into develop.

sloria:
Ah, fetching file comments is very expensive because we send a request to Waterbutler for *every non-OSF storage file* for *every node* requested.

Until https://openscience.atlassian.net/browse/OSF-5725 is addressed, I suggest we remove usage of `related_counts` in this PR altogether, with the caveat that the expand/collapse icons for collections may be incorrect. @caneruguz Can you please make this change?

website/static/js/dashboard.js
```diff
@@ -234,36 +261,53 @@ var FileBrowser = {
...
+                };
+                data.data.push(obj);
+            });
+            m.request({
+                method : 'DELETE',
+                url : collection.links.self + 'relationships/' + 'linked_nodes/',  //collection.links.self + 'node_links/' + item.data.id + '/', //collection.links.self + relationship/ + linked_nodes/
+                config : xhrconfig,
+                data : data
+            }).then(function _removeProjectFromCollectionsSuccess(result){
+                console.log(result);
```
sloria: Remove `console.log` statements.

website/static/js/project-organizer.js
```diff
@@ -94,25 +99,26 @@ function _poModified(item) {
  */
 function _poResolveRows(item) {
     var mobile = window.innerWidth < 767; // true if mobile view
+    var tb = this;
+    var folderIcons = tb.filterOn ? false : true;
```
sloria: Simplify to `!tb.filterOn`.

sloria:
Pass finished. Once https://github.com/CenterForOpenScience/osf.io/pull/4887#issuecomment-183451859 and the minor changes noted above are addressed, this is good to merge onto feature/dashboard and can be put onto staging2 for initial performance testing.

The dashboard code will need further review before we merge it into develop.

pattisdr:
> Until https://openscience.atlassian.net/browse/OSF-5725 is addressed, I suggest we remove usage of related_counts in this PR altogether, with the caveat that the expand/collapse icons for collections may be incorrect. @caneruguz Can you please make this change?

https://openscience.atlassian.net/browse/OSF-5725  ready to review.

sloria:
@caneruguz Dawn's PR has been merged into develop.

sloria:
I am getting an error when trying to load "All My Registrations":

osf___dashboard

The error response for `/v2/users/me/registrations/?related_counts=children&page%5Bsize%5D=60&embed=contributors` I am getting is:

```JSON
{"errors":[{"source":{"parameter":"related_counts"},"detail":"Acceptable values for the related_counts query param are 'true', 'false', or any of the relationship fields; got 'children'"}]}
```

sloria:
I am getting errors when resizing the window. Why are requests being sent on resize?

resizing_requests

sloria:
Turns out the above 400 on resizing was a local error, which I've since resolved (https://github.com/CenterForOpenScience/osf.io/pull/4887#issuecomment-185388328) is still an issue, though.

There's also still the issue of making requests on resizing that needs to be addressed.

caneruguz:
The registrations response is strange, I'm not getting the same locally, and 'children' is one of the legitimate relationships. We might need to look into what's different. 

I can replicate the resizing issue, I'll look into it.

sloria:
Another resizing bug: the highlighting of the active filter gets reset when resizing.

To reproduce: click any filter on the right side. Then resize the window. The highlight goes back to "All my projects".

nav

sloria:
I'm seeing a lot of Raven calls (MockRaven calls to be precise) in my console:

osf___dashboard

sloria:
I'm seeing a lot of requests to `http://localhost:8000/v2/nodes/<nid>/logs/?page%5Bsize%5D=6&embed=nodes&embed=user&embed=linked_node&embed=template_node` which looks like it could be expensive, given all the embedding.

I'm seeing those even though I never clicked the "Activity" tab for any node. Can we avoid these requests?

sloria:
Looks like the Raven calls are happening when hovering over rows:

hover

caneruguz:
Raven calls are part of the missing aspects of the logs, Henrique has finished them so that logs have corresponding messages and the Raven errors wouldn't be received.

The reason it runs on hover is because mithril reruns all views at all redraws. This seems unavoidable but the Raven errors will be cleared after Henrique's logs widget PR.

caneruguz:
There shouldn't be a lot of log requests, right now it asks for the log when item is selected but these are cached, so it shouldn't make the same call again. I can make it so that logs are only requested when Activity is clicked or active but I wanted the logs to be already loaded before users click active. They request 6 logs at a time so I didn't think that was expensive. Let me know what you'd like me to do. But there shouldn't be a lot of these so that's the first thing to figure out.

sloria:
Turns out the "All My Registrations" issue will happen if the user has retracted any of their registrations. The issue is documented here: https://openscience.atlassian.net/browse/OSF-5725 (comments).

sloria:
> The reason it runs on hover is because mithril reruns all views at all redraws.

This needs to be fixed, then; Raven calls should not happen on redraws. Keep in mind that every Raven call is a request to Sentry.

caneruguz:
> The reason it runs on hover is because mithril reruns all views at all redraws.
>> This needs to be fixed, then; Raven calls should not happen on redraws. Keep in mind that every Raven call is a request to Sentry.

Good point, in this case we need to refactor this in the logs. Can we postpone this the log widget PR? That one is further along on the log related code.

caneruguz:
Can you check the latest commit: https://github.com/CenterForOpenScience/osf.io/commit/f5f6fc93188cb4a94db11cf82256ddeeae0ea5f3

Moving the actual loading of collections remove the resize related issues. I thought this solution is a good way to go, and removes the need for the module itself to load it's own data.

sloria:
> Good point, in this case we need to refactor this in the logs. Can we postpone this the log widget PR? That one is further along on the log related code.

I don't follow: how will the log widget fix Raven calls in the hover event handler for the dashboard?

sloria:
> There shouldn't be a lot of log requests, right now it asks for the log when item is selected but these are cached, so it shouldn't make the same call again.

OK, I think it's fine to prefetch the logs when a row is clicked. However, you should still prevent duplicate requests from being sent while the first request is being processed. I was able to break the dashboard by click the same row 3 times in succession.

log_requests

sloria:
I don't believe we use the term "Subcomponents" anywhere else on the site. How about just "Components"?

osf___dashboard

sloria:
Also, keep the casing consistent in the link in the above screenshot.

sloria:
After checking out your last commit, looks like the resize issues are resolved. :+1:

website/static/js/dashboard.js
```diff
@@ -397,26 +441,59 @@ var FileBrowser = {
...
-        }.bind(self);
+        };
 
-        self.sidebarInit = function (element, isInit) {
-            if(!isInit){
-                $('[data-toggle="tooltip"]').tooltip();
-            }
+        // GET COLLECTIONS
+        // Default system collections
+        self.collections = m.prop([].concat(self.systemCollections));
```
sloria: Why concat to an empty Array?

website/static/js/dashboard.js
```diff
@@ -397,26 +441,59 @@ var FileBrowser = {
...
+        self.collectionsPageSize = m.prop(5);
+        // Load collection list
+        self.loadCollections = function _loadCollections (url){
+            var promise = m.request({method : 'GET', url : url, config : xhrconfig});
+            promise.then(function(result){
+                result.data.forEach(function(node){
+                    var count = node.relationships.linked_nodes.links.related.meta.count;
+                    self.collections().push(new LinkObject('collection', { path : 'collections/' + node.id + '/linked_nodes/', query : { 'related_counts' : 'children', 'embed' : 'contributors' }, systemCollection : false, node : node, count : m.prop(count) }, node.attributes.title));
```
sloria: Minor: This can be refactored with `.map`:

```js
self.collection(results.data.map(function() {
    //...
    return new LinkObject(...)
}))
```

website/static/js/dashboard.js
caneruguz: Is that not the right thing to do to copy but not reference?

sloria: Ah ok, didn't realize this was meant to copy; ok, this is fine.

website/static/js/dashboard.js
caneruguz: Collections is making multiple calls to API to append to one list though. Would this not replace the collection list with each page and finally collections would be the last page returned?

website/static/js/dashboard.js
```diff
@@ -178,13 +170,36 @@ var FileBrowser = {
...
+                if(results.actions.POST.category){
+                    self.categoryList = results.actions.POST.category.choices;
+                    self.categoryList.sort(function(a, b){ // Quick alphabetical sorting
+                        if(a.value < b.value) return -1;
+                        if(a.value > b.value) return 1;
+                        return 0;
+                    });
+                }
+            }, function _error(results){
+                console.error('Error loading category names:', results);
```
sloria: This error should be logged to Sentry.

sloria:
Pass finished.

Once https://github.com/CenterForOpenScience/osf.io/pull/4887#issuecomment-185419047 and https://github.com/CenterForOpenScience/osf.io/pull/4887#issuecomment-185420664 addressed, this should be good to merge.

caneruguz:
https://github.com/CenterForOpenScience/osf.io/pull/4887#issuecomment-185419047 should be fixed with one of the latest commits. 

I can't replicate https://github.com/CenterForOpenScience/osf.io/pull/4887#issuecomment-185419047. Clicking double should take you into the project, I guess I don't have an environment with slow load to test this. What is your suggestion?

sloria:
@caneruguz After clicking a row, subsequent clicks should not result in additional requests. You can reproduce the problem by clicking a row very quickly and looking at your Network tab. If you have to, you can throttle your connection in your DevTools:

project_organizer_redesign_by_caneruguz_ _pull_request__4887_ _centerforopenscience_osf_io

caneruguz:
@sloria I understand what you mean. I think the last commit takes care of this, I tested it locally and it doesn't send any further calls.

sloria:
Looks like that specific issue is addressed. 

https://github.com/CenterForOpenScience/osf.io/pull/5044 should fix the "All my registrations" tab when a registration has been retracted, so please get this up to date with develop.

@hmoco says he has a commit that allows the Dashboard to function even if a log template is missing; please update this branch with that change.

Once those are addressed, this should be good to merge.

webpack.common.config.js
```diff
@@ -112,14 +112,16 @@ addons.addons.forEach(function(addonName) {
...
         if (obj.hasOwnProperty(p)) {
-            str += '    \'' + p + '\' : ' + '\'' + obj[p] + '\', \n';
+            str += comma + '    "' + p + '" : ' + '"' + obj[p] + '"\n';
         }
+        comma = ',';
     }
     return str;
 }
-fs.writeFileSync(staticPath('js/_allLogTexts.js'), 'var logActions = {\n' + objToString(mainLogs) + '}; \n module.exports = logActions;');
+fs.writeFileSync(staticPath('js/_allLogTexts.json'), '{\n' + objToString(mainLogs) + '}');
```
sloria: Can you just use `JSON.stringify` here?

caneruguz:
@sloria I made the suggested fixes.

sloria:
I am still unable to access "All my registrations" even after the fix from https://github.com/CenterForOpenScience/osf.io/pull/5044 . I am getting the same 400. @pattisdr Can you please work with Caner to resolve this issue? It's the only blocker to getting this merged.

pattisdr:
Looking at this first thing.

caneruguz:
@sloria , @pattisdr and I worked on this, ready for a check from you.

sloria:
I am unable to expand registrations.

regs_expand

@caneruguz @pattisdr Can you look into this?

website/static/js/dashboard.js
```diff
@@ -657,6 +758,20 @@ var Collections  = {
...
             });
         };
+        self.validateName = function _validateName (val){
+            if (val === 'Bookmarks') {
+                self.isValid(false);
+                self.validationError('"Bookmarks" is a reserved collection name. Please use another name.');
+            } else {
+                self.validationError('');
+                if(val.length > 0) {
+                    self.isValid(true);
```
sloria: Minor: This can be simplified to 

```js
self.isValid(val.length);
```

sloria:
Latest commit fixes the toggling issue. Good to merge!

PR #4888, opened 2016-01-13
Feature: Move comments when moving osfstorage file to a component

samanehsan:
Purpose
-----------
Ensures that when an osfstorage file is moved from a parent project to a component (or between components), the comments on the file are still there.

Changes
------------
If the file is getting moved to a different node, send a `file_moved_node` signal. When the signal is received, update all comments on that file with the new node.


website/files/models/base.py
```diff
@@ -398,9 +399,10 @@ def copy_under(self, destination_parent, name=None):
 
     def move_under(self, destination_parent, name=None):
         self.name = name or self.name
+        if self.parent.node != destination_parent.node:
+            project_signals.file_moved_node.send(self, destination_node=destination_parent.node)
```
sloria: I'm not sure what `name` is here, but it might be useful to pass to the signal. That said , I don't think it's necessary at this point.

sloria:

website/files/models/base.py
samanehsan: When you are renaming a file, `name` is the new name of the file. I didn't pass it to the signal since it's not necessary for updating the comments.

sloria: I see that, but other listeners might find it useful. Again, it's minor, and doesn't need to be done at this point.

samanehsan: Ok, I see what you mean now.

PR #4893, opened 2016-01-13
Let Failed Registration Drafts be deleted

samchrisinger:
See: https://openscience.atlassian.net/browse/OSF-5520

# Purpose

Drafts with failed or rejected registrations could not be deleted. The server was giving a 403 response and the JS failed silently.

# Changes

- Let drafts with failed or rejected registrations be deleted
- Add unit test for this behavior
- Add .fail handler to deleteDraft JS call

### DIFF
--- a/tests/test_registrations/test_views.py
+++ b/tests/test_registrations/test_views.py
@@ -432,6 +432,19 @@ class TestDraftRegistrationViews(RegistrationsTestBase):
         res = self.app.delete(url, auth=self.user.auth, expect_errors=True)
         assert_equal(res.status_code, http.FORBIDDEN)
 
+    @mock.patch('website.archiver.tasks.archive')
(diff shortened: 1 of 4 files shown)


tests/test_registrations/test_views.py
```diff
@@ -432,6 +432,19 @@ def test_delete_draft_registration_registered(self, mock_register_draft):
         res = self.app.delete(url, auth=self.user.auth, expect_errors=True)
         assert_equal(res.status_code, http.FORBIDDEN)
 
+    @mock.patch('website.archiver.tasks.archive')
+    def testn_delete_draft_registration_approved_and_registration_deleted(self, mock_register_draft):
```
sloria: The `testn` might prevent this test from being discovered by `nose`. I'll fix this when I merge.

PR #4897, opened 2016-01-14
Add APS2016

mattspitzer:


scripts/populate_conferences.py
```diff
@@ -614,8 +614,16 @@ def main():
         'poster': True,
         'talk': True,
     },
-}
```
sloria: This bracket is necessary because it matches the opening bracket on L21. I'll fix this when I merge.

PR #4898, opened 2016-01-14
Check if contributors or tags exist for a project

caneruguz:
Another fix for the stagin2 bug

### DIFF
--- a/website/static/js/fileBrowser.js
+++ b/website/static/js/fileBrowser.js
@@ -344,7 +344,7 @@ var FileBrowser = {
             self.users = {};
             self.tags = {};
             self.data().data.map(function(item){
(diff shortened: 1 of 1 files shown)


website/static/js/fileBrowser.js
```diff
@@ -344,7 +344,7 @@ var FileBrowser = {
             self.users = {};
             self.tags = {};
             self.data().data.map(function(item){
-                var contributors = item.embeds.contributors.data;
+                var contributors = item.embeds.contributors.data ? item.embeds.contributors.data : [];
```
sloria: This is fine but can be simplified to

```js
item.embeds.contributors.data || [];
```

PR #4901, opened 2016-01-14
Update comments on file move

samanehsan:
When files move within or between external providers and/or between nodes, update the file comments so that they move with the file. Comments do not move when a file gets copied.

### DIFF
--- a/website/project/views/comment.py
+++ b/website/project/views/comment.py
@@ -1,17 +1,53 @@
 # -*- coding: utf-8 -*-
 import pytz
+from datetime import datetime
 from flask import request
 from modularodm import Q
+from modularodm.exceptions import NoResultsFound
 
 from framework.auth.decorators import must_be_logged_in
+from framework.tasks import app
 
+from website.files.models import FileNode, TrashedFileNode
 from website.notifications.constants import PROVIDERS
 from website.notifications.emails import notify
 from website.models import Comment
-from website.project.decorators import must_be_contributor_or_public
+from website.project.decorators import must_be_contributor_or_public, must_have_permission
+from website.project.model import Node
 from website.project.signals import comment_added, file_moved_node
-from datetime import datetime
+from website import settings
 
+
+@app.task
+@must_be_logged_in
+@must_have_permission('write')
+def update_comment_root_target_file(auth, **kwargs):
+    data = request.get_json()
+    source = data.get('source')
+    destination = data.get('destination')
+    node = Node.load(source.get('nodeId'))
+    destination_node = Node.load(destination.get('nodeId'))
+
+    if source.get('provider') == 'osfstorage':
+        try:
+            old_file = TrashedFileNode.find_one(Q('provider', 'eq', source.get('provider')) &
+                                                Q('node', 'eq', node) &
+                                                Q('path', 'eq', source.get('path')))
+        except NoResultsFound:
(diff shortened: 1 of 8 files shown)


website/project/views/comment.py
```diff
@@ -1,21 +1,54 @@
...
-from website.project.decorators import must_be_contributor_or_public
-from website.project.signals import comment_added, file_moved_node
-from datetime import datetime
+from website.project.decorators import must_be_contributor_or_public, must_have_permission
+from website.project.model import Node
+from website.project.signals import comment_added
+from website import settings
+
+
+@app.task
```
sloria: Does this work with celery turned on? I'm not sure of the side effects of making a view function a celery task, but it could cause trouble.

samanehsan: Yes, I believe I tested it but I will double check.

website/static/js/fangorn.js
```diff
@@ -479,6 +481,11 @@ function doItemOp(operation, to, from, rename, conflict) {
         }
         // no need to redraw because fangornOrderFolder does it
         orderFolder.call(tb, from.parent());
+
+        if (operation === OPERATIONS.MOVE) {
```
sloria: It is brittle to put the responsibility of updating comments' root targets on the front end. Important business rules like this should be handled by the backend application. In other words, when files get moved, the backend should handle the Comment updates without the frontend having to do extra work.

sloria:
Pass finished.

As suggested in https://github.com/CenterForOpenScience/osf.io/pull/4901#discussion_r49893496, the comment updates should be handled transparently on the backend. When a file gets moved, the backend should queue up a job to update the file's comments.

website/project/views/comment.py
sloria: Even if it does work, you aren't calling the function asynchronously (e.g. with `my_task.delay`), so the `@app.task` doesn't actually do anything.

To fix this, I suggest the approach in https://github.com/CenterForOpenScience/osf.io/pull/4901#issuecomment-172059266

website/static/js/fangorn.js
samanehsan: That is the approach that I initially took with osfstorage files, but in the case with external providers a request is getting sent to waterbutler and the file grid is updated with the response. It doesn't seem to go through the osf backend.

sloria: Could you use the webhook that we get back from waterbutler? Looks like there's a signal, too: https://github.com/CenterForOpenScience/osf.io/blob/03daf52befe01e62370ba82d4b239f89f161a482/website/addons/base/views.py#L382

website/project/views/comment.py
samanehsan: This will work with celery turned on. Although while I was testing I found an issue with finding the moved osfstorage file, so I'll push that fix.

sloria:
Discussed in person: in the interest of getting this to QA's hands and released, we're going to go ahead and merge this with a few minor modifications.

- Remove `@app.task` from the new view and remove comments module from celery imports
- Change new route to `/api/v1/project/<pid>/comments/target/` (not discussed, but this path is more easily disambiguated from the comment update route)

@felliott is working on a fix for the Waterbutler hook response so that we'll be able to use that to update comment targets instead of having the client send an extra request. This can be an iterative improvement.

PR #4902, opened 2016-01-15
Do not show contributor gravatar when commenting with anonymous link

samanehsan:
Purpose
-----------
Addresses the issue reported in https://openscience.atlassian.net/browse/OSF-23. 

If a user is logged in and is viewing a project that they are a contributor on using an anonymous VOL, all of the commenters are anonymized. But when the user comments, their name and gravatar appear until refreshing the page.

Changes
-----------
If a contributor is logged in and is viewing his own project through an anonymous VOL, if the user leaves a comment, anonymize their name and do not show their gravatar.

### DIFF
--- a/website/static/js/comment.js
+++ b/website/static/js/comment.js
@@ -282,23 +282,21 @@ var CommentModel = function(data, $parent, $root) {
     self.canEdit = ko.observable(data.attributes.can_edit);
     self.hasChildren = ko.observable(data.attributes.has_children);
 
-    if ('embeds' in data && 'user' in data.embeds) {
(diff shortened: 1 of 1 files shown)


sloria:
Aside from this fix, is there any way the client can access gravatar URLs when commenting with a VOL? Anonymization should be handled on the backend as well as the frontend.

sloria:
^ Discussed on Flowdock. From @samanehsan 

> If the API request has an anonymous link token, then no user information is returned

> The issue was that on the front-end, I was showing the gravatar and name of the current logged in user

PR #4906, opened 2016-01-16
Remove registrations from nodes dashboard

brianjgeiger:
Purpose
-----------
Ensures that the `/v2/users/{user}/nodes/` and `/v2/users/{user}/registrations/`include only the appropriate node types, especially to ensure that the nodes endpoint doesn't include registrations.

Closes https://openscience.atlassian.net/browse/OSF-5555

Changes
------------
* Made a common location for a standard nodes request, so if 'what a node is' changes, it will be updated both in the node list and the users node list.
* Added a /registrations/ link to the base view
* Ensured that the registrations/children view actually returned proper information

Side effects
---------------
List views for nodes and node-like things will not include components that a user sees only because they are the administrator on the node. They will still have access to the node, but won't see them in the list. This will be fixed in https://openscience.atlassian.net/browse/OSF-5563 at some point.

### DIFF
--- a/api/registrations/views.py
+++ b/api/registrations/views.py
@@ -250,6 +250,20 @@ class RegistrationContributorDetail(NodeContributorDetail, RegistrationMixin):
 class RegistrationChildrenList(NodeChildrenList, RegistrationMixin):
     view_category = 'registrations'
     view_name = 'registration-children'
+    serializer_class = RegistrationSerializer
+
+    def get_default_odm_query(self):
+        base_query = (
+            Q('is_deleted', 'ne', True) &
+            Q('is_registration', 'eq', True)
+        )
+        user = self.request.user
+        permission_query = Q('is_public', 'eq', True)
+        if not user.is_anonymous():
+            permission_query = (permission_query | Q('contributors', 'icontains', user._id))
+
+        query = base_query & permission_query
+        return query
 
 
 class RegistrationCommentsList(NodeCommentsList, RegistrationMixin):
--- a/api/users/views.py
+++ b/api/users/views.py
@@ -5,7 +5,6 @@ from django.contrib.auth.models import AnonymousUser
 
 from modularodm import Q
 
-from framework.auth.core import Auth
 from framework.auth.oauth_scopes import CoreScopes
 
 from website.models import User, Node
@@ -16,6 +15,7 @@ from api.base.views import JSONAPIBaseView
 from api.base.filters import ODMFilterMixin
 from api.nodes.serializers import NodeSerializer
 from api.registrations.serializers import RegistrationSerializer
+from api.base.utils import default_node_list_query
 
 from .serializers import UserSerializer, UserDetailSerializer
 from .permissions import ReadOnlyOrCurrentUser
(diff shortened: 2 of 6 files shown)


api/registrations/views.py
```diff
@@ -250,6 +250,20 @@ class RegistrationContributorDetail(NodeContributorDetail, RegistrationMixin):
 class RegistrationChildrenList(NodeChildrenList, RegistrationMixin):
     view_category = 'registrations'
     view_name = 'registration-children'
+    serializer_class = RegistrationSerializer
+
+    def get_default_odm_query(self):
+        base_query = (
+            Q('is_deleted', 'ne', True) &
```
sloria: Should we also exclude retractions?

brianjgeiger: I don't believe so. My understanding is we include retractions but serialize them differently.

api/registrations/views.py
```diff
@@ -250,6 +250,20 @@ class RegistrationContributorDetail(NodeContributorDetail, RegistrationMixin):
...
+            Q('is_registration', 'eq', True)
+        )
+        user = self.request.user
+        permission_query = Q('is_public', 'eq', True)
+        if not user.is_anonymous():
+            permission_query = (permission_query | Q('contributors', 'icontains', user._id))
```
sloria: Avoid `contains` and `icontains` when possible because regex queries are expensive. You should be able to use `eq` instead.

api/registrations/views.py
sloria: I don't have a preference. I think it's fine as is, but was wondering if you had thought about retractions. Same with unapproved registrations.

api/users/views.py
```diff
@@ -291,22 +291,20 @@ class UserNodes(JSONAPIBaseView, generics.ListAPIView, UserMixin, ODMFilterMixin
...
-            Q('is_collection', 'ne', True) &
-            Q('is_deleted', 'ne', True)
-        )
+        current_user = self.request.user
+
+        query = (Q('contributors', 'eq', user) & default_node_list_query())
+
+        permission_query = Q('is_public', 'eq', True)
+        if not current_user.is_anonymous():
+            permission_query = (permission_query | Q('contributors', 'icontains', current_user._id))
```
sloria: Samer here--can you use `eq` instead of `icontains`?

api/users/views.py
```diff
@@ -395,9 +393,16 @@ class UserRegistrations(UserNodes):
...
+        query = (
             Q('is_collection', 'ne', True) &
             Q('is_deleted', 'ne', True) &
-            Q('is_registration', 'eq', True)
+            Q('is_registration', 'eq', True) &
+            Q('contributors', 'icontains', user._id)
         )
+        permission_query = Q('is_public', 'eq', True)
+        if not current_user.is_anonymous():
+            permission_query = (permission_query | Q('contributors', 'icontains', current_user._id))
```
sloria: `icontains` -> `eq`?

sloria:
Pass finished.

api/registrations/views.py
brianjgeiger: I don't believe `eq` works here, as it's a list, but I'll give it a try.

sloria: IIRC, `eq` checks membership in lists.

api/registrations/views.py
brianjgeiger: All the retractions stuff was done well before I got to this. I'm just making sure `children` matches what the main collection does.

sloria: Ah, posted my comment after you posted yours. OK, I think this is fine as is for now.

api/registrations/views.py
brianjgeiger: I feel like I've wasted my life. Okay, I'm fixing all the instances in the API (including a couple not listed here) and running tests. If all goes well, we will be somewhat more efficient in these queries (yay!).

PR #4910, opened 2016-01-19
Fix comment reply notification emails

samanehsan:
Purpose
------------
Fixes JIRA ticket 5558

"Comment reply" notification emails are not getting sent.

Changes
------------
In addition to checking subscribers for node-level subscriptions, check user-level subscriptions as well because "comment replies" is specific to the user, not the project.

### DIFF
--- a/website/notifications/emails.py
+++ b/website/notifications/emails.py
@@ -26,6 +26,8 @@ def notify(event, user, node, timestamp, **context):
     target_user = context.get('target_user', None)
     if target_user:
         target_user_id = target_user._id
+        if event_type in constants.USER_SUBSCRIPTIONS_AVAILABLE:
+            subscriptions = check_target_user(target_user, event_type)
     for notification_type in subscriptions:
         if notification_type != 'none' and subscriptions[notification_type]:
             if user in subscriptions[notification_type]:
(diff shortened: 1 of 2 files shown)


website/notifications/emails.py
```diff
@@ -124,6 +126,11 @@ def check_node(node, event):
     return node_subscriptions
 
 
+def check_target_user(user, event):
```
sloria: This is a misnomer: "check_" implies that the function will return a boolean. How about `filter_by_user_and_event`?

samanehsan: Oh, I see. It's not really filtering through subscriptions though so how about `get_user_subscription` instead?

sloria: Sounds good. Let's go with `get_user_subscriptions` (plural, since this is getting all of `user's` subscriptions, right?)

samanehsan: Yes, it's getting all of the user's subscriptions for that particular event type (but not every event they are subscribed to).

sloria: Cool. Made this change when I merged.

samanehsan: Thanks!

PR #4913, opened 2016-01-19
Refactors user merging verification to require pin

Marcus-Rosti:
# Purpose
This is by no means complete.  This implements the front end of the application.  It produces a pin and attaches it to that email to send to someone.  The pin is produced by creating a 6 digit random number.  Also in this is required a mechanism to verify that the provided pin is correct; however, this functionality (server and client side) has not been implemented by this commit.
 
-----

Essentially what should happen is that when two users (Agrippa and Brutus)  we need to should that Agrippa => Brutus and that Brutus => Agrippa.  To do that, we send Brutus and email with confirm token that directs Brutus to a confirm page that requires a pin to complete the merge.  Agrippa receives an email with a pin number regarding the merge.  If Brutus truly is Agrippa, Brutus will have the pin from Agrippa's email. Brutus will then be fully merged into Agrippa in the same way it works currently.

-----

Thus the original user cannot merge in a random user, and that random user cannot complete a merge without verifying that they initialized the merge.

# Changes
There are a significant number of changes.  The biggest is in framework/auth/core.py with a number of refactors elsewhere to deal with the new interface.  These can probably be simplified and there still needs to be the authorization on the other end but otherwise this functionality plays out.
# Interactions
There's one bug where the User object cannot be saved due to the additional field "pending merges"

### DIFF
--- a/framework/auth/core.py
+++ b/framework/auth/core.py
@@ -46,6 +46,12 @@ logger = logging.getLogger(__name__)
 def generate_confirm_token():
     return security.random_string(30)
 
+def generate_confirm_pin():
+    """Generates a 6 digit pin
+    :rtype: int
+    :return: pin number
+    """
+    return security.random_pin()
 
 def generate_claim_token():
     return security.random_string(30)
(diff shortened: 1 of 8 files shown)


sloria:
There are a couple failing tests; @Marcus-Rosti could you please look into those?

sloria:
Closing this for now until this gets back on the priority queue.

PR #4917, opened 2016-01-19
Use file_updated signal to update comments on file move and rename

samanehsan:
Purpose
-----------
- Refactor logic for updating comments on file move to use the `file_updated` signal sent when a waterbutler log is created on the OSF. 
- In addition to updating comments when a file moves, update comments when a file is renamed
- Update the `Node.commented_files` field with the new file id (fixes issue where comments on files that were moved could not be deleted)

### DIFF
--- a/website/project/views/comment.py
+++ b/website/project/views/comment.py
@@ -50,6 +50,12 @@ def update_comment_root_target_file(self, node, event_type, payload, user=None):
         Comment.update(Q('root_target', 'eq', old_file._id), data={'root_target': new_file})
         Comment.update(Q('target', 'eq', old_file._id), data={'target': new_file})
 
+        # update node record of commented files
+        if old_file._id in source_node.commented_files:
+            destination_node.commented_files[new_file._id] = source_node.commented_files[old_file._id]
+            del source_node.commented_files[old_file._id]
+            source_node.save()
+
 
 @comment_added.connect
 def send_comment_added_notification(comment, auth):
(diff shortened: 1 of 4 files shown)


website/project/views/comment.py
```diff
@@ -7,45 +7,45 @@
...
-        cookie=request.cookies.get(settings.COOKIE_NAME)
-    )
-    if node._id != destination_node._id:
-        Comment.update(Q('root_target', 'eq', old_file._id), data={'node': destination_node})
-
-    Comment.update(Q('root_target', 'eq', old_file._id), data={'root_target': new_file})
-    Comment.update(Q('target', 'eq', old_file._id), data={'target': new_file})
+@file_updated.connect
+def update_comment_root_target_file(self, node, event_type, payload, user=None):
+    source = payload.get('source')
```
sloria: Is it possible for `payload` to be missing the `source` and `destination` keys?  If so, L28 and will fail with an `AttributeError`. If you expect those keys to *always* exist, then better to use `payload['source']`. That way, this function will fail early if that expectation isn't met.

sloria:
Approach looks good. Good to merge once the tests are updated.

website/project/views/comment.py
samanehsan: Good catch! It looks like `source` and `destination` won't always be in the payload (e.g. when uploading a file). But they will be there when a file is moved or renamed, so I can do that check first too.

PR #4918, opened 2016-01-19
[rfr][feature] Update S3 addon to ExternalAccounts [OSF-4936]

mfraezz:
Purpose
=======
To support coming APIv2 addon management, we're updating all of our addon models to utilize the ExternalAccount/Provider architecture. 

Changes
=======
* Update S3 to use `ExternalAccount`
* Dataverse-style modal added for credential creation, for cross-addon UX consistency
* Update frontend/tests/etc accordingly

Side Effects
==========
~~Method for assignment of `provider_id` and `display_name` on `ExternalAccount` objects may require further consideration. It's possible for the migration script to assign them as `user._id` and `user.fullname` from their OSF account if the S3 credentials are invalid, though during normal creation they would be assigned as `<s3_canonical_user_id>` and
 `'S3User<[...]{}>'.format(provider_id[-10:])`, because the canonical ID is guaranteed to exist and be accessible then, or account creation would fail. Consider not migrating invalid accounts.~~ 
Invalid accounts not migrated.

Migration Steps
---------------------
`python -m scripts.s3.migrate_to_external_accounts`

[#OSF-4936]


scripts/s3/migrate_to_external_accounts.py
```diff
@@ -0,0 +1,345 @@
...
+ENCRYPT_UPLOADS = s3_settings.ENCRYPT_UPLOADS_DEFAULT
+
+# set of {ExternalAccount._id: [old_user_settings._id, ...]} mappings
+# with invalid credentials, for logging purposes
+invalid_oauth_creds = {}
+
+
+def verify_user_settings_documents(user_document):
+    try:
+        assert_in('_id', user_document)
```
sloria: I don't think we install nose in production environments. You can just use plain `assert` statements instead.

mfraezz: Past migration scripts have used nose asserts, but I can change it.

sloria: I think we installed nose on the Linode server, but now that OSF is dockerized, `nose` won't get installed

mfraezz: Ah, I see. I'll update this, then.

samanehsan:
Should the authorizing user's name be listed here instead?  
screen shot 2016-01-26 at 11 32 44 am

mfraezz:
@samanehsan, S3's API doesn't provide any information about the user other than `<canonical_user_id>`. Because multiple OSF users can share the same S3 account, it wouldn't make sense to have `User B` authorize S3 and see `Authorized by User A`. I spoke with @sloria about this as it related to the migration script as well.

website/addons/s3/static/user-cfg.js
```diff
@@ -1,61 +1,11 @@
...
-var bootbox = require('bootbox');
-var AddonHelper = require('js/addonHelper');
-var language = require('js/osfLanguage').Addons.s3;
+// Endpoint for S3 user settings
+var url = '/api/v1/settings/s3/accounts/';
 
-$('#s3RemoveAccess').on('click', function() {
-    bootbox.confirm({
-        title: 'Disconnect S3 Account?',
-        message: language.confirmDeauth,
```
samanehsan: Can you remove this code as well, since `language.confirmDeauth` is no longer being used?  https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/osfLanguage.js#L74-#L78

website/addons/s3/static/user-cfg.js
```diff
@@ -1,61 +1,11 @@
...
-        buttons:{
-            confirm:{
-                label:'Disconnect',
-                className:'btn-danger'
-            }
-        }
-    });
-});
+var s3ViewModel = new S3ViewModel(url);
+$osf.applyBindings(s3ViewModel, '#s3AddonScope');
```
samanehsan: Why not initialize `S3UserConfig`, which instantiates `S3ViewModel` and does `applyBindings`?

website/addons/s3/static/s3UserConfig.js
```diff
@@ -0,0 +1,205 @@
...
+                access_key: self.accessKey,
+                secret_key: self.secretKey,
+            })
+        ).done(function() {
+            self.clearModal();
+            $modal.modal('hide');
+            self.updateAccounts();
+
+        }).fail(function(xhr, textStatus, error) {
+            var errorMessage = (xhr.status === 401) ? language.authInvalid : language.authError;
```
samanehsan: `authInvalid` and `authError` are missing from `Addons.s3` in `osfLanguage.js`

website/addons/s3/static/s3UserConfig.js
```diff
@@ -0,0 +1,205 @@
...
+        }).done(function (response) {
+            self.accounts($.map(response.accounts, function(account) {
+                var externalAccount =  new ExternalAccount(account);
+                externalAccount.accessKey = account.oauth_key;
+                externalAccount.secretKey = account.oauth_secret;
+                return externalAccount;
+            }));
+            $('.addon-auth-table').osfToggleHeight({height: 140});
+        }).fail(function (xhr, textStatus, error) {
+            self.changeMessage(language.userSettingsError, 'text-danger');
```
samanehsan: Same as above, `userSettingsError` is not in `Addons.s3` in `osfLanguage.js`

website/addons/s3/static/s3UserConfig.js
```diff
@@ -0,0 +1,205 @@
...
+
+function ViewModel(url) {
+    var self = this;
+
+    self.properName = 'Amazon S3';
+    self.accessKey = ko.observable();
+    self.secretKey = ko.observable();
+    self.account_url = '/api/v1/settings/s3/accounts/';
+    // Whether the initial data has been loaded
+    self.loaded = ko.observable(false);
```
samanehsan: I don't see `self.loaded` getting used anywhere. Can it be removed?

website/addons/s3/static/s3UserConfig.js
```diff
@@ -0,0 +1,205 @@
...
+                url: url,
+                status: status,
+                error: error
+            });
+        });
+        return request;
+    };
+
+    /** Change the flashed status message */
+    self.changeMessage = function(text, css, timeout) {
```
samanehsan: Can you include the `ChangeMessageMixin` in this view model instead of re-defining the change message functions? https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/changeMessage.js#L11

website/addons/s3/static/s3UserConfig.js
```diff
@@ -0,0 +1,205 @@
...
+
+    /** Reset all fields from S3 credentials input modal */
+    self.clearModal = function() {
+        self.message('');
+        self.messageClass('text-info');
+        self.accessKey(null);
+        self.secretKey(null);
+    };
+
+    self.updateAccounts = function() {
```
samanehsan: How is `self.updateAccounts` different from `self.fetch`? They both appear to be making a `GET` request to `'/api/v1/settings/s3/accounts/'` and creating a new `ExternalAccount` for each.

website/addons/s3/static/s3UserConfig.js
```diff
@@ -0,0 +1,205 @@
...
+            setTimeout(function() {
+                self.message('');
+                self.messageClass('text-info');
+            }, timeout);
+        }
+    };
+
+    // Update observables with data from the server
+    self.fetch = function() {
+        $.ajax({
```
samanehsan: From the cosdev docs: "When doing AJAX requests or other async work, it’s often useful to return a promise": http://cosdev.readthedocs.org/en/latest/style_guides/javascript.html#promises

website/addons/s3/static/s3UserConfig.js
```diff
@@ -0,0 +1,205 @@
...
+                error: error
+            });
+        });
+        return request;
+    };
+
+    /** Send POST request to authorize S3 */
+    self.sendAuth = function() {
+        // Selection should not be empty
+        if( !self.accessKey() && !self.secretKey() ){
```
samanehsan: I wouldn't consider this a blocker, but instead of checking all of these conditions, you can add validators to the knockout observables. E.g:

```
self.accessKey = ko.observable().extend({
     required: {
           message: 'Please enter an API access key'
}
```

`citationsList.js` is a good example of a VM that uses knockout validation

website/addons/s3/templates/s3_node_settings.mako
```diff
@@ -64,20 +73,6 @@
       </div>
       </div>
     </div>
-    <div data-bind="if: showCreateCredentials">
```
samanehsan: It looks like this was the only place `showCreateCredentials` was being used. Can you remove it from the `s3NodeConfig` view model?

website/addons/s3/static/s3NodeConfig.js
```diff
@@ -209,37 +214,101 @@ ViewModel.prototype._importAuthConfirm = function() {
...
-ViewModel.prototype.importAuth = function() {
+ViewModel.prototype.connectExistingAccount = function(accountId) {
     var self = this;
-    bootbox.confirm({
-        title: 'Import S3 credentials?',
-        message: 'Are you sure you want to authorize this project with your S3 credentials?',
-        callback: function(confirmed) {
-            if (confirmed) {
-                return self._importAuthConfirm();
+    $osf.putJSON(
```
samanehsan: Functions that do AJAX requests should return a promise: 
http://cosdev.readthedocs.org/en/latest/style_guides/javascript.html#promises

Can you also make this change for `ViewModel.prototype.importAuth`?

website/addons/s3/provider.py
```diff
@@ -0,0 +1,21 @@
+from website.addons.s3.serializer import S3Serializer
+
+class S3Provider(object):
+    """An alternative to `ExternalProvider` not tied to OAuth"""
+
+    name = 'Amazon S3'
+    short_name = 's3'
+    provider_name = 's3'
```
samanehsan: Why is there a `provider_name` in addition to `name` and `short_name`?

website/addons/s3/tests/factories.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+"""Factory boy factories for the Box addon."""
```
samanehsan: This should say "s3" instead of "box"

website/addons/s3/tests/factories.py
```diff
@@ -0,0 +1,34 @@
...
+import mock
+from datetime import datetime
+from dateutil.relativedelta import relativedelta
```
samanehsan: It looks like `mock`, `datetime` and `relativedelata` are not used and can be removed

website/addons/s3/tests/test_serializer.py
```diff
@@ -0,0 +1,66 @@
...
+from website.addons.base.testing.serializers import StorageAddonSerializerTestSuiteMixin
+from website.addons.s3.tests.factories import S3AccountFactory
+from website.addons.s3.serializer import S3Serializer
+
+from tests.base import OsfTestCase
+
+
+class TestS3Serializer(StorageAddonSerializerTestSuiteMixin, OsfTestCase):
+    required_settings = ('userIsOwner', 'nodeHasAuth', 'urls', 'userHasAuth',)
+    required_settings_authorized = ('ownerName', )
```
samanehsan: Do `required_settings` and `required_settings_authorized` need to be defined here? They are already defined in `StorageAddonSerializerTestSuiteMixin`

website/addons/s3/tests/test_view.py
```diff
@@ -9,52 +9,35 @@
 from tests.base import OsfTestCase, get_default_metaschema
 from tests.factories import ProjectFactory, AuthUserFactory
 
+from website.addons.base import testing
+from website.addons.s3.provider import S3Provider
```
samanehsan: I looks like `S3Provider`, `BUCKET_LOCATIONS` and `OsfTestCase` are not being used and can be removed

website/addons/s3/tests/test_view.py
```diff
@@ -219,185 +121,63 @@ def test_user_settings_cant_list(self, mock_can_list):
...
-    def test_s3_node_import_auth_unauthorized(self):
-        url = self.project.api_url_for('s3_node_import_auth')
-        self.node_settings.deauthorize(auth=None, save=True)
-        unauthorized = AuthUserFactory()
-        res = self.app.post(url, auth=unauthorized.auth, expect_errors=True)
-        assert_equal(res.status_code, 403)
+    ## Overrides ##
+
+    @mock.patch('website.addons.s3.views.utils.get_bucket_names')
+    def test_folder_list(self, mock_names):
```
samanehsan: Should this test send a request to the `s3_folder_list` endpoint? Right now it's overriding `TestS3Views.test_folder_list()` but is not making any assertions.

website/addons/s3/tests/test_view.py
```diff
@@ -219,185 +121,63 @@ def test_user_settings_cant_list(self, mock_can_list):
...
-
     def test_s3_remove_node_settings_owner(self):
-        url = self.node_settings.owner.api_url_for('s3_delete_node_settings')
+        url = self.node_settings.owner.api_url_for('s3_deauthorize_node')
         ret = self.app.delete(url, auth=self.user.auth)
-
-        assert_equal(ret.json['has_bucket'], False)
-        assert_equal(ret.json['node_has_auth'], False)
+        result = self.Serializer().serialize_settings(node_settings=self.node_settings, current_user=self.user)
+        assert_equal(result['hasBucket'], False)
```
samanehsan: Why was this changed from checking `ret`?

website/addons/s3/tests/test_model.py
```diff
@@ -6,179 +6,90 @@
 from tests.factories import UserFactory, ProjectFactory
 
 from framework.auth import Auth
-from website.addons.s3.model import AddonS3NodeSettings, AddonS3UserSettings
+from website.addons.base.testing import models
+from website.addons.s3.model import S3NodeSettings, S3UserSettings
```
samanehsan: `S3UserSettings` and `UserFactory` are not being used and can be removed. Same with `from boto.s3.connection import *` (not shown in the diff)

samanehsan:
Can you add tests for `s3/migrate_to_external_accounts.py`?

samanehsan:
Also, it looks like there are merge conflicts that need to be resolved :octocat:

scripts/s3/migrate_to_external_accounts.py
```diff
@@ -0,0 +1,332 @@
...
+            broken_user_settings.append(user_settings_document['_id'])
+            continue
+        if not user_settings_document['access_key'] or not user_settings_document['secret_key']:
+            logger.info(
+                "Found addons3usersettings document (id:{0}) with incomplete or no oauth credentials. It will not be migrated.".format(user_settings_document['_id'])
+            )
+            user_no_oauth_creds.append(user_settings_document['_id'])
+            continue
+        external_account, user, new = migrate_to_external_account(user_settings_document)
+        if not external_account:
```
samanehsan: When would `external_account` be `None`? In `migrate_to_external_account`, it looks like if no EA is found, it will get created.

scripts/s3/migrate_to_external_accounts.py
```diff
@@ -0,0 +1,332 @@
...
+    user_id = utils.get_user_id(access_key=user_settings_document['access_key'], secret_key=user_settings_document['secret_key'])
+    user = User.load(user_settings_document['owner'])
+    if not user_id:
+        return (None, None, None)
+    else:
+        display_name = 'S3User<[...]>{}'.format(user_id[-10:])
+
+    new = False
+    try:
+        external_account = ExternalAccount.find(Q('provider_id', 'eq', user_id))[0]
```
samanehsan: Will this query ever return more than one EA?

scripts/s3/migrate_to_external_accounts.py
```diff
@@ -0,0 +1,332 @@
...
+            )
+        )
+
+    if dry_run:
+        raise RuntimeError('Dry run, transaction rolled back.')
+
+
+def main():
+    dry_run = False
+    remove_old = True
```
samanehsan: The `remove_old` flag isn't being used anywhere -- should it be?

samanehsan:
I tested the migration script and it looks good! One thing I noticed was that the node settings looked pretty weird...but then looked fine after refreshing the page so I'm not sure how much of an issue that is.

samanehsan:
Done with first pass @mfraezz! :snowflake:

website/addons/s3/tests/test_view.py
mfraezz: It's overriding `addons.base.testing.views.OAuthAddonConfigViewsTestCaseMixin.test_folder_list` for the sake of mocking an outgoing API call, and then performing [that test](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/addons/base/testing/views.py#L196-L205)

website/addons/s3/tests/test_view.py
mfraezz: Because that view function no longer returns a JSON response, as it does not need to. `S3Serializer.serialize_settings` does effectively what the old `AddonS3NodeSettings.to_json` did.

scripts/s3/migrate_to_external_accounts.py
mfraezz: This was done mostly because error handling is more complicated for `find_one`, and this query is guaranteed to return at most one result.

scripts/s3/migrate_to_external_accounts.py
mfraezz: It would be `None` when the oauth credentials are invalid, which prevents the creation of an `ExternalAccount` and causes `migrate_to_external_account` to [`return (None, None, None)`](https://github.com/CenterForOpenScience/osf.io/pull/4918/files#diff-f62665d47342aa3f7d5f293328733b4eR69)

scripts/s3/migrate_to_external_accounts.py
mfraezz: Oh, no, I wasn't sure whether or not Steve wanted the old model data removed, but he eventually told me not to. I'll remove this.

mfraezz:
The fact that the node settings page looked funny was a likely a caching issue in your browser.

website/addons/s3/static/s3UserConfig.js
mfraezz: You're right, they're duplicate functions. I'll remove one.

website/addons/s3/static/s3UserConfig.js
mfraezz: It seems like the only way to do that and preserve all three messages would be to keep both the access and secret keys in the same observable, with the validator checking both to determine the correct message. This would mess up the data-binding in the template, however. I'm just going to leave it be for now.

website/addons/s3/provider.py
mfraezz: I thought it was required by one of the mixins, but it looks like it's not. I'll remove it.

website/addons/s3/static/user-cfg.js
mfraezz: :+1:

website/addons/s3/templates/s3_node_settings.mako
mfraezz: Fixed this a different way. `showCreateCredentials` and `showTokenCreateButton` had duplicated logic, removed `showTokenCreateButton` and used `showCreateCredentials` where it should be used.

mfraezz:
CR addressed, migration tests added, and conflicts resolved. :snowman: 

Back to you, @samanehsan

website/addons/s3/static/user-cfg.js
```diff
@@ -1,61 +1,10 @@
-'use strict';
+var S3ViewModel = require('./s3UserConfig.js').S3ViewModel;
```
samanehsan: Looks like this can be removed since it's no longer being used.

website/addons/s3/static/user-cfg.js
```diff
@@ -1,61 +1,10 @@
...
-        }
-    });
-    return false;
-}
-
-$(document).ready(function() {
-    $(window.contextVars.addonSettingsSelector).on('submit', AddonHelper.onSubmitSettings);
-});
+// Load initial S3 data
+s3UserConfig.viewModel.updateAccounts();
```
samanehsan: Since this happens when `s3UserConfig.viewModel` is initialized, maybe move it to the `S3ViewModel` constructor?

website/addons/s3/static/s3UserConfig.js
```diff
@@ -0,0 +1,171 @@
...
+
+    self.properName = 'Amazon S3';
+    self.accessKey = ko.observable();
+    self.secretKey = ko.observable();
+    self.account_url = '/api/v1/settings/s3/accounts/';
+    self.accounts = ko.observableArray();
+
+    // Flashed messages
+    self.message = ko.observable('');
+    self.messageClass = ko.observable('text-info');
```
samanehsan: if you put `ChangeMessageMixin.call(self);` in the constructor, you don't have to define `self.message` and `self.messageClass` either.

website/addons/s3/static/s3UserConfig.js
samanehsan: If there was no access key or secret key, then you could have both of those error messages appear instead of a custom third message. But what you currently have is okay too!

website/addons/s3/tests/test_model.py
samanehsan: Please also remove `from boto.s3.connection import *` if it's not being used!

samanehsan:
When connecting an S3 account on the node settings page, the error message in the "Connect an Amazon S3 Account" modal also appears on the project settings page below. Also, should these error messages be the same as on the user settings page?  

screen shot 2016-01-28 at 11 57 56 am

scripts/tests/test_s3_migrate_to_external_accounts.py
```diff
@@ -0,0 +1,196 @@
...
+        assert_equal(
+            node._backrefs['addons']['s3nodesettings']['owner'][0],
+            node_settings._id
+        )
+
+    def test_migrate(self):
+        migration.migrate(dry_run=False)
+        assert_equal(
+            S3UserSettings.find().count(),
+            6
```
samanehsan: Instead of hard-coding the number can you use the corresponding variables instead? It corresponds to the length of `self.linked_user_settings` +`self.unlinked_user_settings`, right?

scripts/tests/test_s3_migrate_to_external_accounts.py
```diff
@@ -0,0 +1,196 @@
...
+        )
+        assert_equal(
+            S3NodeSettings.find().count(),
+            4
```
samanehsan: Same as above. Is it the length of `self.node_settings_documents` + `self.node_settings_no_encrypt`?

samanehsan:
Done with second pass @mfraezz! :whale:

samanehsan:
Oh, one more thing! For some reason I'm still seeing the user name displayed as `S3User<[...]0123456789>:  
screen shot 2016-01-28 at 1 52 17 pm

mfraezz:
Regarding messages on the node settings page, they should in general be the same as on the user settings page, with a couple exceptions. I'll change it so that the messages are not duplicated on both the page and modal, though.

The old username is because the document was never removed from mongo -- disconnecting an account doesn't remove it from the db, only sets a `deleted` flag.

samanehsan:
Looks good @mfraezz! :tada: Done with CR1, but I think this could use another code review since it's such a large refactor :octocat:

website/addons/base/serializer.py
```diff
@@ -74,7 +74,7 @@ def user_is_owner(self):
             (
```
chrisseto: Additional parens

website/addons/s3/model.py
```diff
@@ -5,104 +5,82 @@
...
+    serializer = S3Serializer
 
     bucket = fields.StringField()
     encrypt_uploads = fields.BooleanField(default=ENCRYPT_UPLOADS_DEFAULT)
-    user_settings = fields.ForeignField(
-        'addons3usersettings', backref='authorized'
-    )
 
     @property
     def folder_name(self):
```
chrisseto: Is the property actually used?

website/addons/s3/model.py
```diff
@@ -5,104 +5,82 @@
...
+
+    @property
     def display_name(self):
         return u'{0}: {1}'.format(self.config.full_name, self.bucket)
 
+    def set_folder(self, folder_id, auth):
+        self.bucket = str(folder_id)
+        self.save()
+
+        self.nodelogger.log(action="bucket_linked", extra={'bucket': folder_id}, save=True)
```
chrisseto: Small but `folder_id` should be cast to a string here as well

website/addons/s3/model.py
```diff
@@ -5,104 +5,82 @@
...
-                'node': self.owner._id,
-            },
-            auth=Auth(user_settings.owner),
-        )
-        if save:
-            self.save()
+        self.nodelogger.log(action='node_authorized', save=save)
+
+    def clear_settings(self):
+        self.bucket = None
```
chrisseto: Clearing and deauthorizing are completely separate?

IE clearing does not deauthorize and deauthorizing does not clear?

website/addons/s3/serializer.py
```diff
@@ -0,0 +1,64 @@
...
+        user_settings = self.node_settings.user_settings or self.user_settings
+
+        result = {
+            'accounts': node.api_url_for('s3_account_list'),
+            'create_bucket': node.api_url_for('create_bucket'),
+            'import_auth': node.api_url_for('s3_import_auth'),
+            'create': node.api_url_for('s3_add_user_account'),
+            'deauthorize': node.api_url_for('s3_deauthorize_node'),
+            'bucket_list': node.api_url_for('s3_folder_list'),
+            'set_bucket': node.api_url_for('s3_get_config'),
```
chrisseto: I realize they are the same url, but this should be `'s3_set_config'`

website/addons/s3/serializer.py
```diff
@@ -0,0 +1,64 @@
...
+            'files': node.web_url_for('collect_file_trees'),
+        }
+        if user_settings:
+            result['owner'] = web_url_for('profile_view_id',
+                uid=user_settings.owner._primary_key)
```
chrisseto: `_id`

website/addons/s3/serializer.py
```diff
@@ -0,0 +1,64 @@
...
+        return result
+
+    def credentials_are_valid(self, user_settings):
+        if user_settings:
+            if len(user_settings.external_accounts) < 1:
+                return False
+            return any([utils.can_list(account.oauth_key, account.oauth_secret)
```
chrisseto: Small but if we do get to a point of having multiple addons this should be a for loop or generator. This will evaluate the entire list, possibly making unnecessary request.

website/addons/s3/views.py
```diff
@@ -0,0 +1,189 @@
...
+    """Verifies new external account credentials and adds to user's list"""
+    try:
+        access_key = request.json['access_key']
+        secret_key = request.json['secret_key']
+    except KeyError:
+        raise HTTPError(httplib.BAD_REQUEST)
+
+    if not (access_key and secret_key):
+        return {
+            'message': ('All the fields above are required.')
```
chrisseto: unnecessary parens

website/addons/s3/views.py
```diff
@@ -0,0 +1,189 @@
...
+        account.save()
+    except KeyExistsException:
+        # ... or get the old one
+        account = ExternalAccount.find_one(
+            Q('oauth_key', 'eq', access_key) &
+            Q('oauth_secret', 'eq', secret_key)
+        )
+    assert account is not None
+
+    if account not in auth.user.external_accounts:
```
chrisseto: If `external_accounts` is a foreign field this might not work all the time as @cwisecarver and myself found out.

website/addons/s3/views.py
```diff
@@ -0,0 +1,189 @@
...
+@must_have_permission('write')
+@must_have_addon(SHORT_NAME, 'user')
+@must_have_addon(SHORT_NAME, 'node')
+@must_be_addon_authorizer(SHORT_NAME)
+@must_not_be_registration
+def s3_set_config(node, auth, user_addon, node_addon, **kwargs):
+    """Saves selected bucket to node settings."""
+    # Fail if user settings not authorized
+    if not user_addon.has_auth:
+        raise HTTPError(httplib.BAD_REQUEST)
```
chrisseto: Should be `httplib.UNAUTHORIZED`

website/addons/s3/views.py
```diff
@@ -0,0 +1,189 @@
...
+
+    # If authorized, only owner can change settings
+    if node_addon.has_auth and node_addon.user_settings.owner != auth.user:
+        raise HTTPError(httplib.BAD_REQUEST)
```
chrisseto: Should be `httplib.FORBIDDEN`

website/addons/s3/views.py
```diff
@@ -0,0 +1,189 @@
...
+            'title': 'Invalid bucket location',
+        }, httplib.NOT_ACCEPTABLE
+
+    try:
+        utils.create_bucket(node_addon, bucket_name, bucket_location)
+    except exception.S3ResponseError as e:
+        return {
+            'message': e.message,
+            'title': 'Problem connecting to S3',
+        }, httplib.NOT_ACCEPTABLE
```
chrisseto: These should all be 400s. Unless its a pain to change the error handling code client side.

chrisseto:
@mfraezz Pass complete nothing major, mostly nitpicky

website/addons/s3/model.py
mfraezz: By Archiver

website/addons/s3/model.py
mfraezz: Clearing does not deauthorize, but deauthorizing clears.

website/addons/s3/views.py
mfraezz: As discussed in person, I don't think that's relevant for this case

mfraezz:
@chrisseto CR addressed, back to you

PR #4920, opened 2016-01-20
Change last saved time 

kch8qx:
in 'Edit draft registration' page to display local time, and GMT time in tooltip. Also change 'Started' and 'Last updated' times in Registrations to display local tme, and GMT time in tooltip.

[OSF-5551]


kch8qx:
New view of 'Edit draft registrations' page (default display local time, tooltip display UTC)

PR #4921, opened 2016-01-20
Feature/merge develop with dashboard autodeploy

brianjgeiger:
Purpose
-----------
Bring dashboard-autodeploy current with develop

Changes
------------
Hopefully nothing other than resolving merge conflicts

Side effects
----------------
The things that didn't cause conflicts seemed to revert some changes. I think they're all tracked down, and a few API tests are still running but all the OSF tests passed as of when I'm writing this. That being said, some tests went away, and hopefully only the appropriate ones, but maybe keep an eye out for inappropriately removed tests especially.


sloria:
I suppose this means we'll need to run the backref migrations on staging2 @mattclark

brianjgeiger:
For sure.

PR #4922, opened 2016-01-20
Handle wb connection error

samanehsan:
Purpose
-------------
Adding error handling for local development

Changes
------------
When calling File.touch, if waterbutler is not running, catch the `ConnectionError` and `return None` so that waterbutler does not always need to be running in order to view projects on the OSF.

### DIFF
--- a/website/files/models/base.py
+++ b/website/files/models/base.py
@@ -529,6 +529,7 @@ class File(FileNode):
         headers = {}
         if auth_header:
             headers['Authorization'] = auth_header
+
         resp = requests.get(
             self.generate_waterbutler_url(revision=revision, meta=True, **kwargs),
             headers=headers,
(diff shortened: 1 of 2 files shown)


website/files/models/base.py
```diff
@@ -529,10 +529,15 @@ def touch(self, auth_header, revision=None, **kwargs):
...
-        resp = requests.get(
-            self.generate_waterbutler_url(revision=revision, meta=True, **kwargs),
-            headers=headers,
-        )
+        try:
+            resp = requests.get(
+                self.generate_waterbutler_url(revision=revision, meta=True, **kwargs),
+                headers=headers,
+            )
+        except requests.ConnectionError:
```
sloria: I'm not sure sure this is the right place to do this; `touch` is used in multiple places besides commenting. If it fails to reach Waterbutler, it should fail loudly most of the time. It might be better to catch this error only if we're trying to get an unread comment count?

sloria:
Pass finished. My one concern is noted above.

PR #4923, opened 2016-01-20
Change 'Last saved' time 

kch8qx:
under 'Edit draft registration' to local time, matched format with 'Started' and 'Last updated' under 'Registrations'

[OSF-5551]

### DIFF
--- a/website/static/js/registrationUtils.js
+++ b/website/static/js/registrationUtils.js
@@ -897,14 +897,11 @@ RegistrationEditor.prototype.init = function(draft) {
         if (!self.draft()) {
             return null;
         }
-        return self.draft().updated;
-    });
-    self.lastSaved = ko.computed(function() {
-        var t = self.lastSaveTime();
-        if (t) {
-            return t.toGMTString();
-        } else {
-            return 'never';
+        if (self.draft().updated) {
+            return self.draft().updated;
+        }
+        else {
+            return 'never';            
         }
     });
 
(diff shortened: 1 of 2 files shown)


kch8qx:
See the first attached screenshot to see how the 'Started' and 'Last updated' times are currently displayed on 'Registrations':

screen shot 2016-01-20 at 2 40 47 pm

See the second attached screenshot to see how the 'Last saved' time under 'Edit draft registration' is now displayed to match the above/existing format:

screen shot 2016-01-20 at 2 41 41 pm

sloria:
What if the draft has never been saved. It should show "never", according to the implementation of `lastSaved`. 

Also, could we get rid of the `lastSaved` if we go with this change? Any objections, @samchrisinger ? Was there a reason to show "Last saved" in UTC time?

samchrisinger:
Can you explain what you mean by "get rid of the lastSaved"? 

Either way this can definitely become a localized time.

sloria:
@samchrisinger There is a `lastSaved` computed and a `lastSavedTime` computed. I am suggesting @kch8qx remove `lastSaved` if it will no longer be used anywhere.

samchrisinger:
Without taking time to throughly browse the KO templates for usages of `lastSaved`, this sounds good to me.

kch8qx:
@sloria I changed the lastSaved to show 'never' if the draft hasn't been saved and removed lastSaved entirely since it isn't used anywhere else. Should I update the PR?

PR #4527, opened 2015-10-28
[Feature] Spam Mixin

Ghalko:
## Proposal
A mixin class that handles spam reports wherever a `StoredObject` uses it. This could be used for:
- Comments (used in this PR)
- Nodes
- Files
- Other user contributed stored objects

The ability to add a way to report spam to any user contributed data without repeating basic functionality makes this DRY. But it also leads to improvements needed for comments and how they are displayed.

## Repercussions
This is a potentially unstable use of MODM, though through testing there doesn't seem to be much instability.

PR #4924, opened 2016-01-20
[WIP][Feature] Spam Report API

Ghalko:
## Proposal

Extracting reports from Comments so that it can be applied to any user provided content.

This will add more data for comment reports, so that the UI can be more flexible.

This PR is based off of #4527 

## Repercussions
Shouldn't impinge on any routes.


Ghalko:
Mostly replaced by #5596.

PR #4926, opened 2016-01-20
Add custom TESS field names for OSF4M

hmoco:
And allow for partial custom fields.


sloria:
Note for deployment of 0.63.0 - `python -m scripts.populate_conferences` will need to be run.

PR #4931, opened 2016-01-21
[hotfix] Raise max file size upload limit

icereval:


sloria:
The FAQ page will need to be updated to be consistent with this change. I can do this when I merge. Pinging @saradbowman to update any internal documentation to reflect the increase in per-size upload limit to 5 GB.

sloria:
Ah, looks like there's a client-side size check in the onboarder here: https://github.com/CenterForOpenScience/osf.io/blob/ec9d1e75d3640b3fa605dfcdcc5ea53d4c311a6f/website/static/js/onboarder.js#L530 . I'll update this when I merge.

PR #4932, opened 2016-01-21
[RFR] Feature/Inherit Contributors When Creating Component

billyhunt:
Purpose
-----------

Allow users to add the current set of contributors when creating a new component.

Current
----------
screen shot 2016-01-21 at 3 19 42 pm
screen shot 2016-01-21 at 3 19 31 pm

Proposed
-------------

screen shot 2016-01-21 at 3 18 06 pm

screen shot 2016-01-21 at 3 17 50 pm


whit537:
Needs a rebase.

website/project/forms.py
```diff
@@ -17,3 +17,4 @@ class NewNodeForm(Form):
     ], widget=BootstrapTextInput())
     description = TextAreaField('Description', widget=BootstrapTextArea())
     category = TextAreaField('Category', widget=BootstrapTextArea())
+    inheritContributors = TextAreaField('Category', widget=BootstrapTextArea())
```
whit537: Why are these all `TextAreaField`s when none of them are textareas?

website/project/views/node.py
```diff
@@ -718,6 +723,9 @@ def _view_project(node, auth, primary=False):
     anonymous = has_anonymous_link(node, auth)
     widgets, configs, js, css = _render_addon(node)
     redirect_url = node.url + '?view_only=None'
+    contributor_ids = []
+    for contributor in node.contributors:
+        contributor_ids.append(contributor._id)
```
whit537: No need to change it on this PR, but, for the future, this would be good place for a list comprehension:

```python
contributor_ids = [contributor._id for contributor in node.contributors]
```

website/project/views/node.py
```diff
@@ -781,6 +789,7 @@ def _view_project(node, auth, primary=False):
             },
             'alternative_citations': [citation.to_json() for citation in node.alternative_citations],
             'has_draft_registrations': node.has_active_draft_registrations,
+            'contributors': contributor_ids
```
whit537: Notice the list comprehension just above, for `alternative_citations`. Could do the same here.

whit537:
Looks purty good! Now how about some ...

alphabet_talphabet_ealphabet_salphabet_talphabet_s

whit537:
##### Punchlist

 - [ ] test coverage for new Python—https://github.com/CenterForOpenScience/osf.io/pull/4932#issuecomment-178111552
 - [ ] explain or change `TextAreaField` for non-textarea—https://github.com/CenterForOpenScience/osf.io/pull/4932#discussion_r51456658

whit537:
Should we vary the success message with the `redirect_url`? I was caught off-guard by "keep working on the component page" while being presented with "Contributors"

screen shot 2016-02-01 at 1 31 25 pm

website/project/views/node.py
whit537: > No need to change it on this PR

Actually, looking at the code in context, I think we should switch to an list comprehension inline at (https://github.com/CenterForOpenScience/osf.io/pull/4932#discussion_r51457390) to conform to the existing style here.

website/project/views/node.py
whit537: > Could

Upgrading to "should," to match existing style of the surrounding code.

website/project/views/node.py
```diff
@@ -203,16 +203,21 @@ def project_new_node(auth, node, **kwargs):
                 http.BAD_REQUEST,
                 data=dict(message_long=e.message)
             )
+        redirect_url = new_component.url
+        if form.inheritContributors.data != '':
+            for contributor in node.contributors:
+                new_component.add_contributor(contributor, permissions=node.get_permissions(contributor), auth=auth)
+            new_component.save()
+            redirect_url = redirect_url + 'contributors/'
```
whit537: This needs a test or two.

```
[osf.io]$ py.test --cov=website.project.views.node --cov-report=annotate tests/test_views.py
```

```
[snip]
>         if form.inheritContributors.data != '':
!             for contributor in node.contributors:
!                 new_component.add_contributor(contributor, permissions=node.get_permissions(contributor), auth=auth)
!             new_component.save()
!             redirect_url = redirect_url + 'contributors/'
[snip]
```

website/project/forms.py
billyhunt: I changed the last two to text fields.  Were you thinking I should create a drop down field and/or boolean or checkbox field, or are the text fields cool enough?

billyhunt:
I think this should be ready to look at again.  Let me know if you have any questions.

billyhunt:
@whit537, you have a chance to look at this PR again?

whit537:
Sorry, let's take a look ...

whit537:
What's the Jira tickets for this? Not jumping out at me on https://openscience.atlassian.net/issues/?filter=11402 ...

whit537:
Is [OSF-4457](https://openscience.atlassian.net/browse/OSF-4457) the right one? Or a subtask of that?

billyhunt:
I think its https://openscience.atlassian.net/browse/OSF-5204

whit537:
Yesss! I've moved it to code review so it shows up in [my queue](https://openscience.atlassian.net/issues/?filter=11402). :)

whit537:
(I'm making excuses for why I missed this. Can you tell? :-P )

billyhunt:
1620910rv1jneaw8b

website/project/forms.py
```diff
@@ -16,4 +17,5 @@ class NewNodeForm(Form):
         validators.Length(max=200, message=u'Title must contain fewer than 200 characters.')
     ], widget=BootstrapTextInput())
     description = TextAreaField('Description', widget=BootstrapTextArea())
-    category = TextAreaField('Category', widget=BootstrapTextArea())
+    category = TextField('Category', widget=BootstrapTextInput())
+    inheritContributors = TextField('Inherit', widget=BootstrapTextInput())
```
whit537: I really don't understand the relationship between the `TextField` and `BootstrapTextInput` classes here, and the HTML that is showing up in the browser: a dropdown and a checkbox!

screen shot 2016-02-10 at 10 45 12 am

whit537: [Looks like](https://github.com/CenterForOpenScience/osf.io/blob/fcbc78f18f390218fd18af962c92e4d672d329b4/framework/forms/__init__.py#L3-L8) we're using [an [old version](https://github.com/CenterForOpenScience/osf.io/blob/fcbc78f18f390218fd18af962c92e4d672d329b4/requirements.txt#L13) of] [WTForms](http://wtforms.readthedocs.org/en/1.0.4/widgets.html).

whit537: Oh, here's why: we're just [writing our own HTML](https://github.com/CenterForOpenScience/osf.io/blob/fcbc78f18f390218fd18af962c92e4d672d329b4/website/templates/project/modal_add_component.mako#L17-L32). :-)

So then what's the point of using WTForms?

whit537: Looks like we are using `NewNodeForm` [for validation](https://github.com/CenterForOpenScience/osf.io/blob/fcbc78f18f390218fd18af962c92e4d672d329b4/website/project/views/node.py#L191-L193).

whit537: We should understand and remove the Python bits of `NewNodeForm` that are extraneous.

whit537:
Punchlist updated. I still need to re-review tests and messaging, more in a bit ...

billyhunt:
I think the test coverage is good, at least on the code you showed me.  Is there other tests you are looking for?

whit537:
Test coverage looks good!

```
>         if form.inheritContributors.data != '':
>             for contributor in node.contributors:
>                 new_component.add_contributor(contributor, permissions=node.get_permissions(contributor), auth=auth)
>             new_component.save()
>             redirect_url = redirect_url + 'contributors/'
>             message = (
>                 'Your component was created successfully. You can edit the contributor permissions, '
>                 'work on your <u><a href=' + new_component.url + '>component</a></u> or return to the <u><a href="{url}">project page</a></u>.'
>             ).format(url=node.url)
```

whit537:
> I think the test coverage is good, at least on the code you showed me. Is there other tests you are looking for?

No. I just hadn't yet verified that the test coverage was good. I have now! :-)

@billyhunt++

billyhunt:
I checked with @sloria and this form is pretty old.  Now, we would use knockout, but with this PR, its not worth writing the whole thing.  So adding the field in our older, legacy way seems the way to go in this case.  

I think everything should be solid.  Let me know if I am missing anything.  

tumblr_naya9cdvsc1sioam6o1_250

whit537:
> Now, we would use knockout, but with this PR, its not worth writing the whole thing.

Sorry, I wasn't asking for a rewrite. I was asking to remove the widget cruft, like so:

```diff
diff --git a/website/project/forms.py b/website/project/forms.py
index 870cc79..bbd3dfd 100644
--- a/website/project/forms.py
+++ b/website/project/forms.py
@@ -1,9 +1,6 @@
 # -*- coding: utf-8 -*-
-from framework.forms import Form, TextField, TextAreaField, validators
-from framework.forms import (
-    BootstrapTextArea,
-    BootstrapTextInput,
-)
+from framework.forms import Form, TextField, validators
+
 
 ###############################################################################
 # Forms
@@ -15,7 +12,7 @@ class NewNodeForm(Form):
         validators.Required(message=u'Title is required'),
         validators.Length(min=1, message=u'Title must contain at least 1 character.'),
         validators.Length(max=200, message=u'Title must contain fewer than 200 characters.')
-    ], widget=BootstrapTextInput())
-    description = TextAreaField('Description', widget=BootstrapTextArea())
-    category = TextField('Category', widget=BootstrapTextInput())
-    inheritContributors = TextField('Inherit', widget=BootstrapTextInput())
+    ])
+    description = TextField('Description')
+    category = TextField('Category')
+    inheritContributors = TextField('Inherit')
```

whit537:
(@billyhunt It's bad form to flail at changing code without understanding *why*—e.g., changing `TextAreaField` to `TextField` in fcbc78f18f390218fd18af962c92e4d672d329b4 was a meaningless change, because it turns out we're not using the WTForm's widget machinery *at all*. We should dig down a level or two if need be to understand *why* we're making a change.)

website/project/views/node.py
```diff
@@ -203,16 +203,25 @@ def project_new_node(auth, node, **kwargs):
                 http.BAD_REQUEST,
                 data=dict(message_long=e.message)
             )
-
+        redirect_url = new_component.url
         message = (
-            'Your component was created successfully. You can keep working on the component page below, '
+            'Your component was created successfully. You can edit contributor permissions below, '
```
whit537: This is the message we deliver when they're *not* adding contributors. Why are we changing it here?

screen shot 2016-02-11 at 3 08 07 pm

website/project/views/node.py
```diff
@@ -203,16 +203,25 @@ def project_new_node(auth, node, **kwargs):
...
             'or return to the <u><a href="{url}">project page</a></u>.'
         ).format(url=node.url)
+        if form.inheritContributors.data != '':
+            for contributor in node.contributors:
+                new_component.add_contributor(contributor, permissions=node.get_permissions(contributor), auth=auth)
+            new_component.save()
+            redirect_url = redirect_url + 'contributors/'
+            message = (
+                'Your component was created successfully. You can edit the contributor permissions, '
```
whit537: Let's change "You can edit the contributor permissions," to "You can edit contributor permissions below," to be parallel with the other message.

screen shot 2016-02-11 at 3 07 42 pm

whit537:
Alright @billyhunt, [punchlist](https://github.com/CenterForOpenScience/osf.io/pull/4932#issuecomment-178113033) updated! :)

billyhunt:
I removed the widget variables and fixed the messaging.  Let me know if there is anything else you need.

whit537:
Cool, all clear from here! :-)

website/project/forms.py
```diff
@@ -1,19 +1,17 @@
...
         validators.Required(message=u'Title is required'),
         validators.Length(min=1, message=u'Title must contain at least 1 character.'),
         validators.Length(max=200, message=u'Title must contain fewer than 200 characters.')
-    ], widget=BootstrapTextInput())
-    description = TextAreaField('Description', widget=BootstrapTextArea())
-    category = TextAreaField('Category', widget=BootstrapTextArea())
+    ])
+    description = TextField('Description')
+    category = TextField('Category')
+    inheritContributors = TextField('Inherit')
```
sloria: Shouldn't this be a Boolean field?

sloria: Also, looking at the other WTForms in the codebase, e.g. https://github.com/CenterForOpenScience/osf.io/blob/65aa012f20c3e3e099bc1a09241a231b8a316501/framework/auth/forms.py#L193, it looks like we follow `lower_camel_case` for field names.

website/project/views/node.py
```diff
@@ -203,16 +203,25 @@ def project_new_node(auth, node, **kwargs):
                 http.BAD_REQUEST,
                 data=dict(message_long=e.message)
             )
-
+        redirect_url = new_component.url
         message = (
             'Your component was created successfully. You can keep working on the component page below, '
             'or return to the <u><a href="{url}">project page</a></u>.'
         ).format(url=node.url)
+        if form.inheritContributors.data != '':
```
sloria: If you make `inheritContributors` a Boolean field, you should be able to check `is True`, although it's probably safest to just check the truthiness.

tests/test_views.py
```diff
@@ -4140,6 +4145,30 @@ def test_creates_a_project(self):
         assert_true(node)
         assert_true(node.title, 'Im a real title')
 
+    def test_create_component_add_contributors(self):
+        url = web_url_for('project_new_node', pid=self.project._id)
+        post_data = {'title': 'New Component With Contributors Title',  'category': '', 'inheritContributors': 'true'}
+        request = self.app.post(url, post_data, auth=self.user1.auth)
```
sloria: This variable name is misleading; `post` returns a response object, not a request. By convention, we use `res` for the return value of webtest requests.

tests/test_views.py
```diff
@@ -4140,6 +4145,30 @@ def test_creates_a_project(self):
...
+        self.project.reload()
+        child = self.project.nodes[0]
+        assert_equal(child.title, 'New Component With Contributors Title')
+        assert_in(self.user1, child.contributors)
+        assert_in(self.user2, child.contributors)
+        # check redirect url to make sure its not going to pid/components
+        assert_in('/contributors/', request)
```
sloria: Judging by the comment, I don't think this assertion is doing what you expect; `assert_in('foo', res)` will check that the text 'foo' is anywhere in the response body. 

You should be able to access the request URL more directly. IIRC, it's something like `res.location` or `res.request.path`.

sloria:
As implemented, this feature will allow read-write contributors to choose the "Inherit" option. As of this writing, only *admin* contributors have been authorized to manage contributors.

What permission level should be required for the "Inherit" option? @billyhunt @erinmayhood

sloria:
Minor: rather than making all this text bold, better to make just the project title bold, as to distinguish static vs. dynamic text.

Also, "contributors" should be lower-cased.

sloria:
Pass finished.

website/project/views/node.py
whit537: Here's [`BooleanField`](https://github.com/wtforms/wtforms/blob/1.0.4/wtforms/fields/core.py#L604-L626). Looks like it's designed to do:

`if form.inheritContributors.data:`

whit537: Or

```
if form.inheritContributors.data is True:
```

:-)

billyhunt:
This should be ready to check again @sloria

sloria:
Looks good! Resolved conflicts and merged into `develop`.

PR #4937, opened 2016-01-22
Only allow osfstorage file comments

samanehsan:
Only allow osfstorage file comments due to inconsistencies with making sure comments stay with addon files when they are moved or renamed.

### DIFF
--- a/addons.json
+++ b/addons.json
@@ -24,5 +24,8 @@
         "github": "partial",
         "s3": "partial",
         "wiki": "full"
-    }
+    },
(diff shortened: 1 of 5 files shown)


sloria:
Two suggestions:

- Block commenting for 
- Make commenting support configurable. Could have a setting in `addons.json` like `addons_commentable` or something like that. This will allow us to incrementally enable comments for providers if need be.

samanehsan:
Done making those changes @sloria! :snowman:

PR #4942, opened 2016-01-25
 Feature/logwidg2

hmoco:
Add the Recent Activity widget and the node_logs endpoint to user.

### ` node_logs ` endpoint:
` /v2/users/<user_id>/node_logs/ `

Only takes in ` GET `  requests. Return logs in the same format as logs are normally returned (paginated). However it grabs the logs from every node the user is currently contributing to. Also takes in a query parameter ` aggregate=1` which then includes a count of how many of the logs returned are in each of four categories (nodes, files, comments, wiki).


pattisdr:
Reviewing...

api/users/pagination.py
```diff
@@ -0,0 +1,60 @@
+from collections import OrderedDict
+from django.core.urlresolvers import reverse
+
+from api.base.pagination import JSONAPIPagination
+from rest_framework.response import Response
+
+class UserNodeLogPagination(JSONAPIPagination):
+
+    node_log_aggregates = None
```
pattisdr: Are you using this anywhere?

api/users/pagination.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def get_paginated_response(self, data):
```
pattisdr: We can probably keep from repeating some of the code in `get_pagingated_response`.  I would add a method to the original pagination class we're using,  JSONAPIPagination, called `get_response_dict` or similar, which just returns the basic paginated response that we use in the majority of the api:
```
def get_response_dict(self, data, url):
        return OrderedDict([
            ('data', data),
            ('links', OrderedDict([
                ('first', self.get_first_real_link(url)),
                ('last', self.get_last_real_link(url)),
                ('prev', self.get_previous_real_link(url)),
                ('next', self.get_next_real_link(url)),
                ('meta', OrderedDict([
                    ('total', self.page.paginator.count),
                    ('per_page', self.page.paginator.per_page),
                ]))
            ])),
        ])
```
Then you call this method in JSONAPIPagination > get_paginated_response.  

Finally, instead of overriding the entirety of `get_paginated_response` in users/pagination, you can override just the `get_response_dict` using the same if/else logic you're already using: 

```
if self.request.query_params.get('aggregate'):
    ...return your dictionary with the aggregate info/last_log_date
else:
   return super(UserNodeLogPagination, self).get_response_dict(data, url)
```

api/users/pagination.py
```diff
@@ -0,0 +1,60 @@
...
+                    ('meta', OrderedDict([
+                        ('total', self.page.paginator.count),
+                        ('per_page', self.page.paginator.per_page),
+                        ('aggregates', OrderedDict([
+                            ('comments', comments),
+                            ('wiki', wiki),
+                            ('nodes', nodes),
+                            ('files', files),
+                        ])),
+                        ('last_log_date', list(self.page.paginator.object_list)[-1].date if self.page.paginator.object_list else 0)
```
pattisdr: In my opinion, the `aggregates` and `last_log_date` information do not belong under the `links` meta information.  Links meta information is really for other "non-standard meta information about the link" http://jsonapi.org/format/#document-links, like total number of links and links per page.

I think this should instead be part of the top-level `meta` member, even with `data` and `links`, so:
screen shot 2016-01-29 at 1 41 46 pm

pattisdr: Related: regarding lines that could be problematic in get_paginated_response https://github.com/CenterForOpenScience/osf.io/blob/develop/api/base/pagination.py#L86
```
if is_anonymized(self.request):
            response_dict['meta'] = {'anonymous': True}
```

If you nest aggregates/last_log_date is under the top-level meta object, the above lines could override it.  Change this so it would add {'anonymous:True} to meta object instead of replace it.

api/users/pagination.py
```diff
@@ -0,0 +1,60 @@
...
+class UserNodeLogPagination(JSONAPIPagination):
+
+    node_log_aggregates = None
+
+    def get_paginated_response(self, data):
+        """
+        Formats paginated response in accordance with JSON API.
+
+        Creates pagination links from the view_name if embedded resource,
+        rather than the location used in the request.
```
pattisdr: Change docstring on method instead of copying it - I was confused and thought you overrode this method to handle embedded. But it seems like you just needed a place to put aggregates/last_log_date info.

api/users/pagination.py
```diff
@@ -0,0 +1,60 @@
+from collections import OrderedDict
+from django.core.urlresolvers import reverse
+
+from api.base.pagination import JSONAPIPagination
+from rest_framework.response import Response
+
```
pattisdr: Add line.

api/users/urls.py
```diff
@@ -6,6 +6,6 @@
     url(r'^$', views.UserList.as_view(), name=views.UserList.view_name),
     url(r'^(?P<user_id>\w+)/$', views.UserDetail.as_view(), name=views.UserDetail.view_name),
     url(r'^(?P<user_id>\w+)/nodes/$', views.UserNodes.as_view(), name=views.UserNodes.view_name),
-    url(r'^(?P<user_id>\w+)/registrations/$', views.UserRegistrations.as_view(), name=views.UserRegistrations.view_name)
-
+    url(r'^(?P<user_id>\w+)/node_logs/$', views.UserNodeLogs.as_view(), name=views.UserNodeLogs.view_name),
```
pattisdr: We have `users/me/nodes` and `users/me/registrations`, but the logs here are shared between nodes and registrations, so the endpoint name is confusing. What about `users/me/logs`?

hmoco: That was something that I thought would imply that the logs belong to the user, which is not accurate, as they both contain logs created by other users, and not contain logs attached to a node that the user is no longer a contributor of.

api/users/pagination.py
hmoco: Yeah, that is set to none so the attribute exists, and the view can attached the get_aggregates method to the pagination class.

api/users/views.py
```diff
@@ -401,8 +405,48 @@ def get_default_odm_query(self):
...
             Q('contributors', 'eq', user._id)
         )
+
         permission_query = Q('is_public', 'eq', True)
         if not current_user.is_anonymous():
             permission_query = (permission_query | Q('contributors', 'eq', current_user._id))
         query = query & permission_query
         return query
+
+class UserNodeLogs(JSONAPIBaseView, generics.ListCreateAPIView, ODMFilterMixin, UserMixin):
```
pattisdr: You need to put documentation on this endpoint so users understand what it is returning, and the fact you can add an aggregate query.

api/users/views.py
```diff
@@ -401,8 +405,48 @@ def get_default_odm_query(self):
...
+    def get_default_odm_query(self):
+        user = self.get_user()
+        if user.is_anonymous():
+            auth = Auth(None)
+        else:
+            auth = Auth(user)
+        raw_nodes = Node.find(
+            Q('contributors', 'eq', user) &
+            Q('is_folder', 'ne', True) &
+            Q('is_deleted', 'ne', True)
```
pattisdr: Question: I haven't been keeping up with all API changes, but what is the `Bookmarks` node for?  Do we want logs related to the Bookmarks to show up here?  Because this is a node that seems to have been automatically created for me.

hmoco: I actually have no clue :/ . Pinging @brianjgeiger

api/users/views.py
```diff
@@ -401,8 +405,48 @@ def get_default_odm_query(self):
...
+        )
+        nodes = [each for each in raw_nodes if each.is_public or each.can_view(auth)]
+        if not nodes:
+            raise NotFound
+        query = [Q('params.node', 'eq', node._id) for node in nodes]
+        return reduce(lambda x, y: x | y, query)
```
pattisdr: Can you explain to me how this works? :) I see that it is returning your list of queries into a query group so you can then use it to query NodeLogs but I don't understand how it's doing that.

api/users/views.py
brianjgeiger: 'Bookmarks' is a collection that is made for people to store quick links to projects. It is essentially what the Dashboard folder is on Production. We currently do not have a use case for logs on collections other than that they are nodes so they get logs. But you will never visit a page on the OSF that shows Bookmarks as a node. Does that answer your question?

pattisdr:
On `users/serializers` we have a `nodes` relationship and a `registrations` relationship, so there should also be a `node_logs` relationship.  As of now, the endpoint exists but we're not advertising it anywhere.

pattisdr:
I don't see API tests - you're adding an endpoint so you need tests.  They can be similar to the tests in `api\api_tests\users\views\test_user_nodes_list.py`.

api/users/views.py
hmoco: Sure! So I'm using ` reduce ` which takes in a function with two arguments, and a list. It takes in the first two elements of the list and applies the function (x is first element, y is second), then the output of that function (in this case x | y) becomes the first argument of the function again, and the second argument is replaced by the third element, and so on until it has gone through the entire list. So example:
```
l  = [ 1, 2, 3, 4]

print reduce(lambda x, y: x+y, l)  #  10
# so it does ((1+2) + 3) + 4)
```
A lambda is just an unnamed function, so my thing above ( ` lambda x, y: x | y `), is equivalent to:
```
def func(a, b):
    return a | b
```
does that clear it up a little bit? Let me know if it doesn't, or if that is not what you were asking about :P

pattisdr:
Your query is `v2/users/{u_id}/node_logs/?aggregate=true`, but the meta information it returns has "aggregates".  I think the query should be changed to `?aggregates=true`.

hmoco:
@pattisdr The tests are written, just forgot to push it, pushing it now. sorry :/

api/users/urls.py
pattisdr: Well  `users/me/nodes` are nodes that the user has contributed to, not just nodes the user created. But in the documentation, we clarify, and say the endpoint is a "Paginated list of nodes that the user contributes to". 

In my opinion, calling them `node_logs` still doesn't clear up the confusion as to who created them, and only complicates things by having the user think that registration logs are excluded. They are not.

pattisdr:
No problem.

api/users/views.py
pattisdr: Should Bookmark nodes be excluded from the query then?  Because under my list of user node_logs, there is a log related to the creation of the bookmark.

api/users/views.py
pattisdr: Cool! I had never seen that before.

api/users/views.py
brianjgeiger: They should be, but. This looks like it hasn't been merged with develop in a while, where folders were renamed to collections and dashboard to bookmarks_collection. Once is_folder becomes is_collection, then the bookmarks will automatically be filtered out. If you have a database with both bookmarks and folders, then that's a mess and you should probably ditch your db and start over.

api/users/urls.py
brianjgeiger: I would go with `logs` and documenting it.

api/users/views.py
pattisdr: My database was wiped clean before starting this review - it was a mess because of node log backrefs :)

api_tests/users/views/test_user_node_logs.py
```diff
@@ -0,0 +1,100 @@
...
+
+        assert_equal(res.status_code, 200)
+
+        aggregates = res.json['links']['meta']['aggregates']
+        assert_equal(aggregates['nodes'], 2)
+        assert_equal(aggregates['comments'], 1)
+        assert_equal(aggregates['wiki'], 1)
+        assert_equal(aggregates['files'], 2)
+
+    def test_logs_from_project_no_longer_being_contributed(self):
```
pattisdr: Good tests, Henrique, especially this one.

api/users/pagination.py
pattisdr: How does setting this attribute allow you to call get_aggregates? How does `node_log_aggregates` map back to `get_aggregates`?

pattisdr:
:sunny: API pass complete - good work, the main requested changes are bringing aggregate information under a top-level `meta` object, even with `data` and `links`, instead of putting it inside the links meta object.  Also, consolidation to avoid some repetition with pagination.

api/users/pagination.py
```diff
@@ -0,0 +1,49 @@
...
+                        ('comments', comments),
+                        ('wiki', wiki),
+                        ('nodes', nodes),
+                        ('files', files),
+                    ])),
+                    ('last_log_date', list(self.page.paginator.object_list)[-1].date if self.page.paginator.object_list else 0)
+                ]))
+            ])
+        else:
+            response_dict = OrderedDict([
```
pattisdr: Can you call super() here since this will just return the normal response_dict?

api/users/pagination.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class UserLogPagination(JSONAPIPagination):
+
+    node_log_aggregates = None
+
+    def get_response_dict(self, data, url):
+        if self.request.query_params.get('aggregates'):
+            comments, nodes, wiki, files = self.node_log_aggregates()
+            response_dict = OrderedDict([
```
pattisdr: You could also call super(UserLogPagination, self).get_response_dict(data, url) here and then add the `meta` object to it.

Or call super() once in  the beginning, return that if no 'aggregates' query, otherwise tack the meta object onto the response_dict.

hmoco: good call!

caneruguz:
I'm starting JS Code review on this @hmoco

website/static/js/logActionsList.js
```diff
@@ -59,7 +59,67 @@ var logActions = {
...
+    's3_file_added' : '${user} added file ${path} to Amazon S3 bucket ${bucket} in ${node}',
+    's3_file_removed' : '${user} removed ${path} in Amazon S3 bucket ${bucket} in ${node}',
+    's3_file_updated' : '${user} updated ${path} in Amazon S3 bucket ${bucket} in ${node}',
+    's3_folder_created' : '${user} created folder ${path} in Amazon S3 in ${node}',
+    's3_node_authorized' : '${user} authorized the Amazon S3 addon for ${node}',
+    's3_node_deauthorized' : '${user} deauthorized the Amazon S3 addon for ${node}',
+    'zotero_folder_selected' : '${user} linked Zotero folder ${folder_name} to ${node}', 
+    'zotero_node_authorized' : '${user} authorized the Zotero addon for ${node}',
+    'zotero_node_deauthorized' : '${user} deauthorized the Zotero addon for ${node}',
 };
```
caneruguz: the way we are generating the log list has changed a bit. We now create json file and addons should be aggregating from their own files. for instance website/addons/dataverse/static/dataverseLogActionList.json is an example for dataverse. 

You need to pull from my component-3 branch and sync to get these examples and make changes. _allLogTexts.json is also now in gitignore so it won't appear.

website/static/js/logTextParser.js
```diff
@@ -92,7 +93,7 @@ var LogPieces = {
     node: {
         view: function (ctrl, logObject) {
             var nodeObject = logObject.embeds.nodes;
-            if(paramIsReturned(nodeObject, logObject)){
+            if(paramIsReturned(nodeObject, logObject) && nodeObject.data[0]){
                 return m('a', {href: nodeObject.data[0].links.html}, nodeObject.data[0].attributes.title);
```
caneruguz: I realize this is my code but it looks suspect when a specific index is used from a data. Can you check why the first item in the array is used?

website/static/js/recentActivityWidget.js
```diff
@@ -0,0 +1,325 @@
...
+        self.dateBegin = moment.utc();
+        self.today = moment.utc();
+        self.sixMonthsAgo = moment.utc().subtract(6, 'months');
+        self.page = 1;
+        self.cache = [];
+        self.loading = false;
+        self.div = 8.64e+7;
+        self.canvasHeight = 40;
+        self.totalEvents = 0;
+        self.eventNumbers = {};
```
caneruguz: Can you add comments to these variables, explaining what they do?

website/static/js/recentActivityWidget.js
```diff
@@ -0,0 +1,325 @@
...
+                }
+                if (!init) {
+                    self.cache.push(result.data.slice(0,9));
+                    self.cache.push(result.data.slice(10,19));
+                    if (!update) {
+                        self.activityLogs(self.cache[self.page - 1]);
+                    }
+                }
+                self.lastPage = (result.links.meta.total / (result.links.meta.per_page/2) | 0) + 1;
+            });
```
caneruguz: It looks like there is no error handling here. Is that correct? Can you add a logical error handling both for us and user if necessary.

website/static/js/recentActivityWidget.js
```diff
@@ -0,0 +1,325 @@
...
+            ctx.lineTo(progBar.offset().left + progBar[0].offsetWidth - $('#rACanvas').offset().left, self.canvasHeight);
+            ctx.strokeStyle = '#E0E0E0 ';
+            ctx.lineWidth = 2;
+            ctx.stroke();
+        };
+        self.categoryColor = function(category){
+            if (category.indexOf('wiki') !== -1){ return '#d9534f'; }
+            if (category.indexOf('comment') !== -1){ return '#5bc0de'; }
+            if (category.indexOf('file') !== -1){ return '#337ab7'; }
+            if (category.indexOf('project') !== -1){ return '#f0ad4e'; }
```
caneruguz: Indexof seems like a strange way to check in this case, category is string right? Can you not do a switch statement?

website/static/js/recentActivityWidget.js
```diff
@@ -0,0 +1,325 @@
...
+        };
+        self.addButtons = function(ele, isInitialized) {
+            if ($('#rALeftButton')){
+                $('#rALeftButton').css('height', $('#rALogs').height());
+            }
+            if ($('#rARightButton')){
+                $('#rARightButton').css('height', $('#rALogs').height());
+            }
+        };
+        self.makeSliderProgress =  function(){
```
caneruguz: Can this not be a variable?

website/static/js/recentActivityWidget.js
```diff
@@ -0,0 +1,325 @@
...
+                query['filter[action]'] = self.eventFilter;
+            }
+            if (reset || init) {
+                query.aggregates = 1;
+            }
+            if (!init) {
+                query['filter[date][lte]'] = self.dateEnd.toISOString();
+                query['filter[date][gte]'] = self.dateBegin.toISOString();
+            }
+            var url = $osf.apiV2Url('users/' + self.userId + '/logs/', { query : query});
```
caneruguz: Is there not a way to get the users logs without having to pass the user id through contextVars? Like ```v2/me/logs```?

website/static/js/logTextParser.js
hmoco: If I remember correctly, nodes is an array of the nodes involved in a nodeLog. Which means the first item will be the node the nodeLog belongs to, which is where all ` ${node} `

website/static/js/recentActivityWidget.js
hmoco: category is a string, but I'm trying to see if those strings are part of that string. For example, category would be something like ` osfstorage_file_added `, which would fall under the ` file ` category, and, please correct me if I'm wrong, switch only accounts for exact string matches

caneruguz: Yes that's correct; as long as this method catches all relevant instances this method works fine. For instance every file category has an action that has the word file in them.

caneruguz:
Can you remove ID's from elements had use class selectors with wrapper? 
For instance make wrapper an option for your widget, this could be an ID. Then every other selector should be:
```
$(wrapper + ' .rALeftButton')....
```

or turn it into a function 
```
widget.select = function (selector) {
    return $(self.args.wrapper + selector);
}
```

website/static/js/recentActivityWidget.js
```diff
@@ -0,0 +1,325 @@
...
+        self.page = 1;
+        self.cache = [];
+        self.loading = false;
+        self.div = 8.64e+7;
+        self.canvasHeight = 40;
+        self.totalEvents = 0;
+        self.eventNumbers = {};
+
+        self.getLogs = function(init, reset, update) {
+            if (!(init || reset || update)  && self.cache[self.page - 1]){
```
caneruguz: Would it not be more readable if you did 
```
self.getLogs = function (state) {
...
```
where state is init, reset or update?

website/static/js/recentActivityWidget.js
```diff
@@ -0,0 +1,325 @@
...
+                    log.attributes.formattableDate = new $osf.FormattableDate(log.attributes.date);
+                });
+                if (init) {
+                    self.lastDay = moment.utc(result.data[0].attributes.date);
+                    self.dateEnd = self.lastDay;
+                    var firstDay = moment.utc(result.links.meta.last_log_date);
+                    self.firstDay = ((firstDay >= self.sixMonthsAgo) ? firstDay : self.sixMonthsAgo).startOf('month');
+                    var dateBegin = moment.utc(result.data[0].attributes.date).subtract(1, 'months');
+                    self.dateBegin = ((dateBegin > self.firstDay) ? dateBegin : self.firstDay).startOf('day');
+                    if ((self.today - self.firstDay)/2629746000 < 1){
```
caneruguz: This number is one month in milliseconds? It's better to make it a constant variable. The increments in this case will not be exactly a month. It's not a big deal but it might become problematic over long term. Ideally it's best to to increments in months.

website/static/js/recentActivityWidget.js
```diff
@@ -0,0 +1,325 @@
...
+                        '<div class="progress-bar"></div>' +
+                '</div>';
+        };
+        self.addSlider = function(ele, isInitialized){
+            var begin = (Number(self.firstDay.format('x'))/self.div | 0);
+            var end = (Number(self.today.format('x'))/self.div | 0);
+            var values = [(Number(self.dateBegin.format('x'))/self.div | 0), (Number(self.dateEnd.format('x'))/self.div | 0)];
+            var canvas = document.getElementById('rACanvas');
+            if (!isInitialized) {
+                $('#rASlider').slider({
```
caneruguz: Can you cache jquery selectors within functions? It saves the program from having to traverse the DOM again to find it. 
```
var sliderElement = $('#rASlider');
```

website/static/js/recentActivityWidget.js
```diff
@@ -0,0 +1,325 @@
...
+        };
+    },
+
+    view: function(ctrl, args){
+
+        var fileEvents = ((ctrl.eventNumbers.files/ctrl.totalEvents)*100 | 0) + (ctrl.eventNumbers.files ? 5 : 0);
+        var commentEvents = ((ctrl.eventNumbers.comments/ctrl.totalEvents)*100 | 0) + (ctrl.eventNumbers.comments ? 5 : 0);
+        var wikiEvents = ((ctrl.eventNumbers.wiki/ctrl.totalEvents)*100 | 0) + (ctrl.eventNumbers.wiki ? 5 : 0);
+        var nodeEvents = ((ctrl.eventNumbers.nodes/ctrl.totalEvents)*100 | 0) + (ctrl.eventNumbers.nodes ? 5 : 0);
+        var otherEvents = 100 - (fileEvents + commentEvents + wikiEvents + nodeEvents);
```
caneruguz: Do these have to refresh at every view? Can you make a function to refresh the counts and then call it when it's needed?

website/static/js/recentActivityWidget.js
```diff
@@ -0,0 +1,325 @@
...
+                m('.row', [
+                    m('.col-xs-10.col-xs-offset-1',
+                        m('#rAProgressBar.progress.category-bar',
+                            ctrl.loading ? m('.progress-bar.progress-bar-success.active.progress-bar-striped', {style: {width: '100%'}}, m('b', {style:{color: 'white'}}, 'Loading')) : ([
+                                m('a.progress-bar' + (ctrl.eventFilter === 'file' ||  ctrl.eventFilter === false ?  '.selected' : ''), {style: {width: fileEvents+'%'},
+                                    onclick: function(){
+                                        ctrl.callLogs('file');
+                                    }}, m('i.fa.fa-file.progress-bar-button')
+                                ),
+                                m('a.progress-bar.progress-bar-warning' + (ctrl.eventFilter === 'project' ||  ctrl.eventFilter === false ?  '.selected' : ''), {style: {width: nodeEvents+'%'},
```
caneruguz: Not super necessary but it looks like this can be unified into a template generating function 
```js
var progressBarTemplate = function (type){
    return m('a.progress-bar' + (ctrl.eventFilter === type ||  ctrl.eventFilter === false ?  '.selected' : ''), {style: {width: eventCount[type]+'%'},
     onclick: function(){
        ctrl.callLogs(type);
    }}, m('i.fa.fa-' + type + '.progress-bar-button')
    ),
}
```
It might need a bit more work than that but check it out.

website/templates/home.mako
```diff
@@ -1,5 +1,30 @@
...
+            <div class="logo-spin logo-xl"></div>
+            <p class="m-t-sm fg-load-message">
+                Loading Recent Activity...
+            </p>
+        </div>
+    </div>
+</%def>
+
+<%def name="stylesheets()">
+  <link rel="stylesheet" href="//code.jquery.com/ui/1.11.4/themes/smoothness/jquery-ui.css">
```
caneruguz: Do you still need the jquery ui theme?

website/templates/home.mako
```diff
@@ -1,5 +1,30 @@
...
+  <link rel="stylesheet" href="/static/vendor/bower_components/jQuery-ui-Slider-Pips/dist/jquery-ui-slider-pips.css">
+  <link rel="stylesheet" href="/static/css/recent-activity-widget.css">
+</%def>
+
+<%def name="javascript_bottom()">
+    ${parent.javascript_bottom()}
+    <script type="text/javascript">
+      window.contextVars = $.extend(true, {}, window.contextVars, {
+          userId: ${userId | sjson, n}
```
caneruguz: Again check if this is necessary to do

caneruguz:
Finished pass.

caneruguz:
@hmoco Can you tie the JIRA ticket to this if it's not done? Thanks!

website/templates/home.mako
hmoco: Yep, it seems some of the pips needs it

website/static/js/recentActivityWidget.js
hmoco: It would here, but the rest of the function would have a bunch of string comparisons, like:
```
if ((state === 'init') || (state === 'update')) {
     ...
}
if (state !== 'init') {
    ...
}
```

website/static/js/recentActivityWidget.js
hmoco: This is only used for the top level labels, so they don't really have to be super consistent. At worst case scenario, the label ` February ` points to January 30th.

PR #4945, opened 2016-01-25
[Feature] Remove remote grant during deauthorization if possible

mfraezz:
[#OSF-5365]

Purpose
=======
Addons that follow the [OAuth RFC](http://tools.ietf.org/html/rfc6749) most closely provide an endpoint to revoke the oauth grant on their end, invalidating any access tokens that may have been discovered by a third-party. We should make use of this feature for addons with an API supporting it.

Changes
=======
* Method and call added to base class that will `pass` when tokens should be revoked.
* Above method overridden in addons that support token revocation (and have been migrated to the `ExternalAccount` architecture already.)
* #4853 has had this method preemptively overriden on the `GitHubUserSettings` model.

Side Effects
==========
None

### DIFF
--- a/tests/test_addons_oauth.py
+++ b/tests/test_addons_oauth.py
@@ -294,6 +294,23 @@ class TestUserSettings(OsfTestCase):
             }
         )
 
+    @mock.patch('tests.test_addons_oauth.MockUserSettings.revoke_remote_oauth_access')
+    @mock.patch('framework.auth.core._get_current_user')
+    def test_revoke_remote_access_called(self, mock_decorator, mock_revoke):
+        mock_decorator.return_value = self.user
+        self.user_settings.delete()
+        assert_equal(mock_revoke.call_count, 1)
+
+    @mock.patch('tests.test_addons_oauth.MockUserSettings.revoke_remote_oauth_access')
+    @mock.patch('framework.auth.core._get_current_user')
+    def test_revoke_remote_access_not_called(self, mock_decorator, mock_revoke):
+        mock_decorator.return_value = self.user
+        user2 = AuthUserFactory()
+        user2.external_accounts.append(self.external_account)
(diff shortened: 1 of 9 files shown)


samanehsan:
Code looks good! :tropical_fish: But, when I tested this out for dropbox I still saw the app listed in my dropbox account settings.

samanehsan:
According to @mfraezz: 
>So the grant revocation ticket was slightly incorrect. As far as I can tell, none of our addons support complete revocation though their API, just token disabling. Which is actually what's being done.

samanehsan:
Tested out disabling tokens for box and dropbox and it is working as expected. After disabling the dropbox and box addons, I cannot make any more requests with the disabled tokens. :+1:

samanehsan:
Done with code review! :whale:

website/addons/box/model.py
```diff
@@ -59,6 +60,21 @@ class BoxUserSettings(AddonOAuthUserSettingsBase):
     oauth_provider = Box
     serializer = BoxSerializer
 
+    def revoke_remote_oauth_access(self, external_account):
+        try:
+            # TODO: write client for box, stop using third-party lib
```
sloria: Why rewrite the client?

tests/test_addons_oauth.py
```diff
@@ -294,6 +294,23 @@ def test_grant_metadata(self):
             }
         )
 
+    @mock.patch('tests.test_addons_oauth.MockUserSettings.revoke_remote_oauth_access')
+    @mock.patch('framework.auth.core._get_current_user')
+    def test_revoke_remote_access_called(self, mock_decorator, mock_revoke):
```
sloria: Minor: `mock_decorator` isn't a descriptive name; `mock_get_current_user` would make more sense.

PR #4947, opened 2016-01-26
Feature: Make project comment_level="public" by default

samanehsan:
Purpose
------------
Addresses JIRA ticket https://openscience.atlassian.net/browse/OSF-5002

Changes
------------
Set default `Node.comment_level = 'public'` so that any OSF user can comment on a public project and update API tests.


sloria:
This will expose comments for private projects via the API. I don't think this is desirable. Comments should only be accessible when the project is made public. @samanehsan Can you look into this?

sloria:
Pass finished.

samanehsan:
@sloria, I still get a `403 Forbidden` error when trying to view comments on a private project (with `comment_level=public`) that I am not a contributor on.

sloria:
@samanehsan Which URL are you hitting? I was referring to the comment detail: `/v2/comments/<id of a comment on a private project>`

samanehsan:
I was hitting the node comments endpoint `/v2/nodes/<node_id>/comments/`. I also get a permissions error when trying to view a comment's detail. 

If the project was previously public and you commented on it as a non-contributor, and then it was made private, then I think you can see the comment detail of the comment that you made.

sloria:
Discussed in person: disregard my original comment; the API correctly returns a 401 when trying to access the comment detail with an unauth'd request.

PR #4949, opened 2016-01-26
[Feature] Use generics for GoogleDrive

mfraezz:
[#OSF-5366]

Purpose
=======
A refactor to addon architecture abstracted much of the code repeated across addons. This PR applies those abstractions to GDrive.

Changes
=======
* `generic_views` used
* models updated accordingly
* tests updated to use mixins

Side Effects
=========
unused `foreign_user_settings` on the `GoogleDriveNodeSettings` model removed.

Migration Steps
---------------------
None


website/addons/googledrive/tests/test_models.py
```diff
@@ -11,10 +11,16 @@
 from tests.base import OsfTestCase
 from tests.factories import UserFactory, ProjectFactory
 from website.addons.base import exceptions
```
samanehsan: It looks like these imports are no longer being used. Can you remove them?

```
import time
from datetime import datetime
from dateutil import relativedelta
from framework.exceptions import PermissionsError
from website.addons.base import exceptions
```

website/addons/googledrive/tests/test_views.py
```diff
@@ -9,147 +9,57 @@
 from tests.base import OsfTestCase
 from tests.factories import AuthUserFactory, ProjectFactory
 
+from website.addons.base.testing import views
 from website.addons.googledrive.client import GoogleDriveClient
+from website.addons.googledrive.serializer import GoogleDriveSerializer
 from website.addons.googledrive.tests.utils import mock_folders as sample_folder_data
+from website.addons.googledrive.tests.utils import GoogleDriveAddonTestCase
 from website.addons.googledrive.tests.factories import GoogleDriveAccountFactory
```
samanehsan: This file also has a bunch of unused imports. Please remove...

```
import httplib as http
from website.util import api_url_for, web_url_for
from tests.factories import AuthUserFactory, ProjectFactory
from website.addons.googledrive.tests.factories import GoogleDriveAccountFactory
```

website/addons/googledrive/model.py
```diff
@@ -96,16 +90,19 @@ def folder_name(self):
...
 
-    def clear_auth(self):
+    def fetch_folder_name(self):
+        return self.folder_name
+
+    def clear_settings(self):
         self.folder_id = None
         self.folder_path = None
-        return super(GoogleDriveNodeSettings, self).clear_auth()
+        self.user_settings = None
```
samanehsan: Why do both `clear_settings` and `clear_auth` set `user_settings` to `None`?

samanehsan:
I'm not sure if this is only a google drive issue, but when I connect an account from the node settings page, I get a message that says "Successfully imported access token from profile" even though I did not import it from user settings.

samanehsan:
Pass complete @mfraezz! :cake:

mfraezz:
The message is expected to say that. When connecting the account, it first gets associated with your user, then gets imported. It could be ambiguated to cover both cases, where you create a new or import an existing account, with language like "Successfully connected account," but that isn't necessarily within the scope of this PR.

samanehsan:
Ok, thanks for the clarification!

samanehsan:
One of the google drive logs is not rendering for me...it looks like the `googledrive_node_deauthorized_no_user` log is missing

samanehsan:
@mfraezz, I noticed a log issue that needs to be fixed, but after that this should be good to go! :evergreen_tree:

samanehsan:
Thanks for making those changes @mfraezz! The log is rendering properly now, and the linked google drive folder remains linked after forking a project. This is good to go after the build passes :cat2:

mfraezz:
:+1:

PR #4950, opened 2016-01-27
[rfr][fix] eliminate backrefs between Nodes and NodeLogs

pattisdr:
:exclamation:  https://github.com/CenterForOpenScience/osf.io/pull/5335 and https://github.com/CenterForOpenScience/osf.io/pull/5066 need to be merged first: 

# Purpose

Issue https://openscience.atlassian.net/browse/OSF-5471

There are many log entries, and projects with many logs have trouble loading their logs - this may be causing problems with file management concurrency.  Removing the backrefs between Nodes and NodeLogs should help.  NodeLogs are currently shared between nodes, registrations, and forks. As of this PR, this is no longer the case.

# Changes
1) Removed logs field off of nodes (but left data in db)
2) Made logs a property of nodes
3) Added `original_node` and `node` fields to NodeLogs
4) Removed `node` property off of NodeLogs
5) Created a script to go through all the NodeLogs, and for every registration/fork in the backrefs, clone the NodeLog with `original_node` set equal to the original node(params['node']) and the `node` set equal to the current node.  For the original log, `original_node` and `node` will be equal, but for the cloned logs, `node` will be either the registration or fork.
6) When making a registration or fork, clone the logs on the original node and point them to the reg/fork
7) Removed API endpoint v2/logs/{log_id}/nodes/ since each log only pertains to one node.
8) Changed the log relationship "nodes", to "node" and "original_node", with a NodeDetail link or RegistrationDetail link, depending on object type.
9) Lots of cleanup.
 
# Side Effects
Will require a migration. 

There should be no UI changes - the goal is to remove the backrefs with no changes to the UI.  There are a few API changes.

For some reason, removing the backrefs necessitates reloading the registration and fork during creation.


tests/test_models.py
```diff
@@ -3122,12 +3122,19 @@ def test_no_parent(self):
     def test_get_recent_logs(self):
         # Add some logs
         for _ in range(5):
-            self.project.logs.append(NodeLogFactory())
+            NodeLogFactory.create(
```
mfraezz: This is a lot of kwargs to pass in every time. It might be better to define a `_create` for [NodeLogFactories](https://github.com/CenterForOpenScience/osf.io/blob/develop/tests/factories.py#L300). Something like

```python
@classmethod
def _create(cls, *args, **kwargs):
    node = kwargs.get('node') or NodeFactory()
    user = kwargs.get('user') or node.creator
    original_node = kwargs.get('original_node') or node
    # etc.
    return NodeLog(**locals)
```

mfraezz: Also, is there any reason that `self.project.add_log()` wouldn't work?

tests/test_views.py
```diff
@@ -750,16 +750,17 @@ def test_get_logs_defaults_to_ten(self):
...
+                user=self.user1,
+                action='file_added',
+                params={'node': self.project._id},
+                node=self.project,
+                original_node=self.project
             )
         self.project.save()
         url = self.project.api_url_for('get_logs')
         res = self.app.get(url, {"page": 1}, auth=self.auth)
+        print res
```
mfraezz: This can probably be removed

website/project/model.py
```diff
@@ -890,6 +882,8 @@ def institution_relationship_url(self):
 
     def __init__(self, *args, **kwargs):
 
+        kwargs.pop('logs', [])
```
mfraezz: Is this necessary? I couldn't find anywhere that we're passing `logs` into the `Node` constructor.

website/project/model.py
```diff
@@ -2359,15 +2371,19 @@ def add_log(self, action, params, auth, foreign_user=None, log_date=None, save=T
             user=user,
             foreign_user=foreign_user,
             params=params,
+            node=self,
+            original_node=params['node']
```
mfraezz: Is it possible that neither `'node'` nor `'project'` would be in `params`, and if so, should this default to `self`, as above?

website/project/model.py
```diff
@@ -2359,15 +2371,19 @@ def add_log(self, action, params, auth, foreign_user=None, log_date=None, save=T
...
         )
 
         if log_date:
             log.date = log_date
+        log.save()
 
-        self.date_modified = log.date.replace(tzinfo=None)
+        if len(self.logs) == 1:
```
mfraezz: This conditional appears to be for optimization, but the `.logs` property already evaluates the MODM queryset to return a list, so I'm not sure that this saves any time. Just 
`self.date_modified = self.logs[-1].date.replace(tzinfo=None)` might be better.

website/project/model.py
```diff
@@ -389,6 +389,8 @@ class NodeLog(StoredObject):
     action = fields.StringField(index=True)
     params = fields.DictionaryField()
     should_hide = fields.BooleanField(default=False)
+    original_node = fields.ForeignField('node', index=True)
+    node = fields.ForeignField('node', index=True)
     __indices__ = [
         {
             'key_or_list': [
```
mfraezz: Should this be removed?

scripts/migration/migrate_backrefs_between_nodes_and_nodelogs.py
```diff
@@ -0,0 +1,73 @@
...
+                logger.exception(error)
+                errored_clones.append([log, clone])
+
+    if dry:
+        raise RuntimeError('Dry run -- transaction rolled back')
+
+
+if __name__ == '__main__':
+
+    dry_run = 'dry' in sys.argv
```
mfraezz: This should probably be `'--dry' in sys.argv`. Flags are usually passed in with a single dash for a single letter, or doubly-dashed for a word. 

On an unrelated note, I know this migration works because I accidentally hard-migrated my db. NodeLogs were still persisting and rendering properly after :+1:

website/project/model.py
pattisdr: okay - I wasn't sure what this was doing.

mfraezz:
Pass complete, no bugs found. :star: 

Did some light manual testing, will do more and a migration on a messier DB after above notes have been addressed.

website/project/model.py
pattisdr: If I understood correctly, I was told to remove logs as a field and make it a property, but not have the migration clear it out of the database.  So old projects before the migration would still have log data in the background.

When __init__ was called on an old project, the logs data would still be there and it would fail here.  Popping off the logs allowed me to get around it and still leave old log data in the db.

pattisdr: More information: 
On `__init__`, kwargs are all the fields on the node.  Popping off `logs` handles projects that were created before the migration.

If you do not pop the logs after the migration, you will get `AttributeError: Guid object has no attribute referent`.

mfraezz: You're correct, the ticket says to leave the data there, and modularodm automatically inserts it when loading up an old node. There's probably a way to specify not to, but that's not a rabbit hole worth going down at the moment.

website/project/model.py
mfraezz: It appears to tell mongo what to index on (as opposed to telling modularodm what to index on via the `index` kwarg in a `fields` constructor), and I haven't noticed any problems removing it locally.

website/project/model.py
pattisdr: I haven't found any examples of where 'node' or 'project' is not in `params` but I will add the case if neither are present, to default to self.

website/project/model.py
pattisdr: This wasn't for optimization - it's been a while, let me remember why I did it!

pattisdr: The first case is `log.date.replace...` and second is self.logs[-1].date.replace...

My thinking was, if self.logs() equals one, then this add_log method created the first log.  So date_modified and date_created should be the same.  

In this example, I just created a project. `log.date` and `self.log.date` are different, the latter being rounded off.
screen shot 2016-02-23 at 3 56 53 pm
screen shot 2016-02-23 at 3 57 08 pm

If this is the first log, there should not be a difference between date_created and date_modified, hence the conditional.

pattisdr: But I guess dates are rounded when they're stored in the db, so date_created and date_modified would be rounded to the second(above), self.logs[-1].date?

I feel like I wouldn't have done this if I hadn't noticed a discrepancy somewhere between the two when there shouldn't have been, but maybe it was just inside this method, there was still a discrepancy before it was saved, and then both were rounded.
screen shot 2016-02-23 at 4 17 14 pm

screen shot 2016-02-23 at 4 17 14 pm

pattisdr: Sorry for the novel :) changed to `self.date_modified = log.date.replace(tzinfo=None)`

pattisdr: Without the check, causes test to fail: tests/test_models/test_date_modified_create_component

Picking up on discrepancy between date_created and date_modified.

tests/test_models.py
```diff
@@ -4012,40 +4019,6 @@ def test_formatted_date(self):
         unparsed = self.log.tz_date
         assert_equal(parsed, unparsed)
 
-    def test_resolve_node_same_as_self_node(self):
```
mfraezz: Could you replace these tests with ones that check `log.node` vs `log.original_node` behavior for forks and registrations?

scripts/migration/migrate_backrefs_between_nodes_and_nodelogs.py
```diff
@@ -0,0 +1,73 @@
...
+    count = 0
+    errored_logs = []
+    logger.info('Adding original_node and node fields to each log and then cloning the log for remaining nodes in the log backref.')
+
+    for log in node_logs:
+        # Two steps. Step 1: Add original_node and node field to log.
+        count += 1
+        original_node = lookup_node(log.params['node'])
+        log.original_node = original_node
+        log.node = lookup_node(log._backrefs['logged']['node']['logs'][0])
```
mfraezz: The script failed here with some old staging data --  the 1829th log had no backrefs. Further investigation on that outdated dump revealed that there were 46 of ~60000 broken in that way, and that they were all comment related. 

Michael checked prod for me:

```python
In [1]: models.NodeLog.find(Q('__backrefs', 'eq', None)).count()
Out[1]: 406
In [2]: bnls = models.NodeLog.find(Q('__backrefs', 'eq', None))

In [3]: set([x.action for x in bnls])
Out[3]:
{u'addon_added',
 u'comment_added',
 u'comment_removed',
 u'comment_updated',
 u'contributor_added',
 u'contributor_removed',
 u'edit_title',
 u'embargo_completed',
 u'file_added',
 u'file_removed',
 u'github_file_added',
 u'made_private',
 u'node_forked',
 u'osf_storage_file_added',
 u'wiki_updated'}
```

mfraezz:
Pass complete. :umbrella: Testing on a slightly more complicated DB revealed no issues, but staging broke the script, and prod will as well. Passing this off to RTM, @sloria can help resolve any issues with the script.

sloria:
Discussed in person with @mfraezz : please ensure that activity points remain accurate after this change.

pattisdr:
Thanks for pointing this out @sloria - I hadn't thought to check this.

Activity points are still calculated the same way before and after the change.  When `add_log` is called for a Node, `increment_user_activity_counters` runs.  This method increments the number of activity points by one.   We don't do a calculation of the number of logs so no changes are necessary.

website/project/model.py
mfraezz: Can confirm, definitely not for optimization. Disregard.

tests/test_models.py
pattisdr: Good idea. Tests added.

website/project/views/node.py
```diff
@@ -897,13 +897,8 @@ def _get_user_activity(node, auth, rescale_ratio):
...
     total_count = len(node.logs)
 
-    # Note: It's typically much faster to find logs of a given node
-    # attached to a given user using node.logs.find(...) than by
-    # loading the logs into Python and checking each one. However,
-    # using deep caching might be even faster down the road.
-
     if auth.user:
-        ua_count = node.logs.find(Q('user', 'eq', auth.user)).count()
+        ua_count = NodeLog.find(Q('node', 'eq', node._id) & Q('user', 'eq', auth.user)).count()
```
chrisseto: Could you add a compound index on `node` and `user`, please?

website/project/model.py
```diff
@@ -950,6 +937,11 @@ def pk(self):
         return self._id
 
     @property
+    def logs(self):
+        """ List of logs associated with this node"""
+        return list(NodeLog.find(Q('node', 'eq', self._id)).sort('date'))
```
chrisseto: This shouldn't be wrapped in `list`.
Calling `list` on a query object loads the entire query into memory, which will get very slow very fast.
returning just `NodeLog.find(...).sort('date')` won't load any logs into memory until they are actually being used and it supports a list like interface. (Minus negative indexing)
And calling `len` on `Node.logs` will run a count query rather than count the list that is loaded.

Negative indexing can be mocked but just return the query sorted on `'-date'`

website/project/views/node.py
pattisdr: I don't understand.  Do I add the compound index to node and user on the NodeLog model?

chrisseto: on `NodeLog`.
When doing a query on two keys, node and user in this case, mongo won't use individual indexes.
A compound index is required for optimizing this call.

```python
 __indices__ = [{
    'key_or_list': [
        ('user', 1),
        ('node', 1),
    ]
}]		
```
Should do the trick, but double check with robomongo or the mongo shell just to make sure.

pattisdr:
Have to take care of negative indexing now that node logs are not a list.

pattisdr:
@chrisseto, do you know why some of these tests are failing?  I removed the wrapping of node logs in a list - and it caused a lot of negative indexing issues, which I mostly fixed by doing `len(node.logs) - 1` (there's probably a better way to do this).  However, some of the tests, I cannot figure out why this change would have caused it.
screen shot 2016-03-09 at 8 59 17 am

chrisseto:
@pattisdr the tests are comparing the two mongo query sets, which is just an identity match of the object like the `is` keyword. In the tests you should wrap the logs in `list` which will evaluate it to a list of nodelog objects that can be compared.

For negative indexing you can just reverse the order of sorting. Which will only be one query to mongo as opposed to the `len`/`count` query and then the actual query.

If `node.logs[0]` is `NodeLog.find(Q('node', 'eq', node)).sort('date')`
then `node.logs[-1]` can be expressed as `NodeLog.find(Q('node', 'eq', node)).sort('-date')`

pattisdr:
@chrisseto, the node.logs[-1] happens hundreds of times in the tests. Is reversing the order of sorting something that should be happening in the tests?

chrisseto:
@pattisdr This should be added as a feature of modular-odm at some point this week.
@felliott or I will ping you when it's ready

pattisdr:
Being able to do node.logs[-1] if it is a query object?

chrisseto:
Yep

pattisdr:
Hooray!

felliott:
@pattisdr, @chrisseto: I've released modular-odm v0.3.0 with this change.

pattisdr:
Thank you @fitz!

On Thu, Mar 10, 2016 at 1:06 PM, Fitz Elliott <notifications@github.com>
wrote:

> @pattisdr <https://github.com/pattisdr>, @chrisseto
> <https://github.com/chrisseto>: I've released modular-odm v0.3.0 with
> this change.
>
> —
> <https://github.com/CenterForOpenScience/osf.io/pull/4950#issuecomment-194980459>
> .
>



-- 
Dawn Pattison
Developer
Center for Open Science
dawn@cos.io

pattisdr:
@chrisseto, tests are passing after your change - thanks

website/project/model.py
```diff
@@ -1843,12 +1843,12 @@ def get_aggregate_logs_query(self, auth):
...
                             for n in self.get_descendants_recursive()
                             if n.can_view(auth)]
-        query = Q('__backrefs.logged.node.logs', 'in', ids) & Q('should_hide', 'ne', True)
+        query = Q('node', 'in', ids) & Q('should_hide', 'ne', True)
         return query
 
     def get_aggregate_logs_queryset(self, auth):
         query = self.get_aggregate_logs_query(auth)
-        return NodeLog.find(query).sort('-_id')
+        return NodeLog.find(query).sort('-date')
```
icereval: we'll need to ensure an index on nodelog for [node, should_hide, -date]

website/project/model.py
```diff
@@ -1964,7 +1964,7 @@ def get_recent_logs(self, n=10):
 
         :param int n: Number of logs to retrieve
         """
-        return list(reversed(self.logs)[:n])
+        return list(reversed(self.logs))[:n]
```
icereval: are we sure we want to `list(...)` the entirety of `self.logs`? seems like it could be a big number...

icereval: won't `[:n]` inside the `list` tell MODM to apply a limit on the cursor/query?

website/project/model.py
```diff
@@ -2196,6 +2195,8 @@ def fork_node(self, auth, title='Fork of '):
             save=False
         )
 
+        forked.save()
```
icereval: for clarification, is the `save` here so we get a guid assigned before calling `add_log` and creating external references?

website/project/model.py
```diff
@@ -2208,7 +2209,13 @@ def fork_node(self, auth, title='Fork of '):
             save=False,
         )
 
-        forked.save()
+        # Clone each log from the original node for this fork.
+        logs = original.logs
+        for log in logs:
+            log.clone_node_log(forked._id)
+
+        forked.reload()
```
icereval: What does the reload do here?

website/project/model.py
```diff
@@ -2299,6 +2311,7 @@ def register_node(self, schema, auth, data, parent=None):
         registered.save()
 
         if settings.ENABLE_ARCHIVER:
+            registered.reload()
```
icereval: same as above, what does the `reload` do?

website/project/model.py
icereval: I hate to keep tagging on this line but I'm really not understanding why the date would originally be incorrect thus the `if/else` being necessary...? Also if this is needed please add a comment as this seems like its saying, if we have `1` log, grab it and get the date, if we have more than one log just grab the last one and use its date, which is the same as if we have one log right... :confused:

website/project/model.py
chrisseto: `reversed` should be used here either, ODM doesn't support it AFAIK.
`self.logs.sort('-date')[:n]` should do the trick here

website/project/model.py
chrisseto: Without digging too much into what is happening `datetime` discrepancies are due to mongo having slightly less precision than python does. When a model if first created it will have the milliseconds attached to the `datetime` but after being saved and loaded from mongo the milliseconds have been dropped. ODM should probably trim datetime at the python level just for consistency. I know we have a decent amount of tests that look weird due to this case.

icereval: log = newest log written, nevermind then, its all so clear now! i think....

website/project/model.py
pattisdr: Yes, without the forked.save()  modularodm raises a `DatabaseError: Record must be loaded`.

website/project/model.py
pattisdr: Without this, modularodm raises a `KeyExistsException` during the creation of a registration and fails.  For some reason, removing the backrefs necessitates reloading the registration and fork during creation.  I don't know enough about modularodm to know why.  I remember someone telling me that removing backrefs on some of the other items had encountered this same issue as well.

website/project/model.py
pattisdr: See below.

website/project/model.py
pattisdr: I don't know what this means.  I already have these indices - 
```
    __indices__ = [
        {
            'key_or_list': [
                ('user', 1),
                ('node', 1)
            ],
        }
    ]
```
Can I add them here?
```
    __indices__ = [
        {
            'key_or_list': [
                ('user', 1),
                ('node', 1),
                ('should_hide', 1),
                ('-date', 1)

            ],
        }
    ]
```

pattisdr:
Looking into failing tests - related to last commit https://github.com/CenterForOpenScience/osf.io/pull/4950/commits/8569cc847debca3be9b8b86766b27dcf5a7fc493

api/logs/permissions.py
```diff
@@ -10,13 +10,4 @@ class ContributorOrPublicForLogs(permissions.BasePermission):
 
     def has_object_permission(self, request, view, obj):
         assert isinstance(obj, (NodeLog)), 'obj must be a NodeLog, got {}'.format(obj)
```
chrisseto: Small, but could you remove the extra `()` around `NodeLog`?

scripts/migration/migrate_backrefs_between_nodes_and_nodelogs.py
```diff
@@ -0,0 +1,70 @@
...
+                {'$set': {
+                    'original_node': log['params']['node'],
+                    'node': log['__backrefs']['logged']['node']['logs'][0]
+                    }
+                },
+                upsert=True
+            )
+            logger.info('{}/{} Log {} updated'.format(count, total_log_count, log['_id']))
+        except KeyError as error:
+            logger.error('Could not migrate nodelog due to error -- likely a lack of __backrefs')
```
chrisseto: Do have have anyway of migrating/handling these failures? Or are they not failures?

scripts/migration/migrate_backrefs_between_nodes_and_nodelogs.py
```diff
@@ -0,0 +1,70 @@
...
+            logger.exception(error)
+            pass
```
chrisseto: This pass is unnecessary

tests/test_models.py
```diff
@@ -3162,12 +3162,14 @@ def test_no_parent(self):
...
         # Add some logs
         for _ in range(5):
-            self.project.logs.append(NodeLogFactory())
+            self.project.add_log('file_added', params={'node': self.project._id}, auth=self.auth)
+
         # Expected logs appears
         assert_equal(
             self.project.get_recent_logs(3),
-            list(reversed(self.project.logs)[:3])
+            list(reversed(self.project.logs))[:3]
```
chrisseto: This should be leveraging ODM's sorting and slicing.
`list(self.project.logs.sort('date')[:3])`

tests/test_models.py
```diff
@@ -3162,12 +3162,14 @@ def test_no_parent(self):
...
         )
+
         assert_equal(
             self.project.get_recent_logs(),
             list(reversed(self.project.logs))
```
chrisseto: Same as the above comment

tests/test_serializers.py
```diff
@@ -178,9 +178,12 @@ class TestNodeLogSerializers(OsfTestCase):
 
     def test_serialize_log(self):
         node = NodeFactory(category='hypothesis')
-        log = NodeLogFactory(params={'node': node._primary_key})
-        node.logs.append(log)
         node.save()
+        log = NodeLogFactory(
+                params={'node': node._primary_key},
```
chrisseto: Just for consistency could you mind changing this to `node._id`? `_primary_key` is an alias to `_id` for internal use in ODM, which is why it's prefixed with a `_`.  `_id` is prefixed because mongodb has decided to prefix it.

scripts/migration/migrate_backrefs_between_nodes_and_nodelogs.py
pattisdr: I don't know the best way to handle this.  It works fine on local data - but there might be broken logs in production/staging.  There should be a nodes on a log's backrefs - could the logs without backrefs be recorded as this script is running and then those logs handled individually?

website/project/model.py
```diff
@@ -460,13 +463,20 @@ def __repr__(self):
...
-            Node.load(self.params.get('node')) or
-            Node.load(self.params.get('project'))
-        )
+    def clone_node_log(self, node_id):
+        """
+        When a node is forked or registered, all logs on the node need to be cloned for the fork or registration.
+        :param node_id:
+        :return: cloned log
+        """
+        original_log = self.load(self._primary_key)
```
chrisseto: `self._id`

website/project/model.py
```diff
@@ -460,13 +463,20 @@ def __repr__(self):
...
+        node = Node.find(Q('_id', 'eq', node_id))[0]
+        log_clone = original_log.clone()
+        log_clone.node = node
+        log_clone.original_node = original_log.original_node
```
chrisseto: Is this and the line below not handled by `.clone()`?

website/project/model.py
```diff
@@ -486,26 +496,10 @@ def formatted_date(self):
...
-        if node.is_fork_of(self.node) or node.is_registration_of(self.node):
-            return node
-        for child in node.nodes:
-            if child.is_fork_of(self.node) or node.is_registration_of(self.node):
-                return child
-        return False
-
     def can_view(self, node, auth):
-        node_to_check = self.resolve_node(node)
+        node_to_check = node
```
chrisseto: Would you mind removing the `node_to_check` variable? It appears to no longer serve a purpose

website/project/model.py
```diff
@@ -823,8 +817,7 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin):
     contributors = fields.ForeignField('user', list=True)
     users_watching_node = fields.ForeignField('user', list=True)
 
-    logs = fields.ForeignField('nodelog', list=True, backref='logged')
-    tags = fields.ForeignField('tag', list=True)
+    tags = fields.ForeignField('tag', list=True, backref='tagged')
```
chrisseto: I'm not sure if the diff is being weird or something, but the `tagged` backref was removed recently. Could you go ahead and remove it from this PR?

website/project/model.py
chrisseto: Please leave a comment explaining why this line is necessary

pattisdr: See above? https://github.com/CenterForOpenScience/osf.io/pull/4950#discussion_r53844090

website/project/model.py
chrisseto: They go into another element of the `__indices__` list
```python
    __indices__ = [{
            'key_or_list': [
                ('user', 1),
                ('node', 1)
            ],
        }, {
            'key_or_list': [
                ('node', 1),
                ('should_hide', 1),
                ('date', -1)  # Reverse indices are specified with a -1 or pymongo.DECENDING. 1 is equivilent to pymongo.ASCENDING
           ]
        }]
```

website/project/model.py
```diff
@@ -880,6 +875,11 @@ def pk(self):
         return self._id
 
     @property
+    def logs(self):
+        """ List of logs associated with this node"""
+        return NodeLog.find(Q('node', 'eq', self._id)).sort('date')
```
chrisseto: Should these be sorted on date descending?

As is `node.logs[0]` will be the node creation log and `node.logs[-1]` will be the latest. Is that the current behavior/desired?

website/static/js/_allLogTexts.json
```diff
@@ -0,0 +1 @@
...ected the Amazon S3 bucket ${bucket} in ${node}","s3_file_added":"${user} added file ${path} to Amazon S3 bucket ${bucket} in ${node}","s3_file_removed":"${user} removed ${path} in Amazon S3 bucket ${bucket} in ${node}","s3_file_updated":"${user} updated ${path} in Amazon S3 bucket ${bucket} in ${node}","s3_folder_created":"${user} created folder ${path} in Amazon S3 in ${node}","s3_node_authorized":"${user} authorized the Amazon S3 addon for ${node}","s3_node_deauthorized":"${user} deauthorized the Amazon S3 addon for ${node}","googledrive_file_added":"${user} added file ${googledrive_path} to Google Drive in ${node}","googledrive_file_removed":"${user} removed file ${googledrive_path} from Google Drive in ${node}","googledrive_file_updated":"${user} updated file ${googledrive_path} in Google Drive in ${node}","googledrive_folder_created":"${user} created folder ${googledrive_path} in Google Drive in ${node}","googledrive_folder_selected":"${user} linked Google Drive folder ${googledrive_folder} to ${node}","googledrive_node_authorized":"${user} authorized the Google Drive addon for ${node}","googledrive_node_deauthorized":"${user} deauthorized the Google Drive addon for ${node}"}
```
chrisseto: I'm fairly certain this is a generated file and should not be in version control.
@samchrisinger might know?

website/views.py
```diff
@@ -164,7 +164,7 @@ def serialize_log(node_log, auth=None, anonymous=False):
         'action': node_log.action,
         'params': sanitize.unescape_entities(node_log.params),
         'date': utils.iso8601format(node_log.date),
-        'node': node_log.node.serialize(auth) if node_log.node else None,
```
chrisseto: Why is this no longer `.node`?

website/static/js/_allLogTexts.json
pattisdr: I will take it out - you are right it should not be in there.

pattisdr:
@chrisseto, starting to look at your comments now.

website/project/model.py
pattisdr: ohh.  thank you.

website/project/model.py
pattisdr: The clone does not appear to have the original_node and user set.

pattisdr: screen shot 2016-04-07 at 11 36 55 am

website/project/model.py
pattisdr: Logs was formerly a field on the Node model, and it was an array. When `node.add_log` was run, it appended the new log onto the end of the logs list.  So oldest logs first, newest logs last.  I was matching previous behavior.

website/project/model.py
chrisseto: What the `.to_storage()` of the clone before those lines?

website/views.py
pattisdr: `original_node` is the node the log was about (formerly the first node on the log backrefs) and `node` is where the log sits currently. `original_node` can be equal to node when the action happened to node A and I am viewing node A's logs but if I am looking at a registration or fork, many times they will be different.

For example, I create a **Node**.  I now have one log, `project_created`.  I then register the node, **Registration**

`project_created` log on **Node**
- original_node =  **Node**
- node =  **Node**

`project_created` log on **Registration**
- original_node =  **Node**
- node = **Registration**

On the front end for the registration, the logs need to refer to the original node.  It's not going to be **Registration** was created, but **Node**. The original node.

website/project/model.py
pattisdr: screen shot 2016-04-07 at 1 34 27 pm

chrisseto:
Regarding the cloning comment from earlier https://github.com/CenterForOpenScience/modular-odm/blob/8dca4a04f9fca20b509617a14652b42ac1537548/modularodm/storedobject.py#L314

pattisdr:
I see some more changes I need to make @chrisseto - I wrote this in January before I worked on the dashboard, and this eliminates the `nodes` relationship on the logs endpoint, in favor of a `node` relationship.

PR #4951, opened 2016-01-27
[Feature] Admin able to Mark Spam

Ghalko:
## Proposal

Needs #4527. Enables the confirming of Spam or Ham by the Admin. This interacts with the OSF db. ADM-26

## Repercussions

Admins with power.


whit537:
> Needs #4527.

It has it! That PR is merged. :-) Do we need a rebase here?

whit537:
Alright, that's all I got in me tonight (spent most of it on #5040 ;). I will plan to pick up with this on Monday! :)

admin/spam/views.py
```diff
@@ -1,47 +1,106 @@
...
+    def get_queryset(self):
+        self.status = self.request.GET.get('status', u'1')
+        query = (
+            Q('reports', 'ne', {}) &
+            Q('reports', 'ne', None) &
+            Q('spam_status', 'eq', int(self.status))
+        )
+        return Comment.find(query).sort(self.ordering)
 
+    def get_context_data(self, **kwargs):
```
cwisecarver: You generally want to call super in any CBV defined method unless you have a specific reason not to. There's so much multiple inheritance that you could lose functionality that you never even knew was included.

In this situation you probably want to do whatever you're doing that's different than the base method in `MultipleObjectMixin` and `ContextMixin` and then splat your context to the super call like they do at the end of `MultipleObjectMixin.get_context_data()` https://ccbv.co.uk/projects/Django/1.9/django.views.generic.list/ListView/#get_context_data-MultipleObjectMixin

admin/spam/views.py
```diff
@@ -1,47 +1,106 @@
...
+            return render(request, self.template_name, context=context)
+        # return super(SpamDetail, self).post(request, *args, **kwargs)
+
+    def get_context_data(self, **kwargs):
+        self.spam_id = kwargs['spam_id']
+        self.item = Comment.load(self.spam_id)
+        kwargs = super(SpamDetail, self).get_context_data(**kwargs)
+        try:
+            kwargs['comment'] = serialize_comment(self.item)
+        except AttributeError:
```
cwisecarver: You shouldn't need to catch this and then rethrow, should you? It should just bubble.

admin/spam/views.py
```diff
@@ -1,47 +1,106 @@
...
+    @method_decorator(login_required)
+    def get(self, request, *args, **kwargs):
+        try:
+            context = self.get_context_data(**kwargs)
+        except AttributeError:
+            return page_not_found(request)  # TODO: 1.9 update to have exception with node/user 404.html will be added
+        self.page = request.GET.get('page', 1)
+        context['page_number'] = self.page
+        context['form'] = self.get_form()
+        return self.render_to_response(context)
```
cwisecarver: Again, unless you're trying to avoid some specific behavior it's good to return super() from CBV methods.

admin/spam/views.py
```diff
@@ -1,47 +1,106 @@
...
+
+    @method_decorator(login_required)
+    def post(self, request, *args, **kwargs):
```
cwisecarver: The super of `post()` calls `self.get_form()` and `form.is_valid()` for you. You should, if you can, be setting `context[]` things in the `get_context_data` method if at all possible.

cwisecarver:
@Ghalko First CR pass done. Swing by if there are questions or problems. Sorry for the massive delay.

admin/spam/views.py
Ghalko: Called super.

admin/spam/views.py
Ghalko: Bubbles.

admin/spam/views.py
Ghalko: Refactored

admin/spam/views.py
Ghalko: Refactored

Ghalko:
@cwisecarver Finished response pass. This was quite old, funny to see how far my understanding of CBV has matured over the last month. It will probably break several of my subsequent PRs, but all for the betterment. The faster we turn this around, the better the other PRs are going to look.

cwisecarver:
Other than the last comment everything looks good. I did a quick functional test too and it appears to be functional. :shipit: Sorry again for the delays @Ghalko.

sloria:
@Ghalko Can you please update this PR with develop? That should fix the build.

PR #4958, opened 2016-01-28
[RFR] Add Contributors to Multiple Components Dialog

billyhunt:
Jira ticket is 
https://openscience.atlassian.net/browse/OSF-5054

Purpose
-----------

Update the Add Contributors node selection dialog.

Current
----------
screen shot 2016-01-28 at 5 22 29 pm

Proposed
-------------

screen shot 2016-01-28 at 5 24 42 pm


sloria:
@billyhunt Can you link the JIRA ticket for this feature in the PR description please?

website/static/js/contribAdder.js
```diff
@@ -377,6 +432,42 @@ var AddContributorViewModel = oop.extend(Paginator, {
...
+        return $.ajax({
+            url: treebeardUrl,
+            type: 'GET',
+            dataType: 'json'
+        }).done(function(response) {
+            self.nodesOriginal = $osf.getNodesOriginal(response[0], self.nodesOriginal);
+            Object.size = function(obj) {
+                var size = 0, key;
+                for (key in obj) {
+                    if (obj.hasOwnProperty(key)) size++;
```
samanehsan: This if statement is missing a `{`

website/static/js/contribAdder.js
```diff
@@ -306,6 +332,35 @@ var AddContributorViewModel = oop.extend(Paginator, {
         }
         return false;
     },
+    selectAllNodes: function() {
+        //select all nodes to add a contributor to.  THe changed variable is set here for timing between
+        // treebeard and knockout
+        var self = this;
+        var nodesStateLocal = ko.toJS(self.nodesState());
```
samanehsan: I'm not too sure what `nodesStateLocal` means. Could this be named `nodesState` instead?

website/static/js/contribAdder.js
samanehsan: Will this always evaluate to true? Instead, could you check the length of `self.nodesOriginal` to determine whether the node has any children?

website/static/js/osfHelpers.js
```diff
@@ -569,6 +569,48 @@ function humanFileSize(bytes, si) {
...
+    for (i=0; i < nodeTree.node.contributors.length; i++) {
+        if (nodeTree.node.contributors[i].is_admin) {
+            adminContributors.push(nodeTree.node.contributors[i].id);
+        }
+        if (nodeTree.node.contributors[i].is_confirmed) {
+            registeredContributors.push(nodeTree.node.contributors[i].id);
+        }
+    }
+    nodesOriginal[nodeId] = {
+        public: nodeTree.node.is_public,
```
samanehsan: This key should be called `isPublic` since the value is a boolean

website/templates/project/modal_add_contributor.mako
```diff
@@ -200,28 +198,27 @@
...
+                    <div class="osf-treebeard">
+                        <div id="addContributorsTreebeard">
+                            <div class="spinner-loading-wrapper">
+                                <div class="logo-spin logo-md"></div>
+                                <p class="m-t-sm fg-load-message"> Loading projects and components...  </p>
                             </div>
                         </div>
-
+                        <div class="help-block" style="padding-left: 15px">
+                            <p id="configureNotificationsMessage"></p>
```
samanehsan: Is this message block being used?

website/project/views/node.py
```diff
@@ -781,6 +784,7 @@ def _view_project(node, auth, primary=False):
             },
             'alternative_citations': [citation.to_json() for citation in node.alternative_citations],
             'has_draft_registrations': node.has_active_draft_registrations,
+            'contributors': contributor_ids
```
samanehsan: Where is this being used? I looked around the code but wasn't able to figure out why this was added in.

billyhunt: I am pretty sure its used in another PR (removing contributors) that used to be a part of this one.  I will pull it.

billyhunt:
Hi @samanehsan, I think I fixed all of your issues.  Want to check it out again?

sloria:
@billyhunt Can you please get this up to date with develop? Your "remove" and "inherit" PRs have been merged, and I assume this will need some work to integrate this PR with those.

samanehsan:
Thanks for making those changes @billyhunt! :palm_tree: Passing this off to @sloria :octocat:

billyhunt:
@sloria Should be good to go.  :facepunch: :dragon: :8ball:

website/static/js/osfHelpers.js
```diff
@@ -569,6 +569,48 @@ function humanFileSize(bytes, si) {
     return bytes.toFixed(1) + ' ' + units[u];
 }
 
+function getNodesOriginal(nodeTree, nodesOriginal) {
```
chennan47: This function is the same as https://github.com/CenterForOpenScience/osf.io/pull/4958/files#diff-0edcea532119f03b4ad90b86dc5ba654R618. Why added it again?

billyhunt: Good catch.  It wasn't merged correctly.  Whoops.  I pushed an update.

website/static/js/contribAdder.js
```diff
@@ -12,6 +12,8 @@ var oop = require('js/oop');
 var $osf = require('js/osfHelpers');
 var osfLanguage = require('js/osfLanguage');
 var Paginator = require('js/paginator');
+var NodeSelectTreebeard = require('js/nodeSelectTreebeard');
+var m = require('mithril');
 
 var NODE_OFFSET = 25;
```
chennan47: This NODE_OFFSET is used for indent original component tree. Since you are using the new one, this is no longer necessary.

chennan47:
Thanks for the fix @billyhunt. Pass done. Look good, @sloria.

website/static/js/contribAdder.js
```diff
@@ -60,10 +79,17 @@ var AddContributorViewModel = oop.extend(Paginator, {
         self.results = ko.observableArray([]);
         self.contributors = ko.observableArray([]);
         self.selection = ko.observableArray();
+        self.contributorIDsToAdd = ko.computed(function() {
+            var contributorIDs = [];
```
sloria: Minor: when building an Array based on values from another Array,  use `.map` rather than a for loop. Also, this computed can be a `pureComputed`.

website/static/js/contribAdder.js
```diff
@@ -306,6 +330,35 @@ var AddContributorViewModel = oop.extend(Paginator, {
...
+        //select no nodes to add a contributor to.  THe changed variable is set here for timing between
+        // treebeard and knockout
+        var self = this;
+        var nodesState = ko.toJS(self.nodesState());
+        for (var key in nodesState) {
+            if (nodesState[key].enabled && nodesState[key].checked) {
+                nodesState[key].checked = false;
+                nodesState[key].changed = nodesState[key].checked !== self.nodesOriginal[key].checked;
+            }
+            m.redraw(true);
```
sloria: The `redraw` in this loop looks expensive and unnecessary. Can this be removed?

website/static/js/contribAdder.js
```diff
@@ -40,6 +40,25 @@ var AddContributorViewModel = oop.extend(Paginator, {
...
+                newValue[key].changed = newValue[key].checked !== self.nodesOriginal[key].checked;
+                if (newValue[key].changed) {
+                    nodesToChange.push(key);
+                }
+            }
+            self.nodesToChange(nodesToChange);
+            m.redraw(true);
+        });
+
+        self.hasChildren = ko.observable(false);
```
sloria: `hasChildren` is derived from `nodesOriginal`, so it should probably be a `computed` rather than an `observable`.

website/static/js/contribAdder.js
```diff
@@ -40,6 +40,25 @@ var AddContributorViewModel = oop.extend(Paginator, {
         self.parentTitle = parentTitle;
         self.async = options.async || false;
         self.callback = options.callback || function() {};
+        self.nodesOriginal = {};
+        //state of current nodes
+        self.nodesToChange = ko.observableArray();
+        self.nodesState = ko.observableArray();
```
sloria: Shouldn't `nodesState` be an `observable` rather than an `observableArray`? The code treats it like an object rather than an Array.

sloria:
Pass finished.

billyhunt:
This should be good to go @sloria

website/static/js/contribAdder.js
```diff
@@ -60,11 +86,18 @@ var AddContributorViewModel = oop.extend(Paginator, {
         self.results = ko.observableArray([]);
         self.contributors = ko.observableArray([]);
         self.selection = ko.observableArray();
+
+        self.contributorIDsToAdd = ko.pureComputed(function() {
+            var selected_ids = self.selection().map(function(user){
```
sloria: As per COSDev, use `lowerCamelCase` for variable names. That said, this variable probably isn't necessary; just return the return value of `.map`.

website/static/js/contribAdder.js
```diff
@@ -40,6 +40,32 @@ var AddContributorViewModel = oop.extend(Paginator, {
...
+                if (newValue[key].changed) {
+                    nodesToChange.push(key);
+                }
+            }
+            self.nodesToChange(nodesToChange);
+            m.redraw(true);
+        });
+
+        self.hasChildren = ko.pureComputed(function() {
+            var index = 0;
```
sloria: You should be able to simplify this function to

```js
Object.keys(self.nodesOriginal).length;
```

website/static/js/contribAdder.js
```diff
@@ -313,6 +344,33 @@ var AddContributorViewModel = oop.extend(Paginator, {
...
+        var self = this;
+        var nodesState = ko.toJS(self.nodesState());
+        for (var key in nodesState) {
+            if (nodesState[key].enabled && nodesState[key].checked) {
+                nodesState[key].checked = false;
+                nodesState[key].changed = nodesState[key].checked !== self.nodesOriginal[key].checked;
+            }
+        }
+        self.nodesState(nodesState);
+        m.redraw(true);
```
sloria: Is this redraw necessary, given that you have a subscriber function on `nodesState` that calls `m.redraw`?

sloria:
Pausing review here. A few more drive-by comments above; will finish up review later today or in the morning.

website/static/js/contribAdder.js
```diff
@@ -60,68 +79,96 @@ var AddContributorViewModel = oop.extend(Paginator, {
...
+
         self.notification = ko.observable('');
         self.inviteError = ko.observable('');
         self.doneSearching = ko.observable(false);
         self.totalPages = ko.observable(0);
-        self.nodes = ko.observableArray([]);
         self.nodesToChange = ko.observableArray();
 
-        self.foundResults = ko.pureComputed(function() {
+        self.foundResults = ko.pureComputed(function () {
```
sloria: Minor: no need to add these extra spaces; they only add noise to the diffs. That said, don't feel you need to remove them at this point.

website/static/js/nodeSelectTreebeard.js
```diff
@@ -0,0 +1,116 @@
+'use strict';
+
+var $ = require('jquery');
+var m = require('mithril');
+var ko = require('knockout');
+var Treebeard = require('treebeard');
+var projectSettingsTreebeardBase = require('js/projectSettingsTreebeardBase');
+
+
+function expandOnLoad() {
```
sloria: The three functions below are copied in 3 different places. See if you can refactor. How about putting these functions in `projectSettingsTreebeardBase`?

website/static/js/osfHelpers.js
```diff
@@ -590,14 +590,18 @@ function getNodesOriginal(nodeTree, nodesOriginal) {
         }
     }
     nodesOriginal[nodeId] = {
-        public: nodeTree.node.is_public,
+        isPublic: nodeTree.node.is_public,
```
sloria: This `getNodesOriginal` should probably go in `projectSettingsTreebeardBase` along with the other shared utility functions (see https://github.com/CenterForOpenScience/osf.io/pull/4958/files#r55267144).

website/static/js/contribAdder.js
```diff
@@ -313,13 +336,39 @@ var AddContributorViewModel = oop.extend(Paginator, {
...
-    submit: function() {
+    selectAllNodes: function () {
+        //select all nodes to add a contributor to.  THe changed variable is set here for timing between
+        // treebeard and knockout
+        var self = this;
+        var nodesState = ko.toJS(self.nodesState());
+        for (var key in nodesState) {
+            if (nodesState[key].enabled) {
+                nodesState[key].checked = true;
+                nodesState[key].changed = nodesState[key].checked !== self.nodesOriginal[key].checked;
```
sloria: Is it necessary to set `changed` here, given that the subscriber will do that here: https://github.com/CenterForOpenScience/osf.io/pull/4958/files#diff-8bccb237dc02931ca4e0314ade877143R54 ?

website/static/js/contribAdder.js
```diff
@@ -373,17 +422,44 @@ var AddContributorViewModel = oop.extend(Paginator, {
...
         var response = JSON.parse(xhr.responseText);
         // Update error message
         this.inviteError(response.message);
+    },
+    /**
+     * get node tree for treebeard from API V1
+     */
+    fetchNodeTree: function () {
+        var self = this;
+        var treebeardUrl = window.contextVars.node.urls.api + 'tree/';
```
sloria: Minor: we generally try to avoid using `contextVars` in re-usable modules because they make the modules more difficult to test (you'd need to set `window.contextVars.node...` before instantiating this class). This usage is a little smelly, but not a dealbreaker.

sloria:
Pass finished.

website/static/js/contribAdder.js
billyhunt: What is the preferred way to set an url @sloria?

website/static/js/contribAdder.js
```diff
@@ -373,17 +428,43 @@ var AddContributorViewModel = oop.extend(Paginator, {
...
+     */
+    fetchNodeTree: function (treebeardUrl) {
+        var self = this;
+        return $.ajax({
+            url: treebeardUrl,
+            type: 'GET',
+            dataType: 'json'
+        }).done(function (response) {
+            self.nodesOriginal = projectSettingsTreebeardBase.getNodesOriginal(response[0], self.nodesOriginal);
+            self.hasChildren = (Object.keys(self.nodesOriginal).length > 1);
```
sloria: This attribute doesn't seem to be initialized anywhere. That said, wouldn't it make this a method rather than an attribute?

```js
self.hasChildren = function() { Object.keys(self.nodesOriginal).length > 1); };
```

sloria:
Pass finished. Once https://github.com/CenterForOpenScience/osf.io/pull/4958#discussion_r55998878 is addressed, this should be good to merge.

PR #4959, opened 2016-01-29
[Feature] Admin OSF User Info

Ghalko:
## Proposal

Display basic info about users and their nodes.

## Repercussions

Does not update any information.


Ghalko:
Close in favor of #4960

PR #4960, opened 2016-01-29
[Feature] Admin Node and User Info

Ghalko:
Replaces #4959 

## Proposal

Add basic info for users and nodes ADM-3, ADM-56 (somewhat intertwined)

## Admin Release Notes

Unlike many other of the admin PRs that I have recently pushed, this one is not tied to previous PRs.


admin/abstract/views.py
```diff
@@ -0,0 +1,49 @@
...
+        self.guid = None
+        super(GuidFormView, self).__init__()
+
+    def get(self, request, *args, **kwargs):
+        self.guid = request.GET.get('guid', None)
+        if self.guid is not None:
+            try:
+                guid_object = self.get_guid_object()
+            except (AttributeError, TypeError):
+                return HttpResponseNotFound(
```
cwisecarver: This could be a bit more user friendly :smiling_imp:

https://docs.djangoproject.com/en/1.9/ref/urls/#handler404 Easy way to create a global django 404 handler.

Ghalko: screen shot 2016-02-01 at 10 45 47 am

Less friendly I think.

Maybe I should go for a signature more like: `admin/nodes/whup2` rather than using the `?guid=`

cwisecarver: You on the right branch? I get the following:
screen shot 2016-02-01 at 12 59 44 pm

admin/nodes/serializers.py
```diff
@@ -0,0 +1,35 @@
...
+
+def parse_permissions(permissions):
+    if len(permissions) == 3:
+        return 'Admin'
+    elif len(permissions) == 2:
+        return 'Write'
+    elif len(permissions) == 1:
+        return 'Read'
+    else:
+        return 'Unknown'
```
cwisecarver: Unknown and permissions in the same sentence gives me the willies. This is just for display, correct? We're not actually using the string value for logic later on down the line? So long as that's the case it's probably fine. I would just hate to encourage string matching.

admin/abstract/views.py
Ghalko: I tried implementing the `page_not_found` suggestion.

admin/templates/users/user.html
```diff
@@ -0,0 +1,78 @@
...
+                        <th>Name</th>
+                        <th>Public</th>
+                        <th># of Contributors</th>
+                    </tr>
+                </thead>
+                <tbody>
+                {% for node in user.nodes %}
+                    <tr>
+                        <td>
+                            <a href="{% url 'nodes:node' %}?guid={{ node.id }}">
```
cwisecarver: If you get to the point where you don't want to type this pattern out anymore django has the concept of both filters and template tags. https://docs.djangoproject.com/en/1.9/howto/custom-template-tags/#custom-template-tags-and-filters Either one would probably work fine in this circumstance, really up to you which you pick. Filters are normally used for smaller transmogrifying text type operations but that isn't a requirement.

Since this is practically the same as your success_url method on the view forms you could define a `reverse_node` and `reverse_user` that took an id and then write two wrapping functions and register them as filters. Or just register the functions as filters and have the success_url methods call those.

This is one of those things that won't really matter until you change your url pattern as you suggested in https://github.com/CenterForOpenScience/osf.io/pull/4960/files#r51434390 and then you are happy you did it.

admin/nodes/serializers.py
Ghalko: This is for just info's sake at this point. The problem is that permissions are a list: `['read'], ['read', 'write'], ['read', 'write', 'admin']`. I guess I will use this: `website.util.permissions` to reduce and test.

admin/abstract/views.py
cwisecarver: Ah, okay. By default, implementing `page_not_found` will just send back a standard 404 with a default message. https://docs.djangoproject.com/en/1.9/ref/views/#django.views.defaults.page_not_found

If you create a 404.html in the root of your templates directory, it will be loaded and you should have the exception raised in the template context. 

Alternatively, you can set handler404 to any FBV that you like and it will be called when a 404 exception is raised. I believe it will be passed 404.html as it's template, just like page_not_found is.

The easiest (and probably most sustainable) thing to do would be to make a 404.html that extends the base layout like the rest of the templates do and displays the exception message and then switch back to `page_not_found`.

admin/nodes/serializers.py
cwisecarver: It's probably fine if it's just for display.

cwisecarver:
Pass done @Ghalko. @cwisecarver for questions or comments.

admin/nodes/serializers.py
Ghalko: Changed to use `reduce_permissions`

admin/templates/users/user.html
Ghalko: Complete. Might still change the signature, but now only have to find one instance of each.

admin/abstract/views.py
Ghalko: Added 404.html. Upgraded to django 1.9 to take advantage of the better `page_not_found` signature.

Ghalko:
Not going to work without upgrading API to 1.9 as well.

.gitignore
```diff
@@ -114,4 +114,5 @@ admin/static/public/
 admin/webpack-stats.json
 # bower
 admin/static/vendor/
-admin/db.sqlite3
```
cwisecarver: This is a good idea. This will make you find errors with your migrations much faster.

admin/base/views.py
```diff
@@ -28,3 +34,44 @@ def password_reset_confirm_custom(request, **kwargs):
...
+
+    def get_context_data(self, **kwargs):
+        self.guid = self.request.GET.get('guid', None)
+        if self.guid is not None:
+            try:
+                guid_object = self.get_guid_object()
+            except AttributeError:
+                raise
+        else:
+            guid_object = None
```
cwisecarver: Just a style thing, I would set `guid_object = None` at the top of the method and lose the else. Then it's None by default and it'll not be None if get_guid_object works.

admin/pre_reg/forms.py
```diff
@@ -36,3 +36,10 @@ class DraftRegistrationForm(forms.Form):
         ),
         required=False
     )
+
+    def __init__(self, *args, **kwargs):
+        prereg_reviewers = itertools.chain(
```
cwisecarver: This should fix the bug that @zkraime was running into trying to query the users table before it existed.

admin_tests/pre_reg/test_views.py
```diff
@@ -1,110 +1,110 @@
...
+#             non_prereg_not_pending.append(draft)
+#
+#         url = reverse('pre_reg:prereg')
+#         res = self.client.get(url)
+#
+#         assert_equal(res.status_code, http.OK)
+#         assert_in('drafts', res.context)
+#         assert_equal(len(res.context['drafts']), len(prereg_pending))
+#         for draft in res.context['drafts']:
+#             assert_in(draft['pk'], [d._id for d in prereg_pending])
```
cwisecarver: I'm guessing @sloria would say say that we've got source control and you could delete this until you're ready to enable it and then pull it out of the history. He's not wrong :wink:

admin/users/serializers.py
```diff
@@ -0,0 +1,22 @@
+"""
+Serialize user
+"""
+# from admin.nodes.serializers import serialize_simple_user
+
+
+def serialize_user(user):
+    return {
+        'name': user.fullname,
+        'nodes': map(serialize_simple_node, user.contributor_to),
```
cwisecarver: Just FYI, with the backref removal user.contributor_to calls User.contributed, which does a query of all Nodes with the user id in their contributors ForeignField(list=True), then it runs through that entire list checking is_deleted and is_dashboard, then you're looping through it again with map. Not saying you shouldn't, just saying this could be a possible bottleneck/performance optimization opportunity.

tests/base.py
```diff
@@ -335,7 +335,9 @@ def setUp(self):
 
 
 class AdminTestCase(DbTestCase, AdminAppTestCase, UploadTestCase, MockRequestTestCase):
-    pass
+    def setUp(self):
```
cwisecarver: Wouldn't pass accomplish the same thing with less LOC?

admin/nodes/templatetags/node_extras.py
```diff
@@ -0,0 +1,9 @@
+from django import template
+from django.core.urlresolvers import reverse
+
+register = template.Library()
+
+
+@register.filter
+def reverse_node(value):
+    return reverse('nodes:node') + '?guid={}'.format(value)
```
cwisecarver: You could do ` return '{}?guid={}'.format(reverse('nodes:node'), value)` instead of using the dreaded plus sign.

admin/users/serializers.py
Ghalko: Unfortunate timing certainly, since this is a major way of being able to confirm user identity.

cwisecarver:
@Ghalko Done. Let me know if there are questions.

admin/base/views.py
Ghalko: Done.

.gitignore
Ghalko: @cwisecarver Not sure if this is an endorsement or a criticism.

admin_tests/pre_reg/test_views.py
Ghalko: Deleted

.gitignore
cwisecarver: Endorsement. I try to make sure I include smileys whenever I include sarcasm.

tests/base.py
Ghalko: Done.

admin/nodes/templatetags/node_extras.py
Ghalko: Done in nodes and users.

admin/base/views.py
```diff
@@ -28,3 +34,43 @@ def password_reset_confirm_custom(request, **kwargs):
...
+        raise NotImplementedError
+
+    def __init__(self):
+        self.guid = None
+        super(GuidFormView, self).__init__()
+
+    def get(self, request, *args, **kwargs):
+        try:
+            return super(GuidFormView, self).get(request, args, kwargs)
+        except AttributeError:
```
sloria: This will swallow AttributeErrors that might not be related to fetching GUIDs, leading to bugs that are hard to find. Why do you expect an `AttributeError` here? https://github.com/CenterForOpenScience/osf.io/pull/4960/files#diff-d0d8d8e2e4760424d31eacb32ecdeb29R63 

It would make sense for concrete implementations of `get_guid_object` to raise exceptions or return an error response rather than handling it in this class.

admin/nodes/serializers.py
```diff
@@ -0,0 +1,25 @@
+from website.project.model import User
+from website.util.permissions import reduce_permissions
+
+from admin.users.serializers import serialize_simple_node
+
+
+def serialize_node(node):
+    user_list = {key: reduce_permissions(value) for key, value in node.permissions.iteritems()}
```
sloria: Minor: it's a bit wasteful to go through two rounds of serialization to serialize contributors. You are looping through the contributors here, then again in the `map` call a few lines down. You should be able to reduce this to one iteration.

admin_tests/users/test_serializers.py
```diff
@@ -0,0 +1,29 @@
+from nose.tools import *  # flake8: noqa
+
+from tests.base import AdminTestCase
+from tests.factories import NodeFactory, UserFactory
+
+from admin.users.serializers import serialize_user, serialize_simple_node
+
+
+class TestUserSerializers(AdminTestCase):
+    def setUp(self):
```
sloria: Minor: This `setUp` doesn't change any behavior, so it can be removed.

sloria:
Is https://github.com/CenterForOpenScience/osf.io/pull/4960#issuecomment-178747193 still true? Does this rely on upgrading Django versions?

sloria:
Pass finished.

sloria:
Also, can you please link to the JIRA ticket for this if there is one?

Ghalko:
Jira tickets mentioned: ADM-3 and ADM-56

Ghalko:
@sloria ^^

admin/base/views.py
Ghalko: This is where the view is returned, so it either has to return a filled in template or a 404 if the serializer comes back with an error. The serializer already raises the exception and returns to the `get_guid_object` early which I guess I could catch and add a new piece of string to, but ultimately it is going to the view.

admin/nodes/serializers.py
Ghalko: Done.

admin_tests/users/test_serializers.py
Ghalko: Removed

sloria:
Looks good!

PR #4962, opened 2016-01-29
[RFR][Feature] Better generics for citation addons

mfraezz:
[#OSF-5369]

Purpose
======
Generalize code for citation addons (Mendeley and Zotero) so that duplication is minimized.

Changes
=======

- [X] Views generalized
- [X] Serializers generalized
- [X] Providers generalized
- [X] Models generalized
- [x] Tests generalized
- [x] Migration script for node settings model written:
   * `<addon>_list_id  --> list_id`

Side Effects
=========
None

Migration Steps
----------------------
`python -m  scripts.migration.migrate_citation_addons_list_id`


felliott:
Minor:  the storage providers all have a `README.md` in their addon directory that explains how to create and configure an application for local testing.  Could you add one each to the mendeley and zotero directories?  The one in the github addon is a good model.

felliott:
Please add docstrings to `website/citations/*`.

felliott:
`GenericCitationViews.import_auth` lost the `@must_not_be_registration` decorator in the refactor.

felliott:
In `CitationsProvider.user_accounts()` the `if user else None` seems odd.  If user is false, won't `user.external_accounts` explode?  If so, shouldn't there be some error handling for that case?

felliott:
In `GenericCitationViews`, why are you setting the names of the anonymous functions to `addon_short_name + original name` when you call them by `original_name`?  Does it help with debugging, a-la Javascript?

felliott:
`website/addons/mendeley/model.py:`  ` _citations_for_folder` calls a method `_citations_for_mendeley_user`, that doesn't seem to exist. Is that supposed to refer to `_citations_for_user`?

website/addons/zotero/tests/test_models.py
```diff
@@ -8,7 +8,12 @@
```
felliott: These two imports don't seem to be used anymore.

website/addons/zotero/tests/test_models.py
```diff
@@ -8,7 +8,12 @@
 
 from tests.base import OsfTestCase
 from tests.factories import UserFactory, ProjectFactory
```
felliott: Ditto.

website/addons/base/testing/utils.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+
+class MockNode(object):
+
```
felliott: This class doesn't seem to be used anywhere.

website/addons/zotero/tests/test_models.py
```diff
@@ -34,337 +42,20 @@ def test_handle_callback(self):
...
-                metadata={'folder': 'fake-folder-id'}
-            )
-        )
-
-        log = self.node.logs[-1]
-        assert_equal(log.action, 'zotero_folder_selected')
-        assert_equal(log.params['folder_id'], folder_id)
-        assert_equal(log.params['folder_name'], folder_name)
-
-    def test_has_auth_false(self):
```
felliott: Is this test implemented in the testing mixin? I couldn't find one similar. (sorry, this is a dupe of one I added to a commit on accident).

website/addons/zotero/tests/test_models.py
```diff
@@ -34,337 +42,20 @@ def test_handle_callback(self):
...
-    def test_clear_auth(self):
-        self.node_settings.external_account = ZoteroAccountFactory()
-        self.node_settings.zotero_list_id = 'something'
-        self.node_settings.user_settings = self.user_settings
-        self.node_settings.save()
-
-        self.node_settings.clear_auth()
-
-        assert_is_none(self.node_settings.external_account)
-        assert_is_none(self.node_settings.zotero_list_id)
```
felliott: This is not asserted in the refactored test mixin. Does it no longer apply?

website/addons/zotero/tests/test_models.py
mfraezz: Several different tests in [`OAuthAddonNodeSettingsTestSuiteMixin`](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/addons/base/testing/models.py#L119) cover what's tested here. `test_complete_has_auth_not_verified`, `test_complete_auth_false`, `test_has_auth`, and `test_clear_auth`.

website/addons/zotero/tests/test_models.py
mfraezz: [`test_clear_settings`](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/addons/base/testing/models.py#L226) secretly does the same thing, as [`AddonCitationsNodeSettings.folder_id`](https://github.com/CenterForOpenScience/osf.io/pull/4962/files#diff-fcda41ab988f395bf849bacc226f5b76R45) just returns `list_id`.

website/addons/mendeley/tests/test_models.py
```diff
@@ -43,406 +33,22 @@ def test_handle_callback(self, mock_get_client):
...
-
-        assert_equal(
-            self.node_settings.selected_folder_name,
-            'Fake Folder'
-        )
-
-    # TODO: Make these tests generic and move to core
-
-    @mock.patch('framework.status.push_status_message')
-    def test_remove_contributor_authorizer(self, mock_push_status):
```
felliott: Are these tests not needed anymore? I couldn't find if they were moved to somewhere else.

mfraezz: Hmm. I thought they were, but looking through `addons/base/testing` I don't see why I thought that. I'll see about generalizing these.

website/addons/base/testing/models.py
```diff
@@ -390,3 +394,193 @@ def test_after_delete(self):
...
+            res = self.provider.client
+            assert_equal(res, self.provider._client)
+            assert_false(mock_get_client.called)
+
+    def test_has_access(self):
+        mock_client = mock.Mock()
+        mock_client.folders.list.return_value = self.ApiExceptionClass({'status_code': 403, 'text': 'Mocked 403 ApiException'})
+        mock_client.collections.return_value = self.ApiExceptionClass({'status_code': 403, 'text': 'Mocked 403 ApiException'})
+        self.provider._client = mock_client
+        self.provider._client
```
felliott: Should this be accessing `.client` to invoke the error? Also, I can't get this test to fail; the assert_raises seems to be a no-op.  Even if I comment out all the setup code, it passes.

felliott:
Hey @mfraezz,

First-pass CR is finished.  It's looking really good!  Let me know when you're ready for me to review again.

mfraezz:
>In CitationsProvider.user_accounts() the if user else None seems odd. If user is false, won't user.external_accounts explode? If so, shouldn't there be some error handling for that case?

I don't think `user` *can* be `None` there, as that function is only used in the `account_list` view, and only seen from the node_settings page, which should require contributorship:
```python
@must_be_logged_in
def _account_list(auth):
    """ List addon accounts associated with the currently logged-in user
    """
    return Provider().user_accounts(auth.user)
```

mfraezz:
Otherwise, CR has been addressed. Back to you, @felliott

felliott:
LGTM, sending this to RTM.

PR #4963, opened 2016-01-31
Pool PyMongo clients

chrisseto:
PyMongo's built in pooling break toku as sockets are acquired per atomic operation, not thread/request.


icereval:
+1

icereval:
@sloria, we'll want to have this get validated by sitting around on a release branch for a full cycle

PR #4965, opened 2016-02-01
include api module in coverage reports

whit537:
I don't find a way to run coverage reports on the `api` module from pytest without this configuration change. Coming from https://github.com/caneruguz/osf.io/pull/17#issuecomment-176989103.


whit537:
Thanks @sloria. :-)

sloria:
Thank you!

PR #4971, opened 2016-02-02
[WIP] Handle file comments on move/rename file

samanehsan:
Update existing FileNode objects with metadata returned from waterbutler. For files in nested folders, just update the path and if necessary, the node and provider.


samanehsan:
Replaced by #/5059

PR #4976, opened 2016-02-03
[WIP][Feature] Upgrade API to 1.9

Ghalko:
## Proposal

Upgrade the API from Django 1.8 to 1.9.

## Repercussions

Although I have started this process there are a few catches that might not be great:
- `api.base.serializer` ~443 has a line that gets back what looks to be a dictionary, but is actually a string.
- `hasattr` is used a few times in the API, but might be causing significant recursion.

Hopefully other issues are related to these, so fixing them will fix many other test failures.


Ghalko:
Stopped build checks because there are obvious failures.

Ghalko:
This is better done by #5242

PR #4981, opened 2016-02-03
Fix conference creation with custom field_names

hmoco:


sloria:
please resend to `release/0.63.0`

PR #4986, opened 2016-02-04
Switch outstanding 'contains' queries to 'eq' queries

cwisecarver:
This moves our list queries to eq queries so that they'll be faster.


sloria:
Much better.

PR #4987, opened 2016-02-04
[Ready to Review] Feature/separate date_last_login and date_last_request

pattisdr:
# Purpose

Issue: https://openscience.atlassian.net/browse/OSF-5523

We have a user field called `date_last_login`, but it actually is the date the user last sent a request.  So currently, we are not actually tracking `date_last_login`. `date_last_login` and `date_last_request` should be separated.  Tracking `date_last_request` separately will allow `date_last_login` to be used for things like loading the number of comments on your project since you last logged in.

# Changes
- Adds a User field called `date_last_request` which will track what `date_last_login` is tracking
- Restricts `date_last_login` to be updated only when a user authenticates with CAS and on initial confirmation
- `date_last_request` is updated when a user makes a request
- Adds script to add `date_last_request` field, initially set equal to `date_last_login`

# Side Effects

Requires a migration.

### DIFF
--- a/scripts/migration/migrate_date_last_request_for_current_users.py
+++ b/scripts/migration/migrate_date_last_request_for_current_users.py
@@ -16,7 +16,7 @@ logger = logging.getLogger(__name__)
 def main(dry=True):
     init_app(routes=False)
     users = models.User.find(Q('date_last_request', 'eq', None))
-    logger.warn('All active users will have a date_last_request field added and set equal to date_last_login.')
+    logger.info('All active users will have a date_last_request field added and set equal to date_last_login.')
 
     for user in users:
         if user.is_active:
--- a/scripts/migration/migrate_date_last_request_for_current_users.py
+++ b/scripts/migration/migrate_date_last_request_for_current_users.py
@@ -1,0 +1,34 @@
+"""
+Used to add users date_last_request, value set equal to date_last_login.
+"""
+
+import sys
+import logging
+from modularodm import Q
+from website.app import init_app
+from website import models
+from scripts import utils as script_utils
+from framework.transactions.context import TokuTransaction
+
+logger = logging.getLogger(__name__)
+
+
+def main(dry=True):
+    init_app(routes=False)
+    users = models.User.find(Q('date_last_request', 'eq', None))
+    logger.warn('All active users will have a date_last_request field added and set equal to date_last_login.')
+
+    for user in users:
+        with TokuTransaction():
+            if user.is_active:
+                user.date_last_request = user.date_last_login
+                if not dry:
+                    user.save()
+                    logger.info('User {0} "date_last_request" added'.format(user._id))
+
+if __name__ == '__main__':
+    dry_run = 'dry' in sys.argv
+    if not dry_run:
+        script_utils.add_file_logger(logger, __file__)
+    main(dry=dry_run)
+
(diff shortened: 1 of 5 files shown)


scripts/migration/migrate_date_last_request_for_current_users.py
```diff
@@ -0,0 +1,34 @@
...
+def main(dry=True):
+    init_app(routes=False)
+    users = models.User.find(Q('date_last_request', 'eq', None))
+    logger.warn('All active users will have a date_last_request field added and set equal to date_last_login.')
+
+    for user in users:
+        with TokuTransaction():
+            if user.is_active:
+                user.date_last_request = user.date_last_login
+                if not dry:
```
mfraezz: Preferred method of handling dry runs is to run the entire migration in a `TokuTransaction` and raise an error at the end `if dry`. Something like:

```python
def main(dry=True):
    #[perform migration]
    if dry:
        raise RuntimeError('Dry run -- transaction rolled back')

if __name__ == '__main__':
    #[perform setup]
    with TokuTransaction():
        main(dry=dry_run)
```

scripts/migration/migrate_date_last_request_for_current_users.py
```diff
@@ -0,0 +1,34 @@
...
+from scripts import utils as script_utils
+from framework.transactions.context import TokuTransaction
+
+logger = logging.getLogger(__name__)
+
+
+def main(dry=True):
+    init_app(routes=False)
+    users = models.User.find(Q('date_last_request', 'eq', None))
+    logger.warn('All active users will have a date_last_request field added and set equal to date_last_login.')
```
mfraezz: Should this be a `.info`? `.warn` and `.error` are usually reserved for when something is going wrong.

mfraezz:
Please add tests for the migration script.

Also, I wouldn't consider this a blocker (as there are currently no tests involving `date_last_login` that aren't targeting mail functionality), but tests that make sure these fields get updated at under the right conditions would be appreciated.

mfraezz:
LGTM. Passing this on to RTM :zap:

sloria:
@pattisdr Is this PR still relevant? Do we still need this field for the Dashboard?

pattisdr:
This was for a feature that was originally going to be in Phase 1, but it was pulled out for a future phase.  We haven't decided if we will need this or not, but I think it's a useful field.

Originally, we were going to have comments since last logged in, activity since last logged in. Last we talked, it had changed to unread comments, unread activity, etc.  So this 'date_last_login' would probably not be used for that, but again, we haven't really decided what's going to happen with this.

This is a useful field though, it is strange to me that date_last_login is currently date_last_request.

sloria:
OK, thanks for the update. I agree that `date_last_login` is a misnomer, since it's really `date_last_request`. That said, there is code--triggered emails, for example--that currently relies on `date_last_login`. Doesn't this code need to be updated?

That said, we can deprioritize work on this for now if we don't need it in the near future.

sloria:
I'm going to close this for now. We can re-open when the need for the new field arises.

pattisdr:
Sounds good.

On Mon, Apr 11, 2016 at 9:33 AM, Steven Loria <notifications@github.com>
wrote:

> I'm going to close this for now. We can re-open when the need for the new
> field arises.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/CenterForOpenScience/osf.io/pull/4987#issuecomment-208345463>
>



-- 
Dawn Pattison
Developer
Center for Open Science
dawn@cos.io

PR #4991, opened 2016-02-04
[Hotfix] Conference Speed

chrisseto:
Migration `python -m scripts.migration.migrate_tags`

### DIFF
--- a/website/conferences/views.py
+++ b/website/conferences/views.py
@@ -233,12 +233,8 @@ def conference_submissions(**kwargs):
         # For efficiency, we filter by tag first, then node
         # instead of doing a single Node query
         projects = set()
-        for tag in Tag.find(Q('_id', 'iexact', conf.endpoint)):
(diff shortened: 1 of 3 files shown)


website/conferences/views.py
```diff
@@ -233,12 +233,8 @@ def conference_submissions(**kwargs):
         # For efficiency, we filter by tag first, then node
         # instead of doing a single Node query
         projects = set()
-        for tag in Tag.find(Q('_id', 'iexact', conf.endpoint)):
-            for node in tag.node__tagged:
-                if not node:
-                    continue
-                if not node.is_public or node.is_deleted:
-                    continue
+        for tag in Tag.find(Q('lower', 'eq', conf.endpoint)):
```
sloria: I don't think `conf.endpoint` is guaranteed to be lower case here; better to use `.lower()` just in case.

sloria:
Merged with minor modifications.

PR #4994, opened 2016-02-04
Add ability for user to remove institution affiliation

hmoco:
tests and other stuff, this might end up only containing api code.


api_tests/users/views/test_user_institutions_relationship.py
```diff
@@ -0,0 +1,170 @@
...
+    def test_type_mistyped(self):
+        res = self.app.delete_json_api(
+            self.url,
+            {'data': [
+                {'type': 'wow', 'id': self.institution1._id}
+            ]},
+            auth=self.user.auth, expect_errors=True
+        )
+
+        assert_equal(res.status_code, 409)
```
hmoco: this is failing. Delete does not go through ` to_internal_value ` method in TypeField

PR #4995, opened 2016-02-04
[Feature] fix esi for jsonapi

cwisecarver:
This disables ESI in **all** cases except where the content-type is that of jsonapi. This can be forced for testing with ?format=jsonapi in the url or by passing the correct jsonapi vendor content type as a header.

### DIFF
--- a/api/base/renderers.py
+++ b/api/base/renderers.py
@@ -1,21 +1,23 @@
 import re
 from rest_framework.renderers import JSONRenderer, BrowsableAPIRenderer
 
-class JSONAPIRenderer(JSONRenderer):
-    format = "jsonapi"
-    media_type = 'application/vnd.api+json'
+
+class JSONRendererWithESISupport(JSONRenderer):
+    format = 'json'
+    media_type = 'application/json'
 
     def render(self, data, accepted_media_type=None, renderer_context=None):
         #  TODO: There should be a way to do this that is conditional on esi being requested and
         #  TODO: In such a way that it doesn't use regex unless there's absolutely no other way.
-        initial_rendering = super(JSONAPIRenderer, self).render(data, accepted_media_type, renderer_context)
+        initial_rendering = super(JSONRendererWithESISupport, self).render(data, accepted_media_type, renderer_context)
         augmented_rendering = re.sub(r'"<esi:include src=\\"(.*?)\\"\/>"', r'<esi:include src="\1"/>', initial_rendering)
         return augmented_rendering
 
 
-class JSONRenderer(JSONAPIRenderer):
-    format = 'json'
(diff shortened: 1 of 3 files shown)


cwisecarver:
The last commit now disables ESI when the content-type not in jsonapi or json.

api/base/views.py
```diff
@@ -74,7 +74,10 @@ def get_serializer_context(self):
             embeds_partials[embed] = self._get_embed_partial(embed, embed_field)
 
         context.update({
-            'enable_esi': is_truthy(self.request.query_params.get('esi', django_settings.ENABLE_ESI)),
+            'enable_esi': (
+                is_truthy(self.request.query_params.get('esi', django_settings.ENABLE_ESI)) and
+                self.request.accepted_media_type in {'application/vnd.api+json', 'application/json'}
```
sloria: Would be good to have these media types go in a settings variable--something like `ESI_MEDIA_TYPES`.

api/base/renderers.py
```diff
@@ -12,6 +12,11 @@ def render(self, data, accepted_media_type=None, renderer_context=None):
         augmented_rendering = re.sub(r'"<esi:include src=\\"(.*?)\\"\/>"', r'<esi:include src="\1"/>', initial_rendering)
         return augmented_rendering
 
+
+class JSONRenderer(JSONAPIRenderer):
```
sloria: Why is this necessary?

sloria:
Pass finished. Just a minor refactor and a question to address, and this is good to merge.

api/base/views.py
cwisecarver: I like it. I'll add it in.

api/base/renderers.py
cwisecarver: There's no way (that I could find) to specify more than one `media_type` per renderer. Since the stock JSONRenderer was not unescaping the ESI tags I subclassed the JSONAPIRender which subclasses the stock JSONRenderer in order to get the same functionality with the smallest amount of code. Looking back on this, I should have probably made a CustomJSONRenderer that inherits from the stock JSONRenderer, implements the render method and specifies json and application/json and then subclassed to get the JSONAPIRenderer. I'm happy to change it to that to save myself embarrassment and if it will more easily answer your question for future readers.

sloria: Ah, I see; I had forgotten that `render` was being implemented in `JSONAPIRenderer`. I think your proposed approach makes sense: implement a `JSONRendererWithESISupport` (or a better name) then have JSONAPIRenderer subclass that. It's a minor change though: I'll leave it up to you to decide if you want to make that.

cwisecarver:
@sloria Once the tests pass all should be well.

PR #4996, opened 2016-02-05
Incorrect Log Permissions and using LogMixin

pattisdr:
# Purpose

Issue: https://openscience.atlassian.net/browse/OSF-5665

<s>API log permissions are incorrect. Users can currently view logs for nodes that don't have permission to view.  Log endpoints are returning 200, regardless of user permissions.

Also, the api endpoint `/v2/logs/{log_id}/nodes/` is giving a 500 error on develop. </s>

<h5> The above was handled in a hotfix. 

This PR now only adds tests for NodeLogsAddedContributors view and changes 'added_contributors' to 'contributors.  If a log action is related to contributors, such as contributor_added, contributor_removed, etc., contributors will appear under the contributors relationship.  This makes the field name more generic. /h5>

# Changes

<s> 1) Updated permissions on LogNodeList, NodeLogDetail, and NodeLogAddedContributors.  The user can only view the log if they have permission to view one of the nodes on the log's backrefs.  This permission check will be even shorter when backrefs are removed between nodes and nodelogs because every log will be associated with exactly one node. </s>

<s>2) Had LogNodeList, and NodeLogDetail inherit from LogMixin so get_log() method is available. This was causing the 500 error. </s>

<s>3) Add tests for the three views. </s>

<h5> Adds tests for NodeLogAddedContributors.  Changes 'added_contributors' language to 'contributors' language. </h5>

### DIFF
--- a/api/logs/views.py
+++ b/api/logs/views.py
@@ -32,11 +32,10 @@ class LogMixin(object):
             )
         return log
 
-
     def check_log_permission(self, log):
-        '''
+        """
         Cycles through nodes on log backrefs.  If user can view any of the nodes, the log can be viewed.
-        '''
+        """
         auth_user = get_user_auth(self.request)
         log_nodes = []
 
--- a/api/logs/views.py
+++ b/api/logs/views.py
@@ -126,7 +126,7 @@ class LogNodeList(JSONAPIBaseView, generics.ListAPIView, LogMixin, ODMFilterMixi
         ]
 
 
-class NodeLogDetail(JSONAPIBaseView, generics.RetrieveAPIView):
+class NodeLogDetail(JSONAPIBaseView, generics.RetrieveAPIView, LogMixin):
     """List of nodes that a given log is associated with. *Read-only*.
 
     Paginated list of nodes that the user contributes to.  Each resource contains the full representation of the node,
@@ -182,10 +182,11 @@ class NodeLogDetail(JSONAPIBaseView, generics.RetrieveAPIView):
     permission_classes = (
         drf_permissions.IsAuthenticatedOrReadOnly,
         base_permissions.TokenHasScope,
+        ContributorOrPublic
     )
 
-    required_read_scopes = [CoreScopes.NODE_BASE_READ]
-    required_write_scopes = [CoreScopes.NODE_BASE_WRITE]
+    required_read_scopes = [CoreScopes.NODE_LOG_READ]
+    required_write_scopes = [CoreScopes.NULL]
 
     serializer_class = NodeLogSerializer
     view_category = 'logs'
@@ -193,13 +194,8 @@ class NodeLogDetail(JSONAPIBaseView, generics.RetrieveAPIView):
 
     # overrides RetrieveUpdateDestroyAPIView
     def get_object(self):
-        log = get_object_or_error(
-            NodeLog,
-            self.kwargs['log_id'],
-            display_name='log'
-        )
-        # May raise a permission denied
-        self.check_object_permissions(self.request, log)
+        log = self.get_log()
+        self.check_log_permission(log)
         return log
 
     # overrides RetrieveUpdateDestroyAPIView
--- a/api/logs/views.py
+++ b/api/logs/views.py
@@ -269,4 +269,3 @@ class NodeLogAssociatedContributors(JSONAPIBaseView, generics.ListAPIView, ODMFi
             return []
         associated_users = User.find(Q('_id', 'in', associated_contrib_ids))
         return associated_users
-
--- a/api/logs/views.py
+++ b/api/logs/views.py
@@ -6,6 +6,9 @@ from modularodm import Q
 from framework.auth.core import User
 
 from framework.auth.oauth_scopes import CoreScopes
+from api.nodes.permissions import (
+    ContributorOrPublic,
+)
 
 from api.base.filters import ODMFilterMixin
 from api.base.utils import get_user_auth, get_object_or_error
@@ -102,6 +105,7 @@ class LogNodeList(JSONAPIBaseView, generics.ListAPIView, LogMixin, ODMFilterMixi
     permission_classes = (
         drf_permissions.IsAuthenticatedOrReadOnly,
         base_permissions.TokenHasScope,
+        ContributorOrPublic
     )
 
     required_read_scopes = [CoreScopes.NODE_LOG_READ]
@@ -114,6 +118,7 @@ class LogNodeList(JSONAPIBaseView, generics.ListAPIView, LogMixin, ODMFilterMixi
 
     def get_queryset(self):
         log = self.get_log()
+        self.check_log_permission(log)
         auth_user = get_user_auth(self.request)
         return [
             node for node in log.node__logged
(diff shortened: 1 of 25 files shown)


sloria:
@pattisdr Could you please add a regression test that fails without this change?

pattisdr:
I am writing lots of tests right now - just haven't pushed.  Logs permissions are incorrect so that will be part of this.

api/logs/views.py
```diff
@@ -97,6 +117,7 @@ class LogNodeList(JSONAPIBaseView, generics.ListAPIView, ODMFilterMixin):
 
     def get_queryset(self):
         log = self.get_log()
+        self.check_log_permission(log)
```
whit537: We don't use the return value here, and `check_log_permission` above returns `True` or `False`. Does it also raise in the event that the user doesn't have permission?

pattisdr: Yes, this method throws a 401 or 403 if the user doesn't have permission.  Checking permissions on logs are tricky, because whether or not the user can view the log is not stored on the log itself, but rather, if the user can view the node the log pertains to.

pattisdr: It is `self.check_object_permissions` specifically that throws the error.

api/logs/views.py
```diff
@@ -29,8 +32,24 @@ def get_log(self):
...
+        log_nodes = []
+
+        for node_id in log._backrefs['logged']['node']['logs']:
+            node = get_object_or_error(Node, node_id, display_name='node')
+            log_nodes.append(node)
+            if node.can_view(auth_user):
+                return True
+
+        self.check_object_permissions(self.request, log_nodes[0])
+        return False
```
whit537: Since the return value isn't used, how about:

 - dropping this `return False`
 - a plain `return` just above, instead of `return True`
 - adding the content at https://github.com/CenterForOpenScience/osf.io/pull/4996/files#r52082314 to the docstring or in comments

whit537: > Yes, this method throws a 401 or 403 if the user doesn't have permission. Checking permissions on logs are tricky, because whether or not the user can view the log is not stored on the log itself, but rather, if the user can view the node the log pertains to.

api/logs/views.py
```diff
@@ -29,8 +32,24 @@ def get_log(self):
...
 
+    def check_log_permission(self, log):
+        """
+        Cycles through nodes on log backrefs.  If user can view any of the nodes, the log can be viewed.
+        """
+        auth_user = get_user_auth(self.request)
+        log_nodes = []
+
+        for node_id in log._backrefs['logged']['node']['logs']:
+            node = get_object_or_error(Node, node_id, display_name='node')
```
whit537: I take it, then, that "error" here is a verb, not a noun?

pattisdr: Seems that way.

api/logs/views.py
whit537: > It is `self.check_object_permissions` specifically that throws the error.

Ah, okay. So maybe a comment there?

```python
self.check_object_permissions(self.request, log_nodes[0])  # will raise 401 or 403 as appropriate
```

api/logs/views.py
whit537: But as you [hint at](https://github.com/CenterForOpenScience/osf.io/pull/4996/files#r52082595), `get_object_or_error` [doesn't error](https://github.com/CenterForOpenScience/osf.io/blob/ea8186dcff172b117b4ebe6632479d532f2ae77a/api/base/utils.py#L55-L86) with errors relevant to checking log permissions.

pattisdr: I think the error for that method is for when the node itself can't be retrieved.

pattisdr: Like if the `id` is invalid, for example.  There are several potential errors- permissions, or retrieving the node itself.

api/logs/views.py
```diff
@@ -29,8 +32,24 @@ def get_log(self):
             )
         return log
```
whit537: It looks like we call the new `check_permissions` method every time that we call `get_log`. That suggest to me that we should push the `check_permissions` call down here into `get_log` itself, to ensure that it's always called, including for any future log node classes.

api_tests/logs/views/test_log_added_contributors.py
```diff
@@ -0,0 +1,74 @@
...
+        self.public_log_contribs_url = self.url + '{}/added_contributors/'.format(self.public_log_add_contributor._id)
+
+    def tearDown(self):
+        NodeLog.remove()
+        Node.remove()
+
+
+class TestLogAddedContributors(LogsTestCase):
+
+    def test_log_detail_private_logged_in_contributor(self):
```
whit537: It's good practice to give test methods sentences for names, so it's clear what they do. In this case, perhaps:

```
def test_log_detail_private_logged_in_contributor_can_access_logs
```

api_tests/logs/views/test_log_detail.py
```diff
@@ -0,0 +1,90 @@
...
+)
+
+from framework.auth.core import Auth
+
+from website.models import NodeLog, Node
+from website.util import permissions as osf_permissions
+from api.base.settings.defaults import API_BASE
+
+
+class LogsTestCase(ApiTestCase):
```
whit537: Why have multiple `LogsTestCase`s? Why not a single base class?

api_tests/logs/views/test_log_added_contributors.py
whit537: And I think these wanna be `test_log_added_contributors_*` instead of `test_log_detail_*`.

whit537: (Though I'm not sure we need the `log_foo_` prefix since it's redundant with the name of the test case, `TestLog{Detail,AddedContributors}`.)

whit537:
Alright, there are some test style changes that could be made, but since this issue is critical I'm not going to block on that. The one thing I would recommend addressing is ...

 - [ ] move `check_log_permissions` down into `get_log`—https://github.com/CenterForOpenScience/osf.io/pull/4996#discussion_r52180652

pattisdr:
Requested changes made.

whit537:
:+1:

pattisdr:
Became hotfix #5004

sloria:
Good to merge when the build passes. :+1:

pattisdr:
I would like to add renaming of route added_contributors to associated_contributors.

pattisdr:
@sloria can you review the changes in https://github.com/pattisdr/osf.io/commit/e481dd178e5595a87201c949cb6121f55f6e298c? 

QA uncovered that the `/added_contributors/` endpoint actually returns contributors for any contributor-related log action. So a log action like contributor_removed will have the removed contributor information under the `added_contributors` relationship.  Changing this to `associated_contributors` handles all contributor-related cases.

Is it OK to include this here? If not, I will pull it out.

sloria:
@pattisdr  +1. It's a good idea to make the field name more generic. However, I think `contributors` is more succinct (without losing clarity).

pattisdr:
@sloria,

pattisdr:
@sloria, `associated_contributors` changed to `contributors`.

PR #5000, opened 2016-02-08
[rfr][feature] Log out users merging or adding email

billyhunt:
Associated with 

Associated with 
https://github.com/CenterForOpenScience/osf.io/pull/5528

### Purpose

Existing users who add an email or merge an account have unexpected behaviors, depending on whom is logged in.  

For example, when adding an email, whomever clicks the link sent when adding an email is automatically logged in which is a security risk.  

Merging is also less then optimal, depending on whom is logged in.  If UserA is merged into UserB.

When UserA is logged in, the merge will succeed.
When UserB is logged in, the merge will fail with no error message.  

### Changes:

* Whomever is logged in will be logged out whenever an email is added or a user is merged.  The user will be directed to a login screen and told which username they should log in as.
* After they log in, they will be given the choice to add the email or merge, with the text of the choice changing depending on whether they are merging or adding an email.
*  If they choose Yes, then the merge or email add occurs.
*  If they choose no, they are removed from the user profile as a user to add.


framework/auth/core.py
```diff
@@ -837,6 +842,23 @@ def _get_unconfirmed_email_for_token(self, token):
...
+        try:
+            verification = self.email_verifications[token]
+        except:
+            return False
+        # Check token for existance and date
+        if (
+            'expiration' in verification and
+            verification['expiration'] < dt.datetime.utcnow() or
+            token not in self.email_verifications
+        ):
```
mfraezz: Can you explicitly group the parts of this conditional for evaluation order? `x and (y or z)` or `(x and y) or z`

billyhunt:
Hey @brianjgeiger and @mfraezz, I am trying to use remove_sessions_for_user(user) instead of logout(), and it seems to not allow a login from my redirect.  It redirects me to the login page.  Have either of you seen that before?

brianjgeiger:
Hmmm. When I use it, I tend to have to try to log in twice. I wasn't sure if it was related to that or some other login changes that have recently happened. It wasn't super-important with what I was doing, but it sounds like it might be key to your process, so you (or someone) might need to poke around with what's going on there.

brianjgeiger:
In the Auth model, there are some functions that deal with taking data from the cookie and finding the session from that (and doing the opposite, grabbing the session data and creating a cookie). My guess is that the problem is in that. It probably needs to handle having a cookie that is associated with a missing session in a little more graceful of a manner.

billyhunt:
I don't think remove_sessions_for_user is right for me in this case.  Its more robust (if a 3rd party were logged in, it would still work).  And it works.

brianjgeiger:
If you want to log out all users logged in as a certain user, this is the way to go. If you're just trying to make sure the user in the current browser session is logged out, then it's not. But, since you have a case where you want to log out the account that's being merged from as well as the account being merged into, you need to use this method in at least some cases.

billyhunt:
I thouuugghhhtt that logout() logged out whomever is currently logged in.  Thats not the case?

brianjgeiger:
Log in through an incognito window and a regular window, use logout, and see if both windows get logged out.

billyhunt:
I implemented @brianjgeiger  snazzy decorator to log out both users when merging.  I am seeing a double login needed.  Apparently this has come up before and might be involved with my local environment.  Should be good to look at again @mfraezz

framework/auth/views.py
```diff
@@ -143,8 +144,8 @@ def auth_login(auth, **kwargs):
         if (campaign == 'institution' and settings.ENABLE_INSTITUTIONS) or campaign != 'institution':
             data['campaign'] = campaign
     data['login_url'] = cas.get_login_url(redirect_url, auto=True)
-
     data['sign_up'] = request.args.get('sign_up', False)
+    data['existing_user'] = request.args.get('existing_user')
```
mfraezz: You should probably define a default: `request.args.get('existing_user', False)`

framework/auth/views.py
```diff
@@ -174,9 +196,12 @@ def confirm_email_get(token, auth=None, **kwargs):
     user = User.load(kwargs['uid'])
     is_merge = 'confirm_merge' in request.args
     is_initial_confirmation = not user.date_confirmed
-
+    existing_user = request.args.get('existing_user')
```
mfraezz: `.get('existing_user', False)`

framework/auth/views.py
```diff
@@ -229,6 +254,126 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+    """Called at login if user cancels their merge or email add.
+    methods: PUT
+    """
+    user = auth.user
+    confirmed_email = request.json.get('address')
+    email_verifications = user.email_verifications.copy()
+    for token in user.email_verifications:
+        if user.confirm_token(token):
+            if user.email_verifications[token]['email'] == confirmed_email:
+                del email_verifications[token]
```
mfraezz: This `copy/del` may be unnecessary, you should just be able to do `user.email_verifications.pop(token)`

framework/auth/core.py
```diff
@@ -855,6 +860,23 @@ def _get_unconfirmed_email_for_token(self, token):
 
         return verification['email']
 
+    def confirm_token(self, token):
+        """Return whether or not a confirmation token is valid for this user.
+        :rtype: bool
+        """
+        try:
+            verification = self.email_verifications[token]
+        except:
```
mfraezz: Should this be excepting a `KeyError`?

framework/auth/views.py
```diff
@@ -163,6 +164,27 @@ def auth_logout(redirect_url=None):
     return resp
 
 
+def auth_email_logout(token, user):
+    """When a user is adding an email or merging an account, add the email to the user and log them out.
+    """
+    redirect_url = web_url_for('auth_login') + '?existing_user=True'
+    if user.confirm_token(token):
+        try:
+            user_merge = User.find_one(Q('emails', 'iexact', user.email_verifications[token]['email']))
```
mfraezz: Don't use `iexact` unless absolutely necessary -- it's very slow. 
`Q('emails', 'eq', user.email_verifications[token]['email'].lower())` should work

framework/auth/views.py
```diff
@@ -229,6 +254,126 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+    methods: GET
+    """
+    # user = User.load(kwargs['uid'])
+    user = auth.user
+    verified_emails = []
+    for token in user.email_verifications:
+        if user.confirm_token(token):
+            if user.email_verifications[token]['confirmed']:
+                try:
+                    user_merge = User.find_one(Q('emails', 'iexact', user.email_verifications[token]['email']))
```
mfraezz: See [this comment](https://github.com/CenterForOpenScience/osf.io/pull/5000/files#r57469484)

framework/auth/views.py
```diff
@@ -247,13 +392,15 @@ def send_confirm_email(user, email):
         merge_target = None
 
     campaign = campaigns.campaign_for_user(user)
-    # Choose the appropriate email template to use
+    # Choose the appropriate email template to use and add existing_user flag if a merge or adding an email.
     if merge_target:
         mail_template = mails.CONFIRM_MERGE
+        confirmation_url += '?existing_user=True'
```
mfraezz: This doesn't really seem like something that should be put in a url. Could you not just check for this when you have to, including when you pass data to the template that uses it?

website/templates/public/login.mako
```diff
@@ -120,12 +118,11 @@
                     <label for="inputName" class="col-sm-4 control-label">Full Name</label>
                     <div class="col-sm-8">
                         <input
-                            ${'autofocus' if sign_up else ''}
-                            type="text"
-                            class="form-control"
-                            id="inputName"
-                            placeholder="Name"
-                            data-bind="
+                                type="text"
```
mfraezz: This indentation change looks wrong.

website/templates/public/login.mako
```diff
@@ -102,16 +103,13 @@
...
-        <form data-bind="submit: submit" class="form-horizontal">
-            <h3 class="m-b-lg"> Create a free account </h3>
+    %if not existing_user:
+        <div id="signUpScope" class="col-sm-5 toggle-box toggle-box-right toggle-box-muted p-h-lg" style="height: auto;">
+            <form data-bind="submit: submit" class="form-horizontal">
+                <h3 class="m-b-lg"> Create a free account </h3>
                 <div
-                    class="form-group"
-                    data-bind="
+                        class="form-group"
```
mfraezz: This indentation shouldn't have changed

website/templates/public/login.mako
```diff
@@ -134,95 +131,96 @@
                         <p class="help-block" data-bind="validationMessage: fullName" style="display: none;"></p>
                     </div>
                 </div>
-            <div
-                class="form-group"
-                data-bind="
+                <div
```
mfraezz: This `<div>` indentation is right, but its attributes are indented too far

website/templates/public/login.mako
```diff
@@ -134,95 +131,96 @@
...
-                        type="text"
-                        class="form-control"
-                        id="inputEmail"
-                        placeholder="Email"
-                        data-bind="
+                >
+                    <label for="inputEmail" class="col-sm-4 control-label">Email</label>
+                    <div class="col-sm-8">
+                        <input
+                                type="text"
```
mfraezz: Dedent this to `data-bind="` by 4 spaces

website/templates/public/login.mako
```diff
@@ -134,95 +131,96 @@
...
+                                class="form-control"
+                                id="inputEmail"
+                                placeholder="Email"
+                                data-bind="
                             value: email1,
```
mfraezz: The contents of this string should be indented by 4 spaces

website/templates/public/login.mako
```diff
@@ -134,95 +131,96 @@
...
+                        >
+                        <p class="help-block" data-bind="validationMessage: email1" style="display: none;"></p>
+                    </div>
                 </div>
-            </div>
-            <div
-                class="form-group"
-                data-bind="
+                <div
+                        class="form-group"
```
mfraezz: Please fix the indentation here

website/templates/public/login.mako
```diff
@@ -134,95 +131,96 @@
...
-                        type="text"
-                        class="form-control"
-                        id="inputEmail2"
-                        placeholder="Re-enter email"
-                        data-bind="
+                >
+                    <label for="inputEmail2" class="col-sm-4 control-label">Confirm Email</label>
+                    <div class="col-sm-8">
+                        <input
+                                type="text"
```
mfraezz: These attributes should be dedented by 4 spaces

website/templates/public/login.mako
```diff
@@ -134,95 +131,96 @@
...
+                                class="form-control"
+                                id="inputEmail2"
+                                placeholder="Re-enter email"
+                                data-bind="
                             value: email2,
```
mfraezz: This string should be indented by 4 spaces

website/templates/public/login.mako
```diff
@@ -134,95 +131,96 @@
...
-                        type="password"
-                        class="form-control"
-                        id="inputPassword3"
-                        placeholder="Password"
-                        data-bind="
+                >
+                    <label for="inputPassword3" class="col-sm-4 control-label">Password</label>
+                    <div class="col-sm-8">
+                        <input
+                                type="password"
```
mfraezz: Attr/str indentation here

framework/auth/views.py
```diff
@@ -229,6 +254,126 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+    verified_emails = []
+    for token in user.email_verifications:
+        if user.confirm_token(token):
+            if user.email_verifications[token]['confirmed']:
+                try:
+                    user_merge = User.find_one(Q('emails', 'iexact', user.email_verifications[token]['email']))
+                except NoResultsFound:
+                    user_merge = False
+
+                #todo remove confirmed, and maybe token, from object
```
mfraezz: >TODO

framework/auth/views.py
```diff
@@ -229,6 +254,126 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+                verified_emails.append({'address': user.email_verifications[token]['email'],
+                                        'token': token,
+                                        'confirmed': user.email_verifications[token]['confirmed'],
+                                        'user_merge': user_merge.email if user_merge else False})
+            else:
+                #todo migrate to remove this hack
```
mfraezz: >TODO

scripts/add_confirmed_to_existing_users_token.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+def do_migration():
+    for user in User.find(Q('email_verifications', 'ne', {})):
+        for token in user.email_verifications:
+            if 'confirmed' not in token:
+                user.email_verifications[token]['confirmed'] = False
```
mfraezz: Are there any cases where this should be set to `True`?

website/static/js/pages/dashboard-page.js
```diff
@@ -40,6 +41,104 @@ request.fail(function(xhr, textStatus, error) {
...
+                        confirm_emails(emails);
+                    }
+                }
+            }
+        });
+
+    }
+}
+
+request = $.getJSON(confirmedEmailURL, function(response) {
```
mfraezz: Is there a way to avoid doing this every time, like some sort of flag that would indicate whether or not it's necessary.

Also, if not, why not do something like this instead of assigning an otherwise unused variable:
```js
$.getJSON(confirmedEmailURL).done(
    function(){...}
).fail(
    function(){...}
);
```

website/templates/public/login.mako
```diff
@@ -48,11 +53,7 @@
     </div>
     %endif
     %if campaign != "institution" or not enable_institutions:
-        %if sign_up:
-            <div class="col-sm-5 col-sm-offset-1 toggle-box toggle-box-left toggle-box-muted p-h-lg">
```
mfraezz: Why was this removed?

scripts/add_confirmed_to_existing_users_token.py
billyhunt: I am only making it exist if it does not.  Its set to later downstream.

framework/auth/views.py
billyhunt: I tried that initially, and was getting this error.

 File "/Users/billyhunt/Dev/osf.io/framework/auth/views.py", line 293, in confirm_email_remove
    for token in user.email_verifications:
RuntimeError: dictionary changed size during iteration

billyhunt:
Ready for another pass @mfraezz 

:neckbeard: :bride_with_veil: :baby_bottle:

framework/auth/core.py
```diff
@@ -855,6 +860,23 @@ def _get_unconfirmed_email_for_token(self, token):
...
+        """
+        try:
+            verification = self.email_verifications[token]
+        except KeyError:
+            return False
+        # Check token for existance and date
+        if (
+            'expiration' in verification and
+            (verification['expiration'] < dt.datetime.utcnow() or
+                token not in self.email_verifications)
```
mfraezz: I think `or token not in self.email_verifications` would be a noop here, as you've already bailed out if `self.email_verifications[token]` raises a `KeyError`.

framework/auth/views.py
```diff
@@ -248,13 +359,15 @@ def send_confirm_email(user, email):
         merge_target = None
 
     campaign = campaigns.campaign_for_user(user)
-    # Choose the appropriate email template to use
+    # Choose the appropriate email template to use and add existing_user flag if a merge or adding an email.
     if merge_target:
         mail_template = mails.CONFIRM_MERGE
+        confirmation_url += '?existing_user={}'.format(user.email)
```
mfraezz: Is `confirmation_url` guaranteed not to have any other query params at this point?

framework/auth/views.py
```diff
@@ -230,6 +255,92 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+
+
+@collect_auth
+def confirm_email_remove(auth=None, **kwargs):
+    """Called at login if user cancels their merge or email add.
+    methods: PUT
+    """
+    user = auth.user
+    confirmed_email = request.json.get('address')
+    email_verifications = user.email_verifications.copy()
```
mfraezz: This [might not be doing](http://stackoverflow.com/questions/3975376/understanding-dict-copy-shallow-or-deep) what you think it's doing. If you just want to manipulate `user.email_verifications` (which seems to be the case), you probably don't need to make a (shallow or deep) copy at all. If you want to be able to make changes to the copied dict without affecting the original object in any way, you probably want to use `copy.deepcopy`.
```python
In [1]: a = {1: [1,2,3]}

In [2]: b = a.copy()

In [3]: b[1].append(4)

In [4]: a
Out[4]: {1: [1, 2, 3, 4]}

In [5]: b[2] = 0

In [6]: a
Out[6]: {1: [1, 2, 3, 4]}

In [7]: b
Out[7]: {1: [1, 2, 3, 4], 2: 0}

In [8]: b.pop(1)
Out[8]: [1, 2, 3, 4]

In [9]: a
Out[9]: {1: [1, 2, 3, 4]}

In [10]: b
Out[10]: {2: 0}
```

framework/auth/views.py
mfraezz: [Seemingly](https://github.com/CenterForOpenScience/osf.io/blob/develop/framework/auth/core.py#L827-L835). Still, prefer string formatting over concatenation:

```python
confirmation_url = '{}?existing_user={}'.format(confirmation_url, user.email)
```

scripts/add_confirmed_to_existing_users_token.py
mfraezz: Right. So there are no cases where it should be set to `True` *here*, this flag just indicates that the user is half-way through a more complicated comfirmation?

scripts/add_confirmed_to_existing_users_token.py
```diff
@@ -0,0 +1,38 @@
...
+from framework.auth import User
+
+
+logger = logging.getLogger(__name__)
+
+
+def do_migration():
+    for user in User.find(Q('email_verifications', 'ne', {})):
+        for token in user.email_verifications:
+            if 'confirmed' not in token:
```
mfraezz: `if 'confirmed' not in user.email_verifications[token]:`

framework/auth/views.py
mfraezz: > > TODO

website/mails/mails.py
```diff
@@ -133,6 +133,7 @@ def send_mail(to_addr, mail, mimetype='plain', from_addr=None, mailer=None,
 CONFIRM_EMAIL_PREREG = Mail('confirm_prereg', subject='Open Science Framework Account Verification, Preregistration Challenge')
 
 CONFIRM_MERGE = Mail('confirm_merge', subject='Confirm account merge')
+CONFIRM_EMAIL_ADD = Mail('confirm_email_add', subject='Add a new email to your OSF account')
```
mfraezz: This template doesn't appear to be used.

website/templates/emails/confirm_email_add.txt.mako
```diff
@@ -0,0 +1,9 @@
+Hello ${merge_target.fullname},
```
mfraezz: > Continuation of [this comment](https://github.com/CenterForOpenScience/osf.io/pull/5000/files#r58400083)

This template appears to just be an unused copy of `website/templates/emails/confirm_merge.txt.mako`

website/static/js/pages/dashboard-page.js
```diff
@@ -40,6 +41,108 @@ request.fail(function(xhr, textStatus, error) {
...
+function confirm_emails(emails) {
+    if (emails.length > 0) {
+        var email = emails.splice(0,1);
+        var title;
+        var mergeMessage;
+        var confirmMessage;
+        var nopeMessage;
+        if (email[0].user_merge) {
+            title =  'Merge account';
+            mergeMessage = 'Would you like to merge \<b>' + email[0].address + '\</b> into your account?  ' +
```
mfraezz: Consider adding an `email` section to `static/js/osfLanguage.js`? Not strictly necessary, but could clean this up a bit.

website/static/js/pages/dashboard-page.js
```diff
@@ -40,6 +41,108 @@ request.fail(function(xhr, textStatus, error) {
...
 });
 
+var confirmedEmailURL = window.contextVars.confirmedEmailURL;
+var removeConfirmedEmailURL = window.contextVars.removeConfirmedEmailURL;
+
+
+//choose which emails/users to add to an account and which to deny
+function confirm_emails(emails) {
+    if (emails.length > 0) {
+        var email = emails.splice(0,1);
```
mfraezz: This line breaks removes the first element of `emails` (which is intended because this function recurses), but assigns `email` to be an array of length 1. Consider instead:
```javascript
var email = emails.splice(0,1)[0];
```

website/templates/public/login.mako
```diff
@@ -60,34 +73,49 @@
             method="POST"
             data-bind="submit: submit"
         >
-            <h3 class="m-b-lg"> Login </h3>
+        <h3 class=${'m-b-lg' if not existing_user else 'm-b-lg m-l-md'}> Login </h3>
+        <div class="form-group">
+            <label for="inputEmail3" class="col-sm-2 control-label">Email</label>
```
mfraezz: `col-sm-[#]` went from 3 / 9 here and below to 2 / 10?

website/static/js/pages/dashboard-page.js
```diff
@@ -40,6 +41,108 @@ request.fail(function(xhr, textStatus, error) {
...
+                        });
+                    }
+                }
+            }
+        });
+
+    }
+}
+
+$.getJSON(confirmedEmailURL).done(function(response) {
```
mfraezz: This looks like it would send this request out every time this page is loaded, though we only want that to happen on login.

Also, given a) you have merge conflicts, b) this is the `dashboard-page`, and c) a huge rewrite to dashboard/projectOrganizer just went out, this will probably need refactor.

framework/auth/views.py
```diff
@@ -230,6 +255,92 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+    for token in user.email_verifications:
+        if user.email_verifications[token]['email'] == confirmed_email:
+            if user.confirm_token(token):
+                email_verifications.pop(token)
+    user.email_verifications = email_verifications
+    user.save()
+    return {
+        'status': 'success',
+        'removed_email': confirmed_email
+    }, 200
```
mfraezz: I'm not sure that this needs to return a `(json, int)` tuple; the only place it's hit has a callback function that accepts no arguments.

website/routes.py
```diff
@@ -330,6 +330,16 @@ def make_url_map(app):
...
     ], prefix='/api/v1')
 
+    process_rules(app, [
+        Rule('/dashboard/confirmed_emails/', 'get', auth_views.confirm_user_get, json_renderer),
+        Rule('/dashboard/confirmed_emails/', 'put', auth_views.add_confirmed_email, json_renderer)
+
+    ], prefix='/api/v1')
+
+    process_rules(app, [
+        Rule('/dashboard/remove_confirmed_emails/', 'put', auth_views.confirm_email_remove, json_renderer)
```
mfraezz: Because this is `remove`-type functionality, it might be better as a `DELETE` request on the same endpoint as `confirm_user_get` and `add_confirmed_email`: 
```python
process_rules(app, [
    Rule('/dashboard/confirmed_emails/', 'get', auth_views.confirm_user_get, json_renderer),
    Rule('/dashboard/confirmed_emails/', 'put', auth_views.add_confirmed_email, json_renderer),
    Rule('/dashboard/confirmed_emails/', 'delete', auth_views.confirm_email_remove, json_renderer)
], prefix='/api/v1')
```

scripts/add_confirmed_to_existing_users_token.py
```diff
@@ -0,0 +1,38 @@
...
+                user.email_verifications[token]['confirmed'] = False
+
+
+def main(dry=True):
+    init_app(set_backends=True, routes=False)  # Sets the storage backends on all models
```
mfraezz: `init_app` is usually called before a `TokuTransaction` is opened, but it's *probably* fine?

framework/auth/views.py
mfraezz: It might be, I'm less familiar with our routing system than I'd like, but I think json responses are usually automatically wrapped with the appropriate status code, headers, etc.

mfraezz:
Pass complete, looking good @billyhunt :sun_with_face: 

The test you added is good, but a couple more should be added to test the other functionality either added or changed in this PR. Frontend tests would also be good, but aren't necessarily blocking. I'll do some manual testing on the next pass after all the merge conflicts from `0.66.0` have been resolved.

website/routes.py
brianjgeiger: The only reason not to is that `DELETE`s don't accept JSON bodies, but instead do things one at a time via their ID. So if you're trying to do multiple removals at once, that's tricky. Otherwise definitely.

mfraezz: I realized this late last night, actually. Billy, disregard.

scripts/add_confirmed_to_existing_users_token.py
billyhunt: How would you suggest I handle it?  Should the flag not exist until it is necessary?

framework/auth/views.py
mfraezz: That behavior can probably be explained by the [above comment](https://github.com/CenterForOpenScience/osf.io/pull/5000/files/443a9f5e1b83eca04a6eb90d065163969cee8962#r58394271)

scripts/add_confirmed_to_existing_users_token.py
mfraezz: `.get('confirmed', False)` should work in places where you need to access it but it might not be defined. However, I just wanted to confirm the semantics behind this flag -- if it's significantly more difficult or ugly to not do this migration, it's fine the way it is.

framework/auth/views.py
billyhunt: When I was working on user.email_verifications directly, I was getting an error about modifying a dictionary I was iterating over.   Soooo, should I rock a deep copy?

website/static/js/pages/dashboard-page.js
billyhunt: I am including the email address in the messaging.  Do you think its worth passing the email variable there to get the message back?

website/templates/public/login.mako
billyhunt: I tweaked the form pretty significantly to increase aesthetics and usability.  

Before
-
screen shot 2016-04-19 at 3 47 58 pm

After
-

screen shot 2016-04-19 at 3 48 34 pm

website/static/js/pages/dashboard-page.js
billyhunt: Good point @mfraezz.  How would you restrict the call to login only?

framework/auth/views.py
billyhunt: @sloria, what do you think?  Do we need to do this return?

scripts/add_confirmed_to_existing_users_token.py
billyhunt: If you don't think we need a migration, then let's not do it.  Its cleaner that way.

website/static/js/pages/dashboard-page.js
mfraezz: You might be able to extend `signIn.js` to do this after a successful login, but before redirecting to the dashboard? I'm not certain.

mfraezz: Checking for the presence of query params indicative of a recently forced login could also work.

billyhunt: What do you mean by forced login?  Also, I think the login could come from multiple places (cas, welcome screen, etc) and at multiple times, so I think you might need to make the call every time you hit the dashboard.

billyhunt: @sloria  was cool with the page requesting every time the dashboard is hit.

framework/auth/views.py
```diff
@@ -190,6 +192,27 @@ def auth_logout(redirect_url=None):
     return resp
 
 
+def auth_email_logout(token, user):
+    """When a user is adding an email or merging an account, add the email to the user and log them out.
+    """
+    redirect_url = web_url_for('auth_login') + '?existing_user={}'.format(user.email)
+    if user.confirm_token(token):
+        try:
+            user_merge = User.find_one(Q('emails', 'eq', user.email_verifications[token]['email'].lower()))
```
mfraezz: Is there any chance of a `MultipleResultsFound` being thrown here?

billyhunt: Since I think this action is rare and the tokens are pretty long and expire quickly, I wouldn't think it would be a problem.

mfraezz: It theoretically could but shouldn't happen, so it's best if it does generate a 500.

mfraezz:
Pass complete, changes look good, behavior looked correct during demo. Dropping this in RTM 🌞

framework/auth/core.py
```diff
@@ -848,6 +854,19 @@ def _get_unconfirmed_email_for_token(self, token):
...
 
+    def clean_email_verifications(self, **kwargs):
+        given_token = kwargs.get('given_token')
+        email_verifications = deepcopy(self.email_verifications)
+        for token in self.email_verifications:
+            try:
+                self.get_unconfirmed_email_for_token(token)
+            except (KeyError, ExpiredTokenError):
+                email_verifications.pop(token)
+                break
```
sloria: Why the `break`? Won't this mean that only one email can get cleaned?

Would be good to have tests for this method.

framework/auth/core.py
```diff
@@ -848,6 +854,19 @@ def _get_unconfirmed_email_for_token(self, token):
 
         return verification['email']
 
+    def clean_email_verifications(self, **kwargs):
+        given_token = kwargs.get('given_token')
```
sloria: Why pull from `kwargs`? The token should probably be a required argument:

```python
def clean_email_verifications(self, token):
     #...
```

tests/test_views.py
```diff
@@ -3413,7 +3415,122 @@ def test_resend_confirmation(self, send_mail):
...
+        remove_res = self.app.delete_json(remove_email_url, email_verifications[0], auth=self.user.auth)
+        self.user.reload()
+        assert_equal(remove_res.json_body['status'], 'success')
+        assert_equal(self.user.get_unconfirmed_emails, [])
+
+    def test_add_expired_email(self):
+        # Do not return expired token and removes it from user.email_verifications
+        email = 'test@example.com'
+        token = self.user.add_unconfirmed_email(email)
+        dt.datetime.utcnow() - dt.timedelta(days=100)
```
sloria: Minor: this line can be removed.

framework/auth/views.py
```diff
@@ -257,6 +301,64 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+        })
+    try:
+        user.confirm_email(token, merge=True)
+    except exceptions.EmailConfirmTokenError as e:
+        raise HTTPError(http.BAD_REQUEST, data={
+            'message_short': e.message_short,
+            'message_long': e.message_long
+        })
+
+    if token:
```
sloria: Won't this always be `True`? You're ensuring that token is provided ~L336. If this is the case you should be able to remove this conditional.

framework/auth/views.py
```diff
@@ -257,6 +301,64 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+    methods: PUT
+    """
+    user = auth.user
+    json_body = request.get_json()
+    try:
+        token = json_body['token']
+    except KeyError:
+        raise HTTPError(http.BAD_REQUEST, data={
+            'message_short': 'Invalid token',
+            'message_long': 'The token provided is invalid'
```
sloria: This is an inaccurate error--in this case, the user did not provide a token at all. The message should be `Must provide a token` or something similar.

framework/auth/views.py
```diff
@@ -257,6 +301,64 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+        })
+    try:
+        user.confirm_email(token, merge=True)
```
sloria: This may raise an `InvalidTokenError`. This needs to be handled.

website/static/js/pages/base-page.js
```diff
@@ -92,6 +93,100 @@ $(document).on('click', '.panel-heading', function(){
...
+        }
+        else {
+            title = 'Add email';
+            requestMessage = 'Would you like to add \<b>' + email.address + '\</b> to your account?';
+            confirmMessage = '\<b>' + email.address + '\</b> has been added into your account.';
+            nopeMessage = 'You have chosen not to add \<b>' + email.address + '\</b> to your account. ' +
+                'If you change your mind, visit the \<a href="/settings/account/">user settings page</a>.';
+        }
+
+        var confirmFailMessage = 'There are a problem adding \<b>' + email.address +
```
sloria: "There was"

website/static/js/pages/base-page.js
```diff
@@ -92,6 +93,100 @@ $(document).on('click', '.panel-heading', function(){
...
+            '\</b>. Please contact <a href="mailto: support@osf.io">support@osf.io</a> if the problem persists.';
+
+        var cancelFailMessage = 'There are a problem removing \<b>' + email.address +
```
sloria: "There was"

framework/auth/views.py
```diff
@@ -257,6 +301,64 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+        })
+    user.clean_email_verifications(given_token=given_token)
+    user.save()
+    return {
+        'status': 'success',
+        'removed_email': json_body['address']
+    }, 200
+
+
+@collect_auth
```
sloria: Minor: `collect_auth` isn't necessary when using `must_be_logged_in`; `auth` will be passed by `must_be_logged_in`.

sloria:
Pass finished.

framework/auth/core.py
billyhunt: I am using it in two contexts in views.py.  In unconfirmed_email_remove I am removing a specific one, in auth_login I am removing all the dead weight of expired tokens so it won't continue to poll for expired tokens on every page load.  Can you think of a better way to architect it?

sloria: Just put the argument in the signature:

```python
def clean_email_verifications(self, token=None):
```

framework/auth/views.py
```diff
@@ -257,6 +301,62 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+    except KeyError:
+        raise HTTPError(http.BAD_REQUEST, data={
+            'message_short': 'Invalid token',
+            'message_long': 'Must provide a token'
+        })
+    try:
+        user.confirm_email(token, merge=True)
+    except exceptions.InvalidTokenError:
+        raise InvalidTokenError(http.BAD_REQUEST, data={
+            'message_short': 'Invalid User Token',
```
sloria: Minor: Keep casing consistent; you use sentence case elsewhere.

framework/auth/views.py
```diff
@@ -257,6 +301,62 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+    """Called at login if user confirms their merge or email add.
+    methods: PUT
+    """
+    user = auth.user
+    json_body = request.get_json()
+    try:
+        token = json_body['token']
+    except KeyError:
+        raise HTTPError(http.BAD_REQUEST, data={
+            'message_short': 'Invalid token',
```
sloria: "Invalid token" is incorrect; how about "missing token".

framework/auth/views.py
```diff
@@ -257,6 +301,62 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+    """Called at login if user cancels their merge or email add.
+    methods: DELETE
+    """
+    user = auth.user
+    json_body = request.get_json()
+    try:
+        given_token = json_body['token']
+    except KeyError:
+        raise HTTPError(http.BAD_REQUEST, data={
+            'message_short': 'Invalid token',
```
sloria: These error messages are incorrect; should be "Missing token" and "Must provide a token".

sloria:
The placement of the "affirmative" vs. "cancel" button for the "Add email" modal is inconsistent with other places on the site.

osf___home

[osf-style's example modal](https://centerforopenscience.github.io/osf-style/style.html#modals)

osf_style___ui_guidelines

Create project

osf___home

framework/auth/core.py
```diff
@@ -1004,6 +1022,26 @@ def display_absolute_url(self):
     def deep_url(self):
         return '/profile/{}/'.format(self._primary_key)
 
+    @property
+    def get_unconfirmed_emails(self):
```
sloria: Not sure where my previous comments went, but this needs to be renamed . Properties should be named like attributes (nouns rather than actions). Also, the docstring needs to be updated.

sloria: Note: renaming to `unconfirmed_emails` isn't entirely accurate, since this returns a list of dicts, not a list of email addresses. Maybe `unconfirmed_email_info`?

framework/auth/core.py
```diff
@@ -1004,6 +1022,26 @@ def display_absolute_url(self):
...
+        """Called at login to see if there are emails to add or users to merge.  Delete expired tokens.
+        methods: GET
+        """
+        unconfirmed_emails = []
+        email_verifications = deepcopy(self.email_verifications)
```
sloria: This deepcopy is unnecessary.

sloria:
Pass finished.

sloria:
I'm seeing this at the home page

osf___home

This line is a suspect: https://github.com/CenterForOpenScience/osf.io/pull/5000/files#diff-611d5f96dde8ca8bcf4c5e6b0a3948a7R18

website/static/css/pages/login-page.css
```diff
@@ -38,3 +38,12 @@
     background-color: #CCC;
     border-color: #BBB;
 }
+.existing-user-signin {
+    /*height: 320px;*/
+    background-color: #EEE;
+    border-radius: 4px;
+    margin-top: 75px;
```
sloria: Why all this extra margin? We should rarely need to override Bootstrap's defaults. In this case, we can remove both the `border-radius` and `margin-top`.

Before:

osf___sign_in_and_login-page_css____projects_cos-projects_osf_website_static_css_pages__-_vim1

After:

osf___sign_in_and_login-page_css____projects_cos-projects_osf_website_static_css_pages__-_vim1

website/static/css/pages/login-page.css
```diff
@@ -38,3 +38,12 @@
...
+}
+.existing-user-signin-button {
+    margin-right: 0px !important;
```
sloria: Again, we should really avoid doing manual spacing like this and let Bootstrap do the job. Otherwise you run into responsiveness problems like this:

osf___sign_in

sloria: This isn't a blocker, but see if you can refactor the markup to use Bootstarp's grid and get rid of this custom CSS.

website/static/js/signIn.js
```diff
@@ -15,10 +15,19 @@ var SignInViewModel = oop.extend(formViewModel.FormViewModel, {
...
-        });
+        var existingUserEmail = decodeURIComponent($osf.urlParams().existing_user);
+        if (existingUserEmail) {
+            self.username = ko.observable(existingUserEmail).extend({
+                required: true,
+                email: true
+            });
+        }
+        else {
+            self.username = ko.observable('').extend({
```
sloria: You can DRY this up a bit:

```js
var existingUser = $osf.urlParams().existing_user
self.username = ko.observable(existingUser ? decodeURIComponent(existingUser) : '').extend({
    required: true,
    email: true
});
```

website/templates/base.mako
```diff
@@ -188,10 +188,13 @@
...
                     locale: ${ user_locale | sjson, n },
                     timezone: ${ user_timezone | sjson, n },
-                    entryPoint: ${ user_entry_point | sjson, n }
+                    entryPoint: ${ user_entry_point | sjson, n },
+                    emailsToAdd: ${ user_email_verifications | sjson, n }
                 },
                 popular: ${ popular_links_node | sjson, n },
-                newAndNoteworthy: ${ noteworthy_links_node | sjson, n }
+                newAndNoteworthy: ${ noteworthy_links_node | sjson, n },
+                confirmedEmailURL: ${ api_url_for('unconfirmed_email_add') | sjson, n }
```
sloria: Minor: This probably isn't necessary, especially because there are no path variables in the URL. It's just `/api/v1//confirmed_emails/`.

sloria:
Pass finished. 

- https://github.com/CenterForOpenScience/osf.io/pull/5000#issuecomment-223392730 is a blocking bug
- https://github.com/CenterForOpenScience/osf.io/pull/5000#discussion_r65601847, https://github.com/CenterForOpenScience/osf.io/pull/5000#discussion_r65603556, and https://github.com/CenterForOpenScience/osf.io/pull/5000#discussion_r65603991 are quick changes
- https://github.com/CenterForOpenScience/osf.io/pull/5000#discussion_r65602730 is worth investigating, but isn't a blocker

website/templates/base.mako
sloria: We want to be judicious with our use of contextVars. In this case, just save the URL to a constant

```js
function confirmEmails(emailsToAdd) {
    if (emailsToAdd.length > 0) {
        var CONFIRMED_EMAILS_URL = '/api/v1/confirmed_emails';
//...
```

website/static/js/signIn.js
```diff
@@ -15,19 +15,10 @@ var SignInViewModel = oop.extend(formViewModel.FormViewModel, {
     constructor: function () {
         var self = this;
         self.super.constructor.call(self);
-        var existingUserEmail = decodeURIComponent($osf.urlParams().existing_user);
```
sloria: If we're removing this, then we don't need to pass this query param anymore:

osf___sign_in

billyhunt: Didn't I remove it?

sloria: It's still showing up in the URL.

sloria: https://github.com/CenterForOpenScience/osf.io/pull/5000/files#diff-9a64b4982d0bd80f7a498c65ddce2023R206

website/static/js/pages/base-page.js
```diff
@@ -92,6 +93,100 @@ $(document).on('click', '.panel-heading', function(){
...
+                    className: 'btn-default',
+                    callback: function () {
+                        $.ajax({
+                            type: 'delete',
+                            url: confirmedEmailURL,
+                            contentType: 'application/json',
+                            dataType: 'json',
+                            data: JSON.stringify(email)
+                        }).done(function () {
+                            $osf.growl('Warning', nopeMessage, 'warning', 8000);
```
sloria: Minor: Is this necessary? Showing a growl after a "cancel" operation seems superfluous.

website/static/js/pages/base-page.js
```diff
@@ -92,6 +93,100 @@ $(document).on('click', '.panel-heading', function(){
...
+                },
+                confirm: {
+                    label: 'Add email',
+                    className: 'btn-success pull-right',
+                    callback: function () {
+                        $osf.putJSON(
+                            confirmedEmailURL,
+                            email
+                        ).done(function () {
+                            $osf.growl('Success', confirmMessage, 'success', 3000);
```
sloria: Minor: This growl message may be superfluous.

billyhunt: How so?

sloria: It obscures the navbar while it's visible, requiring the user to dismiss it to reach the navbar links. But I see now that it's only visible for 3 seconds, so that mitigates a bit. 

Again, really minor, and not a blocker.

sloria:
Created a new feature, branch with this, `feature/logout-users`. Consider this merged. Let's get this to staging!

Send any fixes to that branch.

billyhunt:
tumblr_ni5uj9cslw1t5h21vo6_500

PR #5001, opened 2016-02-08
[Feature] Admin action logs

Ghalko:
## Proposal
- Add logs from admin user's actions
- Display admin user's actions

## Repercussions
- Can make very large items on the logs list.


cwisecarver:
You want to look at the “A few special cases to note about list_display” section of the [django docs](https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_display) regarding the admin site and escaping html.

admin/common_auth/admin.py
```diff
@@ -1,9 +1,14 @@
...
+from django.contrib.admin.models import DELETION
 from django.contrib.auth.admin import UserAdmin
-from django.contrib.auth.models import Permission
 from django.contrib.auth.forms import PasswordResetForm
-from .models import MyUser
+from django.contrib.auth.models import Permission
+from django.core.urlresolvers import reverse
+from django.utils.html import escape
+
+from .logs import OSFLogEntry
```
mfraezz: Prefer full module reference over local reference notation

admin/pre_reg/views.py
```diff
@@ -118,6 +121,10 @@ def approve_draft(request, draft_pk):
 
     user = request.user.osf_user
     draft.approve(user)
+    update_admin_log(
+        request.user.id, draft._id, 'Draft Registration',
+        'approved', ACCEPT_PREREG
```
mfraezz: The `action_flag` kwarg should be passed in with its keyword

admin/spam/views.py
```diff
@@ -101,24 +98,38 @@ def post(self, request, *args, **kwargs):
...
+        item = Comment.load(self.kwargs.get('spam_id', None))
         kwargs = super(SpamDetail, self).get_context_data(**kwargs)
         kwargs.setdefault('page_number', self.request.GET.get('page', 1))
-        kwargs.setdefault('comment', serialize_comment(self.item))
+        kwargs.setdefault('comment', serialize_comment(item))
         kwargs.setdefault('status', self.request.GET.get('status', u'1'))
         return kwargs
 
     def form_valid(self, form):
+        spam_id = self.kwargs.get('spam_id', 'no find')
```
mfraezz: This should probably raise some exception if `spam_id` is missing

admin/spam/views.py
```diff
@@ -101,24 +98,38 @@ def post(self, request, *args, **kwargs):
...
+        item = Comment.load(spam_id)
         if int(form.cleaned_data.get('confirm')) == Comment.SPAM:
-            self.item.confirm_spam(save=True)
+            item.confirm_spam(save=True)
+            log_message = 'Confirmed SPAM: {}'.format(spam_id)
+            log_action = CONFIRM_SPAM
         else:
-            self.item.confirm_ham(save=True)
+            item.confirm_ham(save=True)
+            log_message = 'Confirmed HAM: {}'.format(spam_id)
```
mfraezz: These logs are only visible in the admin interface, correct? We have no plans of ever attaching them in a user-facing way?

admin/common_auth/logs.py
```diff
@@ -0,0 +1,34 @@
...
+    try:
+        OSFLogEntry.objects.log_action(
+            user_id=user_id,
+            content_type_id=None,
+            object_id=object_id,
+            object_repr=object_repr,
+            change_message=message,
+            action_flag=action_flag
+        )
+    except:
```
mfraezz: What exceptions is this expected to raise?

admin/common_auth/logs.py
```diff
@@ -0,0 +1,34 @@
...
+        print 'Failed to log changes to {}'.format(object_id)
+
+
+class OSFLogEntryManager(LogEntryManager):
```
mfraezz: This class appears superfluous

admin/pre_reg/views.py
```diff
@@ -160,11 +170,22 @@ def update_draft(request, draft_pk):
     else:
         schema_data = data.get('schema_data', {})
         data = deepcopy(draft.registration_metadata)
+        log_message = list()
         for key, value in data.items():
```
mfraezz: This `for` seems to be losing comment data. Specifically, nested comments aren't retrieved.

admin/pre_reg/views.py
```diff
@@ -160,11 +170,22 @@ def update_draft(request, draft_pk):
...
-            data[key]['comments'] = schema_data.get(key, {}).get('comments', [])
+            comment = schema_data.get(key, {}).get('comments', [])
+            data[key]['comments'] = comment
+            if comment != []:
+                log_message.append('{}: {}'.format(key, comment[0]['value']))
```
mfraezz: Change `comment` above to `comments`, and replace this line with
```python
for comment in comments:
    log_message.append('{}: {}'.format(key, comment['value']
```

mfraezz:
Pass complete. Please also add some prereg tests, at least to cover the added logging functionality. Back to you, @Ghalko :metal:

admin/common_auth/admin.py
Ghalko: Done, maybe overkill.

admin/pre_reg/views.py
Ghalko: Done.

admin/spam/views.py
Ghalko: This gets caught by the `AttributeError` try/except block in post.

admin/spam/views.py
Ghalko: Correct.

admin/common_auth/logs.py
Ghalko: The `OSFLogEntry` class did not like it when it didn't have its very own Manager class rather than just using the inherited class. Hence why this exists.

admin/pre_reg/views.py
Ghalko: How far nested can they go?

mfraezz: It's just a list, but `comment[0]['value']` only gets the first one.

admin/spam/views.py
mfraezz: So why even add a default?

admin/common_auth/logs.py
mfraezz: Weird django inheritance behavior, I'll accept that.

mfraezz:
LGTM. Dropping this in RTM :cake:

brianjgeiger:
Out of scope for this, but I'd like to propose that NodeLogs be created for any changes to a Node object, even if they're marked as non-showing (I forget what the flag is, but there is one) so that provenance is maintained even if regular users can't see it as things stand. Maybe a follow-up issue?

mfraezz:
`should_hide` is the flag you're thinking of.

brianjgeiger:
Thanks!

Ghalko:
Noted in current ticket. ADM-82

PR #5003, opened 2016-02-08
Make API filtering on a charfield be case insensitive

erinspace:
Relates to https://openscience.atlassian.net/browse/OSF-5029

# Purpose
Filtering by filename in the API fails if you don't use the correct case.

As described on Jira by @brianjgeiger:

https://staging-api.osf.io/v2/nodes/ck8hm/files/s3/?filter[name]=Renamed will always fail, even if there's a file named Renamed, but https://staging-api.osf.io/v2/nodes/ck8hm/files/s3/?filter[name]=renamed will succeed. 

# Changes
- When parsing the filter on a CharField, the API will make the query term lowercase
- Add test to make sure that a search of the incorrect case returns the desired results

# SideEffects
If any API filtering was relying on exact case of search terms for a charField, this will break.


sloria:
@erinspace There's a lot of noise in the history for this PR. Can you please resubmit a new PR with commits containing only your changes?

PR #5013, opened 2016-02-10
[Feature] Admin User Specific Spam

Ghalko:
## Proposal
Connect user profile to spam list so that an administrator can see the flagged/marked/ham from the user in order to evaluate the user's spam potential. ADM-62

## Repercussions
More connectivity == more power

## Order
After #4951


admin/spam/views.py
```diff
@@ -1,47 +1,142 @@
...
+        self.page = request.GET.get('page', 1)
+        context['page_number'] = self.page
+        self.status = request.GET.get('status', 1)
+        context['status'] = self.status
+        context['form'] = self.get_form()
+        if context['form'].is_valid():
+            return self.form_valid(context['form'])
+        else:
+            return render(request, self.template_name, context=context)
+        # return super(SpamDetail, self).post(request, *args, **kwargs)
```
mfraezz: Please remove this comment

admin/spam/views.py
```diff
@@ -1,47 +1,142 @@
...
+
+    def get_context_data(self, **kwargs):
+        self.spam_id = kwargs['spam_id']
```
mfraezz: Could this throw an uncaught `KeyError`?

admin/templates/nodes/node.html
```diff
@@ -0,0 +1,111 @@
...
+                    {% endfor %}
+                    </tbody>
+                    </table>
+                </td>
+            </tr>
+        </tbody>
+        </table>
+        </div>
+    </div>
+{% endblock content %}
```
mfraezz: Newline please

admin_tests/pre_reg/test_views.py
```diff
@@ -1,110 +0,0 @@
...
-from website.models import MetaSchema
-from website.project.model import ensure_schemas
-
-from admin.common_auth.models import MyUser as User
-from admin.pre_reg.views import *  # noqa
-
-PREREG_GROUP = Group.objects.get(name='prereg_group')
-
-
-class PreregViewsTests(AdminTestCase):
```
mfraezz: Why was this file removed?

admin/base/forms.py
```diff
@@ -0,0 +1,5 @@
+from django import forms
+
+
+class GuidForm(forms.Form):
+    guid = forms.CharField(label='Guid', max_length=5, required=True)
```
mfraezz: This should probably specify `min_length=5` too, unless we have any guids less than length 5 (Institutions, maybe?).

Also, it might be good to note that `max_length` will need to be updated whenever we bump the guid length up to 6.

admin_tests/pre_reg/test_views.py
Ghalko: It didn't actually run and wasn't included in Travis for that reason. I commented it out, but it was pointed out that we could just find it in the history of commits and just to delete it for now.

admin/templates/spam/spam.html
```diff
@@ -1,5 +1,6 @@
 {% extends "base.html" %}
```
mfraezz: This file might make more sense named `spam/spam_list.html`. The naming convention in other template directories is that `<model_name>.html` is the view for a `<model>` instance, whereas this is the view for a list of spam.

admin/templates/users/spam.html
```diff
@@ -0,0 +1,85 @@
+{% extends "base.html" %}
```
mfraezz: Would it make more sense for this file to be `/templates/spam/user.html`? We have a `/spam/comment.html` template, not `/comments/spam.html`.

admin/templates/util/pagination.html
```diff
@@ -3,7 +3,7 @@
     <div class="pagination pagination-lg">
         <span>
             {% if items.has_previous %}
-                <a href="?page={{ items.previous_page_number }}"
+                <a href="?page={{ items.previous_page_number }}&status={{ status }}"
```
mfraezz: I'm not sure if this template has implications for non-spam views (maybe pre-reg), but will status always be passed in here?

admin/users/serializers.py
```diff
@@ -0,0 +1,23 @@
+"""
+Serialize user
+"""
+# from admin.nodes.serializers import serialize_simple_user
```
mfraezz: Please remove.

admin_tests/pre_reg/test_views.py
mfraezz: :+1:

admin_tests/nodes/test_views.py
```diff
@@ -0,0 +1,38 @@
...
+            view.get_object()
+
+    def test_load_data(self):
+        node = NodeFactory()
+        guid = node._id
+        request = RequestFactory().get('/fake_path')
+        view = NodeView()
+        view = setup_view(view, request, guid=guid)
+        res = view.get_object()
+        assert_is_instance(res, dict)
```
mfraezz: Please add a few more checks here to ensure you get back the expected result, not just a `dict`. Likewise for other tests of this form.

mfraezz:
Looks good so far, just a few minor issues that need to be addressed. Manual testing revealed no bugs :star: 

Back to you, @Ghalko

admin/spam/views.py
Ghalko: Removed

admin/spam/views.py
Ghalko: Made into a get with default `None`

admin/base/forms.py
Ghalko: Done.

admin/templates/users/spam.html
Ghalko: files moved/renamed

admin/users/serializers.py
Ghalko: removed

admin_tests/nodes/test_views.py
Ghalko: This is tested in test_serializers.py

admin/templates/util/pagination.html
Ghalko: I think this is best left for when we look into refactoring the prereg into a CBV rather than FBV. Definitely a worthwhile consideration.

admin_tests/nodes/test_views.py
mfraezz: You're right. I think I thought this was spam-related, for some reason.

On that note: why are there no spam tests?

admin/templates/util/pagination.html
mfraezz: :+1:

mfraezz:
Please add a few `spam` tests. If not ones for coverage of the whole module, at least for this feature.

Ghalko:
Writing tests for spam in #5045, I would rather not write the same tests for both of these PRs.

mfraezz:
LGTM. Putting this in RTM :crocodile:

sloria:
@Ghalko Can you please get this up-to-date with develop one last time? Once that is done, this is good to merge.

admin/base/views.py
```diff
@@ -74,3 +57,15 @@ def form_valid(self, form):
...
+
+class GuidView(DetailView):
+    def __init__(self):
+        self.guid = None
+        super(GuidView, self).__init__()
+
+    def get(self, request, *args, **kwargs):
+        try:
+            return super(GuidView, self).get(request, *args, **kwargs)
+        except AttributeError:
```
sloria: Why is an `AttributeError` expected here?

admin/nodes/urls.py
```diff
@@ -3,5 +3,8 @@
 from . import views
 
 urlpatterns = [
-    url(r'^$', views.NodeFormView.as_view(), name='node'),
+    url(r'^$', views.NodeFormView.as_view(),
+        name='search'),
+    url(r'^id-(?P<guid>[a-z0-9]+)/$', views.NodeView.as_view(),
```
sloria: Why is the `id-` prefix necessary?

admin/spam/urls.py
```diff
@@ -3,7 +3,10 @@
 from . import views
 
 urlpatterns = [
-    url(r'^$', views.spam_list, name='spam'),
-    url(r'^(?P<spam_id>[a-z0-9]+)/$', views.spam_detail, name='detail'),
-    url(r'^(?P<spam_id>[a-z0-9]+)/email/$', views.email, name='email'),
+    url(r'^$', views.SpamList.as_view(), name='spam'),
+    url(r'^user/id-(?P<user_id>[a-z0-9]+)/$', views.UserSpamList.as_view(),
```
sloria: Again, the `id-` prefix seems unnecessary.

admin/spam/views.py
```diff
@@ -1,47 +1,141 @@
...
-        'comment': serialize_comment(Comment.load(spam_id)),
-        'page_number': request.GET.get('page', 1),
-    }
-    return render(request, 'spam/comment.html', context)
+
+class UserSpamList(SpamList):
+    template_name = 'spam/user.html'
+
+    def __init__(self):
+        self.user_id = None
```
sloria: Why is it necessary to set `user_id` as an attribute; you can always access it from `self.kwargs`.

sloria: I'm seeing this pattern in other views, too. It seems unnecessary to override `__init__` in most cases.

admin/nodes/urls.py
Ghalko: It really isn't. Something that I have picked up, but it is across so many PRs at this point... I would rather change them all at once, along with a good restructuring of many urls.

admin/spam/views.py
Ghalko: Started fixing this in the PRs that I update.

admin/base/views.py
Ghalko: This is where the class will actually return a page_not_found, either in this PR or in subsequent ones.

Ghalko:
@sloria Responded to your concerns. Looks like this is up-to-date and all checks passed.

admin/nodes/urls.py
sloria: :+1:

PR #5015, opened 2016-02-10
[feature-fix] Bibliographic/Non-bibliographic Contributor Logs 

caseyrollins:
- Changes log language from visible/invisible to bibliographic/non-bibliographic.
- Adds custom log language for contributor visibility changes in anonymous view only links.
- Language changes made according to [OSF-5331].

### DIFF
--- a/website/templates/log_list.mako
+++ b/website/templates/log_list.mako
@@ -55,9 +55,19 @@
                                     </span>
                                 </span>
                                 <!-- Log actions are the same as their template name -->
-                                <span data-bind="template: {name: log.action, data: log}"></span>
+                                    <!-- Handle specific language for contributor bibliographic/non-bibliographic on anonymous VOLs -->
+                                    <!-- ko if: (log.action == 'made_contributor_visible' && log.anonymous -->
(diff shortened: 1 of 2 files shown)


website/templates/log_list.mako
```diff
@@ -55,9 +55,19 @@
                                     </span>
                                 </span>
                                 <!-- Log actions are the same as their template name -->
-                                <span data-bind="template: {name: log.action, data: log}"></span>
```
sloria: I've reverted the change in this template because it caused all logs to be the same ("changed a non-bibliographic contributor..."). I'll have QA retest and see if the original issue is still addressed.

website/templates/log_list.mako
```diff
@@ -55,9 +55,19 @@
...
+                                    <!-- Handle specific language for contributor bibliographic/non-bibliographic on anonymous VOLs -->
+                                    <!-- ko if: (log.action == 'made_contributor_visible' && log.anonymous -->
```
caseyrollins: @sloria I think the errant parentheses on this line caused the problem. Fixing it now and then sending a new PR.

PR #5016, opened 2016-02-10
[Feature Fix] OSF links from admin

Ghalko:
Use furl to build from local.py

## Deploy

website.settings DOMAIN should match the actual running osf domain for the deployment


sloria:
Can you use the `settings.DOMAIN` constant that already exists?

Ghalko:
@sloria Fixed.

PR #5017, opened 2016-02-10
Feature/inst2

hmoco:
Combines: https://github.com/CenterForOpenScience/osf.io/pull/4994, https://github.com/CenterForOpenScience/osf.io/pull/4988, https://github.com/CenterForOpenScience/osf.io/pull/4972,
and non lading-page piece of https://github.com/CenterForOpenScience/osf.io/pull/4970


api/institutions/serializers.py
```diff
@@ -12,6 +12,7 @@ class InstitutionSerializer(JSONAPISerializer):
     name = ser.CharField(required=False)
     id = ser.CharField(required=False, source='_id')
     logo_path = ser.CharField()
+    auth_url = ser.CharField()
```
sloria: Is this field required? If so, explicitly pass `required=True`. Also it wouldn't hurt to pass in `help`.

website/settings/defaults.py
```diff
@@ -330,3 +330,5 @@ def parent_dir(path):
 assert (DRAFT_REGISTRATION_APPROVAL_PERIOD > EMBARGO_END_DATE_MIN), 'The draft registration approval period should be more than the minimum embargo end date.'
 
 PREREG_ADMIN_TAG = "prereg_admin"
+
+INSTITUTION_HIDE_UI_FLAG = True
```
sloria: How about `ENABLE_INSTITUTIONS` for consistency with our other feature flags. Also, disabling this should disable the appropriate API endpoint(s), correct (not only hiding the UI)?

website/routes.py
```diff
@@ -75,6 +75,7 @@ def get_globals():
         'login_url': cas.get_login_url(request.url, auto=True),
         'reauth_url': util.web_url_for('auth_logout', redirect_url=request.url, reauth=True),
         'profile_url': cas.get_profile_url(),
+        'show_institutions': not settings.INSTITUTION_HIDE_UI_FLAG,
```
sloria: As suggested in my previous comment, this can be changed to `enable_institutions` for consistency with the settings constant.

website/profile/views.py
```diff
@@ -822,3 +822,13 @@ def redirect_to_twitter(twitter_handle):
         })
 
     return redirect(user.url)
+
+def view_institution(**kwargs):
+    from website.models import Institution
```
sloria: Is it necessary to do this import here? If so, please add a comment explaining why (e.g. `# Avoid circular import`).

website/settings/defaults.py
hmoco: As suggested by Michael, this is a specifically and explicitly for ui hiding. The api endpoints don't need to be hidden (I would think)

website/project/model.py
```diff
@@ -2217,6 +2217,7 @@ def register_node(self, schema, auth, data, parent=None):
         registered.logs = self.logs
         registered.tags = self.tags
         registered.piwik_site_id = None
+        registered.primary_institution = self.primary_institution
```
sloria: Would be good to have a regression test for this somewhere in this test class: https://github.com/CenterForOpenScience/osf.io/blob/8b1e7498253343db3660802a7be82610147c4374/tests/test_models.py#L3776

website/static/js/pages/project-settings-page.js
```diff
@@ -54,8 +54,73 @@ if ($('#wgrid').length) {
         });
     });
 }
+
+
+var institutionsViewModel = function() {
```
sloria: In general, try to keep the `*-page.js` modules relatively "thin"--they should only contain initialization code. ViewModels and other reusable code should be moved to their own modules.

> - Reuseable modules go in website/static/js/. Name modules in lowerCamelCase.
> - Initialization code for a page goes in a module within website/static/js/pages/. Name page modules with lower-dashed-case.

https://cosdev.readthedocs.org/en/latest/osf/js_modules.html#osf-js-modules

sloria: Also, "classes" should be `UpperCamelCase`.

website/static/js/pages/project-settings-page.js
```diff
@@ -54,8 +54,73 @@ if ($('#wgrid').length) {
...
+    ko.punches.enableAll();
```
sloria: Is this needed? I don't see any usage of knockout-punches syntax.

website/static/js/pages/project-settings-page.js
```diff
@@ -54,8 +54,73 @@ if ($('#wgrid').length) {
...
+    var self = this;
+    self.primaryInstitution = ko.observable('None');
+    self.institutionHref = ko.observable('');
+    self.availableInstitutions = ko.observable();
+    $osf.ajaxJSON(
```
sloria: Avoid making requests in constructors; they make classes hard to test.

```js
// I want to unit test InstitutionsViewModel, let's instantiate one
var vm = new InstitutionsViewModel();
// Ah! it's sending an API request. Now I need to mock stuff... ='(
```

Better to separate fetching into methods.

```js
var vm = new InstitutionsViewModel();
vm.fetchUserInstitutions(userID)
    .done(function(institutions) { console.log(institutions); });
```

website/static/js/pages/project-settings-page.js
```diff
@@ -54,8 +54,73 @@ if ($('#wgrid').length) {
...
+        'GET',
+        ctx.apiV2Prefix + 'users/' + ctx.currentUser.id + '/?embed=institutions',
+        {isCors: true}
+    ).done(function(response) {
+        self.availableInstitutions(response.data.embeds.institutions.data);
+    });
```
sloria: Add proper error handling to both these requests (log to Sentry)

website/static/js/pages/project-settings-page.js
```diff
@@ -54,8 +54,73 @@ if ($('#wgrid').length) {
...
+        {isCors: true}
+    ).done(function(response) {
+        if (response.data.embeds.primary_institution.data){
+            self.primaryInstitution(response.data.embeds.primary_institution.data.attributes.name);
+            self.institutionHref(response.data.embeds.primary_institution.data.links.html);
+        }
+    }).fail(function(){});
+    self.submitInst = function() {
+        var inst = $('input[name=primaryInst]:checked', '#selectedInst').val();
+        $osf.ajaxJSON(
```
sloria: Any method that does async work should return a promise. In this case, you can return the return value of ajaxJSON.

website/static/js/pages/project-settings-page.js
```diff
@@ -54,8 +54,73 @@ if ($('#wgrid').length) {
...
+                fields: {xhrFields: {withCredentials: true}}
+            }
+        ).done(function (response) {
+            window.location.reload();
+        }).fail(function (response) {
+            $osf.growl('Error!');
+        });
+    };
+    self.clearInst = function() {
+        $osf.ajaxJSON(
```
sloria: See https://github.com/CenterForOpenScience/osf.io/pull/5017/files#r52519501

website/static/js/pages/login-page.js
```diff
@@ -20,6 +23,45 @@ var activateToggleBox = function () {
     }
 };
 
+
+var InstitutionViewModel = function() {
+    var self = this;
+    self.inst_names = ko.observableArray([]);
```
sloria: As per COSDev, use `lowerCamelCase` for variable names.

website/static/js/pages/login-page.js
```diff
@@ -20,6 +23,45 @@ var activateToggleBox = function () {
...
+        for (var i = 0; i < response.data.length; i++){
+            var name = response.data[i].attributes.name;
+            self.inst_names.push(name);
+            self.insts[name] = response.data[i].attributes.auth_url;
+        }
+    }).fail(function(response){
+    });
+
+    self.instLogin = function(){
+        var key = $('#selected_inst').val();
```
sloria: Prefer data-binding to reading from the DOM like this. You should be able to remove those CSS IDs and use an observable instead.

website/static/js/pages/login-page.js
```diff
@@ -20,6 +23,45 @@ var activateToggleBox = function () {
...
+    $osf.ajaxJSON(
+        'GET',
+        window.contextVars.apiV2Prefix + 'institutions/',
+        {
+            isCors: true
+        }
+    ).done( function(response){
+        for (var i = 0; i < response.data.length; i++){
+            var name = response.data[i].attributes.name;
+            self.inst_names.push(name);
```
sloria: Is this working? IIRC, you need to invoke an observable array to access it:

```js
self.instNames().push(name)
```

sloria: That said, you can refactor this using Array.prototype.map.

```js
self.instNames(response.data.map(function(datum) { return datum.name; }););
```

website/static/js/pages/login-page.js
```diff
@@ -1,15 +1,18 @@
...
+'use strict';
 var $ = require('jquery');
+var ko = require('knockout');
+
+var $osf = require('js/osfHelpers');
 var SignUp = require('js/signUp');
 var LogInForm = require('js/signIn');
 
-new LogInForm.SignIn('#logInForm');
+ko.punches.enableAll();
```
sloria: Again, are you using punches? If not, remove this.

website/static/js/pages/login-page.js
```diff
@@ -20,6 +23,45 @@ var activateToggleBox = function () {
...
+    self.instLogin = function(){
+        var key = $('#selected_inst').val();
+        window.location = self.insts[key];
+    };
+};
+
+
+$(document).ready(function() {
+    var self = this;
+    self.campaign = $('#campaign').val();
```
sloria: The mix of jQuery, Knockout, and mako (https://github.com/CenterForOpenScience/osf.io/pull/5017/files#diff-6aa4ed71920a03b1327bfba3519f9ca0R45) is kludgy. Can you consolidate these and use data-binding instead?

website/static/js/pages/login-page.js
```diff
@@ -20,6 +23,45 @@ var activateToggleBox = function () {
...
+    if (self.campaign === 'institution'){
+        self.viewModel = new InstitutionViewModel();
```
sloria: Why are you assigning to `self`/`this`?

sloria:
I need to pause my pass here. Go ahead an address the comments I've made so far; I will continue my pass tomorrow morning.

hmoco:
@sloria I believe I have responded to your current set of comments

framework/auth/core.py
```diff
@@ -1389,6 +1404,7 @@ class Institution(StoredObject):
     _id = fields.StringField(index=True, unique=True, primary=True)
     name = fields.StringField(required=True)
     logo_name = fields.StringField(required=True)
+    auth_url = fields.StringField(required=False)
```
sloria: Can you add validation to this field? modm has a built-in `URLValidator`.

api/users/views.py
```diff
@@ -425,3 +427,44 @@ def get_default_odm_query(self):
...
+        user = self.request.user
+        current_institutions = {inst._id for inst in user.affiliated_institutions}
+
+        # DELETEs normally dont get type checked
+        # not the best way to do it, should be enforced everywhere, maybe write a test for it
+        for val in data:
+            if val['type'] != self.serializer_class.Meta.type_:
+                raise Conflict()
+        for val in data:
+            if val['id'] in current_institutions:
```
sloria: Is it possible for a `KeyError` to be raised here?

api/users/views.py
```diff
@@ -425,3 +427,44 @@ def get_default_odm_query(self):
...
+
+    def perform_destroy(self, instance):
+        data = self.request.data['data']
+        user = self.request.user
+        current_institutions = {inst._id for inst in user.affiliated_institutions}
+
+        # DELETEs normally dont get type checked
+        # not the best way to do it, should be enforced everywhere, maybe write a test for it
+        for val in data:
+            if val['type'] != self.serializer_class.Meta.type_:
```
sloria: Can a KeyError be raised here?

website/routes.py
```diff
@@ -75,6 +75,7 @@ def get_globals():
         'login_url': cas.get_login_url(request.url, auto=True),
         'reauth_url': util.web_url_for('auth_logout', redirect_url=request.url, reauth=True),
         'profile_url': cas.get_profile_url(),
+        'enable_institutions': settings.ENABLE_INSTITUTION_UI_FLAG,
```
sloria: Minor: it would make sense to make the settings constant consistent with this variable name, so `ENABLE_INSTITUTIONS`.

website/static/js/institutionProjectSettings.js
```diff
@@ -0,0 +1,104 @@
...
+            Raven.captureMessage('Error creating comment', {
+                url: url,
+                status: status,
+                error: error
+            });
+        });
+    };
+    self.submitInst = function() {
+        var url = ctx.apiV2Prefix + 'nodes/' + ctx.node.id + '/relationships/institution/';
+        var inst = $('input[name=primaryInst]:checked', '#selectedInst').val();
```
sloria: You should be able to use data-binding instead of jQuery for this.

website/static/js/pages/login-page.js
```diff
@@ -20,6 +20,17 @@ var activateToggleBox = function () {
     }
 };
 
+$(document).ready(function() {
+    var self = this;
+    self.campaign = $('#campaign').val();
```
sloria: My previous comment disappeared, but I suggested that you refactor this to use data-binding rather than mixing Knockout, jQuery, and Mako. Also, why are you assigning to `self`?

website/static/js/pages/project-settings-page.js
```diff
@@ -54,8 +55,14 @@ if ($('#wgrid').length) {
         });
     });
 }
+
 $(document).ready(function() {
+    var self = this;
     // Apply KO bindings for Project Settings
+    self.institutionProjectSettings = new InstitutionProjectSettings('#institutionSettings');
```
sloria: Why assign to `self` here? `self` isn't very useful in this context.

framework/auth/core.py
```diff
@@ -437,6 +437,14 @@ def absolute_api_v2_url(self):
     def get_absolute_url(self):
         return self.absolute_api_v2_url
 
+    @property
+    def institutions_self_url(self):
```
sloria: Minor: These properties are unnecessary--the `User` and `Node` classes would get pretty bloated if we had a two properties for each relationship.

framework/auth/core.py
```diff
@@ -1361,6 +1369,13 @@ def has_inst_auth(self, inst):
             return True
         return False
 
+    def remove_inst(self, inst_id):
```
sloria: Avoid abbreviations in method names.

framework/auth/core.py
```diff
@@ -1361,6 +1369,13 @@ def has_inst_auth(self, inst):
...
+        inst = Institution.load(inst_id)
```
sloria: Is it necessary to load the whole `Institution` record here? I believe you can pass a PK to `remove`, but you should try it out.

website/static/js/nodeControl.js
```diff
@@ -46,6 +46,18 @@ var ProjectViewModel = function(data) {
     self.user = data.user;
     self.nodeIsPublic = data.node.is_public;
     self.nodeType = data.node.node_type;
+    self.instLogoPath = ko.observable('');
+
+    if (data.node.institution) {
+        $osf.ajaxJSON(
```
sloria: As mentioned in a previous comment, avoid HTTP work upon initialization of ViewModels. Instead, put it in a method. You could then invoke the method in `NodeControl.prototype.init`.

website/static/js/institutionProjectSettings.js
```diff
@@ -0,0 +1,104 @@
...
+                url: url,
+                status: status,
+                error: error
+            });
+        });
+    };
+};
+
+var InstitutionProjectSettings = function(selector)  {
+    this.ViewModel = new ViewModel();
```
sloria: Use `lowerCamelCase` for attribute names.

website/static/js/institutionProjectSettings.js
```diff
@@ -0,0 +1,104 @@
...
+    $osf.applyBindings(this.ViewModel, selector);
+    return this.ViewModel;
```
sloria: Minor: This line is unnecessary.

website/static/js/institutionSignIn.js
```diff
@@ -0,0 +1,36 @@
...
+        {
+            isCors: true
+        }
+    ).done( function(response){
+        for (var i = 0; i < response.data.length; i++){
+            var name = response.data[i].attributes.name;
+            self.instNames.push(name);
+            self.insts[name] = response.data[i].attributes.auth_url;
+        }
+    }).fail(function(response){
```
sloria: Add proper error handling.

website/static/js/institutionSignIn.js
```diff
@@ -0,0 +1,36 @@
...
+var ko = require('knockout');
+
+var $osf = require('js/osfHelpers');
+
+var ViewModel = function() {
+    var self = this;
+    self.instNames = ko.observableArray([]);
+    self.selectedInst = ko.observable();
+    self.insts = {};
+    $osf.ajaxJSON(
```
sloria: As mentioned in a previous comment, avoid making HTTP requests upon instantiation.

sloria: As mentioned in a previous comment, avoid making HTTP requests upon instantiation.

website/static/js/institutionProjectSettings.js
hmoco: wouldnt this be needed to call the fetch requests on  ` -page.js `

website/static/js/institutionSignIn.js
```diff
@@ -0,0 +1,36 @@
...
+        'GET',
+        window.contextVars.apiV2Prefix + 'institutions/',
+        {
+            isCors: true
+        }
+    ).done( function(response){
+        for (var i = 0; i < response.data.length; i++){
```
sloria: Ah! I think a bunch of my previous comments got lost when this code was moved to a new file. I had suggested refactoring this to use `response.data.map`.

sloria:
Pass finished.

api/users/views.py
hmoco: Don't believe so, just wrote tests for it

hmoco:
@sloria, I forgot to ping. Have responded.

website/project/model.py
```diff
@@ -852,15 +855,43 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin):
     # Primary institution node is attached to
     primary_institution = fields.ForeignField('institution')
 
+    affiliated_institutions = fields.ForeignField('institution', list=True)
+
     def add_primary_institution(self, user, inst):
         if inst.auth(user):
             self.primary_institution = inst
+            self.affiliated_institutions.append(inst)
             self.save()
```
sloria: We generally try to avoid implicit saves. This allows us to save computation when making multiple updates at a time

```python
node.add_primary_institution(...)
node.do_something_else()
node.save()
```

website/project/model.py
```diff
@@ -852,15 +855,43 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin):
     # Primary institution node is attached to
     primary_institution = fields.ForeignField('institution')
 
+    affiliated_institutions = fields.ForeignField('institution', list=True)
+
     def add_primary_institution(self, user, inst):
         if inst.auth(user):
             self.primary_institution = inst
+            self.affiliated_institutions.append(inst)
```
sloria: If you can, make this method idempotent, i.e. multiple invocations with the same input should be AOK.

```python
node.add_primary_institution(user, inst)  # adds institution
node.add_primary_instutition(user, inst)  # does not add instution--it is already added
```

website/project/model.py
```diff
@@ -852,15 +855,43 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin):
...
+                    'node': self._primary_key,
+                    'institution': {
+                        'id': inst._id,
+                        'name': inst.name
+                    }
+                },
+                auth=Auth(user)
+            )
             return True
         return False
```
sloria: I'm not sure returning `False` is the best behavior if the user is not affiliated with the institution. After addressing https://github.com/CenterForOpenScience/osf.io/pull/5017/files#r52918842, there would be no way for the caller to distinguish between

1. Institution not added because it has already been added
2. Institution not added because user is not affiliated with the institution

It would be better if an exception were raised for case 2. Case 1 can probably just return `False`.

sloria: Reviewing this led me to investigate this `User#has_inst_auth`  (https://github.com/CenterForOpenScience/osf.io/blob/8fe462a4ab2b9bb6885e0326547566e20e66f827/framework/auth/core.py#L1359-L1362), for which I have a couple minor suggestions:

- Avoid unnecessary abbreviations and keep naming consistent. You have `remove_institution` and `has_inst_auth`.
- The logic can be simplified to `return inst in self.affiliated_institutions`
- Is `has_inst_auth` an accurate name? It looks like it's just checking affiliation. Maybe `is_affiliated_with_instutiton(inst)`?

sloria:
Need to pause review here to look at prod issues.

website/static/js/institutionSignIn.js
```diff
@@ -0,0 +1,47 @@
...
+    self.fetchInstitutions = function() {
+        var url = window.contextVars.apiV2Prefix + 'institutions/';
+        return $osf.ajaxJSON(
+            'GET',
+            url,
+            {
+                isCors: true
+            }
+        ).done(function (response) {
+            response.data.map(function(item){
```
sloria: `.map` returns an array, which you can use to populate `instNames`.

```js
self.instNames(response.data.map(...));
```

website/static/js/nodeControl.js
```diff
@@ -46,6 +46,27 @@ var ProjectViewModel = function(data) {
...
+
+    self.fetchInstitutions = function() {
+        if (data.node.institution) {
+            var url = window.contextVars.apiV2Prefix + 'nodes/' + self._id + '/institution/';
+            return $osf.ajaxJSON(
+                'GET',
+                url,
+                {isCors: true}
+            ).done(function (response) {
+                self.instLogoPath(response.data.attributes.logo_path);
```
sloria: Rather than sending an extra request just to get the logo path, could you add it to the `data` payload that is passed to the constructor. Yes, this will probably mean modifying v1 code, but I think it's justified in this case. By saving a request, we should reduce the "flashing" of the logo to some degree.

website/templates/public/login.mako
```diff
@@ -14,7 +14,39 @@
...
+                <select id="selectedInst" class="form-control" data-bind="value: selectedInst, options: instNames"></select>
+            </div>
+            <div class="form-group">
+                <div class="col-sm-offset-3 col-sm-9">
+                    <button data-bind="click: instLogin" class="btn btn-success pull-right">Sign in</button>
+                </div>
+            </div>
+            <div class="form-group" style="padding-top: 15px">
+                <div class="text-center m-t-lg">
+                    <p>To login normally click <a href="/login/">here</a>.</p>
```
sloria: Missing a comma here.

website/templates/public/login.mako
```diff
@@ -14,7 +14,39 @@
...
+    <p>
+      If your institution has partnered with the Open Science Framework, please
+        select its name below and sign in with your institutional credentials.
+    </p>
+</div>
+%endif
 <div class="row m-t-xl">
+    %if campaign == "institution" and enable_institutions:
+    <div class="col-sm-6 col-sm-offset-3 toggle-box toggle-box-active">
+        <h3 class="m-b-lg"> Login Through Institution</h3>
```
sloria: As per osf-style, use sentence case for headings.

website/templates/public/login.mako
```diff
@@ -186,6 +218,11 @@
 <%def name="javascript_bottom()">
     ${parent.javascript_bottom()}
     <script src=${"/static/public/js/login-page.js" | webpack_asset}></script>
+    <script type="text/javascript">
+        window.contextVars = $.extend(true, {}, window.contextVars, {
```
sloria: Won't this have to go before the script tag for login-page.js?

website/static/js/institutionProjectSettings.js
```diff
@@ -0,0 +1,109 @@
...
+                url: url,
+                status: status,
+                error: error
+            });
+        });
+    };
+    self.clearInst = function() {
+        var url = ctx.apiV2Prefix + 'nodes/' + ctx.node.id + '/relationships/institution/';
+        return $osf.ajaxJSON(
+            'PUT',
```
sloria: This looks incorrect; seems like the API should use `DELETE` to clear an institution.

website/static/js/pages/project-settings-page.js
```diff
@@ -54,8 +55,12 @@ if ($('#wgrid').length) {
         });
     });
 }
+
 $(document).ready(function() {
+    var self = this;
```
sloria: Remove unused line.

website/static/js/pages/project-settings-page.js
```diff
@@ -54,8 +55,12 @@ if ($('#wgrid').length) {
...
     // Apply KO bindings for Project Settings
+    new InstitutionProjectSettings('#institutionSettings');
```
sloria: What happens if `ENABLE_INSTITUTIONS = False`?

website/static/js/institutionProjectSettings.js
```diff
@@ -0,0 +1,109 @@
...
+            url,
+            {
+                'isCors': true,
+                'data': {
+                     'data': {'type': 'institutions', 'id': inst}
+                },
+                fields: {xhrFields: {withCredentials: true}}
+            }
+        ).done(function (response) {
+            window.location.reload();
```
sloria: Why is a reload necessary?

website/project/model.py
```diff
@@ -852,15 +855,43 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin):
...
 
-    def remove_primary_institution(self):
+    def remove_primary_institution(self, user):
+        inst = self.primary_institution
         self.primary_institution = None
+        if inst in self.affiliated_institutions:
+            self.affiliated_institutions.remove(inst)
+        self.add_log(
```
sloria: A log should only be added if the institution was actually removed.

If the institution isn't affiliated with the user, then an exception should probably be raised.

website/project/model.py
```diff
@@ -852,15 +855,43 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin):
...
+            params={
+                'node': self._primary_key,
+                'institution': {
+                    'id': inst._id,
+                    'name': inst.name
+                }
+            },
+            auth=Auth(user)
+        )
         self.save()
```
sloria: Again, avoid implicit saves.

website/templates/nav.mako
```diff
@@ -81,7 +81,10 @@
                   <div class="form-group"><input id="email" class="form-control" type="email" data-bind="value: username" name="username" placeholder="Email" aria-label="Username"></div>
                   <div class="form-group"><input name="password" id="password" class="form-control" type="password" placeholder="Password" data-bind="value: password" aria-label="Password"></div>
                   <div class="form-group"><button type="submit" id="btnLogin" class="btn btn-block btn-primary">Login</button></div>
-                 <div class="text-center m-b-sm"> <a href="/forgotpassword/">Forgot password?</a></div>
+                 %if enable_institutions:
+                  <div class="text-center m-b-sm"> <a href="/login/?campaign=institution">Login through your Institution  <i class="fa fa-arrow-right"></i></a></div>
```
sloria: Lowercase "institution".

sloria:
Non-contributors should not see this: 

osf___fresh_new_project

Like licenses, affiliation should only be shown if it is configured.

I'm not even sure we should show that UI for *contributors* if there is no affiliated institutions. @erinmayhood thoughts?

We should definitely not show that UI if the feature flag (`ENABLE_INSTITUTIONS`) is `False`.

sloria:
Please review your work to ensure that `ENABLE_INSTITUTIONS`  hides all UI associated with institutions.

sloria:
Pass finished.

website/static/js/institutionProjectSettings.js
hmoco: As per json api specs, 1-1 relationships are always updated with a PATCH request: http://jsonapi.org/format/#crud-updating-relationships

website/static/js/institutionProjectSettings.js
hmoco: some content is defined by mako ifs, which is expected to change once a request is succesful

website/templates/project/project.mako
```diff
@@ -120,6 +123,10 @@
                     </ol>
                 % endif
                 </div>
+                %if enable_institutions:
+                    <a class="link-dashed" href="${node['url']}settings/#configureInstitutionAnchor" id="institution">Affiliated Institution:</a>
```
sloria: As per osf-style, use sentence-case

website/templates/project/settings.mako
```diff
@@ -300,6 +304,55 @@
...
+                            </div>
+                            <!-- /ko -->
+                        % endif
+                        % if node['institution']:
+                            <div class="help-block">Your project is currently affiliated with: </div>
+                            <p data-bind="text: primaryInstitution"></p>
+                            <div class="help-block">
+                                Projects affiliated with institutions will show some institutional branding (such as logos), and if public, will be discoverable on OSF institutional landing pages.
+                            </div>
+                            <button data-bind="click: clearInst" class="btn btn-danger">Unaffiliate my project</button>
```
sloria: How about "Remove affiliation"?

website/templates/project/settings.mako
```diff
@@ -300,6 +304,55 @@
...
+                                <div data-bind="foreach: {data: availableInstitutions, as: 'item'}">
+                                    <div>
+                                    <label>
+                                        <input type="radio" data-bind="value: item.id, checked: $parent.selectedInstitution" name="primaryInst">
+                                        <p data-bind="text: item.attributes.name"></p>
+                                    </label>
+                                    </div>
+                                </div>
+                            </div>
+                            <button data-bind="click: submitInst" class="btn btn-success">Affiliate my project</button>
```
sloria: How about just "Affiliate"?

sloria:
This should not be clickable by non-contributors:

osf___institutions_testing

website/project/model.py
```diff
@@ -852,16 +856,48 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin):
...
+    affiliated_institutions = fields.ForeignField('institution', list=True)
+
     def add_primary_institution(self, user, inst):
-        if inst.auth(user):
-            self.primary_institution = inst
-            self.save()
-            return True
-        return False
+        if not inst.auth(user):
+            raise UserNotAffiliatedError('User is not affiliared with {}'.format(inst.name))
```
sloria: Typo: "affiliated"

website/project/model.py
```diff
@@ -852,16 +856,48 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin):
...
+        if inst == self.primary_institution:
+            return False
+        self.primary_institution = inst
+        if inst not in self.affiliated_institutions:
+            self.affiliated_institutions.append(inst)
+        self.add_log(
```
sloria: Shouldn't this log only be added if the user wasn't already affiliated with the institution?

website/project/model.py
```diff
@@ -852,16 +856,48 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin):
...
-    def remove_primary_institution(self):
+    def remove_primary_institution(self, user):
+        inst = self.primary_institution
+        if not inst:
+            return False
         self.primary_institution = None
-        self.save()
+        if inst in self.affiliated_institutions:
+            self.affiliated_institutions.remove(inst)
+        self.add_log(
```
sloria: Same here--this log should only be created if the institution is actually removed.

website/project/model.py
```diff
@@ -852,16 +856,48 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin):
...
-        return False
+        if not inst.auth(user):
+            raise UserNotAffiliatedError('User is not affiliared with {}'.format(inst.name))
+        if inst == self.primary_institution:
+            return False
+        self.primary_institution = inst
+        if inst not in self.affiliated_institutions:
+            self.affiliated_institutions.append(inst)
+        self.add_log(
+            action=NodeLog.INSTITUTION_ADDED,
```
sloria: Will we need to differentiate between "institution added" vs. "primary institution changed"?

website/project/model.py
sloria: Shouldn't `affiliated_institutions` also get copied?

website/static/js/institutionProjectSettings.js
```diff
@@ -0,0 +1,109 @@
...
+        }).fail(function (xhr, status, error) {
+            $osf.growl('Unable to add institution to this node!');
```
sloria: This should follow the text of similar messages on the OSF (no exclamation, show support emails, etc.)

website/project/model.py
hmoco: yep, for all these comments, this was meant for primary institutions only, since we are not currently actually supporting multiple institutions. Instead of changing whether or not add_log gets called, I will change the action name.

api/institutions/serializers.py
```diff
@@ -9,9 +9,10 @@ class InstitutionSerializer(JSONAPISerializer):
         'name'
     ])
 
-    name = ser.CharField(required=False)
-    id = ser.CharField(required=False, source='_id')
+    name = ser.CharField(required=True)
+    id = ser.CharField(required=True, source='_id')
     logo_path = ser.CharField()
+    auth_url = ser.CharField(required=True)
```
sloria: `required=True` doesn't make sense here. `auth_url` is a static value; users should not be able to modify this. You should pass `read_only=True` to the appropriate fields.

sloria:
Please address https://github.com/CenterForOpenScience/osf.io/pull/5017#discussion_r52920844

hmoco:
@sloria, I have: https://github.com/CenterForOpenScience/osf.io/pull/5017/files#diff-0edcd073bdfdcf5618bed0a5939a58d1R1359

api/institutions/serializers.py
```diff
@@ -9,9 +9,10 @@ class InstitutionSerializer(JSONAPISerializer):
         'name'
     ])
 
-    name = ser.CharField(required=False)
-    id = ser.CharField(required=False, source='_id')
+    name = ser.CharField(required=True)
```
sloria: `required=True  doesn't make sense for `name` and `id`; they are read-only.

api/nodes/serializers.py
```diff
@@ -492,7 +492,10 @@ def update(self, instance, validated_data):
                 raise exceptions.NotFound
             if not inst.auth(user):
                 raise exceptions.PermissionDenied
```
sloria: Now that `add_primary_institution` and `remove_primary_institution` check the user's affiliation, this check should be necessary. Instead, you should handle the possible UserNotAffiliated exception that may be raised.

sloria: Discussed in person: `PermissionsError` might not be appropriate; this is more of a validation rather than a permissions error, so the user should receive a 400 rather than a 403. You can use DRF's ValidationError for this.

api/nodes/serializers.py
```diff
@@ -492,7 +492,10 @@ def update(self, instance, validated_data):
...
-        node.primary_institution = inst
+            node.add_primary_institution(inst=inst, user=user)
```
sloria: As suggested in https://github.com/CenterForOpenScience/osf.io/pull/5017/files#r53060585, expected exceptions shuold be handled here, e.g. `UserNotAffiliated`.

tests/factories.py
```diff
@@ -535,6 +535,7 @@ def _create(cls, target_class, _id, name):
             _id=_id,
```
sloria: Is it necessary to override `_create` and `_build`? It'd be better if you could just define the default field names, like you normally would with factories:

```python

class InstitutionFactory(ModularOdmFactory):
    FACTORY_FOR = Institution
    _id = Sequence(lambda n: "S{}".format(n))
    name = Sequence(lambda n: "School{}".format(n))
    logo_name = 'logo.img'
    auth_url='http://thisIsUrl.biz'
```

website/project/model.py
sloria: Sounds good.

sloria:
In response to https://github.com/CenterForOpenScience/osf.io/pull/5017#issuecomment-184831275 : Sorry, didn't see that change. 

The reason I thought my comment was unaddressed is because I saw usages of `inst.auth` (e.g. https://github.com/CenterForOpenScience/osf.io/blob/f1d66f4685bcaea677bce54b743263b93cf38c73/api/nodes/serializers.py#L493). I had assumed `Institution#auth` would go away; it's not necessary to have a method on both Institution and User. I vote to keep `User#is_affiliated_with_institution` and remove `Institution#auth` because the name is better.

website/project/views/node.py
```diff
@@ -779,6 +779,8 @@ def _view_project(node, auth, primary=False):
                 'doi': node.get_identifier_value('doi'),
                 'ark': node.get_identifier_value('ark'),
             },
+            'institution': node.primary_institution.name if node.primary_institution else None,
```
sloria: Minor:

> Prefer namespaced representations to arbitrary prefixes in response data. 

https://cosdev.readthedocs.org/en/latest/osf/guidelines.html#responses

So better to have: 

```json
{
  "institution": {
     "name": ...,
     "logo": ... 
 }
}
```

website/templates/public/login.mako
```diff
@@ -14,7 +14,39 @@
...
+                <select id="selectedInst" class="form-control" data-bind="value: selectedInst, options: instNames"></select>
+            </div>
+            <div class="form-group">
+                <div class="col-sm-offset-3 col-sm-9">
+                    <button data-bind="click: instLogin" class="btn btn-success pull-right">Sign in</button>
+                </div>
+            </div>
+            <div class="form-group" style="padding-top: 15px">
+                <div class="text-center m-t-lg">
+                    <p>To login normally, click <a href="/login/">here</a>.</p>
```
sloria: Was this text decided in the spec/mockups? If not, I suggest a revision:

> For non-institutional login, click here.

sloria:
Do options in this dropdown get sorted?

osf___sign_in

They should be sorted alphabetically by name. In fact, the API endpoint should probably sort alphabetically by name by default.

sloria:
Pass complete.

hmoco:
@sloria have responded

sloria:
Please address this comment: https://github.com/CenterForOpenScience/osf.io/pull/5017/files#r53056625

hmoco:
Good call completely missed that one

api/nodes/views.py
```diff
@@ -1997,7 +1997,7 @@ def get_object(self):
 
 
 class NodeInstitutionRelationship(JSONAPIBaseView, generics.RetrieveUpdateAPIView, NodeMixin):
```
sloria: This view should be read only for registrations--is this true? I don't see a test verifying this.

api/users/permissions.py
```diff
@@ -13,3 +13,16 @@ def has_object_permission(self, request, view, obj):
...
+    """ Check to see if the request is coming from the currently logged in user,
+    and allow non-safe actions if so.
+    """
+    def has_object_permission(self, request, view, obj):
+        assert isinstance(obj, dict)
+        request_user = request.user
+        if request.method in permissions.SAFE_METHODS:
+            return True
+        else:
+            return obj['self'] == request_user
```
sloria: `==` can sometimes be unpredictable when comparing ODM objects. It's safer to compare PKs ( e.g. `request_user._id`)

website/project/model.py
```diff
@@ -852,16 +856,48 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin):
...
+        if inst == self.primary_institution:
+            return False
+        self.primary_institution = inst
+        if inst not in self.affiliated_institutions:
+            self.affiliated_institutions.append(inst)
+        self.add_log(
+            action=NodeLog.PRIMARY_INSTITUTION_ADDED,
+            params={
+                'node': self._primary_key,
+                'institution': {
```
sloria: Just in case we decide to display logs like "change primary institution from X to Y", it would be good to capture the previous institution value (or `None` if there was no previous institution. 

If this is a quick change, go ahead and make it. Otherwise, this can go in a follow-up PR.

website/static/js/institutionProjectSettings.js
```diff
@@ -0,0 +1,109 @@
+'use strict';
+var $ = require('jquery');
+var ko = require('knockout');
+var Raven = require('raven-js');
+
+var $osf = require('js/osfHelpers');
+var ctx = window.contextVars;
```
sloria: `contextVars` should not be used in re-usable modules because it makes them less..re-usable. Instead, pass any necessary variables to the constructor of the wrapper class and view-model.

sloria:
Pass finished. Once the above comments are addressed, this should be good to merge.

hmoco:
@sloria, have responded

PR #5018, opened 2016-02-10
[Ready to Review] Feature/remove home from dashboard

pattisdr:
# Purpose

Issue https://openscience.atlassian.net/browse/OSF-5605

The `/home/` route is unnecessary. It should be changed to `/`, such that
- Visiting `/` when logged in goes to the user 'home' page that contains the "Quick Search", "New and noteworthy", and other relevant widgets
- Visiting `/` goes to the landing page, as it does now

# Changes

Renamed index.mako to landing.mako.  Then created new index.mako.

On url '/', index.mako is rendered.  If user logged in, home.mako is included. If user logged out, landing.mako is included.


pattisdr:
Pretty sure these failing tests are not mine.

pattisdr:
Okay - issues seem to be the collections API still has remnants of checking for `is_folder` and using `FolderFactory`.

whit537:
Let's see here ...

api/collections/serializers.py
```diff
@@ -143,7 +143,7 @@ def make_instance_obj(self, obj):
         return {'data': [
             pointer for pointer in
             obj.nodes_pointer
-            if not pointer.node.is_deleted and not pointer.node.is_folder
+            if not pointer.node.is_deleted and not pointer.node.is_collection
```
whit537: Is the switch here from folder to collection (17829c5fb022bb67c79ea0d9318c706e18d84001, abad36ede5a98d1e10fd5373fd2d8a9417a1b0bd) strictly necessary for this PR? Or is this an extra clean-up while we're at it (cf. https://github.com/caneruguz/osf.io/pull/17#issuecomment-178075559)? Or ... ?

pattisdr: This is cleanup.  The dashboard branch that I am merging into has failing tests - that had nothing to do with my changes.  Collections are the new name for folders.

whit537: Okay, gotcha.

whit537:
I'm getting the expected pages when logged in, logging out, and landing.

screen shot 2016-02-12 at 9 53 40 am
screen shot 2016-02-12 at 9 54 02 am
screen shot 2016-02-12 at 9 54 11 am

website/templates/index.mako
```diff
@@ -1,324 +1,11 @@
...
-            </div>
-          </div>
-        </div>
-      </div>
-
-    </div>
-
-
+<%def name="logged_in(name)">
+    %if name != '':
```
whit537: Need to learn some Mako ... what `name` is this? The function-ish parameter just above, or the HTML-ish attribute value just above (the result of the function-ish call?)?

whit537: http://docs.makotemplates.org/en/latest/defs.html

whit537: Alright, so this evaluates to something like:

```python
def logged_in(name):
    if name != '':
        include('home.mako')
    else:
        include('landing.mako')
```

website/templates/index.mako
```diff
@@ -1,324 +1,11 @@
...
+        <%include file="home.mako"/>
+    %else:
+        <%include file="landing.mako"/>
+    %endif
 </%def>
 
-<%def name="footer()">
+${logged_in(user_name)}
```
whit537: And I infer that when the `user_name` is the empty string here, that's a sufficient indication that we have an anonymous user.

pattisdr: I am using this information: which is available in the template:
https://github.com/CenterForOpenScience/osf.io/blob/develop/website/routes.py#L47

For clarification, when you are asking these questions, are you expecting a response from me, or is this a record of what you're thinking when you review?

whit537: [Looks like it](https://github.com/CenterForOpenScience/osf.io/blob/cef7a1f89871bf624473b556ff70157b6701a1bf/website/routes.py#L47).

website/views.py
```diff
@@ -98,11 +97,6 @@ def _render_nodes(nodes, auth=None, show_path=False):
 
 @collect_auth
 def index(auth):
-    """Redirect to dashboard if user is logged in, else show homepage.
-
-    """
-    if auth.user:
-        return redirect(web_url_for('home'))
     return {}
```
whit537: Is there a way to do away with the `index` function entirely? Seems goofy to have such a stubby little function like this.

website/templates/index.mako
whit537: > For clarification, when you are asking these questions, are you expecting a response from me, or is this a record of what you're thinking when you review?

This is a record of what I'm thinking when I review. You can chime in if you want but you don't have to. If I'm not able to figure something out on my own I'll put it on a punchlist, and when I make a punchlist that's when I'm definitely expecting a response from you. However, this PR is small and tight enough that we may not need a punchlist. :-)

whit537:
#### Punchlist

 - [ ] remove the `/home/` [route](https://github.com/CenterForOpenScience/osf.io/blob/cef7a1f89871bf624473b556ff70157b6701a1bf/website/routes.py#L177) and [view](https://github.com/pattisdr/osf.io/blob/feature/remove_home_from_dashboard/website/views.py#L119-L121)

website/templates/index.mako
whit537: > when I make a punchlist that's when I'm definitely expecting a response from you.

Actually, even when I make a punchlist I'm likely to update it before I'm really expecting a response from you. You can respond at any time, but when I'm done with a review pass and I have the punchlist fully updated I'll tell you so explicitly—when I make a comment like [this](https://github.com/caneruguz/osf.io/pull/17#issuecomment-176877635), *then* I'm definitely expecting a response from you. If you respond before we get to that point that's fine, too. :)

website/views.py
whit537: Looks like routing `Rule`s are [designed](https://github.com/CenterForOpenScience/osf.io/blob/cef7a1f89871bf624473b556ff70157b6701a1bf/framework/routing/__init__.py#L69-L70) to take a `dict` directly for cases such as ours here.

whit537: Do we depend on the `collect_auth` decorator for anything besides the `auth` param to `index`, which is now unnecessary?

whit537:
@pattis An ack/grep of the codebase suggests that the home view may have been used to generate URLs in some places. Hopefully Travis catches the error if it's indeed present. Did you happen to check for that?

website/views.py
whit537: No, because all that `collect_auth` [does](https://github.com/CenterForOpenScience/osf.io/blob/cef7a1f89871bf624473b556ff70157b6701a1bf/framework/auth/decorators.py#L17-L24) is add the `auth` kwarg to the view function call.

Okay! Let's ditch the `index` view entirely.

whit537:
Alright @pattisdr, [punchlist](https://github.com/CenterForOpenScience/osf.io/pull/5018#issuecomment-183373104) updated and the :basketball: is in your court. :-)

website/views.py
pattisdr: We're using the `index` view in the base.mako template https://github.com/CenterForOpenScience/osf.io/blob/develop/website/templates/base.mako#L84

pattisdr: Also using index view in forgot_password.mako template
https://github.com/CenterForOpenScience/osf.io/blob/develop/website/templates/public/forgot_password.mako#L27

whit537: Fair enough.

whit537: How does `web_url_for` work?

pattisdr:
I can't find any instances of the home view generating URL's after this PR.  Prior to this PR, it was used in the index view, but now the index view only returns an empty dict.

website/views.py
whit537: [Good lord](https://github.com/mitsuhiko/flask/blob/0.10.1/flask/helpers.py#L255-L316).

pattisdr: haha :) I was thinking that myself.

whit537:
@pattisdr I was getting distracted by usages in `admin` and `api`, sorry.

:+1:

website/views.py
whit537: A rabbit hole for another day, perhaps. :-)

whit537: Or today ... :rabbit: 

Why does [`OsfWebRenderer__index`](https://github.com/CenterForOpenScience/osf.io/blob/cef7a1f89871bf624473b556ff70157b6701a1bf/website/util/__init__.py#L125) point to the index view function?

whit537: Flask's `url_for` [calls out](https://github.com/mitsuhiko/flask/blob/0.10.1/flask/helpers.py#L304-L305) to Werkzeug's [`build`](https://github.com/mitsuhiko/werkzeug/blob/0.10.4/werkzeug/routing.py#L1666-L1693).

whit537:
```
(Pdb) l
1592                if rv is not None:
1593                    return rv
1594    
1595            # default method did not match or a specific method is passed,
1596            # check all and go with first result.
1597 ->         for rule in self.map._rules_by_endpoint.get(endpoint, ()):
1598                if rule.suitable_for(values, method):
1599                    rv = rule.build(values, append_unknown)
1600                    if rv is not None:
1601                        return rv
1602    
(Pdb) 
1603        def build(self, endpoint, values=None, method=None, force_external=False,
1604                  append_unknown=True):
1605            """Building URLs works pretty much the other way round.  Instead of
1606            `match` you call `build` and pass it the endpoint and a dict of
1607            arguments for the placeholders.
1608    
1609            The `build` function also accepts an argument called `force_external`
1610            which, if you set it to `True` will force external URLs. Per default
1611            external URLs (include the server name) will only be used if the
1612            target URL is on a different subdomain.
1613    
(Pdb) values
{}
(Pdb) method
u'GET'
(Pdb) r
--Return--
> /Users/whit537/workbench/cos/osf.io/env/lib/python2.7/site-packages/werkzeug/routing.py(1601)_partial_build()->(u'', u'/')
-> return rv
(Pdb) rule
<Rule '/' (HEAD, OPTIONS, GET) -> OsfWebRenderer__index>
(Pdb)
```

whit537:
```
(Pdb) rule.__class__
<class 'werkzeug.routing.Rule'>
(Pdb)
```

whit537:
```
(Pdb) pp sorted(self.map._rules_by_endpoint.keys())
<snip>
 'OsfWebRenderer__profile_view',
 'OsfWebRenderer__profile_view_id',
 'OsfWebRenderer__project_new_node',
 'OsfWebRenderer__project_set_privacy',
 'OsfWebRenderer__project_statistics',
 'OsfWebRenderer__project_statistics_redirect',
 'OsfWebRenderer__project_wiki_compare',
 'OsfWebRenderer__project_wiki_edit',
 'OsfWebRenderer__project_wiki_edit_post',
 'OsfWebRenderer__project_wiki_home',
 'OsfWebRenderer__project_wiki_id_page',
 'OsfWebRenderer__project_wiki_view',
 'OsfWebRenderer__reproducibility',
 'OsfWebRenderer__resend_confirmation',
 'OsfWebRenderer__reset_password',
 'OsfWebRenderer__resolve_guid',
 'OsfWebRenderer__user_account',
 'OsfWebRenderer__user_account_password',
 'OsfWebRenderer__user_addons',
 'OsfWebRenderer__user_notifications',
 'OsfWebRenderer__user_profile',
 'OsfWebRenderer__view_project',
 'XMLRenderer__search_share_atom',
 'addon_static',
 'apiv1<path:_>',
 'explore',
 'faq',
 'getting-started',
 'messages__help',
 'news',
 'search',
 'share',
 'share_dashboard',
 'sharehelp',
 'shareregistration',
 'static']
```

whit537: Populated [here](https://github.com/mitsuhiko/werkzeug/blob/0.10.4/werkzeug/routing.py#L1130).

whit537: Yes! Here's where the `OsfWebRenderer__index` string is [computed](https://github.com/CenterForOpenScience/osf.io/blob/cef7a1f89871bf624473b556ff70157b6701a1bf/framework/routing/__init__.py#L159-L162)—we do it ourselves in `framework` (it's not a Flask or Werkzeug thing).

whit537: Okay! Rabbit stew, anyone? :-)

:ramen:

website/views.py
```diff
@@ -98,11 +97,6 @@ def _render_nodes(nodes, auth=None, show_path=False):
 
 @collect_auth
 def index(auth):
```
whit537: @pattisdr At the very least we don't need to `collect_auth` anymore on this view function.

whit537:
> :+1:

@pattisdr I take that back. [Punchlist](https://github.com/CenterForOpenScience/osf.io/pull/5018/files#issuecomment-183373104)! :-)

website/views.py
whit537: @pattisdr Looks like there is a provision for [naming](https://github.com/CenterForOpenScience/osf.io/blob/cef7a1f89871bf624473b556ff70157b6701a1bf/framework/routing/__init__.py#L165-L172) an endpoint that's not a view function.

whit537: So if you wanted to you could remove the `index` view function entirely, pass `{}` to `Rule`, and then update the places where we call `web_url_for('index')` to call `web_url_for('')` instead. But I'm not going to block this PR on that. :-)

pattisdr: I'll make the change.

whit537: !m pattisdr :dancer:

whit537: (Hopefully `web_url_for('')` doesn't already match something else. :)

pattisdr: I am getting BuildErrors with this change.

whit537: Okay, I'll let you decide how far to chase rabbit now. :)

whit537: I'd like to at least see us remove the extraneous `auth` kwarg. I'm fine keeping the stubby view func if it's not in fact trivial to remove it.

pattisdr:
@whit537, removed collect auth and auth kwarg.

whit537:
Thanks. All set here.

:shipit:

PR #5019, opened 2016-02-10
[Ready for Review ]Add template that differentiates between New Account makers and adding a new email to existing account

erinspace:
Addresses https://openscience.atlassian.net/browse/OSF-5697

# Purpose 
Previous cherry-picked commit https://github.com/CenterForOpenScience/osf.io/commit/fee6ee687cd577de8dd2c5cafc2e0b39a76fc5ea modified the CONFIRM_EMAIL template to assume that you're confirming an email added to an existing account. @caileyfitz found that this email is also sent to new registratnt to the OSF. This PR adds a new category for email confirmation to be sent, that checks if this user is not active and sends a specific email then.

# Changes
- New email template that is sent when a person first enters their email into the account signup form
- Calling that email template from ```mails.py```
- Checking if a user is active, sending the simple CONFIRM_EMAIL in that case, and sending them their first ever "you've made an account with this email" if they're not active.

# Side effects
If another case used this generic CONFIRM EMAIL template, but I could not find one.

### DIFF
--- a/website/templates/emails/initial_confirm.txt.mako
+++ b/website/templates/emails/initial_confirm.txt.mako
@@ -1,0 +1,10 @@
+Hello ${user.fullname},
+
+Thank you for registering for an account on the Open Science Framework.
+
+Please verify your email address by visiting this link:
+
+${confirmation_url}
+
+The OSF Team
+Center for Open Science
--- a/framework/auth/views.py
+++ b/framework/auth/views.py
@@ -245,8 +245,10 @@ def send_confirm_email(user, email):
         mail_template = mails.CONFIRM_MERGE
     elif campaign:
         mail_template = campaigns.email_template_for_campaign(campaign)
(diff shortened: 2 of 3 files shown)


mfraezz:
LGTM, sending this to RTM :star:

website/templates/emails/initial_confirm.txt.mako
```diff
@@ -0,0 +1,10 @@
+Hello ${user.fullname},
+
+Thank you for registering for an account on the Open Science Framework.
+
+Please verify your email address by visiting this link:
+
+${confirmation_url}
+
+The OSF Team
```
sloria: There will need to be blank line here in order for there to be a newline in the rendered email. Otherwise, the email will look like

```
The OSF Team Center for Open Science
```

I can do this when I merge.

PR #5020, opened 2016-02-11
[Feature] admin notes

Ghalko:
## Proposal
Admins can keep notes about interactions to facilitate something like case history.


Ghalko:
Deprecated by #5056

PR #5024, opened 2016-02-12
[WIP] Make api lists ESI-based

cwisecarver:
This PR makes list views in the API use ESI. 

Instead of drawing out the full object in JSON the object is replaced by a varnish esi:include tag. This allows cache invalidation to work correctly because every part of the API now depends on a url for each individual object. 

#### Sample Data for http://localhost:8000/v2/nodes/?format=jsonapi&page[size]=3
```json
{
    "data":[
    <esi:include src="http://localhost:8000/v2/nodes/kxfsn/?format=jsonapi"/>,
    <esi:include src="http://localhost:8000/v2/nodes/bnscw/?format=jsonapi"/>,
    <esi:include src="http://localhost:8000/v2/nodes/r7myv/?format=jsonapi"/>
],
    "links":{
       "first":null,
        "last":"http://localhost:8000/v2/nodes/?format=jsonapi&page=115&page%5Bsize%5D=3",
        "prev":null,
        "next":"http://localhost:8000/v2/nodes/?format=jsonapi&page=2&page%5Bsize%5D=3",
        "meta":{"total":345,"per_page":3}}
}
```

#### Special Note
- For list esi:include tags the query parameters on the initial url are compared to a blacklist and appended to the esi url

  - For embed ESI:include tags the query parameters are **not** included because recursion

### DIFF
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -666,6 +666,7 @@ class NodeFileHyperLinkField(RelationshipField):
 class JSONAPIListSerializer(ser.ListSerializer):
 
     def to_representation(self, data):
+        enable_esi = self.context.get('enable_esi', False)
         # Don't envelope when serializing collection
         errors = {}
         bulk_skip_uneditable = utils.is_truthy(self.context['request'].query_params.get('skip_uneditable', False))
@@ -674,9 +675,14 @@ class JSONAPIListSerializer(ser.ListSerializer):
             errors = data.get('errors', None)
             data = data.get('data', None)
 
-        ret = [
-            self.child.to_representation(item, envelope=None) for item in data
-        ]
+        if enable_esi:
+            ret = [
+                self.child.to_esi_representation(item) for item in data
+            ]
+        else:
+            ret = [
+                self.child.to_representation(item, envelope=None) for item in data
+            ]
 
         if errors and bulk_skip_uneditable:
             ret.append({'errors': errors})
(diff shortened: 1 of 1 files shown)


brianjgeiger:
Are there any tests that can be done for this? It looks good otherwise, insofar as.

cwisecarver:
@brianjgeiger The existing integration tests for varnish, that run on travis, will check nearly every combination of api url with esi enabled and esi disabled to make sure the responses match. They also check cache invalidation.

As far as unit tests, I didn't write any, technically, the generated string is invalid json. I could throw some regex against it to confirm what I think is happening is indeed happening but my opinion was that then I would have two problems. :wink: 

I spent the last 2 hours or so functionally testing it which did allow me to find several migrations that I hadn't run that were causing 500s.

brianjgeiger:
Works for me.

api/base/serializers.py
```diff
@@ -769,6 +775,25 @@ def invalid_embeds(self, fields, embeds):
...
+    def to_esi_representation(self, data):
+        href = None
+        query_params_blacklist = ['page[size]', 'format']
+        try:
+            href = data.get_absolute_url()
+        except:
+            representation = super(JSONAPISerializer, self).to_representation(data)
+            return representation
+        else:
+            if href and not href == '{}':
```
brianjgeiger: So, if get_absolute_url succeeds but sends back {} or None, there's nothing that gets returned. Is that the proper case? I think that would return a None into the list. Which might serialize fine, I'm not sure.

brianjgeiger:
Okay, I could only think of one potential issue. Pass complete. :penguin:

api/base/serializers.py
cwisecarver: @brianjgeiger You're right. I stole that from a previous implementation and was returning a SkipField in the else clause. I just added in a failsafe. I should also probably make that exception clause be tighter. I'll figure out what to watch for an change that too.

brianjgeiger: :+1:

sloria:
Merged into `feature/instutution-autodeploy` per @cwisecarver 's request

api/base/serializers.py
```diff
@@ -769,6 +775,27 @@ def invalid_embeds(self, fields, embeds):
...
+    def to_esi_representation(self, data):
+        href = None
+        query_params_blacklist = ['page[size]', 'format']
+        try:
+            href = data.get_absolute_url()
+        except AttributeError:
+            representation = super(JSONAPISerializer, self).to_representation(data)
+            return representation
+        else:
+            if href and not href == '{}':
```
sloria: Minor: `if href and href != '{}':` is a bit more readable.

cwisecarver:
@sloria I'm going to move this back to WIP. I found a bug in it. I'm working on a fix now.

sloria:
@cwisecarver Ah, saw your message after I had merged it locally. Go ahead and send the bugfix to develop when it's ready. I'll hold off on deploying this to staging.

PR #5025, opened 2016-02-12
[Feature][Update] Check-in/out logging [OSF-5056]

mfraezz:
Updates https://github.com/CenterForOpenScience/osf.io/pull/4830

Fixes conflict / circular import
Updated to display logs on `My Projects` too.

screen shot 2016-02-12 at 11 35 34

screen shot 2016-04-04 at 14 51 44

[#OSF-5056]

~~Original~~(Modified) Description:
Purpose
=======
Checking out a file was not logged in recent activity, even though it's a blocking activity. This feature adds that functionality.

Changes
=======
Several files were amended to add this feature. ~~Primarily, the ./website/files/models/osfstorage.py is where most of the action happened. Here I refactored the save method to log a checkin or checkout activity.~~(Model method, hit only in `update` in APIv2, added to set checked-in/out status and log event.) To update the view, I added template for the activity ~~along with signals and function parameters to look for the new activity~~(to both `log_templates.mako` and `logActionsList.json`). `api/logs/serializers.py` also received new serializer field. ~~No code in these sections was edited; rather, I just added the new parameters.~~


sloria:
Note to self: address issues from https://github.com/CenterForOpenScience/osf.io/pull/4830#issuecomment-183399465

website/files/models/osfstorage.py
```diff
@@ -82,9 +83,39 @@ def move_under(self, destination_parent, name=None):
...
+            check that via the if in the middle of the method and log the correct activity
+
+        :return:
+        """
         self.path = ''
         self.materialized_path = ''
-        return super(OsfStorageFileNode, self).save()
+        ret = super(OsfStorageFileNode, self).save()
+
+        if 'checkout' in ret and len(ret) == 1:
```
sloria: I'm not sure about this approach. Typically, we would add a model method that would make field changes and do logging.

```python
file_node.check_out(...)  # sets 'checkout' and adds node log
```

I think this would be cleaner than adding all this logic to `save()`.

sloria:
Pass finished.

website/files/models/osfstorage.py
mfraezz: Slightly more complicated than that -- the `checkout` field is being updated in APIv2, this just detects it and logs accordingly. That being said, it could be done there by 1) detecting that a checkout is occurring and 2) calling a new model method that adds the log. Something like:
```python
#api/files/serializer.py:FileSerializer
    assert isinstance(instance, FileNode), 'Instance must be a FileNode'
+   user = validated_data.get('checkout', None) or instance.checkout
        setattr(instance, attr, value)
    instance.save()
+   if user and len(validated_data) == 1:
+       instance.log_check_out(user)
    return instance

#website/models/files/osfstorage.py:OsfStorageFileNode
def log_check_out(self, user):
    # logic moved from .save
    self.node.add_log(
        action=['checked_in', 'checked_out'][self.is_checked_out],
        params={
            'project': self.node.parent_id,
            'node': self.node._id,
            'urls': {
                'download': "/project/" + self.node._id + "/files/osfstorage/" + self._id + "/?action=download",
                'view': "/project/" + self.node._id + "/files/osfstorage/" + self._id},
            'path': self.materialized_path
        },
        auth=Auth(user),
    )
```

brianjgeiger: Talked with @mfraezz about this. This came about because there were dreams last summer of having file metadata be generically modifiable by the API without doing model changes, thus allowing a new breed of apps in the OSF that may or may not overlap with things like JamDB based apps. The only thing we currently have that takes advantage of this is the file checkout functionality, and we may have decided to abandon that path. This issue makes that decision important.

There are essentially two major potential paths:

1. Make a generic "metadata changed" log event and have message like "Bob change the metadata of Fish Guy.stl from checkout: True to checkout: False." and keep the concept of API modifiable metadata without model changes;
2. Not have file metadata be so generically changeable and make a model method for each type of metadata that gets changed. Then make a specific log message for checking files in and out.

brianjgeiger:
Since you've made a model method, that should be tested independently of the API. Ensure that the logs generated are appropriate in more depth, including having the right user if the admin does a checkin, and generally verifying that checkin and checkout all work correctly.

brianjgeiger:
Also, I am given to understand that logging is changing drastically in 0.66.0, so it would probably be a good idea to target https://github.com/CenterForOpenScience/osf.io/tree/release/0.66.0-develop rather than just develop. Make sure that the new log template format is included (Dawn can point you in the right place if you don't know where that is).

api/files/serializers.py
```diff
@@ -188,8 +188,12 @@ def user_id(self, obj):
 
     def update(self, instance, validated_data):
         assert isinstance(instance, FileNode), 'Instance must be a FileNode'
-        for attr, value in validated_data.items():
-            setattr(instance, attr, value)
+        if len(validated_data) == 1 and 'checkout' in validated_data:
```
brianjgeiger: What happens if you modify two things at once? I don't know if that's currently possible, but if we added another metadata option, my guess is this would break. I think moving the conditional into the for loop is the better way to go.

mfraezz: 1) It's not currently possible, and could only break if on a manual query to the API that tries to update multiple metadata fields, not if the request was made through [the UI (via Fangorn)](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/fangorn.js#L1037-L1052)
2) `len` validation was originally included because `'checkout'` is returned on first `.save()` call, so that would currently work, it could break if/when we migrate more functionality to APIv2

brianjgeiger: Hmmm. I'm having trouble seeing what you're saying in 2. Maybe whiteboarding would help?

And I agree that it's not a problem right now, but it's not an unlikely problem.

brianjgeiger:
Sounds good. I'll mark this as "needs further development," and when 0.66.0 is merged into develop, do those updates. Then we'll re-review.

mfraezz:
@brianjgeiger log template format updated.

brianjgeiger:
Looks good from my perspective.

website/files/models/osfstorage.py
```diff
@@ -102,6 +103,35 @@ def move_under(self, destination_parent, name=None):
             raise exceptions.FileNodeCheckedOutError()
         return super(OsfStorageFileNode, self).move_under(destination_parent, name)
 
+    def check_in_or_out(self, user, checkout, save=False):
+        """
+        Updates self.checkout with the requesting user or None if user has permission
```
brianjgeiger: Except for one thing: update this docstring with the parameter usage, since it's a little odd. https://cosdev.readthedocs.org/en/latest/osf/guidelines.html?highlight=docstring#docstrings

brianjgeiger:
RTM when checks pass, which I expect they will.

website/files/models/osfstorage.py
```diff
@@ -102,6 +103,41 @@ def move_under(self, destination_parent, name=None):
...
+        self.checkout = checkout
+
+        self.node.add_log(
+            action=action,
+            params={
+                'kind': self.kind,
+                'project': self.node.parent_id,
+                'node': self.node._id,
+                'urls': {
+                    'download': "/project/{}/files/{}/{}/?action=download".format(self.node._id, self.provider, self._id),
```
chrisseto: `web_url_for`?

website/files/models/osfstorage.py
```diff
@@ -102,6 +103,41 @@ def move_under(self, destination_parent, name=None):
...
+        iff user has permission to check out file or folder.
+        Adds log to self.node.
+
+
+        :param user:        User making the request
+        :param checkout:    Either the same user or None, depending on in/out-checking
+        :param save:        Whether or not to save the user
+        """
+        if (self.is_checked_out and self.checkout != user and 'admin' not in self.node.permissions.get(user._id, []))\
+           or user._id not in self.node.permissions.keys() or 'write' not in self.node.permissions.get(user._id, []):
```
chrisseto: Use `self.node.get_permissions(user)`

`user._id not in self.node.permissions.keys()` is unnecessary

website/files/models/osfstorage.py
```diff
@@ -102,6 +103,41 @@ def move_under(self, destination_parent, name=None):
...
+            raise exceptions.FileNodeCheckedOutError()
+
+        action = 'checked_out' if checkout else 'checked_in'
```
chrisseto: Constants?

website/files/models/osfstorage.py
chrisseto: Constants?

website/static/js/logActionsList.json
```diff
@@ -37,6 +37,8 @@
     "file_updated":     "${user} updated file in ${node}",
     "file_removed":     "${user} removed file ${path} from ${node}",
     "file_restored":    "${user} restored file ${path} from ${node}",
+    "checked_out":  "${user} checked out ${kind} ${path} from ${node}",
+    "checked_in":   "${user} checked in ${kind} ${path} to ${node}",
```
chrisseto: Just for consistency, could you match the indentation here?

website/files/models/osfstorage.py
mfraezz: Raises `RuntimeError: Attempted to generate a URL without the application context being pushed. This has to be executed when application context is available.`

website/static/js/logActionsList.json
mfraezz: Based on the rest of the file, "one tab out" is the most consistent, but yes.

website/files/models/osfstorage.py
mfraezz: Documented reason for not doing so.

PR #5028, opened 2016-02-12
Feature/allow client to request specific related counts

pattisdr:
# Purpose

Issue: https://openscience.atlassian.net/browse/OSF-5725

 `?related_counts` parameter should be modified to allow the client to request specific relationship accounts, passed as a comma-separated list.

To retain backwards-compat, i.e. passing a boolean should also still work.

# Changes

Modified relationship serializer.

The following now works: 
 - `?related_counts=children`
 - `?related_counts=children,comments`

The following still works:
 - `?related_counts=true`

### DIFF
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -343,6 +343,26 @@ class RelationshipField(ser.HyperlinkedIdentityField):
             )
         )
 
+    def process_related_counts_parameters(self, params):
+        """
+        Processes related_counts parameter.
+
+        Can either be a True/False value for fetching counts on all fields, or a comma-separated list for specifying
+        individual fields.  Ensures specified fields are allowed.
+        """
+        if utils.is_truthy(params) or utils.is_falsy(params):
+            return params
+
+        field_counts_requested = [val for val in params.split(',')]
+        for count_field in field_counts_requested:
+            if count_field not in [field for field in self.parent.fields if getattr(self.parent.fields[field], 'json_api_link', False)]:
+                raise InvalidQueryStringError(
+                    detail="Acceptable values for the related_counts query param are 'true' or 'false' or relationship fields; got '{0}'".format(params),
+                    parameter='related_counts'
+                )
+        return field_counts_requested
+
+
     def get_meta_information(self, meta_data, value):
         """
         For retrieving meta values, otherwise returns {}
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -357,7 +357,7 @@ class RelationshipField(ser.HyperlinkedIdentityField):
         for count_field in field_counts_requested:
             if count_field not in [field for field in self.parent.fields if getattr(self.parent.fields[field], 'json_api_link', False)]:
                 raise InvalidQueryStringError(
-                    detail="Acceptable values for the related_counts query param are 'true' or 'false' or relationship fields; got '{0}'".format(params),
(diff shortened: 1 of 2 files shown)


api/base/serializers.py
```diff
@@ -343,6 +343,25 @@ def resolve(self, resource):
...
+        individual fields.  Ensures field for which we are requesting counts is a relationship field.
+        """
+        if utils.is_truthy(params) or utils.is_falsy(params):
+            return params
+
+        field_counts_requested = [val for val in params.split(',')]
+        for count_field in field_counts_requested:
+            if count_field not in [field for field in self.parent.fields if getattr(self.parent.fields[field], 'json_api_link', False)]:
+                raise InvalidQueryStringError(
+                    detail="Acceptable values for the related_counts query param are 'true', 'false', or any of the relationship fields; got '{0}'".format(params),
```
sloria: :+1: Good error message.

api/base/serializers.py
```diff
@@ -343,6 +343,25 @@ def resolve(self, resource):
...
+
+        Can either be a True/False value for fetching counts on all fields, or a comma-separated list for specifying
+        individual fields.  Ensures field for which we are requesting counts is a relationship field.
+        """
+        if utils.is_truthy(params) or utils.is_falsy(params):
+            return params
+
+        field_counts_requested = [val for val in params.split(',')]
+        for count_field in field_counts_requested:
+            if count_field not in [field for field in self.parent.fields if getattr(self.parent.fields[field], 'json_api_link', False)]:
```
sloria: You can move the list-comp outside of the loop so that it isn't computed more than once. Also, you can make it a set comprehension in order to get constant-time membership checks.

sloria: You might want to use `self.root` rather than `self.parent` to ensure that we get the serializer fields (`parent` could be another field).

sloria:
Minor changes above made during merge.

caneruguz:
@pattisdr @sloria Does this work for linked_nodes in collections? 
```
/v2/collections/?related_counts="linked_nodes"
```
It doesn't work for me locally so I wanted to check if there is a spelling issue?

pattisdr:
No quotes I think?

On Mon, Feb 15, 2016 at 10:52 AM, Caner Uguz <notifications@github.com>
wrote:

> @pattisdr <https://github.com/pattisdr> @sloria
> <https://github.com/sloria> Does this work for linked_nodes in
> collections?
>
> /v2/collections/?related_counts="linked_nodes"
>
> It doesn't work for me locally so I wanted to check if there is a spelling
> issue?
>
> —
> <https://github.com/CenterForOpenScience/osf.io/pull/5028#issuecomment-184267428>
> .
>



-- 
Dawn Pattison
Developer
Center for Open Science
dawn@cos.io

pattisdr:
Yep - http://localhost:8000/v2/collections/?related_counts=linked_nodes. No quotes.

Any relationship field on a serializer should work.

caneruguz:
No quotes is not working for me locally. I'm getting this:
screen shot 2016-02-15 at 1 30 14 pm

pattisdr:
The `self.parent` change to `self.root` before the merge seems to be causing @caneruguz's issue above. @sloria - is it okay to change back to `self.parent`?

sloria:
Made that change. Thanks for looking into it, Dawn. @caneruguz `develop` has the fix.

PR #5030, opened 2016-02-12
Make Comment.target & Comment.root_target Guid objects

samanehsan:
Purpose
-----------
Precursor to handling comments on addon files. 

Changes
------------
Make the comment target and root target guid objects (instead of Node, Comment or StoredFileNode objects) so that when files are updated, the guid referent can get updated with the new stored file node. That way, the comments will not be affected because they target the Guid object.


mfraezz:
Code looks good, but I found a bug: nested comments, whether created before or after migrating, no longer show up. 

After that's resolved, one more quick pass and I'll put this in RTM :camel:

mfraezz:
Please restart the `"osf"` build job. Looks to be a nondeterministic elastic failure.

Otherwise, changes look good. I'll pass this off to RTM whenever the build succeeds. :art:

PR #5035, opened 2016-02-16
Feautre/node archiving children display

erinspace:
https://openscience.atlassian.net/browse/OSF-5591

# Purpose
When registering a project, and a component, if the parent project gets stuck in archiving, the children are presented as being OK and no-longer archiving.

While this is the desired behavior on the backend (the child has a different archive job and so should in fact succeed on its own) it should not be presented this way as that might be confusing to a user who is checking the status of their registration. If the top level is still archiving and so appears "locked," the child registrations should appear to be in the same state.

# Changes
- Add ```archive_tree_in_progress``` to the ``` _get_summary``` method that will display if a node's root is currently still undergoing archiving.
- Use this ```archive_tree_in_progress``` to determine if the archiving badge displayed in the mako template to display instead of its own ```archiving``` property (which is properly reported as finished)
- Add an ```else``` instead of a second if statement in ```render_node.mako```, adjustment found while investigating this issue.

# Side effects
- A new property was added to the ``` _get_summary``` method and no existing code other than the template itself was modified, so things should still work the same
- Other mako templates can now use this provided ```archive_tree_in_progress``` information if they'd like to check the archiving status of the root node.

### DIFF
--- a/website/project/views/node.py
+++ b/website/project/views/node.py
@@ -926,8 +926,7 @@ def _get_summary(node, auth, rescale_ratio, primary=True, link_id=None, show_pat
         'is_pending_retraction': node.is_pending_retraction,
         'embargo_end_date': node.embargo_end_date.strftime("%A, %b. %d, %Y") if node.embargo_end_date else False,
         'is_pending_embargo': node.is_pending_embargo,
-        'archiving': node.archiving,
-        'archive_tree_in_progress': node.root.archiving,
+        'archiving': node.archiving or node.root.archiving,
     }
 
     if node.can_view(auth):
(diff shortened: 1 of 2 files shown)


website/project/views/node.py
```diff
@@ -927,6 +927,7 @@ def _get_summary(node, auth, rescale_ratio, primary=True, link_id=None, show_pat
         'embargo_end_date': node.embargo_end_date.strftime("%A, %b. %d, %Y") if node.embargo_end_date else False,
         'is_pending_embargo': node.is_pending_embargo,
         'archiving': node.archiving,
+        'archive_tree_in_progress': node.root.archiving,
```
mfraezz: Is it possible for the `node.root` archiving job to finish before the `node` archiving job, causing it to be displayed like it's not archiving on that node when, in fact, it is?

Regardless, `summary['archiving']` doesn't appear to be used anywhere other than the affected lines below, so it might be better to set `'archiving': node.archiving or node.root.archiving,` and not even need to alter the template.

sloria: +1 to this suggestion. This has the additional benefit that `node.root.archiving` will not need to be evaluated if `node.archiving` returns `True`.

PR #5036, opened 2016-02-16
reverse sorting fix

KAsante95:
# Purpose

Enable reverse sorting based on user node attributes such as title and date_created

Issue:
https://openscience.atlassian.net/browse/OSF-5237

# Code

Removed fields.pop(0) and simply selected the first index via fields[0].
I then included a reverse sort for the case in which the first character in the field was a '-', thus indicating a reverse sort. To implement the reverse sort, I simply switched the places of the 1 and -1 from the original sort.


sloria:
Please add a regression test that would fail without your fix.

sloria:
Also, can you please link to this issue's corresponding JIRA ticket in the PR description?

api_tests/base/test_filters.py
```diff
@@ -222,3 +226,57 @@ def test_convert_value_float(self):
         field = FakeSerializer._declared_fields['float_field']
         value = self.view.convert_value(value, field)
         assert_equal(value, 42.0)
+
+def sort_multiple(fields):
```
mfraezz: You should be able to just `import` this `from api.base.filters`

api/base/filters.py
```diff
@@ -25,16 +25,26 @@ def sort_multiple(fields):
...
         while fields:
-            field = fields.pop(0)
-            a_field = getattr(a, field)
-            b_field = getattr(b, field)
-            if a_field > b_field:
-                return 1
-            elif a_field < b_field:
-                return -1
+            field = fields[0]
+            if field[0] == '-':
```
mfraezz: You could de-duplicate this part of the function with something like:
```python
field = fields[0]
rev_sort = -1

if field[0] == '-':
    field = field[1:]
    rev_sort = 1

# getattr assignments
# if   > case:
    return rev_sort
# elif < case:
    return -rev_sort

File: api/base/filters.py [hunk 3]
```diff
@@ -25,16 +25,26 @@ def sort_multiple(fields):
...
     def sort_fn(a, b):
         while fields:
-            field = fields.pop(0)
-            a_field = getattr(a, field)
-            b_field = getattr(b, field)
-            if a_field > b_field:
-                return 1
-            elif a_field < b_field:
-                return -1
+            field = fields[0]
```
mfraezz: This will cause an infinite loop if two `objects` with the same `field` are encountered. Please either change it back to `.pop`, or use a `for` loop (e.g. `for field in fields:`). Also, please add a test case that sorts two objects with equal fields.

mfraezz:
Pass complete. Possible infinite loop found. :cloud:

After fixing that, please also add a test to make sure an infinite loop won't happen. There's also some excess code that can be de-duped, but that's not quite as important.

sloria:
@KAsante95 There's a lot of noise in the commit history. Did you pull from master by chance? 

Can you please send a new PR to develop, containing only your changes?

whit537:
This can be closed in favor of https://github.com/CenterForOpenScience/osf.io/pull/5258.

PR #5038, opened 2016-02-16
Feature: Fix comment edit/delete permissions

samanehsan:
Purpose:
------------
Fixes https://openscience.atlassian.net/browse/OSF-5681

If a project's comment level changes from public to private, non-contributors can still edit and delete any comments that they made while the comment level was public.

Changes:
------------
- The `can_edit` field on the CommentSerializer now checks if the user can comment on the project (not just if they are the author).
- The `CommentDetailPermissions.has_object_permission` also checks if the user can comment on the project.


mfraezz:
Pass complete, no bugs found :star2: 

Putting this in RTM when the build passes.

PR #5039, opened 2016-02-17
Move institutions to its own module, add __repr__ method

hmoco:


website/institutions/model.py
```diff
@@ -0,0 +1,58 @@
...
+    @property
+    def pk(self):
+        return self._id
+
+    @property
+    def absolute_url(self):
+        return urlparse.urljoin(settings.DOMAIN, self.url)
+
+    @property
+    def url(self):
```
sloria: This seems incorrect; `Institutions` aren't GUID objects, so they shouldn't have a GUID short URL. Is this actually used anywhere? Can it be removed?

website/institutions/model.py
```diff
@@ -0,0 +1,58 @@
...
+        return '/{}/'.format(self._id)
+
+    @property
+    def deep_url(self):
+        return '/institution/{}/'.format(self._id)
```
sloria: I don't see this anywhere in the routes; is this used?

website/institutions/model.py
```diff
@@ -0,0 +1,58 @@
...
+    def logo_path(self):
+        return '/static/img/institutions/{}/'.format(self.logo_name)
+
+    def get_api_url(self):
+        return self.absolute_api_v2_url
+
+    def get_absolute_url(self):
+        return self.absolute_url
+
+    def view(self):
```
sloria: Can this be removed for now?

website/institutions/model.py
```diff
@@ -0,0 +1,58 @@
...
+        return absolute_reverse('institutions:institution-detail', kwargs={'institution_id': self._id})
+
+    @property
+    def logo_path(self):
+        return '/static/img/institutions/{}/'.format(self.logo_name)
+
+    def get_api_url(self):
+        return self.absolute_api_v2_url
+
+    def get_absolute_url(self):
```
sloria: Remove any unused methods. I'm assuming this is incorrect because the `url` property looks incorrect (see https://github.com/CenterForOpenScience/osf.io/pull/5039/files#r53199195). Can you verify?

sloria:
The build failure is due to a jshint error on develop that has since been fixed. Updating with develop should fix the build.

sloria:
Pass finished.

api/institutions/serializers.py
```diff
@@ -32,10 +32,10 @@ class InstitutionSerializer(JSONAPISerializer):
     )
 
     def get_api_url(self, obj):
-        return obj.get_api_url()
+        return obj.absolute_api_v2_url
 
     def get_absolute_url(self, obj):
```
sloria: Remove the `html` link and this method.

sloria:
Just one last change on this one, and it should be good to merge.

hmoco:
@sloria once tests pass,back to you

PR #5041, opened 2016-02-17
[Feature] Admin can remove user's 2 factor

Ghalko:
## Proposal

This allows the admin to see if the user requesting two factor authentication removal has it, and if they do, remove it. ADM-5

## Repercussions

"Power is of two kinds. One is obtained by the fear of punishment and the other by acts of love. Power based on love is a thousand times more effective and permanent then the one derived from fear of punishment." Mahatma Ghandi


mfraezz:
Two factor can be disabled via user settings page, I'm uncertain of the purpose for this PR. Is it for when  users lock themselves out by misplacing mobile devices?

admin/users/urls.py
```diff
@@ -0,0 +1,12 @@
+from django.conf.urls import url
+
+from . import views
+
+urlpatterns = [
+    url(r'^$', views.UserFormView.as_view(),
+        name='search'),
+    url(r'^id-(?P<guid>[a-z0-9]+)/$', views.UserView.as_view(),
+        name='user'),
+    url(r'^id-(?P<guid>[a-z0-9]+)/rem-2-factor/$', views.remove_2_factor,
```
mfraezz: Consider url of the form `r'^id-(?P<guid>[a-z0-9]+)/two_factor/disable/$'`, or simply `r'^id-(?P<guid>[a-z0-9]+)/two_factor/` that accepts both `POST` and `DELETE` requests. [ADM ticket](https://openscience.atlassian.net/browse/ADM-5) states:

> [After admin removes] User record now has an ADD 2-Factor button (ie: this action can be reversed) (removing two factor deletes the addon, but the user can add it back when they log in.)

This seems to me that if (and only if) it's been disabled through the admin interface, it should be able to be re-enabled? Testing this revealed otherwise.

admin/users/views.py
```diff
@@ -0,0 +1,31 @@
+from django.shortcuts import redirect
+
+from website.project.model import User
+
+from admin.base.views import GuidFormView, GuidView
+from admin.users.templatetags.user_extras import reverse_user
+from .serializers import serialize_user
+
+
+def remove_2_factor(request, guid):
```
mfraezz: See comment https://github.com/CenterForOpenScience/osf.io/pull/5041/files#r54485386 : this method may need refactoring.

admin_tests/users/test_views.py
```diff
@@ -0,0 +1,38 @@
...
+from nose.tools import *  # flake8: noqa
+
+from tests.base import AdminTestCase
+from tests.factories import UserFactory
+from admin_tests.utilities import setup_view
+
+from admin.users.views import UserView
+
+
+class TestUserView(AdminTestCase):
```
mfraezz: Please add a two-factor related test to this class

admin/users/serializers.py
```diff
@@ -0,0 +1,29 @@
+"""
+Serialize user
+"""
+# from admin.nodes.serializers import serialize_simple_user
+
+
+def serialize_user(user):
+    two_factor = user.get_addon('twofactor')
+    if two_factor is None or two_factor is False:
```
mfraezz: Is it `None` if TF has never been configured and `False` if it previously was? If not, are both of these cases necessary?

Alternatively, APIv2 has [helper methods](https://github.com/CenterForOpenScience/osf.io/blob/develop/api/base/utils.py#L17-L33) to determine [truth/fals]yness, it might be good to implement or reuse something like that for ADM.

mfraezz:
The ticket for this PR has a bulleted list of requirements, duplicating that here as a task list:

- [X]     User accounts include a button to remove 2-factor authentication
- [X]     OSF staff member receives a warning after selecting: "You are about to remove 2-factor authentication from [user name's] account. Do you want to continue?" Yes No
- [X]     Clicking the Remove 2-Factor button removes 2-factor for this user
- [ ]     OSF staff member receives verification message: "You have successfully removed 2-factor authentication from [user name's] account. (the refreshed display will show this.) 
* (**CR Note:** Unsure if this is necessary, view updates and indicates that 2F has been removed)
- [ ]     User record now has an ADD 2-Factor button (ie: this action can be reversed) (removing two factor deletes the addon, but the user can add it back when they log in.)
* (**CR Note:** This feature appears to have been abandoned.)
- [ ]     The removal of 2-factor is logged (date, time, and which staff member did it) 
* (**CR Note:** Is this included in the `Log Admin Actions` PR, or will it be added in a follow-up to both?)
- [ ]     OSF user receives an email telling them about the setting change. (manual)  
* (**CR Note:** No email appears to be sent out, and the `Email` button beside user emails is disabled.)

mfraezz:
Pass complete @Ghalko, there are some discrepancies between ticket description and functionality, please address :basketball:

admin/users/urls.py
Ghalko: Reversing this action would require a weird refactor of the OSF, essentially storing the two-factor authentication somewhere, and then bringing it back, basically on a whim. Although possibly preferable in the future, this does not help our more immediate concern of actually being able to disable two factor without @saradbowman having to bug @sloria about deleting this addon.

mfraezz: It's already being stored. The `TwoFactorUserSettings` model inherits (eventually) from `StoredObject`; the only refactor required would be changing how deletion / the clearing of settings is handled.

Ghalko:
To respond (further) to the not checked off items.
- The change is indicated
- Refer to reversing comment above.
- Logging actions will be subsequent tickets to #5001 
- Sending a mail will be in subsequent tickets to #5056 Basically the desk.com integration will allow the admin to send an Email through desk, keeping the desk ticket up-to-date. Implementing a whole new Email function in support of the user might very well confuse them if an Email is sent from admin (without the admin knowing) and then receiving one from desk as well.

In short:
We have plans for better tools and really the ticket is outlining an en total approach rather than a 'make Sara's and Steve's lives easier' approach.

Ghalko:
Tickets created for unmet, potentially difficult, updates to the actions. ADM-5, ADM-79, ADM-80, ADM-81

admin/users/urls.py
Ghalko: So what if the user re-does it themselves?

mfraezz: Currently, it undeletes their old model instance, and then repopulates the cleared fields when they re-enter the 2F verification code. Simplest solution would be to not clear on `delete`, only disable the addon, and then clear when a user re-authorizes but not when an admin does. See [this code](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/addons/twofactor/models.py#L51-L61). 

I don't know of any security threats this approach could pose, but I'm also not as intimately familiar with the two factor addon as I am with most others. May want to run this by Steve.

admin_tests/users/test_views.py
Ghalko: Tests added.

admin/users/views.py
```diff
@@ -0,0 +1,32 @@
...
+    if request.method == 'POST' or request.method == 'DELETE':
```
mfraezz: This conditional isn't necessary unless functionality from ADM-79 is implemented.

admin/users/serializers.py
mfraezz: Please address

admin/users/views.py
mfraezz: Also, this method currently only receives `GET` requests, so this won't work.

admin/templates/users/user.html
```diff
@@ -0,0 +1,146 @@
...
+                                    </div>
+                                    <div class="modal-body">
+                                        Removing this will let the user log in
+                                        with only one form of authentication.
+                                        The only way to reverse this action is
+                                        for the user to re-enable the addon
+                                        after logging in.
+                                    </div>
+                                    <div class="modal-footer">
+                                        <a href="{% url 'users:remove2factor' user.id %}"
```
mfraezz: So far as I can tell, Django doesn't let you specify a request method via the templating language, so the best way to implement that may be via a [redirect](http://django.readthedocs.org/en/1.9.x/topics/http/urls.html#examples) that does some processing on it before altering the method and sending it along to `remove2factor`. That's for ADM-79, though

mfraezz:
Tests look good, but will also need to be changed because it's a `GET` request sent to `remove_2_factor`. 

:ox:  Back to you @Ghalko

Ghalko:
Please merge #5013 before this can proceed.

admin/users/serializers.py
Ghalko: Done

admin/users/views.py
Ghalko: Removed conditional and associated tests.

admin/users/serializers.py
```diff
@@ -5,11 +5,18 @@
...
+        two_factor = False
+    else:
+        two_factor = True
     return {
         'name': user.fullname,
+        'id': user._id,
         'nodes': map(serialize_simple_node, user.contributor_to),
         'emails': user.emails,
         'last_login': user.date_last_login,
+        'two_factor': two_factor,
```
mfraezz: Lines 8-12 could be obviated by changing this line to
```python
'two_factor': bool(user.get_addon('twofactor')),

Comment (sloria) [inline admin/users/serializers.py]: Alternatively: `user.has_addon('twofactor')`

File: admin/users/views.py [hunk 8]
```diff
@@ -1,17 +1,35 @@
...
+from admin.base.views import GuidFormView, GuidView
 from admin.users.templatetags.user_extras import reverse_user
 from .serializers import serialize_user
 
 
+def remove_2_factor(request, guid):
+    user = User.load(guid)
+    try:
+        user.delete_addon('twofactor')
+    except AttributeError:
```
sloria: This is rather indirect. Rather than handling an `AttributeError` here, you can check if `user` is `None`.

```python
if user:
   page_not_found(request)
```

Since this a very common operation, you might want to write a helper along the lines of [get_object_or_error](https://github.com/CenterForOpenScience/osf.io/blob/219c0fce2123489ae21abbac4a7f0d2b4b583985/api/base/utils.py#L56), which allows you to do

```python
user = get_object_or_error(User, guid)
user.delete_addon('twofactor')
```

sloria: This can go in a follow-up PR though.

admin/users/serializers.py
sloria: I'll make this change when I merge.

PR #5042, opened 2016-02-17
[Feature] Admin Send Reset Password Email

Ghalko:
## Proposal

Enable the admin to send a password reset, even with a link some user might still need help.

## Repercussions

"The greater the power, the more dangerous the abuse." Edmund Burke


admin_tests/users/test_views.py
```diff
@@ -0,0 +1,38 @@
...
+from nose.tools import *  # flake8: noqa
+
+from tests.base import AdminTestCase
+from tests.factories import UserFactory
+from admin_tests.utilities import setup_view
+
+from admin.users.views import UserView
+
+
+class TestUserView(AdminTestCase):
```
mfraezz: Please add some `password reset` tests to this file/class.

mfraezz:

Manual testing verified that an admin could send (at least, to the console) an email to a user to reset their password (when followed, the link also worked), and the modal allowed for the selection between any of the user's verified email addresses. I have minor UI concerns regarding the placement of the `Send reset` button, it may make more sense to group it with the `Emails` list. But I don't foresee that being problematic unless we add a lot of buttons to the user detail view.

As soon as some tests are added for this functionality, I'll put it in RTM.

PR #5043, opened 2016-02-17
Add tests for delete non-linked node, and add exitint linked node

hmoco:
for https://openscience.atlassian.net/browse/OSF-5579
Tests follow JSON api specs:
#### POST:
If all of the specified resources can be added to, or are already present in, the relationship then the server MUST return a successful response.
#### DELETE:
If the client makes a DELETE request to a URL from a relationship link the server MUST delete the specified members from the relationship or return a 403 Forbidden response. If all of the specified resources are able to be removed from, or are already missing from, the relationship then the serverMUST return a successful response.

### DIFF
--- a/api_tests/collections/test_views.py
+++ b/api_tests/collections/test_views.py
@@ -1823,14 +1823,12 @@ class TestCollectionRelationshipNodeLinks(ApiTestCase):
         assert_in(self.linked_node._id, ids)
 
     def test_post_node_already_linked(self):
-        number_of_links = len(self.collection.nodes)
         res = self.app.post_json_api(
             self.url, self.payload([self.linked_node._id]),
             auth=self.user.auth
         )
 
-        assert_equal(res.status_code, 201)
-        assert_equal(len(res.json['data']), number_of_links)
+        assert_equal(res.status_code, 204)
 
     def test_put_contributing_node(self):
         res = self.app.put_json_api(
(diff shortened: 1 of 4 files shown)


api_tests/collections/test_views.py
```diff
@@ -1822,6 +1822,16 @@ def test_post_mixed_nodes(self):
...
         assert_in(self.linked_node._id, ids)
 
+    def test_post_node_already_linked(self):
+        number_of_links = len(self.collection.nodes)
+        res = self.app.post_json_api(
+            self.url, self.payload([self.linked_node._id]),
+            auth=self.user.auth
+        )
+
+        assert_equal(res.status_code, 201)
```
sloria: This is suspect: why is the status code 201 (CREATED) if no nodes actually get linked? I would expect a 200.

sloria: It looks like 204 might be more appropriate:

> 204 No Content
> ...
> Note: This is the appropriate response to a POST request sent to a URL from a to-many relationship link when that relationship already exists...

http://jsonapi.org/format/#crud-updating-relationship-responses-204

sloria:
Pass finished.

hmoco:
@sloria responded to your comment, it now returns a 204.

hmoco:
actually I'm gonna write a custom exception for that

sloria:
> actually I'm gonna write a custom exception for that

:+1:

hmoco:
Seems like probabilistic errors, will close and reopen to trigger travis retry

PR #5044, opened 2016-02-18
[WIP] Feature/retractions causing 400 error when requesting specific related counts

pattisdr:
# Purpose

Fixes edge case found by @sloria in https://openscience.atlassian.net/browse/OSF-5725

As of OSF-5725, you can request specific related counts instead of fetching related counts for all the fields.  For example, a response to `v2/registrations/?related_counts=children` will only return counts for the registrations' children, not other countable fields like `contributors` and `node_links`.  However, we hide the majority of the fields for retracted registrations.  So if a retraction is in a list of registrations, `children` is a hidden field in a retraction, and a 400 error would be thrown. 

# Changes
Ignores related_counts for hidden fields.

### DIFF
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -343,7 +343,7 @@ class RelationshipField(ser.HyperlinkedIdentityField):
             )
         )
 
-    def process_related_counts_parameters(self, params):
+    def process_related_counts_parameters(self, params, value):
         """
         Processes related_counts parameter.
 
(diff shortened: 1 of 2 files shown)


api/base/serializers.py
```diff
@@ -353,10 +353,12 @@ def process_related_counts_parameters(self, params):
         if utils.is_truthy(params) or utils.is_falsy(params):
             return params
 
+        is_retraction = getattr(value, 'is_retracted', False)
```
sloria: This doesn't seem like the ideal approach--`Relationship` is meant to be a generic field (not coupled to OSF), and we're adding registration retraction logic here. There may be other times when we'll need to exclude counts (deleted projects, etc.).

Can you come up with a more general solution?

sloria: Is there a way to know whether the counted field is being hidden? A naive approach might be

```python
field.get_attribute(value) is None
```

since `HideIfRegistration`'s and `HideIfRetraction`'s get_attribute return `None` in order to hide the field. Not sure if there's a better way; haven't look too far into it.

sloria:
Pass finished. Would be good to address https://github.com/CenterForOpenScience/osf.io/pull/5044#discussion_r53333107 if possible, but it's not a hard blocker. Maybe spend 30 min to see if you can come up with a solution. Otherwise, just add a `TODO` and we'll defer.

pattisdr:
Okay - working on this.

sloria:
Looks good now.

pattisdr:
Thanks @sloria, much better after your suggestions to make more generic.

PR #5045, opened 2016-02-18
[Feature] Admin Spam Sort

Ghalko:
## Purpose
To sort based on the newest report time rather than the oldest comment. ADM-63

## Repercussions
Migration.


admin/spam/views.py
```diff
@@ -1,47 +1,149 @@
...
+from .forms import ConfirmForm
 
 
-def get_spam_list():
-    query = (
-        Q('reports', 'ne', {}) &
-        Q('reports', 'ne', None)
-    )
-    return Comment.find(query).sort('date_created')
+def key_order_comments(comment):
```
mfraezz: This appears to be the O(n^2) sorting mechanism mentioned in the PR description, but I don't see it being used anywhere. [This line](https://github.com/CenterForOpenScience/osf.io/pull/5045/files#diff-43b817f0cafd09bc0da4a5351442a2daR31) appears to be what's specifying the ordering.

mfraezz: Put a breakpoint in this function and it wasn't hit. Please remove.

Also, confirmed that it was sorting by most recent report first.

scripts/migrate_spam.py
```diff
@@ -0,0 +1,33 @@
...
+        date = None
+        for user, report in comment.reports.iteritems():
+            report_date = report.get('date')
+            if date is None or report_date > date:
+                date = report_date
+        comment.latest_report = date
+    comment.save()
+
+
+def main():
```
mfraezz: Also confirmed that switching between branches swapped sort orderings without breaking anything -- this script may be superfluous.

scripts/migrate_spam.py
```diff
@@ -0,0 +1,33 @@
...
+    comments = Comment.find()
```
mfraezz: `Comment.find(Q('spam_status', 'ne', 0))` should shave some time off this script, as well as obviate some of the mess in `migrate_comment`.

scripts/migrate_spam.py
```diff
@@ -0,0 +1,33 @@
...
+        if len(comment.reports) > 0:
+            comment.spam_status = Comment.FLAGGED
+    if comment.spam_status not in (Comment.FLAGGED, Comment.SPAM,
+                                   Comment.UNKNOWN, Comment.HAM):
+        comment.spam_status = Comment.UNKNOWN
+    try:
+        temp = comment.latest_report
+    except:
+        date = None
+        for user, report in comment.reports.iteritems():
```
mfraezz: `comment.latest_report = max([r.get('date', None) for u, r in list(comment.reports.iteritems()])`

scripts/migrate_spam.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def main():
+    comments = Comment.find()
+    for comment in comments:
+        migrate_comment(comment)
+
+
+if __name__ == '__main__':
+    main()
```
mfraezz: Please use a `TokuTransaction`, add a file logger, and allow for a `dry-run`. See the [devdocs](http://cosdev.readthedocs.org/en/latest/osf/migrations.html?highlight=script#migrations) and [this part of this script](https://github.com/CenterForOpenScience/osf.io/blob/develop/scripts/dropbox/migrate_to_external_accounts.py#L255-L261) for reference

mfraezz:
Please add a few tests, both for the script and for spam ordering.

mfraezz:
Pass complete. A few small changes and this should be RTM :cactus:

admin/spam/views.py
Ghalko: Removed unused function.

Ghalko:
I guess I need a bit of a tutorial on modm. What I glean from this is that it loads the instance and anything that is empty gets filled with `None`?

mfraezz:
`Comment.find()` returns a `MongoQuerySet` that isn't evaluated (i.e. models aren't loaded into memory) until it's iterated through, either via `list(query_result)` or by a `for` loop, list comprehension, etc. `.find` accepts a single `modularodm` query (or, `Q`) which limits the number of results based on ones that satisfy the query. `Q`s accept three parameters: `attribute`, `operator` (`ne` is not equal, `eq` is equals, `contains` or `icontains` perform regexs, etc), and `argument`, which can be anything (`boolean`, `int`, `str`, some form of complicated object).  Note: multiple `Q`'s `&`'d together count as a single `Q`. You can think of MODM queries, in this case, as a thin wrapper for interacting with the database -- significantly more thin than interacting with the model instances themselves.

So, `Comment.find(Q('spam_status', 'ne', 0) & Q('latest_report', 'eq', None))` limits the results to comments with a non-zero `spam_status` and no `latest_report`, which I think is the set of `Comment` objects the script should be interacting with.

Ghalko:
Please merge #5013 before this one can proceed.

admin_tests/spam/test_views.py
```diff
@@ -2,9 +2,49 @@
...
+        self.comment_3.report_abuse(user=self.user_2, save=True,
+                                    category='spam')
+        self.comment_4.report_abuse(user=self.user_2, save=True,
+                                    category='spam')
+        self.comment_5.report_abuse(user=self.user_1, save=True,
+                                    category='spam')
+        self.comment_6.report_abuse(user=self.user_1, save=True,
+                                    category='spam')
+
+    def test_get_spam(self):
```
mfraezz: Should this test check for correct ordering?

scripts/migrate_spam.py
```diff
@@ -0,0 +1,77 @@
...
+                date = report.get('date')
+            elif date < report.get('date'):
+                date = report.get('date')
+        record.latest_report = date
+        record.save()
+        logger.info('Migrated latest_report for comment {}'.format(record._id))
+
+
+def get_no_status_targets():
+    return [c for c in Comment.find(Q('spam_status', 'eq', None))]
```
mfraezz: You can just return `Comment.find(query)` to save evaluation time.

scripts/migrate_spam.py
```diff
@@ -0,0 +1,77 @@
...
+        with TokuTransaction():
+            migrate_status(get_no_status_targets())
+            if dry_run:
+                raise Exception('Dry Run -- Aborting Transaction')
+    if latest:
+        with TokuTransaction():
+            migrate_latest(get_no_latest_targets())
+            if dry_run:
+                raise Exception('Dry Run -- Aborting Transaction')
+    init_app(set_backends=True, routes=False)
```
mfraezz: You should be doing this before this migration methods are run

scripts/migrate_spam.py
```diff
@@ -0,0 +1,77 @@
...
+    if status:
+        with TokuTransaction():
+            migrate_status(get_no_status_targets())
+            if dry_run:
+                raise Exception('Dry Run -- Aborting Transaction')
+    if latest:
+        with TokuTransaction():
+            migrate_latest(get_no_latest_targets())
+            if dry_run:
+                raise Exception('Dry Run -- Aborting Transaction')
```
mfraezz: You can get both migration methods to run during a dry run by changing this to something like:

```python
with TokuTransaction():
    if status:
        migrate_status(get_no_status_targets())
    if latest:
        migrate_latest(get_no_latest_targets())
    if dry_run:
        raise Exception('Dry Run -- Aborting Transaction')
```

scripts/migrate_spam.py
Ghalko: Done with previous vestiges.

scripts/migrate_spam.py
Ghalko: Done.

scripts/migrate_spam.py
Ghalko: Done.

admin_tests/spam/test_views.py
Ghalko: Checked.

scripts/migrate_spam.py
```diff
@@ -0,0 +1,74 @@
...
+    latest = True
+    if '--dry' in sys.argv:
+        dry_run = True
+    if not dry_run:
+        script_utils.add_file_logger(logger, __file__)
+    if '--no_status' in sys.argv:
+        status = False
+    if '--no_latest' in sys.argv:
+        latest = False
+    if not dry_run:
```
mfraezz: This conditional is doubled

Ghalko: This is done.

scripts/migrate_spam.py
```diff
@@ -0,0 +1,72 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def migrate_status(records):
+    for record in records:
+        if len(record.reports) > 0:
+            record.spam_status = Comment.FLAGGED
+        else:
+            record.spam_status = Comment.UNKNOWN
```
sloria: Is there a difference between `spam_status=UNKNOWN` vs `spam_status=None`? Would it be better if `Comment.UNKNOWN=None`?

website/project/spam/model.py
```diff
@@ -39,6 +39,7 @@ class SpamMixin(StoredObject):
     HAM = 4
 
     spam_status = fields.IntegerField(default=UNKNOWN, index=True)
+    latest_report = fields.DateTimeField(default=None, index=True)
```
sloria: `latest_report` is a bit of a misnomer (it's not a report). For consistency with other datetime fields, how about `date_last_reported`?

sloria:
Pass finished. Approach looks good. Just a few minor changes suggested above

website/project/spam/model.py
Ghalko: Done.

scripts/migrate_spam.py
Ghalko: This makes sense. I will reverse the migration to catch any current `UNKNOWN`s.

sloria:
Looks good 🐙 . Merged into release/0.68.0

PR #5047, opened 2016-02-19
[Feature] admin registrations

Ghalko:
## Proposal
Adds a list of registrations, uses #4960 and #5013 to make a simple but effective base.

## Repercussions
Will make some PRs duplicates.


hmoco:
There seems to be no support for horizontal scrolling. This becomes a problem when the screen is really small (not that big of a deal), or when a registration has a retraction. 
Retraction seem to just be dumped as a JSON string, which at times widens the column so much it prevents contributors column from rendering:
screen shot 2016-02-23 at 12 59 04 pm

screen shot 2016-02-23 at 12 59 12 pm

hmoco:
The complete stringified retraction dictionary seems unnecessary, as it is very hard to read and contains information such as approval_token, I would suggest a simplified representation in the list view (which would also help deter the column stretch problem) , and a better representation such as a table on the detail view.

hmoco:
Allowing horizontal scroll would also fix the problem if this is already defined functionality.

hmoco:
Quick question about desired functionality: how come Node tab and registration tabs work very differently? Unless I'm missing something, there is no way to filter the registration list view, which means if I was to look for a specific registration I would need to scroll until finding it. Seems to defeat the purpose of the admin dashboard. I would suggest either some filtering/searching, or a guid input field.

hmoco:
@Ghalko pass complete

Ghalko:
@hmoco To address your comments backwards:
- Desired functionality was for a list of registrations displaying most of the info displayed. Basically this is the desired list. All of these nodes are also in the Node search so if you have a guid you can find it in the nodes interface. As for searching/filtering beyond guids that is definitely a future improvement.
- I didn't know what Sara wants to know about a retraction. On this list view I will just change it to `True`. But I will ask about details wanted in the node view. That should take care of horizontal scrolling issues.
- The empty status is just because I am reusing a paginator that is also used for spam that requires a status. Refactoring the way spam works currently might enable less variable use.

Ghalko:
All that was requested was `is_retracted` for retractions. So value will show up as `True` or `False` no more very long string dumps.

hmoco:
Cool!

PR #5049, opened 2016-02-19
[WIP][Feature] Analytics migration to Keen IO

GageGaskins:
# Purpose
This PR will be used as the base PR for migrating all analytics to KeenIO from Piwik.

# Side Effects
Currently, Piwik will still be primary analytics for customer facing statistics until all of KeenIO capabilities have been researched. This PR just establishes base event tracking to be used later in the migration.


GageGaskins:
@felliott Should be good for CR

package.json
```diff
@@ -29,6 +29,7 @@
     "jquery.cookie": "^1.4.1",
     "json-loader": "^0.5.3",
     "jstimezonedetect": "^1.0.6",
+    "keen-js": "^3.0.0",
```
felliott: You need to a add `uuid` to package.json, too.

website/static/js/keen.js
```diff
@@ -0,0 +1,143 @@
+var keen = require('keen-js');
```
felliott: You should require jquery, since you use it below.  `var $ = require('jquery');`

website/static/js/keen.js
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        this.keenClient.addEvent('pageviews', pageView, function(err){
+            if(err){
+                throw 'Error sending Keen data: ' + err;
+            }
+        });
+    },
+
+    trackCustomEvent: function(eventCollection, eventData){
```
felliott: This function isn't being used.

website/templates/base.mako
```diff
@@ -173,6 +173,15 @@
             </script>
         % endif
 
+%if keen_project_id:
```
felliott: Minor: please clean up the indentation here.

website/static/js/keen.js
```diff
@@ -0,0 +1,143 @@
...
+        this.keenClient = new keen({
+            projectId: keenProjectId,
+            writeKey: keenWriteKey
+        });
+        this.init();
+    },
+
+    createOrUpdateKeenSession: function() {
+        var date = new Date();
+        var min = 25;
```
felliott: Minor: consider giving this a better name, like `expiresInMinutes`.  `min` is quite generic.

website/static/js/keen.js
GageGaskins: I believe jquery is available on every page through webpack and doesn't need to be required when used.

website/static/js/keen.js
GageGaskins: This was intended for future extensibility by other devs who need more specific things tracked. I can take it out for now if needed.

website/static/js/keen.js
```diff
@@ -0,0 +1,116 @@
...
+            projectId: keenProjectId,
+            writeKey: keenWriteKey
+        });
+        this.init();
+    },
+
+    createOrUpdateKeenSession: function() {
+        var date = new Date();
+        var expiresInMinutes = 25;
+        var expDate = date.setTime(date.getTime() + (min*60*1000));
```
felliott: You forgot to rename `min` here.

GageGaskins: Got it

website/static/js/keen.js
felliott: Please take it out for now.

website/static/js/keen.js
```diff
@@ -0,0 +1,116 @@
...
+                'tags': ctx.node.tags
+            };
+        }
+        if(ctx.currentUser){
+            pageView.user = ctx.currentUser;
+        }
+
+        this.keenClient.addEvent('pageviews', pageView, function(err){
+            if(err){
+                throw 'Error sending Keen data: ' + err;
```
felliott: [Per COSdev](http://cosdev.readthedocs.org/en/latest/style_guides/javascript.html#errors), please throw `Error` instances instead of strings.

website/static/js/keen.js
felliott: It'll work, but it's still good practice to be explicit about your dependencies.

felliott: Oh, and don't forget `'use strict';`.

felliott:
CR complete. Looks good!

website/static/js/keen.js
```diff
@@ -0,0 +1,115 @@
...
+        if(!$.cookie('keenUserId')){
+            $.cookie('keenUserId', uuid.v1(), {expires: 365, path: '/'});
+        }
+
+        return $.cookie('keenUserId');
+    },
+
+    trackPageView: function(){
+        this.createOrUpdateKeenSession();
+        var ctx = window.contextVars;
```
sloria: `contextVars` should not be used in re-usable modules. Instead, any data this class depends on should be passed in the constructor.

webpack.common.config.js
```diff
@@ -127,12 +127,14 @@ var resolve = {
...
         'pikaday-css': nodePath('pikaday/css/pikaday.css'),
+        'keen-js': nodePath('keen-js'),
         // Also alias some internal libraries for easy access
         'addons': path.join(__dirname, 'website', 'addons'),
         'tests': staticPath('js/tests'),
         // GASP Items not defined as main in its package.json
         'TweenLite' : nodePath('gsap/src/minified/TweenLite.min.js'),
-        'EasePack' : nodePath('gsap/src/minified/easing/EasePack.min.js')
+        'EasePack' : nodePath('gsap/src/minified/easing/EasePack.min.js'),
+        'uuid': nodePath('uuid')
```
sloria: This shouldn't be necessary. The `uuid` package is installed by npm and properly declares the `main` entry (https://github.com/defunctzombie/node-uuid/blob/master/package.json#L22).

I'm also not sure if the `keen-js` alias is necessary.

scripts/migrate_piwik_historical_data.py
```diff
@@ -0,0 +1,152 @@
+import MySQLdb
+from socket import inet_ntoa
+from binascii import b2a_hex
+import pprint
+from keen import KeenClient
+import urlparse
+
+def main():
+
+    keen_client = KeenClient(
```
sloria: Can you please document any external dependencies that need to be installed in the docstring for this script?

website/templates/base.mako
```diff
@@ -173,6 +173,15 @@
             </script>
         % endif
 
+        %if keen_project_id:
+            <script>
+                window.contextVars = $.extend(true, {}, window.contextVars, {
+                    keenProjectId: '${keen_project_id}',
```
sloria: Glancing at how other contextVars are declared, it looks like the proper way to JSON-encode variables in our mako templates is:

```mako
keenProjectId: ${ keen_project_id | sjson, n }
```

scripts/migrate_piwik_historical_data.py
```diff
@@ -0,0 +1,152 @@
...
+        project_id='56abc1c759949a6734c3f32b',
```
sloria: Shouldn't this be imported from `website.settings`?

website/templates/base.mako
```diff
@@ -173,6 +173,15 @@
...
+                    keenWriteKey: '${keen_write_key}'
```
sloria: Are you 100% certain that this can be publicly shared?

webpack.common.config.js
```diff
@@ -127,12 +127,14 @@ var resolve = {
         'wikiPage': addonsPath('wiki/static/wikiPage.js'),
         'highlight-css': nodePath('highlight.js/styles/default.css'),
         'pikaday-css': nodePath('pikaday/css/pikaday.css'),
+        'keen-js': nodePath('keen-js'),
```
sloria: This is indeed unnecessary. keen-js is installed with npm and properly declares the `browser` entry in its package.json (https://github.com/keen/keen-js/blob/master/package.json#L5).

scripts/migrate_piwik_historical_data.py
GageGaskins: This script won't be used for a while, so that id is just a project id I had for a test on Keen to test the script. It will be implemented correctly later, but now it simply exists as a placeholder until proper migration occurs.

website/templates/base.mako
GageGaskins: The plan for Keen is to do something a la Tornadik with some munger/server middle man that takes all requests which holds the proper Keen settings. However, until that is implemented, there is no way to send data to Keen through js without exposing our keys. If you have a suggestion as to what should be done, I would gladly take the help.

scripts/migrate_piwik_historical_data.py
```diff
@@ -0,0 +1,152 @@
...
+        write_key='248de5ffaf1c829bf74dc6e4e0fc3af6beaa7f5194c2d163699213b66e8cf1fbc41a1adc1dd9b6ec3681ac31584bd0c9847c8d9ca223e44e99e60af219be19c55a9261437020069f92c5825cc2995cba57976ed485e3a9f314e3a1d21d032e53'
+    )
+
+    try:
+        db = MySQLdb.connect(host='127.0.0.1', port=8889, user='root', passwd='root', db='piwikOrig')
```
sloria: These values should not be hardcoded. Maybe put this script in `scripts.keen` and add a `settings` folder where this info can be configured. See `scripts.osfstorage` for an example.

website/static/js/keen.js
```diff
@@ -0,0 +1,115 @@
...
+        var returning = $.cookie('keenUserId') ? true : false;
```
sloria: Minor: The ternary is unnecessary

```python
var returning = Boolean($.cookie('keenUserId'));
```

sloria:
Pass finished.

GageGaskins:
Closing Pull request to recreate without script

PR #4985, opened 2016-02-04
[Feature] institutions auth

chrisseto:

PR #5051, opened 2016-02-22
[feature] institutional auth

icereval:
Replaces #4985


api/institutions/authentication.py
```diff
@@ -0,0 +1,68 @@
...
+        #     "user": {
+        #         "middleNames": "",
+        #         "familyName": "",
+        #         "givenName": "",
+        #         "fullname": "Circle User",
+        #         "suffix": "",
+        #         "username": "user@circle.edu"
+        #     }
+        # }}
+        data = json.loads(payload['data'])
```
sloria: Could a `KeyError` be raised here?

api/institutions/authentication.py
```diff
@@ -0,0 +1,68 @@
...
+        provider = data['provider']
+
+        institution = Institution.load(provider['id'])
+        if not institution:
+            raise AuthenticationFailed('Invalid institution id specified "{}"'.format(provider['id']))
+
+        username = provider['user']['username']
+        fullname = provider['user']['fullname']
+
+        user, created = get_or_create_user(fullname, username)
```
sloria: You might need to lowercase `username` here; it doesn't look like `get_or_create_user` will lowercase it, so it might be possible to create duplicate users with differently-cased usernames.

sloria: Ah, I guess username gets lower-cased in `User#save`, so disregard previous comment.

api/institutions/authentication.py
```diff
@@ -0,0 +1,68 @@
...
+
+        if created:
+            user.given_name = provider['user'].get('givenName')
+            user.middle_names = provider['user'].get('middleNames')
+            user.family_name = provider['user'].get('familyName')
+            user.suffix = provider['user'].get('suffix')
+            user.save()
+
+            # User must be saved in order to have a valid _id
+            user.register(username)
```
sloria: `User#register` does not implicitly save, so you probably need to call `.save()` here.

sloria:
Pass finished.

api/institutions/authentication.py
chrisseto: If the user was just created then the below if block will always be true and user.save will be called.

api/institutions/authentication.py
chrisseto: @icereval

api/institutions/authentication.py
sloria: Ah ok; thanks for the clarification.

PR #5055, opened 2016-02-23
[Hotfix] Send request to import account iff user does not cancel.

mfraezz:
[#OSF-5756]

### DIFF
--- a/website/static/js/oauthAddonNodeConfig.js
+++ b/website/static/js/oauthAddonNodeConfig.js
@@ -171,13 +171,15 @@ var OauthAddonFolderPickerViewModel = oop.extend(FolderPickerViewModel, {
     */
     connectExistingAccount: function(account_id) {
         var self = this;
-
-        return $osf.putJSON(
-            self.urls().importAuth, {
-                external_account_id: account_id
-            }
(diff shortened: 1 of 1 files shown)


sloria:
I don't understand why this is necessary; don't we prevent the request if the user cancels here: https://github.com/CenterForOpenScience/osf.io/pull/5055/files#diff-780e22e6e487060eaba79b328e4c746cR156 ?

mfraezz:
That's a different case -- it's hit when the user has only one ExternalAccount connected for the addon

mfraezz:
[This](https://github.com/CenterForOpenScience/osf.io/pull/5055/files#diff-780e22e6e487060eaba79b328e4c746cR130) is the troublesome case. `connectExistingAccount` is being `.bind`-ed as the callback function. The dropdown's `value` (an `external_account_id`) is passed in if the user clicks `Import`, `null` if `Cancel`. The case where they have only one account assumes that their only account is the one that they choose, and *is* checking that they confirmed the import, but is doing so before `connectExistingAccount` is called.

sloria:
Thanks for the clarification.

PR #5056, opened 2016-02-23
[Feature] Admin Desk [OSF-6258]

Ghalko:
## Proposal
Integrate desk.com into admin to facilitate users/cases continuity

## Deploy
Please insert Desk tokens into `admin.base.settings.local.py`

## Repercussions
A second saved is a second earned. - Not Ben Franklin

## Ticket

https://openscience.atlassian.net/browse/OSF-6258


admin/base/settings/defaults.py
```diff
@@ -219,3 +224,10 @@
 
 ENTRY_POINTS = {'osf4m': 'osf4m', 'prereg_challenge_campaign': 'prereg',
                 'institution_campaign': 'institution'}
+
+
+# Change desk settings in local.py
+DESK_KEY = ''
+DESK_KEY_SECRET = ''
+DESK_TOKEN = ''
```
mfraezz: I know Desk allows this token and `DESK_TOKEN_SECRET` to be manually generated, and used to authenticate on behalf of a Desk user. 

Should these be attached to a user, rather than stored in the settings? This would allow us to have multiple Desk users managing cases, and wouldn't be a horrible security risk because these are useless without the `KEY` and `KEY_SECRET`

admin/desk/urls.py
```diff
@@ -0,0 +1,11 @@
+from django.conf.urls import url
+
+from . import views
```
mfraezz: Please use full module import paths

admin/desk/utils.py
```diff
@@ -0,0 +1,194 @@
...
+"""
+
+import requests
+import json
+from requests_oauthlib import OAuth1Session
+
+from django.conf import settings
+
+
+class DeskError(Exception):
```
mfraezz: Why/Are these overrides necessary?

admin/desk/views.py
```diff
@@ -0,0 +1,64 @@
...
+    template_name = 'desk/customer.html'
+    context_object_name = 'customer'
+
+    def get(self, request, *args, **kwargs):
+        try:
+            return super(DeskCustomer, self).get(request, *args, **kwargs)
+        except (AttributeError, DeskError) as e:
+            return render(request, 'desk/user_not_found.html',
+                          context={
+                              'email': e.status,
```
mfraezz: This will `raise` an `AttributeError` if `isinstance(e, AttributeError)`, as `.status` is being set only in `DeskError`. (`.message` for `AttributeError`s)

admin/desk/utils.py
mfraezz: Also, this should probably go somewhere like `admin/desk/exceptions.py`

admin/base/settings/defaults.py
Ghalko: I believe these credentials are shared between desk admin users. We could get a set for each user that wants to deal with users on Desk. This would then be along the lines of email and password except it would be 4 fields rather than 2.

As an alternative you are welcome to discuss opening a port with Michael in the VPN.

admin/desk/urls.py
Ghalko: Done.

admin/desk/views.py
Ghalko: Changed to message in the class and use.

admin/desk/utils.py
Ghalko: It allows us to have a differentiated exception raised rather than raising a generic exception.

Sticking this in `exceptions.py` is not a standard practice. It would be similar to splitting up models and managers into `models.py` and `managers.py`

admin/base/settings/defaults.py
mfraezz: Two are user-specifc: the tokens. Keys *should* be specific to the `openscience` Desk site.

admin/desk/utils.py
mfraezz: I understand the rationale behind subclassing `Exception`. I don't think the overrides are necessary.

```python
In [29]: e = Exception('Verbose description of error')

In [30]: e.__str__()
Out[30]: 'Verbose description of error'

In [31]: e.__unicode__()
Out[31]: u'Verbose description of error'

In [32]: e.message
Out[32]: 'Verbose description of error'
```

mfraezz:
Why is there spam / user refactor code in this PR?

Ghalko:
@mfraezz Basically this replaces the need for email in spam. Also the buttons to look at desk customers is on the user's detail page.

Ghalko:
Also it seems I need to clean up a few things.

admin/templates/users/user.html
```diff
@@ -79,7 +96,8 @@
                 <tr>
                     <td>OSF account</td>
                     <td>
-                        <a href="">{{ user.id }}</a>
+                        hi {{ osf_link }} bleh
```
mfraezz: >hi  `<link>` bleh

admin/users/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+from django.forms import ModelForm
+
+
+class OSFUser(models.Model):
```
mfraezz: This file appears unused.

admin/templates/users/user.html
Ghalko: Frustrated debugging. Removed.

admin/base/settings/defaults.py
mfraezz: My suggestion is to allow an admin user to enter their own tokens, enabling the Desk UI iff they have (possibly also requiring inclusion in a `desk_admin` 'group' or some such).

Ghalko: That would be a good improvement. This is getting Desk integration to Sara.

admin/desk/utils.py
Ghalko: Removed the duplication.

admin/desk/utils.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def __init__(self, message):
```
mfraezz: Also a superfluous override. 
```python
class DeskError(Exception):
    pass
```
should be sufficient.

admin/base/settings/defaults.py
mfraezz: Alright. I'll let @sloria decide if this is a blocker

admin/templates/users/user.html
```diff
@@ -19,12 +19,29 @@
...
             </a>
             </span>
-            <span class="col-md-2">
-            <a href=""
-               class="btn btn-primary disabled">
-                User notes
-            </a>
-            </span>
+            <div class="btn-group" role="group">
+                <span class="btn btn-primary disabled">Desk</span>
```
mfraezz: As discussed in the demo, please make this not a button.

mfraezz:
LGTM. Dropping this in RTM 🔥

admin/desk/utils.py
```diff
@@ -0,0 +1,185 @@
...
+        if r.status_code != requests.codes.ok:
+            raise DeskError(str(r.status_code))
+        return json.loads(r.content)
+
+    def call_post(self, service, data=None):
+        """ Calls a POST API for the given service name and POST data."""
+        url = self.build_url(service)
+        r = self.oauth.post(url, data=json.dumps(data))
+        if r.status_code >= 400:
+            raise DeskError(str(r.status_code))
```
sloria: Is there more useful information you can gather besides the status code? The response body might be more useful than the status code for debugging.

admin/desk/utils.py
```diff
@@ -0,0 +1,185 @@
...
+        """ Constructs the URL for a given service."""
+        return 'https://%s.%s/%s' % (self.SITE_NAME, DeskClient.BASE_URL, service)
+
+    def call_get(self, service, params=None):
+        """ Calls a GET API for the given service name and URL parameters."""
+        url = self.build_url(service)
+        r = self.oauth.get(url, params=params)
+        if r.status_code != requests.codes.ok:
+            raise DeskError(str(r.status_code))
+        return json.loads(r.content)
```
sloria: Minor: I believe you can do `r.json()` here.

admin/desk/utils.py
```diff
@@ -0,0 +1,185 @@
...
+                customer_data = {
+                    'id': customer['id'],
+                    'name': '{} {}'.format(
+                        customer['first_name'], customer['last_name']),
+                    'emails': customer['emails'],
+                    'background': customer['background'],
+                    'company': customer['company'],
+                    'link': customer['_links']['self']
+                }
+        except DeskError:
```
sloria: Why let this error pass?

admin/desk/utils.py
```diff
@@ -0,0 +1,185 @@
...
+                "level": "super"
+              }
+            }
+        """
+        customer_link = None
+        try:
+            customer_json = self.call_post('customers', data)
+            if not customer_json.get('errors'):
+                customer_link = customer_json['_links']['self']['href']
+        except DeskError:
```
sloria: Again, I'm not sure if letting this error pass is the correct thing to do.

admin/base/settings/defaults.py
sloria: +1 to Matt's suggestion. I wouldn't consider it a blocker to getting this PR merged. @Ghalko Could you create an improvements ticket for a followup PR implementing this?

admin/desk/views.py
```diff
@@ -0,0 +1,64 @@
...
+                          })
+
+    def get_object(self, queryset=None):
+        customer_id = self.kwargs.get('user_id', None)
+        customer = User.load(customer_id)
+        email = customer.emails[0]
+        desk = DeskClient()
+        params = {'email': email}
+        customer = desk.find_customer(params)
+        if customer == {}:
```
sloria: As suggested in previous comments, rather than letting `DeskError` pass silently in `find_customer`, why not allow the exception to be raised and handle it here?

admin/desk/views.py
```diff
@@ -0,0 +1,64 @@
...
+        kwargs.setdefault('desk_case', 'https://{}.desk.com/web/agent/case/'.format(DeskClient.SITE_NAME))
+        kwargs.setdefault('desk_customer', 'https://{}.desk.com/web/agent/customer/'.format(DeskClient.SITE_NAME))
+        return super(DeskCaseList, self).get_context_data(**kwargs)
+
+
+class DeskCustomer(OSFAdmin, DetailView):
+    template_name = 'desk/customer.html'
+    context_object_name = 'customer'
+
+    def get(self, request, *args, **kwargs):
```
sloria: Overriding `get` in a generic CBV is an antipattern; you should only need to override hooks (`get_object`, etc.). Are you sure this is absolutely necessary? Can you handle the AttributeError and DeskError somewhere else?

sloria:
Pass finished.

admin/base/settings/defaults.py
Ghalko: I don't see this as worthy of its own PR. I have already gone back and forth a few times between different options so changing it one more time shouldn't take too long...

sloria: OK, it's your call. If it's a small change, go for it!

Ghalko: Pretty much done.

admin/desk/utils.py
Ghalko: I've added the reason, but there are responses without a body.

admin/desk/views.py
Ghalko: In Django there is only one http-error `Http404` The "antipattern" of catching the error in `ge`t or `post` seems to be a pretty common solution for custom errors. Or better yet to catch the errors in `dispatch`. If a try/except block returns a render then it might try to use this rendered error page as the item it was seeking in `get_queryset` or whatever method the error might occur in. Also a bit DRYer to deal with an issue in one place rather than every place that it might happen.

admin/common_auth/models.py
```diff
@@ -49,6 +49,10 @@ class MyUser(AbstractBaseUser, PermissionsMixin):
     date_joined = models.DateTimeField(auto_now_add=True)
     confirmed = models.BooleanField(default=False)
     osf_id = models.CharField(max_length=5, blank=True)
+    desk_key = models.CharField(max_length=45, blank=True)
```
sloria: As Matt mentioned [here](https://github.com/CenterForOpenScience/osf.io/pull/5056#discussion_r65406994), the key and key secret are specific to the Desk.com API app, not specific to the user. So no need to store these on the user.

admin/desk/utils.py
sloria: The reason does not provide any more information than status code because there is a 1-to-1 mapping between status codes and reason (e.g. 200 = 'OK').

It would still be useful to have the request body in the error if it is available.

One approach is to store the body and the status code as attributes of DeskError.

```python
class DeskError(..):
    def __init__(self, message, status_code=None, content=None):
        super(...).__init__(message)
        self.status_code = status_code
        self.content = content
```

admin/desk/views.py
```diff
@@ -0,0 +1,74 @@
...
+    ordering = 'updated_at'
+    context_object_name = 'cases'
+    paginate_by = 100
+    paginate_orphans = 5
+
+    def dispatch(self, request, *args, **kwargs):
+        try:
+            return super(DeskCaseList, self).dispatch(request, *args, **kwargs)
+        except DeskError as e:
+            return render(request, 'desk/desk_error.html',
```
sloria: If you implement https://github.com/CenterForOpenScience/osf.io/pull/5056/files#r67889102 , you could return the proper status code here (using `e.status_code`).

admin/desk/views.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class DeskCustomer(OSFAdmin, DetailView):
+    template_name = 'desk/customer.html'
+    context_object_name = 'customer'
+
+    def dispatch(self, request, *args, **kwargs):
+        try:
+            return super(DeskCustomer, self).dispatch(request, *args, **kwargs)
+        except AttributeError as e:
```
sloria: Why an `AttributeError`? It looks like the exception handling can be improved here. There are many reasons an AttributeError might be raised, so it doesn't seem correct to catch that here. Instead, a more specific error should be handled.

admin/desk/utils.py
```diff
@@ -0,0 +1,179 @@
...
+            'sort_field': 'created_at',
+            'sort_direction': 'desc'
+        })
+        try:
+            case_list = self.call_get('cases/search', params)
+        except DeskError:
+            raise
+        return case_list[u'_embedded'][u'entries']
+
+    def create_find_customer_by_email(self, email, full_name=None):
```
sloria: Minor: for consistency with other parts of the codebase, it would make sense to name this "find_or_create_customer_by_email".

sloria:
Pass finished.

admin/desk/utils.py
Ghalko: I can certainly do that, but in my experience Desk API really doesn't tell you anything in the body.

admin/desk/utils.py
Ghalko: Raised.

admin/desk/utils.py
Ghalko: Removed extra, possible future utilities.

admin/common_auth/models.py
Ghalko: I think that these are unique for all generated instances, but I will check with Sara once she is out of her meeting.

admin/desk/views.py
Ghalko: I can certainly look at refactoring the admin project.

admin/desk/utils.py
Ghalko: Removed as we don't use this.

admin/common_auth/models.py
mfraezz: That's not how OAuth applications are supposed to work, but it's not impossible that Desk did something weird. `key` and `key_secret` should be associated with an app specific to the COS account, with `token` and `token_secret` being specific to a user in the `openscience` group.

mfraezz: Confirmed with Sara just now, `key` and `key_secret` are associated with the group, not the user.

Ghalko: Then I will change it again.

admin/desk/utils.py
Ghalko: Changed

admin/common_auth/models.py
Ghalko: Split.

admin/desk/views.py
Ghalko: Done.

admin/desk/utils.py
mfraezz: If the only exception you're catching is being reraised, the `try`/`except` is superfluous. Consider removing

admin/desk/views.py
sloria: I still don't understand---what is raising the `AttributeError`?

admin/desk/views.py
sloria: You are not actually returning the status code in the response. You need to pass the `status` argument to `render`.

https://docs.djangoproject.com/en/1.9/topics/http/shortcuts/#render

sloria:
Pass finished. This looks really close to ready to merge. Just a small change (https://github.com/CenterForOpenScience/osf.io/pull/5056#discussion_r69141862). We can defer https://github.com/CenterForOpenScience/osf.io/pull/5056#discussion_r69141862 , but I'd like to understand what's going on there before I merge this.

admin/desk/views.py
Ghalko: The context is the part of the render. The status in the context is passed in as `status` to the template renderer and then when the renderer sees `status` it puts that string in there. See the next following lines for the context.

admin/desk/views.py
Ghalko: `get_object` maybe I should have my own error?

Ghalko: Made my own simple error class that allows me to differentiate the error from an attribute error.

admin/desk/views.py
sloria: Yes, I see that you are passing the status code to the template context. But the response will not have the correct status code unless you pass the `status` argument to `render`.

admin/desk/utils.py
sloria: ^ Matt is correct here; no need to handle the error if all you're doing is raising it.

admin/desk/views.py
```diff
@@ -0,0 +1,86 @@
...
+                          })
+
+    def get_object(self, queryset=None):
+        customer_id = self.kwargs.get('user_id', None)
+        customer = User.load(customer_id)
+        email = customer.emails[0]
+        desk = DeskClient(self.request.user)
+        params = {'email': email}
+        customer = desk.find_customer(params)
+        if customer == {}:
```
sloria: This should never happen, since you are now (correctly) raising a `DeskError` if the customer can't be found. 

What if `find_customer` raised a `DeskCustomerNotFound` error? You could then handle that error in `dispatch` and render the `user_not_found.html` response.

sloria:
Pass finished. Just a couple more small things to address:

- [ ] https://github.com/CenterForOpenScience/osf.io/pull/5056#discussion_r69141862
- [ ] https://github.com/CenterForOpenScience/osf.io/pull/5056/files#r70267764

admin/desk/views.py
Ghalko: Returning 0 results doesn't throw an error, even in Desk API response since it is more of a search. So checking for an empty customer is necessary. Of course there are other implications with this, but for the moment I think these implications are reasonable.

sloria: Even though you don't get an error response back from Desk, it still makes sense to raise an error in `find_customer` rather than raising it here. That way, we can keep the logic in this view thin. Basically, the view should just make the Desk API request and handle errors appropriately.

Not a huge deal though. Even if you do decide to raise the error here, the error should be more specific to Desk users, e.g.

```python
raise DeskCustomerNotFound('Could not find customer with params: {}'.format(params), email=email)
```

admin/templates/base.html
```diff
@@ -75,6 +75,11 @@
                             Admin-User Registration
                         </a>
                     </li>
+                    <li>
```
sloria: Minor: Fix indentation.

admin/users/serializers.py
```diff
@@ -12,6 +18,7 @@ def serialize_user(user):
         'last_login': user.date_last_login,
         'disabled': user.date_disabled if user.is_disabled else False,
         'two_factor': user.has_addon('twofactor'),
+        'osf_link': abs_url.url,
```
sloria: Minor: You can do `user.absolute_url` here.

admin_tests/spam/test_views.py
```diff
@@ -19,7 +17,7 @@
     SpamList,
     UserSpamList,
     SpamDetail,
-    EmailFormView,
+    EmailView,
```
sloria: Typo here.

admin/desk/views.py
Ghalko: Done.

admin/templates/base.html
Ghalko: Thanks...

admin/users/serializers.py
Ghalko: Switched.

admin/desk/views.py
Ghalko: Understood this time. Placed in 3 instances.

PR #5059, opened 2016-02-23
Get & Update guid referents & add addon file comments

samanehsan:
Purpose:
------------
Make guids for addon files remain the same when an addon file/folder is moved or renamed. This will make addon file comments remain on the file even if it is moved/renamed.

Fixes https://openscience.atlassian.net/browse/OSF-5573

Changes:
-------------
This PR allows comments on all of our storage addons. It also handles making sure comments "move" with addon files when they are moved/renamed or a parent folder is moved/renamed:

1. Get file guids by looking up the file's StoredFileNode, or by searching for all StoredFileNodes starting with the folder's materialized path.

2. Update the `Guid.referent` to the new stored file node

3. If a file is being moved between nodes & has comments, update the node field for those comments to the destination node.

**Note this PR requires #5104 to be merged so that guids that belong to deleted StoredFileNodes are not updated**


api/files/serializers.py
```diff
@@ -197,6 +196,15 @@ def update(self, instance, validated_data):
     def is_valid(self, **kwargs):
         return super(FileSerializer, self).is_valid(clean_html=False, **kwargs)
 
+    def get_file_guid(self, obj):
+        if obj:
+            guid = obj.get_guid()
+            if guid:
+                return guid._id
+            else:
```
brianjgeiger: Since the else branch and the default both return the same thing, you shouldn't need the else branch. Just let it fall through.

api/base/serializers.py
```diff
@@ -448,7 +448,11 @@ def get_url(self, obj, view_name, request, format):
                 else:
                     url = self.reverse(view, kwargs=kwargs, request=request, format=format)
                     if self.filter:
-                        url = '{}?filter{}'.format(url, self.format_filter(obj))
+                        formatted_filter = self.format_filter(obj)
+                        if formatted_filter:
```
brianjgeiger: Can you describe to me the circumstance in which `self.filter` is `True` but `formatted_filter` is `False` and why, when both of these are the case, the url is `None` rather than an unfiltered URL or similar? This may be obvious when I get later in the code, but just in case it's not.

api/base/serializers.py
```diff
@@ -470,7 +474,18 @@ def format_filter(self, obj):
         qd = QueryDict(mutable=True)
         filter_fields = self.filter.keys()
         for field_name in filter_fields:
-            qd.update({'[{}]'.format(field_name): self.lookup_attribute(obj, self.filter[field_name])})
+            try:
+                # check if serializer method passed in
+                value = getattr(self.parent, self.filter[field_name])
+            except AttributeError:
+                value = self.lookup_attribute(obj, self.filter[field_name])
+            else:
```
brianjgeiger: I didn't know about try-else before. That's pretty cool.

api/base/serializers.py
```diff
@@ -470,7 +474,18 @@ def format_filter(self, obj):
...
+                value = value(obj)
```
brianjgeiger: Before this line is evaluated, what is `value`? Is it the serializer method? If so, could you store that in a more specifically-named variable and then store the result into value? If not, I'm not sure what's going on here.

samanehsan: That's right! `value` is the serializer method -- I'll give that a better name

api/base/serializers.py
samanehsan: `self.filter` is `True` when the `filter` kwarg is passed into a `RelationshipField`

In this case, I'm passing in `filter={'target': 'get_file_guid'} to the files serializer's comments relationship field. Formatted filter gets the values and formats them into query params. If a file does not have a guid, `formatted_filter` is `None`  so the comments url should also be `None.`

website/files/models/osfstorage.py
```diff
@@ -33,6 +34,25 @@ def get_or_create(cls, node, path):
         # Dont raise anything a 404 will be raised later
         return cls.create(node=node, path=path)
 
+    @classmethod
+    def get_file_guids(cls, materialized_path, provider, guids, node=None):
+        path = materialized_path
```
brianjgeiger: Minor, but it looks like you could probably just strip the `/` here and make the two lines below a tiny bit more readable.

tests/test_comments.py
```diff
@@ -280,3 +292,676 @@ def test_find_unread_does_not_include_deleted_comments(self):
...
+        payload = self._create_payload('move', self.user, source, destination, self.file._id)
+        update_file_guid_referent(self=None, node=destination['node'], event_type='addon_file_moved', payload=payload)
+        self.guid.reload()
+
+        file_node = FileNode.resolve_class(self.provider, FileNode.FILE).get_or_create(destination['node'], self._format_path('{}{}'.format(destination['path'], file_name), file_id=self.file._id))
+        assert_equal(self.guid._id, file_node.get_guid()._id)
+        file_comments = Comment.find(Q('root_target', 'eq', self.guid._id))
+        assert_equal(file_comments.count(), 1)
+
+    @unittest.skip
```
brianjgeiger: Under what circumstances will we remove the skips from these tests?

website/project/views/comment.py
```diff
@@ -17,30 +18,69 @@
...
-
-
-def update_folder_contents(children, source_node, destination_node):
+        file_guids = FileNode.resolve_class(source['provider'], FileNode.ANY).get_file_guids(
+            materialized_path=source['materialized'] if source['provider'] != 'osfstorage' else source['path'],
+            provider=source['provider'],
+            guids=[],
+            node=source_node)
+
+        if event_type == 'addon_file_renamed' and (source['provider'] == 'osfstorage' or source['provider'] == 'box'):
```
brianjgeiger: This looks troublesome to maintain. I saw some lists of providers up top. Could you use those, or make new ones in the same place, to check for providers that can or can't handle these types of events?

website/project/views/comment.py
```diff
@@ -17,30 +18,69 @@
...
+            return
+        if event_type == 'addon_file_moved' and (source['provider'] == destination['provider'] and (source['provider'] == 'osfstorage' or source['provider'] == 'box')) and source_node == destination_node:
```
brianjgeiger: Same here w/r/t the lists

website/project/views/comment.py
```diff
@@ -17,30 +18,69 @@
 
 
 @file_updated.connect
-def update_comment_root_target_file(self, node, event_type, payload, user=None):
-    if event_type == 'addon_file_moved':
+def update_file_guid_referent(self, node, event_type, payload, user=None):
```
brianjgeiger: This is a fairly hefty function, and the logic is fairly involved in the if statements. It might be a good idea to break it up a bit so that the logic is separate from the functionality.

website/project/views/comment.py
```diff
@@ -17,30 +18,69 @@
...
-        else:
-            Comment.update(Q('root_target', 'eq', item._id), data={'node': destination_node})
-            # update node record of commented files
-            if item._id in source_node.commented_files:
-                destination_node.commented_files[item._id] = source_node.commented_files[item._id]
-                del source_node.commented_files[item._id]
-                source_node.save()
-                destination_node.save()
+        if item['kind'] == 'folder':
+            return create_file_from_metadata(item.get('children', []), source, destination, destination_node, obj)
```
brianjgeiger: This is recursive? Could you add a docstring describing what it's doing and why? I'm guessing it's for… moving a folder or file and ensuring you keep the comments?

tests/test_comments.py
samanehsan: The payload format when an addon folder is moved or renamed is not consistent between addons or even within the same addon (with the different possible move/rename actions) -- some include a "children" field which includes the folder's serialized contents and some do not. The tests still pass, but they are not quite accurate so @sloria suggested skipping them.

We can remove the skips if/when waterbutler returns a consistently (or at least predictably) formatted payload.

brianjgeiger: Cool. Maybe a quick note about that in the skiptest decorator and/or a TODO.

website/project/views/comment.py
```diff
@@ -17,30 +18,69 @@
...
+            return
+        if event_type == 'addon_file_moved' and (source['provider'] == destination['provider'] and (source['provider'] == 'osfstorage' or source['provider'] == 'box')) and source_node == destination_node:
+            return
+
+        for guid in file_guids:
+            obj = Guid.load(guid)
+            if source_node != destination_node and Comment.find(Q('root_target', 'eq', guid)).count() != 0:
+                update_comment_node(guid, source_node, destination_node)
+
+            if not (source['provider'] == destination['provider'] and source['provider'] == 'osfstorage'):
```
brianjgeiger: Not a big deal, but this might be a bit easier to follow if it were

```py
if source['provider'] != destination['provider'] or source['provider'] != 'osfstorage':
```

website/project/views/comment.py
```diff
@@ -17,30 +18,69 @@
...
+            data = dict(item)
+            new_file = FileNode.resolve_class(destination['provider'], FileNode.FILE).get_or_create(destination_node, item['path'])
+            if destination['provider'] != 'osfstorage':
+                new_file.update(revision=None, data=data)
+            return new_file
+
+
+def update_comment_node(root_target_id, source_node, destination_node):
+    Comment.update(Q('root_target', 'eq', root_target_id), data={'node': destination_node})
+    destination_node.commented_files[root_target_id] = source_node.commented_files[root_target_id]
```
brianjgeiger: As discussed in person, this concerns me greatly. If we can find a better way that doesn't involve locking the node, I'd be much happier.

brianjgeiger:
Initial pass finished. After you have it working again and have responded to these items, let's take another look at the demo, and then I'll do a follow-up pass. Most of it's minor, with one medium-difficulty refactor, and one big request.

website/project/views/comment.py
samanehsan: Do you have any suggestions for breaking up this function further?

brianjgeiger: Mostly just everything inside the `if` statements would be broken into functions. Maybe just inside the terminal `if` statements.

api/base/serializers.py
brianjgeiger: Makes sense.

website/project/views/comment.py
samanehsan: As discussed on flowdock and with @sloria, `Node.commented_files` can be removed for now since it's used when calculating the number of unread file comments, but that unread count is currently not being used and is expensive to calculate.

samanehsan:
Done making those changes @brianjgeiger! :sunny: I'll remove `Node.commented_files` in a separate PR.

brianjgeiger:
As discussed in person, see if you can find a file rename test in the API tests and add an assert to verify that the file guid doesn't change. It's not necessarily a blocker if it doesn't, but we will want to know for sure.

samanehsan:
@brianjgeiger, there doesn't seem to be any API tests for renaming files

brianjgeiger:
Harrumph.

brianjgeiger:
Could you make a note to QA to have them verify that?

samanehsan:
Sure!

api/nodes/views.py
```diff
@@ -1907,7 +1904,10 @@ def get_queryset(self):
                     del root_target.referent.node.commented_files[root_target._id]
 
             else:
-                url = waterbutler_api_url_for(self.get_node()._id, root_target.referent.provider, root_target.referent.path, meta=True)
+                referent = root_target.referent
+                if referent.provider == 'dropbox':
```
sloria: Why the special case for Dropbox?

samanehsan: The `StoredFileNode.path` for Dropbox includes the root folder, but waterbutler is expecting just the relative path otherwise it will throw a 404 error.

website/files/models/base.py
```diff
@@ -251,6 +251,29 @@ def get_or_create(cls, node, path):
             return cls.create(node=node, path=path)
 
     @classmethod
+    def get_file_guids(cls, materialized_path, provider, guids, node):
```
sloria: It might be good to default `guids` to `None` then do `guids = guids or []`, so that the caller doesn't have to pass an empty list every time this method is invoked.

tests/test_comments.py
```diff
@@ -280,3 +292,678 @@ def test_find_unread_does_not_include_deleted_comments(self):
...
+        payload = self._create_payload('move', self.user, source, destination, self.file._id)
+        update_file_guid_referent(self=None, node=destination['node'], event_type='addon_file_moved', payload=payload)
+        self.guid.reload()
+
+        file_node = FileNode.resolve_class(self.provider, FileNode.FILE).get_or_create(destination['node'], self._format_path('{}{}'.format(destination['path'], file_name), file_id=self.file._id))
+        assert_equal(self.guid._id, file_node.get_guid()._id)
+        file_comments = Comment.find(Q('root_target', 'eq', self.guid._id))
+        assert_equal(file_comments.count(), 1)
+
+    def test_comments_move_when_file_moved_to_osfstorage(self):
```
sloria: There should be no implicit test passes. Either remove these or explicitly skip them with a skip decorator.

samanehsan:
Sure!

tests/test_comments.py
samanehsan: The purpose of this test is to override the one in the mixin, which does not apply to osfstorage files.

samanehsan: Ok!

PR #5061, opened 2016-02-24
[Feature] Make api lists esi based

cwisecarver:
This makes all lists in the v2 API ESI based. Each item in the list will become a separate request that is cached separately.

The easiest way to test this is to request an api url, e.g. `/v2/nodes/?format=jsonapi&embed=contributors` then request that same url with the additional query string parameter `esi=False` and compare the results. If they don't match there's a bug.


cwisecarver:
There's a bug I didn't catch in testing. Deploying it to staging2 let it show up. I'll resubmit when it's fixed.

PR #5063, opened 2016-02-25
[CR][Feature] Analytics migration to Keen IO

felliott:
*This is a resubmission of another PR by @GageGaskins*

# Purpose

This PR will be used as the base PR for migrating all analytics to KeenIO from Piwik.

# Side Effects

Currently, Piwik will still be primary analytics for customer facing statistics until all of KeenIO capabilities have been researched. This PR just establishes base event tracking to be used later in the migration.

Fixes: [#OSF-5763]


sloria:

website/static/js/keen.js
```diff
@@ -0,0 +1,116 @@
...
+
+    trackPageView: function(){
+        this.createOrUpdateKeenSession();
+        var returning = Boolean($.cookie('keenUserId'));
+        var pageView = {
+            'pageUrl': document.URL,
+            'keenUserId': this.getOrCreateKeenId(),
+            'sessionId': $.cookie('keenSessionId'),
+            'pageTitle': document.title,
+            'userAgent': '${keen.user_agent}',
```
sloria: @felliott Is this correct? It looks like mako syntax.

sloria: ...or maybe an ES6 template string? But this file will not be compiled with Babel, so template strings should not be used.

felliott: The code's correct.  That's keen's syntax for [dyanmic placeholders](https://keen.io/docs/api/#dynamic-placeholders).

sloria: Thanks!

PR #5065, opened 2016-02-25
Add get_absolute_html_url method to base serializer and encourage usage

cwisecarver:
This fixes the html api urls pointing to the api instead of osf.


cwisecarver:
@sloria This fixes the HTML api links.

PR #5066, opened 2016-02-25
[Ready to Review] Feature/500 error when accessing log detail

pattisdr:
# Purpose

Issue https://openscience.atlassian.net/browse/OSF-5777

Accessing log detail `/v2/logs/{log_id}/` for a project's logs gives a 500.  

This issue blocks https://openscience.atlassian.net/browse/OSF-5471

# Changes

Instead of using ContributorOrPublic method on log views (which was built for nodes), created a ContributorOrPublicForLogs method which then calls ContributorOrPublic for each of the nodes on the log's backref to see if the user has permission to access the log.

(Will be modified slightly when backrefs are eliminated between nodes and nodelogs).

Tests were already written for these endpoints and were now failing locally before this change.  Are these running on Travis?

### DIFF
--- a/api/logs/permissions.py
+++ b/api/logs/permissions.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from rest_framework import permissions
+
+from website.models import Node, NodeLog
+
+from api.nodes.permissions import ContributorOrPublic
+from api.base.utils import get_object_or_error
+
+
+class ContributorOrPublicForLogs(permissions.BasePermission):
+
+    def has_object_permission(self, request, view, obj):
+        assert isinstance(obj, (NodeLog)), 'obj must be a NodeLog, got {}'.format(obj)
+
+        for node_id in obj._backrefs['logged']['node']['logs']:
+            node = get_object_or_error(Node, node_id, display_name='node')
+            if ContributorOrPublic().has_object_permission(request, view, node):
+                return True
+        return False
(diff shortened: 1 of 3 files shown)


sloria:
Change looks good. @pattisdr Can you please add a regression test?

pattisdr:
@sloria, where do regression tests go?  There are already unittests for this.  (When I originally wrote unittests, they were passing. When I pulled from develop, they were failing. This change makes the unittests pass again.)

sloria:
Oh, ok. So there are tests for the log detail endpoint that fail without this change? I didn't see those. There are other tests that were failing on develop, but those were unrelated.

pattisdr:
Yes, these tests were failing before this change:
https://github.com/CenterForOpenScience/osf.io/blob/develop/api_tests/logs/views/test_log_detail.py

I am still confused about it, because I wrote these tests when I did log permissions, and they were originally passing.  When I pulled from develop, they were failing and this change made them pass again.

pattisdr:
Okay, so now I can't duplicate this ^.  I will try to write another test that fails before this change.

pattisdr:
Thanks for helping me figure out a way to test this @sloria.

Added test that tacks `?format=api` onto the end of the log detail request.   The test will fail prior to this fix.  Test needed because the standard api tests were not duplicating the issue getting seeing in the browser.

api/logs/permissions.py
```diff
@@ -0,0 +1,25 @@
...
+
+        if obj._backrefs.get('logged'):
+            for node_id in obj._backrefs['logged']['node']['logs']:
+                node = get_object_or_error(Node, node_id, display_name='node')
+                if ContributorOrPublic().has_object_permission(request, view, node):
+                    return True
+
+        if getattr(obj, 'node'):
+            if ContributorOrPublic().has_object_permission(request, view, obj.node):
+                return True
```
sloria: Minor: you can simplify to

```python
return ContributorOrPublic().has_object_permission(request, view, obj.node)
```

I can make this change  when I merge.

PR #5033, opened 2016-02-15
[Feature][Update] Project mailing lists [OSF-3975]

mfraezz:
Purpose
=======
Allow users to email `<node_guid>@osf.io` to email all subscribed contributors.

Changes
=======

- [X] Models (`Node` updated, `MLEventLog` added, new subscription type)
- [X] Mailing List logging implemented 
- [X] Script added (node.mailing_enabled migration)
- [X] View functions / routes for ML configuration added
- [X] Tests added
- [X] ML Rejection template added
- [x] ML template for sent mail added
- [x] UI Updated to match wireframes
- [x] Manually tested with ngrok

Side Effects
=========
* `catchall()` route on Mailgun for `.*@osf\.io` may need tweaking -- lower priority
* when guid length is changed (e.g. from 5 to 6), Mailgun route will need to be updated accordingly

Notes
--------
1 - Create Mailgun route to catch any email to `.{5}@osf\.io`, use MG UI.

2 - To migrate `mailing_enabled` for all existing nodes, run
```
python -m scripts.migrate_project_mailing_lists
```

3 - Email templates have not been tested for UI correctness. Sendgrid requires domain verification to send emails `From: "<x@domain>"` , so the `SENDGRID_API_KEY`'s account must have verified `osf.io` and `staging.osf.io`. @sloria, please check this during CR.

h/t @XTech2K for doing the [exploratory work](https://github.com/CenterForOpenScience/osf.io/pull/3934) for this

[#OSF-3975]

PR #5069, opened 2016-02-25
add scripts/* to .coveragerc

whit537:
The `scripts` directory is a Python module, it turns out. It should be possible to run coverage reports against it (e.g., https://github.com/CenterForOpenScience/osf.io/pull/5033/files#r54173051).

cf. #4965


sloria:
Thanks!

PR #5072, opened 2016-02-26
[feature] 'Live File' Links for Archived Files

caseyrollins:
##### Purpose
- Currently, when viewing a file that has been archived in a registration, there is no quick/easy way for a user to see the original version of the archived file. This PR generates a URL when an archived OSFS file is viewed which points to the current live file, provided that it has not been deleted. 

##### Changes
- This PR adds the `archived_from_id` field to the `StoredFileNode` model, which is then set when a file archived. 
- This PR also removes the "This project is a registration of this project/component" page alert ***if*** there is an archived file link alert to be shown, in order to prevent cluttering the page (shown below).
 screen shot 2016-03-16 at 3 39 20 pm

##### Notes
- An archived live file link will ***only*** be generated for OSFS files that are archived *after* this PR is merged. Links will not be generated for previously archived files, or for add-on files, as we cannot guarantee that we are linking to the correct live file (we can only make an educated guess based on comparing file names). 

##### JIRA
- [OSF-5160](https://openscience.atlassian.net/browse/OSF-5160)


website/templates/project/project_header.mako
```diff
@@ -106,8 +106,16 @@
...
-
+                % if not file_name:
+                    <div class="alert alert-info">This ${node['node_type']} is a registration of <a class="link-solid" href="${node['registered_from_url']}">this ${node['node_type']}</a>; the content of the ${node['node_type']} has been frozen and cannot be edited.</div>
+                % else:
+                    % if urls.get('archived_from'):
+                        <div class="alert alert-info">This file is part of a registration and is being shown in its archived version (and cannot be altered).
+                            The <a class="link-solid" href="${urls['archived_from']}">active file</a> is viewable from within the <a class="link-solid" href="${node['registered_from_url']}">live ${node['node_type']}</a>.</div>
+                    % else:
+                        <div class="alert alert-info">This ${node['node_type']} is a registration of <a class="link-solid" href="${node['registered_from_url']}">this ${node['node_type']}</a>; the content of the ${node['node_type']} has been frozen and cannot be edited.</div>
```
samanehsan: Since this is the same as above, is there any way to modify this logic so that you don't have to repeat this?

tests/test_addons.py
```diff
@@ -837,6 +838,28 @@ def test_delete_action_for_folder_deletes_subfolders_and_creates_trashed_file_no
...
+    def test_archived_from_url_without_id(self):
+        file_node = self.get_test_file()
+        registered_node = self.project.register_node(
+            schema=get_default_metaschema(),
+            auth=Auth(self.user),
+            data=None,
+        )
+        archived_from_url = views.get_archived_from_url(registered_node, file_node)
+        assert_false(archived_from_url)
+
```
samanehsan: Can you also add a test to check that if a file node is in `TrashedFileNode`, it's archived_from_url is None?

tests/test_addons.py
```diff
@@ -837,6 +838,28 @@ def test_delete_action_for_folder_deletes_subfolders_and_creates_trashed_file_no
...
+    def test_archived_from_url_with_id(self):
+        file_node = self.get_test_file()
+        file_node.archived_from_id = '12345'
+        registered_node = self.project.register_node(
+            schema=get_default_metaschema(),
+            auth=Auth(self.user),
+            data=None,
+        )
+        archived_from_url = views.get_archived_from_url(registered_node, file_node)
+        assert_true(archived_from_url)
```
samanehsan: Can you also assert that the `archived_from_url` is equal to `self.project.registered_from.web_url_for('addon_view_or_download_file', provider=file_node.provider, path=file_node.archived_from_id)`

website/files/utils.py
```diff
@@ -13,6 +13,7 @@ def copy_files(src, target_node, parent=None, name=None):
     cloned.parent = parent
     cloned.node = target_node
     cloned.name = name or cloned.name
+    cloned.archived_from_id = src._id
```
samanehsan: Is `cloned` always an OSF storage file?

samanehsan:
Pass complete @caseyrollins! :sunny:

website/files/utils.py
caseyrollins: Yes! This function is called when OSFS files are archived/copied -- there are similar functions in Waterbutler that are called for add-on files.

caseyrollins:
@samanehsan, I've added those tests and addressed your comments! I also changed the attribute `archived_from_id` to be named `copied_from_id` since it is set when a file is copied.

whit537:
Taking a look ...

tests/test_addons.py
samanehsan: Can you also assert that the `archived_from_url` is what you expect it to be in this test?

samanehsan:
Thanks for making those changes @caseyrollins! Just one more addition to a test & this is good to merge :tada: :sunny:

tests/test_addons.py
samanehsan: Oh actually...how is this test different from `test_archived_from_url`?

caseyrollins: Ah, you're right. The `views.get_archived_from_url` check should just be added to the `test_archived_from_url` test.

caseyrollins:
@samanehsan Done!

website/files/models/base.py
```diff
@@ -137,6 +139,8 @@ class StoredFileNode(StoredObject):
     node = fields.ForeignField('Node', required=True)
     parent = fields.ForeignField('StoredFileNode', default=None)
 
+    copied_from_id = fields.StringField(default=None)
```
sloria: Why store only the ID, rather than making this a ForeignField?

sloria:
Pass finished. Just one question above. Otherwise, this looks good.

caseyrollins:
@sloria, I changed ```copied_from``` to be a ```ForeignField```, so this could use another round of review.

website/files/models/base.py
```diff
@@ -423,6 +422,18 @@ def parent(self, val):
         self.stored_object.parent = val
 
     @property
+    def copied_from(self):
```
sloria: Why are these properties necessary?

sloria:
@caseyrollins +1 to making copied_from a `ForeignField`, but I'm unsure why the properties are necessary. Can you please clarify?

caseyrollins:
@sloria -- discussed in person -- properties in the ```FileNode``` class are necessary to safely access the ```copied_from``` field of the ```StoredFileNode```.

PR #5075, opened 2016-02-26
[WIP] feature/Add "Read More" link to wiki preview

erinspace:
Relates to https://openscience.atlassian.net/browse/OSF-5450

# Purpose 
Previously, when a wiki was too long to be rendered, the text would be truncated and had a '...' appended to the end. This PR refactors this process a bit, making it smoother and taking out some unnecessary  python HTML parsing before the text is rendered fully in javascript.

This PR adds a "Read More" link to wikis that have additonal content on the home page, OR wikis that  have additional content on pages other than the home page.

A "Read More" link is not present when there is additional content in a component's wiki - this was to avoid blank parent wikis with a confusing "Read More" link when their components had wiki content.

# Changes
- wiki_widget.mako:
    - Remove "pre-rendered" wiki html
        - WIki content was being rendered once, then the whole page jumped as javascript appropriately re-rendered wiki content with truncating and appropriate markdown. Removing this in the template prevents this jumping, and shows the content after it is done loading.
    - Add new div with the "Read More" link.
        - I added a new div instead of placing it in the ```markdownRender``` div to avoid the ReadMore link being replaced when the content is re-rendered by js in ```project-dashboard-page.js```
- wiki/views.py:
     - Set the ```more``` variable up front if there are more wiki pages on the current component
    - Remove unnecessary BeautifulSoup html parsing done in the view

# Side effects
- The wiki preview won't be rendered if javascript is disabled - since javascript is doing the rendering of the text

### DIFF
--- a/website/addons/wiki/templates/wiki_widget.mako
+++ b/website/addons/wiki/templates/wiki_widget.mako
@@ -9,6 +9,12 @@
     % endif
 </div>
 
+<div id="more_link">
+    % if more:
+        <a href="${node['url']}${short_name}/">Read More</a>
+    % endif
+</div>
+
 <% import json %>
 <script>
     window.contextVars = $.extend(true, {}, window.contextVars, {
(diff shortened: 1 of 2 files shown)


erinspace:
Reopened in #5093

PR #5076, opened 2016-02-26
Add loading indicator to contributor search

barbour-em:
## Purpose
Before, the user would see 'No results found' even if the search wasn't over. They should know that contributors are still being searched.

## Changes

Add some extra logic to show 'Searching contributors...'

Had to stop it in the debugger to pull off a screenshot which is why it's dark:
screen shot 2016-02-26 at 3 49 33 pm

## Side effects

It occasionally will "flash" locally (where I have about 4 users to search through) but it's not too bad.

[OSF-5654]


sloria:
Ideally, the loading indicator would only show after a few hundred milliseconds to avoid the "flash". This is fine as is for now, though. Thanks for the quick turnaround.

PR #5077, opened 2016-02-26
OSF-4846 Add new field to logFeed.js and change templates for log to …

adlius:
<h3>
Purpose
</h3>
To dynamically change the log template if the registration is cancelled.

<h3>
Changes
</h3>
In logFeed.js, a new field "registrationCancelled" is added so that the template could tell whether a certain registration is cancelled.
In log_templates.mako, knockout conditionals are added so that the template would render with only the node title, instead of a link, when the registration is cancelled.


chennan47:
Pass done. Look fine.

sloria:
I've cherry-picked this into master so that we can get this to prod ASAP.

PR #5083, opened 2016-02-29
fix the resend emails not returning emails

chennan47:
<b>Purpose</b>
Fix resend confirmation for unconfirmed emails don't return emails list. Solves https://openscience.atlassian.net/browse/OSF-5779#add-comment.


sloria:
Build failure is unrelated to this change.

PR #5085, opened 2016-02-29
[RFR] Check to see if there are multiple contributors to remove

billyhunt:
Related to #4799

This fixes a bug found by QA where if there are multiple nodes but the contributor is not on those nodes, then the multiple remove radio button is displayed.  It should not be.

Steps to Reproduce
---------------------------

1.  have a project with components
2.  add a contributor on the project and not on any of the children
3.  try and remove her

Before Fix
----------------------

screen shot 2016-02-29 at 10 02 33 am

After Fix
----------

you see this dialog

screen shot 2016-02-29 at 10 00 50 am

### DIFF
--- a/website/static/js/contribRemover.js
+++ b/website/static/js/contribRemover.js
@@ -128,7 +128,7 @@ var RemoveContributorViewModel = oop.extend(Paginator, {
 
         self.hasChildrenToRemove = ko.computed(function() {
             //if there is more then one node to remove, then show a simplified page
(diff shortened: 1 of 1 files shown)


billyhunt:
@sloria, @nicipfeiffer  found a bug with removing contributors.  If there are multiple child nodes but a given contributor is not on those nodes, it should not offer the ability to remove the user from the child nodes.  Here is the fix.  If you could get it in develop, that would be the :bomb:

PR #5086, opened 2016-02-29
Add checks needed for new version of dateutil

barbour-em:
The new version of python-dateutil doesn't let you pass an empty string or `None` to the parser without it throwing an error so check for that.

@sloria

### DIFF
--- a/website/files/models/base.py
+++ b/website/files/models/base.py
@@ -574,7 +574,7 @@ class File(FileNode):
         # Transform here so it can be sortted on later
         if data['modified'] is not None and data['modified'] != '':
             data['modified'] = parse_date(
-                data['modified'] or None,
+                data['modified'],
                 ignoretz=True,
                 default=datetime.datetime.utcnow()  # Just incase nothing can be parsed
             )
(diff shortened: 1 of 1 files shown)


website/files/models/base.py
```diff
@@ -737,7 +738,7 @@ def update_metadata(self, metadata, save=True):
         # If its are not in this callback it'll be in the next
         self.size = self.metadata.get('size', self.size)
         self.content_type = self.metadata.get('contentType', self.content_type)
-        if self.metadata.get('modified') is not None:
+        if self.metadata.get('modified') is not None and self.metadata.get('modified') != '':
```
sloria: Minor: it's safer and more concise to check the truthiness of `self.metadata.get('modified')`. I can make this change when I merge.

sloria:
Thanks for the quick response on this. This is good to merge when the build passes.

sloria:
I've squashed these changes (to make them easier to cherry-pick into master) and resolved the remaining issues with the build in https://github.com/CenterForOpenScience/osf.io/commit/381c85189f3fa6d42570d8e6c0101049be4bd1ae

PR #5089, opened 2016-02-29
[feature] Fix Retractions Metadata

caseyrollins:
- Fixed incorrect ```node.is_retracted``` check to correctly display retracted registration metadata.
- [OSF-5765]


mfraezz:
LGTM. Passing this to RTM. :star: 

Side note: Ticket says `date_retracted` should be display as well, but as that's not currently being shown it's out of scope of this bug fix. Follow up ticket will be created.

sloria:
Cherry-picked f5d451ae3595ecd02bc1286c6dedd93ec97216ec into release/0.65.0

PR #5092, opened 2016-02-29
Fix project description box in project settings page 

kch8qx:
so that text area can't be resized outside of parent div

[OSF-5103]


caseyrollins:
LGTM. Moving on to RTM. :star:

PR #5093, opened 2016-02-29
Feature/read more link wiki

erinspace:
Relates to https://openscience.atlassian.net/browse/OSF-5450

# Purpose 
Previously, when a wiki was too long to be rendered, the text would be truncated and had a '...' appended to the end. 

This PR adds a "Read More" link to wikis that have additonal content on the home page, OR wikis that  have additional content on pages other than the home page.

A "Read More" link is not present when there is additional content in a component's wiki - this was to avoid blank parent wikis with a confusing "Read More" link when their components had wiki content.

# Changes
- wiki_widget.mako:
    - Add new div with the "Read More" link.
        - I added a new div instead of placing it in the ```markdownRender``` div to avoid the ReadMore link being replaced when the content is re-rendered by js in ```project-dashboard-page.js```
- wiki/views.py:
     - Set the ```more``` variable up front if there are more wiki pages on the current component
     - update cutoff variable from 500 to 400 to match later JS rendering of truncated text

### DIFF
--- a/website/addons/wiki/views.py
+++ b/website/addons/wiki/views.py
@@ -144,8 +144,8 @@ def wiki_widget(**kwargs):
     use_python_render = False
     if wiki_page and wiki_page.html(node):
         wiki_html = wiki_page.html(node)
-        if len(wiki_html) > 500:
-            wiki_html = BeautifulSoup(wiki_html[:500] + '...', 'html.parser')
+        if len(wiki_html) > 400:
+            wiki_html = BeautifulSoup(wiki_html[:400] + '...', 'html.parser')
             more = True
         else:
             wiki_html = BeautifulSoup(wiki_html)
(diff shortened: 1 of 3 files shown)


mfraezz:
Changes look great, no bugs found :star2: 

Please add a test in `website/addons/wiki/tests/test_wiki.py:TestWikiViews` to make sure that `more` is set correctly when there are multiple pages shorter than the max length (There are already two testing `more` is correct for single page lengths).

erinspace:
Good call @mfraezz! Test added and moved back to CR.

website/addons/wiki/views.py
```diff
@@ -140,16 +140,15 @@ def wiki_widget(**kwargs):
     wiki = node.get_addon('wiki')
     wiki_page = node.get_wiki_page('home')
 
-    more = False
+    more = True if len(node.wiki_pages_current.keys()) >= 2 else False
```
sloria: Be wary of ternary expressions that look like

```python
my_cool_variable = True if some_boolean else False
```

because it can be more simply expressed as

```python
my_cool_variable = some_boolean
```

sloria: I've addressed this in 6f9a53c

PR #5102, opened 2016-03-02
[Feature] Admin - Remove Contributor from Node

Ghalko:
## Proposal

Allow the admin to remove the contributor from a node.

## Repercussions
No way to add them back easily, without cooperation from other users?

## Based on
#5013 Does not include previous thoughts on branches.


admin/common_auth/models.py
```diff
@@ -47,6 +47,10 @@ class MyUser(AbstractBaseUser, PermissionsMixin):
     date_joined = models.DateTimeField(default=datetime.now)
     confirmed = models.BooleanField(default=False)
     osf_id = models.CharField(default=False, max_length=10, blank=True)
+    desk_email = models.EmailField(verbose_name='desk email', max_length=255,
+                                   default='')
+    desk_password = models.CharField(verbose_name='desk password',
+                                     max_length=128, default='')  # TODO: encrypt
```
mfraezz: This shouldn't be a `TODO`, passwords *must* be encrypted before storing them.

admin/nodes/views.py
```diff
@@ -1,17 +1,60 @@
...
+
+from admin.base.views import GuidFormView, GuidView
 from admin.nodes.templatetags.node_extras import reverse_node
 from .serializers import serialize_node
 
 
+def remove_contributor(request, node_id, user_id):
+    user = User.load(user_id)
+    node = Node.load(node_id)
+    node.remove_contributor(user, None)  # TODO: logging not good.
```
mfraezz: Pass in `log=False` here

admin/common_auth/models.py
mfraezz: Though this (`desk`) code is completely unrelated to this feature. Why is it in the PR?

Ghalko: The Desk PR is the furthest along, and was up-to-date as of yesterday, broken again...

As to the password, this basically needs to send the plaintext password to Desk.com, or I just can't find documentation as to what to send instead. There is the possibility of using Oauth, thoughts @mfraezz ?

mfraezz: Use [OAuth](http://dev.desk.com/guides/oauth/#what) if possible. We should never store plaintext passwords

mfraezz: After discussion, it may be best to just the user-specific access keys provided by desk, and keep the application-specific oauth keys in a `/settings/local.py` file.

mfraezz:
Other than the logging issue, the feature-specific code in this PR looks good. (There are other issues with Desk, but I'll put those in the Desk PR) :mount_fuji: 

Please also add a test to make sure the most recent log isn't for removing a contributor after you fix the logging issue. Back to you. @Ghalko

Ghalko:
@mfraezz I am going to take the desk stuff out and base this off of #5013 and bring in registrations rather than slowing this PR down with desk stuff. Bad initial decision.

Ghalko:
Made the commit fiasco much better @mfraezz and the last admin can't be removed.

admin/nodes/urls.py
```diff
@@ -3,5 +3,10 @@
 from . import views
 
 urlpatterns = [
-    url(r'^$', views.NodeFormView.as_view(), name='node'),
+    url(r'^$', views.NodeFormView.as_view(),
+        name='search'),
+    url(r'^id-(?P<guid>[a-z0-9]+)/$', views.NodeView.as_view(),
+        name='node'),
+    url(r'^id-(?P<node_id>[a-z0-9]+)/remove_user-(?P<user_id>[a-z0-9]+)/$',
```
mfraezz: Is there any reason to include `id-` in these routes?

Ghalko: It has become pretty standard. These can be removed, but I would rather remove them all on one PR rather than one of the 13- or 14 open ones. @mfraezz @sloria

sloria:
Merged into release/0.66.0. :+1:

PR #5103, opened 2016-03-02
Fix: Timestamp in OSF4M Date Created in UTC time

barbour-em:
## Purpose 

Make the Date Created column in OSF4M consistent with the Date Created on the project page.

## Changes

In some cases, calling format on the `moment(this.date)` would cause the date to be converted to UTC then formatted. This was because of `moment(...).utc()` vs `moment.utc(...)`. 

screen shot 2016-03-02 at 11 41 40 am

## Side effects

Looks like local and UTC dates elsewhere are still working as expected. 

Chrome, FFox, and Safari all display the same format. 

[OSF-5788]

### DIFF
--- a/website/static/js/osfHelpers.js
+++ b/website/static/js/osfHelpers.js
@@ -480,7 +480,7 @@ var UTC_DATEFORMAT = 'YYYY-MM-DD HH:mm UTC';
 var FormattableDate = function(date) {
 
     if (typeof date === 'string') {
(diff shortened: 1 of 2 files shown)


sloria:
This was fixed in https://github.com/CenterForOpenScience/osf.io/commit/ee66276198ee7a20c0a951ec19780ea7cd5ecefc ; sorry for the duplicated work.

PR #5104, opened 2016-03-02
Feature: Create trashed addon file nodes

samanehsan:
Purpose
-----------
When a folder is moved/renamed, the guids of all of the files are updated with the new StoredFileNode objects. However, there is no way to know whether the original StoredFileNode corresponds to a deleted addon file unless you make a request to waterbutler.

Changes
------------
- When a file is deleted, remove its StoredFileNode and create a TrashedFileNode.
- When resolving a file, if it no longer exists remove the StoredFileNode and create a TrashedFileNode (handles the case where a file is deleted through the addon)

### DIFF
--- a/website/addons/base/views.py
+++ b/website/addons/base/views.py
@@ -392,10 +392,10 @@ def addon_delete_file_node(self, node, event_type, payload, user=None):
         materialized_path = payload['metadata']['materialized']
         if path.endswith('/'):
             folder_children = FileNode.resolve_class(provider, FileNode.FILE).find(
-                                Q('provider', 'eq', provider) &
-                                Q('node', 'eq', node) &
-                                Q('materialized_path', 'startswith', materialized_path))
-
+                Q('provider', 'eq', provider) &
+                Q('node', 'eq', node) &
+                Q('materialized_path', 'startswith', materialized_path)
+            )
             for item in folder_children:
                 if item.kind == 'file' and not TrashedFileNode.load(item._id):
                     item.delete()
--- a/website/addons/base/views.py
+++ b/website/addons/base/views.py
@@ -24,8 +24,7 @@ from framework.transactions.handlers import no_auto_transaction
 from framework.auth.decorators import must_be_logged_in, must_be_signed, collect_auth
 from website import mails
 from website import settings
-from website.files.models import FileNode
-from website.files.models import TrashedFileNode
+from website.files.models import FileNode, TrashedFileNode, StoredFileNode
 from website.project import decorators
 from website.addons.base import exceptions
 from website.addons.base import signals as file_signals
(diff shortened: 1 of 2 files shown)


website/addons/base/views.py
```diff
@@ -384,6 +384,31 @@ def create_waterbutler_log(payload, **kwargs):
...
+                Q('provider', 'eq', provider) &
+                Q('node', 'eq', node) &
+                Q('materialized_path', 'startswith', materialized_path)
+            )
+            for item in folder_children:
+                if item.kind == 'file' and not TrashedFileNode.load(item._id):
+                    item.delete()
+        else:
+            try:
+                file_node = FileNode.resolve_class(provider, FileNode.FILE).find_one(Q('node', 'eq', node) & Q('materialized_path', 'eq', materialized_path))
```
felliott: Minor: please line-wrap this

website/addons/base/views.py
```diff
@@ -384,6 +384,31 @@ def create_waterbutler_log(payload, **kwargs):
     return {'status': 'success'}
 
 
+@file_signals.file_updated.connect
+def addon_delete_file_node(self, node, event_type, payload, user=None):
+    if event_type == 'file_removed' and payload.get('provider', None) != 'osfstorage':
```
felliott: Can you add a docstring explaining why it's important to trash these instead of just deleting them?

website/addons/base/views.py
```diff
@@ -384,6 +384,31 @@ def create_waterbutler_log(payload, **kwargs):
...
+        provider = payload['provider']
+        path = payload['metadata']['path']
+        materialized_path = payload['metadata']['materialized']
+        if path.endswith('/'):
+            folder_children = FileNode.resolve_class(provider, FileNode.FILE).find(
+                Q('provider', 'eq', provider) &
+                Q('node', 'eq', node) &
+                Q('materialized_path', 'startswith', materialized_path)
```
felliott: Does this work for Dropbox with the prepended folder?

felliott: Yes it does.  The storage root is in the path, not the materialized_path, @felliott, you dummy.

samanehsan: Yup! I used `materialized_path` so I would not have to special-case that for dropbox

website/addons/base/views.py
```diff
@@ -384,6 +384,31 @@ def create_waterbutler_log(payload, **kwargs):
...
+            )
+            for item in folder_children:
+                if item.kind == 'file' and not TrashedFileNode.load(item._id):
+                    item.delete()
```
felliott: While you're trashing the files here, can you also hard-delete the folder StoredFileNodes?

felliott:
Looks good.  Just minor fixes needed.

website/addons/base/views.py
samanehsan: Sure! Btw, do you know where we create StoredFileNodes for addon folders? I was trying to test it out.

felliott:
:+1:

PR #5106, opened 2016-03-02
[WIP] Feature/admin inst

hmoco:


website/project/model.py
```diff
@@ -3479,6 +3419,204 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+
+    def __init__(self, node):
+        if node is None:
+            return
+        self.node = node
+        for key, value in self.institution_node_translator.iteritems():
+            setattr(self, key, getattr(node, value))
+
+    def __getattr__(self, item):
+        return getattr(self.node, item)
```
caseyrollins: If ```self``` is an ```Institution``` then wouldn't ```getattr(self, '_id')``` incorrectly return ```self.node._id``` instead of ```self.node.institution_id```?

hmoco: correct. But `__getattr__` does not get called unless the attribute that is being requested is not current an attribute on the instance, which `_id` should be as `__init__` assigns everything on the transform dictionary

website/project/model.py
```diff
@@ -3479,6 +3419,204 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+        for each in super(InstitutionQuerySet, self).__iter__():
+            yield Institution(each)
+
+    def _do_getitem(self, index):
+        item = super(InstitutionQuerySet, self)._do_getitem(index)
+        if isinstance(item, MongoQuerySet):
+            return self.__class__(item)
+        return Institution(item)
+
+class Institution():
```
caseyrollins: Documentation for this class and how it relates to the ```Node``` class would be awesome :tada:

website/project/model.py
```diff
@@ -3479,6 +3419,204 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+
+    institution_node_translator = {
```
caseyrollins: *Possible* name alternatives: ```node_to_institution_map```, ```node_to_institution```

caseyrollins: `'description'` should be included here so its value can be updated when `save()` is called.

website/project/model.py
```diff
@@ -3479,6 +3419,204 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+    _affiliated_institutions = fields.ForeignField('node', list=True)
+
+    @property
+    def affiliated_institutions(self):
+        return SpecialList(self, '_affiliated_institutions', [Institution(node) for node in self._affiliated_institutions])
+
+    def add_primary_institution(self, user, inst, log=True):
+        if not user.is_affiliated_with_institution(inst):
+            raise UserNotAffiliatedError('User is not affiliated with {}'.format(inst.name))
+        if inst.node == self.primary_institution:
```
caseyrollins: If `self.primary_institution` returns an [`Institution`](https://github.com/hmoco/osf.io/blob/feature/admin-inst/website/project/model.py#L3452), isn't this incorrectly comparing an `Institution` with a `Node`?

website/project/model.py
```diff
@@ -3479,6 +3419,204 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+            return False
+        previous = Institution(self.primary_institution) if self.primary_institution else None
```
caseyrollins: Similar to above, isn't `self.primary_institution` already an `Institution`?

website/project/model.py
```diff
@@ -3479,6 +3419,204 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+        self.primary_institution = inst
+        if inst.node not in self.affiliated_institutions:
```
caseyrollins: Similar again, should this be `if inst not in self.affiliated_institutions` since `self.affiliated_institutions` is a list of `Institutions`?

website/project/model.py
```diff
@@ -3479,6 +3419,204 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+        return True
+
+    def remove_primary_institution(self, user, log=True):
+        inst = self.primary_institution
+        if not inst:
+            return False
+        self.primary_institution = None
+        if inst in self.affiliated_institutions:
+            self.affiliated_institutions.remove(inst)
+        inst = Institution(inst)
```
caseyrollins: `inst` is already an `Institution` so this cast shouldn't be necessary.

website/project/model.py
```diff
@@ -3479,6 +3419,206 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+            return
+        self.node = node
+        for key, value in self.attribute_map.iteritems():
+            setattr(self, key, getattr(node, value))
+
+    def __getattr__(self, item):
+        return getattr(self.node, item)
+
+    def __eq__(self, other):
+        if not other:
```
caseyrollins: It would also be good to have a check like `isinstance(other, Institution)` here.

website/project/model.py
```diff
@@ -3479,6 +3419,206 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+    # Primary institution node is attached to
+    _primary_institution = fields.ForeignField('node')
+
+    @property
+    def primary_institution(self):
+        return Institution(self._primary_institution) if self._primary_institution else None
+
+    @primary_institution.setter
+    def primary_institution(self, institution):
+        self._primary_institution = institution.node if institution else None
```
caseyrollins: `isinstance(institution, Institution)` would be good here too.

website/project/model.py
```diff
@@ -3479,6 +3419,206 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+            )
+        return True
+
+    def institution_url(self):
+        return self.absolute_api_v2_url + 'institution/'
+
+    def institution_relationship_url(self):
+        return self.absolute_api_v2_url + 'relationships/institution/'
+
+class SpecialList(list):
```
caseyrollins: Should this class have a specialized `remove` function too?

framework/auth/core.py
```diff
@@ -1376,13 +1380,22 @@ def is_affiliated_with_institution(self, inst):
         return inst in self.affiliated_institutions
 
     def remove_institution(self, inst_id):
-        try:
-            self.affiliated_institutions.remove(inst_id)
-        except ValueError:
-            return False
-        return True
+        node_id = False
+        for node in self._affiliated_institutions:
```
caseyrollins: Is there a reason you're iterating through the protected `_affiliated_institution` instead of `affiliated_institutions`?

caseyrollins:
General first pass done :turtle: 

Only overall comment is to add tests throughout that would catch the errors you found earlier (confirming that `self.primary_institution` always returns an `Institution`, `self._primary_institution` returns a `Node`, `self.attributed_institutions` returns a `SpecialList` of `Institutions`, etc.).

PR #5110, opened 2016-03-02
run scripts/tests/ at Travis

whit537:
I'm not sure if you want this @sloria (depends on how you manage the `scripts` directory, what the lifespan and workflow is for scripts), but over at https://github.com/pattisdr/osf.io/pull/2#issuecomment-191476364 I noticed that we're not actually running `scripts/tests/` at Travis. From the looks of it (and as one might expect), this test suite is in a broken state:

```
============================== 24 failed, 57 passed, 15 error in 142.83 seconds ===============================
```

### DIFF
--- a/tasks/__init__.py
+++ b/tasks/__init__.py
@@ -464,6 +464,11 @@ def test_api():
     test_module(module="api_tests/")
 
 @task
+def test_scripts():
+    """Run the test suite for scripts."""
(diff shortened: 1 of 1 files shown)


sloria:
Other than the scripts that run on cron jobs, the scripts are one-offs that target older versions of the database, so the tests are expected to fail.

PR #5115, opened 2016-03-03
[RFR] Redirect to project instead of component when creating new component

billyhunt:
Jira Ticket:  https://openscience.atlassian.net/browse/OSF-5814

After creating a component, the OSF takes you into that component.  Usually this is not the preferred action, and leaves users confused.  Now users will stay on the project page they were originally working on.

### DIFF
--- a/website/project/views/node.py
+++ b/website/project/views/node.py
@@ -202,7 +202,7 @@ def project_new_node(auth, node, **kwargs):
                 http.BAD_REQUEST,
                 data=dict(message_long=e.message)
             )
-        redirect_url = new_component.url
+        redirect_url = node.url
         message = (
             'Your component was created successfully. You can keep working on the component page below, '
             'or return to the <u><a href="{url}">project page</a></u>.'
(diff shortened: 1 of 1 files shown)


caseyrollins:
LGTM. Just need the build to pass before moving to RTM. :money_with_wings:

website/project/views/node.py
```diff
@@ -202,10 +202,11 @@ def project_new_node(auth, node, **kwargs):
...
             )
-        redirect_url = new_component.url
+        redirect_url = node.url
         message = (
-            'Your component was created successfully. You can keep working on the component page below, '
-            'or return to the <u><a href="{url}">project page</a></u>.'
+            'Your component was created successfully. You can keep working on the project page below, '
+            'or go to the new <u><a href=' + new_component.url +
+            '>component</a></u>.'
         ).format(url=node.url)
```
mfraezz: String formatting is preferred over concatenation, please fix this string/call.

PR #5117, opened 2016-03-03
[WIP] Update notification defaults to "instantly" on node creation

erinspace:
https://openscience.atlassian.net/browse/OSF-4685

# Purpose
When a user creates a project, notification settings are automatically set to "none," and a user might not know notifications even exist. This PR adds a method that automatically adds a node's creator to subscriptions for files and comments so they know the option exists.

# Changes
- Add method that subscribes a nodes' creator to notifications
- Call that method with signals when a node is first saved
- Adjust many tests that were expecting nodes were not configured with notifications.
    - some tests remove notifications and then build them back up as necessary
    - some tests utilize the now-existing notifications set up on node creation or contributor adding

# Side effects and Notes
- the tearDown method for addon tests was removed entirely because of troubles removing specific elements when contributors were added and nodes created mid test . I'm not sure if this was the right - thing to do or if there will be reprocussions, but related tests seem to pass
- Node settings for current projects are NOT touched


tests/test_events.py
```diff
@@ -188,6 +194,11 @@ def setUp(self):
         self.user = factories.UserFactory()
         self.consolidate_auth = Auth(user=self.user)
         self.project = factories.ProjectFactory()
+
+        # Remove default subscriptions
+        for sub in NotificationSubscription.find():
```
mfraezz: Rather than having to do this a bunch of times for different tests, you could change the `setUpClass` [in the base class](https://github.com/CenterForOpenScience/osf.io/blob/develop/tests/base.py#L117), possibly by adding a setting that would disable the creation of these `NotificationSubscriptions` when a `Node` is created. That way you could just re-enable it during `setUp` of notification-type test classes and not have to alter so many different test files.

mfraezz:
According to Jira ticket:
> [All] notification emails must be double checked for their inclusion of the explanation of where to change the user's preferences. 

This should apply to `contributor_added`, `comments`, `comment_replies`, `invite`, and perhaps `new_public_project` in `website/templates/emails/`. Maybe others, as well.

tests/test_events.py
erinspace: Omg @mfraezz you are a genius. This is so much better. 

Thank you!!

erinspace:
Closing this and reopening in the much simpler #5377

PR #5118, opened 2016-03-03
[Feature] Admin Remove Node

Ghalko:
## Proposal

Remove node, potentially spam nodes or nodes by users leaving.

## Repercussions

Does not use Auth, so does not use the remove_node for Node. Certainly hides the node though.


admin_tests/nodes/test_serializers.py
```diff
@@ -7,7 +7,6 @@
 
 
 class TestNodeSerializers(AdminTestCase):
-
     def test_serialize_node(self):
```
mfraezz: Can you update this test (and the `simple_node` test in `users/test_serializers.py`) to include a check for `'deleted'`?

admin/templates/nodes/node.html
```diff
@@ -8,120 +8,135 @@
...
+                    </div>
+                    <div class="modal-body">
+                        This action will be reversible after the fact.
+                    </div>
+                    <div class="modal-footer">
+                        <a href="{% url 'nodes:remove_node' node.id %}"
+                           class="btn btn-danger">
+                            Delete
+                        </a>
+                        <button class="btn btn-default">Cancel</button>
```
mfraezz: This button needs a `data-dismiss="modal"`

mfraezz:
Looks good, @Ghalko. A couple tiny changes and this will be RTM :snake:

admin_tests/nodes/test_serializers.py
Ghalko: Done

admin/nodes/urls.py
```diff
@@ -3,5 +3,12 @@
 from . import views
 
 urlpatterns = [
-    url(r'^$', views.NodeFormView.as_view(), name='node'),
+    url(r'^$', views.NodeFormView.as_view(),
+        name='search'),
+    url(r'^id-(?P<guid>[a-z0-9]+)/$', views.NodeView.as_view(),
+        name='node'),
+    url(r'^id-(?P<guid>[a-z0-9]+)/remove_node/$', views.remove_node,
```
sloria: This can go in a follow-up PR, but you should refactor this so that nodes are removed with a `DELETE` request to `/<guid>/` rather than adding this extra `remove_node` route.

admin/templates/spam/user.html
```diff
@@ -0,0 +1,85 @@
...
+    <title>User Spam</title>
+{% endblock title %}
+{% block content %}
+    <h2>User's List of Spam</h2>
+    <a href="{{ user_id | reverse_user }}" class="btn btn-primary">
+        <i class="fa fa-arrow-left"></i>
+        User details
+    </a>
+    <div class="btn-group">
+    {% if status == '1' %}
```
sloria: Can you pass the constants to the template so that you can do `if status == FLAGGED`, etc.?

sloria:
Went ahead and merged, as there are no major blockers. https://github.com/CenterForOpenScience/osf.io/pull/5118#discussion_r56000044 and 
https://github.com/CenterForOpenScience/osf.io/pull/5118#discussion_r56000604 should still be addressed, but that can happen in a follow-up PR.

admin/nodes/urls.py
Ghalko: Django doesn't like `DELETE` requests per se. Please refer to ADM-82 for more information.

PR #5119, opened 2016-03-03
Dashboard Redesign with Home page elements

caneruguz:
## Purpose
A combined home page, support page and navigation changes to finish off the Stage 1. 

## Changes
Home page is now a layered content based on the wireframes. It includes several sections
- Quick Search
- Recent activity widgets
- New and Noteworthy Items
- Meetings banner

There is a new support page, but getting started is still it's own page with some edits

Navigation has changed based on wireframe

more details coming... 
Tag @pattisdr for Quick search and new and noteworthy, tag @hmoco for Recent activity widgets, tag me for anything else.


caneruguz:
@sloria I started this to check for Travis messages. I will do some more cleanup of it tonight and tomorrow. Feel free to get going with your review.

caneruguz:
Ah I need to merge in prereg fix here. I'll do that tonight!

sloria:
Just a few drive-by comments before I leave for lunch:

- I have many projects, but they aren't loading

osf___home

- In the above screenshot, the "You have no projects" is misaligned
- Can you document how to get the "Discover Public Projects" working?

caneruguz:
@pattisdr Can you respond to this? Thanks!
> Can you document how to get the "Discover Public Projects" working?

pattisdr:
In @sloria's screenshot, @caneruguz has added a Create New Project onto the home page that wasn't there when I said 'Go here to create one.'  We probably don't need to redirect them to the My Projects page, they can create it from here.

Steve, is your api server running?

caneruguz:
Also @sloria I do sometimes see that the content somehow just doesn't load. We need to improve our error reporting but this might be something to look into generally. It feels like the page load kind of gives up.

pattisdr:
To get Discover Public Projects working: 

1) Go to website/settings/local.py and define a POPULAR_LINKS_NODE and a NEW_AND_NOTEWORTHY_LINKS_NODE. Both nodes should be different and public.

2)  Add node links to both POPULAR_LINKS_NODE and NEW_AND_NOTEWORTHY_LINKS_NODE `POST /v2/nodes/{node_id}/node_links/`The UI will display up to five of each.

bower.json
```diff
@@ -19,7 +19,7 @@
     "locales": "https://github.com/citation-style-language/locales.git#d2b612f8a6f764cbd66e67238636fac3888a7736",
     "raven-js": "1.1.17",
     "zeroclipboard": "2.1.6",
-    "osf-style": "https://github.com/CenterForOpenScience/osf-style.git#30057b759f4feaf8faff2478b751ef190a6ae3b1"
-
+    "osf-style": "https://github.com/CenterForOpenScience/osf-style.git#5dcf5997ad2808ffc7c8564f94cde102bf7a4a96",
+    "jQuery-ui-Slider-Pips": "jquery-ui-slider-pips#~1.11.1"
```
sloria: This looks strange; shouldn't it be:

```javascript
"jquery-ui-slider-pips": "~1.11.1"
```

?

sloria:
I got a number of timed-out requests. There also appears to be a problem loading category names:


osf___home

website/static/js/recentActivityWidget.js
```diff
@@ -0,0 +1,347 @@
...
+            ctx.stroke();
+            ctx.beginPath();
+            ctx.moveTo(rightHandle.offsetLeft + (handle.width()/2), 0);
+            ctx.lineTo(progBar.offset().left + progBar[0].offsetWidth - canvasElement.offset().left, self.canvasHeight);
+            ctx.strokeStyle = '#E0E0E0 ';
+            ctx.lineWidth = 2;
+            ctx.stroke();
+        };
+        self.categoryColor = function(category){
+            if (category.indexOf('wiki') !== -1){ return '#d9534f'; }
```
sloria: You can refactor to use an object or a switch statement here.

website/static/js/recentActivityWidget.js
```diff
@@ -0,0 +1,347 @@
...
+            if (category.indexOf('comment') !== -1){ return '#5bc0de'; }
+            if (category.indexOf('file') !== -1){ return '#337ab7'; }
+            if (category.indexOf('project') !== -1){ return '#f0ad4e'; }
+            else { return '#5cb85c'; }
+        };
+        self.addButtons = function(ele, isInitialized) {
+            if (self.select('#rALeftButton')){
+                self.select('#rALeftButton').css('height', self.select('#rALogs').height());
```
sloria: Why so much jQuery here if you're already using mithril to render the view?

sloria:
I'm getting an error from the recent activity widget:

osf___home

website/static/js/recentActivityWidget.js
hmoco: Unless I am incorrect, switch statements expect an exact string match, which this does not.

website/settings/defaults.py
```diff
@@ -228,7 +228,13 @@ def parent_dir(path):
 ALL_MY_REGISTRATIONS_ID = '-amr'
 ALL_MY_PROJECTS_NAME = 'All my projects'
 ALL_MY_REGISTRATIONS_NAME = 'All my registrations'
-NOTEWORTHY_LINKS_NODE = None
+
+# Most Popular and New and Noteworthy Nodes
+POPULAR_LINKS_NODE = '6824t'  # TODO Override in local.py in production.
```
sloria: These should not have hardcoded values like this. Instead, the defaults should be `None` and you should do one of two things:

- Raise an error if this isn't set, causing the app to fail early. This can be done here: https://github.com/CenterForOpenScience/osf.io/blob/a5250be2dadabeec7551e255a75081fda8632798/website/settings/__init__.py#L20
- Treat these as a feature flag and hide the relevant widgets if these are set to `None`.

website/settings/defaults.py
```diff
@@ -228,7 +228,13 @@ def parent_dir(path):
...
+NEW_AND_NOTEWORTHY_LINKS_NODE = '9gyxb'  # TODO Override in local.py in production.
+
+QA_USER_IDS = []  # TODO Override in local.py in production.
+URL_BASE = None  # TODO Override with LOCAL, STAGING, STAGING2, or PRODUCTION in local.py.
```
sloria: Why is this necessary given that we already have `DOMAIN`?

scripts/populate_new_and_noteworthy_projects.py
```diff
@@ -0,0 +1,137 @@
...
+
+logger = logging.getLogger(__name__)
+
+def retrieve_data(url):
+    """ Fetch data and decode json """
+    response = requests.get(url)
+    data = response.json()
+    return data
+
+def get_api_base_path():
```
sloria: Again, why do we need this given that we `settings.DOMAIN` and `settings.API_DOMAIN`? This code is fragile and hard to maintain; we would have to modify these functions every time we make a new deployment (e.g. when we create staging3).

sloria: Also, we probably shouldn't allow requesting data across deployments. For example, local developers shouldn't be requesting data from production or staging--they should only work with their local data.

sloria:
Hm, so my projects finally loaded somehow.

sloria:
The "Create New Project" is misaligned.

osf___home

website/settings/defaults.py
caneruguz: Sorry hard coded values are my bad! They have crept in in one of my commits.

sloria:
The timeout errors seem to be probabilistic.

caneruguz:
Yes, we should zoom in on that problem at some point, let's get some code quality issues out first though.

sloria:
The images can still be optimized quite a bit. I ran bg6.jpg through ImageOptim and got it from 67Kb to 26Kb.

api/users/views.py
```diff
@@ -425,12 +425,85 @@ def get_default_odm_query(self):
...
+
+    ordering = ('-date', )
+
+    def get_default_odm_query(self):
+        user = self.get_user()
+        if user.is_anonymous():
+            auth = Auth(None)
+        else:
+            auth = Auth(user)
+        raw_nodes = list(Node.find(
```
sloria: Avoid casting querysets to list when it isn't necessary.

api/users/views.py
```diff
@@ -425,12 +425,85 @@ def get_default_odm_query(self):
...
+    pagination_class = UserLogPagination
+    view_category = 'users'
+    view_name = 'user-logs'
+
+    ordering = ('-date', )
+
+    def get_default_odm_query(self):
+        user = self.get_user()
+        if user.is_anonymous():
+            auth = Auth(None)
```
sloria: I believe we have a `get_user_auth()` function for this purpose.

api/users/views.py
```diff
@@ -425,12 +425,85 @@ def get_default_odm_query(self):
...
+            Q('is_deleted', 'ne', True)
+        ))
+        nodes = [each for each in raw_nodes if each.is_public or each.can_view(auth)]
+        if not nodes:
+            raise NotFound
+        query = [Q('params.node', 'eq', node._id) for node in nodes]
+        return reduce(lambda x, y: x | y, query)
+
+    def get_queryset(self):
+        self.paginator.node_log_aggregates = self.get_aggregates
```
sloria: `get_aggregates` calls `get_query_from_request`, then `get_query_from_request` gets call in the line after this, which seems inefficient, esp. because there's a lot of querying and filtering going on in `get_default_odm_query`. 

Can you refactor this to save a call to `get_query_from_request`? Maybe use `functools.partial`:


```python
def get_aggregates(self, query):
    #...

def get_queryset(self):
    query = self.get_queryset_from_request()
    self.paginator.node_log_aggregates = partial(self.get_aggregates, query)
```

scripts/populate_new_and_noteworthy_projects.py
```diff
@@ -0,0 +1,137 @@
...
+        base = 'http://127.0.0.1:8000/'
+    return base
+
+def get_popular_nodes():
+    """ Fetch data from url that returns dict with a list of popular nodes from piwik """
+    discover_url = get_api_base_path() + 'api/v1/explore/activity/popular/raw/'
+    return retrieve_data(discover_url)
+
+def get_new_and_noteworthy_nodes():
+    """ Fetches nodes created in the last month and returns 25 sorted by highest log activity """
```
sloria: Why use HTTP for this if you can just use the database?

scripts/populate_new_and_noteworthy_projects.py
```diff
@@ -0,0 +1,137 @@
...
+def is_eligible_node(node):
+    """
+    Check to ensure that node is not the POPULAR or NEW_AND_NOTEWORTHY LINKS_NODE.
+    Ensures QA members are not contributors
+    """
+    if node._id == POPULAR_LINKS_NODE or node._id == NEW_AND_NOTEWORTHY_LINKS_NODE:
+        return False
+
+    for contrib in node.contributors:
+        if contrib._id in QA_USER_IDS:
```
sloria: We might want to use a more generic name for this setting; we might not only exclude QA users. How about something like `NEW_AND_NOTEWORTHY_CONTRIBUTOR_BLACKLIST`?

sloria:
I'm consistently getting request timeouts for this URL: `http://localhost:8000/v2/users/me/logs/?embed=nodes&embed=user&embed=linked_node&embed=template_node&page=1&page%5Bsize%5D=20&aggregates=1`. This looks a very slow endpoint--lots of embedding and [expensive aggregating](https://github.com/CenterForOpenScience/osf.io/blob/a5250be2dadabeec7551e255a75081fda8632798/api/users/views.py#L505-L506). @hmoco @pattisdr Can you look into optimizing this? Removing embeds is low-hanging fruit. Any missing indices? Any expensive filtering going on with Python?

scripts/populate_new_and_noteworthy_projects.py
pattisdr: Good point.

website/static/js/recentActivityWidget.js
sloria: That's right; use an object then.

scripts/populate_new_and_noteworthy_projects.py
```diff
@@ -0,0 +1,112 @@
...
+
+def retrieve_data(url):
+    """ Fetch data and decode json """
+    response = requests.get(url)
+    data = response.json()
+    return data
+
+def get_popular_nodes():
+    """ Fetch data from url that returns dict with a list of popular nodes from piwik """
+    discover_url = DOMAIN + 'api/v1/explore/activity/popular/raw/'
```
sloria: You should be able to use the database rather than HTTP here. Can you reuse the code for the view for this URL?

scripts/populate_new_and_noteworthy_projects.py
```diff
@@ -0,0 +1,112 @@
...
+
+    return True
+
+def update_node_links(designated_node, target_nodes, description):
+    """ Takes designated node, removes current node links and replaces them with node links to target nodes """
+    logger.info('Repopulating {} with latest {} nodes.'.format(designated_node._id, description))
+    user = designated_node.creator
+    auth = Auth(user)
+
+    for i in xrange(len(designated_node.nodes)-1, -1, -1):
```
sloria: To loop through the nodes in reverse, you can do:

```python
for node in reversed(designated_node.nodes):
```

sloria: But, out of curiosity, why is it necessary to go in reverse?

scripts/populate_new_and_noteworthy_projects.py
```diff
@@ -0,0 +1,112 @@
...
+        return False
+
+    for contrib in node.contributors:
+        if contrib._id in NEW_AND_NOTEWORTHY_CONTRIBUTOR_BLACKLIST:
+            logger.info('Node {} skipped because a contributor, {}, is blacklisted.'.format(node._id, contrib._id))
+            return False
+
+    return True
+
+def update_node_links(designated_node, target_nodes, description):
```
sloria: `target_nodes` is a bit of a misnomer because it's actually a list of IDs. It might make more sense to pass the actual nodes in.

scripts/populate_new_and_noteworthy_projects.py
```diff
@@ -0,0 +1,112 @@
...
+    user = designated_node.creator
+    auth = Auth(user)
+
+    for i in xrange(len(designated_node.nodes)-1, -1, -1):
+        pointer = designated_node.nodes[i]
+        if isinstance(pointer, Pointer):
+            designated_node.rm_pointer(pointer, auth)
+
+    for n_id in target_nodes:
+        n = models.Node.find(Q('_id', 'eq', n_id))[0]
```
sloria: The proper way to load an object by ID is `Node.load(n_id)`.

scripts/populate_new_and_noteworthy_projects.py
pattisdr: We store the most popular projects?  I thought you had to fetch them from piwik.

scripts/populate_new_and_noteworthy_projects.py
pattisdr: I am deleting things - if I'm not going in reverse, it is telling me index out of range.

scripts/populate_new_and_noteworthy_projects.py
sloria: Ah, looks like we do store popular projects in piwik. In that case, you should still be able to call the ` popular_activity_json` function directly rather than sending a request to APIv1.

scripts/populate_new_and_noteworthy_projects.py
sloria: Oh I see; `you are modifying `designated_nodes` in place as you loop through it.

It might be cleaner to copy `designated_node.nodes_pointer`, then loop through that. Something like:

```python
# Copy nodes_pointer because we'll be removing pointers from designated_node.nodes
pointers = designated_node.nodes_pointer[:]
for pointer in pointers:
    designated_node.rm_pointer(pointer, auth)
```

sloria: Actually, copying might not even be necessary; it looks like `.nodes_pointer` returns a new list that won't get modfied by `rm_pointer`.

```python
for pointer in designated_node.nodes_pointer:
```

scripts/populate_new_and_noteworthy_projects.py
```diff
@@ -0,0 +1,111 @@
...
+    return retrieve_data(discover_url)
+
+def get_new_and_noteworthy_nodes():
+    """ Fetches nodes created in the last month and returns 25 sorted by highest log activity """
+    today = datetime.datetime.now()
+    last_month = (today - dateutil.relativedelta.relativedelta(months=1))
+    data = db.node.find({'date_created': {'$gt': last_month}, 'is_public': True, 'is_registration': False})
+    node_log_count_mapping = {}
+    for new_node in data:
+        node_log_count_mapping[new_node['_id']] = len(new_node['logs'])
```
sloria: Minor: you can refactor to use a dict comprehension:

```python
node_log_counts = {node['_id']: len(node['logs']) for node in data}
```

sloria: Even better: you don't even needs this intermediate dictionary at all.

```python
return sorted(data, key=lambda node: len(node['logs']), reverse=True)[:25]
```

scripts/populate_new_and_noteworthy_projects.py
```diff
@@ -0,0 +1,111 @@
...
+
+    return True
+
+def update_node_links(designated_node, target_node_ids, description):
+    """ Takes designated node, removes current node links and replaces them with node links to target nodes """
+    logger.info('Repopulating {} with latest {} nodes.'.format(designated_node._id, description))
+    user = designated_node.creator
+    auth = Auth(user)
+
+    for pointer in reversed(designated_node.nodes):
```
sloria: I think my previous comment got lost in the diff, but you can simplify this further by using `nodes_pointer`

```python
for pointer in designated_node.nodes_pointer:
    designated_node.rm_pointer(pointer, auth)
```

website/routes.py
```diff
@@ -175,14 +175,19 @@ def make_url_map(app):
 
     process_rules(app, [
 
-        Rule('/dashboard/', 'get', website_views.dashboard, OsfWebRenderer('dashboard.mako')),
+        Rule('/dashboard/', 'get', website_views.redirect_to_myprojects, OsfWebRenderer('dashboard.mako')),
+        Rule('/myprojects/', 'get', website_views.dashboard, OsfWebRenderer('dashboard.mako')),
```
sloria: Why is this route added?

website/routes.py
```diff
@@ -175,14 +175,19 @@ def make_url_map(app):
 
     process_rules(app, [
 
-        Rule('/dashboard/', 'get', website_views.dashboard, OsfWebRenderer('dashboard.mako')),
+        Rule('/dashboard/', 'get', website_views.redirect_to_myprojects, OsfWebRenderer('dashboard.mako')),
```
sloria: The redirect and the `/myprojects/` route are unnecessary. `/dashboard/` should go directly to the dashboard.

website/routes.py
```diff
@@ -349,6 +354,11 @@ def make_url_map(app):
         ),
     ])
 
+    process_rules(app, [
+        Rule('/explore/activity/popular/raw/', 'get', discovery_views.popular_activity_json, json_renderer),
```
sloria: This route is unnecessary; as suggested in a previous comment, the new_and_noteworthy script can just make requests to piwik directly; no need to go through APIv1.

website/static/js/pages/faq-page.js
```diff
@@ -0,0 +1,141 @@
...
+            return;
+        }
+        $('.clear-search').addClass('clear-active');
+        if (text.length < 2) {
+            return;
+        }
+        var el;
+        var content;
+        $('.support-item').each(function(){
+            el = $(this);
```
sloria: Minor: as per COSDev, prefix jQuery selector variables with `$`, e.g. 

```js
var $el = $(this);
// or
var $this = $(this);
```

website/static/js/pages/faq-page.js
```diff
@@ -0,0 +1,141 @@
...
+        }, 500);
+    }
+
+
+    /**
+     * Affix or restore search layer based on location
+     */
+    function fixSearchLayer () {
+        var topOffset = $(window).scrollTop();
+        var searchLayer = $('.search-layer');
```
sloria: Minor: As per COSDev, prefix jquery elements with `$`.

website/static/js/pages/faq-page.js
```diff
@@ -0,0 +1,141 @@
...
+     * @param item {Object} jQuery object for the .support-item element
+     * @param turnOff {Boolean} true if we are closing and open item
+     * @private
+     */
+    function changeExpandState (item, turnOff) {
+        var body = item.children('.support-body');
+        var head = item.children('.support-head');
+        var icon = head.children('.fa');
+        if (turnOff){
+            body.slideUp();
```
sloria: Minor: the default animation times for these slides are a bit slow. Might be worth speeding these up a bit.

sloria:
On first load of the home page, the user sees nothing.

osf___home

At the very least, a loading indicator or loading text should be shown. An even better solution would be to show some content, e.g. a server-side rendered list of cached user projects to show while the request for all the user's projects is pending.

website/static/js/pages/home-page.js
```diff
@@ -1,216 +1,110 @@
...
+/**
+ * Initialization code for the home page.
+ */
+
+'use strict';
+
+var $ = require('jquery');
+var m = require('mithril');
+
+var quickSearchProject = require('js/quickProjectSearchPlugin');
```
sloria: Minor: keep the casing of components consistent--everywhere else, we use UpperCamelCase.

website/static/js/pages/home-page.js
```diff
@@ -1,216 +1,110 @@
...
+var newAndNoteworthy = require('js/newAndNoteworthyPlugin');
+var meetingsAndConferences = require('js/meetingsAndConferencesPlugin');
```
sloria: Minor: to make things more cohesive, it might make sense to put all these home page components in their own folder, e.g. `js/home-page/quickSearch`, `js/home-page/newAndNoteworthy`.

sloria:
There also doesn't appear to be proper error handling; after I got a JS error, I just see a blank page, like in my previous screenshot.

sloria:
@caneruguz Is this an appropriate loading indicator?

loading

It seemed strange to me to see "Loading projects" even though the project list was already rendered. I thought it was a bug, but I then realized that the loading indicator was for the request to load all my projects for searching. 

Is there a better solution to this? Maybe show a disabled search box input that has "Loading projects for search" as its placeholder text?

sloria:
osf___home

- Text is misaligned
- Icon and exclamation point are unnecessary

sloria:
@caneruguz I'm not sure it's intuitive to have the "x" button be the right-most button for the search bar.

osf___home

The most common pattern is to have search button:

bleep_bloop_-_google_search


bleepbloop_at_duckduckgo

youtube

search_ _fwoeifwjef

--------

...though there are examples of search bars with no button:

centerforopenscience_osf_io__facilitating_open_science


I think it's OK to omit the search button, but it's certainly strange to have the "x" when the user hasn't input anything.

website/static/js/quickProjectSearchPlugin.js
```diff
@@ -0,0 +1,478 @@
...
+                    m('span', {onclick: function() {
+                        ctrl.filter('');
+                        document.getElementById('searchQuery').value = '';
+                        ctrl.quickSearch();
+                    }},  m('button', m('i.fa.fa-times')))
+                ]);
+            }
+        }
+
+        function displayNodes() {
```
sloria: Rather than have these private functions for returning part of the view, it is better to separate these out into separate, stateless subcomponents.

For example:

```js
var NodeList = {
    view: function(ctrl, args) {
        if (args.eligibleNodes().length === 0 && args.filter() != null) {
            return m('.row.m-v-sm', m('.col-sm-10.col-sm-offset-1',
                m('.row',
                    m('.col-sm-12', [m('p.fa.fa-exclamation-triangle'), m('em', 'No results found!')])
                ))
            );
        }
        else {
            return args.eligibleNodes().slice(0, args.countDisplayed()).map(function(n){
                var project = args.nodes()[n];
                var numContributors = project.embeds.contributors.links.meta.total;
                var onClickNode = args.onClickNode.bind(project);
                return m('.m-v-sm.node-styling', {onclick: onClickNode}, m('.row',
                    [
                        // If you want, you could break out single node items into its own component, e.g.
                         // m.component(NodeListItem, {data: project})
                    ]
                ));
            });
        }

    }
}

// ..
m.component(NodeList, {
    nodes: ctrl.nodes(), 
    filter: ctrl.filter(),
    countDisplayed: ctrl.countDisplayed(),
    eligibleNodes: ctrl.eligibleNodes(),
    onClickNode: function(node) {
        // redirect to node
    }
})
```

Spend an hour refactoring in this way. If it takes too much time, feel free to defer, but be thinking about how you can break this code up into smaller components. 

This makes for more efficient (because you're not redeclaring these private functions), readable (separating concerns), testable (small components are easy to test), and reusable code.

website/templates/home.mako
```diff
@@ -1,5 +1,30 @@
...
 <%def name="content()">
-    <h1>Placeholder</h1>
+
+    <div id="osfHome"></div>
+
+</%def>
+
+<%def name="stylesheets()">
+  <link rel="stylesheet" href="/static/css/pages/home-page.css">
+  <link rel="stylesheet" href="//code.jquery.com/ui/1.11.4/themes/smoothness/jquery-ui.css">
```
sloria: What is this stylesheet needed for?

website/templates/home.mako
```diff
@@ -1,5 +1,30 @@
...
+  <link rel="stylesheet" href="/static/vendor/bower_components/jQuery-ui-Slider-Pips/dist/jquery-ui-slider-pips.css">
+  <link rel="stylesheet" href="/static/css/recent-activity-widget.css">
```
sloria: @hmoco Remove this `link` and the CSS file.

website/templates/home.mako
```diff
@@ -1,5 +1,30 @@
...
-    <h1>Placeholder</h1>
+
+    <div id="osfHome"></div>
+
+</%def>
+
+<%def name="stylesheets()">
+  <link rel="stylesheet" href="/static/css/pages/home-page.css">
+  <link rel="stylesheet" href="//code.jquery.com/ui/1.11.4/themes/smoothness/jquery-ui.css">
+  <link rel="stylesheet" href="/static/vendor/bower_components/jQuery-ui-Slider-Pips/dist/jquery-ui-slider-pips.css">
```
sloria: @hmoco Remove this, as it is no longer used. Also remove the library from bower.json

website/templates/prereg_landing_page.mako
```diff
@@ -23,16 +23,10 @@
 
 <%def name="existingPrereg(size=None)">
 <% size = size or '' %>
-<div id="existingPrereg${size}" class="prereg-exisitng-prereg p-md osf-box box-round clearfix m-b-lg" style="display:none">
+<div id="existingPrereg${size}" class="prereg-exisitng-prereg p-md osf-box box-round clearfix m-b-lg" style="display:none; style="width: 100%;"">
```
sloria: This may be outside the scope of this PR, but this usage of a mako variable in a CSS ID is smelly. Relying on the suffix of a CSS ID is a bad idea.

Also, there's a typo in `prereg-existing-prereg`.

website/views.py
```diff
@@ -260,3 +260,11 @@ def redirect_about(**kwargs):
 
 def redirect_howosfworks(**kwargs):
     return redirect('/getting-started/')
+
+def redirect_to_support():
+    # Redirect to support page
+    return redirect('/support/')
+
+def redirect_to_myprojects():
```
sloria: As suggested in a previous comment, this view is unnnecessary.

website/views.py
```diff
@@ -260,3 +260,11 @@ def redirect_about(**kwargs):
 
 def redirect_howosfworks(**kwargs):
     return redirect('/getting-started/')
+
+def redirect_to_support():
```
sloria: This doesn't seem to be used; remove?

sloria:
Pass finished.

pattisdr:
> I think it's OK to omit the search button, but it's certainly strange to have the "x" when the user hasn't input anything.

If we're making this change, the x is in the search bar in the project organizer as well.

website/static/js/quickProjectSearchPlugin.js
pattisdr: @sloria, what do I do about 
```
m('.row',
       [m('.col-sm-4.col-md-5.p-v-xs', m('.quick-search-col', project.attributes.title)),
        m('.col-sm-4.col-md-4.p-v-xs', m('.quick-search-col', $osf.contribNameFormat(project, numContributors, ctrl.getFamilyName))),
        m('.col-sm-4.col-md-3.p-v-xs', m('.quick-search-col', ctrl.formatDate(project)))]
));
```

This was inside the `displayNodes` function. Specifically, `ctrl.formatDate(project)`.    args.formatDate(), but what about the project?

sloria: `project` is within the scope of that loop, so 

```js
m('.row',
       [
      //... 
       m('.col-sm-4.col-md-3.p-v-xs', m('.quick-search-col', args.formatDate(project)))]
));
```

sloria: For another example, see how I wrote `onClickNode` above.

sloria:
One more drive-by comment:

We should be able to make these proper links (i.e. with an `a` tag):

osf___home

This allows the user to hover over the links to see the href, right click and "Copy Link address", etc.

hmoco:
Recent activity widget should be completely removed now, including its unique dependencies and the ` /v2/users/<id>/logs ` endpoint.

website/static/js/quickProjectSearchPlugin.js
pattisdr: oh, right! binding.

pattisdr:
> One more drive-by comment:
We should be able to make these proper links (i.e. with an a tag):

This makes the onhover color the same as the link.  Should the link go white onhover as well?   Then it's not as obvious it can be clicked.
screen shot 2016-03-08 at 6 32 51 pm

website/static/js/quickProjectSearchPlugin.js
pattisdr: I've been stuck on a mithril error for awhile:
`Uncaught Error: Component template must return a virtual element, not an array, string, etc.`

caneruguz:
@pattisdr For links on dark backgrounds you can use this color: #5BC0DE

pattisdr:
#5BC0DE looks nice!

caneruguz:
@pattisdr Can you also take a look at the travis fails when you get a chance?

pattisdr:
OK - question - since I did some of the home page but you did the combination - it is not clear for all of the home page comments, which are my responsibility.  Some are more about the combination, like https://github.com/CenterForOpenScience/osf.io/pull/5119#issuecomment-193908382

caneruguz:
@pattisdr Good point. I'll go through the comments myself as well. You should do the comments that are about your modules only but if you want to pick up something else just add a comment here. I'm responding to QA on the My Projects section now. 

I asked the Travis check as a request, if you have the time it would help if you could try to resolve those.

pattisdr:
Yes I will be looking at Travis issues today as well and trying to finish responding to Steve's review.

pattisdr:
@caneruguz - some things I noticed on My Projects - not sure if they are intended or not, or if some of this is just happening with my local data:

- When you hit clear search button for 'Search all my projects', the text does not disappear from the search bar
- When I add a collection, it doesn't immediately appear in the collections list.  After refreshing several times it's not there.  If I access the collections via the API, it's listed, but the new collection is not appearing on my dashboard.
- If you rename a collection, you have to refresh to see the new name
- When I add project to collection, and then remove it the counts don't update under collections without a refresh
- Sometimes collection counts are wrong:
screen shot 2016-03-09 at 11 07 23 am
- I think contributor counts are confusing.  I have Dawn Pattison (2) and then when I click to filter by my name, four projects come up.  So what does the 2 mean?

pattisdr:
Are spinners on dark backgrounds ok?
screen shot 2016-03-09 at 11 47 36 am

pattisdr:
@caneruguz, if I click the `load more` button on QuickSearch two times, so 30 projects are loaded, it looks like the geometric pattern in the background runs out and then you just see a light blue.

screen shot 2016-03-09 at 2 19 26 pm

website/routes.py
caneruguz: Dashboard is not anymore the dashboard though. The place it goes to is not dashboard, it's a page called My Projects. The meaning of dashboard is not present in current page. Do we have to hold on to the name?

website/templates/prereg_landing_page.mako
caneruguz: The mako variable is part of a previous change. I think changing this here would require some further code change and should probably be handled as a task of its own.

scripts/populate_new_and_noteworthy_projects.py
```diff
@@ -0,0 +1,96 @@
+"""
+This will update node links on POPULAR_LINKS_NODE and NEW_AND_NOTEWORTHY_LINKS_NODE.
+"""
+import sys
+import logging
+import datetime
+import dateutil
+import operator
+import requests
```
sloria: Minor: remove unused imports.

sloria:
Minor: These inputs could use placeholders to guide the user.

osf___home

sloria:
On the home page: this looks like a lot of requests to `/nodes/<nid>/contributors/`

1__tmux

Are you absolutely certain that these are necessary? The "My projects" widget only shows the first few contributors anyway--why do we need to request all of the contributors for each node?

sloria:
"My projects" is loading very slowly, likely due to all the contributor requests shown above. Is there a way we can show the projects incrementally? In other words, once we've loaded all the data for 10 projects, render those while the reset of the node requests are still pending.

This doesn't necessarily need to be implemented before this is merged, but I think it should be done before this is released.

sloria:
The node links are not working:

osf___home

To clarify my previous comment: the whole row should be a hyperlink (not only the name), such that I can see the URL while I'm hovering over the row.

Also, the hrefs on the name are not working: they link to `http://localhost:5000/undefined`

sloria:
osf___home

Why "advanced"? Why not just "Search for more projects"?

pattisdr:
@sloria, we are loading all the contributor names so they can be searched.

pattisdr:
There already is logic in there for loading the first ten nodes, rendering them, and loading the remaining ten in the background.

sloria:
> @sloria, we are loading all the contributor names so they can be searched.

We should not slow down the first render for filtering; better to render the widget as quickly as possible, then load contributor names in the background.

pattisdr:
> Also, the hrefs on the name are not working: they link to http://localhost:5000/undefined

Are you sure? I specifically checked that.

sloria:
osf___home

scripts/populate_new_and_noteworthy_projects.py
```diff
@@ -0,0 +1,94 @@
...
+        n = models.Node.load(n_id)
+        if is_eligible_node(n):
+            designated_node.add_pointer(n, auth, save=True)
+            logger.info('Added node link {} to {}'.format(n, designated_node))
+
+def main(dry_run=True):
+    init_app(routes=False)
+
+    popular_node_ids = popular_activity_json()['popular_node_ids']
+    popular_links_node = models.Node.find(Q('_id', 'eq', POPULAR_LINKS_NODE))[0]
```
sloria: Minor: you can use `.find_one` here, which will error if the record is not found (this is a good thing).

scripts/populate_new_and_noteworthy_projects.py
```diff
@@ -0,0 +1,94 @@
...
+    new_and_noteworthy_links_node = models.Node.find(Q('_id', 'eq', NEW_AND_NOTEWORTHY_LINKS_NODE))[0]
```
sloria: Same here: you can use `.find_one`.

website/discovery/views.py
```diff
@@ -72,3 +72,13 @@ def activity():
         'popular_public_registrations': popular_public_registrations,
         'hits': hits,
     }
+
+
+def popular_activity_json():
```
sloria: Since this isn't a view, this doesn't really belong in this file. It probably makes the most sense to just put this function in the script.

website/routes.py
```diff
@@ -191,14 +191,19 @@ def make_url_map(app):
 
     process_rules(app, [
 
-        Rule('/dashboard/', 'get', website_views.dashboard, OsfWebRenderer('dashboard.mako')),
+        Rule('/dashboard/', 'get', website_views.redirect_to_home, OsfWebRenderer('home.mako')),
+        Rule('/myprojects/', 'get', website_views.dashboard, OsfWebRenderer('dashboard.mako')),
```
sloria: Ping @JeffSpies for route review.

- `/myprojects/` -> project organizer; linked to in navbar

osf___my_projects

- `/` when logged in -> user home, containing "My projects" and "New and noteworthy nodes"; linked to in the top-left of navbar

osf___home

- `/dashboard/` -> redirects to `/` ; included for backwards compat
- `/support/` -> New support page

osf___support

@erinmayhood and I are good with these changes.

sloria:
@caneruguz I've sent a PR to optimize the images here: https://github.com/caneruguz/osf.io/pull/35 .

website/static/js/home-page/meetingsAndConferencesPlugin.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * Meetings and Conferences
+ */
+
+var m = require('mithril');
+
+// CSS
+require('css/quick-project-search-plugin.css');
```
sloria: Why is this required here?

website/static/js/home-page/newAndNoteworthyPlugin.js
```diff
@@ -0,0 +1,166 @@
...
+ * New and Noteworthy Projects
+ */
+
+var $ = require('jquery');
+var m = require('mithril');
+var $osf = require('js/osfHelpers');
+var Raven = require('raven-js');
+
+// CSS
+require('css/quick-project-search-plugin.css');
```
sloria: Why is this required here? Should this module have it's own CSS file?

website/static/js/home-page/newAndNoteworthyPlugin.js
```diff
@@ -0,0 +1,166 @@
...
+                });
+            });
+        }
+
+        function popularProjectsTemplate () {
+            return ctrl.popularNodes().map(function(node){
+                return m.component(NoteworthyNodeDisplay, {
+                    node : node,
+                    getFamilyName: ctrl.getFamilyName,
+                    addToolTip: function(item) {
```
sloria: If the tooltip-adding logic will never vary across usages of `NoteWorthyNodeDisplay`, it doesn't need to be an arg.

website/static/js/home-page/newAndNoteworthyPlugin.js
```diff
@@ -0,0 +1,166 @@
...
+            var url = nodeLink.embeds.target_node.data.relationships.contributors.links.related.href;
+            var promise = m.request({method: 'GET', url : url, config: xhrconfig});
+            promise.then(function(result){
+                var contribNames = [];
+                result.data.forEach(function (contrib){
+                    contribNames.push($osf.findContribName(contrib.embeds.users.data.attributes));
+                });
+                var numContrib = result.links.meta.total;
+                var nodeId = nodeLink.id;
+                self.contributorsMapping[nodeId] = [contribNames, numContrib];
```
sloria: Better to use an object here:

```js
    self.contributorsMapping[nodeId] = {
        names: contribNames,
        total: numContrib
    }
````

File: website/static/js/home-page/newAndNoteworthyPlugin.js [hunk 42]
```diff
@@ -0,0 +1,166 @@
...
+        self.addToolTip = function(line) {
+            var $line = $(line);
+            if (line.offsetWidth < line.scrollWidth) {
+                $line.tooltip('show');
+            }
+        };
+    },
+    view : function(ctrl) {
+        if (ctrl.errorLoading()) {
+            return m('p.text-center.m-v-lg', 'Error loading projects. Please refresh the page.');
```
sloria: For consistency with other error messages on the site, it would be good to include our support email address.

sloria:
Pass finished. Main blocker is fixing the incremental rendering. https://github.com/CenterForOpenScience/osf.io/pull/5119#issuecomment-194529704 should also be implemented, but it can go in a follow-up PR.

website/static/js/home-page/newAndNoteworthyPlugin.js
caneruguz: Yes we should separate these files even if resulting files are very small. They will need one call per page in either case. @pattisdr can you take care of this?

pattisdr: Yes.

website/static/js/home-page/newAndNoteworthyPlugin.js
```diff
@@ -0,0 +1,162 @@
...
+
+var NewAndNoteworthy = {
+    controller: function() {
+        var self = this;
+        self.newAndNoteworthyNodes = m.prop([]); // New and noteworthy nodes list
+        self.popularNodes = m.prop([]); // Most popular nodes list
+        self.contributorsMapping = {}; // Dictionary mapping node id to some of the contrib names and the contrib total.
+        self.SHOW_TOTAL = 5; // Number of new and noteworthy projects displayed in each column
+        self.errorLoading = m.prop(false);  // True if error retrieving projects or contributors.
+        self.someDataLoaded = m.prop(false); // True when the initial request to retrieve projects is complete.
```
sloria: This shouldn't need to be a prop, since it's derived data.

```js
self.someDataLoaded = function() { return Boolean(self.newAndNoteworthyNodes().length); };
```

website/static/js/home-page/quickProjectSearchPlugin.js
```diff
@@ -0,0 +1,496 @@
...
+        self.sortState = m.prop('dateDesc'); //How nodes are sorted - default is date descending - dateDesc
+        self.countDisplayed = m.prop(); // Max number of nodes that can be rendered.  'Load more' increases this by up to ten.
+        self.next = m.prop(); // URL for getting the next ten user nodes. When null, all nodes are loaded.
+        self.loadingComplete = m.prop(false); // True when all user nodes are loaded.
+        self.contributorMapping = {}; // Maps node id to list of contributors for searching
+        self.filter = m.prop(); // Search query from user
+        self.fieldSort = m.prop(); // For xs screen, either alpha or date
+        self.directionSort = m.prop(); // For xs screen, either Asc or Desc
+        self.errorLoading = m.prop(false);  // True if error retrieving projects or contributors.
+        self.someDataLoaded = m.prop(false); // True when the initial request to retrieve projects is complete.
```
sloria: See https://github.com/CenterForOpenScience/osf.io/pull/5119/files#r55742143

sloria:
Discussed in person: https://github.com/CenterForOpenScience/osf.io/pull/5119#issuecomment-194529704 is already addressed, so no action needed there.

whit537:
Go team! Great to see this moving closer to production. :-)

PR #5120, opened 2016-03-03
[WIP] Remove status bar from contributors section of project page

billyhunt:
Jira Ticket:  https://openscience.atlassian.net/browse/OSF-5821

Responding to community request to remove the status bars from the components section of the project and component page.  Apparently, no one uses it and it takes up real estate.

Before
---------
screen shot 2016-03-03 at 5 18 55 pm

After
------
screen shot 2016-03-03 at 5 19 32 pm


caseyrollins:
@billyhunt, code looks good to me :star: 

Can you confirm with community that removing the status bar from all pages (i.e. the rendered node list of a project's registrations as well) is alright?

Also, if that is the case, can you remove the dead code associated with the status bar on the back end? It looks to me like the ```_get_user_activity()``` function in ```node.py``` would no longer be necessary, as well as the ```ua```, ```non_ua```, and ```ua_count``` variables, as they aren't used anywhere else.

billyhunt:
I removed the variables and got the :+1:  from community.  We should be good to go @caseyrollins 

:skull: :shaved_ice: :100: :evergreen_tree:

caseyrollins:
@billyhunt, looks like the ```_get_user_activity()``` function can be removed as well!

billyhunt:
Whoops.  Accidently put it back in.  Got rid of that pesky critter this time.  

:eyeglasses: :tokyo_tower: :neckbeard: :rabbit:

caseyrollins:
Moving over to RTM :dancer:

sloria:
I still see code related to these bars in the code: https://github.com/CenterForOpenScience/osf.io/blob/3e35eff339170011d6201095983ffb7753ad6679/website/templates/util/render_nodes.mako#L9 . Can you take a second look to make sure there aren't any other remnants? Let's get rid of these for good

billyhunt:
OK.  Removed rescale_ratio from everywhere and am always including the logs so the number of contributions can be canceled.  I also put in an if statement so if there is one contribution, it says "1 contribution" and not "1 contributions"

sloria:
Much better. :octopus:

PR #5122, opened 2016-03-04
Hide terms when submitting form

hmoco:


sloria:
@hmoco Is there a JIRA ticket explaining what this is supposed to fix?

PR #5123, opened 2016-03-04
[RfR] Add sign up button to sign in page

billyhunt:
Jira ticket: https://openscience.atlassian.net/browse/OSF-5823

Users should be able to sign up for the OSF directly from anywhere on the OSF.  Putting a sign up button at the top makes sense for us, and it makes sense for America.   

Also, I created a campaign so when the user clicks sign up, they are directed to the login page.  the cool thing is that the "Create a free account" portion of the login screen is active from the button, not "Login" which is the usual default.  (good idea @hmoco)  

Before
--------
screen shot 2016-03-04 at 1 59 23 pm

After
-------

screen shot 2016-03-09 at 10 04 24 am

screen shot 2016-03-09 at 10 05 41 am


billyhunt:
@erinmayhood  and @caneruguz have signed off on the change.  Should be ready to look at.

sloria:
When I click "Sign up", the "email" field of the Login form is focused. 

osf___sign_in

The "Full name" field should be the focused field.

sloria:
I don't really see the value in using the "campaign" code here. Why not just a regular query parameter?

```
/login/?signup=1
```

You can read this query parameter in the [auth_login](https://github.com/CenterForOpenScience/osf.io/blob/6ccd9126c397556b1117682285cbfe862d718824/framework/auth/views.py#L112) view and pass that to the template.

sloria:
Pass finished.

billyhunt:
Made the changes @sloria, :microphone: check it!

website/templates/public/login.mako
```diff
@@ -60,6 +64,16 @@
             <div class="form-group">
                 <label for="inputEmail3" class="col-sm-3 control-label">Email</label>
                 <div class="col-sm-9">
+                    %if sign_up:
+                    <input
```
sloria: Rather than duplicate this input, you should be able to do:

```mako
                    <input
                        type="email"
                        class="form-control"
                        data-bind="value: username"
                        name="username"
                        id="inputEmail3"
                        placeholder="Email"
                        ${'autofocus' if not sign_up else ''}
                    />
```

website/templates/public/login.mako
```diff
@@ -111,7 +130,21 @@
                 >
                     <label for="inputName" class="col-sm-4 control-label">Full Name</label>
                     <div class="col-sm-8">
+                        %if sign_up:
                         <input
+                            autofocus
+                            type="text"
```
sloria: See https://github.com/CenterForOpenScience/osf.io/pull/5123/files#r55689175

framework/auth/views.py
```diff
@@ -116,6 +116,7 @@ def auth_login(auth, **kwargs):
     """
     campaign = request.args.get('campaign')
     next_url = request.args.get('next')
+    sign_up = request.args.get('sign_up')
```
sloria: This can be removed.

sloria:
I can make the final minor changes when I merge.

PR #5124, opened 2016-03-04
Only update last_login when not at an API url

chrisseto:


sloria:
Looks good. Will merge when the build passes.

sloria:
The 2 failing tests are unrelated to this change and have since been fixed on develop.

sloria:
Also cherry-picked this fix into master.

PR #5126, opened 2016-03-06
[hotfix] standardize middleware

icereval:
Standardize usage of framework middleware in both Flake and Django web frameworks


icereval:
wrong branch

PR #5127, opened 2016-03-06
[hotfix] standardize middleware

icereval:
Standardize usage of framework middleware in both Flask and Django web frameworks

### DIFF
--- a/api/base/middleware.py
+++ b/api/base/middleware.py
@@ -1,70 +1,63 @@
-from pymongo.errors import OperationFailure
 from raven.contrib.django.raven_compat.models import sentry_exception_handler
 
+from framework.mongo.handlers import (
+    connection_before_request,
+    connection_teardown_request
+)
 from framework.tasks.handlers import (
     celery_before_request,
     celery_teardown_request
 )
-from framework.transactions import commands, messages, utils
+from framework.transactions.handlers import (
+    transaction_before_request,
+    transaction_after_request,
+    transaction_teardown_request
+)
 
 from .api_globals import api_globals
 
-from api.base import settings
+
+class MongoConnectionMiddleware(object):
+    """MongoDB Connection middleware."""
+
+    def process_request(self, request):
+        connection_before_request()
+
+    def process_response(self, request, response):
+        connection_teardown_request()
+        return response
+
+class CeleryTaskMiddleware(object):
+    """Celery Task middleware."""
+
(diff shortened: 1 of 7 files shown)


framework/transactions/handlers.py
```diff
@@ -46,21 +46,21 @@ def transaction_before_request():
...
-    if response.status_code >= 500:
+    if response.status_code >= base_status_code_error:
         commands.rollback()
     else:
         try:
             commands.commit()
         except OperationFailure as error:
             message = utils.get_error_message(error)
-            if 'lock not granted' in message.lower():
+            if messages.LOCK_ERROR in message:
```
chrisseto: Is taking the `lower()` out going to break anything?

icereval: I was thinking the same, but decided since it appears to be handled here I would reuse the code.

https://github.com/CenterForOpenScience/osf.io/blob/develop/framework/transactions/context.py#L46

framework/mongo/handlers.py
```diff
@@ -33,8 +32,11 @@ def acquire(self, _id=None):
         return self._local[_id]
 
     def release(self, _id=None):
-        self._cache.append(self._local.pop(_id or self.thread_id))
-        self._sem.release()
+        _id = _id or self.thread_id
+
+        if _id in self._local:
```
chrisseto: Release should not be called if this line is false. It should raise an error if you're attempting to release a connection you don't own

api/base/middleware.py
```diff
@@ -1,70 +1,72 @@
...
+
+    def process_exception(self, request, exception):
+        """If an exception occurs, clear the celery task queue so process_response has nothing."""
+        sentry_exception_handler(request=request)
+        celery_teardown_request(error=True)
+        return None
+
+    def process_response(self, request, response):
+        """Clear the celery task queue if the response status code is 500 or above"""
+        celery_after_request(response, base_status_code_error=500)
```
chrisseto: Should this be 400 to match the rollback base status code error?

If 4 nodes are created but one of them 400's the transaction will get rolled back but all the jobs will still run.

icereval: The jobs should die, as for the 400 vs. 500.

A. We didn't clear them at all before.
B. I'm still not certain there isn't  case when a valid task (email?) might be sent w/ a 400'ish response... thoughts?

PR #5128, opened 2016-03-07
[WIP]Feature/meeting view

chennan47:
<b>Purpose</b>
Change the meeting page's project download count to project view and downloads count.

Solves https://openscience.atlassian.net/browse/OSF-4999


website/project/model.py
```diff
@@ -860,6 +860,7 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin):
     primary_institution = fields.ForeignField('institution')
 
     affiliated_institutions = fields.ForeignField('institution', list=True)
+    visit = fields.IntegerField(default=0)
```
mfraezz: Do we want to keep track of visits on all nodes, or just visits to the materials submitted to conferences?

chennan47: just the nodes that belongs to conferences

website/conferences/views.py
```diff
@@ -148,6 +148,7 @@ def _render_conference_node(node, idx, conf):
             ).limit(1)
         ).wrapped()
         download_count = record.get_download_count()
+        view_and_download = download_count + node.visit
```
mfraezz: `download_count` would be superfluous with these changes; it's not used anywhere other than this line, so it could be changed to `view_and_download = record.get_download_count() + node.visit`

scripts/meeting_visit_count.py
```diff
@@ -0,0 +1,46 @@
...
+from website import settings
+from website.app import init_app
+from website.models import Tag, Conference
+from scripts import utils as scripts_utils
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+
+def main(dry_run=True):
```
mfraezz: It seems like it might make more sense to keep track of the visits elsewhere rather than to run this script every day.

chennan47: How?

website/conferences/views.py
chennan47: fixed in 8cb08e8

website/project/model.py
mfraezz: It looks like the only way of determining what is or is not a conference submission is by [node's tags](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/conferences/views.py#L225)

chennan47: Yes

mfraezz: That seems like a fragile connection, but I guess it makes withdrawing something from a conference easy.

Rather than a script to update this every day, it might be better to add a `StringField` indicating whether the node is connected to a conference (`'<conf_name>'` if so, `None` or `''` if not), and then increment `.visit` in the file view function when the submitted materials are viewed iff `node.conference_submission`.

scripts/meeting_visit_count.py
mfraezz: See [this comment](https://github.com/CenterForOpenScience/osf.io/pull/5128/files#r55255193)

mfraezz: The script would still need to do what it's doing, and would also need to update a field on nodes submitted to any conference, but would only have to be run once. I have little experience with piwik, but it seems like we would want to limit the number of times we have to make that request for all Conference-type nodes.

website/project/model.py
mfraezz: You could still calculate submissions by tags, but this would also (going forward) tell you whether or not a give node was ever connected to a conference, and would be helpful in fixing submissions with their tags erroneously removed.

mfraezz: Also, the Jira ticket seems to imply that this should be counting the views on the materials submitted to the conference, not the node containing them.

chennan47: If a conference tag is removed from a project, that project won't even be showing up in the meeting's page. To collect the visit count of that is meaningless then.

mfraezz: That could be fixed with something like
```python
if node.conference_submission and node.conference_submission in node.tags:
    node.visit += 1
    node.save()
```

PR #5133, opened 2016-03-07
[Ready For Review] Add post transaction commit handlers

cwisecarver:
This adds a second level of task runner to the existing `framework/tasks/handlers.py`. These tasks run synchronously, after the mongo transaction has been committed. Instead of celery signatures they are expected to be callable without arguments. (I used functools.partial to prepare my function for calling at a later date.) There is a new `enqueue_postcommit_task` method that enqueues a task to run postcommit.


cwisecarver:
I'm going to make another PR based off of this one that renames the handlers so that they're not awfully name.

sloria:
Closing in favor of #5143

PR #5137, opened 2016-03-07
[#OSF-5827]. remove the welcome message and add the red warning

chennan47:
<b>Purpose</b>
Remove the welcome message and add a red warning to remind user to log out and revisit the link.

<b>Before change:</b>
image

<b>After change:</b>
image


framework/auth/views.py
```diff
@@ -184,7 +184,10 @@ def confirm_email_get(token, auth=None, **kwargs):
                 return redirect(
                     campaigns.campaign_url_for(campaign)
                 )
-            status.push_status_message(language.WELCOME_MESSAGE, 'default', jumbotron=True)
+            if auth.user.emails == 1:
```
sloria: Shouldn't this be `len(auth.user.emails)`?

chennan47: fixed in  e4b43f4

PR #5143, opened 2016-03-08
[Review Changes Completed] Rename framework.tasks to framework.celery_tasks

cwisecarver:
This refactors framework.tasks into framework.celery_tasks and framework.precommit_tasks. It's built off of https://github.com/CenterForOpenScience/osf.io/pull/5133


framework/postcommit_tasks/handlers.py
```diff
@@ -0,0 +1,41 @@
...
+
+def postcommit_queue():
+    if not hasattr(_local, 'postcommit_queue'):
+        _local.postcommit_queue = []
+    return _local.postcommit_queue
+
+def postcommit_before_request():
+    _local.postcommit_queue = []
+
+def postcommit_after_request(response=None, base_status_error_code=500):
```
sloria: Minor: `response` should probably be required, since you are checking `response.status_code` on the first line of the function.

cwisecarver: Yep, totally missed that. I'll make sure response is always going to show up.

cwisecarver: Double-checked that both flask and the django middleware will always send response. It's required now.

tests/test_app.py
```diff
@@ -16,18 +14,32 @@ def test_attach_handlers():
...
+        framework.sessions.after_request,
+    }
+
+    assert_teardown_funcs = {
+        framework.mongo.handlers.connection_teardown_request,
+        framework.celery_tasks.handlers.celery_teardown_request,
+        framework.transactions.handlers.transaction_teardown_request,
+    }
+
+    # Check that necessary handlers are attached and correctly ordered
```
sloria: Is order actually being checked? It looks like you're sorting the sets before you make the assertions.

website/app.py
```diff
@@ -119,6 +123,8 @@ def init_app(settings_module='website.settings', set_backends=True, routes=True,
     init_addons(settings, routes)
     build_js_config_files(settings)
 
+    django.setup()
```
sloria: Is this to make `reverse` work?

Putting this here may lead to this error:

```python
django.core.exceptions.ImproperlyConfigured: Requested setting LOGGING_CONFIG, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.
```

Would it make sense to set a default for `os.environ['DJANGO_SETTINGS_MODULE']` to `api.base.settings` here?

website/settings/local-travis.py
```diff
@@ -42,4 +44,5 @@
 
 VARNISH_SERVERS = ['http://localhost:8080']
 
-ENABLE_VARNISH = True
+# if ENABLE_VARNISH isn't set in python read it from the env var and set it
+locals().setdefault('ENABLE_VARNISH', os.environ.get('ENABLE_VARNISH') == 'True')
```
sloria: Couldn't this just be

```python
ENABLE_VARNISH = os.environ.get('ENABLE_VARNISH') == 'True'
```

?

website/app.py
cwisecarver: It is there to make reverse work. I'm trying to find an alternative because this is breaking admin. We can't have django settings pointing at api to make reverse work and admin to make admin work at the same time.

sloria:
Pass finished.

tests/test_app.py
cwisecarver: You're right. I'm just checking that the lists match but the comment says I'm checking order. I'll fix.

website/settings/local-travis.py
cwisecarver: The setdefault lets me not reset ENABLE_VARNISH if it is set in python already. This is because the travis needs this setting False for the first two passes and True for the third.

sloria: I see; thanks for the clarification.

cwisecarver:
@sloria This is good to go whenever you have time to merge it.

cwisecarver:
Thanks @sloria

PR #5145, opened 2016-03-08
Fixed missing description helper text when a project gets registered.

monikagrabowska:
When there is no description after registration, there is now text that says "No description" rather than a blank space. 

[OSF-5694]

### DIFF
--- a/website/templates/project/project.mako
+++ b/website/templates/project/project.mako
@@ -188,11 +188,14 @@
                 <span data-bind="css: icon"></span>
                 </p>
 
-                % if node['description'] or 'write' in user['permissions']:
-                    <p>
+
+                <p>
+                % if not node['description']:
+                    <span id="description">Description:</span> <span id="nodeDescriptionEditable" class="text-muted" data-type="textarea">${'No description'}</span>
+                % elif node['description'] or 'write' in user['permissions']:
                     <span id="description">Description:</span> <span id="nodeDescriptionEditable" class="node-description overflow" data-type="textarea">${node['description']}</span>
-                    </p>
                 % endif
+                </p>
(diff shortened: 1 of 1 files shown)


sloria:
Nice git flow. :microphone:

website/templates/project/project.mako
```diff
@@ -188,11 +188,14 @@
                 <span data-bind="css: icon"></span>
                 </p>
 
-                % if node['description'] or 'write' in user['permissions']:
-                    <p>
+
+                <p>
+                % if not node['description']:
+                    <span id="description">Description:</span> <span id="nodeDescriptionEditable" class="text-muted" data-type="textarea">${'No description'}</span>
```
sloria: There is no need to do `${'No description'}` because "No description" is just text, not a variable or an expression.

website/templates/project/project.mako
```diff
@@ -188,11 +188,14 @@
                 <span data-bind="css: icon"></span>
                 </p>
 
-                % if node['description'] or 'write' in user['permissions']:
-                    <p>
+
+                <p>
+                % if not node['description']:
```
sloria: I believe we only want to show "No description" for contributors for the project (more specifically, contributors with "write" permissions). This allows contributors to click on the text to edit it.

sloria:
I've just discussed the desired behavior with Nici.

For registrations with no description, the "Description" line should not be displayed for *any* user. There is no need to show "No description" to contributors because they can't edit the description anyway.

Nici is in the process of updating the ticket to reflect this change.

@monikagrabowska Can you please open a new PR implementing this?

monikagrabowska:
Okay!

sloria:
Thanks

PR #5146, opened 2016-03-08
Hotfix/alternative email confirm

chennan47:
<b>Purpose</b>
Fix the remaining problem of https://openscience.atlassian.net/browse/OSF-5827.

### DIFF
--- a/framework/auth/views.py
+++ b/framework/auth/views.py
@@ -184,9 +184,10 @@ def confirm_email_get(token, auth=None, **kwargs):
                 return redirect(
                     campaigns.campaign_url_for(campaign)
                 )
-            if len(auth.user.emails) == 1:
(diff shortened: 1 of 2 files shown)


framework/auth/views.py
```diff
@@ -184,9 +184,10 @@ def confirm_email_get(token, auth=None, **kwargs):
                 return redirect(
                     campaigns.campaign_url_for(campaign)
                 )
-            if len(auth.user.emails) == 1:
+            if len(auth.user.emails) == 1 and len(auth.user.email_verifications) == 0:
```
sloria: This doesn't look correct; won't a new user have exactly one `email_verification`, the one for their primary email?

sloria: Discussed in person: this is actually correct. At this point, the user's primary email will have been confirmed, so there should be no email_verifications.

PR #5148, opened 2016-03-08
[hold]  [feature]Hover description for MD5 and SHA2 for file revision

tnghieu:
## Purpose
The purpose of this pull request is to implement a feature to allow users to understand what the MD5 and SHA2 features are of the Revisions. By having this feature, the user is capable of making file comparisons as well as fully understanding what is provided on nearly half of the screen.

## Changes
Added a question-circle next to MD5 and SHA2. Hovering over the question-circle creates a hovering popup that shows the description of each. Implementation was done on revisions.js on lines 94-98 and 108-113.
screen shot 2016-03-08 at 1 24 58 pm

## Side Effects
Definitions/descriptions of MD5 and SHA2 must be taken into consideration.

### DIFF
--- a/website/static/js/filepage/revisions.js
+++ b/website/static/js/filepage/revisions.js
@@ -90,6 +90,14 @@ var FileRevisionsTable = {
             });
         };
 
+
+        var popOver = function(element, isInit) {
+            if (!isInit) {
+                $(document.getElementById(element.id)).popover();
+            }
+
+        }
+
         self.getTableHead = function() {
             return m('thead', [
                 m('tr', [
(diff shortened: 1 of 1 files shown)


website/static/js/filepage/revisions.js
```diff
@@ -90,6 +90,14 @@ var FileRevisionsTable = {
             });
         };
 
+
+        var popOver = function(element, isInit) {
+            if (!isInit) {
+                $(document.getElementById(element.id)).popover();
```
sloria: You already have access to the DOM element in this function (`element`), so no need to get the element by ID.

```js
$(element).popover();
```

website/static/js/filepage/revisions.js
```diff
@@ -97,12 +105,18 @@ var FileRevisionsTable = {
                     model.hasDate ? m('th', 'Date') : false,
                     model.hasUser ? m('th', 'User') : false,
                     m('th[colspan=2]', 'Download'),
-                    model.hasHashes ? m('th', 'MD5') : false,
-                    model.hasHashes ? m('th', 'SHA2') : false,
+                    model.hasHashes ? m('th', [
+                        'MD5 ', m('#md5pop.fa.fa-question-circle[data-content="Unique 32-digit code for the file."][rel="popover"]' +
```
sloria: It's best to avoid reliance on CSS IDs, as they are brittle (changing the ID will break the code). If you do the refactor suggested in https://github.com/CenterForOpenScience/osf.io/pull/5148/files#r55424839 , you should be able to remove the IDs here.

sloria:
Pass finished. Please address the comments above. There is a also a `jshint` error. You can run jshint locally with `inv jshint`. The builds will fail if jshint finds any errors.

tnghieu:
@sloria I've responded and fixed the issues.

sloria:
@tnghieu Thanks. Is there a JIRA ticket for this change?

hmoco:
@sloria, here is the ticket: https://openscience.atlassian.net/browse/OSF-4443

sloria:
Thanks. Code looks good, but I'd like to get a review from Erin M on what the tooltip text should be. Consider this good to merge, pending that review. :octopus:

erinmayhood:
@tnghieu Let's keep the descriptions super simple:

MD5 is an algorithm used to verify data integrity.

SHA-2 is a cryptographic has function designed by the NSA used to verify data integrity.

Thank you and nice work getting a PR in so quick!

sloria:
Thanks Erin. I've made the language change. Thanks Tom for the PR.

PR #5149, opened 2016-03-08
Hides "Description" after registration if initially blank

monikagrabowska:
Instead of having grey helper text when the description field is left blank, now the description field is hidden from users when left blank.

### DIFF
--- a/website/templates/project/project.mako
+++ b/website/templates/project/project.mako
@@ -188,11 +188,14 @@
                 <span data-bind="css: icon"></span>
                 </p>
 
-                % if node['description'] or 'write' in user['permissions']:
-                    <p>
+
+                <p>
+                % if not node['description']:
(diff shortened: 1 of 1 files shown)


website/templates/project/project.mako
```diff
@@ -188,10 +188,12 @@
                 <span data-bind="css: icon"></span>
                 </p>
 
-                % if node['description'] or 'write' in user['permissions']:
-                    <p>
+                % if not node['description']:
+                    <p></p>
```
sloria: This extra `p` tag will result in unwanted whitespace on the page. Can you simplify your logic so that this can be removed?

sloria:
Pass finished. Just a minor change and this should be good to merge.

monikagrabowska:
Okay! I am working on this now.

PR #5151, opened 2016-03-08
Hide "Description", remove whitespace 

monikagrabowska:
Description field hidden from users after registration if left blank initially - extra whitespace created by previous code removed.

### DIFF
--- a/website/templates/project/project.mako
+++ b/website/templates/project/project.mako
@@ -188,11 +188,14 @@
                 <span data-bind="css: icon"></span>
                 </p>
 
-                % if node['description'] or 'write' in user['permissions']:
-                    <p>
+
+                <p>
+                % if not node['description']:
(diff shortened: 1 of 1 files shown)


website/templates/project/project.mako
```diff
@@ -188,7 +188,9 @@
                 <span data-bind="css: icon"></span>
                 </p>
 
-                % if node['description'] or 'write' in user['permissions']:
+                % if not node['description']:
+                    <% pass %>
```
sloria: Rather than having a conditional that just "passes", can you change the existing conditional so it gives the desired output?

sloria:
Pass finished. See if you can simplify the logic further. Otherwise, behavior looks right.

monikagrabowska:
Okay! Thank you!

PR #5154, opened 2016-03-09
Hides "Description" from all users after registration if initially blank

monikagrabowska:
When there is no description after registration, the field is hidden from all users. 

[OSF-5694]


website/templates/project/project.mako
```diff
@@ -188,7 +188,7 @@
                 <span data-bind="css: icon"></span>
                 </p>
 
-                % if node['description'] or 'write' in user['permissions']:
+                % if node['description']:
```
sloria: With this logic, contributors will not be able to change the description by clicking "No description".

If a project has no description, the desired behavior is:

- A contributor with "write" permission sees "Description: No description", which they can click on to set the description.
- Non-contributors do not see "Description". 

See if you can modify the logic to get this behavior, and make sure to verify the behavior with manual testing.

sloria:
Pass finished. See comment above.

By the way, you don't need to create a new PR each time you make a change. You can just make new commits on your branch, `feature/fix_description`. When you push those commits, the PR will be updated. :v:

website/addons/forward/templates/forward_node_settings.mako
```diff
@@ -53,7 +53,6 @@
                     <input type="submit"
                            class="btn btn-success pull-right"
                            value="Save"
-                           data-bind="disable: !validators.isValid()"
```
sloria: Why was this line removed?

monikagrabowska:
I was working on another task involving one of the buttons on the project settings page, but I think I forgot to switch branches when I committed my changes to the description feature.

sloria:
Ah ok, no worries.

The logic still doesn't look quite right. If I'm understanding it correctly, your change doesn't change the behavior from the previous logic.

The missing bit is that "Description" should not show for registrations if there is not description. From the JIRA ticket:

> The the correct behavior is to hide "Description" for all users if the project is a registration and the description is blank

Can you look into this?

monikagrabowska:
Should contributors with write permissions still be able to see "Description" even if there is no description?

sloria:
Write contributors should always see "Description" for projects but not for registrations.

sloria:
Looks good!

PR #5159, opened 2016-03-09
[Feature] Institutions 2

hmoco:


api/institutions/views.py
```diff
@@ -1,11 +1,14 @@
...
+from rest_framework import status
+from rest_framework.response import Response
 
 from modularodm import Q
 
 from framework.auth.oauth_scopes import CoreScopes
 
-from website.models import Institution, Node, User
+from website.models import Node, User
+from website.project.model import Institution
```
caseyrollins: Not super important, but is there a reason the `Institution` model isn't included in `website.models`?

hmoco: yep, ` website.models ` is reserved for StoredObjects, as this is where modularodm looks for them (I think?)

caseyrollins: :+1:

framework/auth/core.py
```diff
@@ -1376,13 +1380,22 @@ def is_affiliated_with_institution(self, inst):
         return inst in self.affiliated_institutions
 
     def remove_institution(self, inst_id):
-        try:
-            self.affiliated_institutions.remove(inst_id)
-        except ValueError:
-            return False
-        return True
+        node_id = False
+        for node in self._affiliated_institutions:
```
caseyrollins: It would be better to iterate over the institutions in `self.affiliated_institutions` instead of the nodes in the protected `self._affiliated_institutions`.

framework/auth/core.py
```diff
@@ -1376,13 +1380,22 @@ def is_affiliated_with_institution(self, inst):
...
+            if node.institution_id == inst_id:
+                node_id = node._id
+                break
+        if node_id:
+            self._affiliated_institutions.remove(node_id)
```
caseyrollins: `affiliated_institutions.remove()` instead of `_affiliated_institutions.remove()` here.

caseyrollins:
[General comment about tests.](https://github.com/CenterForOpenScience/osf.io/pull/5106#issuecomment-193978403)

website/project/model.py
```diff
@@ -3478,6 +3419,228 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+    def add_primary_institution(self, user, inst, log=True):
+        if not isinstance(inst, Institution):
+            raise TypeError
+        if not user.is_affiliated_with_institution(inst):
+            raise UserNotAffiliatedError('User is not affiliated with {}'.format(inst.name))
+        if inst == self.primary_institution:
+            return False
+        previous = self.primary_institution if self.primary_institution else None
+        self.primary_institution = inst
+        if inst not in self._affiliated_institutions:
```
caseyrollins: Looking for an `Institution` in a list of `Nodes`.
Checking `if inst not in self.affiliated_institutions` would be better.

website/project/model.py
```diff
@@ -3478,6 +3419,228 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+            self.add_log(
+                action=NodeLog.PRIMARY_INSTITUTION_CHANGED,
+                params={
+                    'node': self._primary_key,
+                    'institution': {
+                        'id': inst._id,
+                        'name': inst.name
+                    },
+                    'previous_institution': {
+                        'id': previous._id if previous else None,
```
caseyrollins: Do the logs render/behave correctly when the `previous_institution` values are `None`?

website/static/js/institutionNodes.js
```diff
@@ -0,0 +1,32 @@
...
+    self.fetchInstitutionNodes = function _fetchInstitutionNodes(){
+        return $osf.ajaxJSON(
+            'GET',
+            self.ctx.apiV2Prefix + 'institutions/' + self.ctx.institution.id + '/?embed=nodes',
+            {
+                isCors: true
+            }
+        ).done( function(response){
+            self.allNodes(response.data.embeds.nodes.data);
+        }).fail(function(response){
```
caseyrollins: Handle failure in some way?

website/static/js/pages/institution-page.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+var $ = require('jquery');
+var ko = require('knockout');
+
+ko.punches.enableAll();
+
+var FileBrowser = require('js/dashboard.js').Dashboard;
```
caseyrollins: Why `FileBrowser` instead of `Dashboard`?

website/templates/institution.mako
```diff
@@ -0,0 +1,52 @@
...
+    <div class="alert alert-info"><strong>NOTICE: </strong>Forks, registrations, and uploads will be temporarily disabled while the OSF undergoes a hardware upgrade. These features will return shortly. Thank you for your patience.</div>
+% endif
+    <div id="inst">
+<div class="dashboard-header">
+    <div class="row" style="text-align: center">
+        <div class="col-sm-2"></div>
+        <div class="col-sm-3"><img class="img-circle" height="110" width="110" src=${ logo_path }></div>
+        <div class="col-sm-3">
+            <h2>${ name }</h2>
+            <h4><small class="hidden-sm">${description}</small></h4>
```
caseyrollins: Would it be better to check if there is content in the description first? So that this element doesn't take up space on the page if the institution doesn't have a description.

website/project/model.py
hmoco: yep, the logs render that institution was changed from ` None ` to ` Institution_name `

website/static/js/dashboard.js
```diff
@@ -836,7 +853,7 @@ var Collections = {
                     'data-placement' : 'bottom'
                 }, ''),
                 m('.pull-right', [
-                    m('button.btn.btn-xs.btn-success[data-toggle="modal"][data-target="#addColl"]', m('i.fa.fa-plus')),
+                    !viewOnly ? m('button.btn.btn-xs.btn-success[data-toggle="modal"][data-target="#addColl"]', m('i.fa.fa-plus')) : '',
```
caseyrollins: In addition to not displaying the 'Add Collection' button, should there be a check in the `_addCollection` function to prevent it from being called on an Institution?

website/project/model.py
caseyrollins: :+1:

caseyrollins:
@hmoco, done with general first pass :dog: :star2:

website/templates/institution.mako
hmoco: I think this is enough as it doesn't just hide the button, it prevents it from rendering. Unless there is another use of ` #addColl`

api/institutions/authentication.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def authenticate(self, request):
+        try:
+            payload = jwt.decode(
+                jwe.decrypt(request.body, settings.JWE_SECRET),
+                settings.JWT_SECRET,
+                options={'verify_exp': False},
+                algorithm='HS256'
+            )
+        except (jwt.InvalidTokenError, KeyError, ValueError, TypeError):
```
sloria: You are handling many errors the same way. At the very least, it would be good to have different error messages for different kinds of errors; this will help with debugging. Also, I'm not clear on why all these errors need to be caught. `InvalidTokenError` makes sense, but why `KeyError`, `ValueError` and `TypeError`?

api/institutions/authentication.py
```diff
@@ -0,0 +1,68 @@
...
+        #     "user": {
+        #         "middleNames": "",
+        #         "familyName": "",
+        #         "givenName": "",
+        #         "fullname": "Circle User",
+        #         "suffix": "",
+        #         "username": "user@circle.edu"
+        #     }
+        # }}
+        data = json.loads(payload['data'])
```
sloria: Could a `KeyError` be raised on this line or the next line?

api/nodes/serializers.py
```diff
@@ -526,9 +528,10 @@ def to_representation(self, obj):
         type_ = getattr(meta, 'type_', None)
         assert type_ is not None, 'Must define Meta.type_'
         relation_id_field = self.fields['id']
-        attribute = relation_id_field.get_attribute(obj)
+        attribute = obj.primary_institution._id if obj.primary_institution else None
         relationship = relation_id_field.to_representation(attribute)
-
+        if relationship == 'None':
```
sloria: This is suspect; why is the relationship_id_field serializing to the string `'None'`? Shouldn't it serialize to `None`?

api/institutions/views.py
sloria: For consistency, I think it still makes sense to have `Institution` imported in `website.models`. You don't need to add it to `website.models.MODELS` because that includes all the StoredObjects that get attached to Mongo when we initialize the app.

sloria: Note: If you do this, you'll probably need to add `# flake8: noqa` to avoid the "unused import" error.

api_tests/institutions/views/test_institution_auth.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def test_creates_user(self):
+        username = 'hmoco@circle.edu'
+        assert_equal(User.find(Q('username', 'eq', username)).count(), 0)
+
+        with capture_signals() as mock_signals:
+            res = self.app.post(self.url, self.build_payload(username))
+
+        assert_equal(res.status_code, 204)
+        assert_equal(mock_signals.signals_sent(), set([signals.user_confirmed]))
```
sloria: :+1:

framework/auth/__init__.py
```diff
@@ -79,3 +80,24 @@ def register_unconfirmed(username, password, fullname, campaign=None):
...
+    :return: Tuple of (user, created)
+    """
+    user = get_user(email=address)
+    if user:
+        return user, False
+    else:
+        from website import security  # Avoid circular imports
+        password = str(uuid.uuid4())
+        user = User.create_confirmed(address, password, fullname)
+        user.verification_key = security.random_string(20)
```
sloria: Why does this get set?

sloria: Oh, is the user expected to reset their password after their account is created?

framework/auth/core.py
```diff
@@ -1381,13 +1385,19 @@ def is_affiliated_with_institution(self, inst):
...
+                removed = True
+        return removed
+
+    _affiliated_institutions = fields.ForeignField('node', list=True)
 
-    affiliated_institutions = fields.ForeignField('institution', list=True)
+    @property
+    def affiliated_institutions(self):
+        from website.project.model import Institution, SpecialList
+        return SpecialList(self, '_affiliated_institutions', [Institution(inst) for inst in self._affiliated_institutions])
```
sloria: `SpecialList` is not a descriptive name. How about `ProxiedList`, since `affiliated_institutions` acts as a proxy for `_affiliated_institutions`?

sloria: Also the order of arguments in the constructor isn't intuitive; the API for `SpecialList` should be close to `list`, so I would expect to pass an iterable as the first argument--not the last.

website/project/model.py
```diff
@@ -3478,6 +3420,236 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+        return True
+
+    def institution_url(self):
+        return self.absolute_api_v2_url + 'institution/'
+
+    def institution_relationship_url(self):
+        return self.absolute_api_v2_url + 'relationships/institution/'
+
+
+class SpecialList(list):
```
sloria: Cleverness like this should have a docstring. Be sure to document the use case, ideally with a short code example.

website/project/model.py
```diff
@@ -3478,6 +3420,236 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+    def __init__(self, node, private_target, init=None):
```
sloria: Minor: there's nothing specific to `Node` in this class, so the "node" argument could have a more generic name. Maybe "obj"?

website/project/model.py
```diff
@@ -3478,6 +3420,236 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+        super(SpecialList, self).__init__(init or [])
+        self.node = node
+        self.target = private_target
+
+    def append(self, to_append):
+        junk = getattr(self.node, self.target)
```
sloria: Do your best to be descriptive (and concise) with your variable names. Maybe `private_list`?

website/project/model.py
```diff
@@ -3478,6 +3420,236 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+        for each in super(InstitutionQuerySet, self).__iter__():
+            yield Institution(each)
+
+    def _do_getitem(self, index):
+        item = super(InstitutionQuerySet, self)._do_getitem(index)
+        if isinstance(item, MongoQuerySet):
+            return self.__class__(item)
+        return Institution(item)
+
+class Institution():
```
sloria: In Python 2, all classes should inherit from `object`.

website/project/model.py
```diff
@@ -3478,6 +3420,236 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+    '''
+    "wrapper" class for Node. Together with the find and institution attributes & methods in Node,
```
sloria: :+1:

website/project/model.py
```diff
@@ -3478,6 +3420,236 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+        'domain': 'institution_domain',
+        'name': 'title',
+        'logo_name': 'institution_logo_name',
+        'description': 'description',
+        'email_domain': 'institution_email_domain'
+    }
+
+    def __init__(self, node):
+        if node is None:
+            return
```
sloria: This isn't entirely intuitive, because it will result in instances that may or may not have the `node` attribute.

```python
institution = Institution(node)
institution.node # => Node

institution = Institution()
institution.node # => AttributeError
```

Instead, I would expect:

```python
institution = Institution()
institution.node # => None
```

website/project/model.py
```diff
@@ -3478,6 +3420,236 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+            if getattr(self, key) != getattr(self.node, value):
+                setattr(self.node, value, getattr(self, key))
+        self.node.save()
+
+    @classmethod
+    def find(cls, query=None, **kwargs):
+        if query and getattr(query, 'nodes', False):
+            for node in query.nodes:
+                replacement_attr = cls.attribute_map.get(node.attribute, False)
+                node.attribute = replacement_attr if replacement_attr else node.attribute
```
sloria: Minor: can be be simplified to 

```python
node.attribute = replacement_attr or node.attribute
```

website/project/model.py
```diff
@@ -3478,6 +3420,236 @@ def require_approval(self, user, notify_initiator_on_complete=False):
...
+        query = query & Q('is_institution', 'eq', True) if query else Q('is_institution', 'eq', True)
+        node = Node.find_one(query, allow_institution=True, **kwargs)
+        return cls(node)
+
+    @classmethod
+    def load(cls, id):
+        try:
+            node = Node.find_one(Q('institution_id', 'eq', id) & Q('is_institution', 'eq', True), allow_institution=True)
+            return cls(node)
+        except:
```
sloria: Avoid naked `excepts`; handle a specific exception class instead.

api/institutions/authentication.py
chrisseto: Newp

website/project/views/node.py
```diff
@@ -247,6 +247,14 @@ def node_forks(auth, node, **kwargs):
 @must_be_contributor
 def node_setting(auth, node, **kwargs):
 
+    #check institutions:
+    try:
+        inst = Institution.find_one(Q('email_domain', 'eq', auth.user.username.split('@')[1]))
+        auth.user.affiliated_institutions.append(inst)
```
sloria: @hmoco Is it possible for the same institution to be added more than once here?

Also, is this automatic affiliation the desired behavior? @icereval

website/routes.py
```diff
@@ -40,12 +40,21 @@
...
 def get_globals():
     """Context variables that are available for every template rendered by
     OSFWebRenderer.
     """
     user = _get_current_user()
+    if request.host_url != settings.DOMAIN:
+        try:
+            inst_id = (Institution.find_one(Q('domain', 'eq', request.host)))._id
+            login_url = '{}institution/{}'.format(settings.DOMAIN, inst_id)
+        except:
```
sloria: Same here: avoid naked `except`.

website/project/model.py
sloria: This module is already pretty bloated: Can you move the institution code to its own module, e.g. `website.institutions`? This `Institution` class can go in `websit.institutions.models`.

website/static/js/institutionNodes.js
```diff
@@ -0,0 +1,40 @@
...
+var ko = require('knockout');
+var Raven = require('raven-js');
+
+var $osf = require('js/osfHelpers');
+
+
+var ViewModel = function(context) {
+    var self = this;
+    self.ctx = context;
+    self.allNodes = ko.observable();
```
sloria: Shouldn't this be an `observableArray`?

website/static/js/pages/institution-page.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+var $ = require('jquery');
+var ko = require('knockout');
+
+ko.punches.enableAll();
+
+var Projects = require('js/myProjects.js').MyProjects;
+var LinkObject = require('js/myProjects.js').LinkObject;
```
sloria: I thiiink webpack is smart enough to handle the duplicate import, but just to be sure, better to do

```js
var _myProjects = require('js/myProjects');
var Projects = _myProjects.MyProjects;
//...
```

sloria:
Why the logo image change?

According to http://onmessage.nd.edu/notre-dame-brand-standards/logos/academic-mark/ , we are using the academic mark improperly:

> **The shield may never be separated from the word mark.**

website/views.py
```diff
@@ -75,7 +76,21 @@ def _render_nodes(nodes, auth=None, show_path=False):
...
+        return {
+            'id': inst._id,
+            'name': inst.name,
+            'logo_path': inst.logo_path,
+            'description': inst.description or '',
+            'home': False,
+            'institution': True,
+            'redirect_url': '/institution/{}/'.format(inst._id)
+        }
+    except:
```
sloria: Again: no naked `excepts`.

website/profile/views.py
```diff
@@ -822,3 +822,14 @@ def redirect_to_twitter(twitter_handle):
         })
 
     return redirect(user.url)
+
+
+def view_institution(**kwargs):
+    inst = Institution.load(kwargs.get('id'))
```
sloria: As per COSDev, prefer to put path variables directly in the signature: `view_institution(id, **kwargs)` rather than pulling from kwargs. If you make that change, you might want to rename `id` to avoid shadowing the Python built-in.

sloria: What if an invalid ID is passed? Shouldn't a 404 be returned?

sloria:
Pass finished. After making the suggested changes, please update with the latest `feature/dashboard`.

hmoco:
Logo image is based on the marketing material handed to us by ND, although the cropping was done by me. I believe however is communicating with ND needs to clear up what logo should be used.

website/static/js/institutionNodes.js
hmoco: it doesn't need to be if I remember correctly, as I'm just assigning the whole list to it, rather than iterating and pushing to it.

api/nodes/serializers.py
hmoco: Not exactly sure why this is happening.

hmoco:
@sloria once tests pass, have responded to comments.

website/static/js/institutionNodes.js
sloria: OK, but it's still a good idea to make this an `observableArray`. (1) It communicates that the value of this observable is an array and (2) prevents future devs from becoming sad if they ever do `self.allNodes.push(node)`.

All that said, good call on assigning the whole list at once; that is preferred over `push` anyway, so :+1:

api/nodes/serializers.py
sloria: Can you investigate?

sloria:
Not hard blockers, but I'd still like to see https://github.com/CenterForOpenScience/osf.io/pull/5159#discussion_r55849399 and https://github.com/CenterForOpenScience/osf.io/pull/5159#discussion_r55891222 addressed.

api/nodes/serializers.py
hmoco: So, because it its a CharField (the attribute ` institution_id ` is), it get this:
```
        return six.text_type(value)
```
will continue to investigate if it can be turned into ` null `

hmoco:
@sloria have responded to those 2 comments

PR #5162, opened 2016-03-09
Pop-up added when navigating away from project settings page after editing title/description

NehaTelhan:
**Issue:** https://openscience.atlassian.net/browse/OSF-5104 
# **Purpose**: 
Currently, when a user makes changes to either their category, title, or description in the project settings page and then tries to navigate to another page, their savings won't automatically be made nor are they alerted that they have unsaved changes on the page. This PR will show a pop-up alert when the user is navigating to another page without having saved their changes.

# **Changes**
Shows a pop-up window that tells the user they're leaving a page without having saved the changes made. There are two buttons on this pop-up, the first to 'stay on this page' with the implication that they can save their changes before navigating to another page, and the second to 'leave this page', anyways. This is only enacted if any or all of the following are altered without hitting 'save changes': title, category, description. 

# **Side Effects**
As very few lines of code were added (four lines including and if-statement with a return), no side effects are anticipated.

### DIFF
--- a/website/static/js/pages/project-settings-page.js
+++ b/website/static/js/pages/project-settings-page.js
@@ -230,10 +230,10 @@ $(document).ready(function() {
           return 'The changes on addon setting are not submitted!';
       }
     /* Before closing the page, Check whether changes made to category, title or description are updated or not */
-      if ($('#titleID')[0].value !== (projectSettingsVM).titlePlaceholder  ||
-          $('#descriptionID')[0].value !== (projectSettingsVM).descriptionPlaceholder ||
-          $('#categoryID')[0].value !== (projectSettingsVM).categoryPlaceholder  ) {
-          return 'There are unsaved changes in your project section.';
+      if (projectSettingsVM.title() !== projectSettingsVM.titlePlaceholder ||
+          projectSettingsVM.description() !== projectSettingsVM.descriptionPlaceholder ||
+          projectSettingsVM.selectedCategory() !== projectSettingsVM.categoryPlaceholder) {
+          return 'There are unsaved changes in your project settings.';
       }
     });
 
--- a/website/static/js/pages/project-settings-page.js
+++ b/website/static/js/pages/project-settings-page.js
@@ -233,7 +233,7 @@ $(document).ready(function() {
       if ($('#titleID')[0].value != (projectSettingsVM).titlePlaceholder  ||
           $('#descriptionID')[0].value != (projectSettingsVM).descriptionPlaceholder ||
           $('#categoryID')[0].value != (projectSettingsVM).categoryPlaceholder  ) {
-          return "There are unsaved changes in your project section."
+          return "There are unsaved changes in your project section.";
       }
     });
 
--- a/website/static/js/pages/project-settings-page.js
+++ b/website/static/js/pages/project-settings-page.js
@@ -227,7 +227,13 @@ $(document).ready(function() {
       var unchecked = checkedOnLoad.filter('#selectAddonsForm input:not(:checked)');
 
       if(unchecked.length > 0 || checked.length > 0) {
-        return 'The changes on addon setting are not submitted!';
+          return 'The changes on addon setting are not submitted!';
+      }
+    /* Before closing the page, Check whether changes made to category, title or description are updated or not */
+      if ($('#titleID')[0].value != (projectSettingsVM).titlePlaceholder  ||
+          $('#descriptionID')[0].value != (projectSettingsVM).descriptionPlaceholder ||
+          $('#categoryID')[0].value != (projectSettingsVM).categoryPlaceholder  ) {
+          return "There are unsaved changes in your project section."
       }
     });
 
--- a/website/static/js/pages/project-settings-page.js
+++ b/website/static/js/pages/project-settings-page.js
@@ -230,10 +230,10 @@ $(document).ready(function() {
           return 'The changes on addon setting are not submitted!';
       }
     /* Before closing the page, Check whether changes made to category, title or description are updated or not */
-      if ($('#titleID')[0].value != (projectSettingsVM).titlePlaceholder  ||
-          $('#descriptionID')[0].value != (projectSettingsVM).descriptionPlaceholder ||
-          $('#categoryID')[0].value != (projectSettingsVM).categoryPlaceholder  ) {
-          return "There are unsaved changes in your project section.";
+      if ($('#titleID')[0].value !== (projectSettingsVM).titlePlaceholder  ||
+          $('#descriptionID')[0].value !== (projectSettingsVM).descriptionPlaceholder ||
+          $('#categoryID')[0].value !== (projectSettingsVM).categoryPlaceholder  ) {
+          return 'There are unsaved changes in your project section.';
       }
     });
 
--- a/website/templates/project/settings.mako
+++ b/website/templates/project/settings.mako
@@ -77,24 +77,20 @@
                     <div id="projectSettings" class="panel-body">
                         <div class="form-group">
                             <label>Category:</label>
-                            <select id = categoryID data-bind="attr.disabled: disabled,
-                                               options: categoryOptions,
-                                               optionsValue: 'value',
-                                               optionsText: 'label',
-                                               value: selectedCategory"></select>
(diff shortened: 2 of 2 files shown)


pattisdr:
Please add Issue https://openscience.atlassian.net/browse/OSF-5104 in your description above.

website/static/js/pages/project-settings-page.js
```diff
@@ -227,7 +227,13 @@ $(document).ready(function() {
       var unchecked = checkedOnLoad.filter('#selectAddonsForm input:not(:checked)');
 
       if(unchecked.length > 0 || checked.length > 0) {
-        return 'The changes on addon setting are not submitted!';
+          return 'The changes on addon setting are not submitted!';
+      }
+    /* Before closing the page, Check whether changes made to category, title or description are updated or not */
+      if ($('#titleID')[0].value !== (projectSettingsVM).titlePlaceholder  ||
```
pattisdr: Minor: You shouldn't need parenthesis around projectSettingsVM in L233, L234, L235.

website/static/js/pages/project-settings-page.js
```diff
@@ -227,7 +227,13 @@ $(document).ready(function() {
...
+          $('#descriptionID')[0].value !== (projectSettingsVM).descriptionPlaceholder ||
+          $('#categoryID')[0].value !== (projectSettingsVM).categoryPlaceholder  ) {
+          return 'There are unsaved changes in your project section.';
```
pattisdr: I would change `project section` to `project settings` to be consistent with other error message.  There we are calling them addon settings.

website/templates/project/settings.mako
```diff
@@ -77,24 +77,24 @@
...
                             <label for="title">Title:</label>
-                            <input class="form-control" type="text" maxlength="200" placeholder="Required" data-bind="value: title,
+                            <input id= "titleID" class="form-control" type="text" maxlength="200" placeholder="Required" data-bind="value: title,
                                                                                                       valueUpdate: 'afterkeydown'">
                             <span class="text-danger" data-bind="validationMessage: title"></span>
                         </div>
                         <div class="form-group">
                             <label for="description">Description:</label>
-                            <textarea placeholder="Optional" data-bind="value: description,
+                              <textarea  id= descriptionID placeholder="Optional" data-bind="value: description,
```
pattisdr: Remove extra space before <textarea....   Also, there are two spaces between `textarea` and `id`.

website/templates/project/settings.mako
```diff
@@ -77,24 +77,24 @@
                     <div id="projectSettings" class="panel-body">
                         <div class="form-group">
                             <label>Category:</label>
-                            <select data-bind="attr.disabled: disabled,
-                                                         options: categoryOptions,
-                                                         optionsValue: 'value',
-                                                         optionsText: 'label',
-                                                         value: selectedCategory"></select>
+                            <select id = categoryID data-bind="attr.disabled: disabled,
```
pattisdr: For consistency, remove extra spaces before and after equals sign.

website/templates/project/settings.mako
```diff
@@ -77,24 +77,24 @@
...
+                                               optionsText: 'label',
+                                               value: selectedCategory"></select>
                             <span data-bind="if: disabled" class="help-block">
                               A top-level project's category cannot be changed
                             </span>
                         </div>
                         <div class="form-group">
                             <label for="title">Title:</label>
-                            <input class="form-control" type="text" maxlength="200" placeholder="Required" data-bind="value: title,
+                            <input id= "titleID" class="form-control" type="text" maxlength="200" placeholder="Required" data-bind="value: title,
```
pattisdr: Remove extra spaces after equals sign.

website/static/js/pages/project-settings-page.js
sloria: Drive-by comment: Rather than relying on CSS IDs, which are brittle (i.e. the code will break if the IDs are changed or removed), can you use instead access attributes/observables on the view model? For example: `projectSettingsVM.title()` rather than `$('#titleID')[0].value`

pattisdr:
Okay, @NehaTelhan, first pass complete! :sunny: 

Just requesting minor changes in formatting.  Actual functionality checks out. Well done!

website/static/js/pages/project-settings-page.js
```diff
@@ -227,7 +227,13 @@ $(document).ready(function() {
       var unchecked = checkedOnLoad.filter('#selectAddonsForm input:not(:checked)');
 
       if(unchecked.length > 0 || checked.length > 0) {
-        return 'The changes on addon setting are not submitted!';
+          return 'The changes on addon setting are not submitted!';
+      }
+    /* Before closing the page, Check whether changes made to category, title or description are updated or not */
+      if ($('#titleID')[0].value !== (projectSettingsVM).titlePlaceholder  ||
+          $('#descriptionID')[0].value !== (projectSettingsVM).descriptionPlaceholder ||
+          $('#categoryID')[0].value !== (projectSettingsVM).categoryPlaceholder  ) {
```
pattisdr: Remove extra spaces after categoryPlaceholder.

website/templates/project/settings.mako
```diff
@@ -77,24 +77,24 @@
                     <div id="projectSettings" class="panel-body">
                         <div class="form-group">
                             <label>Category:</label>
-                            <select data-bind="attr.disabled: disabled,
-                                                         options: categoryOptions,
-                                                         optionsValue: 'value',
-                                                         optionsText: 'label',
-                                                         value: selectedCategory"></select>
```
pattisdr: On second thought, I would restore this to how it was before with everything in one line.

pattisdr:
Okay, @NehaTelhan, first pass complete! :sunny: 

Just requesting minor changes in formatting.  Actual functionality checks out. Well done!

pattisdr:
These changes look good to me.  Moving it to ready to merge @sloria.

website/static/js/pages/project-settings-page.js
```diff
@@ -227,7 +227,13 @@ $(document).ready(function() {
       var unchecked = checkedOnLoad.filter('#selectAddonsForm input:not(:checked)');
 
       if(unchecked.length > 0 || checked.length > 0) {
-        return 'The changes on addon setting are not submitted!';
+          return 'The changes on addon setting are not submitted!';
+      }
+    /* Before closing the page, Check whether changes made to category, title or description are updated or not */
+      if (projectSettingsVM.title() !== projectSettingsVM.titlePlaceholder ||
```
sloria: For better cohesion, this logic should be moved to the `ProjectSettings` "class". You can put the `beforeunload` handler into the constructor of the class.

sloria:
Pass finished. Just one minor refactor suggested in https://github.com/CenterForOpenScience/osf.io/pull/5162#discussion_r55842364, and this should be good to merge.

website/static/js/pages/project-settings-page.js
NehaTelhan: Hi, Extern here. I'm not sure I'm understanding what you mean here. I tired a couple things though. 
Right now I made a function in the project class as shown below:

`        navigateAway: function() {
            if (projectSettingsVM.title() !== projectSettingsVM.titlePlaceholder ||
                projectSettingsVM.description() !== projectSettingsVM.descriptionPlaceholder ||
                projectSettingsVM.selectedCategory() !== projectSettingsVM.categoryPlaceholder) {
                return 'There are unsaved changes in your project settings.';
            }
        },`

I then call this function with `ProjectSettings.navigateAway();` in the place this check was previously, within the 'beforeunload' function.
As the functionality is no longer what it should be, I don't think I quite understand what you mean. Can you elaborate some? (I'm also not opposed to talking directly!)

sloria: The change is simpler than you think =) .

This code needs to be relocated to the `ProjectSettings` constructor in `projectSettings.js`. You can add `$(window).on('beforeunload',function() {// your code here}`  there.

NehaTelhan: That's what I thought initially but when I'm testing this out on the localhost, the functionality of it is no longer working.

sloria: Ah ok; no big deal. This is fine as is. :+1:

PR #5164, opened 2016-03-09
Fix the project overview page contributor error

chennan47:
<b>Purpose</b>
Fix the error that the contributors are not showing correctly in project overview page when it is a public project visited under anonymous private link.

Before fix:
screen shot 2016-03-09 at 6 48 06 pm

After fix:
screen shot 2016-03-09 at 6 48 31 pm

Partially fix https://openscience.atlassian.net/browse/OSF-5824

### DIFF
--- a/website/templates/project/project.mako
+++ b/website/templates/project/project.mako
@@ -111,7 +111,7 @@
                     Contributors:
                 % endif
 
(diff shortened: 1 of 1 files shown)


sloria:
Thanks for responding so quickly.

PR #5165, opened 2016-03-10
[Feature] Prevents enter key from refreshing the page within the Create Bucket bootbox

TomHeatwole:
Jira ticket: https://openscience.atlassian.net/browse/OSF-5553

When users enter a name in the S3 create bucket modal, if they hit the "return" key, the modal is dismissed.  This is not the expected behavior.  According to @billyhunt, since this form should be refactored, and fixing the behavior in a timely manner is important, this fix should be put in first to correct the unintended behavior.


sloria:
Sweet!

PR #5166, opened 2016-03-10
Improvement onto icon images: Added ARIA labels to social media icons

sf2ne:
# Purpose
Currently when VoiceOver reads the social media icons in the footer of the OSF pages it just reads "link" for each of them. It has no idea what they are and to a blind or visually impaired person this could be a major problem. This PR allows VoiceOver to read what social media sites the icons represent and relay that information to the user.

# Changes
Adds ARIA labels to the five icons listed under the header "Socialize" in the footer of the OSF. The labels simply state exactly what social media website the icon represents. For example, the icon for twitter has the ARIA label : "Twitter" . An attached image showcases what this looks like in VoiceOver. Additionally, as the label is an ARIA label it would not be visible to sighted users and would only become audible and visible to users using VoiceOver to access the OSF.

# Side Effects
As only 5 lines of code were changed and of those changes only ARIA labels were added, no side effects are anticipated.

screen shot 2016-03-10 at 10 12 54 am

### DIFF
--- a/website/templates/footer.mako
+++ b/website/templates/footer.mako
@@ -29,11 +29,11 @@
             </div>
             <div class="col-sm-3 col-md-2">
                 <h4>Socialize</h4>
-                <a href="http://twitter.com/OSFramework"><i class="fa fa-twitter fa-2x"></i></a>
-                <a href="https://www.facebook.com/OpenScienceFramework"><i class="fa fa-facebook fa-2x"></i></a>
-                <a href="https://groups.google.com/forum/#!forum/openscienceframework"><i class="fa fa-group fa-2x"></i></a>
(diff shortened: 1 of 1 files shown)


sloria:
Sweet!

GaryKriebel:
Was this associated with a JIRA issue?

sf2ne:
This was not associated with a JIRA Issue that I know of.

GaryKriebel:
I created the following JIRA issue to have it follow the dev process and be available for QA's queue.

https://openscience.atlassian.net/browse/OSF-5856

GaryKriebel:
I wouldn't worry about it.  We normally have the issue in JIRA before the pull request and it is in the commit message where you would mention the JIRA number.  

I commented with the JIRA number (as well as commenting on JIRA with the PR number) for clarity with QA.

PR #5169, opened 2016-03-10
Feature: remove Node.commented_files

samanehsan:
Purpose
------------
The `commented_files` field is used to easily calculate the total unread file comments, but updating the field causes the Node to lock. Since the total unread file comments are not being displayed anywhere, this code can be removed for now. 

Changes
------------
- Remove all usages of `Node.commented_files`
- Simplify `Comment.find_n_unread`
- Address code review comments from #5059, mainly minor refactoring


sloria:

PR #5172, opened 2016-03-10
[Feature] Remove project settings addons apply button

TomHeatwole:
On the project settings page, under the checkbox list of addons, there existed an 'Apply' button which was inconvenient for users. This feature removes the apply button and updates the page when the checkboxes are clicked.

Jira ticket: https://openscience.atlassian.net/browse/OSF-5854

## Before
alt text

## After
alt text

### DIFF
--- a/website/static/js/pages/project-settings-page.js
+++ b/website/static/js/pages/project-settings-page.js
@@ -245,6 +245,8 @@ $(document).ready(function() {
                     callback: function(result) {
                         if (!result) {
                             $(that).attr('checked', false);
+                        } else {
+                            $('#selectAddonsForm').submit();
                         }
                     },
                     buttons:{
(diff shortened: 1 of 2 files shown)


website/static/js/pages/project-settings-page.js
```diff
@@ -245,6 +245,8 @@ $(document).ready(function() {
                     callback: function(result) {
                         if (!result) {
                             $(that).attr('checked', false);
+                        } else {
+                            $('#selectAddonsForm').submit();
```
billyhunt: To me, calling the selector seems unnecessary.  I might change the jQuery submit listener into a function.  

something like   
```
function submitBucket() {
```

then call submitBucket()

billyhunt:
Looks good.  Nice work @TomHeatwole  :tada: 

RFM

PR #5173, opened 2016-03-10
[feature] OSF4M Analytics Links

caseyrollins:
- Add link to OSF4M success email which redirects to a Google Analytics link.
- [OSF-4998]

### DIFF
--- a/website/templates/emails/conference_submitted.txt.mako
+++ b/website/templates/emails/conference_submitted.txt.mako
@@ -21,9 +21,9 @@ Get more from the OSF by enhancing your project with the following:
 * Collaborators/contributors to the submission
 * Charts, graphs, and data that didn't make it onto the submission
 * Links to related publications or reference lists
-* Connecting other accounts, like Dropbox, Google Drive, GitHub, figshare and Mendeley via add-on integration. Learn more and read the full list of available add-ons: http://osf.io/getting-started/#addons
+* Connecting other accounts, like Dropbox, Google Drive, GitHub, figshare and Mendeley via add-on integration. Learn more and read the full list of available add-ons: http://osf.io/getting-started/email#addons
 
-To learn more about the OSF, visit: http://osf.io/getting-started
+To learn more about the OSF, visit: http://osf.io/getting-started/email
 
 
 Sincerely yours,
--- a/website/views.py
+++ b/website/views.py
@@ -366,3 +366,7 @@ def redirect_about(**kwargs):
 
 def redirect_howosfworks(**kwargs):
     return redirect('/getting-started/')
+
+
+def redirect_meetings_analytics_link(**kwargs):
+    return redirect('/getting-started/?utm_source=osf4m&utm_medium=email&utm_campaign=success')
--- a/website/routes.py
+++ b/website/routes.py
@@ -198,6 +198,7 @@ def make_url_map(app):
         Rule('/howosfworks/', 'get', website_views.redirect_howosfworks, json_renderer,),
         Rule('/faq/', 'get', {}, OsfWebRenderer('public/pages/faq.mako')),
         Rule('/getting-started/', 'get', {}, OsfWebRenderer('public/pages/getting_started.mako')),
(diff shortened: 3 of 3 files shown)


website/views.py
```diff
@@ -366,3 +366,7 @@ def redirect_about(**kwargs):
 
 def redirect_howosfworks(**kwargs):
     return redirect('/getting-started/')
+
+
+def redirect_meetings_analytics_link(**kwargs):
+    return redirect('/getting-started/?utm_source=osf4m&utm_medium=email&utm_campaign=success')
```
whit537: Do we really want the querystring here?

whit537:
https://openscience.atlassian.net/browse/OSF-4998

website/views.py
whit537: Yes, per https://openscience.atlassian.net/browse/OSF-4998:

> On our side, the links would redirect to:
>
http://osf.io/getting-started/?utm_source=osf4m&utm_medium=email&utm_campaign=success#addons
>
and
>
http://osf.io/getting-started/?utm_source=osf4m&utm_medium=email&utm_campaign=successadd#addons
>
respectively.

whit537: I'll assume for the time being that we don't actually want the `#addons` fragment in the first case, which is the one I'm seeing here. So this is fine. :)

website/templates/emails/conference_submitted.txt.mako
```diff
@@ -21,9 +21,9 @@ Get more from the OSF by enhancing your project with the following:
 * Collaborators/contributors to the submission
 * Charts, graphs, and data that didn't make it onto the submission
 * Links to related publications or reference lists
-* Connecting other accounts, like Dropbox, Google Drive, GitHub, figshare and Mendeley via add-on integration. Learn more and read the full list of available add-ons: http://osf.io/getting-started/#addons
+* Connecting other accounts, like Dropbox, Google Drive, GitHub, figshare and Mendeley via add-on integration. Learn more and read the full list of available add-ons: http://osf.io/getting-started/email#addons
```
whit537: As this PR currently stands, this link will fold down into the base `/getting-started/email` redirect, but [OSF-4998](https://openscience.atlassian.net/browse/OSF-4998) calls for a second redirect specifically for add-ons. Either we should implement the redirect in JavaScript so we can switch on the fragment as originally specified, or we should drop back to `/getting-started/email-addons` and implement the second redirect in Python instead.

whit537:
##### Punchlist—for @caseyrollins 

 - [ ] implement second redirect for addons—https://github.com/CenterForOpenScience/osf.io/pull/5173#discussion_r55923411

whit537:
Okay, @caseyrollins. I made a (short) [punchlist](https://github.com/CenterForOpenScience/osf.io/pull/5173#issuecomment-195769178)! Ball is in your court. :)

website/views.py
caseyrollins: Ah, I didn't notice that the second link was slightly different! Adding second redirect now.

caseyrollins: @whit537, I'm double checking with @saradbowman to see if the analytics links given in the JIRA ticket are correct. I feel like `successadd#addons` at the end of the second link should just be `success#addons`, in which case a second redirect wouldn't be necessary. I'm also assuming that the first link shouldn't have `#addons` at the end.

caseyrollins:
@whit537, I've updated the JIRA ticket with the correct redirect URLs after confirming with @saradbowman that there were some typos. Since the links have the same base URL, I don't think a second redirect is necessary because the browser correctly handles `http://osf.io/getting-started/email#addons` redirecting to `http://osf.io/getting-started/?utm_source=osf4m&utm_medium=email&utm_campaign=success#addons` with the existing redirect function.

whit537:
Circling back here ...

whit537:
@caseyrollins Okay, as long as @saradbowman et al. can get the tracking differentiation they need from `utm_campaign=success` vs. `utm_campaign=success#addons` (i.e., if their metrics toolchain can handle a fragment) then I'm good to go here.

:+1:

PR #5175, opened 2016-03-11
[#OSF-5856] Added ARIA Label for youtube link

sf2ne:
# Purpose
Currently when VoiceOver reads the YouTube link to the video "Getting Started With the Open Science Framework" it just reads "link" for it. It has no idea what that is and to a blind or visually impaired person this could be a major problem. This PR adds an ARIA label to the link so that a blind/visually impaired user would be able to access it easily. This link is found on the home page of the OSF. It is the link located on the screen shot of the OSF with the YouTube play button over it. This can more clearly be seen in the before and after images.

# Changes
Adds an ARIA label to the YouTube link to "Getting Started With the Open Science Framework". This label shows up in the links menu of VoiceOver and reads out "OSF YouTube Video" when the user scrolls to it. This label is not visible to sighted users. Attached are before and after images of how this appears in the VoiceOver menu.

# Side Effects
As only one line of code was changed no side effects are anticipated. 

## Before Image
screen shot 2016-03-11 at 10 10 10 am

## After Image
screen shot 2016-03-11 at 10 13 39 am

### DIFF
--- a/website/templates/index.mako
+++ b/website/templates/index.mako
@@ -38,7 +38,7 @@
         <div id="hero-signup" class="container">
           <div class="row">
             <div class="col-sm-6 hidden-xs">
-              <a class="youtube" href="//www.youtube.com/watch?v=2TV21gOzfhw"><i class="icon icon-play"></i></a>
+              <a class="youtube" href="//www.youtube.com/watch?v=2TV21gOzfhw" aria-label="OSF YouTube Video"><i class="icon icon-play"></i></a>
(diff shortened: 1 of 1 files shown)


sloria:
Thanks!

sloria:
By the way, if you are adding ARIA labels without making any other changes (e.g. to backend code), go ahead and add `[skip ci]` to your commit message before you push. That will prevent the PR from triggering a TravisCI build for simple markup changes like this.

PR #5177, opened 2016-03-11
[Ready to Review] Feature/dashboard event tracking

pattisdr:
# Purpose

Issue https://openscience.atlassian.net/browse/OSF-5721
Adds high level Google Analytics event tracking to new dashboard to gauge what features are being used.

# Changes

<b>Tracks a combined 96 events in project organizer/dashboard and home page.</b>  All high level tracking, basically clicks, if users filtered, etc.  No information being tracked about specific users.

Event tracking format looks like this: 
ga('send', 'event', [eventCategory], [eventAction], [eventLabel], [eventValue], [fieldsObject]);

Using eventCategory, eventAction, eventLabel.  Can explore sending in eventValue with appropriate fields (must be an integer).

Sample event: 
   `$osf.trackClick('recentActivity', 'filter', 'filter-on-files');`

This indicates that someone within the recent activity widget clicked on the word "files" to filter log actions by files.

# For the reviewer
Clicking on some links trigger the same actions and I've chosen to group them together.  For example, there are two places where you can click to add a component on the dashboard. I've grouped them together. Going through the events and determining if there is a better way to group them would be important.

Also, the vast majority of the events that can be tracked are here - but are there any missing that you think we need?


pattisdr:
Moved old PR https://github.com/caneruguz/osf.io/pull/31

website/static/js/home-page/newAndNoteworthyPlugin.js
```diff
@@ -117,7 +117,9 @@ var NewAndNoteworthy = {
         }
 
         function findMoreProjectsButton () {
-            return m('a.btn.btn-default.m-v-lg', {type:'button', href:'/search'}, 'Search for more projects');
+            return m('a.btn.btn-default.m-v-lg', {type:'button', href:'/search', onclick: function() {
+                $osf.trackClick('newAndNoteworthy', 'navigate', 'navigate-to-search-for-more-projects');
```
samanehsan: As discussed over flowdock, please change `newAndNoteworty` to `discoverPublicProjects` so that it's clear where these logged events are happening.

samanehsan:
Done with CR1 @pattisdr! I tested out most of the tracked events and it looks good to me! :sunny:

website/static/js/addProjectPlugin.js
```diff
@@ -86,12 +86,15 @@ var AddProject = {
             self.isValid(false);
         };
     },
-    view : function (ctrl, options) {
+    view : function (ctrl, options, trackingCategory, trackingAction) {
```
sloria: This is counter to the usual signature of component `view` methods, which is `view(ctrl, options, children)` (where `children` is an optional array of mithril elements, as in `m.component(MyPanel, {size: 'large'}, [m('span',{}, 'Hello world')])`.

Better to make `trackingCategory` and `trackingAction` part of the `options` object. This makes for better readability when using the component. Compare

```js
// What do 'foo' and 'bar' mean?
m.component(AddProject, {someParam: 42}, 'foo', 'bar')
```

vs.

```js
m.component(AddProject, {someParam: 42, trackingCategory: 'foo', trackingAction: 'bar'})
```

website/static/js/logTextParser.js
```diff
@@ -72,7 +72,7 @@ var returnTextParams = function (param, text, logObject) {
 };
 
 var LogText = {
-    view : function(ctrl, logObject) {
+    view : function(ctrl, logObject, trackingLocation, action) {
```
sloria: See https://github.com/CenterForOpenScience/osf.io/pull/5177/files#r56344445 . Note: you may need to rename `action` to avoid clashing with `logObject.action`.

```js
var options = $.extend({}, logObject, {trackingLocation: ..., trackingAction: ...});
```

website/static/js/osfHelpers.js
```diff
@@ -871,6 +871,14 @@ var findContribName = function (userAttributes) {
     }
 };
 
+var trackClick = function(category, action, label){
+    window.ga('send', 'event', category, action, label);
+    // TODO remove after event tracking has been reviewed
+    console.log(category, action, label);
```
sloria: This can be removed now.

website/static/js/project-organizer.js
```diff
@@ -364,10 +375,14 @@ var tbOptions = {
                     }
                     tb.filter(event);
                 },
-                onchange: m.withAttr('value', tb.filterText),
+                onchange: function() {
+                    m.withAttr('value', tb.filterText);
```
sloria: Does this still work, i.e. is `filterText` still properly updated? As far as I understand, `m.withAttr` returns *a function*, which is why 

```js
onchange: m.withAttr('value', tb.filterText)
```

works. But now, you are explicitly declaring a callback function here, so I don't think this line is doing anything. I would expect you'd need to do

```js
onchange: function(e) {
    tb.filterText(e.value);
    // ...
}
```

though you should verify this.

sloria:
Pass finished.

- https://github.com/CenterForOpenScience/osf.io/pull/5177/files#r56346457 looks like a possible bug; needs verification
- https://github.com/CenterForOpenScience/osf.io/pull/5177#discussion_r56344445 and https://github.com/CenterForOpenScience/osf.io/pull/5177#discussion_r56344953 are nice refactors but are not a hard requirement. If you think they'll take < 10 mins, go ahead and add them to this PR. If not, they can be done in a follow-up PR.

website/static/js/logTextParser.js
pattisdr: Where do I put this var options?

website/static/js/project-organizer.js
pattisdr: I do not understand what the `m.withAttr('value', tb.filterText);` line is doing.  If I delete it completely nothing seems to change.  Can you explain what it is for?

Search is happening `onkeyup` and if inside the onchange I log `tb.filterText()` and `this.value`, both are the same thing.  If I type the word "carrot" in the search bar and press enter, both this.value and tb.filterText() are equal to 'carrot' without the  `m.withAttr('value', tb.filterText);` line.

pattisdr: There is also value: tb.filterText().

sloria: `m.withAttr` is helper for returning callback functions.

```js
onchange: m.withAttr('value', someProp)
```

is equivalent to

```js
onchange: function(e) { someProp(e.value }
```

m.withAttr returns a function (but does not call it), which is why this line looked strange to me. However, I'm now wondering if this line is even necessary, given that `filterText` is already updated in an `onkeyup` handler.

pattisdr: Yes, I was thinking the same thing, search doesn't seem to be affected if this line is present or not.

sloria: ...so if filterText is updated in `onkeyup`, shouldn't the `trackClick` call go there too?

website/static/js/project-organizer.js
```diff
@@ -364,10 +375,14 @@ var tbOptions = {
...
+                    $osf.trackClick('myProjects', 'filter', 'search-projects');
```
sloria: We probably don't want to track *every* change--the user may be typing a lot here, and we'll get a lot of noise. You can use the [debounce](https://github.com/CenterForOpenScience/osf.io/blob/54418bc5fbce98fdb892267a27a93a560b06433a/website/static/js/osfHelpers.js#L347) function to make sure that trackClick only gets called after the user stops typing for some time interval (e.g. 500ms).

website/static/js/project-organizer.js
pattisdr: I think onchange is more useful - I am tracking it that way in quicksearch too.  I just want to know that they searched. If I am searching my name, and I type 'D-A-backspace-A-W-N'.   There will be 6 trackClicks fired, D, DA, D, DA, DAW, DAWN.  In early testing I had it onkeyup, and I feel tracking it this way inflates the number beyond where it can be useful.   Onchange here will just send one trackClick.

website/static/js/project-organizer.js
pattisdr: Continuation of above.  Onchange does not track every change.  That is one reason why I went with it.  Onchange occurs when the element loses focus, after the content has been changed.

sloria: Ah, ok that makes sense. Thanks for the clarification.

website/static/js/addProjectPlugin.js
```diff
@@ -135,7 +145,10 @@ var AddProject = {
                                             name: 'projectCategory',
                                             value: cat.value,
                                             checked: ctrl.newProjectCategory() === cat.value,
-                                            onchange : m.withAttr('value', ctrl.newProjectCategory)
+                                            onchange : function() {
+                                                m.withAttr('value', ctrl.newProjectCategory);
```
sloria: As suggested in a previous comment, I think this needs to be 

```js
ctrl.newProjectCategory(event.value)
```

website/static/js/project-organizer.js
sloria: Yep, I had a poor understanding of when `onchange` was fire. You're right; the `trackClick` belongs here. So I think the only change to make is to replace the `withAttr` with `tb.filterText(event.value)` .

website/static/js/addProjectPlugin.js
pattisdr: Okay, changing this now.

sloria:
I'll fix the withAttr calls and resolve merge conflicts.

pattisdr:
@sloria, I already did it! Pushing now.

sloria:
Doh; duplicated the work in 3e06a7bf5ed . Sorry bout that. In any case, this is merged!

pattisdr:
Hooray! Now I won't have to keep bringing this up to date!

PR #5178, opened 2016-03-11
[Ready for Review] Resolves misspointed url for registration contributors

cwisecarver:
The urls for registration contributors were pointing to /v2/node/{node_id}/contributors/{contributor_id}/ which did not resolve. This changes them to point to /v2/registrations/{registration_id}/contributors/{contributor_id}/.

In prod the only issue would be that links.self would have been pointing at the wrong place. In staging, with varnish esi enabled, the issue would be that registration contributors would have become `{"errors":[{"title":"Not Found"}]}`.

@caneruguz This should fix your problem.


sloria:
Good to merge when the build fails

PR #5181, opened 2016-03-11
[#OSF-4722] [Ready to Review] Fixed Add Button and Check Button Sizes

sf2ne:
JIRA Ticket: https://openscience.atlassian.net/browse/OSF-4722

# Purpose
Currently when editing one of the projects for which the user has admin permission. In the project's Contributor's page when you try to add a contributor the plus button has a larger width than the minus button does. This PR makes the size of the buttons the same. This is more easily visible in the before and after images.

# Changes
Takes out the section in the plus button code where the width was fixed as the width of the minus button was not fixed. 

# Side Effects
No side effects anticipated. 

## Before
unnamed
unnamed-1

The buttons are different sizes.

## After
screen shot 2016-03-11 at 3 04 41 pm
screen shot 2016-03-11 at 3 05 00 pm

The buttons are now the same size.

### DIFF
--- a/website/templates/project/modal_add_contributor.mako
+++ b/website/templates/project/modal_add_contributor.mako
@@ -62,12 +62,12 @@
                                                     data-bind="visible: !contributor.added,
                                                                click:$root.add,
                                                                tooltip: {title: 'Add contributor'}"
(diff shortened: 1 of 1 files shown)


sloria:
Thank!

PR #5183, opened 2016-03-11
Feature: Add wiki comments to v2 API

samanehsan:
Fixes JIRA ticket OSF-5240 -- Adds ability to comment on wiki pages via the v2 API. A follow-up PR will include the front-end changes to enable wiki comments on the OSF.

<h3>TODO:</h3>

- [x] API: List, create, view, edit, delete, undelete, report, unreport
- [x] API Tests
- [x] Fix embedding comment targets
- [x] Logging
- [x] Notification emails


api/base/serializers.py
```diff
@@ -541,11 +545,13 @@ def resolve(self, resource):
         """
         Resolves the view for target node or target comment when embedding.
         """
-        view_info = self.view_map.get(resource.target._name, None)
+        view_info = self.view_map.get(resource.target.referent._name, None)
```
mfraezz: Could this have unintended consequences for `TargetFields` that don't target a comment?

mfraezz:
>Adds ability to comment on wiki pages via the v2 API.

Does this mean there are intentionally no UI changes to the wiki -- that comments can only be added/modified via APIv2?

api_tests/comments/views/test_comment_detail.py
```diff
@@ -524,18 +465,45 @@ def test_public_node_view_only_link_user_cannot_see_deleted_comment(self):
...
+        expected_type = 'nodes'
+        assert_equal(res.status_code, 200)
+        assert_equal(urlparse(url).path, expected_url)
+        assert_equal(target_type, expected_type)
+
+    def test_public_node_non_contributor_commenter_can_update_comment(self):
+        project = ProjectFactory(is_public=True, comment_level='public')
+        comment = CommentFactory(node=project, user=self.non_contributor)
+        url = '/{}comments/{}/'.format(API_BASE, comment._id)
+        payload = {
```
mfraezz: Can you not use `self._set_up_payload` here?

api/base/serializers.py
samanehsan: A comment's target should always be a `Guid` object -- it's the `Guid.referent` that would refer to the comment, node or file

mfraezz: Right. I guess my question could be better phrased as "Will `TargetField` ever be used for anything other than comments?"

samanehsan: I don't think so -- I wrote it specifically for comment targets

api_tests/comments/views/test_comment_detail.py
samanehsan: I made this change in the new PR!

samanehsan:
Replaced by #5298

PR #5184, opened 2016-03-14
Disable commenting on draft preregistration if approved

barbour-em:
## Purpose

Disable commenting on draft preregistration if approved.
[OSF-5445](https://openscience.atlassian.net/browse/OSF-5445)

## Changes

Check if the registration is approved and remove some elements if it is.

## Side effects

None

### DIFF
--- a/website/project/model.py
+++ b/website/project/model.py
@@ -4495,6 +4495,9 @@ class DraftRegistration(StoredObject):
         return draft
 
     def update_metadata(self, metadata):
+        if self.is_approved():
+            return False
+
         changes = []
         for question_id, value in metadata.iteritems():
             old_value = self.registration_metadata.get(question_id)
--- a/website/project/model.py
+++ b/website/project/model.py
@@ -4471,7 +4471,7 @@ class DraftRegistration(StoredObject):
             else:
                 return self.approval.is_approved
         else:
-            return True
+            return False
 
     @property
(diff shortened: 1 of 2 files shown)


sloria:
@barbour-em Is commenting also disabled on the backend?

barbour-em:
@sloria I'm not 100% on the logic but I think yes. If a registration is approved, it goes through the registration/archiving process and can't be updated. The only way for new comments to be added on the backend is through [DraftRegistration#update_metadata](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/project/model.py#L4497-L4520) which shouldn't get called after approval.

The way I understand it is that the buttons this PR removes just wouldn't do anything once the Draft is approved. That being said, it probably wouldn't hurt to check for approval in #update_metadata.

sloria:
Great; thanks for making that change. This is good to merge when the build is passing.

barbour-em:
@sloria build is (finally) passing

website/project/model.py
```diff
@@ -4495,6 +4495,9 @@ def create_from_node(cls, node, user, schema, data=None):
         return draft
 
     def update_metadata(self, metadata):
+        if self.is_approved:
+            return {}
```
sloria: This method typically returns a `list`, so this should be `[]`. I can make this change when I merge.

PR #5185, opened 2016-03-14
[Ready to Review] Dashboard/quick search while loading

pattisdr:
Issue https://openscience.atlassian.net/browse/OSF-5867

# Purpose

Currently, users must wait until all projects are loaded before they can begin to search.   This can be a long wait if the users have many projects.

# Changes

Allows user to search for projects before all projects are loaded in the background.  As new search results are found, they will be added to the list.  Sorting is still prevented until all nodes are loaded.

Also, adds ball-scale, instead of spinner while loading.  Just an experiment - we need to reach an agreement on the loader.
out

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "list-of-licenses": "git://github.com/samchrisinger/list-of-licenses.git#1f1d9ca8a9aba56d57fca6e27df80b8298048cc8",
     "livedb": "~0.4.8",
     "livedb-mongo": "~0.4.1",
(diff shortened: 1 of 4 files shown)


sloria:
Displaying the loading indicator above the search bar seems unnecessary (and may appear like a bug)--the user doesn't need to know that we are loading more projects in the background unless they actually perform a search. 

Perhaps it makes sense to show a loading indicator under the filtered results while more results are being fetched.

@caneruguz Can you work with Dawn to come up with an appropriate UI for this?

PR #5193, opened 2016-03-14
Feature: load indicator for projects

caneruguz:
## Purpose
Adds a load indicator at top of project organizer that loads initial projects and keeps loading rest on background to avoid long render times.


sloria:
Every load resets the scroll, which makes navigation difficult:

scroll

Can you retain scroll position while items are loading?

website/static/js/project-organizer.js
```diff
@@ -372,6 +375,11 @@ var tbOptions = {
                 $('.db-poFilter>input').val('');
             } }, tb.options.removeIcon())
         ];
+
+        if(tb.options.loadValue() < 100){
```
sloria: I would still like to see this enabled, even if all the projects aren't loaded. Now that there is a loading indicator, is there any reason to keep this disabled?

website/static/js/myProjects.js
```diff
@@ -1329,7 +1344,6 @@ module.exports = {
     MicroPagination : MicroPagination,
     ActivityLogs : ActivityLogs,
     LinkObject: LinkObject,
-    PrivateFunctions : {
-        makeTree : _makeTree
-    }
+    PrivateFunctions : {}
```
sloria: Can this be removed?

sloria:
Pass finished. Main blocker is https://github.com/CenterForOpenScience/osf.io/pull/5193#issuecomment-196837728

PR #5195, opened 2016-03-15
[Ready for Review] Fix comment cache busting

cwisecarver:
Comments have a target and a root target that have individual urls that are cached. When a comment is added/removed/modified now the target and root target urls are being banned in varnish along with the comment url itself.

Risks:
There are some occasions where urls are being banned multiple times in the same request. For instance, when a comment is created the node url is banned once because the node is being saved and a second time because the node is the target and root target of the comment. I'm currently creating a set of urls to be banned per object, that eliminates some duplication. I made a todo to pull the url generation up into the actual postcommit_handler queue so that there could be a set of urls per request.

@sloria this should fix the comment caching issues that @nicipfeiffer has been seeing on stage.


main.py
```diff
@@ -13,7 +13,7 @@
 from website import settings  # noqa
 from website.app import init_app  # noqa
 
-app = init_app('website.settings', set_backends=True, routes=True)
+application = app = init_app('website.settings', set_backends=True, routes=True)
```
cwisecarver: This just enables uwsgi compatibility, it looks for something called "application" by default. It shouldn't have any effect on anything else.

tests/base.py
```diff
@@ -73,6 +73,7 @@ def get_default_metaschema():
     'website.mails',
     'website.search_migration.migrate',
     'website.util.paths',
+    'api.caching.tasks'
```
cwisecarver: This silences the logger which will break travis.

PR #5196, opened 2016-03-15
OSF-5179 Strange wiki compare behavior.

adlius:
<h3>
Purpose
</h3>
The original behavior of the compare function of project wiki was erratic in that it was not clear whether it was comparing texts word by word, line by line or otherwise. This PR make use of new features in diff 2.2.2 and compare wiki texts line by line.

<h3>
Changes
</h3>
Update the package.json file to use newest version of diff (2.2.2).
Change diffTool.js. Specifically, the custom wikiDiff object was deleted and make use of diffLines() in diff 2.2.2.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "bower": "^1.3.12",
     "c3": "^0.4.10",
     "css-loader": "^0.9.1",
-    "diff": "~1.2.2",
+    "diff": "~2.2.2",
     "dropzone": "git://github.com/sloria/dropzone.git#accept-directory",
     "exports-loader": "^0.6.2",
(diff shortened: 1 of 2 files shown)


mfraezz:
I wouldn't consider it a blocker because this code is already untested, but it would be nice to have a few front-end tests to make sure `diffTool` is working.

Otherwise, LGTM. Passing this to RTM :fire:

sloria:
Thanks!

In the future, you only need to include the JIRA ID once in a commit. For example, your first commit would contain "[OSF-5179]". It's also very helpful to link the JIRA ticket in the PR description.

PR #5198, opened 2016-03-15
Feature/password changing security updates

brianjgeiger:
Purpose
-----------
Fixes [sec-3] and [sec-4] by way of [https://openscience.atlassian.net/browse/OSF-5874]. Essentially, resetting a user's password should warn the user and should log out all existing sessions. 

Changes
------------

1. Test for removing sessions when changing password
2. Test for emailing user when changing password
3. Add a stub email template (real email waiting for feedback from community)
4. Move save of user to the change password model
5. Check to verify that this isn't the user's first time adding a password. If it's a change, then email the user and delete existing sessions

Side effects
----------------
Shouldn't be any, although I seem to have to log in twice, especially after the sessions are cleared. I'm not yet sure if that's something I did somehow or if it's just the way the OSF is being right now; possibly something in my local setup.

### DIFF
--- a/framework/auth/core.py
+++ b/framework/auth/core.py
@@ -657,20 +657,21 @@ class User(GuidStoredObject, AddonModelMixin):
         return '{base_url}user/{uid}/{project_id}/claim/?token={token}'\
                     .format(**locals())
 
-    def set_password(self, raw_password):
+    def set_password(self, raw_password, suppress_notification=False):
         """Set the password for this user to the hash of ``raw_password``.
         If this is a new user, we're done. If this is a password change,
         then email the user about the change and clear all the old sessions
         so that users will have to log in again with the new password.
 
         :param raw_password: the plaintext value of the new password
+        :param suppress_notification: Only meant for unit tests to keep extra notifications from being sent
         :rtype: list
         :returns: Changed fields from the user save
         """
         had_existing_password = bool(self.password)
         self.password = generate_password_hash(raw_password)
         return_value = self.save()
-        if had_existing_password:
+        if had_existing_password and not suppress_notification:
             mails.send_mail(
                 to_addr=self.username,
                 mail=mails.PASSWORD_RESET,
--- a/framework/auth/core.py
+++ b/framework/auth/core.py
@@ -658,8 +658,27 @@ class User(GuidStoredObject, AddonModelMixin):
                     .format(**locals())
 
     def set_password(self, raw_password):
-        """Set the password for this user to the hash of ``raw_password``."""
+        """Set the password for this user to the hash of ``raw_password``.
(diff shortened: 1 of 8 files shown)


sloria:
This is similar to #5000 (logging out a user before they confirm an account merge). @billyhunt Were you aware of the `remove_sessions_for_user` function? I haven't reviewed your PR yet, but it seems to be the correct way to log users out. Just want to make sure the approaches are consistent.

framework/auth/core.py
```diff
@@ -657,9 +657,29 @@ def get_claim_url(self, project_id, external=False):
         return '{base_url}user/{uid}/{project_id}/claim/?token={token}'\
                     .format(**locals())
 
-    def set_password(self, raw_password):
-        """Set the password for this user to the hash of ``raw_password``."""
+    def set_password(self, raw_password, suppress_notification=False):
```
sloria: Minor: parameter names that result in double negatives are confusing. How about `notify=True` instead?

website/templates/emails/notify_base.mako
```diff
@@ -60,7 +60,7 @@
                     <tbody>
                         <tr>
                             <td style="border-collapse: collapse;">
-                                <p class="text-smaller text-center" style="text-align: center;font-size: 12px;">Copyright &copy; 2015 Center For Open Science, All rights reserved.
+                                <p class="text-smaller text-center" style="text-align: center;font-size: 12px;">Copyright &copy; 2016 Center For Open Science, All rights reserved.
```
sloria: Good catch.

framework/auth/core.py
```diff
@@ -657,9 +657,29 @@ def get_claim_url(self, project_id, external=False):
...
+        so that users will have to log in again with the new password.
+
+        :param raw_password: the plaintext value of the new password
+        :param suppress_notification: Only meant for unit tests to keep extra notifications from being sent
+        :rtype: list
+        :returns: Changed fields from the user save
+        """
+        had_existing_password = bool(self.password)
         self.password = generate_password_hash(raw_password)
+        return_value = self.save()
```
sloria: We generally avoid implicit saves. I think this can be removed.

sloria:
Pass finished.

framework/auth/core.py
brianjgeiger: The reason I added it in for this case is to ensure that someone doesn't receive a notification for a thing that, for whatever reason, doesn't happen. I can be swayed without too much effort, but there was a reason.

brianjgeiger:
Everything is responded to.

brianjgeiger:
Oh, and don't merge this before we get a proper email from Community. I'll make a note in the description; I almost forgot.

PR #5200, opened 2016-03-16
[Ready to Review] Dashboard/MyProjects Fixes

pattisdr:
# Purpose

Addresses some of the QA issues as logged in the Trello list
https://trello.com/c/zJqcgOL8/75-my-projects-page-spelling-grammar-capitalization
https://trello.com/c/LmHzuIgk/70-resolve-the-repetition-of-my-projects-title-vs-page-in-the-home-page
https://trello.com/c/NRNMvpHc/77-homepage-language-changes
https://trello.com/c/ZAvyPoee/78-info-panel-does-not-show-line-breaks-for-project-description

### DIFF
--- a/api/nodes/serializers.py
+++ b/api/nodes/serializers.py
@@ -72,8 +72,8 @@ class NodeSerializer(JSONAPISerializer):
     id = IDField(source='_id', read_only=True)
     type = TypeField()
 
-    category_choices = Node.CATEGORY_MAP.keys()
-    category_choices_string = ', '.join(["'{}'".format(choice) for choice in category_choices])
+    category_choices = Node.CATEGORY_MAP.items()
+    category_choices_string = ', '.join(["'{}'".format(choice[0]) for choice in category_choices])
 
     title = ser.CharField(required=True)
     description = ser.CharField(required=False, allow_blank=True, allow_null=True)
--- a/website/static/js/addProjectPlugin.js
+++ b/website/static/js/addProjectPlugin.js
@@ -87,15 +87,6 @@ var AddProject = {
         };
     },
     view : function (ctrl, options) {
-        var capitalizeCategories = function(category) {
-            if (category === 'methods and measures') {
-                return 'Methods and Measures';
-            }
-            else {
-                return category.charAt(0).toUpperCase() + category.slice(1);
-            }
-        };
-
         var templates = {
             form : m('.modal-content', [
                 m('.modal-header', [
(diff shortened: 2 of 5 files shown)


website/static/js/addProjectPlugin.js
```diff
@@ -87,6 +87,15 @@ var AddProject = {
         };
     },
     view : function (ctrl, options) {
+        var capitalizeCategories = function(category) {
```
sloria: We have a mapping of internal names to human-readable names here: https://github.com/CenterForOpenScience/osf.io/blob/54418bc5fbce98fdb892267a27a93a560b06433a/website/project/model.py#L707-L719 . 

I'd rather see a mapping like this used (even if it means duplicating it in this JS file) than this capitalize function.

The internal names (e.g. `materials and methods`) are meant to be stable; the human-friendly names may be subject to change (e.g. we might decide to display `Methodology/Materials`). Using a mapping allows us to make this change easily, and prevents unexpected bugs when we add new categories.

website/static/js/myProjects.js
```diff
@@ -946,7 +946,7 @@ var Collections = {
                         m('h3.modal-title', 'Add new collection')
                     ]),
                     body : m('.modal-body', [
-                        m('p', 'Collections are groups of projects that help you organize your work. After you create your collection you can add projects by dragging and dropping projects to the collection. '),
+                        m('p', 'Collections are groups of projects that help you organize your work. After you create your collection, you can add projects by dragging and dropping them to the collection. '),
```
sloria: Minor: One nicety would be to make "create your collection" a link that opens the "Create collection" modal. Also, "dragging and dropping them to the collection" can be simplified to "dragging them into the collection".

Neither of these are blockers, and can be done in a follow-up PR

sloria:
Pass finished.

website/static/js/myProjects.js
pattisdr: If they are reading this text, they are already in the 'create collection' modal.

api/nodes/serializers.py
```diff
@@ -72,8 +72,8 @@ class NodeSerializer(JSONAPISerializer):
     id = IDField(source='_id', read_only=True)
     type = TypeField()
 
-    category_choices = Node.CATEGORY_MAP.keys()
-    category_choices_string = ', '.join(["'{}'".format(choice) for choice in category_choices])
+    category_choices = Node.CATEGORY_MAP.items()
```
sloria: I think this will break validation the `category` field a few lines down because, it is expecting `choices` to be a list of strings.

pattisdr: Why? http://www.django-rest-framework.org/api-guide/fields/#choicefield
> choices - A list of valid values, or a list of (key, display_name) tuples.

Seems to me we weren't taking advantage of the fact that you could have both a value and a display name there.

pattisdr: OPTIONS request to /v2/nodes/ after this change:

screen shot 2016-03-16 at 3 26 50 pm

sloria: Won't the category field use the `choices` field for validation? We only expect to get the internal names as input.

sloria: Ohh, I didn't read the docs closely enough. OK, yeah, I think this is the right approach then. :+1:

PR #5201, opened 2016-03-16
Institutions 2.1

hmoco:


scripts/populate_institutions.py
```diff
@@ -55,7 +55,28 @@ def main(env):
...
+            'domain': ['staging-osf.cos.io'],
+            'description': None,
+            'email_domain': ['cos.io'],
+            'banner_name': 'cos-banner.png'
+        },
+        {
+            'name': 'University of Southern California',
+            '_id': 'usc',
+            'logo_name': 'usc-shield.jpg',
+            'auth_url': 'MICHAEL CHANGE THIS',
```
sloria: ping @icereval

website/static/js/institutionSignIn.js
```diff
@@ -18,11 +18,26 @@ var ViewModel = function() {
...
+            self.instNames(
+                response.data.map(function(item){
+                    var name = item.attributes.name;
+                    self.insts[name] = item.attributes.auth_url + '&target=' + encodeURIComponent(window.contextVars.institution_redirect);
+                    self.insts[item.id] = name;
+                    if (item.attributes.auth_url){
+                        return name;
+                    }
+                }).filter(function (item) {
+                   return typeof item !== 'undefined';
```
sloria: This is suspect; why would the response contain `undefined`?

website/static/js/institutionSignIn.js
```diff
@@ -18,11 +18,26 @@ var ViewModel = function() {
...
+                })
+            );
+            var inst_redirect = decodeURIComponent(decodeURIComponent(window.contextVars.institution_redirect));
```
sloria: As per COSDev, use `lowerCamelCase` for variable names.

website/static/js/institutionSignIn.js
```diff
@@ -18,11 +18,26 @@ var ViewModel = function() {
...
+            if (inst_redirect){
+                var inst_id = inst_redirect.split('institutions')[1];
+                inst_id = inst_id ? inst_id.split('/')[1] : false;
```
sloria: `lowerCamelCase` here, too

website/templates/institution.mako
```diff
@@ -8,17 +8,21 @@
     <div class="alert alert-info"><strong>NOTICE: </strong>Forks, registrations, and uploads will be temporarily disabled while the OSF undergoes a hardware upgrade. These features will return shortly. Thank you for your patience.</div>
 % endif
     <div id="inst">
-<div class="dashboard-header">
+<div class="dashboard-header" style="background: white">
```
sloria: Avoid inline styling. How about adding a new class to `dashboard-page.css`:

```css
.dashboard-header.dashboard-header-institution {
    background: none;
}
```

Then applying

```html
class="dashboard-header dashboard-header-institution"
```

website/static/js/institutionSignIn.js
sloria: Oh, I see now; it's possible that the `map` function will return `undefined`. How about doing the filter first?

```js
response.data.filter(function(item) { return item.attributes.auth_url;}).map(function(item) {
    // ...
})
```

hmoco: Response doesn't. Not returning anything on ` .map ` (which happens if the item does not have an ` auth_url `) adds an undefined object to the list. This is here to clean up those undefineds.

hmoco: Hah, we commented at the same time. Yep that change seems reasonable.

sloria: Updated above comment with more detail.

website/static/js/institutionSignIn.js
```diff
@@ -18,11 +18,26 @@ var ViewModel = function() {
...
         ).done(function (response) {
-            self.instNames(response.data.map(function(item){
-                var name = item.attributes.name;
-                self.insts[name] = item.attributes.auth_url + '&target=' + encodeURIComponent(window.contextVars.institution_redirect);
-                return name;
-            }));
+            self.instNames(
+                response.data.map(function(item){
+                    var name = item.attributes.name;
+                    self.insts[name] = item.attributes.auth_url + '&target=' + encodeURIComponent(window.contextVars.institution_redirect);
```
sloria: What if `auth_url` isn't defined?

website/project/views/node.py
```diff
@@ -249,7 +249,8 @@ def node_setting(auth, node, **kwargs):
 
     #check institutions:
     try:
-        inst = Institution.find_one(Q('email_domain', 'eq', auth.user.username.split('@')[1]))
+        email_domains = [email.split('@')[1] for email in auth.user.emails]
+        inst = Institution.find_one(Q('email_domain', 'in', email_domains))
```
sloria: This looks incorrect, given that `email_domain` is a list field: https://github.com/CenterForOpenScience/osf.io/blob/54418bc5fbce98fdb892267a27a93a560b06433a/website/project/model.py#L3384 .

website/views.py
```diff
@@ -79,15 +80,13 @@ def index():
     try:
         #TODO : make this way more robust
         inst = Institution.find_one(Q('domain', 'eq', request.host))
```
sloria: Do we store `domain` lower-cased? If so, you'll need to do `request.host.lower()`

sloria: This led me to find that `domain` is a [list field](https://github.com/CenterForOpenScience/osf.io/blob/54418bc5fbce98fdb892267a27a93a560b06433a/website/project/model.py#L3381). Therefore, it should probably be renamed to `domains`.

website/project/views/node.py
hmoco: Apparently this works. Have done (minor) testing, and Seto said he has mongo docs that support it.

sloria:
Pass finished.

hmoco:
@sloria have responded to comments.

website/project/views/node.py
sloria: OK. As suggested in a previous comment, `email_domain` should probably be renamed to `email_domains` to make it clear that it's a list.

website/static/js/institutionSignIn.js
```diff
@@ -18,11 +18,28 @@ var ViewModel = function() {
...
-            }));
+            self.instNames(
+                response.data.filter(function (item) {
+                    return item.attributes.auth_url;
+                }).map(function(item){
+                    return item.attributes.name;
+                })
+            );
+            response.data.forEach(function(item){
+                if (item.attributes.auth_url){
```
sloria: Minor: You're already filtering for data that has `auth_url` a few lines up; why not save that to a variable and re-use it?

```js
var validInstitutions = response.data.filter(...);
self.instNames(...map...)
validInstitutions.forEach(...)
```

website/project/model.py
```diff
@@ -3381,7 +3381,8 @@ def require_approval(self, user, notify_initiator_on_complete=False):
     institution_domain = fields.StringField(list=True)
```
sloria: As suggested in a previous comment, this should probably be changed to `institution_domains`

sloria:
Pass finished. Looks like https://github.com/CenterForOpenScience/osf.io/pull/5201/files#r56419747 is the only thing to be addressed.

PR #5204, opened 2016-03-16
remove redundant node text

billyhunt:
With the current text, the word component will be repeated 3 times, which, as I mentioned in the trello card, will invoke Beatlejuice.  No one wants that.  And its also redundant.  

Since the lack of components or projects is implied by the lack of components listed.  I propose we remove the text all together.

Before
---------
screen shot 2016-03-16 at 3 39 28 pm

After
-------

screen shot 2016-03-16 at 4 30 36 pm


sloria:
Yes!

PR #5205, opened 2016-03-16
Editing project description removes existing tags

kch8qx:
[OSF-5802]


sloria:
@kch8qx Since this is a bug and the fix is straight forward, can you promote this to a hotfix please? To do that, just create a new hotfix branch off of master and commit your change there.

sloria:
Also, please add a regression test. A new view test in this class should do the trick: https://github.com/CenterForOpenScience/osf.io/blob/debe8122162851c8817fd265ea3e8ad698b0cecc/api_tests/nodes/views/test_node_detail.py#L236

PR #5206, opened 2016-03-16
[Ready for Review] Add decaching to read comment count timestamp update view

cwisecarver:
This fixes the comment count not being decached when users open the comment slide-out by decaching the users' url. 

The core problem is that the URL that is used to retrieve the users comment count is inside the /v2/nodes namespace and the URL to set/reset the comment count is in the /v1/api/user namespace. When a user posts their comment read timestamp to their /v1/api I purge their user, this PR makes it also purge the Node in question.

@sloria @nicipfeiffer

### DIFF
--- a/website/project/views/comment.py
+++ b/website/project/views/comment.py
@@ -1,8 +1,13 @@
 # -*- coding: utf-8 -*-
+from functools import partial
+
 import pytz
 import markdown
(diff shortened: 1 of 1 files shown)


sloria:
This was cherry-picked into release/0.66.0 with #5207

PR #5210, opened 2016-03-17
[FEATURE] Upgrade DRF

Ghalko:
Hopefully not too much to this one.

## Proposal
Upgrade Django Rest Framework as far as possible (3.3.3)

## Changes
- Fixed some deprecations: `QUERY_PARAMS`
- Removed use of six.text_types for a hyperlink because it was returning a string that looked like a dictionary, but failed to act like a dictionary. The new implementation might be assuming things that are supported in Django models but not the mODM.
- Cleaned up the remaining 6 tests. Three of which were an `ValidationError` was taking the entire error rather than just the message for the details. The last three were a simple deprecation of checking if a field was blank. Used the default.


whit537:
@Ghalko This is in Code Review in [ADM-84](https://openscience.atlassian.net/browse/ADM-84), but Travis is red and the title still says WIP. Is this indeed ready for review?

Ghalko:
@whit537 This is not ready.

api/base/serializers.py
```diff
@@ -490,17 +491,54 @@ def format_filter(self, obj):
 
     # Overrides HyperlinkedIdentityField
     def to_representation(self, value):
-        urls = super(RelationshipField, self).to_representation(value)
-        if not urls:
+        request = self.context.get('request', None)
```
mfraezz: Why was this method changed so dramatically? I'm assuming something in `HyperlinkedIdentityField.to_representation` changed between DRF versions 3.1.1 and 3.3.3?

Ghalko: Correct. Essentially they added a `HyperLink` class that does not play well with our classes and returns what looks to be a dictionary but is actually a string. This really doesn't help in the part where we attempted to access it like a dictionary.

Even when I tried to mitigate the kludge by upgrading to `six` and use `from __future__` we still get a string that looks like a dictionary but isn't. So essentially this is a regression to continue working with the mongodb that we are so attached to.

website/files/models/base.py
```diff
@@ -12,7 +12,6 @@
 from modularodm.exceptions import NoResultsFound
 from dateutil.parser import parse as parse_date
 
-from api.base.utils import absolute_reverse
```
mfraezz: Why'd we stop using `absolute_reverse` in all of these places (but not all places)?

Ghalko: These were all the ones loaded by the flask app, which required DRF, which required `api.base.settings` to be set. Not good when you are just trying to build the JS libraries. I didn't search for the rest because they didn't cause failures.

api/base/filters.py
```diff
@@ -284,7 +284,7 @@ def get_query_from_request(self):
         if self.request.parser_context['kwargs'].get('is_embedded'):
             param_query = None
         else:
-            param_query = self.query_params_to_odm_query(self.request.QUERY_PARAMS)
+            param_query = self.query_params_to_odm_query(self.request.query_params)
```
mfraezz:
```
ipdb> request.QUERY_PARAMS
*** NotImplementedError: `request.QUERY_PARAMS` has been deprecated in favor of `request.query_params` since version 3.0, and has been fully removed as of version 3.2.
```
:+1:

mfraezz:
Pass complete. Manually testing the changes revealed no issues :star:

Putting this in RTM.

website/files/models/base.py
cwisecarver: @chrisseto has a PR where he changes several absolute_reverse calls too. First in wins.

brianjgeiger:
Closing in favor of https://github.com/CenterForOpenScience/osf.io/pull/5242/files

PR #5215, opened 2016-03-17
actually call doNext to update the count

hmoco:


sloria:
This doesn't look correct, `.then` expects to receive a function, so it looked correct before. But maybe I'm missing something?

hmoco:
hm, you are right, which is weird, because functional testing showed this to work. Will investigate

website/static/js/myProjects.js
```diff
@@ -812,7 +812,9 @@ var Collections = {
                             url : collection.data.node.links.self + 'node_links/', //collection.data.node.relationships.linked_nodes.links.related.href,
                             config : xhrconfig,
                             data : dataArray[index]
-                        }).then(doNext, function(result){
+                        }).then(function(){
+                            doNext();
```
sloria: For clarity, would be good to explicitly do `doNext(false)`. Otherwise, another dev might innocently change this back to `.then(doNext)`.

sloria: Alternatively : 

```js
.then(doNext.bind(this, false), function ...)
```

hmoco:
@sloria back to you

PR #5218, opened 2016-03-18
capitalize category on information tab

hmoco:


sloria:
As discussed on https://github.com/CenterForOpenScience/osf.io/pull/5200#discussion_r56380376:

> We have a mapping of internal names to human-readable names here: https://github.com/CenterForOpenScience/osf.io/blob/54418bc5fbce98fdb892267a27a93a560b06433a/website/project/model.py#L707-L719 .

Better to use this mapping than to rely on automatic capitalization.

Could you use the improved `OPTIONS` response described here: https://github.com/CenterForOpenScience/osf.io/pull/5200#discussion_r56399036 ? If not, you could just write the mapping of internal names -> display names in this file.

sloria:
Pass finished.

hmoco:
@sloria, responded!

website/static/js/myProjects.js
```diff
@@ -1324,6 +1324,30 @@ var Filters = {
  */
 var Information = {
     view : function (ctrl, args) {
+        function categoryMap(category) {
```
sloria: This is fine, I suppose, but why not an actual map?

```js
var CATEGORY_MAP = {
    analysis: 'Analysis',
    communication: 'Communication',
    // ...
}

hmoco: so that the default logic is done here rather than something like:
` CATEGORY_MAP['analysis'] if CATEGORY_MAP['analysis']  else 'Uncategorized' `

PR #5220, opened 2016-03-18
Do not embed empty relationships

brianjgeiger:
Purpose
-----------
Fix [https://openscience.atlassian.net/browse/OSF-5373] in which embedding a parent field for a top-level resource would embed an error rather than skipping the field.

Changes
------------
1. Add a test for the problem case
2. Check for the field representation early and, if it doesn't have a representation, bail out early
3. Cache the representation and use it later on when necessary
4. Fix a problem test from the Files views
5. Modify the HideIfRegistration and HideIfRetraction functions to raise SkipField if it's a registration, because the logic isn't always catching None's anymore depending
6. Explicitly skip the registrations field on registrations so it doesn't get serialized incorrectly

Side effects
----------------
Checking to verify the representation is potentially a little cumbersome, but some tests with profiling seems to indicate that it's faster more than it's slower, so that's good.

I'm also a little concerned that we might get some relationship fields added to attributes as None if I did this wrong, but I *think* we're okay there.

### DIFF
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -77,7 +77,10 @@ class HideIfRegistration(ser.Field):
 
     def get_attribute(self, instance):
         if instance.is_registration:
-            return None
+            if isinstance(self.field, RelationshipField):
+                raise SkipField
+            else:
+                return None
         return self.field.get_attribute(instance)
 
     def bind(self, field_name, parent):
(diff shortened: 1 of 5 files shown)


api/base/serializers.py
```diff
@@ -1038,3 +1050,7 @@ def relationship_diff(current_items, new_items):
         'add': {k: new_items[k] for k in (set(new_items.keys()) - set(current_items.keys()))},
         'remove': {k: current_items[k] for k in (set(current_items.keys()) - set(new_items.keys()))}
     }
+
+
+def skip_field():
```
mfraezz: This doesn't appear to be used anywhere.

mfraezz:
Pass complete. Looks good, just remove the unused method and it'll be RTM :100:

api/base/serializers.py
brianjgeiger: Ah, right. I fixed that a different  way. Thank you.

PR #5222, opened 2016-03-18
Trash old addon stored file node after move/rename

samanehsan:
Purpose
-----------
Addresses a commenting issue where if you upload a box file, then move it to a different provider, then re-upload the same file to box & move it again -- that second move fails because there are now two box stored file nodes for that file.

Changes
------------
When an addon file is moved or renamed, after the Guid referent is replaced with a newly created
StoredFileNode, delete the old StoredFileNode so that it is not possible to have multiple records
for the same file.


sloria:
Nice. Cherry-picked into release/0.66.0

PR #5223, opened 2016-03-18
[feature] Consistent Wiki Panel Header Heights 

caseyrollins:
- Adds `panel-height` CSS class and applies it to the wiki menu, edit, and view panel headers. 
- [OSF-3619](https://openscience.atlassian.net/browse/OSF-3619)

### DIFF
--- a/website/static/css/pages/wiki-page.css
+++ b/website/static/css/pages/wiki-page.css
@@ -6,6 +6,9 @@
 .panel-heading.no-background {
     background: transparent;
 }
+.wiki-panel-header {
+    height: 50px;
+}
 .markdown-it-view {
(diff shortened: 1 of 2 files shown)


website/static/css/pages/wiki-page.css
```diff
@@ -6,6 +6,9 @@
 .panel-heading.no-background {
     background: transparent;
 }
+.panel-height {
+    height: 50px;
```
sloria: Class names should describe the UI they are applied to; they should not describe the actual CSS rules that they implement. How about `wiki-panel-header`?

caseyrollins:
@sloria, made that name change :+1:

PR #5216, opened 2016-03-17
[WIP] Phase out usage of KO.punches

abought:
Refs https://openscience.atlassian.net/browse/OSF-5794

## Purpose
Phase out usage of knockout punches template helpers due to reported incompatibilities.

## Task list
- [x] Perform census of all pages affected, and plan best way to decompose this into different sections of the site
- [x] `{{ expression }}` should be replaced with `<!--ko text:expression--><!--/ko-->` or some equivalent
- [x] Remove usages of namespaced dynamic bindings (see the docs): https://mbest.github.io/knockout.punches/
- [ ] Convert pages that rely on wrapped event callbacks (by default, review `click`, `submit`, and `event` bindings)
- [ ] Replace ko.punches text filters (trim) with alternatives.
- [ ] Remove usages of `ko.punches.enableAll();`
- [ ] Remove ko-punches from vendor bundle

### List of affected pages / sections
To be written

### Things to check
- Links should work
- Any text that used to be present should still be present (no fields are missing)
- Long description fields are still truncated
- Should be able to click all buttons on the affected pages

PR #5224, opened 2016-03-18
[ko-1] Remove ko template syntax from non-project pages

abought:
Refs https://openscience.atlassian.net/browse/OSF-5794

## Purpose
Phase out usage of knockout punches template helpers due to reported incompatibilities.

This is the first of three planned PRs. The item touches on settings, profile, and some registration-related project pages. Addon widgets, nodelogs, and project dashboard items will be covered in a separate PR.

This replaces (and divides the work for) https://github.com/CenterForOpenScience/osf.io/pull/5216

## Task list
- [x] `{{ expression }}` should be replaced with `<!--ko text:expression--><!--/ko-->` or some equivalent
- [x] Replace text filters (`fit`) with alternative JS expressions
- [x] Remove usages of namespaced dynamic bindings (see the docs): https://mbest.github.io/knockout.punches/
- [x] Convert pages that rely on wrapped event callbacks (by default, review `click`, `submit`, and `event` bindings)
- [x] Replace ko.punches text filters (trim) with alternatives.
- [x] Remove usages of `ko.punches.enableAll();`

### List of affected pages / sections
- [x] Project settings page: be able to connect and disconnect various addons. Be able to select folders, view contents on files page, and remove/disconnect addons. Confirm that the dataverse addon is configured via a modal and that all buttons and links work. (`http://localhost:5000/<project_id>/settings`). Go to email notifications.
- [x] User/profile settings (`http://localhost:5000/settings/`) (all four tabs)
- [x] Account settings `http://localhost:5000/settings/account/`, particularly 2FA addon (enable, refresh, disable). Also check notifications and addon config. (a few buttons work, email notifications clickable)
- [x] Draft registrations pages: create and edit a draft registration (any form should be fine to test this with)
- [x] Configure citations widgets (mendeley, zotero)
- [x] Log in and sign up for accounts
- [x] Search page (look at search results for both projects and profiles. No extra/missing icons, no links broken, no JS console errors)

### Things to check
- Links should work
- Any text that used to be present should still be present (no fields are missing)
- In search results, long project descriptions (>500 chars) are still truncated
- Should be able to click all buttons on the affected pages, and buttons should work as intended

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -203,3 +203,23 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
+
+
+Helper functions adapted from Knockout.punches
+================================================
+
+Copyright (c) 2016 Michael Best
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+documentation files (the "Software"), to deal in the Software without restriction, including without limitation the
+rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit
+persons to whom the Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of
+the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO
+THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/website/addons/dataverse/static/dataverseNodeConfig.js
+++ b/website/addons/dataverse/static/dataverseNodeConfig.js
@@ -5,14 +5,12 @@
 
 var ko = require('knockout');
 var bootbox = require('bootbox');
-require('knockout.punches');
 var Raven = require('raven-js');
 
 var $osf = require('js/osfHelpers');
 
 var $modal = $('#dataverseInputCredentials');
 
-ko.punches.enableAll();
 
 function ViewModel(url) {
     var self = this;
--- a/website/addons/dataverse/static/dataverseUserConfig.js
+++ b/website/addons/dataverse/static/dataverseUserConfig.js
@@ -4,8 +4,6 @@
 */
 
 var ko = require('knockout');
-require('knockout.punches');
-ko.punches.enableAll();
 var $ = require('jquery');
 var Raven = require('raven-js');
 var bootbox = require('bootbox');
--- a/website/addons/dataverse/templates/dataverse_credentials_modal.mako
+++ b/website/addons/dataverse/templates/dataverse_credentials_modal.mako
@@ -45,7 +45,7 @@
                                 <label for="apiToken">
                                     API Token
                                     <!-- Link to API token generation page -->
-                                    <a href="{{tokenUrl}}"
+                                    <a data-bind="attr: {href: tokenUrl}"
                                        target="_blank" class="text-muted addon-external-link">
                                         (Get from Dataverse <i class="fa fa-external-link-square"></i>)
                                     </a>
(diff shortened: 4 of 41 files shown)


brianjgeiger:
Looking through the code, everything seems fine, and yay for the test on the fit function. Did you have any luck devising a way to be sure that knockout punches was definitely gone from a page, or did you determine it wasn't as necessary as you were thinking?

brianjgeiger:
I wonder what's up with that test. We may have to look into making the JS test more stable at some point.

abought:
> Looking through the code, everything seems fine, and yay for the test on the fit function. Did you have any luck devising a way to be sure that knockout punches was definitely gone from a page, or did you determine it wasn't as necessary as you were thinking?

As discussed with @brianjgeiger in person: 

Good questions, and honestly, I'm still a bit at a loss for a reliable test to see whether punches is present at all. `ko` is imported inside individual modules, so it's hard to check for presence at the console...

We did try adding a handlebars expression (inside a div bound as knockout) to `base.mako`, on the theory that if punches were present, a handlebars expression would be evaluated. Although we got the code in the `base.mako` template to act like knockout, we did *not* get it to respect punches syntax- even if adding the same line to another div (like contributors) caused the expression to be evaluated.

At best guess, there seems to be some weird interaction between punches and the scoping rules of webpack, so there's no single coherent canary-in-the-coal-mine test for whether punches is present. If we're lucky, that means that punches can't affect the entire page either, in return. *shrugs*

Experiment logged below for posterity: 

To base.mako
```
 <div id="knockoutdemo">
    some other text {{2+2}} {{alert('hello,world')}} {{ 'value' }}  {{12}}  after
    <span data-bind="text: value"></span> the  end
</div>
...
in footer:
<script src="${'/static/public/js/uglyhack.js' | webpack_asset}"></script>
```

And uglyhack.js :
```
$osf.applyBindings({'value': 42}, '#knockoutdemo');
```

abought:
> Looking through the code, everything seems fine, and yay for the test on the fit function.

To be fair, this was taken with very minor changes from the ko.punches codebase. :) Let me know if there's anything special we need to do to satisfy licensing issues. (Punches is MIT-licensed)

brianjgeiger:
Whhhaat about $osf.applyBindings()? Could we take advantage of that helper function to do a check for knockout punches, since we're using it for many/most/all application of KO bindings.

brianjgeiger:
Put license information for code that we include/modify in https://github.com/CenterForOpenScience/osf.io/blob/develop/NOTICE

abought:
> Whhhaat about $osf.applyBindings()? Could we take advantage of that helper function to do a check for knockout punches, since we're using it for many/most/all application of KO bindings.

Ah, there's an idea. 

## Short answer
There are potential diagnostics/flags, but artifacts of punches' presence may appear even *before* `enableAll` is called. After a bit more hacking, nothing so robust that I'd recommend it. 

Many of the behaviors we want to target are for templates, rather than hooks accessed through straight JS.

## Sample diagnostic code
The diagnostic below will only help if `knockout.punches` is removed from the vendor section of webpack.commin.config (line 66), and the asset rebuild restarted. It can be combined with the uglyhack notes above:

## A detector for whether punches is present / imported
`ko.punches` gets added when the library is first imported, but in theory, other bits of syntax only start to work their mojo later. To be more specific, I tried the `ko.filters` namespace, which is added by punches (filters can supposedly be turned on or off separately). Putting the following line at the end of `$osf.applyBindings` is therefore a useful diagnostic:

`    console.log('Knockout filters is enabled under selector ', selector, ': ', !!ko.filters);` (or use `ko.punches` directly)


## Caveats
Compare results on the project dashboard vs settings page. Punches won't be present in all binding contexts. Punches may claim to be present even for binding contexts where it is enabled. For the profile settings page, code from #5243:
```
14:22:28.114 Knockout filters is enabled under selector  #knockoutdemo :  false vendor.js:4035:5
14:22:28.634 Knockout filters is enabled under selector  #names :  true vendor.js:4035:5
14:22:28.915 Knockout filters is enabled under selector  #social :  true vendor.js:4035:5
14:22:28.979 Knockout filters is enabled under selector  #jobs :  true vendor.js:4035:5
14:22:29.005 Knockout filters is enabled under selector  #schools :  true vendor.js:4035:5
```

abought:
Did one more experiment, and this version could work as a punches detector. But as discussed in person, I'm not sure it's fully necessary so long as we merge the relevant PRs in a short span of time.

```
var applyBindings = function(viewModel, selector) {
    var elem, cssSelector;
    var $elem = $(selector);
    if (typeof(selector.nodeName) === 'string') { // dom element
        elem = selector;
        // NOTE: Only works with DOM elements that have an ID
        cssSelector = '#' + elem.id;
    } else {
        elem = $elem[0];
        cssSelector = selector;
    }
    if ($elem.length === 0) {
        throw "No elements matching selector '" + selector + "'";  // jshint ignore: line
    }
    if ($elem.length > 1) {
        throw "Can't bind ViewModel to multiple elements."; // jshint ignore: line
    }

    $elem.append("{{alert('Knockout punches is enabled')}}");

    // Ensure that the bound element is shown
    if ($elem.hasClass('scripted')){
        $elem.show();
    }
    // Also show any child elements that have the scripted class
    $(cssSelector + ' .scripted').each(function(elm) {
        $(this).show();
    });
    ko.applyBindings(viewModel, $elem[0]);

    console.log('Knockout filters is enabled under selector ', selector, ': ', !!ko.filters);
};
```

brianjgeiger:
I was going to say this was good, but apparently there are merge conflicts. Aside from that, it looks good to me.

abought:
Replaced by #5416.

PR #5226, opened 2016-03-18
Load search items in parallel with projects

caneruguz:
## Purpose
Have a separate list of flat list items load for the purpose of filtering

### DIFF
--- a/website/static/js/project-organizer.js
+++ b/website/static/js/project-organizer.js
@@ -384,9 +384,11 @@ var tbOptions = {
             tb.updateFolder(allTopLevelProjectsCache(), tb.treeData);
             $('.db-poFilter>input').val('');
         }
+        var filter = $osf.debounce(tb.filter, 800);
         return [ m('input.form-control[placeholder="Search all my projects"][type="text"]', {
             style: 'display:inline;',
             onkeyup: function(event){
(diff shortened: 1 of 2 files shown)


website/static/js/project-organizer.js
```diff
@@ -376,6 +377,13 @@ var tbOptions = {
     hScroll : 300,
     filterTemplate : function() {
         var tb = this;
+        function resetFilter () {
+            $osf.trackClick('myProjects', 'filter', 'clear-search');
+            tb.filterText('');
+            tb.resetFilter.call(tb);
+            tb.updateFolder(allTopLevelProjectsCache(), tb.treeData);
+            $('.db-poFilter>input').val('');
```
sloria: Why do you need to change the value with jQuery? Isn't the input bound to a prop?

sloria:
@caneruguz When the filtering issue you are working on is fixed, please remove "[wip]" from the PR description and ping me on Flowdock.

website/static/js/project-organizer.js
caneruguz: This might be related to a redraw problem, technically tb.filterText('') should clear it. Let's leave this for later to not block the rest of the code.

PR #5227, opened 2016-03-18
Update messages.py

GaryKriebel:
Changed fishare to figshare in two places to fix the issue stated in [#OSF-5888]


sloria:
Cherry-picked into `release/0.66.0`

PR #5230, opened 2016-03-18
Reorder the categories so more recent comes to the top

billyhunt:
The categories on the new dashboard should list in order of most recently updated to least.  This PR sorts according to the most recent date modified.  So far, it does not refresh when a project is added to a category.  I suspect this is due to the redrawing bug on that bar, but am not sure.  It will have to be addressed at some point.  (When you add a project to a collection, that collection should pop to the top)


website/static/js/myProjects.js
```diff
@@ -523,7 +549,8 @@ var MyProjects = {
         self.loadCollections = function _loadCollections (url){
             var promise = m.request({method : 'GET', url : url, config : xhrconfig});
             promise.then(function(result){
-                result.data.forEach(function(node){
+                self.sorted_collections = sortCollections(result.data);
```
sloria: Just a drive-by comment: In general (both in Python and JS), avoid setting attributes outside of a constructor. Also, don't set any unnecessary attributes.

PR #4411, opened 2015-10-07
Add client side validation

billyhunt:
Fixes #3149 

The knockout front end URL validation does not work for observable arrays.  So, bad urls such as 

```
http://&lt;/script&gt;&lt;script&gt;alert(“Now I am doing bad stuff”)&lt;/script&gt;
```
and 
```
&lt;/script&gt;&lt;script&gt;alert(“Now I am doing bad stuff”)&lt;/script&gt;
```

generate either 400 or 500 errors on the server.

This update stops incorrect requests from being sent, giving relevant error messaging.  Note that I put the message at the top and the bottom.  The top message shows where the problem is, the bottom one kicks in if you press Save with invalid url information.  

screen shot 2015-10-07 at 9 17 29 am
Current

screen shot 2015-10-07 at 9 18 42 am
Proposed

PR #5232, opened 2016-03-18
[feature] Client Side Profile URL Validation

caseyrollins:
- Replaces/Continues #4411 

##### Purpose
- Currently, if a user enters and submits an invalid URL in their profile, the server will respond with a 400 error,  and their website field is not saved. However, the user is unaware of this as there is no client-side URL validation, and no error message is presented to the user. 

##### Changes
- This PR adds client-side URL validation so that if a user attempts to enter an invalid website, they will receive an error message which instructs them to enter a valid website. Additionally, if the user attempts to save their profile with an invalid website, they will be presented with another error message that directs them to fix the website field. 

- Error message when entering invalid URL (no previous error message was displayed):
 - screen shot 2016-03-21 at 1 51 22 pm

- Error message when saving profile with invalid URL (previously, "Could not update settings" was displayed):
 - screen shot 2016-03-21 at 1 51 33 pm

##### JIRA
- [OSF-3190](https://openscience.atlassian.net/browse/OSF-3190)


brianjgeiger:
Could you add screen shots of the old and new functionality?

website/static/js/profile.js
```diff
@@ -23,7 +23,13 @@ var socialRules = {
     github: /github\.com\/(\w+)/i,
     researchGate: /researchgate\.net\/profile\/(\w+)/i,
     academia: /(\w+)\.academia\.edu\/(\w+)/i,
-    baiduScholar: /xueshu\.baidu\.com\/scholarID\/(\w+)/i
+    baiduScholar: /xueshu\.baidu\.com\/scholarID\/(\w+)/i,
+    url: '^(https?:\\/\\/)?'+ // protocol
```
brianjgeiger: Does this match something in the Python code somewhere?

website/static/js/profile.js
```diff
@@ -339,7 +345,7 @@ BaseViewModel.prototype.cancel = function(data, event) {
 };
 
 BaseViewModel.prototype.submit = function() {
-    if (this.hasValidProperty() && this.isValid()) {
+     if (this.hasValidProperty() && this.isValid()) {
```
brianjgeiger: I don't believe this space is correct.

caseyrollins:
@brianjgeiger Screenshots added! :camera:

brianjgeiger:
When your first go into the social tab, if you don't have a website, it complains about the lack of a valid website (and gives you a not-particularly-useful X to remove the website). It should only do the validation if you make a change to the website field, or more likely if the website field is not blank. 

screen shot 2016-03-21 at 1 54 39 pm

brianjgeiger:
I'm having some other odd problems which I expect you would have seen, so the above might be because webpack isn't webpacking. Let's see.

caseyrollins:
Hmmm... I'm not seeing the error message upon entering the social tab. I'm only receiving the message when I hit enter after typing in some nonsense.

Also, the highly-confusing X on the side of the website field is always present, and is unrelated to this PR.

brianjgeiger:
Okay, actually running webpack seems to have fixed the blank website erroring issue, so please ignore.

website/static/js/profile.js
brianjgeiger: Looks like it does. `from modularodm.validators import URLValidator` What we will want to avoid is to have someone change this from under us without knowing about it so we can update this, because having incompatible validators would be much worse than not validating the information. I think, in this case, we should ensure that our validation tests are reasonably complete as far as coverage and should be applied to both functions equally. 

So perhaps beef up your JS tests to handle query strings, ip addresses, ports, paths, protocols, etc and then make sure we have python tests that call URL validator that match those tests. The front end can be more restrictive than the back-end, if they have to be different.

brianjgeiger:
Looking good. I think the one space and making testing more extensive and matchy will do it.

brianjgeiger:
Looks great. :tada: :penguin: :tada:

brianjgeiger:
(Pending travis, not counting that one JS test that Steve is fixing)

PR #5233, opened 2016-03-18
Added promise to fix lack of file expansion on treebeard instances

billyhunt:
Responding to code review for 

It is possible that the expandOnLoad function is not ready to rock before the treebeard instance is instantiated.  So I am forcing the sequence of events with a promise.  This issue is a little challenging to track because I cannot reproduce it in my dev environment.


sloria:
1. The native `Promise` class is only available in the newest browsers, and we don't currently have a shim for it. Instead we use jQuery's `Deferred`.
2. I'm not sure this code is actually doing anything. What sequence of events needs to happen before `expandOnLoad` is called?

billyhunt:
I "think" the grid needs to be instantiated.  I am not seeing the problem on my local environment.

billyhunt:
I made the branch wiggity wack.  I'll send a new PR.

PR #5240, opened 2016-03-21
[RFR][Feature]OSF-5847 Update fangorn to use WaterButler v1 API

TomBaxter:
## Purpose:
[OSF-5847](https://openscience.atlassian.net/browse/OSF-5847)
Update fangorn to use WaterButler v1 API

## Changes:
update website/addons/github/static/githubFangornConfig.js for WB v1 upload URL
update website/static/js/fangorn.js for URLS, and lazyLoadPreprocess
update website/static/js/waterbutler.js for WB v1 URLS
update website/static/js/filepage/revisions.js for lazyLoadPreprocess

## ToDo:
Figshare is untested

## Side effects:
Relies on WB v1 API only

#OSF-5847

### DIFF
--- a/website/addons/github/static/githubFangornConfig.js
+++ b/website/addons/github/static/githubFangornConfig.js
@@ -14,7 +14,17 @@ var $osf = require('js/osfHelpers');
 var commandKeys = [224, 17, 91, 93];
 
 function _uploadUrl(item, file) {
-    return waterbutler.buildTreeBeardUpload(item, file, {branch: item.data.branch});
+    // WB v1 update syntax is file_path/?kind=file
+    // WB v1 upload syntax is parent_path/?kind=file&name=filename
+    // If upload target file name already exists don't pass file.name
+    var updateUrl;
+    $.each(item.children, function( index, value ) {
+        if (file.name === value.data.name) {
+            updateUrl = waterbutler.buildTreeBeardUpload(value, {branch: value.data.branch}); 
+            return false;
+        }
+    });
+    return updateUrl || waterbutler.buildTreeBeardUpload(item, {name: file.name, branch: item.data.branch});
 }
 
 // TODO: Refactor, repeating from core function too much
@@ -25,7 +35,7 @@ function _removeEvent (event, items) {
     }
 
     function runDelete (item) {
-        tb.select('.tb-modal-footer .text-danger').html('<i> Deleting...</i>').css('color', 'grey');;
+        tb.select('.tb-modal-footer .text-danger').html('<i> Deleting...</i>').css('color', 'grey');
         // delete from server, if successful delete from view
         $.ajax({
             url: waterbutler.buildTreeBeardDelete(item, {branch: item.data.branch, sha: item.data.extra.fileSha}),
@@ -303,7 +313,7 @@ function _fangornGithubTitle(item, col)  {
         var urlParams = $osf.urlParams();
         
         if (!item.data.branch) {
-            if (urlParams.branch && urlParams.branch != item.data.branch) {
+            if (urlParams.branch && urlParams.branch !== item.data.branch) {
                 item.data.branch = urlParams.branch;
             }
         }
--- a/website/static/js/fangorn.js
+++ b/website/static/js/fangorn.js
@@ -426,26 +426,45 @@ function doItemOp(operation, to, from, rename, conflict) {
     }
     orderFolder.call(tb, from.parent());
 
+    var jsonData;
+    if (operation === OPERATIONS.RENAME) {
+        jsonData = {
+            action: 'rename',
+            rename: rename
+        };
+    } else if (operation === OPERATIONS.COPY) {
+        jsonData = {
+            action: 'copy',
+            path: to.data.path || '/',
+            conflict: conflict,
+            resource: to.data.resource,
+            provider: to.data.provider
+        };
+    } else if (operation === OPERATIONS.MOVE) {
+        jsonData = {
+            action: 'move',
+            path: to.data.path || '/',
+            conflict: conflict,
+            resource: to.data.resource,
+            provider: to.data.provider
+        };
+    }
+
     $.ajax({
         type: 'POST',
         beforeSend: $osf.setXHRAuthorization,
-        url: operation === OPERATIONS.COPY ? waterbutler.copyUrl() : waterbutler.moveUrl(),
+        url: waterbutler.buildTreeBeardFileOp(from),
         headers: {
             'Content-Type': 'Application/json'
         },
-        data: JSON.stringify({
-            'rename': rename,
-            'conflict': conflict,
-            'source': waterbutler.toJsonBlob(from),
-            'destination': waterbutler.toJsonBlob(to),
-        })
+        data: JSON.stringify(jsonData)
     }).done(function(resp, _, xhr) {
         if (to.data.provider === from.provider) {
             tb.pendingFileOps.pop();
         }
         if (xhr.status === 202) {
             var mithrilContent = m('div', [
-                m('h3.break-word', operation.action + ' "' + from.data.materialized + '" to "' + (to.data.materialized || '/') + '" is taking a bit longer than expected.'),
+                m('h3.break-word', operation.action + ' "' + (from.data.materialized || '/') + '" to "' + (to.data.materialized || '/') + '" is taking a bit longer than expected.'),
                 m('p', 'We\'ll send you an email when it has finished.'),
                 m('p', 'In the mean time you can leave this page; your ' + operation.status + ' will still be completed.')
             ]);
@@ -455,7 +474,7 @@ function doItemOp(operation, to, from, rename, conflict) {
             tb.modal.update(mithrilContent, mithrilButtons);
             return;
         }
-        from.data = resp;
+        from.data = tb.options.lazyLoadPreprocess.call(this, resp).data;
         from.data.status = undefined;
         from.notify.update('Successfully ' + operation.passed + '.', 'success', null, 1000);
 
@@ -510,12 +529,7 @@ function doItemOp(operation, to, from, rename, conflict) {
         Raven.captureMessage('Failed to move or copy file', {
             extra: {
                 xhr: xhr,
-                requestData: {
-                    rename: rename,
-                    conflict: conflict,
-                    source: waterbutler.toJsonBlob(from),
-                    destination: waterbutler.toJsonBlob(to),
-                }
+                requestData: jsonData
             }
         });
 
@@ -532,8 +546,22 @@ function doItemOp(operation, to, from, rename, conflict) {
  * @private
  */
 function _fangornResolveUploadUrl(item, file) {
+    // WB v1 update synatx is file path and kind=file
+    // WB v1 upload synatx is parent path and kind=file, name=filename
+    // If upload target file name already exists don't pass file.name
     var configOption = resolveconfigOption.call(this, item, 'uploadUrl', [item, file]); // jshint ignore:line
-    return configOption || waterbutler.buildTreeBeardUpload(item, file);
+    if (configOption) {
+        return configOption;
+    }
+    var updateUrl;
+    $.each(item.children, function( index, value ) {
+        if (file.name === value.data.name) {
+            updateUrl = waterbutler.buildTreeBeardUpload(value);
+            return false;
+        }
+    });
+    
+    return updateUrl || waterbutler.buildTreeBeardUpload(item, {name: file.name});
 }
 
 /**
@@ -731,7 +759,7 @@ function _fangornDropzoneSuccess(treebeard, file, response) {
     // Dataverse : Object, actionTaken : file_uploaded
     revisedItem = resolveconfigOption.call(treebeard, item.parent(), 'uploadSuccess', [file, item, response]);
     if (!revisedItem && response) {
-        item.data = response;
+        item.data = treebeard.options.lazyLoadPreprocess.call(this, response).data;
         inheritFromParent(item, item.parent());
     }
     if (item.data.tmpID) {
@@ -799,7 +827,6 @@ function _fangornDropzoneError(treebeard, file, message, xhr) {
             child.removeSelf();
         }
     }
-    console.log(file);
     treebeard.options.uploadInProgress = false;
     if (msgText !== 'Upload canceled.') {
         addFileStatus(treebeard, file, false, msgText, '');
@@ -869,18 +896,20 @@ function _createFolder(event, dismissCallback, helpText) {
     }
 
     var extra = {};
-    var path = (parent.data.path || '/') + val + '/';
+    var path = parent.data.path || '/';
+    var options = {name: val, kind: 'folder'};
 
     if (parent.data.provider === 'github') {
         extra.branch = parent.data.branch;
     }
 
     m.request({
-        method: 'POST',
+        method: 'PUT',
         background: true,
         config: $osf.setXHRAuthorization,
-        url: waterbutler.buildCreateFolderUrl(path, parent.data.provider, parent.data.nodeId)
+        url: waterbutler.buildCreateFolderUrl(path, parent.data.provider, parent.data.nodeId, options)
     }).then(function(item) {
+        item = tb.options.lazyLoadPreprocess.call(this, item).data;
         inheritFromParent({data: item}, parent, ['branch']);
         item = tb.createItem(item, parent.id);
         orderFolder.call(tb, parent);
@@ -2230,6 +2259,7 @@ tbOptions = {
     uploads : true,         // Turns dropzone on/off.
     columnTitles : _fangornColumnTitles,
     resolveRows : _fangornResolveRows,
+    lazyLoadPreprocess: waterbutler.wbLazyLoadPreprocess,
     hoverClassMultiselect : 'fangorn-selected',
     multiselect : true,
     placement : 'files',
--- a/website/static/js/filepage/revisions.js
+++ b/website/static/js/filepage/revisions.js
@@ -44,6 +44,7 @@ var FileRevisionsTable = {
                 url: self.file.urls.revisions,
                 beforeSend: $osf.setXHRAuthorization
             }).done(function(response) {
+                response = waterbutler.wbLazyLoadPreprocess.call(this, response);
                 m.startComputation();
                 var urlParmas = $osf.urlParams();
                 model.revisions = response.data.map(function(rev, index) {
@@ -80,6 +81,7 @@ var FileRevisionsTable = {
                         url: self.file.urls.metadata,
                         beforeSend: $osf.setXHRAuthorization
                     }).done(function(resp) {
+                        resp = waterbutler.wbLazyLoadPreprocess.call(this, resp);
                         self.canEdit(self.canEdit() && resp.data.extra.canDelete);
                         m.redraw();
                     }).fail(function(xhr) {
--- a/website/static/js/waterbutler.js
+++ b/website/static/js/waterbutler.js
@@ -3,15 +3,14 @@
 var $ = require('jquery');
 var $osf = require('./osfHelpers');
 
-function getViewOnly() {
-    return $osf.urlParams().view_only;
+function buildUrl(path, provider, nid, options) {
+    path = path || '/';
+    var baseUrl = window.contextVars.waterbutlerURL + 'v1/resources/' + nid + '/providers/' + provider + path + '?';
+    return baseUrl + $.param($.extend(getDefaultOptions(), options));
 }
 
-function getDefaultOptions(path, provider) {
-    var options = {
-        path: path,
-        provider: provider
-    };
+function getDefaultOptions() {
+    var options = {};
     var viewOnly = getViewOnly();
     if (viewOnly) {
         options.view_only = viewOnly;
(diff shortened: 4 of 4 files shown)


website/static/js/fangorn.js
```diff
@@ -426,26 +426,47 @@ function doItemOp(operation, to, from, rename, conflict) {
     }
     orderFolder.call(tb, from.parent());
 
+    var jsonData;
+    if (operation === OPERATIONS.RENAME) {
+        jsonData = JSON.stringify({
+        'action': 'rename',
```
felliott: I think your editor's acting up again.  These lines should be indented.

website/static/js/waterbutler.js
```diff
@@ -24,65 +21,57 @@ function getDefaultOptions(path, provider) {
...
+    buildCreateFolderUrl: buildCreateFolderUrl,
     buildRevisionsUrl: buildRevisionsUrl,
-    buildTreeBeardUpload: buildFromTreebeardFile,
-    buildTreeBeardCopy: buildFromTreebeard.bind(this, 'copy?'),
-    buildTreeBeardMove: buildFromTreebeard.bind(this, 'move?'),
-    buildTreeBeardDelete: buildFromTreebeard.bind(this, 'file?'),
+    buildTreeBeardUpload: buildTreeBeardUpload,
+    buildTreeBeardFileOp: buildFromTreebeard.bind(this, ''),
+    // TODO would be nice to prompt for root delete
+    buildTreeBeardDelete: buildFromTreebeard.bind(this, '?confirm_delete=1'),
```
felliott: Since Fangorn doesn't have an interface for storage root deletes, I would leave this out.  Right now a delete against the storage root is probably a programmer error and throwing an error is good.

website/static/js/fangorn.js
```diff
@@ -426,26 +426,47 @@ function doItemOp(operation, to, from, rename, conflict) {
     }
     orderFolder.call(tb, from.parent());
 
+    var jsonData;
+    if (operation === OPERATIONS.RENAME) {
+        jsonData = JSON.stringify({
+            'action': 'rename',
```
felliott: Minor: don't quote JS object keys unless necessary: https://github.com/felixge/node-style-guide#object--array-creation

website/static/js/waterbutler.js
```diff
@@ -24,65 +21,58 @@ function getDefaultOptions(path, provider) {
...
-function buildFromTreebeardFile(item, file, options) {
-    return buildUploadUrl(item.data.path, item.data.provider, item.data.nodeId, file, options);
+function buildFromTreebeard(suffix, item, options) {
+    var path;
+    if ('attributes' in item.data) {
+        path = item.data.attributes.path;
+    } else {
+        path = item.data.path;
+    }
+    return buildUrl(suffix, path, item.data.provider, item.data.nodeId, options);
```
felliott: If `path` can be in either `item.data` or `item.data.attributes`, can't `provider` be in both, too?  Or is that something Fangorn/Treebeard adds explicitly to `item`.

felliott:
How does file update work now?  `PUT /?kind=file&name=foo.txt` used to do create AND update, but now does only create.

felliott:
In `waterbutler.js`: the suffix binding approach was used because actions were part of the url.  With v1, the actions are captured in the HTTP method, and the url is always a path to a file or folder.  Query parameters should be set by updating the `options` param that's passed down.

website/static/js/fangorn.js
```diff
@@ -426,26 +426,47 @@ function doItemOp(operation, to, from, rename, conflict) {
...
-            'conflict': conflict,
-            'source': waterbutler.toJsonBlob(from),
-            'destination': waterbutler.toJsonBlob(to),
-        })
+        data: jsonData
     }).done(function(resp, _, xhr) {
         if (to.data.provider === from.provider) {
             tb.pendingFileOps.pop();
         }
+    var fromMaterialized = from.data.materialized || '/';
```
felliott: I don't think `from.data.materialized` can be undefined or empty, since we always have to have a file or folder to move.

website/static/js/fangorn.js
```diff
@@ -426,26 +426,47 @@ function doItemOp(operation, to, from, rename, conflict) {
...
+    var toMaterialized = to.data.materialized;
```
felliott: If this is only being used once, there's no need to extract it into a variable.  Or if it's being used again, consider putting the alternative initial value (`'/'`) on this line, so you don't have to do it inline below.

website/static/js/fangorn.js
TomBaxter: Will do, thanks for the link. I'll study it.

website/static/js/fangorn.js
```diff
@@ -455,13 +476,14 @@ function doItemOp(operation, to, from, rename, conflict) {
             tb.modal.update(mithrilContent, mithrilButtons);
             return;
         }
-        from.data = resp;
+        from.data = resp.data.attributes;
         from.data.status = undefined;
         from.notify.update('Successfully ' + operation.passed + '.', 'success', null, 1000);
 
+    var fromName = from.data.name || from.data.attributes.name;
```
felliott: Your text editor's indentation is off here and on `var fromKind = ...`.

website/static/js/fangorn.js
```diff
@@ -509,12 +532,7 @@ function doItemOp(operation, to, from, rename, conflict) {
 
         Raven.captureMessage('Failed to move or copy file', {
             xhr: xhr,
-            requestData: {
-                rename: rename,
-                conflict: conflict,
-                source: waterbutler.toJsonBlob(from),
-                destination: waterbutler.toJsonBlob(to),
-            }
+            requestData: jsonData
```
felliott: Oh, I see what you mean when you were asking that question about raven now.  `requestData` should be an object, not a serialized string.  Consider leaving `jsonData` as an object and only serializing it when setting the `POST` body.

website/static/js/waterbutler.js
TomBaxter: Fangorn/Treebeard adds "provider" and "name" at the "item" level. In addition they will also be found at either item.data (for root (googledrive, osfstorage etc.) or item.data.attributes for data received through WB v1 API.

website/static/js/fangorn.js
```diff
@@ -867,18 +884,26 @@ function _createFolder(event, dismissCallback, helpText) {
     }
 
     var extra = {};
-    var path = (parent.data.path || '/') + val + '/';
+    var path = (parent.data.path || '/');
```
felliott: Trivial: no need for the parens.

website/static/js/waterbutler.js
felliott: Ah-ha!  In that case, :+1:

website/static/js/fangorn.js
```diff
@@ -867,18 +884,26 @@ function _createFolder(event, dismissCallback, helpText) {
...
-        url: waterbutler.buildCreateFolderUrl(path, parent.data.provider, parent.data.nodeId)
+        url: waterbutler.buildCreateFolderUrl(path, parent.data.provider, parent.data.nodeId, options)
     }).then(function(item) {
+    var newItem = {attributes:item.data.attributes};
+    Object.keys(item).forEach(function(key,index) {
+        delete item[key];
+    });
+    Object.keys(newItem.attributes).forEach(function(key,index) {
+        item[key] = newItem.attributes[key];
+    });
```
felliott: Can this whole block be replaced with `item = item.data.attributes`?

website/static/js/fangorn.js
```diff
@@ -2228,6 +2253,14 @@ tbOptions = {
     uploads : true,         // Turns dropzone on/off.
     columnTitles : _fangornColumnTitles,
     resolveRows : _fangornResolveRows,
+    // Sets location under root to find treebeard data from API. For WB v1 "attributes"
+    dataLocation : function(value) {
+        if ('attributes' in value) {
+            return value.attributes;
+        } else {
```
felliott: Since the if condition returns, you don't need to wrap the default return in an `else`.

website/static/js/fangorn.js
TomBaxter: Materialized doesn't get set in treebeard/fangorn for the root/non-API items.

website/static/js/fangorn.js
TomBaxter: Done.

website/static/js/filepage/revisions.js
```diff
@@ -58,8 +58,8 @@ var FileRevisionsTable = {
                 if (model.selectedRevision === 0) {
                     self.enableEditing();
                 }
-                model.hasUser = model.revisions[0] && model.revisions[0].extra && model.revisions[0].extra.user;
-                model.hasHashes = model.revisions && model.revisions[0] && model.revisions[0].extra.hashes;
+                model.hasUser = model.revisions[0] && model.revisions[0].attributes.extra && model.revisions[0].attributes.extra.user;
```
felliott: Instead of adding `attributes` to the dereference chain throughout the code, can you extract the attributes when `model.revisions` is created?

website/static/js/fangorn.js
TomBaxter: I'm gonna work on setting up my vim for .js as soon as these PRs are in the clear. I cut out everything .js related after the tabs issue. So, just running base vim for .js at the moment. Sorry for all the styling issues.

website/static/js/fangorn.js
TomBaxter: done

website/static/js/fangorn.js
TomBaxter: Done

website/static/js/fangorn.js
TomBaxter: yes.  :/

website/static/js/fangorn.js
TomBaxter: @felliott So the value sent to a function will be returned if the function doesn't explicitly return something?

felliott: Sorry, no.  I just meant replace `else { return value; }` with `return value;`

TomBaxter: Dur, gotcha, done.

website/static/js/filepage/revisions.js
TomBaxter: Done

website/static/js/waterbutler.js
```diff
@@ -22,67 +21,58 @@ function getDefaultOptions(path, provider) {
...
-    buildDeleteUrl: buildCrudUrl,
-    buildUploadUrl: buildUploadUrl,
-    buildDownloadUrl: function(path, provider, nid, options) {
-        if (window.contextVars.accessToken) {
-            options = $.extend(options || {}, {token: window.contextVars.accessToken});
-        }
-        return buildCrudUrl(path, provider, nid, options);
+    buildDeleteUrl: buildUrl,
+    buildDownloadUrl: buildUrl,
+    buildUploadUrl: function(path, provider, nid, options) {
```
felliott: Minor: name your anonymous functions.  That'll make backtraces easier to follow.  E.g.
````
buildUploadUrl: function _buildUploadUrl(path, provider, nid, options) {
````

File: website/static/js/waterbutler.js [hunk 14]
```diff
@@ -22,67 +21,58 @@ function getDefaultOptions(path, provider) {
...
+        options = $.extend(options || {}, {kind: 'files'});
```
felliott: `$.extend(options, {foo: 'bar'})` will modify `options` in place.  This means whoever calls the function will have the object they pass in modified.  It would probably be better to do something like:
`var wb_options = $.extend({}, options || {}, {foo: 'bar'});`

website/static/js/fangorn.js
```diff
@@ -469,7 +489,8 @@ function doItemOp(operation, to, from, rename, conflict) {
 
         inheritFromParent(from, from.parent());
 
-        if (from.data.kind === 'folder' && from.data.children) {
+    var fromKind = from.data.kind || from.data.attributes.kind;
```
felliott: Minor: editor misindent.

TomBaxter: @felliott Fixed. Also fixed "extend" issue we talked about.

felliott:
I think we talked about this before and I probably gave you bad information.  With the current Fangorn, the user can update a file by uploading a file with the same name.  The conversion to v1 breaks this because of the new stricter v1 API.  We do need to support this, so you'll probably need to search through the treebeard entries, look for a filename match, then use the update url if you find one.

website/static/js/fangorn.js
```diff
@@ -2230,6 +2245,25 @@ tbOptions = {
...
+        if (!$.isArray(obj.data)) {
+            $.each( obj.data.attributes, function( key, value ) {
+                    obj.data[key] = value;
+            });
+            delete obj.data.attributes;
+            return obj;
+        }
+        /* jshint ignore:start */
+        for (var i = 0; i < obj.data.length; i++) {
+            $.each( obj.data[i].attributes, function( key, value ) {
```
felliott: I think you could iterate over `obj.data` with `$.each`, and do `$.extend(entity, entity.attributes)` to copy the properties up a level.

felliott:
I'm seeing JS errors: `TypeError: revision.version is undefined` when viewing a file.

website/static/js/fangorn.js
TomBaxter: Done.

TomBaxter:
Update website/static/js/filepage/revisions.js to use lazyLoadPreprocess.

TomBaxter:
Added check and resolve for existing file name during upload.

website/addons/github/static/githubFangornConfig.js
```diff
@@ -14,7 +14,17 @@ var $osf = require('js/osfHelpers');
 var commandKeys = [224, 17, 91, 93];
 
 function _uploadUrl(item, file) {
-    return waterbutler.buildTreeBeardUpload(item, file, {branch: item.data.branch});
+    // WB v1 update synatx is file path and kind=file
+    // WB v1 upload synatx is parent path and kind=file, name=filename
```
felliott: minor: s/synatx/syntax/.    formatting suggestion: "WB v1 update syntax is /<file_path>?kind=file"

website/addons/github/static/githubFangornConfig.js
```diff
@@ -14,7 +14,17 @@ var $osf = require('js/osfHelpers');
...
+    // If upload target file name already exists don't pass file.name
+    var updateUrl;
+    $.each(item.children, function( index, value ) {
+        if (file.name === value.data.name) {
+        updateUrl = waterbutler.buildTreeBeardUpload(value, {branch: value.data.branch});
```
felliott: needs indent

website/static/js/fangorn.js
```diff
@@ -455,7 +474,7 @@ function doItemOp(operation, to, from, rename, conflict) {
             tb.modal.update(mithrilContent, mithrilButtons);
             return;
         }
-        from.data = resp;
+        from.data = tbOptions.lazyLoadPreprocess.call(this, resp).data;
```
felliott: Instead of calling `lazyLoadPreprocess` on the options global, why not call it on the instantiated treebeard object instead?

website/static/js/fangorn.js
```diff
@@ -532,8 +546,18 @@ function doItemOp(operation, to, from, rename, conflict) {
...
+    var updateUrl;
+    $.each(item.children, function( index, value ) {
+        if (file.name === value.data.name) {
+            updateUrl = waterbutler.buildTreeBeardUpload(value);
+            return false;
+        }
+    });
     var configOption = resolveconfigOption.call(this, item, 'uploadUrl', [item, file]); // jshint ignore:line
-    return configOption || waterbutler.buildTreeBeardUpload(item, file);
+    return configOption || updateUrl || waterbutler.buildTreeBeardUpload(item, {name: file.name});
```
felliott: There's no need to calculate `updateUrl` if `configOption` is set.

website/static/js/fangorn.js
```diff
@@ -2230,6 +2255,21 @@ tbOptions = {
...
     columnTitles : _fangornColumnTitles,
     resolveRows : _fangornResolveRows,
+    // Move obj.data.attributes contents to obj.data , for WB v1 API
+    lazyLoadPreprocess: function(obj) {
+        if (!$.isArray(obj.data)) {
+            $.extend(true, obj.data, obj.data.attributes);
+            delete obj.data.attributes;
+            return obj;
+        }
+        /* jshint ignore:start */
```
felliott: Do you still need the jshint hints?

website/static/js/fangorn.js
```diff
@@ -2230,6 +2255,21 @@ tbOptions = {
     uploads : true,         // Turns dropzone on/off.
     columnTitles : _fangornColumnTitles,
     resolveRows : _fangornResolveRows,
+    // Move obj.data.attributes contents to obj.data , for WB v1 API
+    lazyLoadPreprocess: function(obj) {
+        if (!$.isArray(obj.data)) {
+            $.extend(true, obj.data, obj.data.attributes);
+            delete obj.data.attributes;
```
felliott: Time for me to be parannoying (paranoid + annoying).  If there is a `data.attributes.attributes` attribute, it'll be deleted by this code.  I'd recommend saving `data.attributes`, then deleting it from `data`, then extending `data` with it.  You might also want to break it out into a small anonymous function so that you can reuse it in the for loop below.

website/static/js/filepage/revisions.js
```diff
@@ -44,6 +45,7 @@ var FileRevisionsTable = {
                 url: self.file.urls.revisions,
                 beforeSend: $osf.setXHRAuthorization
             }).done(function(response) {
+                response = fangorn.DefaultOptions.lazyLoadPreprocess.call(this, response);
```
felliott: Can you get `lazyLoadPreprocess` from the treebeard object instead of the global defaults?

website/addons/github/static/githubFangornConfig.js
TomBaxter: Done.

website/addons/github/static/githubFangornConfig.js
TomBaxter: Done.

website/static/js/fangorn.js
TomBaxter: Done.

website/static/js/fangorn.js
TomBaxter: Done.

website/static/js/fangorn.js
TomBaxter: No. Removed.

website/static/js/fangorn.js
TomBaxter: Done.

website/static/js/filepage/revisions.js
TomBaxter: @felliott As near as I can tell there isn't a treebeard object available in revisions.js , am I missing something?

website/static/js/fangorn.js
```diff
@@ -532,8 +546,22 @@ function doItemOp(operation, to, from, rename, conflict) {
...
 function _fangornResolveUploadUrl(item, file) {
+    // WB v1 update synatx is file path and kind=file
+    // WB v1 upload synatx is parent path and kind=file, name=filename
+    // If upload target file name already exists don't pass file.name
+    var updateUrl;
     var configOption = resolveconfigOption.call(this, item, 'uploadUrl', [item, file]); // jshint ignore:line
-    return configOption || waterbutler.buildTreeBeardUpload(item, file);
+    if (configOption) {
+        return configOption;
+    } else {
```
felliott: Minor: you don't need to wrap the rest of the code in an `else` here if the true condition returns.  You can also move the declaration of `updateUrl` down closer to it's assignment.

website/static/js/fangorn.js
```diff
@@ -2230,6 +2259,23 @@ tbOptions = {
     uploads : true,         // Turns dropzone on/off.
     columnTitles : _fangornColumnTitles,
     resolveRows : _fangornResolveRows,
+    // Move obj.data.attributes contents to obj.data , for WB v1 API
+    lazyLoadPreprocess: function(obj) {
+        function moveAttrs(obj) {
```
felliott: Minor: since this is an internal utility function, prefix the name with an underscore.  Also, add a docstring explaining its purpose and mentioning the `obj.attributes.attributes` issue.  Super nitpicky: rename it to `_promoteAttrs` to indicate where/how attributes is being moved.

felliott: Important: you have two variables called `obj` in this scope.  Please rename on of them.

website/static/js/fangorn.js
TomBaxter: Done.

website/static/js/fangorn.js
TomBaxter: @felliott 
Minor:'s done.

I'm not sure I'm following on the Important: . I thought it was the same variable/object(obj) throughout. However, I made my best guess and certainly the renaming is easier to read. Please check if this is what you had in mind.

TomBaxter:
@felliott Moved fangorn.py function lazyLoadPreprocess to waterbutler.js wmLazyLoadPreprocess

website/static/js/fangorn.js
```diff
@@ -532,8 +546,22 @@ function doItemOp(operation, to, from, rename, conflict) {
  * @private
  */
 function _fangornResolveUploadUrl(item, file) {
+    // WB v1 update synatx is file path and kind=file
+    // WB v1 upload synatx is parent path and kind=file, name=filename
+    // If upload target file name already exists don't pass file.name
+    var updateUrl;
```
felliott: Minor: since `updateUrl` isn't used when `configOption` exists, don't declare it until after that branch.  IOW move this declaration to right before the `$.each()`.

website/static/js/waterbutler.js
```diff
@@ -22,67 +21,74 @@ function getDefaultOptions(path, provider) {
...
+    return wb_options;
 }
 
-function toJsonBlob(item, options) {
-    return $.extend(getDefaultOptions(item.data.path || '/', item.data.provider), {nid: item.data.nodeId}, options);
+// Move obj.data.attributes contents to obj.data , for WB v1 API
+function wbLazyLoadPreprocess(obj) {
+    // WB v1 API now stores file info in data.attributes.
+    // Treebeard expects to find it in data
+    function _promoteAttrs(obj_data) {
```
felliott: Sorry, I told you to declare this inside `wbLazyLoadPreprocess` and now I can't remember why.  The name is good, but can you move it outside `wbLazyLoadPreprocess`?  Also, can you expand on the comment?  Show what WB returns, and what format treebeard wants, so that devs who aren't familiar with tb/wb don't need to RTM to understand what this does.  Something like:

    // This function turns a WB structure into a TB structure
    // WB returns a structure like: data: { attributes: { name: 'foo', kind: 'file', ... } }
    // TB wants a structure like: data: { name: 'foo', kind: 'file', ... }

website/static/js/waterbutler.js
```diff
@@ -22,67 +21,74 @@ function getDefaultOptions(path, provider) {
...
+        var saved_attributes = obj_data.attributes;
+        delete obj_data.attributes;
```
felliott: Please add a comment, `// delete before extending in case saved_attributes has its own attributes property`.

website/static/js/fangorn.js
TomBaxter: Done.

website/static/js/waterbutler.js
TomBaxter: Done.

website/static/js/waterbutler.js
TomBaxter: Done.

felliott:
LGTM!

abought:
Possible regression report:

On the old version (develop branch as of this writing), I could select `second-branch` from the files tree page for a project with the Github addon, and get a list of files for the second branch.
http://localhost:7777/data?path=%2F&provider=github&nid=egm7s&ref=second-branch&_=1462200163517

screen shot 2016-05-02 at 10 42 45 am



With this new version checked out, it still shows me the list of files from the `master` branch and ignores my selection.
http://localhost:7777/v1/resources/egm7s/providers/github/?ref=second-branch&meta=&_=1462199577599
screen shot 2016-05-02 at 10 41 16 am

GH project for comparison:
screen shot 2016-05-02 at 10 41 47 am

brianjgeiger:
Closed in favor of #6337

PR #5242, opened 2016-03-21
[Feature] Django Upgrade

Ghalko:
Based on DRF changes in #5210 

Some nice things that will be changed in future PRs.


admin/spam/views.py
```diff
@@ -79,7 +81,13 @@ def get(self, request, *args, **kwargs):
         try:
             context = self.get_context_data(**kwargs)
         except AttributeError:
-            return page_not_found(request)  # TODO: 1.9 update to have exception with node/user 404.html will be added
+            return page_not_found(
+                request,
+                AttributeError(
+                    'Spam with id "{}" not found.'.format(kwargs.get('spam_id',
```
mfraezz: Why is this message wrapped in an `AttributeError` here...

admin/spam/views.py
```diff
@@ -90,7 +98,11 @@ def post(self, request, *args, **kwargs):
         try:
             context = self.get_context_data(**kwargs)
         except AttributeError:
-            return page_not_found(request)  # TODO: 1.9 update to have exception
+            return page_not_found(
+                request,
+                'Spam with id "{}" not found.'.format(kwargs.get('spam_id',
```
mfraezz: ...but not here?

mfraezz:
LGTM. Putting this in RTM :feet:

PR #5249, opened 2016-03-21
feature/Add spinner when comments are loading

erinspace:
# Purpose
Comments can take awhile to load - add a variable that gets set when they're done being requested and a spinner that shows until that variable is set to false

# Changes
Add ko observable variable for when all comments are done loading
use that variable in mako template to show if the comments are done loading or not

# Side effects
shouldn't be any


sloria:
Just one fix to make: the loading indicator should also be hidden when an error occurs. Once that's addressed, this is good to merge.

PR #5254, opened 2016-03-22
Update date_modified explicitly, preventing it to be updated on spam flagging

hmoco:


website/project/model.py
```diff
@@ -167,7 +167,7 @@ class Comment(GuidStoredObject, SpamMixin):
     root_target = fields.AbstractForeignField()
 
     date_created = fields.DateTimeField(auto_now_add=datetime.datetime.utcnow)
-    date_modified = fields.DateTimeField()
+    date_modified = fields.DateTimeField(default=lambda: datetime.datetime.utcnow())
```
sloria: Shouldn't this be `auto_now_add`?

sloria:

PR #5257, opened 2016-03-22
Fixed a case, changed button color and weight

billyhunt:
A couple of small button changes.

So 5 minutes ago
-

screen shot 2016-03-22 at 11 08 28 am
screen shot 2016-03-22 at 11 08 24 am

Fashion forward and case correct
-

screen shot 2016-03-22 at 11 12 01 am

screen shot 2016-03-22 at 11 08 41 am

### DIFF
--- a/website/static/css/pages/home-page.css
+++ b/website/static/css/pages/home-page.css
@@ -35,3 +35,8 @@ h3, h4 {
     background-color: #52A452;
     border-color: #54AC54;
 }
+
+.btn-view-meetings {
+    background-color: #52A452;
+    border-color: #54AC54;
+}
--- a/website/static/js/home-page/meetingsAndConferencesPlugin.js
+++ b/website/static/js/home-page/meetingsAndConferencesPlugin.js
@@ -22,7 +22,7 @@ var MeetingsAndConferences = {
                         ]
                     ),
                     m('.col-md-4.text-center',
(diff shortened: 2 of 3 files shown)


website/static/css/pages/home-page.css
```diff
@@ -35,3 +35,8 @@ h3, h4 {
     background-color: #52A452;
     border-color: #54AC54;
 }
+
+.btn-view-meetings {
```
sloria: Realized this after merging, but this can have a more generic name which describes what the class does, e.g. `btn-success-high-contrast`.

sloria: I'll just make this change.

website/static/js/home-page/meetingsAndConferencesPlugin.js
```diff
@@ -22,7 +22,7 @@ var MeetingsAndConferences = {
                         ]
                     ),
                     m('.col-md-4.text-center',
-                        m('div',  m('a.btn.btn-info.btn-lg.m-v-xl', { style : 'box-shadow: 0 0 9px -4px #000;', type:'button',  href:'/meetings/', onclick: function() {
+                        m('div',  m('a.btn.btn-info.btn-lg.btn-view-meetings.m-v-xl.f-w-xl', { style : 'box-shadow: 0 0 9px -4px #000;', type:'button',  href:'/meetings/', onclick: function() {
```
sloria: I think `btn-info` can be removed, now that you've added a class which overrides the colors.

I can make this change.

PR #5258, opened 2016-03-22
Reverse sorting on /users/nodes/ by title gives 500 [OSF-5237]

KAsante95:
# Purpose

Enable reverse sorting based on user node attributes such as title and date_created

Issue:
https://openscience.atlassian.net/browse/OSF-5237

# Code

Removed fields.pop(0) and simply selected the first index via fields[0].
I then included a reverse sort for the case in which the first character in the field was a '-', thus indicating a reverse sort. To implement the reverse sort, I simply switched the places of the 1 and -1 from the original sort.

### DIFF
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -27,7 +27,7 @@ def sort_multiple(fields):
         rev = 1
         num = 1
         i = 0
-        for field in fields:
+        for index, field in enumerate(fields):
             i += 1
             if field[0] == '-':
                 rev = -1
@@ -39,7 +39,10 @@ def sort_multiple(fields):
             elif a_field < b_field:
                 return -1 * rev
             elif a_field == b_field:
-                continue
+                if index < len(fields) - 1:
+                    continue
+                else:
+                    return 0
         return 0
     return sort_fn
 
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -25,7 +25,10 @@ def sort_multiple(fields):
     fields = list(fields)
     def sort_fn(a, b):
         rev = 1
+        num = 1
+        i = 0
         for field in fields:
+            i += 1
             if field[0] == '-':
                 rev = -1
                 field = field[1:]
@@ -36,9 +39,10 @@ def sort_multiple(fields):
             elif a_field < b_field:
                 return -1 * rev
             elif a_field == b_field:
-                return 0
+                continue
         return 0
     return sort_fn
+
 
 
 class ODMOrderingFilter(OrderingFilter):
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -35,7 +35,6 @@ def sort_multiple(fields):
                 return 1 * sort_direction
             elif a_field < b_field:
                 return -1 * sort_direction
-            continue
         return 0
     return sort_fn
 
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -35,8 +35,7 @@ def sort_multiple(fields):
                 return 1 * sort_direction
             elif a_field < b_field:
                 return -1 * sort_direction
-            elif a_field == b_field:
-                continue
+            continue
         return 0
     return sort_fn
 
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -25,7 +25,7 @@ def sort_multiple(fields):
     fields = list(fields)
     def sort_fn(a, b):
         sort_direction = 1
-        for index, field in enumerate(fields):
+        for field in fields:
             if field[0] == '-':
                 sort_direction = -1
                 field = field[1:]
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -25,7 +25,6 @@ def sort_multiple(fields):
     fields = list(fields)
     def sort_fn(a, b):
         rev = 1
-        num = 1
         i = 0
         for index, field in enumerate(fields):
             i += 1
@@ -46,11 +45,8 @@ def sort_multiple(fields):
         return 0
     return sort_fn
 
-
-
 class ODMOrderingFilter(OrderingFilter):
     """Adaptation of rest_framework.filters.OrderingFilter to work with modular-odm."""
-
     # override
     def filter_queryset(self, request, queryset, view):
         ordering = self.get_ordering(request, queryset, view)
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -24,19 +24,17 @@ from api.base.serializers import RelationshipField, TargetField
 def sort_multiple(fields):
     fields = list(fields)
     def sort_fn(a, b):
-        rev = 1
-        i = 0
+        sort_direction = 1
         for index, field in enumerate(fields):
-            i += 1
             if field[0] == '-':
-                rev = -1
+                sort_direction = -1
                 field = field[1:]
             a_field = getattr(a, field)
             b_field = getattr(b, field)
             if a_field > b_field:
-                return 1 * rev
+                return 1 * sort_direction
             elif a_field < b_field:
-                return -1 * rev
+                return -1 * sort_direction
             elif a_field == b_field:
                 if index < len(fields) - 1:
                     continue
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -36,10 +36,7 @@ def sort_multiple(fields):
             elif a_field < b_field:
                 return -1 * sort_direction
             elif a_field == b_field:
-                if index < len(fields) - 1:
-                    continue
-                else:
-                    return 0
+                continue
         return 0
     return sort_fn
 
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -24,26 +24,19 @@ from api.base.serializers import RelationshipField, TargetField
 def sort_multiple(fields):
     fields = list(fields)
     def sort_fn(a, b):
+        rev = 1
         for field in fields:
             if field[0] == '-':
+                rev = -1
                 field = field[1:]
-                a_field = getattr(a, field)
-                b_field = getattr(b, field)
-                if a_field > b_field:
-                    return -1
-                elif a_field < b_field:
-                    return 1
-                elif a_field == b_field:
-                    return 1
-            else:
-                a_field = getattr(a, field)
-                b_field = getattr(b, field)
-                if a_field > b_field:
-                    return 1
-                elif a_field < b_field:
-                    return -1
-                elif a_field == b_field:
-                    return 1
+            a_field = getattr(a, field)
+            b_field = getattr(b, field)
+            if a_field > b_field:
+                return 1 * rev
+            elif a_field < b_field:
+                return -1 * rev
+            elif a_field == b_field:
+                return 0
         return 0
     return sort_fn
 
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -24,8 +24,7 @@ from api.base.serializers import RelationshipField, TargetField
 def sort_multiple(fields):
     fields = list(fields)
     def sort_fn(a, b):
-        while fields:
-            field = fields[0]
+        for field in fields:
             if field[0] == '-':
                 field = field[1:]
                 a_field = getattr(a, field)
@@ -34,6 +33,8 @@ def sort_multiple(fields):
                     return -1
                 elif a_field < b_field:
                     return 1
+                elif a_field == b_field:
+                    return 1
             else:
                 a_field = getattr(a, field)
                 b_field = getattr(b, field)
@@ -41,6 +42,8 @@ def sort_multiple(fields):
                     return 1
                 elif a_field < b_field:
                     return -1
+                elif a_field == b_field:
+                    return 1
         return 0
     return sort_fn
 
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -25,15 +25,25 @@ def sort_multiple(fields):
     fields = list(fields)
     def sort_fn(a, b):
         while fields:
-            field = fields.pop(0)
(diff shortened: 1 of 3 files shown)


api_tests/base/test_filters.py
```diff
@@ -222,3 +229,51 @@ def test_convert_value_float(self):
...
+    def test_filter_queryset_fails(self):
+        class query:
+            title = ' '
+            def __init__(self, title):
+                self.title = title
+            def __str__(self):
+                return self.title
+        query_s = []
+        q1 = query('NewProj')
+        query_s.append((q1))
```
whit537: What's up with the extra parens here? Let's drop them here and throughout these two new tests.

api_tests/base/test_filters.py
```diff
@@ -222,3 +229,51 @@ def test_convert_value_float(self):
...
+        q2 = query('Zip')
+        query_s.append((q2))
+        q3 = query('Proj')
+        query_s.append((q3))
+        q4 = query('Activity')
+        query_s.append((q4))
+        sort_q = sorted(query_s, cmp=filters.sort_multiple(['-title']))
+        sortt = []
+        for i in sort_q:
+            sortt.append(str(i))
```
whit537: This is a great place for a list comprehension:

```python
sortt = [str(i) for i in sort_q]
```

whit537:
#### Punchlist

 - [ ] remove extra parens—https://github.com/CenterForOpenScience/osf.io/pull/5258#discussion_r57052684
 - [ ] use a list comprehension—https://github.com/CenterForOpenScience/osf.io/pull/5258#discussion_r57052793

api_tests/base/test_filters.py
```diff
@@ -222,3 +229,51 @@ def test_convert_value_float(self):
...
+                return self.title
+        query_s = []
+        q1 = query('NewProj')
+        query_s.append((q1))
+        q2 = query('Zip')
+        query_s.append((q2))
+        q3 = query('Activity')
+        query_s.append((q3))
+        q4 = query('Activity')
+        query_s.append((q4))
```
whit537: You can greatly abbreviate this code by combining a couple Pythonic idioms:

```python
query_s = [query(x) for x in 'NewProj Zip Activity Activity'.split()]
```

api/base/filters.py
```diff
@@ -24,17 +24,30 @@
...
+            if field[0] == '-':
+                field = field[1:]
+                a_field = getattr(a, field)
+                b_field = getattr(b, field)
+                if a_field > b_field:
+                    return -1
+                elif a_field < b_field:
+                    return 1
+                elif a_field == b_field:
+                    return 1
```
whit537: Do you really mean `1` here? Usually when comparing, equality results in `0`, as in line 47.

whit537: So, actually, the way this function had been implemented depends on the `0` case falling outside the loop. Why did you move the equality comparison inside the loop?

api_tests/base/test_filters.py
```diff
@@ -222,3 +229,51 @@ def test_convert_value_float(self):
...
+        sort_q = sorted(query_s, cmp=filters.sort_multiple(['-title']))
+        sortt = []
+        for i in sort_q:
+            sortt.append(str(i))
+        assert_equal((sortt) , ['Zip', 'Proj', 'NewProj', 'Activity'])
+    
+    def test_filter_queryset_duplicate_fails(self):
+        class query:
```
whit537: Since we duplicate this helper class definition, let's move it up into the scope of the containing class, so we can access it as `self.query` in both tests.

api/base/filters.py
whit537: The (an?) error here is in sorting based on multiple fields; here's a field for a failing test (also illustrating the suggestion at https://github.com/CenterForOpenScience/osf.io/pull/5258#discussion_r57056259):

```
diff --git a/api_tests/base/test_filters.py b/api_tests/base/test_filters.py
index 7bcfefb..a9a4d4d 100644
--- a/api_tests/base/test_filters.py
+++ b/api_tests/base/test_filters.py
@@ -277,3 +277,15 @@ class TestODMOrderingFilter(ApiTestCase):
             sortt.append(str(i))
         assert_equal((sortt) , ['Zip', 'NewProj', 'Activity', 'Activity'])
 
+
+    class obj:
+        def __init__(self, **kw):
+            self.__dict__.update(**kw)
+
+    def test_filter_queryset_handles_multiple_fields(self):
+        objs = [self.obj(title='NewProj', n=10),
+                self.obj(title='Zip', n=20),
+                self.obj(title='Activity', n=30),
+                self.obj(title='Activity', n=40)]
+        actual = [x.n for x in sorted(objs, cmp=filters.sort_multiple(['title', '-number']))]
+        assert_equal(actual, [40, 30, 10, 20])
```

api/base/filters.py
```diff
@@ -24,17 +24,30 @@
...
+            else:
+                a_field = getattr(a, field)
+                b_field = getattr(b, field)
+                if a_field > b_field:
+                    return 1
+                elif a_field < b_field:
+                    return -1
+                elif a_field == b_field:
+                    return 1
```
whit537: I'd like to see [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)er code here. We should be able to implement this without duplicating the `getattr` calls, nor the `>` and `<` comparisons (hint: set a flag to `1` or `-1` based on the `field[0]` comparison and multiply the `return` value by that).

whit537:
Ok @KAsante95, pass done and I've left you a [punchlist](https://github.com/CenterForOpenScience/osf.io/pull/5258#issuecomment-199978625) for you. Ping me when you're ready for more review! :-)

api_tests/base/test_filters.py
```diff
@@ -222,3 +229,26 @@ def test_convert_value_float(self):
...
+        def __init__(self, title):
+            self.title = title
+        def __str__(self):
+            return self.title
+
+    def test_filter_queryset_fails(self):
+        query_s = [self.query(x) for x in 'NewProj Zip Proj Activity'.split()]
+        sort_q = sorted(query_s, cmp=filters.sort_multiple(['-title']))
+        sortt = [str(i) for i in sort_q]
+        assert_equal((sortt) , ['Zip', 'Proj', 'NewProj', 'Activity'])
```
whit537: We've still got extra parentheses around `sortt` here.

api/base/filters.py
```diff
@@ -24,26 +24,19 @@
...
-                elif a_field == b_field:
-                    return 1
+            a_field = getattr(a, field)
+            b_field = getattr(b, field)
+            if a_field > b_field:
+                return 1 * rev
+            elif a_field < b_field:
+                return -1 * rev
+            elif a_field == b_field:
+                return 0
```
whit537: This will still fail a test like the one at https://github.com/CenterForOpenScience/osf.io/pull/5258#discussion_r57059485. Let's go ahead and add that test to the test suite ... and then make it pass! :-)

whit537:
[Punchlist](https://github.com/CenterForOpenScience/osf.io/pull/5258#issuecomment-199978625) updated! Back over to you, @KAsante95. :-)

api/base/filters.py
```diff
@@ -24,20 +24,29 @@
 def sort_multiple(fields):
     fields = list(fields)
     def sort_fn(a, b):
-        while fields:
-            field = fields.pop(0)
+        rev = 1
+        i = 0
```
brianjgeiger: You're not using `i` at all, so you should be able to remove it.

api/base/filters.py
```diff
@@ -24,20 +24,29 @@
 def sort_multiple(fields):
     fields = list(fields)
     def sort_fn(a, b):
-        while fields:
-            field = fields.pop(0)
+        rev = 1
```
brianjgeiger: Let's make rev something more readable. How about rename it into `sort_direction`, then define `FORWARD = 1` and `REVERSE = -1`. Then you can set `sort_direction = REVERSE` and `return 1 * sort_direction`, for example.

api_tests/base/test_filters.py
```diff
@@ -222,3 +229,43 @@ def test_convert_value_float(self):
...
+        title = ' '
+        number = 0
+        def __init__(self, title, number):
+            self.title = title
+            self.number = number
+        def __str__(self):
+            return self.title
+
+    def test_filter_queryset_fails(self):
+        query_s = [self.query(x) for x in 'NewProj Zip Proj Activity'.split()]
```
brianjgeiger: I am not a big fan of vague variable names. Could you change `query_s`, `sort_q`, and `sortt` into things that are very clear as to their purpose? Same with anything else similar.

brianjgeiger:
Make sure you have explicit tests for forward and reverse sorting, and also add an API view test to check explicitly for the 500 error from [OSF-5237]. You may have to try the test on the develop branch to see that it 500s there, then ensure that the test passes on your branch.

brianjgeiger:
:penguin: Pass done.

hmoco:
@sloria, as pointed out by Dawn, the bug I fixed in my PR seems to be fixed better and more throughly tested here. Could we see that this gets merged?

sloria:
@hmoco Sure. Will merge this after it's through CR.

api_tests/base/test_filters.py
```diff
@@ -222,3 +229,58 @@ def test_convert_value_float(self):
...
+            self.number = number
+        def __str__(self):
+            return self.title
+
+
+    def test_filter_queryset_forward(self):
+        query_to_be_sorted = [self.query(x) for x in 'NewProj Zip Proj Activity'.split()]
+        sorted_query = sorted(query_to_be_sorted, cmp=filters.sort_multiple(['title']))
+        sorted_output = [str(i) for i in sorted_query]
+        assert_equal((sorted_output) , ['Activity', 'NewProj', 'Proj', 'Zip'])
```
whit537: Why are we still using extra parens here?

api/base/filters.py
```diff
@@ -24,20 +24,27 @@
...
-                return 1
+                return 1 * sort_direction
             elif a_field < b_field:
-                return -1
+                return -1 * sort_direction
+            elif a_field == b_field:
+                if index < len(fields) - 1:
+                    continue
+                else:
+                    return 0
```
whit537: I believe the last conditional here is unnecessary (our tests pass without it); let's remove it.

That is, in the case where `index == len(fields) - 1`, we will be at the end of the `for` loop, and we'll hit the immediately subsequent `return 0`. In the `<` case, the explicit `continue` doesn't have any effect because we're already at the end of the `for` block; there are no additional statements that we're skipping by explicitly `continue`ing.

whit537: (See also: https://github.com/CenterForOpenScience/osf.io/pull/5258#discussion_r57055035.)

whit537:
[Punchlist](https://github.com/CenterForOpenScience/osf.io/pull/5258#issuecomment-199978625) updated! Back to you, @KAsante95 ... almost there! :-)

api/base/filters.py
```diff
@@ -24,20 +24,24 @@
...
             a_field = getattr(a, field)
             b_field = getattr(b, field)
             if a_field > b_field:
-                return 1
+                return 1 * sort_direction
             elif a_field < b_field:
-                return -1
+                return -1 * sort_direction
+            elif a_field == b_field:
+                continue
```
whit537: The last branch here is extraneous. Let's remove it.

api_tests/base/test_filters.py
```diff
@@ -222,3 +229,58 @@ def test_convert_value_float(self):
...
+            self.number = number
+        def __str__(self):
+            return self.title
+
+
+    def test_filter_queryset_forward(self):
+        query_to_be_sorted = [self.query(x) for x in 'NewProj Zip Proj Activity'.split()]
+        sorted_query = sorted(query_to_be_sorted, cmp=filters.sort_multiple(['title']))
+        sorted_output = [str(i) for i in sorted_query]
+        assert_equal(sorted_output , ['Activity', 'NewProj', 'Proj', 'Zip'])
```
whit537: There shouldn't be a space between "`sorted_output`" and the comma.

whit537: (Sorry for not flagging this earlier.)

whit537:
[Punchlist](https://github.com/CenterForOpenScience/osf.io/pull/5258#issuecomment-199978625) updated.

api/base/filters.py
```diff
@@ -24,20 +24,23 @@
...
+                field = field[1:]
             a_field = getattr(a, field)
             b_field = getattr(b, field)
             if a_field > b_field:
-                return 1
+                return 1 * sort_direction
             elif a_field < b_field:
-                return -1
+                return -1 * sort_direction
+            continue
```
whit537: `continue` is unnecessary at the end of a loop. It would only be used if you wanted to skip some additional code in the block of the loop. When there's nothing to skip, it's unneeded. Let's remove it! :-)

api/base/filters.py
```diff
@@ -24,20 +24,22 @@
 def sort_multiple(fields):
     fields = list(fields)
     def sort_fn(a, b):
-        while fields:
-            field = fields.pop(0)
+        sort_direction = 1
+        for index, field in enumerate(fields):
```
whit537: Gosh, I hate to keep nitpicking ... but we no longer need the enumeration here, since we're no longer using `index` as of be85a83ccbda1f2ee12ea7d42facd0d160a19cab.

I promise we're close, @KAsante95! ☺️

whit537:
Okay! We're a little weak on the unit tests, but some is better than none. I'm good to go here. 👍

PR #5261, opened 2016-03-22
Reorder Categories from Newest to Oldest

billyhunt:
The categories on the new dashboard should list in order of most recently updated to least. This PR sorts according to the most recent date modified. So far, it does not refresh when a project is added to a category. I suspect this is due to the redrawing bug on that bar, but am not sure. It will have to be addressed at some point. (When you add a project to a collection, that collection should pop to the top)


sloria:
Could you also modify the API call here (https://github.com/CenterForOpenScience/osf.io/blob/98df8b4b87a68456e806b668aac3929e829df034/website/static/js/myProjects.js#L613) to return the collections sorted by date_modified in descending order? You might still have to change the position of "Bookmarks" to be on top in the front-end, but at least you wouldn't have to sort the other results.

sloria:
Pass finished.

website/static/js/myProjects.js
```diff
@@ -554,6 +559,22 @@ var MyProjects = {
             self.breadcrumbs().push(linkObject);
         };
 
+        var sortCollections = function(collections) {
+            var collectionsToSort = [];
+            var bookmarksCollection = {};
+            for (var i=0; i<collections.length; i++) {
```
sloria: You shouldn't need to loop through all the collections in order to remove the Bookmarks collection.

Also, since all the other items should already be sorted, you probably don't need to check `date_modified`. Maybe try this: https://stackoverflow.com/questions/23921683/javascript-move-an-item-of-an-array-to-the-front

sloria:
Pass finished.

billyhunt:
I couldn't find a faster way to pull based on an attribute of an object (not the entire thing), but I removed the sort and kick out of the loop when I find bookmarks.

sloria:
@billyhunt Did you see the SO page I linked to above? Couldn't you do something like:

```js
var first = 'should be first';
data.sort(function(x,y){ return x.someAttr == first ? -1 : y.someAttr == first ? 1 : 0; });
```

website/static/js/myProjects.js
```diff
@@ -899,6 +920,7 @@ var Collections = {
                             }
                             if(!skipCount){
                                 collection.data.count(collection.data.count()+1);
+                                self.loadCollections;
```
sloria: Why was this added?

billyhunt:
I am probably missing something, but it seems like your solution would be slower (has to go through the whole list) and harder to debug later.

billyhunt:
This is going to be put on hold while the code is in flux and pushed back to the next release.

PR #5262, opened 2016-03-22
Release/Fix unread count when moving file between nodes

samanehsan:
Purpose:
------------
When moving a file to a different component, the `node` field for all of its comments get updated to the destination node. This updated the `date_modified` field, which resulted in all of the comments appearing as if they were unread.

Requires: https://github.com/CenterForOpenScience/osf.io/pull/5254 (removes the auto-update of `date_modified`)

Changes
------------
Update the `user.comments_viewed_timestamp` so that `user.comments_viewed_timestamp[destination_node._id]['files'][root_target_id] == user.comments_viewed_timestamp[source_node._id]['files'][root_target_id]`

### DIFF
--- a/website/project/views/comment.py
+++ b/website/project/views/comment.py
@@ -23,7 +23,7 @@ from website import settings
 
 
 @file_updated.connect
-def update_file_guid_referent(self, node, event_type, payload, user=None):
+def update_file_guid_referent(self, node, user, event_type, payload):
     if event_type == 'addon_file_moved' or event_type == 'addon_file_renamed':
         source = payload['source']
         destination = payload['destination']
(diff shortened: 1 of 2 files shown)


website/project/views/comment.py
```diff
@@ -23,7 +23,7 @@
 
 
 @file_updated.connect
-def update_file_guid_referent(self, node, event_type, payload, user=None):
+def update_file_guid_referent(self, node, user, event_type, payload):
```
sloria: Good call on making `user` a required argument :+1: .

sloria:
Pass finished. Looks good to merge when the build passes. :octopus: 

As discussed in person, this change will be unnecessary if we change `comments_viewed_timestamp` to be keyed on GUID rather than node ID. Saman will send a follow-up PR implementing this refactor.

samanehsan:
The follow-up PR is #5268

samanehsan:
Replaced by #5268

PR #5268, opened 2016-03-22
Feature: Change structure of User.comments_viewed_timestamps

samanehsan:
Purpose
-----------
Since `user.comments_viewed_timestamps` organizes timestamps under a `node_id` top-level key, when a file moves nodes the field needs to be updated and that locks the user. Instead, make the timestamps a dictionary of `{target_id: timestamp}`

Changes
------------
- Change structure of user.comments_viewed_timestamps from {'node_id': {'node': timestamp, 'files': {'file_id': timestamp}}} to {'node_id: timestamp, file_id: timestamp, ...}.
- Need to run `scripts.update_comments_viewed_timestamp`

### DIFF
--- a/scripts/update_comments.py
+++ b/scripts/update_comments.py
@@ -1,6 +1,5 @@
 """
-Update User.comments_viewed_timestamp field & comments model.
-Accompanies https://github.com/CenterForOpenScience/osf.io/pull/1762
+Update User.comments_viewed_timestamp field.
 """
 import logging
 import sys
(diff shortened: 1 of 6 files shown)


scripts/update_comments.py
```diff
@@ -19,25 +18,29 @@
...
+    users = User.find(Q('comments_viewed_timestamp', 'ne', None) & Q('comments_viewed_timestamp', 'ne', {}))
     for user in users:
         if user.comments_viewed_timestamp:
-            for node in user.comments_viewed_timestamp:
-                user.comments_viewed_timestamp[node] = {'node': user.comments_viewed_timestamp[node]}
-            user.save()
-            logger.info('Migrated timestamp for user {0}'.format(user._id))
+            timestamps = {}
+            for node_id in user.comments_viewed_timestamp:
+                node_timestamps = user.comments_viewed_timestamp[node_id]
```
sloria: Minor: You can do

```python
for node_id, node_timestamps in user.comments_viewed_timestamp.items():
```

but it's not a big deal.

PR #5271, opened 2016-03-22
fix loading not ending when the user has no institutions + fixes

hmoco:


website/static/js/institutionProjectSettings.js
```diff
@@ -20,10 +20,8 @@ var ViewModel = function(data) {
             url,
             {isCors: true}
         ).done(function (response) {
-            if (response.data.embeds.institutions.data.length){
-                self.availableInstitutions(response.data.embeds.institutions.data);
-                self.loading(false);
-            }
+            self.availableInstitutions(response.data.embeds.institutions.data.length ? response.data.embeds.institutions.data: []);
+            self.loading(false);
```
sloria: Should these also be set when the request fails?

hmoco: Not necessarily. if the request fails, we get a message (through Raven), and the settings page never loads. The alternative is to render that there is an error, with an error message in place of whatever would go in the box.

sloria: Ah I see. Thanks for the clarification. Yes, there should be a an error message displayed to the user if this fails. Can you add that to this PR?

sloria:
Pass finished.

hmoco:
@sloria responded, also a few small fixes thrown in here. The Search wont properly work for a bit though.

hmoco:
also commits are super small each, can squash if needed.

website/templates/project/settings.mako
```diff
@@ -306,32 +306,39 @@
                     <div class="panel-body">
                         % if not node['institution']['name']:
                             <div data-bind="visible: !loading()">
-                                <div data-bind="visible: availableInstitutions().length">
+                                <div data-bind="visible: error()">
                                     <div class="help-block">
-                                        Projects affiliated with institutions will show some institutional branding (such as logos) and if public, will be discoverable on OSF institutional landing pages.
-
-                                        You are authorized to affiliate your projects with the following institutions:
+                                        Failed to load up your institutions. Please try again.
```
sloria: Keep your error messages consistent with the rest of the site. For example, the message should include a mailto link to the support email.

sloria:
Any Trello cards associated with these changes? Please follow guidelines for PR descriptions.

sloria:
Pass finished.

sloria:
Jshint error was unrelated to this PR.

@hmoco Please move any relevant Trello cards if need be.

PR #5272, opened 2016-03-22
feature/load comments incrementally

erinspace:
# Purpose

At the moment, comments don't load unless all of the API requests have been made. This PR makes it so that comments are revealed as they come back from the API - and the requests continue in the background

# Changes
Remove the check that there are no more links to load from the API - resolves the request and appends to the comments array, and makes it visible on the very first request that comes back. if there are more requests to make, it still runs recursively and continues to add to the array of comments already displayed.

# Side Effects
- Makes comment loading a bit jumpier than just loading them all at once when all the API requests had been made.

### Questions
- This still uses deferred promises - could this be done smoother by removing the promise components altogether?

### DIFF
--- a/website/static/js/comment.js
+++ b/website/static/js/comment.js
@@ -137,10 +137,6 @@ BaseComment.prototype.fetch = function() {
 /* Go through the paginated API response to fetch all comments for the specified target */
 BaseComment.prototype.fetchNext = function(url, comments) {
     var self = this;
-    var deferred = $.Deferred();
-    if (self._loaded) {
-        deferred.resolve(self.comments());
-    }
     var request = osfHelpers.ajaxJSON(
         'GET',
         url,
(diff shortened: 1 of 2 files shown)


website/static/js/comment.js
```diff
@@ -147,19 +147,20 @@ BaseComment.prototype.fetchNext = function(url, comments) {
...
+        if (self._loaded !== true) {
             self._loaded = true;
             self.loadingComments(false);
         }
+        self.comments(
+            ko.utils.arrayMap(comments, function(comment) {
+                return new CommentModel(comment, self, self.$root);
+            })
+        );
+        deferred.resolve(self.comments());
```
sloria: This promise shouldn't get resolved until all the requests have finished, i.e. when there is no more next link.

website/static/js/comment.js
```diff
@@ -147,19 +147,20 @@ BaseComment.prototype.fetchNext = function(url, comments) {
...
+        self.configureCommentsVisibility();
```
sloria: This method should also only need to be run after all the requests have finished.

website/static/js/comment.js
```diff
@@ -147,19 +147,20 @@ BaseComment.prototype.fetchNext = function(url, comments) {
...
-                })
-            );
-            deferred.resolve(self.comments());
-            self.configureCommentsVisibility();
+        if (self._loaded !== true) {
             self._loaded = true;
             self.loadingComments(false);
         }
+        self.comments(
+            ko.utils.arrayMap(comments, function(comment) {
```
sloria: I wonder if the "jumpiness" is due to the fact that we are serializing *all* comments every time a new page of comments is received. Instead, you should only need to serialize comments as they come.

sloria:
Pass finished.

sloria:
https://trello.com/c/WwqicbTk/50-load-comments-incrementally

PR #5273, opened 2016-03-22
[WIP] Prelim Keen IO Analytics Page

GageGaskins:
# Purpose
Adds preliminary data to analytics page using Keen. Every project uses a scoped key as to avoid potential read abuse by users.

# Side Effects
None


website/templates/project/statistics.mako
```diff
@@ -2,34 +2,76 @@
...
+                <div class="text-center">
+                    <div class="logo-spin logo-lg"></div>
+                </div>
+            </div>
         </div>
     </div>
-% else:
-    % if not node['is_public']:
-        <div class='alert alert-warning'>
-            <strong>Note:</strong> Usage statistics are collected only for public resources.
```
felliott: We should still be showing this message if the project is not public.

website/static/js/statistics.js
```diff
@@ -0,0 +1,121 @@
...
+    self.init = function () {
+        self.visitsByDay();
+        self.topReferrers();
+        self.visitsServerTime();
+    };
+
+
+};
+
+function Statistics() {
```
felliott: `Statistics` is too generic a name.  How about `PublicProjectUsageStats` or something like that?

website/templates/project/statistics.mako
felliott: I'm an idiot.  I thought this was saying statistics are only *shown* for public resources.  You were correct to remove it.

PR #5274, opened 2016-03-22
toggle breadcrumbs to use use component or project modal

billyhunt:
Currently, if you add a component via breadcrumbs, it will pop a Add project dialog.  Not the oh so needed Add component dialog.  This should make that magic happen.

:(
-

screen shot 2016-03-22 at 5 02 30 pm

:)
-

screen shot 2016-03-22 at 5 02 39 pm

### DIFF
--- a/website/static/js/myProjects.js
+++ b/website/static/js/myProjects.js
@@ -383,6 +383,7 @@ var MyProjects = {
                 self.data(value.data);
             }
             if(!value.data[0]){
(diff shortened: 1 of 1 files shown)


sloria:
https://trello.com/c/DbG2MeuR/132-create-component-button-now-incorrectly-leads-to-new-project-modal

sloria:
Thanks!

PR #5277, opened 2016-03-23
Inital attempt to increase performance for embeds

chrisseto:


brianjgeiger:
This seems pretty cool. Since you're going through the view, and since all the tests pass, most of my potential concerns are allayed. Out of curiosity, what's an example of DRF junk we're avoiding?

chrisseto:
@brianjgeiger about 40K calls to `copy.deepcopy`

PR #5278, opened 2016-03-23
[Feature] Admin search for users

Ghalko:
## Proposal

Connect into elastic search to be able to search for users. This will also display a list of the found users.

## Repercussions

No longer need to find users on the OSF by using its search utility. Centralizing authority.


Ghalko:
Closing for now. Definitely not a priority since this is on the OSF.

PR #5280, opened 2016-03-23
[RfM] Add ability to attach multiple files to the preregistration form [OSF-5586]

barbour-em:
Refs: https://openscience.atlassian.net/browse/OSF-5586

## Purpose
Allow users of the prereg form to add more than one file to an answer

## Changes

In the form:
screen shot 2016-03-23 at 10 27 49 am

In the preview:
screen shot 2016-03-23 at 10 44 20 am

### DIFF
--- a/website/static/js/registrationEditorExtensions.js
+++ b/website/static/js/registrationEditorExtensions.js
@@ -62,8 +62,6 @@ var osfUploader = function(element, valueAccessor, allBindings, viewModel, bindi
         if (item.kind === 'file') {
             viewModel.selectedFiles.push(item);
             item.css = 'fangorn-selected';
-        } else {
-            viewModel.selectedFiles([]);
         }
     };
     var fw = new FilesWidget(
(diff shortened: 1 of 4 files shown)


mfraezz:
This seems to not allow the removal of the last selected file, even though it can be deleted through the treebeard interface.
screen shot 2016-03-25 at 11 28 08

website/static/js/registrationEditorExtensions.js
```diff
@@ -169,45 +167,73 @@ var Uploader = function(question) {
...
-    self.unselectFile = function() {
-        self.selectedFile(null);
-        question.extra({
-            selectedFileName: NO_FILE
-        });
+    self.unselectFile = function(fileToRemove) {
+
+        var files = question.extra();
+
+        for (var i = 0; i < files.length - 1; i++) {
```
mfraezz: This prevents a user from removing the last selected file. `i < files.length` or `i <= files.length - 1` would fix.

website/static/js/registrationUtils.js
```diff
@@ -182,7 +182,13 @@ var Question = function(questionSchema, data) {
     self.properties = questionSchema.properties || [];
     self.match = questionSchema.match || '';
 
-    self.extra = ko.observable(self.data.extra || {});
+    self.extra = ko.observableArray(self.data.extra || []);
+
+    self.formattedFileList = ko.computed(function() {
```
mfraezz: This should be a `ko.pureComputed`, otherwise the file info doesn't get saved.

mfraezz:
Confirmed that, with some changes, everything is stored properly. A couple of minor issues to address, otherwise looks good. :cactus:

This also may create an issue with the admin module's prereg curation -- the attached files are not displayed. You may want to collaborate with @Ghalko for that.

barbour-em:
@mfraezz what is the expected behavior for the attached files in the admin module? I see this when viewing a submitted draft:

screen shot 2016-03-28 at 9 28 47 am

Other issues are addressed in dd14a9

mfraezz:
Hmm. Links to attached files aren't showing up for me -- I attached one or more files to each question that accepted them, but all I get is the answer I put: 
screen shot 2016-03-29 at 10 27 02

Confirmed that everything is stored properly in the database, so it could just be something weird with my local config. Regardless, if it is an issue it seems like an admin module issue, probably not within scope for this PR.

Otherwise, changes look good. Dropping this in RTM :deciduous_tree:

PR #5281, opened 2016-03-23
[Ready for review] Fix file comment banning

cwisecarver:
This should fix the bug with dataverse and figshare (among others) comment unread counts not updating. 

It would be nice if when we replaced the v1 route for updating the timestamps for unread comment counts it was related more directly to the object being commented upon instead of the user. The current setup causes many logic branches to handle the fact that the URL/object being modified is the user but the object we're expected to see change is the File, Node, etc.

@sloria @nicipfeiffer @samanehsan


samanehsan:
I just sent in a PR that simplified `user.comments_viewed_timestamp` so that it's just a dictionary of `{root_id: timestamp}` so this endpoint won't have as many logic branches.

sloria:
cherry picked into release/0.66.0.

PR #4952, opened 2016-01-27
[feature] Refactor Retraction to Withdrawal

caseyrollins:
- Changes 'retracted' to 'withdrawn' in user-facing language.
- Refactors routes to /withdraw/ instead of /retraction/
- [OSF-5542]

PR #5285, opened 2016-03-23
[feature] Refactor Retract to Withdrawal  

caseyrollins:
- Refactor `retract` to `withdraw` in api and admin code.
- Relates to #4952


sloria:
Are there any API docs that also need to be updated?

sloria:
This is fine to merge for now, but please update any API docs as necessary.

PR #5286, opened 2016-03-23
Open 'edit' panel on project wiki by default if Home wiki is empty

kch8qx:
[OSF-5363]


mfraezz:
Pass complete, no bugs found :fire: 

Please add a test (or an assert to an existing test) to ensure that both panels are open when they should be.

mfraezz:
LGTM. Dropping this in RTM :shipit:

PR #4979, opened 2016-02-03
[rfr][feature] improve dataverse generics/structure [OSF-5367]

mfraezz:
[#OSF-5367]

Purpose
=======
A refactor to addon architecture abstracted much of the code repeated across addons. This PR applies those abstractions to Dataverse, when possible.

Changes
=======
Addon structure made more similar to other addons.
- [x] `generic_views` used when possible
- [x] models updated accordingly
- [x] tests updated to use mixins
- [x] Write migration script for `ExternalAccount` -- `.display_name` and `.oauth_key`
    * `if host=='dataverse-demo.iq.harvard.edu'` -> `host='demo.dataverse.org'`

Side Effects
=========
Legacy fields (`foreign_user_settings` and others) on the `AddonDataverseNodeSettings` model removed

Migration Steps
----------------------
`python -m  scripts.dataverse.migrate_demo_host`

PR #5287, opened 2016-03-23
Remove misc unused backrefs

abought:
Refs various, general parent is https://openscience.atlassian.net/browse/OSF-5674 

## Purpose
We are attempting to reduce the number of unused backrefs on our models, as backrefs create side effects (record locking) that are not ideal.

This PR is intended to encompass the low-hanging fruit: backrefs that aren't used at all, in various places.

## Summary of changes
- Remove various unused backrefs (see diff)

## Side effects/ testing notes
Should be no side effects because these backrefs were unused. Migration probably unnecessary. Where relevant, below are notes on how to hit endpoints where these models are used.

- The `MailRecord` model is used to track email submissions to conferences (OSF For Meetings); I wasn't able to test this locally.

- Also try creating embargoed and retratcted registrations, as well as approving a registration when asked to do so via email

- Very quickly confirm the box addon works as expected (usersettings recognized)

## Subtasks

https://openscience.atlassian.net/browse/OSF-5897
https://openscience.atlassian.net/browse/OSF-5902
https://openscience.atlassian.net/browse/OSF-5903
https://openscience.atlassian.net/browse/OSF-5905
~~https://openscience.atlassian.net/browse/OSF-5909~~ (moved to a different PR)
https://openscience.atlassian.net/browse/OSF-5910
https://openscience.atlassian.net/browse/OSF-5918

### DIFF
--- a/website/addons/badges/model/badges.py
+++ b/website/addons/badges/model/badges.py
@@ -161,7 +161,7 @@ class BadgeAssertion(StoredObject):
     @property
     def recipient(self):
         return {
-            'idenity': self.node._id,
+            'idenity': self.node._id,  # TODO: An unknown amount of code may depend on this typo
             'type': 'osfnode',  # TODO Could be an email?
             'hashed': False
         }
--- a/website/addons/box/model.py
+++ b/website/addons/box/model.py
@@ -81,9 +81,7 @@ class BoxNodeSettings(StorageAddonBase, AddonOAuthNodeSettingsBase):
     oauth_provider = Box
     serializer = BoxSerializer
 
-    foreign_user_settings = fields.ForeignField(
-        'boxusersettings', backref='authorized'
-    )
+    foreign_user_settings = fields.ForeignField('boxusersettings')
     folder_id = fields.StringField(default=None)
     folder_name = fields.StringField()
     folder_path = fields.StringField()
--- a/website/addons/box/model.py
+++ b/website/addons/box/model.py
@@ -81,7 +81,6 @@ class BoxNodeSettings(StorageAddonBase, AddonOAuthNodeSettingsBase):
     oauth_provider = Box
     serializer = BoxSerializer
 
-    foreign_user_settings = fields.ForeignField('boxusersettings')
     folder_id = fields.StringField(default=None)
     folder_name = fields.StringField()
     folder_path = fields.StringField()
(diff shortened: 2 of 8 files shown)


website/addons/badges/model/badges.py
```diff
@@ -161,7 +161,7 @@ def verify(self, vtype='hosted'):
     @property
     def recipient(self):
         return {
-            'idenity': self.node._id,
+            'idenity': self.node._id,  # TODO: An unknown amount of code may depend on this typo
```
mfraezz: A quick grep indicates that no code depends on this typo, but still :neutral_face:

website/addons/box/model.py
```diff
@@ -81,9 +81,7 @@ class BoxNodeSettings(StorageAddonBase, AddonOAuthNodeSettingsBase):
     oauth_provider = Box
     serializer = BoxSerializer
 
-    foreign_user_settings = fields.ForeignField(
-        'boxusersettings', backref='authorized'
-    )
+    foreign_user_settings = fields.ForeignField('boxusersettings')
```
mfraezz: Actually, please remove this field entirely. It was last relevant during a migration that changed how `BoxNodeSettings` and `BoxUserSettings` objects are associated with each other. There was supposed to be a clean-up PR for this and other addons, but that may have been lost in a flood of other priorities.

The same can be done for the backreffed field on the `AddonDataverseNodeSettings` model as well.

website/addons/badges/model/badges.py
abought: Thanks; that was a quick note before I realized badges were irrelevant anyway. (this addon is turned off) :ghost: 

When it involves potentially outdated production data, I'm never quite sure whether a field is present at some level. :)

website/addons/box/model.py
abought: Fields deleted in 70bc6aa148dde232b5cf6915910f15fced3a25b0

mfraezz:
LGTM. RTM as soon as the build passes :fire:

brianjgeiger:
Looks like you did the same thing with Dataverse that you did with Box. Should QA test that equally?

abought:
According to my notes, some dataverse work is also being done in this open PR: https://github.com/CenterForOpenScience/osf.io/pull/4979

I'll defer to @mfraezz to answer whether the overlap in tickets has any implications for QA.

mfraezz:
Realistically, neither Box nor Dataverse should require QA due to this PR. The one removed field was used exclusively for a migration that (among other things) removed the dependency that the model had for that field. At most, a check to make sure a user's [addon] account can be imported to a node after being connected via the user settings page should suffice.

If you're feeling adventurous, @brianjgeiger, [#4979](https://github.com/CenterForOpenScience/osf.io/pull/4979) has been RTM for 42 days

PR #5293, opened 2016-03-24
[hold] Hide "withdrawn" registrations

mfraezz:
https://trello.com/c/dIxjFwIN/134-retracted-registrations-should-not-appear-on-my-projects-page
`.retracted` -> `.withdrawn`


sloria:
The "withdrawn" language is only added in develop, not the 0.66 release. I'll cherry-pick this into develop.

sloria:
Cancel that; the dashboard isn't yet on `develop`. Cherry-picked into `release/0.66.0-develop`. This change will need to be remade after 0.66.0 is released.

PR #5295, opened 2016-03-24
Improve application of text filters

abought:
Related to, and possible alternative to, #5292 

## Purpose

Improve application of the `strip_ko` filter. We can reasonably apply this to all templates, because it is separate from the markupsafe flag.

## Developer / Deployment notes

When testing the fix, try these steps to ensure that your mako cache is cleared and templates reflect the newest filters. Because this is a global  without help.

1. Pull code
2. Stop any running OSF server
3. Clear contents of `/tmp/mako_modules` directory
4. Restart server

## QA testing notes

I can walk you through in person.


abought:
Replaced with #5296 , to master.

PR #5301, opened 2016-03-25
Fix collaborator icons on wiki page

asanfilippo7:
From [https://openscience.atlassian.net/browse/OSF-4264](url)

- Create new fields on the WikiEditor ViewModel to display up to 18 active users, and a message saying "and X more"

- Update edit.mako template to display up to 18 icons and the "more" message

- Update wiki-page.css to change the height of the Ace Wiki-Editor. (Shortening to 400px prevents the editor from being pushed off screen and doesn't affect UI).

Additionally, from [https://openscience.atlassian.net/browse/OSF-5722](url)

- Small logic change in contribAdder.js addAllVisible method. This makes it so that the addAll button does not appear before searching or if a search returns no results, but does appear if a search returns results


website/addons/wiki/static/WikiEditor.js
```diff
@@ -42,18 +42,35 @@ ko.bindingHandlers.ace = {
...
+        if (self.activeUsers().length > 18) {
+            return self.activeUsers().slice(0,18);
+        }
+        return self.activeUsers();
+    });
+    
+    // Show text that says "and # more" collaborators
+    self.andOthersMessage = ko.computed(function() {
+        if (self.activeUsers().length > 18) {
+            var leftovers = self.activeUsers().slice(18, self.activeUsers().length);
```
mfraezz: Why not just subtract `18` from `self.activeUsers().length`? Faster than making a copy of the last `n-18` elements.

website/addons/wiki/static/WikiEditor.js
```diff
@@ -42,18 +42,35 @@ ko.bindingHandlers.ace = {
...
+    // Display the icons of up to the first 18 collaborators on a page
+    self.showCollaborators = ko.computed(function() {
+        if (self.activeUsers().length > 18) {
+            return self.activeUsers().slice(0,18);
+        }
+        return self.activeUsers();
+    });
+    
+    // Show text that says "and # more" collaborators
+    self.andOthersMessage = ko.computed(function() {
```
mfraezz: This should probably be a [`pureComputed`](http://knockoutjs.com/documentation/computed-pure.html)

website/addons/wiki/static/WikiEditor.js
```diff
@@ -42,18 +42,35 @@ ko.bindingHandlers.ace = {
...
     self.currentText = viewText; //from wikiPage's VM
     self.activeUsers = ko.observableArray([]);
+    
     self.status = ko.observable('connecting');
     self.throttledStatus = ko.observable(self.status());
     self.autocom = ko.observable(false);
 
-    self.displayCollaborators = ko.computed(function() {
-       return self.activeUsers().length > 1;
+   self.displayCollaborators = ko.computed(function() {
```
mfraezz: This should be indented by one space

website/addons/wiki/static/WikiEditor.js
```diff
@@ -42,18 +42,35 @@ ko.bindingHandlers.ace = {
...
+        }
+        return self.activeUsers();
+    });
+    
+    // Show text that says "and # more" collaborators
+    self.andOthersMessage = ko.computed(function() {
+        if (self.activeUsers().length > 18) {
+            var leftovers = self.activeUsers().slice(18, self.activeUsers().length);
+            return '...and ' + leftovers.length + ' more';
+        }
```
mfraezz: Some browsers might get upset if you don't return an empty string in the `self.activeUsers().length <= 18` case.

mfraezz:
The build is failing because `addAllVisible` is now returning what it's supposed to, but the tests were apparently not written to handle that. The failing test in `website/static/js/tests/contributors.test.js` needs to provide an `id` to `vm.results` that is neither a contributor nor selected for the `Add All` button to show up.

Also, please clean up the extraneous whitespace -- there shouldn't be spaces on otherwise blank lines. Pass complete :cat:

mfraezz:
I'm sending this back to "Additional Development", but only for [blank](https://github.com/CenterForOpenScience/osf.io/pull/5301/files#diff-9980126467484d9aac72552bded1999bR189) [lines](https://github.com/CenterForOpenScience/osf.io/pull/5301/files#diff-914211b5fa509220396a91ff07113de3R45) [with](https://github.com/CenterForOpenScience/osf.io/pull/5301/files#diff-914211b5fa509220396a91ff07113de3R56) [spaces](https://github.com/CenterForOpenScience/osf.io/pull/5301/files#diff-914211b5fa509220396a91ff07113de3R64) [like](https://github.com/CenterForOpenScience/osf.io/pull/5301/files#diff-914211b5fa509220396a91ff07113de3R73) [these](https://github.com/CenterForOpenScience/osf.io/pull/5301/files#diff-58e49a58eaa869974d65487796f91b4bR163). Otherwise looks good, and I'll drop it in RTM when that's resolved. :guitar: 

(Sorry if it seems like I'm picking nits; normally Steve fixes whitespace things on merge but he's away and I don't have that power.)

asanfilippo7:
No worries! I think I got all the whitespaces out? I'm still not entirely sure when a space is/isn't extraneous...so if it still looks wrong let me know and I'll do it again. Sorry for the trouble!

mfraezz:
General rule about whitespace: blank lines for readability are fine, but there should only be additional characters on those lines if it's a line of code or a comment -- there shouldn't be tabs or spaces on an otherwise blank line (unless you're using a certain [esoteric programming language](https://en.wikipedia.org/wiki/Whitespace_(programming_language))).

I tried to include example code here, but GH automatically strips tabs and spaces from blank lines within \`\`\` tags.

PR #5303, opened 2016-03-25
[RFR] Fix transaction rollback error overriding original exception.

cwisecarver:
Logging and returning correct error in case of transaction rollback failure

### DIFF
--- a/framework/transactions/handlers.py
+++ b/framework/transactions/handlers.py
@@ -36,10 +36,12 @@ def transaction_before_request():
         return None
     try:
         commands.rollback()
-        logger.error('Transaction already in progress; rolling back.')
+        logger.error('Transaction already in progress prior to request; rolling back.')
     except OperationFailure as error:
+        #  expected error, transaction shouldn't be started prior to request
         message = utils.get_error_message(error)
         if messages.NO_TRANSACTION_ERROR not in message:
+            #  exception not a transaction error, reraise
             raise
     commands.begin()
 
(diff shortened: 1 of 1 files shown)


framework/transactions/handlers.py
```diff
@@ -52,7 +53,14 @@ def transaction_after_request(response, base_status_code_error=500):
     if view_has_annotation(NO_AUTO_TRANSACTION_ATTR):
         return response
     if response.status_code >= base_status_code_error:
-        commands.rollback()
+        original_exception = sys.exc_info()
+        try:
+            commands.rollback()
+        except OperationFailure:
+            logger.error('Transaction Rollback failed, original exception was: ', exc_info=original_exception)
```
icereval: Can this be made to only display the error for the transaction rollback if it truly errors? Checking in the after_request/teardown_request depending on order of call to determine if was have not already done the rollback, even if we do decide to attempt it once again we in this case we don't need to report the error and can ignore it.

icereval:
first pass :pizza:

framework/transactions/handlers.py
icereval: You'll want to verify, but calls are made in the following order so we'll need to make the framework abstractions idempotent. A solution might also be to determine if Django has called process_exception from process_response to avoid calling them out of order from flask...

Django
https://github.com/CenterForOpenScience/osf.io/blob/develop/api/base/middleware.py#L54
1. process_exception --> transaction_teardown_request
2. process_response --> transaction_after_request

Flask
https://github.com/CenterForOpenScience/osf.io/blob/develop/framework/transactions/handlers.py#L84
1. after_request --> transaction_after_request
2. teardown_request --> transaction_teardown_request

cwisecarver: I think this explains the ordering of django: https://docs.djangoproject.com/en/1.9/topics/http/middleware/#hooks-and-application-order

cwisecarver: We could just set a property on the response object in `process_exception`, if one isn't already being set, and look at it in `process_response`.

cwisecarver:
@icereval I'm fairly certain this would pass if @chrisseto's javascript wasn't breaking the build right now.

PR #5306, opened 2016-03-25
[feature] Allow Read + Write Contributors to Add Components with Inherited Contributors

caseyrollins:
##### Purpose 
- Currently, when a r+w contributor creates a new component within a project, they are not given the option to 'add contributors from the parent project'. However, a r+w contributor will become an admin on a component they create, and therefore would be able to later add contributors anyways. 

##### Changes
- Add 'add contributors' checkbox in add component modal for a r+w contributor.
- Add tests for r+w contributors to add components with inherited contributors.
- If a r+w contributor creates a new component with inherited contributors, their permission will be set to admin on the new component, instead of being inherited as r+w. 

##### JIRA 
- [OSF-5887](https://openscience.atlassian.net/browse/OSF-5887)

### DIFF
--- a/website/project/views/node.py
+++ b/website/project/views/node.py
@@ -32,7 +32,7 @@ from website.project.decorators import (
     http_error_if_disk_saving_mode
 )
 from website.tokens import process_token_or_pass
-from website.util.permissions import ADMIN, READ, WRITE
+from website.util.permissions import ADMIN, READ, WRITE, CREATOR_PERMISSIONS
 from website.util.rubeus import collect_addon_js
 from website.project.model import has_anonymous_link, get_pointer_parent, NodeUpdateError, validate_title
 from website.project.forms import NewNodeForm
@@ -207,11 +207,10 @@ def project_new_node(auth, node, **kwargs):
             'Your component was created successfully. You can keep working on the project page below, '
             'or go to the new <u><a href={component_url}>component</a></u>.'
         ).format(component_url=new_component.url)
-        if form.inherit_contributors.data and (node.has_permission(user, ADMIN) or node.has_permission(user, WRITE)):
+        if form.inherit_contributors.data and node.has_permission(user, WRITE):
             for contributor in node.contributors:
-                new_component.add_contributor(contributor, permissions=node.get_permissions(contributor), auth=auth)
-                if contributor is user and not node.has_permission(contributor, 'admin'):
-                    new_component.set_permissions(contributor, ['read', 'write', 'admin'])
+                perm = CREATOR_PERMISSIONS if contributor is user else node.get_permissions(contributor)
+                new_component.add_contributor(contributor, permissions=perm, auth=auth)
 
             new_component.save()
             redirect_url = new_component.url + 'contributors/'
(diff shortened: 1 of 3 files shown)


brianjgeiger:
Just to be sure, "If a r+w contributor creates a new component with inherited contributors, **their** permission will be set to admin" means "If a r+w contributor creates a new component with inherited contributors, **the component creator's** permission will be set to admin", yes?

caseyrollins:
@brianjgeiger Correct!

website/project/views/node.py
```diff
@@ -207,9 +207,12 @@ def project_new_node(auth, node, **kwargs):
             'Your component was created successfully. You can keep working on the project page below, '
             'or go to the new <u><a href={component_url}>component</a></u>.'
         ).format(component_url=new_component.url)
-        if form.inherit_contributors.data and node.has_permission(user, ADMIN):
+        if form.inherit_contributors.data and (node.has_permission(user, ADMIN) or node.has_permission(user, WRITE)):
             for contributor in node.contributors:
                 new_component.add_contributor(contributor, permissions=node.get_permissions(contributor), auth=auth)
+                if contributor is user and not node.has_permission(contributor, 'admin'):
```
brianjgeiger: This doesn't hurt anything, but if you left out `node.has_permission(contributor, 'admin')`, the functionality wouldn't change, would it? It would just be a redundant permissions set?

brianjgeiger: I'm going to be super picky here primarily for readability. What if we re-do this for loop to set a permissions variable to the current permissions at the top, and if the contributor is the user, set the permissions to CREATOR_PERMISSIONS (which you'll have to import alongside WRITE and ADMIN).  Then add the contributor to the component with permissions=permissions. This will keep from doing multiple permissions sets and should be a little more clear to people coming in what is going on.

website/project/views/node.py
```diff
@@ -207,9 +207,12 @@ def project_new_node(auth, node, **kwargs):
...
+                    new_component.set_permissions(contributor, ['read', 'write', 'admin'])
```
brianjgeiger: Regardless of the above, `CREATOR_PERMISSIONS` would be better than the list here, in case something changes.

brianjgeiger:
This looks good. Refactor that one for loop and I'll give it another pass.

website/project/views/node.py
```diff
@@ -207,9 +207,11 @@ def project_new_node(auth, node, **kwargs):
             'Your component was created successfully. You can keep working on the project page below, '
             'or go to the new <u><a href={component_url}>component</a></u>.'
         ).format(component_url=new_component.url)
-        if form.inherit_contributors.data and node.has_permission(user, ADMIN):
+        if form.inherit_contributors.data and node.has_permission(user, WRITE):
             for contributor in node.contributors:
-                new_component.add_contributor(contributor, permissions=node.get_permissions(contributor), auth=auth)
+                perm = CREATOR_PERMISSIONS if contributor is user else node.get_permissions(contributor)
```
brianjgeiger: Nice.

brianjgeiger:
Looks great; just waiting for the tests to run. :tada: :penguin: :tada:

PR #5322, opened 2016-03-29
[Feature Fix] Fix and QA suggestions

Ghalko:
## Proposal
- Returns a web-page for re-enabling a user
- Allows child nodes to be deleted from parent node
- Indicates the status of child node with action button
- Remove registration deletion
- Indicates registration pending

## Repercussions
A bit more robust.


mfraezz:
LGTM. :shipit:

PR #5324, opened 2016-03-29
0.66 Fix my projects issue with adding new projects not updating view

caneruguz:
This solution makes another call to server to get the new item with the embedded users.

### DIFF
--- a/website/static/js/myProjects.js
+++ b/website/static/js/myProjects.js
@@ -874,7 +874,6 @@ var MyProjects = {
                     title: 'Create new project',
                     categoryList: ctrl.categoryList,
                     stayCallback: function () {
-                        console.log(this.saveResult());
                         var savedItem = this.saveResult().data;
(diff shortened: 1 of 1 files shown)


website/static/js/myProjects.js
```diff
@@ -875,8 +875,20 @@ var MyProjects = {
                     title: 'Create new project',
                     categoryList: ctrl.categoryList,
                     stayCallback: function () {
-                        ctrl.allProjectsLoaded(false);
-                        ctrl.updateList(ctrl.breadcrumbs()[ctrl.breadcrumbs().length - 1]);
+                        var savedItem = this.saveResult().data;
+                        var url = $osf.apiV2Url('nodes/' + savedItem.id + '/', { query : { 'related_counts' : 'children', 'embed' : 'contributors' }});
+                        m.request({method : 'GET', url : url, config : xhrconfig, background: true}).then(function(r){
+                            var item = r.data;
+                            ctrl.nodes.projects.treeData.data.unshift(item);
```
chrisseto: `ctrl.nodes` has been removed

PR #5325, opened 2016-03-29
[feature] Wiki comments

samanehsan:
Fixes JIRA ticket OSF-5240 -- Adds ability to comment on wiki pages via the v2 API. This PR also includes the front-end changes to enable wiki comments on the OSF to facilitate testing & review.

Also moves some API logic out of the comment views & into the comment target models by adding a `Commentable` mixin with common properties & methods that a comment target must have. This is to avoid having to update the API whenever a new comment target is added.

<h3>Changes:</h3>
Wiki comments:
- API: List, create, view, edit, delete, undelete, report, unreport
- API Tests
- Fix embedding comment targets
- Logging
- Notification emails
- Add comment pane to wiki pages
- Display number of unread comments on wiki pages

Refactor:
- Added `Commentable` mixin
- Removed checking whether a file with comments exists

Notes:
Because there are no wiki API endpoints, I updated the API to handle embedding items in the response data that do not have an API view (so that embedding comment targets do not throw an error when the target is a wiki page) and added the number of unread comments to the response links metadata when the request includes `related_counts=True&filter[target]=target_id.`


website/project/model.py
```diff
@@ -146,13 +145,31 @@ class MetaData(GuidStoredObject):
...
+
+    @property
+    def root_target_page(self):
+        raise NotImplementedError
+
+    @property
+    def is_deleted(self):
+        raise NotImplementedError
+
+    def belongs_to_node(self, node):
```
mfraezz: `node` arg should be `node_id` --  all the overrides use the latter.

mfraezz: Could also consider adding default behavior here, as the same behavior is given in three of four overrides.

mfraezz:
Bug found: nested comments don't load. The API throws a `500`, and the browser console displays a CORS issue (likely a red herring).

`[30/Mar/2016 16:45:41]"GET /v2/nodes/zera5/comments/?related_counts=True&embed=user&filter[target]=xhksdby3tv6p&_=1459370724149 HTTP/1.1" 500 93764`

screen shot 2016-03-30 at 16 49 34

website/project/model.py
samanehsan: I discussed adding a default `belongs_to_node` with Steve -- he suggested that those methods be explicitly defined in the models even if there is some repetition.

mfraezz:
Changes look good; above bug resolved and no additional bugs found during testing. Dropping this in RTM :money_with_wings:

api/base/pagination.py
```diff
@@ -89,6 +93,21 @@ def get_paginated_response(self, data):
 
         response_dict = self.get_response_dict(data, reversed_url)
 
+        if view_name == 'nodes:node-comments' and self.request.query_params.get('related_counts', False):
```
sloria: This special-casing of one view looks suspicious. Could you use a subclass instead?

```python
class CommentPagination(JSONAPIPagination):
    def get_paginated_response(...):
        #...

class NodeCommentsList(...):
    pagination_class = CommentPagination
   #...
```

api/comments/serializers.py
```diff
@@ -131,25 +126,12 @@ def get_validated_target_type(self, obj):
 
     def get_target(self, node_id, target_id):
         target = Guid.load(target_id)
-        if not target:
+        if not target or not getattr(target.referent, 'belongs_to_node'):
```
sloria: `getattr` will raise an `AttributeError` unless you pass a default as the 3rd argument.

website/project/model.py
```diff
@@ -148,13 +147,31 @@ class MetaData(GuidStoredObject):
     date_created = fields.DateTimeField(auto_now_add=datetime.datetime.utcnow)
     date_modified = fields.DateTimeField(auto_now=datetime.datetime.utcnow)
 
+class Commentable(object):
```
sloria: We're trying to prevent `website/project/model.py` from growing much bigger than it already is, so how about moving this into a separate module, e.g. `website.project.commentable`.

Eventually we'll want to move the Comment model outside of this module as well, e.g. `website.comments.model.Comment` (and then we could move `Commentable` to `website.comments.commentable.Commentable`. But that can wait; you've probably had enough refactoring for this PR =)

sloria: Also, this class could use a brief docstring so that devs know when to use it. Something simple like `"""Abstract class that defines the interface for models that can have comments attached to them."""`

website/project/model.py
```diff
@@ -148,13 +147,31 @@ class MetaData(GuidStoredObject):
...
 
-class Comment(GuidStoredObject, SpamMixin):
+    @property
+    def target_type(self):
+        raise NotImplementedError
+
+    @property
+    def root_target_page(self):
```
sloria: Docstrings for these properties would also be helpful. It isn't obvious to me what `root_target_page` is supposed to return.

website/project/model.py
```diff
@@ -248,13 +288,16 @@ def create(cls, auth, **kwargs):
...
-        elif isinstance(comment.root_target.referent, StoredFileNode):
-            log_dict['file'] = {'name': comment.root_target.referent.name, 'url': comment.get_comment_page_url()}
-            comment.page = Comment.FILES
-        else:
+        page = getattr(comment.root_target.referent, 'root_target_page', None)
+        if not page:
             raise ValueError('Invalid root target.')
+        comment.page = page
+
+        if comment.page == Comment.FILES:
```
sloria: Rather than these conditionals, could you expose a hook on `Commentable`?

```python
class StoredFileNode(..., Commentable):
    def create_log_metadata(self, comment):
        """Return extra data to pass as ``params`` to ``Node.add_log`` when a new comment is created."""
        return {'file': {'name': self.name, 'url': comment.get_comment_page_url()}}
```

sloria: Ah, I see that the same params are added for all types of comment-related logs (create, edit, etc.). Therefore, better to use a more generic name for the hook, .e.g `get_extra_log_params(self, comment)`.

website/static/js/comment.js
```diff
@@ -642,8 +626,8 @@ CommentListModel.prototype.initListeners = function() {
     });
 };
 
-var onOpen = function(page, rootId, nodeApiUrl) {
-    if (osfHelpers.urlParams().view_only){
+var onOpen = function(page, rootId, nodeApiUrl, currentUserId) {
+    if (osfHelpers.urlParams().view_only ||currentUserId === '') {
```
sloria: It's probably safest to check the truthiness of `currentUserID`.

```js
 if (osfHelpers.urlParams().view_only || !currentUserID'') 
```

website/static/js/pages/wiki-edit-page.js
```diff
@@ -167,3 +168,25 @@ $(document).ready(function () {
     // Tooltip
     $('[data-toggle="tooltip"]').tooltip();
 });
+
+var $comments = $('.comments');
+if ($comments.length && window.contextVars.wiki.wikiID !== 'None') {
+    var currentUser = {
```
sloria: Do you need to reserialize `currentUser`? Is there a reason not to pass `window.contextVars.currentUser` to `Comment.init`?

website/templates/include/comment_pane_template.mako
```diff
@@ -15,9 +15,12 @@
                 <i class="fa fa-times"></i>
             </button>
             <h4>
-                <span data-bind="if: page() == 'node' ">${node['title']} Discussion</span>
                 %if file_name:
                     <span data-bind="if: page() == 'files'">Files | ${file_name} Discussion</span>
+                %elif wiki_name:
+                    <span data-bind="if: page() == 'wiki'">Wiki | ${wiki_name} Discussion</span>
```
sloria: This is fine for now, but we'll probably want to get rid of the Mako in favor of Knockout for rendering `file_name`, `wiki_name`, etc.

sloria: Also, do you need the mako conditionals when there are already KO `ifs`?

```js
<span data-bind="if: page() == 'wiki' && wikiName()">Wiki | ${wiki_name} Discussion</span>
```

website/static/js/comment.js
```diff
@@ -122,22 +123,26 @@ BaseComment.prototype.setupToolTips = function(elm) {
 
 BaseComment.prototype.fetch = function() {
     var self = this;
+    var setUnread = getTargetType(self) !== 'comments' && !osfHelpers.urlParams().view_only && self.author.id !== '';
```
sloria: Since `getTargetType` relies on `this`/`self`, it should be a method rather than a function.

sloria:
Minor: I'm not sure we should hide the comment panel for pages that don't have content. Users may want to discuss before they write any content. That said, this is a nitpick; I don't think we need to change the implemented behavior.

sloria:
Pass finished.

website/static/js/pages/wiki-edit-page.js
samanehsan: Good catch! I'll need to change the currentUser info I construct from API JSON data as well since the url is under the `urls.profiles` key.

website/templates/include/comment_pane_template.mako
samanehsan: When I remove the mako conditionals, mako throws an error because `file_page` is undefined (unless you're viewing the file detail page)

sloria: Ah ok, so we'd have to render file_name, etc. with KO, too (like in my code snippet above).

PR #5326, opened 2016-03-29
[Fix] Only filter retractions and pending registrations on the 'All my registrations' page

mfraezz:
https://trello.com/c/qSscVKea/136-pending-and-retracted-registrations-should-not-show-on-my-projects-page


brianjgeiger:
Looks good, and the demo worked as expected. :tada: :penguin: :tada:

PR #5329, opened 2016-03-30
[RFR][Feature] OSF-5883 MFR render failures. Show a better error message.

TomBaxter:
## Purpose:
[OSF-5883](https://openscience.atlassian.net/browse/OSF-5883)
When MFR render fails. Show a better error message.

## Changes:
update website/addons/base/views.py update for more palatable error messages

update website/files/models/base.py add "def get_guid" to "class TrashedFileNode" to allow for better message in website/addons/base/views.py

## Side effects:
None

#5883


website/addons/base/views.py
```diff
@@ -472,6 +494,16 @@ def addon_deleted_file(auth, node, **kwargs):
...
     # Allow file_node to be passed in so other views can delegate to this one
     trashed = kwargs.get('file_node') or TrashedFileNode.load(kwargs.get('trashed_id'))
+    if kwargs.get('path'):
+        file_path = kwargs.get('path')
+    else:
+        file_path = trashed.path
+    if trashed.name:
+        file_name = trashed.name
+    else:
+        file_name = os.path.basename(file_path)
```
erinspace: These if else statements can be made a little less verbose with some ```.get()``` and some ```getattr()``` use with defaults! Something like:

```python
file_path = kwargs.get('path', trashed.path)
file_name = getattr(trashed, 'name', os.path.basename(file_path))
```

website/addons/base/views.py
```diff
@@ -480,6 +512,7 @@ def addon_deleted_file(auth, node, **kwargs):
 
     ret = serialize_node(node, auth, primary=True)
     ret.update(rubeus.collect_addon_assets(node))
+    retError = ERROR_MESSAGES[kwargs.get('error_type')].format(file_name=file_name, provider=trashed.provider)
```
erinspace: Running this locally and trying to access a deleted file throws a ```KeyError``` for me - looks like ```error_type``` isn't showing up in ```kwargs``` so is defaulting to ```None```

website/addons/base/views.py
```diff
@@ -548,19 +585,21 @@ def addon_view_or_download_file(auth, path, provider, **kwargs):
 
     if version is None:
         if file_node.get_guid():
-            # If this file has been successfully view before but no longer exists
-
-            # Move file to trashed file node
+            # If this file has been successfully view before, or OSF, but no
```
erinspace: What does "or OSF" mean? Also if you're updating this already you change "view" to "viewed"?

website/files/models/base.py
```diff
@@ -91,6 +91,19 @@ def restore(self, recursive=True, parent=None):
...
 
+    def get_guid(self, create=False):
+        """Attempt to find a Guid that points to this object.
+        One will be created if requested.
+        :rtype: Guid
+        """
+        try:
+            # Note sometimes multiple GUIDs can exist for
+            # a single object. Just go with the first one
+            return Guid.find(Q('referent', 'eq', self))[0]
```
erinspace: you can use ```find_one``` instead of indexing into this queryset and then except ```NoResultsFound```

website/files/models/base.py
```diff
@@ -91,6 +91,19 @@ def restore(self, recursive=True, parent=None):
...
+        except IndexError:
+            if not create:
+                return None
+        return Guid.generate(self)
```
erinspace: Unless I'm mistaken, this will generate a guid regardless of the ```create``` argument, right? Should this be in the ```except``` block?

website/files/models/base.py
```diff
@@ -573,10 +586,10 @@ def touch(self, auth_header, revision=None, **kwargs):
             self.generate_waterbutler_url(revision=revision, meta=True, **kwargs),
             headers=headers,
         )
+        # Returns None to addon_view_or_download_file 'version'
```
erinspace: Having a bit of trouble understanding this comment - seems like the ```File.touch``` method is used elsewhere than to just return to the specific view -- I could be wrong though!

erinspace:
Hello @TomBaxter ! Pass one finished!

TomBaxter:
Thanks @erinspace , working on getting my dev environment in order, then will fix up per your comments.

website/addons/base/views.py
TomBaxter: Works great for file_path. Unfortunately, trashed.name is set to None, so doesn't hit default.

website/addons/base/views.py
TomBaxter: I've made the comment more succinct. What I was trying to get at is that OSF storage files get GUIDs whether they are viewed or not.

website/files/models/base.py
TomBaxter: Having trouble with this.
I switched...         return Guid.find(Q('referent', 'eq', self))[0]
to...                     return Guid.find_one(Q('referent', 'eq', self))
but that returns "None"
In the examples I looked at the object of find_one were either "cls" or "FileNode.resolve_class(self.provider, kind)"
Do I need to use the resolve_class method? If so it looks like I would need to introduce a numerical "kind" to the code.

website/files/models/base.py
TomBaxter: I think by default, if "create=True" isn't passed to the method that this will return "None" if the guid isn't found.

  Thank you for this comment as I just copied this method from "class StoredFileNode" and didn't think about all the possibilities. I assume that, if a TrashedFileNode doesn't have a guid, under no circumstances would we want to create one.

website/files/models/base.py
TomBaxter: Removed. That was just a note to myself while I was trying to figure out the code. I missed it when I was cleaning up.

TomBaxter:
@erinspace  New code pushed.

website/addons/base/views.py
TomBaxter: Fixed by adding default error_type.

website/files/models/base.py
TomBaxter: After our discussion I was able to make the change.

website/addons/base/views.py
```diff
@@ -516,21 +550,21 @@ def addon_view_or_download_file(auth, path, provider, **kwargs):
         raise HTTPError(httplib.BAD_REQUEST)
 
     if not isinstance(node_addon, StorageAddonBase):
-        raise HTTPError(httplib.BAD_REQUEST, {
+        raise HTTPError(httplib.BAD_REQUEST, data={
             'message_short': 'Bad Request',
-            'message_long': 'The add-on containing this file is no longer connected to the {}.'.format(node.project_or_component)
+            'message_long': 'The add-on {} containing {} is no longer connected to {}.'.format(provider, path, node.project_or_component)
```
felliott: Minor copyedit: for this and the others below, can you change the wording to "The {} add-on containing"?  I think "The github add-on containing foo.txt" sounds better than "The add-on github containing foo.txt".

felliott:
Right now both error messages state that the provider is unresponsive and suggest verifying the deletion through the provider's website.  This doesn't make sense for files that were deleted via the OSF, since they should have storedfilenodes or trashedfilenodes that can confirm that they were deleted.  In that case the error message should be something like "This file has been deleted".  We'll fill in the metadata for it later.  All deleted osfstorage files should show this message.  The other message is for when we don't have any record of the file and can't speak to what happened to it.  Perhaps rename `FILE_GONE` to `FILE_DELETED`, to show that it's an affirmative status, not a placeholder for the unknown.

Similarly, I think the default for `error_type` in `addon_deleted_file` should be `FILE_NOT_FOUND`.  If we are able to find a TrashedFileNode with the given `trashed_id`, then update it to `FILE_DELETED`.

website/addons/base/views.py
```diff
@@ -40,15 +40,52 @@
...
 <style>
 .file-download{{display: none;}}
 .file-share{{display: none;}}
 .file-delete{{display: none;}}
 </style>
 <div class="alert alert-info" role="alert">
-This link to the file "{file_name}" is no longer valid.
-</div>'''
+<p>
+This {provider} link to the file "{file_name}" was delete from the OSF.
```
felliott: s/delete/deleted/.

website/addons/base/views.py
```diff
@@ -40,15 +40,52 @@
...
+</p>
+<p>
+
+It was delete by User: {deleted_by} On: {deleted_on}
```
felliott: "It was delete**d** by <a href="/{deleted_by}">{deleted_by}</a> on {deleted_on}}"

website/addons/base/views.py
```diff
@@ -466,20 +503,42 @@ def addon_view_or_download_file_legacy(**kwargs):
...
     """
     # Allow file_node to be passed in so other views can delegate to this one
-    trashed = kwargs.get('file_node') or TrashedFileNode.load(kwargs.get('trashed_id'))
-    if not trashed:
-        raise HTTPError(httplib.NOT_FOUND, {
-            'message_short': 'Not Found',
-            'message_long': 'This file does not exist'
-        })
+    file_node = kwargs.get('file_node') or TrashedFileNode.load(kwargs.get('trashed_id'))
+    if isinstance(file_node, TrashedFileNode):
```
felliott: You can simplify this conditional block by initializing `deleted_by` and `deleted_on` to None before the block and only setting them if isa TrashFileNode && tfn.deleted_by is true.

website/addons/base/views.py
```diff
@@ -548,19 +611,18 @@ def addon_view_or_download_file(auth, path, provider, **kwargs):
...
-            # Show a nice error message
-            return addon_deleted_file(file_node=file_node, **kwargs)
-
-        raise HTTPError(httplib.NOT_FOUND, {
-            'message_short': 'Not Found',
-            'message_long': 'This file does not exist'
-        })
+            # Trash file if not already found in trash.
+            file_node = TrashedFileNode.load(file_node._id) or file_node.delete()
+            # This GUID's file is no longer accessable, trash it.
```
felliott: Please remove the commented-out code.

website/addons/base/views.py
TomBaxter: Done.

website/addons/base/views.py
TomBaxter: Done

website/addons/base/views.py
TomBaxter: Done

website/addons/base/views.py
TomBaxter: Done.

website/addons/base/views.py
```diff
@@ -40,15 +40,52 @@
 # import so that associated listener is instantiated and gets emails
 from website.notifications.events.files import FileEvent  # noqa
 
-FILE_GONE_ERROR_MESSAGE = u'''
+# TODO - make distiction between filenodes that had GUIDs and were trashed
```
felliott: Does this TODO still apply?  If not, please delete it.

website/addons/base/views.py
```diff
@@ -40,15 +40,52 @@
...
 <style>
 .file-download{{display: none;}}
 .file-share{{display: none;}}
 .file-delete{{display: none;}}
 </style>
 <div class="alert alert-info" role="alert">
-This link to the file "{file_name}" is no longer valid.
-</div>'''
+<p>
+This {provider} link to the file "{file_name}" was deleted from the OSF.
```
felliott: Copy edit: `The file "{file_name}" stored on {provider} was deleted via the OSF.`  Also do we have access to the non-shortname of the storage provider here?  i.e. can we display "Google Drive" instead of "googledrive"?  I think it's available via the addon models.

website/addons/base/views.py
```diff
@@ -40,15 +40,52 @@
...
+</p>
+<p>
+
+It was delete by <a href="/{deleted_by}">{deleted_by}</a> on {deleted_on}
```
felliott: 1.) s/delete/deleted/   2.) Needs a period at the end.  3.) Can you format `deleted_on`?  Drop the millisecond component and display the timezone (UTC is fine for now).

website/addons/base/views.py
```diff
@@ -40,15 +40,52 @@
 # import so that associated listener is instantiated and gets emails
 from website.notifications.events.files import FileEvent  # noqa
 
-FILE_GONE_ERROR_MESSAGE = u'''
+# TODO - make distiction between filenodes that had GUIDs and were trashed
+# through the OSF/WB and those that were trashed due to not being found at the
+# provider. Currently both cases go through 'FILE_GONE_HAD_GUID'.
+ERROR_MESSAGES = {'FILE_GONE': u'''
 <style>
 .file-download{{display: none;}}
```
felliott: What are `.file-download`, `.file-share`, and `.file-delete` used for?  If they're obsolete, please delete them.

website/addons/base/views.py
```diff
@@ -466,20 +503,41 @@ def addon_view_or_download_file_legacy(**kwargs):
...
-    if not trashed:
-        raise HTTPError(httplib.NOT_FOUND, {
-            'message_short': 'Not Found',
-            'message_long': 'This file does not exist'
-        })
+    file_node = kwargs.get('file_node') or TrashedFileNode.load(kwargs.get('trashed_id'))
+    deleted_by, deleted_on = None, None
+    if isinstance(file_node, TrashedFileNode):
+        # TODO test for None once hotfix is in
+        print('attrs are ' + str(dir(file_node)))
```
felliott: 1.) Does this comment still apply?  2.) Please remove the debugging code.

website/addons/base/views.py
```diff
@@ -466,20 +503,41 @@ def addon_view_or_download_file_legacy(**kwargs):
...
+        deleted_by = str(file_node.deleted_by)
+        deleted_on = str(file_node.deleted_on)
+        if deleted_by == "None":
+            error_type = 'BLAME_PROVIDER'
+        else:
+            error_type = 'FILE_GONE'
+    else:
+        error_type = 'DONT_KNOW'
+    file_path = kwargs.get('path', file_node.path)
+    if file_node.name:
```
felliott: `file_name = file_node.name or os.path.basename(file_path)`

website/addons/base/views.py
```diff
@@ -466,20 +503,41 @@ def addon_view_or_download_file_legacy(**kwargs):
...
-            'message_long': 'This file does not exist'
-        })
+    file_node = kwargs.get('file_node') or TrashedFileNode.load(kwargs.get('trashed_id'))
+    deleted_by, deleted_on = None, None
+    if isinstance(file_node, TrashedFileNode):
+        # TODO test for None once hotfix is in
+        print('attrs are ' + str(dir(file_node)))
+        deleted_by = str(file_node.deleted_by)
+        deleted_on = str(file_node.deleted_on)
+        if deleted_by == "None":
```
felliott: `error_type = 'BLAME_PROVIDER' if file_node.deleted_by is None else 'FILE_GONE'`

website/files/models/base.py
```diff
@@ -101,6 +101,18 @@ def restore(self, recursive=True, parent=None):
         TrashedFileNode.remove_one(self)
         return restored
 
+    def get_guid(self):
+        """Attempt to find a Guid that points to this object.
+
+        :param TrashedFileNode or StoredFileNode file_node
```
felliott: Delete this line, `file_node` is not a param.

website/files/models/base.py
```diff
@@ -192,7 +204,10 @@ def wrapped(self):
     def get_guid(self, create=False):
         """Attempt to find a Guid that points to this object.
         One will be created if requested.
-        :rtype: Guid
+
+        :param TrashedFileNode or StoredFileNode file_node
```
felliott: Ditto.

website/files/models/base.py
```diff
@@ -192,7 +204,10 @@ def wrapped(self):
...
+        :param Boolean create: True will generate GUID if there isn't one
```
felliott: copyedit: "Should we generate GUID if there isn't one? Default: False"

website/addons/base/views.py
TomBaxter: No. Removed.

website/addons/base/views.py
TomBaxter: Done and yes, done.

website/addons/base/views.py
TomBaxter: Complete.

website/addons/base/views.py
TomBaxter: Looks like they suppress the display of buttons, without them you get...
err_msg_scrreny

website/addons/base/views.py
TomBaxter: No, removed. Done.

website/addons/base/views.py
TomBaxter: Done.

website/files/models/base.py
TomBaxter: I'm not sure. This is how the def was setup when I came to it. I only added the usage. My best guess is we should leave it be, as I'm not sure what all calls this.

felliott: Sorry, by "copyedit" I just meant to change the docstring to the string above.

website/addons/base/views.py
felliott: :+1: Silly me and good catch!

website/addons/base/views.py
```diff
@@ -40,15 +40,49 @@
...
 </style>
 <div class="alert alert-info" role="alert">
-This link to the file "{file_name}" is no longer valid.
-</div>'''
+<p>
+The file "{file_name}" stored on {provider} was deleted via the OSF.
+</p>
+<p>
+
+It was deleted by <a href="/{deleted_by}">{deleted_by}</a> on {deleted_on}.
```
felliott: The url and the link text should be different.  This currently tries to build a link the the user's name.  Instead this should create something like `<a href="/mst3k">John Doe</a>`.

website/files/models/base.py
TomBaxter: Done

website/addons/base/views.py
TomBaxter: Done

website/addons/base/views.py
```diff
@@ -516,21 +570,21 @@ def addon_view_or_download_file(auth, path, provider, **kwargs):
...
-            'message_long': 'The add-on containing this file is no longer authorized.'
+            'message_long': 'The {} add-on containing {} is no longer authorized.'.format(provider, path)
         })
 
     if not node_addon.complete:
-        raise HTTPError(httplib.BAD_REQUEST, {
+        raise HTTPError(httplib.BAD_REQUEST, data={
             'message_short': 'Bad Request',
-            'message_long': 'The add-on containing this file is no longer configured.'
+            'message_long': 'The {} add-on containing {} is no longer configured.'.format(provider, path)
```
abought: HTTP error messages (error.mako) are treated as trusted text.

The path (really, any user entered text inserted via `string.format` in HTTPErrors) will also need to be escaped with `markupsafe.escape` accordingly.

TomBaxter: @felliott @abought  Done and pushed.

felliott:
LGTM. :+1:

website/addons/base/views.py
```diff
@@ -466,20 +501,35 @@ def addon_view_or_download_file_legacy(**kwargs):
...
+        error_type = 'BLAME_PROVIDER' if file_node.deleted_by is None else 'FILE_GONE'
+    else:
+        error_type = 'DONT_KNOW'
+    file_path = kwargs.get('path', file_node.path)
+    file_name = file_node.name or os.path.basename(file_path)
+    file_name_title, file_name_ext = os.path.splitext(file_name)
+    provider_full = settings.ADDONS_AVAILABLE_DICT[file_node.provider].full_name
     ret = serialize_node(node, auth, primary=True)
     ret.update(rubeus.collect_addon_assets(node))
+    retError = ERROR_MESSAGES[error_type].format(file_name=markupsafe.escape(file_name),
```
abought: Because this is a bit of an unusual error reporting mechanism, could you be sure to test this with a malformed google docs filename of the sort we discussed?

In particular, I want to make sure that special characters aren't being double-escaped (once backend, once frontend)

Due to a PR that landed earlier today, Markupsafe rules recently became far more widespread... so for most certain results, consider pulling from develop. Let me know if I can clarify in person.

TomBaxter: Reviewed with Andy, looks good for now.

website/files/models/base.py
sloria: This is still using `.find()`; @TomBaxter did you mean to change this to `.find_one`? Keep in mind, you'll also need to handle `NoResultsFound` if you make that change.

sloria:
@TomBaxter Can you please get this up to date with develop?

website/files/models/base.py
fitcosz: @sloria 
  @erinspace originally wanted find_one but then reconsidered. If I remember correctly, the retraction was to do with handling NoResultsFound as you mentioned. I'm happy to go either way with it.
  Should I switch to find_one?

TomBaxter: @sloria @erinspace  Ha! fitcosz is an auth used for waterbutler testing. Sorry for any confusion.

TomBaxter:
@sloria Conflicts resolved and latest develop merged.

PR #5330, opened 2016-03-30
[Fix] Ensure Archiver-required field returns correct value w/ GH

mfraezz:
Purpose
======
Fix registration failures on staging caused by the `getattr` in [archive_folder_name](https://github.com/CenterForOpenScience/osf.io/blob/release/0.66.0-develop/website/addons/base/__init__.py#L727). GH doesn't quite fit the `StorageAddon` generalization, but is close enough to me made to work, and `user/repo` is the closest approximation to `folder_name` that `GitHubNodeSettings` has.

Changes
=======
Make field used by archiver a property that returns the GitHub approximation of what that field should be: 
* `folder_name` <-- `user/repo`

Side Effects
=========
None expected.

### DIFF
--- a/website/addons/github/model.py
+++ b/website/addons/github/model.py
@@ -96,17 +96,17 @@ class GitHubNodeSettings(StorageAddonBase, AddonOAuthNodeSettingsBase):
 
     @property
     def folder_id(self):
-        return self.repo
+        return self.repo or None
 
     @property
     def folder_name(self):
-        return self.repo
-
-    @property
-    def folder_path(self):
         if self.complete:
             return '{}/{}'.format(self.user, self.repo)
         return None
+
+    @property
+    def folder_path(self):
+        return self.repo or None
 
     @property
     def has_auth(self):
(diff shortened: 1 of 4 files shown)


brianjgeiger:
Were you able to create a unit test that would cause the archive test to fail that was fixed by your update? I saw some test changes in there, but I'm not sure how thoroughly it tests the issue being addressed.

mfraezz:
Test changes in this are (currently) only to keep `/addons/github/tests/` passing. Some unknown number of archiver tests have noops (`mock_obj.assert_anything(False)` returns `True`), and while fixing that test suite is important, it's not as important as quickly fixing a single known case of failure so that QA can test registrations. (But someone who actually knows archiver should probably do that soon.)

felliott:
Applying this will require a manual migration of the staging and staging2 databases:

````
    db['githubnodesettings'].update({}, {'$unset': {'folder_id': None, 'folder_name': None, 'folder_path': None}}, multi=True)
    commit()
````

These fields were added as part of the refactor, but were inferred from the model.  Now that they've been removed, future migrations should not add them.

PR #5336, opened 2016-03-30
OSF-5199 No forbidden error when accessing settings page

adlius:
<h3>
Purpose
</h3>
When trying to access a public project's settings as a non-contributor (through typing url in the browser), no Forbidden error was shown. Instead, a blank project page was rendered. This PR corrects this problem.

<h3>
Changes
</h3>
This is an interesting bug. If the code reviewers don't mind the long story of how I discovered what exactly the problem is, here's the long story:
According to the problem description and actual recreation of the bug, it is obvious that when a user tries to access the settings page of a public project, the view renders successfully. The reason it shows a blank page is that the template decides what and what not to show according to the user's permissions. In this case, the template decides to show nothing since user's permission match none of the conditionals in the template file.
But the problem is that permission control should not be handled by the template file. The fact that the server decide to return a view object to this supposedly forbidden request means the permission decorator, specifically the "@must_be_contributors" decorator, preceding the node_settings() function in the node.py file does not handle permission correctly.
So I chased to the decorators.py file to look for answers. At first, it seems that the _must_be_contributor_factory() function doesn't check whether the user is a contributor at all! However, such a huge "mistake" should be indicated by failed unit tests that checks the decorator. It seems highly unlikely that both the unit test and the decorator function is implemented incorrectly.
Closer inspection shows that the problems lies with "assumptions". 
<h5>
TL:DR? Please start reading from here.
</h5>
I believe the original implementations of the decorators are based on the assumption that being "public" and "only contributors can view this" are mutually exclusive. In other words, if a project is public, then everyone, even if it is not a contributor, can view this. This is further exemplified by the "@must_be_contributors_or_public" and "@must_be_contributors_or_public_but_not_anonymized" decorators generated by the factory function.
Thus, the "@must_be_contributor" decorator is also generated by the factory function based on the same assumptions. However, what "must be contributor" means here for the settings page is actually "must be contributor even if the project is public", which contradicts with the general assumption that "project is public" and"project can only be viewed by contributors" are mutually exclusive.
<h5>
"Enough with your story, so what exactly are the changes?" Fine, real TL:DR starts here.
</h5> 
In decorators.py, I add another decorator "@must_be_contributor_even_if_public".
In node.py, I make use of the new decorator.

<h3>
P.S.
</h3>
I am obviously really excited to find this bug and maybe I am a bit too excited. But this is truly educational for me as a newbie developer in that it shows how different assumptions may change the result. So I take some time and effort to write the long story as a reflection to remind myself. I hope I correctly diagnose the problem. If you guys have any better ideas on how to fix this please comment.

### DIFF
--- a/website/project/decorators.py
+++ b/website/project/decorators.py
@@ -257,7 +257,6 @@ def must_be_contributor_even_if_public(func):
 
     @functools.wraps(func)
     def wrapped(*args, **kwargs):
-        response = None
         _inject_nodes(kwargs)
         node = kwargs['node']
 
--- a/website/project/decorators.py
+++ b/website/project/decorators.py
@@ -247,6 +247,31 @@ must_be_contributor_or_public = _must_be_contributor_factory(True)
 must_be_contributor_or_public_but_not_anonymized = _must_be_contributor_factory(include_public=True, include_view_only_anon=False)
 
 
+def must_be_contributor_even_if_public(func):
+    """Decorator that verifies whether user is a contributor even if the node is public
+
+    :returns: Decorator function
+    :raises: http.Forbidden if user is not contributor and the node is public
+
+    """
+
+    @functools.wraps(func)
+    def wrapped(*args, **kwargs):
+        response = None
+        _inject_nodes(kwargs)
+        node = kwargs['node']
+
+        kwargs['auth'] = Auth.from_kwargs(request.args.to_dict(), kwargs)
+        user = kwargs['auth'].user
+
+        if not node.is_contributor(user):
+            raise HTTPError(http.FORBIDDEN)
+
+        return func(*args, **kwargs)
+
+    return wrapped
+
+
 def must_have_addon(addon_name, model):
     """Decorator factory that ensures that a given addon has been added to
     the target node. The decorated function will throw a 404 if the required
--- a/website/project/views/node.py
+++ b/website/project/views/node.py
@@ -25,7 +25,7 @@ from website.project import new_node, new_private_link
 from website.project.decorators import (
     must_be_contributor_or_public_but_not_anonymized,
     must_be_contributor_or_public,
-    must_be_contributor,
+    must_be_contributor_even_if_public,
     must_be_valid_project,
     must_have_permission,
     must_not_be_registration,
@@ -290,7 +290,7 @@ def node_forks(auth, node, **kwargs):
 
 @must_be_valid_project
 @must_be_logged_in
-@must_be_contributor
+@must_be_contributor_even_if_public
 def node_setting(auth, node, **kwargs):
 
     ret = _view_project(node, auth, primary=True)


website/project/decorators.py
```diff
@@ -247,6 +247,30 @@ def wrapped(*args, **kwargs):
 must_be_contributor_or_public_but_not_anonymized = _must_be_contributor_factory(include_public=True, include_view_only_anon=False)
 
 
+def must_be_contributor_even_if_public(func):
```
mfraezz: I'm not sure that this is right approach. This file has a [decorator factory](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/project/decorators.py#L194) intended to handle contributor / public and anonymous behavior. Three of the four possible behavior patterns are [defined above](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/project/decorators.py#L244-L247), and it seems like the one we want for this issue is the yet-undefined one --  `(include_public=False, include_view_only_anon=False)`. 

However, `include_view_only_anon` shouldn't matter for this case. As a non-contributor is making it through the `@must_be_contributor` decorator, there's apparently a bug in the decorator factory.

mfraezz: There's one other use of the `@must_be_contributor` decorator, in the `project_remove_contributor` view function. Also not something we want A/VOL users to have access to.

It seems like either a) the default for `include_view_only_anon` in the decorator factory should be `False`, and passed in as `True` for `must_be_contributor_or_public` or b) it should be explicitly defined in the three places that factory is used.

Also, additional logic needs to be added to raise a `403 FORBIDDEN` in the correct case (`not (allow_anon & user_is_anon) & not public_node & not user_is_contrib`)

mfraezz: Side note: the contributors page gets around this by checking that the user has at least `read` permissions, presumably because the `@must_be_contributor` decorator doesn't actually check that the user must be a contributor. This works because public projects allow non-contribs to view, but they don't explicitly have `read` access, the minimum permission granted to a contributor.

This would also be an acceptable workaround, but if you go this route please add a bug ticket indicating that `_must_be_contributor_factory` is flawed.

mfraezz:
Pass complete, back to you @adlius :octocat:

adlius:
@mfraezz This is a really tricky bug to follow. As I have mentioned in my first comment, I have realized that there might be some problem with the decorator factory function in that it doesn't explicitly check whether the user "is_contributor()" or not. But I thought such a huge mistake would definitely be caught by the unit tests. However, I looked at the unit test (https://github.com/CenterForOpenScience/osf.io/blob/develop/tests/test_auth.py#L208) and it does check the decorator function, albeit not every one of them. The test "test_must_be_contributor_when_user_is_not_contributor_raises_error" does pass. However when I change the project to be a public one in the setup function, the test failed. Hence my conclusion that the problem lies with what "must_be_contributor" means. I believe the original assumption was that if a project is public, then it is viewable to everyone including non-contributors. But for the settings page, the "must_be_contributor" decorator means "must_be_contributor_even_if_public", which is why I add such a decorator.
Leaving all the above aside, when I read your comment, my first thought was "did or did not the decorator function really check for contributors?" Because if it does check for contributors, albeit not explicitly, it means the decorator factory function was implemented correctly. So I poked around more. The decorator factory function calls the "check_can_access()" function (https://github.com/CenterForOpenScience/osf.io/blob/develop/website/project/decorators.py#L228) which in turn calls the "can_view()" function () (https://github.com/CenterForOpenScience/osf.io/blob/develop/website/project/decorators.py#L174). The "can_view()" would then check whether the user has "read" permission. (https://github.com/CenterForOpenScience/osf.io/blob/519d8aab2a5ae946a50122359e5172c212c4fcc4/website/project/model.py#L1075). If the "read" permission is the minimum permission granted to a contributor, then the factory function does check whether the user is a contributor, just not explicitly and the code could definitely need some refactoring. However, all this prove that the decorator factory function works fine under the original assumption, which is "if a node is public, then everyone can access it".
Also, I wonder why the "node_contributors()" function (which returns a view for contributor page of a project) use the "must_have_permission(READ)" decorator instead of the "must_be_contributor" decorator. Does that mean "must_have_permission(READ)" is synonymous to "must_be_contributor (no matter public or not)"? If so, then I should revert my change and use the same decorator for the settings page.
Either way, there seems to be some confusion regarding the permissions and authentications. I could definitely add a ticket to JIRA for refactoring the code. What do you think is the best solution? I could use some advice or you may correct me if I am wrong.
I apologize for the lengthy feedback. But I feel the need to do so due to the "complexity" of this bug.

mfraezz:
Lots of feedback, but I agree that it's complex. I'll try to address everything as best I can --

> I thought such a huge mistake would definitely be caught by the unit tests.

You're not the first person to ever make that erroneous assumption, don't feel bad. In light of this, I'd suggest first refactoring the linked [test class](https://github.com/CenterForOpenScience/osf.io/blob/develop/tests/test_auth.py#L208) to actually test all cases of intended functionality for all three of the decorators (:tada: TDD :tada:). I can help flesh out "all cases" if you want, but I don't think I need to enumerate them here.

> [W]hen I change the project to be a public one in the setup function, [`test_must_be_contributor_when_user_is_not_contributor_raises_error`] failed.

The `must_be_contributor` decorator shouldn't care whether or not the node is public, that's what the `must_be_contributor_or_public` decorator is for.

>[T]he "must_be_contributor" decorator means "must_be_contributor_even_if_public", which is why I add such a decorator

We're in agreement regarding what the decorator means, but I think the best approach is to fix the behavior. This could be done in one of two ways -- by fixing the factory (in a manner similar to my original comments), or by replacing the factory with three targeted decorator functions of the same names as what the factory is being used to create. Either approach merits increasing test coverage, as detailed above.

>[M]y first thought was "did or did not [`must_be_contributor`] really check for contributors?" [Further investigation into that question]

Short answer: No, but it should.

>If the "read" permission is the minimum permission granted to a contributor, then the factory function does check whether the user is a contributor, just not explicitly and the code could definitely need some refactoring. 

Not quite... it's worth noting that there's a distinction between being granted `read` access as a contributor versus being able to view a node because it's public -- viewing a node doesn't necessarily mean you have `read` access. 

>However, all this prove that the decorator factory function works fine under the original assumption, which is "if a node is public, then everyone can access it".

That's what the more widely-used `must_be_contributor_or_public` decorator is for. The lack of `_or_public` signifies that it should be more restrictive, and used for things that you wouldn't want to allow noncontributors to do on public nodes. The only two things that are actually decorated with `must_be_contributor` are a) viewing the settings page and b) removing contributors, neither of which should be available to noncontributors.

>What do you think is the best solution?

The decorators need to be fixed (or removed/replaced), but not necessarily now. If you're feeling overwhelmed by the scope of this, just change it to use the same `must_have_permission(READ)` decorator that the `node_contributors` page uses (which, incongruously, actually does a better job of checking for contributorship than does `must_be_contributor`), and add a Jira ticket for this `must_be_contributor` bug.

mfraezz:
As always, let me know you have any other questions or concerns, or if I failed to address any existing ones. :grimacing:

PR #5348, opened 2016-03-31
Make requirement invoking more convenient

brianjgeiger:
Purpose
-----------
This is a backwards compatible update to invoking requirements that will make it easy to regularly update all of the normal development requirements with a single command.

Changes
------------
The invoke requirements task will take multiple flags. The only thing that won't work in a group with everything else is release, which only works with addons. If you want to do base and addons, you could do `invoke requirements --base --addons`. Or you could do `invoke requirements --quick` which will do, in order, addons, dev, and base.

Side effects
----------------
Shouldn't be any. Invoking requirements with only one option still only runs the one item requested, and invoking with no options will just run the base requirements.


tasks/__init__.py
```diff
@@ -426,28 +426,42 @@ def flake():
...
     Examples:
+        inv requirements
+        inv requirements --quick
 
-        inv requirements --dev
-        inv requirements --addons
-        inv requirements --release
-        inv requirements --metrics
+    Quick requirements are, in order, addons, dev and the base requirements. They should do for day-to-day development
+    work so you get everything that someone might have added that would keep you from running the software properly.
```
samchrisinger: This second sentence is hurting my brain.

tasks/__init__.py
```diff
@@ -426,28 +426,42 @@ def flake():
...
+
+    By default, base requirements will run. However, if any set of addons, release, dev, or metrics are chosen, base
```
samchrisinger: > By default, base requirements will run.
Except `base=False`, and if I want to include requirements for addons, dev, or metrics the base requirements are not installed.

Maybe I'm not  reading the code correctly, but for:
> Or you could do invoke requirements --quick which will do, in order, addons, dev, and base.

I'd think the else branch on L455 would get hit with: `dev=True, metrics=True, base=False`

tasks/__init__.py
brianjgeiger: Fixed on next push.

tasks/__init__.py
brianjgeiger: L447 sets `base` to true if you are not specifying anything else.

brianjgeiger: And yes, if you set `--dev --metrics` then you would only run the dev and metrics branches.

Most of the complication comes from the backwards compatibility for the release branch, incidentally. It would have been much simpler otherwise.

samchrisinger: But setting `--quick` means L448 will never run, consequently `base == False`

tasks/__init__.py
```diff
@@ -426,28 +426,42 @@ def flake():
...
     Examples:
+        inv requirements
+        inv requirements --quick
 
-        inv requirements --dev
-        inv requirements --addons
-        inv requirements --release
-        inv requirements --metrics
+    Quick requirements are, in order, addons, dev and the base requirements. You should be able to use --quick for
+    day to day development.
```
samchrisinger: Thank you :smiley: :revolving_hearts:

tasks/__init__.py
brianjgeiger: L444

tasks/__init__.py
```diff
@@ -426,28 +426,42 @@ def flake():
...
         req_file = os.path.join(HERE, 'requirements', 'release.txt')
-    elif dev:  # then dev requirements
-        req_file = os.path.join(HERE, 'requirements', 'dev.txt')
-    elif metrics:  # then dev requirements
-        req_file = os.path.join(HERE, 'requirements', 'metrics.txt')
-    else:  # then base requirements
-        req_file = os.path.join(HERE, 'requirements.txt')
-    run(pip_install(req_file), echo=True)
+        run(pip_install(req_file), echo=True)
+    else:
```
samchrisinger: Sadly GH obliterated our old comments.

But if you ever use the `--quick` flag _minus_ `--base` L448 will never run, and `base` will remain false

samchrisinger: What about just adding `base = True` to the `if quick:` block?

samchrisinger: I can't read

PR #5364, opened 2016-04-01
Fix search page "unresolved" error for malformed queries [OSF-4538]

abought:
Refs https://openscience.atlassian.net/browse/OSF-4538

## Purpose
The search page was returning "unable to resolve" errors for certain kinds of malformed queries, when it should have said "bad search query". This was traced to a problem where elasticsearch returned the wrong error code (not 400) in certain cases.

## Summary of changes
- Capture ES failures  as a generic search exception (if specific cases not already handled)
- Show a more user-friendly error message

## Testing notes
Should no longer get "unable to resolve" errors in a growl box. See sample strings in https://openscience.atlassian.net/browse/OSF-4538?focusedCommentId=30463&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-30463

## Side effects
Since the ES error code is less specific than it should be, users may get a less helpful error message than intended. We will continue to log these exceptions to Sentry until a more permanent fix can be arranged.

This is a bit tricky to unit-test, because the error does not happen locally for me. (though it does happen on staging). Suggestions welcome.

### DIFF
--- a/website/search/elastic_search.py
+++ b/website/search/elastic_search.py
@@ -14,6 +14,7 @@ from elasticsearch import (
     Elasticsearch,
     NotFoundError,
     RequestError,
+    TransportError,
     helpers,
 )
 from modularodm import Q
@@ -83,6 +84,9 @@ def requires_search(func):
             except RequestError as e:
                 if 'ParseException' in e.error:
                     raise exceptions.MalformedQueryError(e.error)
+                raise exceptions.SearchException(e.error)
+            except TransportError as e:
+                # Catch and wrap generic ES 500 codes. TODO: Improve fix for https://openscience.atlassian.net/browse/OSF-4538
                 raise exceptions.SearchException(e.error)
 
         sentry.log_message('Elastic search action failed. Is elasticsearch running?')
--- a/website/search/elastic_search.py
+++ b/website/search/elastic_search.py
@@ -2,23 +2,22 @@
 
 from __future__ import division
 
-import re
 import copy
-import math
-import logging
-import unicodedata
 import functools
+import logging
+import math
+import re
+import unicodedata
 
-import six
-
-from modularodm import Q
 from elasticsearch import (
-    Elasticsearch,
-    RequestError,
-    NotFoundError,
     ConnectionError,
+    Elasticsearch,
+    NotFoundError,
+    RequestError,
     helpers,
 )
+from modularodm import Q
+import six
 
 from framework import sentry
 from framework.celery_tasks import app as celery_app
@@ -27,11 +26,11 @@ from framework.mongo.utils import paginated
 from website import settings
 from website.filters import gravatar
 from website.models import User, Node
+from website.project.licenses import serialize_node_license_record
 from website.search import exceptions
 from website.search.util import build_query
 from website.util import sanitize
 from website.views import validate_page_num
-from website.project.licenses import serialize_node_license_record
 
 logger = logging.getLogger(__name__)
 
--- a/website/search/elastic_search.py
+++ b/website/search/elastic_search.py
@@ -86,7 +86,7 @@ def requires_search(func):
                     raise exceptions.MalformedQueryError(e.error)
                 raise exceptions.SearchException(e.error)
             except TransportError as e:
-                # Catch and wrap generic ES 500 codes. TODO: Improve fix for https://openscience.atlassian.net/browse/OSF-4538
+                # Catch and wrap generic uncaught ES error codes. TODO: Improve fix for https://openscience.atlassian.net/browse/OSF-4538
                 raise exceptions.SearchException(e.error)
 
         sentry.log_message('Elastic search action failed. Is elasticsearch running?')
--- a/website/search/views.py
+++ b/website/search/views.py
@@ -14,6 +14,7 @@ from modularodm import Q
 from framework.auth.decorators import collect_auth
 from framework.auth.decorators import must_be_logged_in
 from framework.exceptions import HTTPError
+from framework import sentry
 from website import settings
 from website.models import Node, User
 from website.project.views.contributor import get_node_contributors_abbrev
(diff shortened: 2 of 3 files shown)


whit537:
##### Punchlist—for @abought 

 - [ ] make Travis happy

whit537:
##### Investigations—@whit537

 - [ ] reproduce the error locally
 - [ ] verify the workaround

abought:
> make Travis happy

Thanks. Per CI tests, we're waiting on one more PR to fix failing admin user tests. (you'll notice that CI has been a little bumpy Friday through today)

I'll get this restarted as soon as that PR is in.

website/search/elastic_search.py
```diff
@@ -85,6 +85,9 @@ def wrapped(*args, **kwargs):
                 if 'ParseException' in e.error:
                     raise exceptions.MalformedQueryError(e.error)
                 raise exceptions.SearchException(e.error)
+            except TransportError as e:
+                # Catch and wrap generic ES 500 codes. TODO: Improve fix for https://openscience.atlassian.net/browse/OSF-4538
+                raise exceptions.SearchException(e.error)
 
         sentry.log_message('Elastic search action failed. Is elasticsearch running?')
         raise exceptions.SearchUnavailableError("Failed to connect to elasticsearch")
```
whit537: While we're here, how about a light refactor to make the logic clearer? This exception could go up under an `if es is None` at the top, and then the `try`/`except` could be dedented a level.

whit537: Nevermind. This is marked critical, let's not spend the time.

whit537:
This reads fine to me for a quick fix, @abought. I am going to spend a little more time trying to trigger the error and the fix locally, but if I don't get to that before this is otherwise ready to move forward then don't wait for me. :+1:

website/search/elastic_search.py
```diff
@@ -85,6 +85,9 @@ def wrapped(*args, **kwargs):
                 if 'ParseException' in e.error:
                     raise exceptions.MalformedQueryError(e.error)
                 raise exceptions.SearchException(e.error)
+            except TransportError as e:
```
whit537: Since `RequestError` is a [subclass](https://github.com/elastic/elasticsearch-py/blob/1.3.0/elasticsearch/exceptions.py#L85) of `TransportError`, we could broaden the previous `except` and drop this one (the risk of matching a non-`RequestError` `ParseException` seems slight).

website/search/elastic_search.py
```diff
@@ -85,6 +85,9 @@ def wrapped(*args, **kwargs):
...
+                # Catch and wrap generic ES 500 codes. TODO: Improve fix for https://openscience.atlassian.net/browse/OSF-4538
```
whit537: As I read the elasticsearch client ([one](https://github.com/elastic/elasticsearch-py/blob/1.3.0/elasticsearch/connection/http_requests.py#L69-L72), [two](https://github.com/elastic/elasticsearch-py/blob/1.3.0/elasticsearch/connection/base.py#L91-L102)), `TransportError` or a [subclass](https://github.com/elastic/elasticsearch-py/blob/1.3.0/elasticsearch/exceptions.py#L77-L103) is used for any response code greater than or equal to 300.

website/search/elastic_search.py
whit537: I've made a "nice to have" punchlist. :)

website/search/views.py
```diff
@@ -51,6 +48,16 @@ def wrapped(*args, **kwargs):
...
             })
+        except exceptions.SearchException:
+            # Interim fix for issue where ES fails with 500 in some settings- ensure exception is still logged until it can be better debugged. See OSF-4538
+            sentry.log_exception()
+            sentry.log_message('Elasticsearch returned an unexpected error response')
+            # TODO: Add a test; may need to mock out the error response due to inability to reproduce error code locally
+            raise HTTPError(http.BAD_REQUEST, data={
+                'message_short': 'Could not perform search query',
+                'message_long': ('Please check our help (the question mark beside the search box) for more information '
+                                 'on advanced search queries.'),
```
whit537: Is this really the long message we want to give folks in this case? If it is, then we could/should [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) this up with the `MalformedQueryError` case (on the pattern of our [underlying](https://github.com/CenterForOpenScience/osf.io/blob/c71d014da1739fd6858bfde5d7c3bac5f730241b/website/search/elastic_search.py#L84-L87) `RequestError` ([&rarr; `TransportError`](https://github.com/CenterForOpenScience/osf.io/pull/5364#discussion_r58397540)) handling, since `MalformedQueryError` subclasses `SearchException`). If not, then we should change it. :-)

abought: Per DRYing the error message, I intentionally made `message_short` less specific, and kept the exception generic. Because we have already failed to catch a known error (parsing), we are forced to catch a generic 500 error, which could occur for a range of reasons. I wanted to de-emphasize the importance of syntax vs the text in line 49, to avoid misleading the user.

That said, moving `message_long` to `website/language.py` would be entirely reasonable. I'll do this once CI is de-borked. (I need an excuse to make CI rerun, anyway!)

Long term: I'd take the error message out of the backend entirely, but that would require more substantial changes to front-end code. Spots like line 49 (backend code containing raw HTML that must be trusted as written) are a bad idea; if possible user-facing presentation text should live in the front end, eg `osfLanguage.js`.

website/search/elastic_search.py
abought: This is a tough call. `RequestErrors` specifically indicate a 400-series error code, whereas `TransportErrors` can be less specific.

I agree the block seems more redundant- but I also feel that the true fix for this ticket would involve understanding why we get an incorrect error code at all. That will probably involve dealing with server logs, though.

Not sure if it was the right choice, but I wrote this code as a separate block so it would be easier to remove cleanly later.

website/search/elastic_search.py
abought: Well, greater than 400, anyway. Will tweak the comment.

https://github.com/elastic/elasticsearch-py/blob/1.3.0/elasticsearch/exceptions.py#L27

whit537: Well, if I'm reading the code right, the docstring is wrong:

https://github.com/elastic/elasticsearch-py/blob/1.3.0/elasticsearch/connection/http_requests.py#L69-L72

website/search/elastic_search.py
whit537: > I wrote this code as a separate block so it would be easier to remove cleanly later.

Fair enough.

abought:
I have made the requested changes to DRY error language text and moved this back into CR, with passing tests. 9e240be

whit537:
:+1:

whit537:
I've updated the punchlist. I'm all set here. Go science! :-)

PR #5367, opened 2016-04-01
Fix JS tests- restore code lost during merge commit

abought:
Code lost in: d08b649142834c3c3e3df873c86bfae761cb2f50

Fix JSHint/ Karma failure in CI- function is called, but code was lost in merge commit.


brianjgeiger:
Looks good. When tests pass, or if someone more familiar with the code in question *ahem* @caneruguz *ahem* code reviews it, I'll merge it in.

PR #5368, opened 2016-04-01
RFR Remove Admin Self Redirect

billyhunt:
Jira issue# https://openscience.atlassian.net/browse/OSF-5947

Fixes problems where users cannot remove themselves from multiple nodes in production.  Also fixes redirect when users can no longer see the node.


icereval:
closed by #5369

PR #5370, opened 2016-04-04
[Feature] Admin node refactor

Ghalko:
## Proposal

A cleanup of the admin node interface from improvements accumulated over development.

ADM-82

## Repercussions

More security, using groups and `login_required`. Otherwise no other repercussions.


admin/base/utils.py
```diff
@@ -0,0 +1,22 @@
+"""
+Utility functions and classes
+"""
+from django.contrib.auth.views import redirect_to_login
+from django.core.exceptions import PermissionDenied
+from django.contrib.auth.mixins import UserPassesTestMixin
+
+
+class OSFAdmin(UserPassesTestMixin):
+    login_url = '/admin/auth/login'
```
mfraezz: Is this necessary? Aside from the lack of a trailing slash, it's the same as `admin_settings.LOGIN_URL`.

Ghalko: You are correct.

admin/nodes/urls.py
```diff
@@ -1,18 +1,18 @@
 from django.conf.urls import url
+from admin.nodes import views
 
-from . import views
 
 urlpatterns = [
     url(r'^$', views.NodeFormView.as_view(),
         name='search'),
-    url(r'^id-(?P<guid>[a-z0-9]+)/$', views.NodeView.as_view(),
+    url(r'^(?P<guid>[a-z0-9]+)/$', views.NodeView.as_view(),
```
mfraezz: :+1:

admin/nodes/views.py
```diff
@@ -1,39 +1,29 @@
-from datetime import datetime
+from __future__ import unicode_literals
 
-from django.views.generic import ListView
+from django.views.generic import ListView, DeleteView
+from datetime import datetime
```
mfraezz: `datetime` is standard library, should be imported in the first section of imports (with but below `__future__`)

admin/nodes/views.py
```diff
@@ -42,16 +32,150 @@ def success_url(self):
...
+            flag = None
+            osf_flag = None
+            message = None
+            if node.is_deleted:
+                node.is_deleted = False
+                node.deleted_date = None
+                flag = NODE_RESTORED
+                message = 'Node {} restored.'.format(node.pk)
+                osf_flag = NodeLog.NODE_CREATED
+            elif not node.is_registration:
```
mfraezz: So, registrations can be deleted but not undeleted through the admin interface?

admin/templates/403.html
```diff
@@ -0,0 +1,11 @@
...
+{% load static %}
+{% block title %}
+    <title>403</title>
+{% endblock title %}
+{% block content %}
+    <h1><strong>403</strong></h1>
+
+    You do not have permission.
+    <h2>{{ exception | capfirst }}</h2>
+{% endblock content %}
```
mfraezz: `\n` please.

admin/nodes/views.py
Ghalko: No, they can be restored, at the moment, but not deleted.

admin/templates/403.html
Ghalko: What does the shorthand mean? For EOF?

mfraezz: A newline at EOF, yes.

screen shot 2016-04-05 at 15 41 35

admin/nodes/views.py
mfraezz: That's what I meant. :+1:

mfraezz:
You have a `flake8` issue causing the build to fail:
>./admin/base/urls.py:1:1: F401 'patterns' imported but unused

Otherwise, looking good. Just want to do some manual testing to make sure the refactored functionality is still good, then it should be RTM :penguin:

admin/nodes/views.py
Ghalko: Done.

admin/templates/403.html
Ghalko: Done.

mfraezz:
Thanks for those changes. Manual testing looks good :shipit:

PR #5371, opened 2016-04-04
[Feature Fix] Fix failing admin tests

Ghalko:
Move back to using Django's `Testcase` for admin app.

### DIFF
--- a/api/base/settings/defaults.py
+++ b/api/base/settings/defaults.py
@@ -27,6 +27,7 @@ AUTHENTICATION_BACKENDS = (
 
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = osf_settings.DEBUG_MODE
+DEBUG_PROPAGATE_EXCEPTIONS = True
 
 ALLOWED_HOSTS = [
     '.osf.io'
--- a/tests/base.py
+++ b/tests/base.py
@@ -19,6 +19,7 @@ from faker import Factory
 from nose.tools import *  # noqa (PEP8 asserts)
 from pymongo.errors import OperationFailure
 from modularodm import storage
+from django.test import SimpleTestCase, override_settings
 from django.test import TestCase as DjangoTestCase
 
 
@@ -242,7 +243,8 @@ class TestAppJSONAPI(TestApp, JSONAPIWrapper):
     delete_json_api = json_api_method('DELETE')
 
 
-class ApiAppTestCase(DjangoTestCase):
+@override_settings(DEBUG_PROPAGATE_EXCEPTIONS=True)
+class ApiAppTestCase(SimpleTestCase):
     """Base `TestCase` for OSF API v2 tests that require the WSGI app (but no database).
     """
     def setUp(self):
--- a/tests/base.py
+++ b/tests/base.py
@@ -20,6 +20,7 @@ from nose.tools import *  # noqa (PEP8 asserts)
 from pymongo.errors import OperationFailure
 from modularodm import storage
 from django.test import SimpleTestCase, override_settings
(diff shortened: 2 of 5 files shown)


brianjgeiger:
Looks good. Will merge when tests pass.

brianjgeiger:
So hopefully merging develop into this branch will fix the problems you're seeing?

Ghalko:
This is develop.

brianjgeiger:
Sorry, I thought something had been merged recently into develop, but I was looking at the wrong thing. Ignore me.

tests/factories.py
```diff
@@ -718,3 +719,69 @@ def _create(cls, *args, **kwargs):
...
         return super(NodeLicenseRecordFactory, cls)._create(*args, **kwargs)
+
+
+def render_generations_from_parent(parent, creator, num_generations):
+    current_gen = parent
+    for generation in xrange(0, num_generations):
+        next_gen = NodeFactory(
+            parent=current_gen,
+            creator=creator,
+            title=fake.science_sentence(),
```
mfraezz: `fake` here is just `faker.Factory`, which doesn't normally have `.science_<things>`. This looks like it came from `create_fakes.py`, which also called `fake.add_provider(Sciencer)`.

samchrisinger: Attempted to address this in https://github.com/CenterForOpenScience/osf.io/pull/5382/commits/64532c5040cf47eb7d3a366c7d733c05b767f74f

mfraezz: Whoops, didn't see that

tests/factories.py
```diff
@@ -718,3 +719,69 @@ def _create(cls, *args, **kwargs):
...
+        else:
+            new_parent = render_generations_from_parent(parent, creator, node_number)
+    return new_parent
+
+
+def create_fake_user():
+    email = fake.email()
+    name = fake.name()
+    parsed = impute_names(name)
+    user = UserFactory(username=email, fullname=name,
```
abought: Do we need all this boilerplate? Why can't we just use `UserFactory()` to create the instance directly? (any arguments not provided should be taken from the defaults)

Ghalko:
Copied Sam's closed PR, all builds are green.

api/base/settings/defaults.py
```diff
@@ -27,6 +27,7 @@
 
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = osf_settings.DEBUG_MODE
+DEBUG_PROPAGATE_EXCEPTIONS = True
```
cwisecarver: If you install the [django_extensions](http://django-extensions.readthedocs.org/en/latest/runserver_plus.html) package and remove this and run `python manage.py runserver_plus` you'll get werkzeug and it's interactive debugging console like you do with flask. There's also several other fancy things.

tests/base.py
```diff
@@ -251,8 +252,7 @@ def setUp(self):
         self.app = TestAppJSONAPI(api_django_app)
 
 
-@override_settings(DEBUG_PROPAGATE_EXCEPTIONS=True)
```
cwisecarver: This should be specific to the test cases or the yellow and grey exception error page with the trace turns into plain/text and everyone is :disappointed:

api/base/settings/defaults.py
Ghalko: Noted.

tests/base.py
icereval: @Ghalko, plz move this setting back to the decorator only so it only applies to the test cases and ping me with the pr (against 0.67.0) so i can merge, thx :octocat:

PR #5374, opened 2016-04-04
[feature] Fix Logs for Files Moved from Private Components [OSF-4582]

caseyrollins:
##### Purpose
- Currently on the OSF, if user A moves a file from within private component X (which user B **does not** have access to) to parent project or other component Y (which user B **does** have access to), the file moved log for project/component Y will display the name of private component X to user B.
- This PR checks if the viewing user (user B) should be able to view the name of private component X in the file moved log on project/component Y. 
 - If user B does not have access to private component X, the file moved log on project/component Y will read (generally) "user A moved (file name) from a private component to (name of project/component Y).
- Screenshots provided below for clarification. 
 - user A = Me
 - user B = Harry Styles
 - private component X = Private -- Casey
 - project/component Y = 1D

##### Before
- My view:
screen shot 2016-04-04 at 3 57 50 pm

- Harry Styles' view (bad): 
screen shot 2016-04-04 at 3 57 20 pm

##### After
- My view (unchanged):
screen shot 2016-04-04 at 3 54 47 pm

- Harry Styles' view (good):
screen shot 2016-04-04 at 3 55 47 pm

##### JIRA
- [OSF-4582](https://openscience.atlassian.net/browse/OSF-4582)

##### Notes
- @brianjgeiger, I think the ```can_view```/```canView``` variables used in this PR could use more descriptive names but I've struggled to come up with something that isn't too verbose -- suggestions would be much appreciated.


brianjgeiger:
The only suggestion I would make for can_view is to make it more descriptive of what you can view. If it means the user can view the thing that the variable is stored on, then it's fine. If it means they can view something else, then add what that is to the end.

mfraezz:
Just from the `After` view, it seems like Harry's view should also include a middle log reading
> Casey Rollins moved test.pdf from OSFStorage in **1D** to a private component

mfraezz:
Also via Jira:

> The name and link of the private component also appears in the email notifications that go out for file updates, which are dependent on the logs.

Please add one or more test(s) (probably to both `tests/test_notifications.py:TestSendEmails` for transactional emails and `:TestSendDigest` for digest emails) to make sure that:
1) the proper logs are being sent out in this case
2) `<private_component.title> not in <email.message>`

brianjgeiger:
Since there'll be more development on this, I'm going to move this to 0.68.0

saradbowman:
I think the into/out from logs need to show on the nodes, but without the private components' names. That is to say, I agree with @caseyrollins implementation above (provided that @mfraezz's comment that Harry Styles should see a middle log is included)

Example: I am an admin on both of Casey's nodes above. 
I upload a file to public project, 1D. Then I move it to Private -- Casey. To a non-contributor, the file does not appear in the project. The logs must reflect that the file has been moved from the project, otherwise there is a mysterious disappearing file.

caseyrollins:
@saradbowman @mfraezz I originally left out the middle log regarding the file storage provider because it gives additional information about a private component that a non-contributing user wouldn't know otherwise (i.e. if the middle log says that a file was moved from Dropbox, the non-contributing user now knows that Dropbox is enabled on the private component -- whether this really matters or not, I don't know, but I erred on the side of being overly private). But I also agree it seems inconsistent without the middle log reading, and I can definitely include it.

website/views.py
```diff
@@ -165,10 +165,16 @@ def serialize_log(node_log, auth=None, anonymous=False):
         'params': sanitize.unescape_entities(node_log.params),
         'date': utils.iso8601format(node_log.date),
         'node': node_log.node.serialize(auth) if node_log.node else None,
-        'anonymous': anonymous
+        'anonymous': anonymous,
+        'can_view': can_view(node_log, auth) if node_log.action == NodeLog.FILE_MOVED else True,
     }
 
 
+def can_view(node_log, auth):
```
mfraezz: In light of [Sara's comment](https://github.com/CenterForOpenScience/osf.io/pull/5374#issuecomment-205920955) I'm not sure if this is the right approach. (Also, while I haven't verified this, it seems like if you were to request a 'restricted' log from the API, you would still get all the restricted info, plus a flag saying you can't view.)

Instead, consider adding a helper function that, when `<log>.params` are serialized, checks the permissions of the requesting user and does any replacement there, rather than when the logs are rendered. Pseudocode:
```python
def sanitize_params(auth, log):
    if log.action not in UNSAFE_ACTIONS:
        return sanitize.unescape_entities(log.params)
    safe_params = copy.deepcopy(sanitize.unescape_entities(log.params))
    for key in safe_params.keys():
        if key in UNSAFE_KEYS:
            if not check_auth:
                 safe_params[key] = whatever_sanitization_process
    return safe_params
```

saradbowman:
After discussion with @caseyrollins, we agree Harry should see the middle log, in the form: 

Casey Rollins moved test.pdf from a private component to test.pdf in OSF Storage in 1D

Harry should not see the storage provider name for the private component nor the private components name.

mfraezz:
>if the middle log says that a file was moved from Dropbox, the non-contributing user now knows that Dropbox is enabled on the private component

I'm not sure if that's true. Compare the newest log in your view to the newest log in Harry's -- the bracketed (with `[]`) text in `moved <filename> from [<provider> in <private_component.title>] to <provider> in <viewable_node.title>` is replaced with `a private component`. The only info they receive about what provider the file is stored in is (or at least, should be) info for the node they have read access to.

Either way, with my suggested [params sanitization method](https://github.com/CenterForOpenScience/osf.io/pull/5374#discussion_r58588636), you shouldn't need to worry about that.

brianjgeiger:
I've chatted with @mfraezz about the sanitization approach, and it seems the most sane of the options that came up. Keeping it on the model is good, and ensuring that there are enough layers of logic to prevent unnecessary calculations is also good. Details may change, but it seems a sound approach.

caseyrollins:
Just for clarity in this thread, I believe @mfraezz and I were talking about different things when referring to the mysterious "middle log", and therefore my [previous comment](https://github.com/CenterForOpenScience/osf.io/pull/5374#issuecomment-205924244) can be disregarded.

mfraezz:
Yeah, my bad. `"(hidden|unshown) log"` would've probably been clearer.

PR #5375, opened 2016-04-04
Allow read-write users to remove linked  [OSF-440]

abought:
Refs: https://openscience.atlassian.net/browse/OSF-440

## Purpose
Allow read-write users to remove links to other projects. Should address an issue where no one, including the destination project owner, could delete a link to a private project.

Additionally, allow users with write access to delete links, *even if they don't know what the link is pointing at*. (see screenshot) This was deemed slightly preferable to an unremovable link owned by someone who is no longer a contributor.
screen shot 2016-04-05 at 1 05 31 pm

Tested "private item deletion" with a private fork, and a private node, using two users (one admin, one with write access).

## Summary of changes
In the screenshots below: an admin user created 1 link to a public project, and a RW user created 1 link to a private project (that the admin wasn't a contributor on)

For the admin, BEFORE / AFTER (same!)
screen shot 2016-04-04 at 3 40 07 pm

For the RW user, BEFORE/AFTER
screen shot 2016-04-04 at 3 40 13 pm
screen shot 2016-04-04 at 3 41 09 pm

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -54,7 +54,7 @@ from website.project.views.contributor import (
 )
 from website.project.views.node import _should_show_wiki_widget, _view_project, abbrev_authors
 from website.util import api_url_for, web_url_for
-from website.util import rubeus
+from website.util import permissions, rubeus
 
 
 class Addon(MockAddonNodeSettings):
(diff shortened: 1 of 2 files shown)


abought:
This one is clear to restart CI, provided that @brianjgeiger is ok with the behavior/ permissions change.

brianjgeiger:
That's odd that the admin user can get the title of the project. That seems improper, but not, strictly speaking, part of this issue. We'll have to file a new ticket for that.

abought:
> That's odd that the admin user can get the title of the project. That seems improper, but not, strictly speaking, part of this issue. We'll have to file a new ticket for that.

There are two links. Project P is the secret project- the admin isn't seeing that title. 

Am I missing something?

brianjgeiger:
Ah, okay. So the problem is that the admin can't *see* the link at all. If you're admin or rw, you should probably be able to see a "<Private Link>" placeholder that would give them the ability to delete.

abought:
> Ah, okay. So the problem is that the admin can't see the link at all. If you're admin or rw, you should probably be able to see a "" placeholder that would give them the ability to delete.

Per PR description:

> Exception to this rule: if the user does not have permission to see the linked node, do not show delete controls for that link. (It would be hard to figure out what they were deleting!)

If you can figure out a good way around that design hurdle, I'd be happy to work it in. Currently, I don't think the front-end gets any info about the node at all (ID or title)- only that a link to it exists.

brianjgeiger:
I'm not sure what the front end gets, but what they *could* get is the pointer's id, since pointers are (ugh) their own nodes. Which would let people with permissions delete them. No, it's not ideal for people to delete things that they don't know what they are, but it's more ideal than not letting them have access to know that their node is pointing to something that they don't know about. Let me know if you definitely don't have access to the pointer ID. We can chat with Sara once we understand the extent of the difficulty. We may move the rest of this to an enhancement issue.

brianjgeiger:
I've verified with Sara: RW and Admin users should have a delete control for the "Private Link" placeholder. So mostly it's a question of how much work it'll be to get there.

brianjgeiger:
Looks good. That's all mako code, right? Could you… write some webtests for that to ensure that the controls appear under the proper circumstances?

abought:
> Looks good. That's all mako code, right? Could you… write some webtests for that to ensure that the controls appear under the proper circumstances?

Good idea- given how *specialized* render_node.mako is, having a test here would put my mind at ease.

I've pushed the changes, but Github doesn't want to admit it; I'll check in the AM once they've sorted their outage.

abought:
Incidentally, this PR also includes some cleanup of `test_views`, including removing a duplicate test. I've done that in separate commits so that those changes can be examined in isolation.

mfraezz:
Changes look good, tests look good, and it works :star2: 

Dropping this in RTM

PR #5376, opened 2016-04-04
[feature] @mentions [OSF-5658]

leb2dg:
## Purpose

Related ticket: https://openscience.atlassian.net/browse/OSF-5658

## Changes
* At mention names appear as a link to the mentioned contributor
* An error message is shown and the submit button is disabled if a comment is over 500 characters but the validator on the model allows up to 1000 characters
    * padding is added to the model validator to account for html elements that are converted to markdown before they are saved

Mentioning a contributor:

## Possible additions
* Emojis PR #5418 
* Global notification settings for @ mentions (whether they would like to be emailed)
    * In progress
    * Note: in the related ticket, the notification settings for mentions are pulled from the comment notification settings but due to inconsistency in opinions on the mention notification functionality, what currently is implemented is everyone is notified if mentioned regardless of their notification setting for comments -- essentially functionality is up for discussion

## Side Effects
* add dependency At.js
    * provide @/+ functionality
* add new_mentions & old_mentions to the comment model
   * new_mentions is a required field so it must be passed as an attribute or the request will fail
* a user could submit a comment with up to 1000 characters using the API


whit537:
##### Investigations—for @whit537

 - [ ] run code
 - [ ] read code

whit537:
The move to `contenteditable` appears to be having unintended consequences.

##### Steps

1. Type "foo" [enter] "bar"
1. Click "Comment"

##### Expected 

```
foo
bar
```

##### Actual

```
foo<div>bar</div>
```

screen shot 2016-04-11 at 2 31 06 pm

whit537:
##### Punchlist—for @leb2dg 

 - [ ] justify move to contenteditable—

whit537:
I'm surprised that bolded text is not escaped, whereas divs are!

whit537:
screen shot 2016-04-11 at 2 42 55 pm

whit537:
`<span>` and `<br>` are also bleeding through:

screen shot 2016-04-11 at 2 44 37 pm

leb2dg:
Saved comments are rendered as markdown but the user shouldn't be able to write html. The `<div>` inserted after enter and the `<br>` inserted after shift + enter are bugs that need to be fixed. However, manually typing any html is expected to render as typed (comment text is bleached before saving to the db).

The switch to `contenteditable` is necessary to prevent the user from deleting half of a mention and to keep track of which users have been mentioned. The user's guid is added as an attribute to @ mention span element and then parsed out on save.

whit537:
> The switch to `contenteditable` is necessary

Ftr it looks like both GitHub and Jira use `textarea`, and they both have mentions.

screen shot 2016-04-11 at 3 19 41 pm

screen shot 2016-04-11 at 3 20 50 pm

whit537:
> However, manually typing any html is expected to render as typed (comment text is bleached before saving to the db).

Okay. I hadn't tried that yet, fwiw.

leb2dg:
Both Github and Jira insert uniq user IDs into the textarea so they are easily parsed. This is an option for the OSF but guids are not very user friendly.

whit537:
> Both Github and Jira insert unique user IDs into the textarea so they are easily parsed. This is an option for the OSF but guids are not very user friendly.

My hunch is that it is going to prove difficult to force a `contenteditable` to truly behave like a `textarea`. Could we store a mapping of username to userid as a plain ol' JavaScript object somewhere, rather than using the DOM (of the `contenteditable`) to encode that info?

leb2dg:
> My hunch is that it is going to prove difficult to force a contenteditable to truly behave like a textarea. 

Both Facebook and GroupMe use  `contenteditable` instead of a `textarea` since they allow multiple users with the same name. Unfortunately, `contenteditable` does require a little more logic. 
> Could we store a mapping of username to userid as a plain ol' JavaScript object somewhere, rather  than using the DOM (of the contenteditable) to encode that info?

OSF users do not have a username associated with them other than their full name so a mapping from their full name to their guid would not cover all of the use cases. For instance:
* a contributor makes an edit to their full name after they have been mentioned in a comment
* there is more than one contributor with the same full name

whit537:
I'm getting some @mentions!

screen shot 2016-04-13 at 4 33 53 pm

whit537:
Gosh, I guess I don't want to do QA's job, but the UX here feels kinda brittle to me. Here are two more examples (Chrome 49 on Mac OS):

---------------

##### Steps

 1. Type these characters: `@` `f` `o` `<tab>` `!` `<left arrow>`

##### Expectation

Cursor moves to the left of the `!`.

##### Actual

Cursor remains to the right of the `!`.

screen shot 2016-04-13 at 5 29 23 pm

-----------------

##### Steps

 1. Start with the above.
 1. `<left arrow>` (cursor is now on the left of the `!`)
 1. `<left arrow>` (cursor briefly moves one whitespace to the left, then jumps to the beginning of the line, before the `@`)
 1. `<right arrow>`

##### Expectation

Cursor jumps to right after the second `o` (`@Foo| !`), or perhaps after the whitespace (`@Foo |!`).

##### Actual

Cursor jumps to right after the `!`.

screen shot 2016-04-13 at 5 30 21 pm

leb2dg:
Thanks for being thorough, @whit537! I'm going to fix these bugs, take care of some edge cases brought up by QA, and then move this ticket back to code review.

whit537:
Sure thing! I'm just getting started, though—I haven't even read any code yet! :-)

bower.json
```diff
@@ -19,6 +19,7 @@
     "locales": "https://github.com/citation-style-language/locales.git#d2b612f8a6f764cbd66e67238636fac3888a7736",
     "raven-js": "2.1.0",
     "zeroclipboard": "2.1.6",
-    "osf-style": "https://github.com/CenterForOpenScience/osf-style.git#5dcf5997ad2808ffc7c8564f94cde102bf7a4a96"
+    "osf-style": "https://github.com/CenterForOpenScience/osf-style.git#5dcf5997ad2808ffc7c8564f94cde102bf7a4a96",
+    "At.js": "jquery.atwho#^1.5.0"
```
whit537: Ah, here's the secret! :-)

website/project/model.py
```diff
@@ -264,6 +277,13 @@ def create(cls, auth, **kwargs):
             save=False,
         )
 
+        if (comment.new_mentions):
+            comment.new_mentions = [mention for mention in comment.new_mentions if mention not in comment.old_mentions and validate_contributor(mention, comment.node.contributors)]
```
whit537: Let's try to avoid long lines like this. Try abbreviating `mention` to `m` in the list comprehension, and if it's still longer than 100 chars (I haven't seen a local style guide on this, but usually 80 or maybe 100 are the standard) then maybe factor out the condition into a lambda:

```python
foo = lambda a: a or b or c
bar = [x for x in bar if foo(x)]
```

[`filter`](https://docs.python.org/2/library/functions.html#filter) is arguably more concise than a list comprehension here:

```python
bar = filter(foo, bar)
```

Some ideas. :-)

website/project/signals.py
```diff
@@ -2,6 +2,7 @@
 
 signals = blinker.Namespace()
 comment_added = signals.signal('comment-added')
+mention_added = signals.signal('mention-added')
```
whit537: Eep. Generally not a huge fan of signals, but it seems to be the established pattern for the comments code, so I guess it makes sense to use that on this PR.

website/static/js/comment.js
```diff
@@ -16,6 +16,77 @@ var osfHelpers = require('js/osfHelpers');
...
+    displayTpl: '<li>${name} <small>${fullName}</small></li>',
+    limit: 6,
+    callbacks: callbacks
+};
+
+var plus_config = {
+    at: '+',
+    headerTpl: '<div class="atwho-header">Contributors<small>&nbsp;↑&nbsp;↓&nbsp;</small></div>',
+    insertTpl: '+${name}',
+    displayTpl: '<li>${name} <small>${fullName}</small></li>',
```
whit537: It might be worth factoring out `headerTpl` and `displayTpl` since they're the same between the two configs. Not sure how to [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) it up further than that.

website/templates/emails/mention_replies.html.mako
```diff
@@ -0,0 +1,17 @@
+<table class="comment-row" border="0" cellpadding="8" cellspacing="0" width="100%" align="center" style="font-size: 13px;background: #fff;border: 1px solid #eee;border-radius: 5px;margin-bottom: 10px;padding: 0px !important;">
+    <tr>
+        <td width="40" class="icon" valign="middle" style="border-collapse: collapse;font-size: 24px;color: #999;"> <img class="avatar" src="${gravatar_url}" width="48" alt="avatar" style="border: 0;height: auto;line-height: 100%;outline: none;text-decoration: none;border-radius: 25px;"> </td>
```
whit537: Is there a reason to have styles inline instead of in a stylesheet?

website/static/js/koHelpers.js
```diff
@@ -396,6 +396,41 @@ ko.bindingHandlers.datePicker = {
...
+        $element.html(initialValue);
+        $element.on('change input paste keyup blur', function() {
+            var observable = valueAccessor();
+            // limit input to maxlength
+            var charLimit = $element.attr('maxlength');
+            if (charLimit) {
+                if ($element.text().length < charLimit) {
+                    observable($element.html());
+                } else {
+                    $element.html(observable());
```
whit537: This has the effect of blocking me from pasting long passages of text. I would expect instead for the text to be cut off instead of disallowed entirely.

##### Steps

1. `python -c "print('a'*499, end='')" | pbcopy`
1. paste into comment box

##### Expected

The first 499 bytes are pasted.

##### Actual

Nothing is pasted.

whit537: I've updated the report based on the behavior I'm observing in production.

website/static/js/koHelpers.js
```diff
@@ -396,6 +396,41 @@ ko.bindingHandlers.datePicker = {
...
+            } else {
+                observable($element.html());
+            }
+        });
+    },
+    update: function(element, valueAccessor) {
+        var $element = $(element);
+        var initialValue = ko.utils.unwrapObservable(valueAccessor());
+        if (initialValue === '') {
+            $(element).html(initialValue);
```
whit537: What's the point of this? Why do we need to set `html()` if initialValue is `''`?

leb2dg:
You are welcome to continue looking at the code but there are some decent changes coming shortly and I wouldn't want anyone to have to do duplicate work. Changes include:
* logic for which email will be sent
    * users should not get multiple if there is a new comment and they are mentioned
* contenteditable behavior which I hadn't come across
    * the arrow key navigation
    * pasting html
    * enter key inserting html elements
* only confirmed contributors can be mentioned
    * so non-confirmed user's emails cannot be spammed
* full names will appear in dropdown and users can search based off of first or last name in list

whit537:
Sounds good @leb2dg, thanks for the heads-up. I've updated the [punchlist](https://github.com/CenterForOpenScience/osf.io/pull/5376#issuecomment-208488985) based on a first pass through the code. I'll hold off on further review until the next round of changes. Go science! :-)

website/static/js/koHelpers.js
leb2dg: It clears the box after a comment has been submitted. Without the update function the submitted comment remains in the box and without checking if the initialValue is `''` the cursor would get reset to the beginning of the box with every input. There may be a better way to handle this.

website/templates/emails/mention_replies.html.mako
leb2dg: The email templates don't have associated stylesheets. I believe it has to do with the way they are sent.

whit537: Gah, sorry. Missed that we are in an email context here. Inline styles are appropriate here.

website/static/js/koHelpers.js
```diff
@@ -396,6 +396,48 @@ ko.bindingHandlers.datePicker = {
...
+        });
+    },
+    update: function(element, valueAccessor, allBindings, bindingContext) {
+        var $element = $(element);
+        var initialValue = ko.utils.unwrapObservable(valueAccessor());
+        var bindings = bindingContext;
+        var charLimit = $element.attr('maxlength');
+        if (charLimit && bindingContext.underMaxLength) {
+            var inputTextLength = $element[0].innerText.length;
+            var errorExceedsCharacterLimit = 'Exceeds character limit. Please reduce to ' + charLimit + ' characters or less.';
```
whit537: This message is still visible after canceling the edit.

screen shot 2016-04-19 at 6 31 21 pm

website/templates/include/comment_template.mako
```diff
@@ -69,12 +69,12 @@
                         -->
                         <div data-bind="template {if: editing, afterRender: autosizeText}">
                             <div class="form-group" style="padding-top: 10px">
-                                <textarea class="form-control" data-bind="value: content, valueUpdate: 'input', attr: {maxlength: $root.MAXLENGTH}"></textarea>
+                                <div class="form-control atwho-input" placeholder="Add a comment" data-bind="editableHTML: content, attr: {maxlength: $root.MAXLENGTH}" contenteditable="true"></div>
```
whit537: The placeholder should be "Edit comment" in this case.

screen shot 2016-04-19 at 6 33 26 pm

website/static/js/koHelpers.js
whit537: (It's also goofy that the error message is so visually disassociated with the comment field it applies to, but I guess that's out of scope here.)

screen shot 2016-04-19 at 6 36 36 pm

whit537: I don't expect the error message to be shown when I paste 500 characters, but it does. It goes away with 499 characters. (I'm [counting](https://github.com/CenterForOpenScience/osf.io/pull/5376#discussion_r59874695) right, right?)

screen shot 2016-04-19 at 6 38 59 pm

website/static/js/comment.js
```diff
@@ -16,6 +16,187 @@ var osfHelpers = require('js/osfHelpers');
...
+// make sure br is always the lastChild of contenteditable so return works properly
+var lastElementBr = function(){
+    if (!this.lastChild || this.lastChild.nodeName.toLowerCase() !== 'br') {
+        this.appendChild(document.createElement('br'));
+    }
+};
+
+// ensure that return inserts a <br> in all browsers
+var onReturn = function (e) {
+    var doxExec = false;
```
whit537: Did you mean for this to be `doxExec`? I would've expected `docExec`.

whit537:
Alright, @leb2dg, I've reviewed your changes and have updated the [punchlist](https://github.com/CenterForOpenScience/osf.io/pull/5376#issuecomment-208488985) as well as my own [investigations](https://github.com/CenterForOpenScience/osf.io/pull/5376#issuecomment-208429637) todo. Gonna leave this in code review in Jira so I don't forget about it, but I won't be back until Friday so if you want to work on it in the mean time feel free to self-assign.

The `contenteditable="false"` on the span seems to have tightened up the experience considerably. Good find! :-)

leb2dg:
@whit537, I am closing this PR in favor of #5518 which has the the proper notification settings and functionality for @ mentions. Sorry, that the PR just got larger 😬

PR #5377, opened 2016-04-04
Update notification defaults to "instantly" on node creation [OSF-4685]

erinspace:
*note*: this is a redo of #5117, I made a new PR because changes were pretty extensive 

https://openscience.atlassian.net/browse/OSF-4685

# Purpose
When a user creates a project, notification settings are automatically set to "none," and a user might not know notifications even exist. This PR adds a method that automatically adds a node's creator to subscriptions for files and comments so they know the option exists.  It also will automatically subscribe any contributor added to the project.

# Changes
- Add method that subscribes a user to notifications
- Call that method with signals when a node is first saved with the nodes' creator
- Also call that method when a contributor is added with that contributor
- Add a setting that decides if a notification subscription should be automatically created (by default yes)
- Adjust that setting to False in a special case for NotificationSubscrtiption realted tests that don't create the subscriptions by default
- Disable automatic subscriptions on tests that benefit from creating them manually
- Alter text on contributor added, invite, forward invite, and public project emails to explain notifications and link to profile settings to update them

# Side effects and Notes
- Node settings for current projects are NOT touched


mfraezz:
Reminder to address [comment](https://github.com/CenterForOpenScience/osf.io/pull/5117#issuecomment-201465660) from original PR:

> According to Jira ticket:
> > [All] notification emails must be double checked for their inclusion of the explanation of where to change the user's preferences. 
>
>This should apply to `contributor_added`, `comments`, `comment_replies`, `invite`, and perhaps `new_public_project` in `website/templates/emails/`. Maybe others, as well.

tests/test_notifications.py
```diff
@@ -33,6 +33,12 @@
 
 class TestNotificationsModels(OsfTestCase):
 
+    @classmethod
+    def setUpClass(cls):
```
mfraezz: If you've got a bunch of different test classes that need to set this setting to `True` (especially since they're all in the same file), you might want to make a wrapper class for `OsfTestCase` that just does does this step. Something like
```python
class NotificationTestCase(OsfTestCase):
    @classmethod
    def setUpClass(cls):
        pass  # modify setting

    @classmethod
    def tearDownClass(cls):
       pass  # unmodify setting

class TestNotificationsModels(NotificationTestCase):
    # setUp, tests, etc.

class TestSubscriptionView(NotificationTestCase):
    # ...
```

tests/base.py
```diff
@@ -134,6 +134,9 @@ def setUpClass(cls):
         cls._original_bcrypt_log_rounds = settings.BCRYPT_LOG_ROUNDS
         settings.BCRYPT_LOG_ROUNDS = 1
 
+        cls._original_enable_notification_subscription_creation = settings.ENABLE_NOTIFICATION_SUBSCRIPTION_CREATION
+        settings.ENABLE_NOTIFICATION_SUBSCRIPTION_CREATION = False
```
mfraezz: There should also be a line in `tearDownClass` assigning this to the `_original` value

website/project/model.py
```diff
@@ -1475,6 +1497,9 @@ def save(self, *args, **kwargs):
                 save=True,
             )
 
+            if self.creator and settings.ENABLE_NOTIFICATION_SUBSCRIPTION_CREATION:
```
mfraezz: Should this be happening every time a node is saved?

mfraezz: Disregard, didn't see that this was in an `if first_save` conditional

website/templates/emails/contributor_added.txt.mako
```diff
@@ -1,7 +1,13 @@
+<%!
+    from website import settings
+%>
+
 Hello ${user.fullname},
 
 ${referrer_name + ' has added you' if referrer_name else 'You have been added'} as a contributor to the project "${node.title}" on the Open Science Framework: ${node.absolute_url}
 
+You will automatically be subscribed to notification emails for comments and updated files for this project. To change your email notification preferences, visit your project or your user settings: ${settings.DOMAIN + "settings/notifications/"}
```
mfraezz: Should this also include a link to the project's settings (ditto other emails with this language)? I'm not certain it should; this may be more confusing for the user than its worth

website/project/model.py
```diff
@@ -1335,6 +1335,28 @@ def set_node_license(self, license_id, year, copyright_holders, auth, save=False
         if save:
             self.save()
 
+    def subscribe_user_to_notifications(self, user):
+        """ Update the notification settings for the creator or contributors
+
+        :param user: User to subscribe to notifications
+        """
+        from website.notifications import utils as notification_utils
```
mfraezz: Because this is used only once here, can you do `from website.notifications.utils import to_subscription_key`?

mfraezz: Also, I'm assuming these imports are *here* to avoid circularity issues?

mfraezz:
Pass complete @erinspace, looking good :facepunch: 

In addition to the few minor things to address, please also resolve the merge conflicts. Back to you

erinspace:
@mfraezz ready for another round!

Some notable changes:
After talking with BrianG and Andy a bit, agreed that the default OsfTestCase should behave like the OSF does, and in special cases, like wanting exact control over which subscriptions are created and where for testing, should disable notification creation and do it manually. I double checked the test running times - they were both about 15minutes locally for me with both implementations.  So here's this new iteration. We can of course keep discussing it!

All that being said, Brian also pointed out an important thing - at the moment there is no good way to globally turn off all project or file notifications. There is a ticket for it - https://openscience.atlassian.net/browse/OSF-5748 - but no implementation yet. Should this be done without the other?

mfraezz:
>[T]here is no good way to globally turn off all project or file notifications. [...] Should this be done without the other?

Probably not. The annoyance potential for this is too great without a way for users to prevent auto-subscriptions to notifications indefinitely. Please close this in favor of #5452, which both includes this PR and fixes [that issue](https://openscience.atlassian.net/browse/OSF-5748).

erinspace:
Closing in favor of #5452 !

sloria:
Reopening for CR

sloria:
Ah, scratch that. Will review this in https://github.com/CenterForOpenScience/osf.io/pull/5452

PR #5385, opened 2016-04-05
Change privacy on project with publicly editable wiki [OSF-5482]

brianjgeiger:
Purpose
-----------
Allow public projects with publicly editable wikis to be made private via the API.

Closes [OSF-5482]

Changes
------------
Check within push_status_message() to see if we're running inside a Flask context. If so, under most circumstances let the message go, unless it's an error, in which case raise a 400 and pass along the error message.

Also, test all that, including the case directly related to the purpose. h/t to @mfraezz for the mock code in the unit test.

Side effects
----------------
push_status_message() isn't actually supposed to be called from anything that the API can get to, but it's littered throughout the model. If it failed, it would 500 for sure, so this is almost certainly better, but we only have the one case right now. The worst would be if someone were using 'warning' when it's really an error or something, because then it should error but might fail silently. *Probably* not, because if it fails it should fail anyways. So it should be fine, but.

### DIFF
--- a/framework/status/__init__.py
+++ b/framework/status/__init__.py
@@ -3,7 +3,6 @@
 from collections import namedtuple
 
 from framework.sessions import session
-# from api.base.api_globals import api_globals
 
 Status = namedtuple('Status', ['message', 'jumbotron', 'css_class', 'dismissible', 'trust'])  # trust=True displays msg as raw HTML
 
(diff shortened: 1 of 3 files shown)


framework/status/__init__.py
```diff
@@ -26,10 +26,20 @@ def push_status_message(message, kind='warning', dismissible=True, trust=True, j
...
     :param trust: Whether the text is safe to insert directly into HTML as given. (useful if the message includes
         custom code, eg links) If false, the message will be automatically escaped as an HTML-safe string.
-    :param jumbotron: Should this be in a jumbstron element rather than an alert
+    :param jumbotron: Should this be in a jumbotron element rather than an alert
     """
     # TODO: Change the default to trust=False once conversion to markupsafe rendering is complete
-    statuses = session.data.get('status')
+    try:
+        statuses = session.data.get('status')
+    except RuntimeError:
```
mfraezz: My only concern is that this _might_ be a bit too greedy with error handling. Is it possible for this to raise a `RuntimeError` anywhere other than what this is intended to fix, and if so, do we want to prevent failure there? If the answer to either of those is `Well, maybe...` or stronger, is there a way of verifying that *this* particular `RuntimeError` happened because of an API call?

brianjgeiger: I mean, it's only around the one line, and that's when you try to get the status from session data. But, let's see if I can find out anything else about that runtime error

brianjgeiger: Okay, there's a message I can check. If the message doesn't match, I can re-raise the error. I'll fix that and e if I can add another test case to verify.

mfraezz: It's not my biggest concern about code we maintain (and it's _very_ well documented), so I wouldn't necessarily consider it a blocker, but if it's easy to tighten up I think we should.

brianjgeiger: I can probably re-raise the exception, but I cannot think of a way of testing that the re-raising works properly. I can test all of the other branches, but not that one.

mfraezz: Try using [mock side effects](http://www.voidspace.org.uk/python/mock/mock.html#mock.Mock.side_effect)

mfraezz:
LGTM. RTM whenever the build passes :sweat_drops:

PR #5386, opened 2016-04-05
[hold][fix] Fix Keen session logic and error logging.

felliott:
The keen session cookie expiration date was being incorrectly set and silently ignored, leading to session inconsistencies in the logged data. This patch correctly sets the session time to 25min. It also explicitly logs to Sentry pageview events that Keen rejects, so that we can inspect them later.

### DIFF
--- a/website/static/js/keen.js
+++ b/website/static/js/keen.js
@@ -17,15 +17,11 @@ var KeenTracker = oop.defclass({
     },
 
     createOrUpdateKeenSession: function() {
-        var date = new Date();
+        var expDate = new Date();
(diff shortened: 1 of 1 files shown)


icereval:
@felliott, can you update with the latest `release/0.67.0` code, that should make it pass tests for a moment... maybe :-/

felliott:
Done.  Travis is passing now.

icereval:
Thanks for updating, looks good I'll merge, if you need this in prod for stats feel free to submit a hotfix (its very constrained update) :+1:

PR #5387, opened 2016-04-05
[fix] Ensure Archiver-required field returns correct value for GH

mfraezz:
Original PR (against now-closed branch): #5330

Purpose
======
Fix registration failures on staging caused by the `getattr` in [archive_folder_name](https://github.com/CenterForOpenScience/osf.io/blob/release/0.66.0-develop/website/addons/base/__init__.py#L727). GH doesn't quite fit the `StorageAddon` generalization, but is close enough to me made to work, and `user/repo` is the closest approximation to `folder_name` that `GitHubNodeSettings` has.

Changes
=======
Make field used by archiver a property that returns the GitHub approximation of what that field should be: 
* `folder_name` <-- `user/repo`
* related fields (`folder_id`, `folder_path`) properties as well.

Side Effects
=========
None expected.

### DIFF
--- a/website/addons/github/model.py
+++ b/website/addons/github/model.py
@@ -96,17 +96,17 @@ class GitHubNodeSettings(StorageAddonBase, AddonOAuthNodeSettingsBase):
 
     @property
     def folder_id(self):
-        return self.repo
+        return self.repo or None
 
     @property
     def folder_name(self):
-        return self.repo
-
-    @property
-    def folder_path(self):
         if self.complete:
             return '{}/{}'.format(self.user, self.repo)
         return None
+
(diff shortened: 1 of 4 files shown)


mfraezz:
Important comment from original PR (via @felliott):

> Applying this will require a manual migration of the staging and staging2 databases:
> 
>     db['githubnodesettings'].update({}, {'$unset': {'folder_id': None, 'folder_name': None, 'folder_path': None}}, multi=True)
>     commit()
> 
> These fields were added as part of the refactor, but were inferred from the model. Now that they've been removed, future migrations should not add them.
>

PR #5391, opened 2016-04-05
Don't allow users to view Collections as Projects [OSF-5843]

erinspace:
https://openscience.atlassian.net/browse/OSF-5843

# Purpose
Right now, users can access a collection by visiting it's guid as if it were a normal project. This PR adds a check to the ```@must_be_valid_project``` decorator that checks to see if it's a collection, and throws a 404 error with relevant message if it is.

# Changes
- Add check to ```@must_be_valid_project``` to see if guid being accessed is a collection

# Side Effects
None I can think of

### DIFF
--- a/website/project/decorators.py
+++ b/website/project/decorators.py
@@ -88,7 +88,7 @@ def must_be_valid_project(func=None, retractions_valid=False):
 
             if getattr(kwargs['node'], 'is_collection', True):
                 raise HTTPError(
-                    http.BAD_REQUEST,
+                    http.NOT_FOUND,
                     data=dict(message_long='Viewing collections is not permitted')
                 )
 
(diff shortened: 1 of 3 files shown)


website/project/decorators.py
```diff
@@ -86,6 +86,12 @@ def wrapped(*args, **kwargs):
 
             _inject_nodes(kwargs)
 
+            if getattr(kwargs['node'], 'is_collection', True):
+                raise HTTPError(
+                    http.BAD_REQUEST,
```
mfraezz: As per the Jira ticket, this should be a `404 NOT FOUND`

mfraezz:
Minor inconsistency with the expected behavior described on the ticket. Otherwise, LGTM. :fire:

website/project/decorators.py
brianjgeiger: There's a similar error with retracted registrations that treated the same, so we were leaning towards matching that.

brianjgeiger: It'll probably be quick to change either way, though, if retractions feel significantly different from this.

mfraezz:
Changes look good, dropping this in RTM :pencil2:

PR #5393, opened 2016-04-05
[hotfix] Move loadingComments ko observabale to BaseComment

erinspace:
# Purpose
@felliott  noticed that production was throwing an error because the loadingComments ko observable was defined only on the ListCommentModel and not also on the CommentModel: https://sentry.cos.io/share/issue/31312e32393036/

 This PR also adds it to the BaseModel to make sure it's good everywhere

# Changes 
- Add ```self.loadingComments = ko.observable(true);``` to the ```BasetModel``` in ```website/static/js/comment.js```

# Side Effects
None I can think of!


icereval:
thanks!

PR #5394, opened 2016-04-05
[Ready to Review][OSF-5945] Feature/new logs audit

pattisdr:
Issue https://openscience.atlassian.net/browse/OSF-5945

# Purpose

Audits new logs on My Projects page.  For the most part, tries to match current logs.  As a general rule, files will provide links whether or not the resource has been deleted. (If it has been deleted, the link will return a Gone or Not Found).  Logs referencing nodes, however, will not provide links if the resource has been deleted.   

# Changes

Numerous small fixes to logs that appear in My Projects page - no change to project organizer logs.  Log phrasing changed, links added, etc.  For full list of changes see google doc link in Issue https://openscience.atlassian.net/browse/OSF-5945


icereval:
fyi, a new log type will be introduced w/ https://github.com/CenterForOpenScience/osf.io/pull/5025

pattisdr:
> fyi, a new log type will be introduced w/ #5025

Thanks - looks like Matt took care of adding this log action to `logActionsList.json`.  The only change I made to this was make the file checked in/checked out a link.

website/static/js/logTextParser.js
```diff
@@ -159,10 +174,12 @@ var LogPieces = {
     node: {
         view: function (ctrl, logObject) {
             var nodeObject = logObject.embeds.original_node;
+            var logAction = logObject.attributes.action;
```
hmoco: this doesn't seem to be used

website/static/js/logTextParser.js
```diff
@@ -173,11 +190,20 @@ var LogPieces = {
                 else if (nodeObject.data.attributes) {
                     return m('span', nodeObject.data.attributes.title);
                 }
-            } else {
+            }
+            // Original node has been deleted
+            else if (!paramIsReturned(nodeObject, logObject)) {
+                var deletedNode = logObject.attributes.params.params_node;
+                if (paramIsReturned(deletedNode, logObject)){
```
hmoco: Unless I'm incorrect, if this fails (returns false), this will return nothing, over the ` return m('span', 'a project'); ` in the else. Same is true for line 181.
``` 
function func(){
    if(false){} 
    else if (true){var text='text'}
    else { return 'yeppers'}
    return text
}
```
this will always return 'text'. This series of ` else if `s is fine if the only condition to return a value is at the top level, but if there are other conditions within it, once those fail, it will not move to the next ` else if ` or ` else `, which is intended. @sloria any thoughts?

hmoco: An alternative is simply removing the ` else ` on line 201.

website/static/js/logTextParser.js
```diff
@@ -242,26 +327,71 @@ var LogPieces = {
...
+                    if (nodeObject.data.links && nodeObject.data.attributes) {
+                        return nodeObject.data.links.html;
+                    }
+                }
+                return null;
+            };
+        },
         view: function (ctrl, logObject) {
-            return returnTextParams('title_new', 'a title', logObject);
+            var url = ctrl.returnLinkForPath();
```
hmoco: Why is this a function being called here and not just a variable with values ` null ` or `nodeObject.data.links.html` ?

website/static/js/logTextParser.js
```diff
@@ -242,26 +327,71 @@ var LogPieces = {
...
     // external identifiers on node
     identifiers: {
         view: function (ctrl, logObject) {
-            return returnTextParams('identifiers', 'identifier(s)', logObject);
+            external_ids = logObject.attributes.params.identifiers;
+            if (paramIsReturned(external_ids, logObject)) {
+                var doi = external_ids.doi;
+                var ark = external_ids.ark;
+                if (doi && ark) {
+                    return m('span', 'doi:' + doi + ' and ark:' + ark);
```
hmoco: Is there any use on returning just one or the other? Or is this behavior defined by the project log templates?

website/static/js/logTextParser.js
```diff
@@ -242,26 +327,71 @@ var LogPieces = {
...
+                    if (paramIsReturned(self.page_id, logObject)){
+                        return '/' + self.page_id;
+                    }
+                }
+                return null;
+            };
+        },
         view: function (ctrl, logObject) {
-            return returnTextParams('page', 'a title', logObject);
+            var url = ctrl.returnLinkForPath();
```
hmoco: same question as above

website/static/js/logTextParser.js
```diff
@@ -285,7 +415,11 @@ var LogPieces = {
         view: function (ctrl, logObject) {
             var source = logObject.attributes.params.source;
             if(paramIsReturned(source, logObject)){
-                return m('span', [source.materialized, ' in ', source.addon]);
+                var sourceMaterialized = source.materialized;
```
hmoco: why not just ` var sourceMaterialized = stripBackslash(source.materialized); `, since stripBacklash already checks if the value ends with a backlash?

website/static/js/project-organizer.js
```diff
@@ -53,7 +53,10 @@ function _poTitleColumn(item) {
  * @private
  */
 function _poContributors(item) {
-    var contributorList = item.data.embeds.contributors.data;
+    var contributorList = [];
+    if (item.data && item.data.embeds && item.data.embeds.contributors) {
+        contributorList = item.data.embeds.contributors.data;
+    }
     if(!contributorList){
```
hmoco: due to your changes, this will never return `''`, since `[]` evaluates to ` true `. This is taken of by line 63:
```
if (contributorList.length === 0) {
         return '';
     }
```, so I think this `if` can be removed.

hmoco:
@pattisdr, pass complete. Lots of things you caught that I missed when I was working on logs, so thank you!

pattisdr:
Closed in favor of https://github.com/CenterForOpenScience/osf.io/pull/5513

pattisdr:
@hmoco, responded to your CR.  Has been made a PR to 68.

PR #5395, opened 2016-04-05
[rfr][fix] Fixing esi embeds for drf 3.3.3

cwisecarver:
As far as I can tell, I'll look more in the morning, this fixes ESI embeds for DRF 3.3.3. It looks like the structure of related links changed in the way they were being passed to `to_esi_representation`. After a few key checks and digging deeper into the object I got the correct href and passed it to thing that generates the ESI url. The key checks don't seem very pythonic and I'll try to clean them up in the morning but I think this will fix it. The output on local matches the output on staging and the error I was getting on `/myprojects/` locally is gone.

The bit in `api/base/views.py` just solidifies that the API really is trying to serve JSON on this particular request instead of just checking the global allowed content types. h/t @chrisseto 

@icereval, @brianjgeiger, @Ghalko

### DIFF
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -513,8 +513,8 @@ class RelationshipField(ser.HyperlinkedIdentityField):
 
     def to_esi_representation(self, value, envelope='data'):
         relationships = self.to_representation(value)
-        if relationships is not None and 'related' in relationships.keys():
-            href = relationships['related']
+        if relationships is not None and 'links' in relationships.keys() and 'related' in relationships['links'].keys():
+            href = relationships['links']['related']['href']
             if href and not href == '{}':
                 if self.always_embed:
                     envelope = 'data'
@@ -1007,7 +1007,6 @@ class JSONAPISerializer(ser.Serializer):
                         else:
                             try:
                                 # If a field has an empty representation, it should not be embedded.
-                                field.to_representation(attribute)
                                 result = self.context['embed'][field.field_name](obj)
                             except SkipField:
                                 result = None
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -513,8 +513,11 @@ class RelationshipField(ser.HyperlinkedIdentityField):
 
     def to_esi_representation(self, value, envelope='data'):
         relationships = self.to_representation(value)
-        if relationships is not None and 'links' in relationships.keys() and 'related' in relationships['links'].keys():
+        try:
             href = relationships['links']['related']['href']
+        except KeyError:
+            raise SkipField
+        else:
             if href and not href == '{}':
                 if self.always_embed:
                     envelope = 'data'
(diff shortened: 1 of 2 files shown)


api/base/serializers.py
```diff
@@ -513,8 +513,8 @@ def get_url(self, obj, view_name, request, format):
 
     def to_esi_representation(self, value, envelope='data'):
         relationships = self.to_representation(value)
-        if relationships is not None and 'related' in relationships.keys():
-            href = relationships['related']
+        if relationships is not None and 'links' in relationships.keys() and 'related' in relationships['links'].keys():
```
brianjgeiger: Just verifying that, for our use case, there's nothing alternate paths of this if statement that interest us (such as there is a link in the relationship.keys() but it's not a related link). I'm guessing no, because it's the esi representation and that doesn't care about self links or whatever, but just making sure.

brianjgeiger:
Looks good to me, and tests passing help. I'm willing to get this on staging so we can verify it works with the rest of the services as well if there are no objections.

api/base/serializers.py
cwisecarver: The self link should end up coming back in the request for the relationships.links[].related.href response.

api/base/views.py
```diff
@@ -129,7 +129,7 @@ def get_serializer_context(self):
         context.update({
             'enable_esi': (
                 is_truthy(self.request.query_params.get('esi', django_settings.ENABLE_ESI)) and
-                self.request.accepted_media_type in django_settings.ESI_MEDIA_TYPES
+                self.request.accepted_renderer.media_type in django_settings.ESI_MEDIA_TYPES
```
cwisecarver: This, BTW, is why the varnish part of the travis tests were passing. Since self.request.accepted_media_type decided to stop existing sometime between 3.1.1 and 3.3.3 ESI was being disabled by the varnish tests even though it was being enabled through the setting and the correct content type. So the varnish tests were comparing data where ESI was false against data where ESI was false and passing. @brianjgeiger @icereval @Ghalko

Ghalko: @cwisecarver So you need a test to make sure that doesn't happen again.

cwisecarver: I can't think of any way to write a test to accomplish this. If DRF changes their API ... I don't know how we can test against every potential change they could make. 

The point I was hoping to get across was that "tests passing" isn't a stamp of approval that the actual underlying functionality works. As we use more packages and frameworks that we didn't write we're going to have to pay more attention to functional, integration, and regression testing.

Unit testing is great for units of code but it won't find a integration problem that occurs in the final output (html/json) that is built using several different independent frameworks.

brianjgeiger: On the plus side, even if this had been an obscure error and not one that hit you in the face immediately on login, it would have (and was) caught by integration tests. But yes, more of these, and more maintenance on them, would be awesome.

PR #5396, opened 2016-04-06
[hotfix] Add missing '/' to start of materialized path

samanehsan:
Fixes issue where the materialized paths returned from waterbutler when renaming a file in dropbox,
github, googledrive and s3 did not start with '`/'` and caused errors when renaming the file again.
Because the materialized path did not match the one used to look up the file, the files received new guids.

### DIFF
--- a/website/project/views/comment.py
+++ b/website/project/views/comment.py
@@ -59,6 +59,10 @@ def create_new_file(obj, source, destination, destination_node):
         new_file = FileNode.resolve_class(destination['provider'], FileNode.FILE).get_or_create(destination_node, destination['path'])
         if destination['provider'] != 'osfstorage':
             new_file.update(revision=None, data=data)
+            # TODO: Remove when materialized paths are fixed in the payload returned from waterbutler
+            if not new_file.materialized_path.startswith('/'):
+                new_file.materialized_path = '/' + new_file.materialized_path
+                new_file.save()
     else:
         new_file = find_and_create_file_from_metadata(destination.get('children', []), source, destination, destination_node, obj)
         if not new_file:
@@ -68,7 +72,7 @@ def create_new_file(obj, source, destination, destination_node):
                 new_path = obj.referent.materialized_path.replace(source['materialized'], destination['materialized'])
             new_file = FileNode.resolve_class(destination['provider'], FileNode.FILE).get_or_create(destination_node, new_path)
             new_file.name = new_path.split('/')[-1]
-            new_file.materialized_path = new_path
+            new_file.materialized_path = '/' + new_path
             new_file.save()
     return new_file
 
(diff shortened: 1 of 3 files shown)


icereval:
@felliott @chrisseto , please confirm you are onboard with this temporary fix?

https://github.com/CenterForOpenScience/osf.io/pull/5396/files#diff-df7300cf6eb9b0215bd17f29e54b259eR62

felliott:
I am.  @samanehsan and I discussed it yesterday.

website/project/views/comment.py
```diff
@@ -68,7 +72,7 @@ def create_new_file(obj, source, destination, destination_node):
                 new_path = obj.referent.materialized_path.replace(source['materialized'], destination['materialized'])
             new_file = FileNode.resolve_class(destination['provider'], FileNode.FILE).get_or_create(destination_node, new_path)
             new_file.name = new_path.split('/')[-1]
-            new_file.materialized_path = new_path
+            new_file.materialized_path = '/' + new_path
```
chrisseto: Should this have an if guard in it as well?

samanehsan: Good point -- I'll make that change!

website/project/views/comment.py
```diff
@@ -59,6 +59,10 @@ def create_new_file(obj, source, destination, destination_node):
         new_file = FileNode.resolve_class(destination['provider'], FileNode.FILE).get_or_create(destination_node, destination['path'])
         if destination['provider'] != 'osfstorage':
             new_file.update(revision=None, data=data)
+            # TODO: Remove when materialized paths are fixed in the payload returned from waterbutler
+            if not new_file.materialized_path.startswith('/'):
```
chrisseto: Could this check just be done once at the beginning of the method?

samanehsan: Probably...I'll take a look!

chrisseto:
lg;mt

icereval:
@brianjgeiger input plz

icereval:
migration? :dancers:

samanehsan:
@icereval, should I add a migration script that goes through all `StoredFileNode` materialized paths and makes sure they begin with a `'/'`?

brianjgeiger:
A change to materialized path will potentially break scripts for anyone who uses that field in their scripts. We will need to inform API users when this happens with a little time.

felliott:
Just to clarify, it's not the WB user responses that are missing the slash, it's the logging payloads that WB is posting back to the OSF.   Users should only encounter this when trying to parse the contents of a NodeLogs v1 query.  We're already inconsistent there.

A migration is probably a good idea, though. Addon StoredFileNodes created since 0.67 may or may not have the leading slash.

brianjgeiger:
Thanks, Fitz. Then I have no API concerns.

icereval:
@samanehsan, please add a migration script when time allows, thanks!

samanehsan:
I added in the migration script @icereval! 😎

PR #5307, opened 2016-03-25
Remove backrefs for OSF-[5896, 5898, 5899, 5900, 5901, 5906, 5907, 5909]

abought:
Refs various, general parent is https://openscience.atlassian.net/browse/OSF-5674 

## Purpose
We are attempting to reduce the number of unused backrefs on our models, as backrefs create side effects (record locking) that are not ideal.

This PR is intended to encompass the low-hanging fruit: simple backrefs that can be easily replaced with small queries.

## Summary of changes
- Remove various unused backrefs (see diff)
- Replace usages of these backrefs with direct database queries

## Side effects/ testing notes
If all goes well, there will be no user-facing side effects, and OSF functionality will not change. Below is a list of features that could be affected by these changes.

- Make sure the list of active registrations appears correctly (eg for registrations tab on project pages)
- Watch (or unwatch) a project. Reload the page and make sure the number of watchers updates correctly. Make sure that you correctly receive updates about projects you are watching.
- Do something that would trigger a notification email (transactional or digest) on a project. Make sure that you receive the email correctly with no errors.
- Fork a project. Ensure the list of forks displays correctly, and the fork button updates with the correct fork count.
- Create a new project, using another project as a template. (fork button--> "copy project structure"). Verify that the "templated from" counter in the modal updates to show the correct count. 
- Node links should still work

There may be some issues with deletion as well. We don't really delete many things (users and nodes just get hidden via a flag), but try things like unwatching a node, and see if that gives errors.

## Subtasks
https://openscience.atlassian.net/browse/OSF-5896
https://openscience.atlassian.net/browse/OSF-5896
https://openscience.atlassian.net/browse/OSF-5898
https://openscience.atlassian.net/browse/OSF-5899
https://openscience.atlassian.net/browse/OSF-5900
https://openscience.atlassian.net/browse/OSF-5901
https://openscience.atlassian.net/browse/OSF-5906 (badges, unused)
https://openscience.atlassian.net/browse/OSF-5907 (badges, unused)
https://openscience.atlassian.net/browse/OSF-5909

PR #5398, opened 2016-04-06
[hold] Small speed ups for the view_node endpoint

chrisseto:


website/addons/base/__init__.py
```diff
@@ -160,22 +160,17 @@ def _include_to_static(self, include):
 
     @property
     def icon(self):
-
         try:
             return self._icon
-        except:
+        except AttributeError:
```
brianjgeiger: :+1:

website/addons/base/__init__.py
```diff
@@ -160,22 +160,17 @@ def _include_to_static(self, include):
...
             static_path = os.path.join('website', 'addons', self.short_name, 'static')
             static_files = glob.glob(os.path.join(static_path, 'comicon.*'))
-            image_files = [
+            return next((
```
brianjgeiger: Okay, general idea of this is that the old version made a list and was expecting no more than one item and returns that item or None. So you've replaced it with a generator that either returns the one item or None. Yes?

website/project/model.py
```diff
@@ -160,7 +160,7 @@ class Comment(GuidStoredObject, SpamMixin):
 
     user = fields.ForeignField('user', required=True)
     # the node that the comment belongs to
-    node = fields.ForeignField('node', required=True)
+    node = fields.ForeignField('node', required=True, index=True)
```
brianjgeiger: You and your fancy wanting to index fields.

website/project/model.py
```diff
@@ -1004,6 +1004,11 @@ def ids_above(self):
         return {p._id for p in parents}
 
     @property
+    def has_children(self):
+        if not self.nodes:
```
brianjgeiger: This doesn't look to return anything in the true case and is also very negative. If there's something fancier going on here, just add a comment describing it.

website/project/model.py
```diff
@@ -1091,8 +1096,11 @@ def is_registration_of(self, other):
     @property
     def forks(self):
         """List of forks of this node"""
-        return list(self.node__forked.find(Q('is_deleted', 'eq', False) &
-                                           Q('is_registration', 'ne', True)))
+        return Node.find(
+            Q('is_deleted', 'eq', False) &
+            Q('is_registration', 'ne', True) &
+            Q('forked_from', 'eq', self._id)
```
brianjgeiger: I would have expected this to have been handled already. Was this missing from a backref removal PR or is there an open PR that's going to have a merge conflict?

website/addons/base/__init__.py
chrisseto: Yes

website/project/model.py
chrisseto: Whoops....

website/project/model.py
```diff
@@ -1786,19 +1789,27 @@ def pointing_at(self, pointed_node_id):
...
+            x._backrefs['parent']['node']['nodes'][0]: x._id
+            for x in Pointer.find(Q('_id', 'in', zip(*self.pointed)[0]))
+        }
+
+        nodes = {
+            pointers[n._id]: n
+            for n in
+            Node.find(
+                Q('_id', 'in', pointers.keys()) &
+                Q('is_deleted', 'ne', not deleted) &
```
cwisecarver: Would you get the same result/performance if you changed to?:
```
    Q('is_deleted', 'eq', deleted) &
    Q('is_collection', 'eq', folders)
```
just to make it easier for us mere mortals to not have to noodle through the double-negative?

website/project/model.py
```diff
@@ -1690,18 +1698,14 @@ def rm_pointer(self, pointer, auth):
...
+        return [x[0] for x in self.to_storage()['nodes'] if x[1] == 'node']
 
     @property
     def nodes_primary(self):
-        return [
-            node
-            for node in self.nodes
-            if node.primary
-        ]
+        ids = [x[0] for x in self.to_storage()['nodes'] if x[1] == 'node']
```
brianjgeiger: Is this faster because it avoids having to load each of the nodes in the list comprehension?

website/project/model.py
```diff
@@ -1786,19 +1789,27 @@ def pointing_at(self, pointed_node_id):
...
-            if not folders and pointer_node.is_collection:
-                continue
-            if not deleted and pointer_node.is_deleted:
-                continue
-            if resolve:
-                ret.append(pointer_node)
-            else:
-                ret.append(each)
-        return ret
+    def get_points(self, folders=False, deleted=False):
```
brianjgeiger: How much faster is this? It reduces readability significantly, so we'll want to it be measurable.

website/project/model.py
```diff
@@ -1786,19 +1789,27 @@ def pointing_at(self, pointed_node_id):
...
+        if not self.pointed:
+            return []
+
+        # Note: Due to GUIDs being unique there will never be an overlap between Nodes and pointers
+        pointers = {
+            x._backrefs['parent']['node']['nodes'][0]: x._id
```
brianjgeiger: You will want to check this versus develop, and this may explain the other backref question above. https://github.com/CenterForOpenScience/osf.io/pull/5307 has cleared away almost all of the backrefs.

website/util/rubeus.py
```diff
@@ -266,21 +267,14 @@ def collect_addon_js(node, visited=None, filename='files.js', config_entry='file
     :return list: List of JavaScript include paths
 
     """
-    # NOTE: must coerce to list so it is JSON-serializable
-    visited = visited or []
-    visited.append(node._id)
-    js = set()
-    for addon in node.get_addons():
+    js = []
+    for addon_config in settings.ADDONS_AVAILABLE_DICT.values():
```
brianjgeiger: I'm not super-familiar with the addon system. Is is definitely equivalent to `node.get_addons()`?

tests/test_rubeus.py
```diff
@@ -353,30 +353,3 @@ def test_serialize_node(self):
                 'fetch': None,
             },
         )
-
-    def test_collect_js_recursive(self):
```
brianjgeiger: What changes lead to these two tests being removable?

brianjgeiger:
Pass complete. Tests failing. I would feel a lot more comfortable if this were to develop or release/0.67.0 than to master considering other things that are also happening. And it would be good to get this run through QA when the time comes, considering.

website/project/model.py
chrisseto: It has to be `ne` because AFAIK and as the rest of the code indicates `is_deleted` and `is_collection` may not exist on non-collection or non-deleted nodes.

sloria:
@chrisseto What's the status of this PR?

chrisseto:
@sloria kind of broken. There are a couple choices that need to be made still. I just haven't gotten the time to look at it.

PR #5400, opened 2016-04-06
[hold] Resolve twitter handles [OSF-3617]

kch8qx:
# Purpose
Currently, when editing social fields, users are able to enter Twitter handles containing '@' (e.g. '@example'). When someone then clicks on their Twitter handle from their profile, they are sent to their Twitter page (e.g. https://twitter.com/@example). This is not a problem when the handle contains only one leading '@', since pages like the above work correctly on Twitter.
The problem occurs when a user enters a Twitter handle containing more than one leading '@' (e.g. '@@@example'). In this case, this will send whoever clicks on the handle to a bad address like https://twitter.com/@@@example.

# Changes
This PR removes all instances of leading '@' from user-entered Twitter handles. Twitter profiles accessed by both the user profile and the search page will lead to the correct Twitter page without any leading '@' (e.g. https://twitter.com/example for a user-entered handle of '@@@example'). 
This PR also includes a migration script that any leading '@' from existing user Twitter handles, and a test of this migration script.

[OSF-3617]


mfraezz:
via Jira ticket:
>Twitter handles do not always resolve properly. Example: osf.io/@BrianNosek 
[...]
They will resolve properly if you include the extra @ symbol. 
Example: osf.io/@@BrianNosek [...]
I believe at some point, we required the @ symbol in user profile.

I'm unable to reproduce this issue as described. On `develop`, twitter handles can be entered with or without an `@` and resolved correctly when clicked. Two or more `@`'s failed to resolve.

Regardless, these changes are good going forward, but it doesn't solve the `"existing users included @@ at some point in the past"` issue. A [migration script](http://cosdev.readthedocs.org/en/latest/osf/migrations.html) is needed for that. I have a script template locally, here's that but with an appropriate query for this case:

```python
import logging
import sys

from modularodm import Q

from framework.transactions.context import TokuTransaction
from website.models import User
from scripts import utils as script_utils

def get_targets():
    return User.find(Q('social.twitter', 'contains', '@'))

def migrate(targets, dry_run=True):
    # iterate over targets
    # log things

    if dry_run:
        raise RuntimeError('Dry run, transaction rolled back.')

def main():
    dry_run = False
    if '--dry' in sys.argv:
        dry_run = True
    if not dry_run:
        script_utils.add_file_logger(logger, __file__)
    init_app(set_backends=True, routes=False)
    with TokuTransaction():
        migrate(targets=get_targets(), dry_run=dry_run)

if __name__ == "__main__":
    main()
```

Also, please follow our standard [PR description template](http://cosdev.readthedocs.org/en/latest/process/pull_requests.html#sample-pull-request-description)

brianjgeiger:
I love that you have tests on your migration. Let's see if we can add in as many crazy tests as we can think of (migrations are very tricky on production data). Try to think of situations like "users who don't have twitter handles", "users who have many more `@`s in their handle", "users who have `@`s in the middle of their handle", "users who have a perfectly valid twitter handle", "users who have a unicode twitter handle", and so on. Whatever you can think of that might cause problems, go ahead and check to make sure your script does what you would expect it to.

scripts/migration/migrate_twitter_handles.py
```diff
@@ -0,0 +1,66 @@
...
+    users = targets
+    logging.info(len(users))
+    for user in users:
+        twitter = user.social['twitter'].replace("@", "")
+        logger.info('Setting `social.twitter` field of user {0} to {1}'.format(user._id, twitter))
+        if not dry_run:
+            user.social['twitter'] = twitter
+            user.save()
+    '''
+    if dry_run:
```
brianjgeiger: Go ahead and removed commented out lines. Also, you should comment with `#` rather than `'''`.

scripts/migration/migrate_twitter_handles.py
```diff
@@ -0,0 +1,66 @@
...
+        raise RuntimeError('Dry run, transaction rolled back.')
+    '''
+
+def main():
+    dry_run = False
+    if '--dry' in sys.argv:
+        dry_run = True
+    if not dry_run:
```
brianjgeiger: Not a big deal, but you could change this to `else:` instead, since it's right there.

brianjgeiger:
For the issue description, it's nice when we can show screenshots of what changes so QA can test. Could you do a couple of sets of screenshots, one with the current functionality, and one with what will be different with your changes?

brianjgeiger:
Pass done. 🐧

kch8qx:
Adding screenshots to clarify before and after behavior!

## Before:
I enter a Twitter handle in 'Profile Settings' that I know won't resolve:
screen shot 2016-04-27 at 4 22 30 pm
I click 'Save'.
I go to my profile, and see that my profile's Twitter handle is exactly the same as I entered it:
screen shot 2016-04-27 at 4 26 29 pm
When I click on the Twitter link (https://twitter.com/@@@khewtart), I get:
screen shot 2016-04-27 at 4 27 09 pm

## After:
I enter my Twitter handle exactly the same as before:
screen shot 2016-04-27 at 4 22 30 pm
I click 'Save', go to my profile, and see that my profile's Twitter handle has been resolved:
screen shot 2016-04-27 at 4 22 42 pm
When I click on this link, it goes to my actual Twitter page, https://twitter.com/khewtart and no 404 error

The above process shows how clicking 'Save' when editing profile settings and social fields automatically resolves Twitter handles. The migration script covers those cases where users have already entered Twitter handles that can't be resolved.

mfraezz:
The script fails on unicode characters in twitter handles, but none of the targets that would be hit on prod contain unicode (and I'm fairly certain that Twitter doesn't support non-ascii characters in handles).

Otherwise, LGTM. Dropping this into RTM 👓

scripts/migration/migrate_twitter_handles.py
```diff
@@ -0,0 +1,110 @@
...
+        script_utils.add_file_logger(logger, __file__)
+    init_app(set_backends=True, routes=False)
+    with TokuTransaction():
+        migrate(targets=get_targets(), dry_run=dry_run)
+
+if __name__ == "__main__":
+    main()
+
+class TestMigrateTwitterHandles(OsfTestCase):
```
sloria: Tests need to go in `scripts/tests`; otherwise the script will fail due to import errors (we don't install test requirements in production and staging deploys).

scripts/migration/migrate_twitter_handles.py
```diff
@@ -0,0 +1,110 @@
...
+        self.user3 = UserFactory()
+        self.user3.social['twitter'] = '@@user3'
+        self.user3.save()
+        # User 4 has many leading '@' signs, should be changed to 'user4' after migration
+        self.user4 = UserFactory()
+        self.user4.social['twitter'] = '@@@@@@@@@@@@@@@@@user4'
+        self.user4.save()
+        # User 5 has '@' signs in the middle of their handle, should be changed to 'user5' after migration
+        self.user5 = UserFactory()
+        self.user5.social['twitter'] = 'user@@@@@@@5'
```
sloria: Is there a reason to remove non-leading "@"s? Are you certain that these aren't valid. Seems safer to only strip leading "@"'s anyway.

scripts/migration/migrate_twitter_handles.py
```diff
@@ -0,0 +1,110 @@
...
+        self.user5.save()
+        # User 6 has '@' signs at the end of their handle, should be changed to 'user5' after migration
+        self.user6 = UserFactory()
+        self.user6.social['twitter'] = 'user6@@@@@@'
```
sloria: See https://github.com/CenterForOpenScience/osf.io/pull/5400/files#r62684339

scripts/migration/migrate_twitter_handles.py
brianjgeiger: `@`s are not valid in twitter handles. I did some testing way back when.

website/static/js/profile.js
```diff
@@ -673,6 +673,8 @@ SocialViewModel.prototype.serialize = function() {
             return value;
         }
     );
+    var twitter = serializedData.twitter;
+    serializedData.twitter = twitter.replace(/@/g, '');
```
sloria: It would be good to avoid special-casing in this method. Instead, could you modify the regex in `socialRules.twitter` to exclude leading `@` signs?

sloria:
Pass finished.

scripts/migration/migrate_twitter_handles.py
sloria: OK, then disregard my comment.

kch8qx:
@sloria my original plan was to try to modify socialRules.twitter's regex to disallow `@` signs, but I don't have a lot of experience with regex and couldn't get it to work. I'll go back now and try and make the change there instead of in the serializedData function. 

Besides that change and moving the migration tests to the `scripts/tests`, that's all I need to fix correct?

sloria:
@kch8qx Yep, just those two minor changes, and this should be good to go.

kch8qx:
Note: migration script test has been moved to scripts/tests. 

To do: use regex to disallow '@' in socialRules.twitter

sloria:
I'm going to hold off on merging this for now. Updating this many users' profile info will result in many nodes getting updated in elastic search, which will queue many jobs in celery. Let's wait until we have multiple celery queues in production before this is merged.

website/static/js/profile.js
```diff
@@ -673,6 +673,8 @@ SocialViewModel.prototype.serialize = function() {
             return value;
         }
     );
+    var twitter = serializedData.twitter;
+    //serializedData.twitter = twitter.replace(/@/g, '');
```
whit537: Looks like we can take out these two lines, ya?

whit537:
##### Punchlist

- [ ] back out no-op changes—https://github.com/CenterForOpenScience/osf.io/pull/5400#discussion_r68682014

whit537:
As I read this PR, we don't need any logic changes after all, we're *just* doing the migration. Ya?

brianjgeiger:
It's Autumn Cleaning time, so I'm closing this PR. Feel free to re-open when

PR #5401, opened 2016-04-06
[Feature] admin user refactor

Ghalko:
## Proposal
This is an accumulation of many small items for user refactor similar to #5370 

ADM-70

## Repercussions
Many simplifications thanks to Django 1.9.


admin/users/views.py
```diff
@@ -13,36 +15,104 @@
...
+            if flag is not None:
+                update_admin_log(
+                    user_id=self.request.user.id,
+                    object_id=user.pk,
+                    object_repr='User',
+                    message=message,
+                    action_flag=flag
+                )
+        except AttributeError:
+            return page_not_found(
```
mfraezz: This doesn't need to happen now, but given how many times this exact `return` happens, it might be nice to add a helper function to do so.

mfraezz:
LGTM. Tested to verify that both disabling/enabling a user and removing two factor still worked. Dropping this in RTM

PR #5404, opened 2016-04-06
[hotfix] remove unused hgrid css from meetings page

felliott:
hgrid was removed from bower.json by c07269d0f2 on Feb. 24.  We no
longer use or serve it and it 404s.


icereval:
nice!

PR #5405, opened 2016-04-06
Do not block UI when registration fails [OSF-5967]

abought:
Refs, but does not completely fix, https://openscience.atlassian.net/browse/OSF-5809

Fixes https://openscience.atlassian.net/browse/OSF-5967

## Purpose
When a registration of a draft fails, the UI does not unblock, and the user is left without a course of action.

Instead, the UI should be unblocked correctly, and the user should be taken back to the registration page after dismissing the modal.

## Testing notes
See the linked PR for an example test case of a failing registration. (you can fork that project if you don't directly have access). Be warned that the rare failure case in question takes ~1min of waiting before it fails.

For local machine code review purposes, you can force a failure much faster by modifying `projects.project.views.drafts.register_draft_registration` to return a 400 immediately.

### DIFF
--- a/website/static/js/registrationUtils.js
+++ b/website/static/js/registrationUtils.js
@@ -692,7 +692,12 @@ Draft.prototype.register = function(url, data) {
             bootbox.alert({
                 title: 'Registration failed',
                 message: language.registerFail,
-                callback: $osf.unblock
+                callback: function() {
+                    $osf.unblock();
+                    if (self.urls.registrations) {
+                        window.location.assign(self.urls.registrations)
+                    }
+                }
             });
         })
         .always(function() {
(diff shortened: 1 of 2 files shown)


mfraezz:
LGTM. Dropping this in RTM :star:

website/static/js/registrationUtils.js
```diff
@@ -693,9 +689,20 @@ Draft.prototype.register = function(url, data) {
...
+                callback: function() {
+                    $osf.unblock();
+                    if (self.urls.registrations) {
+                        window.location.assign(self.urls.registrations);
+                    }
+                }
+            });
         })
-        .always($osf.unblock);
+        .always(function() {
```
icereval: not sure why there needs to be another anon function call here, a direct call out to a promise response is 100% ok.

icereval:
lgtm

PR #5408, opened 2016-04-07
[feature] speedup conferences submission view

cwisecarver:
The in query will speed it up a little. There has to be a way to stop looping over all the conferences but I haven't been able to figure it out so far.

### DIFF
--- a/scripts/indices.py
+++ b/scripts/indices.py
@@ -26,3 +26,10 @@ db['node'].create_index([
     ('is_registration', ASCENDING),
     ('contributors', ASCENDING),
 ])
+
+db['node'].create_index([
+    ('tags.$', ASCENDING),
+    ('is_public', ASCENDING),
+    ('is_deleted', ASCENDING),
+    ('institution_id', ASCENDING),
+])
--- a/scripts/indices.py
+++ b/scripts/indices.py
@@ -28,7 +28,7 @@ db['node'].create_index([
 ])
 
 db['node'].create_index([
-    ('tags.$', ASCENDING),
+    ('tags', ASCENDING),
     ('is_public', ASCENDING),
     ('is_deleted', ASCENDING),
     ('institution_id', ASCENDING),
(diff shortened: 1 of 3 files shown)


icereval:
Can you add a manual index:
https://github.com/CenterForOpenScience/osf.io/blob/develop/scripts/indices.py

I believe you found index's made w/ `tags.$` don't actually work, could probably comment it out or remove it w/ a note to the manual creation file.
https://github.com/CenterForOpenScience/osf.io/blob/develop/website/project/model.py#L680

cwisecarver:
@icereval Added the index and removed it from MODM.

scripts/indices.py
```diff
@@ -26,3 +26,10 @@
     ('is_registration', ASCENDING),
     ('contributors', ASCENDING),
 ])
+
+db['node'].create_index([
+    ('tags.$', ASCENDING),
```
icereval: don't need the `.$`

website/conferences/views.py
```diff
@@ -230,13 +230,19 @@ def conference_submissions(**kwargs):
...
         projects = set()
-        for tag in Tag.find(Q('lower', 'eq', conf.endpoint.lower())):
-            for node in Node.find(Q('tags', 'eq', tag._id) & Q('is_public', 'eq', True) & Q('is_deleted', 'eq', False)):
-                projects.add(node)
+
+        tags = Tag.find(Q('lower', 'eq', conf.endpoint.lower())).get_keys()
+        nodes = Node.find(
+            Q('tags', 'in', tags) &
+            Q('is_public', 'eq', True) &
+            Q('is_deleted', 'eq', False)
```
chrisseto: `is_deleted` queries should `'ne', True`. AFAIK we still have some nodes floating around without `is_deleted` properly set.

cwisecarver: I'm happy to change it @chrisseto but that would be changing original functionality past just optimizing the queries. It was `is_deleted eq False` prior. She was dead when I got there...

cwisecarver:
@chrisseto All dun

PR #5409, opened 2016-04-07
Registration

KAsante95:
# Purpose
Display the correct files link in the registration serializer.

Issue:
https://openscience.atlassian.net/browse/OSF-5892

# Code

Because both RegistrationRegistrationsList and RegistrationProvidersList had the same view name, 'registration-providers', I changed the view name for RegistrationRegistrationsList to 'registration-registrations'.


sloria:
@KAsante95 Does this bug exist on master? If so, can you resend this as a hotfix?

PR #5411, opened 2016-04-07
[feature] Rate limit password reset view [OSF-5960]

brianjgeiger:
Purpose
-----------
Fix https://openscience.atlassian.net/browse/OSF-5960 which also fixes [SEC-5]

Currently, we have broad rate limiting over everything in the OSF, but someone reported a vulnerability with the forgot password field that would let malicious users email someone with a "reset password" link thousands of times in a minute. This adds a rate limit to that field to prevent it from happening nearly as frequently

Changes
------------
1. Add a test for the condition
2. Add a field to the user model to track when the last time a request was made
3. Add a setting for how frequently we'll allow people to request a new password (currently set to 30 seconds, which seems decent to me)
4. Add a new webtest function to ensure that html wasn't in a view, not just that it was in the view

Side effects
---------------
The malicious user will be able to lock the user object and write to the DB with each forgotten password request. This is less than ideal, but was discussed with Michael. In the long term, we'll add some better rate limiting at other, more appropriate places, but we want to make sure users are not harmed and we don't get marked as spam in the meantime.

### DIFF
--- a/framework/auth/core.py
+++ b/framework/auth/core.py
@@ -291,6 +291,8 @@ class User(GuidStoredObject, AddonModelMixin):
     # verification key used for resetting password
     verification_key = fields.StringField()
 
+    forgot_password_last_post = fields.DateTimeField()
+
     # confirmed emails
     #   emails should be stripped of whitespace and lower-cased before appending
     # TODO: Add validator to ensure an email address only exists once across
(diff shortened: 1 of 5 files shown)


mfraezz:
Build failed, otherwise LGTM. Discussed side effect and limit interval in person, both are acceptable. DB locking would already occur from this abuse, so it's nothing new. Dropping in RTM whenever TCI passes :fireworks:

brianjgeiger:
Tests are passing, @mfraezz

PR #5415, opened 2016-04-07
Speed up User PUT and PATCH Methods by modifying User properties [OSF-5338]

erinspace:
https://openscience.atlassian.net/browse/OSF-5338

# Purpose
```PUT``` and ```PATCH``` methods are very slow on the OSF API at the moment. @brianjgeiger noticed that was because the ```contributor_to``` method was a list comprehension relying on a previous property and not a faster ModularODM Query. This PR changes the ```contributor_to``` property to return a ModularODM query with the appropriate nodes.

# Changes 
- Move the  ```contributed``` property to be closer to the others
- Update the ```contributor_to``` property to return a ModularODM queryset instead of relying on the ```contributed``` property

# Side Effects
- The places that use those properties right now don't seem to be affected by this change, but these properties are now returning query sets instead of acting as generators so I might have missed something....

### DIFF
--- a/framework/auth/core.py
+++ b/framework/auth/core.py
@@ -1074,12 +1074,11 @@ class User(GuidStoredObject, AddonModelMixin):
 
     @property
     def contributor_to(self):
-        return (
-            node for node in self.contributed
-            if not (
-                node.is_deleted
-                or node.is_bookmark_collection
-            )
+        from website.project.model import Node
+        return Node.find(
+            Q('contributors', 'eq', self._id) &
+            Q('is_deleted', 'eq', False) &
+            Q('is_bookmark_collection', 'eq', False)
         )
 
     @property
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: framework/auth/core.py [hunk 1]
```diff
@@ -1073,13 +1068,17 @@ def profile_url(self):
         return '/{}/'.format(self._id)
 
     @property
+    def contributed(self):
+        from website.project.model import Node
```
brianjgeiger: Circular import?

brianjgeiger:
We'll need visible_contributor_to() as well.

Also, I presume these are covered by tests, yes? Could you verify that and comment or make new tests if, by chance, these are untested? They don't necessarily need to be directly tested, but at least a view test that calls these methods eventually.

brianjgeiger:
Oh, and this is for 0.67.0, so please target the release branch instead of develop. Sorry for not mentioning earlier.

erinspace:
Closing this PR to open on a new branch against the release!

PR #5243, opened 2016-03-21
[ko-2] Remove ko template syntax from project pages

abought:
Refs https://openscience.atlassian.net/browse/OSF-5889
Extension of #5216 , #5224

## Purpose
Phase out usage of knockout punches template helpers due to reported incompatibilities.

This is the second of three planned PRs. This covers addon widgets, contributor management, (lots of) nodelogs, and project dashboard items.

## Task list
- [x] `{{ expression }}` should be replaced with `<!--ko text:expression--><!--/ko-->` or some equivalent
- [x] Replace text filters (`fit`) with alternative JS expressions
- [x] Remove usages of namespaced dynamic bindings (see the docs): https://mbest.github.io/knockout.punches/
- [x] Convert pages that rely on wrapped event callbacks (by default, review `click`, `submit`, and `event` bindings)
- [x] Remove usages of `ko.punches.enableAll();`

## Testing notes
If pages do not start showing the new log templates as expected, consider deleting `_log_templates.mako` (the underscore-prefied file is automatically generated, usually when the OSF server is manually restarted). If old template cache hangs around, pages might not correctly reflect new syntax.

### List of affected pages / sections
- Project dashboard
  - [x] Node logs for ALL addons (I did not check all these exhaustively)
    - Box
    - Dataverse
    - Dropbox
    - Figshare
    - Forward
    - Github
    - Googledrive
    - Mendeley
    - OSF storage
    - S3
    - Zotero
  - [x] Dataverse dashboard widget
  - [x] Forward link dashboard widget
  - [x] Contributors list
  - Survey of miscellaneous buttons and content
- [x] File and project comments: Read, unread, and deleted comments.
- [x] Project contributors page, esp the add and remove contributors modals
- [x] Project "registrations" page (top level)

### Things to check
- Links should work
- Node log entries should not be missing fields (across all addons and osfstorage)
- Dataverse, zotero, and link widgets on project dashboard pages should render correctly
- Any text that used to be present should still be present (no fields are missing)
- Long project descriptions (>500 chars) are still truncated
- Should be able to click all buttons on the affected pages, and buttons should work as intended (no console errors)

PR #5416, opened 2016-04-07
[hold for 0.68] Remove ko punches usages throughout site

abought:
Replaces #5224 and #5243 .
Refs https://openscience.atlassian.net/browse/OSF-5969

## Purpose
Phase out usage of knockout punches template helpers due to reported incompatibilities.

This PR modifies all pages on the entire site so that no uses of knockout punches should remain. It incorporates and updates the two PRs it is intended to replace.

## Summary of changes and testing notes
See closed PRs #5224 and #5243 for details.  This will require regression testing throughout the site to verify links, logs, display text, and button functionality.

Note to reviewers: all commits up to and including b5bc727 already passed through code review, in the two linked precursor PRs.

## Deployment notes
This PR removes a top-level mako filter- it affects the behavior of templates in a way that does not trigger an auto-rebuild. You may encounter errors when rendering mako pages, unless you clear the mako template cache:

1. Stop the server
2. Delete the contents of `/tmp/mako_modules`
3. Restart the OSF server instance


abought:
Brought PR up to date and removed a few more ko punches usages that crept in.

(including some `log_templates` changes in merge commit b216134 ; sorry about that)

website/static/js/koHelpers.js
```diff
@@ -209,6 +209,46 @@ ko.bindingHandlers.fadeVisible = {
     }
 };
 
+var fitHelper = function(value, length, replacement, trimWhere) {
```
brianjgeiger: If one or both of these are the code we're using from Punches, we should label it as such.

brianjgeiger:
Looks fine. We should add in a comment or two above where we grabbed code. We've added the license already, so that's cool. Everything else is essentially the same thing over and over with one notable exception where Andy had to add a slash. But it seems fine and should be ready for some intensive QA.

brianjgeiger:
Replaced by #5491

PR #5417, opened 2016-04-07
[fix] Speed up User PUT and PATCH Methods by modifying User properties [OSF-5338]

erinspace:
https://openscience.atlassian.net/browse/OSF-5338

# Purpose
```PUT``` and ```PATCH``` methods are very slow on the OSF API at the moment. @chrisseto noticed that was because the ```contributor_to``` and ```visible_contributor_to``` methods were list comprehensions relying on a previous property and not a faster ModularODM Query. This PR changes the ```contributor_to``` and and ```visible_contributor_to``` properties to return a ModularODM query with the appropriate nodes.

# Changes 
- Move the  ```contributed``` property to be closer to the others
- Update the ```contributor_to``` property to return a ModularODM queryset instead of relying on the ```contributed``` property
- Update the ```visible_contributor_to``` property to return a ModularODM queryset instead of relying on the ```contributed``` property
- Add tests for the ```contributed_to``` and ```visible_contributor_to``` properties

# Side Effects
- The places that use those properties right now don't seem to be affected by this change, but these properties are now returning query sets instead of acting as generators so I might have missed something....

Moved from #5415!

### DIFF
--- a/framework/auth/core.py
+++ b/framework/auth/core.py
@@ -1086,7 +1086,7 @@ class User(GuidStoredObject, AddonModelMixin):
         from website.project.model import Node
         return Node.find(
             Q('contributors', 'eq', self._id) &
-            Q('is_deleted', 'eq', False) &
+            Q('is_deleted', 'ne', True) &
             Q('is_bookmark_collection', 'ne', True) &
             Q('visible_contributor_ids', 'eq', self.user._id)
         )
--- a/framework/auth/core.py
+++ b/framework/auth/core.py
@@ -1074,12 +1074,11 @@ class User(GuidStoredObject, AddonModelMixin):
 
     @property
     def contributor_to(self):
-        return (
-            node for node in self.contributed
-            if not (
-                node.is_deleted
-                or node.is_bookmark_collection
-            )
+        from website.project.model import Node
+        return Node.find(
+            Q('contributors', 'eq', self._id) &
+            Q('is_deleted', 'eq', False) &
+            Q('is_bookmark_collection', 'eq', False)
         )
 
     @property
--- a/framework/auth/core.py
+++ b/framework/auth/core.py
@@ -1088,7 +1088,7 @@ class User(GuidStoredObject, AddonModelMixin):
             Q('contributors', 'eq', self._id) &
             Q('is_deleted', 'ne', True) &
             Q('is_collection', 'ne', True) &
-            Q('visible_contributor_ids', 'eq', self.user._id)
+            Q('visible_contributor_ids', 'eq', self._id)
         )
 
     def get_summary(self, formatter='long'):
--- a/framework/auth/core.py
+++ b/framework/auth/core.py
@@ -1083,9 +1083,12 @@ class User(GuidStoredObject, AddonModelMixin):
 
     @property
     def visible_contributor_to(self):
-        return (
-            node for node in self.contributor_to
-            if self._id in node.visible_contributor_ids
+        from website.project.model import Node
+        return Node.find(
+            Q('contributors', 'eq', self._id) &
+            Q('is_deleted', 'eq', False) &
+            Q('is_bookmark_collection', 'eq', False) &
+            Q(self._id, 'eq', 'visible_contributor_ids')
         )
 
     def get_summary(self, formatter='long'):
--- a/framework/auth/core.py
+++ b/framework/auth/core.py
@@ -1078,7 +1078,7 @@ class User(GuidStoredObject, AddonModelMixin):
         return Node.find(
             Q('contributors', 'eq', self._id) &
             Q('is_deleted', 'ne', True) &
-            Q('is_bookmark_collection', 'ne', True)
+            Q('is_collection', 'ne', True)
         )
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: framework/auth/core.py [hunk 1]
```diff
@@ -1073,20 +1068,27 @@ def profile_url(self):
...
-        return (
-            node for node in self.contributed
-            if not (
-                node.is_deleted
-                or node.is_bookmark_collection
-            )
+        from website.project.model import Node
+        return Node.find(
+            Q('contributors', 'eq', self._id) &
+            Q('is_deleted', 'eq', False) &
```
chrisseto: IIRC all `is_deleted` queries must be `ne` because only deleted nodes have the `is_deleted` field.

framework/auth/core.py
```diff
@@ -1073,20 +1068,27 @@ def profile_url(self):
...
+            Q('is_bookmark_collection', 'eq', False)
```
chrisseto: I think the attribute is `is_collection` here, isn't `is_bookmark_collection` a property?

`is_collection` queries must also be `ne`

framework/auth/core.py
```diff
@@ -1073,20 +1068,27 @@ def profile_url(self):
...
     def visible_contributor_to(self):
-        return (
-            node for node in self.contributor_to
-            if self._id in node.visible_contributor_ids
+        from website.project.model import Node
+        return Node.find(
+            Q('contributors', 'eq', self._id) &
+            Q('is_deleted', 'eq', False) &
+            Q('is_bookmark_collection', 'eq', False) &
+            Q(self._id, 'eq', 'visible_contributor_ids')
```
chrisseto: This query is reversed

framework/auth/core.py
```diff
@@ -1073,20 +1068,27 @@ def profile_url(self):
...
 
     @property
     def visible_contributor_to(self):
-        return (
-            node for node in self.contributor_to
-            if self._id in node.visible_contributor_ids
+        from website.project.model import Node
+        return Node.find(
+            Q('contributors', 'eq', self._id) &
+            Q('is_deleted', 'eq', False) &
```
chrisseto: See above comments

erinspace:
Thank you @chrisseto!! 📐

framework/auth/core.py
```diff
@@ -1073,20 +1068,27 @@ def profile_url(self):
...
-            node for node in self.contributed
-            if not (
-                node.is_deleted
-                or node.is_bookmark_collection
-            )
+        from website.project.model import Node
+        return Node.find(
+            Q('contributors', 'eq', self._id) &
+            Q('is_deleted', 'ne', True) &
+            Q('is_bookmark_collection', 'ne', True)
```
chrisseto: This will need to query on `is_collection` as well.
```python
    # Project Organization
    is_bookmark_collection = fields.BooleanField(default=False, index=True)
    is_collection = fields.BooleanField(default=False, index=True)
```

chrisseto: Maybe just `is_collection` @pattisdr or @brianjgeiger ?

brianjgeiger: `is_collection` will also catch `is_bookmark_collection`, so that's probably the best way to go.

tests/models/test_user.py
```diff
@@ -124,6 +124,27 @@ def test_contributed_property(self):
         projects_contributed_to = project.model.Node.find(Q('contributors', 'eq', self.user._id))
         assert_equal(list(self.user.contributed), list(projects_contributed_to))
 
+    def test_contributor_to_property(self):
+        # Make sure there's at least one deleted project and bookmark collection
+        factories.ProjectFactory(creator=self.user, is_deleted=True)
+        factories.BookmarkCollectionFactory(creator=self.user)
+
+        contributor_to = project.model.Node.find(
```
brianjgeiger: This is not an ideal approach. Rather than re-creating the query, set up a node with a mix of visible and non-visible contributors and ensure that the correct contributors are in the correct lists, and the ones who aren't supposed to be are not.

erinspace: Makes sense! I knew this wasn't ideal and this sounds much more sound.

tests/models/test_user.py
```diff
@@ -124,6 +124,34 @@ def test_contributed_property(self):
         projects_contributed_to = project.model.Node.find(Q('contributors', 'eq', self.user._id))
         assert_equal(list(self.user.contributed), list(projects_contributed_to))
 
+    def test_contributor_to_property(self):
+        normal_node = factories.ProjectFactory(creator=self.user)
+        deleted_node = factories.ProjectFactory(creator=self.user, is_deleted=True)
+        bookmark_collection_node = factories.BookmarkCollectionFactory(creator=self.user)
+        collection_node = factories.CollectionFactory(creator=self.user)
+        contributor_to_nodes = [node._id for node in self.user.contributor_to]
```
brianjgeiger: Almost there. Could you ensure that the an invisible contributor will also be returned via contributor_to, since you have the other side of that covered in the next test?

PR #5420, opened 2016-04-07
[fix] Add serializer to InstitutionAuth to fix API Docs [OSF-5957]

erinspace:
https://openscience.atlassian.net/browse/OSF-5957

# Purpose

API Swagger docs were throwing a server error:
```
AssertionError: 'InstitutionAuth' should either include a `serializer_class` attribute, or override the `get_serializer_class()` method.
```

# Changes
Add a serializer (the most basic one I could think of, the ```JSONAPISerializer```) 

# Side Effects
- None anticipated, unless the way that view was intended to be serialized is dramatically different

### DIFF
--- a/api/institutions/views.py
+++ b/api/institutions/views.py
@@ -12,6 +12,7 @@ from website.models import Node, User, Institution
 from api.base import permissions as base_permissions
 from api.base.filters import ODMFilterMixin
 from api.base.views import JSONAPIBaseView
+from api.base.serializers import JSONAPISerializer
 from api.base.utils import get_object_or_error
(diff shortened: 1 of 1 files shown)


brianjgeiger:
Looks good. Simple. Works. Doesn't break tests. Shouldn't break anything. My only question for @icereval is whether institutional endpoints should also be dev only at this point. That can be a separate issue though.

icereval:
@brianjgeiger, devmode should be use very sparingly as we discussed, originally it was only intended for the api's very first release to limit its scope a bit, but its certainly usable in special case by case scenarios going forward :-)

PR #4279, opened 2015-09-11
[WIP] Convert misc pages to markupsafe (markupsafe-4)

abought:
Refs [#OSF-4426]

## Purpose
Continues the markupsafe series. This PR focuses on single / miscellaneous pages, including search, meetings, SHARE, login, and dashboard functionality. All testing notes from previous items in the series continue to apply.

## Summary of changes / routes affected
Some of these routes are static or otherwise make minimal use of mako variables. Routes particularly expected to be affected by this conversion are marked with a (*). 

TODO: Write list once conversion complete

PR #5421, opened 2016-04-07
Convert remainder of site to markupsafe (markupsafe-4)

abought:
Refs https://openscience.atlassian.net/browse/OSF-4426
Replaces and extends #4279 

## Purpose
Improve handling of user-entered text across many sections of the site, and help resolve display artifacts.

## Summary of changes
- Convert all routes not covered in previous PRs.

- Try to reduce the number of language strings that contain HTML outright (cut down on presentation concerns in the backend)

- If we know that some / all status message text contains HTML, then make sure any user-entered params in that string are escaped before settings `trust=True`

## Testing notes and list of affected routes
See JIRA ticket for notes and affected route list. QA has a pre-existing list of sample content that can be used to confirm pages render as expected.


framework/auth/forms.py
```diff
@@ -3,6 +3,8 @@
 # TODO: Remove me
 
 from wtforms import ValidationError
+
+from framework import auth
```
brianjgeiger: Don't reverse it or anything, but especially for a PR of this size, it's a good idea not to do irrelevant formatting or line-shuffling changes. I mean, if this fixes something that was broken, cool, but otherwise it's just another potential merge conflict. Best to save those for otherwise tiny PRs if it's important but not urgent.

framework/auth/views.py
```diff
@@ -80,10 +80,10 @@ def forgot_password_post():
         if user_obj:
             #TODO: Remove this rate limiting and replace it with something that doesn't write to the User model
             now = datetime.datetime.utcnow()
-            last_attempt = user_obj.forgot_password_last_post or now - datetime.timedelta(seconds=FORGOT_PASSWORD_MINIMUM_TIME)
+            last_attempt = user_obj.forgot_password_last_post or now - datetime.timedelta(seconds=settings.FORGOT_PASSWORD_MINIMUM_TIME)
```
brianjgeiger: Like this, for example. Saman has another PR which renames this variable to something more generic, and it's going to conflict as soon as it gets merged in. (I admit that I should have seen settings in there and done it right the first time; that's totally on me.)

website/addons/badges/templates/badges_user_settings.mako
```diff
@@ -96,6 +96,7 @@ You have not created any Badges.
 
 //todo fetch url from server
   var jsonToList = function(badge, id) {
+      // TODO: Rewrite this substantially before this feature is re-activated
```
brianjgeiger: This could use some explanation for future developer archaeologists. Maybe a link to the markup safe issue or a comment about what needs re-writing?

website/addons/base/__init__.py
```diff
@@ -1,24 +1,26 @@
 # -*- coding: utf-8 -*-
-import os
 import glob
 import importlib
 import mimetypes
-from bson import ObjectId
-from modularodm import fields
-from mako.lookup import TemplateLookup
+import os
```
brianjgeiger: Blah blah more potential merge conflicts etc. You get the idea.

website/addons/wiki/templates/edit.mako
```diff
@@ -354,42 +353,39 @@
...
-    var canEditPageName = isContributor && ${json.dumps(
-        wiki_id and wiki_name != 'home'
-    )};
+    var canEditPageName = isContributor && ${(wiki_id and wiki_name != 'home') | sjson, n };
 
     window.contextVars = window.contextVars || {};
     window.contextVars.wiki = {
         canEdit: canEditBody,
         canEditPageName: canEditPageName,
-        usePythonRender: ${json.dumps(use_python_render)},
```
brianjgeiger: Could you briefly explain why we used to need json.dumps here and we don'y any more?

website/profile/views.py
```diff
@@ -353,9 +354,10 @@ def user_account_password(auth, **kwargs):
         user.change_password(old_password, new_password, confirm_password)
         user.save()
     except ChangePasswordError as error:
-        push_status_message('<br />'.join(error.messages) + '.', kind='warning')
+        for m in error.messages:
+            push_status_message(m, kind='warning', trust=False)
```
brianjgeiger: This looks like it will result in a series of error messages rather than one message. This is inconvenient, and may actually be a regression. Can we combine these without using HTML rather than having multiple?

website/templates/project/nodes_privacy.mako
```diff
@@ -1,4 +1,4 @@
-<div id="nodesPrivacy" class="modal fade" div style="display: none;">
```
brianjgeiger: Was `display: none` not necessary or did you move it somewhere else?

website/templates/util/render_form.mako
```diff
@@ -2,14 +2,14 @@
     % if id:
         id="${id}"
     % endif
-    name="${name}" method="${method_string}" ${"action=\""+action_string+"\"" if action_string else ""} class="${form_class}">
+    name="${name}" method="${method_string}" action="${action_string}" class="${form_class}">
```
brianjgeiger: Much better. I'm afraid to check to see if I did the terrible version you've replaced.

website/templates/util/render_keys.mako
```diff
@@ -1,3 +1,4 @@
+<!-- TODO: File is unused; candidate for deletion? -->
```
brianjgeiger: Probably so. This was probably what we used for developer keys back in the v1 days. I say "used" but I don't think it saw much use outside of an early version of SHARE and maybe a couple of edge cases.

brianjgeiger:
By and large, seems fine. A couple of questions to answer, a request for a more useful TODO, one small change for error message pushing, and please take out the changes to the rate limiting code for forgotten passwords. 🐧

website/addons/wiki/templates/edit.mako
abought: `json.dumps` was deprecated in favor of the `sjson, n` pair. See JIRA ticket for link to conversion guidelines.

Also see function definition for a brief description of vectors the newer filter is designed to handle. The `n` part of the filter pair ensures that we do not apply HTML-specific rules to a variable that is not HTML.
https://github.com/CenterForOpenScience/osf.io/blob/develop/website/util/sanitize.py#L121

website/profile/views.py
abought: Multiple individual status messages seems like it is more in line with behavior elsewhere on the site. (eg registrations with public/private addons of various types) Each individual problem with the form input is given its own space.

The reason for the HTML was to display them cleanly on separate lines. I'm very reticent to move responsibility for HTML generation further into the backend (eg if we added more markupsafe escaping for all variables before sending a user-generated string).

`push_status_message` could be used to send a separate CSS class instead using some [whitespace tricks](http://stackoverflow.com/questions/2703601/how-to-give-line-break-from-css-without-using-br)... but this could introduce display/browser anomalies of its own, and it still leaves the issue of display concerns being folded into the backend, plus special casing. We can weigh UI options tomorrow when I'm back in the office.

website/templates/project/nodes_privacy.mako
abought: See line 3. This appeared to be a markup error, judging from the div attribute on the div tag.

I do recall verifying that the page looked the same for me when viewing the page directly.

website/templates/util/render_keys.mako
abought: Thought so. :) I wouldn't mind a follow-on cleanup PR, but wanted to limit the scope of this one and not also go deleting things. Too many balls in the air at once. :)

framework/auth/views.py
abought: Wouldn't this be an extremely easy merge conflict to resolve? I'm not sure I see the advantage of removing here....?

brianjgeiger: Probably, but why ask for trouble? Saman already did this plus the other related changes. If someone messes up a merge conflict, it may not be immediately noticed, and the nature of this particular PR is one that a lot of little merge conflicts will build up that may be hard to distinguish.

abought: I'll change this if absolutely necessary- but as you say, merge conflicts are going to be a pain point for this PR no matter what.

My strong preference would be to resolve this by giving the developer a heads-up , rather than hacking at the code. The developer can (and should) be trusted to examine the final diff for a merge conflict before moving to review.

website/addons/badges/templates/badges_user_settings.mako
abought: Per policy, reply to this one has been added to JIRA ticket.
https://openscience.atlassian.net/browse/OSF-4426?focusedCommentId=31238&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-31238

framework/auth/views.py
brianjgeiger: Yes, please change these three lines back to how they were.

abought:
Pushed requested change. Per the 0.68 branch announcement... how do we want to go about resolving these merge conflicts? @sloria ?

brianjgeiger:
I believe this is going into develop not the feature branch. It will be released after 0.68.0 last I talked with Steve.

sloria:
Correct; since this is a 0.69.0 feature, you were correct to merge `develop`. 👍

website/profile/views.py
brianjgeiger: I'm preeeetty sure this was something we had to fix at some point for a similar endpoint, but not necessarily in this view. Ideally, of course, the front-end should be validating and not sending the request until it's pretty much ready, which it may be and this might not really be a problem. 

The least impactful would be to join with `'. '` instead. Hmmm. Let's keep it the way you have it and, if it's terrible, we'll fix it properly later.

brianjgeiger:
Seems good to me. 🎉 🐧 🎉

website/templates/public/pages/getting_started.mako
```diff
@@ -73,6 +73,7 @@
                 </ul>
             </div>
         </div>
+    </div>
```
sloria: Good catch

website/templates/util/render_keys.mako
sloria: Yep, this is unused; I'd welcome a PR removing it =)

PR #5427, opened 2016-04-08
[RfR] [OSF-5654] Fix: Show 'loading contributors...' in more cases 

barbour-em:
## Purpose
Cover the case where a user has searched for a contributor, selects that contributor, and then searches for a second contributor with a different name.

## Changes
Make sure to reset `doneSearching` when a subsequent search happens.
## Refs
https://openscience.atlassian.net/browse/OSF-5654

[OSF-5654]


mfraezz:
LGTM. Dropping this in RTM :star:

PR #5428, opened 2016-04-08
[hotfix] Pass user to file delete

samanehsan:
When deleting a file node, pass in the user so that the `TrashedFileNode.deleted_by` gets set.

### DIFF
--- a/website/addons/base/views.py
+++ b/website/addons/base/views.py
@@ -384,7 +384,7 @@ def create_waterbutler_log(payload, **kwargs):
 
 
 @file_signals.file_updated.connect
-def addon_delete_file_node(self, node, event_type, payload, user=None):
+def addon_delete_file_node(self, node, user, event_type, payload):
     """ Get addon StoredFileNode(s), move it into the TrashedFileNode collection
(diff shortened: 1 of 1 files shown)


website/addons/base/views.py
```diff
@@ -552,7 +552,7 @@ def addon_view_or_download_file(auth, path, provider, **kwargs):
 
             # Move file to trashed file node
             if not TrashedFileNode.load(file_node._id):
-                file_node.delete()
+                file_node.delete(user=auth.user)
```
felliott: `user` should not be passed here.  This is the case where we have a StoredFileNode, but can't find in it in the metadata response.  i.e. it disappeared.

samanehsan: ok, that makes sense

felliott:
Tested locally, works for me!  Thanks, @samanehsan!

icereval:
@felliott, @samanehsan test seem to be unhappy https://travis-ci.org/CenterForOpenScience/osf.io/jobs/121767087#L3319 :-/

PR #5291, opened 2016-03-24
[fix] Use local instances of expandOnLoad

billyhunt:
Updated version of https://github.com/CenterForOpenScience/osf.io/pull/5233
Related to https://github.com/CenterForOpenScience/osf.io/pull/4958/files

When adding contributors, the top level is not expanding on staging.  This seems to be correlated to a move of multiple expandOnLoad functions to a global helper.  We are going to go back to using local instances for now.

PR #5429, opened 2016-04-08
Feature/contributors add expand tree

caneruguz:
Updating this PR https://github.com/CenterForOpenScience/osf.io/pull/5291
because of merge conflicts after 0.66 

Tested on the project page treebeard that changes make public 
Tested on the add contributor modal, 

things seems to work fine but across these PR some context might be lost.


billyhunt:
The two places that were not working are not @caner, Thanks!

PR #5423, opened 2016-04-08
Add get_absolute_url method to FileVersionSerializer

cwisecarver:
This fixes [NotImplementedError exceptions for file version lists](https://sentry.cos.io/sentry/osf-iy/issues/479/).

PR #5433, opened 2016-04-08
Force all serializers to have a get_absolute_url method

cwisecarver:
This test checks that all serializers not based in `api_test` or `test` modules have a get_absolute_url method. 

I also had to add a get_absolute_url method to `NodeAlternativeCitationSerializer` that doesn't do anything, other than throw an exception, because AFAIK Node Alternative Citations don't have a url.


cwisecarver:
@icereval This will pass if #5423 and #5426 are merged prior and it's rebased. I can either merge those into this PR and we can cross our fingers on a merge conflict or you can merge #5423, #5426, and then #5433 and I promise it's tests will pass (this statement has no legal liability).

PR #5436, opened 2016-04-09
secure DOIs

katrinleinweber:
The DOI foundation has [stopped listing the `dx` resolver as "preferred", and started to support HTTPS](https://www.doi.org/doi_handbook/3_Resolution.html#3.8). I hope I correctly understood your code to generate/display DOI links, not parse them.

I omitted the same change in `citeproc.js`, because [an upstream PR](https://bitbucket.org/fbennett/citeproc-js/pull-requests/8/generate-secure-doi-links/diff) is pending.


sloria:
Thanks @katrinleinweber !

PR #5438, opened 2016-04-11
0.67.0/fix qa info logging

brianjgeiger:
Purpose
-----------
The github 'Merge and Squash' command broke the git_logs task because. This fixes the problem and makes it a bit nicer for QA because they have clickable links to go to the github pr.

Changes
------------
1. Added a meta script module for gathering info such as this
2. Added a gatherer script to get the last 1000 commits from git and compare those to the last 100 PRs from github. Whatever overlaps there gets its json dumped into a json file. 
3. Added a GITHUB_API_TOKEN setting
4. Added a knockout script to get the json data that was created by the script and show it when you click on the 'Warning: dev mode' banner (and hide it again if you click again)

qa helper

screen shot 2016-04-11 at 9 10 04 am
screen shot 2016-04-11 at 9 10 07 am

Side effects
--------------
- It seems to work fine out of dev mode. Well, it should. I tested the else branch of the mako by removing everything from the if branch that didn't match, and it doesn't error or anything. So it should be good.
- If you don't have a github api token in your local.py, then you won't get any cool stats, and the script will 404 in the console, but otherwise will not show any problems.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@ node_modules
 # Allow robots.txt to be deployment-specific
 website/static/robots.local.txt
 /website/static/git_logs.txt
+/website/static/gitlogs.json
 
 # OSF-specific
 ##############
--- a/.gitignore
+++ b/.gitignore
@@ -55,7 +55,7 @@ node_modules
 # Allow robots.txt to be deployment-specific
 website/static/robots.local.txt
 /website/static/git_logs.txt
-/website/static/gitlogs.json
+/website/static/git_logs.json
 
 # OSF-specific
 ##############
--- a/.gitignore
+++ b/.gitignore
@@ -54,8 +54,6 @@ website/static/vendor/bower_components/
 node_modules
 # Allow robots.txt to be deployment-specific
 website/static/robots.local.txt
-/website/static/git_logs.txt
-/website/static/git_logs.json
 
 # OSF-specific
 ##############
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -20,6 +20,3 @@ livereload==2.3.2
 
 # Avoid eating cpu with live reloading
 watchdog==0.8.3
-
-# QA Helper
-requests==2.5.3
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -20,3 +20,6 @@ livereload==2.3.2
 
 # Avoid eating cpu with live reloading
 watchdog==0.8.3
+
+# QA Helper
+requests==2.5.3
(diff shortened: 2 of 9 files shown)


brianjgeiger:
I have a build issue I need to fix with my new script module. Once I fix that, tests should pass again.

requirements/dev.txt
```diff
@@ -20,3 +20,6 @@ livereload==2.3.2
 
 # Avoid eating cpu with live reloading
 watchdog==0.8.3
+
+# QA Helper
```
icereval: Any reason the same version in the root `requirements.txt` does not work? https://github.com/brianjgeiger/osf.io/blob/0.67.0/fix-QA-info-logging/requirements.txt#L51

brianjgeiger: No idea. This didn't actually work, though, and I ended up having to move the import inside the function. I don't have any good explanation for it, but these two lines should be removable without affecting anything.

.gitignore
```diff
@@ -55,6 +55,7 @@ node_modules
 # Allow robots.txt to be deployment-specific
 website/static/robots.local.txt
 /website/static/git_logs.txt
+/website/static/git_logs.json
```
sloria: How about putting git_logs in `website/static/built/`, which is already ignored?

brianjgeiger: Sure

scripts/meta/gatherer.py
```diff
@@ -0,0 +1,42 @@
+import json
+from website.settings import GITHUB_API_TOKEN
+
+the_filename = 'website/static/git_logs.json'
+
+
+def gather_pr_data():
+    import requests
```
sloria: Requests is part of our core `requirements.txt`, so no need to import in the function.

scripts/meta/gatherer.py
```diff
@@ -0,0 +1,42 @@
+import json
+from website.settings import GITHUB_API_TOKEN
+
+the_filename = 'website/static/git_logs.json'
```
sloria: Minor: Uppercase constants. Also, no need for articles; how about `GIT_LOGS_FILE`. Also, best to use `os.path.join` for building filepaths.

website/static/js/pages/base-page.js
```diff
@@ -111,6 +112,7 @@ $(function() {
         $osf.initializeResponsiveAffix();
     }
     new NavbarControl('.osf-nav-wrapper');
+    new DevModeHelper('.dev-mode-helper');
```
sloria: Minor: Better to bind by ID rather than class (on the off-chance the the class is used more than once).

sloria:
Pass finished; just minor changes to make and this is good to merge.

scripts/meta/gatherer.py
brianjgeiger: Travis disagrees with you on this. https://travis-ci.org/CenterForOpenScience/osf.io/builds/122245666

PR #5439, opened 2016-04-11
[RFR] Allow cross-origin requests to APIv2

samchrisinger:
# Purpose

Third party sites using OAuth or personal access tokens will want to make authenticated requests to our APIv2. The current CORS policy makes this impossible, so this PR aims to allow cross-origin requests when:
- the client includes the Authorization header (auth still must be checked later in the middleware pipe) AND
- the client is not sending cookies with the request

# Changes

Slight refactor of the code for dynamically whitelisting Institutions. Given the nature of the CorsMiddleware source, I think `origin_not_found_in_white_lists` is an elegant place to add custom logic, but checking for cross-origin requests requires having the `request` in context. So there's a little hackery here to bind the request into that method of the class instance, but it seems to work pretty nicely.


api/base/middleware.py
```diff
@@ -107,11 +109,30 @@ class CorsMiddleware(corsheaders.middleware.CorsMiddleware):
...
-        not_found = super(CorsMiddleware, self).origin_not_found_in_white_lists(origin, url)
-        if not_found:
-            not_found = Institution.find(Q('domains', 'eq', url.netloc.lower())).count() == 0
-        return not_found
+    def process_request(self, request):
+        def origin_not_found_in_white_lists(self, request, origin, url):
+            not_found = super(CorsMiddleware, self).origin_not_found_in_white_lists(origin, url)
+            if not_found:
+                # Check if origin is in the dynamic Institutions whitelist
+                if Institution.find(Q('domains', 'eq', url.netloc.lower())).count() != 0:
```
cwisecarver: It would be nice if we could generate this list on application start so we're not doing this query on every request but it's probably out of scope for this PR.

cwisecarver:
One tiny comment, it was already that way prior to your changes so it's probably irrelevant. lgtm @samchrisinger  👍

api/base/middleware.py
samchrisinger: +1. Could be an easy change

samchrisinger: @cwisecarver this may be intentional to allow live-loading new institutions?

samchrisinger: But yeah doing this on every request seems excessive

cwisecarver: I talked to @icereval about it a while ago and he was okay loading them on app start. I started working on it loading up inside settings but then :shipit: and I fell into something else and never looped back. It's not a big deal either way. I just remember seeing the number of calls to this query in new relic and thinking it didn't need to be there.

icereval: agreed, if we can cache this :star: :+1:

samchrisinger:
@cwisecarver if the build passes I think I'm all done here.

Right now I'm not unit testing scenarios for preflight requests. Any concerns here to look out for?

api/base/middleware.py
```diff
@@ -26,13 +25,13 @@
...
     celery_teardown_request
 )
-from website.models import Institution
 from framework.transactions.handlers import (
     transaction_before_request,
     transaction_after_request,
     transaction_teardown_request
 )
 from .api_globals import api_globals
+from api.base import setttings as api_settings
```
cwisecarver: extra t's in settings @samchrisinger

api/base/wsgi.py
```diff
@@ -6,7 +6,10 @@
 For more information on this file, see
 https://docs.djangoproject.com/en/1.8/howto/deployment/wsgi/
 """
-from website import settings
+import itertools
+
+from website import settings, models
+from api.base import settings as api_settings
```
cwisecarver: proper number of t's in settings 👍

api/base/settings/__init__.py
```diff
@@ -22,3 +23,11 @@
     from . import defaults
     for setting in ('JWE_SECRET', 'JWT_SECRET'):
         assert getattr(local, setting, None) and getattr(local, setting, None) != getattr(defaults, setting, None), '{} must be specified in local.py when DEV_MODE is False'.format(setting)
+
+def load_institutions():
+    global INSTITUTION_ORIGINS_WHITELIST
+    from website import models
+    INSTITUTION_ORIGINS_WHITELIST = tuple(itertools.chain(*[
+        institution.domains
```
cwisecarver: Would you want to .lower() here too just to be safe? User input is unfortunately done most of the time by users.

samchrisinger: Good call

cwisecarver:
@samchrisinger Just loaded the newest ref of the PR under gunicorn with 24 workers. Only got one "INSTITUTIONS LOADED!" print message. Seems like this is good. Lmk if you decide to refactor to a class level var and I'll retry.

samchrisinger:
Thanks for checking this out, I'll take another shot in the morning.

PR #5441, opened 2016-04-11
[rfr] Prevent changing attached DraftRegistration file

barbour-em:
## Purpose

Prevent users from changing an attached file when prereg form is in progress. 

## Changes
- Any file attached to a DraftReg becomes "checked-out" when the preregistration form is submitted.
- The files become "checked-in" when the registration is approved or sent back as rejected.
- If the user tries to check-in an attached file before the DraftReg is approved, they receive a warning that this action will cancel their submission. If they continue, any DraftReg with that file attached will be rejected and all files attached to it become checked-in.

## Side effects
- Expose /reject/ route on front end. I don't like this so I'd like suggestions on the best way to handle it. 
- Checking in an attached file can do a lot if the user doesn't read the warning. 

## Ticket
https://openscience.atlassian.net/browse/OSF-5585
Depends on #5280


barbour-em:
Well this has been a disaster

PR #5445, opened 2016-04-11
Private/View Only Rework [#OSF-5792]

alexschiller:
Note: see https://github.com/CenterForOpenScience/osf.io/pull/4509 for original. This PR is for the rework per some requested changes. All changes in the linked PR apply to this one as well.

## Purpose
Make view-only links actually view-only, make anonymized view-only links not show public components. 

## Changes
Massively strip functionality from the project page if an active view-only link is on the URL:
1. Rework the nav bar to explain view-only link and give a way to return to the dashboard
2. Remove access to all editing functionality
3. Remove all menu bar info except Files and Wiki
4. Anonymize public components if on an anonymous link. 
5. Only allow access to components selected when creating the link, even if the component is public. This makes it so nothing is implicitly added to the view-only link unless the creator of the link makes it explicit that they wanted to include it. 
6. Logged in users and anonymous users now see exactly the same view of a project when viewing it through a view only link. 
7. If a view only link is active but a person is on something not within the scope of that link they will receive a redirect and the current view_only link will be removed. This can happen 1) if not in a project and you try a view only link, 2) if on a node that the view_only link does not contain, 3) use an invalid view_only link

## Side Effects
1. ~~One strange situation could still occur if someone were to add the view-only link to a different OSF page. Currently it will ONLY strip functionality from projects, but it will replace the top menu bar with the view only bar. I'm not sure there is a good way to change this.~~  This has been modified so that view_only links now redirect to not include the view_only link if the link is not valid

## QA Considerations
1. Both Public and Private components should be tested for view access.
2. Both Logged in (with access) and un-logged in users should be tested
3. Both anonymized and non-anonymized links should be tested
4. There should be no way to modify the project while the link is active
5. Users should not be able to gain access to other project information using a private link on a different project (Whenever I test locally it redirects to login, but I can't confirm how it will work on staging). 
6. Private links with institutions should not show institution access (I wasn't sure how to test this locally, but the code looked right)
7. Testing that view_only links redirect properly when the link is not valid for the current url location should be tested.


website/static/css/style.css
```diff
@@ -648,4 +653,4 @@ button.close {
 }
 .breaklines {
     white-space: pre-wrap;
-}
\ No newline at end of file
+}
```
mfraezz: 🎉

website/templates/footer.mako
```diff
@@ -1,34 +1,36 @@
...
-                <a href="https://www.facebook.com/OpenScienceFramework" aria-label="Facebook"><i class="fa fa-facebook fa-2x"></i></a>
-                <a href="https://groups.google.com/forum/#!forum/openscienceframework" aria-label="Google Group"><i class="fa fa-group fa-2x"></i></a>
-                <a href="https://www.github.com/centerforopenscience" aria-label="GitHub"><i class="fa fa-github fa-2x"></i></a>
-                <a href="https://plus.google.com/103557785986281627994" aria-label="Google Plus" rel="publisher"><i class="fa fa-google-plus fa-2x"></i></a>
-            </div> <!-- column -->
         </div>
-    </div>
-</footer><!-- end footer -->
+    </footer><!-- end footer -->
+%endif
```
mfraezz: 😞

mfraezz:
`Analytics` still shows up for public projects visited with AVOLs: 

screen shot 2016-04-12 at 10 13 37

`Settings` can still be accessed by modifying the url, but only when the viewing user is a contributor, and only `Email Notifications` is displayed:
 
screen shot 2016-04-12 at 10 15 53

Other nodes can be accessed with the same `PrivateLink.key`, with weird behavior:

screen shot 2016-04-12 at 10 18 02

But this doesn't happen if the user isn't logged in (contributorship on the node likely matters as well): 

screen shot 2016-04-12 at 10 20 29

mfraezz:
Please add or update tests to make sure that requests with `PrivateLinks` behave as expected. The fact that these changes didn't break any tests means we aren't testing this functionality thoroughly enough.

Also, please resolve merge conflicts.

framework/flask/__init__.py
```diff
@@ -70,6 +70,9 @@ def redirect(location, code=302):
     flask.redirect to ensure the correct behavior of view-only
     links.
     """
+    if strip_view_only:
+        return flask_redirect(location, code=code)
+
     view_only = request.args.get('view_only', '')
     if view_only:
```
mfraezz: Could this be simplified with `if view_only and not strip_view_only`?

framework/sessions/__init__.py
```diff
@@ -44,6 +47,26 @@ def add_key_to_url(url, scheme, key):
     parsed_redirect_url = parsed_url._replace(**replacements)
     return urlparse.urlunparse(parsed_redirect_url)
 
+def get_node_id(url):
+    try:
+        node_id = [x for x in url.split("/") if x][2]
+        # Avoid circular import
+        from website.models import Node
+        if Node.find(Q('_id', 'eq', node_id)).count() == 1:
```
mfraezz: Loading an invalid guid just returns `None`; this can be simplified.

Also, it seems like there should be a better way of doing this.

framework/sessions/__init__.py
```diff
@@ -44,6 +47,26 @@ def add_key_to_url(url, scheme, key):
     parsed_redirect_url = parsed_url._replace(**replacements)
     return urlparse.urlunparse(parsed_redirect_url)
 
+def get_node_id(url):
+    try:
+        node_id = [x for x in url.split("/") if x][2]
```
mfraezz: This isn't guaranteed to work, as we support the older `/project/<guid>/` url form, as well as just `/<guid>/`.

framework/sessions/__init__.py
```diff
@@ -58,11 +81,28 @@ def prepare_private_key():
     if request.method != 'GET':
         return
 
-    # Done if private_key in args
+    node_id = get_node_id(request.url)
```
mfraezz: `request.params` would give you a smaller string to attempt to parse

mfraezz: Also, you don't need to do this unless you know there's a private link. Otherwise this will slow every request down.

framework/sessions/__init__.py
```diff
@@ -58,11 +81,28 @@ def prepare_private_key():
...
     key_from_args = request.args.get('view_only', '')
-    if key_from_args:
+
+    # Done if not on a node; If not on a node remove the key
+    if not node_id:
+        if key_from_args:
+            return redirect(request.base_url, code=http.TEMPORARY_REDIRECT, strip_view_only=True)
```
mfraezz: It seems like some of these conditionals could be DRY'd up.

framework/sessions/__init__.py
```diff
@@ -76,11 +116,11 @@ def prepare_private_key():
         scheme = None
         key = None
 
-    # Update URL and redirect
+    # Update URL and redirect`
```
mfraezz: >`

framework/sessions/__init__.py
mfraezz: If it can't be avoided, you should probably be iterating over `request.path.split('/')`, possibly checking length to see if it could be a guid, and returning the first successful `Node.load`'s `id`

mfraezz:
Also, this seems to break file storage -- at least OSFS: 

screen shot 2016-05-02 at 12 22 22

whit537:
### Punchlist

- [ ] fix Travis

whit537:
#### Investigations

- [ ] skim code
- [ ] run code
- [ ] read code

whit537:
I've added @mfraezz's comments to the [punchlist](https://github.com/CenterForOpenScience/osf.io/pull/5445#issuecomment-222708955). Reviewing recent commits ...

alexschiller:
@whit537 Passing on Travis now. I believe everything else on the punchlist was completed in my last couple commits.

tests/test_views.py
```diff
@@ -119,12 +119,44 @@ def test_can_be_anonymous_for_public_project(self, mock_property):
...
+    @mock.patch('framework.sessions.get_node_id')
+    def test_has_private_link_key(self, mock_get_node_id):
+        mock_get_node_id.return_value = self.project
         res = self.app.get(self.project_url, {'view_only': self.link.key})
         assert_equal(res.status_code, 200)
 
+    def test_has_bad_private_link_key_redirects_properly(self):
+        res = self.app.get(self.project_url, {'view_only': 'not_valid_link'})
+        assert_is_redirect(res)
+        res = res.follow(expect_errors=True)
```
mfraezz: Here and other places in this file, the `expect_errors` flag should only need to be present if errors are expected -- that is, if the response status code is in the 4XX-5XX range.

tests/test_views.py
```diff
@@ -155,37 +189,40 @@ def test_prepare_private_key(self):
         assert_equal(res.status_code, 200)
         assert_equal(res.request.GET['key'], self.link.key)
 
-    def test_cannot_access_registrations_or_forks_with_anon_key(self):
+    @mock.patch('framework.sessions.get_node_id')
+    def test_cannot_access_registrations_or_forks_with_anon_key(self, mock_get_node_id):
```
mfraezz: These tests don't seem to cover all the possibly-hidden endpoints. Consider adding more for the other tabs that are hidden (Analytics, Settings, etc.) to ensure the correct error code is raised. This would help prevent regressions.

mfraezz:
Just a couple drive-by comments for now. Looking good 🌴

framework/sessions/__init__.py
```diff
@@ -45,6 +48,27 @@ def add_key_to_url(url, scheme, key):
     parsed_redirect_url = parsed_url._replace(**replacements)
     return urlparse.urlunparse(parsed_redirect_url)
 
+def get_node_id(url):
+    try:
+        # Avoid circular import
+        from website.models import Node
+        for seg in furl.furl(url).path.segments:
+            if len(seg) == 5:  # this will need to change if guids == 5 or 6
```
mfraezz: Maybe change this to `if len(seg) >= 5:` for maintenance ease.

framework/sessions/__init__.py
```diff
@@ -45,6 +48,27 @@ def add_key_to_url(url, scheme, key):
     parsed_redirect_url = parsed_url._replace(**replacements)
     return urlparse.urlunparse(parsed_redirect_url)
 
+def get_node_id(url):
```
mfraezz: I think this will break if a user tries to view a file with a VOL. 

Files can have their own guid, but `if file.node in private_link.nodes`, the link should still be granted view permission.

mfraezz:
A quick pass of functional review indicates that behavior is mostly correct, but files are broken as expected. Back to you, @alexschiller 🎣

mfraezz:
This is currently off the roadmap, at least until DjangOSF goes in. Product will re-evaluate priorities later

PR #5446, opened 2016-04-11
[Feature] Admin Spam Refactor

Ghalko:
## Proposal
Refactor many small issues that are listed in ADM-83

## Repercussions
Cleaner and more security.


admin/spam/views.py
```diff
@@ -1,96 +1,124 @@
...
+    update_admin_log,
+    CONFIRM_HAM,
+    CONFIRM_SPAM,
+    USER_EMAILED,
+)
 from admin.spam.serializers import serialize_comment
 from admin.spam.forms import EmailForm, ConfirmForm
+from admin.spam.templatetags.spam_extras import reverse_spam_detail
 
+STATUS = dict(
```
cwisecarver: @Ghalko My fifteen minutes turned into several hours. I just started looking at this now. I'm going to pick it up again first thing in the morning, fresh. My first note would be to take a look at django's [ChoiceField](https://docs.djangoproject.com/en/1.9/ref/forms/fields/#choicefield). You can make a tuple of tuples in your [model](https://docs.djangoproject.com/en/1.9/ref/models/fields/#field-choices) and give choices there. It's the same concept as what I think you're doing but it's bound to the model and by setting this to a ChoiceField on the ModelForm it automatically selects a widget based on whether multiple select is allowed and creates the right html. You can also use it on multiple field types, Integer, Char, whatever. If you want to get really fancy there's [this](https://django-model-utils.readthedocs.org/en/latest/utilities.html#choices). It's part of a third-party package I've used in the past. That package has a ton of other nice features around model managers that cut out a lot of boiler-plate code.

Ghalko: The model is a modm model in OSF. I'll take a look at the 3rd party code.

cwisecarver: Ah, okay. The third party stuff will only work with django models. I saw the spam/model.py at the bottom of this commit and assumed it was django.

admin/spam/views.py
```diff
@@ -1,96 +1,124 @@
...
-                {'item': self.spam}
+                {'item': spam}
             ).content,
         }
         return super(EmailFormView, self).get_initial()
 
     def form_valid(self, form):
+        message = form.cleaned_data.get('message')
+        email = form.cleaned_data.get('email')
+        update_admin_log(
```
cwisecarver: Would you maybe want to log after you email? If there's an exception sending the mail we will have a log of something that didn't happen.

admin/spam/views.py
```diff
@@ -1,96 +1,124 @@
...
+from admin.base.utils import OSFAdmin
+from admin.common_auth.logs import (
+    update_admin_log,
+    CONFIRM_HAM,
+    CONFIRM_SPAM,
+    USER_EMAILED,
+)
 from admin.spam.serializers import serialize_comment
 from admin.spam.forms import EmailForm, ConfirmForm
+from admin.spam.templatetags.spam_extras import reverse_spam_detail
```
cwisecarver: I saw these template tags. It seems like you could consolidate them into one tag that takes a view name, kwargs, and querystring kwargs. Sorta like this but one that supports more than one qs kwarg: 
```python

from django.utils.http import urlencode

def reverse_querystring(view, urlconf=None, args=None, kwargs=None, current_app=None, query_kwargs=None):
    '''Custom reverse to handle query strings.
    Usage:
        reverse('app.views.my_view', kwargs={'pk': 123}, query_kwargs={'search', 'Bob'})
    '''
    base_url = reverse(view, urlconf=urlconf, args=args, kwargs=kwargs, current_app=current_app)
    if query_kwargs:
        return '{}?{}'.format(base_url, urlencode(query_kwargs)
    return base_url
```
https://gist.github.com/benbacardi/227f924ec1d9bedd242b

admin/spam/views.py
```diff
@@ -1,96 +1,124 @@
...
-        self.spam_id = spam_id
-        self.page = request.GET.get('page', 1)
         try:
-            self.spam = serialize_comment(Comment.load(spam_id))
-        except (AttributeError, TypeError):
-            return HttpResponseNotFound(
-                '<h1>Spam comment ({}) not found.</h1>'.format(spam_id)
+            return super(EmailFormView, self).get(request, *args, **kwargs)
+        except AttributeError:
+            return page_not_found(
```
cwisecarver: I think you want to [return an `HttpResponseNotFound` and let django call `page_not_found`](https://docs.djangoproject.com/en/1.9/topics/http/views/#returning-errors). That gives you the ability to [set a custom 404 view method](https://docs.djangoproject.com/en/1.9/topics/http/views/#customizing-error-views) in your urls file at any point in the future and all of your 404s will call that method instead of having to go track them all down and replace them.

admin/spam/views.py
```diff
@@ -1,96 +1,124 @@
...
-        spam_id = kwargs.get('spam_id', None)
-        self.spam_id = spam_id
-        self.page = request.GET.get('page', 1)
         try:
-            self.spam = serialize_comment(Comment.load(spam_id))
-        except (AttributeError, TypeError):
-            return HttpResponseNotFound(
-                '<h1>Spam comment ({}) not found.</h1>'.format(spam_id)
+            return super(EmailFormView, self).get(request, *args, **kwargs)
+        except AttributeError:
```
cwisecarver: I assume this is attempting to catch any attribute errors in `get_context_data`. Is there a reason you're not just doing that there? You could remove `get()` and `post()` entirely, right?

cwisecarver: If it's because you can't return a response from there you can just raise an HTTP404 error and django will handle it for you.

admin/spam/views.py
```diff
@@ -126,23 +159,18 @@ def get_queryset(self):
         return Comment.find(query).sort(self.ordering)
 
     def get_context_data(self, **kwargs):
-        queryset = kwargs.pop('object_list', self.object_list)
```
cwisecarver: Django's ListView would handle all of this stuff for you if we didn't have to call serialize_comment... We could make a wrapper around the ODM queryset that calls serialize when something's actually accessed and then use that and your view code would end up with get_queryset and not much else.

website/project/spam/model.py
```diff
@@ -85,19 +86,20 @@ def report_abuse(self, user, save=False, **kwargs):
...
-        :param date: Date report submitted
         :param save: Save changes
         :param kwargs: Should include category and message
         :raises ValueError: if user is reporting self
         """
         if user == self.user:
             raise ValueError('User cannot report self.')
         self.flag_spam()
-        report = {'date': datetime.utcnow(), 'retracted': False}
+        date = datetime.utcnow()
```
cwisecarver: Intermediate variable that's only used once.

cwisecarver:
Pass dun. Back to @Ghalko

admin/spam/views.py
Ghalko: Done.

website/project/spam/model.py
Ghalko: Removed

admin/spam/views.py
Ghalko: This is taken care of with the `raise Http404`

admin/spam/views.py
Ghalko: This would then be defined in base as a template tag and could in fact be used for all current extras that I have defined throughout admin.

cwisecarver: Nice!

admin/spam/views.py
Ghalko: Although I think this would be a good idea, it should be for a future PR. Although not as concise as possible and with a bit of repetition I think it is not too terrible. Really this is a request for using the API serializers which I might explore a bit more after the main section of this is done.

cwisecarver: Agreed.

cwisecarver: An alternative would be to skip ODM entirely and build something that talks to the API but acts like a django queryset. I have one of these things. It would have to be modified to be able to write and to work with our api but it wouldn't be hard to modify.

Ghalko:
@cwisecarver Response done. Back to you you.

cwisecarver:
lgtm @Ghalko :shipit:

admin/base/templatetags/base_extras.py
```diff
@@ -0,0 +1,8 @@
+from django.core.urlresolvers import reverse
+from django.utils.http import urlencode
+
+
+def reverse_qs(view, urlconf=None, args=None, kwargs=None, current_app=None, query_kwargs=None):
```
sloria: This is useful, but why is this in `templatetags` ? Would it make more sense in `admin.base.utils`?

Also minor: `query_kwargs` is a bit of a misnomer because you don't pass them as keyword arguments. How about just `query`?

sloria:
Looks good! Just one comment, but it's not blocking: https://github.com/CenterForOpenScience/osf.io/pull/5446#discussion_r61790792 . Feel free to make that change in a follow-up PR.

PR #5448, opened 2016-04-11
Small fixes to landing page

hmoco:
## Purpose
small clean ups and fixes to landing page
## Changes
* Ordering happens at the node list view, to allow for proper fetching of additional nodes without adding duplicates
* Prevent showing help text that implies that the user is looking at the my projects page (including the tooltips for collections and contributors)
* Current user should be a filter in institution's contributors if they are a contibutor to one of those institutions nodes


sloria:

PR #5451, opened 2016-04-11
[WIP] Add Keen support for osf4m [#OSF-5979]

jeffreyliu3230:
## Purpose

Send osf4m logs to Keen server.

## Changes

New feature.

## Side effects

Changes add_poster_by_email function.

## Ticket

https://openscience.atlassian.net/browse/OSF-5979

### DIFF
--- a/website/conferences/views.py
+++ b/website/conferences/views.py
@@ -4,6 +4,7 @@ import json
 import httplib
 import logging
 from datetime import datetime
+from keen.client import KeenClient
 
 from modularodm import Q
 from modularodm.exceptions import ModularOdmException
@@ -27,6 +28,32 @@ from website.conferences.model import Conference
 
 
 logger = logging.getLogger(__name__)
+
+
+# Setup Keen client
+def get_keen_client(project_id, write_key):
+    if project_id and write_key:
+        return KeenClient(
+            project_id=project_id,
+            write_key=write_key,
+        )
+    else:
+        return
+
+
+keenclient = get_keen_client(project_id=settings.KEEN_PROJECT_ID, write_key=settings.KEEN_WRITE_KEY)
+
+
+def _dt_to_iso8601(value):
+    """Helper function for formatting UTC time.
+    """
+    iso8601 = value.isoformat()
+    if iso8601.endswith('+00:00'):
+        iso8601 = iso8601[:-9] + 'Z'  # offset upped to 9 to get rid of 3 ms decimal points
+    else:
+        iso8601 = iso8601[:-3] + 'Z'
+
+    return iso8601
 
 
 @no_auto_transaction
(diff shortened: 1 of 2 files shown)


website/conferences/views.py
```diff
@@ -28,6 +29,22 @@
 
 logger = logging.getLogger(__name__)
 
+# Setup Keen client
+keen.project_id = settings.KEEN_PROJECT_ID
+keen.write_key = settings.KEEN_WRITE_KEY
+
+
+def _dt_to_iso8601(value):
```
icereval: @sloria, do we have a helper anywhere for this?

jeffreyliu3230: @icereval FYI i copied the function from here: https://github.com/CenterForOpenScience/osf.io/blob/develop/api_tests/files/views/test_file_detail.py#L21 and changed it a bit. I would guess there is no helper.

website/conferences/views.py
```diff
@@ -138,6 +155,27 @@ def add_poster_by_email(conference, message):
...
+                "api": "/api/v1/profile/{}/".format(user._id)
+            },
+            "timezone": user.timezone,
+            "fullname": user.fullname,
+            "isActive": user.is_active,
+            "id": user._id
+        },
+        "meeting": message.conference_name,
+        "userCreated": user_created,
+        "timeSubmitted": _dt_to_iso8601(datetime.utcnow())
```
felliott: Can you make the above three attributes under an "osf4m" top-level key?  Could you also add a static field product.id?  `"product": { "id": "osf4m" }`

website/conferences/views.py
```diff
@@ -138,6 +155,27 @@ def add_poster_by_email(conference, message):
     if node_created and user_created:
         signals.osf4m_user_created.send(user, conference=conference, node=node)
 
+    # Send log to Keen
+    keen.add_event('osf4m', {
+        "user": {
+            "username": user.username,
+            "userEmail": message.sender_email,
```
felliott: Please make this a subproperty of the "osf4m" key and rename it to "receivedFrom".

website/conferences/views.py
icereval: Might be worth adding this to `osfHelpers.js` in the vicinity this other 8601 helper.

sloria: Would this work? https://github.com/CenterForOpenScience/osf.io/blob/8225f315d414e89dd72c0f277bca0f8a6823273e/framework/utils.py#L7-L9

felliott: @sloria, it would need millisecond support, but yes.  That function could use some timezone-verification, though.

felliott:
The code is good, but @jeffreyliu3230 discovered that this information is already available through User `system_tags`, so this PR is no longer needed.

PR #5452, opened 2016-04-11
[feature] allow user to select global notification settings [OSF-5748]

leb2dg:
Note: this PR replaces PR #5377 and is piggy backing off of #5733

## Purpose
* New Projects get default notifications set as instantly
* Global project notification settings on project page

screen shot 2016-06-13 at 10 39 38 am
screen shot 2016-06-13 at 10 38 39 am
screen shot 2016-06-13 at 10 39 00 am
screen shot 2016-06-13 at 10 39 07 am

## Changes
Summary across the three PRs this one incorporates:

* Add method that subscribes a user to notifications
* Call that method with signals when a node is first saved with the nodes' creator
* Also call that method when a contributor is added with that contributor
* Add a setting that decides if a notification subscription should be automatically created (by default yes)
* Adjust that setting to False in a special case for NotificationSubscription related tests that don't create the subscriptions by default
* Disable automatic subscriptions on tests that benefit from creating them manually
* Alter text on contributor added, invite, forward invite, and public project emails to explain notifications and link to profile settings to update them
* add global subscription types for comments added and files updated to `USER_SUBSCRIPTIONS_AVAILABLE` in `subscribe_user_to_notifications` check for global settings before defaulting to `email_transactional`
* Update settings heading and add tooltip explaining global preferences settings

## Side effects
* only impacts new project settings

## Ticket


mfraezz:
Code and behavior both look good @leb2dg 🌟 , and h/t @erinspace for all the initial work 👏 

Please add a few more tests for global notification settings, to make sure that users without any global `NotificationSubscriptions` have the correct default subscriptions when creating a new `Node`, and that modifying their global subscription preferences changes the subscription settings of newly created `Nodes`

leb2dg:
@mfraezz I've added a couple more tests to cover those cases. Thanks!

mfraezz:
LGTM. Dropping this in RTM 🌵 

Side Note: May conflict slightly with #5033, which also uses `NotificationSubscriptions`

website/project/model.py
```diff
@@ -1387,6 +1387,34 @@ def set_node_license(self, license_id, year, copyright_holders, auth, save=False
...
+    def subscribe_user_to_notifications(self, user):
+        """ Update the notification settings for the creator or contributors
+
+        :param user: User to subscribe to notifications
+        """
+        from website.notifications.utils import to_subscription_key
+        from website.notifications.utils import get_global_notification_type
+        from website.notifications.model import NotificationSubscription
+
+        events = ['file_updated', 'comments']
```
sloria: Use the constants available to you in `notifications.constants`.

sloria:
Will this require any migration steps?

leb2dg:
Only if we want it to impact projects that already exist. This PR just takes the global settings and applies them to new projects, if there are no global settings it defaults to `email_transactional` which is instant.

tests/test_notifications.py
```diff
@@ -117,6 +116,100 @@ def test_check_user_private_node_child_permissions_excludes_pointers(self):
...
+
+        factories.NotificationSubscriptionFactory(
+            _id=user._id + '_' + 'global_file_updated',
+            owner=user,
+            event_name='global_file_updated'
+        ).add_user_to_subscription(user, 'none')
+
+        node = factories.NodeFactory(creator=user)
+
+        user_subscriptions = [x for x in utils.get_all_user_subscriptions(user)]
```
sloria: Minor: This is the same as `list(utils.get_all_user_subscriptions(user)`.

website/project/model.py
```diff
@@ -1527,6 +1556,9 @@ def save(self, *args, **kwargs):
                 save=True,
             )
 
+            if self.creator and settings.ENABLE_NOTIFICATION_SUBSCRIPTION_CREATION:
+                self.subscribe_user_to_notifications(user=self.creator)
```
sloria: This is OK, but it's reinventing the wheel a bit. We typically use blinker signals for this case. You could emit a signal for `node_created` here, then have a listener that created the subscription. Something like

```python
class Node(...):
    @contrib_added.connect
    @node_created.connect
    def _subscribe_user_to_notifications_listener(...):
        ...
```

sloria: ^ This refactor isn't a blocker; feel free to make this change in a follow-up  PR

website/templates/emails/contributor_added.txt.mako
```diff
@@ -1,7 +1,13 @@
+<%!
+    from website import settings
+%>
+
 Hello ${user.fullname},
 
 ${referrer_name + ' has added you' if referrer_name else 'You have been added'} as a contributor to the project "${node.title}" on the Open Science Framework: ${node.absolute_url}
 
+You will automatically be subscribed to notification emails for comments and updated files for this project. To change your email notification preferences, visit your project or your user settings: ${settings.DOMAIN + "settings/notifications/"}
```
sloria: Is this to spec? Do we need to tell the user this?

sloria: Every time we add a new notification type, we'll need to remember to update this email text.

Can we just add a more generic statement at the end of the email?

```
Sincerely,

.....

To change your email preference, visit ....
```

@saradbowman

website/templates/emails/new_public_project.html.mako
```diff
@@ -8,7 +8,7 @@
         Congratulations on making a public project on the Open Science Framework (OSF)! Now that your project “${project_title}" is public, you’ll be able to take advantage of more OSF features:
         <ul>
             <li><a href="${osf_url}${nid}/files/">The number of downloads of your files will be automatically logged for you</a></li>
-            <li><a href="${osf_url}${nid}/settings/">You can enable other OSF users to comment on your work</a></li>
+            <li><a href="${osf_url}${nid}/settings/">You can enable other OSF users to comment on your work or update your own comment notification settings</a></li>
```
sloria: Also, is this to spec? Seems unnecessary (there's a lot that you can do on the settings page that don't need to mentioned in this email).

website/project/model.py
```diff
@@ -2952,6 +2984,9 @@ def add_contributor(self, contributor, permissions=None, visible=True,
                 while len(user.recently_added) > MAX_RECENT_LENGTH:
                     user.recently_added.pop()
 
+            if contrib_to_add.is_registered and settings.ENABLE_NOTIFICATION_SUBSCRIPTION_CREATION:
```
sloria: The `is_registered` check should probably be moved to the `subscribe_user_to_notifications` method so that the caller doesn't have to do this check every time.

website/project/model.py
```diff
@@ -1527,6 +1556,9 @@ def save(self, *args, **kwargs):
                 save=True,
             )
 
+            if self.creator and settings.ENABLE_NOTIFICATION_SUBSCRIPTION_CREATION:
```
sloria: Just curious: why is it necessary to check if `self.creator`? Looks like we're already assuming that `self.creator` exists here: https://github.com/CenterForOpenScience/osf.io/blob/0512b299781bdfbb89044075f499495c7795129c/website/project/model.py#L1507-L1510

website/notifications/utils.py
```diff
@@ -361,6 +363,21 @@ def get_parent_notification_type(node, event, user):
...
 
+def get_global_notification_type(global_subscription, user):
+    """
+    Given a global subscription (e.g. NotificationSubscription object with event_type
+    'global_file_updated'), find the user's notification type.
+    :param obj global_subscription: modular odm NotificationSubscription object
+    :param obj user: modular odm User object
+    :return: str notification type (e.g. 'email_transactional')
+    """
+    if global_subscription and isinstance(global_subscription, NotificationSubscription):
```
sloria: Why is this necessary? The caller should always pass a `NotificationSubscription` as the first argument, right? If not, this function should probably fail.

sloria:
Pass finished.

Pending a few minor changes and a few spec checks from @saradbowman (see https://github.com/CenterForOpenScience/osf.io/pull/5452#discussion_r61477997 and https://github.com/CenterForOpenScience/osf.io/pull/5452#discussion_r61476645), this should be good to merge.

website/templates/emails/contributor_added.txt.mako
saradbowman: @sloria How about we say to the user,

You will automatically be subscribed to notification emails for updates to this project. To change your email notification preferences, visit your project or your user settings: ${settings.DOMAIN + "settings/notifications/"}

sloria: Sure. @leb2dg  Please make that change in this and the other emails.

website/templates/emails/new_public_project.html.mako
saradbowman: The default setting for commenting is now that other users can comment on your project, so I actually think we should remove this bullet and replace it with: 

<li><a href="${osf_url}${nid}/analytics/">You can see analytics on how many others are accessing your projects</a></li>

I know this is out of scope/not included in the spec, so if you'd like me to put in a separate ticket I'm happy to do so.

website/project/model.py
leb2dg: @erinspace do you have any feedback on these comments?

website/notifications/utils.py
leb2dg: Yeah, the check isn't necessary. I'll remove it.

website/templates/emails/new_public_project.html.mako
leb2dg: @sloria, what would you like done here?

leb2dg:
@mfraezz, this PR contains the most recent spec changes 🎈

website/templates/emails/contributor_added.txt.mako
```diff
@@ -1,7 +1,13 @@
+<%!
+    from website import settings
+%>
+
 Hello ${user.fullname},
 
 ${referrer_name + ' has added you' if referrer_name else 'You have been added'} as a contributor to the project "${node.title}" on the Open Science Framework: ${node.absolute_url}
 
+You will automatically be subscribed to notification emails for updates to this project. To change your email notification preferences, visit your project or your user settings: ${settings.DOMAIN + "settings/notifications/"}
```
mfraezz: From other PR:

>Is this necessarily true? If a user sets their global default to no notifications, this line would probably be confusing. Consider passing a var to the email's template constructor, and adding logic to say either will not be subscribed or will automatically be subscribed depending on their setting.

mfraezz:
Two discrepancies:
* Ticket indicates that the default value for each type of notification should be `Instantly`, but `Replies to your comments` defaults to `Never`. (This will affect tests)
* Language on the ticket makes it seem like the `NOTE` should go inside the `Configure Notification Preferences` above `Default Global Notification Settings` on (possibly only) the user settings page, not in a tooltip. Pinging @saradbowman for clarification.

screen shot 2016-06-06 at 15 56 55

saradbowman:
As usual, I agree with @mfraezz - can we make the NOTE: part persistent text above the Default Global Notification settings text?

nicipfeiffer:
I also agree. The tooltip is covering the text for the settings and not ideal for users who might need to see both pieces of info at once.
How about we model the text similar to how we have it on the project settings + email notifications:
screenshot 2016-06-06 16 33 50

leb2dg:
@saradbowman @nicipfeiffer can you confim the below notes to make sure I understand the changes: 
* the tooltips should be removed

* `NOTE: Regardless of your selected preferences, OSF will continue to provide transactional and administrative service emails.` should be above the *Default Global Notification Settings* an below the *Configure Notification Preferences* header

* `These are default settings for new projects you create or are added to. Modifying these settings will not modify settings on existing projects.` should be below *Default Global Notification Settings*

* What, if any help text should go below *Project Notifications*

mfraezz:
Only the `NOTE` text should be removed; it should be `text-info` and go between *Configure Notification Preferences* and *Default Global etc*. What was previously in the tooltip should stay where it is.

leb2dg:
Ah, got it. Is what is in the Project Notification tooltip okay? I added that text because it was previously the same as the Global Notification Preferences.

leb2dg:
Alright, I think the UI is correct now. Back to you @mfraezz

website/project/model.py
```diff
@@ -1411,6 +1411,36 @@ def set_node_license(self, license_id, year, copyright_holders, auth, save=False
...
 
+    def subscribe_user_to_notifications(self, user):
+        """ Update the notification settings for the creator or contributors
+
+        :param user: User to subscribe to notifications
+        """
+        from website.notifications.utils import to_subscription_key
+        from website.notifications.utils import get_global_notification_type
+        from website.notifications.model import NotificationSubscription
+        from website.notifications.constants import NODE_SUBSCRIPTIONS_AVAILABLE
```
mfraezz: Trivial, but `[this line] as events` obviates the var assignment line immediately below.

website/templates/emails/contributor_added.txt.mako
```diff
@@ -1,7 +1,17 @@
+<%!
+    from website import settings
+%>
+
 Hello ${user.fullname},
 
 ${referrer_name + ' has added you' if referrer_name else 'You have been added'} as a contributor to the project "${node.title}" on the Open Science Framework: ${node.absolute_url}
 
+% if all_global_subscriptions_none:
```
mfraezz: This could be simplified with something like
```
You will ${'not be ' if all_global_subscriptions_none else ''} automatically subscribed [...]
```

website/project/views/contributor.py
```diff
@@ -462,7 +462,7 @@ def send_claim_email(email, user, node, notify=True, throttle=24 * 3600):
 
 
 @contributor_added.connect
-def notify_added_contributor(node, contributor, auth=None, throttle=None):
+def notify_added_contributor(node, contributor, auth=None, throttle=None, all_global_subscriptions_none=None):
```
mfraezz: `all_global_subscriptions_none` appears to be used as a boolean, so it should default to `False` instead of `None`

website/notifications/utils.py
```diff
@@ -371,13 +373,40 @@ def get_parent_notification_type(node, event, user):
...
+
+def check_if_all_global_subscriptions_are_none(user):
+    all_global_subscriptions_none = False
+    user_sunscriptions = get_all_user_subscriptions(user)
+    for user_subscription in user_sunscriptions:
+        if 'global' in user_subscription.event_name:
+            all_global_subscriptions_none = True
+            global_notification_type = get_global_notification_type(user_subscription, user)
+            if global_notification_type != 'none':
+                all_global_subscriptions_none = False
```
mfraezz: This conditional branch could just `return False`

website/project/views/contributor.py
mfraezz: Also, as a kwarg it's somewhat big and clunky, albeit semantically explicit. I can't complain without suggesting something better, so ¯\\\_(ツ)_/¯

mfraezz:
A few non-blocking nit-picky comments, otherwise LGTM 🌟 

Dropping this in RTM, feel free to address those small things without needing to send it back to me.

leb2dg:
Moving the text out of the tooltips has been deferred due to the difficulty so this should be ready to go 👍

website/settings/defaults.py
```diff
@@ -69,6 +69,7 @@ def parent_dir(path):
 CONFIRM_REGISTRATIONS_BY_EMAIL = True
 ALLOW_REGISTRATION = True
 ALLOW_LOGIN = True
+ENABLE_NOTIFICATION_SUBSCRIPTION_CREATION = True
```
sloria: Minor: This is a bit of a misnomer; in addition to subscribing a user on project creation, users are also subscribed on when they are added as a contributor. How about something like `SUBSCRIBE_CONTRIBUTORS_TO_NOTIFICATIONS`?

website/static/js/notificationsTreebeard.js
```diff
@@ -76,6 +76,11 @@ function displayParentNotificationType(item){
     return '';
 }
 
+var popOver = function(element, isInit) {
```
sloria: Minor misnomer: how about `tooltipConfig`?

website/templates/emails/contributor_added.txt.mako
```diff
@@ -1,7 +1,13 @@
+<%!
+    from website import settings
+%>
+
 Hello ${user.fullname},
 
 ${referrer_name + ' has added you' if referrer_name else 'You have been added'} as a contributor to the project "${node.title}" on the Open Science Framework: ${node.absolute_url}
 
+You will ${'not be ' if all_global_subscriptions_none else ''} automatically subscribed to notification emails for updates to this project. To change your email notification preferences, visit your project or your user settings: ${settings.DOMAIN + "settings/notifications/"}
```
sloria: @saradbowman Is this to spec? If the user explicitly unsubscribed globally, this sentence is probably unnecessary. It's kinda like a dinner menu that says, "Your hamburger will not be served with cheese."

website/templates/emails/forward_invite.txt.mako
```diff
@@ -14,7 +18,7 @@ You have been added by ${referrer.fullname} as a contributor to the project "${n
 
 ${claim_url}
 
-Once you have set a password, you will be able to make contributions to ${node.title}.
+Once you have set a password, you will be able to make contributions to ${node.title}. You will automatically be subscribed to notification emails for updates to this project. To change your email notification preferences, visit your project or your user settings: ${settings.DOMAIN + "settings/notifications/"}
```
sloria: Minor: as per COSDev, use singlequotes for strings.

website/project/model.py
```diff
@@ -1411,6 +1411,35 @@ def set_node_license(self, license_id, year, copyright_holders, auth, save=False
...
+                event_id = to_subscription_key(target_id, event)
+                global_event_id = to_subscription_key(user._id, 'global_' + event)
+                global_subscription = NotificationSubscription.load(global_event_id)
+
+                subscription = NotificationSubscription.load(event_id)
+                if not subscription:
+                    subscription = NotificationSubscription(_id=event_id, owner=self, event_name=event)
+                if global_subscription:
+                    global_notification_type = get_global_notification_type(global_subscription, user)
+                    subscription.add_user_to_subscription(user, global_notification_type)
```
sloria: What if the global notification type is `none`? Don't we want to prevent this call?

sloria: ...unless `add_user_to_subscription` is responsible for handling `none`, which on a quick glance, it does.

sloria:
osf___notifications

* How about just "Default Notification Settings"? @saradbowman 
* I realize this might be out of scope, but the coloring on "No configured projects" is inappropriate. it should just be grey (`.text-muted`).

Neither of these are blockers to merging this.

sloria:
The tooltip is getting cut off.

osf___notifications

This is worth looking into <=15 minutes. Unless @saradbowman objects, I suggest we defer adding these tooltips until a future release.

website/static/js/notificationsTreebeard.js
```diff
@@ -118,11 +123,30 @@ function ProjectNotifications(data) {
...
+                                m('p', [
+                                    m('b', item.data.node.title + ':  '),
+                                    item.data.node.title === 'Default Global Notification Settings' ?
+                                        m('span[class="fa fa-info-circle"]', {
+                                            'data-toggle': 'tooltip',
+                                            'title':globalNotificationsMessage,
+                                            'config': popOver,
+                                            'data-placement': 'bottom'
+                                        }) :
+                                        m('span[class="fa fa-info-circle"]', {
```
sloria: There's some duplication that can be removed here:

```js
m('span[class="fa fa-info-circle"]', {
    'data-toggle': 'tooltip',
    'title': item.data.node.title === 'Default...' ? globalNotificationsMessage : projectNotificationsMessage,
    'config': popOver,
    'data-placement': 'bottom'
})
```

leb2dg:
I can't reproduce the tooltip getting cut off in chrome, firefox, or safari. Did the tooltip display correctly at any point?

website/static/js/notificationsTreebeard.js
sloria: Or, even better: add the help text to this function: https://github.com/CenterForOpenScience/osf.io/blob/6e449d62a128163a6e9519a5d51ca2ed28ccf503/website/notifications/utils.py#L403 and use it here.

```python
    return [
        {
            'node': {
                'id': user._id,
                'title': 'Default Global Notification Settings',
                'help': 'These are the default settings...'
            },
        },
```

website/project/model.py
```diff
@@ -2991,7 +3027,8 @@ def add_contributor(self, contributor, permissions=None, visible=True,
             if save:
                 self.save()
 
-            project_signals.contributor_added.send(self, contributor=contributor, auth=auth)
+            all_global_subscriptions_none = check_if_all_global_subscriptions_are_none(contributor)
```
sloria: It would be good if we could move as much notification code out of the project model as possible. This check, for example, can happen in the listener (https://github.com/CenterForOpenScience/osf.io/blob/6e449d62a128163a6e9519a5d51ca2ed28ccf503/website/project/views/contributor.py#L465) rather than passing it to the signal.

website/project/model.py
```diff
@@ -1411,6 +1411,35 @@ def set_node_license(self, license_id, year, copyright_holders, auth, save=False
         if save:
             self.save()
 
+    def subscribe_user_to_notifications(self, user):
+        """ Update the notification settings for the creator or contributors
+
+        :param user: User to subscribe to notifications
+        """
+        from website.notifications.utils import to_subscription_key
```
sloria: As suggested in a previous comment, it would be good to move as much notifications-specific code out of the project model. All these circular imports smell of mixed concerns.

A common solution we've been using is signals. How about creating listeners for the `contributor_added` (already exists) and `project_created` (doesn't yet exist) signals? 

The code will look something like:

```python
# website.notifications.listeners

from website.project.signals import project_created, contributor_added

def subscribe_user_to_notifications(node, user):
    # ...

@project_created.connect
def subscribe_creator(node):


@contributor_added.connect
def subscribe_contributor(node, contributor, *args, **kwargs):
    subscribe_user_to_notifications(node, contributor)
```

`subscribe_user_to_notifications` will essentially replace the method you have here.

sloria: One additional benefit of this approach is that it would allow us to get rid of the `ENABLE_NOTIFICATION_SUBSCRIPTION_CREATION` setting, which seems to exist solely for testing.

Our base test class already has a way to disconnect signals, so you'd be able to do something like:

```python
class NotificationTestCase(OsfTestCase):
    DISCONNECTED_SIGNALS = {
        contributor_added: [notify_added_contributor, subscribe_contributor]
        project_created: [subscribe_creator]
    }
```

sloria:
Minor: Could use a tad more padding above the "NOTE: ..." text. See other panels for reference.

osf___notifications

sloria:
> Did the tooltip display correctly at any point?

No, it hasn't rendered correctly for me locally (on either Chrome or Firefox). That said, I have seen it render correctly on both your and Erin's computers... so might be a personal problem; I'll look into it.

sloria:
Pass finished.

mfraezz:
>I have seen it render correctly on both your and Erin's computers

Mine as well.

tests/base.py
```diff
@@ -335,6 +336,30 @@ class AdminTestCase(DbTestCase, DjangoTestCase, UploadTestCase, MockRequestTestC
...
+    """
+    DISCONNECTED_SIGNALS = {
+        # disconnect signals so that add_contributor does not send "fake" emails in tests
+        contributor_added: [subscribe_contributor],
+        project_created: [subscribe_creator]
+    }
+
+    def setUp(self):
+        super(NotificationTestCase, self).setUp()
+        for signal in self.DISCONNECTED_SIGNALS:
```
sloria: Minor: This should be unnecessary because this is handled by `OsfTestCase`.

website/notifications/utils.py
```diff
@@ -371,13 +373,68 @@ def get_parent_notification_type(node, event, user):
...
+    for notification_type in constants.NOTIFICATION_TYPES:
+        if user in getattr(global_subscription, notification_type):
+            return notification_type
+
+
+def check_if_all_global_subscriptions_are_none(user):
+    all_global_subscriptions_none = False
+    user_sunscriptions = get_all_user_subscriptions(user)
+    for user_subscription in user_sunscriptions:
+        if 'global' in user_subscription.event_name:
```
sloria: Minor: It's theoretically possible for a non-global notification to have "global" in its name. To prevent possible clashing, it's safer to do `if user.subscription.startswith('global_')`.

This reminds me: it would be worth documenting the significance of "global_" in the field documentation (by convention, a code comment above the field).

website/templates/emails/forward_invite.txt.mako
```diff
@@ -14,7 +18,7 @@ You have been added by ${referrer.fullname} as a contributor to the project "${n
 
 ${claim_url}
 
-Once you have set a password, you will be able to make contributions to ${node.title}.
+Once you have set a password, you will be able to make contributions to ${node.title}. You will automatically be subscribed to notification emails for updates to this project. To change your email notification preferences, visit your project or your user settings: ${settings.DOMAIN + 'settings/notifications/'}
```
sloria: Minor: To be more concise: "You will automatically be subscribed to notification emails for this project."

website/templates/emails/invite.txt.mako
```diff
@@ -1,10 +1,14 @@
...
+%>
+
 Hello ${fullname},
 
 You have been added by ${referrer.fullname} as a contributor to the project "${node.title}" on the Open Science Framework. To set a password for your account, visit:
 
 ${claim_url}
 
-Once you have set a password, you will be able to make contributions to ${node.title} and create your own projects.
+Once you have set a password, you will be able to make contributions to ${node.title} and create your own projects. You will automatically be subscribed to notification emails for updates to this project. To change your email notification preferences, visit your project or your user settings: ${settings.DOMAIN + "settings/notifications/"}
```
sloria: See https://github.com/CenterForOpenScience/osf.io/pull/5452/files#r66835906

website/notifications/constants.py
```diff
@@ -4,7 +4,9 @@
 }
 
 USER_SUBSCRIPTIONS_AVAILABLE = {
-    'comment_replies': 'Replies to your comments'
+    'global_comment_replies': 'Replies to your comments',
```
sloria: This reminds me: We should have a field validator for `NotificationSubscription.event_name` and `NotificationDigest.event` that verifies that the value is one of the keys in `USER_SUBSCRIPTIONS_AVAILABLE` or `NODE_SUBSCRIPTIONS_AVAILABLE`.

But that's out of the scope of this feature...I'll create a ticket.

sloria: Up for grabs https://openscience.atlassian.net/browse/OSF-6499

website/notifications/utils.py
```diff
@@ -175,7 +175,7 @@ def get_configured_projects(user):
         # If the user has opted out of emails skip
         node = subscription.owner
 
-        if not isinstance(node, Node) or (user in subscription.none and not node.parent_id):
+        if not isinstance(node, Node) or (user in subscription.none and not node.parent_id) or node.title == 'Bookmarks':
```
sloria: You can check `node.is_bookmark_collection` rather than comparing against the title.

website/notifications/utils.py
```diff
@@ -327,6 +327,8 @@ def serialize_event(user, subscription=None, node=None, event_description=None):
         event_type = event_description
     if node and node.node__parent:
         notification_type = 'adopt_parent'
+    elif 'global' in event_type:
```
sloria: Same here: `event_type.startswith('global_')` is a tad bit safer.

website/notifications/listeners.py
```diff
@@ -0,0 +1,12 @@
+from website.notifications.utils import subscribe_user_to_notifications
+from website.project.signals import contributor_added, project_created
+from website.project.views.contributor import notify_added_contributor
+
+@project_created.connect
```
sloria: 👍

website/templates/emails/contributor_added.txt.mako
saradbowman: @sloria I'd prefer, "You will not receive notification emails for updates to this project. To change your email notification preferences, visit your project or user settings"

sloria:
🐛  The global/default notification settings don't seem to be taking effect.

## Steps

1. In user notification settings, set all event types to "Instantly"
2. Create a new project
3. Go to project notification settings

## Expected 

All notification types are set to "Instantly"

## Actual

All notification types are set to "Never"

## Possible cause

I wonder if your new `listeners.py` file is never getting imported and therefore the signals never get connected. You might need to import the module here: https://github.com/CenterForOpenScience/osf.io/blob/e5d6f2e64a1dad279cab66f14b54c98de2868e7c/website/app.py#L28-L31

sloria:
There is not enough padding around the help text.

osf___sub__settings


Looks like the `.panel-body` div is missing here: https://github.com/CenterForOpenScience/osf.io/blob/e5d6f2e64a1dad279cab66f14b54c98de2868e7c/website/templates/project/settings.mako#L381

After adding div.panel-body

osf___sub__settings

sloria:
Pass finished. Once https://github.com/CenterForOpenScience/osf.io/pull/5452#issuecomment-225663829 is resolved, this is good to go. Feel free to also address the nitpicks, but they're not blockers.

website/templates/emails/contributor_added.txt.mako
```diff
@@ -1,7 +1,13 @@
+<%!
+    from website import settings
+%>
+
 Hello ${user.fullname},
 
 ${referrer_name + ' has added you' if referrer_name else 'You have been added'} as a contributor to the project "${node.title}" on the Open Science Framework: ${node.absolute_url}
 
+You will ${'not receive ' if all_global_subscriptions_none else 'be automatically subscribed to '} notification emails for updates to this project. To change your email notification preferences, visit your project or your user settings: ${settings.DOMAIN + "settings/notifications/"}
```
sloria: Minor: I think Sara's suggested text is "You will not receive notification emails for this project." 

I can make this change when I merge.

website/templates/profile/notifications.mako
```diff
@@ -54,6 +54,10 @@
         </div>
         <div class="panel panel-default">
             <div class="panel-heading clearfix"><h3 class="panel-title">Configure Notification Preferences</h3></div>
+            <div class="panel-body">
+                <div class="help-block" style="padding-left: 15px">
```
sloria: Minor: With `panel-body` added, the extra padding should no longer be necessary. I can remove this when I merge.

leb2dg: The padding aligns the edge of the help text with the rest of the body. The `panel-body` just adds padding to the top.

website/templates/emails/contributor_added.txt.mako
leb2dg: >
@sloria I'd prefer, "You will not receive notification emails for updates to this project. To change your email notification preferences, visit your project or user settings"
>

website/templates/profile/notifications.mako
leb2dg: Just kidding. It does look a little weird without the padding though so your call.

website/templates/emails/contributor_added.txt.mako
sloria: Ahhhh fair enough. I don't think she'll mind that I changed it. What she doesn't know can't hurt her. Shh...

saradbowman: I'm right here!!!

PR #5453, opened 2016-04-11
fix bugs in FormattableDate JS tests

felliott:
## Purpose

Our karma tests for FormattableDates had two consistent failures.  One was a bad test, the other was a bug in the underlying library.  Both tests happen to pass on Travis because the local timezone on their server is GMT.  These tests will fail in any other timezone.

## Changes

The first error was caused by FormattableDate assuming that date-only input
represents midnight in local-time.  Our test was comparing the return
value to midnight in UTC, which means the test would fail anywhere but
in +0.  We should probably fix up FormattableDate to consistently return
midnight UTC, but that will require a lot more verification of our
consuming code.  For now, add a new test function that only verifies
the date component of the returned datetime.

The second bug was caused by moment.js not recognizing the '+00' method
of indicating timezone offset.  moment.js has now been pinned to
v2.10.6, and the tests updated to test both positive and negative
offsets.

## Side effects

Upgrading moment.js from `2.9.0` to `2.10.6` has the potential to cause errors elsewhere.  I've looked through the code and didn't see any obvious problems, I didn't encounter any JS errors or misformatted dates while running the code locally.  QA should check share events, activity logs, and wiki timestamps.

## Ticket

[#OSF-5950|https://openscience.atlassian.net/browse/OSF-5950]

I couldn't reproduce the S3 issues locally, so that is not part of this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "markdown-it-video": "~0.2.1",
     "mime-types": "~2.0.12",
     "mithril": "0.2.0",
-    "moment": "^2.9.0",
+    "moment": "^2.10.6",
     "morgan": "^1.5.1",
     "object-assign": "^2.0.0",
     "pikaday": "^1.3.2",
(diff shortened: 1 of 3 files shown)


mfraezz:
Also unable to reproduce the S3 issues locally, confirmed that `karma` now passes. 

LGTM :shipit:

PR #5454, opened 2016-04-11
[Ready to Review] hotfix/myprojects-null-checking

pattisdr:
# Purpose

Getting 'TypeError Cannot read property 'contributors' of undefined'.  API not returning contributor information.

Also handles when user makes project stored in Discover Public Projects private.

# Changes

Use lodash.get for dashboard null checking

### DIFF
--- a/website/static/js/myProjects.js
+++ b/website/static/js/myProjects.js
@@ -268,11 +268,12 @@ function _formatDataforPO(item) {
     if (contributorsData){
         item.embeds.contributors.data.forEach(function(c){
             var attr;
-            if (c.embeds.users.data) {
-                attr = c.embeds.users.data.attributes;
+            var users = lodashGet(c, 'embeds.users.data', null);
+            if (users) {
+                attr = users.attributes;
             }
             else {
-                attr = c.embeds.users.errors[0].meta;
+                attr = users.errors[0].meta;
             }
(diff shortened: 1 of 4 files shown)


sloria:
I'm seeing an error when loading All my projects:

osf___my_projects

I've confirmed that this error doesn't occur on `master`.

website/static/js/myProjects.js
```diff
@@ -263,14 +264,16 @@ function _formatDataforPO(item) {
...
             var attr;
-            if (c.embeds.users.data) {
-                attr = c.embeds.users.data.attributes;
+            var users = lodashGet(c, 'embeds.users.data', null);
+            if (users) {
+                attr = users.attributes;
             }
             else {
-                attr = c.embeds.users.errors[0].meta;
+                attr = users.errors[0].meta;
```
sloria: The error described in https://github.com/CenterForOpenScience/osf.io/pull/5454#issuecomment-208941401 seems to be happening here. We get to this branch when users are not embedded and there are no errors.

This should fix it:


```js
            var users = lodashGet(c, 'embeds.users.data', null);
            if (users) {
                if (users.errors) {
                    attr = users.errors[0].meta;
                } else {
                    attr = users.attributes;
                }
            }
            if (attr) {
                item.contributors += attr.full_name + ' ' + attr.middle_names + ' ' + attr.given_name + ' ' + attr.family_name + ' ' ;
            }
```

sloria: I'll make this change when I merge.

PR #5455, opened 2016-04-12
[RFR] Limit Varnish banning to once per url

cwisecarver:
## Purpose
Limit the banning of each api v2 url to once per request. So if the Node xyz12 is saved 7,123 times during one request it will be banned from Varnish once.

## Changes
- Remove passing fields_changed because as a list it's unhashable and it wasn't being used anyway
- Changed enqueue_postcommit_task to take a tuple comprised of a function and a tuple of arguments
- Changed the save listener to call enqueue_postcommit_task with a function and the instance of the object saved

## Side effects
- Could completely break decaching in varnish.
    - I've functionally tested with a debugger attached in both the API and the OSF

### DIFF
--- a/framework/postcommit_tasks/handlers.py
+++ b/framework/postcommit_tasks/handlers.py
@@ -30,12 +30,7 @@ def postcommit_after_request(response, base_status_error_code=500):
     return response
 
 def enqueue_postcommit_task(function_and_args):
-    try:
-        postcommit_queue().add(function_and_args)
-    except RuntimeError:
-        # if we can't add to the queue just log and run it
-        logger.error('Couldn\'t add {} to the postcommit queue, running syncronously'.format(function_and_args))
-        function_and_args[0](*function_and_args[1])
+    postcommit_queue().add(function_and_args)
 
 
 handlers = {
--- a/framework/postcommit_tasks/handlers.py
+++ b/framework/postcommit_tasks/handlers.py
@@ -22,7 +22,6 @@ def postcommit_after_request(response, base_status_error_code=500):
         return response
     try:
         if postcommit_queue():
-            pcq = postcommit_queue()
             threads = [gevent.spawn(func, *args) for func, args in postcommit_queue()]
             gevent.joinall(threads)
     except AttributeError:
--- a/framework/postcommit_tasks/handlers.py
+++ b/framework/postcommit_tasks/handlers.py
@@ -18,7 +18,7 @@ def postcommit_before_request():
 
 def postcommit_after_request(response, base_status_error_code=500):
     if response.status_code >= base_status_error_code:
-        _local.postcommmit_queue = set()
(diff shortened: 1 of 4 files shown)


framework/postcommit_tasks/handlers.py
```diff
@@ -1,38 +1,41 @@
...
-                task()
+            threads = [gevent.spawn(func, *args) for func, args in postcommit_queue()]
+            gevent.joinall(threads)
     except AttributeError:
         if not settings.DEBUG_MODE:
             logger.error('Post commit task queue not initialized')
     return response
 
-def enqueue_postcommit_task(f):
+def enqueue_postcommit_task(function_and_args):
```
chrisseto: `def enqueue_postcommit_task(func, *args, **kwargs):`?

framework/postcommit_tasks/handlers.py
```diff
@@ -1,38 +1,41 @@
...
     try:
-        if f not in postcommit_queue():
-            postcommit_queue().append(f)
+        postcommit_queue().add(function_and_args)
     except RuntimeError:
-        f()
```
chrisseto: When does this raise a `RuntimeError`?

framework/postcommit_tasks/handlers.py
```diff
@@ -1,38 +1,41 @@
...
 def postcommit_after_request(response, base_status_error_code=500):
     if response.status_code >= base_status_error_code:
-        _local.postcommmit_queue = []
+        _local.postcommmit_queue = set()
         return response
     try:
         if postcommit_queue():
-            for task in postcommit_queue():
-                task()
+            threads = [gevent.spawn(func, *args) for func, args in postcommit_queue()]
```
chrisseto: Could order ever possibly matter here?

Does this work locally when running without gevent?

cwisecarver: I don't think order should ever matter. There's one URL for each object. The set is a list of every unique thing that was saved during the request so they should all be removed. 

It won't work without gevent but gevent is in the base requirements.txt now so it shouldn't need to.

framework/postcommit_tasks/handlers.py
```diff
@@ -1,38 +1,36 @@
...
     return _local.postcommit_queue
 
 def postcommit_before_request():
-    _local.postcommit_queue = []
+    _local.postcommit_queue = set()
 
 def postcommit_after_request(response, base_status_error_code=500):
     if response.status_code >= base_status_error_code:
-        _local.postcommmit_queue = []
+        _local.postcommmit_queue = set()
```
sloria: Typo here: `postcommit_queue`

cwisecarver: Good catch @sloria. That'll teach me to use made up words as variable names.

PR #5456, opened 2016-04-12
[feature] Remove Addon-related backrefs

mfraezz:
## Purpose
Remove addon related backrefs

#### QA Notes
Things to test:

1. Previously configured addons still work and display properly
    * on the user settings page
    * on the node settings page
2. Connecting and listing new (and combinations of new/old) addons still works properly
    * on the user settings page
    * on the node settings page

Full addon regressions are probably unnecessary, though it is theoretically possible that a previously-connected model will be changed but not updated in memory. See Side Effect \# 1. 

## Changes
* backrefs removed
* tests fixed
* `test_addons` call added to `test` invoke task

## Side effects
1) `.get_addons` or similar now no longer return the same object-in-memory (i.e. the same objects are at different locations on the stack). Objects accessed by a backref are cached due to the way MODM handles these things, but querysets returned by a `modularodm.Q(uery)` are not. Anything that requires these objects to be the same (e.g. mocking) will fail, but this is unlikely to affect anything aside from tests.
2) Conflicts with `release/0.67.0` expected (S3 and GH refactors). **HOLD** until then.

## Ticket
[\[OSF-5914\]](https://openscience.atlassian.net/browse/OSF-5914)

### DIFF
--- a/tests/test_addons_oauth.py
+++ b/tests/test_addons_oauth.py
@@ -96,14 +96,15 @@ class TestNodeSettings(OsfTestCase):
             external_account=self.external_account,
             user=self.user
         )
+        self.user_settings.reload()
 
         assert_equal(
             self.node_settings.external_account,
             self.external_account
         )
         assert_equal(
-            self.node_settings.user_settings,
-            self.user_settings
+            self.node_settings.user_settings._id,
+            self.user_settings._id
         )
         assert_in(
             self.project._id,
@@ -118,6 +119,12 @@ class TestNodeSettings(OsfTestCase):
             external_account=self.external_account,
             user=self.user
         )
+        self.user_settings.reload()
+        assert_equal(
+            self.user_settings.oauth_grants,
+            {self.project._id: {self.external_account._id: {}}}
+        )
+
         self.user_settings.revoke_oauth_access(self.external_account, auth=Auth(self.user))
         self.user_settings.reload()
 
--- a/tests/test_registrations/test_archiver.py
+++ b/tests/test_registrations/test_archiver.py
@@ -536,6 +536,13 @@ class TestArchiverTasks(ArchiverTestCase):
                 )
                 patch.start()
                 patches.append(patch)
+                n_patch = mock.patch.object(
+                    n,
+                    'get_addon',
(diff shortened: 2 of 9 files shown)


tests/test_addons_oauth.py
```diff
@@ -118,6 +119,12 @@ def test_revoke_auth(self, mock_decorator, mock_deauth):
             external_account=self.external_account,
             user=self.user
         )
+        self.user_settings.reload()
+        assert_equal(
+            self.user_settings.oauth_grants,
+            {self.project._id: {self.external_account._id: {}}}
```
brianjgeiger: Is `external_account.id == {}` a function of it being a mocked test? (The real question I'm asking is: is this right? I mean, I presume so, but it looks a little odd)

tests/test_registrations/test_archiver.py
```diff
@@ -536,6 +536,13 @@ def test_archive_success_with_components(self):
                 )
                 patch.start()
                 patches.append(patch)
+                n_patch = mock.patch.object(
```
brianjgeiger: For the tests that you had to modify their mocking with, do you have contentious tests that will test both the positive and negative conditions to show that it's not just covering up the failing behavior but is definitely working? Not that I'm suspicious of Mock or anything.

brianjgeiger:
Seems fine. Let me know about the couple of items, especially the if the tests cover enough. Also, let QA know in the Purpose section how they should focus their testing (user settings page, node settings page, add and remove addons and etc) :penguin:

tests/test_addons_oauth.py
mfraezz: This is right. `oauth_grants` is a dict of the form
```python
{
    <node_id> : {
        <externalaccount_id> : {
            # Optional metadata -- typically folder info
            <meta_name> : <meta_val>,
            [...]
        },
        [...]
    },
    [...]
}
```

tests/test_registrations/test_archiver.py
brianjgeiger: Discussed in person; don't worry about this one.

PR #5457, opened 2016-04-12
[HotFix]RegistrationSerializer

KAsante95:
## Purpose
Display the correct files link in the registration serializer.

## Changes
Because both RegistrationRegistrationsList and RegistrationProvidersList had the same view name, 'registration-providers', I changed the view name for RegistrationRegistrationsList to 'registration-registrations'.

## Side effects

Possibly a broken path in the serializer.

## Ticket

https://openscience.atlassian.net/browse/OSF-5892

### DIFF
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -24,15 +24,27 @@ from api.base.serializers import RelationshipField, TargetField
 def sort_multiple(fields):
     fields = list(fields)
     def sort_fn(a, b):
-        while fields:
-            field = fields.pop(0)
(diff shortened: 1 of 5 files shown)


brianjgeiger:
This looks correct and shouldn't cause any problems.

icereval:
thx

sloria:
This looks to have a merge of develop in it. To prevent any unwanted changes from getting into master, I've made this change in 53b339c7aff1dba88b937692cb89e4ccc605ed86 . Thanks for the fix, @KAsante95

PR #5458, opened 2016-04-12
Fix comments not saving

Ghalko:
## Purpose

To fix commenting on prereg submissions for admin users.

## Changes

Removed `deepcopy` copy and used given data. `deepcopy` was not actually copying the dictionary correctly.

## Side effects
Commenting can actually happen

## Ticket

https://openscience.atlassian.net/browse/ADM-86


sloria:

PR #5335, opened 2016-03-30
[hotfix] Registration initiated log entries store wrong node [OSF-4922]

chennan47:
<b>Purpose</b>
Fix the registration logs redirecting to the registration instead of the original project. 
Solves https://openscience.atlassian.net/browse/OSF-4922.

PR #5462, opened 2016-04-12
[Ready to Review] Add registration info to params

pattisdr:
# Purpose

The migration in https://github.com/CenterForOpenScience/osf.io/pull/5335 drops references to registrations in registration logs.  Also, params['node'] points to the registration in some logs, and the node in others.

The overall goal is consistency among registration-related logs.  params['node'] will be set to `registered_from`.  params['registration'] will be the registration.  The later migration in https://github.com/CenterForOpenScience/osf.io/pull/4950 will add `original_node` fields (set equal to params['node']) and `node` fields, initially set equal to the nodes on the logs' backrefs, and afterwards, pertaining to the node the log resides on. 

# Changes
- Adds params['registration'] for all registration-related logs.
- Changes params['node'] to point to registered_from instead of registration.
- Updates script migrate_registration_logs.py to migrate past logs.

**Example**
Node is ABC
Registration is DEF
original_node and node fields are added in https://github.com/CenterForOpenScience/osf.io/pull/4950.  `original_node` is the node the log originally belonged to and `node` is where the log currently belongs. On a project, its original_node and node_fields will be the same.  But if I register that project, `original_node` will point to the project, and `node` will be the registration.

```
Log 123 on Node ABC

action = registration_initiated
original_node = ABC 
node = ABC 
params = {
   "node": "ABC",
   "registration": "DEF",
   "registration_approval": "sdfjkasf..."
}
```

Working off of staging DB before https://github.com/CenterForOpenScience/osf.io/pull/5335 and https://github.com/CenterForOpenScience/osf.io/pull/4950 migrations
### Before running updated scripts.migrate_registration_logs
screen shot 2016-04-13 at 2 26 05 pm
### After: 
screen shot 2016-04-13 at 7 15 11 pm


website/project/model.py
```diff
@@ -906,7 +906,7 @@ def registered_user_id(self):
 
     @property
     def registered_from_id(self):
-        """The ID of the user who registered this node if this is a registration, else None.
+        """The ID of the node that was registered, else None.
```
sloria: Good catch

sloria:
@pattisdr Changes look good. As discussed in person, the migration script only needs to handle *production* data, not incorrectly-migrated data on staging.

QA: Be aware that registration logs created before this PR's migration is run may link to the incorrect node. All logs created after the script is run should be correct. @nicipfeiffer

scripts/migrate_registration_logs.py
```diff
@@ -1,36 +1,75 @@
...
+        Q('action', 'eq', 'embargo_approved') |
+        Q('action', 'eq', 'embargo_completed') |
+        Q('action', 'eq', 'embargo_cancelled')
     )
     return logs
 
 
+def get_registered_from(registration):
+    """
+    Gets node registration was registered from.  Handles deleted registrations where registered_from is null.
```
sloria: :+1: This docstring is helpful.

pattisdr:
@slora, last commit inserts id as a string when clone is being created.  Thanks @chrisseto.

scripts/migrate_registration_logs.py
```diff
@@ -48,4 +87,4 @@ def main(dry_run):
     dry_run = 'dry' in sys.argv
```
sloria: Minor: it looks like the convention is to use `--dry` instead of `dry` (based on the other scripts in https://openscience.atlassian.net/browse/OSF-5986).

scripts/migrate_registration_logs.py
```diff
@@ -1,36 +1,75 @@
...
+def get_registered_from(registration):
+    """
+    Gets node registration was registered from.  Handles deleted registrations where registered_from is null.
+
+    """
+    if registration.registered_from:
+        return registration.registered_from_id
+    else:
+        first_log = db['node'].find_one({'_id': registration._id})['logs'][0]
+        return NodeLog.load(first_log).params['node']
```
sloria: Running on a recent dump of prod, I get a `KeyError: 'node'` here. It appears that there are projects in the database whose first log doesn't contain `node`, but they do contain a `project` key.

sloria: Based on 638f73167fb, there was a time when we didn't add the `node` param for top-level projects. So you can use `params['project']` in those cases.

sloria:
Will the migration work if it is run more than once, i.e. if a log has already been migrated, will it break if the script is run again?

pattisdr:
I haven't thought much about it, but I would think it would work.  It would be very slow though.

If the migration has already been run on a log, and I run it again, it checks to see if params_node.is_registration.  It should not be a registration so it will move to the else branch and set the params['registration'] by getting it off registration approval.  Fetching registrations from nodes is a lot slower than getting nodes from registrations. 

Maybe there should be a check that skips the log if both params['node'] is a node and params['registration'] is a registration.

sloria:
Ah, yeah, I noticed the script was running very slowly because of this line in the script:

```python
            log.params['registration'] = RegistrationApproval.load(log.params['registration_approval_id'])._get_registration()._id
```

`_get_registration()` just does a query for 

```python
return Node.find_one(Q('registration_approval', 'eq', self))
```

Adding an index to `registration_approval` makes this run *much* faster.

pattisdr:
Commit https://github.com/CenterForOpenScience/osf.io/pull/5462/commits/8426ef9cbb73f96a378b0127f935f3a75f250ad8 might fix some registrations having zero logs.

sloria:
Ok, reviewed with @pattisdr and @chrisseto . The steps for migration will need to be:

- checkout old code: `git checkout 0.66.46` (or latest hotfix)
- `python -m scripts.ensure_backrefs`
- `git checkout master`
- `python -m scripts.migrate_registration_logs`
- `python -m scripts.migrate_backrefs_between_nodes_and_nodelogs`

sloria:
Running the ensure_backref code locally now. If that works, I'll add that code to this PR and merge.

pattisdr:
@sloria, @chrisseto, ran ensure_backrefs before running `scripts.migrate_registration_logs` and `scripts.migrate_backrefs_between_nodes_and_nodelogs`.  

Post-migration on local copy of staging, there are now 0 nodes without logs, and all params['node'] point to nodes and params['registrations'] point to registrations!

pattisdr:
```
[__main__]  INFO: For log action project_registered, 14241/14241 of params[node] are nodes and 14241/14241 of params[registration] are registrations
[__main__]  INFO: For log action registration_initiated, 7091/7091 of params[node] are nodes and 7091/7091 of params[registration] are registrations
[__main__]  INFO: For log action registration_approved, 2281/2281 of params[node] are nodes and 2281/2281 of params[registration] are registrations
[__main__]  INFO: For log action registration_cancelled, 252/252 of params[node] are nodes and 252/252 of params[registration] are registrations
[__main__]  INFO: For log action retraction_initiated, 614/614 of params[node] are nodes and 614/614 of params[registration] are registrations
[__main__]  INFO: For log action retraction_approved, 356/356 of params[node] are nodes and 356/356 of params[registration] are registrations
[__main__]  INFO: For log action retraction_cancelled, 118/118 of params[node] are nodes and 118/118 of params[registration] are registrations
[__main__]  INFO: For log action embargo_approved, 853/853 of params[node] are nodes and 853/853 of params[registration] are registrations
[__main__]  INFO: For log action embargo_completed, 380/380 of params[node] are nodes and 380/380 of params[registration] are registrations
[__main__]  INFO: For log action embargo_cancelled, 278/278 of params[node] are nodes and 278/278 of params[registration] are registrations

[__main__]  INFO: 54410/54410 nodes checked. 0 have no logs.
[__main__]  INFO: Finished analysis
```

PR #5463, opened 2016-04-12
Monkey patch out cached properties

chrisseto:

### DIFF
--- a/api/base/views.py
+++ b/api/base/views.py
@@ -71,11 +71,11 @@ class JSONAPIBaseView(generics.GenericAPIView):
             ser = CACHE[self.request._request][view.get_serializer_class()]
 
             try:
+                ser._context = view.get_serializer_context()
+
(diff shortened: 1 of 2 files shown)


brianjgeiger:
By and large this looks good, except for not really having a good way of understanding the potential implications. Could you add an API test that tests this? @cwisecarver suggested making a request, checking for the object that should be disposed of, making another request, and ensuring that the object is disposed of? I imagine that will be tricky not causing the test to force the object to stick around, so if you can't, you can't, but if you can think of a way, that would be nice. Otherwise it should be good to go as near as I can tell.

sloria:
Going to merge this for now, as it fixes a critical memory leak on prod. A test would be nice, but isn't high priority, since this changes an implementation detail rather than changing an API.

brianjgeiger:
Go for it.

PR #5465, opened 2016-04-13
OSF-5199 No forbidden error when accessing settings page

adlius:
<h3>
Purpose
</h3>
Original PR is [here](https://github.com/CenterForOpenScience/osf.io/pull/5336). I closed that PR and created this one to reflect new changes to fix the bug of no forbidden error when accessing settings page.

<h3>
Changes
</h3>
Change permission of the settings page to make use of the decorator "must_have_permission(READ)".
Add lots of unit tests to test all three decorator functions.

### DIFF
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -26,7 +26,10 @@ from framework.auth.decorators import must_be_logged_in
 from website import mails
 from website import settings
 from website.project.decorators import (
-    must_have_permission, must_be_contributor,
+    must_have_permission,
+    must_be_contributor,
+    must_be_contributor_or_public,
+    must_be_contributor_or_public_but_not_anonymized,
     must_have_addon, must_be_addon_authorizer,
 )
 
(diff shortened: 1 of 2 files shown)


adlius:
Oh, I almost forgot. Shout-out to @mfraezz for your patience on the last PR.

samanehsan:
@adlius, please fix the failing build :octocat:

tests/test_auth.py
```diff
@@ -193,27 +196,55 @@ class TestMustBeContributorDecorator(AuthAppTestCase):
...
-    def test_must_be_contributor_no_user(self):
+    def test_must_be_contributor_no_user_and_public_project_redirect(self):
+        res = view_that_needs_contributor(
+            pid=self.public_project._primary_key,
+            user=None,
+        )
+        assert_is_redirect(res)
+        # redirects to login url
+        redirect_url = res.headers['Location']
+        login_url = cas.get_login_url(service_url='http://localhost/')
```
mfraezz: This probably shouldn't be explicitly defined, but built from `settings.DOMAIN` or similar

mfraezz: ...though I see that we're already doing it elsewhere, and it would require parsing out the `:5000` part of `settings.DOMAIN`. This is fine, disregard.

mfraezz:
LGTM. Tests are good, and confirmed that the user receives a `403 FORBIDDEN`. Dropping this in RTM.

Side note: The only other case where `must_be_contributor` is used is `website.project.views.contributor:project_remove_contributor`. This isn't an issue for that method because it checks for admin permissions and, without them, raises a `403` unless the requesting user is trying to remove themselves; authenticated non-contributors can't get through. While entirely phasing out the `must_be_contributor` decorator is technically out of scope for this issue, given how small of a change it would be, it may be worthwhile to include the `must_have_permission(READ)` change there.

PR #5466, opened 2016-04-13
[WIP] [Feature] Add Admin Sales Analytics to OSF [#OSF-5932]

cslzchen:
## Purpose

With the pageview tracking code that we've added to {{staging.osf.io}} it should be possible to start generating metrics that are interesting from a sales perspective. Blow is a list of metrics we add to OSF admin server:

- Average user session length (past month)
- Average MAU session length (past month)
- Number of users that view 2+ products in the same year
- Number of users that view 2+ products in the same month

## Changes

Add a new tab: "OSF Sales Analytics" to OSF Admin.

## Side effects

No side effects.

## Ticket

[OSF-5923](https://openscience.atlassian.net/browse/OSF-5923)


cslzchen:
Pull request closed. Move to  [#5608](https://github.com/CenterForOpenScience/osf.io/pull/5608).

PR #5468, opened 2016-04-13
Add API /v2/guid/{guid}/ endpoint [OSF-6003]

samanehsan:
## Purpose
Add `/v2/guid/{guid}/` endpoint that redirects to the appropriate view depending on the guid referent.
E.g. `/v2/guid/{user_id}/` redirects to `/v2/users/{user_id}`.

If the guid exists but corresponds to a view that does not exist yet (e.g. wikis), the user gets a 501 not implemented error: 
screen shot 2016-04-13 at 11 23 46 am

If the guid does not exist, the user gets a 404 error:
screen shot 2016-04-13 at 11 22 44 am

## Changes
Add a `/v2/guid/{guid}/` endpoint and a `GuidRedirect` class that redirects to `referent.absolute_api_v2_url` or raises an error.

## Side effects
None

## Ticket
https://openscience.atlassian.net/browse/OSF-6003


brianjgeiger:
This looks great. Could you add tests for all of the known successful guid types, one 404 test, and one 501 test?

samanehsan:
Done adding tests @brianjgeiger! ☕

api/base/urls.py
```diff
@@ -22,7 +22,8 @@
                 url(r'^files/', include('api.files.urls', namespace='files')),
                 url(r'^docs/', include('rest_framework_swagger.urls')),
                 url(r'^institutions/', include('api.institutions.urls', namespace='institutions')),
-                url(r'^collections/', include('api.collections.urls', namespace='collections'))
+                url(r'^collections/', include('api.collections.urls', namespace='collections')),
+                url(r'^guid/', include('api.guid.urls', namespace='guid'))
```
brianjgeiger: this prooobably should be `guids`

brianjgeiger: Despite what I said on Jira. That guy clearly had no idea what he was talking about.

api/guid/views.py
```diff
@@ -0,0 +1,39 @@
...
+
+from framework.guid.model import Guid
+from framework.auth.oauth_scopes import CoreScopes, ComposedScopes
+from api.base.exceptions import NotImplementedError
+from api.base import permissions as base_permissions
+from api.base.views import JSONAPIBaseView
+
+
+class GuidRedirect(JSONAPIBaseView):
+
```
brianjgeiger: We should do a quick bit of documentation on this endpoint for the API docs. Maybe:

    """Find an item by its guid.

    This endpoint will redirect you to the most appropriate URL given its OSF GUID. If you give it a GUID for a node,
    you will be redirected to /v2/nodes/{node-id}, for example, while giving it a user's GUID will redirect you to
    /v2/users/{user-id}. If the GUID does not resolve, you will receive a 410 GONE response.

    #This Request/Response

    """

brianjgeiger:
Okay, other than docs and changing the collection name and url to `guids`, this does everything I could want. Yay! :penguin:

api/base/exceptions.py
```diff
@@ -83,6 +83,10 @@ def json_api_exception_handler(exc, context):
     return response
 
 
+class NotImplementedError(APIException):
```
sloria: This will shadow Python's built-in `NotImplementedError`. How about `EndpointNotImplementedError` or `HTTPNotImplementedError`?

api/base/exceptions.py
```diff
@@ -83,6 +83,10 @@ def json_api_exception_handler(exc, context):
...
+    status_code = status.HTTP_501_NOT_IMPLEMENTED
+    default_detail = _('This view is not yet implemented.')
```
sloria: The term "view" is Python-specific and shouldn't be returned to the client. How about "This endpoint is not yet implemented."

api/base/exceptions.py
brianjgeiger: `EndpointNotImplementedError` or `APINotImplementedError` might be most appropriate. (Django has an `HTTPNotImplementedError`)

sloria:
Just a couple drive-by comments to add to Brian's suggestions. After those are addressed, this is good to merge.

brianjgeiger:
Super-cool. :tada: :penguin: :tada:

PR #5469, opened 2016-04-13
[feature] Add .configured to <Addon>NodeSettings

mfraezz:
## Purpose
Fixes API bug where authorized storage addons without a selected folder were throwing `503`'s when links were followed

## Changes
* Add `.configured` property to base classes
* Use `.configured` instead of `.complete` in API call
* Tests

## Side effects
None, although there may be other places where similar bugs exist. This only fixes a very specific case.

## Ticket
[\[OSF-6019\]](https://openscience.atlassian.net/browse/OSF-6019)

### DIFF
--- a/api/nodes/views.py
+++ b/api/nodes/views.py
@@ -1519,7 +1519,7 @@ class NodeProvidersList(JSONAPIBaseView, generics.ListAPIView, NodeMixin):
             for addon
             in self.get_node().get_addons()
             if addon.config.has_hgrid_files
-            and addon.complete
+            and addon.configured
         ]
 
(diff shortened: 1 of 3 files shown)


brianjgeiger:
Since this is intended to fix a bug that is visible in an API view, could you also make a view test? If you have to reinvent the wheel with mocking, don't, but if you can just add the case where the addon is added but not configured, watch it fail without you code, and watch it pass with, that would be awesome. :penguin:

mfraezz:
As discussed in person, that class isn't being tested currently, and it's a simple enough change that it isn't worth doing now. :shipit:

PR #5476, opened 2016-04-14
Do not copy deleted nodes when templating project [OSF-5942]

samanehsan:
## Purpose

Fix [OSF-5942]

If a user deletes a component in a project, and then uses that project as a template to create a new project, the deleted component shows up in the new project's logs.  E.g. "Component 2" was deleted in the original project, but shows up in the logs for the new project: 
screen shot 2016-04-14 at 11 19 35 am

## Changes
When templating a project with components, do not copy components that are deleted.

## Side effects

None.

## Ticket

https://openscience.atlassian.net/browse/OSF-5942

### DIFF
--- a/website/project/model.py
+++ b/website/project/model.py
@@ -1638,7 +1638,7 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin, Commentable):
         new.nodes = [
             x.use_as_template(auth, changes, top_level=False)
             for x in self.nodes
-            if x.can_view(auth)
+            if x.can_view(auth) and not x.is_deleted
         ]
 
         new.save()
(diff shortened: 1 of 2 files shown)


website/project/model.py
```diff
@@ -1638,7 +1638,7 @@ def use_as_template(self, auth, changes=None, top_level=True):
         new.nodes = [
             x.use_as_template(auth, changes, top_level=False)
             for x in self.nodes
-            if x.can_view(auth)
+            if x.can_view(auth) and not x.is_deleted
```
samanehsan: @brianjgeiger, @sloria should `Node.can_view` return `False` if the project is deleted?

brianjgeiger: I don't *think* so, but I can see the argument for it. I think there are likely a lot of code assumptions around `can_view` being permissions-only that changing that might cause trouble. Philosophically, probably it should.

sloria: I also think not. Our code will need to distinguish between an auth error vs. a deleted error (e.g. 403 vs 410).

samanehsan: Ok, thank you both for the clarification!

brianjgeiger:
This looks great. And while it would be fantastic to clean up the logs on the problem templated nodes, I think the benefit:effort ratio would be far too low to make that worth doing. :tada: :penguin: :tada:

PR #5477, opened 2016-04-14
Override get_serializer_class() in GuidRedirect view [OSF-6003]

samanehsan:
## Purpose

DRF-Swagger throws and error because `GuidRedirect` does not define a `serializer_class`.

## Changes

Override `get_serializer_class()` and have it return `None` because setting `serializer_class = None` directly is not allowed. Also add a test to ensure all views either define a `serializer_class` or override `get_serializer_class()`

## Side effects
None.

## Ticket

https://openscience.atlassian.net/browse/OSF-6003

### DIFF
--- a/api/guids/views.py
+++ b/api/guids/views.py
@@ -32,6 +32,9 @@ class GuidRedirect(JSONAPIBaseView):
     view_category = 'guids'
     view_name = 'guid-detail'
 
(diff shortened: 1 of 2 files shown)


brianjgeiger:
This is good to merge when tests pass.

PR #5479, opened 2016-04-14
Correctly catch errors when filter type does not match field type

abought:
## Purpose

Certain APIv2 views exit with a 500 error when the filter has a different datatype than the field.

This was evident in the `/v2/nodes/<nid>/files/<provider>` endpoint, when filtering by `last_touched`: this fails for osfstorage, because that provider does not store a value for the field. (by design)

## Changes
- When filtering fails due to a `TypeError`, Provide a JSONAPI-formatted error message and a 400 error, instead of returning an unhelpful 500.

```
{
    "errors": [
        {
            "source": {
                "parameter": "filter"
            }, 
            "meta": {}, 
            "detail": "Could not apply filter to specified field"
        }
    ]
}

```

## Side effects
Should be none, as this is merely handling a previously uncaught exception.

I can create a custom exception class if preferred.

## Ticket
https://openscience.atlassian.net/browse/OSF-4967

### DIFF
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -374,10 +374,17 @@ class ListFilterMixin(FilterMixin):
                 if params['value'].lower() in getattr(item, field_name, {}).lower()
             ]
         else:
-            return_val = [
-                item for item in default_queryset
-                if self.FILTERS[params['op']](getattr(item, field_name, None), params['value'])
-            ]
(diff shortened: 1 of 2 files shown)


sloria:
The test failure is unrelated to these changes, and has since been fixed on `develop`. This is good to merge.

PR #5482, opened 2016-04-15
make institution.find not get deleted_insts

hmoco:
## Purpose

prevent deleted institutions from being accidently found.

## Changes

added argument to .find on institution that by default adds ` Q('is_deleted', 'ne', True) `

## Side effects

To find a deleted inst, one would have to call `.find` with a deleted=True argument.

### DIFF
--- a/api_tests/institutions/views/test_institution_list.py
+++ b/api_tests/institutions/views/test_institution_list.py
@@ -3,6 +3,7 @@ from nose.tools import *  # flake8: noqa
 from tests.base import ApiTestCase
 from tests.factories import InstitutionFactory
 
+from website.models import Node
 from api.base.settings.defaults import API_BASE
 
 class TestInstitutionList(ApiTestCase):
@@ -11,6 +12,10 @@ class TestInstitutionList(ApiTestCase):
         self.institution = InstitutionFactory()
         self.institution2 = InstitutionFactory()
         self.institution_url = '/{}institutions/'.format(API_BASE)
+
+    def tearDown(self):
+        super(TestInstitutionList, self).tearDown()
+        Node.remove()
 
     def test_return_all_institutions(self):
         res = self.app.get(self.institution_url)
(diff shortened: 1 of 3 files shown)


hmoco:
@icereval, let me know if there was a ticket for this

sloria:
Hm, I'm not sure if it's desirable to add an arbitrary query (`is_deleted=False`) by default. After all, `Institution.find()` is meant to behave like `StoredObject.find()`. It might be slightly better to just add the `is_deleted=False` query to the `Institution.find(..)` calls where appropriate.

That said, I don't feel strongly against this PR's approach.

hmoco:
although I agree that ` .find ` methods should behave similarly (ideally exactly the same), I can't think of a single scenario where returning a deleted institution is desired, and there isn't one in the current institutions codebase, meaning ` is_deleted=False ` would have to be added to every ` Institution.find() ` call.

hmoco:
@sloria ^ that's my thought process for this, only added the deleted argument to prevent completely disallowing returning deleted Institutions. Ultimately, your call, I can see the benefits of both.

sloria:
Your rationale is valid. You could also argue that `is_deleted` is an implementation detail and that deleted Institutions/Nodes should be treated as if they didn't exist in the database, in which case excluding them by default makes sense. Let's go with that.

Can you add a test for your change? Once that's done, this is good to merge.

hmoco:
added an api test, do you think there should be a model test for the deleted find argument?

sloria:
A model test wouldn't hurt; would be good to test the default (Institition.find() doesn't return deleted institutions) as well as the non-default (.find(deleted=True)).

hmoco:
there they are!

sloria:
Thanks; good to merge when build passes.

GaryKriebel:
@sloria .... I'm guessing 0.66.0 is no longer valid?

sloria:
@GaryKriebel Hotfixes are going into 0.66.0

GaryKriebel:
Ah, yes... makes sense

PR #5483, opened 2016-04-15
Load comments incrementally as you scroll [#OSF-5891]

erinspace:
## Purpose

At the moment, all comments are loaded all at once when the comment pane is open. This can result in many unnecessary API calls when a user does not wish to see all the comments. 

This PR makes it so that new comments aren't loaded until the user scrolls to the bottom of the currently displayed list of comments from the first call. When you scroll to the bottom, new comments will be loaded (and will stay loaded).

## Changes
- new osfHelper function that will perform an action when the specified element scrolls to the bottom
- add ```urlForNext``` ko Observable that keeps track of when there is a new URL to load comments
- new GetMoreComments method that's called when there is a new URL to call and when the previous request is finished
- call the new osfHelper on the comments window to trigger the getMoreComments method
- Change the loading icon to be a placeholder blank comment template to replace the current spinner:
screen shot 2016-04-26 at 9 37 25 am
- Add a button/link to load more comment replies if there are more than 10:
screen shot 2016-04-26 at 3 21 51 pm

## Side effects
- Other places can now use the onScrollToBottom function to load more sparingly!

## Ticket

https://openscience.atlassian.net/browse/OSF-5891


website/static/js/comment.js
```diff
@@ -119,6 +143,23 @@ BaseComment.prototype.setupToolTips = function(elm) {
     });
 };
 
+BaseComment.prototype.get_url = function() {
```
samanehsan: `get_url` should be in camelCase

samanehsan: It looks like `get_url` isn't being used anywhere. Did you mean to use it in `BaseComment.prototype.fetch`?

website/static/js/comment.js
```diff
@@ -161,16 +202,33 @@ BaseComment.prototype.fetchNext = function(url, comments, setUnread) {
...
             self.loadingComments(false);
         }
     }).fail(function () {
         self.loadingComments(false);
     });
 };
 
+BaseComment.prototype.getMoreComments = function() {
+    var self = this;
+    var next_url = self.urlForNext();
```
samanehsan: `next_url` should also be in camelCase

samanehsan:
After the first 10 comments load and I scroll to the bottom to load more, I think all of the remaining comments load instead of only the next 10.

website/static/js/comment.js
```diff
@@ -91,6 +100,21 @@ var BaseComment = function() {
         return self.submittingReply() ? 'Commenting' : 'Comment';
     });
 
+    self.updateViewportDimensions = ko.computed(function () {
+        var commentSidebarRef = $('.cp-sidebar-content');
+        var commentRef = $('.comment-body').first();
+
+        var commentsWidth = self.atBottomOfScroll() ? commentSidebarRef.width() : 300;
```
samanehsan: Can you replace the hard-coded values with variables? That will make it easier to tell where the numbers came from!
e.g. `var COMMENT_SIDEBAR_DEFAULT_WIDTH = $('.cp-sidebar-content').width();`

samanehsan:
Pass complete @erinspace! ☀️

samanehsan:
Can you update the docstring for `fetchNext` now that it no longer gets all of the comments at once? https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/comment.js#L142

website/static/js/comment.js
```diff
@@ -161,16 +163,25 @@ BaseComment.prototype.fetchNext = function(url, comments, setUnread) {
...
 };
 
+BaseComment.prototype.getMoreComments = function() {
+    var self = this;
+    var nextUrl = self.urlForNext();
+    var comments = self.comments();
+    var setUnread = self.getTargetType() !== 'comments' && !osfHelpers.urlParams().view_only && self.author.id !== '';
+
+    if (self.urlForNext()) {
+        if (!self.loadingComments()) {
```
samanehsan: Can these two if statements be combined onto one line?

samanehsan:
This looks awesome @erinspace!!! Just a couple of minor comments to address then this should be good to merge ☀️

samanehsan:
When I click to view a comment's replies, I only get the last 10. Is it possible to do the same thing as top-level comments and load more when you scroll to the bottom?

samanehsan:
Pass complete @erinspace! 🎉

website/static/js/osfHelpers.js
```diff
@@ -844,6 +844,18 @@ var trackClick = function(category, action, label){
     return true;
 };
 
+
+// Call a function when scrolled to the bottom of the element
+/// Useful for triggering an event at the bottom of a window, like infinite scroll
+function onScrollToBottom(element, callback) {
+    $(element).scroll(function() {
+        var $this = $(this);
+        if ($this.scrollTop() + $this.innerHeight() >= $this[0].scrollHeight) {
```
brianjgeiger: This may be a little excessive, but since you had so much trouble with this, it might be a good idea to pull this boolean out into its own function and then unit test that function. If you could throw all the things that you ran into along the way that didn't work into the test, that would save future folks who try to make this fancier from messing everything up.

Alternately, you could try unit testing this function, but you'd have to fake the element properly, which might not be so bad.

samanehsan:
This looks great @erinspace! 🚀 

Pinging @nicipfeiffer and @saradbowman -- this might be relevant to the product spec for commenting changes!

PR #5484, opened 2016-04-15
Generalize forgot password email throttle & throttle email endpoints [OSF-6000]

samanehsan:
## Purpose

Currently, when sending the forgot password email, the "last sent" time updates on form submit, not when the email gets sent. 

Also, throttle endpoints where the user action results in an email getting sent with no limitations as to how frequently they are able to do so. This includes resending a confirmation email to an unconfirmed account email, sending an account export request and sending an account deactivation request.

## Changes
- Instead of having a user field saving the time each type of email was last sent (e.g. forgot
password email), have one generic field -- `User.email_last_sent` and a generic flag `SEND_EMAIL_THROTTLE.`
- When sending the forgot password email, only update `User.email_last_sent` after the email
is sent.
- Only allow users to resend a confirmation email, send an account export request or send an account deactivation request once every 30 seconds.

## Side effects
None.

## Ticket

https://openscience.atlassian.net/browse/OSF-5960


brianjgeiger:
Looks great! :tada: :penguin: :tada:

website/profile/views.py
```diff
@@ -102,6 +104,8 @@ def resend_confirmation(auth):
     data = request.get_json()
 
     validate_user(data, user)
+    if not throttle_period_expired(user.email_last_sent, settings.SEND_EMAIL_THROTTLE):
+        raise HTTPError(httplib.BAD_REQUEST, data={'message_long': 'Please wait 30 seconds before sending another confirmation email.'})
```
brianjgeiger: Let's leave the message vague as to how long to wait, and maybe what kind of email. Maybe something like, 'Too many requests. Please wait a while before trying again.'

brianjgeiger: Ah, this is in the view, right. So maybe, 'Too many requests. Please wait a while before sending another confirmation email.'

website/profile/views.py
```diff
@@ -777,21 +782,39 @@ def unserialize_schools(auth, **kwargs):
 
 @must_be_logged_in
 def request_export(auth):
+    user = auth.user
+    if not throttle_period_expired(user.email_last_sent, settings.SEND_EMAIL_THROTTLE):
+        raise HTTPError(httplib.BAD_REQUEST,
+                        data={'message_long': 'Please wait 30 seconds before sending another account export request.',
```
brianjgeiger: Vague up the time here as well, please.

website/profile/views.py
```diff
@@ -777,21 +782,39 @@ def unserialize_schools(auth, **kwargs):
...
     return {'message': 'Sent account export request'}
 
 
 @must_be_logged_in
 def request_deactivation(auth):
+    user = auth.user
+    if not throttle_period_expired(user.email_last_sent, settings.SEND_EMAIL_THROTTLE):
+        raise HTTPError(http.BAD_REQUEST,
+                        data={
+                            'message_long': 'Please wait 30 seconds before sending another account deactivation request.',
```
brianjgeiger: More vague, please, with respect to the time.

website/project/views/contributor.py
```diff
@@ -365,7 +366,12 @@ def get_timestamp():
 
 
 def throttle_period_expired(timestamp, throttle):
```
brianjgeiger: Looks like this doesn't have any tests. Could you add a few? I wouldn't go crazy, but cover the major cases.

brianjgeiger:
Looks good. A few messages to alter, a few tests to add. And were you able to manually test as well? 🐧

samanehsan:
Done making those changes! I manually tested those endpoints 🌷

tests/test_contributors_views.py
```diff
@@ -131,3 +134,17 @@ def test_get_contributors_from_parent(self):
...
         )
+
+    def test_throttle_period_expired_no_timestamp(self):
+        is_expired = throttle_period_expired(timestamp=None,  throttle=30)
+        assert_true(is_expired)
+
+    def test_throttle_period_expired_using_datetime(self):
+        timestamp = datetime.utcnow()
+        is_expired = throttle_period_expired(timestamp=timestamp,  throttle=30)
+        assert_false(is_expired)
```
brianjgeiger: Could you add a couple more assertions on these last two tests. Maybe timestamp + 29 seconds is false, and timestamp + 31 seconds is true? Just to ensure that the base functionality of the function works.

brianjgeiger:
Okay, add those extra assertions and we'll be all done. Don't forget to throw the jira issue into code review when it's ready. 🐧

brianjgeiger:
Perfect. 🎉 🐧 🎉

framework/auth/views.py
```diff
@@ -26,10 +26,10 @@
 from website import mails
 from website import language
 from website import security
+from website.project.views.contributor import throttle_period_expired
```
sloria: Minor: `throttle_period_expired` doesn't really belong in `views.contributor`. It'd be better to go in `website.utils.time`. Feel free to make this change. Or not. Up to you.

sloria: Actually, I'll just make this change when I merge.

sloria:
Merged into release/0.68.0

framework/auth/views.py
samanehsan: Thanks for making that change @sloria! 🐙

PR #5480, opened 2016-04-14
[#OSF-58348]Email help desk with stuck registrations 

chennan47:
## Purpose

Change the scripts which handle failed registration not to cleanup the failed one but to collect the failed registration daily and send all the info to help desk.

## Changes

1. remove the code in https://github.com/CenterForOpenScience/osf.io/blob/6ccd9126c397556b1117682285cbfe862d718824/scripts/cleanup_failed_registrations.py which cleanup the failed registration

2. log all the failed registration and send it to the help desk

3. set it up as a daily task

## Ticket

https://openscience.atlassian.net/browse/OSF-5848

PR #5485, opened 2016-04-15
[#OSF-58348]Email help desk with stuck registrations

chennan47:
## Purpose

Change the scripts which handle failed registration not to cleanup the failed one but to collect the failed registration daily and send all the info to help desk. Replace of https://github.com/CenterForOpenScience/osf.io/pull/5480.

## Changes

1.remove the code in https://github.com/CenterForOpenScience/osf.io/blob/6ccd9126c397556b1117682285cbfe862d718824/scripts/cleanup_failed_registrations.py which cleanup the failed registration

2.log all the failed registration and send it to the help desk

3.set it up as a daily task through celerybeat

## Ticket

https://openscience.atlassian.net/browse/OSF-5848


scripts/get_failed_registrations.py
```diff
@@ -0,0 +1,82 @@
+
+# -*- coding: utf-8 -*-
+import sys
+import logging
+
+from datetime import datetime
+from django.core.mail import send_mail
```
sloria: Use the OSF's `send_mail` function rather than Django's. Otherwise, this script will likely crash.

sloria:
Please test that the script works, and that the correct email output is logged to the celery logs.

sloria:
Pass finished.

scripts/get_failed_registrations.py
chennan47: fixed in fa197cc

PR #5486, opened 2016-04-15
Feature/resend confirm email throttle [OSF-6000]

samanehsan:
Note: This PR relies on changes made in #5484, which must get merged first! 

## Purpose
When a user adds an additional email to their account (which can be any email address), a confirmation email gets sent to that email address. Users can resend that confirmation email any number of times, which can potentially be used for spamming. 

## Changes

Only allow users to click "Resend confirmation email" once every 30 seconds.

## Side effects

None.

## Ticket


samanehsan:
Cherry-picked 2f8c22a into #5484

PR #5487, opened 2016-04-15
Unregistered contributors in the Dashboard/My Projects page [OSF-6057] 

pattisdr:
❗ https://github.com/CenterForOpenScience/osf.io/pull/5513 should be merged first

# Purpose

Jira https://openscience.atlassian.net/browse/OSF-6057

1) Solves Trello issue https://trello.com/c/QAT7HpYH/241-unregistered-contributor-on-a-certain-project-shows-up-as-enter

Unregistered contributors can be known as different names across different projects until the user claims the account.  The MyProjects and Dashboard page were not taking this into account and just using the user name in the db - which would be whatever the first user to add them as a contributor decided to call them.  

If I call an unregistered contributor "Robert" in my project, and you call him "Bob", while they are the same user in the background, until the user claims the account (and decides that he wants to be known as Robert), he should be called Robert in my projects, and Bob in yours.

2) Tweaks Discover Public Projects.  Currently, projects do not show up in Discover Public Projects if the user does not have access to them.  However, if the project has been made private by the user, it should also not show up in this list.  Also, loops through Discover Public Project results until have 5 to display, or have exceeded the number of results. 

# Changes

Adds an unregistered_contributor field to the NodeContributors serializer so we can use this value in My Projects and Dashboard.  Also creates a serializer to be used in the NodeLogContributors endpoint.  This is used in the new logs, when the log action is contributor_added.

For filtering, If I am a contributor on two different projects calling the unregistered contributor different things, the filter will show up as "Bob Jackson a.k.a. Robert Jackson".  For search, searching "Bob" only returns Bob projects, and "Robert" returns Robert projects because it is searching text.  This is different than filtering.

**Project Organizer and Filtering**
screen shot 2016-04-15 at 5 21 10 pm

**Dashboard**
screen shot 2016-04-15 at 3 22 23 pm

**Logs**
screen shot 2016-04-15 at 3 23 04 pm

screen shot 2016-04-15 at 3 22 38 pm

# Side effects

A few API changes.


website/static/js/myProjects.js
```diff
@@ -709,14 +726,24 @@ var MyProjects = {
             self.nameFilters = [];
 
             var userFinder = function(lo) {
+                if (lodashGet(u2, 'unregistered_contributors')) {
```
samchrisinger: Might be clearer to just use:
`if (u2.unregistered_contributors)`

website/static/js/myProjects.js
```diff
@@ -709,14 +726,24 @@ var MyProjects = {
...
               return lo.label === u2.data.embeds.users.data.attributes.full_name;
             };
 
             // Add to lists with numbers
             for (var user in self.users) {
                 var u2 = self.users[user];
                 if (u2.data.embeds.users.data) {
-                  var link = oldNameFilters.find(userFinder) || new LinkObject('contributor', {id: u2.data.id, count: u2.count, query: { 'related_counts' : 'children' }}, u2.data.embeds.users.data.attributes.full_name, options.institutionId || false);
+                    var name;
+                    if (lodashGet(u2, 'unregistered_contributors')) {
```
samchrisinger: ditto

website/static/js/myProjects.js
```diff
@@ -268,18 +268,22 @@ function _formatDataforPO(item) {
...
+                var users = lodashGet(c, 'embeds.users.data', null);
+                if (users) {
+                    if (users.errors) {
+                        attr = users.errors[0].meta;
+                    } else {
+                        attr = users.attributes;
+                    }
+                }
+                if (attr)
+                    item.contributors += attr.full_name + ' ' + attr.middle_names + ' ' + attr.given_name + ' ' + attr.family_name + ' ' ;
```
samchrisinger: Likely outside the scope of this PR but I'm noticing some duplication between this code and project-organizer.js, in particular for mapping API responses into a consistent User interface. It might be beneficial to encapsulate this as JS class to keep all of this logic in once place.

website/static/js/project-organizer.js
```diff
@@ -67,13 +69,17 @@ function _poContributors(item) {
         var familyName;
         var givenName;
         var fullName;
-        if (person.embeds.users.data) {
+
+        if (lodashGet(person, 'attributes.unregistered_contributor')) {
+            familyName = person.attributes.unregistered_contributor;
+        }
+        else if (person.embeds.users.data) {
             familyName = person.embeds.users.data.attributes.family_name;
```
samchrisinger: https://github.com/CenterForOpenScience/osf.io/pull/5487/files#r60628526

api/logs/views.py
```diff
@@ -220,7 +221,7 @@ class NodeLogContributors(JSONAPIBaseView, generics.ListAPIView, ODMFilterMixin,
     required_read_scopes = [CoreScopes.USERS_READ]
     required_write_scopes = [CoreScopes.NULL]
 
-    serializer_class = UserSerializer
+    serializer_class = UnregisteredContributorSerializer
```
samchrisinger: I don't really like the semantics of this, but I'm guessing it was a decision made a while ago-- just feels me that the exceptional case (unregistered contributor) appears as the `serializer_class`. '

The existence of `UnregisteredContributorSerializer` doesn't make a ton of sense to me. It's just a user + one extra field. But I'm probably just being overly zealous. 

Not a big deal, but this definitely threw me at first.

website/static/js/home-page/newAndNoteworthyPlugin.js
```diff
@@ -43,12 +43,17 @@ var NewAndNoteworthy = {
...
         var newAndNoteworthyPromise = m.request({method: 'GET', url: newAndNoteworthyUrl, config: xhrconfig, background: true});
         newAndNoteworthyPromise.then(function(result){
-            var numNew = Math.min(result.data.length, self.SHOW_TOTAL);
+            var numNew = result.data.length;
             for (var l = 0; l < numNew; l++) {
                 var data = result.data[l];
                 if (lodashGet(data, 'embeds.target_node.data', null)) {
-                    self.newAndNoteworthyNodes().push(result.data[l]);
-                    self.fetchContributors(result.data[l]);
+                    if (data.embeds.target_node.data.attributes.public === true) {
```
samchrisinger: I'm trying to understand:
> projects do not show up in Discover Public Projects if the user does not have access to them. 

why does this list include non-public projects in the first place?

api/logs/views.py
samchrisinger: Another option might be to add the `unregistered_contributor` to the `UserSerializer`, and let that serializer figure out the appropriate representation for that user (claimed or unclaimed).  Then 
```
 +        modified_users = []
 +        for contrib in associated_users:
 +            if contrib.unclaimed_records.get(log.original_node._id):
 +                contrib.unregistered_name = contrib.unclaimed_records[log.original_node._id]['name']
 +            modified_users.append(contrib)
 +        return modified_users
```
Could go away. Really what this ^  does is tamper with serialization in the view. which is not ideal.

samchrisinger: Thoughts on conditionally omitting a field from the serialized representation: http://stackoverflow.com/questions/27015931/remove-null-fields-from-django-rest-framework-response

samchrisinger:
@pattisdr 1st pass done 🌞

website/static/js/home-page/newAndNoteworthyPlugin.js
pattisdr: This is the case if we add a project to Discover Public Projects, and then that user makes the project private.  It is still in the list (because it is a node link), but now it's a private link.

pattisdr: So if you made your project private, it wouldn't show up for me in the Discover Public Projects list, but it would still show up for you.  QA pointed this out - then you might think everyone else can see the project you decided to make private.

samchrisinger: Oh I see. I was forgetting the "Discover Public" collection in generated in a script.

Thanks for the clarification!

samchrisinger: @pattisdr wondering how hard it would be do actually update that collection when a project is made private? Could be a hook in `Node.set_privacy`? 

Or if we're already filtering the result-set backend should we just filter out all non-public projects there?

api/logs/views.py
pattisdr: Thinking about this...

Users and contributors are different things.  If I just navigate to a user detail endpoint `/v2/users/me`, the unregistered_contributor field doesn't apply here.  Unregistered contributor names only come into play when I'm asking for contributors on a node.  You might have called the user Robert, I called him Bob, and until the user claims the account, his name is different on different nodes.  But the different names are associated with the nodes.

samchrisinger: I think this is a fair argument. Given this, can you explain why we're not using `NodeContributorSerializer` (https://github.com/CenterForOpenScience/osf.io/blob/develop/api/nodes/serializers.py#L276) ?

pattisdr: The reason why UserSerializer was being used instead of the NodeContributor serializer in the first place was that we just wanted the name of the contributor.  It was the simplest serializer.

This is for a very specific endpoint - `logs/{log_id}/contributors`. It is fetching the contributors that were added as a part of this log action (contributors_added).  If this log has nothing to do with adding contributors, it will return an empty list.    The whole reason behind this endpoint even existing was that we needed the contributors' names that were added in the My Projects page to formulate the logs, without needing an additional request. 

In summary - to use Node Contributor serializer I need to return the rest of the info about the contributor in this view https://github.com/CenterForOpenScience/osf.io/blob/develop/api/nodes/views.py#L539-543.  That seems like a lot of information for just needing the user's name.

Alternatively, we're returning a bunch of things in log params now - maybe we could get the names of the users and return them in params and do away with this endpoint altogether.

samchrisinger: Thanks again for the explanation-- clearly there's some history here that I missed.

I really don't want to block this PR, and would be OK with a comment explaining the need for `UnclaimedUserSerializer`.

But IMO doing away with such a specific endpoint seems like a great idea to me. Probably needs approval from @brianjgeiger to go forward here.

pattisdr: @samchrisinger, @brianjgeiger - I went ahead and did the work to remove the `logs/{log_id}/contributors` endpoint.  When I put it in in the first place, in CR it came up that this wasn't the best way to handle it.   If you think the endpoint should stay I will add it back in.

I moved the user information to params - just the information that we need for the front end - names, is_active attribute, and link for the user.   Logs are a weird endpoint, having to handle lots of different types of information.  In my opinion, it is okay to return the bare minimum information we need under params to formulate the logs.  For example, if a node has been deleted, we need to access the node's title for the logs because accessing the node in the API will return Gone.  So we stick the bare minimum: node title and id in the log params.

brianjgeiger: I meant to respond to this last night, but didn't have a chance. Sorry for the delay. Having a contributors sub-route or not is fine by me; we do it in the nodes endpoint, but if it's not necessary, that's fine. The important thing is that serialization of any user information has to go through either the users serializer or it has to be otherwise anonymizable. Be sure there's a test for "contributor information does not show if anonymous view only link" or similar. There's a wrapper that will handle the work on a field-by-field basis; you'll see it I in the user serializer.

In general, I was hoping to force all serialization of user information to go through the User serializer, so the longer we can keep that, the better. I'm not going to be super-strict about it as long as that's taken care of.

website/static/js/home-page/newAndNoteworthyPlugin.js
pattisdr: Are you talking about changing how node links work?  The projects under Discover Public Projects are node links.  Current behavior, it someone makes a node link private, the project stays linked, you just no longer have access to it.  

We are fetching the node links on the node via API v2 and some could be private.

api/logs/views.py
pattisdr: Okay, I didn't know the bit about needing to be anonymizable.

As a reminder this is for the contributors_added log, and we need to get the names of the contributors added.   In the params, the contributors are a list.  So I have an array of contributors and I need to get their names.  I just changed it to stick their names under params but I don't know how to use the User Serializer there.

The only way I know how to use the User Serializer is to revert back to what it was before, with it's own view, own endpoint, etc.

pattisdr: Params themselves don't seem to show up when viewing through an anonymous link, however? The contributor information would be under params.  Is this sufficient, or do I definitely need the UserSerializer?


screen shot 2016-04-22 at 10 24 48 am

brianjgeiger: My only concern is that user information not be serialized when there's an anonymous link. This seems fine, but please write some tests so if someone decides that they need log params when there's an anonymous link, that user info is not revealed,

pattisdr: Added test. 

I am checking is_anonymized in the contributors field within the params, so even if someone changes this later and params are shown in anonymous links, the contributors are still hidden.

So here, I've temporarily added 'params' to non_anonymized_fields and the contributors return an empty list.

screen shot 2016-04-22 at 1 05 31 pm

pattisdr:
@samchrisinger, do you think I can move this to Ready To Merge or should I get another reviewer?

samchrisinger:
@pattisdr, sorry I didn't know this was ready for another pass. Will take a look this morning.

pattisdr:
My fault - I had forgotten about this one.

samchrisinger:
lgtm 👍

sloria:
Just needs one last update with develop. @pattisdr . Thanks!

PR #5488, opened 2016-04-15
[Feature] throttle account export request [OSF-6000]

samanehsan:
Note: This PR is branched off of #5487, which will need to get merged into develop first

## Purpose
Users can request an export of their account data as frequently as they would like -- each time that button is clicked an email gets sent to the osf help account.

## Changes
Only allow users to request an export of their account data once every 30 seconds. 

## Side effects
None.

## Ticket


samanehsan:
Cherry-picked 25feabe into #5487

PR #5489, opened 2016-04-15
Feature/Fix retractions [OSF-6056]

pattisdr:
See Ticket https://openscience.atlassian.net/browse/OSF-6056 for more info

## Purpose

Retractions used to be included under the `/nodes/` route; that is no longer the case.  The corresponding API tests were not updated so are not testing the correct thing.  

For example, I want to make sure I can't access a retraction's logs.  The test was looking at `/v2/nodes/{retraction_id}/logs/` instead or `/v2/registrations/{retraction_id}/logs/`.  The first would 404, the second 403. 

## Changes

- Consolidates retractions tests and uses /registrations/ endpoint instead of /nodes/ endpoint.
- Changes remaining places in the API to use 'withdrawn' language instead of 'retracted' language


brianjgeiger:
By and large, this looks good. But, and I hate to say this, but apparently we've renamed 'retractions' to 'withdrawals' or something on the front end, and since you've gone and re-factored all this anyways, it would probably be good to match. Could you verify with @saradbowman the proper words to use for this and update accordingly? Also, since I'm not sure aside from looking at tests what the rules are for withdrawn registrations, we should probably mention them in the docs. That will let QA test and API consumers know what to expect. Just a quick set of bullet points about what's different between withdrawn registrations and normal ones would be fine.

pattisdr:
I know we've renamed retractions to withdrawals, but you are saying you want the phrasing to also match in the tests?  And fields, like HideIfRetraction?

brianjgeiger:
Yeah, let's keep the language the same from the front-end to the visible part of the API to the code when we can. Since this isn't something people are using yet in any great numbers, if we get it now, then devs won't have to wonder why it's called something different in one place than another.

brianjgeiger:
Don't change it in the Node model, though. That would be a big pain we don't want to deal with right now. Just as far back as the API.

pattisdr:
Why has collection formerly dashboard been removed from the API docs for the nodes?  Can I add this?

brianjgeiger:
Collection should be out of the node serializer entirely. Is there still a field for it?

pattisdr:
Yes, it is still in the Node Serializer.

brianjgeiger:
Hmmm. It should probably not be in the node serializer, because you can't serialize a collection through the node endpoint. It will always be false. We took out the registration flag, right?

pattisdr:
Registration flag is still there too.

brianjgeiger:
Okay. Let's leave the fields and document them with a note that they'll always be false and not to use them and that they may be going away in a future version of the API.

pattisdr:
Skimming through the API docs - " Top-level nodes may have a category other than project".  In the new dashboard/my projects page, a top-level project has to have the category "project"; you're not given the option to change it.  

Should we restrict the API so in the future all top-level nodes must have category "project"?

brianjgeiger:
There's an open issue where Sara and I are discussing whether that should be a real limitation or not.

api/nodes/views.py
```diff
@@ -106,39 +107,41 @@ def get_file_object(self, node, path, provider, check_object_permissions=True):
...
+    a category other than project, and children nodes may have a category of project.
 
     ##Node Attributes
 
     <!--- Copied Attributes from NodeDetail -->
 
     OSF Node entities have the "nodes" `type`.
 
-        name           type               description
+        name                            type               description
```
brianjgeiger: Did you mean to put all those spaces in there?

brianjgeiger: Oh, I see you did. Sorry, the view obscured it. I will delete these comments.

brianjgeiger:
Sorry, I never managed to finish up on this after you did all the renaming. If this is useful before monday, please find someone to do a quick pass on it.

brianjgeiger:
Okay, Community has come back and told us that the API is correct on top-level nodes rather than the front end, so the front-end will change.

api/nodes/views.py
```diff
@@ -708,35 +738,44 @@ def perform_destroy(self, instance):
...
-        pending_embargo_approval        boolean            Is the associated Embargo awaiting approval by project admins?
+        title                           string             title of the registered project or component
+        description                     string             description of the registered node
+        category                        string             bode category, must be one of the allowed values
+        date_created                    iso8601 timestamp  timestamp that the node was created
+        date_modified                   iso8601 timestamp  timestamp when the node was last updated
+        tags                            array of strings   list of tags that describe the registered node
+        current_user_permissions        array of strings   list of strings representing the permissions for the current user on this node
+        fork                            boolean            is this project a fork?
+        registration                    boolean            has this project been registered? (always true - may be deprecated in future versions)
```
brianjgeiger: I think this should be "is this node a registration".

api_tests/registrations/views/test_withdrawn_registrations.py
```diff
@@ -0,0 +1,173 @@
...
+            'pending_withdrawal': None,
+            'pending_registration_approval': None,
+            'pending_embargo_approval': None,
+            'embargo_end_date': None,
+            'registered_meta': None,
+            'current_user_permissions': None,
+            'registration_supplement': registration.registered_schema[0].name
+        }
+
+        for attribute in expected_attributes:
```
brianjgeiger: The big gap in testing is that, if someone adds a new field to the NodeSerializer but doesn't mark it explicitly in the RegistrationSerializer, then it's going to be serialized. That will be very easy to miss. Can you either add a new test or refactor this test to account for that? It might make more sense as a Serializer test. 

The other option is to refactor the hidden fields to work more like the Anonymous fields, where we have to explicitly list the fields that are okay to show. Mostly, though, just make sure people won't accidentally reveal information by adding a new field to the Node serializer. Probably we'll need to do something similar with the relationships fields.

brianjgeiger:
Pass done. One tiny thing, one big thing. 🐧

api_tests/registrations/views/test_withdrawn_registrations.py
pattisdr: That is a very good point.

pattisdr: I added a test that looks at all the fields in the Node Serializer and if they are not whitelisted fields for withdrawals they must be wrapped in a HideIfWithdrawal field on the Registration Serializer.  It caught that template_from was being serialized (I'm pretty sure we shouldn't show that for withdrawals but I might be wrong).

api_tests/base/test_serializers.py
```diff
@@ -52,6 +53,27 @@ def null_field(*args, **kwargs):
...
+    """
+    All fields on the Node Serializer other than the few we can serialize for withdrawals must be redeclared on the
+    Registration Serializer and wrapped in HideIfWithdrawal
+    """
+    def test_registration_serializer(self):
+
+        # fields that are visible for withdrawals
+        visible_on_withdrawals = ['contributors', 'date_created', 'description', 'id', 'links', 'registration', 'title', 'type']
+        # fields that do not appear on registrations
+        non_registration_fields = ['registrations']
```
brianjgeiger: Is there anything worth testing on this independently of the Withdrawn tests?

brianjgeiger:
Just the one comment on the serializer test. 🐧

api_tests/base/test_serializers.py
pattisdr: I will add some tests for the HideIfRegistration field.  I assumed I had added them when I added the field, but it looks like my original tests were just centered around retractions hmph.

brianjgeiger:
Looks great! 🎉 🐧 🎉

sloria:

pattisdr:
@sloria, build passing.

PR #5490, opened 2016-04-15
[Feature] throttle account delete request [OSF-6000]

samanehsan:
Note: This PR is branched off of #5487, which will need to get merged into develop first

## Purpose

Users can request account deactivation as frequently as they wish, and  each time that button is clicked an email gets sent to the osf help account.

## Changes

Only let users request account deactivation once every 30 seconds. 

## Side effects

None.

## Ticket


samanehsan:
Cherry-picked 7a34932 into #5487

PR #5494, opened 2016-04-18
[WIP] [RfR] Migration for DraftRegistration question.extra

barbour-em:
## Purpose

Having multiple files in a DraftReg requires that the extra field be a list but looking at data on staging most are still dicts. I didn't notice this problem locally which is why this migration wasn't in the original PR. 

## Changes
``` 
'q1' : {                         'q1': {
  'value': 'foo',      ->          'value': 'foo',
  'extra': {}                      'extra': []
}                                 }
'q2': {                          'q2': {
  'value': 'bar',                  'value': 'bar',
  'extra': {                       'extra': [{
    'file': 'data'       ->          'file': 'data'
  }                                 }]
}                                 }
```

## Side effects

Shouldn't be any noticeable side effects other than fixing js errors that might be currently happening on staging. 

## Ticket
No ticket other than [the parent](https://openscience.atlassian.net/browse/OSF-5586)

### DIFF
--- a/scripts/migration/migrate_registration_extra.py
+++ b/scripts/migration/migrate_registration_extra.py
@@ -1,0 +1,53 @@
+"""
+Changes existing question.extra on a draft to a list
+required for multiple files attached to a question
+"""
+import sys
+import logging
+
+from website.app import init_app
+from scripts import utils as scripts_utils
+from website.models import DraftRegistration
+from framework.transactions.context import TokuTransaction
+
+
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+def migrate_drafts(dry):
+    draft_registrations = DraftRegistration.find()
+    count = 0
+    for r in draft_registrations:
+        data = r.registration_metadata
+        for q, ans in data.iteritems():
+            try:
+                files = ans['extra']
+            except (TypeError, KeyError) as e:
+                # Non-prereg Registrations
+                logger.exception(e)
+            else:
+                if type(files) is dict:
+                    if len(files.keys()) == 0:
+                        ans['extra'] = []
+                    else:
+                        ans['extra'] = [files]
+        count += 1
+        if not dry:
+            r.save()
+    logger.info('Done with {0} drafts migrated.'.format(count))
+
+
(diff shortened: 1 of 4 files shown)


sloria:
@samchrisinger Can you review this when you get a chance?

sloria:
Note to whoever merges this--this will need to be merged into `release/0.68.0`.

@barbour-em Be sure to **not** merge develop into this branch.

samchrisinger:
Yes

scripts/migration/migrate_registration_extra.py
```diff
@@ -0,0 +1,53 @@
...
+from website.models import DraftRegistration
+from framework.transactions.context import TokuTransaction
+
+
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+def migrate_drafts(dry):
+    draft_registrations = DraftRegistration.find()
```
samchrisinger: Is there a reason not to filter this set ahead of time? Like:
https://github.com/CenterForOpenScience/osf.io/blob/847824cf86ddd7403a9b18b1ff3278cd8399ab63/website/prereg/views.py#L26

samchrisinger:
One comment, otherwise lgtm

scripts/migration/migrate_registration_extra.py
barbour-em: No, there isn't. Done in fddd5e6

barbour-em:
@nicipfeiffer how's this?
screen shot 2016-04-25 at 8 55 09 am

nicipfeiffer:
@barbour-em Looks good, thank you!

samchrisinger:
@barbour-em do we need to migrate metadata for existing Prereg registrations?

barbour-em:
Thinking about displaying the registration supplement, yes. What should that query look like?

```python
Node.find(
    Q('is_registered', 'eq', True) &
    Q(PREREG_CHALLENGE_METASCHEMA, 'in', 'registered_schema') 
)
```
?

samchrisinger:
^ Maybe add a `&

PR #4992, opened 2016-02-04
User Registration and Admin Updates

zkraime:
originally: #4718 and https://github.com/CenterForOpenScience/osf.io/pull/4796

PR #5478, opened 2016-04-14
[Feature] Admin prereg refactor

Ghalko:
## Purpose
Overhaul the backend of prereg.

## Changes
Overhaul of the backend to make several things easier. Plus a better initial list with modals.

## Side effects
Hopefully easier to add David's requests.

## Ticket

ADM-87

PR #5496, opened 2016-04-18
[Feature] Admin Registration/Passwords Refactor

Ghalko:
## Purpose
Finish #4992
## Changes

Cleaning up and conforming to Rest of Admin module

## Ticket
ADM-41, ADM-47


admin/common_auth/admin.py
```diff
@@ -39,22 +39,14 @@ def send_email_invitation(self, request, queryset):
             reset_form = PasswordResetForm({'email': user.email}, request.POST)
             assert reset_form.is_valid()
             reset_form.save(
-                #subject_template_name='templates/emails/account_creation_subject.txt',
-                #email_template_name='templates/emails/invitation_email.html',
+                subject_template_name='common_auth/emails/account_creation_subject.txt',
```
cwisecarver: Neither of these appear to be there. Looks like the emails directory doesn't actually exist in common_auth unless I'm looking in the wrong root directory.

admin/common_auth/models.py
```diff
@@ -39,14 +39,14 @@ class MyUser(AbstractBaseUser, PermissionsMixin):
...
     )
 
-    first_name = models.CharField(max_length=30, blank=True)
-    last_name = models.CharField(max_length=30, blank=True)
+    first_name = models.CharField(max_length=30, blank=False)
+    last_name = models.CharField(max_length=30, blank=False)
     is_active = models.BooleanField(default=True)
     is_admin = models.BooleanField(default=False)
     is_staff = models.BooleanField(default=False)
     date_joined = models.DateTimeField(default=datetime.now)
```
cwisecarver: Django has a way to [do this on add](https://docs.djangoproject.com/en/1.9/ref/models/fields/#django.db.models.DateField.auto_now_add). If you're migrating older data you'll have to decide what to do about the old ones.

admin/common_auth/models.py
```diff
@@ -39,14 +39,14 @@ class MyUser(AbstractBaseUser, PermissionsMixin):
...
     confirmed = models.BooleanField(default=False)
-    osf_id = models.CharField(default=False, max_length=10, blank=True)
+    osf_id = models.CharField(default='', max_length=5, blank=True)
```
cwisecarver: I believe setting default and blank to True throws a warning.

admin/common_auth/views.py
```diff
@@ -47,3 +55,45 @@ def get_success_url(self):
...
+
+
+class RegisterUser(SuperUser, FormView):
+    form_class = UserRegistrationForm
+    template_name = 'register.html'
+
+    def form_valid(self, form):
+        osf_id = form.cleaned_data.get('osf_id')
+        try:
+            osf_user = User.load(osf_id)
```
cwisecarver: Unless you expect `osf_user = User.load(osf_id)` to throw an AttributeError I'd pull it out of the try. It just limits the possibility you'll catch an exception that you weren't expecting.

admin/common_auth/views.py
```diff
@@ -47,3 +55,45 @@ def get_success_url(self):
...
+            osf_user.system_tags.append(PREREG_ADMIN_TAG)
+        except AttributeError:
+            return page_not_found(
```
cwisecarver: Again, I'd raise `Http404('OSF user with id "{}" not found. Please double check.'.format(osf_id))` just for modularity.

cwisecarver:
@Ghalko Clicking on "Forgot my password" in an incognito window gives me this:
screen shot 2016-04-26 at 12 18 15 pm

admin/base/urls.py
```diff
@@ -10,21 +9,19 @@
...
+    url(
+        base_pattern,
         include([
             url(r'^$', views.home, name='home'),
-            url(r'^django_admin/', include(admin.site.urls)),
+            url(r'^admin/', include(admin.site.urls)),
             url(r'^spam/', include('admin.spam.urls', namespace='spam')),
-            url(r'^auth/', include('admin.common_auth.urls', namespace='auth')),
+            url(r'^account/', include('admin.common_auth.urls', namespace='auth')),
+            url(r'^password', include('password_reset.urls')),
```
cwisecarver: Think this needs a trailing slash.

admin/pre_reg/utils.py
```diff
@@ -5,10 +5,14 @@
...
 
 def get_prereg_reviewers():
-    return MyUser.objects.filter(
-        groups__name='prereg_group'
-    ).annotate(
-        fuller_name=Concat('first_name', Value(' '), 'last_name')
-    ).values_list(
-        'email', 'fuller_name'
-    )
+    # Note - fixes django.db.utils.OperationalError: no such table' error if ever one erases a db table and tries to remigrate
```
cwisecarver: I think, if you moved the call for this from the form __init__ method to the form's get_initial method you wouldn't have this exception happen. Either way, the clause is too wide. You should trim it down to just OperationError. Lots of other things could go wrong in this statement and the wide exception handling would make it quite the mystery.

cwisecarver:
Pass done. Back to you @Ghalko

admin/common_auth/admin.py
Ghalko: Changed strings to reflect actual files.

admin/common_auth/models.py
Ghalko: Changed

admin/common_auth/models.py
Ghalko: Removed default

admin/common_auth/views.py
Ghalko: Right, done.

admin/common_auth/views.py
Ghalko: Done.

admin/base/urls.py
Ghalko: Fixed, seems to have fixed the error above too.

admin/pre_reg/utils.py
Ghalko: I think this would be better addressed in #5478, and I would put it in the `FormView`'s `get_initial`.

Ghalko:
@cwisecarver This is done.

cwisecarver:
👍

sloria:
Can you please update this with develop one last time, @Ghalko ?

sloria:

PR #5502, opened 2016-04-19
Add mendeley auto_refresh_url [OSF-5853]

samanehsan:
## Purpose

Some projects on production (e.g. https://osf.io/fx86p/) are not loading Mendeley citations. @mfraezz, believes this is due to expired account credentials. This PR auto-refreshes credentials (which expire after 1 hour).

## Changes

Add `auto_refresh_url` to `Mendeley` model and call `refresh_oauth_key()` to get a new access token if `_verify_client_validity` fails.

## Side effects
None.

## Ticket
https://openscience.atlassian.net/browse/OSF-5853


website/addons/mendeley/model.py
```diff
@@ -68,11 +70,14 @@ def _verify_client_validity(self):
         try:
             self._client.folders.list()
         except MendeleyApiException as error:
-            self._client = None
-            if error.status == 403:
-                raise HTTPError(403)
+            if error.status == 401 and 'Token has expired' in error.message:
+                self.refresh_oauth_key()
```
mfraezz: This can raise an `InvalidAuthError`, and it isn't guaranteed to succeed; either another Mendeley API call needs to be made to see if it worked, or `refresh_oauth_key` needs to be refactored to return a boolean if it doesn't raise anything.

samanehsan: Can I do a try/except here?

```
try:
    self.refresh_oauth_key()
except InvalidAuthError:
    raise HTTPError(401)

**Comment by samanehsan** [comment]
Done making that change @mfraezz! 🎉

File: website/addons/mendeley/model.py [hunk 2]
```diff
@@ -68,11 +72,17 @@ def _verify_client_validity(self):
         try:
             self._client.folders.list()
         except MendeleyApiException as error:
-            self._client = None
-            if error.status == 403:
-                raise HTTPError(403)
+            if error.status == 401 and 'Token has expired' in error.message:
+                try:
+                    self.refresh_oauth_key()
```
mfraezz: Still needs to ensure that this actually refreshes the key.

From the outdated comment:
>[This call] isn't guaranteed to succeed; either another Mendeley API call needs to be made to see if it worked, or `refresh_oauth_key` needs to be refactored to return a boolean if it doesn't raise anything.

samanehsan: Could `refresh_oauth_key` do this check to make sure a new key was returned?

mfraezz: Yes, it should be a fairly simple refactor; `return False` in the three initial failure cases, don't change the error handling, and `return True` if `self.account.<oauth_things>` are updated properly.

website/oauth/models/__init__.py
```diff
@@ -423,10 +423,13 @@ def refresh_oauth_key(self, force=False, extra={}, resp_auth_token_key='access_t
         except InvalidGrantError:
             raise InvalidAuthError()
         else:
-            self.account.oauth_key = token[resp_auth_token_key]
-            self.account.refresh_token = token[resp_refresh_token_key]
-            self.account.expires_at = resp_expiry_fn(token)
-            self.account.save()
+            if self.account.oauth_key != token[resp_auth_token_key]:
```
mfraezz: Do you need to do this check here? It should fail and raise an exception if if the refresh request didn't work.

It shouldn't be possible if third parties follow the OAuth2 RFC, but I could imagine a poorly-constructed service that, instead of generating a new token, just extends the expiry time of the first.

website/addons/mendeley/model.py
```diff
@@ -68,11 +71,19 @@ def _verify_client_validity(self):
...
             self._client.folders.list()
         except MendeleyApiException as error:
-            self._client = None
-            if error.status == 403:
-                raise HTTPError(403)
+            if error.status == 401 and 'Token has expired' in error.message:
+                try:
+                    refreshed_key = self.refresh_oauth_key()
+                except InvalidAuthError:
+                    raise HTTPError(401)
```
mfraezz: You should still be setting `self._client = None` in failure cases

website/oauth/models/__init__.py
samanehsan: Ok, in that case we probably don't need to do that check

mfraezz:
LGTM. Dropping this in RTM 💾

sloria:
Since this is affecting users on prod. I'm going to wrap this into a hotfix.

PR #5505, opened 2016-04-20
"Publish" confirmation modal should say the name of the correct Dataverse [OSF-4556]

asanfilippo7:
## Purpose

Currently, when a user clicks the "Publish" button to publish a Dataverse or Dataverse dataset through the OSF, the popup modal asks if they want to publish to "The Harvard Dataverse," regardless of which Dataverse is actually hosting their project.

## Changes

- Added an Ajax call that gets the Dataverse host when a user hits "Publish"
- Changed the text in the modal to display this host instead of automatically displaying Harvard
- Added an error message in the modal and console if the Ajax call fails

## Side effects

Right now the host name that is displayed is in url format, e.g. "demo.dataverse.org," "dataverse.harvard.edu," etc. It would be nice to have this in more of an actual name format, but because there are many Dataverse repositories out there and we probably don't know exactly how to match their urls to their names, I'm not sure if this is possible at the moment.

## Ticket

[https://openscience.atlassian.net/browse/OSF-4556](url)


website/addons/dataverse/static/dataverseFangornConfig.js
```diff
@@ -29,29 +29,55 @@ var _dataverseItemButtons = {
             tb.dropzone.hiddenFileInput.click();
             tb.dropzoneItemCache = item;
         }
-        function dataversePublish(event, item, col) {
+        // Get the external Dataverse host before publishing
+        function getDataverseHost(event, item, col) {
+            var host;
+            $.ajax({
```
samanehsan: Instead of sending a request to get the Dataverse host, you can pass in that information from the view so that it's available under `item.data` (`item.data.dataverseIsPublished` is a good example of this)

website/static/js/_allLogTexts.json
```diff
@@ -0,0 +1 @@
...} removed ${path} in Amazon S3 bucket ${bucket} in ${node}","s3_file_updated":"${user} updated ${path} in Amazon S3 bucket ${bucket} in ${node}","s3_folder_created":"${user} created folder ${path} in Amazon S3 in ${node}","s3_node_authorized":"${user} authorized the Amazon S3 addon for ${node}","s3_node_deauthorized":"${user} deauthorized the Amazon S3 addon for ${node}","s3_node_deauthorized_no_user":"Amazon S3 addon for ${node} deauthorized","googledrive_file_added":"${user} added file ${googledrive_path} to Google Drive in ${node}","googledrive_file_removed":"${user} removed file ${googledrive_path} from Google Drive in ${node}","googledrive_file_updated":"${user} updated file ${googledrive_path} in Google Drive in ${node}","googledrive_folder_created":"${user} created folder ${googledrive_path} in Google Drive in ${node}","googledrive_folder_selected":"${user} linked Google Drive folder ${googledrive_folder} to ${node}","googledrive_node_authorized":"${user} authorized the Google Drive addon for ${node}","googledrive_node_deauthorized":"${user} deauthorized the Google Drive addon for ${node}","googledrive_node_deauthorized_no_user":"Google Drive addon for ${node} deauthorized"}
```
samanehsan: Was this file committed accidentally?

samanehsan:
Pass complete @asanfilippo7! 🏇

website/addons/dataverse/static/dataverseFangornConfig.js
asanfilippo7: @samanehsan I tried doing this. Unless I'm missing something, though, none of the information stored in item.data actually references the host. For example, here is item.data for a dataset on apitest.dataverse.org:

screen shot 2016-04-25 at 2 40 38 pm

samanehsan: @asanfilippo7, the info in `item.data` is coming from a view function somewhere in the dataverse addon code (probably `dataverse/views.py`, so even though the host info is not currently in `item.data`, it's possible to add it in by updating what's returned by that view function. That way, you don't have to make an additional request to get that info.

asanfilippo7:
@samanehsan this should be fixed now

samanehsan:
Thanks @asanfilippo7! Can you pls fix the failing build?

samanehsan:
This is good to merge once the build passes! 🌂

PR #5506, opened 2016-04-20
[Feature][Migration] Add SECURE flag for OSF, API and ADMIN cookie [#OSF-4901, #SEC-16]

cslzchen:
### Main
Add `SECURE` flag in OSF, API and ADMIN cookie, which ensures that cookies are transferred over https.
### Side
Add HTTPS for osf (Flask) and apiserver, admin (Django) as an option for local development.
### Note
`HTTPONLY` flag is addressed in another ticket [OSF-6339](https://openscience.atlassian.net/browse/OSF-6339).

### Local HTTP

To run OSF normally, please update `local.py` by adding the following code:
```
SECURE_MODE = False  # disable osf cookie secure for local development
PROTOCOL = 'https://' if SECURE_MODE else 'http://'
DOMAIN = PROTOCOL + 'localhost:5000/'
API_DOMAIN = PROTOCOL + 'localhost:8000/'
SESSION_COOKIE_SECURE = SECURE_MODE
```

### Local Https

For how to run HTTPS for local development, refer to [OSF-4901](https://openscience.atlassian.net/browse/OSF-4901). Our DevDocs will be updated accordingly when deployed.

### Staging, Staging2, Production, Test

Double check all (osf, admin, api, waterbutler, mfr, etc.) run https by default (http will 301 redirect to https) and make sure that `DEBUG_MODE=False` and add the following or similar to respective settings:

```
SECURE_MODE = not DEBUG_MODE  # Set secure cookie
SESSION_COOKIE_SECURE = SECURE_MODE
# TODO: Change to True after ticket #OSF-6339 has been resolved
SESSION_COOKIE_HTTPONLY = False
```

## Side effects

Need to update `local.py`.

##  QA Test

On staging, staging2, test and production:
- Nothing breaks
- Any http:// request to ".osf.io" should not contain OSF cookie. The following redirected https:// request should contain the cookie (if there is one).

## Ticket

[OSF-4901](https://openscience.atlassian.net/browse/OSF-4901)
[SEC-16](https://openscience.atlassian.net/browse/SEC-16)


abought:
Just to be safe, we should also set the appropriate django and flask config variables in the appropriate places. (based on the new config value you chose here)

tasks/__init__.py
```diff
@@ -66,7 +66,13 @@ def server(host=None, port=5000, debug=True, live=False, gitlogs=False):
         server.watch(os.path.join(HERE, 'website', 'static', 'public'))
         server.serve(port=port)
     else:
-        app.run(host=host, port=port, debug=debug, threaded=debug, extra_files=[settings.ASSET_HASH_PATH])
+        if https:
+            # for local https test, generate self-signed cert (and key if necessary) and put the path in local.py
+            context = (settings.OSF_SERVER_CERT, settings.OSF_SERVER_KEY)
+            app.run(host=host, port=5000, debug=debug, threaded=debug, extra_files=[settings.ASSET_HASH_PATH], ssl_context=context)
```
abought: DRY with line 74 as discussed.

website/settings/defaults.py
```diff
@@ -79,11 +79,16 @@ def parent_dir(path):
 SHARE_ELASTIC_INDEX_TEMPLATE = 'share_v{}'
 
 # Sessions
+COOKIE_NAME = 'osf'
 # TODO: Override OSF_COOKIE_DOMAIN in local.py in production
 OSF_COOKIE_DOMAIN = None
-COOKIE_NAME = 'osf'
+# TODO: Override OSF_COOKIE_SECURE in local.py in production
+COOKIE_SECURE = False
```
abought: Might be better to rename this to something more generic, so that one setting controls all behavior required to turn on SSH for local development.

cslzchen:
@abought @brianjgeiger Ready for review.

admin/base/settings/defaults.py
```diff
@@ -20,6 +20,15 @@
 DEBUG = osf_settings.DEBUG_MODE
 DEBUG_PROPAGATE_EXCEPTIONS = True
 
+# secure mode for using https in local development
+SECURE = osf_settings.SECURE_MODE
+LOCAL = osf_settings.LOCAL_MODE
```
abought: These variables can probably be removed: `LOCAL` is unused, and `SECURE` makes more sense as a direct reference to the parent setting.

api/base/settings/defaults.py
```diff
@@ -29,6 +29,14 @@
 DEBUG = osf_settings.DEBUG_MODE
 DEBUG_PROPAGATE_EXCEPTIONS = True
 
+# secure mode for using https in local development
+SECURE = osf_settings.SECURE_MODE
+LOCAL = osf_settings.LOCAL_MODE
```
abought: See above note; these two variables can probably be removed.

tasks/__init__.py
```diff
@@ -66,7 +66,12 @@ def server(host=None, port=5000, debug=True, live=False, gitlogs=False):
         server.watch(os.path.join(HERE, 'website', 'static', 'public'))
         server.serve(port=port)
     else:
-        app.run(host=host, port=port, debug=debug, threaded=debug, extra_files=[settings.ASSET_HASH_PATH])
+        if settings.SECURE_MODE:
+            context = (settings.OSF_SERVER_CERT, settings.OSF_SERVER_KEY)
+        else:
+            context = None
+        app.run(host=host, port=port, debug=debug, threaded=debug, extra_files=[settings.ASSET_HASH_PATH], ssl_context=context)
```
abought: Can you clarify why you need a pre-made SSL context for this to work? (eg, is this a requirement because of the different servers working together?)

I notice that flask's `app.run` has an "ad-hoc" mode to generate this on the fly, and am wondering why we chose not to go that route. (it seems like it wuld save a step for anyone using this feature)

Adhoc mode is per the documentation of the thing flask's `app.run` is based on:
http://werkzeug.pocoo.org/docs/0.11/serving/#werkzeug.serving.run_simple
> ssl_context – an SSL context for the connection. Either an ssl.SSLContext, a tuple in the form (cert_file, pkey_file), the string 'adhoc' if the server should automatically create one, or None to disable SSL (which is the default).

abought:
Yesterday there was some discussion of whether we also needed to set `CSRF_COOKIE_SECURE` for Django/ admin. Can you clarify what decision was made and why?
https://docs.djangoproject.com/en/1.9/ref/settings/#csrf-cookie-secure

(I don't see that setting used in this PR)

requirements/secure.txt
```diff
@@ -0,0 +1,6 @@
+-r ./dev.txt
+
+# Requirements that are used in the DEVELOPMENT and LOCAL environment only
+
+# django-sslserver
```
abought: This comment is probably unnecessary given the line below. :)

requirements/secure.txt
```diff
@@ -0,0 +1,6 @@
+-r ./dev.txt
```
abought: Can you explain why `dev.txt` dependencies are included in this file?

(as opposed to this being a totally separate flag)

admin/base/settings/defaults.py
```diff
@@ -20,6 +20,15 @@
...
+
+# session:
+SESSION_COOKIE_HTTPONLY = True
```
abought: Can you please add a TODO to change this to `settings.SESSION_COOKIE_HTTPONLY` in the future? I understand the reasons for hard-coding the value for the admin tool... but we don't want to forget to make things consistent later.

Be sure to cite the relevant JIRA ticket number in the source code comment, so we can link the TODO to a known issue in our tracker.

website/settings/defaults.py
```diff
@@ -79,11 +87,15 @@ def parent_dir(path):
 SHARE_ELASTIC_INDEX_TEMPLATE = 'share_v{}'
 
 # Sessions
-# TODO: Override OSF_COOKIE_DOMAIN in local.py in production
-OSF_COOKIE_DOMAIN = None
 COOKIE_NAME = 'osf'
-# TODO: Override SECRET_KEY in local.py in production
+OSF_COOKIE_DOMAIN = None
```
abought: The removed TODO items served as deployment notes- was there any particular reason for removing them?

website/settings/local-dist.py
```diff
@@ -32,9 +39,15 @@
...
-OSF_COOKIE_DOMAIN = None
 COOKIE_NAME = 'osf'
-SECRET_KEY = "CHANGEME"
+OSF_COOKIE_DOMAIN = None
+SECRET_KEY = 'CHANGEME'
+SESSION_COOKIE_SECURE = SECURE_MODE
+SESSION_COOKIE_HTTPONLY = False
+
+# certificate for local SECURE_MODE development
+OSF_SERVER_KEY = 'PATH_TO_LOCAL_PRIVATE_KEY'
```
abought: Are these paths to a file?

If so, can you please change the default value to `None`, and make "path to local private key" a source code comment instead? 

Explicitly indicating that no default value is provided can help avoid confusing error messages that happen when some placeholder text gets treated like an actual value.

abought:
Pass complete. There are a few places where I think the code can be cleaned up or simplified, none of them very major.

Could you take care of those items, and also answer questions where noted? (eg regarding the `CSRF_COOKIE_SECURE` setting)

cslzchen:
@abought Thanks a bunch for your review. It is very helpful. I will answer those questions as soon as I finish my current job on Keen.

abought:
Also: thanks for the helpful testing and deployment notes. Confirmed that I can run as https locally and that OSF cookie now has secure flag.

Is there a separate issue tagging the changes to CAS cookie?

cslzchen:
I created another issue [DEV-93](https://openscience.atlassian.net/browse/DEV-93). It is in the CAS repo not OSF.

admin/base/settings/defaults.py
cslzchen: `LOCAL` will be removed.
I prefer to keep admin and osf sharing the same settings since they share other services. `SECURE` can be removed, `SESSION_COOKIE_SECURE = osf_settings.SECURE_MODE`.

admin/base/settings/defaults.py
cslzchen: Agree. 
\[FYI\] Django's default value for `SESSION_COOKIE_HTTPONLY` is true. It is not an issue. But we still want to make the two consistent.

requirements/secure.txt
cslzchen: As we discusses, there no need to recursively include dev.txt since `secure` is specific and rare use case.

tasks/__init__.py
cslzchen: As we discussed, it is OK for us to use Flask `ad_hoc` to generate certs on the fly. However, this will cause the browser raise trust dialog overtime server restarts. In additions, django-ssl server have a hardcoded cert which is very dangerous to trust.

website/settings/defaults.py
cslzchen: I didn't realize that I had removed them and will fix this.

website/settings/local-dist.py
cslzchen: Agree.

requirements/secure.txt
```diff
@@ -1,6 +1,3 @@
--r ./dev.txt
+# Requirements for running https in local development (apiserver and amdinserver)
```
abought: Possible typo- should this be `adminserver`?

website/settings/defaults.py
```diff
@@ -52,12 +52,15 @@ def parent_dir(path):
 DEV_MODE = False
 DEBUG_MODE = False
 
-# Should be True in local development and False on staging, staging2, production
+# True if in local development and False if on staging, staging2, production
```
abought: Note to self: the distinguishing feature of local mode is that someone is running via the built-in testing server of the framework (`app.run`, `manage.py runserver`, etc) rather than an external server via wsgi.

requirements/secure.txt
cslzchen: typo it is, will fix this

abought:
Changes look ok to me. (typo aside)

One other request: this PR will affect how the app is deployed, so it might take a bit more effort than the usual automatic deployment behavior on staging.  Can you please add deployment notes to the PR description, and/or coordinate with either @sloria or @mattclark before this gets merged?

cslzchen:
Sure, I will do this now. Thanks again for the review.

tasks/__init__.py
```diff
@@ -79,7 +84,11 @@ def git_logs():
 def apiserver(port=8000, wait=True, host='127.0.0.1'):
     """Run the API server."""
     env = os.environ.copy()
-    cmd = 'DJANGO_SETTINGS_MODULE=api.base.settings {} manage.py runserver {}:{} --nothreading'.format(sys.executable, host, port)
+    cmd = 'DJANGO_SETTINGS_MODULE=api.base.settings {} manage.py runserver {} --nothreading'.format(sys.executable, port)
```
sloria: Why was `host` removed?

tasks/__init__.py
```diff
@@ -940,6 +963,7 @@ def generate_self_signed(domain):
     ).format(domain)
     run(cmd)
 
+
```
sloria: FWIW, our flake config explicitly ignores the "2-newline" rule since it leads to unnecessary whitespace. This is fine though.

sloria:
Discussed in person: even when `SECURE_MODE = False`, I am unable to log in with fakecas because the login redirects to a https URL.

sloria:
Pass finished.

tasks/__init__.py
cslzchen: Discussed in person: `host` was not there then, looks like merge conflicts, will fix this.

sloria:
After correctly setting the PROTOCOL and DOMAIN settings in my local.py, login no longer redirects to a https URL, but I still can't log in.

cslzchen:
I only tested realCAS. Need to test fakeCAS. And I will add note on updating `local.py`. Thanks for the review.

cslzchen:
@sloria
fakeCAS works without any issue, could be `local.py` issue. Here is the changes for running local http:
```
SECURE_MODE = not DEBUG_MODE  # disable osf cookie secure and httponly
PROTOCOL = 'https://' if SECURE_MODE else 'http://'
DOMAIN = PROTOCOL + 'localhost:5000/'
API_DOMAIN = PROTOCOL + 'localhost:8000/'
SESSION_COOKIE_SECURE = SECURE_MODE
```

sloria:
I am a go with this when @mattclark is ready with the necessary config.

PR #5507, opened 2016-04-20
Remove "publish" option for draft datasets that have an associated published dataset [OSF-4553]

asanfilippo7:
## Purpose

Currently, when a user has a published Dataverse dataset connected to their project, they are given the option to "Publish" this dataset again. This occurs via a "Publish" button in the UI that appears when the user selects the "Draft" version in the dropdown menu. Users shouldn't be allowed to double-publish a dataset, so this button should not appear.

## Changes

- Separated the logic that displays the "Publish" button into a conditional statement. This statement only executes if the dataset doesn't have an associated published version, regardless of which version the user is actually viewing.

## Ticket

[https://openscience.atlassian.net/browse/OSF-4553](url)


samanehsan:
Looks good to me @asanfilippo7! ☀️

PR #4236, opened 2015-09-03
[Ready] Fixes the osfToggleHeight initialization in user settings 

caneruguz:
## Purpose
Fixes JIRA issue: OSF-4265
Where the application of the osfToggleHeight plugin to addon node list was unreliable. 

## Changes 
There are two types of data load methods for addons
1. Those that get added by backend and mako file addon_permissions.mako
-- For these the document.ready function was used to apply
2. Asyncronous loading with ajax after load 
-- For these the initialization was added where nodes have been applied 

## Side effects
None 

screen shot 2015-09-03 at 12 00 12 pm

PR #5508, opened 2016-04-20
OSF-4265 Fix strange behavior of user settings page expand/collapse caret

adlius:
## Purpose

User settings page's section for configuring addons have strange behaviors about the expand/collapse caret of each addon. (see https://openscience.atlassian.net/browse/OSF-4265)

## Changes

@caneruguz 's original PR(https://github.com/CenterForOpenScience/osf.io/pull/4236) on this bug was supposed to deal with the fact that different addon's are loaded differently as 

>There are two types of data load methods for addons
>1. Those that get added by backend and mako file addon_permissions.mako
>-- For these the document.ready function was used to apply
>2. Asyncronous loading with ajax after load 
>-- For these the initialization was added where nodes have been applied

and, judging from QA's feedback in the ticket, the fix did work at first. The ticket was re-opened during regression testing. I checked the current develop and the changes made in the PR are not showing. It is possible that some later PR reverted the merge.
Thus I replicate the changes made in the PR by Caner. (All such changes are included in the first commit) Somehow the expand/collapse caret's behavior for all addons are still unpredictable on each refresh of the page, and the fix does not seem to work on Safari at all (No carets will show). As QA have noticed, the caret's will show upon resizing the page. Thus, in the "profile-settings-addon-page.js", a line is added to programmatically trigger a resize event of the browser after the page is fully loaded. (The second commit)

## Side effects

Safari seems to block the resize event so that the fix still does not work on it. Works on Chrome and Firefox. Need more tests on IEs.

Don't know why the fix in the original PR worked before but not now. Maybe other components related to this bug changed between then and now. And the fix in the second commit is quite hackish I'll admit.

## Ticket

https://openscience.atlassian.net/browse/OSF-4265

### DIFF
--- a/website/static/js/pages/profile-settings-addons-page.js
+++ b/website/static/js/pages/profile-settings-addons-page.js
@@ -122,6 +122,10 @@ $(document).ready(function(){
     });
 });
 
+$(window).load(function () {
+    $(window).trigger('resize');
+});
+
 /* Before closing the page, Check whether the newly checked addon are updated or not */
 $(window).on('beforeunload',function() {
     //new checked items but not updated
--- a/website/static/js/pages/profile-settings-addons-page.js
+++ b/website/static/js/pages/profile-settings-addons-page.js
@@ -115,7 +115,11 @@ for (var i=0; i < addonEnabledSettings.length; i++) {
 }
 
 $(document).ready(function(){
-    $('.addon-auth-table').osfToggleHeight({height: 140});
+// Separately load addons whose data is loaded on the backend as opposed to async javascript.
+    var makoLoadedAddons = ['dropbox', 'github', 'box', 'figshare', 'googledrive', 's3'];
+    makoLoadedAddons.forEach(function(addon){
+        $('#'+ addon + '-header').osfToggleHeight({height: 140});
+    });
 });
 
 /* Before closing the page, Check whether the newly checked addon are updated or not */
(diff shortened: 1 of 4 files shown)


samanehsan:
@adlius, can you look into how to fix this on Safari?

website/static/js/pages/profile-settings-addons-page.js
```diff
@@ -135,6 +143,8 @@ $(window).on('beforeunload',function() {
 ****************/
 
 $('.addon-oauth').each(function(index, elem) {
+// Applies to Mendeley, Zotero.  There is a separate code for dataverse in DataverseUserConfig.js
```
samanehsan: I think almost all of the addons are using `OAuthAddonSettingsViewModel` now  (except Dataverse and s3 I believe) so this comment isn't quite right

samanehsan:
Does `s3UserConfig.js` also have to be updated? https://github.com/CenterForOpenScience/osf.io/blob/develop/website/addons/s3/static/s3UserConfig.js#L54

samanehsan:
Pass complete @adlius! ☀️

website/static/js/pages/profile-settings-addons-page.js
adlius: Ah, that's probably the culprit. I'll look into it and see what I can do.

PR #5511, opened 2016-04-21
Bold warning on access token creation [#OSF-6058]

erinspace:
## Purpose
When creating an access token on the OSF on your settings page, it's not clear that the first time the token is created is the last time you'll be able to view it. This PR adds a bolder warning to the first time the token is displayed, as well as a "copy to clipboard" button to help facilitate using the code later on in an actual application.

## Changes
- Add clipboardjs to npm installable packages
    - copies text to the clipboard without flash!
- Add warning text to first time token is displayed
- Move Cancel, Save, and Delete Buttons to above token display
    - These buttons deal directly with any potential token changes, not with the new token that was just created. 
- Small CSS style changes

## Side effects
- Not sure about how well the copy to clipboard button will work on IE - tested locally as much as possible

## Ticket
https://openscience.atlassian.net/browse/OSF-6058

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "bootbox": "^4.4.0",
     "bower": "^1.3.12",
     "c3": "^0.4.10",
+    "clipboard": "^1.5.10",
     "css-loader": "^0.9.1",
     "diff": "~2.2.2",
     "dropzone": "git://github.com/sloria/dropzone.git#accept-directory",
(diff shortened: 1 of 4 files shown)


mfraezz:

The code looks fine, but after consulting about the UI, it's been requested that you make it look more like GitHub's:

screen shot 2016-04-21 at 11 43 56
screen shot 2016-04-21 at 11 45 07

mfraezz:
LGTM. Dropping this in RTM :fire:

sloria:
Ah, merged too soon; we are already using ZeroClipboard for clipboard access; @erinspace can you look into using this so we don't add another dependency?

erinspace:
@sloria Looked into that for sure - the reason I switched to clipboard js is it's pure javascript - ZeroClipboard uses "an invisible Adobe Flash movie," and I thought this was a way nicer solution.

PR #5512, opened 2016-04-21
Add user entry point tracking [#OSF-6070]

jeffreyliu3230:
## Purpose

Keep track of the user entry points (OSF, OSF for Meetings, Preregistration, Institution) on Keen.

## Changes

Add user_entry_point to global variables. Add entry_point property to currentUser in base.mako.

## Ticket

https://openscience.atlassian.net/browse/OSF-6070


felliott:
LGTM!

sloria:
Why do we need to put this into the templates/JS? Couldn't we just use a script that queries the database for system tags?

jeffreyliu3230:
We can. Do you mean write a function in python to get the system tags then pass it as context?

felliott:
@sloria, we added it to keen so that we can do user-origin focused queries entirely within keen.  E.g. "What are the most popular pages for institutional users?"  "Do users who join via OSF4M check out other parts of our site"

website/util/metrics.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+
+
+def get_entry_point(system_tags):
```
sloria: Minor: it would be more intuitive to pass a User as the input here; that way, if we decide to store the entry point in some other way that doesn't use system_tags, we can do so without changing calls to this function.

sloria:
@felliott Thanks for the clarification

@jeffreyliu3230 Just one minor suggestion above; otherwise good to merge.

jeffreyliu3230:
Updated!

jeffreyliu3230:
@sloria I changed the name of the context variable to follow the naming rule. Should be ready to merge.

PR #5515, opened 2016-04-21
[Feature] Admin Meetings/Conferences

Ghalko:
## Purpose

Add the ability for an admin user to view/update/create meetings for OSF

## Changes

Makes a new module that serializes current meetings, allows for updates, and allows for new meetings to be created.

## Side effects

No more running a gigantic script full of past meetings in order to populate meetings.

## Ticket
ADM-57


admin/meetings/forms.py
```diff
@@ -0,0 +1,128 @@
...
+        label='Message body for mail',
+        widget=forms.TextInput(attrs={'size': '60'}),
+    )
+    field_mail_attachment = forms.CharField(
+        label='Mail attachment message',
+        widget=forms.TextInput(attrs={'size': '60'}),
+    )
+
+    def clean_endpoint(self):
+        data = self.cleaned_data['endpoint']
```
cwisecarver: Just a style thing, I might rename `data` to `endpoint` or something, the same with `data` in `clean_admins`, change to `emails`. Just so it's more readable.

admin/meetings/views.py
```diff
@@ -0,0 +1,151 @@
...
+        kwargs.setdefault('meetings', map(serialize_meeting, queryset))
+        kwargs.setdefault('page', page)
+        return super(MeetingListView, self).get_context_data(**kwargs)
+
+
+class MeetingFormView(OSFAdmin, FormView):
+    template_name = 'meetings/detail.html'
+    form_class = MeetingForm
+
+    def get(self, request, *args, **kwargs):
```
cwisecarver: These could probably go away and you could raise the correct `HttpXXX` exception with an error message.

admin/meetings/views.py
```diff
@@ -0,0 +1,151 @@
...
+                                           self.kwargs.get('endpoint'))
+
+    def get_context_data(self, **kwargs):
+        kwargs.setdefault('meeting', serialize_meeting(
+            Conference.get_by_endpoint(self.kwargs.get('endpoint'))
+        ))
+        return super(MeetingFormView, self).get_context_data(**kwargs)
+
+    def get_initial(self):
+        self.initial = serialize_meeting(
```
cwisecarver: I don't think you need to get and serialize the conference by endpoint both here and in `get_context_data`. Is there a reason?

admin/meetings/views.py
```diff
@@ -0,0 +1,151 @@
...
+            Conference.get_by_endpoint(self.kwargs.get('endpoint'))
+        )
+        self.initial.setdefault('edit', True)
+        return super(MeetingFormView, self).get_initial()
+
+    def form_valid(self, form):
+        custom_fields, data = get_custom_fields(form.cleaned_data)
+        # Form validation already catches if there is no conference
+        conf = Conference.get_by_endpoint(data.get('endpoint'))
```
cwisecarver: I feel like you could get this conference somewhere early in the request cycle and stick it onto self for the duration and just use it everywhere.

admin/meetings/views.py
```diff
@@ -0,0 +1,151 @@
...
+        return reverse('meetings:detail',
+                       kwargs={'endpoint': self.kwargs.get('endpoint')})
+
+
+class MeetingCreateFormView(OSFAdmin, FormView):
+    template_name = 'meetings/create.html'
+    form_class = MeetingForm
+
+    def get_initial(self):
+        data = Conference.DEFAULT_FIELD_NAMES
```
cwisecarver: Maybe default_field_names instead of data?

admin/meetings/forms.py
Ghalko: Done.

cwisecarver:
@Ghalko do I need to run any migrations for this to work correctly? I get 404s for all the meetings on the first page of the listing and if I go to the second page I get a 500 with this trace in the console.
```
  File "/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/wsgiref/handlers.py", line 85, in run
  File "/Users/chriswisecarver/.venvs/osf/lib/python2.7/site-packages/django/contrib/staticfiles/handlers.py", line 63, in __call__
  File "/Users/chriswisecarver/.venvs/osf/lib/python2.7/site-packages/django/core/handlers/wsgi.py", line 177, in __call__
  File "/Users/chriswisecarver/.venvs/osf/lib/python2.7/site-packages/django/core/handlers/base.py", line 230, in get_response
    response = self.handle_uncaught_exception(request, resolver, sys.exc_info())
  File "/Users/chriswisecarver/.venvs/osf/lib/python2.7/site-packages/django/core/handlers/base.py", line 149, in get_response
  File "/Users/chriswisecarver/.venvs/osf/lib/python2.7/site-packages/django/core/handlers/base.py", line 147, in get_response
  File "/Users/chriswisecarver/.venvs/osf/lib/python2.7/site-packages/django/views/generic/base.py", line 68, in view
  File "/Users/chriswisecarver/.venvs/osf/lib/python2.7/site-packages/django/contrib/auth/mixins.py", line 116, in dispatch
    return super(UserPassesTestMixin, self).dispatch(request, *args, **kwargs)
  File "/Users/chriswisecarver/.venvs/osf/lib/python2.7/site-packages/django/views/generic/base.py", line 88, in dispatch
    return handler(request, *args, **kwargs)
  File "/Users/chriswisecarver/.venvs/osf/lib/python2.7/site-packages/django/views/generic/list.py", line 174, in get
    context = self.get_context_data()
  File "/Users/chriswisecarver/Dropbox/cos-dev/osf.io/admin/meetings/views.py", line 31, in get_context_data
    kwargs.setdefault('meetings', map(serialize_meeting, queryset))
  File "/Users/chriswisecarver/Dropbox/cos-dev/osf.io/admin/meetings/serializers.py", line 10, in serialize_meeting
    'admins': ', '.join([u.emails[0] for u in meeting.admins]),
IndexError: list index out of range
```

cwisecarver:
I think it's because I have a User object that doesn't have any `emails`.

admin/meetings/views.py
Ghalko: Done.

admin/meetings/views.py
Ghalko: Nope, simplified `get_context_data` to only set endpoint if it isn't already set.

admin/meetings/views.py
Ghalko: Stuck it in `dispatch`. This also reduces the number of `Http404`s that need to be called. Removed the `handle_error` function since there is only one instance now.

admin/meetings/views.py
Ghalko: Done.

Ghalko:
As discussed, yes, without emails the user would be ineligible.

Also as for using the `get_user(email)` it has this line in it: https://github.com/CenterForOpenScience/osf.io/blob/develop/framework/auth/core.py#L120

So it is looking at the username as well as the emails. Though this line is also not very heartening as it seems it doesn't deal with lists very well.

website/conferences/model.py
```diff
@@ -27,19 +27,18 @@ class Conference(StoredObject):
...
+        'submission2': 'talk',
+        'submission1_plural': 'posters',
+        'submission2_plural': 'talks',
+        'meeting_title_type': 'Posters & Talks',
+        'add_submission': 'poster or talk',
+        'mail_subject': 'Presentation title',
+        'mail_message_body': 'Presentation abstract (if any)',
+        'mail_attachment': 'Your presentation file (e.g., PowerPoint, PDF, etc.)'
+    }
+    field_names = fields.DictionaryField(default=DEFAULT_FIELD_NAMES)
```
sloria: Mutable values should not be passed as the `default` argument, for the [same reason](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments) one shouldn't use mutable values for default keyword arguments. You can pass a callable to `default`, e.g. `default=lambda: DEFAULT_FIELD_NAMES`.

admin/meetings/forms.py
```diff
@@ -0,0 +1,129 @@
...
+    field_mail_attachment = forms.CharField(
+        label='Mail attachment message',
+        widget=forms.TextInput(attrs={'size': '60'}),
+    )
+
+    def clean_endpoint(self):
+        endpoint = self.cleaned_data['endpoint']
+        edit = self.cleaned_data['edit']
+        if not edit:
+            if Conference.find(Q('endpoint', 'iexact', endpoint)).count() > 0:
```
sloria: `iexact` queries are inefficient because they translate to Mongo regex queries. IIRC, all `endpoint` values should be lowercased, so you should just use `eq` and lowercase endpoint.

sloria: Actually, isn't this the same as `get_by_endpoint`? This branch and the `else` branch seem to be doing the same thing. Can you refactor?

admin/meetings/views.py
```diff
@@ -0,0 +1,129 @@
...
+        new_conf.save()
+        return super(MeetingCreateFormView, self).form_valid(form)
+
+    def get_success_url(self):
+        return reverse('meetings:detail',
+                       kwargs={'endpoint': self.kwargs.get('endpoint')})
+
+
+def get_custom_fields(data):
+    """Return two dicts, one with 'field' stripped
```
sloria: Why not just name the form fields the same as the db fields so that you don't have to do this preprocessing?

admin/meetings/forms.py
```diff
@@ -0,0 +1,129 @@
...
+from framework.auth.core import get_user
+from website.models import Conference
+from website.conferences.exceptions import ConferenceError
+
+
+class MultiEmailField(forms.Field):
+    def to_python(self, value):
+        if not value:
+            return []
+        return [r.strip() for r in value.split(',')]
```
sloria: The emails should probably be lowercased as well.

admin/meetings/urls.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import absolute_import
+
+from django.conf.urls import url
+
+from admin.meetings import views
+
+urlpatterns = [
+    url(r'^$', views.MeetingListView.as_view(), name='list'),
+    url(r'^create/$', views.MeetingCreateFormView.as_view(), name='create'),
```
sloria: Prefer RESTful routing. You shouldn't need a `/meetings/create/` endpoint; just `POST` to `/meetings/`.

sloria:
Pass finished.

admin/meetings/forms.py
Ghalko: Unfortunately the first assumption is incorrect. `https://osf.io/view/NACIII/` uses an uppercase endpoint and it is stored as such.

The request, though, is completely doable.

admin/meetings/views.py
Ghalko: The fields are in a dictionary. I am basically bringing them to the top so that they can be form fields, but can be distinguished easily as going back into the dictionary. Another way would be to have a map between the names and their membership in the dictionary. Preference @sloria?

website/conferences/model.py
Ghalko: Done, possibly.

admin/meetings/forms.py
Ghalko: Done.

admin/meetings/forms.py
Ghalko: Done.

admin/meetings/urls.py
Ghalko: This suggestion is a confluence of three requests: `GET: list of meetings`, `GET: blank meeting form`, `POST: new meeting`. Should I change the list to be on another URL?

website/conferences/model.py
Ghalko: Had to move the `DEFAULT` outside of the class definition for that tow work. Corrected associated instances of variable use.

admin/meetings/urls.py
sloria: No. The routes should be

```
GET /meetings/ -> get list of meetings
POST /meetings/ -> create a new meeting
GET /meetings/<endpoint> -> get detail for meeting
PATCH /meetings/<endpoint> -> update meeting
```

admin/meetings/views.py
sloria: Ok, I understand now. This is ok, but seems brittle and non-obvious to read. How about checking if the keys are in `DEFAULT_FIELD_NAMES`, like so:

```python
field_names = { }
form_data = deepcopy(form.cleaned_data)
for key, val in form.cleaned_data.iteritems():
    if key in DEFAULT_FIELD_NAMES:
        field_names[key] = form_data.pop(key)
#... Conference(field_names=field_names, **form_data)

```

sloria:
Pass finished.

admin/meetings/views.py
Ghalko: Done. Fixed function that this was done in.

admin/meetings/urls.py
Ghalko: How do you get a blank form @sloria?

Ghalko: With this not using DRF I think we should leave it the way it is, since there are no ListCreateView classes within Django. @sloria

Ghalko:
Unless we want to switch to the Django Rest Framework I think this is gtg. @sloria

admin/meetings/urls.py
sloria: Yep, this is fine for now.

admin/meetings/views.py
```diff
@@ -0,0 +1,128 @@
...
+    def get_success_url(self):
+        return reverse('meetings:detail',
+                       kwargs={'endpoint': self.kwargs.get('endpoint')})
+
+
+def get_custom_fields(data):
+    """Return two dicts, one of field_names and the other regular fields."""
+    data_copy = deepcopy(data)
+    field_names = {}
+    for key, value in data.iteritems():
```
sloria: Much nicer 👍

admin_tests/spam/test_views.py
```diff
@@ -177,7 +179,8 @@ def setUp(self):
         self.view = setup_form_view(self.view, self.request, self.form,
                                     spam_id=self.comment._id)
 
-    def test_get_context_data(self):
+    @mock.patch('admin.spam.views.render')
```
sloria: Why is this mock necessary?

sloria:
Just one question here: https://github.com/CenterForOpenScience/osf.io/pull/5515#discussion_r64082757, but it's not a blocker. 🐙

admin_tests/spam/test_views.py
Ghalko: It prevents it from calling the render which goes out of its way. I believe in the desk PR #5056 I remove this, this was to fix an issue with the tests.

PR #5520, opened 2016-04-22
OSF-5239 Two Factor Auth bug

adlius:
## Purpose

Two factor authentication code still exists after disabling and enabling. Fixed

## Changes

In "twoFactorUserConfig.js", added a line so that it resets the field after disabling.

## Side effects

None. At least not that I may think of.

## Ticket

OSF-5239
https://openscience.atlassian.net/browse/OSF-5239

### DIFF
--- a/website/addons/twofactor/static/twoFactorUserConfig.js
+++ b/website/addons/twofactor/static/twoFactorUserConfig.js
@@ -107,6 +107,7 @@ ViewModel.prototype.disableTwofactorConfirm = function() {
             self.isEnabled(false);
             self.isConfirmed(false);
             $(self.qrCodeSelector).html('');
+            self.tfaCode('');
         })
         .fail(function(xhr, status, error) {
             Raven.captureMessage('Failed to disable two-factor.', {


whit537:
Quick fix. I'd like to see a test but our js tests are weak. I guess we'll let QA handle it.

whit537:
Okay, okay. I felt guilty about not functional-testing this one. However, I don't have a smart phone, so I can't use Google Authenticator or Authy or whatever ... sooooo I hacked a fun little script to solve for acceptable 2fa verification codes, given the secret key:

```python
#!/usr/bin/env python2
from __future__ import absolute_import, division, print_function, unicode_literals

import sys
from base64 import b32decode
from binascii import hexlify

from oath import accept_totp


class GloriousHack(object):
    def __str__(self):
        class BelieveIt(str):
            def __eq__(self, other):
                print(other)
        return BelieveIt()

accept_totp(hexlify(b32decode(sys.argv[1])), GloriousHack())
```

Per [discussion](https://www.flowdock.com/app/cos/osf-dev/threads/nXtbEzMi6rQf84xAxy75I0nP6TP) in Flowdock, this script works for both initial configuration and subsequent login attempts (just be sure to hang on to the secret key!).

```
osf.io| $ python ~/Desktop/solve-totp.py SZS46AHELO5PV6IT
802608
819414
585652
osf.io| $
```

whit537:
Alright, fix confirmed. :-)

whit537:
👍

PR #5522, opened 2016-04-22
Feature/add spinner to component logs

billyhunt:
## Purpose

There is no spinner on the recent activity under components in the project page.  We NEED MORE SPINNERS!  (And it bothers users that activity is blank when you open it)

Before
-----
no-spinner

## Changes

Added a spinner that becomes *invisible* after the data loads.

## Ticket

https://openscience.atlassian.net/browse/OSF-4380?jql=issuetype%20%3D%20Improvement%20AND%20status%20%3D%20Open%20AND%20text%20~%20loading


billyhunt:
Bad PR

PR #5523, opened 2016-04-22
[RFR] Add Spinner to Component Details on Project Page [OSF-4537]

billyhunt:
## Purpose

There is no spinner on the recent activity under components in the project page.  We NEED MORE SPINNERS!  (And it bothers users that the details are missing until it loads)

Before
-----
no-spinner

After
--

apr-28-2016 10-46-24

## Changes

Added a spinner that becomes *invisible* after the data loads.

## Ticket

https://openscience.atlassian.net/browse/OSF-4537


brianjgeiger:
Do you have an 'After' gif you could post?

brianjgeiger:
Did a quick functional test. Two items:
1) The spinner kept spinning even after the log items had loaded. Well, log item, as there was only one. There may be a timing issue related to that.

forever_spin

2) Rumor has it that the OSF spinners are out and we should be using something different, based on @caneruguz's recent lunch presentation which may have happened while you were out of the state. I'm not sure if the style guide has been updated to reflect this info, but we might need to change loading indicators to something more subtle and able to be replicated on the page twelve times without being weird.

brianjgeiger:
It _looks_ like everything should be okay with regards to not doing things out of order vis-a-vis item 1 above, but apparently something is happening there.

billyhunt:
Looks like I was only dismissing the loader if the logs were successfully completed.  That is fixed now.  I will check with caner about the current loader style and with casey if this effects the need for this PR.

billyhunt:
I checked with @caseyrollins, seems like the loader will be able to be used with her code as well when its up and running.  Also, I used the new loader.  This PR is good for another pass @brianjgeiger

mfraezz:
LGTM. This should perhaps be resubmitted against `develop`, but I'll leave that up to Steve. Dropping this in RTM 🌾

website/static/js/project.js
```diff
@@ -13,6 +13,7 @@ var LogFeed = require('js/logFeed.js');
 
 var ctx = window.contextVars;
 var NodeActions = {}; // Namespace for NodeActions
+require('css/style.css');
```
sloria: This shouldn't be necessary; IIRC, style.css is loaded with a `link` tag on every page.

website/static/css/style.css
```diff
@@ -666,4 +666,8 @@ button.close {
 
 .comma-separated:last-of-type:after {
     content: "";
-}
\ No newline at end of file
+}
+
+#render-node .ball-pulse-sync>div, #render-node .ball-pulse>div, #render-node .ball-scale>div, #render-node .ball-beat-sync>div, #render-node .ball-beat>div, #render-node .ball-beat>div {
```
sloria: Why is this necessary?

website/static/js/project.js
```diff
@@ -217,12 +218,18 @@ NodeActions.openCloseNode = function(nodeId) {
     var $logs = $('#logs-' + nodeId);
     if (!$logs.hasClass('active')) {
         if (!$logs.hasClass('served')) {
+            $('.ball-scale').show();
```
sloria: This will affect other `.ball-scale` elements on the page. Better to use an ID.

sloria: Also, as per COSDev, cache jQuery objects to variables.

```
var $loader = $(....);
```

sloria:
Pass finished.

website/static/css/style.css
billyhunt: The new loader is only in my-projects.css.  I would think we would want to be able to reference it everywhere.  Am I missing something @sloria?

sloria: Yes, the ball-scale loader should be usable everywhere; I am referring more to this selector--it's seems overly specific, due mostly to the fact that it's using a CSS ID.

I see now that this is a copy-and-paste of https://github.com/CenterForOpenScience/osf.io/blob/45caed51a62ec0cebafa36202b78f34796230995/website/static/css/my-projects.css#L329-L331 . Instead of relying on IDs, can you make a new class name that describes its usage (something like `.ball-scale-dark`, `.ball-scale-light`...)

website/static/css/pages/project-page.css
```diff
@@ -142,3 +142,7 @@
 span#nodeDescriptionEditable{
     white-space:pre-line;
 }
+
+#render-node  .ball-scale>div {
+    background-color: #BED1E2;
```
sloria: Is there a way to DRY this up with https://github.com/CenterForOpenScience/osf.io/pull/5523/files/9b4687efef370dc11e0bbe678c0fb3fa0dab6231#r62506263 ?

sloria:
Looking good. Once https://github.com/CenterForOpenScience/osf.io/pull/5523/files#r62534463 is addressed, this is good to merge.

sloria:
Merged and cherry-picked into release/0.69.0

PR #5524, opened 2016-04-22
Release/fix fork and registration wikis [OSF-6095]

samanehsan:
## Purpose
When a node is forked or registered, the `wiki_pages_current` and `wiki_pages_versions` lists are copied over, but not the `NodeWikiPage` objects themselves. As a result, a fork or registration can include a wiki version where its `node` field is the original project. Because the wiki node id does not match the fork/registration id, the API throws an error when a user tries to comment on one of those wiki pages.

## Changes
When a project is forked or registered, create a copy of each wiki page in `wiki_pages_versions` and change the node to the fork or registration instead.

## Side effects
More NodeWikiPages!

Requires migration:

```
python -m scripts.clone_wiki_pages
```

## Ticket
https://openscience.atlassian.net/browse/OSF-6095

### DIFF
--- a/scripts/clone_wiki_pages.py
+++ b/scripts/clone_wiki_pages.py
@@ -17,10 +17,10 @@ logger = logging.getLogger(__name__)
 
 def main():
     nodes = Node.find()
-    clone_wiki_pages(nodes)
+    update_wiki_pages(nodes)
 
 
-def clone_wiki_pages(nodes):
+def update_wiki_pages(nodes):
     for i, node in enumerate(nodes):
         if node.wiki_pages_versions:
             cloned_wiki_pages = {}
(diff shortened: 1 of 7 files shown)


scripts/clone_wiki_pages.py
```diff
@@ -0,0 +1,62 @@
...
+def clone_wiki_pages(nodes):
+    for i, node in enumerate(nodes):
+        if node.wiki_pages_versions:
+            cloned_wiki_pages = {}
+            for key in node.wiki_pages_versions:
+                cloned_wiki_pages[key] = []
+                for wiki_id in node.wiki_pages_versions[key]:
+                    node_wiki = NodeWikiPage.load(wiki_id)
+                    if node_wiki.node._id != node._id:
+                        clone = node_wiki.clone()
```
sloria: As mentioned on FD: can you refactor this into a method of NodeWikiPage to DRY up the code?

scripts/clone_wiki_pages.py
```diff
@@ -0,0 +1,62 @@
...
+                        clone.node = node
+                        clone.user = node_wiki.user
```
sloria: Is this line necessary?

samanehsan: Yes, `user` gets set to `None` when the `NodeWikiPage` is cloned so this line fixes that.

sloria:
Fixed up the migration in https://github.com/CenterForOpenScience/osf.io/commit/c91d2649a9838e6beec022dab15710f4bdfd65b6 and merged this into release/0.68.0. 👍 🐙

PR #5526, opened 2016-04-22
Populate New and Noteworthy Round Two [OSF-6049] 

pattisdr:
Jira https://openscience.atlassian.net/browse/OSF-6049

## Purpose

Script to populate new and noteworthy needed some work.  It ran one time and then we turned it off until we could come up with a better way to populate the list.  I've been manually populating it for the last few weeks and came up with a list of items that I thought would help better pick the projects we should highlight.

The original script just pulled recently created nodes and ordered by highest # log actions.

## Changes

This is what script is pulling:
- Highest number of unique log actions - people that are using a lot of features in the OSF
- Projects from the last month
- Top level projects (not components! not registrations!)
- Projects selected should not have common contributors
- Projects selected should not have "test", "photo", "workshop", or "data" in the title.

### DIFF
--- a/scripts/populate_new_and_noteworthy_projects.py
+++ b/scripts/populate_new_and_noteworthy_projects.py
@@ -18,9 +18,9 @@ from website.settings import POPULAR_LINKS_NODE, NEW_AND_NOTEWORTHY_LINKS_NODE, 
 
 logger = logging.getLogger(__name__)
 
-
 def popular_activity_json():
     """ Return popular_public_projects node_ids """
+
     activity_json = activity()
     popular = activity_json['popular_public_projects']
     popular_ids = {'popular_node_ids': []}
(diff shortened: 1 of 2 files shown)


whit537:
Code looks fine. Could use some tests, but this is a low-risk area. I'm good to go here. 👍

scripts/populate_new_and_noteworthy_projects.py
```diff
@@ -18,23 +18,59 @@
...
+        if set(added_node['contributors']).intersection(node['contributors']) != set():
+            return False
+    return True
+
+def acceptable_title(node):
+    """ Omit projects that have certain words in the title """
+
+    omit_titles = ['test', 'photo', 'workshop', 'data']
+    if any(word in str(node['title']).lower() for word in omit_titles):
+        return False
```
sloria: Minor: This can be simpified to 

```python
return not any(word in str(...))
```

sloria: Definitely not a blocker though. Good to merge!

PR #5528, opened 2016-04-25
[rfr][feature] Log out users merging or adding email

billyhunt:
Associated with 

And a based on https://github.com/CenterForOpenScience/osf.io/pull/5000

### Purpose

Existing users who add an email or merge an account have unexpected behaviors, depending on whom is logged in.  

For example, when adding an email, whomever clicks the link sent when adding an email is automatically logged in which is a security risk.  

Merging is also less then optimal, depending on whom is logged in.  If UserA is merged into UserB.

When UserA is logged in, the merge will succeed.
When UserB is logged in, the merge will fail with no error message.  

### Changes:

* Whomever is logged in will be logged out whenever an email is added or a user is merged.  The user will be directed to a login screen and told which username they should log in as.
* After they log in, they will be given the choice to add the email or merge, with the text of the choice changing depending on whether they are merging or adding an email.
*  If they choose Yes, then the merge or email add occurs.
*  If they choose no, they are removed from the user profile as a user to add.


framework/auth/core.py
```diff
@@ -843,6 +848,22 @@ def _get_unconfirmed_email_for_token(self, token):
 
         return verification['email']
 
+    def confirm_token(self, token):
```
sloria: This is a bit of a misnomer: I would expect a method called `confirm_token` to return a confirm token, not validate it. How about `validate_token`?

framework/auth/core.py
```diff
@@ -843,6 +848,22 @@ def _get_unconfirmed_email_for_token(self, token):
...
+        """Return whether or not a confirmation token is valid for this user.
+        :rtype: bool
+        """
```
sloria: Because this method returns a boolean, there's no way to differentiate between (a) token doesn't exist and (b) token is expired.

Instead of returning `False` what if you return different custom exceptions?

framework/auth/views.py
```diff
@@ -150,7 +152,7 @@ def auth_login(auth, **kwargs):
 
     status_message = request.args.get('status', '')
     if status_message == 'expired':
-        status.push_status_message('The private link you used is expired.')
+        status.push_status_message('The private link you used is expired.  Please Resend email.')
```
sloria: Lowercase "resend". Also, could the message guide the user a little more? Like "Please click the *Resend email* button" or something similar?

framework/auth/core.py
sloria: This method could use a unit test in `test_models.py`.

website/routes.py
```diff
@@ -337,6 +337,17 @@ def make_url_map(app):
             json_renderer,
         )
     ], prefix='/api/v1')
+
+    process_rules(app, [
+        Rule('/dashboard/confirmed_emails/', 'get', auth_views.confirm_user_get, json_renderer),
```
sloria: Why are these nested under `/dashboard/`? Seems unnecessary.

website/static/css/pages/login-page.css
```diff
@@ -38,3 +38,35 @@
...
     border-color: #BBB;
 }
+.existing-user-signin {
+    /*height: 320px;*/
+    background-color: #EEE;
+    border-radius: 4px;
+    margin-top: 75px;
+}
+
+.logo-no-spin {
```
sloria: This appears unused.

website/static/css/pages/login-page.css
```diff
@@ -38,3 +38,35 @@
...
+.login-forgot-password {
+    padding-left: 10px;
+}
+
+.login-title {
+    font-weight: bold;
+    margin-top: 10px;
+}
+
+.login-logo {
```
sloria: This also appears unused.

website/static/css/pages/login-page.css
```diff
@@ -38,3 +38,35 @@
...
+
+.logo-no-spin {
+    background: url(../img/logo_spin.png);
+}
+
+.login-forgot-password {
+    padding-left: 10px;
+}
+
+.login-title {
```
sloria: This is also unused; please check if others are unused as well.

website/static/css/pages/login-page.css
```diff
@@ -38,3 +38,35 @@
...
+  display: inline-block;
+  vertical-align: middle;
+  background-image: url(/static/img/cos-logo-150-px.png);
+  background-size: 100%;
+  width: 150px;
+  height: 150px;
+}
+
+.remember-me-checkbox {
+    margin-left: -20px !important;
```
sloria: These hardcoded margins are suspicious; usually they aren't necessary, and we should just use Bootstrap's grid system to place things where they belong. Can you look into this?

website/static/js/osfHelpers.js
```diff
@@ -280,15 +280,28 @@ var isEmail = function(value) {
   * If `str` is falsy, return {}.
   * Modified from getQueryParameters plugin by Nicholas Ortenzio (MIT Licensed).
   */
-var urlParams = function(str) {
+var urlParams = function(str, isEmail) {
```
sloria: `isEmail` appears to be unused.

website/static/js/osfHelpers.js
```diff
@@ -280,15 +280,28 @@ var isEmail = function(value) {
...
     var stringToParse = str || document.location.search;
     if (!stringToParse) {
         return {};
     }
     return (stringToParse).replace(/(^\?)/,'').split('&')
-        .map(function(n){return n = n.split('='),this[n[0]] = decodeURIComponent(n[1]).replace(/\+/g, ' '),this;}.bind({}))[0];
+        .map(function(n){return n = n.split('='),this[n[0]] = decodeURIComponent(n[1]),this;}.bind({}))[0];
```
sloria: Why was this changed?

website/static/js/osfHelpers.js
```diff
@@ -280,15 +280,28 @@ var isEmail = function(value) {
...
 };
 
+/**
+  * Get email address as an object.
+  * If `str` is falsy, return {}.
+  * Modified from getQueryParameters plugin by Nicholas Ortenzio (MIT Licensed).
+  */
+var urlEmail = function(str) {
```
sloria: This shouldn't be necessary. You should URI-encode the email addresses in the query parameters:

```js
encodeURIComponent('ßteve@test.com')  // "%C3%9Fteve%40test.com"
```

Then when you read the query param, decode:

```js
decodeURIComponent(urlParams().email)   // 'ßteve@test.com'
```

website/static/js/pages/home-page.js
```diff
@@ -6,13 +6,116 @@
...
+var bootbox = require('bootbox');
+
+var $osf = require('js/osfHelpers');
+require('loaders.css/loaders.min.css');
+
+var confirmedEmailURL = window.contextVars.confirmedEmailURL;
+var removeConfirmedEmailURL = window.contextVars.removeConfirmedEmailURL;
+
+//choose which emails/users to add to an account and which to deny
+function confirm_emails(emails) {
```
sloria: As per COSDev, use `lowerCamelCase` for function names.

website/static/js/pages/home-page.js
```diff
@@ -6,13 +6,116 @@
...
+        }
+        else {
+            title = 'Add email';
+            requestMessage = 'Would you like to add \<b>' + email.address + '\</b> to your account?';
+            confirmMessage = '\<b>' + email.address + '\</b> has been added into your account.';
+            nopeMessage = 'You have chosen not to add \<b>' + email.address + '\</b> to your account.' +
+             'If you change your mind, visit the \<a href="/settings/account/">user settings page</a>.';
+        }
+
+        var failSuccessMessage = 'There are a problem adding \<b>' + email.address +
```
sloria: Minor: this is a confusing variable name; how about `confirmFailMessage` (and `cancelFailMessage` for consistency)?

framework/auth/views.py
```diff
@@ -185,6 +188,27 @@ def auth_logout(redirect_url=None):
     return resp
 
 
+def auth_email_logout(token, user):
+    """When a user is adding an email or merging an account, add the email to the user and log them out.
+    """
+    redirect_url = web_url_for('auth_login') + '?existing_user={}'.format(escape_html(user.email))
```
sloria: As suggested in a previous comment, you should URI-encode the email (not escape it). You can use `urllib.quote_plus` for this.

framework/auth/views.py
```diff
@@ -251,6 +278,91 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+                except NoResultsFound:
+                    user_merge = False
+
+                verified_emails.append({'address': user.email_verifications[token]['email'],
+                                        'token': token,
+                                        'confirmed': user.email_verifications[token]['confirmed'],
+                                        'user_merge': user_merge.email if user_merge else False})
+            else:
+                user.email_verifications[token]['confirmed'] = False
+    user.save()
```
sloria: This is suspicious; `GET` requests should not be doing any writes to the database.

website/static/js/pages/home-page.js
```diff
@@ -6,13 +6,116 @@
...
+            onEscape: function() {},
+            backdrop: true,
+            closeButton: true,
+            buttons: {
+                confirm: {
+                    label: 'Add email',
+                    className: 'btn-success',
+                    callback: function () {
+                        $osf.putJSON(
+                            confirmedEmailURL,
```
sloria: I am unsure how this is working. Doesn't that endpoint only support `GET` requests?

framework/auth/core.py
sloria: Actually, there seems to already be a method that implements this behavior, `get_confirmation_token`. It will raise an `ExpiredTokenError` if the token is expired, and a KeyError if the token doesn't exist. Can you use that instead?

sloria:
Finishing my initial pass here. I haven't gone through the entire flow yet and I can't say I understand how the pieces fit together quite yet. I'll take a close look at this tomorrow. In the meantime, feel free to address the comments above.

website/routes.py
```diff
@@ -337,6 +337,17 @@ def make_url_map(app):
...
+        Rule('/dashboard/confirmed_emails/', 'put', auth_views.add_confirmed_email, json_renderer)
```
sloria: Rather than having 2 duplicate rules, you can pass a list of supported methods, e.g. `['get', 'put']`.

website/static/js/pages/home-page.js
sloria: Ah, I just saw that PUT is supported by a duplicate `Rule`. See https://github.com/CenterForOpenScience/osf.io/pull/5528/files#r61086751

website/routes.py
```diff
@@ -337,6 +337,17 @@ def make_url_map(app):
...
+
+    ], prefix='/api/v1')
+
+    process_rules(app, [
+        Rule('/dashboard/remove_confirmed_emails/', 'put', auth_views.confirm_email_remove, json_renderer)
```
sloria: Use HTTP verbs rather than arbitrary naming conventions in URLs (see http://stackoverflow.com/a/2022938/1157536, points 1 and 2).

You should only need one URL to achieve the functionality you need:

```
GET /api/v1/emails/  -> list confirmed emails
POST /api/v1/emails/ -> add a new email
DELETE /api/v2/emails/ with JSON payload {"emails": "mick%40stones.com"} -> remove "mick@stones.com" from confirmed emails
```

website/routes.py
sloria: `PUT` is incorrect here. `POST` is used for creating objects. See my comment below for the proposed routes.

sloria:
Discussed in person: to make sure the user sees the merge/add email confirmation regardless of the page they are redirected to after login, the code to display the modal will be moved to `base-page.js` and will use `contextVars.currentUser` in base.mako to save the emails to be merged/added.

framework/auth/views.py
```diff
@@ -191,7 +192,7 @@ def auth_logout(redirect_url=None):
 def auth_email_logout(token, user):
     """When a user is adding an email or merging an account, add the email to the user and log them out.
     """
-    redirect_url = web_url_for('auth_login') + '?existing_user={}'.format(escape_html(user.email))
+    redirect_url = web_url_for('auth_login') + '?existing_user={}'.format(urllib.quote_plus(user.email))
```
sloria: Is this query param still necessary, since "existing user" is now passed as a context var?

website/routes.py
```diff
@@ -337,6 +339,14 @@ def make_url_map(app):
             json_renderer,
         )
     ], prefix='/api/v1')
+
+    process_rules(app, [
+        Rule('/dashboard/confirmed_emails/', 'get', auth_views.verified_email_get, json_renderer),
```
sloria: Is this GET endpoint used? Can it be removed?

sloria: Not sure where my previous comment went, but why nest under `/dashboard/`?

framework/auth/views.py
```diff
@@ -251,6 +279,88 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+    }, 200
+
+
+@collect_auth
+def verified_email_add(auth=None, **kwargs):
+    """Called at login if user confirms their merge or email add.
+    methods: PUT
+    """
+    user = auth.user
+    confirmed_email = request.json.get('address')
```
sloria: Why is it necessary for the client to pass the email address? It doesn't appear to be used except for a check on ~L345. It looks like the `token` will suffice to confirm the email.

framework/auth/views.py
```diff
@@ -251,6 +279,88 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+                                        'token': token,
+                                        'confirmed': user.email_verifications[token]['confirmed'],
+                                        'user_merge': user_merge.email if user_merge else False})
+    return verified_emails
+
+
+@collect_auth
+def verified_email_remove(auth=None, **kwargs):
+    """Called at login if user cancels their merge or email add.
+    methods: PUT
```
sloria: This should accept `DELETE`, not `PUT`.

website/static/js/pages/base-page.js
```diff
@@ -92,6 +93,101 @@ $(document).on('click', '.panel-heading', function(){
     }
 });
 
+//choose which emails/users to add to an account and which to deny
+function confirmEmails(emailsToAdd) {
+    if (emailsToAdd.length > 0) {
+        var confirmedEmailURL = window.contextVars.confirmedEmailURL;
+        var email = emailsToAdd.splice(0, 1)[0];
```
sloria: It's generally a bad idea for functions to mutate the arguments passed to them. Why not use `Array.prototype.slice` (which doesn't mutate arrays) when you make the recursive calls instead?

framework/auth/core.py
```diff
@@ -843,6 +848,22 @@ def _get_unconfirmed_email_for_token(self, token):
...
         return verification['email']
 
+    def verify_token(self, token):
+        """Return whether or not a confirmation token is valid for this user.
+        :rtype: bool
+        """
+        try:
+            verification = self.email_verifications[token]
+        except KeyError:
+            return False
```
sloria: As suggested in a previous comment: because this method return a boolean, there is no way to distinguish between (1) if `token` is invalid and (2) `token` is expired. 

This method should probably mirror the behavior of `get_confirmation_token`: raise an `ExpiredTokenError` for expired tokens and raise a `KeyError` if the token is not a valid token for the user.

framework/auth/views.py
```diff
@@ -196,9 +221,12 @@ def confirm_email_get(token, auth=None, **kwargs):
     user = User.load(kwargs['uid'])
     is_merge = 'confirm_merge' in request.args
     is_initial_confirmation = not user.date_confirmed
-
+    existing_user = request.args.get('existing_user', None)
```
sloria: As suggested in a previous comment, it doesn't appear necessary to pass `existing_user` anymore. If you need a query param to logout the user, then you could just accept a bool, `?logout=1`

framework/auth/views.py
```diff
@@ -251,6 +279,88 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+    return verified_emails
+
+
+@collect_auth
+def verified_email_remove(auth=None, **kwargs):
+    """Called at login if user cancels their merge or email add.
+    methods: PUT
+    """
+    user = auth.user
+    confirmed_email = request.json.get('address')
```
sloria: A 400 error should be returned if this isn't provided.

Also, `request.json` is [deprecated](http://flask.pocoo.org/docs/0.10/api/#flask.Request.json). Use `.get_json()` instead.

website/templates/public/login.mako
```diff
@@ -80,176 +86,198 @@
...
+                        >
                 </div>
             </div>
-            <div class="form-group">
-                <div class="col-sm-offset-3 col-sm-9">
-                    <div class="checkbox">
-                    <label><input type="checkbox"> Remember me</label>
+            %if existing_user:
+                <div class="col-sm-12 m-t-sm">
+                <div class="form-group col-sm-7 remember-me-checkbox">
```
sloria: The grid CSS here looks off. I don't think `col-*` classes are meant to be used on the same element as `form-group`. Also, each row needs to have 12 columns. Can you look into fixing this up?

Using the grid properly will hopefully allow you to remove one-off classes like `remember-me-checkbox`.

website/templates/public/login.mako
```diff
@@ -80,176 +86,198 @@
...
-        </form>
-    </div>
-        %if sign_up:
-            <div id="signUpScope" class="col-sm-5 toggle-box toggle-box-right toggle-box-active p-h-lg" style="height: auto;">
-        %else:
-            <div id="signUpScope" class="col-sm-5 toggle-box toggle-box-right toggle-box-muted p-h-lg" style="height: auto;">
-        %endif
-        <form data-bind="submit: submit" class="form-horizontal">
-            <h3 class="m-b-lg"> Create a free account </h3>
+                <div class="col-sm-8 login-forgot-password">
```
sloria: CSS looks off here too; see https://github.com/CenterForOpenScience/osf.io/pull/5528/files#r61775823

website/templates/public/login.mako
```diff
@@ -80,176 +86,198 @@
...
+</div>
+</div>
 <%def name="javascript_bottom()">
     ${parent.javascript_bottom()}
     <script type="text/javascript">
         window.contextVars = $.extend(true, {}, window.contextVars, {
             'campaign': ${campaign or '' | sjson, n},
-            'institution_redirect': ${institution_redirect or '' | sjson, n}
+            'institution_redirect': ${institution_redirect or '' | sjson, n},
+            'existing_user': ${existing_user or '' | sjson, n}
```
sloria: Is this used?

website/static/js/signIn.js
```diff
@@ -15,10 +15,19 @@ var SignInViewModel = oop.extend(formViewModel.FormViewModel, {
     constructor: function () {
         var self = this;
         self.super.constructor.call(self);
-        self.username = ko.observable('').extend({
-            required: true,
-            email: true
-        });
+        var urlEmail = $osf.urlEmail();
```
sloria: I see now; the `existing_user` param is used to prepopulate the input field. That said, `urlEmail` seems unnecessary. You should be able to read the parameter the regular way, as in

```js
var existingUserEmail = decodeURIComponent($osf.urlParams().existing_user);

File: website/static/js/osfHelpers.js [hunk 31]
```diff
@@ -289,6 +289,19 @@ var urlParams = function(str) {
         .map(function(n){return n = n.split('='),this[n[0]] = decodeURIComponent(n[1]).replace(/\+/g, ' '),this;}.bind({}))[0];
 };
 
+/**
+  * Get query string arguments as an object.
+  * If `str` is falsy, return {}.
+  * Modified from getQueryParameters plugin by Nicholas Ortenzio (MIT Licensed).
+  */
+var urlEmail = function(str) {
```
sloria: This probably isn't necessary, and it duplicates a lot from `urlParams`. See https://github.com/CenterForOpenScience/osf.io/pull/5528/files#r61776758

tests/test_views.py
```diff
@@ -3370,6 +3371,89 @@ def test_resend_confirmation(self, send_mail):
...
+        assert_equal(res.json_body['status'], 'success')
+        assert_equal(self.user.emails[1], 'test@example.com')
+
+    def test_add_email_merge(self):
+        email = "copy@cat.com"
+        dupe = UserFactory(
+            username=email,
+            emails=[email]
+        )
+        dupe.set_password("copycat")
```
sloria: Is it necessary to set the password in this test? This is minor, but bcrypt-ing passwords can slow down tests significantly.

sloria:
Is there a test for adding an email with non-ascii characters?

website/templates/public/login.mako
```diff
@@ -42,34 +46,36 @@
...
+                        <p>For non-institutional sign in, click <a href="/login/">here</a>.</p>
                     %endif
                 </div>
             </div>
         </div>
     </div>
+%endif
+%if campaign != "institution" or not enable_institutions:
+    %if existing_user:
+        <div class="col-sm-6 col-sm-offset-3 existing-user-signin p-b-md m-b-m">
```
sloria: Check the grid here too; the other `elif` and `else` branches have an element that takes 6 columns; this element takes 9, which is suspicious.

website/static/js/pages/home-page.js
```diff
@@ -6,13 +6,27 @@
...
+
 
 var QuickSearchProject = require('js/home-page/quickProjectSearchPlugin');
 var NewAndNoteworthy = require('js/home-page/newAndNoteworthyPlugin');
 var MeetingsAndConferences = require('js/home-page/meetingsAndConferencesPlugin');
 
 var columnSizeClass = '.col-md-10 col-md-offset-1 col-lg-8 col-lg-offset-2';
 
+
+var Raven = require('raven-js');
```
sloria: All of these imports and variables are unused.

website/static/js/pages/home-page.js
```diff
@@ -6,13 +6,27 @@
 
 var $ = require('jquery');
 var m = require('mithril');
+var bootbox = require('bootbox');
```
sloria: This is also unused.

website/routes.py
```diff
@@ -65,6 +66,7 @@ def get_globals():
         'user_timezone': user.timezone if user and user.timezone else '',
         'user_url': user.url if user else '',
         'user_gravatar': profile_views.current_user_gravatar(size=25)['gravatar_url'] if user else '',
+        'user_email_verifications': auth_views.verified_email_get(uid=user._id) if user else [],
```
sloria: You shouldn't need to call a view function; just use a normal function or a property of `User`. As suggested in a previous comment, the `verified_email_get` endpoint doesn't appear to be used anywhere. Just turn it into a property, and you'll be golden.

sloria:
Pass finished.

website/routes.py
billyhunt: @sloria, I use currently verified_email_get for testing.  Is there a better way to get the list of users to add in tests?

sloria: As suggested in https://github.com/CenterForOpenScience/osf.io/pull/5528/files/2cb0c6854f964a4381be36744e6744300b5ea7d6#r61778717 , just make it a helper function or a property. There's no need for a view.

website/templates/public/login.mako
billyhunt: This is because the login page is centered in this view.

Before
---
![screen shot 2016-05-03 at 11 09 30 am]
(https://cloud.githubusercontent.com/assets/6232068/14997434/b33d6348-114c-11e6-96dc-f08e745d0cf8.png)

After
---
screen shot 2016-05-03 at 4 34 54 pm

billyhunt:
Should be good to go.  When I was reviewing, I saw I had removed the apiOauth2Application dictionary in osfLanguage.js

It seems like an error so I put it back.  Can you double check that that was the right move?

framework/auth/views.py
```diff
@@ -280,40 +290,16 @@ def confirm_email_get(token, auth=None, **kwargs):
...
-
-@collect_auth
 def verified_email_remove(auth=None, **kwargs):
     """Called at login if user cancels their merge or email add.
-    methods: PUT
+    methods: DELETE
     """
     user = auth.user
-    confirmed_email = request.json.get('address')
+    confirmed_email = request.get_json('address')
```
sloria: Is this working? `get_json` doesn't expect to receive a key in its arguments: http://flask.pocoo.org/docs/0.10/api/#flask.Request.get_json

billyhunt: It does work, but there is no need to have the arguments, so I removed it.

framework/auth/core.py
```diff
@@ -843,6 +849,52 @@ def _get_unconfirmed_email_for_token(self, token):
...
 
+    def clean_email_verifications(self):
+        email_verifications = deepcopy(self.email_verifications)
+        for token in self.email_verifications:
+            try:
+                self.verify_token(token)
+            except (KeyError, ExpiredTokenError):
+                email_verifications.pop(token)
+        self.email_verifications = email_verifications
+        self.save()
```
sloria: Avoid implicit saves in methods. The user of your method might want to do other updates before they save:

```python
user.clean_email_verifications()
user.do_something_else()
user.save()
```

framework/auth/core.py
```diff
@@ -843,6 +849,52 @@ def _get_unconfirmed_email_for_token(self, token):
...
+
+    def verified_email_get(self):
```
sloria: This name needs to be updated. As discussed in a previous comment, you could make this a property, e.g. `verfied_emails`

framework/auth/core.py
```diff
@@ -843,6 +849,52 @@ def _get_unconfirmed_email_for_token(self, token):
...
+                except NoResultsFound:
+                    user_merge = False
+
+                verified_emails.append({'address': self.email_verifications[token]['email'],
+                                        'token': token,
+                                        'confirmed': self.email_verifications[token]['confirmed'],
+                                        'user_merge': user_merge.email if user_merge else False})
+        return verified_emails
+
+    def verify_token(self, token):
```
sloria: This seems to do almost the exact same thing as `User._get_unconfirmed_email_for_token`. It'd probably make sense to just rename that method to "unprivatize" it, i.e. remove the leading underscore, and just use that instead of reimplementing it here.

framework/auth/views.py
```diff
@@ -150,7 +154,7 @@ def auth_login(auth, **kwargs):
 
     status_message = request.args.get('status', '')
     if status_message == 'expired':
-        status.push_status_message('The private link you used is expired.')
+        status.push_status_message('The private link you used is expired.  Please <a href="/settings/account/resend email.')
```
sloria: This is not valid HTML.

Also, you should pass the `trust=True` param to make it clear that this message is safe to render as HTML as-is.

framework/auth/views.py
```diff
@@ -185,6 +190,33 @@ def auth_logout(redirect_url=None):
...
+def auth_email_logout(token, user):
+    """When a user is adding an email or merging an account, add the email to the user and log them out.
+    """
+    redirect_url = web_url_for('auth_login') + '?existing_user={}'.format(urllib.quote_plus(user.email))
+    try:
+        user.verify_token(token)
+    except KeyError:
+        raise HTTPError(http.BAD_REQUEST, data={
+            'message_short': "Bad token",
+            'message_long': "There was a key error with the token"
```
sloria: The user will not know what a "key error" means. How about "The provided token is invalid."?

framework/auth/views.py
```diff
@@ -185,6 +190,33 @@ def auth_logout(redirect_url=None):
...
+        })
+    except ExpiredTokenError:
+        status.push_status_message('The private link you used is expired.')
+    try:
+        user_merge = User.find_one(Q('emails', 'eq', user.email_verifications[token]['email'].lower()))
```
sloria: As suggested in a previous comment, you can get rid of `verify_token` in favor of using `get_unconfrimed_email_for_token`. That way you can avoid accessing `user.email_verifications` directly here.

framework/auth/views.py
```diff
@@ -251,6 +287,62 @@ def confirm_email_get(token, auth=None, **kwargs):
     ))
 
 
+@collect_auth
+def verified_email_remove(auth=None):
+    """Called at login if user cancels their merge or email add.
+    methods: DELETE
+    """
+    user = auth.user
+    confirmed_email = request.get_json()
```
sloria: Minor: this seems like a misnomer. Reading through this view it, looks like this is a dict of the form `{'token': <some token>}`, not an email. How about just `json_body`?

framework/auth/views.py
```diff
@@ -251,6 +287,62 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+    email_verifications = deepcopy(user.email_verifications)
+    for token in user.email_verifications:
+        if token == confirmed_email['token']:
```
sloria: What if `token` isn't in the json body? A proper 400 should be returned to the client. You can do this validation before you go into the loop.

```python
json_body = request.get_json()
try:
    given_token = json_body['token']
except KeyError:
    raise HTTPError(...)
for token ...
```

framework/auth/views.py
```diff
@@ -251,6 +287,62 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+    }, 200
+
+
+@collect_auth
+def verified_email_add(auth=None):
+    """Called at login if user confirms their merge or email add.
+    methods: PUT
+    """
+    user = auth.user
+    token = request.json.get('token')
```
sloria: Use `get_json`.

website/static/js/pages/base-page.js
```diff
@@ -92,6 +93,101 @@ $(document).on('click', '.panel-heading', function(){
...
+                confirm: {
+                    label: 'Add email',
+                    className: 'btn-success',
+                    callback: function () {
+                        $osf.putJSON(
+                            confirmedEmailURL,
+                            email
+                        ).done(function () {
+                            $osf.growl('Success', confirmMessage, 'success', 3000);
+                            confirmEmails(emailsToAdd.slice(1));
```
sloria: There's an additional call to `confirmEmails` a few lines down; won't this result in duplicate confirmation steps?

framework/auth/core.py
sloria: Actually, `verified_emails` would also be a misnomer because this returns *un*verified emails, correct? So how about `unconfirmed_emails`? The variable in `get_globals` as well as the JS variable in contextVars should reflect this name change (e.g. `user_unconfirmed_emails` and `currentUser.unconfirmedEmails`).

framework/auth/views.py
```diff
@@ -251,6 +287,62 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+        user.save()
+        return {
+            'status': 'success',
+        }, 200
+
+    # Redirect to CAS and authenticate the user with a verification key.
+    user.verification_key = security.random_string(20)
+    user.save()
+
+    return redirect(cas.get_login_url(
```
sloria: Does this redirect actually do anything? It appears incorrect because this is a JSON endpoint.

website/templates/public/login.mako
sloria: Even when you are centering things, every row must still add up to 12 columns; this ensures the grid responds correctly on mobile devices, etc. You just need to adjust your offsets.

sloria:
Pass finished.

framework/auth/views.py
```diff
@@ -251,6 +288,68 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+    methods: DELETE
+    """
+    user = auth.user
+    if user is None:
+        raise HTTPError(http.NOT_FOUND)
+    json_body = request.get_json()
+    try:
+        given_token = json_body['token']
+    except KeyError:
+        raise HTTPError(http.NOT_FOUND)
```
sloria: NOT_FOUND (404) is inappropriate for this type of error (the client didn't provide a token)--a 400 should be returned instead. Also, it's good practice to send useful error messages to the client so they know how to correct the error. You can pass `message_short` and `message_long` to the `data` param of `HTTPError`.

framework/auth/views.py
```diff
@@ -251,6 +288,68 @@ def confirm_email_get(token, auth=None, **kwargs):
...
 
 
+@collect_auth
+def unconfirmed_email_remove(auth=None):
+    """Called at login if user cancels their merge or email add.
+    methods: DELETE
+    """
+    user = auth.user
+    if user is None:
+        raise HTTPError(http.NOT_FOUND)
```
sloria: 404 isn't appropriate here; a 401 is used for when auth creds aren't passed. Why not use `must_be_logged_in`?

sloria: That said, I just discovered the behavior for `must_be_logged_in` is incorrect for JSON requests. I've filed a ticket (https://openscience.atlassian.net/browse/OSF-6317). I still think it's still ok to use that decorator for now though.

framework/auth/views.py
```diff
@@ -251,6 +288,68 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+    email_verifications = deepcopy(user.email_verifications)
+    for token in user.email_verifications:
+        if token == given_token:
+            email_verifications.pop(token)
+    user.email_verifications = email_verifications
+    user.save()
+    return {
+        'status': 'success',
+        'removed_email': json_body
```
sloria: This looks wrong--won't `json_body` look like `{'token': <token>}`?

framework/auth/views.py
```diff
@@ -251,6 +288,68 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+
+
+@collect_auth
+def unconfirmed_email_add(auth=None):
+    """Called at login if user confirms their merge or email add.
+    methods: PUT
+    """
+    user = auth.user
+    if user is None:
+        raise HTTPError(http.NOT_FOUND)
```
sloria: See previous comment; use `must_be_logged_in`.

framework/auth/views.py
```diff
@@ -251,6 +288,68 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+            'status': 'success',
+        }, 200
+
+    # Redirect to CAS and authenticate the user with a verification key.
+    user.verification_key = security.random_string(20)
+    user.save()
+
+    return {
+        'status': 'success',
+        'removed_email': json_body
```
sloria: See previous comment-- "removed_email" is not a removed email.

framework/auth/views.py
```diff
@@ -251,6 +288,68 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+    methods: PUT
+    """
+    user = auth.user
+    if user is None:
+        raise HTTPError(http.NOT_FOUND)
+    json_body = request.get_json()
+    try:
+        token = json_body['token']
+    except KeyError:
+        raise HTTPError(http.NOT_FOUND)
```
sloria: See https://github.com/CenterForOpenScience/osf.io/pull/5528/files#r63067074

framework/auth/views.py
```diff
@@ -251,6 +288,68 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+            'message_long': e.message_long
+        })
+
+    if token:
+        user.save()
+        return {
+            'status': 'success',
+        }, 200
+
+    # Redirect to CAS and authenticate the user with a verification key.
```
sloria: Where does the redirect happen?

framework/auth/views.py
```diff
@@ -185,6 +191,33 @@ def auth_logout(redirect_url=None):
     return resp
 
 
+def auth_email_logout(token, user):
+    """When a user is adding an email or merging an account, add the email to the user and log them out.
+    """
+    redirect_url = web_url_for('auth_login') + '?existing_user={}'.format(urllib.quote_plus(user.email))
+    try:
+        unconfirmed_email = user.get_unconfirmed_email_for_token(token)
+    except KeyError:
```
sloria: This seems to catching the wrong error; `get_unconfirmed..` raises either a `InvalidTokenError` or a `ExpiredTokenError`.

framework/auth/views.py
```diff
@@ -185,6 +191,33 @@ def auth_logout(redirect_url=None):
...
+        raise HTTPError(http.BAD_REQUEST, data={
+            'message_short': "Bad token",
+            'message_long': "The provided token is invalid."
+        })
+    except ExpiredTokenError:
+        status.push_status_message('The private link you used is expired.')
```
sloria: Nothing is stopping execution of the rest of this function, so `user.email_verifications[token]['confirmed']` still gets set to `True`, which seems incorrect.

framework/auth/views.py
```diff
@@ -185,6 +191,33 @@ def auth_logout(redirect_url=None):
...
+    try:
+        user_merge = User.find_one(Q('emails', 'eq', unconfirmed_email))
+    except NoResultsFound:
+        user_merge = False
+    if user_merge:
+        remove_sessions_for_user(user_merge)
+    user.email_verifications[token]['confirmed'] = True
```
sloria: I'm not sure if this should happen here. What happens if the user removes the `?logout=1` query param from their confirmation link? Does that break the flow? It's probably not a big deal if the user doesn't get logged out, but are there other side effects if the `confirmed` flag doesn't get set to `True`?

website/static/js/accountSettings.js
```diff
@@ -217,13 +217,13 @@ var UserProfileViewModel = oop.extend(ChangeMessageMixin, {
...
-                                ' You will receive a confirmation email at ' + '<em>' + addrText + '</em>.' +
-                                ' Please log out of this account and check your email to confirm this action.',
+                                message: '<b>' + addrText + '</b>' + ' was added to your account.' +
+                                ' You will receive a confirmation email at ' + '<b>' + addrText + '</b>.' +
+                                ' Please check your email to confirm this action.',
                                 buttons: {
                                     ok: {
                                         label: 'Close',
-                                        className: 'btn-default'
+                                        className: 'btn-primary'
```
sloria: Why was this changed? Other "Close" buttons seem to use `btn-default`: https://github.com/CenterForOpenScience/osf.io/blob/4f618ad25b8fbca86213bea604e708450021b1fd/website/static/js/pages/project-settings-page.js#L115-L118

framework/auth/core.py
```diff
@@ -843,6 +849,35 @@ def _get_unconfirmed_email_for_token(self, token):
...
+    def clean_email_verifications(self):
+        email_verifications = deepcopy(self.email_verifications)
+        for token in self.email_verifications:
+            try:
+                self.get_unconfirmed_email_for_token(token)
+            except (KeyError, ExpiredTokenError):
+                email_verifications.pop(token)
+        self.email_verifications = email_verifications
+
+    def unconfirmed_email_get(self):
```
sloria: You've made a view into a method, so make sure to change the name and docstring accordingly (e.g. "methods: GET" no longer applies). This should probably be made a property, since it's a...property...of the user =).  

Whether it's a property or a getter method, this should not do any writes, i.e. clean_email_verifications should not be called here. Users of class will not expect `user.get_unconfirmed_emails()` (getter method) or `user.unconfirmed_emails` (property) to write to the database.

framework/auth/views.py
```diff
@@ -150,7 +154,8 @@ def auth_login(auth, **kwargs):
 
     status_message = request.args.get('status', '')
     if status_message == 'expired':
-        status.push_status_message('The private link you used is expired.')
+        status.push_status_message('The private link you used is expired.  Please <a href="/settings/account/"> +'
```
sloria: Typo here; remove the `+`.

sloria: Can we reduce a step for the user here and resend the link when they click "resend email" rather than directing them to the account settings?

This is nbd, though, you can add this in a followup PR if you want.

sloria:
Pass finished.

framework/auth/views.py
billyhunt: This would take a bit of effort to use a direct link.  Seems like a separate PR would be a good idea.

framework/auth/core.py
billyhunt: @sloira, where would be a good place to put clean_email_verifications?  auth_login?

sloria: I think that makes sense.

framework/auth/views.py
billyhunt: I tested this.  And you log in and get a message saying the user was not added, and revisit the link.  The merge won't happen until you go and click the correct link.

website/static/js/pages/base-page.js
billyhunt: Fixed.

framework/auth/views.py
```diff
@@ -185,6 +193,34 @@ def auth_logout(redirect_url=None):
...
+    try:
+        unconfirmed_email = user.get_unconfirmed_email_for_token(token)
+    except InvalidTokenError:
+        raise HTTPError(http.BAD_REQUEST, data={
+            'message_short': "Bad token",
+            'message_long': "The provided token is invalid."
+        })
+    except ExpiredTokenError:
+        status.push_status_message('The private link you used is expired.')
+        return
```
sloria: This implicitly returns `None`, which seems incorrect, because this means the view will return `None` here: https://github.com/CenterForOpenScience/osf.io/pull/5528/files#diff-9a64b4982d0bd80f7a498c65ddce2023R241 . Probably better to raise an `HTTPError` so that the user sees an error page.

framework/auth/views.py
```diff
@@ -185,6 +193,34 @@ def auth_logout(redirect_url=None):
...
 
+def auth_email_logout(token, user):
+    """When a user is adding an email or merging an account, add the email to the user and log them out.
+    """
+    redirect_url = web_url_for('auth_login') + '?existing_user={}'.format(urllib.quote_plus(user.email))
+    try:
+        unconfirmed_email = user.get_unconfirmed_email_for_token(token)
+    except InvalidTokenError:
+        raise HTTPError(http.BAD_REQUEST, data={
+            'message_short': "Bad token",
```
sloria: Very minor: Use singlequotes for strings, as per COSDev.

framework/auth/views.py
```diff
@@ -137,8 +140,11 @@ def auth_login(auth, **kwargs):
         # Only allow redirects which are relative root or full domain, disallows external redirects.
         if not (next_url[0] == '/' or next_url.startsWith(settings.DOMAIN)):
             raise HTTPError(http.InvalidURL)
-
     if auth.logged_in:
+        # remove expired email verifications
+        if not campaign:
+            auth.user.clean_email_verifications(auth.user)
```
sloria: I am getting a TypeError here because `clean_email_verifications` does not expect arguments.

```
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/flask/app.py", line 1836, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/flask/app.py", line 1820, in wsgi_app
    response = self.make_response(self.handle_exception(e))
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/flask/app.py", line 1403, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/flask/app.py", line 1381, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/sloria/projects/cos-projects/osf/framework/routing/__init__.py", line 118, in wrapped
    data = fn(*args, **kwargs)
  File "/Users/sloria/projects/cos-projects/osf/framework/auth/decorators.py", line 22, in wrapped
    return func(*args, **kwargs)
  File "/Users/sloria/projects/cos-projects/osf/framework/auth/views.py", line 146, in auth_login
    auth.user.clean_email_verifications(auth.user)
TypeError: clean_email_verifications() takes exactly 1 argument (2 given)
```

website/static/js/pages/base-page.js
```diff
@@ -92,6 +93,100 @@ $(document).on('click', '.panel-heading', function(){
...
+                'This action is irreversable.';
+            confirmMessage = '\<b>' + email.address + '\</b> has been merged into your account.';
+            nopeMessage = 'You have chosen to not merge \<b>' + email.address + '\</b>  into your account. ' +
+                'If you change your mind, visit the \<a href="/settings/account/">user settings page</a>.';
+        }
+        else {
+            title = 'Add email';
+            requestMessage = 'Would you like to add \<b>' + email.address + '\</b> to your account?';
+            confirmMessage = '\<b>' + email.address + '\</b> has been added into your account.';
+            nopeMessage = 'You have chosen not to add \<b>' + email.address + '\</b> to your account.' +
```
sloria: Typo: need a space after the period.

framework/auth/views.py
```diff
@@ -251,6 +291,68 @@ def confirm_email_get(token, auth=None, **kwargs):
...
+    user = auth.user
+    json_body = request.get_json()
+    try:
+        given_token = json_body['token']
+    except KeyError:
+        raise HTTPError(http.BAD_REQUEST, data={
+            'message_short': 'Invalid token',
+            'message_long': 'The token provided is invalid'
+        })
+    email_verifications = deepcopy(user.email_verifications)
```
sloria: This is the same as `user.clean_email_verifications()`

framework/auth/views.py
sloria: Ah, I see. I didn't see the conditional here: https://github.com/CenterForOpenScience/osf.io/pull/5528/files#diff-9a64b4982d0bd80f7a498c65ddce2023R243 that would prevent the merge/email add from happening.

sloria:
Added a few drive-by comments while I was looking into https://github.com/CenterForOpenScience/osf.io/pull/5528/files#r63072843 , but stopping here until you give me the go-ahead to do next pass. 👍

framework/auth/views.py
```diff
@@ -201,12 +200,15 @@ def auth_email_logout(token, user):
         unconfirmed_email = user.get_unconfirmed_email_for_token(token)
     except InvalidTokenError:
         raise HTTPError(http.BAD_REQUEST, data={
-            'message_short': "Bad token",
-            'message_long': "The provided token is invalid."
+            'message_short': 'Bad token',
+            'message_long': 'The provided token is invalid.'
         })
     except ExpiredTokenError:
         status.push_status_message('The private link you used is expired.')
```
sloria: Probably shouldn't have a status message *and* and error page.

PR #5294, opened 2016-03-24
[rfr] EmbargoTerminationApproval

samchrisinger:
:revolving_hearts: 
https://openscience.atlassian.net/browse/OSF-5641
:bouquet: :confetti_ball: :bouquet: 

# Purpose

We want a way for users to be able to end embargoed registrations public early. Borrowing from the patterns for Registrations, Embargoes, Retractions, etc, we will notify admins on the registration tree and request that they either approve or disapprove this action. If an admin fails to respond within 48hrs, they have granted implicit approval. If a user is the only project administrator, then the Embargo will be lifted immediately. 

# Changes

- Sanction and its subclasses are moved to their own module (website.project.sanctions)
- add new Sanction subclass EmbargoTerminationApproval
- add NodeLog.EMBARGO_TERMINATED/EMBARGO_TERMINATION_APPROVED/ EMBARGO_TERMINATION_CANCELLED and supporting templates (TODO)
- update nodesPrivacy.js/mako to support making Embargoes public
- move NodePrivacy instantiation into NodeControl, and pass onPrivacyChanged handler into NodePrivacy
- add tests for Node sanction states (is_pending_registration, is_embargoed, etc)

# Side Effects

The logic for restricting what can be made public/private is fairly scattered, so I wouldn't be surprised if some places cause issues (see: https://github.com/samchrisinger/osf.io/commit/0d1821dae05a24c2f5e1c2ea869ca1f837b0d2e8#diff-7a6d21a50624153d15e63a5f643401c7). Hopefully testing will resolve most or all of these.

Additionally this adds a new auto-approve script (like Embargoes and ReigstrationApprovals), that will needed to be added to celery-beat (?).

# Testing Notes

### Creating an Embargoed registration locally:
1. Create a project or navigate to an existing project
2. Create a registration of that project
  1. go to the 'Registrations' tab
  2. click 'New registration', pick 'Open Ended ...'
  3. Press 'Preview for submission'
  4. Press 'Register' and choose 'Enter registration into Embargo'; choose a valid end date
  5. Watch the log output of your `celery_worker`, it should dump an email that reads like:
```
[2016-04-11 13:47:24,410: DEBUG/Worker-4] Sending email...
[2016-04-11 13:47:24,411: DEBUG/Worker-4] To: sam@cos.io
From: openscienceframework-noreply@osf.io
Subject: Registration pending for one of your projects.
Message: Hello sam,

[...]

PR #5529, opened 2016-04-25
Embargo Terminations

samchrisinger:
Replaces: https://github.com/CenterForOpenScience/osf.io/pull/5294


samchrisinger:
@mfraezz- PR is updated with a clean output from MockChecker. If the build passes I think this is ready for another look.

scripts/approve_embargo_terminations.py
```diff
@@ -0,0 +1,68 @@
...
+            ))
+        else:
+            count += 1
+            logger.info("Ending the Embargo ({0}) of Registration ({1}) early. Making the registration and all of its children public now.".format(embargo._id, registration._id))
+            request._on_complete()
+            registration.reload()
+            assert registration.is_embargoed is False
+            assert registration.is_public is True
+    logger.info("Auto-approved {0} of {1} embargo termination requests".format(count, len(pending_embargo_termination_requests)))
+
```
sloria: For this to work with celery beat, you'll need to declare a new celery task using `@celery_app.task(...)`.

scripts/approve_embargo_terminations.py
```diff
@@ -0,0 +1,70 @@
...
+        Q('state', 'eq', models.EmbargoTerminationApproval.UNAPPROVED)
+    )
+
+def main():
+    pending_embargo_termination_requests = get_pending_embargo_termination_requests()
+    count = 0
+    for request in pending_embargo_termination_requests:
+        registration = models.Node.find_one(Q('embargo_termination_approval', 'eq', request))
+        if not registration.is_embargoed:
+            raise RuntimeError("Registration {0} associated with this embargo termination request ({0}) is not embargoed.".format(
```
sloria: This will terminate the script early and may cause other (valid) termination requests to not be approved. It might make more sense to log the error and `continue`.

website/project/model.py
```diff
@@ -3695,776 +3763,6 @@ def to_json(self):
             "anonymous": self.anonymous
         }
 
-
-class Sanction(StoredObject):
```
sloria: Big 👍  for moving these classes to a separate module.

website/tokens/handlers.py
```diff
@@ -53,8 +69,8 @@ def sanction_handler(kind, action, payload, encoded_token, auth, **kwargs):
         err_code = http.BAD_REQUEST
         err_message = 'There is no {0} associated with this token.'.format(Model.DISPLAY_NAME)
     elif sanction.is_approved:
-        err_code = http.BAD_REQUEST if kind in ['registration', 'embargo'] else http.GONE
-        err_message = "This registration is not pending {0}.".format(sanction.DISPLAY_NAME)
+        # Simply strip query params and redirect if already approved
```
sloria: 👍  Good improvement

scripts/approve_embargo_terminations.py
```diff
@@ -40,13 +40,13 @@ def main():
     for request in pending_embargo_termination_requests:
         registration = models.Node.find_one(Q('embargo_termination_approval', 'eq', request))
         if not registration.is_embargoed:
-            raise RuntimeError("Registration {0} associated with this embargo termination request ({0}) is not embargoed.".format(
+            logger.warning("Registration {0} associated with this embargo termination request ({0}) is not embargoed.".format(
                 registration._id,
                 request._id
             ))
```
sloria: You should probably also `continue` here, just to be safe. At the very least, it will prevent incorrect log statements.

sloria:
Note to self for deployment to staging: will need to set `EMBARGO_TERMINATION_PENDING_TIME` to a lower value.

PR #5532, opened 2016-04-25
Addon management via APIv2 [OSF-5987]

mfraezz:
## Purpose
Allow users to programmatically manage their addons via APIv2. Only GoogleDrive has full functionality at the start -- others vary between partial and none.

## Changes
* Endpoints added to API:
    * Node:
         * `<Addon>NodeSettings` list -- view only
         * `<Addon>NodeSettings` detail -- also manages configuration
         * list possible folders for addon -- **GDrive only**
    * User:
        * `<Addon>UserSettings` list and detail -- view only
        * `ExternalAccount` list and detail -- view only
* Serializers for handling `<Addon><Node|User>Settings`, `ExternalAccounts`, addon folders
* Generalized tests for all addons
* Browsable API Documentation
* Follow up ticket created to fix side effect 1 listed below: [OSF-6120](https://openscience.atlassian.net/browse/OSF-6120) 
* **QA**: Minor change to the addon FolderPicker, please test that it works with Drive and at least one other addon.

## Swagger Docs 
With unrelated endpoints removed and duplicated docs collapsed:
[Swagger UI - Addons.pdf](https://github.com/CenterForOpenScience/osf.io/files/296220/Swagger.UI.-.Addons.pdf) or 
swagger ui - addons

## Postman Examples: `GET`
1. Public addon description endpoint
screen shot 2016-06-03 at 11 38 00
2. List addons connected to node. (Only GD here)
screen shot 2016-06-03 at 13 04 11
3. Detail of node-connected addon (equivalent to 9)
screen shot 2016-06-03 at 13 04 34
4. List folders available for addon, starting at root.
screen shot 2016-06-03 at 13 05 00
5. Follow `children` link above -- folder nesting.
screen shot 2016-06-03 at 13 06 19
6. Node detail of a non-enabled addon
screen shot 2016-06-03 at 13 06 52
7. Node detail of an enabled but non-connected addon
screen shot 2016-06-03 at 13 07 43
8. Node detail of a connected but non-configured addon
screen shot 2016-06-03 at 13 08 26
9. Node detail of a configured addon (equivalent to 3)
screen shot 2016-06-03 at 13 08 39
10. Folder list for addon that doesn't yet support it
screen shot 2016-06-08 at 09 28 22
11. List of users connected addons
screen shot 2016-06-03 at 13 15 11
12. Specific addon ExternalAccount (DV has weird variable assignments because of their API)
screen shot 2016-06-03 at 13 16 00
13. Addon with multiple accounts connected
screen shot 2016-06-03 at 13 16 25
14. Specific addon ExternalAccount
screen shot 2016-06-03 at 13 24 13

 

## Side effects
1. Some `TODO`s added. Enabling for addons other than just GDrive requires refactoring the related OSF models, generic_views, tests, and UI, which was out of scope for this task. See ticket [6120](https://openscience.atlassian.net/browse/OSF-6120)
* 854 tests added. If TCI job time increases significantly, consider moving `test_addons` from the `else` job to the `osf` job:
```
~/D/osf.io ❯ DJANGO_SETTINGS_MODULE=api.base.settings nosetests api_tests/users/views/test_user_addons.py api_tests/nodes/views/test_node_addons.py
......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
Ran 854 tests in 109.488s

OK
```

## Ticket

[OSF-5987](https://openscience.atlassian.net/browse/OSF-5987)


api/addons/serializers.py
```diff
@@ -0,0 +1,36 @@
...
+    class Meta:
+        type_ = 'node_addon_folders'
+
+    id = ser.CharField(source='provider', read_only=True)
+    kind = ser.CharField(default='folder', read_only=True)
+    name = ser.CharField(read_only=True)
+    path = ser.CharField(read_only=True)
+    folder_id = ser.CharField(read_only=True)
+
+    links = LinksField({
```
brianjgeiger: This should probably be a relationship field. See https://github.com/CenterForOpenScience/osf.io/blob/develop/api/nodes/serializers.py#L107-L111

api/base/utils.py
```diff
@@ -137,3 +137,9 @@ def default_node_permission_query(user):
 
 def extend_querystring_params(url, params):
     return furl.furl(url).add(args=params).url
+
+def extract_object_from_dict(self, obj):
```
brianjgeiger: This is not immediately clear why it's doing what it is, so please add a docstring.

api/nodes/serializers.py
```diff
@@ -266,6 +269,111 @@ def update(self, node, validated_data):
         return node
 
 
+class NodeAddonSettingsSerializer(JSONAPISerializer):
+    """
+    Overrides AddonSettingsSerializer to return node-specific fields
```
brianjgeiger: This says it overrides AddonSettingsSerializer but it subclasses from JSONAPISerializer.

api/nodes/serializers.py
```diff
@@ -266,6 +269,111 @@ def update(self, node, validated_data):
...
+
+    id = ser.CharField(source='_id', read_only=True)
+    enabled = ser.BooleanField(required=True)
+    auth_id = ser.CharField(source='object.external_account._id', allow_null=True)
+    folder_id = ser.CharField(source='object.folder_id', allow_null=True)
+    folder_path = ser.CharField(source='object.folder_path', required=False, allow_null=True)
+    has_auth = ser.BooleanField(source='object.has_auth', read_only=True)
+    complete = ser.BooleanField(source='object.complete', read_only=True)
+    configured = ser.BooleanField(source='object.configured', read_only=True)
+    node = ser.CharField(source='object.owner._id', read_only=True)
```
brianjgeiger: This should likely be a relationship link to a node rather than just the node id unless that's super inconvenient for the major use case / every use case.

api/nodes/serializers.py
```diff
@@ -266,6 +269,111 @@ def update(self, node, validated_data):
...
+class NodeAddonSettingsSerializer(JSONAPISerializer):
+    """
+    Overrides AddonSettingsSerializer to return node-specific fields
+    """
+    class Meta:
+        type_ = 'node_addons'
+
+    id = ser.CharField(source='_id', read_only=True)
+    enabled = ser.BooleanField(required=True)
+    auth_id = ser.CharField(source='object.external_account._id', allow_null=True)
```
brianjgeiger: Please rename this to something else. `auth` is highly overloaded in the code.

api/nodes/serializers.py
```diff
@@ -266,6 +269,111 @@ def update(self, node, validated_data):
...
+    folder_id = ser.CharField(source='object.folder_id', allow_null=True)
+    folder_path = ser.CharField(source='object.folder_path', required=False, allow_null=True)
+    has_auth = ser.BooleanField(source='object.has_auth', read_only=True)
```
brianjgeiger: If this doesn't mean 'the current user is authorized to use this', then please rename it as well.

api/nodes/serializers.py
```diff
@@ -266,6 +269,111 @@ def update(self, node, validated_data):
...
+    complete = ser.BooleanField(source='object.complete', read_only=True)
+    configured = ser.BooleanField(source='object.configured', read_only=True)
+    node = ser.CharField(source='object.owner._id', read_only=True)
+
+    def to_representation(self, data, envelope='data'):
+        return super(NodeAddonSettingsSerializer, self).to_representation(data, envelope=envelope)
+
+    def update(self, instance, validated_data):
```
brianjgeiger: _Note to self to return to this once there's documentation and when I have more time._

api/nodes/views.py
```diff
@@ -1367,6 +1371,168 @@ def get_object(self):
...
+class NodeAddonList(JSONAPIBaseView, generics.ListAPIView, ListFilterMixin, NodeMixin):
+    """List of addons connected to this node *Read-only*
+
+    TODO: More Docs
+    """
+
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        ContributorOrPublic,
+        ExcludeRetractions,
```
brianjgeiger: This will probably change to ExcludeWithdrawals or similar soon. Keep an eye out when you merge with develop.

api/nodes/views.py
```diff
@@ -1367,6 +1371,168 @@ def get_object(self):
...
+            try:
+                obj = addon_settings.find(Q('owner', 'eq', self.get_node()._id) & Q('deleted', 'eq', False))[0]
+                enabled = True
+            except IndexError:
+                pass
+            qs.append(
+                {
+                    '_id': addon,
+                    'enabled': enabled,
+                    'object': obj
```
brianjgeiger: This seems like it could be a *bit* more descriptive than "object".

mfraezz: ಠ_ಠ 

rn

api/nodes/serializers.py
mfraezz: I'm not certain that there *is* a use case for this field, aside from perhaps following a link to the connected node.

brianjgeiger: Okay, then either remove it or turn it into a relationship field, please. If you were re-writing the front-end to use these endpoints, and you don't have a use for that field, then leave it out.

api/nodes/views.py
brianjgeiger: Your grief is legitimate. I totally should have caught it then.

api/nodes/serializers.py
mfraezz: Its meaning is closer to 'current node is authorized to use this'

api/nodes/serializers.py
```diff
@@ -266,6 +269,111 @@ def update(self, node, validated_data):
...
+    enabled = ser.BooleanField(required=True)
+    auth_id = ser.CharField(source='object.external_account._id', allow_null=True)
+    folder_id = ser.CharField(source='object.folder_id', allow_null=True)
+    folder_path = ser.CharField(source='object.folder_path', required=False, allow_null=True)
+    has_auth = ser.BooleanField(source='object.has_auth', read_only=True)
+    complete = ser.BooleanField(source='object.complete', read_only=True)
+    configured = ser.BooleanField(source='object.configured', read_only=True)
+    node = ser.CharField(source='object.owner._id', read_only=True)
+
+    def to_representation(self, data, envelope='data'):
```
samchrisinger: Can this go away?

api/nodes/serializers.py
```diff
@@ -266,6 +269,111 @@ def update(self, node, validated_data):
...
+        return super(NodeAddonSettingsSerializer, self).to_representation(data, envelope=envelope)
+
+    def update(self, instance, validated_data):
+        addon_name = instance.get('_id', None)
```
samchrisinger: Can you please explain this usage of `_id` as something other than the record's primary key?

api/nodes/serializers.py
```diff
@@ -266,6 +269,111 @@ def update(self, node, validated_data):
...
+        if addon_name not in ADDONS_FOLDER_CONFIGURABLE:
+            raise exceptions.MethodNotAllowed('Requested addon not currently configurable via API.')
+
+        auth = get_user_auth(self.context['request'])
+        node_settings = instance.get('object', None)
+
+        try:
+            auth_id = validated_data['object']['external_account']['_id']
```
samchrisinger: `auth_id = node_settings['external_account']['_id']`

api/addons/serializers.py
```diff
@@ -0,0 +1,36 @@
...
+        'children': 'get_child_folders'
+    })
+
+    def get_child_folders(self, obj):
```
samchrisinger: For some reason I'm thought `obj` was a class instance, but later it's being accessed like a <dict>. Is this correct? Should we be using `getattr` instead?

api/addons/serializers.py
```diff
@@ -0,0 +1,36 @@
+from rest_framework import serializers as ser
+from api.base.serializers import JSONAPISerializer, LinksField
+from api.base.utils import absolute_reverse
+
+class NodeAddonFolderSerializer(JSONAPISerializer):
+    """
+    Overrides AddonSettingsSerializer to return node-specific fields
```
samchrisinger: ditto for Brian's later comment

api/nodes/serializers.py
mfraezz: I know of no immediate plans to refactor the front-end to use this part of the API. Removing

api/nodes/permissions.py
```diff
@@ -22,7 +24,9 @@ def has_object_permission(self, request, view, obj):
 class AdminOrPublic(permissions.BasePermission):
 
     def has_object_permission(self, request, view, obj):
-        assert isinstance(obj, (Node, User, Institution)), 'obj must be a Node, User or Institution, got {}'.format(obj)
+        if isinstance(obj, dict):
```
samchrisinger: Your helper does this conditional implicitly. Maybe consider removing the `if` here and above.

api/base/utils.py
mfraezz: Added in https://github.com/CenterForOpenScience/osf.io/pull/5532/commits/3583141ebb4975a2e1ebe54b7e83b0ee2201e50c

api/nodes/serializers.py
mfraezz: `node_settings` is an `<Addon>NodeSettings` object, and it may or may not have an `ExternalAccount` connected to it already. Either way, we want to get the EA_id that the user is specifying, not the old one that may or may not be there

api/nodes/serializers.py
mfraezz: It was done at the request of @brianjgeiger 

I think the rationale was something along the lines of "only one settings object per addon per node; two birds with one field"

api/nodes/serializers.py
samchrisinger: I'm mostly saying that since you're assigning `node_settings = instance['object']` on L298, you might as well use the variable (possible readability improvement?).

api/nodes/serializers.py
samchrisinger: 😢 
> only one settings object per addon per node;

This field really is the `addon_short_name` correct? I'm not hugely in favor of mixing that concept with the primary key of the record. `_id` has a very specific meaning every other place on the OSF.

api/addons/serializers.py
mfraezz: This serializes folder data returned from the third party API of any given addon. As noted in this [`TODO`](https://github.com/CenterForOpenScience/osf.io/pull/5532/files#diff-51eef6ea2a6c35e2b080c9a6fe591e38R1487), the `<Addon>NodeSettings` models (and likely tests, generic views) require a moderate refactor in order to make this generalizable via APIv2. Given that the only existing use case we have for this feature is for GDrive, and that it was requested a while ago, and that it will need post-model-refactoring refactoring, I wrote this part of the feature somewhat hastily.

api/nodes/serializers.py
mfraezz: `instance` is a dict passed in from the view, grabbed from the database, which contains the node settings object.
`validated_data` is a dict passed in from the user's request, parsed and validated by DRF, which contains data that the user wants to set on the node settings object.

You could maybe get an `external_account_id` from `node_settings`, but it wouldn't be one the user requested to connect, unless it was already connected.

api/nodes/serializers.py
mfraezz: I wasn't either, but it's what the API Czar requested.

api/nodes/serializers.py
samchrisinger: Sorry I was doing a bad job of reading. Thanks for the explanation.

api/nodes/serializers.py
samchrisinger: @brianjgeiger we should really not do this.

brianjgeiger: It doesn't have to be called `_id`, but we need something in the ID field for this API item, and the short name is what makes the most sense in this context. In the view we can call it whatever.

samchrisinger: So at the absolute very least we need to reconcile what `_id` is used for here: https://github.com/CenterForOpenScience/osf.io/pull/5532/files/3f80960e58d2009349c79294ed4e8b0f09486544#diff-7a6d21a50624153d15e63a5f643401c7R279
and what it's used for in this method.

Can we force the inclusion of `short_name` in PUTs/PATCHes? Then `_id` can be the primary key of the `*NodeSettings` instance and `short_name` can provide the context needed for efficient lookups.

api/nodes/serializers.py
```diff
@@ -266,6 +269,104 @@ def update(self, node, validated_data):
...
+                folder_path = None
+            folder_id = {
+                'id': folder_id,
+                'path': folder_path
+            }
+
+        enabled = validated_data.get('enabled', None)
+
+        if enabled is False and (folder_id or external_account_id):
+            raise Conflict('Cannot disable the addon and set the folder/authorization at the same time.')
```
samchrisinger: Given this it could be good to factor these branches into separate helper methods on the class, or at the very least throw in a return after L332.

api/nodes/serializers.py
```diff
@@ -266,6 +269,104 @@ def update(self, node, validated_data):
...
+            if not external_account:
+                raise exceptions.NotFound('Unable to find requested account.')
+            if external_account not in auth.user.external_accounts:
+                raise exceptions.PermissionDenied('Requested action requires account ownership.')
+            if node_settings.external_account and external_account_id != node_settings.external_account._id:
+                node_settings.deauthorize(auth=auth)
+            node_settings.set_auth(external_account, auth.user)
+
+        if (folder_id
+            and not (instance['settings']
```
samchrisinger: :feelsgood: this is hurting my brain some.

api/nodes/serializers.py
```diff
@@ -266,6 +269,104 @@ def update(self, node, validated_data):
...
+        except KeyError:
+            folder_id = None
+            set_folder = False
+
+        if folder_id and addon_name == 'googledrive':
+            try:
+                folder_path = validated_data['settings']['folder_path']
+            except KeyError:
+                folder_path = None
+            folder_id = {
```
samchrisinger: Maybe choose a different name if this is not actually the `folder_id`, but a more complex structure.

api/nodes/serializers.py
```diff
@@ -266,6 +269,104 @@ def update(self, node, validated_data):
...
+                raise exceptions.PermissionDenied('Requested action requires account ownership.')
+            if node_settings.external_account and external_account_id != node_settings.external_account._id:
+                node_settings.deauthorize(auth=auth)
+            node_settings.set_auth(external_account, auth.user)
+
+        if (folder_id
+            and not (instance['settings']
+                and instance['settings'].get('folder_id', False)
+                and (instance['settings']['folder_id'] == folder_id
+                    or instance['settings']['folder_id'] == folder_id.get('folder_id', False)
```
samchrisinger: so `folder_id` can be a str or a dict? 

I wish I had more constructive thoughts on this, but this block is pretty hard to grok.

api/nodes/views.py
```diff
@@ -1367,6 +1371,168 @@ def get_object(self):
...
+    def get_default_queryset(self):
+        qs = []
+        for addon in ADDONS_OAUTH:
+            addon_settings = ADDONS_AVAILABLE_DICT[addon].settings_models.get('node', None)
+            obj = None
+            enabled = False
+            try:
+                obj = addon_settings.find(Q('owner', 'eq', self.get_node()._id) & Q('deleted', 'eq', False))[0]
+                enabled = True
+            except IndexError:
```
samchrisinger: It could be argued the the `enabled = True` belongs in an `else` block in this try/except. I don't have strong feelings about this but it's worth thinking about.

api/nodes/views.py
```diff
@@ -1367,6 +1371,168 @@ def get_object(self):
...
+        base_permissions.TokenHasScope,
+    )
+
+    serializer_class = NodeAddonSettingsSerializer
+    view_category = 'nodes'
+    view_name = 'node-addon-detail'
+
+    def get_object(self):
+        addon = self.kwargs['provider']
+        if addon in ADDONS_OAUTH:
```
samchrisinger: It would be nice to DRY this logic with the block in NodeAddonList.get_default_queryset

api/nodes/views.py
```diff
@@ -1367,6 +1371,168 @@ def get_object(self):
...
+        base_permissions.TokenHasScope,
+    )
+
+    serializer_class = NodeAddonFolderSerializer
+    view_category = 'nodes'
+    view_name = 'node-addon-folders'
+
+    def _to_hgrid(self, item, addon, path):
+        """
+        :param item: contents returned from Google Drive API
```
samchrisinger: Please update this docstring.

Additionally, it would be really nice to not introduce new `hgrid` references into our backend code. Can you explain why we have to do serialization in the view?

api/nodes/views.py
```diff
@@ -1367,6 +1371,168 @@ def get_object(self):
...
+        # TODO: MVP deadlines make code bad
+        from website.addons.base.exceptions import InvalidAuthError
+        from website.addons.googledrive.client import GoogleDriveClient
+
+        try:
+            access_token = node_addon.fetch_access_token()
+        except InvalidAuthError:
+            raise PermissionDenied('Invalid credentials.')
+
+        client = GoogleDriveClient(access_token)
```
samchrisinger: IMO the whole point of refactoring the addons was so we never had to do this in the API. 

If we don't have it already, there needs to be a standardized interface on the *NodeSettings classes to fetch the list of available folders.

If there is a genuine rush to push this feature through, can you please fill me in as to why?

api/users/serializers.py
```diff
@@ -113,6 +113,30 @@ def update(self, instance, validated_data):
...
+
+    def account_links(self, obj):
+        l = {}
+        for account in obj.external_accounts:
+            l[account._id] = {
+                'account': absolute_reverse('users:user-external_account-detail', kwargs={'user_id': obj.owner._id, 'provider': obj.config.short_name, 'account_id': obj._id}),
+                'nodes_connected': [n.absolute_api_v2_url for n in obj.get_attached_nodes(account)]
+            }
+
+        return l
```
samchrisinger:
```python 
return {
    account._id: {
        'account': absolute_reverse('users:user-external_account-detail', kwargs={'user_id': obj.owner._id, 'provider': obj.config.short_name, 'account_id': obj._id}),
        'nodes_connected': [n.absolute_api_v2_url for n in obj.get_attached_nodes(account)]
    }
    for account in obj.external_accounts
}
```

api/users/views.py
```diff
@@ -229,6 +231,95 @@ def get_serializer_context(self):
...
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        base_permissions.TokenHasScope,
+        CurrentUser,
+    )
+
+    serializer_class = AddonAccountSerializer
+    view_category = 'users'
+    view_name = 'user-external_accounts'
+
+    def get_queryset(self):
```
samchrisinger: Please consolidate this with UserAddonDetail.get_object

api_tests/nodes/views/test_node_addons.py
```diff
@@ -0,0 +1,747 @@
...
+            auth=self.user.auth,
+            expect_errors=wrong_type)
+        addon_data = self.get_response_for_addon(res)
+        if not wrong_type:
+            assert_equal(addon_data['external_account_id'], None)
+            assert_false(addon_data['node_has_auth'])
+        if wrong_type:
+            assert_equal(addon_data, None)        
+
+    def test_settings_list_raises_error_if_not_GET(self):
```
samchrisinger: Slight preference for three separate test cases here.

api_tests/nodes/views/test_node_addons.py
```diff
@@ -0,0 +1,747 @@
...
+            assert_equal(addon_data['path'], '/')
+            assert_equal(addon_data['kind'], 'folder')
+            assert_equal(addon_data['name'], '/ (Full Google Drive)')
+            assert_equal(addon_data['folder_id'], 'FAKEROOTID')
+        if wrong_type:
+            assert_equal(res.status_code, 405)
+
+
+    def test_folder_list_raises_error_if_not_GET(self):
+        put_res = self.app.put_json_api(self.folder_url, {
```
samchrisinger: ditto w/ https://github.com/CenterForOpenScience/osf.io/pull/5532/files#r61129207

api_tests/users/views/test_user_addons.py
```diff
@@ -0,0 +1,406 @@
...
+            pass
+        res = self.app.get(
+            self.setting_list_url,
+            auth=self.user.auth)
+
+        addon_data = res.json['data']
+        assert_equal(addon_data, [])
+
+    def test_settings_list_raises_error_if_not_GET(self):
+        put_res = self.app.put_json_api(self.setting_list_url, {
```
samchrisinger: same as https://github.com/CenterForOpenScience/osf.io/pull/5532/files#r61129207

samchrisinger:
@mfraezz pass finished

api/nodes/serializers.py
mfraezz: English translation of this logic would be something like `If we received a folder_id to set, and the settings object wasn't already configured with that folder_id, then [...]`

api/nodes/views.py
mfraezz: From [comment](https://github.com/CenterForOpenScience/osf.io/pull/5532#discussion_r61101918):
>As noted in this [`TODO`](https://github.com/CenterForOpenScience/osf.io/pull/5532/files#diff-51eef6ea2a6c35e2b080c9a6fe591e38R1487), the `<Addon>NodeSettings` models (and likely tests, generic views) require a moderate refactor in order to make [listing of available folders] generalizable via APIv2. Given that the only existing use case we have for this feature is for GDrive, and that it was requested a while ago, and that it will need post-model-refactoring refactoring, I wrote this part of the feature somewhat hastily.

api_tests/nodes/views/test_node_addons.py
mfraezz: I combined them to minimize `setUp` time. This is ran once for each addon, and `setUp/tearDown` seems to take the most time.

samchrisinger: I'll let you make the call, but IMO it's bad testing semantics to combine these into a single case.

api/nodes/views.py
samchrisinger: Discussed in person. For this iteration GoogleDrive specific code is acceptable.

api/nodes/views.py
samchrisinger: If the `_to_hgrid` convention gets propagate via copy/paste into the API, the blood is on your hands. 😺  🔪 🍅

api/nodes/serializers.py
samchrisinger: Basically I'm saying the readability here is pretty poor. You might consider breaking the clauses into separate branches or descriptive variables.

samchrisinger:
As discussed in person, any GDrive specific code is fine for this pass.

Would be nice to see some cleanup in other places, but overall lgtm.

api/nodes/views.py
mfraezz: Renamed to `_from_third_party_format` 😬

api/nodes/serializers.py
mfraezz: The problem with keying `_id` to `<Addon>*Settings._id` is that sometimes the "settings object" that we want to return is actually `None` (see `NodeAddonList` and `NodeAddonDetail`). It can be modified so that the JSONAPI `id` is sourced from `<settings_config>.short_name`, but `id` will ultimately still be `addon_short_name`

api/users/views.py
mfraezz: Fixed in https://github.com/CenterForOpenScience/osf.io/pull/5532/commits/69477f1239f7ac7cc7e000436847d89c5b878661

mfraezz:
CR responded to. Will continue with writing browsable API docs, but this should be ready for another pass in the interim.

mfraezz:
CR responded to, browsable API docs added. This should be ready for another pass @samchrisinger @brianjgeiger 🔕

samchrisinger:
@brianjgeiger do we plan to make this part of the public API right away?  Or wait until latter to show the browsable docs?

mfraezz:
@samchrisinger, GDrive-specific code removed from API. Still the only one that it will work with currently, but aside from changing settings and tests, the API shouldn't require much (hopefully, any) refactoring to enable this feature for additional addons.

brianjgeiger:
Per our meeting, let's remove the faked serialization of unconfigured addons. Instead, implement a /v2/addons/ route that lists all available addons and let the POST to the nodes/addons route with a type of `addon` or whatever and in the attributes we'd have a field for the provider which would be required.

brianjgeiger:
`/v2/addons/` should look like:
```json
{
    "type": "addon",
    "id": shortname,
    "attributes": {
        "name": name,
        "description"; marketing copy for why this is an awesome addon,
        "url": url of the service
    }
}

```

framework/auth/oauth_scopes.py
```diff
@@ -71,6 +77,9 @@ class ComposedScopes(object):
     USERS_READ = (CoreScopes.USERS_READ,)
     USERS_WRITE = USERS_READ + (CoreScopes.USERS_WRITE,)
 
+    USERS_FULL_READ = USERS_READ + (CoreScopes.USERS_ADDON_READ,)
```
abought: `full_read`and `full_write` scopes can be requested by third parties (outside the OSF).

Unless there is a very strong use case to give third parties power over addons, we should limit addon management endpoints to the `osf.admin` scope. (eg, can only be managed via the OSF)

For example, a one-time-only script could be run by taking the user's cookie value. (which is unrestricted access)

Oauth tokens are generally intended to be issued to third parties, and scopes limit what power those parties have on behalf of that user. I'd recommend restricting this to admin to start; it's much harder to take away power than it is to give it.

@brianjgeiger  , is there such a use case?

website/static/js/folderPickerNodeConfig.js
```diff
@@ -510,7 +510,21 @@ var FolderPickerViewModel = oop.defclass({
                     custom : FolderPicker.selectView
                 }
             ];
-            }
+            },
+            xhrconfig : function(xhr) {
```
abought: Take a look at `$osf.setXHRAuthorization` as a possible pre-existing helper

website/static/js/folderPickerNodeConfig.js
```diff
@@ -510,7 +510,21 @@ var FolderPickerViewModel = oop.defclass({
...
+                xhr.withCredentials = true;
+            },
+            lazyLoadPreprocess: function(data) {
+                // Also handle data from API -- squash `attributes` to what TB expects
+                if (data.data) {
+                    for (var i = 0; i < data.data.length; i++) {
+                        $.each(data.data[i].attributes, function(key, value) {
```
abought: Per in person conversation: Tom Baxter mentions a possible solution to this problem in candidate PR #5240 .

api/base/settings/defaults.py
```diff
@@ -189,3 +189,6 @@
 ENABLE_ESI = osf_settings.ENABLE_ESI
 VARNISH_SERVERS = osf_settings.VARNISH_SERVERS
 ESI_MEDIA_TYPES = osf_settings.ESI_MEDIA_TYPES
+
+ADDONS_FOLDER_CONFIGURABLE = ['box', 'dropbox', 's3', 'googledrive']
```
abought: Is hard-coding this consistent with our configuration mechanism for addons in general?

api/base/utils.py
```diff
@@ -137,3 +137,14 @@ def default_node_permission_query(user):
 
 def extend_querystring_params(url, params):
     return furl.furl(url).add(args=params).url
+
+def extract_object_from_dict(obj):
```
abought: As discussed, this function will likely be going away.

api/nodes/views.py
```diff
@@ -1367,6 +1371,213 @@ def get_object(self):
...
+
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        ContributorOrPublic,
+        ExcludeRetractions,
+        base_permissions.TokenHasScope,
+    )
+
+    required_read_scopes = [CoreScopes.NODE_ADDON_READ]
+    required_write_scopes = [CoreScopes.NODE_ADDON_WRITE]
```
abought: This endpoint is read-only; there's probably not much value in defining an allowed write scope. Consider using, eg `CoreScopes.NULL` instead.

abought:
While running, found an issue using a project that had one of everything configured. See https://github.com/mfraezz/osf.io/pull/3 for one possible solution (not thoroughly explored).

api/base/settings/defaults.py
mfraezz: No. However, this will go away when the minor refactors necessary to enable this for other addons go in. The alternative would be either doing all of those in this PR, or adding a flag to the models. Neither option is particularly appealing.

framework/auth/oauth_scopes.py
brianjgeiger: Yes, the reason why we started implementing this was to allow a user to create a set of projects and add their google drive folders to each project based on where in the structure the project existed. There are also use cases that Jeff has requested which will need this capability for third party apps.

In reality, there's very little that we're likely to do with the API that won't be exposed to apps that want the capability.

abought: Remember: scopes are not about what a user can do. (cookies are an all-access pass)

Scopes are about what you'd let a third-party do on your behalf. (eg importing tokens into a specific project; of course the third party has no way of authorizing google drive without you) 

For now, we may want to revisit putting this in the narrow bucket of `USERS_FULL_READ` vs just the very powerful  `FULL_READ`. User-related scopes were originally envisioned for sensitive profile data, not addon management. (viewing someone's email address shouldn't require being able to authorize addons)

abought: Other than that, if there's a use case for third parties, that makes sense to not lock it down completely; I'm just encouraging us to think about the boundaries around data as we go, so that it will be easier to release more restrictive scopes later when we need them.

brianjgeiger: I'm working on scopes right now to try to make them usable.

Also, email addresses are not currently available through the API. But changing your twitter handle might not give the app the rights to connect an addon.

framework/auth/oauth_scopes.py
```diff
@@ -81,6 +87,9 @@ class ComposedScopes(object):
     USERS_READ = (CoreScopes.USERS_READ,)
     USERS_WRITE = USERS_READ + (CoreScopes.USERS_WRITE,)
 
+    USERS_FULL_READ = USERS_READ + (CoreScopes.ADDONS_READ,)
+    USERS_FULL_WRITE = USERS_FULL_READ + (CoreScopes.USERS_ADDON_WRITE,)
```
abought: These variables are defined, but never used.

api/nodes/views.py
```diff
@@ -1514,6 +1519,225 @@ def get_object(self):
...
+
+    required_read_scopes = [CoreScopes.ADDONS_READ]
+    required_write_scopes = [CoreScopes.NODE_ADDON_WRITE]
+
+    serializer_class = NodeAddonFolderSerializer
+    view_category = 'nodes'
+    view_name = 'node-addon-folders'
+
+    def get_queryset(self):
+        # TODO: refactor this/NS models to be generalizable
```
abought: Per discussion on another recent PR, we're trying to encourage all TODO items to be triaged for appropriate followup at the time they are created, rather than at some future date.

Can you do that for this, and other TODOs that you expect to stick around after this gets merged?

api/users/views.py
```diff
@@ -229,6 +231,189 @@ def get_serializer_context(self):
...
+    view_category = 'users'
+    view_name = 'user-external_account-detail'
+
+    def get_object(self):
+        user_settings = self.get_addon_settings()
+        account_id = self.kwargs['account_id']
+
+        account = ExternalAccount.load(account_id)
+        if not (account and account in user_settings.external_accounts):
+            raise NotFound('Requested addon unavailable')
```
abought: I get this message for `http://localhost:8000/v2/users/me/addons/figshare/`, despite this account having at least one project with figshare connected. Is that the expected behavior?

api_tests/nodes/views/test_node_addons.py
```diff
@@ -0,0 +1,714 @@
...
+        assert_equal(res.status_code, 403)
+
+
+class NodeAddonTestSuiteMixin(NodeAddonListMixin, NodeAddonDetailMixin, NodeAddonFolderMixin):
+    def set_urls(self):
+        self.set_setting_list_url()
+        self.set_setting_detail_url()
+        self.set_folder_url()
+
+    def should_expect_errors(self, success_types=['CONFIGURABLE', 'OAUTH']):
```
abought: Even for a simple use case such as this, it's a good idea to avoid mutable default argument values. Eg use a tuple instead of a list for `success_types`.

api/nodes/views.py
```diff
@@ -1514,6 +1519,225 @@ def get_object(self):
...
+        if addon not in ADDONS_OAUTH:
+            raise NotFound('Requested addon unavailable')
+
+        node = self.get_node()
+        if node.has_addon(addon):
+            raise InvalidModelValueError(
+                detail='Add-on {} already enabled for node {}'.format(addon, node._id)
+            )
+
+        serializer.save()
```
abought: Consider calling `super` for the parent method directly, instead of duplicating the code for that method when you override.

From DRF,
```
class CreateModelMixin(object):
    """
    Create a model instance.
    """
...
    def perform_create(self, serializer):
        serializer.save()
```


I mention this because we've been bitten before when DRF refactored internals....

api/nodes/views.py
```diff
@@ -1514,6 +1519,225 @@ def get_object(self):
...
+    #This Request/Response
+    """
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        ContributorOrPublic,
+        ExcludeWithdrawals,
+        base_permissions.TokenHasScope,
+    )
+
+    required_read_scopes = [CoreScopes.ADDONS_READ]
```
abought: This is a tricky one for oauth scopes.

In general we try to draw boundaries around kinds of related data, and make sure that all the types of data returned by a view are specified.

It seems that this returns a list of folder names (even if not specific files?). Depending on how powerful this endpoint is, we may want to add more than one `CoreScope` to this list, to ensure that the user must be allowed to see filesystem AND addon information in order to use this view.

There aren't many `CoreScopes` dealing with file data at present; as far as I can see from the list, `CoreScopes.NODE_FILE_READ` seems a relevant option.

tests/base.py
```diff
@@ -330,6 +332,89 @@ def setUp(self):
...
+        from tests.factories import (
+            ProjectFactory,
+            AuthUserFactory,
+        )
+        from website.addons.base import (
+            AddonOAuthNodeSettingsBase, AddonNodeSettingsBase,
+            AddonOAuthUserSettingsBase, AddonUserSettingsBase
+        )
+        # TODO(mfraezz:) handle type-specific setUp for each type individually
+        assert self.addon_type in ('CONFIGURABLE', 'OAUTH', 'MANAGEABLE', 'UNMANAGEABLE', 'INVALID')
```
abought: Some of these options seem like they could overlap. Are they test-specific, or should the allowed values be documented somewhere?

(I'm trying to flag anything that seems like it could lead to mystery schemas)

api_tests/users/views/test_user_addons.py
```diff
@@ -0,0 +1,422 @@
...
+    def AccountFactory(self):
+        pass
+
+
+class UserUnmanageableAddonTestSuiteMixin(UserAddonTestSuiteMixin):
+    addon_type = 'UNMANAGEABLE'
+
+# UNMANAGEABLE
+
+class TestUserForwardAddon(UserUnmanageableAddonTestSuiteMixin, ApiAddonTestCase):
```
abought: This is certainly an... intricate.... set of test cases. It seems like it will impose certain maintenance burdens as addons are refactored, in order to conform to what is a fairly complex hierarchy of mixins and special properties. 

Am I correct in assuming that the complexity is driven by addons not all being ported to the newer architecture? How necessary is all this specialized behavior?

website/addons/base/__init__.py
```diff
@@ -69,7 +69,7 @@ def __init__(self, short_name, full_name, owners, categories,
                  node_settings_model=None, user_settings_model=None, include_js=None, include_css=None,
                  widget_help=None, views=None, configs=None, models=None,
                  has_hgrid_files=False, get_hgrid_data=None, max_file_size=None, high_max_file_size=None,
-                 accept_extensions=True,
+                 accept_extensions=True, description='', url=None,
```
abought: I don't think this PR is the straw that broke the camel's back... but whenever I see a constructor with this many arguments, I start thinking about ways to refactor and encourage separation of concerns. 

As an actionable example, user-facing text seems like it would be a property of the addon itself, not the configuration object. I count 20+ arguments (plus kwargs) here, which is something to consider going forward.

website/addons/googledrive/tests/test_views.py
```diff
@@ -47,7 +47,7 @@ def test_folder_list_not_root(self, mock_drive_client_folders):
         self.node_settings.set_auth(external_account=self.external_account, user=self.user)
         self.node_settings.save()
 
-        url = self.project.api_url_for('googledrive_folder_list', folderId=folderId)
+        url = self.project.api_url_for('googledrive_folder_list', folder_id=folderId)
```
abought: 👍

website/addons/googledrive/utils.py
```diff
@@ -27,6 +28,7 @@ def to_hgrid(item, node, path):
         'kind': 'folder',
         'name': safe_name,
         'addon': 'googledrive',
+        'provider': 'googledrive',  # TODO: remove duplication -- APIv2
```
abought: Particularly given that we're adding APIv2 endpoints, what is the advantage of retrofitting an older method with a new field?

abought:
Pausing review for tonight. I'll check in with Matt in the AM on a few questions that can't be answered just by skimming. For now, quick requests include revisiting the permissions on views, and tagging and triaging remaining TODO items.

I've confirmed that my previously reported bug is fixed in the newest changeset though.

api/users/views.py
mfraezz: Yes. Only addons that use the `ExternalAccount` architecture are supported, and figshare does not yet.

api_tests/users/views/test_user_addons.py
mfraezz: To test that any addon specified by an incoming behaves as expected; necessary. This is less burdensome than it would be if these were repeated for every addon -- it's the same rationale that went into creating generic OSF tests for addons (see `/website/addons/base/testing/`). 

After resolution of the follow-up ticket, one of these types will go away, as the behavior of `CONFIGURABLE` and `OAUTH` will be equivalent. `UNMANAGEABLE` are addons that you can't interact with via APIv2, and `INVALID` are ones that don't exist.

I realize now that `MANAGEABLE` isn't actually used; two of these types will go away.

website/addons/googledrive/utils.py
mfraezz: Language usage needs to be normalized -- APIv2 requires the service provider to be keyed to `'provider'`, current front-end / v1 code's key is `'addon'`. Updating this everywhere seemed out of scope.

tests/base.py
mfraezz: Test-specific, shouldn't be used outside of addon-related API tests. Above `TODO` is possibly unnecessary.

website/addons/base/__init__.py
mfraezz: The configuration object *is* the addon. It holds all the meta-information about the addon, such as name, category (`'storage'`, `'security'`, `'documentation'`), associated models, etc.

As for :camel: straws, I agree that the quantity of `args` is pushing absurdity, but that refactor seemed more like an `Improvement` than a `Feature`.

api/nodes/views.py
mfraezz: This is not a powerful endpoint. `NODE_FILE_READ` may be appropriate for the `NodeAddonFolderList` view, but this view and the related detail view only return configuration info about the selected folder for this addon, if one has been selected. It's nothing that someone who can view the node wouldn't be able to see.

api/nodes/views.py
mfraezz: ...Like the [follow-up ticket](https://openscience.atlassian.net/browse/OSF-6120) created to resolve these TODOs, you mean?

brianjgeiger: The other part of that is to tag in the code. `# TODO: [OSF-1234 Imshould do the thing with the code.]`

abought: Thanks for making the ticket! For @brianjgeiger 's sake, I'd encourage you to tag all remaining TODOs in the code with a reference to the followup ticket before this is merged.

(applies throughout this PR, not just to one line)

website/addons/googledrive/utils.py
abought: I see. I assume this is because Google Drive specifically was inconsistent, while the other addons already contained a key named `provider`?

(if every provider has the same apiv1/apiv2 difference, an alternative would be to specify the `source=` argument in the DRF serializer field definition. Different ways of approaching the duplicate code problem)

brianjgeiger: Is `to_hgrid` sending things to the v2 api serializer? That seems less than ideal. You could just use the source field on the v2 serializer to tell it which field to use. Or am I misunderstanding?

framework/auth/oauth_scopes.py
```diff
@@ -110,9 +116,9 @@ class ComposedScopes(object):
     ORGANIZER_WRITE = ORGANIZER_READ + (CoreScopes.ORGANIZER_COLLECTIONS_BASE_WRITE, CoreScopes.NODE_LINKS_WRITE)
 
     # Privileges relating to editing content uploaded under that node # TODO: Add wiki etc when implemented
-    NODE_DATA_READ = (CoreScopes.NODE_FILE_READ, )
+    NODE_DATA_READ = (CoreScopes.NODE_FILE_READ, CoreScopes.ADDONS_READ)
```
abought: Should allowing a third party to read project files also inherently mean letting them see addon configuration info?

Even though scopes happen on top of existing permissions controls, it doesn't seem that addon endpoints need to be accessible for everyday file usage.

For the future, it seems like what we really want is a separate scope that handles node configuration management... if that's something we allow third parties to do at all.

website/addons/base/__init__.py
abought: Like I said, not a blocker for this PR in any way- more a general impression. The very existence of a constructor implies that this data lives somewhere else already (addon settings presumably)...

abought:
Discussed comments with dev; passing ticket back while he makes the requested changes.

website/addons/googledrive/utils.py
mfraezz: >Is `to_hgrid` sending things to the v2 api serializer? That seems less than ideal. You could just use the source field on the v2 serializer to tell it which field to use. Or am I misunderstanding?

Yes, kind of. There's some amount of normalization required between receiving folder info from any third party, most addons have some helper to do this, some do it in the same function that makes the `GET` request. Be not alarmed, `rubeus` isn't actually being used.

`source=` will probably work.

abought:
Tests are failing, including some that appear related to this PR.

framework/auth/oauth_scopes.py
```diff
@@ -124,8 +128,8 @@ class ComposedScopes(object):
     NODE_ALL_WRITE = NODE_ALL_READ + NODE_METADATA_WRITE + NODE_DATA_WRITE + NODE_ACCESS_WRITE
 
     # Full permissions: all routes intended to be exposed to third party API users
-    FULL_READ = NODE_ALL_READ + USERS_READ + ORGANIZER_READ + GUIDS_READ + (CoreScopes.INSTITUTION_READ, )
-    FULL_WRITE = NODE_ALL_WRITE + USERS_WRITE + ORGANIZER_WRITE + GUIDS_READ
+    FULL_READ = NODE_ALL_READ + USERS_READ + ORGANIZER_READ + GUIDS_READ + (CoreScopes.INSTITUTION_READ, CoreScopes.ADDONS_READ, )
+    FULL_WRITE = NODE_ALL_WRITE + USERS_WRITE + ORGANIZER_WRITE + GUIDS_READ + (CoreScopes.NODE_ADDON_WRITE, )
```
abought: Under what circumstances would we want a third party to be able to reconfigure an addon on behalf of a user

In particular, because the tokens we collect are issues on behalf of the entire account (not a folder), it seems that this would let a third party effectively browse the entire contents of the user's account, beyond the folders they intended to make visible.

See doc comment on route capabilities here: https://github.com/CenterForOpenScience/osf.io/pull/5532/files#diff-51eef6ea2a6c35e2b080c9a6fe591e38R1589

Based on flowdock discussion, I'll refer this to @brianjgeiger to revisit the use case, and see if we can't work out any way to mitigate this.

api/nodes/views.py
```diff
@@ -1514,6 +1519,225 @@ def get_object(self):
...
+    Attributes other than `enabled` will be null if the addon is not enabled for this node.
+
+    ##Permissions
+
+    <Addon>NodeSettings that are attached to public Nodes will give read-only access to everyone. Private nodes require explicit read
+    permission. Write and admin access are the same for public and private nodes. Administrators on a parent node have
+    implicit read permissions for all child nodes.
+
+    Any users with write or admin access to the node are able to deauthorize an enabled addon, but only the addon authorizer is able
+    to change the configuration (i.e. selected folder) of an alread-configured <Addon>NodeSettings entity.
```
abought: Typo, should be "already".

Spellchecking prose is *always* recommended throughout; I'll leave that as an exercise to the developer.

api/addons/views.py
```diff
@@ -0,0 +1,76 @@
...
+        categories  list        List of categories this addon belongs to
+
+    #This Request/Response
+    """
+    permission_classes = (
+        drf_permissions.AllowAny,
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        TokenHasScope, )
+
+    required_read_scopes = [CoreScopes.ADDONS_READ]
```
abought: Why does the addon list endpoint rely on the same core scope as the node addon detail endpoint?

One gives generic information about what an addon is, and the other gives information about how the addon is configured for a node. (folders etc)

Core scopes represent the kind of data that someone can see from that view. If two views give out different data with no overlap, they should use different core scopes.

api/nodes/views.py
abought: Since this gives out information about folder names (which might not be exposed via other means), I still lean more in favor of some extra restriction here.

framework/auth/oauth_scopes.py
```diff
@@ -38,6 +40,8 @@ class CoreScopes(object):
     NODE_FILE_READ = 'nodes.files_read'
     NODE_FILE_WRITE = 'nodes.files_write'
 
+    NODE_ADDON_WRITE = 'nodes.addon_write'
```
abought: Per above comment, you may want a `NODE_ADDON_READ` scope here.

abought:
Passing back to dev for now while we try to nail down the access controls.

framework/auth/oauth_scopes.py
brianjgeiger: We've been requested to consider, among other use cases, allowing someone to make an application that will allow a user to pick a github repo they control and, with minimal clicking, create an OSF user, link the github account to their OSF account, and create a project with that github repo attached. Possibly some other flourishes in addition.

That being said, it is true that said scope could allow an app to link to, say, the root of someone's dropbox, create a project, archive all the info from that dropbox into the new project, add another user to the project, and remove the authorizing user from that project. So we might want to have a process in place for blessing apps that do that kind of thing.

abought: > We've been requested to consider, among other use cases, allowing someone to make an application that will allow a user to pick a github repo they control and, with minimal clicking, create an OSF user, link the github account to their OSF account, and create a project with that github repo attached. Possibly some other flourishes in addition.

TLDR: Because we have already started handing these tokens out as the default option, we need to seriously revisit how we're handling `full_read`. Realistically, there should be a totally separate scope for things that touch on sensitive account management.

1. What you describe is an extremely powerful capability- especially given the way github encourages me to mix my personal and professional identity under the same account. I don't want to connect for science, only to find that a client's trade secrets are now vulnerable.

2. At the absolute minimum, we should not give this kind of thing out by default in the absence of a concrete and specific need, and even then, only **after** a careful process is in place. 

3. It is easier to withhold permissions until actually needed than to take them away following a crisis.

4. Making `full_read` less than the full set of powers may admittedly be confusing, but it's a painful consequence of the current system being live and public-facing. In an extreme case, there are mechanisms in place to support deprecating token names entirely.

mfraezz: Fixed in https://github.com/CenterForOpenScience/osf.io/pull/5532/commits/c0b28126dffd2b5175259811ab693adde0a591e2

Added to `ComposedScopes.ADMIN_LEVEL`

brianjgeiger:
Could you include request/response json in the changes section as well as screenshots of the swagger docs related to this PR?

mfraezz:
@brianjgeiger, screenshots added.

brianjgeiger:
What's the status code on this? I think 404 is probably appropriate. The error detail is still useful to keep, though.

screen shot 2016-06-07 at 4 18 55 pm

For this one, if we're going to support in the future, 501 might be more appropriate. I think 405 is for something like issuing a DELETE to a read-only endpoint. If it's never appropriate for some addons, then 404 is probably the way to go.

screen shot 2016-06-07 at 4 26 40 pm

api/nodes/permissions.py
```diff
@@ -11,7 +12,9 @@
 class ContributorOrPublic(permissions.BasePermission):
 
     def has_object_permission(self, request, view, obj):
-        assert isinstance(obj, (Node, Pointer)), 'obj must be a Node or Pointer, got {}'.format(obj)
+        if isinstance(obj, AddonSettingsBase):
+            obj = obj.owner
+        assert isinstance(obj, (Node, Pointer)), 'obj must be a Node, Pointer, or AddonSettings got {}'.format(obj)
```
brianjgeiger: Super-minor: 'or AddonSettings; got {}'

api/nodes/permissions.py
```diff
@@ -30,7 +33,7 @@ def has_object_permission(self, request, view, obj):
 class AdminOrPublic(permissions.BasePermission):
 
     def has_object_permission(self, request, view, obj):
-        assert isinstance(obj, (Node, User, Institution)), 'obj must be a Node, User or Institution, got {}'.format(obj)
+        assert isinstance(obj, (Node, User, Institution, AddonSettingsBase)), 'obj must be a Node, User, Institution, or AddonSettings got {}'.format(obj)
```
brianjgeiger: Same here with the semicolon.

api/nodes/serializers.py
```diff
@@ -292,6 +295,124 @@ def update(self, node, validated_data):
...
+                ))):
+            if auth.user._id != node_settings.user_settings.owner._id:  # And the user is allowed to do this
+                raise exceptions.PermissionDenied('Requested action requires addon ownership.')
+            return True
+        return False
+
+    def update(self, instance, validated_data):
+        addon_name = instance.config.short_name
+        if addon_name not in ADDONS_FOLDER_CONFIGURABLE:
+            raise exceptions.MethodNotAllowed('Requested addon not currently configurable via API.')
```
brianjgeiger: 501 as mentioned earlier.

api/nodes/views.py
```diff
@@ -1529,6 +1534,225 @@ def get_object(self):
...
+
+    required_read_scopes = [CoreScopes.NODE_ADDON_READ, CoreScopes.NODE_FILE_READ]
+    required_write_scopes = [CoreScopes.NULL]
+
+    serializer_class = NodeAddonFolderSerializer
+    view_category = 'nodes'
+    view_name = 'node-addon-folders'
+
+    def get_queryset(self):
+        # TODO: [OSF-6120] refactor this/NS models to be generalizable
```
brianjgeiger: 👍

website/static/js/folderPickerNodeConfig.js
```diff
@@ -509,7 +509,20 @@ var FolderPickerViewModel = oop.defclass({
                     custom : FolderPicker.selectView
                 }
             ];
-            }
+            },
+            xhrconfig: $osf.setXHRAuthorization,
+            lazyLoadPreprocess: function(data) {
```
brianjgeiger: Since this is OSF user-facing, it should be called out in changes so QA knows to check for it, yes?

brianjgeiger:
I think I asked this a month or two ago, but you know your tests work, right? If something is implemented incorrectly in an addon, tests will fail appropriately? Obvs don't write tests for the tests or anything, just verifying.

brianjgeiger:
Pass done. If there are no quibbles, once changes are made and tests pass, move to RTM. This will likely have to follow a feature process for going through QA rather than an improvements process.

mfraezz:
>What's the status code on this? I think 404 is probably appropriate

404, raised by `AddonSettingsMixin.get_addon_settings`.

>For this one, if we're going to support in the future, 501 might be more appropriate. I think 405 is for something like issuing a DELETE to a read-only endpoint. If it's never appropriate for some addons, then 404 is probably the way to go.

A 404 will be raised if it's not in `ADDONS_OAUTH`, changing the failure at folder listing to raise a 501 and modifying the detail for clarity.

api/nodes/serializers.py
```diff
@@ -292,6 +296,124 @@ def update(self, node, validated_data):
...
+        type_ = 'node_addons'
+
+    id = ser.CharField(source='config.short_name', read_only=True)
+    external_account_id = ser.CharField(source='external_account._id', required=False, allow_null=True)
+    folder_id = ser.CharField(required=False, allow_null=True)
+    folder_path = ser.CharField(required=False, allow_null=True)
+    node_has_auth = ser.BooleanField(source='has_auth', read_only=True)
+    configured = ser.BooleanField(read_only=True)
+
+    links = links = LinksField({
```
sloria: Minor typo here.

website/static/js/folderPickerNodeConfig.js
```diff
@@ -509,7 +509,20 @@ var FolderPickerViewModel = oop.defclass({
...
+                // Also handle data from API -- squash `attributes` to what TB expects
+                // TODO: [OSF-6384] DRY this up when PR #5240 goes in
+                if (data.data) {
+                    $.each(data.data, function(i, obj) {
+                        var saved_attributes = obj.attributes;
```
sloria: Minor: As per COSDev, use lowerCamelCase for variable names.

I can fix this when I merge.

sloria:
This is awesome. Once this is up to date with develop, this is good to go to staging2.

sloria:
This is now on `feature/apiv2-addons`. Consider this merged. 🐙

PR #5534, opened 2016-04-25
Feature/APIv2 Fork Functionality [OSF-3897]

pattisdr:
JIRA https://openscience.atlassian.net/browse/OSF-3897

## Purpose

Adds ability for fork projects and get list of forks via the API v2.  

## Changes

Functionality added: 

- GET /v2/nodes/{node_id}/forks/
- POST /v2/nodes/{node_id}/forks/
- GET /v2/registrations/{registration_id}/forks/
- POST /v2/registrations/{registration_id}/forks/

To create a fork, issue a POST request to the node forks or the registration forks list endpoint.  Request body should look like this:
```
{
    "data": {
        "type": "nodes", # MANDATORY
        "attributes": {
            "title": "{insert_title}" # OPTIONAL
        }
    }
}

```
`forks` relationship link added to Node and Registration serializers.  Also, the title for the fork can be specified up-front, rather than having to edit the title.  If no title supplied, default is still 'Fork of ' + current node title.

### DIFF
--- a/api/nodes/serializers.py
+++ b/api/nodes/serializers.py
@@ -131,6 +131,11 @@ class NodeSerializer(JSONAPISerializer):
         related_view_kwargs={'node_id': '<forked_from_id>'}
     )
 
+    forks = RelationshipField(
+        related_view='nodes:node-forks',
+        related_view_kwargs={'node_id': '<pk>'}
+    )
+
     node_links = RelationshipField(
         related_view='nodes:node-pointers',
         related_view_kwargs={'node_id': '<pk>'},
--- a/api/nodes/serializers.py
+++ b/api/nodes/serializers.py
@@ -273,6 +273,30 @@ class NodeDetailSerializer(NodeSerializer):
     id = IDField(source='_id', required=True)
 
 
+class NodeForksSerializer(NodeSerializer):
+
+    category_choices = Node.CATEGORY_MAP.items()
+    category_choices_string = ', '.join(["'{}'".format(choice[0]) for choice in category_choices])
+
+    title = ser.CharField(required=False)
+    category = ser.ChoiceField(read_only=True, choices=category_choices, help_text="Choices: " + category_choices_string)
+    forked_date = ser.DateTimeField(read_only=True)
+
+    def create(self, validated_data):
+        node = validated_data.pop('node')
+        fork_title = validated_data.pop('title', 'Fork of ')
+        request = self.context['request']
+        auth = get_user_auth(request)
+        fork = node.fork_node(auth, title=fork_title)
+
+        try:
+            fork.save()
+        except ValidationValueError as e:
+            raise InvalidModelValueError(detail=e.message)
+
+        return fork
+
+
 class NodeContributorsSerializer(JSONAPISerializer):
     """ Separate from UserSerializer due to necessity to override almost every field as read only
     """
--- a/api/nodes/serializers.py
+++ b/api/nodes/serializers.py
@@ -127,7 +127,7 @@ class NodeSerializer(JSONAPISerializer):
     )
 
     forked_from = RelationshipField(
-        related_view='nodes:node-detail',
+        related_view=lambda n: 'registrations:registration-detail' if getattr(n, 'is_registration', False) else 'nodes:node-detail',
         related_view_kwargs={'node_id': '<forked_from_id>'}
     )
 
--- a/api/nodes/views.py
+++ b/api/nodes/views.py
@@ -21,6 +21,7 @@ from api.users.views import UserMixin
 from api.nodes.serializers import (
     NodeSerializer,
     NodeLinksSerializer,
+    NodeForksSerializer,
     NodeDetailSerializer,
     NodeProviderSerializer,
     NodeContributorsSerializer,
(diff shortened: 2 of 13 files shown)


api/nodes/serializers.py
```diff
@@ -127,10 +127,15 @@ class NodeSerializer(JSONAPISerializer):
     )
 
     forked_from = RelationshipField(
-        related_view='nodes:node-detail',
+        related_view=lambda n: 'registrations:registration-detail' if getattr(n, 'is_registration', False) else 'nodes:node-detail',
```
brianjgeiger: Look who's gotten all _fancy_!

api/nodes/serializers.py
```diff
@@ -273,6 +278,30 @@ class NodeDetailSerializer(NodeSerializer):
     id = IDField(source='_id', required=True)
 
 
+class NodeForksSerializer(NodeSerializer):
+
+    category_choices = Node.CATEGORY_MAP.items()
```
brianjgeiger: These don't carry over from the NodeSerializer?

api/nodes/serializers.py
pattisdr: :) @chrisseto helped me with implement this elsewhere awhile back.  Remember how `node` and `original_node` can be nodes or registrations on logs?  So we used this and then I put a check in the RelationshipField to see if the view is a callable and then fetch the node-detail or registration-detail view.

api/nodes/serializers.py
pattisdr: I didn't originally carry them over and it was saying the category items weren't defined.

api/nodes/serializers.py
```diff
@@ -273,6 +278,30 @@ class NodeDetailSerializer(NodeSerializer):
...
+    category_choices_string = ', '.join(["'{}'".format(choice[0]) for choice in category_choices])
+
+    title = ser.CharField(required=False)
+    category = ser.ChoiceField(read_only=True, choices=category_choices, help_text="Choices: " + category_choices_string)
+    forked_date = ser.DateTimeField(read_only=True)
+
+    def create(self, validated_data):
+        node = validated_data.pop('node')
```
brianjgeiger: Functional testing seems to indicate that fields that are not the title are not carrying over. 


```json
{
    "data": {
        "type": "nodes",
        "attributes": {
            "title": "Quicker test with category and description",
            "category": "project",
            "description": "A super-quick test"
        }
    }
}
```
Is creating a node with a new title but not a new category/description.

brianjgeiger: Reading through the docs, it appears that it's not supported to change anything but the title on creation. That's probably fine then. We don't support more than that on the front end, do we?

pattisdr: Yes, I was trying to more-or-less match front-end functionality.  Actually when creating a fork, you can't even change the title, it is automatically 'Fork of + ' current node title.  But the method for creating a fork already had the title as an argument so I thought forking would be improved by allowing you to change the title on creation.  But to mach the front end completely, no fields change on creation.

brianjgeiger: Okay, cool, let's keep it like this.

api/nodes/views.py
```diff
@@ -1069,6 +1071,109 @@ def perform_destroy(self, instance):
...
+    required_read_scopes = [CoreScopes.NODE_FORKS_READ]
+    required_write_scopes = [CoreScopes.NODE_FORKS_WRITE]
+
+    serializer_class = NodeForksSerializer
+    view_category = 'nodes'
+    view_name = 'node-forks'
+
+    # overrides ListCreateAPIView
+    def get_queryset(self):
+        all_forks = self.get_node().forks
```
brianjgeiger: Is `forks` a list on the model?

brianjgeiger: Ah, no, it's not. It's a property. Phew.

api_tests/nodes/views/test_node_forks_list.py
```diff
@@ -0,0 +1,278 @@
...
+        assert_equal(res.status_code, 401)
+        assert_equal(res.json['errors'][0]['detail'], 'Authentication credentials were not provided.')
+
+    def test_cannot_fork_private_node_logged_in_non_contributor(self):
+        res = self.app.post_json_api(self.private_project_url, self.fork_data, auth=self.user_two.auth, expect_errors=True)
+        assert_equal(res.status_code, 403)
+        assert_equal(res.json['errors'][0]['detail'], 'You do not have permission to perform this action.')
+
+    def test_can_fork_private_node_logged_in_contributor(self):
+        res = self.app.post_json_api(self.private_project_url + '?embed=children&embed=node_links&embed=logs&embed=contributors&embed=forked_from', self.fork_data, auth=self.user.auth, expect_errors=True)
```
brianjgeiger: You shouldn't have 'expect_errors' if you don't.

brianjgeiger:
I wonder if we should hold this until the withdrawal PR is through, so we can have tests for ensuring you can't fork a withdrawn registration.

pattisdr:
I do have a test for that here, but the withdrawal PR will group all the retractions items together.

website/project/model.py
```diff
@@ -2119,7 +2119,11 @@ def fork_node(self, auth, title='Fork of '):
                 if forked_node is not None:
                     forked.nodes.append(forked_node)
 
-        forked.title = title + forked.title
+        if title == 'Fork of ' or title == '':
```
brianjgeiger: Was this necessary because you're the first person using the "provide your own title" functionality?

brianjgeiger:
Well, whichever direction it goes, you'll probably have to handle the conflict or error. So it should be fine. :)

website/project/model.py
pattisdr: Yes. The method is written to prepend whatever title is supplied to the beginning of the original node title.  When no title is supplied, 'Fork of ' is added to the original node title.  If the method is forking your components, an empty string is added to the original node title.

api_tests/registrations/views/test_registration_forks.py
```diff
@@ -0,0 +1,295 @@
...
+        assert_equal(res.status_code, 401)
+        assert_equal(res.json['errors'][0]['detail'], 'Authentication credentials were not provided.')
+
+    def test_cannot_fork_private_registration_logged_in_non_contributor(self):
+        res = self.app.post_json_api(self.private_registration_url, self.fork_data, auth=self.user_two.auth, expect_errors=True)
+        assert_equal(res.status_code, 403)
+        assert_equal(res.json['errors'][0]['detail'], 'You do not have permission to perform this action.')
+
+    def test_can_fork_private_registration_logged_in_contributor(self):
+        res = self.app.post_json_api(self.private_registration_url + '?embed=children&embed=node_links&embed=logs&embed=contributors&embed=forked_from', self.fork_data, auth=self.user.auth, expect_errors=True)
```
brianjgeiger: Also here with the expect_errors

brianjgeiger:
Okay, one tiny problem that showed up twice, otherwise great. 🐧

pattisdr:
Hooray! Thanks for reviewing so quickly!

brianjgeiger:
This looks good. I did functional tests on nodes but not registrations due to the poor state of my local environment. You are confident in the functioning of the registration endpoints?

pattisdr:
Yes I am confident. Registrations work like nodes, except in the case of retractions where attempting to fork will fail with a 403.

pattisdr:
I did a lot of testing on my local environment with registrations specifically.

website/project/model.py
sloria: In order to avoid special-casing `"Fork of"` how about making `title` default to `None`?

```python
def fork_node(..., title=None):
    PREFIX = 'Fork of '
    if title is None:
         title = PREFIX + forked.title
```

api/nodes/serializers.py
sloria: `category_choices` and `category_choices_string` *will* be inherited from `NodeSerializer`, but they won't be in scope for the `category` field. Therefore, you'll have to access them off of `NodeSerializer`.

```python
category = ser.ChoiceField(read_only=True, choices=NodeSerializer.category_choices, ...)
```

api/nodes/views.py
```diff
@@ -1069,6 +1071,109 @@ def perform_destroy(self, instance):
...
+        auth = get_user_auth(self.request)
+        return sorted([node for node in all_forks if node.can_view(auth)], key=lambda n: n.forked_date, reverse=True)
```
sloria: Rather than sorting with Python, can you do it in the database?

```python
self.get_node().forks.sort('-forked_date')
```

sloria:
Looks good. Just one suggestion, https://github.com/CenterForOpenScience/osf.io/pull/5534/files#r62691200, but it isn't blocking. @pattisdr Feel free to address that in a follow up PR.

website/project/model.py
pattisdr: Still have to handle components being forked. Component title would be an empty string ('').  With your suggestion, component title would be empty string, not None, which would make component title be empty string instead of the forked title.

pattisdr: @sloria 
```
PREFIX = 'Fork of '
if title is None:          # Regular case - top level node becomes Fork of + original node title.
    forked.title = PREFIX + forked.title
elif title == '':            # Component being forked - forked component title same as original component title
    forked.title = forked.title
else:
    forked.title = title  # Case where title specified
```

sloria: Looks good to me. Would be helpful to doc those semantics in the docstring.

pattisdr: Okay.  Will be a part of https://github.com/CenterForOpenScience/osf.io/pull/5611

PR #5538, opened 2016-04-26
Set the X-Frame-Options header to disallow embedding into frames

chrisseto:
## Purpose

Set the `X-Frame-Options` header for all requests.

## Changes

Set the `X-Frame-Options` header for all requests.

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-6113
https://openscience.atlassian.net/browse/SEC-15

### Before:
screen shot 2016-04-26 at 12 06 06

### After:
screen shot 2016-04-26 at 12 04 37

### DIFF
--- a/framework/sessions/__init__.py
+++ b/framework/sessions/__init__.py
@@ -174,4 +174,7 @@ def after_request(response):
     if session.data.get('auth_user_id'):
         session.save()
 
+    # Disallow embeding in frames
+    response.headers['X-Frame-Options'] = 'SAMEORIGIN'
+
     return response


icereval:
@chrisseto, I believe this breaks MFR when I tested last (may need to fake subdomains in /etc/hosts)

chrisseto:
@icereval This should only affect the osf going into iframes. IIRC The mfr iframe is on the mfr domain?
I'll double check.

chrisseto:
@icereval, MFR is unaffected as long as `ALLOWED_PROVIDER_DOMAINS` contains the osf domain.
screen shot 2016-04-28 at 11 29 51

icereval:
LGTM

PR #5539, opened 2016-04-26
Use constraints file to resolve dependency conflicts

abought:
## Purpose
Speed up installation of python requirements, by implementing some basic dependency resolution.

### Options for future scope creep
- Pin dependencies for certain addons that are currently allowed to auto-upgrade to newest
- Remove dependency on personal repositories (Lyndsy, Josh, etc)
- Modernize very old dependencies (after thorough review)

## Changes
- Add a constraints file. Dependency choices were based on analysis of logs without constraints. In general, we chose versions to match the version that was present after final installation- since each run of pip happens separately, the version in OSF base requirements "wins".
- Update all invoke tasks that use pip so that they recognize the constraints file
- Add a reminder into the base requirements file

## Side effects
In general, the final result WITH the constraints file should be the same set of packages/versions as if we ran with no constraints file- but it should happen faster (because we don't need to wait for lxml etc to install twice). See artifacts on linked ticket for audit.

There are some disadvantages:

- Developer virtual environments must be upgraded to pip >= 7.1 (released June 30, 2015)
- When upgrading packages to a newer pinned version, we will need to check if it is defined in `constraints.txt`. 

It's a slight maintenance headache, but provides a simple mechanism for resolving conflicts caused by external dependencies whose requirements we don't control.

## Ticket
https://openscience.atlassian.net/browse/OSF-5959

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 # Base requirements for running the OSF.
 # NOTE: This does not include addon, development, metrics or release requirements.
+# NOTE: When updating pinned version, you may also need to update constraints.txt
 # To install addon requirements: inv requirements --addons
 # To install dev requirements: inv requirements --dev
 # To install metrics requirements: inv requirements --metrics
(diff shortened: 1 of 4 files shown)


sloria:
Very nice

PR #5542, opened 2016-04-26
[hotfix] Remove forward slashes from wiki page names

samanehsan:
## Purpose

There are several wiki pages on production with a "/" in the title and because those titles are invalid the wiki pages cannot be saved.

## Changes

Remove the "/" from wiki page titles so that they can be handled by #5524 

## Ticket

Related to https://github.com/CenterForOpenScience/osf.io/pull/5524

### DIFF
--- a/scripts/remove_wiki_title_forward_slashes.py
+++ b/scripts/remove_wiki_title_forward_slashes.py
@@ -1,0 +1,61 @@
+"""
+Remove forward slashes from wiki page titles, since it is no longer an allowed character and
+breaks validation.
+"""
+import logging
+import sys
+
+from framework.mongo import database as db
+from framework.transactions.context import TokuTransaction
+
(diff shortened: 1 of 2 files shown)


scripts/remove_wiki_title_forward_slashes.py
```diff
@@ -0,0 +1,61 @@
...
+
+from website.app import init_app
+from website.project.model import Node
+from scripts import utils as script_utils
+
+logger = logging.getLogger(__name__)
+
+
+def main():
+    wiki_pages = db.nodewikipage.find({}, {'_id': True, 'page_name': True, 'node': True})
```
sloria: Mongodb allows for regex queries, so you should be able to do

```python
db.nodewikipage.find({'page_name': {'$regex': '/'}}, ...
```

PR #5543, opened 2016-04-26
Fix error messages on tag deletion [OSF-5257]

abought:
## Purpose
Files and projects allow tagging. Hitting the "delete" key on an empty tag box on the file or project page should not generate a red growl box / 400 server error.

## Changes
- DON'T let the user try to delete empty tags (suppress server request for files and project dashboard page when hitting delete key in empty box)
- DO warn when they try to do so for a project (emulating how [osfstorage files](https://github.com/abought/osf.io/blob/268c3eeea2b0a31fe84a226b7bfd0d7896a09b06/website/files/models/osfstorage.py#L238-L238) handle it), by making the server [send back](https://github.com/abought/osf.io/blob/3524cd3eb055b4f57e393a5b8a98c484087433fc/website/addons/osfstorage/views.py#L258-L258) a 400 error
- DO show a growl box on the project dashboard when tag deletion fails (similar to file detail page)
- DO add regression tests for situations that should generate a 400 / 409 error

Updates Node `remove_tag` method to mimic the behavior of osfstorage file's `remove_tag` method. Specifically, the method will return false as an indication that something failed to delete.

Provide two separate error codes: `400` (generic invalid request) or `409` (couldn't delete tag due to [conflict](https://httpstatuses.com/409)). The latter will help us to avoid giving users a spurious error message- if the tag couldn't be deleted, we can just treat it as gone from a UI perspective without bothering anyone.

## Side effects
None expected.

There is one [further use](https://github.com/abought/osf.io/blob/f34edce523e12156d4b31f81935efccc3f2ed62d/api/nodes/serializers.py#L254-L254) of `Node.remove_tag` that might benefit from being retrofitted with this new behavior, but assessing that is outside the scope of this PR.

## Ticket
https://openscience.atlassian.net/browse/OSF-5257


brianjgeiger:
I am having trouble figuring out how to test this functionally. Sometime this afternoon, could you demonstrate?

website/static/js/pages/project-dashboard-page.js
```diff
@@ -145,17 +145,15 @@ $(document).ready(function () {
...
-                dataType: 'JSON',
-                data: data
-            });
+            // Don't try to delete a blank tag (would result in a server error)
+            if (!tag) {
+                return false;
+            }
+            var request = $osf.ajaxJSON('DELETE', url, {'data': {'tag': tag}});
             request.fail(function(xhr, textStatus, error) {
+                $osf.growl('Error', 'Could not remove tag');
```
brianjgeiger: I'm not sure if this is better. Currently, if two people have the same project page open, and one person removes a tag, followed by another person removing the same tag on their browser, they get an error. Technically, no, the second person couldn't remove the tag, but that's just because the tag was gone. From the user's standpoint, the tag is gone, so everything is good.

brianjgeiger:
Overall, the code looks good with perhaps the one use case up for debate. I haven't been able to delete a tag using the delete key, so I don't know how to replicate the original issue, but this certainly looks like it should fix it. A JS test or three wouldn't go amiss. 🐧

brianjgeiger:
As discussed, I think this needs to either become simpler or more ambitious. We can either catch the exact use case described in the docs, were you can't try to delete a tag that's deleted, or we can also ensure that the error returned from the server only raises up to the UI level if the requested change will not result in a state that the user expects (i.e. deleting a non-existent tag should just have the tag go away silently, not complain that, yes, the tag is gone, but you were not the one who deleted it).

abought:
Well... maintenance coding isn't about taking shortcuts, and the best solution is the one that helps us carry forward any lessons learned. Do you mind if I run an APIv2 scenario by you while it's still fresh on both our minds? (I'm having a smidge of truble testing locally as the API is sassing me about missing fields, even on a patch request.)
https://github.com/abought/osf.io/blob/f34edce523e12156d4b31f81935efccc3f2ed62d/api/nodes/serializers.py#L251-L251

I notice that it gets the list of "old tags" from the node, and "current tags" from the user request. From an initial readthrough, this seems like APIv2 might have a bug that causes it to resurrect already-deleted tags. 

The central issue is that the APIv2 tags field relies on bulk updates of individual data, and bulk updates can include stale information.

Scenario:
- Users A and B load a node dashboard with two tags: `redshirt` and `browncoat`.
- User A deletes tag `redshirt` by sending a request to the server, with `current_tags = ["browncoat"]`
- According to the server, the node has only one tag, so for the next request, `node.tags = ["browncoat"]`
- Several minutes later, User B sends a delete request to the server, deleting `browncoat`. The payload they send is constructed based on the data available to the front-end at the time (they still think that `tags = ["redshirt", "browncoat"]`)
    - According to that logic, the payload they send says that after deletion, there is one tag left: `current_tags = ["redshirt"]`
    - The server compares `node.tags` to `current_tags`. It adds one tag present on the request but not the node (redshirt) and removes one tag present on the node but not in the request (browncoat)
    - According to the server, the node now has only one tag: the previously deleted field `redshirt` has been restored due to stale request data.

brianjgeiger:
Yes, that is what will happen and is not unexpected. Doing a fresh request for node info would be wise in these circumstances.

abought:
> Yes, that is what will happen and is not unexpected. Doing a fresh request for node info would be wise in these circumstances.

That seems... problematic. At best, the strategy would reduce the time window for a collision to occur- it wouldn't really eliminate the issue of stale data entirely. APIv1 treats these as atomic add/remove options, which seems a bit more robust. 

For comparison, other bulk node fields (like contributors) are updated in separate requests.

abought:
EDIT to above: that's outside the scope of this PR. But to ensure correct capture of future requirements, I'll go ahead and extend this PR as follows:

- 409 for remove operations if the user tries to remove a tag that is already gone
- 400 for other forms of bad request (like deleting a blank tag)
- UI suppresses warning if it receives a 409 on deletion.

brianjgeiger:
The problem is the data structure for tags is not one that supports a nice interface with a REST API. We could turn them into a collection and give them IDs which would allow us to treat them individually, but honestly, they're tags. I'm not super-concerned about someone having to add/remove a tag.

abought:
Test failure in varnish doesn't appear related; can someone please restart the test?

website/files/models/osfstorage.py
```diff
@@ -238,14 +239,20 @@ def add_tag(self, tag, auth, save=True, log=True):
     def remove_tag(self, tag, auth, save=True, log=True):
         from website.models import Tag, NodeLog  # Prevent import error
         tag = Tag.load(tag)
-        if tag and tag in self.tags and not self.node.is_registration:
+        if not tag:
+            raise InvalidTagError
+        elif tag not in self.tags:
+            raise TagNotFoundError
+        elif self.node.is_registration:
```
mfraezz: Should you be checking `is_registration` first? It seems like that error would take priority over tag validity.

abought: That's a good point, and this change would make the behavior of the project and file endpoints slightly more consistent.

Commit pushed.

brianjgeiger:
Looks good; functions quite nicely. I've restarted tests. Once they pass, this is good to merge. 🎉 🐧 🎉

abought:
As Matt and I discussed, we've been seeing a lot of varnish failures lately in CI. Pinging @cwisecarver for insights into what we can do to get that part passing.

cwisecarver:
tl;dr: our PCRE library is coming up missing.

@abought @mfraezz It looks like from the error messages that varnish isn't starting up properly on Travis. 
```
[0KError:
Message from VCC-compiler:
/tmp/varnish/usr/sbin/varnishd: symbol lookup error: /tmp/varnish/usr/lib/varnish/libvarnish.so: undefined symbol: pcre_free_study
Running VCC-compiler failed, exited with 127
VCL compilation failed
```

Looks like it's being caused by:
```
dpkg-deb: error: `libpcre3_8.31-2ubuntu2.1_amd64.deb' is not a debian format archive
```

Which is in fact true, because it's a [404 error page](http://security.ubuntu.com/ubuntu/pool/main/p/pcre3/libpcre3_8.31-2ubuntu2.1_amd64.deb).

I'll work with @icereval when he gets in to fix the travis config unless @mattclark is bored and knows how to wrangle about travis.

abought:
Thanks, @cwisecarver . Could you ping me when that change lands? It seems that this PR is blocked mainly on varnish failures... I'm not sure how to move this forward short of endlessly hitting "rerun" on travis until I get lucky.

PR #5544, opened 2016-04-27
[RFR] Fix ESI content negotiation for Applications and Tokens API segments

cwisecarver:
## Purpose

[With ESI enabled Application and Token API segments would return `""` for each application or token.](https://www.flowdock.com/app/cos/osf-dev/threads/fRTR6af6NTCnxThJApnh0CAm87i) This fixes that problem.

## Changes

In order to disable the browseable API for these endpoints the list of renderers was specified on the view. The renderer specified didn't process the response, removing quotes from around ESI tags. I changed the renderers list to the two renderers that will remove those quotes.

## Side effects

I functionally tested the endpoints for applications and tokens using both the `?format=json` and the content-type request header. They both returned correct responses and the endpoints did not appear in the browseable API locally. My local setup runs through varnish. It's possible that something is amiss @abought and @samchrisinger should probably take a look at it once it gets on staging to make sure it's secure.

## Ticket

Flowdock link of the [conversation](https://www.flowdock.com/app/cos/osf-dev/threads/fRTR6af6NTCnxThJApnh0CAm87i) regarding this.

### DIFF
--- a/api/applications/views.py
+++ b/api/applications/views.py
@@ -6,6 +6,7 @@ from rest_framework import generics
 from rest_framework import renderers
 from rest_framework import permissions as drf_permissions
 
+from api.base.renderers import JSONAPIRenderer, JSONRendererWithESISupport
 from modularodm import Q
 
 from framework.auth import cas
@@ -50,7 +51,7 @@ class ApplicationList(JSONAPIBaseView, generics.ListCreateAPIView, ODMFilterMixi
     view_category = 'applications'
     view_name = 'application-list'
 
-    renderer_classes = [renderers.JSONRenderer]  # Hide from web-browsable API tool
+    renderer_classes = [JSONRendererWithESISupport, JSONAPIRenderer,]  # Hide from web-browsable API tool
 
     def get_default_odm_query(self):
 
@@ -90,7 +91,7 @@ class ApplicationDetail(JSONAPIBaseView, generics.RetrieveUpdateDestroyAPIView, 
     view_category = 'applications'
     view_name = 'application-detail'
 
-    renderer_classes = [renderers.JSONRenderer]  # Hide from web-browsable API tool
+    renderer_classes = [JSONRendererWithESISupport, JSONAPIRenderer,]  # Hide from web-browsable API tool
 
     def get_object(self):
         return self.get_app()
@@ -128,7 +129,7 @@ class ApplicationReset(JSONAPIBaseView, generics.CreateAPIView, ApplicationMixin
 
     serializer_class = ApiOAuth2ApplicationResetSerializer
 
-    renderer_classes = [renderers.JSONRenderer]  # Hide from web-browsable API tool
+    renderer_classes = [JSONRendererWithESISupport, JSONAPIRenderer, ]  # Hide from web-browsable API tool
 
     view_category = 'applications'
     view_name = 'application-reset'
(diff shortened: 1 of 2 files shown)


api/applications/views.py
```diff
@@ -50,7 +50,7 @@ class ApplicationList(JSONAPIBaseView, generics.ListCreateAPIView, ODMFilterMixi
     view_category = 'applications'
     view_name = 'application-list'
 
-    renderer_classes = [renderers.JSONRenderer]  # Hide from web-browsable API tool
+    renderer_classes = [JSONRendererWithESISupport, JSONAPIRenderer, ]  # Hide from web-browsable API tool
```
samchrisinger: AFIK it isn't actually possible to hide stuff from the API docs. Maybe we can simplify this?

@brianjgeiger ?

abought: This doesn't hide the endpoint (because at minimum it has to respond to JSON requests)... it just makes it less user-friendly.

If we move to swagger, I agree that will likely ignore any custom rendering settings we have in place, in which case the value of custom behavior for this class goes way down.

samchrisinger: In that case, I think we may be able to safely remove `renderer_classes` from each of these.

abought:
One testing suggestion (until I can find my notes for how to set up varnish locally...)

The tokens endpoint has very specialized behavior- it is intended to hide the token value after the first visit. (and never serialize it again)

We should make sure that varnish doesn't inadvertently cache and return more information than we expect to be present, for subsequent requests to the token endpoint.

api/applications/views.py
cwisecarver: I'm happy to remove them. I was just trying to keep existing functionality. 👉 👃

cwisecarver:
@abought The VCL (which is in the private-edge-configs repo) is "pass"ing anything based in the tokens namespace. It acts as nothing more than a proxy.

```
    # it's tokens, pass
    if (req.url ~ "^/v2/tokens/") {
        std.log("DEBUGGING: PASS -- TOKENS");
        return (pass);
    }
```

api/applications/views.py
cwisecarver: @abought @samchrisinger @brianjgeiger @pattisdr Decision on removing the renderer_classes which would allow token and applications to appear in DRF browseable API. The (current implementation) alternative is that these urls will work but do not return HTML, just JSON.

abought: Our real goal was to keep this an undocumented (private) API not intended for external use- hence the visual difference. But swagger has a different mechanism for excluding things.

I'm ok with removing this if we switch to swagger, as the current mechanism of hiding things is pretty weak anyway. (and creates varnish maintenance headaches) Usage will still be controlled separately, by permissions classes.

Can you please add a TODO or somesuch, so that we remember to hide the endpoint in a future doc revamp?

cwisecarver: 👍  I'll leave the renderer_classes in here, which would have been here regardless of varnish, and throw in a todo for future documentarians.

cwisecarver:
@abought Added todos.

cwisecarver:
@sloria and @icereval This is required for varnish on prod. It'd be nice if we could get this merged so QA could start testing it Monday or even later today on staging. 

If it's a problem, let me know and I can turn it into a hotfix on top of 0.68.0. I thought it was going to be in 0.69.0 so I based it off develop. It should merge pretty cleanly regardless.

@icereval I undid the yapf formatting ;)

samchrisinger:
:shipit:

samchrisinger:
@cwisecarver @sloria this is becoming a bit of a blocker for me. Travis failures seem unrelated. Any reason not to merge?

PR #5545, opened 2016-04-27
[fix] [RfR] Limit five files per question on Draft upload

barbour-em:
## Purpose

Limit five files to be attached to a file upload question.

## Changes
- Check if user is trying to add a sixth file and give a pop up warning.
- Add description to question telling the user they can only upload 5 files.

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-5587

### DIFF
--- a/website/static/js/registrationEditorExtensions.js
+++ b/website/static/js/registrationEditorExtensions.js
@@ -177,7 +177,16 @@ var Uploader = function(question) {
         question.value(question.formattedFileList());
     });
 
+    self.UPLOAD_LANGUAGE = 'You may attach up to 5 files to this question. You may attach files that you already have' +
+        'in this OSF project, or upload a new file from your computer. Uploaded files will automatically be added to this project' +
+        'so that they can be registered.';
+
     self.addFile = function(file) {
+        if(self.selectedFiles().length === 5) {
+            bootbox.alert('Too many files. Cannot attach more than 5 files to a question.');
+            return false;
+        }
+
         if(self.fileAlreadySelected(file))
             return false;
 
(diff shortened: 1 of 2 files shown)


website/static/js/registrationEditorExtensions.js
```diff
@@ -177,7 +177,16 @@ var Uploader = function(question) {
         question.value(question.formattedFileList());
     });
 
+    self.UPLOAD_LANGUAGE = 'You may attach up to 5 files to this question. You may attach files that you already have' +
+        'in this OSF project, or upload a new file from your computer. Uploaded files will automatically be added to this project' +
+        'so that they can be registered.';
```
leb2dg: There are some missing spaces between each of these string concatenations:
screen shot 2016-04-27 at 12 27 16 pm

website/static/js/registrationEditorExtensions.js
```diff
@@ -177,7 +177,16 @@ var Uploader = function(question) {
...
+
     self.addFile = function(file) {
+        if(self.selectedFiles().length === 5) {
```
leb2dg: The `length === 5` logic should probably be modified because it causes multiple error messages if more than 5 files are selected for upload:

uploadfiles

leb2dg:
@barbour-em I think this should be ready to go once the spacing and multiple error messages are fixed :)

barbour-em:
@leb2dg Comments addressed. Spacing is fixed. As for multiple warnings, each question will warn the user once about the file limit via the modal, then will ignore attempts after that.

sloria:
Cherry-picked into release/0.68.0 🐙

PR #5548, opened 2016-04-27
Feature/fix displaying non-bibliographic contributors on "My Projects" and "Popular Public Projects" [OSF-6032]

samanehsan:
## Purpose

Non-bibliographic contributors are listed when viewing projects listed under "Popular Public Projects" on the dashboard page and when viewing a project that does not belong to the current user (e.g. a bookmarked public project) on the my projects page.

## Changes

- Only list bibliographic contributors for projects displayed under "Popular Public Projects"
- On the "My Projects" page, show all contributors if the current user is a contributor, otherwise show only  bibliographic contributors.
- Only list non-bibliographic users under the "Contributors" filter section if the current user has a project with that user. 

## Side effects

None.

## Ticket

https://openscience.atlassian.net/browse/OSF-6032


brianjgeiger:
For the second bullet point in the list of changes, that's "show all contributors if the current user is a contributor, otherwise show only bibliographic contributors", yes?

samanehsan:
Yes! I'll update the description.

brianjgeiger:
Also, I know that this is what was requested by QA and Caner, but could you verify this list of rules with Community? It's a little… complicated to know what to expect when you're looking at any given section.

website/static/js/myProjects.js
```diff
@@ -664,13 +664,17 @@ var MyProjects = {
...
                 var contributors = lodashGet(item, 'embeds.contributors.data', []);
-
+                var isContributor = $osf.userIsContributor(window.contextVars.currentUser.id, contributors);
                 if (contributors) {
                     for(var i = 0; i < contributors.length; i++) {
                         var u = contributors[i];
                         if ((u.id === window.contextVars.currentUser.id) && !(self.institutionId)) {
                           continue;
                         }
+                        if (!isContributor && !u.attributes.bibliographic) {
```
brianjgeiger: Ah, I see why you wrote it that way in the description. :)

website/static/js/osfHelpers.js
```diff
@@ -838,6 +838,17 @@ var findContribName = function (userAttributes) {
     }
 };
 
+// Check if a user is in the list of project contributors returned by the v2 API.
+function userIsContributor(userId, contributors) {
```
brianjgeiger: Yay, new js helper function! Could you… add tests for it?

api/base/pagination.py
```diff
@@ -152,3 +152,17 @@ def get_paginated_response(self, data):
...
+
+    def get_paginated_response(self, data):
+        """ Add number of bibliographic contributors to links.meta"""
+        response = super(NodeContributorPagination, self).get_paginated_response(data)
+        response_dict = response.data
+        kwargs = self.request.parser_context['kwargs'].copy()
+        node_id = kwargs.get('node_id', None)
+        node = Node.load(node_id)
+        total_bibliographic = len(node.visible_contributor_ids)
+        response_dict['links']['meta']['total_bibliographic'] = total_bibliographic
```
brianjgeiger: Could you add a test or two for this as well.

brianjgeiger:
Nothing terrible in the code. Could use some tests (definitely the ones called out, and if there are any more you can add for the project organizer, it would be good; that's under-tested as it is, if it's tested at all). 

Also, I presume you have done a complete set of manual tests, yes?

samanehsan:
Done adding those tests @brianjgeiger! ☔

brianjgeiger:
Cool. I think resolve the conflicts and this should be ready to merge. 🎉 🐧 🎉

website/static/js/osfHelpers.js
```diff
@@ -839,6 +839,17 @@ var findContribName = function (userAttributes) {
...
+    var isContributor = false;
+    for (var i = 0; i < contributors.length; i++) {
+        if (contributors[i].id === userId) {
+            isContributor = true;
```
sloria: Once you get here, you should be able to `return true`.

website/static/js/project-organizer.js
```diff
@@ -61,6 +62,18 @@ function _poContributors(item) {
         return '';
     }
     var totalContributors = item.data.embeds.contributors.links.meta.total;
+    var isContributor = $osf.userIsContributor(window.contextVars.currentUser.id, contributorList);
+    if (!isContributor) {
+        // bibliographic contributors
+        var visibleContributors = [];
+        for (var i=0; i < contributorList.length; i++) {
+            if (contributorList[i].attributes.bibliographic) {
```
sloria: You should be able to use `Array.prototype.filter` for this.

website/static/js/osfHelpers.js
sloria: There's nothing really specific about contributors about this function; I wonder if it can be generalized.

How about implementing a simplified version of [_.find](https://lodash.com/docs#find)? 

> Iterates over elements of collection, returning the first element predicate returns truthy for.

I also wouldn't be opposed to adding `lodash.find` as a dependency.

sloria:
Pass finished.

website/static/js/osfHelpers.js
samanehsan: @sloria, can I use `lodash.find` instead of implementing a similar `_.find` method?

sloria: Yep, I actually edited my comment a few seconds after I posted to suggest just that.

samanehsan: ok, yay! I can get rid of some code!

samanehsan:
Done making those changes @sloria! ⛅

PR #5557, opened 2016-04-28
[feature] fix typos and add node authorization and deauthorization actions [OSF-5371]

leb2dg:
## Purpose

Fix typos and add node authorization and deauthorization actions to API docs. Related to the `v2/nodes/<node_id>/logs` endpoint.

## Changes

* biliography --> bibliography
* permissions_update --> permissions_updated
* added `node_authorized` and `node_deauthorized` to ticket

Note: `file_added`, `file_updated`, etc. were not changed to `addon_file_added` because the logs say the name of the addon that was added. So, if the user filters off of `addon_file_added` no results would appear because the actual logs say something like `github_file_added`.

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-5371


samanehsan:
Looks good to me @leb2dg! 🐋

PR #5513, opened 2016-04-21
Feature/new logs audit [OSF-5945]

pattisdr:
Issue https://openscience.atlassian.net/browse/OSF-5945

Moved from https://github.com/CenterForOpenScience/osf.io/pull/5394

# Purpose

Audits new logs on My Projects page.  For the most part, tries to match current logs.  As a general rule, files will provide links whether or not the resource has been deleted. (If it has been deleted, the link will return a Gone or Not Found).  Logs referencing nodes, however, will not provide links if the resource has been deleted.   

# Changes

Numerous small fixes to logs that appear in My Projects page - no change to project organizer logs.  Log phrasing changed, links added, etc.  For full list of changes see google doc link in Issue https://openscience.atlassian.net/browse/OSF-5945

PR #5558, opened 2016-04-28
[Feature] Fix embargo logs [OSF-6005]

samanehsan:
## Purpose

Logs about registration actions should link to the registered project. Eg. Instead of "Embargo for {Registration} approved, the log should say "Embargo of registration of {Project} approved."

## Changes

- Change the "embargo_approved_no_user" log to "Embargo of registration of {Project} approved" and pass in the registered project instead of the registration in the `embargo_registrations.py` script.
- Change the "embargo_completed_no_user" log to "Embargo of registration of {Project} completed"
(the registered project is already being passed to the log).
- Remove the unused "embargo_completed" log, since embargo completion is currently done automatically

## Side effects
None.

## Ticket
https://openscience.atlassian.net/browse/OSF-6005


scripts/embargo_registrations.py
```diff
@@ -46,7 +46,8 @@ def main(dry_run=True):
                         parent_registration.registered_from.add_log(
                             action=NodeLog.EMBARGO_APPROVED,
                             params={
-                                'node': parent_registration._id,
+                                'node': parent_registration.registered_from_id,
+                                'registration': parent_registration._id,
```
mfraezz: Is this used by the logs?

mfraezz:
1. For consistency, we might also want to change `embargo_approved` language to `<User> approved embargo of registration of <Node>` instead of `[...] embargoed registration of [...]`
* Should this fix existing logs of this type that were created incorrectly? If so, please add a migration script / tests for it.
* Please add a regression test that ensures the log links to the correct node.

Pass complete. 🀄

scripts/embargo_registrations.py
samanehsan: I added it in because the other function that creates an `EMBARGO_APPROVED` log has it, but looking at the log template itself I don't think it's used. I can remove it

samanehsan:
Should I also change `embargo_initiated` and `embargo cancelled`?

scripts/embargo_registrations.py
mfraezz: I'm not certain that it needs to be removed -- it could be useful to store information about which registration this log refers to, even if it's not displayed currently.

mfraezz:
Maybe? @saradbowman might have opinions about language

pattisdr:
Remember that there are logs in two places now - the My Projects page has them except they use the API v2 to pull logs and have slightly different log text.

If log_templates.mako is being updated, https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/logActionsList.json needs to be updated.  Eventually we will move to using these new logs everywhere.  If something looks off in `logActionsList.json` I also have a log audit PR that is going in soon https://github.com/CenterForOpenScience/osf.io/pull/5513.

samanehsan:
Just made that change @pattisdr! Thanks for letting me know!

mfraezz:
Pass complete, changes look good. Migration failed for me with `AttributeError: Node object has no attribute registered_from_id`, but that could be because this staging dump is outdated.

scripts/fix_embargo_approved_logs.py
```diff
@@ -0,0 +1,47 @@
...
+def main():
+    targets = get_targets()
+    fix_embargo_approved_logs(targets)
+
+
+def get_targets():
+    return NodeLog.find(Q('action', 'eq', NodeLog.EMBARGO_APPROVED) & Q('params.user', 'eq', None))
+
+
+def fix_embargo_approved_logs(targets):
```
sloria: This function could use some logging--at the very least, log what records were touched and a count of migrated logs.

website/static/js/logActionsList.json
```diff
@@ -45,10 +45,9 @@
     "comment_updated":  "${user} updated a comment ${comment_file} on ${node}",
     "embargo_initiated":    "${user} initiated an embargoed registration of ${node}",
     "embargo_approved":     "${user} approved embargoed registration of ${node}",
-    "embargo_approved_no_user": "Embargo for ${node} approved",
+    "embargo_approved_no_user": "Embargo of registration of ${node} approved",
```
sloria: @saradbowman Is this desired language? The log will display like

> Embargo of registration of Project Explicit approved

scripts/fix_embargo_approved_logs.py
sloria: Ah, I didn't catch the logging.debug statement a few lines down. You should use `logger.info` instead so that the logs are written to a file (which is what add_file_logger does).

scripts/fix_embargo_approved_logs.py
```diff
@@ -0,0 +1,47 @@
...
+    for log in targets:
+        node_id = log.params['node']
+        node = Node.load(node_id)
+        if node.is_registration:
+            log.params['node'] = node.registered_from_id
+            log.params['registration'] = node._id
+            log.save()
+            logging.debug('Updated node param of log {}'.format(log._id))
```
sloria: As mentioned in my previous comment, use `logger.info`. Also, it would be useful to log out what params was before and after the migration.

Also, prefer to log *before* making a db change (e.g. "Updating param of log {} ..."). That way, if the script crashes on the `.save()` we can diagnose which record failed.

sloria:
Pass finished. Only minor changes to make.

samanehsan:
Done making those changes @sloria! 🐫

scripts/fix_embargo_approved_logs.py
```diff
@@ -26,14 +26,18 @@ def get_targets():
...
 def fix_embargo_approved_logs(targets):
+    count = 0
     for log in targets:
         node_id = log.params['node']
         node = Node.load(node_id)
         if node.is_registration:
+            log_params = log.params
             log.params['node'] = node.registered_from_id
             log.params['registration'] = node._id
+            logger.info('Updating params of log {} from {} to {}'.format(log._id, log_params, log.params))
```
sloria: Since log_params just points to `log.params`, they will be logged as identical. You'd need to copy the original params to fix this. I can make this change when I merge.

scripts/fix_embargo_approved_logs.py
```diff
@@ -26,14 +26,18 @@ def get_targets():
...
             log.save()
-            logging.debug('Updated node param of log {}'.format(log._id))
+            count += 1
+    logger.info('{} logs migrated').format(count)
```
sloria: Typo here: `.format` needs to be called on the string; will fix when I merge.

samanehsan: oh nooo! good catch!

PR #5559, opened 2016-04-28
Make log action happen when prereg registration is initiated

hmoco:


website/templates/log_templates.mako
```diff
@@ -204,15 +204,13 @@ to
...
-        It was submitted for review to the Preregistration Challenge on
-    <!-- /ko -->
-    <!-- ko ifnot: params.is_prereg -->
         registered
         <a class="log-node-title-link overflow" data-bind="attr: {href: nodeUrl}, text: nodeTitle"></a>
-    <!-- /ko -->
+</script>
+
+<script type="text/html" id="prereg_registration_initiated">
+    submitted for review to the Preregistration Challenge a registration of
```
sloria: Grammar fix: should be 

"submitted <name> for review to the Preregistration Challenge."

website/project/views/drafts.py
```diff
@@ -130,6 +133,16 @@ def submit_draft_for_review(auth, node, draft, *args, **kwargs):
         save=True
     )
 
+    if prereg_utils.get_prereg_schema() == draft.registration_schema:
+
+        node.add_log(
```
sloria: Is `node` the registration or the original node here?

PR #5561, opened 2016-04-28
Fix KO click events broken by KO punches removal

abought:
## Purpose
Fix various buttons that stopped working when KO punches was removed.

## Changes
Bind the correct context to various buttons. For list of buttons affected and problem details, see [#OSF-6132]. Incorporates previous reg fixes by Eric.

This fixes several broken buttons so that they no longer give JS console errors. The results of clicking should now be consistent with the intended behavior.

## Side effects
This was a limited audit for one common failure case; there may be other broken buttons.

## Ticket
https://openscience.atlassian.net/browse/OSF-6132

### DIFF
--- a/website/templates/include/profile/jobs.mako
+++ b/website/templates/include/profile/jobs.mako
@@ -22,7 +22,7 @@
                         </span>
                         <a
                                 class="text-danger pull-right"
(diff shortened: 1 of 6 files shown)


abought:
Goodness. How very prompt.

PR #5563, opened 2016-04-28
Update COS inst and allow redirect_url to persist between login screens

hmoco:


scripts/populate_institutions.py
```diff
@@ -83,7 +83,7 @@ def main(env):
             {
                 '_id': 'usc',
                 'name': 'University of Southern California',
-                'description': 'Projects must abide by <a href="http://policy.usc.edu/info-security/">USC\'s Information Security Policy</a>. Data stored for human subject research repositories must abide by <a href="http://policy.usc.edu/biorepositories/">USC\'s Biorepository Policy</a>. The OSF may not be used for storage of Personal Health Information that is subject to <a href="http://policy.usc.edu/hipaa/">HIPPA regulations</a>.',
+                'description': '<h5>Projects must abide by <a href="http://policy.usc.edu/info-security/">USC\'s Information Security Policy</a>. Data stored for human subject research repositories must abide by <a href="http://policy.usc.edu/biorepositories/">USC\'s Biorepository Policy</a>. The OSF may not be used for storage of Personal Health Information that is subject to <a href="http://policy.usc.edu/hipaa/">HIPPA regulations</a>.</h5>',
```
sloria: Will all descriptions be an `h5`? If so, then that should probably go in the template.

website/templates/public/pages/active_nodes.mako
```diff
@@ -27,7 +27,7 @@
             <section id='newPublicProjects'>
                 <h3 class='anchor'>Newest public projects</h3>
                 <div class='project-list'>
-                    ${node_list(recent_public_projects, prefix='newest_public', metric='date_created')}
+                    ${node_list(recent_apublic_projects, prefix='newest_public', metric='date_created')}
```
sloria: Typo here

sloria:
Pass finished.

scripts/populate_institutions.py
hmoco: for now those make sense as h5, however, future ones might have more/less text. As suggested by Michael, since we are allowing html, it would make sense to decouple formatting the description   from the template.

website/templates/public/pages/active_nodes.mako
hmoco: oops good catch

hmoco:
@sloria back to you

scripts/populate_institutions.py
icereval: don't include any formatting just html related to links, i've already patched in text-muted text-small into the mako template, this text is a disclaimer (description is maybe a bad name but we can fix that later) and should be displayed accordingly see osf.usc.edu for an example

scripts/populate_institutions.py
```diff
@@ -53,7 +53,7 @@ def main(env):
             {
                 '_id': 'cos',
                 'name': 'Center For Open Science',
-                'description': None,
+                'description': '<h5><i> COS is a non-profit technology company providing free and open services to increase inclusivity and transparency of research. COS supports shifting incentives and practices to align more closely with scientific values. Find out more at <a href="cos.io">cos.io</a>.</i></h5>',
```
icereval: same as below

scripts/populate_institutions.py
hmoco: ok

scripts/populate_institutions.py
icereval: no italics plz

PR #5555, opened 2016-04-27
[feature] Update Getting Started Links to Guides Page

caseyrollins:
- Also adds redirects:
 - `/getting-started/` -> `help.osf.io`
 - `/getting-started/email/` -> `help.osf.io`
 - `/howosfworks/` -> `help.osf.io`

PR #5568, opened 2016-04-28
Redirecting /help/ page to /faq/ [#OSF-4039]

dsb9ef:
## Purpose

The OSF has a /help/ page, but it is only an unstyled placeholder.  This change redirects /help/ to /faq/.  

## Changes

- Added a rule in website/routes.py to redirect /help/ to faq

- Added a test case in test_views.py that checks if the page was redirected

 

## Ticket
https://openscience.atlassian.net/browse/OSF-4039

### DIFF
--- a/website/routes.py
+++ b/website/routes.py
@@ -262,12 +262,6 @@ def make_url_map(app):
             json_renderer
         ),
         Rule(
-            '/help/',
-            'get',
-            website_views.redirect_help_to_faq,
-            json_renderer
-        ),
-        Rule(
             '/support/',
             'get',
             {},
--- a/website/routes.py
+++ b/website/routes.py
@@ -262,6 +262,12 @@ def make_url_map(app):
             json_renderer
         ),
         Rule(
+            '/help/',
+            'get',
+            website_views.redirect_help_to_faq,
+            json_renderer
+        ),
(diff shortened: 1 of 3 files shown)


website/routes.py
```diff
@@ -262,6 +262,12 @@ def make_url_map(app):
             json_renderer
         ),
         Rule(
+            '/help/',
+            'get',
+            website_views.redirect_help_to_faq,
```
mfraezz: I'm not certain this is the right approach. It seems like it would be simpler to make the `/faq/` `Rule` also include `/help/`, and save the user a request. In any case, [the other `/help/` route](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/routes.py#L280) needs to be disabled.

sloria: I agree with @mfraezz 's suggestion. See this recent, relevant change: https://github.com/CenterForOpenScience/osf.io/pull/5555/files#diff-758f58f82207e3674b3105d2b3d4ce7aR215 . You should just be able to add `/help/` redirect to the Guides pages. @dsb9ef Can you please send a new PR branched off of https://github.com/CenterForOpenScience/osf.io/tree/caseyrollins-feature/getting-started-to-help?

PR #5571, opened 2016-04-28
Revert veer

sloria:
## Purpose

Reverts Veer schema PRs because they depend on #5280 , which is getting pulled from 0.68.0.

NOTE: this include #5570


sloria:
doh; wrong branch

PR #5578, opened 2016-04-29
[Hotfix] Encode wiki page_name when creating absolute url [OSF-6141]

samanehsan:
## Purpose

If a wiki has a page name with spaces in it (e.g. "Test Wiki"), it's absolute url is `http://osf.io/<project_id/wiki/Test Wiki`, which redirects to `http://osf.io/<project_id/wiki/Test` because of the space.

## Changes

%encode the wiki page name so that spaces or special characters do not break the url.

## Side effects

None.

## Ticket

https://openscience.atlassian.net/browse/OSF-6141

### DIFF
--- a/website/addons/wiki/model.py
+++ b/website/addons/wiki/model.py
@@ -3,6 +3,7 @@
 import datetime
 import functools
 import logging
+import urllib
 
 from bleach import linkify
 from bleach.callbacks import nofollow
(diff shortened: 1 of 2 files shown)


website/addons/wiki/model.py
```diff
@@ -211,7 +212,7 @@ def get_extra_log_params(self, comment):
 
     # used by django and DRF - use v1 url since there are no v2 wiki routes
     def get_absolute_url(self):
-        return '{}wiki/{}/'.format(self.node.absolute_url, self.page_name)
+        return '{}wiki/{}/'.format(self.node.absolute_url, urllib.quote(self.page_name))
```
sloria: Wouldn't you need to use `quote_plus`, to properly encode spaces?

samanehsan: But wouldn't `quote_plus` convert spaces to `+` instead of `%20`?

sloria: Discussed on FD: this is correct as is 👍 .

sloria:
Good to merge when the build passes

PR #5581, opened 2016-04-29
Throw API error if addon is not configured [OSF-4947]

samanehsan:
## Purpose

Going to the API endpoint `/v2/node_id/files/<provider>/` for an addon that is not configured on the project throws a 503 service unavailable error.

## Changes

Throw a 404 NotFound error.

## Side effects

None.

## Ticket
https://openscience.atlassian.net/browse/OSF-4947

### DIFF
--- a/api/nodes/utils.py
+++ b/api/nodes/utils.py
@@ -25,7 +25,7 @@ def get_file_object(node, path, provider, request):
             )
         return obj
 
-    if not node.has_addon(provider):
+    if not node.get_addon(provider) or not node.get_addon(provider).configured:
         raise NotFound('The {} provider is not configured for this project.'.format(provider))
 
     url = waterbutler_api_url_for(node._id, provider, path, meta=True)
--- a/api/nodes/utils.py
+++ b/api/nodes/utils.py
@@ -25,6 +25,9 @@ def get_file_object(node, path, provider, request):
             )
         return obj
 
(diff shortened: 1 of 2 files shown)


samanehsan:
The failing tests are also failing on develop, so this is ready for CR!

api/nodes/utils.py
```diff
@@ -25,6 +25,9 @@ def get_file_object(node, path, provider, request):
             )
         return obj
 
+    if not node.has_addon(provider):
```
mfraezz: This doesn't work if the addon is authorized but not configured.

screen shot 2016-05-02 at 09 55 41
screen shot 2016-05-02 at 09 55 54


You can check that with `node.get_addon(provider).configured`

samanehsan:
Done making that change @mfraezz!

api/nodes/utils.py
```diff
@@ -25,6 +25,9 @@ def get_file_object(node, path, provider, request):
             )
         return obj
 
+    if not node.get_addon(provider).configured:
```
mfraezz: This can raise an `AttributeError` if `get_addon` returns `None`

samanehsan: Yeah, I just realized that when my test failed :( This is fixed now!

mfraezz:
LGTM. Dropping this in RTM

PR #5582, opened 2016-04-29
Make project organizer buttons visible on small screens [OSF-5322]

billyhunt:
## Changes

The buttons in the project organizer go off screen on small displays, because the minimum size of the treebeard panel was set to 400px.  Now there is no minimum.

Before
---
screen shot 2016-04-29 at 2 34 02 pm

After
---
screen shot 2016-04-29 at 2 34 29 pm

## Side effects

The project window can now be smaller than 400px.  This should probably be checked in all browsers to make sure it doesn't get all wiggity wack.  

## Ticket

[#osf-5322]


brianjgeiger:
Looks good. :tada: :penguin: :tada:

PR #5583, opened 2016-04-29
Hotfix/fix banning of wiki comments

cwisecarver:
## Purpose

This fixes wiki comment decaching.

@sloria @icereval This is a replacement for #5554 that shouldn't cause any tests to fail.

## Changes

An unhandled AttributeError was being caused during the decaching process of NodeWikiPages. This caused none of the objects modified during the request to be decached, including the comments URL. This checks the original object for an absolute_api_v2_url. If that does not exist then it returns an empty list of urls to ban. There are also try/except clauses around checking target.referents and root_target.referents. They check for an AttributeError and pass, continuing with the other potential things to ban.

## Side effects

If there are many things without absolute_api_v2_urls there could be a lot of logs in sentry. These would have been causing attribute errors in sentry previous to this PR so ... unlikely.

### DIFF
--- a/api/caching/tasks.py
+++ b/api/caching/tasks.py
@@ -42,7 +42,6 @@ def get_bannable_urls(instance):
                                                                   path=parsed_target_url.path)
                 bannable_urls.append(url_string)
 
-
             try:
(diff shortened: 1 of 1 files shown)


sloria:
Merged as a hotfix into master, develop, and release/0.68.0

PR #5585, opened 2016-04-29
Feature/Stricter Password Requirements [#OSF-3457]

erinspace:
## Purpose

When signing up for the OSF or changing your password, there are not very strict password strength requirements at the moment. This PR will add some stricter security requirements, and also some password strength suggestions, using Dropbox's zxcvbn library https://github.com/dropbox/zxcvbn

## Changes

- Add zxcvbn library to requirements
- Add password strength bar to front page, create account login page, forgot password page and reset password page
    - relies on the strength assigned 0-4 by zxcvbn
- CSS classes for password width added to OSF helpers based on those strength numbers
- only allows passwords with strength 2-4 as calculated by zxcvbn (yellow indicator)
- On new accounts and password changes:
    - password cannot be the same as the email address
- Redo reset password form to use KO validation over WTForms
- After password reset, user is redirected to login instead of logging in automatically

With the password "bbbbbb" --
screen shot 2016-05-23 at 1 34 54 pm

With the password "password" -- 
screen shot 2016-05-23 at 1 42 49 pm

With the password "horsesrule" --
screen shot 2016-05-23 at 1 35 08 pm

With the password asdfghjkl;'
screen shot 2016-05-23 at 1 45 23 pm

With username same as your password:
screen shot 2016-05-23 at 2 04 08 pm

## Side effects
- It'd be awesome if someone could take a look at https://github.com/dropbox/zxcvbn#performance and double check that I haven't loaded the library in a way that will cause performance issues.  For example, I require it in ```koHelpers``` to add reusable custom KO validation - will this be slow?
- Advice on how these strength indicators are placed are much welcome! Are they in the way? Is the feedback placed well? Any quirks?

NOTE: This does NOT deal with password reset in the admin. 

## Ticket
https://openscience.atlassian.net/browse/OSF-3457


framework/auth/views.py
```diff
@@ -48,10 +49,10 @@ def reset_password(auth, **kwargs):
         }
         raise HTTPError(400, data=error_data)
 
-    if request.method == 'POST' and form.validate():
+    if request.method == 'POST':  # TODO - do we need more than KO password validation?
```
mfraezz: > # TODO - do we need more than KO password validation?

Probably not. `minLength` and `maxLength` are being validated in KO, which is all the validation being previously being done with the form.

website/static/js/koHelpers.js
```diff
@@ -195,6 +196,12 @@ ko.validation.rules.mustEqual = {
     message: 'The field does not match the required input.'
 };
 
+ko.validation.rules.complexity = {
```
mfraezz: Is this validation rule used anywhere outside of `setPassword.js`? If not, it might be better to set this there so that `zxcvbn` isn't `require`'d where it isn't needed.

framework/auth/views.py
mfraezz: Though if it's possible to circumvent the front-end and `POST` directly to this endpoint, then yes.

website/static/js/setPassword.js
```diff
@@ -0,0 +1,206 @@
...
+//    - change: change your password when you know your old one
+var ViewModel = function(passwordViewType, submitUrl, campaign) {
+
+    var self = this;
+
+    self.typedPassword = ko.observable('');
+
+    self.passwordInfo = ko.computed(function() {
+        if (self.typedPassword()) {
+            return zxcvbn(self.typedPassword());
```
mfraezz: While testing very long passwords, somewhere around 180 characters my browser stopped responding. the [zxcvbn performance page](https://github.com/dropbox/zxcvbn#performance) suggests sending no more than 100 characters to zxcvbn to minimize runtime latency.

mfraezz:
>When you've forgotten your password, I had trouble checking to see if they didn't set their username as their password, which is a new requirement on the signup page and the change password page.

I think you can modify `framework.auth.views.reset_password` to return the username (email) associated with the user, in addition to just the `verification_key` on a `GET`. This should make it available in the template / validator.

mfraezz:
>Can I more effectively refactor the code anywhere to make things more modular? It feels like I repeated a lot of code but am at a loss how to fix that. 

I'm also not sure if all the extra front-end code can be DRY'd up, at least not without normalizing all the password changing forms. Out of scope for this task, but I think it is ticketed.

mfraezz:
> * Are the strength indicator bars placed in a good spot? Any design changes?
> * Is the password strength feedback helpful/in an OK place?

Placement / design / etc. LGTM. 👍

mfraezz:
In the cases where email *is* available, could you have the bar turn red and give a more descriptive error message when password and email match?

screen shot 2016-05-12 at 14 38 04

mfraezz:
Bug found: On the `resetpassword/<verification_key>/` page, I'm able to bypass validation -- typed `qwerty` and it was accepted. Verified that this was not the case on the landing or sign-up pages.

mfraezz:
Pass complete, back to you @erinspace 👾

framework/auth/core.py
```diff
@@ -658,6 +660,8 @@ def set_password(self, raw_password, notify=True):
         """
         had_existing_password = bool(self.password)
         self.password = generate_password_hash(raw_password)
+        if self.username == raw_password:
+            raise HTTPError(httplib.CONFLICT)
```
mfraezz: Please include a descriptive error message, e.g. `HTTPError(httplib.CONFLICT, message='You cannot set your password to your email')`

framework/auth/views.py
```diff
@@ -36,10 +35,12 @@
 
 @collect_auth
 def reset_password(auth, **kwargs):
+    """ If get request, show reset password page. If POST, attempt to reset password if
```
mfraezz: Part of this docstring should go to the function below.

On a related note, spoke in person about why this function was split up. :+1: from me

mfraezz:
The claim account page seems not to use `zxcvbn` validation -- `'abc123'` was accepted, and the password strength bar never changed from blank:

screen shot 2016-05-20 at 11 34 15

mfraezz:
:bug: Ditto for the forgot password flow:

screen shot 2016-05-20 at 11 41 36

Tested to see if the above `email != password` validation had been fixed, but it looks like the routing change may have broken something -- submitting the form led to:

screen shot 2016-05-20 at 11 42 14

mfraezz:
Confirmed that validation still works on the landing page

website/static/js/osfHelpers.js
```diff
@@ -269,6 +269,47 @@ var mapByProperty = function(list, attr) {
 
 
 /**
+  * Return CSS colors and percentage of a progress bar based on a value.
+  * Used to return a value for password complexity
+  */
+var valueProgressBar = function(value) {
+    if (value >= 0) {
+        if (value === 0) {
```
sloria: You can refactor this to use a mapping.

```js
complexity_css_map = {
    0: {className: 'progress-bar ....', style: 'width: 20%'},
    1: ...
}
```

Also, better to use `className` rather than `class`, since `class` is a reserved word in newer version of JS.

website/static/js/setPassword.js
```diff
@@ -0,0 +1,234 @@
...
+    });
+
+    self.campaign = ko.observable(campaign);
+
+    // Preserve object of validated fields for use in `submit`
+    var validatedFields = {
+        password: self.password
+    };
+
+    if (passwordViewType === 'reset' || passwordViewType === 'change') {
```
sloria: Just a thought: it would be cool to refactor out this page-specific (passwordViewType) code. I can think of two ways.

1) Add a way to extend an observable with password-checking, a la knockout-validation. Something like

```js
var password = ko.observable('').extend({passwordChecking: true});
//...
// You would access the below methods in the HTML
password.passwordComplexity() // => 3
password.passwordSuggestions() // ['Too easy', ...]
```

```html
<p data-bind="foreach: password.passwordSuggestions()"> ...
```

I think this is the ideal API, as it allows very easy reuse of password validation logic. I haven't given much thought to implementation, though.

Alternatively...

2) Add a custom binding handler


```html
<input type="password" data-bind="passwordValidation: {observable: password, messages: pwMessages, complexity: pwComplexity}" />
```

```js
function ViewModel(..) {
    self.password = ko.obs();
    self.pwMessages = ko.obsArray();
    self. pwComplexity = ko.obs();
}
```

This API is less ergonomic than 1), but *might* be more straightforward to implement.


These are just thoughts, though. Refactoring isn't a blocker to get this merged, but it's worth investigating.

sloria:
Just a few drive-by comments. Bye for now =)

website/templates/landing.mako
```diff
@@ -79,7 +83,7 @@
                           </div>
                           <!-- /ko -->
                           <div>
-                              <button type="submit" class="btn btn-warning" data-bind="visible: !submitted()" id="signupSubmit">Sign up free</button>
+                              <button type="submit" class="btn btn-warning" data-bind="visible: !submitted(), css: {disabled: !password.isValid()}" id="signupSubmit">Sign up free</button>
```
mfraezz: At least on some pages (this one), this causes the button to not look clickable until both:

1. the password is valid
2. the user has clicked somewhere since the password became valid

screen shot 2016-05-31 at 16 04 12

Not sure if this is/should be a blocker, but it seems like it might be confusing to the user.

mfraezz:
Confirmed that validation worked on the landing, settings, and forgot password pages. :tada:
Also, confirmed that the above `email != password` bug for the forgot password flow was no longer present:

screen_shot_2016-05-31_at_16_22_04

Unfortunately... this happens when you submit the email you're claiming as your password on the claim account page 🐝:

screen shot 2016-05-31 at 16 20 00

Pass complete, back to you @erinspace 🌔

erinspace:
Just to clarify, this error on the claim page only happened if the user ignored all warnings and the greyed out button and hit enter, but ah well, can't control that eh. Should be good now.

mfraezz:
LGTM. Confirmed expected behavior on password-setting pages. Dropping this in RTM 🌠

framework/auth/views.py
```diff
@@ -49,26 +49,33 @@ def reset_password(auth, **kwargs):
...
     }
 
 
+def reset_password_post(verification_key):
+    """ attempt to reset password if request form passed
+    """
+    user_obj = get_user(verification_key=verification_key)
+    if not user_obj:
+        error_data = {
+            'message_short': 'Invalid url.',
```
sloria: +1 for custom error messages.

Super minor: Remove the period and uppercase "URL".

framework/auth/views.py
```diff
@@ -49,26 +49,33 @@ def reset_password(auth, **kwargs):
...
+            'message_long': 'The verification key in the URL is invalid or has expired.'
+        }
+        raise HTTPError(400, data=error_data)
+
+    user_obj.verification_key = security.random_string(20)
+    try:
+        user_obj.set_password(request.json['password'])
+    except HTTPError as e:
```
sloria: `set_password` should not raise an `HTTPError`; HTTP is outside of the responsibilities of the models. Instead, `set_password` should raise a `modularodm.exception.ValidationError` and that should be re-raised as an HTTPError here.

framework/auth/core.py
```diff
@@ -663,6 +665,8 @@ def set_password(self, raw_password, notify=True):
         """
         had_existing_password = bool(self.password)
         self.password = generate_password_hash(raw_password)
+        if self.username == raw_password:
```
sloria: See https://github.com/CenterForOpenScience/osf.io/pull/5585/files#r67396994

website/static/js/osfHelpers.js
```diff
@@ -271,6 +271,21 @@ var mapByProperty = function(list, attr) {
 
 
 /**
+  * Return CSS colors and percentage of a progress bar based on a value.
+  * Used to return a value for password complexity
+  */
+
+var valueProgressBar = {
```
sloria: Minor: I'm not sure this belongs here. The values (0-5) are really specific to password complexity, so it might make sense to just put this in `setPassword.js`.

website/static/js/setPassword.js
```diff
@@ -0,0 +1,267 @@
...
+//    - signup: required fields for Name, email, email confirmation, password, and password strength
+//    - reset: fields for password, strength, and password confirmation (for when you don't know your old password)
+//    - change: change your password when you know your old one
+var ViewModel = function(passwordViewType, submitUrl, campaign, redirectUrl) {
+
+    var self = this;
+
+    self.typedPassword = ko.observable('');
+
+    self.passwordInfo = ko.computed(function() {
```
sloria: Minor: this can be a `pureComputed`. Usually this doesn't matter much, but I'm not sure how efficient zxcvbn is, so it might be worth it here.

website/static/js/setPassword.js
sloria: After further investigation, approach 1) is totally possible! KO allows you to define custom "extenders": http://knockoutjs.com/documentation/extenders.html .

But again, this doesn't need to happen before this is merged. I've created a follow-up ticket for this: https://openscience.atlassian.net/browse/OSF-6514

website/static/js/pages/login-page.js
```diff
@@ -4,7 +4,7 @@
 'use strict';
 var $ = require('jquery');
 
-var SignUp = require('js/signUp');
+var SignUp = require('js/setPassword');
```
sloria: For consistency with other usages, this should be 

```js
var SetPassword = require('js/setPassword');
```

website/static/js/pages/profile-account-settings-page.js
```diff
@@ -3,12 +3,15 @@
 var $ = require('jquery');
 var $osf = require('js/osfHelpers.js');
 var accountSettings = require('js/accountSettings.js');
+var changePassword = require('js/setPassword');
```
sloria: Same here; keep the import names consistent.

```js
var SetPassword = require('js/setPassword');
```

website/static/js/pages/resetpassword-page.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Reset Password page
+ */
+'use strict';
+var $ = require('jquery');
+
+var SetPassword = require('js/setPassword');
+var verificationKey = window.contextVars.verification_key;
+var resetUrl = '/api/v1/resetpassword/' + verificationKey + '/';
+var redirectrUrl = '/login/';
```
sloria: Typo here: `redirectURL`.

website/static/js/setPassword.js
```diff
@@ -0,0 +1,267 @@
...
+        if (self.typedPassword()) {
+            return zxcvbn(self.typedPassword().slice(0, 100));
+        }
+    });
+
+    self.passwordFeedback = ko.computed(function () {
```
sloria: This can also be a `pureComputed`.

framework/auth/views.py
```diff
@@ -49,26 +49,33 @@ def reset_password(auth, **kwargs):
...
+    if not user_obj:
+        error_data = {
+            'message_short': 'Invalid url.',
+            'message_long': 'The verification key in the URL is invalid or has expired.'
+        }
+        raise HTTPError(400, data=error_data)
+
+    user_obj.verification_key = security.random_string(20)
+    try:
+        user_obj.set_password(request.json['password'])
```
sloria: `request.json` is deprecated in Flask. Use `request.get_json()` instead.

website/static/js/setPassword.js
```diff
@@ -0,0 +1,267 @@
...
+    self.flashMessage = ko.observable('');
+    self.flashMessageClass = ko.observable('');
+    self.flashTimeout = null;
+
+    self.trim = function(observable) {
+        observable($.trim(observable()));
+    };
+
+    /** Change the flashed message. */
+    self.changeMessage = function(message, messageClass, text, css, timeout, timeoutClock) {
```
sloria: This can be simplified; why pass both `message` and `text`? You should just be able to pass one of these. Passing a string message as the first argument.

```js
self.changeMessage('Password reset successful!', ...)
```

You already have an observable for storing the flashed message.

```js
self.changeMessage(message, ...) {
     self.flashMessage(message);
}

Comment (sloria) [inline website/static/js/setPassword.js]: Same goes with `messageClass` / `css`; you shouldn't need to pass both.

File: website/static/js/setPassword.js [hunk 20]
```diff
@@ -0,0 +1,267 @@
...
+    self.isValid = ko.computed(function() {
+        return self.validatedFields.isValid();
+    });
+
+    self.submitSuccess = function(response) {
+        self.changeMessage(
+            self.flashMessage,
+            self.flashMessageClass,
+            response.message,
+            'text-info p-xs'
```
sloria: As per osf-style, success messages should be green (`.text-success`).

caneruguz:
Hey @erinspace , Steve asked me to take a look at the UI. I have some comments on that:

- The progress bar seems a little too thick to me. Maybe the progress bar sizes we have in the style guide might help: https://centerforopenscience.github.io/osf-style/style.html#progress. It would be good to use heights in relation to placement. Probably mostly you want the md or sm sizes depending on general line height. They need to be much shorter than the input fields. 
- I am concerned that there is no indication that these are password strength metrics. Normally it would be accompanied by "Password Strength" or something similar. I understand that we don't want to add another line. It looks like most of our bars are long due to using the full width of the input placement. So how about splitting the bar line 60-40,  60% would have the bar length and the other 40% would have the words Weak, Medium, Strong or some indication of strength. Another option is to use the full bar to add these words on it but I don't like it because it can split the text when its at 50% or so. 
- On the main log in page, I am concerned about the blockiness of the password and help text fields. It loses the grouping somewhat that this is all about the password. I would make progress bar height less and not put background on the helper text but increase the space between all these and the bottom a bit. 

Let me know if you need clarification on any of this. 

Thanks for making password selection easier.

sloria:
Thanks @caneruguz 

Pass finished. Just a few minor changes. If Caner's suggestions are easy to implement, go ahead and throw those in, too, but I wouldn't say there are any major blockers.

sloria:
Did some plumbing and made a new PR #5928

PR #5588, opened 2016-05-02
[RfR] Disable create account btn after submission [OSF-5816]

barbour-em:
## Purpose

Prevent registration of the same account multiple times.

## Changes

Disable create account when the form is submitted

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-5816

[OSF-5816]

### DIFF
--- a/website/static/js/signUp.js
+++ b/website/static/js/signUp.js
@@ -103,6 +103,10 @@ var ViewModel = function(submitUrl, campaign) {
     };
 
     self.submit = function() {
+        if (self.submitted()) {
+            $osf.growl('Already submitted', 'You cannot sign up more than once.', 'error');
+            return false;
+        }
         // Show errors if invalid
         if (!self.isValid()) {
             // Ensure validation errors are displayed
--- a/website/static/js/signUp.js
+++ b/website/static/js/signUp.js
@@ -104,7 +104,7 @@ var ViewModel = function(submitUrl, campaign) {
 
     self.submit = function() {
         if (self.submitted()) {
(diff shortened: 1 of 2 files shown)


samanehsan:
Looks good @barbour-em! As an additional level of precaution, in the `submit` function, can you show an error message if the user already hit submit?

barbour-em:
@samanehsan added in bb20b1e. Hopefully the unrelated tests will pass this time.

website/static/js/signUp.js
```diff
@@ -103,6 +103,10 @@ var ViewModel = function(submitUrl, campaign) {
     };
 
     self.submit = function() {
+        if (self.submitted()) {
+            $osf.growl('Already submitted', 'You cannot sign up more than once.', 'error');
```
samanehsan: The `type` here should be `danger` instead of `error`. Also, the function defaults to `danger` so it's not necessary to pass it in. https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/osfHelpers.js#L26

samanehsan: Actually instead of a growlbox this error message can go in the form and replace the "Registration successful" message

barbour-em: ✔️

samanehsan:
Pass complete @barbour-em! 🌴

barbour-em:
@samanehsan, I missed [this comment](https://github.com/CenterForOpenScience/osf.io/pull/5588#discussion_r62053248) at first. Changes made.

samanehsan:
Thanks for making those changes @barbour-em! This is good to merge! :octocat:

PR #5589, opened 2016-05-02
Pass VOL query param to redirect_url [OSF-6161]

samanehsan:
## Purpose

The API guid redirect route throws an error when viewing a private project/file/etc. even if there is a view_only parameter included in the url.

## Changes

Add the `view_only parameter` to the redirect url.

## Side effects

None.

## Ticket

https://openscience.atlassian.net/browse/OSF-6161

### DIFF
--- a/api/guids/views.py
+++ b/api/guids/views.py
@@ -37,7 +37,10 @@ class GuidRedirect(JSONAPIBaseView):
 
     def get(self, request, **kwargs):
         url = self.get_redirect_url(**kwargs)
+        view_only_token = self.request.query_params.get('view_only')
         if url:
+            if view_only_token:
+                url += '?view_only={}'.format(view_only_token)
             return http.HttpResponseRedirect(url)
         raise NotFound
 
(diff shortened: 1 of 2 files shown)


brianjgeiger:
This looks great. 🎉 🐧 🎉

api/guids/views.py
```diff
@@ -37,7 +37,10 @@ def get_serializer_class(self):
 
     def get(self, request, **kwargs):
         url = self.get_redirect_url(**kwargs)
+        view_only_token = self.request.query_params.get('view_only')
         if url:
+            if view_only_token:
+                url += '?view_only={}'.format(view_only_token)
```
sloria: Is it possible for there to be query params on `url` before this point? If that's the case, then the `view_only` param will need to be appended with a `&`. You can use `furl` to take care of this if you want.

sloria:
Pass finished. Just one thing to address: https://github.com/CenterForOpenScience/osf.io/pull/5589#discussion_r62679322 . Also, as discussed on FD, let's get this in as a hotfix. Thanks!

api/guids/views.py
brianjgeiger: I don't *think* there are, but best to be sure. Good catch.

samanehsan: @sloria it's not possible, since `url` is the `absolute_api_v2_url` of the guid referent. Should I append all of the query params to the url or just the view only param?

samanehsan:
Re-opened as a hotfix https://github.com/CenterForOpenScience/osf.io/pull/5634 🔥

PR #5591, opened 2016-05-02
[0.69] Fix S3 addon selector and and deauth modals [#OSF-6156]

abought:
## Purpose
Fix JS console errors that occurred after authorizing S3 addon on a project settings page. These errors prevented selection of a bucket to be associated with that project.

Also identified and fixed some related errors on user settings page when deauthorizing an addon for a specific node, as well as disconnecting an addon account for the user.

Thanks to Nici for the help getting S3 working locally.

## Changes
- Fix s3 addon for node settings page
- Fix "Remove addon" modal dialogs on the user settings page; should now be able to remove an addon by clicking the red "x" next to a project name, without the "remove" button generating JS errors. Special cases addressed in this PR are noted as separate items:
  - Dataverse
  - S3
  - Settings for other addons
- Fix "Disconnect account" links on the user settings page. Clicking the remove button in the modal that comes up should no longer give a JS error. (h/t @tombaxter for reporting)
- Fix error rendering a "comment updated" log entry when commenting on wikis. The log entry would appear on the project dashboard.

## Side effects
None expected. Let me know if any addon-related buttons are broken on either the project or user settings "configure addons" pages.

## Ticket
https://openscience.atlassian.net/browse/OSF-6156

### DIFF
--- a/website/addons/dataverse/static/dataverseUserConfig.js
+++ b/website/addons/dataverse/static/dataverseUserConfig.js
@@ -135,7 +135,7 @@ function ViewModel(url) {
             title: 'Disconnect Dataverse Account?',
             message: '<p class="overflow">' +
                 'Are you sure you want to disconnect the Dataverse account on <strong>' +
-                $osf.htmlEscape(account.name) + '</strong>? This will revoke access to Dataverse for all projects associated with this account.' +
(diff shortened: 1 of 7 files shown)


abought:
The parent account "disconnect" buttons also have problems; investigating and will bundle fix into this PR.

samanehsan:
Tested out that this works! Looks good to me @abought! 🌲

PR #5592, opened 2016-05-03
Allow user to set category even for top-level projects. [OSF-5833]

abought:
Product team has requested that this be merged alongside #5617 

## Purpose
Top level projects should be allowed to have any category. All parts of the OSF should follow these rules consistently.

*From now on, a "project" is defined as something that has no parent. A "component" is something that does have a parent (even if the category is "project"- the category field is misleading).*

Most parts of the OSF (incl APIv2) should already be following these rules following last year's node refactor. This PR is intended to catch stragglers. It adds internal changes, and one small UI improvement.

## Changes
- The project settings page now allows you to set category for top-level projects (consistent with how APIv2 works). Top-level projects no longer have to be of category "project".

- Consolidated how a project was defined into a single place. This means that different parts of the OSF should give more consistent counts. (eg the profile page "# of public projects" count should match the number actually shown)
  - Updated project queries to exclude new things that are not projects (`is_collection`, `is_bookmark_collection`, etc)

- Altered metrics scripts to use new project definition

- Performance improvement for profile page: `len(list(queryset))` --> `queryset.count()`

## Side effects
The list of projects shown- and counts- will be subtly different than before. I have performed an initial audit on staging DB, before and after the code change. See spreadsheet link in JIRA ticket for some sample things that will be affected.

* Search results / category breakdown
* Recent activity page: `/explore/activity/` (also used by new and noteworthy script)
* Profile page projects list (counts, as well as what shows up in the projects vs components panels)
* Benchmarks run nightly
* Admin app metrics tool

## Ticket
https://openscience.atlassian.net/browse/OSF-5833


admin_tests/metrics/test_utils.py
```diff
@@ -25,25 +26,25 @@ class TestMetricsGetProjects(AdminTestCase):
...
         count = get_projects(public=True)
         nt.assert_equal(count, 1)
 
     def test_get_registrations(self):
         count = get_projects(registered=True)
         nt.assert_equal(count, 1)
 
-    def test_time(self):
-        time = self.node.date_created - timedelta(seconds=1)
+    def test_no_results_before_old_time(self):
```
brianjgeiger: This maybe could be disambiguated a tiny bit more. Is it, "No nodes before nodes were created", essentially? "No nodes a week ago because this test should only take a few seconds to run"? :)

abought: Yeah- the change was to make the time delay a bit longer, to discourage weird nondeterministic failures if something got out of sync.

admin_tests/metrics/test_utils.py
```diff
@@ -52,23 +53,22 @@ class TestMetricsGetDaysStatistics(AdminTestCase):
     def setUp(self):
         super(TestMetricsGetDaysStatistics, self).setUp()
         Node.remove()
-        NodeFactory(category='project')  # makes 2 nodes
-        NodeFactory(category='data')  # probably makes 1 data 1 project?
-        print get_projects()
+        NodeFactory(category='project')  # makes Node, plus parent
```
brianjgeiger: Not gonna lie: I kind of hate the NodeFactory and the need to have a comment about what it's doing. How about we replace it with something more explicit, like a project factory and then add a component to it? Since we're messing around here anyways.

abought: I removed it where it touched on the functionality of the PR (because I could get away with it), but otherwise tried to avoid scope creep.  @Ghalko concurs that `ProjectFactory` would be a better choice in future admin test work.

scripts/analytics/benchmarks.py
```diff
@@ -217,15 +209,15 @@ def main():
     projects_forked = get_projects_forked()
     projects_registered = get_projects_registered()
 
-    number_projects = len(projects)
+    number_projects = projects.count()
```
brianjgeiger: My python internals isn't as strong as it should be, but given that len() is super-pythonic and using it is disastrous from a performance standpoint, I wonder if we shouldn't make a maintenance ticket to override `__len__` for query sets to do what count() does.

abought: You're actually pretty close to the truth- ODM  already treats `len` and `count` the same way in ODM- unless you call `list` on something first. (this is an ODM behavior rather than python internals per se)

In this case, I'm deliberately being stubborn and using the queryset method, specifically to discourage us from treating lists and querysets as interchangeable. (given our problems in the past, I'm trying to make *sure* that this exact syntax breaks if given a list)

website/profile/views.py
```diff
@@ -57,15 +56,14 @@ def get_public_projects(uid=None, user=None):
 def get_public_components(uid=None, user=None):
     user = user or User.load(uid)
     # TODO: This should use User.visible_contributor_to?
-
+    # TODO: Should this be limited for users with many projects / components?
```
brianjgeiger: And/or does it has to be cast to a list? Get public projects isn't being cast to a list.

website/project/utils.py
```diff
@@ -6,16 +6,33 @@
 
 # Alias the project serializer
 from website.project.views.node import _view_project
-serialize_node = _view_project
+serialize_node = _view_project  # TODO: Seriously?
+
+CONTENT_NODE_QUERY = (
+    # Can encompass accessible projects, registrations, or forks
+    Q('is_bookmark_collection', 'ne', True) &
```
brianjgeiger: You can't have a bookmark collection without it also being a collection, so you can just use `is_collection` here.

website/profile/views.py
abought: Good questions!

`website.profile.utils.get_projects(...)` was only ever used to get a count, so the list call was totally unnecessary. There, the return value of the function could just be a queryset and there was a big potential performance gain.

By contrast, `get_public_projects` and `get_public_components` are API endpoints designed to return all entries, and they pass the results to `render_nodes` (which immediately turns the result into a list regardless).   Either way, `get_public_projects` actually does convert the result to a list first, in line 53- the functions work the same, just with different syntax.

abought: Also, protip: If I had to guess, I'd say the single best performance optimization we could make for profiles would be to use a datasource with pagination, so it doesn't have to load all that DB data at once. ;)

brianjgeiger: Perhaps some sort of new api? A second version, maybe.

brianjgeiger:
Okay, pass done. 🐧 You should make maintenance tickets for the todo items and link them to the todo lines as well.

brianjgeiger:
- [ ] Rename the one test to something slightly more descriptive
- [ ] Remove check for is_bookmark_collection in the one query

abought:
Pushed requested changes; I'll check CI in the AM.

brianjgeiger:
Changes are good. When tests pass, then we are ready to merge. 🎉 🐧 🎉

abought:
Restarted build for unrelated CI failures that crept in to develop last night.

Ran `flake8`, `jshint`, and OSF unit tests locally, without failures. Fingers crossed.

brianjgeiger:
Restarted second branch which failed again.

PR #5593, opened 2016-05-03
Fix guid redirect view scopes [OSF-6148]

samanehsan:
## Purpose

Users get the error "The user has not authorized access to this view" when viewing the `/v2/guids/<guid>` API route if they are authenticated with a personal access token.

## Changes

Create a `CoreScope` for the guids view and add it to `ComposedScopes.FULL_READ` instead of directly requiring `ComposedScope.FULL_READ`.

Also check to ensure that the required scopes are lists of `CoreScopes`, not `ComposedScopes`

## Side effects

None.

## Ticket

https://openscience.atlassian.net/browse/OSF-6148

### DIFF
--- a/api/base/permissions.py
+++ b/api/base/permissions.py
@@ -53,6 +53,11 @@ class TokenHasScope(permissions.BasePermission):
             except AttributeError:
                 raise ImproperlyConfigured('TokenHasScope requires the view to define the '
                                            'required_read_scopes attribute')
+            assert isinstance(view.required_read_scopes, list), 'The required_read_scopes must be a list of CoreScopes'
+            if view.required_read_scopes and type(view.required_read_scopes[0]) == tuple:
+                raise ImproperlyConfigured('TokenHasScope requires the view to define the '
+                                           'required_read_scopes attribute using CoreScopes')
+
             return read_scopes
         else:
             # A write operation implicitly also requires access to read scopes
@@ -61,7 +66,10 @@ class TokenHasScope(permissions.BasePermission):
             except AttributeError:
                 raise ImproperlyConfigured('TokenHasScope requires the view to define the '
                                            'required_write_scopes attribute')
-
+            assert isinstance(view.required_write_scopes, list), 'The required_write_scopes must be a list of CoreScopes'
+            if view.required_write_scopes and type(view.required_write_scopes[0]) == tuple:
+                raise ImproperlyConfigured('TokenHasScope requires the view to define the '
+                                           'required_write_scopes attribute using CoreScopes')
             return write_scopes
 
 
--- a/api/base/permissions.py
+++ b/api/base/permissions.py
@@ -7,6 +7,7 @@ from framework.auth import oauth_scopes
 from framework.auth.cas import CasResponse
 
 from website.models import ApiOAuth2Application, ApiOAuth2PersonalToken
+from website.util.sanitize import is_iterable_but_not_string
 
 
 # Implementation built on django-oauth-toolkit, but  with more granular control over read+write permissions
@@ -53,8 +54,9 @@ class TokenHasScope(permissions.BasePermission):
             except AttributeError:
                 raise ImproperlyConfigured('TokenHasScope requires the view to define the '
                                            'required_read_scopes attribute')
-            assert isinstance(view.required_read_scopes, list), 'The required_read_scopes must be a list of CoreScopes'
-            if view.required_read_scopes and type(view.required_read_scopes[0]) == tuple:
+            assert is_iterable_but_not_string(view.required_read_scopes), \
+                'The required_read_scopes must be an iterable of CoreScopes'
+            if view.required_read_scopes and isinstance(view.required_read_scopes[0], tuple):
                 raise ImproperlyConfigured('TokenHasScope requires the view to define the '
                                            'required_read_scopes attribute using CoreScopes rather than ComposedScopes')
 
@@ -66,8 +68,9 @@ class TokenHasScope(permissions.BasePermission):
             except AttributeError:
                 raise ImproperlyConfigured('TokenHasScope requires the view to define the '
                                            'required_write_scopes attribute')
-            assert isinstance(view.required_write_scopes, list), 'The required_write_scopes must be a list of CoreScopes'
-            if view.required_write_scopes and type(view.required_write_scopes[0]) == tuple:
+            assert is_iterable_but_not_string(view.required_read_scopes), \
+                'The required_write_scopes must be an iterable of CoreScopes'
+            if view.required_write_scopes and isinstance(view.required_write_scopes[0], tuple):
                 raise ImproperlyConfigured('TokenHasScope requires the view to define the '
                                            'required_write_scopes attribute using CoreScopes rather than ComposedScopes')
             return write_scopes
--- a/api/base/permissions.py
+++ b/api/base/permissions.py
@@ -56,7 +56,7 @@ class TokenHasScope(permissions.BasePermission):
             assert isinstance(view.required_read_scopes, list), 'The required_read_scopes must be a list of CoreScopes'
             if view.required_read_scopes and type(view.required_read_scopes[0]) == tuple:
                 raise ImproperlyConfigured('TokenHasScope requires the view to define the '
-                                           'required_read_scopes attribute using CoreScopes')
+                                           'required_read_scopes attribute using CoreScopes rather than ComposedScopes')
 
             return read_scopes
         else:
@@ -69,7 +69,7 @@ class TokenHasScope(permissions.BasePermission):
             assert isinstance(view.required_write_scopes, list), 'The required_write_scopes must be a list of CoreScopes'
             if view.required_write_scopes and type(view.required_write_scopes[0]) == tuple:
                 raise ImproperlyConfigured('TokenHasScope requires the view to define the '
-                                           'required_write_scopes attribute using CoreScopes')
+                                           'required_write_scopes attribute using CoreScopes rather than ComposedScopes')
             return write_scopes
 
 
--- a/api_tests/base/test_auth.py
+++ b/api_tests/base/test_auth.py
@@ -195,3 +195,19 @@ class TestOAuthScopedAccess(ApiTestCase):
 
         res = self.app.get(url, params=payload, auth='some_valid_token', auth_type='jwt', expect_errors=True)
         assert_equal(res.status_code, 403)
+
+    @mock.patch('framework.auth.cas.CasClient.profile')
+    def test_full_read_scope_can_read_guid_view(self, mock_user_info):
+        project = ProjectFactory()
+        mock_user_info.return_value = self._scoped_response(['osf.full_read'])
+        url = api_v2_url('guids/{}/'.format(project._id), base_route='/', base_prefix='v2/')
+        res = self.app.get(url, auth='some_valid_token', auth_type='jwt', expect_errors=True)
+        assert_equal(res.status_code, 302)
+
+    @mock.patch('framework.auth.cas.CasClient.profile')
+    def test_full_write_scope_can_read_guid_view(self, mock_user_info):
+        project = ProjectFactory()
+        mock_user_info.return_value = self._scoped_response(['osf.full_write'])
+        url = api_v2_url('guids/{}/'.format(project._id), base_route='/', base_prefix='v2/')
+        res = self.app.get(url, auth='some_valid_token', auth_type='jwt', expect_errors=True)
+        assert_equal(res.status_code, 302)
--- a/api_tests/base/test_auth.py
+++ b/api_tests/base/test_auth.py
@@ -9,6 +9,7 @@ from nose.tools import *  # flake8: noqa
 from framework.auth import cas
 from website.util import api_v2_url
 from website.addons.twofactor.tests import _valid_code
+from website.settings import API_DOMAIN
 
 from tests.base import ApiTestCase
 from tests.factories import AuthUserFactory, ProjectFactory, UserFactory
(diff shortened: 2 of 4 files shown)


api/base/permissions.py
```diff
@@ -53,6 +53,11 @@ def _get_scopes(self, request, view):
             except AttributeError:
                 raise ImproperlyConfigured('TokenHasScope requires the view to define the '
                                            'required_read_scopes attribute')
+            assert isinstance(view.required_read_scopes, list), 'The required_read_scopes must be a list of CoreScopes'
+            if view.required_read_scopes and type(view.required_read_scopes[0]) == tuple:
+                raise ImproperlyConfigured('TokenHasScope requires the view to define the '
+                                           'required_read_scopes attribute using CoreScopes')
```
brianjgeiger: Maybe mention also "rather than ComposedScopes" since very few people really know what they're doing with scopes.

api/base/permissions.py
```diff
@@ -61,7 +66,10 @@ def _get_scopes(self, request, view):
             except AttributeError:
                 raise ImproperlyConfigured('TokenHasScope requires the view to define the '
                                            'required_write_scopes attribute')
-
+            assert isinstance(view.required_write_scopes, list), 'The required_write_scopes must be a list of CoreScopes'
+            if view.required_write_scopes and type(view.required_write_scopes[0]) == tuple:
+                raise ImproperlyConfigured('TokenHasScope requires the view to define the '
+                                           'required_write_scopes attribute using CoreScopes')
```
brianjgeiger: Same here with additional message.

api_tests/base/test_auth.py
```diff
@@ -195,3 +195,19 @@ def test_node_write_scope_cant_read_user_view(self, mock_user_info):
 
         res = self.app.get(url, params=payload, auth='some_valid_token', auth_type='jwt', expect_errors=True)
         assert_equal(res.status_code, 403)
+
+    @mock.patch('framework.auth.cas.CasClient.profile')
+    def test_full_read_scope_can_read_guid_view(self, mock_user_info):
+        project = ProjectFactory()
+        mock_user_info.return_value = self._scoped_response(['osf.full_read'])
+        url = api_v2_url('guids/{}/'.format(project._id), base_route='/', base_prefix='v2/')
+        res = self.app.get(url, auth='some_valid_token', auth_type='jwt', expect_errors=True)
```
brianjgeiger: Are you expecting errors?

brianjgeiger: Also, if you allow redirect, you can make sure it goes through to the final location.

api_tests/base/test_auth.py
```diff
@@ -195,3 +195,19 @@ def test_node_write_scope_cant_read_user_view(self, mock_user_info):
...
+        assert_equal(res.status_code, 302)
+
+    @mock.patch('framework.auth.cas.CasClient.profile')
+    def test_full_write_scope_can_read_guid_view(self, mock_user_info):
+        project = ProjectFactory()
+        mock_user_info.return_value = self._scoped_response(['osf.full_write'])
+        url = api_v2_url('guids/{}/'.format(project._id), base_route='/', base_prefix='v2/')
+        res = self.app.get(url, auth='some_valid_token', auth_type='jwt', expect_errors=True)
```
brianjgeiger: Same as above

framework/auth/oauth_scopes.py
```diff
@@ -111,7 +113,7 @@ class ComposedScopes(object):
     NODE_ALL_WRITE = NODE_ALL_READ + NODE_METADATA_WRITE + NODE_DATA_WRITE + NODE_ACCESS_WRITE
 
     # Full permissions: all routes intended to be exposed to third party API users
-    FULL_READ = NODE_ALL_READ + USERS_READ + ORGANIZER_READ
+    FULL_READ = NODE_ALL_READ + USERS_READ + ORGANIZER_READ + (CoreScopes.GUIDS_READ, )
     FULL_WRITE = NODE_ALL_WRITE + USERS_WRITE + ORGANIZER_WRITE
```
brianjgeiger: Since you didn't make a GUIDS composed scope, and since it looks like GUIDS_READ isn't in the above scopes, I'm pretty sure you'll need to add your GUIDS_READ scope here as well.

brianjgeiger:
Pass done. Your verified your assert shows up when something is misconfigured and you did manual testing? 🐧

samanehsan:
Yes, I did manual testing to check that an error gets raised if the required scopes are not lists or if the view requires a composed scope.

samanehsan:
Done making those changes @brianjgeiger! 🌂

api_tests/base/test_auth.py
```diff
@@ -195,3 +196,23 @@ def test_node_write_scope_cant_read_user_view(self, mock_user_info):
...
+
+    @mock.patch('framework.auth.cas.CasClient.profile')
+    def test_full_read_scope_can_read_guid_view(self, mock_user_info):
+        project = ProjectFactory()
+        mock_user_info.return_value = self._scoped_response(['osf.full_read'])
+        url = api_v2_url('guids/{}/'.format(project._id), base_route='/', base_prefix='v2/')
+        res = self.app.get(url, auth='some_valid_token', auth_type='jwt')
+        redirect_url = '{}{}nodes/{}/'.format(API_DOMAIN, API_BASE, project._id)
+        assert_equal(res.status_code, 302)
+        assert_equal(res.location, redirect_url)
```
brianjgeiger: Since this request relies on a header, you'll probably want to verify that the whole transaction works (i.e. you can access the thing you're trying to access via the redirect). It can be an additional res and set of assertions, but I can see where just getting the URL won't necessarily get you the proper access to the resource.

api/base/permissions.py
```diff
@@ -53,6 +53,11 @@ def _get_scopes(self, request, view):
             except AttributeError:
                 raise ImproperlyConfigured('TokenHasScope requires the view to define the '
                                            'required_read_scopes attribute')
+            assert isinstance(view.required_read_scopes, list), 'The required_read_scopes must be a list of CoreScopes'
```
abought: Strictly speaking, this can also be an iterable. (I'm totally ok with the scopes field being a tuple instead of a list)

There is a relevant function `is_iterable_but_not_string`, though it inexplicably lives in `website.util.sanitize`...

api/base/permissions.py
```diff
@@ -53,6 +53,11 @@ def _get_scopes(self, request, view):
...
+            if view.required_read_scopes and type(view.required_read_scopes[0]) == tuple:
```
abought: Why use `type == tuple` instead of `isinstance(tuple)`?

api/base/permissions.py
```diff
@@ -61,7 +66,10 @@ def _get_scopes(self, request, view):
             except AttributeError:
                 raise ImproperlyConfigured('TokenHasScope requires the view to define the '
                                            'required_write_scopes attribute')
-
+            assert isinstance(view.required_write_scopes, list), 'The required_write_scopes must be a list of CoreScopes'
+            if view.required_write_scopes and type(view.required_write_scopes[0]) == tuple:
```
abought: See above question regarding `isinstance`

framework/auth/oauth_scopes.py
```diff
@@ -111,11 +116,11 @@ class ComposedScopes(object):
...
 
     # Full permissions: all routes intended to be exposed to third party API users
-    FULL_READ = NODE_ALL_READ + USERS_READ + ORGANIZER_READ
-    FULL_WRITE = NODE_ALL_WRITE + USERS_WRITE + ORGANIZER_WRITE
+    FULL_READ = NODE_ALL_READ + USERS_READ + ORGANIZER_READ + GUIDS_READ
+    FULL_WRITE = NODE_ALL_WRITE + USERS_WRITE + ORGANIZER_WRITE + GUIDS_READ
 
     # Admin permissions- includes functionality not intended for third-party use
-    ADMIN_LEVEL = FULL_WRITE + APPLICATIONS_WRITE + TOKENS_WRITE + COMMENT_REPORTS_WRITE
+    ADMIN_LEVEL = FULL_WRITE + APPLICATIONS_WRITE + TOKENS_WRITE + COMMENT_REPORTS_WRITE + GUIDS_READ
```
brianjgeiger: `ADMIN_LEVEL` has `FULL_WRITE` which has `GUIDS_READ`, so you don't need that here.

abought: Strictly speaking, this is already included in `FULL_WRITE` and may not need to be added to ADMIN a second time.

You can confirm this by running `oauth_scopes.py` as a standalone python file and seeing if it is in the admin scope list, without being added in this line.

api_tests/base/test_auth.py
samanehsan: It's possible that the user has correct scopes for the redirect view but does not have permission to view the project. So do you mean this test should include additional assert statements to check the user's permissions on the redirected view?

brianjgeiger: What I meant was that we should have the user redirect to a node that they have permissions for and ensure, using a single request, that the node information loads. So:

- User creates node with guid X
- User goes to /v2/guids/X using their auth token with follow_redirect=true
- User gets the proper project title with no more work

However, since you mention it, you should probably also do that where they are given a 403/401 when appropriate. If that works, I'm not super concerned about scopes. But if you want to go mad with power, see what happens if their token has scope for the redirect but not for the resource behind it.

samanehsan:
Done updating the tests @brianjgeiger! 🌻

api/base/permissions.py
```diff
@@ -53,6 +54,12 @@ def _get_scopes(self, request, view):
             except AttributeError:
                 raise ImproperlyConfigured('TokenHasScope requires the view to define the '
                                            'required_read_scopes attribute')
+            assert is_iterable_but_not_string(view.required_read_scopes), \
+                'The required_read_scopes must be an iterable of CoreScopes'
+            if view.required_read_scopes and isinstance(view.required_read_scopes[0], tuple):
```
brianjgeiger: Unless I'm missing something, wouldn't it be better to check each of the required_read_scopes items and verified `isinstance(scope, CoreScope)`? Sorry I missed the earlier discussion. That seems the most direct way of determining if everything is correct, yes?

brianjgeiger:
Okay, one more thing for me, and it should be ready to go. 🐧

api/base/permissions.py
samanehsan: The `view.required_read_scopes` will always contain `CoreScopes` from what I've seen. For example, if I set `required_read_scopes = [ComposedScopes.FOO]`, where `ComposedScopes.FOO = (CoreScopes.BAR, CoreScopes.FIZZ, )`, then `view.required_read_scopes = [(CoreScopes.BAR, CoreScopes.FIZZ)]`

brianjgeiger: Ah, okay. Makes sense. Is it worth separating everything out into items and checking not isinstance ComposedScope?

abought: We're not using instances of ComposedScopes; we're treating the values as strings.

(Abusing a class to act more like an enum, so all the possible choices are under one namespace)

abought: > @brianjgeiger, the view.required_read_scopes will always contain CoreScopes from what I've seen. 

And correct: each item in `required_read_scopes` should be a value (which for argument's sake is usually a string), and that value should be attached as one of the class-level attributes of `CoreScopes`. Eg:

`required_read_scopes = [CoreScopes.INSTITUTION_READ, CoreScopes.USERS_READ]`

Where `INSTITUTION_READ = 'institutions_read'`

abought:
Since the question keeps coming up, would you consider adding documentation comments to `oauth_scopes.py`, as shown below?

Other notes: In general, a scope describes the permissions needed for that individual view- most API endpoints serialize a specific kind of data, and with minimal overlap between endpoints. In that picture, APIs are typically used to fetch data, not web pages- and the mapping to Core Scopes makes sense (because it describes only the very narrow set of data needed for that endpoint)

The GUID endpoint is extremely unusual, in that it can return just about anything in the site. Hopefully most use cases would never require anything as complicated as a composed scope.

```
class CoreScopes(object):
    """
    The smallest units of permission that can be granted- all other scopes are built out of these.
    Each named constant is a single string.
    """
    # IMPORTANT: All views should be based on the smallest number of Core scopes required to describe the data in that view
    ...
```

```
class ComposedScopes(object):
    """
    Composed scopes, listed in increasing order of access (most restrictive first). Each named constant is a tuple.
    """
    # IMPORTANT: Composed scopes exist only as an internal implementation detail. All views should be based on selections from CoreScopes, above
    ...
```

brianjgeiger:
I'm not opposed to the documentation, but I expect it wouldn't have helped in this instance. I think the error will be the most useful.

samanehsan:
Back to you @brianjgeiger! ☔

brianjgeiger:
Looks great. 🎉 🐧 🎉

api/base/permissions.py
abought: One other note: it seems that we're trying to check whether the class definition is valid.

That seems like the kind of safeguard that would only need to happen once, to warn the developer they were making a mistake locally... but this middleware class will run the method on every single request. 

Is that performance overhead really necessary? Can we improve that, eg by checking once when the class is first instantiated?  (I'm not at all certain, but I *think* Django uses singletons where possible)

brianjgeiger: How long does it take to run isinstance()?

abought: Longer than it takes to not run it, and with much the same effect.

Performance is the cumulative result of many small design decisions. This one won't make or break us by itself, but it's worth asking ourselves the question for every new feature we push.

brianjgeiger: Yeah, but this change would have been much better three rounds of code review ago. Since it took so long, and since it's such a small amount of optimization, I do not recommend trying to move this code. Since it will be a while to  this merged in, replacing it with a try-except would be fine.

samanehsan:
Re-opened as a hotfix #5633 🔥

PR #5594, opened 2016-05-03
[hotfix] Sort registrations by date, most recent first

aaxelb:
## Purpose

Ensure that project registrations are shown in order of registration date, most recent first.

## Changes

Change the get_registrations endpoint to explicitly sort registrations by registered_date, instead of assuming they were already sorted and simply reversing the order.

Add a regression test therefor. 

## Side effects

None I can see.

## Ticket

https://openscience.atlassian.net/browse/OSF-6008


sloria:
Looks good to merge when the build passes.

PR #5596, opened 2016-05-03
[Feature] Add OSF spam comment rendering [ADM-27]

Ghalko:
## Purpose
This simply displays the comments that are spam and/or deleted.

## Changes
Changes `is_abuse` to `has_report` and uses `is_abuse` to signal if it is flagged or confirmed as spam.

## Side effects

Displays everything except singleton deleted comment, same as before.

## Ticket
ADM-27


admin/spam/views.py
```diff
@@ -174,6 +174,8 @@ def form_valid(self, form):
         try:
             if int(form.cleaned_data.get('confirm')) == Comment.SPAM:
                 item.confirm_spam(save=True)
+                item.is_deleted = True
```
mfraezz: Is there ever a case where we would want to `confirm_spam` and not delete the comment?

Ghalko: The problem is that `SpamMixin` doesn't know what it is going to be mixed in with. Right now it is only comments, which has the `is_deleted`, but if we apply this to other models it might not be. Either we will have to modify those models with a migration, or adjust how it is shown to be removed.

mfraezz: My vote would be moving towards a generalized way of indicating deletion, but this is fine for now.

admin/spam/views.py
```diff
@@ -174,6 +174,8 @@ def form_valid(self, form):
         try:
             if int(form.cleaned_data.get('confirm')) == Comment.SPAM:
                 item.confirm_spam(save=True)
```
mfraezz: You don't need to specify `save` here if you're saving it two lines down.

website/static/js/comment.js
```diff
@@ -273,6 +273,11 @@ var CommentModel = function(data, $parent, $root) {
     self.isAbuse = ko.observable(data.attributes.is_abuse);
     self.canEdit = ko.observable(data.attributes.can_edit);
     self.hasChildren = ko.observable(data.attributes.has_children);
+    self.hasReport = ko.observable(data.attributes.has_report);
+    self.isDeletedAbuse = ko.observable(self.isDeleted() && self.isAbuse());
+    self.isDeletedNotAbuse = ko.observable(self.isDeleted() && !self.isAbuse());
+    self.isAbuseNotDeleted = ko.observable(!self.isDeleted() && self.isAbuse());
```
mfraezz: These 'calculated' ones should probably be `ko.pureComputed`, not `.observable`. Could explain some UI bugs I was seeing

mfraezz:
Pass complete, couple of small issues found. Back to you, @Ghalko 💪

admin/spam/views.py
Ghalko: Removed and moved `item.save()` outside the if-else block.

website/static/js/comment.js
Ghalko: Moved all the `&&` observables to `pureComputed`.

Ghalko:
In order for me to test on my machine I need to get real CAS working, but I am going to push these fixes up if you want to take a look @mfraezz.

website/templates/include/comment_template.mako
```diff
@@ -21,14 +31,18 @@
...
+                        <span class="text-muted">
+                            <em>Comment reported.</em>
+                        </span>
+                        <span data-bind="if: hasChildren()" class="comment-actions pull-right">
                             <i data-bind="css: toggleIcon, click: toggle"></i>
                         </span>
-                        Comment reported.
                     </div>
-                    <a data-bind="click: submitUnreportAbuse">Not abuse</a>
+                    <div data-bind="if: hasReport">
```
mfraezz: `hasReport` is only set when data is received from the server, and I think it's semantically equivalent to `isAbuseNotDeleted`. This should be removed to maintain UI consistency.

Ghalko: No, has report refers to the person viewing the comments. If they have a report submitted then they can remove the report. @mfraezz

Ghalko: Added a `self.hasReport(true);` on successful submission of abuse to manually update.

Ghalko:
Closing in favor of one against develop.

PR #5598, opened 2016-05-03
Only show cancel file upload button while uploading [OSF-5972]

samanehsan:
## Purpose

If I cancel a file upload right when its upload progress reaches 100%, a growlbox pops up saying that the upload was cancelled. But when I refresh the page the file is still there.

## Changes

Only show the cancel upload button when upload progress is less than 100%

## Side effects

None.

## Ticket

https://openscience.atlassian.net/browse/OSF-5972

### DIFF
--- a/website/static/js/fangorn.js
+++ b/website/static/js/fangorn.js
@@ -123,36 +123,41 @@ var uploadRowTemplate = function(item){
     var columns = [{
         data : '',  // Data field name
         css : '',
-        custom : function(){ return m('row.text-muted', [
-            m('.col-xs-7', {style: 'overflow: hidden;text-overflow: ellipsis;'}, [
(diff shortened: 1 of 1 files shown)


brianjgeiger:
If you really wanted to eliminate the potential lag problem, you could set the disable to something short of 100%. That reduces their window to cancel, but it helps to eliminate thinking they cancelled when they won't have a chance.

Other than that, looks good. Up to you if you want to try reducing the window. You'll have a better sense from testing if that's good or no.

samanehsan:
@brianjgeiger, reducing a user's window to cancel won't make much of a difference for small files that load very quickly, and will probably look strange for larger files that load slowly (it would be obvious that the "X" is gone before the file has finished uploading) so I'll leave it as is.

brianjgeiger:
Cool.

brianjgeiger:
🎉 🐧 🎉

PR #5602, opened 2016-05-03
[Feature] Add 30-day Session Timeout for OSF [OSF-6099][SEC-10]

cslzchen:
## Purpose

OSF does not have a server-side session timeout validation. It relies on two client-side assumptions:
* Browser follows the protocol and removes the cookie when quits/expires
* The cookie is not stolen.

We add a server side validation which checks the age of the target session associated with a cookie. User is redirected to CAS if the session is older than 30 days, and the outdated session is removed from our database. Users will be automtically logged in if TGC is still alive.

## Changes

The validation is added to `before_request()` in `framework/sessions/__init__.py`:

```
if not util_time.throttle_period_expired(session.date_created, settings.COOKIE_TIMEOUT):
    ...
else
    remove_session(session)
```

## Side effects

No side effects

## Ticket

[OSF-6099](https://openscience.atlassian.net/browse/OSF-6099)
[SEC-10](https://openscience.atlassian.net/browse/SEC-10)


framework/sessions/__init__.py
```diff
@@ -166,9 +165,13 @@ def before_request():
             session = Session.load(session_id) or Session(_id=session_id)
         except itsdangerous.BadData:
             return
-        if session.data.get('auth_user_id') and 'api' not in request.url:
-            database['user'].update({'_id': session.data.get('auth_user_id')}, {'$set': {'date_last_login': datetime.utcnow()}}, w=0)
-        set_session(session)
+
+        delta = (datetime.utcnow() - session.date_created).total_seconds()
```
brianjgeiger: There's a helper function for this. `website.util.time.throttle_period_expired()`

brianjgeiger: I mean, I know it's super simple, but if time things get weird, and they sometimes do, at least this is tested and code centralized.

brianjgeiger:
Can you think of a way to make a test for this?

brianjgeiger:
Okay, so potential test and replace some code with a call to a helper function. Otherwise looks good. 🐧

framework/sessions/__init__.py
cslzchen: Sure, didn't know there is a helper function for it.

brianjgeiger: Yup. Saman didn't either before she implemented it recently, which is how I found out about it. :)

cslzchen:
I don't have a good test in mind now, but will figure it out.

brianjgeiger:
If not, no big deal. But if you think of something clever, cool. The throttling command itself is tested, and that's probably the only big thing. The other would be to separate out some of the functionality into a function and test that separately. But I'm not super-concerned.

tests/test_auth_basic_auth.py
```diff
@@ -79,3 +82,16 @@ def test_valid_credential_twofactor_valid_otp(self):
...
+        self.app.set_cookie(settings.COOKIE_NAME, str(cookie))
+        res = self.app.get(self.reachable_url)
+        assert_equal(res.status_code, 200)
+
+    def test_expired_cookie(self):
+        self.session = SessionFactory(user=self.user1, date_created=(datetime.utcnow() - timedelta(seconds=settings.OSF_SESSION_TIMEOUT)))
+        cookie = self.user1.get_or_create_cookie()
+        self.app.set_cookie(settings.COOKIE_NAME, str(cookie))
+        res = self.app.get(self.reachable_url)
+        assert_equal(res.status_code, 302)
```
mfraezz: It might be good to `assert_in('login', res.location)` as well, but it's probably fine as is.

mfraezz:
Minor nitpicking aside, LGTM. Dropping this in RTM ⭐

cslzchen:
@mfraezz Thanks for the review. @sloria Is it still ok for me to add the assertion for redirection url when it is in RTM?

cslzchen:
Add `assert_in('login', res.location)` as suggested by MattF. Drop to RTM directly.

website/util/time.py
```diff
@@ -10,7 +11,25 @@ def get_timestamp():
...
         return True
-    elif isinstance(timestamp, datetime.datetime):
-        return (datetime.datetime.utcnow() - timestamp).total_seconds() > throttle
+    elif isinstance(timestamp, datetime):
+        return (datetime.utcnow() - timestamp).total_seconds() > throttle
     else:
         return (get_timestamp() - timestamp) > throttle
+
+
+def generate_expiration_time(duration, base=datetime.utcnow(), utc=False):
```
sloria: The `datetime.utcnow()` in the signature is incorrect. Python will evaluate the call to `utcnow` whenever this file is imported, so you'll end up with an arbitrary default value. It should be changed to `None`.

website/util/time.py
```diff
@@ -10,7 +11,25 @@ def get_timestamp():
...
+    """
+    Generate an expiration timestamp
+    :param duration: duration in seconds
+    :param base: optional base timestamp
+    :param utc: optional boolean flag which determines return type
+    :return: a timestamp object (default) or a utc string (if utc is set to True)
+    """
+    if not base:
+        base = datetime.utcnow()
+    elif not isinstance(base, datetime):
```
sloria: This branch doesn't seem necessary. The user should only pass either a datetime or `None`.

website/util/time.py
sloria: Actually, this function doesn't appear to be used anywhere. Can this be removed?

tests/test_auth_basic_auth.py
```diff
@@ -95,3 +95,4 @@ def test_expired_cookie(self):
         self.app.set_cookie(settings.COOKIE_NAME, str(cookie))
         res = self.app.get(self.reachable_url)
         assert_equal(res.status_code, 302)
+        assert_in('login', res.location)
```
sloria: This is fine. To be even more specific, I believe you can do a direct comparison against `res.request.path`. But it's no big deal.

sloria:
Pass finished.

website/util/time.py
cslzchen: Yes, this should be removed. It was used for cookie flag `Max-Age` and `Expired` which is now removed.

cslzchen:
Thanks @sloria. Move to RTM after Travis.

PR #5603, opened 2016-05-04
[Feature] Admin Render removed user as unregistered [ADM-67]

Ghalko:
## Purpose

Addresses the issue that a user account that has been deactivated still has live links in contributor lists. Clicking on this dead link gives a 'resource removed' message which is annoying. By removing this link it removes this possibility.

## Changes

Make a user unregistered, `is_registered = False` when user account is disabled. And `is_registered = True` when reactivated.

## Side effects
Using the `is_registered` for this purpose might seem as though the variable is being misused, however, a `User` object has several other boolean fields that do much the same things.

## Ticket

[ADM-67]


Ghalko:
Wrong branch.

PR #5608, opened 2016-05-04
[Feature] Sales Analytics Page and Dashboard

jeffreyliu3230:
## Purpose

Add sales analytics page to the admin app.
Add analytics dashboard to the admin app.
## Changes

Dashboard on the sales-analytics page using bootstrap and keen's template. Add db metrics and visualization. 

## Side effects

Keen visualization is currently hidden on the dashboard. We haven't decided what to show from Keen.

## Ticket

https://openscience.atlassian.net/browse/OSF-5923


admin/sales_analytics/keen.py
```diff
@@ -0,0 +1,25 @@
...
+
+if KEEN_CREDENTIALS['keen_ready']:
+    KEEN_CREDENTIALS.update({
+        'keen_project_id': KEEN_PROJECT_ID,
+        'keen_read_key': KEEN_READ_KEY,
+        'keen_write_key': KEEN_WRITE_KEY
+    })
+
+
+# TODO: when Keen is ready, use of settings instead
```
Ghalko: Shouldn't settings be in `defaults.py` There is certainly a way that the credentials chain can work with defaults and local.py in the admin and the OSF.

admin/sales_analytics/metrics.py
```diff
@@ -0,0 +1,113 @@
+from admin.base.settings import ENTRY_POINTS
+from admin.sales_analytics.utils import get_entry_point, get_sorted_index
+from datetime import datetime, timedelta
+from framework.mongo import database as db
+
+
+def get_user_count(db=db, entry_points=ENTRY_POINTS):
```
Ghalko: The name of this file should conform to the module standards (mostly Django standards) This looks like either a `utils.py` or a `serializers.py` file.

admin/sales_analytics/metrics.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    Get the number of users created from each entry point: osf, osf4m, prereg, and institution.
+    """
+    count_list = []
+    percent_list = []
+    tags = ['osf4m', 'prereg', 'institution', 'osf']
+    total = db.user.find({}).count()
+    for i in entry_points:
```
Ghalko: Generally you should reserve i, j, and k for numerical iterators, and single character variables for list comprehension. I would suggest `entry_point` as it is the singular.

admin/sales_analytics/keen.py
cslzchen: I planned to wait until Keen becomes ready in OSF. But we can definitely change to the commented code below to use OSF settings now.

jeffreyliu3230: @cslzchen worked on this part. Thoughts?

Ghalko: This code though looks an awful lot like it should be in `defaults.py` You can comment it out there too. Having settings spread throughout the code is not a good idea.

admin/sales_analytics/metrics.py
```diff
@@ -0,0 +1,113 @@
...
+        count = db.user.find({'system_tags': i}).count()
+        percent = round(float(count) / float(total), 2)
+        count_list.append(count)
+        percent_list.append(percent)
+    osf_count = total - sum(count_list)
+    osf_percent = 1 - sum(percent_list)
+    count_list.append(osf_count)
+    percent_list.append(osf_percent)
+    sorted_index = get_sorted_index(count_list)
```
Ghalko: This is only used once and it is one line, unless you intend to use it or test it thoroughly (not a bad idea) then I would suggest just moving it into here.

Ghalko: Also it might be better to sort the list initially before using it in the for loop. Then you might not have to go through all of this post loop code.

admin/sales_analytics/keen.py
cslzchen: @jeffreyliu3230 just remove the code above and uncommented the code following TODO, thanks. the keen settings are already in OSF

admin/sales_analytics/metrics.py
```diff
@@ -0,0 +1,113 @@
...
+from framework.mongo import database as db
+
+
+def get_user_count(db=db, entry_points=ENTRY_POINTS):
+    """
+    Get the number of users created from each entry point: osf, osf4m, prereg, and institution.
+    """
+    count_list = []
+    percent_list = []
+    tags = ['osf4m', 'prereg', 'institution', 'osf']
```
Ghalko: It seems like these tags mirror the entry points with the addition of 'osf.' This means that every time you update `ENTRY_POINTS` you will have to update tags. And, if I am not mistaken, keep them in the same order, or risk mislabeling data. I would suggest that you use `ENTRY_POINTS` as a basis list and then add 'osf' after you are done with the for loop.

admin/sales_analytics/metrics.py
```diff
@@ -0,0 +1,113 @@
...
+    pipeline = [
+        {'$match': {'date': {'$gt': start_date}}},
+        {'$group': {'_id': '$user', 'node_id': {'$addToSet': '$params.node'},
+                    'action': {'$addToSet': '$action'}}}
+    ]
+    user_node = db.nodelog.aggregate(pipeline)['result']
+    multi_product_count = 0
+    cross_product_count = 0
+    multi_action_count = 0
+    for i in user_node:
```
Ghalko: Again, `i` should be reserved for numerical iterators. Also it might be good to change `user_node` to a plural. Or maybe even a name that doesn't step on the Node nomenclature of the OSF.

admin/sales_analytics/metrics.py
```diff
@@ -0,0 +1,113 @@
...
+                else:
+                    repeat_action_count += 1
+                    date_registered = db.user.find({'_id': user_id}).next()['date_registered']
+                    age = (date - date_registered).days
+                    repeat_action_user_age.append(age)
+                    break
+    return {'repeat_action_count': repeat_action_count, 'repeat_action_age': repeat_action_user_age}
+
+
+user_count = get_user_count(db, ENTRY_POINTS)
```
Ghalko: I am not entirely sure why you have these global variables.

Ghalko:
Remove `sales_analytics/models.py` as it isn't used.

admin/sales_analytics/urls.py
```diff
@@ -0,0 +1,11 @@
+from django.conf.urls import url
+
+from . import views
+
+urlpatterns = [
+    url(r'^dashboard$', views.dashboard, name='dashboard'),
+    url(r'^user_session', views.user_session, name='user_session'),
+    url(r'^products_view', views.products_view, name='products_view'),
+    url(r'^products_usage', views.products_usage, name='products_usage'),
+    url(r'^debug_test', views.debug_test, name='debug_test'),
```
Ghalko: `debug_test` should be removed unless useful for staging testing.

admin/sales_analytics/views.py
```diff
@@ -0,0 +1,31 @@
+from django.shortcuts import render
+from admin.sales_analytics import keen
+from admin.sales_analytics.metrics import user_count, multi_product_metrics_yearly, multi_product_metrics_monthly, repeat_action_user_monthly
+
+
+def dashboard(request):
```
Ghalko: Even though this is behind a VPN we should consider putting security on these. It also might be nice to take a CBV approach in order to make this simpler to do and test. Please look at other admin modules in order to get an idea.

admin/sales_analytics/urls.py
cslzchen: Currently all keen related analytics such as `user_session`, `products_view` and `debug_test` are turned off for staging by setting `KEEN_READY` to `False`, but the code remains there. I had planned the last one to serve for user customized queries. Maybe rename it? Or should we remove all keen stuff and only leave dashboard there?

admin/static/js/sales_analytics/dashboard.js
```diff
@@ -0,0 +1,430 @@
...
+        self.keenClient.run(query, function(error, response) {
+            if (error) {
+                document.getElementById('keen-chart-average-session-user').innerHTML = 'Keen Query Failure';
+                console.log(error);
+            }
+            else {
+                var dataSet = self.extractDataSet(response);
+                var result = ss.mean(dataSet);
+                self.drawChart(chart, 'metric', 'Minutes', result/1000/60);
+                console.log('average user session length is ' + result + " ms" );
```
Ghalko: Are these console.log statements for debugging? If so, remove.

cslzchen: Yes and we will remove them.

admin/static/js/sales_analytics/dashboard.js
```diff
@@ -0,0 +1,430 @@
...
+                    categories: userCount.tags,
+                },
+                rotated: true
+            },
+            bar: {
+                width: {
+                    ratio: 0.5 // this makes bar width 50% of length between ticks
+                }
+                // or
+                //width: 100 // this makes bar width 100px
```
Ghalko: Remove if not using.

admin/static/js/sales_analytics/dashboard.js
```diff
@@ -0,0 +1,430 @@
...
```
Ghalko: Remove if not using.

admin/static/js/sales_analytics/dashboard.js
```diff
@@ -0,0 +1,430 @@
...
+        self.getCrossProductCountMonthly(multiProductMetricsMonthly);
+        self.getMultiActionCountYearly(multiProductMetricsYearly);
+        self.getMultiActionCountMonthly(multiProductMetricsMonthly);
+        self.getRepeatActionCountMonthly(repeatActionUserMonthly);
+        self.getTotalUserCount(userCount);
+    };
+
+    self.clean = function() {
+        console.log('clean');
+    };
```
Ghalko: Okay? Does this do anything?

admin/static/js/sales_analytics/dashboard.js
```diff
@@ -0,0 +1,430 @@
...
+
+        var beginTime;
+        var endTime;
+        var deltaSet = [];
+
+        for (var i in keenResult.result) {
+            var session = keenResult.result[i];
+            if (session.hasOwnProperty('result')) {
+                if (session.result.length === 1) {
+                    // TODO: take care of the situation where there is only one 'keen.timestamp'
```
Ghalko: Why is this not doable now?

admin/sales_analytics/metrics.py
jeffreyliu3230: Those variables are used by views.py to be put in context. This way I could import variables in views.py instead of the functions and all the variables needed to pass to the functions. I thought this makes views.py much cleaner.

admin/static/js/sales_analytics/sa-products-view.js
```diff
@@ -0,0 +1,138 @@
...
+            property_name: 'user.id',
+            operator: 'ne',
+            property_value: null
+        },
+        inactiveUserFilter: {
+            property_name: 'user.id',
+            operator: 'ne',
+            property_value: ''
+        }
+    };
```
Ghalko: How much of this is the same as the other file? Can you make a parent 'class'?

admin/static/js/sales_analytics/sa-user-session.js
```diff
@@ -0,0 +1,212 @@
...
+            property_name: 'user.id',
+            operator: 'ne',
+            property_value: null
+        },
+        inactiveUserFilter: {
+            property_name: 'user.id',
+            operator: 'ne',
+            property_value: ''
+        }
+    };
```
Ghalko: Again looks like quite a bit of repetition between files. Please consider the DRY principle.

admin/static/js/sales_analytics/sa-user-session.js
```diff
@@ -0,0 +1,212 @@
...
+                deltaSet.push(endTime - beginTime);
+            }
+        }
+        return deltaSet;
+    };
+
+};
+
+var salesAnalytics = new SalesAnalytics();
+salesAnalytics.init();
```
Ghalko: End of file needs extra line.

admin/static/js/sales_analytics/sales-analytics.js
```diff
@@ -0,0 +1,293 @@
...
+            property_name: 'user.id',
+            operator: 'ne',
+            property_value: null
+        },
+        inactiveUserFilter: {
+            property_name: 'user.id',
+            operator: 'ne',
+            property_value: ''
+        }
+    };
```
Ghalko: I am repeating myself quite alot, but Don't Repeat Yourself. (DRY)

Ghalko:
Remove `static/js/sales_analytics/utils.js` if not used.

admin/templates/base.html
```diff
@@ -140,4 +141,4 @@
     {% endblock %}
 
   </body>
-</html>
+</html>
```
Ghalko: New line at the end of the file.

admin/sales_analytics/metrics.py
jeffreyliu3230: Good points about updating the list. I didn't have tags in the beginning, but names in the ENTRY_POINTS are not good for labelling the data. so maybe changing ENTRY_POINTS to a list of tuples that contains the original system tags and their display names?

admin/templates/base_analytics.html
```diff
@@ -0,0 +1,52 @@
...
+
+    {% block title_analytics %}
+
+    {% endblock title_analytics %}
+
+    {% block content_analytics %}
+
+    {% endblock content_analytics %}
+
+{% endblock content %}
```
Ghalko: EOF, add newline.

admin/templates/sales_analytics/debug_test.html
```diff
@@ -0,0 +1,6 @@
+{% extends 'base_analytics.html' %}
+{% load static %}
+
+{% block title_analytics %}
+    <h1>Debug and Test</h1>
+{% endblock title_analytics %}
```
Ghalko: Newline.

admin/templates/sales_analytics/products_usage.html
```diff
@@ -0,0 +1,6 @@
+{% extends 'base_analytics.html' %}
+{% load static %}
+
+{% block title_analytics %}
+    <h1>OSF Products Usage</h1>
+{% endblock title_analytics %}
```
Ghalko: newline

admin/templates/sales_analytics/products_view.html
```diff
@@ -0,0 +1,17 @@
...
+{% block content_analytics %}
+    <div id='keen-chart-osf-products-view-mt2'>...</div>
+    <div id='keen-chart-osf-products-view-mee'>...</div>
+    <div id='keen-chart-osf-products-view-pre'>...</div>
+    <div id='keen-chart-osf-products-view-ins'>...</div>
+{% endblock content_analytics %}
+
+{% block bottom_js %}
+    <script src="{% static 'public/js/sa-products-view.js' %}"></script>
+{% endblock bottom_js %}
```
Ghalko: newline

admin/templates/sales_analytics/user_session.html
```diff
@@ -0,0 +1,19 @@
...
+    <div id='keen-chart-average-session-user'>...</div>
+    <h3>Average MAU Session Length of Previous 7 days</h3>
+    <div id='keen-chart-average-session-mau'>...</div>
+    <h3>Average User/MAU Session Length History of past 12 weeks</h3>
+    <div id='keen-chart-average-session-history'>...</div>
+{% endblock content_analytics %}
+
+{% block bottom_js %}
+    <script src="{% static 'public/js/sa-user-session.js' %}"></script>
+{% endblock bottom_js %}
```
Ghalko: newline

admin/templates/sales_analytics/products_usage.html
sloria: FWIW, we explicitly ignore the newline rule in our flake8 config, as I don't think it significantly affects readability.

admin/static/js/sales_analytics/dashboard.js
jeffreyliu3230: Probably not. Will remove it.

admin/static/js/sales_analytics/dashboard.js
jeffreyliu3230: @cslzchen what was exactly the issue here?

admin/static/js/sales_analytics/sa-products-view.js
jeffreyliu3230: Yes lots of repetitions here and there. all the js files other than dashboard.js should be deleted in this case.

admin/static/js/sales_analytics/dashboard.js
cslzchen: It was in the general design but ended up not being used. Safe to remove.

admin/static/js/sales_analytics/dashboard.js
cslzchen: This is doable, but it is one of the many issues related staging data and incorrect data uploads. That's why mine needs to wait until Gage's Keen is fully ready (and start to run for a while). The best way for yours to be merged is to remove all my code except the structure you need. I already take notes of Bryan's review and will fix them soon.

cslzchen: I was half wrong. This single timestamp issue can happen a lot when we only track page refresh / new page event. Gage is not adding any new tracking based on user actions right now. So I still need to fix this part. :)

admin/static/js/sales_analytics/sa-products-view.js
cslzchen: There used to be only one file, but I move them to different pages. They are suppose be in {{utils.js}} and exported for other js to require. But I haven't got time to do it yet.

jeffreyliu3230:
@Ghalko @sloria Thanks for all the comments! I just pushed the new commits again but somehow Travis  doesn't start testing the changes?

jeffreyliu3230:
@Ghalko All tests passed. BTW the superuser I created does not have the access to the pages with OSFAdmin class view. Anyway I could fix it?

Ghalko:
@jeffreyliu3230 You need to go into the django admin interface (/admin) and give your user the osf_admin group.

admin/base/settings/defaults.py
```diff
@@ -197,3 +198,23 @@
         ]
     },
 }
+
+# Keen.io settings in local.py
+KEEN_PROJECT_ID = 'CHANGEME'
+KEEN_READ_KEY = 'CHANGEME'
+KEEN_WRITE_KEY = 'CHANGEME'
```
Ghalko: If these are going to be defined in OSF settings local.py, then we should point at osf settings.

jeffreyliu3230: right now osf settings only have KEEN_PROJECT_ID and KEEN_WRITE_KEY defined. I'll go ahead and add KEEN_READ_KEY as well then point the settings to osf_settings

jeffreyliu3230: Updated

jeffreyliu3230:
@sloria Travis might be stuck again?

Ghalko:
@jeffreyliu3230 No there are just many, many things in the queue. If you submit quickly you can have a stack of them. I canceled your 'change settings' commit build so that the second one could build.

jeffreyliu3230:
@Ghalko Oh cool thanks!

admin/sales_analytics/utils.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+# Helper functions
+def get_entry_point(system_tags, entry_points=ENTRY_POINTS):
+    """
+    Given the user system_tags, return the user entry point (osf, osf4m, prereg, institution)
+    In case of multiple entry_points existing in the system_tags, return only the first one.
+    """
+    entry_points = entry_points.keys()
+    for i in system_tags:
```
Ghalko: This loop makes a few assumptions: That there is only one system tag from the `entry_points` and that it will be in order.

Also it is standard practice to only use `i` as an integer variable in a loop. Please consider using `tag` or some other somewhat descriptive name.

jeffreyliu3230: Yes that is the assumption. Occasionally users would have more than one entry point but it is very rare (and nobody knows why). Therefore, using the first one is probably good enough.

admin/sales_analytics/utils.py
```diff
@@ -0,0 +1,137 @@
...
+                    age = (date - date_registered).days
+                    repeat_action_user_age.append(age)
+                    break
+    return {'repeat_action_count': repeat_action_count, 'repeat_action_age': repeat_action_user_age}
+
+
+user_count = get_user_count(db, ENTRY_POINTS)
+multi_product_metrics_yearly = get_multi_product_metrics()
+multi_product_metrics_monthly = get_multi_product_metrics(timedelta=timedelta(days=30))
+repeat_action_user_monthly = get_repeat_action_user_count()
```
Ghalko: Not convinced that this is the best way to "export" these. Basically they are run every time this file is imported, although that is only once it does also mean that you could have stale data.

jeffreyliu3230: Probably. I'll import the functions instead

Ghalko:
@jeffreyliu3230 Pass finished.

jeffreyliu3230:
@Ghalko Thank you. Updated.

PR #5610, opened 2016-05-04
Feature: Wiki v2 API read-only routes [OSF-6168]

samanehsan:
Purpose
------------
Add read-only routes for wiki pages.

Changes
------------
Added `wikis` as a relationship field on the NodeSerializer
Added wiki routes:
- `/v2/nodes/node_id/wikis/`
- ~~`/v2/nodes/node_id/wikis/wiki_id`~~
- `/v2/registrations/registration_id/wikis/`
- ~~`/v2/registrations/registration_id/wikis/wiki_id`~~
- `/v2/wikis/wiki_id/`: 

Only current versions of wiki pages are viewable through the API. The node/registration wiki lists are filterable by "name" and "date_modified."

The wiki serializer returns metadata about the wiki page, and includes a "download" link to retrieve the contents of the wiki page. The link is an additional API v2 route that uses a custom renderer.

Here is an example output for the wiki detail route (the node wiki list route will have a list of these wikis):
screen shot 2016-05-10 at 4 23 47 pm

Side Effects
----------------
Update the wiki comment API code that was written when the wiki routes did not exist.

JIRA Ticket
----------------
https://openscience.atlassian.net/browse/OSF-6168


api/base/urls.py
```diff
@@ -24,7 +24,8 @@
                 url(r'^docs/', include('rest_framework_swagger.urls')),
                 url(r'^institutions/', include('api.institutions.urls', namespace='institutions')),
                 url(r'^collections/', include('api.collections.urls', namespace='collections')),
-                url(r'^guids/', include('api.guids.urls', namespace='guids'))
+                url(r'^guids/', include('api.guids.urls', namespace='guids')),
+                url(r'^wikis/', include('api.wikis.urls', namespace='wikis'))
```
brianjgeiger: Totally minor, but feel free to include the `,` at the end of this line because python.

api/nodes/views.py
```diff
@@ -2006,3 +2010,151 @@ def update(self, request, *args, **kwargs):
...
+
+    ordering = ('-date', )  # default ordering
+
+    # overrides ODMFilterMixin
+    def get_default_odm_query(self):
+        node = self.get_node()
+        wiki_pages = node.wiki_pages_current
+        node_wiki_pages = []
+        for key in wiki_pages:
+            node_wiki_pages.append(wiki_pages[key])
```
brianjgeiger: If you wanted to do a list comprehension here, it's an ideal use case.

api/wikis/serializers.py
```diff
@@ -0,0 +1,71 @@
...
+    ])
+
+    id = IDField(source='_id', read_only=True)
+    type = TypeField()
+    name = ser.CharField(source='page_name')
+    path = ser.SerializerMethodField()
+    materialized = ser.SerializerMethodField(method_name='get_path')
+    date_modified = ser.DateTimeField(source='date')
+    content_type = ser.SerializerMethodField()
+    extra = ser.SerializerMethodField(read_only=True, help_text='Additional metadata about this wiki')
```
brianjgeiger: I'm not seeing the `kind` field.

api/wikis/serializers.py
```diff
@@ -0,0 +1,71 @@
...
+        'name',
+        'date_modified'
+    ])
+
+    id = IDField(source='_id', read_only=True)
+    type = TypeField()
+    name = ser.CharField(source='page_name')
+    path = ser.SerializerMethodField()
+    materialized = ser.SerializerMethodField(method_name='get_path')
+    date_modified = ser.DateTimeField(source='date')
```
brianjgeiger: I believe this should just be `modified` (see files serializer in case the docs that I copied this from were wrong)

api/wikis/serializers.py
brianjgeiger: Also no `size` field. I forget if we talked about skipping that one, though shouldn't be hard to implement I think.

api/wikis/views.py
```diff
@@ -0,0 +1,145 @@
...
+    URL. By default, fetches the wiki page based on the wiki_id kwarg.
+    """
+
+    serializer_class = WikiSerializer
+    wiki_lookup_url_kwarg = 'wiki_id'
+
+    def get_wiki(self, check_permissions=True):
+        pk = self.kwargs[self.wiki_lookup_url_kwarg]
+        try:
+            wiki = NodeWikiPage.find_one(Q('_id', 'eq', pk))
```
brianjgeiger: Can find_one() be replaced with load()?

api_tests/wikis/views/test_wiki_detail.py
```diff
@@ -0,0 +1,170 @@
...
+                             NodeWikiFactory, PrivateLinkFactory)
+
+
+class TestWikiDetailView(ApiTestCase):
+    def setUp(self):
+        super(TestWikiDetailView, self).setUp()
+        self.user = AuthUserFactory()
+        self.non_contributor = AuthUserFactory()
+
+    def _add_project_wiki_page(self, node, user):
```
brianjgeiger: These all look very similar. Can you make an ApiWikiTestCase that holds all the setup methods?

brianjgeiger:
Unless I missed it, I think this will also need some tests for the download link and renderer.

brianjgeiger:
Finally, could you put the output of the various routes in the Changes section of the wiki so QA will know what they're looking for? Pass finished! 🐧

api/nodes/views.py
samanehsan: Oh! I can simplify this even further with just `node_wiki_pages.values()`

api/wikis/serializers.py
samanehsan: It's `date_modified` in the `FileSerializer`: https://github.com/CenterForOpenScience/osf.io/blob/develop/api/files/serializers.py#L103

api/wikis/serializers.py
samanehsan: Should `kind = file`?

brianjgeiger: Yup.

api/wikis/serializers.py
brianjgeiger: Okay, carry on, then. I suspect the docs are wrong or I copied and pasted _very_ badly.

api/wikis/serializers.py
samanehsan: Cool, and any suggestions on getting the `size` of the wiki file? For files, size is included with the metadata but this info is not stored on the wiki model

brianjgeiger: It should just be `sys.getsizeof(string)`.

samanehsan:
Done making those changes @brianjgeiger! 🎉

api/base/renderers.py
```diff
@@ -31,3 +31,14 @@ def get_context(self, *args, **kwargs):
...
+
+class PlainTextRenderer(StaticHTMLRenderer):
+    """
+    Renders plain text.
+
+    Inherits from StaticHTMLRenderer for error handling.
+    """
+
+    media_type = 'text/markdown'
+    format = '.txt'
```
brianjgeiger: What all does this variable do?

samanehsan: I'm not so sure, but it's required according to the docs http://www.django-rest-framework.org/api-guide/renderers/#custom-renderers:
> To implement a custom renderer, you should override BaseRenderer, set the .media_type and .format properties, and implement the .render(self, data, media_type=None, renderer_context=None)

samanehsan: But...I think it should be `txt` and not `.txt`

brianjgeiger: If it affects the extension that the browser uses for downloads, it may be we want it to be `markdown` or `md`.

samanehsan: Oh okay! It's hard to tell the difference because the route is only rendering wiki content and not downloading it.

brianjgeiger: So, as discussed in person, `txt` is probably the best thing here.

api/nodes/views.py
```diff
@@ -2159,3 +2163,148 @@ def update(self, request, *args, **kwargs):
...
+    def get_default_odm_query(self):
+        node = self.get_node()
+        node_wiki_pages = node.wiki_pages_current.values() if node.wiki_pages_current else []
+        return Q('_id', 'in', node_wiki_pages)
+
+    def get_queryset(self):
+        return NodeWikiPage.find(self.get_query_from_request())
+
+
+class NodeWikiDetail(JSONAPIBaseView, generics.RetrieveAPIView, WikiMixin):
```
brianjgeiger: Does this need to exist, or could we just use the WikiDetail view instead?

samanehsan: This seems to be the pattern for other node relations that have both a `/nodes/node_id/object/object_id/` and an `object/object_id/` route. Also, this view does an additional check to make sure that the requested wiki belongs to the node.

website/addons/wiki/model.py
```diff
@@ -175,13 +176,18 @@ class NodeWikiPage(GuidStoredObject, Commentable):
...
 
+    # For Django compatibility
+    @property
+    def pk(self):
+        return self._id
+
     @property
     def deep_url(self):
-        return '{}wiki/{}/'.format(self.node.deep_url, self.page_name)
+        return '{}wiki/{}/'.format(self.node.deep_url, urllib.quote(self.page_name))
```
brianjgeiger: Is this for our various markup safe changes?

api/nodes/views.py
brianjgeiger: We definitely do it for contributors, because they don't have a life outside of their nodes, and we don't for node children, because that would get annoying with the infinite nesting. I think we may for files, but that may just be because we don't necessarily have a guid for files. Do we have anything not weird that definitely has a guid in this situation? I just suspect it's extra tests and code that we don't need.

brianjgeiger: Logs. Logs are the closest to this, and they do not have a nodes/node_id/logs/log_id route, just the list view, because they have unique IDs that match properly. I would skip the node detail and just use the absolute detail view.

brianjgeiger: I also have a sneaking suspicion that we have too many routes for node comments based on this, but we'll probably just have to leave that be at this point.

samanehsan: Yeah, files do follow this pattern as well as Institutions and AlternativeCitations. I can't seem to find an example of a single view that is used by two routes.

brianjgeiger: I'm proposing we also ditch the route. I know I told you to do the route and for that I apologize. But I think it won't be necessary and simpler overall.

brianjgeiger:
Okay, pass done, and nearly there. 🐧 

- [ ] Change `.txt` to `txt`
- [ ] Remove redundant route and views for `/v2/nodes/[node_id]/wikis/[wiki_id]/` and point everything to `/v2/wikis/[wiki_id]/` for the detail view
- [ ] Verify that urllib.quote is the right thing there.

samanehsan:
@brianjgeiger, do you mean `/v2/nodes/[node_id]/wikis/[wiki_id]/` should use the `WikiDetail` view? Or that `/v2/nodes/[node_id]/wikis/[wiki_id]/` and the `NodeWikiDetail` view should be deleted?

website/addons/wiki/model.py
samanehsan: This fixes a bug with how the url is constructed. If there's a space in the wiki page name, there will be a space in the url since it's not being encoded. Originally, I fixed it for just one of the wiki url property methods so I changed it for these as well.

brianjgeiger:
I mean they should be deleted. And their tests should be deleted.

api/registrations/serializers.py
```diff
@@ -86,10 +86,10 @@ class RegistrationSerializer(NodeSerializer):
         related_view_kwargs={'node_id': '<pk>'}
     ))
 
-    wikis = RelationshipField(
+    wikis = HideIfWithdrawal(RelationshipField(
```
brianjgeiger: Yay for that test! I'm glad we added it in.

brianjgeiger:
What happens if you access a wiki via guid that belongs to a registration that's been withdrawn?

brianjgeiger:
Okay, pass finished. That's my only question. We may need to have a permissions class for that sort of data, but shooould be pretty easy.

samanehsan:
Good catch! The wiki detail page is still visible on withdrawn registrations. This is also the case for the comment detail and file detail routes.

brianjgeiger:
Once the Flake8 error is done, this should be good.

samanehsan:
Done fixing that error @brianjgeiger! ☔

api/wikis/serializers.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def get_size(self, obj):
+        return sys.getsizeof(obj.content)
+
+    def get_content_type(self, obj):
+        return 'text/markdown'
+
+    def get_extra(self, obj):
+        return {
+            'version': obj.version
```
sloria: Any reason `version` shouldn't be a "top-level" attribute? @brianjgeiger

sloria:
Just had one question: https://github.com/CenterForOpenScience/osf.io/pull/5610#discussion_r64267406 , but it's not a blocker. Looks good. 🐙

api/wikis/serializers.py
brianjgeiger: Yeah. We're copying the format from files, and versions in files are an extra, so we're keeping that true here as well. The goal will be, when we change wikis from being a Mongo thing to a WB thing, existing code shouldn't break.

PR #5611, opened 2016-05-04
Feature/APIv2 Create Registrations [OSF-3087]

pattisdr:
Issue https://openscience.atlassian.net/browse/OSF-3087
## Purpose

Adds ability to create registrations in the APIv2.  This is a 2-3 step process.  First, create a draft registration with mandatory registration form.  Second, update draft registration incrementally until the form is complete.  Finally, create a registration using the draft registration.

## Changes

Adds endpoints: 
-GET `/v2/nodes/{node_id}/draft_registrations/` (Gets list of all draft registrations)
-POST  `/v2/nodes/{node_id}/draft_registrations/` (Create draft registration)

-GET `/v2/nodes/{node_id}/draft_registrations/{draft_registration_id}/` (Gets a single draft registration)
-PUT/PATCH `/v2/nodes/{node_id}/draft_registrations/{draft_registration_id}/` (Update a draft registration)
-DELETE `/v2/nodes/{node_id}/draft_registrations/{draft_registration_id}/` (Delete a draft registration)

-POST `/v2/nodes/{node_id}/registrations` (Create a registration from a draft registration)
-PUT `/v2/registrations/{registration_id}/`  (Make a registration public)

-GET `/v2/metaschemas` (Get all metaschemas)
-GET `/v2/metaschemas/{metaschema_id}/` (Get a specific metaschema).


pattisdr:
@brianjgeiger, do you have any idea what might be causing these failing tests?

brianjgeiger:
No. Restarting the build to see how consistent, though I do see that it's happening regularly with your stuff. Are you up to date with develop?

pattisdr:
I was seven commits behind but up to date now.

brianjgeiger:
Okay, see if tests still pass locally.

brianjgeiger:
(So you know, the second of the three sets of unit tests are post merging-with-the-target-branch, so if that fails, it may mean that there's an incompatible but not conflicting thing in the target branch.)

pattisdr:
The third set of unit tests was failing similarly in one of Andy's PR's https://github.com/CenterForOpenScience/osf.io/pull/5543 before a change by @cwisecarver.

cwisecarver:
@pattisdr The error on [this](https://travis-ci.org/CenterForOpenScience/osf.io/builds/128873288) build is `ImportError: No module named jsonschema`

pattisdr:
:) @cwisecarver - thanks for pointing that out!

pattisdr:
Tests finally passing. Thank you @brianjgeiger and @cwisecarver for your help!

api/nodes/serializers.py
```diff
@@ -555,3 +559,97 @@ def error_checker(self, data):
     def get_absolute_url(self, obj):
         #  Citations don't have urls
         raise NotImplementedError
+
+
+class DraftRegistrationSerializer(JSONAPISerializer):
+
+    schema_choices = list(ACTIVE_META_SCHEMAS)
```
samchrisinger: The schemas should be identified by primary key not by name. Is this possible?

api/nodes/serializers.py
```diff
@@ -555,3 +559,97 @@ def error_checker(self, data):
...
+
+    def validate_metadata(self, draft, metadata):
+        """
+        Validates registration_metadata field.  Called in update and create methods because the draft is
+        needed in the context.
+        """
+        reviewer = is_prereg_admin_not_project_admin(self.context['request'], draft)
+        schema = create_jsonschema_from_metaschema(draft, is_reviewer=reviewer)
+        # Required fields are only required when creating the actual registration, not updating the draft.
+        if schema.get('required'):
```
samchrisinger: The required field should be applied on a property-by-property basis (rather than at the top-level). So it may make more sense to pass a flag into the `create_jsonschema_from_metaschema` method that toggles whether or not 'required' is ever added to the generated schema.

samchrisinger: > The required field should be applied on a property-by-property basis (rather than at the top-level). 
I hadn't read the helper method yet, please ignore this part.

api/registrations/utils.py
```diff
@@ -0,0 +1,214 @@
...
+    required = []
+
+    for page in metaschema['pages']:
+        for question in page['questions']:
+            if is_required(question):
+                required.append(question['qid'])
+            json_schema['properties'][question['qid']] = {
+                "type": "object",
+                "additionalProperties": False,
+                "properties": extract_question_values(question, is_reviewer)
```
samchrisinger: There is a slight caveat here where 'object' type questions specify on a property-by-property basis which parts are required (each leaf node determines whether or not it is required). In jsonschema you can do this:

```json
{
  "foo": {
     "type": "object",
     "properties": { ... }

  }

**Comment by pattisdr** [comment]
Something else I am also adding after talking to @brianjgeiger is the ability to update registrations, public field only.

File: api/nodes/views.py [hunk 4]
```diff
@@ -704,11 +738,209 @@ def perform_destroy(self, instance):
...
+
+    serializer_class = DraftRegistrationSerializer
+    view_category = 'nodes'
+    view_name = 'node-draft-registrations'
+
+    # overrides ListCreateAPIView
+    def get_queryset(self):
+        node = self.get_node()
+        drafts = DraftRegistration.find(Q('branched_from', 'eq', node))
+        return [draft for draft in drafts if not draft.registered_node or draft.registered_node.is_deleted]
```
samchrisinger: Could move the `not draft.registered_node` into the modm query:
`Q('registered_node', 'eq', None)`

api/nodes/views.py
```diff
@@ -74,6 +78,36 @@ def get_node(self, check_object_permissions=True):
...
+        node_id = self.kwargs['node_id']
+        if draft_id is None:
+            draft_id = self.kwargs['draft_id']
+        draft = get_object_or_error(DraftRegistration, draft_id)
+
+        if not draft.branched_from._id == node_id:
+            raise ValidationError('This draft registration is not created from the given node.')
+
+        if self.request.method not in drf_permissions.SAFE_METHODS:
+            registered_and_deleted = draft.registered_node and draft.registered_node.is_deleted
```
samchrisinger: :+1:

samchrisinger:
@pattisdr pass done

api/nodes/views.py
pattisdr: But then the modm query won't return drafts where registered_node is deleted?

api/nodes/serializers.py
pattisdr: Yes, I can change this. I remember you saying something about both primary key and name, but I thought if I took the name they specified, and then looked up the version 2 schema for that name, we would be using the correct schema.

But I will change it so primary key is used.

api/registrations/utils.py
pattisdr: Yes you're right, it will be a lot better to do it that way.

api/nodes/views.py
samchrisinger: Good point!

website/project/model.py
```diff
@@ -2149,8 +2150,10 @@ def fork_node(self, auth, title='Fork of '):
                 if forked_node is not None:
                     forked.nodes.append(forked_node)
 
-        if title == 'Fork of ' or title == '':
-            forked.title = title + forked.title
+        if title is None:
```
pattisdr: Fixes some fork issues from https://github.com/CenterForOpenScience/osf.io/pull/5534

pattisdr:
Requested changes made, @samchrisinger!

api/registrations/serializers.py
```diff
@@ -139,6 +154,44 @@ def get_registration_url(self, obj):
     def get_absolute_url(self, obj):
         return self.get_registration_url(obj)
 
+    def create(self, validated_data):
+        auth = get_user_auth(self.context['request'])
+        draft = validated_data.pop('draft')
+        registration_choice = validated_data.pop('registration_choice', 'immediate')
+        embargo_lifted = validated_data.pop('lift_embargo', None)
+
+        self.all_required_questions_answered(draft)
```
samchrisinger: Really minor semantic change-- `all_required_questions_answered` to me indicates a method returning a boolean value, when in fact it always returns a falsey value. This may be clearer as `check_required_questions_answered`

samchrisinger:
One tiny suggestion, otherwise lgtm. @pattisdr can you please move the ticket to ready-to-merge when you're ready?

pattisdr:
Moved to ready to merge!

sloria:
@pattisdr Can you update this with develop one last time?

pattisdr:
Updating again.

api/metaschemas/serializers.py
```diff
@@ -0,0 +1,22 @@
...
+
+    links = LinksField({
+        'self': 'get_absolute_url'
+    })
+
+    def get_absolute_url(self, obj):
+        return obj.absolute_api_v2_url
+
+    class Meta:
+        type_ = 'meta_schemas'
```
sloria: "Metaschemas" is one word, so this should be `metaschemas`.

api/nodes/serializers.py
```diff
@@ -640,3 +644,94 @@ def error_checker(self, data):
...
+        return obj.absolute_url
+
+    def create(self, validated_data):
+        node = validated_data.pop('node')
+        initiator = validated_data.pop('initiator')
+        metadata = validated_data.pop('registration_metadata', None)
+
+        schema_id = validated_data.pop('registration_schema').get('_id')
+        schema = get_object_or_error(MetaSchema, schema_id)
+        if schema.schema_version != 2 or schema.name not in ACTIVE_META_SCHEMAS:
```
sloria: What is the significance of version 2? Won't different schemas have different numbers of versions?

pattisdr: I was mirroring what is in https://github.com/CenterForOpenScience/osf.io/blob/develop/website/project/views/drafts.py#L259 in only wanting to use the latest version of the schema. version 2 is hardcoded many places here - I guess we currently don't have higher than 2? I could look through all the schemas with that name and make sure it is the highest number instead of using 2?

api/registrations/utils.py
```diff
@@ -0,0 +1,224 @@
...
+    metaschema = draft.registration_schema.schema
+    json_schema = base_metaschema(metaschema)
+    required = []
+
+    for page in metaschema['pages']:
+        for question in page['questions']:
+            if is_required(question) and required_fields:
+                required.append(question['qid'])
+            json_schema['properties'][question['qid']] = {
+                "type": "object",
```
sloria: As per COSDev, use singlequotes for strings.

api/nodes/serializers.py
```diff
@@ -640,3 +644,94 @@ def error_checker(self, data):
...
+    def validate_metadata(self, draft, metadata):
+        """
+        Validates registration_metadata field.  Called in update and create methods because the draft is
+        needed in the context.
+        """
+        reviewer = is_prereg_admin_not_project_admin(self.context['request'], draft)
+        # Required fields are only required when creating the actual registration, not updating the draft.
+        schema = create_jsonschema_from_metaschema(draft, required_fields=False, is_reviewer=reviewer)
+        try:
+            jsonschema.validate(metadata, schema)
```
sloria: This is pretty neat. It might be good to move this validation logic to the model, so we can hide these implementation details. The API would be something like

```python
draft.validate(metadata, ....)
```

api/nodes/serializers.py
pattisdr: We are currently forcing a version 2, anyways.

api/nodes/views.py
```diff
@@ -84,6 +88,35 @@ def get_node(self, check_object_permissions=True):
...
+
+            if draft.registered_node and not registered_and_deleted:
```
sloria: Minor: checking `draft.registered_node` is unnecessary because it's already checked in `registered_and_deleted`.

api/registrations/serializers.py
```diff
@@ -192,6 +264,13 @@ class RegistrationDetailSerializer(RegistrationSerializer):
     """
     Overrides NodeSerializer to make id required.
     """
+
+    def __init__(self, *args, **kwargs):
+        remove_fields = ['draft_registration', 'registration_choice', 'lift_embargo']
```
sloria: Removing fields in a subclass feels like improper encapsulation. Also, I'm not sure if there are side effects to accessing `self.fields` this early (this might bite us when we switch to Django models). How about using base class?

```python
class BaseRegistrationSerializer(Serializer):
    ...

class RegistrationSerializer(BaseRegistrationSerializer):
    draft_registration = ....
    registration_choice = ...

class RegistrationDetailSerializer(BaseRegistrationSerializer):
    id = IDField(..)
```

api/nodes/serializers.py
pattisdr: What do you mean by hide the implementation details?

api/nodes/serializers.py
sloria: Ah ok, I didn't realize all the schemas were on the same version. Thanks for clarifying. It might be worth using a constant, e.g. `website.project.metadata.LATEST_SCHEMA_VERSION`.

sloria:
Pass finished.

api/nodes/serializers.py
sloria: The use of the `jsonschema` library is an implementation detail in that the user of the model classes doesn't need to know about it. The user just wants to validate input data against the metaschema. 

By hiding the use of `jsonschema` behind a method, `validate_metadata(metadata)`, we can safely change the implementation (e.g. use a different jsonschema validation library) without affecting usages of `validate_metadata`.

sloria: After giving this more thought, it would make more sense for the `validate_metadata` method to go on a `MetaSchema` not on a `DraftRegistration`.

sloria: If you want, you could define it on both, by delegating:

```python
class DraftRegistration(..):
     def validate_metadata(self, *args, **kwargs):
         return self.schema.validate_metadata(*args, **kwargs)
```

but it isn't strictly necessary.

api/nodes/serializers.py
pattisdr: Do you want to replace all versions of looking up Q('schema_version', 'eq', 2).  There are numerous instances.

api/nodes/views.py
pattisdr: Well I can't say `if not registered_and_deleted` because that would mean that would require that every draft has a node that has been registered and then deleted.  I want to check if there's a registered node in the first place.

pattisdr: Maybe I just say `if draft.registered_node and not draft.registered_node.is_deleted` and don't try to use the var i defined.  I was just trying to reuse some code, but then it gets confusing.

api/registrations/serializers.py
pattisdr: Good idea.

pattisdr:
@sloria, requested changes made.

api/nodes/views.py
```diff
@@ -104,7 +104,7 @@ def get_draft(self, draft_id=None):
         if self.request.method not in drf_permissions.SAFE_METHODS:
             registered_and_deleted = draft.registered_node and draft.registered_node.is_deleted
 
-            if draft.registered_node and not registered_and_deleted:
+            if draft.registered_node and not draft.registered_node.is_deleted:
```
sloria: Minor: this is fine, but I was more suggesting you just check `if registered_and_deleted`.

pattisdr: ooohh :)

pattisdr: Wait..if_registered_and_deleted is fine though.  You're allowed to get a draft that had a registration but then it was deleted.

sloria: Ah, fair enough. 👍

sloria:
@pattisdr Can you please look into the failing test?

api/nodes/views.py
```diff
@@ -1213,9 +1481,9 @@ class NodeForksList(JSONAPIBaseView, generics.ListCreateAPIView, NodeMixin, ODMF
 
     # overrides ListCreateAPIView
     def get_queryset(self):
-        all_forks = self.get_node().forks
+        all_forks = self.get_node().forks.sort('-forked_date')
```
sloria: Good catch

website/project/model.py
```diff
@@ -118,6 +120,31 @@ def requires_consent(self):
...
+
+    # used by django and DRF
+    def get_absolute_url(self):
+        return self.absolute_api_v2_url
+
+    def validate_metadata(self, *args, **kwargs):
+        """
+        Validates registration_metadata field.
+        """
+        reviewer = kwargs.pop('reviewer', False)
```
sloria: Better to explicitly put these in the method signature rather than pulling off of kwargs. This makes the code more readable and easier for static analysis (e.g. IDEs can understand the signature better).

website/project/model.py
```diff
@@ -118,6 +120,31 @@ def requires_consent(self):
...
+    @property
+    def absolute_api_v2_url(self):
+        path = '/metaschemas/{}/'.format(self._id)
+        return api_v2_url(path)
+
+    # used by django and DRF
+    def get_absolute_url(self):
+        return self.absolute_api_v2_url
+
+    def validate_metadata(self, *args, **kwargs):
```
sloria: Would be good to have unit tests for this method, but those can be added in a follow up PR.

website/project/model.py
sloria: On a related note, `metadata` should probably be a required argument.

website/project/model.py
pattisdr: I agree that it can go in a follow-up PR. Even though I don't have any tests specifically for this new "validate_metadata" method that has been moved on the model there are still dozens of APIv2 tests for creating and updating drafts, and creating drafts from registrations, all which are testing that this functionality is working.

website/project/model.py
```diff
@@ -3980,3 +4028,9 @@ def reject(self, user):
     def add_status_log(self, user, action):
         log = DraftRegistrationLog(action=action, user=user, draft=self)
         log.save()
+
+    def validate_metadata(self, *args, **kwargs):
+        """
+        Validates draft's metadata
+        """
+        return self.registration_schema.validate_metadata(*args, **kwargs)
```
pattisdr: When I am creating or updating a draft, I am running validate_metadata before I run update_metadata.  In the instance where I am creating a draft, there won't be any metadata yet. Where I am updating the draft, the metadata will be outdated.  This only works when I am validating the metadata when creating the actual registration and the draft's metadata is not changing.

sloria: Oh ok. thanks for the clarifcation.

sloria:
Pass finished.

website/project/model.py
```diff
@@ -4029,8 +4026,8 @@ def add_status_log(self, user, action):
         log = DraftRegistrationLog(action=action, user=user, draft=self)
         log.save()
 
-    def validate_metadata(self, *args, **kwargs):
+    def validate_metadata(self, metadata, reviewer, required_fields):
```
sloria: It's not intuitive for the delegated `validate_method` to make `reviewer` and `required_fields` required arguments (when they are optional in `Metaschema.validate_metadata`). Since you're just delegating, you can just pass on *args, and **kwargs

```python
def validate_metadata(self, *args, **kwargs):
    return ...validate_metadata(*args, **kwargs)
```

PR #5620, opened 2016-05-06
hotfix/Make prereg the first option in registration template

erinspace:
## Purpose

Community reported that prereg challenge participants were often mistakenly selecting the default "open ended registration" on the registration page, and requested that the prereg challenge become the top and default option.

## Changes
- Move prereg challenge to the top of the list of schemas that renders on the registration options

## Side effects
- None on the dev side, but new registrations may now accidentally be prereg

## Ticket

https://openscience.atlassian.net/browse/OSF-6122

[#OSF-6122]


sloria:

PR #5622, opened 2016-05-06
Fix: remove dupe description text [OSF-6278]

barbour-em:
## Purpose

Fix registration supplements showing double helper text.
## Changes

`RegistrationEditor#unwrap` now appends the question description so we don't need it in the template. 
## Side effects

N/A 

## Ticket

https://openscience.atlassian.net/browse/OSF-6278

[OSF-6278]

### DIFF
--- a/website/templates/project/register.mako
+++ b/website/templates/project/register.mako
@@ -30,7 +30,6 @@
             <div data-bind="foreach: {data: page.questions, as: 'question'}">
               <div class="row">
                 <h4 data-bind="attr: {id: question.id}, text: question.title"></h4>
-                <small><em data-bind="text: question.description"></em></small>
                 <div class="col-md-12 well">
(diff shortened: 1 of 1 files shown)


website/templates/project/register.mako
```diff
@@ -30,7 +30,6 @@
             <div data-bind="foreach: {data: page.questions, as: 'question'}">
               <div class="row">
                 <h4 data-bind="attr: {id: question.id}, text: question.title"></h4>
-                <small><em data-bind="text: question.description"></em></small>
```
samchrisinger: @barbour-em this may be better to fix in the previewQuestion method-- in particular don't include description the in the preview for questions with only depth 1 in their structure.

barbour-em: I'm not sure I understand. Does that mean for non-object questions?

samchrisinger: @barbour-em sorry, I'm wrong here. 

This fix lgtm 👍

PR #5626, opened 2016-05-09
Do not log comment actions on a registered node [OSF-5961]

barbour-em:
## Purpose

In order to keep registration logs frozen, comment actions shouldn't be logged but comments should be possible. 

## Changes

Do not log comment actions if the node is a registration.

After adding, editing and deleting one comment, then adding a second:

screen shot 2016-05-09 at 10 05 03 am

## Side effects

No logs for comment actions. I tried to make sure every log that would try to reference a comment log won't error trying to reference it.

## Ticket

https://openscience.atlassian.net/browse/OSF-5961

[OSF-5961]

### DIFF
--- a/tests/test_comments.py
+++ b/tests/test_comments.py
@@ -31,7 +31,7 @@ from tests.base import (
     capture_signals
 )
 from tests.factories import (
-    UserFactory, ProjectFactory, AuthUserFactory, CommentFactory, NodeFactory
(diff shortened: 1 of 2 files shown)


barbour-em:
Closed per [this discussion.](https://openscience.atlassian.net/browse/OSF-5961)

PR #5632, opened 2016-05-10
Create wiki without restoring deleted

kch8qx:
## Purpose

Currently, if you delete a wiki file, and then create a new wiki file with the same name as the deleted file, the new file still maintains all of the edits from the old file. These can be found under the 'Compare' panel in the project wiki. (In this example, "hello" has been created, edited, deleted, and re-created)
image
The purpose of this PR is when a user deletes a project wiki, the wiki (and all of its former edits) are permanently deleted. When a new wiki is created with the same name as a deleted wiki, it will not restore the old wiki's edits. (Again "hello" has  been created, edited, deleted, and re-created, but the old edits no longer persist in "hello"'s versions)
image

## Changes

When a wiki is deleted currently, the page is just deleted from the project's wiki_pages_current. In order to prevent the deleted page from being restored when a new wiki of the same name is created, the page is now also deleted from the project's wiki_pages_versions upon deletion.

This change alone would only prevent newly deleted pages from being restored. If a user has already deleted a page before this change is made, and then creates a new wiki page of the same name, those previous edits would be restored. To fix this, this PR also includes a migration script that removes all wiki pages from a project's wiki_pages_versions that are not also pages in that project's wiki_pages_current. This prevents any page that has been deleted but does not currently exist as a project wiki from being restored when the user creates a new wiki page of the same name.

## Side effects

There is really only one case (I can think of) that this PR wouldn't resolve. For example, let's say:
1. You create a wiki page "hello"
2. You make some edits to that page
3. You delete that page
4. You create another wiki page called "hello", and those previous edits have been restored
This PR will not retroactively un-restore those edits from the first version of "hello". Instead, if you now delete "hello", then create a new "hello" after this PR, the previous edits will not persist.

## Ticket

https://openscience.atlassian.net/browse/OSF-5968


scripts/migration/migrate_deleted_wikis.py
```diff
@@ -0,0 +1,48 @@
...
+def get_targets():
+    return NodeLog.find(Q('action', 'eq', NodeLog.WIKI_DELETED))
+
+def migrate(targets, dry_run=True):
+    # iterate over targets
+    logs = targets
+    nodes = set()
+    for log in logs:
+        nodes.add(log.node)
+    for node in nodes:
```
brianjgeiger: It looks like you're doing more work than you need to here. Although I'm all for readability, you could probably just replace everything up to this point with:

```
for log in targets:
    node = log.node
```

scripts/migration/migrate_deleted_wikis.py
```diff
@@ -0,0 +1,48 @@
...
+        versions = node.wiki_pages_versions
+        current = node.wiki_pages_current
+        updated_versions = {}
+        for wiki in node.wiki_pages_versions:
```
brianjgeiger: Since you've saved `node.wiki_pages_versions` to a `versions`, go ahead and use `versions` here.

scripts/tests/test_migrate_deleted_wikis.py
```diff
@@ -0,0 +1,46 @@
...
+        super(TestMigrateDeletedWikis, self).setUp()
+        self.user = UserFactory()
+        self.auth = Auth(user=self.user)
+        self.project = ProjectFactory(creator=self.user)
+        # Update home wiki (which can't be deleted) an a second wiki
+        self.project.update_node_wiki('home', 'Hello world', self.auth)
+        self.project.update_node_wiki('second', 'Hola mundo', self.auth)
+        self.project.update_node_wiki('second', 'Hola mundo 2', self.auth)
+        # Delete the second wiki to populate targets
+        self.project.delete_node_wiki('second', self.auth)
```
brianjgeiger: Since you have the new code in delete_node_wiki, why isn't this call removing the wiki page versions?

brianjgeiger:
This needs tests for the changes you made to the actual code (rather than just the migration).

scripts/tests/test_migrate_deleted_wikis.py
```diff
@@ -0,0 +1,46 @@
...
+        # Initial targets should include: user2, user3, user4, user5, user6 (5 in total)
+        logs = get_targets()
+        # assert len is equal to 1 log (deleting 'second' wiki on project)
+        assert_equal(len(logs), 1)
+
+    def test_migrate(self):
+        # Assert 'home' has 2 versions
+        # Assert 'second' has 2 versions
+        assert_equal(len(self.versions['home']), 1)
+        assert_equal(len(self.versions['second']), 2)
```
brianjgeiger: Let's break these two asserts into a `test_pre_migration_has_expected_counts()` method to clean this up a bit.

scripts/tests/test_migrate_deleted_wikis.py
```diff
@@ -0,0 +1,46 @@
...
+        self.current = self.project.wiki_pages_current
+
+    def test_get_targets(self):
+        # Initial targets should include: user2, user3, user4, user5, user6 (5 in total)
+        logs = get_targets()
+        # assert len is equal to 1 log (deleting 'second' wiki on project)
+        assert_equal(len(logs), 1)
+
+    def test_migrate(self):
+        # Assert 'home' has 2 versions
```
brianjgeiger: You are only asserting 1 version for home (which seems appropriate, so I'm guessing it's just the comment that's inaccurate).

scripts/tests/test_migrate_deleted_wikis.py
```diff
@@ -0,0 +1,46 @@
...
+        # Assert 'second' has 2 versions
+        assert_equal(len(self.versions['home']), 1)
+        assert_equal(len(self.versions['second']), 2)
+        logs = get_targets()
+        migrate(logs, dry_run=False)
+        self.project.reload()
+        # Assert that 'home' has same versions as before
+        self.versions = self.project.wiki_pages_versions
+        assert_equal(len(self.versions['home']), 1)
```
brianjgeiger: Let's break this into a `test_migration_does_not_affect_home()` method.

scripts/tests/test_migrate_deleted_wikis.py
```diff
@@ -0,0 +1,46 @@
...
+        # need to assert that versions no longer contains 'second'
+        assert_equal(len(self.versions), 1)
+        # create another wiki with name 'second'
+        self.project.update_node_wiki('second', 'Hola mundo 3', self.auth)
+        # Make sure old versions not restored
+        assert_equal(len(self.versions['second']), 1)
```
brianjgeiger: L43-L46 could be their own test as well.

brianjgeiger:
Pass done. 🐧 To do:

- [ ] Break up the migration tests into a few more focused tests
- [ ] Add a test or two to ensure that the non-migration code works
- [ ] Simplify the beginning of the migration
- [ ] Use the variable you made in the for loop

scripts/tests/test_migrate_deleted_wikis.py
jnayak1: Since delete_node_wiki() was changed to delete the versions, 'second' should not be in self.versions. I am removing that line.

PR #5634, opened 2016-05-10
[Hotfix] Pass VOL query param to redirect_url [OSF-6161]

samanehsan:
## Purpose

The API guid redirect route throws an error when viewing a private project/file/etc. even if there is a view_only parameter included in the url.

## Changes

Add the `view_only parameter` to the redirect url.

## Side effects

None.

## Ticket

https://openscience.atlassian.net/browse/OSF-6161

### DIFF
--- a/api/guids/views.py
+++ b/api/guids/views.py
@@ -37,7 +37,10 @@ class GuidRedirect(JSONAPIBaseView):
 
     def get(self, request, **kwargs):
         url = self.get_redirect_url(**kwargs)
+        view_only_token = self.request.query_params.get('view_only')
         if url:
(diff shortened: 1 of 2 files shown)


api/guids/views.py
```diff
@@ -38,6 +39,8 @@ def get_serializer_class(self):
     def get(self, request, **kwargs):
         url = self.get_redirect_url(**kwargs)
         if url:
+            if self.request.query_params:
+                url = furl.furl(url).add(query_params=self.request.query_params)
```
sloria: `url` is a `furl` instance. You will need to use `.url` to get the string URL.

PR #5637, opened 2016-05-10
[0.69-ish] Fix text in project addon removal modal

abought:
## Purpose

Fix double escaped text in addon removal modals; name should not be rendered as `<li>figshare</li>`

## Changes
Remove redundant use of `htmlEscape`.

## Side effects
None expected. Profile addon removal modals code is different. (there may be some dead code in the latter file, but outside scope of this PR). Shouldn't have any addon names that will introdsuce display artifacts.

## Ticket
https://openscience.atlassian.net/browse/OSF-6298


abought:
Thanks. 

I've canceled the CI build to make way for someone else. :)

PR #5638, opened 2016-05-10
[Hotfix] Logical signal placement, related check-in/out fix

mfraezz:
## Purpose
De-duplicate the `contributor_removed` signal, put project-related signals in `website.project.signals`. 
Contributors that are removed individually (`remove_contributor()` instead of just `remove_contributors()`) will now have any checked-out files checked-in when removed.

## Changes
* Remove duplicate `contributor_removed` from `framework.auth.signals`
* Put `node_deleted` signal in `website.project.signals` rather than `framework.auth.signals`
* Update senders, listeners, tests to use correct signals.
* Minor: `return all(<bool_list>)` when trying to determine if any items in `<bool_list>` are `False`

## Side effects
None

## Ticket
None that I know of ¯\\\_(ツ)_/¯ 

### To test that this fix works, 

1) Have a write contributor check out a file
2) Remove the write contributor via Contributors page
3) Ensure file was checked back in.

### To test that this doesn't break anything,

A1) Make node, configure some notification settings for it. 
A2) Ensure node is present on user settings page -> Notifications -> Configure Notification Preferences
A3) Delete node.
A4) Ensure node is not present on user settings page -> Notifications -> Configure Notification Preferences

B1) Make node with multiple contributors, configure notifications for contribs (e.g. commenting)
B2) Ensure emails are sent out according to notification settings.
B3) Delete one of the contributors who would receive a notification email for an event
B4) Perform event that would trigger email
B5) Ensure: removed contributor does not receive email, any subscribed contribs still do.

C1) Check out a file as one user
C2) Merge that user into another user
C3) Ensure checkout is now owned by the merged-into user.

==========================================================================
#### Review Note
May require migration to fix any files checked-out by a now-noncontributor, though project admins can still check those in.

### DIFF
--- a/website/project/model.py
+++ b/website/project/model.py
@@ -2755,7 +2755,7 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin, Commentable):
         self.save()
 
         #send signal to remove this user from project subscriptions
-        project_signals.contributor_removed.send(contributor, node=self)
+        project_signals.contributor_removed.send(self, user=contributor)
 
         return True
 
--- a/website/project/model.py
+++ b/website/project/model.py
@@ -26,7 +26,6 @@ from framework.mongo import StoredObject
 from framework.mongo import validators
 from framework.addons import AddonModelMixin
 from framework.auth import get_user, User, Auth
-from framework.auth import signals as auth_signals
 from framework.exceptions import PermissionsError
 from framework.guid.model import GuidStoredObject, Guid
 from framework.auth.utils import privacy_info_handle
@@ -2098,7 +2097,7 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin, Commentable):
         self.deleted_date = date
         self.save()
 
-        auth_signals.node_deleted.send(self)
+        project_signals.node_deleted.send(self)
 
         return True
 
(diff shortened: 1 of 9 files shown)


sloria:
Can you add a regression test for the checkin bug? Also, can you include a brief description of what QA should look at once this is deployed?

website/project/model.py
```diff
@@ -2756,7 +2755,7 @@ def remove_contributor(self, contributor, auth, log=True):
         self.save()
 
         #send signal to remove this user from project subscriptions
-        auth_signals.contributor_removed.send(contributor, node=self)
+        project_signals.contributor_removed.send(contributor, node=self)
```
sloria: This is fine for now to keep compat, but it probably makes more sense to make `self` the sender, not the contributor.

sloria:
Pass finished.

website/project/model.py
mfraezz: Listeners on the `auth` vs `project` signals had reversed `args` order, I picked one format arbitrarily. Trivial to update if you think it makes more sense the other way.

sloria: Since contributor_removed is a project-related signal, it'd make more sense to have the project/node be the sender. Go ahead and make the change if it's trivial; otherwise nbd.

mfraezz:
Regression test added, QA notes added, `args` ordering improved. Initial (still had not started) TCI build cancelled. 

Back to you, @sloria

PR #5642, opened 2016-05-10
[Fix] Fix inconsistent redirection for "dashboard" [OSF-6272]

cslzchen:
## Purpose

When users visit the dashboard page without a cookie (deleted by browser or expired by server), OSF should redirect them to CAS for authentication. After auto/manual (depending on CAS TGC) login, users will land on the dashboard page instead of our general home page.

## Changes

- Refactor the naming for view, rule and mako of myprojects page.
- Add view and rule for dashboard page to enforce `must_be_logged_in` for correct CAS authentication and redirection.

## Side effects

No side effects

## Tickets

[OSF-6272](https://openscience.atlassian.net/browse/OSF-6272)


website/project/views/contributor.py
```diff
@@ -282,8 +282,8 @@ def project_manage_contributors(auth, node, **kwargs):
     except ValueError as error:
         raise HTTPError(http.BAD_REQUEST, data={'message_long': error.message})
 
-    # If user has removed herself from project, alert; redirect to user
-    # dashboard if node is private, else node dashboard
+    # If user has removed herself from project, alert; redirect to
+    # user's my projects page if node is private, else node dashboard
```
brianjgeiger: "node dashboard" should be "node summary".

website/project/views/contributor.py
```diff
@@ -292,7 +292,7 @@ def project_manage_contributors(auth, node, **kwargs):
         )
         if node.is_public:
             return {'redirectUrl': node.url}
-        return {'redirectUrl': web_url_for('dashboard')}
+        return {'redirectUrl': web_url_for('my_projects')}
```
brianjgeiger: Is this correct, or should it actually go to the dashboard? A question for @saradbowman or @nicipfeiffer I think. Since the default login location is the dashboard, should all default redirects be the dashboard? Is the dashboard our home page, essentially, while my_projects might be something that always has to be explicitly chosen? Or are there rules for what circumstances will direct you to the dashboard and what other circumstances take you to my_projects?

My preference is for always to the dashboard unless you click explicitly on my_projects (but my preference isn't the important one).

website/project/views/contributor.py
```diff
@@ -344,8 +344,8 @@ def project_remove_contributor(auth, **kwargs):
             raise HTTPError(http.BAD_REQUEST, data={
                 'message_long': 'Could not remove contributor.'})
 
-        # On parent node, if user has removed herself from project, alert; redirect to user
-        # dashboard if node is private, else node dashboard
+        # On parent node, if user has removed herself from project, alert; redirect to
+        # user's my projects page if node is private, else node dashboard
```
brianjgeiger: "node dashboard" -> "node summary" (I know you didn't do that, but since you're there, let's get the whole sentence consistent with modern usage)

website/views.py
```diff
@@ -101,13 +101,19 @@ def find_bookmark_collection(user):
...
 def dashboard(auth):
+    return redirect('/')
+
+
+@must_be_logged_in
+def my_projects(auth):
     user = auth.user
-    dashboard_folder = find_bookmark_collection(user)
-    dashboard_id = dashboard_folder._id
+    my_projects_folder = find_bookmark_collection(user)
```
brianjgeiger: This should be `bookmark_collection` rather than `my_projects_folder`. Same with the similar variables.

framework/auth/views.py
```diff
@@ -113,7 +113,7 @@ def forgot_password_get(auth, *args, **kwargs):
     """Return forgot password page upon.
     """
     if auth.logged_in:
-        return redirect(web_url_for('dashboard'))
+        return redirect(web_url_for('my_projects'))
```
brianjgeiger: This probably should still be `dashboard` (see below).

website/project/views/contributor.py
cslzchen: I tried to keep what it was before. The old code `web_url_for('dashboard')` before targets the `/myprojects` page. But you are right, I will ask Sara and Niki on what they think.

brianjgeiger:
Pass done. 🐧

website/project/views/contributor.py
cslzchen: Sara agrees with you. :) This should have been changed to the dashboard when we added the new dashboard. I expect test case failures for the change and I will fix them as well.

cslzchen:
Thanks.

website/project/views/contributor.py
```diff
@@ -356,7 +356,7 @@ def project_remove_contributor(auth, **kwargs):
                 redirect_url = {'redirectUrl': node.url}
             # Else stay on current page
             else:
-                redirect_url = {'redirectUrl': web_url_for('dashboard')}
+                redirect_url = {'redirectUrl': web_url_for('my_projects')}
```
cslzchen: @brianjgeiger this redirection should be changed back to `/dashboard` as well

brianjgeiger:
Just in case, be sure you're getting all the places you want to test. If you decide it's good, move it to Ready to Merge. 🎉 🐧 🎉

cslzchen:
@brianjgeiger Fixed all local test failures. Move to RTM after Travis.

website/templates/nav.mako
```diff
@@ -24,7 +24,7 @@
     <div id="navbar" class="navbar-collapse collapse navbar-right">
       <ul class="nav navbar-nav">
         % if user_name:
-            <li id="osfNavDashboard"><a href="/">Dashboard</a></li>
+            <li id="osfNavDashboard"><a href="/dashboard">Dashboard</a></li>
```
sloria: Why was this changed? Why link to `/dashboard` if the user will be redirected to `/`?

cslzchen: We want every `/dashboard` to go through the new view `dashboard` which has a @must_be_logged_in` decorator before the redirection. This can trigger consistent CAS redirection (auto or manual depending on the status OSF cookie `osf` and CAS cookie `TGC`).

sloria: OK; thanks for the clarification.

PR #5533, opened 2016-04-25
[Feature] Admin Prereg upgrade

Ghalko:
## Purpose

Upgrades to make prereg more usable 

## Changes

Based on #5478 

## Side effects

Much more flexibility in viewing and interacting with Prereg projects.

## Ticket

ADM-73, ADM-74, ADM-75, ADM-76, ADM-77

PR #5644, opened 2016-05-11
[Feature] Admin Prereg Upgrade

Ghalko:
## Purpose

Replaces #5533 

## Changes

Adding features for better prereg use.

## Side effects

None.

## Ticket

OSF-6260, OSF-6261, OSF-6262, OSF-6263, OSF-6264

### DIFF
--- a/admin/pre_reg/serializers.py
+++ b/admin/pre_reg/serializers.py
@@ -52,8 +52,12 @@ def serialize_draft_registration(draft, json_safe=True):
         'embargo': embargo,
         'registered_node': node_url,
         'status': status,
-        'logs': draft.logs,
+        'logs': map(serialize_draft_logs(draft.status_logs)),
     }
+
+
+def serialize_draft_logs(log):
+    return '{} on {}'.format(log.action, log.date)
 
 
 def get_url(draft):
--- a/admin/pre_reg/serializers.py
+++ b/admin/pre_reg/serializers.py
@@ -21,18 +21,16 @@ def serialize_user(user):
 
 # TODO: Write and use APIv2 serializer for this
 def serialize_draft_registration(draft, json_safe=True):
-    if draft.branched_from is not None:
-        url = furl(OSF_DOMAIN)
-        url.path.add(draft.branched_from.url)
-        node_url = url.url
+    node_url = get_url(draft)
+
+    embargo = get_embargo(draft, json_safe)
+
+    if draft.is_approved:
+        status = get_approval_status(draft)
+    elif draft.is_rejected:
+        status = 'Rejected'
     else:
-        node_url = None
-    registration_choice = draft.approval.meta.get('registration_choice', None)
-    if registration_choice == EMBARGO:
-        time = parser.parse(draft.approval.meta['embargo_end_date'])
-        embargo = iso8601format(time) if json_safe else time
-    else:
-        embargo = IMMEDIATE
+        status = 'Pending approval'
 
     return {
         'pk': draft._id,
@@ -53,4 +51,35 @@ def serialize_draft_registration(draft, json_safe=True):
         'title': draft.registration_metadata['q1']['value'],
         'embargo': embargo,
         'registered_node': node_url,
+        'status': status,
     }
+
+
+def get_url(draft):
+    url = furl(OSF_DOMAIN)
+    if draft.registered_node is not None:
+        url.path.add(draft.registered_node.url)
+        return url.url
+    elif draft.branched_from is not None:
+        url.path.add(draft.branched_from.url)
+        return url.url
+    return None
+
+
+def get_embargo(draft, json_safe):
+    registration_choice = draft.approval.meta.get('registration_choice', None)
+    if registration_choice == EMBARGO:
+        time = parser.parse(draft.approval.meta['embargo_end_date'])
+        return iso8601format(time) if json_safe else time
+    return IMMEDIATE
+
+
+def get_approval_status(draft):
+    if draft.registered_node is not None:
+        if draft.registered_node.is_deleted:
+            return 'Approved but canceled'
+        if draft.registered_node.retraction is None:
+            return 'Approved and registered'
+        else:
+            return 'Approved but withdrawn'
+    return 'Approved but not registered'
--- a/admin/pre_reg/serializers.py
+++ b/admin/pre_reg/serializers.py
@@ -52,6 +52,7 @@ def serialize_draft_registration(draft, json_safe=True):
         'embargo': embargo,
         'registered_node': node_url,
         'status': status,
+        'logs': draft.logs,
     }
 
 
--- a/admin/pre_reg/serializers.py
+++ b/admin/pre_reg/serializers.py
@@ -1,4 +1,5 @@
 from furl import furl
+from datetime import datetime
 
 from framework.utils import iso8601format
 from dateutil import parser
@@ -50,7 +51,7 @@ def serialize_draft_registration(draft, json_safe=True):
 
 
 def serialize_draft_logs(log):
-    return '{} on {}'.format(log.action, log.date)
+    return '{} on {}'.format(log.action, datetime.strftime(log.date, '%c'))
 
 
 def get_url(draft):
--- a/admin/pre_reg/serializers.py
+++ b/admin/pre_reg/serializers.py
@@ -52,7 +52,7 @@ def serialize_draft_registration(draft, json_safe=True):
         'embargo': embargo,
         'registered_node': node_url,
         'status': status,
-        'logs': map(serialize_draft_logs(draft.status_logs)),
+        'logs': map(serialize_draft_logs, draft.status_logs),
     }
 
 
--- a/admin/pre_reg/serializers.py
+++ b/admin/pre_reg/serializers.py
@@ -25,13 +25,6 @@ def serialize_draft_registration(draft, json_safe=True):
 
     embargo = get_embargo(draft, json_safe)
 
-    if draft.is_approved:
-        status = get_approval_status(draft)
-    elif draft.is_rejected:
-        status = 'Rejected'
-    else:
-        status = 'Pending approval'
-
     return {
         'pk': draft._id,
         'initiator': serialize_user(draft.initiator),
@@ -51,7 +44,7 @@ def serialize_draft_registration(draft, json_safe=True):
         'title': draft.registration_metadata['q1']['value'],
         'embargo': embargo,
         'registered_node': node_url,
-        'status': status,
+        'status': get_approval_status(draft),
         'logs': map(serialize_draft_logs, draft.status_logs),
     }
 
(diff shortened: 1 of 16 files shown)


website/project/model.py
```diff
@@ -3815,6 +3815,8 @@ class DraftRegistration(StoredObject):
 
     approval = fields.ForeignField('draftregistrationapproval', default=None)
 
+    status_logs = fields.DictionaryField(default=dict)
```
sloria: Please document the expected format for this field.

website/project/model.py
```diff
@@ -3879,6 +3881,13 @@ def is_rejected(self):
         else:
             return False
 
+    @property
+    def logs(self):
+        return [
+            '{} on {}'.format(v.title(), k)
```
sloria: Try to separate presentation/formatting code from the model. This should probably be a helper function, e.g. `format_status_logs(draft)`.

website/project/model.py
```diff
@@ -3918,13 +3927,19 @@ def update_metadata(self, metadata):
         self.registration_metadata.update(metadata)
         return changes
 
+    def add_status_log(self, action):
+        self.status_logs.update({
+            datetime.datetime.utcnow().strftime('%Y-%m-%d %H:%M'): action
```
sloria: Why are status logs represented as a dictionary keyed by date? Why optimize for accessing by a single date? It might make sense to follow the same modeling as NodeLog/Node.logs.

website/static/js/registrationUtils.js
```diff
@@ -11,7 +11,7 @@ var bootbox = require('bootbox');
 var moment = require('moment');
 var History = require('exports?History!history');
 
-require('js/koHelpers');
+var koHelpers = require('js/koHelpers');
```
sloria: This var isn't used anywhere. We only require `koHelpers` in order to register our custom binding handlers.

sloria:
Pass finished.

website/project/model.py
Ghalko: This is a really simple log for David just to be able to see how many stages one of these went through. It would certainly be possible to make a new model of a stripped down log, but I thought that it would be simpler to have a list, however a list really isn't supported in modm beyond using other models.

website/static/js/registrationUtils.js
Ghalko: Removed.

website/project/model.py
Ghalko: This hinges on the answer to https://github.com/CenterForOpenScience/osf.io/pull/5644#discussion_r62885659

website/project/model.py
Ghalko: This hinges on the answer to https://github.com/CenterForOpenScience/osf.io/pull/5644#discussion_r62885659

Ghalko:
@sloria should I make a new model just to keep these simple logs?

sloria:
A new model would make sense.

website/project/model.py
Ghalko: Replaced with serializer in prereg admin.

admin/pre_reg/serializers.py
```diff
@@ -21,18 +21,16 @@ def serialize_user(user):
...
-    registration_choice = draft.approval.meta.get('registration_choice', None)
-    if registration_choice == EMBARGO:
-        time = parser.parse(draft.approval.meta['embargo_end_date'])
-        embargo = iso8601format(time) if json_safe else time
+    node_url = get_url(draft)
+
+    embargo = get_embargo(draft, json_safe)
+
+    if draft.is_approved:
+        status = get_approval_status(draft)
```
mfraezz: Minor, but why not have this helper determine approval status in the other cases as well?

Ghalko: Done.

admin/pre_reg/utils.py
```diff
@@ -0,0 +1,43 @@
...
+    'n_initiator': '-initiator',
+    'title': 'title',
+    'n_title': '-title',
+    'date': 'date',
+    'n_date': '-date',
+    'state': 'state',
+    'n_state': '-state',
+}
+
+VIEW_STATUS = {
```
mfraezz: This appears unused. It's included in the context for `DraftListView`, but is unused by the template.

website/project/model.py
```diff
@@ -3782,6 +3782,30 @@ def to_json(self):
...
+    @classmethod
+    def add_log(cls, action):
+        log = cls(
+            action=action,
+        )
+        log.save()
+        return log
+
+    def __repr__(self):
+        return 'DraftRegistrationLog: {}'.format(self._id)
```
sloria: This is ok, but it should be consistent with other `__repr__` methods. It would also be helpful to include the date and action.

It should look something like

```
<DraftRegistrationLog(date=datetime.datetime(...), action='foo') with _id 'abc12'>
```

website/project/model.py
```diff
@@ -3782,6 +3782,30 @@ def to_json(self):
...
+    field - _id - primary key
+    field - date - date of the action took place
+    field - action - simple action to track what happened
+    """
+    _id = fields.StringField(primary=True, default=lambda: str(ObjectId()))
+    date = fields.DateTimeField(default=datetime.datetime.utcnow, index=True)
+    action = fields.StringField(index=True)
+
+    @classmethod
+    def add_log(cls, action):
```
sloria: This doesn't seem necessary, and it's confusingly named; you don't add logs to a DraftRegistrationLog.  It would make more sense to have `DraftRegistration#add_status_log`.

website/project/model.py
```diff
@@ -3782,6 +3782,30 @@ def to_json(self):
...
+class DraftRegistrationLog(StoredObject):
+    """ Simple log to show status changes for DraftRegistrations
+
+    field - _id - primary key
+    field - date - date of the action took place
+    field - action - simple action to track what happened
+    """
+    _id = fields.StringField(primary=True, default=lambda: str(ObjectId()))
+    date = fields.DateTimeField(default=datetime.datetime.utcnow, index=True)
+    action = fields.StringField(index=True)
```
sloria: It would be helpful to have constants for the valid values of `action` (similar to `NodeLog`).

admin/pre_reg/views.py
```diff
@@ -17,19 +18,18 @@
...
 
 class DraftListView(PreregAdmin, ListView):
     template_name = 'pre_reg/draft_list.html'
-    paginate_by = 10
-    paginate_orphans = 1
-    ordering = '-approval.initiation_date'
+    ordering = '-date'
     context_object_name = 'draft'
 
     def get_queryset(self):
```
sloria: I know the below code is from a previous PR, but there is a helper function for getting the prereg metaschema: `website.prereg.utils.get_prereg_schema`.

admin/pre_reg/utils.py
```diff
@@ -0,0 +1,36 @@
+
+SORT_BY = {
```
sloria: Why is this necessary? Couldn't the client just pass `-initiator` instead of `n_initiator`?

admin/pre_reg/views.py
```diff
@@ -54,8 +63,42 @@ def get_context_data(self, **kwargs):
                 for d in query_set
             ],
             'page': page,
+            'p': self.get_paginate_by(query_set),
+            'SORT_BY': SORT_BY,
+            'order': self.get_ordering(),
+            'status': self.request.GET.get('status', 'all'),
         }
 
+    def get_paginate_by(self, queryset):
```
sloria: The `queryset` param is unused.

website/project/model.py
```diff
@@ -3782,6 +3782,30 @@ def to_json(self):
             "text": self.text
         }
 
+
+class DraftRegistrationLog(StoredObject):
+    """ Simple log to show status changes for DraftRegistrations
+
+    field - _id - primary key
+    field - date - date of the action took place
+    field - action - simple action to track what happened
```
sloria: It might also make sense to store the user who generated the logged action.

sloria:
Pass finished.

website/project/model.py
Ghalko: Done.

website/project/model.py
Ghalko: Done.

website/project/model.py
Ghalko: Done.

admin/pre_reg/views.py
Ghalko: Done.

admin/pre_reg/views.py
Ghalko: It is a default in Django.

admin/pre_reg/utils.py
Ghalko: This is a consolidation so that the template renderer doesn't need to deal with `-` since this dict is used in sorting the lists.

website/project/model.py
Ghalko: Added

Ghalko:
@sloria This is ready for final pass/merging.

website/project/model.py
```diff
@@ -3815,6 +3841,8 @@ class DraftRegistration(StoredObject):
 
     approval = fields.ForeignField('draftregistrationapproval', default=None)
 
+    status_logs = fields.ForeignField('draftregistrationlog', list=True)
```
sloria: Rather than DraftReg having a forward-ref to DraftRegLog, why not flip the relationship (Log-> DraftReg)? This would be consistent with NodeLog -> Node.

sloria:
Pass finished. Just one suggestion: https://github.com/CenterForOpenScience/osf.io/pull/5644#discussion_r64462912

website/project/model.py
Ghalko: Consistent, yes, however, what is the general rationale for this architecture?
- Does having a list of logs imply a new set of automatic queries? If so, why do we even have this option? In what case would this ever be optimal?
- Why is a backref, assuming the previous question is reasonable, any better? Won't it load the DraftRegistration with each log loaded?
-

mfraezz: >Does having a list of logs imply a new set of automatic queries? 

Assuming I understand your question, you could still have a property on this model that just returns `DraftRegistrationLog.find(Q('owner', 'eq', self._id))` or similar. You probably shouldn't need to change `'logs': map(serialize_draft_logs, draft.status_logs)` above.

> If so, why do we even have this option? In what case would this ever be optimal?

In the case where there exist a large number of DRLogs, you gain the benefit of not needing to load each into memory when loading the DraftRegistration.

> Why is a backref, assuming the previous question is reasonable, any better? 

It's not a *backref*, exactly. It's a forward ref in the other direction.

> Won't it load the DraftRegistration with each log loaded?

Yes, which is better than loading every DRLog when the associated DraftRegistration is loaded -- one reason is above. Another reason is that the DR will likely be cached already if you're loading a DRLog; the reverse does not necessarily hold true.

sloria:
Merged into `develop`. 👍

PR #3934, opened 2015-08-04
[WIP] [Feature] Project Mailing Lists

XTech2K:
## Purpose
Allows contributors on a project to communicate to each other using email.

## Changes
* Add local mailing list info to Nodes
* Add functions using Celery to update the external mailing list on Mailgun
* Add a task to cron.py that will ensure that Celery updates were successful and fix any problems that emerge
* Add settings to the settings page for these mailing lists (enabling/disabling and subscription changes)

### UI
Currently, there is a working UI with most of the originally imagined functionality. However, after taking another look at the proposal and seeing how vague it was, I think that the current UI is not the ideal way for it to work. Some Knockout should probably be added, and in general it feels like there is not enough UI connection with related features.

UI files which were changed, but should probably be looked at to be redone in a better way:
* website/static/js/contribManager.js
* website/static/js/pages/project-settings-page.js
* website/static/js/project.js
* website/static/js/projectSettings.js
* website/templates/project/contributors.mako
* website/templates/project/project.mako
* website/templates/project/settings.mako

### Emails
While the rejection email simply needs to be looked at normally for quality, one of the sections that I did not end up being able to implement were the other emails associated with the feature:
* An email when you create a project that will have a mailing list, and will inform you of the address of the list (perhaps by sending through the list)
* Add a note about the mailing list to the add-contributor email if Mailing is enabled on that project/node

### Roll-out
Essentially this entire section of the Feature Proposal has been left undone, and the only thing that is currently coded which relate to feature roll-out is the migration script for nodes and users. A test file and class have been added, but they do not actually do much testing right now, since I wasn't able to figure out in time how exactly to imitate a pre-change node to test on.

### Project Mailing vs. Email Discussions
Currently, I have very inconsistently used the terms "Project Mailing" and "Email Discussions" throughout both the front and back ends of this feature, largely due to the issue mentioned below about future integrations vs. current clarity.

The use of both of these is fine, but a more consistent use pattern needs to be used for it; for example, "Project Mailing" could be used for functionality specific to this feature while "Discussions" could be used for things that will likely be used in the future by other parts of the larger discussions forums feature.

Alternatively, all mention of "Discussions" could be removed and simply re-applied whenever more features are added that want to integrate with this one.

*Note that more may be added before the end of the day today, August 14th, 2015*

## Important Notes
[...]

PR #5645, opened 2016-05-11
[Feature][Update] (External) Project mailing lists [OSF-3975]

mfraezz:
Updates #5033 , #3934

Purpose
=======
Allow users to email `<node_guid>@osf.io` to email all subscribed contributors.

Changes
=======

- [X] Models (`Node` updated, `MLEventLog` added, new subscription type)
- [X] Mailing List logging implemented 
- [X] Script added (node.mailing_enabled, create_list migration)
- [X] View functions / routes for ML configuration added
- [X] Tests added
- [x] UI Updated to match wireframes
- [x] Manually tested with ngrok
- [x] Celery tasks to manage remote list added
- [x] Cron job to sync lists weekly added
- [x] `mailing_list` tasks added to highest priority celery queue

UI Changes
=========
1) Mailing enabled on a node (admin view)
screen shot 2016-06-08 at 16 18 48
2) Mailing disabled on a node (admin view)
screen shot 2016-06-08 at 16 18 57
3) Contributors page
screen shot 2016-06-08 at 16 20 27
4) Dashboard modal button location
screen shot 2016-06-08 at 16 20 37
5) Modal layout
screen shot 2016-06-08 at 16 20 46

Side Effects
=========
None

Notes
--------
1 - Create Mailgun route to catch any email to `mailing_list_robot@osf\.io` using MG UI, routed to `osf.io/api/v1/mailing_list/messages/`

2 - If we want to allow users to unsubscribe via the email itself, need to:
* Enable unsubscribe tracking in MG (in `Domains`)
* Customize email template in MG to include `%mailing_list_unsubscribe_url%` (Also, probably **never** use MG for sending anything else)
* Specify a webhook to route unsubscribe info to `osf.io/api/v1/mailing_lists/hooks/unsubscribe`

3 - To migrate `mailing_enabled` for all existing nodes and create the mailing lists, run
```
python -m scripts.migrate_project_mailing_lists
```
* Optionally, append `dry` for a dry run. This will not create mailing lists on MG

## Ticket
[OSF-3975](https://openscience.atlassian.net/browse/OSF-3975)


abought:
There are a variety of new TODO items in this code. BrianG has indicated that he plans a big push to tag and address as many TODO items as possible- in support of this, could you please go through and:

- Remove the ones that are no longer relevant, 
- Resolve the ones that are blockers for this PR, and
- Create followup issues to deal with any remaining TODOs that are outside current scope (and add the JIRA ticket number next to the TODO text)

The idea is that the person who tagged the item should create the issues where possible, so we preserve some sense of context for what the comment is asking for.

scripts/migrate_project_mailing_lists.py
```diff
@@ -0,0 +1,111 @@
...
+
+EVENT = 'mailing_list_events'
+SUBSCRIPTION_TYPE = 'email_transactional'
+
+def fake(*args, **kwargs):
+    # Override actual method to prevent outgoing calls
+    return
+
+def get_targets():
+    return Node.find()
```
abought: The file docstring says this is for top-level projects, but this query doesn't seem to match. Can you clarify? Below is a comparison, using the "has no parent node" definition of top-level projects that was recently applied elsewhere in the OSF.

```
>>> Node.find().count()
62877
>>> Node.find(Q('parent_node', 'eq', None)).count()
53697
```

abought: Also, there is now a predefined query for this sort of thing:

https://github.com/abought/osf.io/blob/7f1d5c78f8811141876e8e42fb30d56ba9b67bbc/website/project/utils.py#L24-L24

In particular, it seems like this definition would create mailing lists for bookmarks, collections, registrations, and deleted nodes- is that intentional?

```
>>> Node.find(utils.TOP_LEVEL_PROJECT_QUERY).count()
3035
```

scripts/migrate_project_mailing_lists.py
```diff
@@ -0,0 +1,111 @@
...
+                        # users added on `create_list`
+
+                subscription.save()
+                node.save()
+                successful_enables.append(node._id)
+            except Exception as e:
+                unknown_failures[node._id] = e
+            else:
+                create_list(node._id)
+        if i % 100 == 0:
```
abought: What is the purpose of this section, if I might ask?

scripts/migrate_project_mailing_lists.py
mfraezz: That part of the spec was thrown out. It should apply to every node that isn't registered, deleted, or a collection.

scripts/migrate_project_mailing_lists.py
```diff
@@ -0,0 +1,111 @@
...
+
+    logger.info(
+        "Successfully disabled {0} new mailing lists for registrations/dashboards:\n{1}".format(
+            len(successful_disables), successful_disables
+        )
+    )
+
+    if unknown_failures:
+        logger.error('Handled {0} unknown exceptions while creating lists:\n{1}'.format(
+                len(unknown_failures), ['({}, {})'.format(nid, unknown_failures[nid]) for nid in unknown_failures.keys()]
```
abought: Tiny syntax note for future reference- you can use `[v for k,v in somedict.iteritems()]` in place of `[somedict[k] for k in somedict.keys()]`

It'll actually be slightly better in python 2, insofar as iteritems is an iterator, whereas I think `.keys` creates a temporary list all at once.

scripts/migrate_project_mailing_lists.py
mfraezz: Clears the cache so that it doesn't crash migrating the ~70K nodes on prod. Something similar was done when the Node/Nodelog backref removal script was run.

scripts/migrate_project_mailing_lists.py
mfraezz: >it seems like this definition would create mailing lists for bookmarks, collections, registrations, and deleted nodes- is that intentional?

Explicitly disabling it is intentional:
```
+        if node.is_registration or node.is_collection or node.is_deleted:
+            try:
+                logger.info('({0}/{1})Disabling mailing list for registration/dashboard {2}'.format(i+1, ncount, node._id))
+                node.mailing_enabled = False    
+                node.save()
+                successful_disables.append(node._id)
```

abought: > That part of the spec was thrown out. It should apply to every node that isn't registered, deleted, or a collection.

Good to know. Clarifying the docstring might be helpful- but there's still a wide gulf.

```
>>> Node.find(utils.PROJECT_QUERY).count()
6251
```

scripts/migrate_project_mailing_lists.py
mfraezz: Not terribly concerned -- this won't be hit on an ideal database. When this was last tested with prod data, there were two `unknown_failures` that turned out to be old `NodeFile` guids which were somehow grabbed.

scripts/update_project_mailing_lists.py
```diff
@@ -0,0 +1,61 @@
+# -*- coding: utf-8 -*-
+
```
abought: For posterity, a docstring describing the purpose of the script would be welcome.

website/mailing_list/model.py
```diff
@@ -0,0 +1,17 @@
...
+
+@unique_on(['destination_node', '_id'])
+class MailingListEventLog(StoredObject):
+    _id = fields.StringField(primary=True, default=lambda: str(ObjectId()))
+
+    date_created = fields.DateTimeField(default=datetime.datetime.utcnow, index=True)
+
+    email_content = fields.StringField()
+    sending_user = fields.ForeignField('user', index=True)
+    destination_node = fields.ForeignField('node', index=True)
```
abought: Note to self: pausing review here for tonight.

scripts/migrate_project_mailing_lists.py
mfraezz: As long as `Node.find(Q('is_collection', 'eq', True)).count() + Node.find(Q('is_registration', 'eq', True)).count()` sums to `7439`, I'm not concerned. It is test data, after all.

abought: Iterating over the records in python seems like a very slow way to meet this constraint... 

Though as far as I can tell, the version of TokuMx we use may be too old to support bulk operations on the database side. Whee.

website/mailing_list/utils.py
```diff
@@ -0,0 +1,422 @@
...
+    from website.models import User  # avoid circular import
+    user = User.load(user_id)
+
+    for email in user.emails:
+        resp = requests.delete(
+            '{}/{}/members/{}'.format(MAILGUN_BASE_LISTS_URL, address(node_id), email),
+            auth=('api', settings.MAILGUN_API_KEY)
+        )
+        if resp.status_code not in [200, 404]:
+            raise HTTPError(resp.status_code, data={'message_long': resp.json() or ''})
```
abought: What kind of data does mailgun put in its error response payloads?

I *think* this function only gets called in celery... but by raising an HttpError, you're sending the entire error response payload out to whoever made the request. If this function ever connects to something in the OSF website front-end, we would want to avoid sending potentially sensitive data to the website user.  (who shouldn't even have to know we're using mailgun at all, much less see a detailed mailgun error)

website/mailing_list/utils.py
```diff
@@ -0,0 +1,422 @@
...
+                # Update title if necessary
+                update_title(node_id, node.title)
+
+            if remote_subscribes != local_subscribes:
+                # Push local changes
+                members_list = jsonify_users_list(node.contributors, unsubs=get_unsubscribes(node))
+                update_multiple_users_in_list(node_id, members_list)
+
+            # Delete any noncontribs
+            from website.models import User  # avoid circular import
```
abought: I'm all for avoiding circular imports, but can we perhaps group all the dependencies at the top of the function?  (node, user, etc)

website/mailing_list/utils.py
mfraezz: >What kind of data does mailgun put in its error response payloads?

I can't say for all cases, I didn't see any example errors in [their docs](https://documentation.mailgun.com/api-mailinglists.html#mailing-lists). During testing, I saw errors like `Duplicate user` and `List does not exist`, but none that returned sensitive info.

website/mailing_list/utils.py
mfraezz: The intent was to only import them if necessary, but I can do that.

website/mailing_list/utils.py
abought: As long as there are no future plans to connect this to front-end, this seems ok then. Thanks for checking.

website/mailing_list/utils.py
```diff
@@ -0,0 +1,422 @@
...
+
+    :param str node_id: The id of the node in question
+    :param list members: List of json-formatted user dicts to add/enable/update
+    """
+    resp = requests.post(
+        '{}/{}/members.json'.format(MAILGUN_BASE_LISTS_URL, address(node_id)),
+        auth=('api', settings.MAILGUN_API_KEY),
+        data={
+            'upsert': True,
+            'members': json.dumps(members)
```
abought: According to the mailgun API, there is a limit of 1,000 members per request.

https://documentation.mailgun.com/api-mailinglists.html#mailing-lists
> <POST /lists/<address>/members.json
>> Adds multiple members, up to 1,000 per call, to a Mailing List.

For large projects, this could be problematic- especially once we start offering `project-all` lists that include child components.

Suggested improvement: all calls to the mailgun API should plan for rate limiting behavior. We are planning several large reproducibility initiatives, as well as targeting larger institutions... so this edge case is not too far-fetched.

website/mailing_list/utils.py
```diff
@@ -0,0 +1,422 @@
...
+    node.mailing_updated = True
+    node.save()
+    celery_delete_list(node._id)
+
+
+###############################################################################
+# Mailing List Helper Functions
+###############################################################################
+
+def jsonify_users_list(users, unsubs=[]):
```
abought: Avoid mutable default argument values in function definitions- this can lead to some truly *annoying* bugs.

Prefer `unsubs=()` instead.

website/mailing_list/utils.py
```diff
@@ -0,0 +1,422 @@
...
+    elif '@' in long_email:
+        return long_email.lower().strip()
+    return None
+
+def log_message(**kwargs):
+    """ Acquires and logs messages sent through Mailgun"""
+    from website.models import Node  # avoid circular imports
+    message = request.form
+    target = find_email(message['To'])
+    node = Node.load(re.search(r'[a-z0-9]*@', target).group(0)[:-1])
```
abought: It looks like `find_email` can return `None`,  which would break this regex.

https://github.com/CenterForOpenScience/osf.io/pull/5645/files#diff-0c7586e1fb715dfe41216abd7f7fd1eaR374

```
>>> import re; re.search(r'[a-z0-9]*@', None)
  File "<input>", line 1, in <module>
  File "/Users/andyboughton/code/cos/osf.io/env/lib/python2.7/re.py", line 146, in search
    return _compile(pattern, flags).search(string)
TypeError: expected string or buffer
```

website/mailing_list/utils.py
```diff
@@ -0,0 +1,422 @@
...
+
+    sender_email = find_email(message['From'])
+    sender = get_user(email=sender_email)
+
+    # Create a log of this mailing event
+    MailingListEventLog(
+        email_content=message,
+        destination_node=node,
+        sending_user=sender,
+    ).save()
```
abought: Generally for functions whose sole purpose is to create a DB object, it's helpful to return the resulting object- a nice convention of API design when building a library of code.

website/mailing_list/utils.py
```diff
@@ -0,0 +1,422 @@
...
+        if email_match:
+            return email_match.groups()[0].lower().strip()
+    elif '@' in long_email:
+        return long_email.lower().strip()
+    return None
+
+def log_message(**kwargs):
+    """ Acquires and logs messages sent through Mailgun"""
+    from website.models import Node  # avoid circular imports
+    message = request.form
```
abought: By picking the message off from the `flask.request.form` parameter (rather than acting on data passed in as an argument), you couple this function tightly to a running flask application.

In the future we may want all work to be done by an API (eg Django). When possible, I recommend writing all utility functions so that they act on data passed explicitly in- this will make it easier to decouple the code or use it in a different backend later.

website/mailing_list/utils.py
```diff
@@ -0,0 +1,422 @@
...
+        destination_node=node,
+        sending_user=sender,
+    ).save()
+
+def unsubscribe_user_hook(*args, **kwargs):
+    """ Hook triggered by MailGun when user unsubscribes.
+    See `Unsubscribes Webhook` below https://documentation.mailgun.com/user_manual.html#tracking-unsubscribes
+    for possible kwargs
+    """
+    message = request.form
```
abought: See note above; can we make this utility function not depend on flask?

website/mailing_list/utils.py
```diff
@@ -0,0 +1,422 @@
...
+    unsub = message.get('recipient')
+    mailing_list = message.get('mailing-list')
+    if not unsub or not mailing_list:
+        raise Exception()
```
abought: A more specific exception class would be helpful here. `Exception` is too generic.

website/mailing_list/utils.py
```diff
@@ -0,0 +1,422 @@
...
+    from website.models import User, NotificationSubscription  # avoid circular imports
+    user = User.find_one('username', 'eq', unsub)
```
abought: This hook should handle the case where no user matches the query. `find_one` will raise a `NoResultsFound` exception.

website/mailing_list/utils.py
```diff
@@ -0,0 +1,422 @@
...
+    node_id = mailing_list.split('@')[0]
+    subscription = NotificationSubscription.load(to_subscription_key(node_id, 'mailing_list_events'))
```
abought: If this endpoint hits a node that is no longer accessible, `.load` could return `None` and raise an exception in the next line. How should we handle this case? (eg do we just log that mailgun's picture of the OSF is out of date? Do we try to actively correct the problem?)

website/mailing_list/utils.py
```diff
@@ -0,0 +1,422 @@
...
+    subscription.add_user_to_subscription(user, 'none', save=True)
+
+def get_recipients(node):
+    if node.mailing_enabled:
+        from website.models import NotificationSubscription  # avoid circular import
+        subscription = NotificationSubscription.load(to_subscription_key(node._id, 'mailing_list_events'))
+        return [u for u in subscription.email_transactional]
```
abought: A docstring (or more specific function name) would be helpful here. For example, it's unclear why only transactional email users qualify as recipients for this function. 

Separately, is there a reason why we explicitly convert the queryset into a list before returning? Both can be iterated, but returning a queryset is "lazy"- it lets us defer the work of fetching records until later. If a subsequent check finds that we don't need the results, then we save ourselves the performance cost of the extra DB work.

website/mailing_list/views.py
```diff
@@ -0,0 +1,86 @@
...
+
+unsubscribe_user = utils.unsubscribe_user_hook
+log_message = utils.log_message
+
+def format_node_data_recursive(nodes, user):
+    items = []
+
+    for node in nodes:
+
+        can_read = node.has_permission(user, 'read')
```
abought: In place of hard-coded values, could you please use the named constant? (several places throughout project)

`from website.util.permissions import ADMIN, READ` or somesuch.

website/mailing_list/views.py
```diff
@@ -0,0 +1,86 @@
...
+    node.save()
+    return {'message': 'Successfully enabled mailing lists', 'success': True}
+
+
+@must_be_valid_project
+@must_have_permission(ADMIN)
+@must_not_be_registration
+def disable_mailing_list(node, **kwargs):
+    utils.celery_delete_list(node._id)
+    node.mailing_enabled = False
```
abought: It seems like there is a situation where this could get out of sync- would you mind walking me through how we handle the following scenario?

1. The user disables the mailing list from the web page.
2. This view immediately marks the list as disabled.
3. The delete operation fails for some unknown reason (eg Mailgun returns a 500 error)
4. Now, the OSF thinks the list is gone... but according to mailgun, the list is still there. (inconsistent state across two sources of record)

Questions:
a. Do we have a mechanism to retry deletion later if it fails?
b. Will users still be able to receive emails from a list that they thought was gone?
c. Are there any other issues that could arise here?

website/profile/views.py
```diff
@@ -224,6 +225,13 @@ def update_user(auth):
             for list_name, subscription in user.mailchimp_mailing_lists.iteritems():
                 if subscription:
                     mailchimp_utils.unsubscribe_mailchimp_async(list_name, user._id, username=user.username)
+
+            for node in user.node__contributed:
+                if node.mailing_enabled:
+                    node.mailing_updated = True
+                    node.save()
+                    celery_update_single_user_in_list(node._id, user._id, email=username, old_email=user.username)
```
abought: See question elsewhere about possible OSF/mailgun inconsistency- this scenario would apply wherever we set a flag (and schedule a celery task for later).

website/project/model.py
```diff
@@ -1573,9 +1582,35 @@ def save(self, *args, **kwargs):
...
+                subscription = NotificationSubscription(
+                    _id=to_subscription_key(self._id, 'mailing_list_events'),
+                    owner=self,
+                    event_name='mailing_list_events'
+                )
+                subscription.save()
+            except KeyExistsException:
+                subscription = NotificationSubscription.load(to_subscription_key(self._id, 'mailing_list_events'))
+            else:
+                subscription.add_user_to_subscription(self.creator, 'email_transactional', save=True)
```
abought: `else` is fired if the try block succeeds- do we really need to be saving the `subscription` twice in a row?

website/project/model.py
```diff
@@ -1641,6 +1676,7 @@ def use_as_template(self, auth, changes=None, top_level=True):
         )
 
         new.save(suppress_log=True)
+        new.add_contributor(contributor=auth.user, permissions=CREATOR_PERMISSIONS, log=False, save=False)
```
abought: When a "copy project structure" (create a new project from a template), it seems I already appear as a listed contributor, even without this line.  (eg Line 1657)

Can you clarify why this change is necessary?

website/routes.py
```diff
@@ -1596,6 +1597,53 @@ def make_url_map(app):
             json_renderer,
         ),
 
+        Rule(
+            [
+                '/project/<pid>/mailing_list/',
```
abought: Standard confirmation question: have all new routes been confirmed by the routing tsars?

website/static/js/mailingListTreebeard.js
```diff
@@ -0,0 +1,115 @@
...
+var $ = require('jquery');
+var m = require('mithril');
+var Treebeard = require('treebeard');
+var $osf = require('js/osfHelpers');
+var $tb = require('js/treebeardHelpers');
+var projectSettingsTreebeardBase = require('js/projectSettingsTreebeardBase');
+
+function subscribe(item, notification_type, reload) {
+    var id = item.data.node.id;
+    var url = '/api/v1/project/' + id + '/mailing_list/';
```
abought: I'm actually a bit surprised that we're adding new endpoints/ functionality to APIv1 at this point. Can you please clarify why it's not being written from scratch for APIv2?

(since presumably we'd expect a rewrite anyway)

Fitz was compiling a list of api v1 endpoints that are missing from apiv2. I don't think apiv2 should block this PR, but we should be sure to get these new endpoints added to his list.

website/static/js/notificationsTreebeard.js
```diff
@@ -165,16 +141,21 @@ function ProjectNotifications(data) {
...
             else {
+                options = [
+                    m('option', {value: 'none', selected : item.data.event.notificationType === 'none' ? 'selected': ''}, 'Never'),
+                    m('option', {value: 'email_transactional', selected : item.data.event.notificationType === 'email_transactional' ? 'selected': ''}, 'Instantly'),
+                ];
+                if (item.data.event.title !== 'mailing_list_events') {
+                    options.push(m('option', {value: 'email_digest', selected : item.data.event.notificationType === 'email_digest' ? 'selected': ''}, 'Daily'));
+                    options.push(m('option', {value: 'adopt_parent', selected: item.data.event.notificationType === 'adopt_parent' ? 'selected' : ''},
+                                                 'Adopt setting from parent project ' + displayParentNotificationType(item)));
+                }
```
abought: Non-blocking comment:

There seems to be a lot of near-duplicate code (see line 116). For future maintainability, I'd encourage us to DRY this if possible.

website/templates/project/project_base.mako
```diff
@@ -32,6 +32,7 @@
 ##  % endif
 % endif
 
+<%include file="modal_mailing_list_contributors.mako"/>
```
abought: The mailing list modal does not appear to be used on every project page. Putting it in the project_base template may add bloat to pages that don't (and can't) display it. I suggest adding it to the project dashboard and settings pages specifically.

website/templates/project/settings.mako
```diff
@@ -378,8 +378,31 @@
                     <div class="panel-heading clearfix">
                         <h3 class="panel-title">Email Notifications</h3>
                     </div>
+                    % if node['mailing_list_enabled'] or user['is_admin']:  ## Begin Project Mailing Lists Settings
```
abought: There are several display anomalies on the new project settings section:

1. Sorting icons are misaligned
2. Text overlaps with project title
3. The treebeard widget now has several extra columns (age actions etc); I'm not sure what the columns do, and they are not specified in the wireframes. They don't seem to update when I toggle mailing list settings either.
screen shot 2016-05-20 at 1 40 35 pm

abought:
Extremely crude first pass complete. There are a few questions to be discussed with the developer to see how edge cases are handled.

Minimum notes where action is recommended:
- [ ] Several places where error handling could be improved
- [ ] Where possible, decouple utility functions from flask by passing parameters, rather than picking values off the flask global request object
- [ ] All mailgun API calls should take into account mailgun API limits
- [ ] Some visual glitches with layout of new project settings panes in Firefox
- [ ] Notify Fitz of new endpoints for his APIv1--> APIv2 checklist

website/mailing_list/utils.py
mfraezz: It should only ever return `None` when it's used by OSF4M -- this is well structured data that comes directly from an email service provider. Still, I'll add some checks

website/mailing_list/utils.py
mfraezz: There are no data passed in as arguments. This is triggered when Mailgun routes a message to an endpoint, with information passed in via the request.

website/mailing_list/utils.py
mfraezz: See response above -- not in its current form, no.

website/mailing_list/utils.py
mfraezz: Which should never happen -- non-users won't be added to the list.

website/mailing_list/utils.py
mfraezz: Generally, yes. However, this isn't called internally, it's triggered by an email being sent to the list.

website/mailing_list/utils.py
mfraezz: If this happens, it will be logged to sentry. So, yes.

website/mailing_list/utils.py
mfraezz: > it's unclear why only transactional email users qualify as recipients for this function. 

Because the options for `mailing_list_events` are `transactional` and `none`, and users subscribed to `none` shouldn't be considered `recipients`.

website/mailing_list/views.py
mfraezz: >the following scenario

Valid concern.

>a. Do we have a mechanism to retry deletion later if it fails?

Yes, but it requires `mailing_enabled` being set to `True`. Will update.

>b. Will users still be able to receive emails from a list that they thought was gone?

Only while the delete is pending / unsuccessful

>c. Are there any other issues that could arise here?

*Here*? A large number of admins toggling list enablement status repeatedly could maybe cause queuing issues

website/profile/views.py
mfraezz: `mailing_updated` will cause it to be checked for consistency when the celerybeat task is run weekly.

website/project/model.py
mfraezz: Probably not, but we do need to here.

website/project/model.py
mfraezz: It's moved down to below the first save so that the subscription can be created (something something `_primary_key`s needing to exist before `ForeignField`s are populated. See previous PR for a more in-depth explanation).

website/routes.py
mfraezz: /me nods vigorously

website/static/js/mailingListTreebeard.js
mfraezz: This̯͓̭ͯ̚ fé̟̳̞̗̮ͅatuͫ͏̳̠͕̖͈̪̥rě̪ͣ͑̉̀͊ͫ͡ ha̲͍͚͈̠͍̎ͬ̀̃ͅs a d̴͉͙ͅe̱͗̍̈e̵ͫ̚p̵͚͙̙̜ and s̢̳̠͔̳̣̫̑̾ͣͅė̡͂͋҉͎̭̘c̵̼̫̫̺̗̗̺̲͖ͦ͆͂ͣͪͦ́r̦̥̾ͭͤ̔ͮͯ̃e̸̼͔͆͒̄͛́͟ṭ̢͍̠̥͕̙̗͚͊̓͆̎̄̌͊͂̀͟ h̤̩͎̠̭͚͈͔͍̘͔̰͇̮͛̒͋̀͐̍͐̈̀͌̄̐́̕͞i̴̵̧̲̯̝̗͓͖͌͑̈́ͧͮ̀̆͊̓̿ͯ͝S̓̉̈́̓̔ͨ̿̉̌͏̮͇̲̜̮̳̖͈͔̭̼T̸̨͕̙͇̯͆̐̈́̚͢͢ͅO̵̧̢̻̩̣͕̼̰͈̰͖̝͐̈̀ͭ̓ͅr̴̡̪̱͕̰̰̤̾͊̎͗̔̊̈ͦ̐ͯ̅ͪ͑ͩ͑͛͛̈̐͠Y̴̸̞̳̬̙͖̭͕͂͂̀͐ͧ͂̽̐̿ͮ̃̽̒̆ͫ͆̀͟. 

====

I shall ensure an update is ticketed.

website/templates/project/settings.mako
mfraezz: They weren't present several months ago when I first wrote this that part of the code. I'll do a bit of digging

website/mailing_list/model.py
mfraezz: screen shot 2016-05-20 at 16 39 38
>11:39 PM
> pausing

screen shot 2016-05-20 at 16 40 01
>11:44 PM

ಠ ͜ರೃ

website/static/js/notificationsTreebeard.js
mfraezz: Fixed in https://github.com/CenterForOpenScience/osf.io/pull/5645/commits/8729daf5d5844f694cb3ae1db068644bc87cd3e1

website/templates/project/settings.mako
mfraezz: I'm not sure that I know what "new project settings section" you're referring to.

screen shot 2016-05-20 at 23 03 58

abought:
One additional question to consider, while this is being worked on: a guid that looks like a word is a rare event, but at very least we might want a followup ticket.

Is there a list of "actual" OSF email addresses already in use? (help, support, contact, noreply@osf.io, etc) Do we have a mechanism to ensure that auto-generated project mailing list names do not overlap with those addresses? (including in the future when we move to longer guids)

Sample options include a whitelist/blacklist, or making project emailing list names more distinctive by adding a namespace prefix, eg:
`project-a1b2c@osf.io`

mfraezz:
>One additional question to consider, while this is being worked on: a guid that looks like a word is a rare event, but at very least we might want a followup ticket.

We have a long `BlacklistGuid` list that is intended prevent words or word-like strings from being generated as guids.

>Is there a list of "actual" OSF email addresses already in use?

Yes, and it has been taken into account

website/mailing_list/views.py
```diff
@@ -0,0 +1,94 @@
...
+@must_not_be_registration
+def disable_mailing_list(node, **kwargs):
+    utils.celery_delete_list(node._id)
+    node.mailing_updated = True
+    node.mailing_enabled = False
+    node.save()
+
+def flask_unsubscribe_user(*args, **kwargs):
+    message = request.form
+    utils.unsubscribe_user_hook(message)
```
abought: This is a step forward, but can still be improved a bit: The utility function should operate on data without depending on one framework (so we can support flask + django endpoints in the future). But there is no guarantee that another framework would create a `request.form` object that looks or acts the same way.

So to keep this framework agnostic, the utility function call signature should be something like `def unsubscribe_hook(recipient, message, headers)`- the view function is responsible for unpacking the form into data, and the utility function shouldn't depend on the internal structure of special flask objects at all.

scripts/migrate_project_mailing_lists.py
chrisseto:
```python
 Node._cache.data.clear()
 Node._object_cache.data.clear()
```

scripts/update_project_mailing_lists.py
```diff
@@ -0,0 +1,63 @@
...
+from website.app import init_app
+from website.mailing_list.utils import full_update
+from website.project.model import Node
+
+logger = logging.getLogger(__name__)
+
+def main(dry_run=True):
+    if dry_run:
+        def full_update(*args, **kwargs):
```
chrisseto: Does this work?
`full_update` will only be defined as an empty function in the scope of `main`.
Anywhere else it *should* be the original one from `website.mailing_list.utils`

website/mailing_list/model.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+import datetime
+
+from modularodm import fields, StoredObject
+
+from framework.mongo import ObjectId
+from framework.mongo.utils import unique_on
+
+@unique_on(['destination_node', '_id'])
```
chrisseto: Does this actually do anything?
`_id` is already unique due to being a primary key.
This will result in no constraint being applied as `_id` will always be a newly generated unique value.

website/mailing_list/model.py
```diff
@@ -0,0 +1,17 @@
...
+class MailingListEventLog(StoredObject):
+    _id = fields.StringField(primary=True, default=lambda: str(ObjectId()))
+
+    date_created = fields.DateTimeField(default=datetime.datetime.utcnow, index=True)
```
chrisseto: Only index fields that you plan on querying on. If you are not querying on these fields alone it's a wasted index.
And a note for dates. You generally want to index them in descending order rather than ascending, that way the latest dates will be the first items in the index, see `NodeLog` for an example.

mfraezz: Currently this model isn't queried on at all, it exists in preparation for possible future features / UI.

website/mailing_list/utils.py
```diff
@@ -0,0 +1,429 @@
...
+    :param list unsubs: unsubscribed users
+    """
+    members_list = [[]]
+    page = 0
+    for member in users:
+        for email in member.emails:
+            # Mailgun allows 1000 user upserts per request
+            if floor(len(members_list[page]) / 999):
+                page += 1
+                members_list[page] = []
```
abought: It might be a good idea to unit test this function.

Also, for future work, it would be preferable to have one function that flattens the list, and a separate one that breaks it into chunks. The pagination mechanism seems a bit complicated. (see http://stackoverflow.com/questions/312443/how-do-you-split-a-list-into-evenly-sized-chunks-in-python or `itertools.islice` for alternate and more readable ways of paginating a long list)

website/mailing_list/model.py
abought: Good indexing advice; thanks Ostriches.

website/project/model.py
```diff
@@ -852,6 +852,10 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin, Commentable):
     # TODO: Add validator
     comment_level = fields.StringField(default='public')
 
+    # Project Mailing
+    mailing_enabled = fields.BooleanField(default=True, index=True)
+    mailing_updated = fields.BooleanField(default=False, index=True)
```
abought: Please add a comment describing what each field is, and the difference.

(especially since there are other kinds of emails than mailing lists- the node class has too many unexplained fields as is)

scripts/update_project_mailing_lists.py
```diff
@@ -0,0 +1,63 @@
...
+            # Override actual method to prevent outgoing calls
+            return
+
+    # reload the node to ensure that it is current
+    node.reload()
+
+    assert not node.is_registration and not node.is_collection
+
+    # Reset mailing_updated now in case of a user-change during this update
+    node.mailing_updated = False
```
abought: It seems that this line confirms the update has been completed, before it actually completes. Why isn't this in the `else` block of the try/except, instead?

(in the odd case where something causes the script to fail partway through, this represents confirmation of an action before the event has actually successfully completed)

abought: As discussed on flowdock, there may be a case where another celery task triggers an action that needs to be done, while this script is running.

This could trigger an edge case where mailgun and the db were not consistent; one solution would be for the short-lived celery tasks to update the DB record when they fail, to indicate work needs to be done to clean up after a problem.

Will discuss more in person. I'm flying a bit blind without a clear sense of all the interconnected failure modes these tasks are designed to handle.

website/mailing_list/utils.py
```diff
@@ -0,0 +1,425 @@
...
+            'name': list_title(node),
+            'access_level': 'members'
+        }
+    )
+    if resp.status_code != 200:
+        raise HTTPError(resp.status_code, data={'message_long': resp.json() or ''})
+
+    members_list = []
+    members_list = jsonify_users_list(node.contributors, unsubs=get_unsubscribes(node))
+    members_list[-1].append({'address': address('mailing_list_robot'), 'subscribed': True})  # Routing robot
```
abought: If the list had one page of 1000 records, wouldn't this make 1001?

This is one argument for separating the jsonify serializer from the pagination mechanism- the serializer has no way of knowing about subsequent changes made to the page mechanism.

website/templates/project/settings.mako
abought: Confirmed anomalies are gone after recent changes.

website/mailing_list/utils.py
mfraezz: Pages intentionally stop at 999.

abought: I still think paginating the results before you've fully constructed the list of addresses is asking for trouble.

This seems like a case of a method trying to be too clever about too specific a use case. But to meet the current QA deadline, we should put a pin in this and focus our energies on the celery tasks to keep data internally consistent.

website/templates/project/settings.mako
mfraezz: ...I didn't change anything about this.

website/routes.py
```diff
@@ -1596,6 +1597,55 @@ def make_url_map(app):
             json_renderer,
         ),
 
+        # Mailing List Routes
+        # TODO: [OSF-6400] Update to APIv2
```
chrisseto: Any reason to not just go ahead with this?

website/mailing_list/utils.py
```diff
@@ -0,0 +1,426 @@
...
+ANGLE_BRACKETS_REGEX = re.compile(r'<(.*?)>')
+
+MAILGUN_BASE_LISTS_URL = '{}/lists'.format(settings.MAILGUN_API_URL)
+
+###############################################################################
+# Decorator
+###############################################################################
+
+def require_mailgun(func):
+    """ Execute MailGun API calls iff API key is set """
```
chrisseto: typo

website/mailing_list/utils.py
```diff
@@ -0,0 +1,426 @@
...
+    """ Returns information about the mailing list from Mailgun
+    :param str node_id: ID of the node in question
+    :returns dict info: mailing list info
+    """
+    resp = requests.get(
+        '{}/{}'.format(MAILGUN_BASE_LISTS_URL, address(node_id)),
+        auth=('api', settings.MAILGUN_API_KEY),
+    )
+    if resp.status_code == 200:
+        return json.loads(resp.text)
```
chrisseto: `resp.json()` is preferred

website/mailing_list/utils.py
mfraezz: http://www.merriam-webster.com/dictionary/iff

abought: More to the point:
https://xkcd.com/1033/

website/static/js/mailingListTreebeard.js
```diff
@@ -0,0 +1,115 @@
...
+    var method = notification_type === 'enabled' ? 'POST' : 'DELETE';
```
abought: I'm having trouble finding where the `get_node_mailing_list` method is used- this seems to be the only function that references the URL, and this function only does POST and DELETE requests (not `GET`).  Can you clarify?

website/static/js/mailingListTreebeard.js
abought: One smaller request: in general, we try to avoid passing hard-coded URLs into knockout code- but more so when we know a refactor/route change is coming.

Ideally the URL for this route would be passed from the backend to JS, rather than hardcoded as a string inside this function.

scripts/update_project_mailing_lists.py
```diff
@@ -0,0 +1,63 @@
...
+        logger.error(e)
+        node.mailing_updated = True
+        node.save()
+    else:
+        logger.info('Successfully updated node {}'.format(node._id))
+
+@celery_app.task(name='scripts.update_project_mailing_list')
+def run_main(dry_run=True):
+    if not dry_run:
+        scripts_utils.add_file_logger(logger, __file__)
```
abought: `scripts_utils` is undefined in this module, so the main method would never be able to run.

(There is a `scripts.utils` module, but you don't import it.)

website/mailing_list/utils.py
```diff
@@ -0,0 +1,426 @@
...
+    """
+    from website.models import Node, User  # avoid circular import
+    node = Node.load(node_id)
+
+    info, members = get_list(node_id)
+
+    if node.is_deleted:
+        delete_list(node_id)
+    elif node.mailing_enabled:
+        if 'list' in info.keys():
```
abought: You can just write `if 'list' in info` here.

Not only is it simpler, but looking up a key in a hash (dict) is faster than searching through elements of a list (`dict.keys()`).

website/mailing_list/utils.py
```diff
@@ -0,0 +1,426 @@
...
+            info = info['list']
+            remote_subscribes = [member['address'] for member in members['items'] if member['subscribed']].sort()
+            local_subscribes = [user.username for user in get_recipients(node)].sort()
+
+            if info['name'] != list_title(node):
+                # Update title if necessary
+                update_title(node_id, node.title)
+
+            if remote_subscribes != local_subscribes:
```
abought: General tip, but not a huge issue for this PR: 

Instead of comparing two sorted lists, you can use sets (hashes with faster comparisons built in)

scripts/update_project_mailing_lists.py
```diff
@@ -0,0 +1,63 @@
...
+    node.mailing_updated = False
+    node.save()
+
+    try:
+        full_update(node._id)
+    except Exception as e:
+        logger.error(
+            'Unexpected error raised when updating list of '
+            'node {}. Continuing...'.format(node._id))
+        logger.error(e)
```
abought: These two lines can be simplified to `logger.exception`

```
logging.exception(msg[, *args[, **kwargs]])
Logs a message with level ERROR on the root logger. The arguments are interpreted as for debug(), except that any passed exc_info is not inspected. Exception info is always added to the logging message. This function should only be called from an exception handler.
```

website/mailing_list/utils.py
```diff
@@ -0,0 +1,426 @@
...
+        subscription = NotificationSubscription.load(to_subscription_key(node._id, 'mailing_list_events'))
+    if subscription:
+        subscription.remove_user_from_subscription(user)
+        node.mailing_updated = True
+        node.save()
+    celery_remove_user_from_list(node._id, user._id)
+
+
+@user_confirmed.connect
+def resubscribe_on_confirm(user):
```
abought: Why is this called "resubscribe"? Is an unconfirmed user being added to the mailing list?

If so: What is the intent of subscribing someone twice? (once when added, once when they confirm an account) Could this lead to a situation where someone finds them back on a mailing list after opting out?

website/mailing_list/utils.py
```diff
@@ -0,0 +1,426 @@
...
+
+def address(node_id):
+    return '{}@{}'.format(node_id, furl(settings.DOMAIN).host)
+
+def list_title(node):
+    return '{} Mailing List'.format(node.title)
+
+def find_email(long_email):
+    # allow for both "{email}" syntax and "{name} <{email}>" syntax
+    if '<' in long_email:
```
abought: `regex.search` will return `None` if no matches are found.

Rather than using two different ways of testing the same rule, it would be a good idea to use the existing regex behavior consistently if possible.

website/static/js/mailingListTreebeard.js
mfraezz: A `GET` is used to populate the Treebeard initially

mfraezz: See the `ProjectMailingList` constructor in `project-settings-page.js`

website/static/js/pages/project-settings-page.js
```diff
@@ -13,28 +13,48 @@ require('css/addonsettings.css');
...
+                url: notificationsURL, status: status, error: error
+            });
+        });
+    }
+}
+initNotificationsTB();
+
+//Initialize treebeard grid for node admins
+var ProjectMailingList = require('../mailingListTreebeard.js');
+var mailingListSettingsURL = ctx.node.urls.api + 'mailing_list/';
```
abought: Same comment as elsewhere: can we avoid hard-coding URLs in JS, and instead pass them from front-end? It's harder to migrate usage of an api endpoint if we have trouble finding where it's used.

website/mailing_list/utils.py
```diff
@@ -0,0 +1,426 @@
...
+                'address': email,
+                'subscribed': member not in unsubs and email == member.username,
+                'vars': {'_id': member._id, 'primary': email == member.username}
+            })
+
+    # Mailgun allows 1000 user upserts per request
+    return [members_list[i:i + 999] for i in range(0, len(members_list), 999)]
+
+def address(node_id):
+    return '{}@{}'.format(node_id, furl(settings.DOMAIN).host)
```
abought: Per discussion: All mailgun configuration should go in the local settings file, so that changes made to the code do not override required mailgun configuration.

abought:
When running migration script locally, got 4 unknown errors- need to improve script logging so we can capture and address these problems.

abought:
Could you please move all mailing list scripts to `scripts/mailing_lists` and update celery usages, commands etc? That directory is getting rather messy.

abought:
During testing, we accidentally simulated celery downtime, and discovered that a project created while celery is on the fritz will never get a mailing list.

Based on reasonable experience, that's not so much an edge case as it is a Tuesday- is there a good way to handle this situation?

abought:
During local testing, no celery task was sent when creating new components, even with `MAILGUN_API_KEY` set and `USE_CELERY = True` in local.py. Did a hard restart of celery. No errors or warning. Matt is investigating.

abought:
Manually set a flag on one node and ran update script:

scripts/migrate_project_mailing_lists.py
mfraezz: @sloria instructed me to clear only those keys, not the entire `_cache`

website/mailing_list/utils.py
mfraezz: No.

```python
def subscribe_contributor_to_mailing_list(node, contributor, auth=None):
    if node.mailing_enabled and contributor.is_active:
        [...]
```

I'll update the language.

website/static/js/pages/project-settings-page.js
mfraezz:
```
Searching 1611 files for "mailing_list/"

/Users/matt/Devel/osf.io/website/routes.py:
 1604:                 '/project/<pid>/mailing_list/',
 1605:                 '/project/<pid>/node/<nid>/mailing_list/',
 1614:                 '/project/<pid>/mailing_list/',
 1615:                 '/project/<pid>/node/<nid>/mailing_list/',
 1624:                 '/project/<pid>/mailing_list/',
 1625:                 '/project/<pid>/node/<nid>/mailing_list/',
 1634:                 '/mailing_list/messages/',
 1642:                 '/mailing_list/hooks/unsubscribe/',

/Users/matt/Devel/osf.io/website/static/js/mailingListTreebeard.js:
   13:     var url = '/api/v1/project/' + id + '/mailing_list/';

/Users/matt/Devel/osf.io/website/static/js/pages/project-settings-page.js:
   43: var mailingListSettingsURL = ctx.node.urls.api + 'mailing_list/';

10 matches across 3 files
```

I'll add `TODO`s with the ticket number, to make sure these are discoverable.

abought: TODOs are a fallback option, because they assume the solution is to hard code a different URL (same bat-issue, different bat-time)- the best choice would be to pass in the URL along with any other data used to initialize the object.

That said, if the rest of the PR is ready this evening I'll try to wrap my part of the review then, and trust the person merging to make a decision on whether other pending changes to celery are blockers.

website/mailing_list/utils.py
mfraezz: Addressed in https://github.com/CenterForOpenScience/osf.io/pull/5645/commits/33d1ea37cee7b145592aa980ec22c2b4742b8641

website/static/js/pages/project-settings-page.js
mfraezz: Pending celery changes are a blocker. They're scheduled for more rigorous QA today, but preliminary evaluation with Celery Flower and RabbitMQ-management indicates that they're good to go. So, it shouldn't be a blocker for long, and the update for that will be as trivial as appending `'website.mailing_list'` and `'scripts.mailing_lists'` to the list of high-priority modules.

abought:
Ran code locally with new settings. Confirmed that celery tasks are now firing when a user is
a) Added to a project,
b) Stops subscription to project mailing list when a user turns off notifications in that project, and 

When I add an unregistered contributor, no celery task fires. Is this intentional?

abought:
This PR has developed merge conflicts; can you make sure it is updated before consideration for merge? Thanks.

abought:
> During testing, we accidentally simulated celery downtime, and discovered that a project created while celery is on the fritz will never get a mailing list.

I tried to test this by setting `USE_CELERY=False` in my configuration, saving, the hard-restarting both rabbitmq, celery, and my local OSF.... but a mailing list got created remotely on mailgun anyway.

Can we triage this question, eg
a) Whether creating a mailing list when celery is off is intentional, and
b) The remaining question wrt whether a mailing list is ever created if the first attempt fails (due to celery, mailgun errors, etc)

website/static/js/pages/project-settings-page.js
```diff
@@ -40,7 +40,7 @@ initNotificationsTB();
 
 //Initialize treebeard grid for node admins
 var ProjectMailingList = require('../mailingListTreebeard.js');
-var mailingListSettingsURL = ctx.node.urls.api + 'mailing_list/';
+var mailingListSettingsURL = ctx.node.urls.api + 'mailing_list/';  // TODO [OSF-6400]: Update to V2
```
abought: It's a start, but all this really does is move the maintenance burden down the line. 

Rather than hardcoding URLs into a single line of JS this way, our usual convention is to pass in the URL as a parameter to the consuming JS (when we instantiate the viewmodel, or whatever)

abought:
Pausing for tonight pending merge conflicts and a few questions/ errata.

website/static/js/pages/project-settings-page.js
mfraezz: This is where the viewmodel is instantiated. I can prevent the other one by passing this url in to the viewmodel.

mfraezz:
>When I add an unregistered contributor, no celery task fires. Is this intentional?

Yes. They get added to the list on all projects they're contributors for when the account is claimed.

>This PR has developed merge conflicts

I was waiting for `feature/multiple-celery-queues` to go in to develop (as we want these tasks to be high priority). Now that it has, I've updated.

>a) Whether creating a mailing list when celery is disabled is intentional,

I can't think of a situation where this would be the case. I'll sleep on it.

>b) The remaining question wrt whether a mailing list is ever created if the first attempt fails 

This could be mitigated by setting the default for `mailing_updated` to `True`, causing every newly created node to be picked up by the celerybeat job the Sunday after it's made.

scripts/mailing_lists/migrate_project_mailing_lists.py
```diff
@@ -0,0 +1,115 @@
...
+        create_list = real_create_list
+
+    successful_enables = []
+    successful_disables = []
+    unknown_failures = {}
+    nodes = get_targets()
+    ncount = len(nodes)
+    logger.info('Preparing to migrate {} nodes.'.format(ncount))
+    for i, node in enumerate(nodes):
+        if node.is_registration or node.is_collection or node.is_deleted:
```
abought: I'm a little concerned that if we add node types (registration/project/fork/collection etc) in the future, we'll end up creating mailing lists by accident, because of the need to manually check for a flag in the affirmative. 

Whether in this PR or soon, we should look for ways to make this maintainable, by centralizing the definition of what constitutes a project- scattering references to each new flag throughout the codebase is an invitation to problems down the line.

(I recently worked on a PR involving queries for projects- there were many inconsistent definitions, all scattered throughout the OSF)

abought:
I've done a very light look tonight. Requested changes appear to have been made, and the functionality I could test locally is working with the new celery priority queues. Assigning to Steve as he has requested to take a look (per MattF).

I could not test inbound mailgun hooks (like mailing list event logs) while running locally. 

Other notes:
- If there aren't yet official QA testing notes, I encourage you to write some up, and particularly to include anything that might trigger inconsistency between mailgun and the OSF- I'd feel much better if that was included in the testing! 
- The need to keep data consistent with the external service is the big challenge of this PR (esp if there's a case where the celery task may not be responsible for updating status flags).

Lastly: I know we've probably committed to the design, but I'd still strongly prefer namespacing the mailing list names somehow (eg `project-abc12@osf.io`)... the current scheme limits our ability to add new OSF email addresses (if a similar GUID exists). Effectively, we're completely reliant on a relatively opaque process (the GUID blacklist script) to avoid collisions.

scripts/mailing_lists/migrate_project_mailing_lists.py
```diff
@@ -0,0 +1,115 @@
...
+                        # users added on `create_list`
+
+                subscription.save()
+                node.save()
+                successful_enables.append(node._id)
+            except Exception as e:
+                logger.exception('Error while handling node {}'.format(node._id))
+                unknown_failures[node._id] = e
+            else:
+                create_list(node._id)
```
sloria: An uncaught HTTPError could occur here; perhaps we should add to unknown_failures or a different list altogether (`mailgun_failures`) and roll back the changes to the node.

scripts/mailing_lists/update_project_mailing_lists.py
```diff
@@ -0,0 +1,64 @@
...
+@celery_app.task(name='scripts.mailing_lists.update_project_mailing_list')
+def run_main(dry_run=True):
+    if not dry_run:
+        scripts_utils.add_file_logger(logger, __file__)
+    with TokuTransaction():
+        main(dry_run=dry_run)
+
+if __name__ == '__main__':
+    with TokuTransaction():
+        main()
```
sloria: +1 for supporting manual usage. It would also be useful to support the `--dry` argument in case we need to run this script manually.

sloria: Also, add the file logger if not in dry mode.

tests/test_models.py
```diff
@@ -3310,7 +3310,7 @@ def test_replace_contributor(self):
     def test_permission_override_on_readded_contributor(self):
 
         # A child node created
-        self.child_node = NodeFactory(parent=self.project, creator=self.auth)
+        self.child_node = NodeFactory(parent=self.project, creator=self.auth.user)
```
sloria: Good catch.

website/mailing_list/utils.py
sloria: > I raised it so that it would be caught by framework.celery_tasks.error_handler and logged to sentry. Neither this nor any of the other tasks in this file should ever be called directly by the front-end. 

Minor: If these functions are not meant to return HTTP responses to the client, then raising an `HTTPError` is a bit misleading. Might be better to use a custom error (`MailingListServiceError` or something like that). That said, I think the current implementation is fine as long as the errors get logged to sentry.

website/static/js/mailingListTreebeard.js
```diff
@@ -0,0 +1,114 @@
+'use strict';
+
+var $ = require('jquery');
+var m = require('mithril');
+var Treebeard = require('treebeard');
+var $osf = require('js/osfHelpers');
+var $tb = require('js/treebeardHelpers');
+var projectSettingsTreebeardBase = require('js/projectSettingsTreebeardBase');
+
+function subscribe(item, notification_type, reload, url) {
```
sloria: Minor: as per COSDev, use lowerCamelCase for variable names.

website/mailing_list/utils.py
```diff
@@ -0,0 +1,413 @@
...
+        destination_node=node,
+        sending_user=sender,
+    ).save()
+
+def unsubscribe_user_hook(unsub=None, mailing_list=None):
+    """ Hook triggered by MailGun when user unsubscribes.
+    See `Unsubscribes Webhook` below https://documentation.mailgun.com/user_manual.html#tracking-unsubscribes
+    for possible kwargs
+    """
+    if not unsub or not mailing_list:
```
sloria: Why not just make the arguments required?

website/profile/views.py
```diff
@@ -224,6 +225,13 @@ def update_user(auth):
             for list_name, subscription in user.mailchimp_mailing_lists.iteritems():
                 if subscription:
                     mailchimp_utils.unsubscribe_mailchimp_async(list_name, user._id, username=user.username)
+
+            for node in user.node__contributed:
```
sloria: This backref has been removed in favor of `user.contributed`.

website/static/js/treebeardHelpers.js
```diff
@@ -0,0 +1,39 @@
...
+
+function openAncestors (tb, item) {
+    var parent = item.parent();
+    if(parent && parent.id > 0) {
+        tb.updateFolder(null, parent);
+        openAncestors(tb, parent);
+    }
+}
+
+module.exports = window.$.tb = {
```
sloria: Why do you need to attach to `window.$.tb`?

scripts/mailing_lists/migrate_project_mailing_lists.py
```diff
@@ -0,0 +1,115 @@
+# -*- coding: utf-8 -*-
```
sloria: In order to run these scripts, there needs to be an `__init__.py` file in `scripts/mailing_lists/`.

scripts/mailing_lists/migrate_project_mailing_lists.py
sloria: Any risk we'll hit a rate limit? There are currently 97795 total nodes on production

sloria: Also, are you gracefully handling the state before projects are migrated? If we don't have downtime during this migration, what will users see on projects that haven't been migrated?

sloria:
Pass finished. No major blockers. Just some small changes and a few questions about the migration.

scripts/mailing_lists/migrate_project_mailing_lists.py
mfraezz: Added in https://github.com/CenterForOpenScience/osf.io/pull/5645/commits/dc99ed6994df891f29345cdd4d9b2289b5b2976c

scripts/mailing_lists/migrate_project_mailing_lists.py
mfraezz: As far as I can tell from their documentation, we shouldn't hit a rate limit. Their Webhooks and Domains APIs are rate limited, but the docs don't say that the Mailing Lists API is. As for state, that isn't explicitly handled. The simplest solution would be to just not roll out the UI changes until after the migration.

scripts/mailing_lists/migrate_project_mailing_lists.py
```diff
@@ -0,0 +1,173 @@
...
+            except Exception as e:
+                logger.exception('Error while handling node {}'.format(node._id))
+                unknown_failures[node._id] = e
+        else:
+            subscription = NotificationSubscription.load(to_subscription_key(node._id, EVENT))
+            if subscription:
+                assert node.mailing_enabled
+                logger.info('({0}/{1}) Mailing list for node {2} already enabled'.format(i, ncount, node._id))
+                successful_noops.append(node._id)
+                continue
```
sloria: What if the script is canceled or fails in the middle of subscribing contributors? Wouldn't we be bailing out prematurely here?

brianjgeiger:

PR #5646, opened 2016-05-11
[hotfix] Fix prereg form rendering files [OSF-6305]

barbour-em:
## Purpose

Fix file representations that were causing errors with the prereg form.
## Changes

Update migration script so that they fit the template.

## Side effects

Shouldn't be any

## Ticket

https://openscience.atlassian.net/browse/OSF-6305

### DIFF
--- a/scripts/migration/migrate_registration_extra.py
+++ b/scripts/migration/migrate_registration_extra.py
@@ -7,6 +7,7 @@ import logging
 
 from modularodm import Q
 from website.app import init_app
+from website.files.models import FileNode
 from scripts import utils as scripts_utils
 from website.models import Node
 from website.prereg.utils import get_prereg_schema
@@ -16,15 +17,38 @@ from framework.transactions.context import TokuTransaction
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
+def migrate_file_representation(bad_file):
+    view_url = bad_file.get('viewUrl', '')
+    fid = view_url.split('/')[-2]
+    f = FileNode.load(fid)
+    data = {
+        'data': {
+            'kind': 'file',
+            'name': bad_file['selectedFileName'],
+            'path': f.path,
+            'extra': {},
+            'sha256': f.get_version().metadata['sha256']
+        }
+    }
+    bad_file.update(data)
+    logger.info('Migrated file representation of File: {0}'.format(fid))
+
+
 def migrate_file_meta(question):
     files = question.get('extra')
+    migrated = False
+    if files and isinstance(files, list):
+        for f in files:
+            if not f.get('data', None):
+                migrate_file_representation(f)
+                migrated = True
     if isinstance(files, dict):
         if len(files) == 0:
             question['extra'] = []
         else:
             question['extra'] = [files]
-        return True
-    return False
+        migrated = True
+    return migrated
 
 
 def migrate():
@@ -46,6 +70,7 @@ def migrate():
         reg.save()
         if migrated:
             count += 1
+            logger.info('Migrated preregistration with id: {0}'.format(reg._id))
     logger.info('Done with {0} preregistrations migrated.'.format(count))
 
 if __name__ == '__main__':
--- a/scripts/migration/migrate_registration_extra_drafts.py
+++ b/scripts/migration/migrate_registration_extra_drafts.py
@@ -7,6 +7,7 @@ import logging
 
 from modularodm import Q
 from website.app import init_app
+from website.files.models import FileNode
 from scripts import utils as scripts_utils
 from website.models import DraftRegistration
 from website.prereg.utils import get_prereg_schema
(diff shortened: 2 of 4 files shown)


scripts/migration/migrate_registration_extra.py
```diff
@@ -16,15 +17,38 @@
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
+def migrate_file_representation(bad_file):
+    view_url = bad_file.get('viewUrl', '')
+    fid = view_url.split('/')[-2]
```
sloria: An `IndexError` will occur here if `view_url` is `''`; is `viewUrl` guaranteed to be in `bad_file` and will it always have a `/`? If so, then you should use `bad_file['viewUrl']` rather than `.get`.

barbour-em: All the data I found on staging has `viewUrl` and IIRC the way we used to do it required that field. I'll change it to `bad_file['viewUrl']`

scripts/migration/migrate_registration_extra.py
```diff
@@ -16,15 +17,38 @@
...
+        'data': {
+            'kind': 'file',
+            'name': bad_file['selectedFileName'],
+            'path': f.path,
+            'extra': {},
+            'sha256': f.get_version().metadata['sha256']
+        }
+    }
+    bad_file.update(data)
+    logger.info('Migrated file representation of File: {0}'.format(fid))
```
sloria: Better to log *before* making the update to bad_file, e.g. `Migrating file representation...`. That way, if the script crashes on any of the update steps, we can diagnose the `fid` that failed.

sloria:
Got an error when running the script:

```python

    exec code in run_globals
  File "/Users/sloria/projects/cos-projects/osf/scripts/migration/migrate_registration_extra.py", line 84, in <module>
    raise RuntimeError('Dry run, rolling back transaction.')
  File "/Users/sloria/projects/cos-projects/osf/scripts/migration/migrate_registration_extra.py", line 82, in <module>
    migrate()
  File "/Users/sloria/projects/cos-projects/osf/scripts/migration/migrate_registration_extra.py", line 67, in migrate
    migrated = migrate_file_meta(value)
  File "/Users/sloria/projects/cos-projects/osf/scripts/migration/migrate_registration_extra.py", line 43, in migrate_file_meta
    migrate_file_representation(f)
  File "/Users/sloria/projects/cos-projects/osf/scripts/migration/migrate_registration_extra.py", line 28, in migrate_file_representation
    'path': f.path,
AttributeError: 'NoneType' object has no attribute 'path'
```

sloria:
After moving the logging statement to the beginning of the ` migrate_file_representation` function, I get this log before the traceback:

```
Migrating file representation of File: osfstorage
```

barbour-em:
Hmmm, this might be because some viewUrls have a trailing slash while others don't?

Maybe:
```python
view_url = bad_file['viewUrl'][:-1]
fid = view_url.split('/')[-1]
```
would be better?

sloria:
Now I've updated the logging statement to log the whole file:

```python
def migrate_file_representation(bad_file):
    logger.info('Migrating file representation of File: {0}'.format(bad_file))
```

And now I get

```python
[__main__]  INFO: Migrating file representation of File: {u'viewUrl': u'/project/<snip>/files/osfstorage/56cae3499...', u'hasSelectedFile': True, u'sha256': u'...', u'selectedFileName': <snip>, u'nodeId': <snip>}
...
```

barbour-em:
Hmmm, this might be because some `viewUrl`s have a trailing slash and others don't?

Maybe 
```python
view_url = bad_file['viewUrl'].rstrip('/')
fid = view_url.split('/')[-1]
```
would be better?

sloria:
That looks right; I've changed it to

```python
def migrate_file_representation(bad_file):
    logger.info('Migrating file representation of File: {0}'.format(bad_file))
    view_url = bad_file['viewUrl'].rstrip('/')
    fid = view_url.split('/')[-1]
    f = FileNode.load(fid)
    assert f is not None, 'Could not load FileNode with id {}'.format(fid)
```

And now I get an AssertionError--looks like an invalid fid--investigating now..

sloria:
Hm, the fid that the script fails on looks valid but doesn't exist in the storedfilenode collection; are deleted files handled?

sloria:
Ah sure enough--the fid *does* exist in the trashedfilenode collection. So should we try to load from trashedfilenode if we can't find it in storedfilenode?

barbour-em:
If we think we should allow users to attached trashedfilesnodes then I would say yes but I don't know enough about the relationship between them to say either way.

sloria:
What is the alternative?

barbour-em:
If the FileNode can't be loaded, remove the `bad_file` dict entirely? If a user deleted a FileNode, I can't imagine they'd expect it to stay attached to a draft.

sloria:
For drafts, perhaps, but the above error is happening in the `migrate_registration_extra`. How should existing registrations be handled? It's possible that a user deleted a file after a project is registered.

barbour-em:
Existing registrations should try to load from trashed, I don't think we should be removing something from an already registered project.

sloria:
I agree. TrashedFileNode does not implement `get_version`, so you'll need to figure out a different way to get the correct sha256, ie. the sha for the registered version. Can you look into this?

PR #5648, opened 2016-05-11
[featurefix] Fix missing file names in log templates

felliott:
## Purpose

File/folder names are not being displayed for osfstorage file delete actions:

screen shot 2016-05-11 at 1 32 30 pm

because the expected display string isn't being assigned to the `data-bind`'s `text` attribute.

## Changes

Add `text: ` to `data-bind` in the osfstorage delete template and the dataverse study_linked template.

## Side effects

Nope.

## Ticket

No ticket.


sloria:
Good catch

PR #5655, opened 2016-05-16
add a check for files

hmoco:
## Purpose

Prevent ` folder has no download count ` error to be triggered in sending osf4m queued mails.

## Changes

Makes the presend find the first file, rather than the first whatever under osfstorage root.


sloria:
Is this bug on master? If so, please make this a hotfix.

PR #5656, opened 2016-05-16
A banner to notify of possible problems with adblock on analytics page [OSF-6307]

SSJohns:
## Purpose

fix JIRA problem OSF-6307

## Changes

A banner shows up if the user has adblock on the analytics page

## Side effects

There shouldn't be any.

## Ticket

### DIFF
--- a/website/static/js/ads.js
+++ b/website/static/js/ads.js
@@ -1,0 +1,1 @@
+var canRunAds = true;
--- a/website/static/js/ads.js
+++ b/website/static/js/ads.js
@@ -1,1 +1,6 @@
+/* http://stackoverflow.com/questions/4869154/how-to-detect-adblock-on-my-website
+ * Most adblocks will see that a file with ads in the title is attempting to be
+ * imported and block it. In this case the variable won't exist, which means we will
+ * enter the if.
+*/
 var canRunAds = true;
--- a/website/templates/project/statistics.mako
+++ b/website/templates/project/statistics.mako
@@ -5,6 +5,22 @@
   <h2 class="text-300">Analytics</h2>
 </div>
 
+<div id="hiddenFlag"></div>
+<script src="/static/js/ads.js"></script>
+<script>
+  if( window.canRunAds === undefined ){
+    var banner = document.createElement('div');
+    banner.className += 'm-b-md p-md osf-box-lt box-round';
+    var node = document.createTextNode("The use of adblocking software may prevent site analytics from loading properly. For more information go");
+    banner.appendChild(node);
+    var link = document.createElement('a');
+    link.href = "https://openscience.atlassian.net/browse/NCP-821";
+    link.innerHTML = " here";
+    banner.appendChild(link);
+    document.getElementById('hiddenFlag').appendChild(banner);
+  }
+</script>
+
 <%
     if user['is_contributor']:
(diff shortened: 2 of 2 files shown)


website/templates/project/statistics.mako
```diff
@@ -5,6 +5,23 @@
   <h2 class="text-300">Analytics</h2>
 </div>
 
+<div id="hiddenFlag"></div>
+<script src="/static/js/ads.js"></script>
+<script>
+  if( window.canRunAds === undefined ){
+    var banner = document.createElement('div');
```
brianjgeiger: This is a little unwieldy, because building html in javascript is hard to read. However, the general idea seems good. Instead, make a CSS class called `.ad-hidden` or similar. Put the HTML in directly with the class of `.ad-hidden`. For the javascript, if `window.canRunAds` then change the style of .ad-hidden to `display: hidden`. Does that make sense?

website/static/js/ads.js
```diff
@@ -0,0 +1 @@
+var canRunAds = true;
```
brianjgeiger: Is there some documentation about why this works? If so, add a comment with a link to that documentation (a stack overflow question or ad-blocker FAQ or whatever) so if it stops working in the future we'll have an idea where to check.

website/templates/project/statistics.mako
```diff
@@ -5,6 +5,23 @@
...
+    banner.className += 'm-b-md p-md osf-box-lt box-round';
+    var node = document.createTextNode("The use of adblocking software may prevent site analytics from loading properly. For more information go");
+    banner.appendChild(node);
+    var link = document.createElement('a');
+    link.href = "https://openscience.atlassian.net/browse/NCP-821";
```
brianjgeiger: This link won't be useful for anyone who doesn't work here (try opening it in an incognito window to see). Check through http://help.osf.io to see if there's a page you can link to there, and if not, ask @saradbowman if she has a recommended place to link.

brianjgeiger:
Good start, but needs a couple changes. Pass done. 🐧

website/templates/project/statistics.mako
```diff
@@ -5,6 +5,24 @@
   <h2 class="text-300">Analytics</h2>
 </div>
 
+<div id="hiddenFlag"></div>
+<script src="/static/js/ads.js"></script>
+<script>
+  if( window.canRunAds === undefined ){
+    var banner = document.createElement('div');
+    banner.className += 'm-b-md p-md osf-box-lt box-round text-center';
```
brianjgeiger: Not quite what I was saying. You have the hiddenFlag div above, so just put the content in there rather than making it with Javascript.

In base.css, you could make a .adBlockVisible class. That could look something like:
```css
.ad_block_visible: {
    display: hidden
}
```

Then your hidden flag div could look like
```html
<div id="hiddenFlag" class='ad_block_visible m-b-md p-md osf-box-lt box-round text-center' >
    The use of adblocking software may prevent site analytics from loading properly.
    <!-- Look at <a href="http://help.osf.io/page/does/not/exist/yet">our help pages</a> for more info. -->
</div>
````

And your javascript would change display: hidden to display: block. Something in that range? This way HTML is HTML, CSS is CSS, and all the javascript does is make it visible.

brianjgeiger:
Pass done. 🐧

PR #5657, opened 2016-05-16
Allow users to read license information on a node [OSF-6292]

brianjgeiger:
## Purpose

Allow users of the v2 API to access license information for a node and details about available licenses. (Read only)

## Changes

1. Added `/v2/licenses/` and `/v2/licenses/[license-id]` routes to show the licenses available to the user
2. Added a license relationship to the node serializer that would link to the above licenses
3. Added node_license attribute to the node serializer that would give year and copyright holder information related to the attached license
4. Ensured that withdrawn registrations do not serialize license information
5. Ensure that the license relationship can be viewed with anonymized VOL but not the node_license data
6. Allow default query on ODM Filter Views to be `None`
7. Add license read scope
8. Documentation and tests

screen shot 2016-05-16 at 7 33 33 pm

screen shot 2016-05-16 at 7 33 11 pm

## Side effects

Shouldn't be any.

## Ticket

https://openscience.atlassian.net/browse/OSF-6292

### DIFF
--- a/api/licenses/views.py
+++ b/api/licenses/views.py
@@ -87,4 +87,3 @@ class LicenseList(JSONAPIBaseView, generics.ListAPIView, ODMFilterMixin):
     def get_queryset(self):
         queryset = NodeLicense.find(self.get_query_from_request())
         return queryset
-
--- a/api/licenses/views.py
+++ b/api/licenses/views.py
@@ -1,0 +1,90 @@
+from rest_framework import generics, permissions as drf_permissions
+from framework.auth.oauth_scopes import CoreScopes
+
+from api.base.filters import ODMFilterMixin
+from api.base import permissions as base_permissions
+from api.base.utils import get_object_or_error
+from api.licenses.serializers import LicenseSerializer
+from api.base.views import JSONAPIBaseView
+
+from website.project.licenses import NodeLicense
+
+
+class LicenseDetail(JSONAPIBaseView, generics.RetrieveAPIView):
+
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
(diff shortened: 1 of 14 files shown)


api/licenses/views.py
```diff
@@ -0,0 +1,89 @@
...
+
+    #This Request/Response
+
+    """
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        base_permissions.TokenHasScope,
+    )
+
+    required_read_scopes = [CoreScopes.LICENSE_READ]
```
mfraezz: Note: discussed in person, these endpoints should have an `ALWAYS_PUBLIC` scope, which will be implemented in OSF-6360. Fine for now.

mfraezz:
Please add tests to ensure "`Changes`" points 4 and 5 above:

> 4\. Ensured that withdrawn registrations do not serialize license information
> 5\. Ensure that the license relationship can be viewed with anonymized VOL but not the node_license data

Otherwise, LGTM. Back to you @brianjgeiger

brianjgeiger:
# 4 test: there's a registration test that ensures that only specific fields are visible when the registration is withdrawn. That catches these problems automatically rather than writing individual tests which would, if someone forgot to decorate the field, also cause forget to write the test. So it's covered.

brianjgeiger:
Number five is handled by a similar situation, where we have to define the list of things that get shown in a VOL. There's a bunch of infrastructure around it already, so I *think* testing each field individually would be not super useful, but you might check me on that.

mfraezz:
Discussed in person, above concerns are addressed. Dropping this in RTM :shipit:

PR #5661, opened 2016-05-17
Indentation matters

samchrisinger:
# Purpose

Improper indentation in the `find_question` util was breaking things.

# Changes

Fix the indentation

### DIFF
--- a/website/archiver/utils.py
+++ b/website/archiver/utils.py
@@ -227,8 +227,8 @@ def find_question(schema, qid):
             q['qid']: q
             for q in page['questions']
         }
-    if qid in questions:
-        return questions[qid]
+        if qid in questions:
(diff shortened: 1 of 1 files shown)


website/archiver/utils.py
```diff
@@ -221,14 +221,22 @@ def find_registration_files(values, node):
...
         questions = {
             q['qid']: q
             for q in page['questions']
         }
-    if qid in questions:
-        return questions[qid]
+        if root in questions:
+            item = questions[root]
+    title = item.get('title')
+    while len(path):
```
sloria: Won't this be an infinite loop? I don't see anything that changes the length of path.

samchrisinger: 😿

PR #5664, opened 2016-05-17
Fix checking for wikis with no ids

samanehsan:
## Purpose

Fixes regression where wiki comment view model was getting instantiated even if there was no wiki.

## Changes
Only instantiate the wiki comment view model if `window.contextVars.wiki.wikiID !== null`.

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-6349

Fix [#OSF-6349]


samanehsan:
Sent to the release branch: https://github.com/CenterForOpenScience/osf.io/pull/5665

PR #5667, opened 2016-05-18
Wiki help bar that adds syntax guidelines [OSF-2080]

SSJohns:
## Purpose

Addition of a question button to the wiki markup bar in order toi help users navigate to the page describing how markdown is done. When pressed it just pops-up with a modal box that users will see and get a link from.

## Changes

Edited the Markdown.Editor.js to add the button and give it a data-target that points to the wiki-bar-modal-help.mako page which is what will popup, Also changed the edit.mako page to include the wiki-bar-modal-help.mako page when the page is loaded.
## Side effects

If the edit bar is used anywhere else in the code then the bar will pop up with a link to the wiki pop up markdown page. Talked to QA and they said the only place that the edit bar is used is the wiki so it should be fine.

## Ticket

https://openscience.atlassian.net/browse/OSF-2080

### DIFF
--- a/website/addons/wiki/static/pagedown-ace/Markdown.Editor.js
+++ b/website/addons/wiki/static/pagedown-ace/Markdown.Editor.js
@@ -1656,6 +1656,34 @@ var Range = ace.require('ace/range').Range;
                 buttonRow.appendChild(button);
                 return button;
             };
+            var makeHelpButton = function(id,title,XShift){
+              var button = document.createElement("li");
+              //Based off the search button classes
+              //<button type=button class="btn osf-search-btn" data-toggle="modal" data-target="#search-help-modal"><i class="fa fa-question fa-lg"></i></button>
+              //$(".dropdown-toggle").attr("data-toggle", "dropdown");
+              button.className = "wmd-button";
+              // change those attributes
+              button.setAttribute('data-toggle','modal');
+              button.setAttribute('data-target','#wiki-help-modal');
+
+              togg = button.getAttribute('data-toggle');
+              targ = button.getAttribute('data-target');
+              console.log(togg);
+              console.log(targ);
+              //button.data.target = "#wiki-help-modal"
+              button.style.left = xPosition + "px";
+              xPosition += 25;
+              var buttonImage = document.createElement("span");
+              button.id = id + postfix;
+              button.appendChild(buttonImage);
+              button.title = title;
+              button.XShift = XShift;
+              //if (textOp)
+              //    button.textOp = textOp;
+              setupButton(button, true);
+              buttonRow.appendChild(button);
+              return button;
+            }
             var makeCheckBox = function (div_id, cb_id, XShift, text) {
                 var li = document.createElement("li");
                 li.id = div_id;
@@ -1716,7 +1744,7 @@ var Range = ace.require('ace/range').Range;
             makeCheckBox("wmd-autocom-toggle", "autocom", "-240px", "Autocomplete");
 
             makeSpacer(4);
-            buttons.help = makeButton("wmd-help-button", getStringAndKey("help"), "-240px", bindCommand("helpPopUp"));
+            buttons.help = makeHelpButton("wmd-help-button", getStringAndKey("help"), "-240px");
 
             //helpOptions = true;
 
@@ -2466,3 +2494,10 @@ var Range = ace.require('ace/range').Range;
 
 
 })();
+//<button type=button class="btn osf-search-btn" data-toggle="modal" data-target="#search-help-modal"><i class="fa fa-question fa-lg"></i></button>
+//$(".dropdown-toggle").attr("data-toggle", "dropdown");
+$(function(){
+  console.log($("wmd-help-button"));
+  $("wmd-help-button").attr("data-toggle","modal");
+  $("wmd-help-button").attr("data-target","#wiki-help-modal");
+});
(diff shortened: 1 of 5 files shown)


website/addons/wiki/static/pagedown-ace/Markdown.Editor.js
```diff
@@ -1649,6 +1656,34 @@ var Range = ace.require('ace/range').Range;
                 buttonRow.appendChild(button);
                 return button;
             };
+            var makeHelpButton = function(id,title,XShift){
+              var button = document.createElement("li");
+              //Based off the search button classes
+              //<button type=button class="btn osf-search-btn" data-toggle="modal" data-target="#search-help-modal"><i class="fa fa-question fa-lg"></i></button>
```
brianjgeiger: Probably we don't need this explanation text. The reason we added it to the other PR was because we were pulling from something done externally, and it was something that was based on logic that's not necessarily obvious. In this case, you're just doing something like we do it elsewhere on the site, so you don't need to document it.

Our dev guide might help to decide when to comment, but generally rarely. https://cosdev.readthedocs.io/en/latest/style_guides/python.html#on-comments

website/addons/wiki/static/pagedown-ace/Markdown.Editor.js
```diff
@@ -1649,6 +1656,34 @@ var Range = ace.require('ace/range').Range;
...
+              //$(".dropdown-toggle").attr("data-toggle", "dropdown");
+              button.className = "wmd-button";
+              // change those attributes
+              button.setAttribute('data-toggle','modal');
+              button.setAttribute('data-target','#wiki-help-modal');
+
+              togg = button.getAttribute('data-toggle');
+              targ = button.getAttribute('data-target');
+              console.log(togg);
```
brianjgeiger: Be sure to remove console logging before making a pull request.

website/addons/wiki/static/pagedown-ace/Markdown.Editor.js
```diff
@@ -1649,6 +1656,34 @@ var Range = ace.require('ace/range').Range;
...
+              console.log(targ);
+              //button.data.target = "#wiki-help-modal"
```
brianjgeiger: Commented out code should be removed.

website/addons/wiki/static/pagedown-ace/Markdown.Editor.js
```diff
@@ -1649,6 +1656,34 @@ var Range = ace.require('ace/range').Range;
...
+              var button = document.createElement("li");
+              //Based off the search button classes
+              //<button type=button class="btn osf-search-btn" data-toggle="modal" data-target="#search-help-modal"><i class="fa fa-question fa-lg"></i></button>
+              //$(".dropdown-toggle").attr("data-toggle", "dropdown");
+              button.className = "wmd-button";
+              // change those attributes
+              button.setAttribute('data-toggle','modal');
+              button.setAttribute('data-target','#wiki-help-modal');
+
+              togg = button.getAttribute('data-toggle');
```
brianjgeiger: Whole words are a good idea with variables. Partially because there's no character limit, so if it helps readability and isn't excessive, then yay! Also partially because people using IDEs will get spell check squiggles if it's not made up of real words, so no need for extra squiggles.

website/addons/wiki/static/pagedown-ace/Markdown.Editor.js
```diff
@@ -2441,5 +2482,22 @@ var Range = ace.require('ace/range').Range;
...
+          message:
+              '<p>Wiki uses the Markdown snytax. This gives you many options, but can be very simple as well. ' +
+              'To see more information and examples go to our <a href="http://help.osf.io/m/collaborating/l/524109-using-the-wiki">guides</a>' +
+              '</p>'
+      });
+    }
+
 
 })();
+//<button type=button class="btn osf-search-btn" data-toggle="modal" data-target="#search-help-modal"><i class="fa fa-question fa-lg"></i></button>
```
brianjgeiger: No commented out code here (there are some lines I skipped, so be sure to check everywhere).

website/addons/wiki/templates/wiki-bar-modal-help.mako
```diff
@@ -0,0 +1,18 @@
+<div class="modal fade" id="wiki-help-modal" tabindex="-1" role="dialog" aria-labelledby="myModalLabel" aria-hidden="true">
+        <div class="modal-dialog">
+            <div class="modal-content">
+                <div class="modal-header">
+                    <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true"></span>&times;</button>
+                    <h3 class="modal-title">Wiki Syntax Help</h3>
+                </div>
+                <div class="modal-body">
+                  <p>
+                    Wiki uses the Markdown snytax. This gives you many options, but can be very simple as well. To see more information and examples go to our <a href="http://help.osf.io/m/collaborating/l/524109-using-the-wiki">guides</a>
```
brianjgeiger: Why is there this *and* a bootbox?

website/static/js/ads.js
```diff
@@ -0,0 +1,6 @@
+/* http://stackoverflow.com/questions/4869154/how-to-detect-adblock-on-my-website
+ * Most adblocks will see that a file with ads in the title is attempting to be
+ * imported and block it. In this case the variable won't exist, which means we will
+ * enter the if.
+*/
+var canRunAds = true;
```
brianjgeiger: This shouldn't be here. This was the problem you were having with branches? You'll want to get this cleared up. It will be a bit of a pain to do, I'm afraid.

brianjgeiger:
By and large, it's looking good. Get rid of the things that don't belong in this branch, ditch the commented out code and other extra comments, and make the variable names a bit more descriptive. I'll give it another look after that. Functionally, though, it seems good so far. Pass complete. 🐧

PR #5668, opened 2016-05-18
Make Github delete confirmation modal match standard delete confirmation modal [OSF-6346]

cameronblandford:
## Purpose

The purpose of this PR is to resolve [[OSF-6346]](https://openscience.atlassian.net/browse/OSF-6346): a style inconsistency stemming from the fact that the GitHub add-on uses a different fangorn helper than the other add-ons. The specific problem was had three parts:
1. A discrepancy in modal button style
2. Incorrect placement of the modal header content (placed in body instead of in header)
3. If a non-GitHub file foo.bar was deleted earlier in the same session, "Delete foo.bar?" would be in the modal header. Otherwise, the header would be empty.

## Changes

1. To fix button style, I changed the classes of the buttons to align with what's in `fangorn.js`.
2. To fix the placement of the header content, I moved the header text to the modal .update() function, consistent with `fangorn.js`.
3. This part was resolved via change 2 - By updating the header before each display of the modal, no old modal headers are now shown.

## Side effects

* None that I can foresee.

### Before

image

### After

image
image

## Ticket

[[OSF-6346]](https://openscience.atlassian.net/browse/OSF-6346)

### DIFF
--- a/website/addons/github/static/githubFangornConfig.js
+++ b/website/addons/github/static/githubFangornConfig.js
@@ -17,7 +17,6 @@ function _uploadUrl(item, file) {
     return waterbutler.buildTreeBeardUpload(item, file, {branch: item.data.branch});
 }
 
-// TODO: Refactor, repeating from core function too much
 function _removeEvent (event, items) {
     var tb = this;
     function cancelDelete() {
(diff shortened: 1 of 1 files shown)


brianjgeiger:
Minor items: for the future, anything that is branched off of develop make your branch `feature/branch_name` instead of just `branch_name` per gitflow https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow/ . Also go ahead an make a descriptive title for the branch (can be similar to or the same as the Jira ticket title) before the [OSF-1234] token (I've changed this one already).

brianjgeiger:
Onces tests pass, this is good to merge. 🎉 🐧 🎉

PR #5670, opened 2016-05-18
Improvements for the analytics dashboard [OSF-6348]

jeffreyliu3230:
## Purpose

Improvements for the analytics dashboard.

## Changes

- Query results are now saved to the admin db.
- Added user count history for each product and timeseries visualization.
- Added daily new users count and repetitive actions user age (average).
- Added constraint (date_disabled: null) to user count.
- Updated visualization design and layout.
- jquery for hiding or revealing keen analytics tabs on the dashboard.

## Ticket

https://openscience.atlassian.net/browse/OSF-6348


felliott:
Could you resubmit this with just the relevant commits against `develop`?  I've done this in my repo so I can look at it (https://github.com/felliott/osf.io/tree/feature/query-improvement).  You're welcome to use that or do it yourself, whichever you prefer.  Cleaning up the PR will make reviewing the changes much easier.

felliott:
Is there a way to squash the django migrations in the PR into one?  Since they haven't been applied to production yet, there shouldn't be a need for intermediate steps (I think?).

felliott:
Two issues with `admin.sales.utils.get_user_analytics`:  

1.) you can get the appropriate query for active users from https://github.com/CenterForOpenScience/osf.io/blob/750125e9cbaaf51ca55938ab1380166e133801d8/scripts/analytics/benchmarks.py#L25
2.) You may need to check that the timezone for the `date_registered` is being specified.  IIRC, python doesn't set timezones on certain datetime calls, meaning the since date will be dependent on the timezone of the server.

brianjgeiger:

PR #5671, opened 2016-05-18
[Feature]Registrations with Dataverse fail [OSF-5801]

TomBaxter:
## Purpose:
[OSF-5801](https://openscience.atlassian.net/browse/OSF-5801)
Fix failing Dataverse registrations

## Changes:
Update website/archiver/tasks.py to avoid duplicate copy calls (draft and published versions were stumbling over each other)

## Side effects:
None

[#OSF-5801]

### DIFF
--- a/website/archiver/tasks.py
+++ b/website/archiver/tasks.py
@@ -201,6 +201,8 @@ def archive_addon(addon_short_name, job_pk, stat_result):
     addon_name = addon_short_name
     if 'dataverse' in addon_short_name:
         addon_name = 'dataverse'
+        revision = 'latest' if addon_short_name.split('-')[-1] == 'draft' else 'latest-published'
+        folder_name_suffix = 'draft' if addon_short_name.split('-')[-1] == 'draft' else 'published'
     create_app_context()
     job = ArchiveJob.load(job_pk)
     src, dst, user = job.info()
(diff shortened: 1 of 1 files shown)


website/archiver/tasks.py
```diff
@@ -201,6 +201,8 @@ def archive_addon(addon_short_name, job_pk, stat_result):
     addon_name = addon_short_name
     if 'dataverse' in addon_short_name:
         addon_name = 'dataverse'
+        revision = 'latest' if addon_short_name.split('-')[-1] == 'draft' else 'latest-published'
```
brianjgeiger: Don't change this, because it's done elsewhere in the addon and is out of scope, but getting the published status from the short name would be better as a function than by taking the last item from the split just in case it changes at some point, then it's both explicit and localized to change. I just mention because it gives me something to talk about.

brianjgeiger:
Okay, this looks great. Note for Steve: There'll be a hotfix along shortly for this in case, you know, you want a hotfix instead.

brianjgeiger:
🎉 🐧 🎉 

Also, hotfix at https://github.com/CenterForOpenScience/osf.io/pull/5698

website/archiver/tasks.py
TomBaxter: Agreed! I was thinking that if we get around to refactoring this, that we should treat draft and published as separate addon instances throughout the code base, in both WaterButler and OSF. As it is they are in an awkward in between. Which I assume is what led to this bug in the first place.

brianjgeiger: Cool. Could you make an improvement ticket in jira (with the component Maintenance) with you thoughts on a potential refactor?

TomBaxter:
Cancelling. Code added via hotfix #5698

PR #5672, opened 2016-05-18
"End embargo early" log has incorrect outdated language [OSF-6341]

el1t:
## Purpose

The formatting of embargo end logs is inconsistent with other logs. This fixes the trailing period.
screen shot 2016-05-18 at 2 38 32 pm

## Changes

- ~~Added words to the log template to match other logs~~
- Removed the trailing period
- ~~Change node to `registered_from_id` (previously registration id)~~

## Side effects

None

## Ticket

[OSF-6341 "End embargo early" log has incorrect outdated language](https://openscience.atlassian.net/browse/OSF-6341)

### DIFF
--- a/website/project/model.py
+++ b/website/project/model.py
@@ -3081,7 +3081,7 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin, Commentable):
         :param permissions: A string, either 'public' or 'private'
         :param auth: All the auth information including user, API key.
         :param bool log: Whether to add a NodeLog for the privacy change.
-        :param bool meeting_creation: Whther this was creayed due to a meetings email.
+        :param bool meeting_creation: Whether this was created due to a meetings email.
         """
         if auth and not self.has_permission(auth.user, ADMIN):
             raise PermissionsError('Must be an admin to change privacy settings.')
--- a/website/templates/log_templates.mako
+++ b/website/templates/log_templates.mako
@@ -39,7 +39,7 @@ initiated an embargoed registration of
 <script type="text/html" id="embargo_terminated_no_user">
 Embargo for
 <a class="log-node-title-link overflow" data-bind="text: nodeTitle, attr: {href: nodeUrl}"></a>
-ended.
+ended
 </script>
 
 ## Retraction related logs


brianjgeiger:
There are two places that logs are generated. One is in the project summary page, which is using the widget that you updated. The other is in the my_projects page, which uses a newer widget. Verify that the change is made in that widget as well.

website/project/model.py
```diff
@@ -3484,7 +3484,7 @@ def terminate_embargo(self, auth):
             action=NodeLog.EMBARGO_TERMINATED,
             params={
                 'project': self._id,
-                'node': self._id,
+                'node': self.registered_from_id,
```
brianjgeiger: Because you're changing the format of the saved log information, this will require a migration as well.

brianjgeiger:
Pass done. Looks good so far, but there's a couple of things still left to do. 🐧

brianjgeiger:
Hold off on writing the migration until Sara has commented on the Jira ticket. It's possible that at least one of the changes is inappropriate.

website/project/model.py
brianjgeiger: Based on review from Product, revert this line and don't worry about doing a migration.

el1t:
I've pushed to revert the link, although github is taking a while to update.

website/templates/log_templates.mako
```diff
@@ -37,9 +37,8 @@ initiated an embargoed registration of
 </script>
 
 <script type="text/html" id="embargo_terminated_no_user">
-Embargo for
-<a class="log-node-title-link overflow" data-bind="text: nodeTitle, attr: {href: nodeUrl}"></a>
-ended.
+Embargo of registration of
```
brianjgeiger: I think we're reverting this language change, so yeah, just the `.` and fixing the typos in the docstring above should do it.

brianjgeiger:
Nearly there. The conversation on the ticket wouldn't have made as much sense without some context, so I've been specific. Pass done. 🐧

el1t:
Pushed and squashed, only change is the trailing period 😄

brianjgeiger:
Perfect. Pass done. 🎉 🐧 🎉

sloria:
Failing build is unrelated to this change. Good to merge.

PR #5673, opened 2016-05-18
[Feature] Multiple celery queues [#DEV-100]

mfraezz:
## Purpose
Allow celery tasks to be queued by priority.

## Changes
* Pin to a fork of Kombu with a fix that is not yet in a stable release
* Default queue now defined explicitly
* Three queues added (`low`, `med`, `high`)
* Automatic task router added. Note: future tasks will need their modules added to a priority level, or will be put in the default queue between `low` and `med`
* Tasks with names that didn't follow the `module.path.to.function` convention fixed.

## Side effects
Names of commented-out celerybeat tasks (for easier local dev) will need to be updated manually on staging/prod to queue those tasks properly (`LOW` rather than `DEFAULT`). 

## Ticket

[#DEV-100](https://openscience.atlassian.net/browse/DEV-100)

### DIFF
--- a/framework/celery_tasks/routers.py
+++ b/framework/celery_tasks/routers.py
@@ -1,34 +1,8 @@
 # -*- coding: utf-8 -*-
-from website.settings import DEFAULT_QUEUE, LOW_QUEUE, MED_QUEUE, HIGH_QUEUE
-
-LOW_PRI_MODULES = {
-    'framework.analytics.tasks',
-    'framework.celery_tasks',
-    'scripts.osfstorage.usage_audit',
-    'scripts.osfstorage.glacier_inventory',
-    'scripts.analytics.tasks',
-    'scripts.osfstorage.files_audit',
-    'scripts.osfstorage.glacier_audit',
-    'scripts.populate_new_and_noteworthy_projects',
-    'website.search.elastic_search',
-}
-
-MED_PRI_MODULES = {
-    'framework.email.tasks',
-    'scripts.send_queued_mails',
-    'scripts.triggered_mails',
-    'website.mailchimp_utils',
-    'website.notifications.tasks',
-}
-
-HIGH_PRI_MODULES = {
-    'scripts.approve_embargo_terminations',
-    'scripts.approve_registrations',
-    'scripts.embargo_registrations',
-    'scripts.refresh_box_tokens',
-    'scripts.retract_registrations',
-    'website.archiver.tasks',
-}
+from website.settings import (
+    DEFAULT_QUEUE, LOW_QUEUE, MED_QUEUE, HIGH_QUEUE,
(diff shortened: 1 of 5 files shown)


framework/celery_tasks/routers.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
+from website.settings import DEFAULT_QUEUE, LOW_QUEUE, MED_QUEUE, HIGH_QUEUE
+
+LOW_PRI_MODULES = {
```
cwisecarver: I might move these into the settings file if it won't cause a horrendous circular import. All these have to be modified when a new task is added, just like the list of celery task. It might be more discoverable if they're all in the same place. Just a suggestion.

requirements.txt
```diff
@@ -44,6 +44,10 @@ git+https://github.com/CenterForOpenScience/modular-odm.git@develop
 # Python markdown extensions for comment emails
 git+git://github.com/aleray/mdx_del_ins.git
 
+# Kombu with the ability to specify queue priority
```
cwisecarver: I'm sure you tested this but kombu, celery, and rabbitmq tend to have pretty tight version binding. I've upgraded rabbitmq before which broke celery, then upgraded celery (which didn't specify a dependency on a specific version of kombu) which then broke the upgraded celery forcing me to manually upgrade kombu. That was a fun four hours I'll never get back.

cwisecarver:
lgtm. 

It might be worth thinking about some configuration that specifies the number of celery workers per queue. That'll probably have to go in the private repo but one of us could work with you to set it up. Something else to think about is putting long running tasks in separate queues from short running ones. Long running tasks should probably have more workers than short running ones. It'll help with queues filling up when there aren't enough workers to handle everything in the queue.

requirements.txt
mfraezz: It's the same version of Kombu that we were using, with one additional commit cherry-picked in: https://github.com/celery/kombu/commit/c20f85446926cf0cfdadaa196113835282a33612

It seemed not to break anything for me, and I used both the default `amqp` and `rabbitmqctl` to verify these queues were being created with the proper priority levels.

requirements.txt
```diff
@@ -44,6 +44,10 @@ git+https://github.com/CenterForOpenScience/modular-odm.git@develop
...
+# TODO: Remove this when Kombu has a stable release including commit c20f854
+git+git://github.com/mfraezz/kombu.git@v3.0.36
```
sloria: We're trying to move away from using forks on devs' accounts. When you get a chance, can you transfer ownership of your fork to me so I can then transfer it to the COS org account?

sloria:
Created `feature/multiple-celery-queues` (consider this merged). Will get this onto staging2 soon.

PR #5678, opened 2016-05-18
A banner to notify of possible problems with adblock on analytics page [OSF-6307]

SSJohns:
## Purpose

fix JIRA problem OSF-6307

## Changes

A banner shows up if the user has adblock on the analytics page

## Side effects

There shouldn't be any.

## Ticket


website/templates/project/statistics.mako
```diff
@@ -5,6 +5,16 @@
   <h2 class="text-300">Analytics</h2>
 </div>
 
+<div id="hiddenFlag" class="ad_block_display">
+  <div class="m-b-md p-md osf-box-lt box-round text-center">The use of adblocking software may prevent site analytics from loading properly.</div>
+</div>
+<script src="/static/js/ads.js"></script>
+<script>
+  if( window.canRunAds === undefined ){
+    document.getElementById('hiddenFlag').style.display = "block";
```
brianjgeiger: If you replace out https://github.com/CenterForOpenScience/osf.io/pull/5678/files#diff-8b5fde9e5d5d19f47a0415fa762de65bR6 with this line, and instead of `document.getElementById('hiddenFlag')` you use `$.('.ad_block_display)`, then you can entirely remove this script block. That means that anything that has the class `ad_block_display` and that imports the ads.js script will do what you want it to do. Nice and reusable.

Are you familiar with jQuery? If not, take a look at it before implementing the change. It's super-handy and great for maintaining cross-browser compatibility.

brianjgeiger:
Pass done. 🐧  I know it's a lot of revision cycles, but it'll be very simple and reusable once it's all done.

website/templates/project/statistics.mako
SSJohns: Can I ask what the link goes to? It just keeps going to this file for me and do you mean line 14? Are these two different things you are saying I should do or is this all one thing?

Also I am familiar with jquery but I was sure you still needed to put it into a script tag in order for it to work. I do not think jquery can be run in html like you are suggesting.

brianjgeiger: When you click the link, look for the gold highlighted line. It points to line 6 in ads.js.

brianjgeiger:
Okay, are you super-tired of this issue and ready to move on, or would you like to explore other front-end technologies that will make it better (and potentially prepare you for other issues). I'm good to do either.

brianjgeiger:
Actually, I'm not even sure it would be better. So the problem that I've introduced with my last suggestion is that I've made you flip when it's hidden vs. when it's shown, so there'll be a "flash" when the page loads that will show the error and then remove it when the js loads.

I think your previous way will be a better experience, even if it's a bit more work. So keep the jquery but otherwise go back to where it was hidden by default and you check for the variable. So almost exactly what you had before, but with `$('.ad_block_display').` rather than `document.getElementById('hiddenFlag').` to select the element. 

I apologize for all the back and forth.

brianjgeiger:
🎉 🐧 🎉

website/static/css/style.css
```diff
@@ -692,3 +692,7 @@ button.close {
 div.ball-scale-blue>div {
     background-color: #BED1E2;
 }
+
+.ad_block_display{
```
sloria: As per COSDev, use `lowercase-and-dashes` for CSS classes.
https://cosdev.readthedocs.io/en/latest/style_guides/html_css.html

website/templates/project/statistics.mako
```diff
@@ -5,6 +5,16 @@
   <h2 class="text-300">Analytics</h2>
 </div>
 
+<div id="hiddenFlag" class="ad_block_display">
+  <div class="m-b-md p-md osf-box-lt box-round text-center">The use of adblocking software may prevent site analytics from loading properly.</div>
+</div>
+<script src="/static/js/ads.js"></script>
+<script>
+if( window.canRunAds === undefined ){
+  $('.ad_block_display').css({'display':'block'});
```
sloria: Better to select on ID for specificity. You can also use `$(..).show()`.

brianjgeiger: This one was my suggestion for reusability. If we want to do the same sort of thing with adblocking, there's no reason why we wouldn't do it to everything on that page that we want shown.

sloria:
Just minor changes to make; I'll make these changes when I merge.

sloria:
We already have a helper CSS class for hiding content, `.scripted`. I'll use that when I make my changes.

SSJohns:
Ah, okay. I didn't know that. The suggestion to make another css class wasn't mine. I originally did it all in a single js function.

sloria:
Merged.

PR #5679, opened 2016-05-18
Feature/Add read only Identifier views to the API [#OSF-6291]

erinspace:
## Purpose

- John's Hopkins needed routes to access identifiers via the API - this PR adds those routes. Access a node's identifiers or node detail according to its unique identifier. Right now, identifiers are only able to be assigned for registrations.

## Identifier List
```GET /v2/nodes/[node_id]/identifiers/```

screen shot 2016-05-24 at 12 31 08 pm

### Filter Identifier List by category

    /v2/registrations/<node_id>/identifiers/?filter[category]=doi

## Identifier Detail
```GET /v2/identifiers/[identifier_id]/```

screen shot 2016-05-24 at 12 32 17 pm

### Identifier JSON
```
{
    "relationships": {
        "referent": {
            "links": {
                "related": {
                    "href": "http://localhost:8000/v2/registrations/j32ys/",
                    "meta": {}
                }
            }
        }
    },
    "links": {
        "self": "http://localhost:8000/v2/identifiers/5735f648669aec1069278b2d/"
    },
    "attributes": {
        "category": "doi",
        "value": "10.5072/FK2OSF.IO/J32YS"
    },
    "type": "identifiers",
    "id": "5735f648669aec1069278b2d"
}
```

## Changes

- Add url to access identifiers in node routes
- Add Identifier URLs to access individual identifiers
- add Identifiers serializer
- add IdentifierList view
- add IdentifierDetail view
- tests for these two
- Filter for identifiers list by category
- Move identifier factory to more central place
- Allow Identifier factory to specificy category to create more than one
- sneak in @felliott 's change to swagger docs

## Side effects
- None I can think of!

## Ticket

https://openscience.atlassian.net/browse/OSF-6291

### CODE REVIEW

File: api/base/settings/defaults.py [hunk 1]
```diff
@@ -168,8 +168,8 @@
 
 # TODO: Revisit methods for excluding private routes from swagger docs
 SWAGGER_SETTINGS = {
+    'api_path': '/',
```
brianjgeiger: 👍

api/base/urls.py
```diff
@@ -24,7 +24,8 @@
                 url(r'^docs/', include('rest_framework_swagger.urls')),
                 url(r'^institutions/', include('api.institutions.urls', namespace='institutions')),
                 url(r'^collections/', include('api.collections.urls', namespace='collections')),
-                url(r'^guids/', include('api.guids.urls', namespace='guids'))
+                url(r'^guids/', include('api.guids.urls', namespace='guids')),
+                url(r'^identifiers/', include('api.identifiers.urls', namespace='identifiers'))
```
brianjgeiger: Go ahead and add a comma at the end of this item. I know, tiny thing. If this ends up being it for the CR, feel free to ignore. :)

brianjgeiger:
Could you add to the Changes section a copy of the JSON generated by this PR's changes and maybe a screenshot of the relevant documentation?

api/identifiers/views.py
```diff
@@ -0,0 +1,98 @@
...
+    ##Query Params
+
+     Identifiers may be filtered by their category.
+
+    #This Request/Response
+
+    """
+
+    permission_classes = (
+        IsPublic,
```
brianjgeiger: Is this accurate? I don't think we use this much, so I don't have a good feel for when we would use it. Is this saying that we can only read identifiers for public resources?

brianjgeiger: Just looked up the code. Looks like it means IsViewable rather than IsPublic. I may have to make a ticket to rename that.

api/identifiers/views.py
```diff
@@ -0,0 +1,98 @@
...
+    required_read_scopes = [CoreScopes.IDENTIFIERS_READ]
+    required_write_scopes = [CoreScopes.NULL]
+
+    serializer_class = IdentifierSerializer
+    view_category = 'identifiers'
+    view_name = 'node-identifier-detail'
+
+    def get_object(self):
+        identifier = self.kwargs['node_identifier']
+        return Identifier.find_one(Q('value', 'eq', identifier))
```
brianjgeiger: Do you have the object's ID here? Because if so, use `load` rather than `find_one` because it will be much faster.

api_tests/identifiers/views/test_identifier_list.py
```diff
@@ -0,0 +1,74 @@
...
+        paths = [
+            urlparse.urlparse(
+                item['relationships']['referent']['links']['related']['href']
+            ).path for item in self.data
+        ]
+        assert_in('/{}registrations/{}/'.format(API_BASE, self.registration._id), paths)
+
+    def test_identifier_list_returns_correct_categories(self):
+        categories = [identifier.category for identifier in self.all_identifiers]
+        categories_in_response = [identifier['attributes']['category'] for identifier in self.data]
```
brianjgeiger: Since you aren't specifying a sort order, my guess is that if the default sort order changes, this test will fail. Might be good to tighten this up just a bit.

api_tests/identifiers/views/test_identifier_list.py
```diff
@@ -0,0 +1,74 @@
...
+
+        assert_items_equal(categories_in_response, categories)
+
+    def test_identifier_list_returns_correct_values(self):
+        categories = [identifier.value for identifier in self.all_identifiers]
+        categories_in_response = [identifier['attributes']['identifier']['self'] for identifier in self.data]
```
brianjgeiger: Same with this on the sort order.

framework/auth/oauth_scopes.py
```diff
@@ -92,6 +94,9 @@ class ComposedScopes(object):
     # Guid redirect view
     GUIDS_READ = (CoreScopes.GUIDS_READ, )
 
+    # Identifier views
+    IDENTIFIERS_READ = (CoreScopes.IDENTIFIERS_READ, )
```
brianjgeiger: Add this into `NODE_METADATA_READ` as well.

tests/factories.py
```diff
@@ -771,6 +772,20 @@ def _create(cls, *args, **kwargs):
...
+class IdentifierFactory(ModularOdmFactory):
+    class Meta:
+        model = Identifier
+
+    referent = SubFactory(RegistrationFactory)
+    value = Sequence(lambda n: 'carp:/2460{}'.format(n))
+
+    @classmethod
+    def _create(cls, *args, **kwargs):
+        kwargs['category'] = kwargs.get('category', 'carpid')
```
brianjgeiger: Didn't feel like changing these to `erid`s?

brianjgeiger:
Couple of small things, but one big thing: When you're looking at a registration record, how do you see what its identifiers are? Pass done. 🐧

api/identifiers/views.py
```diff
@@ -0,0 +1,98 @@
...
+from website.identifiers.model import Identifier
+
+
+class IdentifierList(JSONAPIBaseView, generics.ListAPIView, RegistrationMixin, ODMFilterMixin):
+    """List of identifiers for a specified node. *Read-only*.
+
+
+   ##Identifier Attributes
+
+    OSF License entities have the "licenses" `type`.
```
felliott: Copy-n-paste doc error!

api/identifiers/views.py
erinspace: I have the assigned doi/ark/whatever from the but not the _id as stored -- not sure how to get it there (without instead asking for the mongoid in the url which seems not-to-spec) but maybe I'm missing something

api_tests/identifiers/views/test_identifier_list.py
erinspace: that's why I was using ```assert_items_equal```, as I understand this order does not matter

api_tests/identifiers/views/test_identifier_list.py
erinspace: same same! ^

tests/factories.py
erinspace: wanted to keep the carp name alive!

api/identifiers/views.py
brianjgeiger: Didn't you change the id of the object to the mongo_id last week, though? That's what should be being passed around.

api/nodes/urls.py
```diff
@@ -23,6 +24,7 @@
     url(r'^(?P<node_id>\w+)/relationships/institution/$', views.NodeInstitutionRelationship.as_view(), name=views.NodeInstitutionRelationship.view_name),
     url(r'^(?P<node_id>\w+)/node_links/$', views.NodeLinksList.as_view(), name=views.NodeLinksList.view_name),
     url(r'^(?P<node_id>\w+)/node_links/(?P<node_link_id>\w+)/', views.NodeLinksDetail.as_view(), name=views.NodeLinksDetail.view_name),
+    url(r'^(?P<node_id>\w+)/identifiers/$', identifier_views.IdentifierList.as_view(), name=identifier_views.IdentifierList.view_name),
```
brianjgeiger: Since you aren't adding the relationship field to the node serializer, this shouldn't be necessary. Go ahead and take it out.

api_tests/identifiers/views/test_identifier_list.py
```diff
@@ -0,0 +1,79 @@
...
+
+class TestIdentifierList(ApiTestCase):
+    def setUp(self):
+        super(TestIdentifierList, self).setUp()
+        self.user = AuthUserFactory()
+        self.user_two = AuthUserFactory()
+
+        self.registration = RegistrationFactory(creator=self.user, is_public=True)
+        self.identifier = IdentifierFactory(referent=self.registration)
+        self.url = '/{}nodes/{}/identifiers/'.format(API_BASE, self.registration._id)
```
brianjgeiger: This should be registrations, not nodes. How does this even work? You shouldn't be able to access a registration through a node sub-route. (Don't research that; it was rhetorical)

brianjgeiger:
Okay, pull out nodes stuff, update PR description. Looks like things are tested. Pass done. 🐧

brianjgeiger:
Awesome! 🎉 🐧 🎉

PR #5682, opened 2016-05-19
Feature/APIv2 500 Error Update Contributors [OSF-6376]

pattisdr:
Ticket https://openscience.atlassian.net/browse/OSF-6376
## Purpose

ValueError not being handled properly.  Getting a 500 error when user attempts an update or bulk update request on contributors that causes there to be no bibliographic contributors. 

In example below, there are two contributors and I am attempting to make them both non-bibliographic, which isn't allowed.  Instead of handling the error properly, a server error is returned.
```
BULK PATCH /v2/nodes/<node_id>/contributors/
{
    "data": [{
        "id": "12345",
        "type": "contributors",
        "attributes": {
            "bibliographic": "False"
        }
    },
      {
        "id": "abcde",
        "type": "contributors", 
        "attributes": {
            "bibliographic": "False"
        }
        
    }]
}

```
## Changes

Properly handles 500 error by throwing a validation error.

### DIFF
--- a/api/nodes/serializers.py
+++ b/api/nodes/serializers.py
@@ -386,6 +386,8 @@ class NodeContributorDetailSerializer(NodeContributorsSerializer):
             node.update_contributor(contributor, permission, visible, auth, save=True)
         except NodeStateError as e:
             raise exceptions.ValidationError(detail=e.message)
+        except ValueError as e:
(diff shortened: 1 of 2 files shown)


brianjgeiger:
Looks great. 🎉 🐧 🎉

PR #5683, opened 2016-05-19
[Feature] Display comments on OSF correctly

Ghalko:
originally #5596 

## Purpose

To render comments with the correct information for people viewing comments on reporting and confirmed spam.

## Changes

- Add to API variables to more completely show the status of a reported comment
- Add `pureComputed` variables to JS in order to display the correct information
- Adjust template to accommodate language and display changes

## Side effects

You can now see removed comments so that there is some semblance of flow if a comment is removed.

## Ticket

https://openscience.atlassian.net/browse/OSF-6265


mfraezz:
LGTM. Dropping this in RTM 🐹

PR #5684, opened 2016-05-19
Warn users they cannot search by email address [OSF-5639]

zachjanicki:
## Purpose

Currently users can search when adding contributors by email address. They can then invite a user, who may already exist in the database by the email address entered, which can cause confusion when it says they already have an account. These changes will no longer enable users to search by email address.

## Changes

If a user attempts to search an email address in the add contributors page, it will show a message that searching by email address is not allowed, and it will not give them the option to add that email address as an unregistered contributor.

BEFORE:
screen shot 2016-05-19 at 11 53 50 am

AFTER:
screen shot 2016-05-19 at 11 56 43 am

## Side effects

None.

## Ticket

### DIFF
--- a/website/static/js/tests/contributors.test.js
+++ b/website/static/js/tests/contributors.test.js
@@ -163,6 +163,37 @@ describe('addContributors', () => {
                    });
                });
            });
+
+           describe('emailSearch', () => {
+               it('should return true with an email address entered', () => {
+                   vm.query(
+                       'a1234@gmail.com'
+                   );
+                   assert.isTrue(vm.emailSearch());
+               });
+               it('should return false with a name entered', () => {
+                   vm.query(
+                       faker.name.findName()
+                   );
+                   assert.isFalse(vm.emailSearch());
+               });
+           });
+
+           describe('nameSearch', () => {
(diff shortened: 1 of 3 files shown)


brianjgeiger:
Tests are failing, so be sure they are passing before putting into Code Review.

brianjgeiger:
Also, could you add before and after screenshots of your changes?

website/static/js/tests/contributors.test.js
```diff
@@ -163,6 +163,37 @@ describe('addContributors', () => {
                    });
                });
            });
+
+           describe('emailSearch', () => {
```
brianjgeiger: Yay tests!

brianjgeiger:
Pass done. 🐧

brianjgeiger:
You'll need to change this branch name. Use the gitflow convention of feature/branch-name, so for this something like `feature/error-searching-by-email-address`. Close this PR and open a new one for the new branch.

As a general rule, we would just keep the current branch name and change it for next time, but you'll want to keep `develop` and `master` free or it could cause some issues later on, and I can't guarantee this PR will be closed before you're going to need to have it available.

PR #5686, opened 2016-05-19
[Feature] Render deactivated users in projects correctly [OSF-6266]

Ghalko:
## Purpose

Removes links of deactivated users. Replaces #5604

## Changes

- Remove link in listed contributors
- Remove link in comments
- Remove link in logs

## Side effects

Possibly an issue with users that are removed from a project, after they are deactivated?

## Ticket

https://openscience.atlassian.net/browse/OSF-6266


website/templates/log_list.mako
```diff
@@ -46,12 +46,17 @@
...
+                                    <span data-bind="if: log.user.registered">
                                     <span data-bind="if: log.userURL">
                                         <a class="overflow" data-bind="text: log.userFullName || log.apiKey, attr: {href: log.userURL}"></a>
                                     </span>
                                     <span data-bind="ifnot: log.userURL">
                                         <span class="overflow" data-bind="text: log.userFullName"></span>
                                     </span>
+                                    </span>
+                                    <span data-bind="ifnot: log.user.registered">
+                                        Deactivated User
```
mfraezz: This should be sentence-cased, as in the anonymized view above:

`A deactivated user [...]`

Also, this only works for logs that the deactivated user is the creator of, not the target of:

screen shot 2016-05-23 at 11 22 56

It's unclear whether or not this is even in the spec, so consider either removing or fixing, for consistency.

website/static/js/comment.js
```diff
@@ -298,7 +298,12 @@ var CommentModel = function(data, $parent, $root) {
             'gravatarUrl': userData.links.profile_image
         };
     } else {
-        self.author = self.$root.author;
+        self.author = {
+            'id': null,
+            'urls': {'profile': ''},
+            'fullname': 'Deactivated user',
```
mfraezz: AVOLs cause the username to read as `A User`, so this should be `A Deactivated User`.

mfraezz: Although, given that we're not changing the name as it appears on the contributors list, perhaps we should not be changing it elsewhere, either. Or at most changing it to something like 
`'{} (deactivated)'.format(userData.attributes.full_name)`

Ghalko: The problem is that the data doesn't come through with deactivated users. I'll look further into it, but for now `A deactivated user`

website/templates/log_list.mako
Ghalko: I think this is going to be replaced soon with the API v2 log that is on the Dashboard. Then it just works. But changed to sentence case for now.

website/static/js/comment.js
mfraezz: Ah, yes. Seemingly because `api/base/utils.py:get_object_or_error` raises an error if a disabled user is requested.

mfraezz: Still, sufficient `meta` information is returned for proper display:

screen shot 2016-05-23 at 13 11 53

website/templates/log_list.mako
mfraezz: Not certain that's correct; it appears to not work at all on the activity log.
<hr>
screen shot 2016-05-23 at 13 14 33
==============
screen shot 2016-05-23 at 13 14 43

mfraezz:
Couple of UI concerns, otherwise LGTM. 

Back to you, @Ghalko 🐁

website/static/js/comment.js
Ghalko: Got it, dug into the error to extract the full_name.

website/templates/log_list.mako
Ghalko: The two logs are using the two different APIs. `logTextParser.js` is the newer one using API v2 and will replace the one in the Project dashboard *soon.

mfraezz:
LGTM. RTM when the build passes. 🐭

admin/users/views.py
```diff
@@ -40,19 +40,19 @@ class UserDeleteView(OSFAdmin, DeleteView):
...
                 message = 'User account {} disabled'.format(user.pk)
             else:
                 user.date_disabled = None
                 subscribe_on_confirm(user)
+                user.is_registered = True
                 flag = USER_RESTORED
                 message = 'User account {} reenabled'.format(user.pk)
             user.save()
-            if flag is not None:
+            if flag:
```
mfraezz: Is this conditional necessary? It looks like flag can't be falsy here.

mfraezz:
>Possibly an issue with users that are removed from a project, after they are deactivated?

Didn't appear to be problematic during my manual testing. Did you test this and notice any odd behavior?

admin/users/views.py
```diff
@@ -61,15 +61,11 @@ def delete(self, request, *args, **kwargs):
                     action_flag=flag
                 )
         except AttributeError:
```
mfraezz: What might be raising an `AttributeError` in the above `try` block?

If nothing aside from the lack of a `user` is expected to do so, you should drop the `try/except` and just do a check after `self.get_object`, raising the `404` if it returned `None`

admin/users/views.py
```diff
@@ -100,15 +96,11 @@ def delete(self, request, *args, **kwargs):
                 action_flag=USER_2_FACTOR
             )
         except AttributeError:
```
mfraezz: Ditto above comment:

>What might be raising an AttributeError in the above try block?
>
>If nothing aside from the lack of a user is expected to do so, you should drop the try/except and just do a check after self.get_object, raising the 404 if it returned None

admin/users/views.py
```diff
@@ -132,47 +124,30 @@ def get_object(self, queryset=None):
...
-                    )
-                )
-            )
+    context_object_name = 'user'
 
     def get_context_data(self, **kwargs):
         user = User.load(self.kwargs.get('guid'))
         try:
             self.initial.setdefault('emails', [(r, r) for r in user.emails])
         except AttributeError:
```
mfraezz: Mostly ditto above comment:

>What might be raising an AttributeError in the above try block?
>
>If nothing aside from the lack of a user is expected to do so, you should drop the try/except and just do a check after self.get_object, raising the 404 if it returned None

Also, is there interplay between this and `form_valid`?

mfraezz:
Pass complete. Should be RTM after nitpicky comments are addressed 🎲

admin/users/views.py
Ghalko: This is doing duck-typing. A `None` return value doesn't quack. This was a major part of refactors after users were refactored. I could change this in all previous/subsequent modules...

admin/users/views.py
Ghalko: Ditto my reply above. https://github.com/CenterForOpenScience/osf.io/pull/5686/files#r64436601

admin/users/views.py
Ghalko: It is a double check that the user found with an email is the same one found with the guid.

admin/users/views.py
mfraezz: We usually try to limit the amount of code in a `try` block that isn't expected to raise an error.

Duck-typing is generally intended to determine whether or not an object is suitable based on the methods/properties/attributes it has, but any `User` object should have all of these.

admin/users/views.py
mfraezz: >It is a double check that the user found with an email is the same one found with the guid.

"It" being `form_valid`, I take it, as that's not what this is doing. Also, should `form_valid` then be raising something other than a `404`, then, as it might not be the case that an object wasn't found? `400` or `409` both seem more correct.

admin/users/views.py
Ghalko: Moved the `update_admin_log` out of the try-block. All other operations within the if statement and the `user.save()` are either assignment statements or attribute manipulation.

admin/users/views.py
Ghalko: 409 is more of a conflict of the database from what I've read. The fact that it isn't included in the Django framework but is in the DRF is interesting. And 400 is more for malformed requests. I think a 'not found' describes the: "I didn't find the original user with that email address." Which I can certainly update this to say.

mfraezz: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#409

> ##### 409 Conflict
> Indicates that the request could not be processed because of conflict in the request

My experience from API work is that this "conflict in the request" is frequently that some of the data passed in conflicts with assumptions about said data -- such as a passed-in email and id not being associated with the same user, though both data are assumed to map to the same user.

That being said, probably not a big deal.

Ghalko: Does the change in language from the latest commit address your concerns? @mfraezz https://github.com/CenterForOpenScience/osf.io/pull/5686/commits/a8eb2f16399f4ddf22f6b34b95ee4594301edd49

brianjgeiger: 409 seems correct for this error code. You are claiming a request that has contradictory info: That the user has this user id and this email address, but both of those pieces of information are not true at the same time. Hence conflict.

Ghalko: Returned 409

admin/users/views.py
mfraezz: Above comment had some resolving action, this seems to not have.

website/templates/log_list.mako
```diff
@@ -46,12 +46,17 @@
                                 <span class="contributor-anonymous">A user</span>
                             </span>
                                 <span data-bind="ifnot:log.anonymous">
+                                    <span data-bind="if: log.user.registered">
```
sloria: Logs on the "My projects" page are rendered with separate code than the logs on project Overview pages. See `logTextParser.js` and `logActionsList.json`. You may need to make a corresponding change there to make the logs display consistently.

sloria:
Pass finished. Good to merge after https://github.com/CenterForOpenScience/osf.io/pull/5686/files#r64758445 is addressed

website/templates/log_list.mako
Ghalko: It is taken care of with the API. :) @sloria Removed users return an error and it is caught in `logTextParser.js` so it only displays the name.

sloria: Oh neat!

PR #5692, opened 2016-05-19
Dev/feat/math typesetting for share search results [OSF-6020]

HalcyonChimera:
## Purpose
Allow math ml to render in a beautiful and pleasing manner in the results of share searches.

## Changes
included mathjax script on the page, set up hooks to add certain elements to the mathjax queue once they're loaded into the dom.

## Side effects
It'll typeset any markup in tags or contributor's names. It'll probably break in ie6.

## Ticket
https://openscience.atlassian.net/browse/OSF-6020


brianjgeiger:
Looks like it's failing flake8. The details link above will tell you more.

PR #5693, opened 2016-05-19
[migration] Create wiki without restoring deleted [OSF-5968]

jnayak1:
## Purpose

Currently, if you delete a wiki file, and then create a new wiki file with the same name as the deleted file, the new file still maintains all of the edits from the old file. These can be found under the 'Compare' panel in the project wiki. (In this example, "hello" has been created, edited, deleted, and re-created)
Alt text

The purpose of this PR is when a user deletes a project wiki, the wiki (and all of its former edits) are permanently deleted. When a new wiki is created with the same name as a deleted wiki, it will not restore the old wiki's edits. (Again "hello" has been created, edited, deleted, and re-created, but the old edits no longer persist in "hello"'s versions)
Alt text
## Changes
Changes

When a wiki is deleted currently, the page is just deleted from the project's wiki_pages_current. In order to prevent the deleted page from being restored when a new wiki of the same name is created, the page is now also deleted from the project's wiki_pages_versions upon deletion.

This change alone would only prevent newly deleted pages from being restored. If a user has already deleted a page before this change is made, and then creates a new wiki page of the same name, those previous edits would be restored. To fix this, this PR also includes a migration script that removes all wiki pages from a project's wiki_pages_versions that are not also pages in that project's wiki_pages_current. This prevents any page that has been deleted but does not currently exist as a project wiki from being restored when the user creates a new wiki page of the same name.

## Side effects

There is really only one case (I can think of) that this PR wouldn't resolve. For example, let's say:
1. You create a wiki page "hello"
2. You make some edits to that page
3. You delete that page
4. You create another wiki page called "hello", and those previous edits have been restored
This PR will not retroactively un-restore those edits from the first version of "hello". Instead, if you now delete "hello", then create a new "hello" after this PR, the previous edits will not persist.

## Ticket

https://openscience.atlassian.net/browse/OSF-5968


brianjgeiger:
We should also test the basic behavior of the code in a non-migration test, so we'll need a unit test that Creates, edits, deletes, creates, and edits a wiki and ensures that the proper number of revisions exists at each point. That would go in `website/addons/wiki/tests/test_wiki.py`.

scripts/migration/migrate_deleted_wikis.py
```diff
@@ -0,0 +1,45 @@
...
+    # iterate over targets
+    for log in targets:
+        node = log.node
+        versions = node.wiki_pages_versions
+        current = node.wiki_pages_current
+        updated_versions = {}
+        for wiki in versions:
+            if wiki in current:
+                updated_versions[wiki] = versions[wiki]
+        if not dry_run:
```
brianjgeiger: Since everything is wrapped in a transaction, we *should* be able to leave this if statement out and just always save. We've had a couple of really rough migrations recently, though, so let's let @sloria rule on that before taking it out.

website/addons/wiki/tests/test_wiki.py
```diff
@@ -569,6 +569,20 @@ def test_project_wiki_delete_w_valid_special_characters(self, mock_sharejs):
         self.project.reload()
         assert_not_in(to_mongo_key(SPECIAL_CHARACTERS_ALLOWED), self.project.wiki_pages_current)
 
+    @mock.patch('website.addons.wiki.utils.broadcast_to_sharejs')
+    def test_project_wiki_versions_after_delete(self, mock_sharejs):
+        # Ensure wiki versions do not reappear when a wiki is deleted and a new
```
brianjgeiger: Try to use more descriptive test titles instead of comments. https://cosdev.readthedocs.io/en/latest/process/testing.html#general-testing-guidelines

scripts/migration/migrate_deleted_wikis.py
sloria: Yeah, this shouldn't be necessary. Raising an uncaught exception is currently the preferred way to implement dry_runs. 

One minor refactor would be to raise the error in the with block:

```python
with TokuTransaction():
     migrate(...)
      if dry_run:
           raise RuntimeError(...)
```

This puts all the transaction code in the same place.

brianjgeiger:
Pass done. 🐧

scripts/migration/migrate_deleted_wikis.py
```diff
@@ -0,0 +1,44 @@
...
+        versions = node.wiki_pages_versions
+        current = node.wiki_pages_current
+        updated_versions = {}
+        for wiki in versions:
+            if wiki in current:
+                updated_versions[wiki] = versions[wiki]
+        with TokuTransaction():
+            node.wiki_pages_versions = updated_versions
+            node.save()
+            if dry_run:
```
brianjgeiger: This wasn't exactly what Steve said. The if dry_run should be in the main() function's `with TokuTransaction()` otherwise it'll only do one pass through the loop and stop.

scripts/migration/migrate_deleted_wikis.py
```diff
@@ -0,0 +1,44 @@
...
+    # iterate over targets
+    for log in targets:
+        node = log.node
+        versions = node.wiki_pages_versions
+        current = node.wiki_pages_current
+        updated_versions = {}
+        for wiki in versions:
+            if wiki in current:
+                updated_versions[wiki] = versions[wiki]
+        with TokuTransaction():
```
brianjgeiger: You also don't need a withTokuTransaction() in this function because the call to `migrate()` is wrapped in a transaction from the call in the main() function.

brianjgeiger:
Pass done. Nearly there. 🐧

brianjgeiger:
Cool. When this passes Travis it should be ready. Since there's a migration, Steve may do another pass. 🎉 🐧 🎉

scripts/migration/migrate_deleted_wikis.py
```diff
@@ -0,0 +1,43 @@
...
+    # iterate over targets
+    for log in targets:
+        node = log.node
+        versions = node.wiki_pages_versions
+        current = node.wiki_pages_current
+        updated_versions = {}
+        for wiki in versions:
+            if wiki in current:
+                updated_versions[wiki] = versions[wiki]
+        node.wiki_pages_versions = updated_versions
```
sloria: This migration could use some logging. At the very least, any nodes whose wiki_pages_versions was changed should be logged. Something like:

```
Modified Node 123ab wiki_pages_versions from {'abc123..': [....} to {....}
```

It would also be nice to include a count of modified nodes.

sloria:
Thank you for being thorough in your PR description 👪

scripts/migration/migrate_deleted_wikis.py
sloria: I can do this when I merge.

scripts/migration/migrate_deleted_wikis.py
```diff
@@ -0,0 +1,43 @@
...
+        node.save()
+
+
+def main():
+    dry_run = False
+    if '--dry' in sys.argv:
```
sloria: Very minor: This can be simplified to

```python
dry_run = '--dry' in sys.argv
```

I'll change this when I merge.

sloria:
This is good to go. I've created a branch with this and the minor logging additions. 🐙

PR #5694, opened 2016-05-19
[OSF-6273] Fix last update timestamp on draft registrations to not update for al…

Tav0:
## Purpose
Fix so that the only the draft registration being updated will have a new time stamp.

## Changes
Added a boolean check before the meta schema gets saved.

## Side effects
None.

## Ticket
https://openscience.atlassian.net/browse/OSF-6273

### DIFF
--- a/website/templates/project/edit_draft_registration.mako
+++ b/website/templates/project/edit_draft_registration.mako
@@ -48,7 +48,7 @@
 
                 <div class="row" style="margin-bottom: 10px;">
                   <span>
-                    Last saved: <span data-bind="text: lastSaveTime"></span>
+                    Last auto-saved: <span data-bind="text: lastSaveTime"></span>
                   </span>
                   <!-- ko if: onLastPage -->
                   <span data-bind="if: onLastPage() && hasValidationInfo()" class="pull-right">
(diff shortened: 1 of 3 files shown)


Tav0:
Added a minor edit to the Draft Registration editor so that the user knows that what he/she is editing is being auto-saved without clicking on "Save Draft"

website/templates/project/edit_draft_registration.mako
```diff
@@ -48,7 +48,7 @@
 
                 <div class="row" style="margin-bottom: 10px;">
                   <span>
-                    Last saved: <span data-bind="text: lastSaveTime"></span>
+                    Last auto-saved: <span data-bind="text: lastSaveTime"></span>
```
brianjgeiger: Was this change suggested by someone?

brianjgeiger:
Code seems good, but I haven't tried it out. Make an arrangement to demo this with @nicipfeiffer and/or @saradbowman. I will drop into the Product Demo column of Jira.

website/templates/project/edit_draft_registration.mako
Tav0: @brianjgeiger Yes, before making the change I talked to @jinluyuan about this (:

brianjgeiger: Cool, thank you. When you do a review with Nici and/or Sara, please mention this explicitly and get their feedback as well, just in case.

brianjgeiger:
Looks great. 🎉 🐧 🎉

PR #5695, opened 2016-05-19
Warn users they cannot search by email address [OSF-5639] 

zachjanicki:
## Purpose

Currently users can search when adding contributors by email address. They can then invite a user, who may already exist in the database by the email address entered, which can cause confusion when it says they already have an account. These changes will no longer enable users to search by email address.

## Changes

If a user attempts to search an email address in the add contributors page, it will show a message that searching by email address is not allowed, and it will not give them the option to add that email address as an unregistered contributor.

### BEFORE:

image

### AFTER:
image

## Side effects

None.

## Ticket

### DIFF
--- a/website/static/js/contribAdder.js
+++ b/website/static/js/contribAdder.js
@@ -97,7 +97,7 @@ AddContributorViewModel = oop.extend(Paginator, {
         self.childrenToChange = ko.observableArray();
 
         self.emailSearch = ko.pureComputed(function () {
-            if(String(self.query()).indexOf("@") >= 0) {
+            if(String(self.query()).indexOf('@') >= 0) {
                 return true;
             } else {
                 return false;
--- a/website/static/js/contribAdder.js
+++ b/website/static/js/contribAdder.js
@@ -96,11 +96,23 @@ AddContributorViewModel = oop.extend(Paginator, {
         self.totalPages = ko.observable(0);
         self.childrenToChange = ko.observableArray();
 
+        self.emailSearch = ko.pureComputed(function () {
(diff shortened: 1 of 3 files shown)


brianjgeiger:
I'm guessing your before and after photos are mixed up?

zachjanicki:
Ah yes they were. It is corrected.

website/static/js/contribAdder.js
```diff
@@ -96,10 +96,17 @@ AddContributorViewModel = oop.extend(Paginator, {
         self.totalPages = ko.observable(0);
         self.childrenToChange = ko.observableArray();
 
+        self.emailSearch = ko.pureComputed(function () {
+            var emailRegex = new RegExp('[a-zA-Z0-9]+@[a-zA-Z0-9]+\.[a-z]+');
```
brianjgeiger: We would never use this for real email address validation, but it should be fine for the purpose.

brianjgeiger: No, wait, I don't think that would work for some cases that QA will almost certainly use. (i.e. nici+testuser1@cos.io) Well maybe.

Still, take a look at http://stackoverflow.com/a/1373724 and maybe do the practical one (include a comment to the link to that answer so people can see where it came from and why you used it)

brianjgeiger:
Okay, one small correction, and please add updated screenshots showing each of the conditions you've modified. Then I'll have Sara verify and we'll be done. 🐧

mfraezz:
Other than the small UI change, LGTM. This should be ready soon 🌵

PR #5696, opened 2016-05-19
[OSF-6383] Fix for text wrapping

Tav0:
## Purpose
Fix for License text does not wrap in Firefox and Chrome

## Changes
Added support for Chrome, Firefox and Opera.

## Side effects
None

## Ticket
https://openscience.atlassian.net/browse/OSF-6383

### DIFF
--- a/website/static/css/pages/project-page.css
+++ b/website/static/css/pages/project-page.css
@@ -143,7 +143,7 @@ span#nodeDescriptionEditable{
     white-space:pre-line;
 }
 
-pre {
+.license-text {
     white-space: pre-wrap;       /* CSS 3 */
     white-space: -pre-wrap;      /* Opera 4-6 */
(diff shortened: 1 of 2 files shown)


website/static/css/pages/project-page.css
```diff
@@ -142,3 +142,9 @@
 span#nodeDescriptionEditable{
     white-space:pre-line;
 }
+
+pre {
```
sloria: This selector is too general; it might have side effects for other `pre` tags on the page. Can you apply a CSS class to the specific pre tag you want to style this way?

sloria:
Pass finished.

website/static/css/pages/project-page.css
Tav0: My bad, you are right! Just pushed again. Thanks!

sloria:
Thanks!

PR #5697, opened 2016-05-20
[OSF-3622] CSL citation fix

RomanZWang:
For CSL files with no specified bibliography information, default citation data is provided

### DIFF
--- a/.bash_profile
+++ b/.bash_profile
@@ -1,0 +1,3 @@
+PATH=$PATH:/usr/local/sbin
+source "/usr/bin/virtualenvwrapper.sh"
+export WORKON_HOME="/opt/virtual_env/“
--- a/.bash_profile
+++ b/.bash_profile
@@ -1,3 +1,1 @@
-PATH=$PATH:/usr/local/sbin
-source "/usr/bin/virtualenvwrapper.sh"
-export WORKON_HOME="/opt/virtual_env/“
+
--- a/.bash_profile
+++ b/.bash_profile
@@ -1,3 +1,0 @@
-PATH=$PATH:/usr/local/sbin
-source "/usr/bin/virtualenvwrapper.sh"
-export WORKON_HOME="/opt/virtual_env/“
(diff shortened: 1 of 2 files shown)


mfraezz:
I'm not sure if this is the best approach, given the differences between how `U Schyłku Starożytności (Polish)` looks on the Zotero style page and how it looks with these changes.

Ideally, we'd be able to prevent styles that we can't necessarily render correctly from being options presented to the user, rather than attempting to parse some json that may or may not be indicative of what the style is supposed to look like.

Three main points to achieve this:

- Add flag on the `CitationStyle` model -- indicates presence of `bibliography` tag in the xml tree
- Modify the default query to exclude models without a bib
- Modify script to set flag on models.

mfraezz:
Also, please follow the default template when submitting a PR:

```
## Purpose

<!-- Describe the purpose of your changes -->

## Changes

<!-- Briefly describe or list your changes  -->

## Side effects

<!--Any possible side effects? -->


## Ticket

<!-- Link to JIRA ticket, if applicable e.g. https://openscience.atlassian.net/browse/OSF-1234 -->
```

mfraezz:
Furthermore, the build is failing due to syntax errors (see the bottom of the [log](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/132299276)).

You can run these checks locally with `invoke flake` and `invoke jshint`

.bash_profile
```diff
@@ -0,0 +1 @@
+
```
mfraezz: Please don't commit `.dot_files`

mfraezz:
Pass complete @RomanZWang. Alternate approach proposed, happy to do a demo / talk about concerns touched on in [that comment](https://github.com/CenterForOpenScience/osf.io/pull/5697#issuecomment-221315994). 🎶

sloria:
Just a drive-by comment: there's a lot of noise in the history for this PR. @RomanZWang when you finish addressing Matt's review, can you create a new branch and commit only the necessary changes in a single commit or two?

RomanZWang:
Okay I will make these changes. Thank you for the feedback

PR #5700, opened 2016-05-20
[feature fix] Uploads do not continue after Dataverse upload cancel [OSF-5459]

el1t:
## Purpose
Cancelling pending uploads to Dataverse and Github (those that require synchronous uploads) failed to cancel uploads. Instead, **fangorn** would attempt to upload a cancelled file, which would either succeed (undesired) or raise an error, stopping all uploads (also undesired). However, it would appear cancelled to the user because **fangorn** removes all uploads the user cancels, cancelled or not, from the UI. This was because uploads were removed from the UI regardless of whether they were found in `dropzone`.

In addition, when cancelling an upload in progress by a synchronous provider, the next upload that gets processed is not cancellable (even though the old code searched `dropzone.files`) due to `dropzone` not adding it to the `files` list.

## Changes

- Add robust searching through `syncFileCache` in addition to `dropzone.files` in `cancelUploads`
 - Remove all pending uploads from `syncFileCache` when `row` is not specified
 - Search through `syncFileCache` for `row` if provider is part of `SYNC_UPLOAD_ADDONS`
- Prevent removing from UI (`tb.deleteNode`) unless file to cancel is found
- Optimize filtering `dropzone.files` for cancellable files
- Add comments to `cancelUploads`
- Manually append next file in `syncFileCache` to `dropzone.files` when queued upload starts

_Note that files with synchronous providers may not actually appear in `syncFileCache` if they are currently being uploaded by `dropzone`._

## Side effects

Could affect uploading to asynchronous providers, but unlikely—it should be _more_ reliable, since the file is not removed from the UI unless it is actually cancelled. Tested working with Box.

## Ticket

https://openscience.atlassian.net/browse/OSF-5459

### DIFF
--- a/website/static/js/fangorn.js
+++ b/website/static/js/fangorn.js
@@ -10,6 +10,7 @@ var m = require('mithril');
 var URI = require('URIjs');
 var Raven = require('raven-js');
 var Treebeard = require('treebeard');
+var Dropzone = require('dropzone');
 
 var $osf = require('js/osfHelpers');
 var waterbutler = require('js/waterbutler');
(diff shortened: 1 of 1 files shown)


el1t:
Perhaps this should be a hot fix, since the issues are present on [production](osf.io)?

whit537:
#### Investigations—for @whit537

- [ ] skim code
- [ ] run code
 - [ ] configure an external storage provider?
- [ ] read code

whit537:
#### Punchlist—for @el1t

whit537:
I'm working on reproducing the [bug report](https://openscience.atlassian.net/browse/OSF-5459). I've created an account at https://apitest.dataverse.org/ and have connected it to a project in my local OSF instance. I have a simple text file that I'm trying to upload to the Dataverse draft folder on the project overview page. When I upload the file I get an "Error: Internal Server Error" message in the UI, and a `temp.zip` file in Dataverse. Investigating ...

whit537:
Per @felliott in Flowdock, retrying with `demo.dataverse.org` ...

whit537:
That worked! Proceeding with bug repro ...

whit537:
Alright, bug reproduced. Now to evaluate the fix ...

whit537:
Fix seems to work! Now to read code ...

website/static/js/fangorn.js
```diff
@@ -88,31 +89,70 @@ function findByTempID(parent, tmpID) {
...
-            tb.deleteNode(row.parentID,row.id);
-            if(row.data.tmpID === j.tmpID){
-                tb.dropzone.removeFile(j);
+    var cancelableStatuses = [Dropzone.UPLOADING, Dropzone.QUEUED];
+    // Select files that are uploading, queued, or rejected (!accepted)
+    var filesArr = tb.dropzone.files.filter(function(file) {
+        return cancelableStatuses.indexOf(file.status) > -1 || !file.accepted;
+    });
+    // Remove all queued files
+    if (row === undefined) {
```
whit537: I do find where `cancelUploads` is [called](https://github.com/el1t/osf.io/blob/86896d674ba92087da624f33a4f8860651950a93/website/static/js/fangorn.js#L1823) without a `row` (though I haven't figured out how to get there in the UI).

👍

el1t: Yeah that's actually another bug, fangorn doesn't show the cancel all uploads button when viewing synchronous providers

whit537: This method is getting unwieldy. Let's go ahead and split into two methods, one for canceling all uploads, and one for canceling one upload.

website/static/js/fangorn.js
```diff
@@ -88,31 +89,70 @@ function findByTempID(parent, tmpID) {
...
+            if (file.tmpID === row.data.tmpID) {
+                tb.deleteNode(row.parentID, row.id);
+                tb.dropzone.removeFile(file);
+                return true;
+            }
+        });
     }
-    tb.isUploading(false);
+    tb.isUploading(handled && filesArr.length > 1);
 }
```
whit537: This method is getting unwieldy. Let's go ahead and split it into two methods, one for canceling all uploads, and one for canceling one upload.

website/static/js/fangorn.js
whit537: @el1t How do I trigger `cancelUploads(undefined)` through the UI?

el1t: You can change the call from `cancelUploads.call(tb, item);` [here](https://github.com/el1t/osf.io/blob/86896d674ba92087da624f33a4f8860651950a93/website/static/js/fangorn.js#L192) to `cancelUploads.call(tb)`

whit537:
Alright @el1t, [punchlist](https://github.com/CenterForOpenScience/osf.io/pull/5700#issuecomment-223611756) updated. A couple things to get started ...

website/static/js/fangorn.js
whit537: > Yeah that's actually another bug, fangorn doesn't show the cancel all uploads button when viewing synchronous providers

Ah, hrm. Okay, that explains that. :-P

whit537: > You can change the call from `cancelUploads.call(tb, item);` [here](https://github.com/el1t/osf.io/blob/86896d674ba92087da624f33a4f8860651950a93/website/static/js/fangorn.js#L192) to `cancelUploads.call(tb)`

Right, missed your previous comment, sorry.

whit537: With that manual change I am seeing that cancelling all seems to work.

whit537:
I'm debating whether to ask for regression tests for this. It could be complicated since this involves external services. @el1t Waddya think? :)

whit537:
Bringing over from private Flowdock:

>> What do regression tests entail?
>
> A regression is a fancy name for a bug. A regression test is a test for a bug. The idea is that you would write a failing test for the bug reported in the Jira ticket, and then write the fix.

It's too late to write the test first, but we could at least write the test second. :-)

whit537:
@el1t I'm good on this. If you want to write a test or two that'd be great, but I'm not going to block on it.

👍

sloria:
Nice work!

PR #5701, opened 2016-05-20
Feature/name order [OSF-5948]

zamattiac:
## Purpose

To alphabetically sort the bibliographic contributors of publicized projects on the dashboard (under "Discover").

## Changes

Added an ordering specification to NodeContributorsList that it be by last name.

## Side effects

The change doesn't appear to mess with manually-specified contributor ordering on the actual node page.

## Ticket

### DIFF
--- a/api/nodes/views.py
+++ b/api/nodes/views.py
@@ -556,6 +556,8 @@ class NodeContributorsList(JSONAPIBaseView, bulk_views.BulkUpdateJSONAPIView, bu
         base_permissions.TokenHasScope,
     )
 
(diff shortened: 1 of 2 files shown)


brianjgeiger:
Unfortunately, the ordering requested was not alphabetically, but instead to sort by the order chosen by the admins of the project. Because these projects are meant to represent work for academic papers, the contributors list is meant to represent the order of work that went into the project. This is represented by the order of the contributors in the list on the project model.

The trick with this one will be getting DRF to recognize the order that is set by the python list. Pass done. 🐧

PR #5702, opened 2016-05-20
Fix the display of unrendered markdown on wiki and project page [OSF-6361]

Tav0:
## Purpose
Fix the unwanted unrendered markdown on the wiki and project page.

## Changes
Added CSS so the unwanted unrendered markdown is not displayed until it is rendered.

## Side effects
None.

## Ticket
https://openscience.atlassian.net/browse/OSF-6361

### DIFF
--- a/website/addons/wiki/model.py
+++ b/website/addons/wiki/model.py
@@ -169,7 +169,7 @@ class NodeWikiPage(GuidStoredObject, Commentable):
 
     page_name = fields.StringField(validate=validate_page_name)
     version = fields.IntegerField()
-    date = fields.DateTimeField(auto_now_add=datetime.datetime.utcnow)
+    date = fields.DateTimeField(default=datetime.datetime.utcnow)
     content = fields.StringField(default='')
 
     user = fields.ForeignField('user')
--- a/website/addons/wiki/model.py
+++ b/website/addons/wiki/model.py
@@ -170,7 +170,7 @@ class NodeWikiPage(GuidStoredObject, Commentable):
 
     page_name = fields.StringField(validate=validate_page_name)
     version = fields.IntegerField()
-    date = fields.DateTimeField(default=datetime.datetime.utcnow)
+    date = fields.DateTimeField(auto_now_add=datetime.datetime.utcnow)
     content = fields.StringField(default='')
 
     user = fields.ForeignField('user')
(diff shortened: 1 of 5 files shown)


brianjgeiger:
Looks great. 🎉 🐧 🎉

website/addons/wiki/model.py
```diff
@@ -169,7 +169,7 @@ class NodeWikiPage(GuidStoredObject, Commentable):
 
     page_name = fields.StringField(validate=validate_page_name)
     version = fields.IntegerField()
-    date = fields.DateTimeField(auto_now_add=datetime.datetime.utcnow)
+    date = fields.DateTimeField(default=datetime.datetime.utcnow)
```
sloria: Why was this changed? `auto_now_add` will ensure that `date` is only set on creation, which is what we want.

website/static/css/pages/project-page.css
```diff
@@ -152,3 +152,7 @@ span#nodeDescriptionEditable{
     white-space: -pre-wrap;      /* Opera 4-6 */
     white-space: -o-pre-wrap;    /* Opera 7 */
 }
+
+#markdownRender {
```
sloria: We try to avoid selecting CSS IDs in our CSS rules because they aren't reusable. Also, we already have a CSS helper class to hide UI that is rendered client-side, `.scripted`. Can you use that instead?

website/static/js/pages/project-dashboard-page.js
```diff
@@ -207,7 +207,10 @@ $(document).ready(function () {
                 var truncatedText = $.truncate(renderedText, {length: 400});
                 markdownElement.html(truncatedText);
                 mathrender.mathjaxify(markdownElement);
+                markdownElement.css('display', 'inherit');
```
sloria: Minor: you should be able to do `markdownElement.show()`, which is more readable and possibly more correct.

website/addons/wiki/static/wikiPage.js
```diff
@@ -84,16 +84,19 @@ function ViewWidget(visible, version, viewText, rendered, contentURL, allowMathj
 
                 request.done(function (resp) {
                     if(self.visible()) {
+                        var markdownElement = $('#wikiViewRender');
```
sloria: > All variables that are used to store/cache jQuery objects should have a name prefixed with a $.

http://lab.abhinayrathore.com/jquery-standards/ via [COSDev](https://cosdev.readthedocs.io/en/latest/style_guides/javascript.html#jquery)

sloria:
Pass finished. Just a few more minor changes to make.

website/addons/wiki/model.py
Tav0: @sloria I believe I fixed that on my last commit `clean code`

PR #5703, opened 2016-05-20
Remove auto-forward from External Link add-on [OSF-4884]

cameronblandford:
## Purpose

Remove the auto-forward feature from the External Link add-on. Require user confirmation before following the link.

## Changes

To accomplish this, I completely removed the timer system from the add-on, and updated the modal text to reflect the lack of a countdown / auto-redirect. I removed the form input for "seconds until auto-redirect," as it no longer had any effect. I removed the tests that only tested the auto-redirect feature, and updated the other tests to not look for any of the (no longer existent) variables involved in the auto-redirect feature. Because a feature was being removed, I didn't add any tests.

## Side effects

There are no side effects that I foresee occurring. Everything not mentioned explicitly by the ticket should work as it always has.

### Before
image
image

### After
image
image

## Ticket

[[OSF-4884]](https://openscience.atlassian.net/browse/OSF-4884)


sloria:
cursor_and_remove_auto-forward_from_external_link_add-on__osf-4884__by_cameronblandford_ _pull_request__5703_ _centerforopenscience_osf_io

❤️  😍  💟  😻

sloria:
You should also be able to remove the redirect-related fields from the model as well.

sloria:
Pass finished. Just a little more code to remove and this is good to merge (Travis-willing).

website/addons/forward/model.py
```diff
@@ -18,10 +18,6 @@ class ForwardNodeSettings(AddonNodeSettingsBase):
     url = fields.StringField(validate=URLValidator())
     label = fields.StringField(validate=sanitized)
     redirect_bool = fields.BooleanField(default=True, validate=True)
```
sloria: Can you also remove this field?

cameronblandford: The redirect boolean?

cameronblandford: Redirect boolean now completely removed from the add-on. Add-on will now always automatically redirect.

website/addons/forward/tests/factories.py
```diff
@@ -12,6 +12,4 @@ class Meta:
         model = ForwardNodeSettings
 
     owner = SubFactory(ProjectFactory)
-    url = 'http://frozen.pizza.reviews/'
-    redirect_bool = True
-    redirect_secs = 15
+    url = 'http://frozen.pizza.reviews/'
```
mfraezz: Trivial, but generally prefer a single newline at EOF. Some editors get upset if it's not there.

mfraezz:
Build is failing due to flake8 errors: https://travis-ci.org/CenterForOpenScience/osf.io/jobs/132628180

You can run this locally with `invoke flake8`

website/addons/forward/tests/factories.py
cameronblandford: Good to know. POSIX standard?

mfraezz: Yup. From http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206 :

> 3.206 Line
>
> A sequence of zero or more non- \<newline> characters plus a terminating \<newline> character.

This doesn't really matter for 99% of modern systems, but it's nice to be standards-compliant in the trivially-simple cases. Also, it can sometimes lead to odd terminal behavior:

```
~/D/example ❯ cat a.txt
foo% 
~/D/example ❯ cat b.txt
bar
~/D/example ❯ cat *.txt
foobar
```

`zsh` is smart enough to put a `%\n` after line two in this case, most shells are not.

mfraezz:
LGTM. Dropping this in RTM 📟

PR #5704, opened 2016-05-20
[Feature] API v2 add file tags to serialization [OSF-6379]

Ghalko:
## Purpose

Add file tags to API v2 serialization in order to facilitate Ember work.

## Changes

- add tags to file serialization ala node serialization
- allow tags to be updated if in osfstorage
- make sure that files serialize with tags (testing)

screen shot 2016-06-01 at 10 35 41 am

screen shot 2016-06-01 at 10 55 39 am

## Side effects

Will not allow updates for non-osfstorage accounts

## Ticket

https://openscience.atlassian.net/browse/OSF-6379


brianjgeiger:
Could you add a screenshot of the /v2/docs endpoint related to this as well as a copy of some sample JSON (request and response under various circumstances would be ideal) in the changes section?

api/files/serializers.py
```diff
@@ -88,6 +111,7 @@ class FileSerializer(JSONAPISerializer):
         'size',
         'provider',
         'last_touched',
+        'tags',
```
brianjgeiger: Could you verify that tags are indexed in ModularODM? I mean, it would be crazy if they weren't, because why tag if you aren't going to be able to search, but just in case.

api/files/serializers.py
```diff
@@ -188,6 +213,23 @@ def user_id(self, obj):
...
+            elif self.partial:
+                current_tags = set(old_tags)
+            else:
+                current_tags = set()
+
+            for new_tag in (current_tags - old_tags):
+                instance.add_tag(new_tag, auth=auth)
+            for deleted_tag in (old_tags - current_tags):
+                instance.remove_tag(deleted_tag, auth=auth)
+        if instance.provider != 'osfstorage' and 'tags' in validated_data:
```
brianjgeiger: I think if you bump this to the top of the method just below the assert, then you won't need the lines above this to be in an if branch.

api/nodes/serializers.py
```diff
@@ -264,8 +264,7 @@ def update(self, node, validated_data):
         auth = get_user_auth(self.context['request'])
         old_tags = set([tag._id for tag in node.tags])
         if 'tags' in validated_data:
-            current_tags = set(validated_data.get('tags'))
-            del validated_data['tags']
+            current_tags = set(validated_data.pop('tags', []))
```
brianjgeiger: Look who's all fancy. 👍

brianjgeiger:
This could use some more tests if at all possible.

brianjgeiger:
Pass complete. Refactor `if`, add tests if possible, verify index, and make documentation of issue a little more complete. 🐧

api/files/serializers.py
Ghalko: So tags... not indexed on `FileNode` or even `Node.` They are indexed with `elasticsearch` but of course we don't have an API route to that, though it would be interesting if we directed some of the filtering to `elasticsearch` through middleware that could take the filters and then do a search.

brianjgeiger: There is an issue to do that, but yeah, not for this. Okay, could you add the indices?

Ghalko: To both?

brianjgeiger: Yeah, if we're filtering on the field, we should have an index.

api/files/serializers.py
Ghalko: Done.

api/files/serializers.py
Ghalko: Indexed.

brianjgeiger:
Could you add a screenshot of the /v2/docs endpoint related to this as well as a copy of some sample JSON (request and response under various circumstances would be ideal) in the changes section?

website/files/models/base.py
```diff
@@ -199,7 +199,7 @@ class StoredFileNode(StoredObject, Commentable):
     checkout = fields.AbstractForeignField('User')
 
     #Tags for a file, currently only used for osfStorage
-    tags = fields.ForeignField('Tag', list=True, index=True)
+    tags = fields.ForeignField('Tag', list=True)
```
brianjgeiger: Have you talked with anyone about the performance implications of this? Is it otherwise handled somehow, or is filtering on tags just not going to work well?

api_tests/files/views/test_file_detail.py
```diff
@@ -361,3 +361,41 @@ def test_read_only(self):
...
+
+class TestFileTagging(ApiTestCase):
+    def setUp(self):
+        super(TestFileTagging, self).setUp()
+        self.user = AuthUserFactory()
+        self.node = ProjectFactory(creator=self.user)
+        self.file1 = api_utils.create_test_file(
+            self.node, self.user, filename='file1')
+
+    def test_tags_add_and_remove_properly(self):
```
brianjgeiger: I know this was copied almost verbatim from an existing test, but it should _probably_ be broken up into three tests. I'd have felt worse about this if the git blame said my name, but it says Dawn's, so maybe it wasn't entirely my fault.

Also, do file tags generate any logs in general? If so, please check for the logs.

website/files/models/base.py
Ghalko: `raise OperationFailure(details["err"], details["code"])
OperationFailure: Cannot transition from not multi key to multi key in multi statement transaction`

This is the failure when a list is indexed.

brianjgeiger:
- [ ] Update PR description with screenshot of docs and JSON of requests/responses
- [ ] Break up test into three tests
- [ ] Verify if logs are generated and test for logs if applicable
- [ ] Verify performance implications (or lack thereof) of the inability to add the index

Mostly good. Pass done. 🐧

website/files/models/base.py
brianjgeiger: I know, I saw the link in the commit message. But what does it mean as far as performance? Seto can probably answer the question quickly.

api_tests/files/views/test_file_detail.py
Ghalko: Done and done.

api_tests/files/views/test_file_detail.py
```diff
@@ -382,20 +380,51 @@ def test_tags_add_and_remove_properly(self):
...
-        res = self.app.put_json_api(url, payload, auth=self.user.auth)
+
+    def test_tags_add_and_remove_properly(self):
+        self.app.put_json_api(self.url, self.payload, auth=self.user.auth)
+        self.payload['data']['attributes']['tags'] = []
+        res = self.app.put_json_api(self.url, self.payload, auth=self.user.auth)
         assert_equal(res.status_code, 200)
         assert_equal(len(res.json['data']['attributes']['tags']), 0)
+
+    def test_put_wo_tags_doesnt_remove_tags(self):
```
brianjgeiger: Typo with `wo`?

api_tests/files/views/test_file_detail.py
```diff
@@ -382,20 +380,51 @@ def test_tags_add_and_remove_properly(self):
...
+        assert_equal(res.status_code, 200)
+        # Ensure adding tag data is correct from the PUT response
+        assert_equal(len(res.json['data']['attributes']['tags']), 1)
+        assert_equal(res.json['data']['attributes']['tags'][0], 'goofy')
+
+    def test_add_tag_adds_log(self):
+        count = len(self.node.logs)
+        self.app.put_json_api(self.url, self.payload, auth=self.user.auth)
+        assert_equal(len(self.node.logs), count + 2)
+        assert_equal(NodeLog.FILE_TAG_ADDED, self.node.logs[-2].action)
```
brianjgeiger: Did the decorator not work well for this?

brianjgeiger:
Looks good. 🎉 🐧 🎉

api_tests/files/views/test_file_detail.py
Ghalko: No. 'without' without the '/'

brianjgeiger: Oh, I see.

api/base/filters.py
```diff
@@ -377,6 +377,13 @@ def get_filtered_queryset(self, field_name, params, default_queryset):
                 item for item in default_queryset
                 if params['value'].lower() in getattr(item, field_name, {}).lower()
             ]
+        elif isinstance(field, JSONAPIListField):
+            return_val = [
+                item for item in default_queryset
+                if params['value'].lower() in [
+                    i.lower for i in getattr(item, field_name, [])
```
sloria: Doesn't this need to be `i.lower()`?

api_tests/files/views/test_file_detail.py
sloria: @Ghalko Why are you checking to the second-to-last log rather than the latest log?

api/files/serializers.py
sloria: @Ghalko Where do you add the indices?

sloria:
Pass finished. Just a couple questions and a possible bug https://github.com/CenterForOpenScience/osf.io/pull/5704/files#r66259600 (might also want to double check that the tests are covering that bug).

api/base/filters.py
samchrisinger: In Bryan's absence I'm going to fix and test this.

api/files/serializers.py
brianjgeiger: He had added the indices, then it broke all kinds of things and he had to remove them. There's details in the PR somewhere.

brianjgeiger: Or I thought so. I know we had chatted about it, but I can't find the discussion or the point at which tests failed and he changed the code back. Hmm.

api/base/filters.py
```diff
@@ -377,6 +377,13 @@ def get_filtered_queryset(self, field_name, params, default_queryset):
                 item for item in default_queryset
                 if params['value'].lower() in getattr(item, field_name, {}).lower()
             ]
+        elif isinstance(field, JSONAPIListField):
```
samchrisinger: @brianjgeiger do you think this could just be `ser.ListField`? Doesn't look like there's any practical difference between the two classes here

brianjgeiger: That's probably fine.

api/nodes/serializers.py
```diff
@@ -264,8 +264,7 @@ def update(self, node, validated_data):
         auth = get_user_auth(self.context['request'])
         old_tags = set([tag._id for tag in node.tags])
         if 'tags' in validated_data:
-            current_tags = set(validated_data.get('tags'))
-            del validated_data['tags']
```
sloria: Why was this line removed?

sloria: It seems like it is necessary for the check ~L279 to be correct, but I'm not sure.

samchrisinger: @sloria he's popping the key out of the dict

sloria: Oh, fair enough. Sorry for the noise.

api/files/serializers.py
sloria: It's not enough to set `index=True` on the field because the field is a list field, and Toku doesn't like indices on list fields before there are data in the database (I suspect you saw test failures).

For special cases like this, we document the indices in scripts/indices.py.

brianjgeiger: Okay, cool. Yeah, that was my big question about that.

PR #5705, opened 2016-05-20
Remove overflow scrollbar on Wiki Edit pane [OSF-5970]

baylee-d:
## Purpose

On the Wiki Edit pane, the setup of the Revert/Save buttons made the box have an overflow.  This was unnecessary for the interface, and didn't make it very user-friendly.  
scroll_bar

The purpose of this PR is to remove the scrollbar on the Edit pane.  The scrollbar appeared by default on the pane when the page loaded.  

## Changes

Currently when a user goes to the Edit pane on their Wiki page, the scrollbar will appear by default on the Edit pane.  All of the changes made for this PR involve fixing the styling on the Edit pane in order to remove the scrollbar.  

The main changes included getting rid of the fixed height and overflow properties for the Edit pane. 
screen shot 2016-05-19 at 4 39 57 pm
 

## Side effects

Shouldn't be any

## Ticket

<a href="https://openscience.atlassian.net/browse/OSF-5970">https://openscience.atlassian.net/browse/OSF-5970</a>

### DIFF
--- a/website/static/css/style.css
+++ b/website/static/css/style.css
@@ -373,13 +373,18 @@ a {
 
 /* OSF-panel */
 
-/* 
-formatting for the osf-panel in the Wiki.
-*/
 .osf-panel {
     border: 1px solid #ddd;
     display: block;
     position: relative;
+	overflow-x: hidden;
+	overflow: auto;
+	height: 764px;
+}
+
+/* removes the automatic height and overflow properties from the Edit Wiki pane */
+.osf-panel-edit {
+	height: auto;
 }
 
 @media (max-width: 991px) {
--- a/website/static/css/style.css
+++ b/website/static/css/style.css
@@ -375,16 +375,16 @@ a {
 
 .osf-panel {
     border: 1px solid #ddd;
+    height: 764px;
+    overflow: auto;
     display: block;
     position: relative;
-	overflow-x: hidden;
-	overflow: auto;
-	height: 764px;
+    overflow-x: hidden;
 }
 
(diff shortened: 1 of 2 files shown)


brianjgeiger:
Do a search on the code to see if .osf-panel is being used elsewhere. Also, you could add a class to this particular pane (say .collaborative-edit-panel) and apply your CSS changes to that instead. 

Pass done. 🐧

baylee-d:
I did a quick search on the page, and I found a few other panels that use the same .osf-panel class.  Instead, I changed the .osf-panel class back to the way it originally was and instead just made a new class for that individual panel with the new CSS changes in it.
Hopefully that'll solve any issues that might have come from changing the .osf-panel class.
________________________________

website/static/css/style.css
```diff
@@ -372,13 +372,19 @@ a {
...
 /* OSF-panel */
+
 .osf-panel {
     border: 1px solid #ddd;
-    height: 764px;
-    overflow: auto;
     display: block;
     position: relative;
-    overflow-x: hidden;
+	overflow-x: hidden;
```
brianjgeiger: Spacing here is off. Also, if you revert it back to what it was before rather than moving/reordering the lines, you won't have the unnecessary diff and you'll be less likely to cause a merge conflict.

brianjgeiger:
Other than that, this seems to be working the way I'd expect. Nearly there, pass done. 🐧

mfraezz:
LGTM. RTM when the build passes ⭐

PR #5707, opened 2016-05-20
unconfigured add-ons display widget to non-contributors [OSF-4161]

zachjanicki:
## Purpose

Currently users who do not have the ability to contribute to a project can see when that project has unconfigured add-ons. After this change, they will not be able to see what is not configured. 

## Changes

non-contributors can no longer see add-ons that are not configured

BEFORE:
screen shot 2016-05-20 at 2 39 54 pm

AFTER: (with one add-on set to configured)
screen shot 2016-05-20 at 2 40 14 pm

## Side effects

None.

## Ticket

https://openscience.atlassian.net/browse/OSF-4161

### DIFF
--- a/website/templates/project/addon/widget.mako
+++ b/website/templates/project/addon/widget.mako
@@ -1,5 +1,5 @@
-<div class="panel panel-default" name="${short_name}">
-
+% if complete or 'write' in user['permissions']:
+    <div class="panel panel-default" name="${short_name}" id="div">
             <div class="panel-heading clearfix">
                 <h3 class="panel-title">${full_name}</h3>
                 <div class="pull-right">
@@ -9,6 +9,9 @@
 
                 </div>
             </div>
+    % else:
+        <div>
+    % endif
 
     % if complete:
 
@@ -16,7 +19,7 @@
             ${self.body()}
         </div>
 
-    % else:
+    % elif not complete and 'admin' in user['permissions']:
 
         <div mod-meta='{
                 "tpl": "project/addon/config_error.mako",
--- a/website/templates/project/addon/widget.mako
+++ b/website/templates/project/addon/widget.mako
@@ -19,7 +19,7 @@
             ${self.body()}
         </div>
 
-    % elif not complete and 'admin' in user['permissions']:
+    % elif not complete and 'write' in user['permissions']:
 
         <div mod-meta='{
                 "tpl": "project/addon/config_error.mako",


brianjgeiger:
You have your other PR mixed in with this one. Get that separated out first, please. Pass done. 🐧

website/templates/project/addon/widget.mako
```diff
@@ -1,5 +1,5 @@
-<div class="panel panel-default" name="${short_name}">
-
+% if complete or 'write' in user['permissions']:
```
brianjgeiger: The 'write' in user['permissions'] is throwing me here. Can you explain why this block shows for all write users even if it's not complete?

zachjanicki: I might have a misunderstanding of OSF permissions... The idea is that the widget will show if it has been completely configured and the user has permission to edit the project. Should it instead be 'admin'? Do contributors have 'admin' rights or just 'write' rights?

brianjgeiger: Contributors potentially have `read` permissions. But because your `if` is an `or`, it's showing if it's complete or if it's not complete but the user has write permissions. Proooobably this should just be `if complete`. However… (see comment on the code farther down that I'm about to put in)

website/templates/project/addon/widget.mako
```diff
@@ -9,14 +9,17 @@
...
+    % endif
 
     % if complete:
 
         <div class="panel-body">
             ${self.body()}
         </div>
 
-    % else:
+    % elif not complete and 'admin' in user['permissions']:
```
brianjgeiger: …users with write permission can have the ability to configure addons, I believe. So play around with what you can do as a user with different permissions as far as configuring addons and then revisit this logic.

zachjanicki: Are you saying that this line should read 
```% elif not complete and 'write' in user['permissions']```
?

Comment (brianjgeiger) [inline website/templates/project/addon/widget.mako]: No, I'm asking under what circumstances that block should show so we can figure out what the logic should be. Could be the `or` is wrong, could be the `write` is wrong, could be both, could be something else.

Comment (zachjanicki) [inline website/templates/project/addon/widget.mako]: the block should be showing if the addon is properly configured or the current user has permission to configure the addon. So I would think that as long as the user has write permission or it is complete, it should be displayed.

Comment (brianjgeiger) [inline website/templates/project/addon/widget.mako]: Okay, then it looks like you know what to do to change that.

File: website/templates/project/addon/widget.mako (hunk 1 above)
Comment (sloria) [inline website/templates/project/addon/widget.mako]: Nici has commented on the desired UI for contributors of the different permission levels here: https://openscience.atlassian.net/browse/OSF-4161 . 

I think this implements that behavior, but i'm not completely sure. @zachjanicki Can you please verify?

File: website/templates/project/addon/widget.mako [hunk 3]
```diff
@@ -9,14 +9,17 @@
...
+    % endif
 
     % if complete:
 
         <div class="panel-body">
             ${self.body()}
         </div>
 
-    % else:
+    % elif not complete and 'write' in user['permissions']:
```
brianjgeiger: Strictly speaking, the `not complete` in this isn't really doing anything, since you wouldn't get to this branch unless complete were false. This could be re-written as `elif 'write' in user['permissions']:`. It's a trivial change either way.

website/templates/project/addon/widget.mako
```diff
@@ -9,14 +9,17 @@
 
                 </div>
             </div>
+    % else:
```
brianjgeiger: When I looked at this as a whole file and not in files changed view, I realized that the %else blocks etc could be moved to the left one tab. The first if block starts all the way to the left, and everything else is indented one extra level, so it took me a minute to figure out what was going on. This is also super minor.

brianjgeiger:
A couple of tiny, tiny things, but let's go ahead and make this perfect while we're at it. Be sure to manually test after the changes are done, but if it still works and tests pass, this will be ready to go. 🐧

brianjgeiger:
Based on these test failures, might be worth pulling from develop and ensuring api tests (and other tests) run locally before doing the push.

PR #5708, opened 2016-05-20
[Feature] Allow embargo registrations at 3 days out [OSF-4304]

mluke93:
## Purpose

When attempting to embargo a Registration, if the user attempts to pick a date 3 days away (as is required to Embargo) after 8pm, the system informs them that they must pick a date at least 3 days away, and won't let them pick that date (ie. On August 31st 8:36pm, the calendar does not recognize September 3rd as a valid date). The problem lied in the timezone difference between local time and UTC time. There is a four hour difference (hence why it had a problem starting at 8pm). So the valid date calculation needed to account for the time zone difference.

## Changes

- Added the Timezone Offset when calculating the chosen end date in milliseconds
- Fixed some of formatting.

## Side effects

- The max length (4 years) might have a similar issue to this.

## Ticket

https://openscience.atlassian.net/browse/OSF-4304

### DIFF
--- a/website/static/js/registrationModal.js
+++ b/website/static/js/registrationModal.js
@@ -48,27 +48,25 @@ var RegistrationViewModel = function(confirm, prompts, validator) {
         return moment(new Date(self.pikaday())).subtract(utcOffset, 'minutes');
     });
 
-    self.minimumTimeValidation = function (embargoLocalDateTime) {
+    self.minimumTimeValidation = function (x, y, embargoLocalDateTime) {
         var minEmbargoMoment = getMinimumDate(embargoLocalDateTime),
             endEmbargoMoment = self.embargoEndDate();
         return minEmbargoMoment.isBefore(endEmbargoMoment);
     };
 
-    self.maximumTimeValidation = function (embargoLocalDateTime) {
+    self.maximumTimeValidation = function (x, y, embargoLocalDateTime) {
         var maxEmbargoMoment = getMaximumDate(embargoLocalDateTime),
             endEmbargoMoment = self.embargoEndDate();
         return maxEmbargoMoment.isAfter(endEmbargoMoment);
     };
 
     var validation = [{
-        validator: function () {
-            return self.minimumTimeValidation();
-        },
+        validator: self.minimumTimeValidation
+        ,
         message: 'Embargo end date must be at least three days in the future.'
     }, {
-        validator: function () {
-            return self.maximumTimeValidation();
-        },
+        validator: self.maximumTimeValidation
+        ,
         message: 'Embargo end date must be less than four years in the future.'
     }];
     if(validator) {
--- a/website/static/js/registrationModal.js
+++ b/website/static/js/registrationModal.js
@@ -1,9 +1,9 @@
 var ko = require('knockout');
+var moment = require('moment');
 var pikaday = require('pikaday');
 require('pikaday-css');
 var bootbox = require('bootbox');
 var $ = require('jquery');
-var moment = require('moment');
 var $osf = require('js/osfHelpers');
 var language = require('js/osfLanguage').registrations;
 
@@ -11,10 +11,6 @@ var template = require('raw!templates/registration-modal.html');
 $(document).ready(function() {
     $('body').append(template);
 });
-
-// TODO(hrybacki): Import min/max dates from website.settings
-var TWO_DAYS_FROM_TODAY_TIMESTAMP = new Date().getTime() + (2 * 24 * 60 * 60 * 1000);
-var FOUR_YEARS_FROM_TODAY_TIMESTAMP = new Date().getTime() + (1460 * 24 * 60 * 60 * 1000);
 
 var MAKE_PUBLIC = {
     value: 'immediate',
@@ -24,6 +20,7 @@ var MAKE_EMBARGO = {
     value: 'embargo',
     message: 'Enter registration into embargo'
 };
+var utcOffset = moment().utcOffset();
 var today = new Date();
 
 var RegistrationViewModel = function(confirm, prompts, validator) {
@@ -48,21 +45,31 @@ var RegistrationViewModel = function(confirm, prompts, validator) {
         }
     );
     self.embargoEndDate = ko.computed(function() {
-        return new Date(self.pikaday());
+        return moment(new Date(self.pikaday())).subtract(utcOffset, 'minutes');
     });
 
+    self.minimumTimeValidation = function (embargoLocalDateTime) {
+        var minEmbargoMoment = getMinimumDate(embargoLocalDateTime),
+            endEmbargoMoment = self.embargoEndDate();
+        return minEmbargoMoment.isBefore(endEmbargoMoment);
+    };
+
+    self.maximumTimeValidation = function (embargoLocalDateTime) {
+        var maxEmbargoMoment = getMaximumDate(embargoLocalDateTime),
+            endEmbargoMoment = self.embargoEndDate();
+        return maxEmbargoMoment.isAfter(endEmbargoMoment);
+    };
+
     var validation = [{
-        validator: function() {
-            var endEmbargoMoment = moment(self.embargoEndDate()).subtract(moment().utcOffset(), 'm');
-            return endEmbargoMoment.valueOf() > TWO_DAYS_FROM_TODAY_TIMESTAMP;
+        validator: function () {
+            return self.minimumTimeValidation();
         },
         message: 'Embargo end date must be at least three days in the future.'
     }, {
-	validator: function() {
-            var endEmbargoMoment = moment(self.embargoEndDate()).subtract(moment().utcOffset(), 'm');
-        return endEmbargoMoment < FOUR_YEARS_FROM_TODAY_TIMESTAMP;
-	},
-	message: 'Embargo end date must be less than four years in the future.'
+        validator: function () {
+            return self.maximumTimeValidation();
+        },
+        message: 'Embargo end date must be less than four years in the future.'
     }];
     if(validator) {
         validation.unshift(validator);
@@ -104,10 +111,20 @@ RegistrationViewModel.prototype.show = function() {
 RegistrationViewModel.prototype.register = function() {
     this.confirm({
         registrationChoice: this.registrationChoice(),
-        embargoEndDate: this.embargoEndDate()
+        embargoEndDate: this.embargoEndDate(),
+        minimumTimeValidation: this.minimumTimeValidation(),
+        maximumTimeValidation: this.maximumTimeValidation()
     });
 };
 
 module.exports = {
     ViewModel: RegistrationViewModel
 };
+
+function getMinimumDate(embargoLocalDateTime) {
+    return moment(embargoLocalDateTime).add(2, 'days');
+}
+
+function getMaximumDate(embargoLocalDateTime) {
+    return moment(embargoLocalDateTime).add(4, 'years').subtract(1, 'days');
+}
--- a/website/static/js/registrationModal.js
+++ b/website/static/js/registrationModal.js
@@ -52,14 +52,14 @@ var RegistrationViewModel = function(confirm, prompts, validator) {
 
     var validation = [{
         validator: function() {
-            var endEmbargoDateTimestamp = self.embargoEndDate().getTime();
+            var endEmbargoDateTimestamp = self.embargoEndDate().getTime() + (self.embargoEndDate().getTimezoneOffset() * 60000);
             return endEmbargoDateTimestamp > TWO_DAYS_FROM_TODAY_TIMESTAMP;
         },
         message: 'Embargo end date must be at least three days in the future.'
     }, {
 	validator: function() {
             var endEmbargoDateTimestamp = self.embargoEndDate().getTime();
-	    return endEmbargoDateTimestamp < FOUR_YEARS_FROM_TODAY_TIMESTAMP;
+	        return endEmbargoDateTimestamp < FOUR_YEARS_FROM_TODAY_TIMESTAMP;
 	},
 	message: 'Embargo end date must be less than four years in the future.'
     }];
--- a/website/static/js/registrationModal.js
+++ b/website/static/js/registrationModal.js
@@ -61,12 +61,10 @@ var RegistrationViewModel = function(confirm, prompts, validator) {
     };
 
     var validation = [{
-        validator: self.minimumTimeValidation
-        ,
+        validator: self.minimumTimeValidation,
         message: 'Embargo end date must be at least three days in the future.'
     }, {
-        validator: self.maximumTimeValidation
-        ,
+        validator: self.maximumTimeValidation,
         message: 'Embargo end date must be less than four years in the future.'
     }];
     if(validator) {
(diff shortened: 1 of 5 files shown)


website/static/js/registrationModal.js
```diff
@@ -52,14 +52,15 @@ var RegistrationViewModel = function(confirm, prompts, validator) {
 
     var validation = [{
         validator: function() {
-            var endEmbargoDateTimestamp = self.embargoEndDate().getTime();
+            var timeZoneOffset = self.embargoEndDate().getTimezoneOffset() * (60 * 1000);
```
brianjgeiger: We should still be using moment.js in the site. There's probably a better way to do this using moment. The things I want to avoid are 1) using magic numbers; and 2) anything by hand related to time because time is the *worst*.

website/static/js/registrationModal.js
```diff
@@ -52,14 +52,15 @@ var RegistrationViewModel = function(confirm, prompts, validator) {
...
+            var endEmbargoDateTimestamp = self.embargoEndDate().getTime() + timeZoneOffset;
             return endEmbargoDateTimestamp > TWO_DAYS_FROM_TODAY_TIMESTAMP;
         },
         message: 'Embargo end date must be at least three days in the future.'
     }, {
 	validator: function() {
             var endEmbargoDateTimestamp = self.embargoEndDate().getTime();
-	    return endEmbargoDateTimestamp < FOUR_YEARS_FROM_TODAY_TIMESTAMP;
+	        return endEmbargoDateTimestamp < FOUR_YEARS_FROM_TODAY_TIMESTAMP;
```
brianjgeiger: Is this too many extra spaces?

brianjgeiger:
Pass done. Mostly just look into http://momentjs.com

website/static/js/registrationModal.js
mluke93: It looked wrong on my end, so I changed something. Let me go back and fix that, because you're definitely right. That doesn't look right at all.

mluke93:
Made changes to the validation so it now uses a moment to calculate the time chosen on the calendar (including UTC offsets). Also, made the same change for the 'Four Years' case, so it is consistent (before, you could sometimes pick exactly four years out, and sometimes you couldn't). You now must pick a date under 4 years out.

website/static/js/registrationModal.js
```diff
@@ -52,14 +53,14 @@ var RegistrationViewModel = function(confirm, prompts, validator) {
...
-            return endEmbargoDateTimestamp > TWO_DAYS_FROM_TODAY_TIMESTAMP;
+            var endEmbargoMoment = moment(self.embargoEndDate()).subtract(moment().utcOffset(), 'm');
+            return endEmbargoMoment.valueOf() > TWO_DAYS_FROM_TODAY_TIMESTAMP;
         },
         message: 'Embargo end date must be at least three days in the future.'
     }, {
 	validator: function() {
-            var endEmbargoDateTimestamp = self.embargoEndDate().getTime();
-	    return endEmbargoDateTimestamp < FOUR_YEARS_FROM_TODAY_TIMESTAMP;
+            var endEmbargoMoment = moment(self.embargoEndDate()).subtract(moment().utcOffset(), 'm');
```
brianjgeiger: Spacing in this function block seems all wrong. Four spaces to indent. https://cosdev.readthedocs.io/en/latest/style_guides/javascript.html#style

brianjgeiger:
Okay, *since* we're in here already and I'm having a hard time figuring out for sure what's going on, maybe it's worth doing just a bit more to make it better. Right now, limits are being set at https://github.com/CenterForOpenScience/osf.io/pull/5708/files#diff-9f8b0649d71a07ff6a3d91a6348fcb92R16 which looks like it's going to happen at instantiation rather than at validation, so if you load the page, wait three days, and then try to embargo for 30 minutes from now, it'll work, because you'll be past the three day wait time on account of waiting. Not the worst thing, but there'll be some other inconsistencies and there's no reason for it.

More importantly, it's really hard to tell from what's happening here what's in local time and what's in UTC time at which point, and so maybe we can make this a little bit better now that we have moment. Finally, there are no tests, which makes it especially difficult for me to be sure it's fixed properly. So add some story points to the issue and let's see if we can do this correctly.

Moment can add dates together by doing super readable things like `moment().add(7, 'days').add(1, 'months');` so we can take advantage of that. If we name the validator functions instead of declaring them anonymously in the validator, then we can write some unit tests that verify what will happen in different conditions.

So essentially I'm asking you to make:

- minimumTimeValidation(embargoLocalDateTime) function
- maximumTimeValidation(embargoLocalDateTime) function
- Unit tests for the two validation functions above to test normal functioning, plus some likely problems when people are using the browser in different time zones. (You can pass in the embargoLocalDateTime in whichever timezone that the person requests, that way you're not relying on someone upstream to convert from localtime - you want all the conversion to happen behind the tested function.)

Does this all make sense? It might not work exactly like I'm suggesting, but the important thing is to make it more readable and to make it tested.

brianjgeiger:
Pass done. 🐧

brianjgeiger:
You might also consider having separate functions to set the minimum time and the maximum time outside of the validator functions, and have your tests use that as well, so if we change the minimum time and maximum time in the future, the tests won't have to be adjusted. Since there's a todo near the code that sets the min/max times to improve that, we know it has a chance of being changed, so let's be prepared.

mluke93:
I think I see what you're going for. I will definitely go ahead and take a
crack at this. If I have any further questions I'll drop by and ask you.

On Tue, May 24, 2016 at 9:07 AM, Brian J. Geiger <notifications@github.com>
wrote:

> You might also consider having separate functions to set the minimum time
> and the maximum time outside of the validator functions, and have your
> tests use that as well, so if we change the minimum time and maximum time
> in the future, the tests won't have to be adjusted. Since there's a todo
> near the code that sets the min/max times to improve that, we know it has a
> chance of being changed, so let's be prepared.
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/CenterForOpenScience/osf.io/pull/5708#issuecomment-221262494>
>



-- 
Luke Marsh

brianjgeiger:
So you know, I'm working from home today but will be in tomorrow. However, there are folks who are ready to mentor, so just ask on osf-dev for them.

mluke93:
Good to know! Thanks for the advice.

mluke93:
Let me know if something isn't formatted correctly, or if you still want anything changed. The validation functions are inside of the registrationViewModel, while the getMin/MaxTime are outside. If you have a preference on location of these functions, I can go ahead and change that. Also added 6 tests (2 regular valid, 2 regular invalid, 2 timezone-based). The last two should work in whatever timezone the computer is in.

website/static/js/registrationModal.js
```diff
@@ -47,21 +45,31 @@ var RegistrationViewModel = function(confirm, prompts, validator) {
...
+        var maxEmbargoMoment = getMaximumDate(embargoLocalDateTime),
+            endEmbargoMoment = self.embargoEndDate();
+        return maxEmbargoMoment.isAfter(endEmbargoMoment);
+    };
+
     var validation = [{
-        validator: function() {
-            var endEmbargoDateTimestamp = self.embargoEndDate().getTime();
-            return endEmbargoDateTimestamp > TWO_DAYS_FROM_TODAY_TIMESTAMP;
+        validator: function () {
```
brianjgeiger: I would expect this could just be `validator: self.minimumTimeValidation,` or similar.

website/static/js/registrationModal.js
```diff
@@ -47,21 +45,31 @@ var RegistrationViewModel = function(confirm, prompts, validator) {
...
+            return self.minimumTimeValidation();
         },
         message: 'Embargo end date must be at least three days in the future.'
     }, {
-	validator: function() {
-            var endEmbargoDateTimestamp = self.embargoEndDate().getTime();
-	    return endEmbargoDateTimestamp < FOUR_YEARS_FROM_TODAY_TIMESTAMP;
-	},
-	message: 'Embargo end date must be less than four years in the future.'
+        validator: function () {
```
brianjgeiger: And here. The advantage of this is that, when you're debugging, you don't have an anonymous function in your call stack. It makes things much easier.

website/static/js/registrationModal.js
```diff
@@ -103,10 +111,20 @@ RegistrationViewModel.prototype.show = function() {
...
+        minimumTimeValidation: this.minimumTimeValidation(),
+        maximumTimeValidation: this.maximumTimeValidation()
     });
 };
 
 module.exports = {
     ViewModel: RegistrationViewModel
 };
+
+function getMinimumDate(embargoLocalDateTime) {
```
brianjgeiger: 👍

website/static/js/tests/registrationModal.test.js
```diff
@@ -104,4 +107,44 @@ describe('registrationModal', () => {
...
+            validDateTwoDaysTZ.setMinutes(validDateTwoDaysTZ.getMinutes() - validDateTwoDaysTZ.getTimezoneOffset());
+            var validDateTZFuture = new Date(validDateTwoDaysTZ);
+            validDateTZFuture.setDate(validDateTwoDaysTZ.getDate() + 3);
+            vm.pikaday(validDateTZFuture);
+            assert.isTrue(vm.minimumTimeValidation(validDateTwoDaysTZ));
+        });
+        it('returns true for date less than 4 years in the future, regardless of time zone', () => {
+            var validDateFourYearsTZ = new Date(2016, 0, 1);
+            validDateFourYearsTZ.setMinutes(validDateFourYearsTZ.getMinutes() - validDateFourYearsTZ.getTimezoneOffset());
+            console.log(validDateFourYearsTZ);
```
brianjgeiger: No console logs in PRs, please.

website/static/js/tests/registrationModal.test.js
```diff
@@ -104,4 +107,44 @@ describe('registrationModal', () => {
...
+            var validDateTZFutureYears = new Date(validDateFourYearsTZ);
+            validDateTZFutureYears.setDate(validDateTZFutureYears.getDate() + 1460);
+            console.log(validDateTZFutureYears);
```
brianjgeiger: Same here

brianjgeiger:
Tests are mostly there. Do the false versions of the tests with timezones, and do a couple variants of timezones (such as one that's west of UTC and one that's east).

brianjgeiger:
Overall, this is looking much more maintainable and comprehensible. Very good. Pass done. 🐧

mluke93:
Whoops! Sorry about the console logs. That was an oversight on my part. I used them for my own personal testing. I'll fix up those validator functions and add in the extra tests.

brianjgeiger:
Heh. Happens all the time. There's one developer who has to put money in a jar each time. Since this is your first offense, we won't do that to you. :)

mluke93:
Alright, let's have another go. Sorry about the delay, been working on our project. So, I've removed the anonymous function, but did have to add in the parameter workaround on minimumTimeValidation and maximumTimeValidation. From everything I've tried, it seems we either get an anonymous function or those extra two parameters we don't need. The tests have been updated and should reflect varying time zones. I believe that's all I've changed.

brianjgeiger:
Looks like you have some problems with tests. At the very least a few JSHint errors. 

screen shot 2016-06-01 at 2 49 47 pm

website/static/js/registrationModal.js
```diff
@@ -47,21 +45,29 @@ var RegistrationViewModel = function(confirm, prompts, validator) {
...
+        return maxEmbargoMoment.isAfter(endEmbargoMoment);
+    };
+
     var validation = [{
-        validator: function() {
-            var endEmbargoDateTimestamp = self.embargoEndDate().getTime();
-            return endEmbargoDateTimestamp > TWO_DAYS_FROM_TODAY_TIMESTAMP;
-        },
+        validator: self.minimumTimeValidation
+        ,
```
brianjgeiger: I imagine this was one of the hint errors, but move this comma up to the previous line.

mluke93:
Oh. Huh. Ran fine on my end. I'l check it out. Sorry about that.

website/static/js/tests/registrationModal.test.js
```diff
@@ -104,4 +107,90 @@ describe('registrationModal', () => {
...
+        });
+        it('returns true for date less than 4 years in the future, in a western timezone', () => {
+            var validDateFourYearsTZWest = moment().utcOffset(-4).toDate();
+            validDateFourYearsTZWest.setMinutes(validDateFourYearsTZWest.getMinutes() - validDateFourYearsTZWest.getTimezoneOffset());
+            var validDateTZFutureWest = new Date(validDateFourYearsTZWest);
+            validDateTZFutureWest.setDate(validDateFourYearsTZWest.getDate() + 1459);
+            vm.pikaday(validDateTZFutureWest);
+            assert.isTrue(vm.maximumTimeValidation(null, null, validDateFourYearsTZWest));
+        });
+        it('returns true for date less than 4 years in the future, in a western timezone', () => {
```
brianjgeiger: I think this should be `eastern`

website/static/js/tests/registrationModal.test.js
```diff
@@ -104,4 +107,90 @@ describe('registrationModal', () => {
...
+        });
+        it('returns false for date less than 2 days in the future, regardless of time zone', () => {
+            var invalidDateTwoDaysTZ = new Date(2016, 0, 1);
+            invalidDateTwoDaysTZ.setMinutes(invalidDateTwoDaysTZ.getMinutes() - invalidDateTwoDaysTZ.getTimezoneOffset());
+            var invalidDateTZFuture = new Date(invalidDateTwoDaysTZ);
+            invalidDateTZFuture.setDate(invalidDateTwoDaysTZ.getDate() + 1);
+            vm.pikaday(invalidDateTZFuture);
+            assert.isFalse(vm.minimumTimeValidation(null, null, invalidDateTwoDaysTZ));
+        });
+        it('returns false for date at least 4 years in the future, regardless of time zone', () => {
```
brianjgeiger: I think 'regardless of time zone' might not be the right thing to say here. Can you rephrase, or explain if the phrasing is exactly perfect?

brianjgeiger:
Super close. Pass don. 🐧

mluke93:
Hopefully this should do it.

brianjgeiger:
If travis passes, this is good to merge. 🎉 🐧 🎉

mluke93:
Woohoo!

brianjgeiger:
We have a travis test that runs tests on the code after it's been merged with develop. In this case, it looks like a Javascript test is failing in the post-merge scenario. 

screen shot 2016-06-01 at 5 50 32 pm

If you pull in develop and run Karma tests, you should see it and be able to fix it. Everything else looks good.

mluke93:
Huh. Interesting. Ill get on that first thing in the morning.

brianjgeiger:
This is good. It was a random test failure.

website/static/js/tests/registrationModal.test.js
```diff
@@ -104,4 +107,58 @@ describe('registrationModal', () => {
             assert.isTrue(vm.requestingEmbargo());
         });
     });
+    describe('#timeValidation', () => {
+        it('returns true for date more than 2 days in the future', () => {
```
sloria: Thanks for being thorough with testing.

PR #5710, opened 2016-05-20
[Feature] Recognizes URLs in project comments [OSF-598]

TomHeatwole:
## Purpose

Allow users to comment usable links by URL.

## Changes

Before comments are displayed, a function is run which searches the comment for URLs and replaces them with a link to that URL.

Before:

After:

## Side effects

It's possible that the regex expression could pick up things that are not URLs and try to create links. This is pretty unlikely and would be very rare if there is an error.

## Ticket

https://openscience.atlassian.net/browse/OSF-598


mfraezz:
I'm not sure that this is the right approach, but it's close. Your regex currently only detects well-formed urls. Matching any possible url with a regex is a [surprisingly](http://stackoverflow.com/questions/1410311/regular-expression-for-url-validation-in-javascript/1411800#1411800) [hard](http://stackoverflow.com/questions/827557/how-do-you-validate-a-url-with-a-regular-expression-in-python/827621#827621) [task](http://stackoverflow.com/questions/226505/question-about-url-validation-with-regex/226709#226709). Consider instead using a library like [linkifyjs](https://github.com/SoapBox/linkifyjs) to do this work for you instead.

PR #5714, opened 2016-05-23
add Saving state to save button on profile

billyhunt:
## Purpose

The save button has no signifier while a profile save is taking place.  Users do not know that the save request is being made. 

Now, after clicking save, The save button will grey and change to "Saving" until the request is completed, then it will change back to save.

saving
before request returns

save
after request returns

Also fixed a small bug in the social section of the profile where "Settings Updated" was on the same line as the save button.

## Ticket

https://openscience.atlassian.net/browse/OSF-3912

### DIFF
--- a/website/static/js/profile.js
+++ b/website/static/js/profile.js
@@ -729,7 +729,6 @@ SocialViewModel.prototype.submit = function() {
         );
     } else {
         this.showMessages(true);
-
     }
 };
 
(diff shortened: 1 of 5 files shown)


sloria:
You should only have to set `saving` in a few places, i.e. before and after the submit request completes.

```js
submit() {
    self.saving(true);
    postJSON()
        ...
        .always(function() { self.saving(false); })
}
```

billyhunt:
Good idea.  Also, should I do this all through the profile?  The update email and whatever.  Can you think of a way to make a helper that everyone can easily use?

sloria:
In the interest of keeping the PR focused, it's fine to change just this one button.

I'm not sure how a helper would work or if it'd even be helpful. If you have an idea, go for it, but I think the above suggested approach is fine.

PR #5715, opened 2016-05-23
API V2 Node file list filtering by size not working [OSF-4948]

adlius:
## Purpose
This PR fix the problem of API V2 not being able to filter by size of the file

## Changes
The root cause of this bug is that when the user tries to query by filesize, a parameter is passed through the url. In filters.py, the parameter passed by the url is read as a string, instead of an int, thus causing the filter function "get_filtered_queryset()" not being able to match by size as size should be an int. Change is made to convert the parameter from string to int.

In addition, one unit test and a helper method for the unit test is added to verify filtering by size works.

## Side effects
None

## Ticket
https://openscience.atlassian.net/browse/OSF-4948


api_tests/nodes/views/test_node_files_list.py
```diff
@@ -358,6 +358,16 @@ def test_node_files_are_filterable_by_kind(self):
...
 
+    def test_node_files_are_filterable_by_size(self):
+        self.file = api_utils.create_test_file_for_size_filter(self.project, self.user)
+
+        url = '/{}nodes/{}/files/osfstorage/?filter[size]=123'.format(API_BASE, self.project._id,)
+        res = self.app.get(url, auth=self.user.auth)
+        print(res)
+        assert_equal(res.status_code, 200)
+        assert_equal(res.json['data'][0]['attributes']['name'], 'test_file')
+        assert_equal(res.json['data'][0]['attributes']['size'], 123)
```
brianjgeiger: Can you get the size off of self.file directly rather than using `123`? The problem is that someone could change the size in the helper function and then the test would fail.

api_tests/utils.py
```diff
@@ -15,3 +15,30 @@ def create_test_file(node, user):
         'contentType': 'img/png'
     }).save()
     return test_file
+
+
+def create_test_file_for_size_filter(node, user):
```
brianjgeiger: This is super handy, though we have factories for most things (UserFactory, ProjectFactory, etc). Look at making a factory like the others rather than doing it this way.

brianjgeiger:
Pass done. Good start, but let's look at making it a bit more standard with our other test fixtures. 🐧

api_tests/utils.py
adlius: @brianjgeiger I did some search within our existing codebase and no FileFactory exists for OsfStorageFile and I kept digging to try to figure out why. It seems that a root node of a folder exists as the root for all OsfFileStorageNodes. Making a factory would only generate an instance of OsfStorageFile. However for the tests to work, we need to append the instance to the root node. The existing "append_file()" method of the root node only takes in a filename as an argument, therefore creating an instance with such filename and append it to the root node. And there is no other method that would take a File instance as an argument and append it to the root node. Creating such a method would mean adding an extra method to the codebase which would only be used for the unit tests.
The best way I can think of is moving the methods in the api_tests/utils.py to the test file itself and make those them private methods for the test classes. I checked and discovered such solution is consistent with other test files regarding osfstorage.

brianjgeiger: Cool, thank you.

api_tests/utils.py
```diff
@@ -14,4 +14,4 @@ def create_test_file(node, user):
         'size': 1337,
         'contentType': 'img/png'
     }).save()
-    return test_file
+    return test_file
```
brianjgeiger: Did you run Flake8 on this before you pushed? This looks like it might fail on Travis.

brianjgeiger:
It did *not* fail on Travis for Flake8, although it did fail on a weird test that shouldn't have anything to do with this. Re-running to see if it happens again. Otherwise this is good. 🎉 🐧 🎉

brianjgeiger:
Oh, I know that error. It happens 1 in a 1000 times. Not a concern. It should pass easily.

api/base/filters.py
```diff
@@ -368,7 +368,7 @@ def get_filtered_queryset(self, field_name, params, default_queryset):
         if isinstance(field, ser.SerializerMethodField):
             return_val = [
                 item for item in default_queryset
-                if self.FILTERS[params['op']](self.get_serializer_method(field_name)(item), params['value'])
+                if self.FILTERS[params['op']](self.get_serializer_method(field_name)(item), int(params['value']))
```
sloria: Can we assume that the value should always be cast to an `int`? This is handling `SerializerMethodFields`, which may return any type of data.

sloria:
Pass finished. One issue to address: https://github.com/CenterForOpenScience/osf.io/pull/5715/files#r64802584

api/base/filters.py
brianjgeiger: I talked with Fitz about this, and currently all the providers are sending back ints. We could, theoretically, sign on a provider that returns size in something non-integral, but a) that would be dumb; and b) we could cross that bridge when we came to it.

brianjgeiger: 'Cause really, who's going to send their size via API in TB or something dumb like that? Not even [pick your least favorite provider] does that.

sloria: Even if all the values we are currently receiving are ints, there's nothing guaranteeing that the SerializerMethodField will return an int. This might lead to a hard-to-trace bug in the future.

brianjgeiger: Yeah, Steve is right. Ideally the size field should be a more specific field type if possible.

brianjgeiger: The `extra` field should be a RestrictedDictSerializer and the size field should be a NumberField or similar. That should help. See /api/logs/serializers.py for examples of the RestrictedDictSerializer.

api/base/filters.py
```diff
@@ -365,11 +366,16 @@ def get_filtered_queryset(self, field_name, params, default_queryset):
...
         field_name = self.convert_key(field_name, field)
 
-        if isinstance(field, ser.SerializerMethodField):
+        if isinstance(field, ser.SerializerMethodField) and not isinstance(field, SerializerIntegerField):
             return_val = [
                 item for item in default_queryset
-                if self.FILTERS[params['op']](self.get_serializer_method(field_name)(item), params['value'])
+                if self.FILTERS[params['op']](self.get_serializer_method(field_name)(item), (params['value']))
             ]
+        elif isinstance(field, ser.SerializerMethodField) and isinstance(field, SerializerIntegerField):
```
brianjgeiger: Minor, but you can take out the first clause and just do `elif isinstance(field, SerializerIntegerField):`

api/files/serializers.py
```diff
@@ -14,6 +14,12 @@
 from api.base.serializers import Link, JSONAPISerializer, LinksField, IDField, TypeField
 from website.util import api_v2_url
 
+
+class SerializerIntegerField(ser.SerializerMethodField):
```
brianjgeiger: Let's rename this to `SerializerMethodIntegerField` so it's clear where it came from and how it behaves.

mfraezz:
Please also pull from develop and resolve the merge conflicts.

brianjgeiger:
Make those small changes and merge with develop and this should be ready. 🐧

PR #5717, opened 2016-05-23
Publish confirmation modals should say name of correct Dataverse [OSF-4556]

acshi:
## Purpose
The modal dialogs used to confirm that a user wants to publish a dataverse or dataset used to always talk about the Harvard dataverse, rather than the dataverse provider specifically being used.

## Changes
The host provider (for example, demo.dataverse.org) is found and this name is used in the modals instead of "Harvard".

After changes:
screen shot 2016-05-23 at 2 33 34 pm

## Side effects
No side effects. However, it should be noted that there are actually two modals that can be shown, and that without the fixes for OSF-4553 and OSF-6397, the second modal will never be able to be shown. (For when a dataverse is published but the specific dataset is not).

## Ticket
https://openscience.atlassian.net/browse/OSF-4556


mfraezz:
LGTM. Dropping this in RTM 💰

website/addons/dataverse/views.py
```diff
@@ -252,6 +252,7 @@ def _dataverse_root_folder(node_addon, auth, **kwargs):
     }
 
     # Get the dataverse host
+    # (stored in oauth_key because dataverse doesn't use that)
```
sloria: 👍  Thanks for the docco; this would otherwise have been really confusing.

PR #5721, opened 2016-05-23
Updates to Dataverse Publishing flow and UI [OSF-6397]  [OSF-4553] 

acshi:
(These two tickets were so similar that the code changes were essentially the same... and thus the pull requests were also merged)
## Purpose
Previous to this change, the publish option for a dataset would only be available if the entire dataverse had not yet been published. If the user publishes their dataverse/dataset once, then the publish option would be removed. Further changes to the draft dataverse would not be able to be published through the Fangorn interface, and a user would have to go to the dataverse website to do this.

## Changes
The behavior is implemented as follows:
Initially, with an unpublished dataverse/dataset, the publish option is available.
Once the dataset is published, the publish option is removed.
If the user uploads or deletes a file from the draft dataset, the publish option is again enabled to publish these new changes.
Further publishing and further changing continue this cycle.

## Side effects
This fix enables full testing of both modals in ticket https://openscience.atlassian.net/browse/OSF-4556 by making the publish button appear in the circumstances to make both modal versions appear.

## Ticket
https://openscience.atlassian.net/browse/OSF-4553
Old PR merged with this one: https://github.com/CenterForOpenScience/osf.io/pull/5719

## Purpose
To allow the publish button to appear at the correct times, and to prevent the "view"published" option from appearing unless the published version actually exists.

## Changes
This additional change just makes sure that the "view published" option is only available for published datasets.

## Side effects
No side effects.

## Ticket
https://openscience.atlassian.net/browse/OSF-6397

### DIFF
--- a/website/static/js/fangorn.js
+++ b/website/static/js/fangorn.js
@@ -757,7 +757,6 @@ function _fangornDropzoneSuccess(treebeard, file, response) {
 
     if (item.data.provider === "dataverse") {
         item.parent().data.datasetDraftModified = true;
-        
     }
 
     treebeard.redraw();
--- a/website/static/js/fangorn.js
+++ b/website/static/js/fangorn.js
@@ -754,6 +754,11 @@ function _fangornDropzoneSuccess(treebeard, file, response) {
     }
     var url = item.data.nodeUrl + 'files/' + item.data.provider + item.data.path;
     addFileStatus(treebeard, file, true, '', url);
+
+    if (item.data.provider === "dataverse") {
+        item.parent().data.datasetDraftModified = true;
+    }
+    
     treebeard.redraw();
 }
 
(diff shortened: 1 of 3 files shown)


mfraezz:
This appears to duplicate #5719 -- the last commit SHA is different, but contains the same changes. If this is the case, please close one of the PRs and update the other to indicate that it addresses both tickets (add a link in the `Ticket` section and add the number to the PR title).

website/static/js/fangorn.js
```diff
@@ -754,6 +754,11 @@ function _fangornDropzoneSuccess(treebeard, file, response) {
     }
     var url = item.data.nodeUrl + 'files/' + item.data.provider + item.data.path;
     addFileStatus(treebeard, file, true, '', url);
+
+    if (item.data.provider === "dataverse") {
```
mfraezz: This and below will make `jshint` fail with the message `Strings must use singlequote`.

mfraezz:
RTM whenever the build passes.

website/static/js/fangorn.js
mfraezz: This one as well. Not just below

PR #5722, opened 2016-05-24
Improve file widget "Rename" & "Create folder" buttons [OSF-6288]

el1t:
## Purpose

There was an inconsistency between the icon on the "rename" button when in the main menu vs. while entering a name. Also, low resolutions would reflow the text inside the buttons while entering a name, causing a malformed button. 

### Before

screen shot 2016-05-24 at 1 17 21 pm

screen shot 2016-05-24 at 1 17 27 pm

screen shot 2016-05-24 at 1 17 38 pm

### After

main menu
rename view
create folder view

- Use pencil icon for rename in both instances
- Remove text from "rename" and "create" buttons that appear when entering text

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-6288

### DIFF
--- a/website/static/js/fangorn.js
+++ b/website/static/js/fangorn.js
@@ -1654,7 +1654,7 @@ var FGItemButtons = {
                         onclick: function () {
                             mode(toolbarModes.RENAME);
                         },
-                        icon: 'fa fa-font',
+                        icon: 'fa fa-pencil',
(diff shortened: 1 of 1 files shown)


brianjgeiger:
Looks good unless @saradbowman says otherwise. 🎉 🐧 🎉

PR #5723, opened 2016-05-24
Add "project preview" link to unregistered user invite email [OSF-3211] 

Johnetordoff:
## Purpose

To allow an invited unregistered user to view the project or component they were invited to via a public or view-only link. After clicking this link they are sent to a to the linked project/component page where a non-dismissible alert appearing urging them to join the project with the set password token embedded in a link. 

## Changes

- Modifies HTML/JS in add_contributor_modal/contribAdder.js to display, pick and send an invite link to the server side.

- Modifes contributor.py's send_claim_email to send INVITE, INVITE_FORWARD, PENDING_INVITE_VERIFICATION with different arguments, including the invite link passed by the referrer. It also adds arguments to the url which allow the non-dismissible alert to be displayed.

- Modifies INVITE, INVITE_FORWARD, PENDING_INVITE_VERIFICATION email templates to include invite link, also refactors to make variable names a little more intuitive.

- Modifies node.py's _view_project to read url arguments passed in invitation link and add them to the user object where they can be read in project_header.mako and displayed.

- Modifes tests to accommodate new email parameters. 

## Side effects

- Changed z-index on add view-only modal so it pops over add contributor modal, so you add a view-only during the invite process.

- Not really a side effect, but in my rambling I noticed, add_unregistered_contributor isn't used anywhere in the code base except for tests, this is counterintuitive and should be changed, but I didn't want to creep on it.

## Ticket

https://openscience.atlassian.net/browse/OSF-3211

### DIFF
--- a/website/static/js/contribAdder.js
+++ b/website/static/js/contribAdder.js
@@ -231,15 +231,15 @@ AddContributorViewModel = oop.extend(Paginator, {
             dataType: 'json'
         }).done(function (response) {
 
-            response.node.private_links = response.node.private_links.map(function (pirvateLink) {
-                pirvateLink.date_created = new $osf.FormattableDate(pirvateLink.date_created);
-                pirvateLink.url = response.node.absolute_url + "?view_only=" +  pirvateLink.key;
+            response.node.private_links = response.node.private_links.map(function (privateLink) {
+                privateLink.date_created = new $osf.FormattableDate(privateLink.date_created);
+                privateLink.url = response.node.absolute_url + "?view_only=" +  privateLink.key;
 
-                pirvateLink.components = pirvateLink.nodes.map(function(component){
+                privateLink.components = privateLink.nodes.map(function(component){
                     return component.title;
                 }).join(", ");
 
-                return pirvateLink;
+                return privateLink;
             });
 
             self.inviteLinks(response.node);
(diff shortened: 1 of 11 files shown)


website/templates/emails/forward_invite.txt.mako
```diff
@@ -1,21 +1,23 @@
-Hello ${referrer.fullname},
+Hello ${refferer.fullname},
```
el1t: Just passing through, but this is not the correct spelling of referrer.

website/static/js/contribAdder.js
```diff
@@ -220,6 +221,33 @@ AddContributorViewModel = oop.extend(Paginator, {
...
+
+            response.node.private_links = response.node.private_links.map(function (pirvateLink) {
+                pirvateLink.date_created = new $osf.FormattableDate(pirvateLink.date_created);
+                pirvateLink.url = response.node.absolute_url + "?view_only=" +  pirvateLink.key;
+
+                pirvateLink.components = pirvateLink.nodes.map(function(component){
+                    return component.title;
+                }).join(", ");
+
+                return pirvateLink;
```
el1t: And `privateLink` :wink:

website/templates/emails/forward_invite.txt.mako
Johnetordoff: @el1t much appreciated!

PR #5726, opened 2016-05-24
Add Institutions to Multiple Nodes

billyhunt:
## Purpose

Add the ability to add or remove an institution to either a single node, or a node and its children, depending on what the user would like.  Also the UI has been modified.

## Changes

See images.

Added a dialog to allow the addition of an institution to multiple nodes.

screen shot 2016-05-24 at 12 18 25 pm
screen shot 2016-05-24 at 12 18 20 pm
screen shot 2016-05-24 at 12 18 13 pm

## Ticket


hmoco:
@billyhunt whats the behavior if a user only has admin permissions on a subset of components?

website/static/js/tests/institutionsProjectSettings.test.js
```diff
@@ -0,0 +1,100 @@
...
+'use strict';
+var assert = require('chai').assert;
+var $ = require('jquery');
+var m = require('mithril');
+var InstitutionProjectSettings = require('js/institutionProjectSettings');
+var bootbox = require('bootbox');
+
+var expectedTitle = 'Sample Project';
+
+    window.contextVars = $.extend(true, {}, window.contextVars, {
```
sloria: Tests should not have to use window.contextVars, but I realize that you have to do this because of existing code (outside the scope of this PR), so this is fine for now.

sloria: Minor: Fix indentation.

website/static/js/tests/institutionsProjectSettings.test.js
```diff
@@ -0,0 +1,100 @@
...
+        name: 'Sample Institution'
+    };
+
+    var modifyStub;
+    var modifyInstStub;
+    var bootboxStub;
+
+    var viewModel = new InstitutionProjectSettings.ViewModel(data);
+
+    it('user variables set', (done) => {
```
sloria: Minor: You only need to use `done` for async tests.

```js
it('does something rad', (done) => {
    someAsyncThing()
        .then(function(result) {
            assert.equal(....)
            done()
        });
});
```

website/static/js/tests/projectSettings.test.js
```diff
@@ -137,4 +139,34 @@ describe('ProjectSettings', () => {
...
+        var modifyStub;
+        var modifyInstStub;
+        var bootboxStub;
+
+        before(() => {
+            modifyInstStub = sinon.stub(viewModel, '_modifyInst');
+            bootboxStub = sinon.stub(bootbox, 'dialog');
+        });
+        after(() => {
+            viewModel.modifyChildrenDialog.restore();
```
sloria: modifyChildrenDialog isn't stubbed; is this supposed to be `bootboxStub.restore()`?

website/templates/project/settings.mako
```diff
@@ -316,46 +316,40 @@
...
-                                        </div>
-                                    </div>
-                                    <button data-bind="click: submitInst, css: {disabled: selectedInstitution() == null}" class="btn btn-success">Affiliate</button>
-                                </div>
-                           </div>
-                        % endif
+                            </br>
+                        <!-- ko if: availableInstitutions().length > 0 -->
+                        <label>Available Institutions: </label>
+                        <!-- /ko -->
```
sloria: Should the `<table>` be included in the `ko if`?

website/static/js/institutionProjectSettings.js
```diff
@@ -111,9 +195,39 @@ var ViewModel = function(data) {
...
+    var treebeardUrl = window.contextVars.node.urls.api + 'tree/';
+    self.viewModel.fetchNodeTree(treebeardUrl);
     $osf.applyBindings(this.viewModel, selector);
+
 };
 
-module.exports = InstitutionProjectSettings;
+module.exports = {
+    InstitutionProjectSettings: InstitutionProjectSettings,
+    ViewModel: ViewModel
```
sloria: 👍

billyhunt:
@hmoco, in reference to https://github.com/CenterForOpenScience/osf.io/pull/5726#issuecomment-221352393 It will only change components it can, and if the user does not have permission on the current node, then they cannot change anything, even if they have admin privileges on the children.

website/templates/project/settings.mako
billyhunt: I *think* that the institutions should show without the ability to change them.  @saradbowman is that correct?

website/static/js/institutionProjectSettings.js
```diff
@@ -2,105 +2,189 @@
...
 
 var $osf = require('js/osfHelpers');
+var projectSettingsTreebeardBase = require('js/projectSettingsTreebeardBase');
 
 var ViewModel = function(data) {
     var self = this;
     self.loading = ko.observable(true);
     self.showAdd = ko.observable(false);
     self.institutionHref = ko.observable('');
     self.userInstitutions = window.contextVars.currentUser.institutions;
```
sloria: This is the line that is causing an error in the karma tests you added. As suggested in my previous comment, we really shouldn't be using `contextVars` in this module, but that refactor is outside the scope of this PR. I've created a new ticket https://openscience.atlassian.net/browse/OSF-6416 . 

For now, you could use `lodash.get` which allows you to do nested attribute access;

```js
    self.userInstitutions = lodashGet('window.contextVars.currentUser.institutions') || [];
```

website/templates/project/settings.mako
sloria: Hm, but if `availableInstitutions().length` is `0`, then there will be nothing to iterate through in the `foreach` below, so all you'll have is 

```html
<table class="table">
    <tbody>
    </tbody>
</table>
```

which doesn't seem right.

website/templates/project/settings.mako
```diff
@@ -306,8 +306,8 @@
                     </div>
                     <div class="panel-body">
                         <div class="help-block">
-                            Projects can be affiliated with institutions that have created OSF for Institution accounts.
                             <!-- ko if: affiliatedInstitutions().length == 0 -->
+                            Projects can be affiliated with institutions that have created OSF for Institution accounts.
```
sloria: Typo (that existed before); "Institutions".

website/static/js/projectSettingsTreebeardBase.js
```diff
@@ -41,6 +41,11 @@ function getNodesOriginal(nodeTree, nodesOriginal) {
             registeredContributors.push(nodeTree.node.contributors[i].id);
         }
     }
+    var nodeInstitutions = [];
+    for (i=0; i < nodeTree.node.affiliated_institutions.length; i++) {
+            nodeInstitutions.push(nodeTree.node.affiliated_institutions[i].id);
```
sloria: Minor: overindentation.

website/project/views/node.py
```diff
@@ -962,10 +962,15 @@ def node_child_tree(user, node_ids):
                 'is_confirmed': contributor.is_confirmed
             })
 
+            affiliated_institutions = []
```
sloria: This variable shouldn't be in the `for` loop. Also, the for loop below can be refactored into a list comprehension.

```python
affiliated_institutions_serialized = [{'id': ....} for institution in node.affiliated_institutions]
```

website/project/views/node.py
```diff
@@ -993,6 +999,7 @@ def node_child_tree(user, node_ids):
             'category': node.category,
             'permissions': {
                 'view': can_read,
+                'is_admin': node.has_permission(user, 'read')
```
sloria: Minor: There's redundancy and confusing nesting here. There's `item['node']['is_admin']` and `item['permissions']['is_admin']`. "Write" permissions are under `item['node']` but read permissions are under `item['permissions']['view']`.

That said, this code was already in place. Feel free to refactor this in a followup PR.

website/static/js/projectSettingsTreebeardBase.js
```diff
@@ -41,6 +41,11 @@ function getNodesOriginal(nodeTree, nodesOriginal) {
             registeredContributors.push(nodeTree.node.contributors[i].id);
         }
     }
+    var nodeInstitutions = [];
+    for (i=0; i < nodeTree.node.affiliated_institutions.length; i++) {
```
sloria: This can be refactored using `.map`.

website/static/js/institutionProjectSettings.js
```diff
@@ -111,9 +195,39 @@ var ViewModel = function(data) {
     };
 };
 
+/**
+ * get node tree for treebeard from API V1
+ */
+ViewModel.prototype.fetchNodeTree = function(treebeardUrl) {
+        var nodesOriginal = {};
```
sloria: Minor: overindentation.

website/static/js/institutionProjectSettings.js
sloria: Ah, looks like this won't be an issue. Henrique made the refactor in https://github.com/CenterForOpenScience/osf.io/pull/5744 to remove usage of contextVars, so you'll no longer have to set contextVars in your tests.

Can you please update your tests to reflect this?

website/static/js/institutionProjectSettings.js
```diff
@@ -2,105 +2,189 @@
...
+            buttonLabel: self.isAddInstitution() ? 'Add institution' : 'Remove institution',
+            buttonColor: self.isAddInstitution() ? 'btn-success' : 'btn-danger'
+        };
+    });
+
+    self.modifyChildrenDialog = function (item) {
+        var message;
+        var addToOneMessage;
+        var addToAllMessage;
+        addToOneMessage = 'Add <b>' + item.name + '</b> to <b>' +  window.contextVars.node.title + '</b>.',
```
sloria: After you pull the latest changes, make sure to remove this use of contextVars.

website/static/js/institutionProjectSettings.js
```diff
@@ -111,9 +195,39 @@ var ViewModel = function(data) {
...
+        var nodesOriginal = {};
+        var self = this;
+        return $.ajax({
+            url: treebeardUrl,
+            type: 'GET',
+            dataType: 'json'
+        }).done(function (response) {
+            nodesOriginal = projectSettingsTreebeardBase.getNodesOriginal(response[0], nodesOriginal);
+            self.nodeParent = response[0].node.id;
+            self.hasChildren(Object.keys(self.nodesOriginal).length > 1);
```
sloria: I am seeing this modal even for projects that don't have components.

osf___osf4i_test_settings

The problem is this line; you are getting the keys for `self.nodesOriginal`, which is an `observable` that happens to have 5 methods, so this will always be `true`. I think this should be changed to `nodesOriginal`.

sloria:
Bug: When you have more than one institution to add and you click to add the first one, the UI adds the second one. When you refresh the page, it shows that the first one is added.

inst_bug

sloria:
Not a bug, but you can also see from the above gif there is a pause between when I click "Add" and when the UI updates. This is because the UI is waiting for the API request to finish.

It's better UX to have the UI update "optimistically", i.e. update immediately and assume that the request will succeed (the UI would revert the change if an error occurs).

This would be nice, but it isn't necessary before this PR is merged. A stopgap solution would be to disable the "Add" button while the request is happening and change the button text to "Adding".

website/static/js/institutionProjectSettings.js
```diff
@@ -2,105 +2,189 @@
...
+        var url = data.apiV2Prefix + 'institutions/' + item.id + '/relationships/nodes/';
+        var ajaxJSONType = self.isAddInstitution() ? 'POST': 'DELETE';
+        var nodesToModify = [];
+        if (self.modifyChildren()) {
+            for (var node in self.nodesOriginal()) {
+                nodesToModify.push({'type': 'nodes', 'id': self.nodesOriginal()[node].id});
+            }
+        }
+        else {
+                nodesToModify.push({'type': 'nodes', 'id': self.nodeParent});
```
sloria: Minor: overindentation.

website/static/js/institutionProjectSettings.js
```diff
@@ -2,105 +2,189 @@
...
+                }
+                self.affiliatedInstitutions(self.affiliatedInstitutions());
+            }
+            self.availableInstitutions(self.availableInstitutions());
+
         }).fail(function (xhr, status, error) {
-            $osf.growl('Unable to remove institution from this node. Please try again. If the problem persists, email <a href="mailto:support@osf.io.">support@osf.io</a>');
-            Raven.captureMessage('Unable to remove institution from this node!', {
+            $osf.growl('Unable to modify the institution on this node. Please try again. If the problem persists, email <a href="mailto:support@osf.io.">support@osf.io</a>');
+            Raven.captureMessage('Unable to modufy this institution!', {
```
sloria: Minor: typo

website/static/js/institutionProjectSettings.js
```diff
@@ -2,105 +2,189 @@
...
                 fields: {xhrFields: {withCredentials: true}}
             }
-        ).done(function (response) {
+        ).done(function () {
             var indexes = self.affiliatedInstitutions().map(function(each){return each.id;});
-            var removed = self.affiliatedInstitutions().splice(indexes.indexOf(item.id), 1)[0];
-            if ($.inArray(removed.id, self.userInstitutionsIds) >= 0){
-                self.availableInstitutions().push(removed);
+            if (self.isAddInstitution()) {
+                var index = indexes.indexOf(self.selectedInstitution());
```
sloria: Can you check this logic? This seems a likely culprit for this bug: https://github.com/CenterForOpenScience/osf.io/pull/5726#issuecomment-221649965

sloria:
Pass finished.

website/static/js/institutionProjectSettings.js
```diff
@@ -13,16 +12,15 @@ var ViewModel = function(data) {
...
+    self.userInstitutions = data.currentUser.institutions;
+    self.userInstitutionsIds = self.userInstitutions.map(function(item){return item.id;});
     self.selectedInstitution = ko.observable();
-    self.affiliatedInstitutions = ko.observable(window.contextVars.node.institutions);
-    self.affiliatedInstitutionsIds = self.affiliatedInstitutions().map(function (item) {
-        return item.id;
-    });
+    self.affiliatedInstitutions = ko.observable(data.node.institutions);
+
+    self.affiliatedInstitutionsIds = self.affiliatedInstitutions().map(function(item){return item.id;});
```
sloria: Does `self.affiliatedInstitutions` ever change? If it does, then I think affiliatedInstitutionIds will need to be computed.

website/static/js/institutionProjectSettings.js
```diff
@@ -13,16 +12,15 @@ var ViewModel = function(data) {
...
+    self.availableInstitutions = ko.observable(self.userInstitutions.filter(function(each){
```
sloria: Same here; I think this will need to be a computed if `affiliatedInstitutions` changes over time.

website/static/js/institutionProjectSettings.js
```diff
@@ -5,102 +5,174 @@ var Raven = require('raven-js');
...
+                });
+            }
         });
     };
+
+    self.pageTitle = ko.computed(function () {
+        return self.isAddInstitution() ? 'Add institution' : 'Remove institution';
+    });
+
+    self.hasThingsToAdd = ko.computed(function () {
```
sloria: Minor: This is probably unnecessary. You can just check 

```js
if (self.availableInstitutions.length)
```

where appropriate.

sloria:
Just a few more drive-by comments above. Mostly nits. The major blocker is https://github.com/CenterForOpenScience/osf.io/pull/5726#issuecomment-221649965.

website/static/js/institutionProjectSettings.js
```diff
@@ -5,102 +5,174 @@ var Raven = require('raven-js');
...
+                var added = self.availableInstitutions().splice(index, 1)[0];
+                self.affiliatedInstitutions().push(added);
+            }
+            else {
+                index = self.affiliatedInstitutionsIds().indexOf(item.id);
+                var removed = self.affiliatedInstitutions().splice(index, 1)[0];
                 self.availableInstitutions().push(removed);
-                self.availableInstitutions(self.availableInstitutions());
             }
+            self.availableInstitutions(self.availableInstitutions());
```
sloria: Does this line and the line below it do anything?

sloria:
Bug: when removing an institution from a project with components, the text next to the choices is incorrect:

osf___institution_testing_settings

website/templates/project/settings.mako
```diff
@@ -306,8 +306,8 @@
                     </div>
                     <div class="panel-body">
                         <div class="help-block">
-                            Projects can be affiliated with institutions that have created OSF for Institution accounts.
                             <!-- ko if: affiliatedInstitutions().length == 0 -->
+                            Projects can be affiliated with Institutions that have created OSF for Institution accounts.
```
sloria: The first "institutions" should not be capitalized.

billyhunt: I thought that was the typo.  What is the typo from https://github.com/CenterForOpenScience/osf.io/pull/5726#discussion-diff-64600674

billyhunt: Ohhh the second Institution should be plural.

website/static/js/institutionProjectSettings.js
sloria: Discussed on Flowdock: to correctly modify observableArrays, you need to make sure to not call them

```js
self.affiliInsti.push(...)   // not self.affiliInsti().push(...)
self.affiliInsti.splice(...) // not self.affiliInst().splice(...)
```

website/static/js/institutionProjectSettings.js
```diff
@@ -5,102 +5,173 @@ var Raven = require('raven-js');
...
-    };
+    //Has child nodes
+    self.hasChildren = ko.observable(false);
 
-    self.submitInst = function() {
-        var url = data.apiV2Prefix + 'nodes/' + data.node.id + '/relationships/institutions/';
-        var inst = self.selectedInstitution();
+    //user chooses to delete all nodes
+    self.modifyChildren = ko.observable(false);
+    self.title = 'Add Institution';
```
sloria: Is this used anywhere?

sloria:
Merged into `feature/multi-institution`.

PR #5729, opened 2016-05-24
[Feature] File view pane, edit panel shows earlier version [OSF-6300]

TomBaxter:
## Purpose:
[OSF-6300](https://openscience.atlassian.net/browse/OSF-6300)
Fix to get latest version in file editor panel

## Changes:
Update website/addons/base/views.py add version to key for file_guid_to_share_uuids

## Credits
Just like Brian said all along.

# Warning!
It may be best to stop the sharejs server during the merging of this code. If there is an existing editing session before merging and a new user starts editing afterwards, then the new session will not share(js) with the old session. Unexpected results may occur.

[#OSF-6300]

### DIFF
--- a/website/addons/base/views.py
+++ b/website/addons/base/views.py
@@ -662,12 +662,12 @@ def addon_view_file(auth, node, file_node, version):
 
     ret = serialize_node(node, auth, primary=True)
 
-    if file_node._id not in node.file_guid_to_share_uuids:
-        node.file_guid_to_share_uuids[file_node._id] = uuid.uuid4()
+    if file_node._id + '-' + version._id not in node.file_guid_to_share_uuids:
+        node.file_guid_to_share_uuids[file_node._id + '-' + version._id] = uuid.uuid4()
         node.save()
 
     if ret['user']['can_edit']:
-        sharejs_uuid = str(node.file_guid_to_share_uuids[file_node._id])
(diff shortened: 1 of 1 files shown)


brianjgeiger:
Does this work with wikis as well?

brianjgeiger:
Could you do some experimentation with the changeover to be sure the right procedure, then update the warning with a step-by-step?

brianjgeiger:
Pass done. 🐧

TomBaxter:
I did a quick audit of wiki code. This PR will have no effect on the wiki, nor does it appear to me that the issue this PR fixes could happen in the wiki. This issue is due to a mismatch between the stored add-on file and the sharejs instance. The wiki does not use OSF Storage or any other add-on. It stores all wiki contents in Mongo. I guess it's possible that it could run into a similar issue with Mongo/sharejs mismatch. But after about 30 minutes of testing I was unable to get it to lose track of the correct version. There are fewer layers of abstraction with the wiki implementation, so harder to to lose track.

brianjgeiger:
As discussed in person:

1. Create a wiki using the web ui
2. Leave that ShareJS session going
3. Update the wiki to a new version with different content using the osf shell (`invoke shell`)
4. Connect an incognito window to create a new ShareJS session to the same wiki

Let me know if the two windows have a) the same information and b) the correct information.

PR #5732, opened 2016-05-24
[Feature] Hides account deactivation button upon refresh if user has requested to deactivate. [OSF-2735]

TomHeatwole:
## Purpose
This prevents confusion when a user requests to deactivate their account then refreshes the page and sees that the deactivate button is still available.

## Changes
By adding a requestedDeactivation to the User class, the page checks if the account is pending deactivation on load, and in the case that the user has requested deactivation, it displays a message informing the user rather than the button.

Before:

After:

## Side effects
N/A

## Ticket
https://openscience.atlassian.net/browse/OSF-2735

### DIFF
--- a/framework/auth/core.py
+++ b/framework/auth/core.py
@@ -407,6 +407,9 @@ class User(GuidStoredObject, AddonModelMixin):
     # user language and locale data (e.g. 'en_US')
     locale = fields.StringField(default='en_US')
 
+    # whether the user has requested to deactivate their account
+    requestedDeactivation = fields.BooleanField(default=False)
+
     _meta = {'optimistic': True}
 
     def __repr__(self):
--- a/tests/models/test_user.py
+++ b/tests/models/test_user.py
@@ -363,7 +363,8 @@ class TestUserMerging(base.OsfTestCase):
             'mailing_lists',
             'verification_key',
             '_affiliated_institutions',
-            'contributor_added_email_records'
+            'contributor_added_email_records',
+            'requestedDeactivation'
         ]
 
         calculated_fields = {
@@ -410,14 +411,18 @@ class TestUserMerging(base.OsfTestCase):
         # from the explicit rules above, compile expected field/value pairs
         expected = {}
         expected.update(calculated_fields)
+        print(expected)
         for key in default_to_master_user_fields:
             expected[key] = getattr(self.user, key)
-
+        print(set(expected.keys()))
+        print("/n")
+        print(set(self.user._fields))
         # ensure all fields of the user object have an explicit expectation
         assert_equal(
             set(expected.keys()),
             set(self.user._fields),
         )
+
 
         # mock mailchimp
         mock_client = mock.MagicMock()
--- a/tests/models/test_user.py
+++ b/tests/models/test_user.py
@@ -411,18 +411,14 @@ class TestUserMerging(base.OsfTestCase):
         # from the explicit rules above, compile expected field/value pairs
         expected = {}
         expected.update(calculated_fields)
-        print(expected)
         for key in default_to_master_user_fields:
             expected[key] = getattr(self.user, key)
-        print(set(expected.keys()))
-        print("/n")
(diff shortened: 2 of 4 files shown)


tests/models/test_user.py
```diff
@@ -410,15 +411,19 @@ def test_merge(self, mock_get_mailchimp_api):
         # from the explicit rules above, compile expected field/value pairs
         expected = {}
         expected.update(calculated_fields)
+        print(expected)
```
abought: Please remove print statements before advancing this for code review.

website/profile/views.py
```diff
@@ -343,7 +343,8 @@ def user_account(auth, **kwargs):
         'user_id': user._id,
         'addons': user_addons,
         'addons_js': collect_user_config_js([addon for addon in settings.ADDONS_AVAILABLE if 'user' in addon.configs]),
-        'addons_css': []
+        'addons_css': [],
+        'requestedDeactivationgit ': user.requestedDeactivation
```
abought: Typo in variable name.

website/templates/profile/account.mako
```diff
@@ -129,9 +129,15 @@
...
                     <div class="panel-heading clearfix"><h3 class="panel-title">Deactivate Account</h3></div>
                     <div class="panel-body">
-                        <p class="alert alert-warning"><strong>Warning:</strong> This action is irreversible.</p>
+                        %if not requestedDeactivation:
+                            <p class="alert alert-warning"><strong>Warning:</strong> This action is irreversible.</p>
+                        %endif
                         <p>Deactivating your account will remove you from all public projects to which you are a contributor. Your account will no longer be associated with OSF projects, and your work on the OSF will be inaccessible.</p>
-                        <a class="btn btn-danger" data-bind="click: submit, css: success() === true ? 'disabled' : ''">Request deactivation</a>
+                        %if not requestedDeactivation:
+                             <a class="btn btn-danger" data-bind="click: submit, css: success() === true ? 'disabled' : ''">Request deactivation</a>
```
brianjgeiger: The minor downside to this method is that the view is different on reloading the page than when the request successfully returns. Better would be to ensure that knockout has access to requestedDeactivation and handle the view from there. While you're waiting to hear back from the server, deactivating the button is proper, but once you do, it would be better to switch to the requestedDeactivation branch.

Also, while we're here, what happens if you get an error from the server?

If this will take more than a day or so, let me know, because having this change is better than waiting too long on the ideal version.

brianjgeiger:
Pass done. Recommending looking at tightening up the UI just a bit more if it won't take long. Otherwise we'll go with this and open a new issue for the upgrade. 🐧

TomHeatwole:
Since we've broken up into project teams, I'll go ahead and submit this one as is and a second ticket can be created as needed.

brianjgeiger:
Then this is great. 🎉 🐧 🎉

framework/auth/core.py
```diff
@@ -407,6 +407,9 @@ class User(GuidStoredObject, AddonModelMixin):
     # user language and locale data (e.g. 'en_US')
     locale = fields.StringField(default='en_US')
 
+    # whether the user has requested to deactivate their account
+    requestedDeactivation = fields.BooleanField(default=False)
```
sloria: As per COSDev, use snake_case for variable names.

website/profile/views.py
```diff
@@ -343,7 +343,8 @@ def user_account(auth, **kwargs):
         'user_id': user._id,
         'addons': user_addons,
         'addons_js': collect_user_config_js([addon for addon in settings.ADDONS_AVAILABLE if 'user' in addon.configs]),
-        'addons_css': []
+        'addons_css': [],
+        'requestedDeactivation': user.requestedDeactivation
```
sloria: For consistency with the other keys, use snake_case.

sloria:
Pass finished; just minor casing issues. I'll fix those when I merge.

PR #5733, opened 2016-05-24
Feature/global notification settings [#OSF-5748]

erinspace:
Again piggybacking off of #5452 to implement small spec changes

screen shot 2016-05-26 at 10 19 12 am

## Purpose
- New Projects get default notifications set as instantly
- Global project notification settings on project page

### Summary across the three PRs this one incorporates:
- Add method that subscribes a user to notifications
- Call that method with signals when a node is first saved with the nodes' creator
- Also call that method when a contributor is added with that contributor
- Add a setting that decides if a notification subscription should be automatically created (by default yes)
- Adjust that setting to False in a special case for NotificationSubscrtiption related tests that don't create the subscriptions by default
- Disable automatic subscriptions on tests that benefit from creating them manually
- Alter text on contributor added, invite, forward invite, and public project emails to explain notifications and link to profile settings to update them
- add global subscription types for comments added and files updated to ```USER_SUBSCRIPTIONS_AVAILABLE```
in ```subscribe_user_to_notifications``` check for global settings before defaulting to email_transactional
- Update settings heading and add tooltip explaining global preferences settings

## Side effects
- Will not change settings for existing projects

## Ticket


website/project/model.py
```diff
@@ -2977,6 +3009,9 @@ def add_contributor(self, contributor, permissions=None, visible=True,
                 while len(user.recently_added) > MAX_RECENT_LENGTH:
                     user.recently_added.pop()
 
+            if contrib_to_add.is_registered and settings.ENABLE_NOTIFICATION_SUBSCRIPTION_CREATION:
```
mfraezz: From @sloria on the previous PR:

>The `is_registered` check should probably be moved to the `subscribe_user_to_notifications` method so that the caller doesn't have to do this check every time.

website/project/model.py
```diff
@@ -1551,6 +1580,9 @@ def save(self, *args, **kwargs):
                 save=True,
             )
 
+            if self.creator and settings.ENABLE_NOTIFICATION_SUBSCRIPTION_CREATION:
```
mfraezz: Also from [previous PR](https://github.com/CenterForOpenScience/osf.io/pull/5452#discussion_r61478806):

>Just curious: why is it necessary to check `if self.creator`? Looks like we're already assuming that `self.creator` exists [...]

website/templates/emails/contributor_added.txt.mako
```diff
@@ -1,7 +1,13 @@
+<%!
+    from website import settings
+%>
+
 Hello ${user.fullname},
 
 ${referrer_name + ' has added you' if referrer_name else 'You have been added'} as a contributor to the project "${node.title}" on the Open Science Framework: ${node.absolute_url}
 
+You will automatically be subscribed to notification emails for updates to this project. To change your email notification preferences, visit your project or your user settings: ${settings.DOMAIN + "settings/notifications/"}
```
mfraezz: Is this necessarily true? If a user sets their global default to no notifications, this line would probably be confusing. Consider passing a var to the email's template constructor, and adding logic to say either `will not be subscribed` or `will automatically be subscribed` depending on their setting.

mfraezz:
From the ticket -- desired settings, options, and their default values:

>Comments Added to Your Projects: (Options: Never, Daily Digest, Instantly) - Default is Instantly
>Files Updated on Your Projects: (Options: Never, Daily Digest, Instantly) - Default is instantly
>Replies to your comments: (Options: [Never,] Daily Digest, Instantly) - Default is Instantly

Please change the default for `Comments Added` to `Instantly`.

mfraezz:
Recently requested on the ticket:

>Also, please include text in this Configure box:
>
>"NOTE: Regardless of your selected preferences, OSF will continue to provide transactional and administrative service emails."

mfraezz:
Pass complete. Back to you, @erinspace 🌃

sloria:
Closing in favor of #5452

PR #5736, opened 2016-05-24
Feature/fix regressed registration logs[OSF-6409]

pattisdr:
Ticket https://openscience.atlassian.net/browse/OSF-6409

## Purpose

Some registration logs have regressed.  Registration-related logs were supposed to have a params['node'] field and a params['registration'] field.  Now some logs do not have params['registration'] field and params['node'] field points to the registration. This means original_node is incorrect since it is supposed to be the same as params['node'].

## Changes

Restores params['registration'] fields where missing.  Points params['node'] fields to node instead of registration where incorrect.  Original_node should be set to params['node'].

**params['registration'] may be missing and params['node'] and original_node incorrectly pointing to the registration for the following log actions:**
- embargo_approved
- embargo_cancelled
- embargo_terminated
- retraction_approved
- retraction_cancelled
- registration_cancelled

**params['registration'] may be missing**

- registration_approved 

## Side effects
Requires migration.

### DIFF
--- a/scripts/migrate_regressed_registration_logs.py
+++ b/scripts/migrate_regressed_registration_logs.py
@@ -1,0 +1,105 @@
+"""
+Some registration-related logs have regressed where there is 1) no params['registration'] field, 2) params['node'] field points to registration,
+and original node field is incorrect(since original_node = params['node'].
+
+This is causing some logs to link to registrations instead of nodes.
+"""
+
+import logging
+import sys
+from modularodm import Q
+
+from framework.transactions.context import TokuTransaction
+
+from website.models import Node, NodeLog, RegistrationApproval
+from website.app import init_app
+from scripts import utils as script_utils
+
+logger = logging.getLogger(__name__)
+
+
+def main():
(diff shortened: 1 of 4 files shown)


scripts/migrate_regressed_registration_logs.py
```diff
@@ -0,0 +1,105 @@
...
+    logger.info('Migrating registration_cancelled, registration_approved, retraction_cancelled, embargo_approved, embargo_cancelled, and embargo_terminated logs.')
+    count = 0
+    for log in targets:
+        node_id = log.params['node']
+        node = Node.load(node_id)
+        if node.is_registration:
+            log.params['node'] = get_registered_from(node)
+            log.params['registration'] = node._id
+            log.original_node = log.params['node']
+            logger.info('Updating params of log {}. params[node]={}, params[registration]={}, and original_node = {}'.format(log._id, log.params['node'], log.params['registration'], log.original_node))
```
sloria: Very useful. 👍

Issue #5737, opened 2016-05-25
Update invoke tasks for 'apiserver' and 'adminserver' to listen on public interfaces.


emetsger:
Currently the definitions for the `apiserver` and `adminserver` invoke tasks do not support listening on an interface other than `127.0.0.1`.  This is problematic when running the OSF services from Docker containers managed by a `docker-machine`, because the `docker-machine` will not be able to proxy requests to a container that isn't listening on a public-facing interface.  

If the `apiserver` and `adminserver` tasks are updated to accept a host name for binding, and maintain their default behavior when a host name is absent (i.e. bind to `127.0.0.1`), these two tasks can be instructed to listen on a public interface, and participate in a docker ecosystem that is managed by a `docker-machine`.

PR #5738, opened 2016-05-25
Updates the 'apiserver' and 'adminserver' tasks to enable listening on …

emetsger:
## Purpose

Enables the 'apiserver' and 'adminserver' invoke tasks to listen on a public interface, allowing them to participate in a `docker-machine`-managed ecosystem.  Requests from the local machine can be successfully proxied by a `docker-machine` to a Docker container running the apiserver if the apiserver is able to listen on a public interface.

## Changes

Updates the `apiserver` and `adminserver` tasks to accept an optional `host` argument, which defaults to a value of `127.0.0.1` (thereby maintaining current behavior when no `host` argument is supplied).  The format of the Python command is updated to supply the host value.

## Side effects

None that I can tell.

## Ticket

Proposed fix for issue [#5737](https://github.com/CenterForOpenScience/osf.io/issues/5737).

### DIFF
--- a/tasks/__init__.py
+++ b/tasks/__init__.py
@@ -76,10 +76,10 @@ def git_logs():
 
 
 @task
-def apiserver(port=8000, wait=True):
+def apiserver(port=8000, wait=True, host='127.0.0.1'):
     """Run the API server."""
     env = os.environ.copy()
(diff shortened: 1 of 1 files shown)


sloria:
Thanks again!

Issue #5739, opened 2016-05-25
Dev build fails with: "UnicodeEncodeError: 'ascii' codec can't encode characters in position 18-20: ordinal not in range(128)"


emetsger:
This error occurs when building the OSF codebase within Docker containers.  It seems to be an odd interaction between the terminal environment and the ascii characters used when emitting color output from the `webpack` task when in development mode.

When the `webpack` command is updated by removing the `--colors` argument, this error goes away, and builds in the development environment using Docker containers works.

PR #5740, opened 2016-05-25
Updates the 'webpack' task, removing the use of the --colors arg when…

emetsger:
## Purpose

Removes the `--colors` argument from the webpack task, avoiding an interaction with the terminal environment that causes a development build to fail with:

> UnicodeEncodeError: 'ascii' codec can't encode characters in position 18-20: ordinal not in range(128)

Color output will not be emitted by the `webpack` task in a development environment.  Builds are a bit more dreary, but won't fail when executed from a Docker container.

## Changes

Updates the `webpack` to be invoked with `--progress` no matter the environment (development or production).

## Side effects

The webpack task will be invoked with `--progress`, even in a development build.

## Ticket

Proposed fix for issue #5739.

### DIFF
--- a/tasks/__init__.py
+++ b/tasks/__init__.py
@@ -897,10 +897,7 @@ def webpack(clean=False, watch=False, dev=False):
         clean_assets()
     webpack_bin = os.path.join(HERE, 'node_modules', 'webpack', 'bin', 'webpack.js')
     args = [webpack_bin]
-    if settings.DEBUG_MODE and dev:
-        args += ['--colors']
-    else:
-        args += ['--progress']
+    args += ['--progress']
     if watch:
         args += ['--watch']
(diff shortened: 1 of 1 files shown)


sloria:
Thanks @emetsger . Looks like a good change; we certainly don't want webpack builds failing just to support colors. 

That said, it's nice to have colors in environments that support it; it makes errors very obvious while developing. Maybe we add a `color` parameter to the invoke task that defaults to `False`?

emetsger:
> Maybe we add a color parameter to the invoke task that defaults to False?

+1 

That sounds great!

sloria:
Added the `--colors` argument in https://github.com/CenterForOpenScience/osf.io/commit/587283336b460140fd93d76e3c867a26b8aa5f8a . Thanks for the PR.

emetsger:
Thanks @sloria!

PR #5742, opened 2016-05-25
Add Project Name to Component in New Dashboard [OSF-6027]

zamattiac:
## Purpose

Displays the root project of components showing up standalone on the quick view of projects and components. The view can be sorted and searched by root project as well. 

## Changes

- Embedded root project information in endpoint call to access its title and display it.
- Added column to the table and adjusted column width to maintain table width.
- Added sorting and searching by root project values
- New view:
image

and private projects: 
image

## Side effects

- In xs and sm view, sort can be by project and date.
- If there is some other reason that a root title request returns an error, it will still display as private project.

## Ticket

https://openscience.atlassian.net/browse/OSF-6027
[OSF-6027]


brianjgeiger:
I'm not super-excited about misrepresenting the privacy of a project because of an error response. It would probably be better if it indicated an error instead. _Project name unavailable_ or similar. I mean, the error shouldn't happen, but if it does, we should be clear about what happened.

website/static/js/home-page/quickProjectSearchPlugin.js
```diff
@@ -545,13 +570,26 @@ var QuickSearchNodeDisplay = {
             return m('.', args.eligibleNodes().slice(0, args.countDisplayed()).map(function(n){
                 var project = args.nodes()[n];
                 var numContributors = project.embeds.contributors.links.meta.total;
+                var title = project.attributes.title;
+                try {
+                    var root = project.embeds.root.data.attributes.title;
+                }
+                catch (err) {
+                    var root = "Private Project";
```
brianjgeiger: You *should* be able to get the error details in here to determine if the root is private or not. If not, obviously that's more complicated and we may have to skip it. But most likely you'll have `project.embeds.root.error` or similar that you can check.

brianjgeiger:
Pass done. 🐧 Let's see if we can eliminate the one side effect. If you haven't, you should also arrange for a demo to the Product team so they can verify it's working ideally. I suspect the styling may need to change a bit so that the most significant title is always bolded or similar (which looks like it could be a little more complicated). Mostly looks good, though.

brianjgeiger:
I had a chat with Nici, and let's ensure the case of not showing a real title for the root, make it italicized, and do the thing about always having the title of the actual node you'd visit be bolded so they're all consistently styled.

brianjgeiger:
And jshint is unhappy with your changes.

screen shot 2016-05-26 at 11 59 34 am

zamattiac:
## Current view:

1b5f6500-234c-11e6-9717-844121667783

brianjgeiger:
Looks good. 🎉 🐧 🎉

sloria:
The current implementation shows the root--rather than the parent--for components. This might be misleading for deeply-nested components.

For example, say I have two projects, `RPCB / Coffee causes cancer / Analysis` and `RPCB / Coffee cures cancer / Analysis`.

Both of these will be displayed as `RPCB / Analysis`.

It seems desirable to show the parent rather than the root. @saradbowman Thoughts?

website/static/js/home-page/quickProjectSearchPlugin.js
```diff
@@ -291,6 +291,15 @@ var QuickSearchProject = {
             return self.preSelectDirection() === sort ? 'selected' : 'not-selected';
         };
 
+        // Filtering on root project
+        self.rootMatch = function (node) {
+            try {
+                var root = project.embeds.root.data.attributes.title;
```
sloria: This variable doesn't appear to be used.

website/static/js/home-page/quickProjectSearchPlugin.js
```diff
@@ -291,6 +291,15 @@ var QuickSearchProject = {
...
+                return (node.embeds.root.data.attributes.title.toUpperCase().indexOf(self.filter().toUpperCase()) !== -1);
```
sloria: Rather than doing a try catch, you can use `lodash.get` to get a deeply nested attribute.

```js
var lodashGet = require('lodash.get');
//..
var rootTitle = lodashGet(node, 'embeds.root.data.attributes.title') || '';
return rootTitle.toUpperCase()....
```

website/static/js/home-page/quickProjectSearchPlugin.js
```diff
@@ -545,13 +555,35 @@ var QuickSearchNodeDisplay = {
             return m('.', args.eligibleNodes().slice(0, args.countDisplayed()).map(function(n){
                 var project = args.nodes()[n];
                 var numContributors = project.embeds.contributors.links.meta.total;
+                var title = project.attributes.title;
+                var root;
+                try {
+                    root = project.embeds.root.data.attributes.title;
```
sloria: Try to keep `try` blocks as constrained as possible to avoid catching an error that isn't meant to be caught and mishandling it. 

Also, as mentioned in a previous comment, you can use `lodash.get` to get deeply nested attributes.

brianjgeiger:
For root vs. parent, I don't think there's one particular way that's better. Well, more specifically, I don't think there's one way that will catch all cases. 

In the case of RPP, parent might be better. But if you do a lot of projects with a similar 3+ deep nesting structure, then root will be better. But yes, happy to let @saradbowman decide. Or to have someone do some research into what's more commonly used right now. 

Also, what information is shown will definitely affect how people use it, so we do have the option of pushing people towards one way or another.

sloria:
Pass finished. 

@zamattiac Wait for Sara's decision before addressing https://github.com/CenterForOpenScience/osf.io/pull/5742#issuecomment-223094780 . In the meantime, go ahead and make the other minor changes.

zamattiac:
As far as I am aware, it was decided to use the root.

brianjgeiger:
That's a very passive way of describing it. Did @saradbowman or @nicipfeiffer say to use the root, or were there rumors that you overheard, or something else?

nicipfeiffer:
Let's go with root, please

brianjgeiger:
JSHint Error
screen shot 2016-06-09 at 10 03 28 am

brianjgeiger:
What's happening with commit https://github.com/CenterForOpenScience/osf.io/pull/5742/commits/fb716e99a162859efed403e15a882ec8e8bf6398 ? We decided to go with root rather than parent per https://github.com/CenterForOpenScience/osf.io/pull/5742#issuecomment-223656785 and I didn't see any of Steve's other comments that might have been related to that. Most of the things I was expecting to change was just the lodash items, not using a variable, and an overly aggressive try-except block.

zamattiac:
Sara and Nici requested it on the JIRA ticket 3 days ago and approved it
when I showed it to them on JIRA yesterday.

On Thu, Jun 9, 2016 at 4:52 PM, Brian J. Geiger <notifications@github.com>
wrote:

> What's happening with commit fb716e9
> <https://github.com/CenterForOpenScience/osf.io/pull/5742/commits/fb716e99a162859efed403e15a882ec8e8bf6398>
> ? We decided to go with root rather than parent per #5742 (comment)
> <https://github.com/CenterForOpenScience/osf.io/pull/5742#issuecomment-223656785>
> and I didn't see any of Steve's other comments that might have been related
> to that. Most of the things I was expecting to change was just the lodash
> items, not using a variable, and an overly aggressive try-except block.
>
> —
> You are receiving this because you modified the open/close state.
> Reply to this email directly, view it on GitHub
> <https://github.com/CenterForOpenScience/osf.io/pull/5742#issuecomment-225022713>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/APHeVYouYs6pLphLeXhi8xMabUTvNMs5ks5qKHz_gaJpZM4ImnXt>
> .
>



-- 
Matthew Keitelman
571-269-4724
mak2vr@virginia.edu

brianjgeiger:
@zamattiac Is the screenshot up to date with the new functionality?

zamattiac:
It isn't sorry. Here it currently is:
image

and private projects: 
image

website/static/js/home-page/quickProjectSearchPlugin.js
```diff
@@ -539,6 +539,29 @@ var QuickSearchProject = {
     }
 };
 
+function ancestorName(node, ancestor, title) {
+    var name;
+    var request = lodashGet(node, 'embeds.' + ancestor + '.data.attributes.title', '');
```
brianjgeiger: This is just a little harder to follow than I'd like, and it's probably because of variable naming, especially `request` and `title`, although `ancestor` might also be more descriptive, such as `ancestorDescriptor` or something (`ancestor` sounds as though it might be an object, for example).

brianjgeiger:
Logic seems good, just a request to clarify how the one function works probably mostly by changing variable names. 🐧 Once that's in, it should be RTM.

brianjgeiger:
Looks great. 🎉 🐧 🎉

website/static/js/home-page/quickProjectSearchPlugin.js
```diff
@@ -46,7 +47,7 @@ var QuickSearchProject = {
         self.templateNodes.start();
 
         // Load up to first ten nodes
-        var url = $osf.apiV2Url('users/me/nodes/', { query : { 'embed': 'contributors'}});
+        var url = $osf.apiV2Url('users/me/nodes/', { query : { 'embed': ['contributors','root', 'parent']}});
```
sloria: Embedding can be very expensive. @zamattiac Can you do a bit of performance testing to check that our response times aren't slowed down too much by this change?

brianjgeiger: I'm not opposed to checking, but it's, what, 10 records at most? If we can't do that easily, we have bigger troubles.

sloria: We're still loading all of the user's projects, 10 at a time. Sure, the user will see content after the first 10 load, but the user won't be able to filter their all their projects until they're all loaded, so we still want that to be relatively fast.

website/static/js/home-page/quickProjectSearchPlugin.js
```diff
@@ -531,6 +539,29 @@ var QuickSearchProject = {
     }
 };
 
+function getAncestorDescriptor(node, nodeTitle, ancestor) {
+    var ancestorDescriptor;
+    var ancestorTitleRequest = lodashGet(node, 'embeds.' + ancestor + '.data.attributes.title', '');
+    var errorRequest = lodashGet(node, 'embeds.' + ancestor + '.errors[0].detail', '');
+        switch(errorRequest) {
```
sloria: Minor: Overindentation.

website/static/js/home-page/quickProjectSearchPlugin.js
brianjgeiger: Ah, right, because of the searching. Well, worth profiling, I guess, although it shouldn't affect the UX.

website/static/js/home-page/quickProjectSearchPlugin.js
```diff
@@ -531,6 +539,29 @@ var QuickSearchProject = {
...
+            case '':
+                if (ancestorTitleRequest === nodeTitle | ancestorTitleRequest === '') {
```
sloria: Minor: Use `||` for logical OR.

sloria:
Pass finished. Just a little performance testing and a few nitpicks. Once those are done, this is good to merge.

zamattiac:
The performance testing is in: at most, the increase is .05 seconds.

sloria:
Great! Thanks for doing that. This is good to merge then. Will merge this after today's release.

PR #5745, opened 2016-05-25
Redirect users from unmaintained osf.io/news, to maintained cos.io/news [OSF-6417]

Johnetordoff:
## Purpose

Make sure user aren't exposed to out of date news at osf.io/news

## Changes

New website view function redirects to cos.io/news which is maintained. Deletes unused news.mako.

## Side effects

None that I know of.

## Ticket

No Ticket, but BrianN said I should do this on flowdock, and seems like a simple fix.


website/views.py
```diff
@@ -261,3 +261,7 @@ def redirect_getting_started(**kwargs):
 def redirect_to_home():
     # Redirect to support page
     return redirect('/')
+
+def redirect_to_cosNews(**kwargs):
```
sloria: use snake_case, e.g. `cos_news`.

website/views.py
```diff
@@ -261,3 +261,7 @@ def redirect_getting_started(**kwargs):
...
+    # Redirect to COS News page
+    return redirect('http://cos.io/news/')
```
sloria: Use `https` links.

sloria:
Pass finished.

PR #5748, opened 2016-05-25
[Feature/versions] Surface Versions on files [OSF-6026]

SSJohns:
## Purpose

The changes introduced a feature that would let users see what version of the file they were looking at on the files page. In addition when they clicked that version they would be taken to the files-view page on the Revisions tab, and be able to see what version of the file they would be looking at. Clicking the "Version #" at the top of the page would take them to the Revisions tab and further clicking of it would keep them on that page.

## Changes

I made changes to a lot of files, on the front end I made a versions column on the files tab and in the files view page the version number is displayed alongside the name of the file. 

I was not able to check for dropbox or any other addons, however the workings should be similar. A problem with my machine prevented local testing of those functions. 

All of the features are purely front end changes so it would be hard to implement unit test for them. 

### Before
Before Uno

Before Dos

### After
After One

After Two

### Mobile View
screen shot 2016-07-18 at 3 27 51 pm

screen shot 2016-07-18 at 3 28 06 pm

There was a strange thing happening in the revisions tab, but I doubt that is the fault of any changes I made. It looks like it more has to do with the providers mfr will accept.

screen shot 2016-07-18 at 3 28 24 pm

## Side effects

If the fangorn files are used as a template in any other places the the default columns could be initialized as Name and Version. This behavior might be  undesirable. 

Editied a few inhouse functions that are used in other places. Mainly -fangornTitleColumn to be host to a helper function. This should still work in the way the previous one did in all other places in the code that it is used.

## Ticket

https://openscience.atlassian.net/browse/OSF-6026


website/static/js/fangorn.js
```diff
@@ -1176,36 +1176,45 @@ function gotoFileEvent (item) {
...
-        return m('a.fg-file-links',{ href: '/' + item.data.nodeID.toString() + '/'},
-                item.data.name);
-    }
-    return m('span', item.data.name);
+  var tb = this;
+  return _fangornTitleColumnHelper(tb,item,col,item.data.name,'/','fg-file-links');
+}
+
+function _fangornVersionColumn(item,col) {
+  var tb = this;
```
brianjgeiger: Spacing throughout should be 4 spaces per level of indent.

website/static/js/fangorn.js
```diff
@@ -1157,11 +1157,11 @@ function _fangornUploadMethod(item) {
     return configOption || 'PUT';
 }
 
-function gotoFileEvent (item) {
+function gotoFileEvent (item, toUrl) {
```
brianjgeiger: Don't feel the need to change this, but if you'd wanted to you could have checked to see if toUrl was defined and, if not, set it to '/'. That would have kept the interface the same for existing usages while still allowing you to add the new functionality.

SSJohns: I checked around and I think goToFileEvent is only used in this file, that's why I did it that way.

website/static/js/filepage/index.js
```diff
@@ -363,6 +363,64 @@ var FileViewPage = {
         }, 1000);
 
         self.mfrIframeParent = $('#mfrIframeParent');
+        function goToRevisions(e){
+          var editable = self.editor && self.editor.selected;
+          var viewable = self.mfrIframeParent.is(':visible');
+          if (editable || viewable){
```
brianjgeiger: Since you're checking for both viewable and editable individually inside this block, and you aren't doing anything else inside this block, I suspect this if statement can be safely removed.

website/static/js/filepage/index.js
```diff
@@ -363,6 +363,64 @@ var FileViewPage = {
...
+          }
+          self.revisions.selected = true;
+        }
+
+        function changeVersionHeader(value){
+          m.render(document.getElementById('version-link'), m('a', {onclick: goToRevisions},'(Version:  ' + String(value) + ')'));
+        }
+
+        function httpGetAsync(theUrl, callback) {
+            var xmlHttp = new XMLHttpRequest();
```
brianjgeiger: Why XML HTTP? I think all of the endpoints we talk to regularly are json, yes?

website/static/js/filepage/index.js
```diff
@@ -363,6 +363,64 @@ var FileViewPage = {
...
+            xmlHttp.send(null);
+        }
+
+        function setVersionHeader(responseText) {
+            var obj = JSON.parse(responseText);
+            changeVersionHeader(obj.data[0].version);
+        }
+
+        //anchor checking hack that will select if true
+        var idx = window.location.href.indexOf('?');
```
brianjgeiger: Check out URI.js - it should be better than string manipulation and we already use it in our stack. https://medialize.github.io/URI.js/

website/templates/project/view_file.mako
```diff
@@ -15,6 +15,7 @@
     <h2 class="break-word">
       ## Split file name into two parts: with and without extension
       ${file_name_title | h}<span id="file-ext">${file_name_ext | h}</span>
+      <a id="version-link"><span id="version-number"></span></a>
```
brianjgeiger: This seems a little odd and unlike the way we're doing things elsewhere in the file (such as right above and right below), but I'm not positive because I'm not big into mithril and am not sure if this is proper.

brianjgeiger:
Okay, a couple of places where there are better tools that you could be using, and one place where I'm not sure if you're handling the templating portion right or not. Pass done. 🐧 Once you've finished with these changes and get it working again, please schedule a product demo with Nici and/or Sara before putting it back into code review so they can approve your implementation.

website/static/js/filepage/index.js
SSJohns: I think the if statement is made to uncheck the tabs that are not being used. So if I am moving things to the revisions tab I need this if to make sure that the views and edits tab are not going to stay highlighted.

brianjgeiger: Right, but inside this block you have an `if (viewable)` and and `if (editable)` and that's it. So you should be able to safely remove the outer if statement and just keep the inner block. I.e.
```
var editable = self.editor && self.editor.selected;
var viewable = self.mfrIframeParent.is(':visible');
if (viewable){
    self.mfrIframeParent.toggle();
}
if (editable) {
    self.editor.selected = false;
}
self.revisions.selected = true;
```
*should* work exactly the same as the code you have, yes?

website/static/js/filepage/index.js
SSJohns: Implemented.

website/static/js/filepage/index.js
SSJohns: I don't understand, that is the way to get a json response. Are you saying I should use JSONRequest instead? This code just issues a request and returns data that is parsed as a json. The method to get that json shouldn't matter. Unless you specifically want me to use Ajax or jQuery this method works without relying on importing any more libraries or using too much overhead.

website/static/js/filepage/index.js
SSJohns: Umm, yeah I can do it that way. I just had it this way because elsewhere in the file I saw that they did this and I assumed it was to avoid extraneous if checks so I just followed that style. I can change it though.

website/static/js/filepage/index.js
SSJohns: Nvm, eliminated this function completely.

website/static/js/filepage/index.js
brianjgeiger: Please do change it. It's entirely trivial from a performance view either way, but in this case the code is entirely redundant.

website/addons/base/views.py
```diff
@@ -571,6 +571,7 @@ def addon_deleted_file(auth, node, error_type='BLAME_PROVIDER', **kwargs):
         'file_path': file_path,
         'file_name_title': file_name_title,
         'file_name_ext': file_name_ext,
+        'version_id': version.identifier,
```
mfraezz: `version` isn't defined here. It might be `kwargs`, but even if so this will cause issues.

SSJohns: Yeah, I'm trying to fix that now so it will pass the Travis test. Someone suggested I rewrite the test because I added it, does that sound like it's along the right tracks?

website/addons/base/views.py
```diff
@@ -571,7 +571,6 @@ def addon_deleted_file(auth, node, error_type='BLAME_PROVIDER', **kwargs):
         'file_path': file_path,
         'file_name_title': file_name_title,
         'file_name_ext': file_name_ext,
-        'version_id': version.identifier,
```
brianjgeiger: As @mfraezz indicated, this was failing because deleted files don't have a version, so when it tried to access version.identifier, there was an error. This is _proooobably_ okay, but another way to have fixed it would have been to do `'version_id': None,` instead, which would keep the option serialized but ensure that it was null when it made it to the outside world.

All that being said, I don't think anything's using the version field on a deleted file, although I also don't know who would be using `addon_deleted_file()` at all. Have you gone through and found out what uses `addon_deleted_file()` and done manual testing to ensure that it won't be problematic?

SSJohns: I've at least gone through and made sure no file that uses `addon_deleted_file() `need the version. In the way it is now the code for this part is exactly how it was before I started work on this project, any errors that would result would not be related to my ticket. Saying that I've checked and nothing I changed relies on `addon_deleted_file()` and there are no errors that I have seen.

website/static/js/fangorn.js
```diff
@@ -1280,6 +1291,13 @@ function _fangornResolveRows(item) {
     });
 
     if (item.data.kind === 'file') {
+      default_columns.push(
+      {
+        data: 'version',
```
brianjgeiger: What are you using for an editor or IDE? It should be able to use the project's .jshintrc to know how many spaces there are per tab in a javascript file so you don't have to deal with it manually. See if you can get that fixed and then go ahead and clean up the spaces from this commit.

brianjgeiger:
Pass done. 🐧 

- [ ] Verify you're not breaking anything by not having a version serialized for 
- [ ] Fix your setup to use the .jshintrc
- [ ] Fix spaces

brianjgeiger:
Fantastic. Could you add Before and After screenshots to the Changes section so QA knows what to look for? Then this should be good to go. 🎉 🐧 🎉

website/static/js/filepage/index.js
```diff
@@ -363,6 +364,37 @@ var FileViewPage = {
...
+                self.mfrIframeParent.toggle();
+            }
+            if (editable) {
+                self.editor.selected = false;
+            }
+            self.revisions.selected = true;
+        }
+
+        function changeVersionHeader(){
+            m.render(document.getElementById('version-link'), m('a', {onclick: goToRevisions}, document.getElementById('version-link').innerHTML));
```
sloria: As per COSDev, use `lowerCamelCase` for CSS IDs.

website/static/js/filepage/index.js
```diff
@@ -363,6 +364,37 @@ var FileViewPage = {
...
+        }
+
+        //anchor checking hack that will select if true
+        var idx = window.location.href.indexOf('?');
+        var idx_uri = (idx > 0) ? URI.parseQuery(window.location.href.slice(idx)) : '';
```
sloria: As per COSDev, use `lowerCamelCase` for variable names.

sloria: We have a helper for parsing query params, `osfHelpers.urlParams`. Use this instead.

SSJohns: Above Brian said to use URI.js because we also already use this on our stack. Is the osfhelpers on top of uri?

sloria: We use URI.js for building URLs, but not for reading query params.

brianjgeiger: As a general rule, especially for using our helper functions, @sloria will give better advice than I do.

website/static/js/filepage/index.js
```diff
@@ -363,6 +364,37 @@ var FileViewPage = {
         }, 1000);
 
         self.mfrIframeParent = $('#mfrIframeParent');
+        function goToRevisions(e){
+            var editable = self.editor && self.editor.selected;
```
sloria: Rather than doing all this work to manipulate the DOM and modify the controller state, could you just link to `?show=revision`?

We modify the user's query params in other parts of the site. `website/statis/js/search.js` should have examples of this.

This will allow the user to use the "Back" button to go back to the default file view. It will also allow the user to copy and paste the URL for revisions.

sloria:
Pass finished.

website/static/js/filepage/index.js
SSJohns: Coo

website/static/js/filepage/index.js
SSJohns: Are you saying redirect to `?show=revision` when ever this is called?

sloria: Essentially. Use history.js to modify the query parameters without doing a page refresh. If our components are wired correctly, simply changing that query parameter should enable the revisions view.

brianjgeiger:
Looks like there are merge conflicts and a request, "From the Product Consults flow: the link should toggle the revisions view, i.e. clicking alternately shows and hides the revisions view. Shaquille Johnson Can you please update your PR with this behavior?"

sloria:
@SSJohns This is not working after the latest updates. When I click the "Version: " link, a JS error occurs.

osf___dads_jpg

and the Revisions tab does not get displayed.

sloria:
I'll continue my CR when this is in a working state.

sloria:
Pinging @pattisdr to make her aware of this feature, since she is working on porting the file detail page.

caneruguz:
@SSJohns this adds a column to the files view. Did you check if the mobile view is still working well with the columns? Maybe put a screenshot. Also the project summary files widget screen shot would also be useful.

SSJohns:
Hello, and yeah other changes in develop have added columns as well. It may be getting slightly cramped in the view though, but I don't think a column based webpage is really meant to be viewed in mobile.

caneruguz:
If I'm not mistaken around 6-8% of our users do use their mobile phones. Whether we think a column based web page is meant to be used in mobile or not we do need to provide a working version to the users who are accessing their files from their phones. Cramped view may itself not be a problem, but we do need to address how the view is. Can you please add the screenshots? 

@sloria

SSJohns:
Oh sorry, I didn't mean to make it sound like I wasn't adding the screenshot, I added the screen shot right before I posted my comment. I just meant to point out that other people have been adding to the file view and I was not sure how I would address this because it is very out of line with my ticket and PR.

caneruguz:
@SSJohns That's fair, and why it's all the more important to highlight this point. The sizing looks fine for the moment, but we will keep an eye on what else is being added. We will probably look into how to make the column names clear when truncated, but that's not necessarily part of this PR, probably another ticket would need to be made. I'll look into that. Thanks!

PR #5749, opened 2016-05-26
[RFR] Feature/fix locking by moving tasks to celery [OSF-6294]

cwisecarver:
## Purpose

This should fix tasks locking user objects in mongo. 

## Changes

It adds the ability for the postcommit handler decorator to make tasks run in celery instead of running postcommit, inside or after the request. It also decorates several tasks that were causing locks.

## Side effects

If celery is down or backed up some of the tasks may be delayed to the point that users may notice the discrepancy in their activity counters.

## Ticket
This is a second pass at fixing OSF-6294 without so many side effects.

https://openscience.atlassian.net/browse/OSF-6294

### DIFF
--- a/website/notifications/utils.py
+++ b/website/notifications/utils.py
@@ -1,5 +1,6 @@
 import collections
 
+from framework.postcommit_tasks.handlers import run_postcommit
 from modularodm import Q
 from modularodm.exceptions import NoResultsFound
 
@@ -8,6 +9,10 @@ from website.notifications import constants
 from website.notifications import model
 from website.notifications.model import NotificationSubscription
 from website.project import signals
+
+from framework.celery_tasks import app
+from framework.celery_tasks.handlers import queued_task
+from framework.transactions.context import transaction
 
 
 class NotificationsDict(dict):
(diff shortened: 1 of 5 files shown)


brianjgeiger:
If you're queueing the request, does it need to be post-commit now?

brianjgeiger:
Discussed in person. I'm not doing a real code review, but the conceptual issues seem good to me.

website/notifications/utils.py
```diff
@@ -69,6 +72,12 @@ def remove_contributor_from_subscriptions(node, user):
 
 @signals.node_deleted.connect
 def remove_subscription(node):
+    remove_subscription_task(node._id)
+
+@run_postcommit(once_per_request=False, celery=True)
+@app.task(max_retries=5, default_retry_delay=60)
```
sloria: Why was this changed to be celery task?

cwisecarver: Because it was causing transactions to lock and fail on bulk api requests. The main point of this PR is switching three functions to run as celery tasks to work around the locking issue. I would switch the entirety of the postcommit handler to do that but the varnish purging tasks need to run after the transaction is committed but prior to the response being sent to the user so they have to run directly after the database commit.

PR #5753, opened 2016-05-26
Fix broken "remove contributor" modal caused by knockout async problems [OSF-6109]

cameronblandford:
## Purpose

To remove a bug where asynchronous knockout.js data-bind processing is too slow and results in a broken modal being displayed.

## Changes

This is a problem that has been encountered before, and I used the same solution that's employed across OSF: using the helper method osfHelpers.applyBindings() instead of the default ko.applyBindings(). As per the usual implementation, I updated the modal content divs to also include the "scripted" class with which the osfHelpers applyBindings method interacts. These changes, both in theory and in all observations thus far, have fixed the bug, and have updated the code to be consistent with the rest of OSF.

## Side effects

Though not specified in a ticket, I found the same issue in the code for the Add Contributors modal, and updated it to use the osfHelpers implementation of applyBindings, instead of the default knockout implementation. This should prevent similar bugs with the Add Contributors modal.

### Normal Behavior

image

### Behavior caused by OSF-6109

image

## Notes

Because of the specific nature of this bug, the probability of any action reproducing it is extremely low (since it depends on a particular and rare sequences of asynchronous events). I'd only been able to replicate it three times, after ~800 tries. After locally implementing this fix, I gave it another ~1000 tries, and could not reproduce it. This, in addition to how the fix was implemented (in that it's worked in the past and widely used), leads me to believe that this fix works. However, I technically have no way of confirming this 100%. But I truly believe I have tested for it to the best of my ability.

## Ticket

[[OSF-6109]](https://openscience.atlassian.net/browse/OSF-6109)

### DIFF
--- a/website/static/js/contribAdder.js
+++ b/website/static/js/contribAdder.js
@@ -508,7 +508,7 @@ ContribAdder.prototype.init = function() {
     self.viewModel.fetchNodeTree(treebeardUrl).done(function(response) {
         new NodeSelectTreebeard('addContributorsTreebeard', response, self.viewModel.nodesState);
     });
-    ko.applyBindings(self.viewModel, self.$element[0]);
+    $osf.applyBindings(self.viewModel, self.$element[0]);
     // Clear popovers on dismiss start
     self.$element.on('hide.bs.modal', function() {
         self.$element.find('.popover').popover('hide');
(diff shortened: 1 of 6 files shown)


brianjgeiger:
Could you merge with develop then `invoke karma`? There's a post-merge js test that's failing that may, somehow, be related to what you're doing.

brianjgeiger:
Other than the failing test, looks good. 🐧 Pass done.

mfraezz:
LGTM. Dropping this in RTM ⚓

PR #5754, opened 2016-05-26
Abillity to add institutions on project creation

hmoco:
## Purpose

Allows user to add institutions to a project on the creation modal.

## Changes

instcreate

## Ticket

https://openscience.atlassian.net/browse/OSF-6315


website/static/js/addProjectPlugin.js
```diff
@@ -35,6 +34,9 @@ var AddProject = {
         self.newProjectDesc = m.prop('');
         self.newProjectCategory = m.prop(self.defaultCat);
         self.newProjectTemplate = m.prop('');
+        self.newProjectInstitutions = window.contextVars.currentUser.institutions.map(
```
sloria: Can you avoid using `window.contextVars` easily? Pass institutions as a `param`?

website/static/js/addProjectPlugin.js
```diff
@@ -144,6 +153,54 @@ var AddProject = {
...
+                                $('.img-circle.fa.fa-check').removeClass('hidden');
+                                ctrl.newProjectInstitutions = window.contextVars.currentUser.institutions.map(
+                                    function(inst){return inst.id;}
+                                );
+                            }, style: {float: 'right', marginRight: '5px'}}, 'Select all'),
+                            m('table', m('tr', window.contextVars.currentUser.institutions.map(
+                                function(inst){
+                                    return m('td',
+                                        m('a', {onclick: function(){
+                                            $('#instLogo' + inst.id).toggleClass('hidden');
```
sloria: This dependence on CSS IDs and manual DOM updates should be unnecessary. Consider refactoring (store state on ctrl and update view accordingly...).

website/static/js/addProjectPlugin.js
```diff
@@ -144,6 +153,54 @@ var AddProject = {
...
+                                            [
+                                            m('img.img-circle.text-muted',
+                                                {
+                                                    src: inst.logo_path, width: '45px', height: '45px',
+                                                }
+                                            ),
+                                            m('i.img-circle.fa.fa-check#instLogo' + inst.id,
+                                                {
+                                                    style: {
+                                                        color: 'lightgreen',
```
sloria: Use the osf-style color palette whenever possible. https://centerforopenscience.github.io/osf-style/style.html#colors

How about #C7FFC7?

website/static/js/addProjectPlugin.js
hmoco: I went with contextVars because addProject is instantiated in multiple places, all of which would require pointing to the same piece of contextVars.

sloria:
Pass finished. No major blockers. But if the above refactors can be done quickly, better to do them now to avoid duplicated QA effort.

website/static/js/addProjectPlugin.js
sloria: OK, that's fine for now.

website/static/js/addProjectPlugin.js
```diff
@@ -89,6 +96,15 @@ var AddProject = {
...
                 self.viewState('success');
                 self.goToProjectLink(result.data.links.html);
+                if (self.newProjectInstitutions.length > 0){
+                    var newNodeApiUrl = $osf.apiV2Url('nodes/' + result.data.id + '/relationships/institutions/');
+                    var data = {data: self.newProjectInstitutions.filter(
+                        function(id){return self.checkedInstitutions[id];}
+                    ).map(
+                        function(id){return {type: 'institutions', id: id};}
+                    )};
+                    m.request({method: 'POST', url: newNodeApiUrl, data: data, config: xhrconfig});
```
sloria: This needs error handling

sloria: Also, what happens if the user navigates away from the page before this request finishes?

It might be better to only show the success alert once both requests have finished. You can do this by chaining promises.

```js
// send create project request,
var request = m.request(....)
if user has selected institutions:
    request.then(sendAffiliationRequest)
request.then(handleSuccess, handleError)
````

File: website/static/js/addProjectPlugin.js [hunk 5]
```diff
@@ -35,6 +35,13 @@ var AddProject = {
         self.newProjectDesc = m.prop('');
         self.newProjectCategory = m.prop(self.defaultCat);
         self.newProjectTemplate = m.prop('');
+        self.checkedInstitutions = {};
+        self.newProjectInstitutions = window.contextVars.currentUser.institutions.map(
```
sloria: Minor:  I still think this should be refactored to be passed as an option, but in the interest of getting this onto staging soon, this is ok. That said, you could easily future-proof by doing

```js
self.institutions = options.institutions || window.contextVars.currentUser.institutions || [];
```

This will make the future refactor easier.

website/static/js/addProjectPlugin.js
```diff
@@ -7,6 +7,7 @@ require('css/add-project-plugin.css');
 var $ = require('jquery');
 var m = require('mithril');
 var $osf = require('js/osfHelpers');
+var InstitutionImg = require('js/components/institution');
```
sloria: Minor: This is a bit misleading, because you're not importing the `InstititionImg` component; you're importing the `institution` module. You could do

```js
var institutionComponents = require('js/components/institution');
var CheckableInst = institutionComponents.CheckableInstitution;
```

website/static/js/components/institution.js
```diff
@@ -41,6 +41,31 @@ var InstitutionImg = {
     }
 };
 
+var CheckableInst = {
```
sloria: Minor: For naming consistency, don't abbreviate `Institution` -> `CheckableInstitution`.

website/static/js/components/institution.js
```diff
@@ -18,7 +18,7 @@ var InstitutionImg = {
             muted: false
         };
         if (!opts.width && !opts.height) {
-            throw new Error('InstitionImg requires width and/or height option.');
```
sloria: 👍  Thanks

sloria:
Pass finished. Mostly nits, but https://github.com/CenterForOpenScience/osf.io/pull/5754#discussion_r64953232 should be addressed

website/static/js/addProjectPlugin.js
```diff
@@ -89,6 +96,15 @@ var AddProject = {
             var success = function _success (result) {
                 self.viewState('success');
                 self.goToProjectLink(result.data.links.html);
+                if (self.newProjectInstitutions.length > 0){
```
sloria: I'm a bit confused why `newProjectInstitutions` is necessary; couldn't you just use `self.checkInstitutions`?

sloria: Even when you create a project with no institutions selected, a request gets sent to the /relationships/institutions endpoint. I think you can avoid this with something like

```js
if (Object.keys(self.checkedInstitutions).length) 
```

PR #5755, opened 2016-05-26
Add user messaging on Analytics tab regarding use with Adblocking [OSF-6307]

SSJohns:
## Purpose

Further PR that gets rid of webpages ability to sense ads and just displays the block to all users.

## Changes

Removed ad sensing code
## Side effects

None

## Ticket


sloria:
As per https://openscience.atlassian.net/browse/OSF-6307, the alert should not be shown on private projects. Also, could you look into making the alert dismissible? @SSJohns

mfraezz:
Build is failing due to jshint (javascript syntax) errors: https://travis-ci.org/CenterForOpenScience/osf.io/jobs/134488389

You can run this locally with `invoke jshint` to see if anything is wrong before you push. `invoke flake8` is the python equivalent.

mfraezz:
LGTM. Verified intended behavior for projects both private and public, and for registrations. Dropping this RTM when the build passes. 🐧

website/static/js/pages/base-page.js
```diff
@@ -123,4 +123,15 @@ $(function() {
             new KeenTracker(window.contextVars.keenProjectId, window.contextVars.keenWriteKey, params);
         }
     }
+
+    // Ad script alert
+    var adBlockPersistKey = 'adBlock';
+    var $adBlock = $('#adBlock').on('closed.bs.alert', function() {
```
sloria: This code doesn't need to run on every page; base-page.js is included on every page. This should only be executed on the "Analytics" page. To do this, you can add a new `-page` JS module for the analytics page (statistics.mako). See http://cosdev.readthedocs.io/en/latest/osf/js_modules.html#building-and-using-modules

sloria:
Pass finished. Just one comment to address: https://github.com/CenterForOpenScience/osf.io/pull/5755#discussion_r65943529

PR #5759, opened 2016-05-27
Feature/share files

Rytiggy:
## Purpose

Restricted functionality to share window and styling 

## Changes

adjusted css and added If statements that change the page accordingly


Rytiggy:
Restricted functionality to share window and styling

PR #5761, opened 2016-05-27
Fix registrations files provider relationship [#OSF-6413]

erinspace:
## Purpose
In the API, registration files were being redirected to ```/node/```  instead of ```/registrations/``` and were so causing an error. This adds a registration file provider serializer to send the correct information

## Changes
- Add RegistrationProviderSerializer
- Use that serializer

## Side effects
- nope

## Ticket
https://openscience.atlassian.net/browse/OSF-6413

### DIFF
--- a/api/registrations/serializers.py
+++ b/api/registrations/serializers.py
@@ -4,7 +4,7 @@ from rest_framework import exceptions
 
 from api.base.utils import absolute_reverse
 from api.files.serializers import FileSerializer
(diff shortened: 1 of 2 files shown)


brianjgeiger:
Looks great! 🎉 🐧 🎉

brianjgeiger:
Could you add a copy of the old json vs. the new in the Changes section?

erinspace:
Closing in favor of the hotfix #5763

PR #5766, opened 2016-05-27
Hotfix/email help desk stuck registrations [#OSF-5848]

erinspace:
## Purpose
There is an existing script that is not being run currently to find stuck registrations, delete them, and send an email to both Desk.com and the user telling them there is a problem. 

This PR changes this script to only email the help desk so we can deal with stuck registrations on a case by case basis, and NOT email the user.

## Changes
- Add ```email_users``` flag to ```handle_archive_fail``` function
- Set that flag to ```False``` to only email the Help desk with stuck registrations

## Side effects
- None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-5848

### DIFF
--- a/scripts/cleanup_failed_registrations.py
+++ b/scripts/cleanup_failed_registrations.py
@@ -6,8 +6,6 @@ import logging
 from modularodm import Q
 
 from website.archiver import (
-    ARCHIVER_UNCAUGHT_ERROR,
-    ARCHIVER_FAILURE,
     ARCHIVER_INITIATED
 )
 
@@ -17,7 +15,6 @@ from website import (
 )
 
 from website.settings import ARCHIVE_TIMEOUT_TIMEDELTA
-from website.archiver.utils import handle_archive_fail
 from website.archiver.model import ArchiveJob
 
 from website.app import init_app
@@ -54,10 +51,10 @@ def notify_desk_about_failed_registrations(dry_run=True):
             # Send an email to the OSF Help Desk
             mails.send_mail(
                 to_addr=settings.SUPPORT_EMAIL,
-                mail=mails.ARCHIVE_UNCAUGHT_ERROR_DESK,
+                mail=mails.ARCHIVE_REGISTRATION_STUCK_DESK,
                 user=f.creator,
                 src=f.registered_from,
-                results=f.archive_job.target_info(),
+                archive_job=f.archve_job,
             )
 
             count += 1
--- a/scripts/cleanup_failed_registrations.py
+++ b/scripts/cleanup_failed_registrations.py
@@ -10,6 +10,12 @@ from website.archiver import (
     ARCHIVER_FAILURE,
     ARCHIVER_INITIATED
 )
+
+from website import (
+    mails,
+    settings
+)
+
 from website.settings import ARCHIVE_TIMEOUT_TIMEDELTA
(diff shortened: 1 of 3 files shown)


scripts/cleanup_failed_registrations.py
```diff
@@ -50,7 +50,8 @@ def remove_failed_registrations(dry_run=True):
                 f.registered_from,
                 f,
                 f.creator,
-                f.archive_job.target_info()
+                f.archive_job.target_info(),
+                email_users=False
```
sloria: As per the ticket, we only want to be notified on Desk about stuck registrations; there should be no db writes. So rather than adding a new parameter to `handle_archive_fail` (which appears to do db writes), why not just remove the call to `handle_archive_fail` and send the desk emails?

sloria:
Pass finished.

scripts/cleanup_failed_registrations.py
```diff
@@ -40,26 +48,27 @@ def remove_failed_registrations(dry_run=True):
...
-                ARCHIVER_UNCAUGHT_ERROR,
-                f.registered_from,
-                f,
-                f.creator,
-                f.archive_job.target_info()
+
+            # Send an email to the OSF Help Desk
+            mails.send_mail(
+                to_addr=settings.SUPPORT_EMAIL,
+                mail=mails.ARCHIVE_UNCAUGHT_ERROR_DESK,
```
sloria: We should send a different type of email for these "stuck" registrations. Otherwise, it'll be hard to distinguish these from the ones sent here: https://github.com/CenterForOpenScience/osf.io/blob/4a417cc1490052fd019918b5f6e5a809e36104e1/website/archiver/utils.py#L72-L75 . 

How about a new email that says:

```
User: <FULL NAME> (<USERNAME>) [<USER_ID>]

Registration <TITLE> [<NODE_ID>] is stuck in archiving.

Archive Job: [<ARCHIVE_JOB_ID>]

<JSON REPRESENTATION OF ARCHIVE_JOB (can use archive_job.to_storage()) >

Automatically sent from scripts/cleanup_failed_registrations.py
```

scripts/cleanup_failed_registrations.py
```diff
@@ -10,6 +10,12 @@
     ARCHIVER_FAILURE,
```
sloria: Minor: remove any unused imports.

sloria:
Pass finished.

scripts/cleanup_failed_registrations.py
```diff
@@ -40,26 +45,27 @@ def remove_failed_registrations(dry_run=True):
...
-                f.creator,
-                f.archive_job.target_info()
+
+            # Send an email to the OSF Help Desk
+            mails.send_mail(
+                to_addr=settings.SUPPORT_EMAIL,
+                mail=mails.ARCHIVE_REGISTRATION_STUCK_DESK,
+                user=f.creator,
+                src=f.registered_from,
+                archive_job=f.archve_job,
```
sloria: Typo here: `f.archive_job`.

sloria:
@erinspace Did you verify that this works locally?

sloria:
Tested locally; seems to work as expected.

sloria:
Merged into develop. 👍

PR #5770, opened 2016-05-31
Hotfix/ Fix API referent error on get_paginated_response [#OSF-6427

erinspace:
## Purpose
When filtering by  node id, and using related counts, a non found node would throw a 500 error. This PR changes the filter function to check if a node being filtered on exists before trying to paginate.

## Changes
- Add check for root_target to filter for pagination
- Add test to make sure that no results are found when filtering on a fake node id with related counts

## Side effects
- none anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-6427


sloria:

PR #5772, opened 2016-05-31
[RFR][Feature][OSF-6300]File view pane, edit panel shows earlier version

TomBaxter:
## Purpose:
[OSF-6300](https://openscience.atlassian.net/browse/OSF-6300)
Fix to get latest version in file editor panel
Add code to refresh active sharejs sessions. For future use in v2 API wiki updating

## Changes:
Update website/addons/base/views.py add version to key for file_guid_to_share_uuids
Update website/addons/wiki/model.py add def update_active_sharejs for use in future API updates
Update website/addons/wiki/shareServer.js add reload call
Update website/addons/wiki/static/ShareJSDoc.js add reload case

# Warning!
1. This code updates sharejs code and will require sharejs to be bounced.
2. If sharejs restart is deferred, any new shared editing sessions initiated from the file view after the code is updated will not share(js) with sessions initiated before the code is updated. This does not effect the wiki sharejs sessions

[#OSF-6300]

### DIFF
--- a/website/addons/base/views.py
+++ b/website/addons/base/views.py
@@ -662,12 +662,12 @@ def addon_view_file(auth, node, file_node, version):
 
     ret = serialize_node(node, auth, primary=True)
 
-    if file_node._id not in node.file_guid_to_share_uuids:
-        node.file_guid_to_share_uuids[file_node._id] = uuid.uuid4()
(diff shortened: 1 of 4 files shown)


brianjgeiger:
Looks good. 🎉 🐧 🎉

PR #5773, opened 2016-05-31
[RFR] Give warning message to users with childless nodes

billyhunt:
Related to https://github.com/CenterForOpenScience/osf.io/pull/5726

## Purpose

Users working on a project without components were not getting a warning if they were removing a non-affiliated institution.  Also

1.  A bug where an institution could sometimes show up in the add list even if it shouldn't. 
2.  A div was not properly closed, leading to a red text disaster!

## Ticket

### DIFF
--- a/website/static/js/institutionProjectSettings.js
+++ b/website/static/js/institutionProjectSettings.js
@@ -48,6 +48,7 @@ var ViewModel = function(data) {
         var message;
         var modifyOneMessage;
         var modifyAllMessage;
+        var htmlMessage;
         if (self.isAddInstitution()) {
             message = 'Add <b>' + item.name + '</b> to <b>' + data.node.title + '</b> or to <b>' +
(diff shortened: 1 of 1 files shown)


website/static/js/institutionProjectSettings.js
```diff
@@ -165,7 +174,9 @@ var ViewModel = function(data) {
             else {
                 index = self.affiliatedInstitutionsIds().indexOf(item.id);
                 var removed = self.affiliatedInstitutions.splice(index, 1)[0];
-                self.availableInstitutions.push(removed);
+                if (self.userInstitutionsIds.indexOf(removed.id) > -1) {
```
sloria: Won't this always be false? IIRC, `indexOf` returns `-1` if the item isn't in the list.

sloria: Oh, nvm, this is right.

PR #5775, opened 2016-06-01
[Feature] Checkout files on prereg submission [OSF-6411]

Ghalko:
## Purpose

On submission to the prereg challenge, this takes the files that the user has indicated they want to submit for review and checks out those files prereg user's account (David). Really this is only a stop-gap measure to warn the user that the files are checked out. They can still make changes to them if they are admin and can force check-in files.

## Changes

- Add three methods to Draft Registrations
- `get_metadata_files` - returns a list of files from the deep wells of metadata.
- `checkout_files` - checks out all files from previous method
- `checkin_files` - checks in all files from `get_metadata_files

## Side effects

Staging: add a user guid to `PREREG_FILE_CHECKOUT_USER` in `local.py` in order to facilitate testing.
Production: add David's guid to `PREREG_FILE_CHECKOUT_USER` for this feature to work.

## Ticket


admin_tests/pre_reg/test_views.py
```diff
@@ -25,7 +25,8 @@
 
 
 class TestDraftListView(AdminTestCase):
-    def setUp(self):
+    @mock.patch('admin.pre_reg.views.DraftRegistration.get_metadata_files')
```
mfraezz: Why does this need to be mocked? It's not making any outgoing calls, and you're not assigning any side effects / return values / etc. to it here.

mfraezz: This applies to all added uses of `@mock.patch` in this PR

website/project/model.py
```diff
@@ -4000,16 +4003,45 @@ def approve(self, user):
...
         self.approval.reject(user)
         self.add_status_log(user, DraftRegistrationLog.REJECTED)
         self.approval.save()
+        self.checkin_files()
 
     def add_status_log(self, user, action):
         log = DraftRegistrationLog(action=action, user=user, draft=self)
         log.save()
 
+    def get_metadata_files(self):
```
mfraezz: I'm generally opposed to adding new things to `website/project/model.py` unless absolutely necessary. That being said, I'm not certain that there's a good way around this without a refactor to how registrations work. It *may* be possible to modify the `PreregCallbackMixin` in `website/project/sactions.py` instead.

website/project/model.py
```diff
@@ -4000,16 +4003,45 @@ def approve(self, user):
...
+        for q in ['q7', 'q16', 'q11', 'q13', 'q12', 'q19']:
+            for file_info in self.registration_metadata[q]['value']['uploader']['extra']:
+                if file_info['data']['provider'] != 'osfstorage':
+                    continue
+                fid = file_info['data']['path'].split('/')[1]
+                yield OsfStorageFileNode.load(fid)
+
+    def checkout_files(self, user, save=False):
+        """Check out all metadata files"""
+        for item in self.get_metadata_files():
```
mfraezz: Isn't this supposed to happen iff it's a submission to the Prereg challenge, not for just any registration?

website/project/model.py
```diff
@@ -4000,16 +4003,45 @@ def approve(self, user):
...
+            try:
+                item.checkout = user
+            except AttributeError:
+                continue  # ignore files that have changed
+            if save:
+                item.save()
+
+    def checkin_files(self):
+        """Check in all metadata files"""
+        for item in self.get_metadata_files():
```
mfraezz: Ditto [this comment](https://github.com/CenterForOpenScience/osf.io/pull/5775/files#r67401076)

>Isn't this supposed to happen iff it's a submission to the Prereg challenge, not for just any registration?

website/project/model.py
```diff
@@ -3978,6 +3979,8 @@ def submit_for_review(self, initiated_by, meta, save=False):
         approval.save()
         self.approval = approval
         self.add_status_log(initiated_by, DraftRegistrationLog.SUBMITTED)
+        prereg_user = User.load(settings.PREREG_FILE_CHECKOUT_USER)
```
mfraezz: Given that this should only ever be used inside of `checkout_files` (perhaps more aptly named `checkout_preregistered_files`), could this be loaded inside of that function instead, after a check is done to ensure that this reg is a prereg?

mfraezz:
Pass complete, back to you @Ghalko. Please also add a few tests for new behavior. 🌐

admin_tests/pre_reg/test_views.py
Ghalko: I don't need to have all the file data in the draft registration for all of these operations, yet I must submit for review which now does a `get_metadata_files` in setup. Therefore it is mocked in setup.

website/project/model.py
Ghalko: I suppose so. Hmmm...

Ghalko: Done.

website/project/model.py
Ghalko: Done.

website/project/model.py
Ghalko: Done.

admin/pre_reg/views.py
```diff
@@ -102,10 +102,9 @@ class DraftDetailView(PreregAdmin, DetailView):
     context_object_name = 'draft'
 
     def get_object(self, queryset=None):
+        draft = DraftRegistration.load(self.kwargs.get('draft_pk'))
```
mfraezz: Without a default, `.get` can raise an `AttributeError`, which is probably why it was previously all being done in the `try` block.

admin_tests/pre_reg/test_views.py
mfraezz: I still think it would be good to not mock it. You're preventing the testing of possible live behavior by mocking it. 

If you don't need to have file data in the draft (as many users might), not having file data in the draft shouldn't break the submit process. This may mean better error handling in `get_metadata_files` is required (it looks like `KeyErrors` might be raised without files).

website/project/model.py
```diff
@@ -4004,16 +4008,51 @@ def approve(self, user):
         self.approval.approve(user)
         self.add_status_log(user, DraftRegistrationLog.APPROVED)
         self.approval.save()
+        self.checkin_files(save=True)
```
mfraezz: If multiple users need to `.approve` a registration before it goes through, could this potentially trigger a check-in too early?

website/project/model.py
```diff
@@ -4004,16 +4008,51 @@ def approve(self, user):
...
         self.add_status_log(user, DraftRegistrationLog.REJECTED)
         self.approval.save()
+        self.checkin_files(save=True)
 
     def add_status_log(self, user, action):
         log = DraftRegistrationLog(action=action, user=user, draft=self)
         log.save()
 
+    def get_metadata_files(self):
+        for q in ['q7', 'q16', 'q11', 'q13', 'q12', 'q19', 'q26']:
```
mfraezz: What happens if we release a version `3` of the prereg schema, with more/different Q's accepting files?

website/project/model.py
```diff
@@ -4004,16 +4008,51 @@ def approve(self, user):
...
+            for file_info in self.registration_metadata[q]['value']['uploader']['extra']:
```
mfraezz: What happens if one or more of these keys are missing, if a question is submitted without a file attached?

admin/pre_reg/views.py
Ghalko: Default for `.get` and `.load`
screen shot 2016-06-21 at 1 31 01 pm
 is `None`

mfraezz: You're right, disregard.

website/project/model.py
Ghalko: The only place `.approve` is called is in prereg admin.

website/project/model.py
Ghalko: I can make a version test for that, or scrape the json to get these values. Which one is better? Scraping json would allow this to be for any registration. But do we need it for any other registration type? @brianjgeiger

Ghalko:
In favor of #5907

PR #5776, opened 2016-06-01
[RFR][Feature][OSF-6429]Make 409 mismatch error details more useful

TomBaxter:
## Purpose:
[OSF-6429](https://openscience.atlassian.net/browse/OSF-6429)
Add error specific details to three 409 Conflict exceptions

## Changes:
Update api/base/serializers.py add custom details to 409 Conflict calls

## Side effects
None

[#OSF-6429]


brianjgeiger:
While tests are running, can you add to the PR description examples of json requests and responses for these?

brianjgeiger:
Also, please put in the expected response from the call. This is for QA so they know what sorts of things they should send and what they should expect to get back. An example of each type would be super-handy for them.

api/base/serializers.py
```diff
@@ -225,7 +225,7 @@ def to_internal_value(self, data):
             if request.method in utils.UPDATE_METHODS and not utils.is_bulk_request(request):
                 id_field = getattr(self.root.instance, self.source, '_id')
                 if id_field != data:
-                    raise Conflict()
+                    raise Conflict(detail=('The resource id you specified "' + data + '" does not match the id of the resource you specified "' + id_field + '".'))
```
brianjgeiger: I think this could be more useful. In this case, one of these comes from the URL and the other comes from the JSON, yes? So maybe, `The id you used in the URL "axkrt" does not match the id you used in the json body's id field "axktr".`

api/base/serializers.py
```diff
@@ -249,7 +249,7 @@ def to_internal_value(self, data):
             type_ = self.root.Meta.type_
 
         if type_ != data:
-            raise Conflict()
+            raise Conflict(detail=('The resource type you specified "' + data + '" does not match the type of the resource you specified "' + type_ + '".'))
```
brianjgeiger: This could probably still stand to be more useful. `This resource has a type of "collections", but you set the json body's type field to "collection". You probably need to change the type field to match the resource's type.`

api/base/serializers.py
```diff
@@ -267,7 +267,7 @@ def __init__(self, **kwargs):
 
     def to_internal_value(self, data):
         if self.target_type != data:
-            raise Conflict()
+            raise Conflict(detail=('The resource type you specified "' + data + '" does not match the type of the target resource you specified "' + self.target_type + '".'))
```
brianjgeiger: This can probably be the same as the one above or similar.

api/base/serializers.py
brianjgeiger: You'll want to verify I'm correct about the second sentence though. I'm pretty sure I am, but if this error hits before the 404, that would be bad advice.

api/base/serializers.py
brianjgeiger: And since this is a relationship field, if you know which relationship it belongs to, we can probably make it even better.

api/base/serializers.py
```diff
@@ -225,7 +225,7 @@ def to_internal_value(self, data):
             if request.method in utils.UPDATE_METHODS and not utils.is_bulk_request(request):
                 id_field = getattr(self.root.instance, self.source, '_id')
                 if id_field != data:
-                    raise Conflict()
+                    raise Conflict(detail=('The id you specified in the URL "' + id_field + '" exists, but does not match the id you specified in the JSON body "' + data + '".'))
```
brianjgeiger: Use the string class's format method rather than concatenating strings. Sorry I didn't catch this last time. 
```python
'The id you specified in the url "{}" exists, but does not match  the id you specified in the JSON body "{}"'.format(id_field, data)
```

brianjgeiger: Also, are you just using your email notifications? Best to check the discussion; I had made some edits a minute or two after setting the original message that I think got lost. https://github.com/CenterForOpenScience/osf.io/pull/5776#discussion_r65555351

brianjgeiger: screen shot 2016-06-02 at 3 52 34 pm

brianjgeiger: Unless you checked and I'm wrong about the 404, in which case let me know and disregard all these extra messages.

api/base/serializers.py
```diff
@@ -249,7 +249,8 @@ def to_internal_value(self, data):
             type_ = self.root.Meta.type_
 
         if type_ != data:
-            raise Conflict()
+            raise Conflict(detail=('The type you specified in the URL "' +
```
brianjgeiger: This doesn't match what I suggested. Could it not be implemented that way or was it incorrect?
screen shot 2016-06-02 at 3 56 20 pm

TomBaxter: I'll change the error messages to your wording.

TomBaxter:
@brianjgeiger I apologize that I lost your original comments. I rebased after updating to latest develop. I want to double check before I make another commit. Are these messages correct?

```
{
"errors": [
{ "detail": "The id you used in the URL, \"q7tp2\", does not match the id you used in the json body's id field, \"blurt\". The object \"q7tp2\" exists, otherwise you'd get a 404, so most likely you need to change the id field to match." }
]
}
------------
{
"errors": [
{ "detail": "This resource has a type of \"nodes\", but you set the json body's type field to \"blurt\". You probably need to change the type field to match the resource's type." }
]
}
-----------
{
"errors": [
{ "detail": "The target resource has a type of \"users\", but you set the json body's type field to \"blurt\". You probably need to change the type field to match the target resource's type." }
]
}
```

  If not can you please repost the verbiage. Sorry for the rework.

brianjgeiger:
No problem. That looks about right. Two things to be sure of (which you may have done):

1. Verify the bit about the 404. I'm fairly certain, but best to doublecheck. You could even write a test for it.
2. Some relationships may be more complicated than this. Comments, primarily, can be attached to nodes or files or wikis. Make sure this error message works in those cases, or if something else happens with that. It may be you can't cause that message to appear because of how you link the items, I'm not sure. But just in case.

TomBaxter:
Yes, checked for 404 earlier. Just checked for comments. Looks like comment creation doesn't run through this check. Here's what I get for a comment added to node with type mismatch.

{
  "errors": [
    {
      "detail": "Invalid target type. Expected \"nodes\", got \"blurt.\""
    }
  ]
}

brianjgeiger:
Might as well fix that one while you're here.

brianjgeiger:
Looks great. 🎉 🐧 🎉

PR #5777, opened 2016-06-01
[Production and Staging] Fixed jumpy user and project settings side nav bar [#OSF-3825]

sf2ne:
## Purpose

In osf.io when on the user settings page or on the project settings page the side nav bar jumps to affix to the bottom of the page when the page is refreshed. This problem happens most consistently on Chrome and Firefox but I have found it also happens sometimes on Safari. 

User Settings Page
screen shot 2016-05-31 at 10 34 05 pm

Project Settings Page
screen shot 2016-05-31 at 10 39 27 pm

These changes were implemented to improve the user experience and have the side nav bar actually return to its proper place when the page is refreshed.

## Changes

The problem was found to be that in bootstrap when something like a panel is affixed both to the top and the bottom, there are instances of conflict between the two affixes. I was unable to find an HTML or CSS solution as simply removing the bottom affix causes the panel to run over the footer and trying to translate the panel up and down the y-axis merely stopped the panel from scrolling with the page. 

Instead a jquery solution was found consisting of a function called sticky_element which affixes the side nav panel to the bottom footer and then calculates the top affix based on the top of the screen window and not some affixed value through bootstrap. This bypasses the bootstrap affix complication entirely. To fix some minor bugs discovered in the process of writing that function two resize functions were put in to handle if a window is resized bigger or smaller. 

This is how the side nav bar looks now when refreshed at the bottom of the page:
screen shot 2016-06-01 at 11 23 09 am

## Side effects

The only side effect I could think of is that if the user resizes the screen to the mobile size and then refreshes the page. If after that they then resize back to a normal larger screen size the side nav bar stops scrolling.

Aside from that the only other thing added that should affect anything else is that I added div IDs to the user and project settings side nav panels and to the footer.

## Ticket

https://openscience.atlassian.net/browse/OSF-3825


website/templates/project/settings.mako
```diff
@@ -11,17 +11,81 @@
 ##        }
 ##    }'></div>
 
+<!-- The following jquery functions are to fix the bootstrap affix issue
```
mfraezz: Could you [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) this code up? 

Wrapping all of this in a function that accepts the target DOM element's id (and perhaps also configuration options -- e.g. offsets -- but that's not a blocker now as it's currently the same between both uses), and putting that in `osfHelpers.js` would probably be best. That way, you could just import and instantiate it; something like 
```js
var StickyElement = require('js/osfHelpers').StickyElement;
new StickyElement('#whicheversettingspanel'); 
```
should do all of this work.

sf2ne:
I committed again to put the jquery methods in the right place. Sorry, I had to undo one of my commits.

brianjgeiger:
Looks like you will need to handle some merge conflicts.

website/static/js/osfHelpers.js
```diff
@@ -857,6 +857,78 @@ function onScrollToBottom(element, callback) {
...
+ *  @param {String} panelID:
+ *  @param {Integer} panel_offset:
+ */
+var stickIt = function stickIt(panelID, panel_offset){
+    this.elementID = panelID;
+    this.offsetElement = panel_offset;
+
+    var lastWidth;
+    $(window).load(function(){
+        if($(window).width() > 769){
```
brianjgeiger: Magic numbers are generally a bad idea. Can you re-work this to use bootstrap? If we're looking for transitions from states, we should be checking what states transitions will cause a problem and maybe looking at what the values for those transitions are in a more generic way. That being said, that sounds like a lot of trouble for not much effort.

If we keep it in osfHelpers, it should probably be pretty generic, even if it's just a matter of passing some parameters from the calling javascript so that different pages with different rules will be able to pass that in properly. Alternately, we could move this function to the page that would use it (if it's only in one place) and maybe define the number as a constant somewhere in the js.

I lean towards the "passing in a parameter" approach, but think about what makes sense to you.

website/static/js/osfHelpers.js
```diff
@@ -857,6 +857,78 @@ function onScrollToBottom(element, callback) {
...
+        var stickyE   = panelID,
+            bottomE   = "#pagefooter";
+        if($( stickyE ).length){
+            $( stickyE ).each(function(){
+                var fromTop = $( this ).offset().top - panel_offset, // number of pixels to the top of element
+                // offset from the top of the screen
+                fromBottom = $( document ).height()-($( this ).offset().top + $( this ).outerHeight()),
+                //number of pixels from the top of the bottom element, adding height to take into account padding/borders
+                stopOn = $( document ).height()-( $( bottomE ).offset().top)+($( this ).outerHeight() - $( this ).height());
+                if( (fromBottom-stopOn) > 200){
```
brianjgeiger: Another magic number. Best to make a constant or a function parameter as mentioned above.

brianjgeiger:
A couple of items with magic numbers. Also, is the side effect you mentioned a thing, and if so, can you make it go away? It looks like your logic could be augmented to handle it. Pass done. 🐧

PR #5782, opened 2016-06-01
Change check mark to opacity toggle on create node with institution

hmoco:

### DIFF
--- a/website/static/js/components/institution.js
+++ b/website/static/js/components/institution.js
@@ -47,7 +47,7 @@ var SelectableInstitution = {
             filter: 'grayscale(100%)'
         };
         if (!checked) {
-            opts.style = mutedStyle;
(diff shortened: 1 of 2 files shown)


website/static/js/components/institution.js
```diff
@@ -41,24 +41,15 @@ var InstitutionImg = {
...
-                }
-            ) : '',
-        ]);
+        var mutedStyle = {
+            opacity: '0.25',
+            '-webkit-filter': 'grayscale(100%)',
+            filter: 'grayscale(100%)'
+        };
+        if (!checked) {
+            opts.style = mutedStyle;
```
sloria: This will override all styles passed to INstitutionImg; better to do 

```js
opts.style = $.extend({}, mutedStyle, opts.style);
```

PR #5784, opened 2016-06-02
[RFR][Feature][OSF-6440][Multiple Institutions] Add check for affiliation on dashboard

TomBaxter:
## Purpose:
[OSF-6440](https://openscience.atlassian.net/browse/OSF-6440)
Update user's affiliated email domain institution before user gets queried to create initial project.

## Changes:
Update framework/auth/core.py add "def update_affiliated_institutions_by_email_domain" to "class User"

Update framework/auth/__init__.py update "def authenticate" to call User.update_affiliated_institutions_by_email_domain

Update  website/project/views/node.py  update def node_setting to call User.update_affiliated_institutions_by_email_domain

## Side effects
None

[#OSF-6440]

### DIFF
--- a/api/users/views.py
+++ b/api/users/views.py
@@ -300,7 +300,6 @@ class UserNodes(JSONAPIBaseView, generics.ListAPIView, UserMixin, ODMFilterMixin
     # overrides ODMFilterMixin
     def get_default_odm_query(self):
         user = self.get_user()
-        user.update_affiliated_institutions_by_email_domain()
         query = Q('contributors', 'eq', user) & default_node_list_query()
         if user != self.request.user:
             query &= default_node_permission_query(self.request.user)
--- a/api/users/views.py
+++ b/api/users/views.py
@@ -300,6 +300,7 @@ class UserNodes(JSONAPIBaseView, generics.ListAPIView, UserMixin, ODMFilterMixin
     # overrides ODMFilterMixin
     def get_default_odm_query(self):
         user = self.get_user()
+        user.update_affiliated_institutions_by_email_domain()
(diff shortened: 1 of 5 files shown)


api/users/views.py
```diff
@@ -300,6 +300,7 @@ class UserNodes(JSONAPIBaseView, generics.ListAPIView, UserMixin, ODMFilterMixin
     # overrides ODMFilterMixin
     def get_default_odm_query(self):
         user = self.get_user()
+        user.update_affiliated_institutions_by_email_domain()
```
sloria: This probably isn't the place to so this; a `GET` request should not result in writes to the database.

How about putting this in the view for the Dashboard?

sloria:
Pass finished.

api/users/views.py
TomBaxter: @sloria Done.

sloria:
The code that checks for affiliation on the project settings page should use your added method.

framework/auth/core.py
```diff
@@ -937,6 +937,22 @@ def update_search_nodes_contributors(self):
...
+        :return:
+        """
+        # Avoid circular import
+        from website.project.model import Institution
+        try:
+            email_domains = [email.split('@')[1] for email in self.emails]
+            inst = Institution.find_one(Q('email_domains', 'in', email_domains))
+            if inst not in self.affiliated_institutions:
+                self.affiliated_institutions.append(inst)
+                self.save()
```
sloria: Methods should not make implicit saves. I'll fix this when I merge.

sloria:
Squashed and promoted to a hotfix in https://github.com/CenterForOpenScience/osf.io/pull/5793

PR #5785, opened 2016-06-02
Feature/fix institution karma test

billyhunt:
Related to #5726

## Purpose

Fix a karma test which needed the v1 url being called.

## Ticket


sloria:

PR #5787, opened 2016-06-02
[CR] switch user-facing analytics to Keen from Piwik

felliott:
## Purpose

Replace our Piwik-based analytics with Keen.io.  This will apply to both internal and user-facing analytics.

### Add anonymized pageviews

Keen can restrict access to sets of events, but not the content within those events.  For that reason, we need to create anonymized pageviews suitable for displaying to users.   These **must not** contain any identifying information (user id, email, ip address).  This PR updates the current Keen code to create both public and private pageview events and to log them to separate projects.  Public pageviews are only logged after a project is made public.

### Turn off Piwik tracking

Migrating the Piwik historical data will be done in the next phase.  We want to avoid double counting, so we're turning off Piwik logging at the same time we're turning on Keen logging.  Piwik as a whole has not been removed, since there is current functionality that relies on its data.  The rest of the Piwik code will be removed after the migration is run.

### Add keen-based user-analytics visualizations

Replace the current Piwik charts and graphs on the Analytics tab of public projects with new visualizations using Keen's dataviz libraries.  These will be sparsely populated until the historical data has been migrated.

## Side effects

Supersedes issue https://github.com/CenterForOpenScience/osf.io/pull/5273.  That PR will be split into two separate PRs:

1. Turn off Piwik logging and switch the user-facing analytics to Keen. (this PR)
2. Migrate the Piwik historical data into Keen and purge/replace the last remaining remnants of Piwik from the OSF codebase FOREVER!

Until the second PR is merged the Discover Public Projects functionality will still be based off of Piwik and will be stale.  Until the piwik historical data has been migrated and/or the keen analytics have had a chance to collect some data, the user-analytics are going to be *awfully* bare.

The library used to provide anonymized location data is [python-geoip-geolite2](https://pypi.python.org/pypi/python-geoip-geolite2).  It converts an IP address into a geographical location.  While writing the migration scripts, I found that repeated lookups can be slow.  This will probably be unnoticeable for individual requests, but it might slow down testing.  We skip the lookup if the request doesn't have a `remote_addr` property, but I don't know if the tests are mocking that.

## Migrations

Some local configuration will need to be updated.  In `websites/settings/local.py`, these config keys:

```
KEEN_PROJECT_ID = None
KEEN_WRITE_KEY = None
KEEN_READ_KEY = None
```
will need to be replaced with:
```
KEEN = {
    'public': {
        'project_id': None,
        'master_key': None,
        'write_key': None,
    },
    'private': {
        'project_id': None,
        'write_key': None,
        'read_key': None,
    },
}
```
I will provide the actual keys. There is also a migration script that needs to be run to generate read keys for already-public projects:

`python -m scripts.migrate_piwik.initialize_keen_read_keys`

## Ticket

Converting Analytics to Piwik: [#OSF-6069](https://openscience.atlassian.net/browse/OSF-6069)

**Next**: Migrating Piwik data to Keen [#OSF-5283](https://openscience.atlassian.net/browse/OSF-5283)

**Next, Next**: No ticket for it yet, but @caneruguz was kind enough to perform a UI consult. There will be some follow-up PRs to add some proper polish to the Analytics page.

## Credit

All praise to @GageGaskins, who kicked this whole thing off!  I stole authorship for some of his commits while rebasing this to split the PRs, but rest assured, they were his and they were *divine*.


website/project/model.py
```diff
@@ -1417,6 +1420,25 @@ def set_node_license(self, license_id, year, copyright_holders, auth, save=False
...
+        api_readkey = scoped_keys.encrypt(settings.KEEN['public']['master_key'], options={
+            'filters': [{
+                'property_name': 'node.id',
+                'operator': 'eq',
+                'property_value': str(self._id)
+            }],
+            'allowed_operations': ['read']
+        })
+        self.keenio_read_key = api_readkey
+        self.save()
```
sloria: We generally try to avoid implicit saves within model methods. Can this be removed?

website/routes.py
```diff
@@ -103,8 +106,10 @@ def get_globals():
         'reauth_url': util.web_url_for('auth_logout', redirect_url=request.url, reauth=True),
         'profile_url': cas.get_profile_url(),
         'enable_institutions': settings.ENABLE_INSTITUTIONS,
-        'keen_project_id': settings.KEEN_PROJECT_ID,
-        'keen_write_key': settings.KEEN_WRITE_KEY,
+        'keen_public_project_id': settings.KEEN['public']['project_id'],
```
sloria: Minor: We generally prefer namespaced representations over prefixes

```python
'keen': {
    'public_project_id': ...,
     ....
}

Comment (sloria) [inline website/routes.py]: Or even:


```python
'keen': settings.KEEN
```

website/static/js/keen.js
```diff
@@ -1,127 +1,242 @@
...
+
+            self.trackPageView = function () {
+                var self = this;
+                if (window.contextVars.node && window.contextVars.node.isPublic) {
+                    self.trackPublicEvent('pageviews', {});
+                }
+                self.trackPrivateEvent('pageviews', {});
+            };
+
+            self.trackPrivateEvent = function(collection, event) {
```
sloria: Minor: These methods can be simplified to

```js
self.trackPrivateEvent = _trackCustomEvent.bind(self, self._privateClient);
```

website/static/js/statistics.js
```diff
@@ -0,0 +1,213 @@
...
+            'params': {
+                event_collection: 'pageviews',
+                timeframe: 'this_7_days',
+                target_property: 'anon.id',
+                group_by: 'time.local.hour_of_day',
+            }
+        };
+
+        var dataviz = new keenDataviz()
+                .el('#serverTimeVisits')
```
sloria: It would be good to refactor out these hardcoded CSS IDs. Instead, these should be passed at initialization time (in the page module).

Something like

```js
// statistics-page.js

projectUsageStats.
    visitsServerTime('#serverTimeVisits')
    visitsByDay('#visits')
    // ...
```

website/static/js/statistics.js
```diff
@@ -0,0 +1,213 @@
...
+        self.visitsByDay();
+        self.topReferrers();
+        self.visitsServerTime();
+        self.popularPages();
+    };
+
+
+};
+
+function ProjectUsageStatistics() {
```
sloria: Not sure this wrapper class is necessary.

website/templates/base.mako
```diff
@@ -189,7 +185,11 @@
                     timezone: ${ user_timezone | sjson, n },
                     entryPoint: ${ user_entry_point | sjson, n },
                     institutions: ${ user_institutions | sjson, n},
-                    emailsToAdd: ${ user_email_verifications | sjson, n }
+                    emailsToAdd: ${ user_email_verifications | sjson, n },
+                    anon: {
```
sloria: See my previous comment re: nested, namespaced representations. If `get_globals` is changed, this can be simplified to

```js
anon: ${ anon | sjson, n}
```

sloria:
Pass finished. Lookin good! No major blocking changes here, other than getting this up to date with develop.

website/routes.py
felliott: We have keys in the settings we don't want to risk leaking into mako, but point taken.  I'll update this to `keen: { public: {}, private: {}}`.

website/static/js/keen.js
felliott: Ooooh.  I've seen `.bind()` but never used it.  Thanks for pointing that out!

website/project/model.py
felliott: We did it this way to assign read keys on-demand, as public pages are visited for the first time.  I suppose we could factor it out into a migration, instead.  Would that be preferable?

felliott: Nevermind, answered my own question.  Yes, this should be a migration or else `set privacy` will start acting weird.

website/static/js/keen.js
felliott: Doh! `.bind()` resolves `self._privateClient` before it's defined.

website/static/js/statistics.js
felliott: Good catch, will remove.

website/static/js/statistics.js
felliott: Will do!

website/static/js/keen.js
sloria: Ah, ok. This is fine as is, then.

scripts/migrate_piwik/initialize_keen_read_keys.py
```diff
@@ -0,0 +1,35 @@
...
+from website.app import init_app
+
+
+logger = logging.getLogger(__name__)
+
+
+if __name__ == '__main__':
+    init_app(routes=False)
+
+    public_nodes = Node.find(Q('is_public', 'eq', True))
```
sloria: Minor: We can probably skip deleted projects. Not a big deal, though.

webpack.common.config.js
```diff
@@ -47,6 +47,7 @@ var entry = {
     'profile-settings-personal-tokens-list-page': staticPath('js/pages/profile-settings-personal-tokens-list-page.js'),
     'profile-settings-personal-tokens-detail-page': staticPath('js/pages/profile-settings-personal-tokens-detail-page.js'),
     'sharing-page': staticPath('js/pages/sharing-page.js'),
+    'statistics-page': staticPath('js/pages/statistics-page.js'),
```
sloria: There is already a `statistic-page` entry point, so this can be removed.

website/static/js/statistics.js
```diff
@@ -0,0 +1,242 @@
...
+                            }
+                        }
+                    },
+                });
+
+        var munger = function() {
+            this.dataset.updateColumn(0, function(value, index, column) {
+                var title = value.replace(/^OSF \| /, '');
+                // Strip off the project title, if present at beginning of string
+                if (title.startsWith(window.contextVars.node.title)) {
```
sloria: We generally try to avoid using contextVars in re-usable modules. Can this information be passed in as arguments?

This isn't a blocker.

website/static/js/pages/statistics-page.js
```diff
@@ -13,3 +19,78 @@ $(function(){
         $adBlock.show();
     }
 });
+
+keenAnalysis.ready(function(){
+    function drawAllCharts(statsStartDate, statsEndDate) {
+        projectUsageStats.visitsByDay('#visits', startDate, endDate);
```
sloria: 👍

website/static/js/pages/statistics-page.js
```diff
@@ -13,3 +19,78 @@ $(function(){
...
+        projectUsageStats.topReferrers('#topReferrers', startDate, endDate);
+        projectUsageStats.popularPages('#popularPages', startDate, endDate);
+        projectUsageStats.visitsServerTime('#serverTimeVisits', startDate, endDate);
+    }
+
+    function updateStartDate() {
+        startPicker.setStartRange(startDate);
```
sloria: Minor: It's a bit confusing to use variables that are outside the scope of this function (`startPicker` and `startData`). In contrast, `drawAllCharts` receives its dependencies as arguments. Better to make this consistent. Alternatively, you could create a class that stores the dependencies as attributes.

```js
var DatePicker = oop.defclass({
    constructor: function(elem) {
        var self = this;
        self.startDate = ...
        self._startPicker = new pikaday(..)
    },
    updateStartDate: function(start) {
        self._startPicker.setStartRange(start)
        ...
    }
})
```

Certainly not a blocker, though.

website/static/js/statistics.js
```diff
@@ -0,0 +1,242 @@
...
+    var self = this;
+
+    /**
+    * Manages the connections to Keen.  Needs a project id and valid read key for that project.
+    *
+    * @property keenClient
+    * @type {keenAnalysis}
+    */
+    self.keenClient = new keenAnalysis({
+        projectId: window.contextVars.keen.public.projectId,
```
sloria: Minor: Same here--can these be passed as arguments rather than pulling from contextVars? contextVars should only be accessed in the `-page` modules.

sloria:
Pass finished. Again, no major blockers. I'll check back in on this PR in an hour or so and get it merged. In the meantime, feel free to address any of the nits above.

felliott:
Will do.  Thanks, @sloria!

website/static/js/dateRangePicker.js
```diff
@@ -0,0 +1,62 @@
...
+'use strict';
+
+require('pikaday-css');
+
+var oop = require('js/oop');
+var pikaday = require('pikaday');
+
+
+var DateRangePicker = oop.defclass({
+    constructor: function(startPickerElem, startDate, endPickerElem, endDate, minDate, maxDate) {
```
sloria: Minor: This is ok for now, but with that many arguments, it's more readable to use an options object to pass arguments:

```
new DateRangePicker({
    startPicker: '#startPicker',
    startDate: new Date(...),
     //...
});
```

sloria:
Created a feature branch for this: `feature/keen`. Consider this merged!

PR #5789, opened 2016-06-03
[hotfix]Allow deletion of Institutions;  don't show private nodes on OSF4I landing pages

hmoco:
## Purpose

Allow for deletion of institutions using the `is_deleted` flag, and making sure deleted intitutions dont have a landing page and can't be found in search.

## Changes
Institution now has is_deleted attribute. Update_search looks for that to delete the elastic doc.

## Ticket
Also this https://openscience.atlassian.net/browse/OSF-6453

### DIFF
--- a/website/institutions/model.py
+++ b/website/institutions/model.py
@@ -129,7 +129,7 @@ class Institution(object):
     def load(cls, key):
         from website.models import Node
         try:
-            node = Node.find_one(Q('institution_id', 'eq', key) & Q('is_deleted', 'ne', True), allow_institution=True)
+            node = Node.find_one(Q('institution_id', 'eq', key), allow_institution=True)
             return cls(node)
         except NoResultsFound:
             return None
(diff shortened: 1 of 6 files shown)


website/institutions/model.py
```diff
@@ -128,7 +129,7 @@ def find_one(cls, query=None, deleted=False, **kwargs):
     def load(cls, key):
         from website.models import Node
         try:
-            node = Node.find_one(Q('institution_id', 'eq', key), allow_institution=True)
+            node = Node.find_one(Q('institution_id', 'eq', key) & Q('is_deleted', 'ne', True), allow_institution=True)
```
icereval: do we want to prevent loading the institution model by id? seems like we should be filtering on view/query right?

sloria: I think I agree. We typically don't try to hide `is_deleted` at the model level. Querying at the view layer will keep consistency with other models.

PR #5790, opened 2016-06-03
change redirect url to cas

billyhunt:
Related to PR 5000

## Purpose

Change login from login.mako to cas login.

## Ticket

https://openscience.atlassian.net/browse/OSF-4542

### DIFF
--- a/framework/auth/views.py
+++ b/framework/auth/views.py
@@ -199,10 +199,10 @@ def auth_logout(redirect_url=None):
     return resp
 
 
-def auth_email_logout(token, user):
+def auth_email_logout(token, user, redirect_url=None):
     """When a user is adding an email or merging an account, add the email to the user and log them out.
     """
-    redirect_url = web_url_for('auth_login')
+    redirect_url = cas.get_login_url(redirect_url)
     try:
         unconfirmed_email = user.get_unconfirmed_email_for_token(token)
     except InvalidTokenError:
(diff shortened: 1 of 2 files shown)


framework/auth/views.py
```diff
@@ -199,10 +199,10 @@ def auth_logout(redirect_url=None):
     return resp
 
 
-def auth_email_logout(token, user):
+def auth_email_logout(token, user, redirect_url=None):
     """When a user is adding an email or merging an account, add the email to the user and log them out.
     """
-    redirect_url = web_url_for('auth_login')
+    redirect_url = cas.get_login_url(redirect_url)
```
sloria: `redirect_url` will always be `None` here, since you're never passing it to `auth_email_logout`. I don't think `get_login_url` handles `None` as a service URL. Look into other usages of `get_login_url` to figure out the right thing to do here.

sloria:
Pass finished.

framework/auth/views.py
billyhunt: In every case I see, get_login_url is using a redirect_url that is set to none or pulled from a request.  How do you think I should handle it?  Set request_url=None in get_login_url?  Set request_url to '/' or whatever works in auth_email_logout?

sloria: Oh, ok, maybe it does handle `None`. Yeah, I think setting redirect_url to `/` makes sense in the call to `get_login_url` makes sense.

PR #5795, opened 2016-06-03
Hotfix/Fix Locking By Moving Tasks to Celery

cwisecarver:
## Purpose

Hotfix this into master

## Side effects

As of right now, there is a chance that things will still run concurrently because I'm calling delay instead of chaining the tasks. There is a retry on them so if they fail because of a lock they should retry after 60 seconds when the lock will be long gone. I have a working PR to fix this problem but I haven't submitted it because it's getting a maximum recursion depth exceeded error on extremely large changes like adding a component with 100+ contributors. This PR does not have that problem.


framework/analytics/__init__.py
```diff
@@ -72,6 +75,8 @@ def build_page(rex, kwargs):
         return None
 
 
+@run_postcommit(once_per_request=False, celery=True)
```
sloria: We removed this due to lack of request context. I reckon that's why there's a merge conflict.

cwisecarver: I think you're right @sloria. I just tried to merge master into my feature branch and it just reverted all my changes and had a bunch of merge conflicts. I'm trying again.

PR #5796, opened 2016-06-04
Hand-patching codes back in for celery tasks

cwisecarver:
## Purpose

This should fix the locking problem and analytics not updating.

## Side effects

Because I'm just calling delay on the celery tasks it could still create locking but the retries and the 60 second wait should mitigate that until I can get a PR in that fixes it for good. Because of the retries things put into celery could potentially run out of order.


icereval:
🎉

PR #5798, opened 2016-06-06
Fix external sites opening in new tabs [#OSF-6426]

erinspace:
## Purpose
- External links from the profile page, OSF for meetings pages, or forward widgets open tabs in a new window, which could lead to some undesirable effects (see tickets for more detail)

## Changes

- Remove "_blank" target from link in forward widget on configured projects
- Remove "_blank" target from log link for changed forward URL
- Remove "_blank" target from link in external websites in the user profile
- Remove "_blank" target from external conference website URL

## Side effects
- None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-6426

### DIFF
--- a/website/addons/forward/templates/forward_widget.mako
+++ b/website/addons/forward/templates/forward_widget.mako
@@ -22,7 +22,7 @@
 
         <div>
             This project contains a forward to
-            <a data-bind="attr: {href: url}, text: linkDisplay" target="_blank"></a>.
+            <a data-bind="attr: {href: url}, text: linkDisplay"></a>.
         </div>
 
         <div class="spaced-buttons m-t-sm">
--- a/website/addons/forward/templates/log_templates.mako
+++ b/website/addons/forward/templates/log_templates.mako
@@ -1,6 +1,6 @@
 <script type="text/html" id="forward_url_changed">
 changed forward URL to
-<a target="_blank" class="overflow log-file-link" data-bind="attr: {href: params.forward_url}, text: params.forward_url"></a>
+<a class="overflow log-file-link" data-bind="attr: {href: params.forward_url}, text: params.forward_url"></a>
 in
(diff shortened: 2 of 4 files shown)


mfraezz:
Manual testing indicated that putting `rel="noreferrer"` in the anchor tags instead of removing `target="_blank"` might be a better solution -- it effectively strips `window.opener` from the new tab without removing desired functionality. Given that we're not using `window.opener` anywhere, this shouldn't cause any issues.

brianjgeiger:
I think @billyhunt  checked with @saradbowman, and I believe everyone was okay with losing that functionality in these instances. I'm good with the solution as implemented unless anyone objects.

brianjgeiger:
Unless this solution is less secure, but I don't believe it is.

erinspace:
As I understood it, it also wasn't losing functionality so much as providing a clearer UI by not using a new tab. Happy to change it to whatever though.

mfraezz:
Sara confirmed, desired behavior is implemented. Dropping this in RTM 🀄

sloria:
@erinspace The article you cite is referring specifically to opening new browser *windows* (not tabs), which is 100% terrible.

As far as opening external links in new *tabs*, that's up for debate:

- Point: http://uxmovement.com/navigation/why-external-links-should-open-in-new-tabs/
- Counterpoint: http://www.aphoenix.ca/why-external-links-should-not-open-in-new-tabs-01/

I was actually surprised to find a Google Docs link that opened a new tab today:

2016_06_06_-_google_docs

I don't feel strongly either way, and I hate to be the one to open bikeshedding season. So unless Sara or Nici have objections, I think this is good to merge.

erinspace:
I am OK with anything - just someone tell me which do to.

sloria:
Ah, just saw Matt's comment. OK, looks like this is good to merge

PR #5799, opened 2016-06-06
[hold] [Hotfix] remove withdrawn registrations from search [OSF-6456]

leb2dg:
## Purpose

Withdrawn registrations and their files currently appear in search results. This PR deletes the elastic search index when the registration is withdrawn. This isn't really a bug but more of a change in spec.

## Changes

If the node is retracted, the elastic document is treated as if it is deleted or private.

## Side effects

Withdrawn registrations and their files will no longer appear in search results. This will only effect nodes that go through `update_node` or `update_file`.

## Ticket

https://openscience.atlassian.net/browse/OSF-6456

### DIFF
--- a/tests/test_elastic.py
+++ b/tests/test_elastic.py
@@ -279,14 +279,14 @@ class TestRegistrationRetractions(SearchTestCase):
         ) as registration:
             self.registration = registration
 
-    def test_retraction_is_searchable(self):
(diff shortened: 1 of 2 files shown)


sloria:
The current production behavior is actually correct. See https://openscience.atlassian.net/browse/OSF-6456

PR #5801, opened 2016-06-06
Social media sharing bar [OSF-2052]

acshi:
## Purpose
Enable social media sharing with buttons that connect to the appropriate services.

## Changes
Project and component pages have a "share" button that shows a selection of social media sharing buttons.
screen shot 2016-07-13 at 10 47 08 am

The existing file share dialog now also has these social media buttons.
screen shot 2016-07-12 at 9 26 39 am

When any of these buttons are clicked, a new window with the appropriate service is shown.
image

For each of the options, the title of the project, component, or file is given, along with the appropriate link url: main urls for projects and components, and direct MFR views for files. Twitter additionally has "via @OSFramework" added to the text.

## Side effects
There should not be any side effects.

## Ticket
https://openscience.atlassian.net/browse/OSF-2052


website/static/font-awesome/css/font-awesome.css
```diff
@@ -0,0 +1,2199 @@
+/*!
```
mfraezz: This file shouldn't need to be here; we already have `fa` icons and css classes.

website/static/js/filepage/index.js
```diff
@@ -53,7 +53,28 @@ var SharePopover =  {
...
-                        m('.tab-pane.active#share', m('.input-group', [
+                        m('.tab-pane.active#share', {
+                            config: function(element) {
+                                var shareContents = $('#shareContentHolder').html();
+                                if ($(element).find('.shareButtons').length === 0) {
+                                    var contentUrl = $('#share input').val();
+                                    var encodedUrl = encodeURIComponent(contentUrl);
+                                    var safeUrl = $('<div>').text(contentUrl).html();
+
+                                    shareContents = shareContents.replace('CONTENT_URL', safeUrl);
```
mfraezz: Generally, prefer mako templating over manual HTML replacing.

website/static/js/project.js
```diff
@@ -313,6 +313,10 @@ $(document).ready(function() {
         window.location = '/search/?q=(tags:"' + $(e.target).text().toString().trim()+ '")';
     });
 
+    $('#shareDropDownMenu li a').on('click', function(e) {
```
mfraezz: This appears to be the same (excepting the UI selector string) as something in `filepage/index.js`, consider [DRYing](https://en.wikipedia.org/wiki/Don't_repeat_yourself) this up.

website/static/js/project.js
```diff
@@ -313,6 +313,10 @@ $(document).ready(function() {
...
+        window.open(this.href, '', 'menubar=no,toolbar=no,resizable=yes,scrollbars=yes,width=600,height=400');
```
mfraezz: It's generally considered bad UI practice to force something to open in a new window -- has this been approved by the product team?

mfraezz:
This UI could look better:

screen shot 2016-06-07 at 14 16 57

Ideally, I think the button container should be centered below the `Share` button, but the social buttons themselves should at least have consistent whitespace around them (the margin on the right is too big). Might want to confirm that (and other UI) with @saradbowman or @nicipfeiffer; UI/UX isn't my specialty

nicipfeiffer:
@mfraezz @acshi Is the share/embed popover no longer happening when the user clicks the share button?

acshi:
@nicipfeiffer The elements shown above are indeed both popovers; is this what you meant?

nicipfeiffer:
@acshi Sorry for the vague question...I was specifically referring to this:
screenshot 2016-06-07 16 34 33
Where the share link is displayed and the embed option resides, is this still on the file detail page?

mfraezz:
LGTM. Please pull from `develop` and resolve the conflcts, after that and a successful TCI build this should be ready to merge 🉐

website/templates/project/project.mako
```diff
@@ -51,8 +51,10 @@
                             <a class="btn btn-default" href="#nodesPrivacy" data-toggle="modal">Make Private</a>
                         % endif
                         <button class="btn btn-default disabled">Public</button>
-                    % endif
                     </div>
+
+                    % endif
```
mfraezz: It looks like the `% if` this is closing is contained within a `<div class="btn-group">`, so the closing `</div>` should be below this.

mfraezz: This is trivial, though. @sloria can probably just fix on merge

website/static/css/pages/share-buttons.css
```diff
@@ -0,0 +1,41 @@
+div.dropdown .dropdown-menu li, .shareButtons {
```
sloria: As per COSDev, use `lower-dashed-case` for CSS classes.

website/static/css/pages/share-buttons.css
```diff
@@ -0,0 +1,41 @@
...
+    margin: 3px 20px;
+}
+
+div.dropdown .dropdown-menu li a, .shareButtons a {
```
sloria: Is it necessary to select `div.dropdown...`? I'm concerned that overriding Bootstrap's defaults might have unwanted side effects.

website/templates/project/view_file.mako
```diff
@@ -1,4 +1,7 @@
 <%inherit file="project/project_base.mako"/>
+
+<link rel="stylesheet" href="/static/css/pages/share-buttons.css">
```
sloria: This should go in the `stylesheets` def.

website/templates/project/project.mako
sloria: @acshi Go ahead and fix this when you make the other minor changes.

website/static/js/filepage/index.js
```diff
@@ -53,7 +53,15 @@ var SharePopover =  {
                         m('li', m('a[href="#embed"][data-toggle="tab"]', 'Embed'))
                     ]), m('br'),
                     m('.tab-content', [
-                        m('.tab-pane.active#share', m('.input-group', [
+                        m('.tab-pane.active#share', {
+                            config: function(element) {
+                                if ($(element).find('.shareButtons').length === 0) {
+                                    var shareContents = $('#shareContentHolder').html();
```
sloria: Manipulating the DOM manually like this should be unnecessary. Also, this approach has duplication--the HTML for the social media buttons need to be copied into every template they're used.

Since mithril is already used here, why not create a mithril component for the SHARE buttons?

```javascript
// in js/components/socialShare.js
var ShareButtons = {
    view: function(ctrl, options) {
        return m('div.share-buttons', {}, [
            m('a', {href: 'https://twitter...'})
            // ...
        ])
    }
}

// filepage/index.js
m('.tab-page.active#share', {}, [
    m.component(ShareButtons, {})
])
```

sloria: This would allow you to reuse the component on both the file view page and the project overview.

```js
// in js/components/socialShare.js
var ShareDropdown = {
    view: function(ctrl, options) {
        return m('.btn-group dropdown', {}, [
            //...
            m.component(ShareButtons)
        ])
    }
}
```
```mako
// project.mako
<div id="shareDropdown"></div>
```

// project-dashboard-page.js
if (window.contextVars.node.isPublic) { m.mount(document.getElementById('shareDropdown'), ShareDropDown); }
```

**Comment by sloria** [comment]
Pass finished.

**Comment by mfraezz** [comment]
Please resolve merge conflicts and fix the build error 🐼

File: website/static/js/pages/project-dashboard-page.js [hunk 10]
```diff
@@ -285,4 +286,19 @@ $(document).ready(function () {
     $('a.btn').mouseup(function(){
         $(this).blur();
     });
+
+    if (window.contextVars.node.isPublic) {
+        m.mount(document.getElementById('shareDropdown'),
+                m.component(SocialShare.ShareDropdown,
+                    {file_name: window.contextVars.node.title, share_url: window.location.href}));
+
+        // Make sure the drop down share menu is not obscured
```
sloria: I've just made a fix on develop that should make these two event handlers unnecessary. The tooltip should properly disappear when hovering off of the "Duplicate" button.

@acshi Can you verify this after you pull from develop?

website/static/css/pages/share-buttons.css
```diff
@@ -0,0 +1,37 @@
...
+    display: inline-block;
+}
+
+.share-buttons .fa {
+    color: #08c;
+    font-size: 1.6em;
+    line-height: 1.6em;
+}
+
+#share .share-buttons {
```
sloria: I don't see any elements with an id of "share". Can this rule be removed?

website/static/css/pages/share-buttons.css
```diff
@@ -0,0 +1,37 @@
...
+    white-space: nowrap;
+    background-color: #f8f8f8;
+}
+
+#shareDropdown .dropdown-menu a:hover {
+    background-image: none;
+    background-color: #f8f8f8;
+}
+
+#shareContentHolder {
```
sloria: I don't see elements with this ID either; remove?

website/static/js/components/socialshare.js
```diff
@@ -0,0 +1,52 @@
...
+var utils = require('js/components/utils');
+var required = utils.required;
+
+var ShareButtons = {
+    openLinkInPopup: function(href) {
+        window.open(href, '', 'menubar=no,toolbar=no,resizable=yes,scrollbars=yes,width=600,height=400');
+        return false;
+    },
+    view: function(ctrl, options) {
+        var file_name = encodeURIComponent(required(options, 'file_name'));
```
sloria: As per COSDev, use lowerCamelCase for variable names.

sloria: Note: The passed options should also be lowerCamelCase.

sloria: Also, `file_name` is a misnomer; you are passing the node title to it. How about just `name`?

website/static/js/components/socialshare.js
```diff
@@ -0,0 +1,52 @@
...
+        var share_url = encodeURIComponent(required(options, 'share_url'));
```
sloria: Minor: `share_url` is a bit redundant (this is a ShareButtons component); how about just `url`?

website/templates/project/view_file.mako
```diff
@@ -214,7 +215,8 @@
       window.contextVars.file.urls.external = window.contextVars.file.extra.webView;
     </script>
 
-    <link href="/static/css/pages/file-view-page.css" rel="stylesheet">
+    <link href="/static/css/pages/file-view-page.css" rel="stylesheet" />
+    <link href="/static/css/pages/share-buttons.css" rel="stylesheet" />
```
sloria: Since these styles are tightly tied to the mithril components you created, it would make sense to import this css file (using `require`) in `socialshare.js`.

website/static/js/components/socialshare.js
```diff
@@ -0,0 +1,52 @@
...
+                m('i.fa.fa-facebook-square[aria-hidden=true]')),
+            m('a', {href: linkedinHref, target: '_blank', onclick: this.openLinkInPopup.bind(this, linkedinHref)},
+                m('i.fa.fa-linkedin-square[aria-hidden=true]')),
+            m('a', {href: emailHref, target: '_blank', onclick: this.openLinkInPopup.bind(this, emailHref)},
+                m('i.fa.fa-envelope-square[aria-hidden=true]')),
+        ]);
+    }
+};
+
+var ShareDropdown = {
```
sloria: Nice work on this.

website/static/js/components/socialshare.js
```diff
@@ -0,0 +1,52 @@
...
+    view: function(ctrl, options) {
+        return [
+            m('a.btn.btn-default[data-toggle=dropdown]', 'Share'),
+            m('ul.dropdown-menu.pull-right[role=menu]#shareDropdownMenu', {}, [
```
sloria: Is the `shareDropdownMenu` CSS ID necessary? If not, remove it.

sloria:
Pass finished. Looking much better. Just a few details to iron out.

caneruguz:
Hey @acshi , @sloria asked me to do a UI review. Here are some comments:
- You are saying that this is using popover but it has the dropdown style. Are you sure you are implementing bootstrap popover correctly? For instance why does it not have the arrows pointing to the button? Can you check your implementation? http://getbootstrap.com/javascript/#popovers

- In terms of design would something like this work better? It would use non-square versions. In this example I changed the first two and kept the last two for comparison. 
screen shot 2016-07-06 at 12 46 40 pm

- It looks like linked in shares also have text like twitter. I don't know if it would be possible to provide a template text for that as well. Did you look into that? 

- Not a concern for this PR but we have a button next to Share that has an icon commonly associated with sharing or sending. We might have to rethink that one down the line.

website/static/css/pages/share-buttons.css
acshi: The "share" element is defined in filepage/index.js:53. Those css rules are in fact necessary unlike some of the others!
m('li.active', m('a[href="#share"][data-toggle="tab"]', 'Share')),

website/static/js/pages/project-dashboard-page.js
acshi: Verified that your changes are effective.

acshi:
@caneruguz 
- I actually never intended to make a popover. Perhaps you got my component (the ShareDropdown) mixed up with the pre-existing SharePopover that was already on the file page. If you think a popover would be preferable, I would be open to changing it, but I choose a dropdown because I wanted a simple menu of choices that would drop down for use and then disappear when it lost focus. I wouldn't want the sharing menu to stay on the screen if the user clicks elsewhere to do other things (which is how the SharePopover currently behaves).

- I am not an expert on UI design: I choose the squares because they looked like buttons to me. I've changed the style to make them look like yours now.

- I actually do provide the same information to LinkedIn as I do to Twitter (and the other options), however the LinkedIn dialog does not display it as a template. Both Twitter and Facebook try to visit the shared link and make a view of it to show the user. However, running on localhost the links cannot be reached by them, so we don't see the same thing as we would if this were being run on staging or production. In addition, Twitter would perform url shortening if the links were not to localhost.

sloria:
The GUID URLs don't seem to be displaying for files.

When I click the twitter button on the file detail page, it only shows the filename:

post_a_tweet_on_twitter

website/static/css/pages/share-buttons.css
sloria: Ah ok. Even so, we try to avoid using CSS IDs in selectors because they aren't reusable. Is it necessary to include `#share` in the selector?

website/static/css/pages/share-buttons.css
```diff
@@ -0,0 +1,35 @@
...
+
+#share .share-buttons {
+    display: inline-block;
+    margin-top: 12px;
+    margin-left: 4px;
+    margin-right: 4px;
+    vertical-align: top;
+}
+
+#shareButtonsPopoverContent {
```
sloria: Can you use a CSS class here instead?

sloria:
Pass finished.

Issue #18, opened 2013-05-31
add files link displayed on a public project, even if I am not a contributor


SamPortnow:
If you are not a contributor to a public project, you can still go to the project page and the add file link is still visible. However, if you follow that link and upload a file, an error message is produced:

 SyntaxError: JSON.parse: unexpected character

I would expect that the link to add a file would not even show up.

PR #5804, opened 2016-06-07
[Fix][Improvement][Migration] Use CAS for all authentications and refactor Auth logic [#OSF-6438]

cslzchen:
## Purpose

Submitting username and password from OSF directly bypass the CAS Login CSRF protection. This PR moves all authentication (with a few safe exceptions) to be handled directly by CAS.

UI co-contributer: @billyhunt

## Main Ticket

https://openscience.atlassian.net/browse/OSF-6438

## Related Ticket

https://openscience.atlassian.net/browse/SEC-26
https://openscience.atlassian.net/browse/OSF-6450
https://openscience.atlassian.net/browse/SEC-10 Part 3

## Related CAS and fakeCAS PR

https://github.com/CenterForOpenScience/cas-overlay/pull/18
https://github.com/CenterForOpenScience/fakecas/pull/4

### OSF Front-end

- Home Page
    - Replace drop-down sign in form with a button which redirects users to CAS login. "Forgot Your Password" and "Institution Login" links are moved to CAS as well.

home-page

- Signup Page

    - Remove redundant login from Signup Page. Add "Already have an account?" and "Login through your institution" to redirect users to CAS login.

signup-page

- Forgot Password Notification

    - Stay on `forgot_password.mako` with push notification instead of redirecting to Sign Up page

forgot-password-notification

- Resend Confirmation Notification

    - Whether unconfirmed email exists or not, use the same notification. This is consistent with Forgot Password.

resend-confirmation-notification

### OSF Back-end

- Refactor CAS `get_login_url()`:

cas-verification-key

- Claimed contributor will be redirected to the project page instead of profile page.

- Hide security implementation by using new helper function `generate_verification_key()` and increase key length.

- Refactor auth and contributor views + related views, routes, forms and tests.
    - Divide `auth_login` into `auth_login` and `auth_register`
    - Combine `forgot_password_get` and `forgot_password_put` into `forgot_password`
    - Write comprehensive unit test + security for forgot and reset password
    - Improve comments and TODO list
    - Remove obsolete views, routes, tests, forms and fix broken routes

### CAS Front-end

- CAS login page serves as the only place where user submits username and password to sign in.
    - Stayed Signed In (Persistent CAS TGC option: https://openscience.atlassian.net/browse/SEC-10)
    - Forgot Password
    - Institution Login
    - Back to OSF
    - Create Account

cas-login-page

### CAS Back-end

- Disable explicit auto login in `casLoginView.jsp` that previously causes CSRF vulnerability.

- Handle `(username, verification_key)` authentication internally, including:
    - forgot password reset
    - unregistered contributor set password
    - create account
    - merge user or add email
    - send and resend confirmation

### Other

- Forgot Password, Create Account, Set Password (Unregistered Contributor) are still handled by OSF. It is safe since we use a one-time verification key as our CSRF token.

- Institution Login & Prereg unchanged.

- Compatible with two-factor login

## Side effects

- This breaks current fakeCAS for local development. Here is the [fakeCAS-#4](https://github.com/CenterForOpenScience/fakecas/pull/4).

- This works with current CAS, but it is recommend to deploy this PR with [CAS-#18](https://github.com/CenterForOpenScience/cas-overlay/pull/18) together.

- For deployment of new CAS: add new OSF `/register` url to CAS
cas-verification-key


website/landing_pages.py
```diff
@@ -1,34 +1,20 @@
...
-    try:
-        data[0]['title_text'] = title
-        data[0]['content_template_path'] = content_path
-        data[0]['login_url'] = cas.get_login_url(redirect_to, auto=True)
-    except TypeError:
-        pass
-
-    return data
+@decorators.must_be_logged_in
+def _landing_page(auth, title, redirect_to, **kwargs):
```
icereval: lets kill off this one line helper

website/project/views/contributor.py
```diff
@@ -633,7 +633,6 @@ def claim_user_form(auth, **kwargs):
             # Redirect to CAS and authenticate the user with a verification key.
```
icereval: Please make a utility/helper out of line 627 `user.verification_key = security.random_string(20)` looks like this was already begun for confirm/claim tokens @ https://github.com/CenterForOpenScience/osf.io/blob/develop/framework/auth/core.py#L45

Searching the code base this shows up in a few locations:
https://github.com/CenterForOpenScience/osf.io/search?utf8=%E2%9C%93&q=security.random_string

@abought do you have any concern or recommendation on the length/complexity of the token generated.

icereval:
@cslzchen first pass complete, looks great!

@billyhunt thanks for your help on the UI consultations!

website/project/views/contributor.py
abought: There might be some length issues; see checklist below.

There are two basic rules for randomly generated tokens:

- All possibilities have to be equally likely, eg the (pseudo)RNG is good. This helper function uses `SystemRandom`, which is in line with what the [docs](https://docs.python.org/2/library/random.html#random.SystemRandom) recommend.

- The chances of guessing accepted credentials should be extremely small; this seems much shorter than the recommended limits accepted by security professionals elsewhere. As one metric for comparison, RFC6749 (for OAuth) [suggests](https://tools.ietf.org/html/rfc6749#section-10.10):
> The probability of an attacker guessing generated tokens (and other
   credentials not intended for handling by end-users) MUST be less than
   or equal to 2^(-128) and SHOULD be less than or equal to 2^(-160).

Some back of the envelope calculations: In this case, the character set of the function (letters + numbers) = 62 characters to choose from, so the number of possible combinations is something like `(62)**20 ~= (64)**20 = (2**6)**20 = (2**120)`- a 20-character string falls 256x short of the minimum randomness.

I'd recommend a longer string- 30 characters long at least, and often we use longer. For OAuth client secrets, for example, [we use](https://github.com/abought/osf.io/blob/6c26058b338712fb159b4993b85c74a9cb13aa20/website/oauth/models/__init__.py#L470-L470) a 40-character string, which is to say `~= (2**6)**40 = 2**240`. For access tokens, I think the string is even longer- 70 characters.

cslzchen: @icereval Looks like the helper is already there, I can just used `generate_confirm_token` and `generate_claim_token` for our purpose (with token length).

@abought 30-charactor is OK but `62^30 <= 64^32 = 2^160`. I am for 40-character string.

cslzchen: Just realized that the helper is for token. I do need new one for verification key. :) @icereval

abought: > Just realized that the helper is for token. I do need new one for verification key. :) @icereval

Also, it goes without saying that any confirmation token that shows up in a URL will show up in server logs, so of course this token should be one-time use only. But I think it already is.

(technically, `math.log(2**160, 62) = 26.87180463312671`. But I'm all for 40 characters too; anyone trying to guess credentials is Not Our Friend, so let's not pretend to be nice to them.)

framework/auth/views.py
```diff
@@ -113,21 +117,25 @@ def forgot_password_post():
...
-###############################################################################
 @collect_auth
 def auth_login(auth, **kwargs):
-    """If GET request, show login page. If POST, attempt to log user in if
-    login form passsed; else send forgot password email.
-
+    # TODO: auth_login is no longer the entry point for OSF login, need to refactor
+    """
+    This view serves several purposes:
+        GET request with '/login/':                         show sign up page for user to create account;
```
sloria: Is this accurate? Shouldn't `/login/` go to the login page?

sloria:
Discussed in person with @cslzchen : 

- `/login/` should go to the CAS login page.
- `/register/` should go to the sign up form.
- Campaigns should still be handled by the login route, e.g. `/login/?campaign=prereg`. 

These changes will require `login.mako` to be split into 2 templates (one for campaign logins, one for the sign up form).

sloria:
Pass finished.

framework/auth/cas.py
```diff
@@ -55,20 +55,15 @@ class CasClient(object):
...
         self.BASE_URL = base_url
 
-    def get_login_url(self, service_url, auto=False, username=None, password=None, verification_key=None, otp=None):
+    def get_login_url(self, service_url, username=None, verification_key=None):
         url = furl.furl(self.BASE_URL)
         url.path.segments.append('login')
         url.args['service'] = service_url
-        if auto:
+        if username and verification_key:
+            # TODO: remove auto=True from url when new CAS (PR#18) is deployed
```
cslzchen: Note for CR: `auto=True` works for both current CAS and CAS PR-#18.

framework/auth/views.py
```diff
@@ -51,88 +57,134 @@ def reset_password(auth, **kwargs):
...
+    if request.method == 'POST':
+        if form.validate():
+            # new random verification key, allows CAS to authenticate the user w/o password, one-time only.
+            # this overwrite also invalidates the verification key generated by forgot_password_post
+            user_obj.verification_key = generate_verification_key()
+            user_obj.set_password(form.password.data)
+            user_obj.save()
+            status.push_status_message('Password reset', kind='success', trust=False)
+            # redirect to CAS and authenticate the user with the one-time verification key.
+            # TODO: fix CAS that this verification key is destroyed once used
```
cslzchen: Note for CR: the last `verification_key` for CAS remains in mongo after user successfully resets password.

framework/auth/views.py
```diff
@@ -51,88 +57,134 @@ def reset_password(auth, **kwargs):
...
-        # Redirect to CAS and authenticate the user with a verification key.
-        return redirect(cas.get_login_url(
-            web_url_for('user_account', _absolute=True),
-            auto=True,
-            username=user_obj.username,
-            verification_key=user_obj.verification_key
-        ))
-
-    forms.push_errors_to_status(form.errors)
+    if request.method == 'GET':
```
cslzchen: Note for CR: explicit `if` flow for better understanding. Same for `forgot_password`.

framework/auth/views.py
```diff
@@ -51,88 +57,134 @@ def reset_password(auth, **kwargs):
...
     return {
-        'verification_key': verification_key,
+        'verification_key': verification_key
     }
 
 
-def forgot_password_post():
-    """Attempt to send user password reset or return respective error.
+@collect_auth
+def forgot_password(auth):
```
cslzchen: Note for CR: combine `forgot_password_get` and `forgot_password_post`.

framework/auth/views.py
```diff
@@ -51,88 +57,134 @@ def reset_password(auth, **kwargs):
...
 
 @collect_auth
-def forgot_password_get(auth, *args, **kwargs):
-    """Return forgot password page upon.
+def auth_login(auth):
     """
-    if auth.logged_in:
-        return redirect(web_url_for('dashboard'))
-    return {}
+    This view serves as the entry point for OSF login and campaign login.
```
cslzchen: Note for CR: this comment is a 'reverse-engineering' of how the following code work.

tests/test_views.py
```diff
@@ -3222,11 +3222,6 @@ def test_explore(self):
         res = self.app.get("/explore/").maybe_follow()
         assert_equal(res.status_code, 200)
 
-    def test_forgot_password_get(self):
```
cslzchen: This is moved `webtest_tests.py`.

website/templates/nav.mako
```diff
@@ -82,21 +82,8 @@
...
-                            <div class="form-group"><input name="password" id="password" class="form-control" type="password" placeholder="Password" data-bind="value: password" aria-label="Password"></div>
-                            <div class="form-group"><button type="submit" id="btnLogin" class="btn btn-block btn-primary">Login</button></div>
-                            %if enable_institutions:
-                                <div class="text-center m-b-sm"> <a href="/login/?campaign=institution">Login through your institution  <i class="fa fa-arrow-right"></i></a></div>
-                            %endif
-                            <div class="text-center m-b-sm"> <a href="/forgotpassword/">Forgot password?</a></div>
-                        </form>
-                    </ul>
+                    <a href="${web_url_for('auth_register')}" class="btn btn-success btn-top-signup m-r-xs">Sign Up</a>
+                    <a href="${login_url}" class="btn btn-info btn-top-login p-sm">Sign In</a>
```
cslzchen: Note for CR: I had planned to replace `${login_url}` with `${web_url_for('auth_login')` which serves as a entry point view. However, this breaks several redirections back from CAS. These redirections are calculated in `routes.py` as globals.

framework/auth/views.py
```diff
@@ -149,14 +191,12 @@ def auth_login(auth, **kwargs):
             raise HTTPError(http.InvalidURL)
 
     if auth.logged_in:
-        if not request.args.get('logout'):
+        if not log_out:
```
cslzchen: Note for CR: remove ambiguity from shadowing logout.

tests/webtest_tests.py
```diff
@@ -877,45 +848,168 @@ def test_newest_public_project_and_registrations_show_in_explore_activity(self):
...
+    #  logged-in user should be automatically logged out upon before reset password
     def test_reset_password_logs_out_user(self):
-        another_user = AuthUserFactory()
-        # visits reset password link while another user is logged in
-        res = self.app.get(self.url, auth=another_user.auth)
-        assert_equal(res.status_code, 200)
-        # We check if another_user is logged in by checking if
-        # their full name appears on the page (it should be in the navbar).
-        # Yes, this is brittle.
-        assert_not_in(another_user.fullname, res)
```
cslzchen: Note for CR: `assert_in('logout?service=', location)` and `assert_not_in('reauth', location)` guarantee that user will be logged out.

framework/auth/views.py
```diff
@@ -51,88 +54,126 @@ def reset_password(auth, **kwargs):
...
-            if throttle_period_expired(user_obj.email_last_sent, settings.SEND_EMAIL_THROTTLE):
-                user_obj.verification_key = security.random_string(20)
-                user_obj.email_last_sent = datetime.datetime.utcnow()
-                user_obj.save()
-                reset_link = furl.urljoin(
-                    settings.DOMAIN,
-                    web_url_for(
-                        'reset_password',
-                        verification_key=user_obj.verification_key
+    if request.method == 'GET':
```
sloria: This can be reoved.

cslzchen: Sure, I will remove all explicit `GET` checks which are redundant.

framework/auth/cas.py
icereval: Go ahead and remove the `auto` option, we will be releasing both at the same time

framework/auth/cas.py
```diff
@@ -173,6 +168,7 @@ def revoke_tokens(self, payload):
         else:
             self._handle_error(resp)
 
+
```
icereval: single line break for class methods (https://www.python.org/dev/peps/pep-0008/#blank-lines)

framework/auth/views.py
```diff
@@ -51,88 +54,126 @@ def reset_password(auth, **kwargs):
...
+                username=user_obj.username,
+                verification_key=user_obj.verification_key
+            ))
+        else:
+            forms.push_errors_to_status(form.errors)
+            # Don't go anywhere
+
     return {
-        'verification_key': verification_key,
+        'verification_key': verification_key
```
icereval: should have a trailing comma (https://www.python.org/dev/peps/pep-0008/#indentation - bottom)

framework/auth/views.py
icereval: Seems this should be a `return`?

framework/auth/cas.py
cslzchen: The following method `parse_auth_header()` does not belong to the class.

framework/auth/views.py
cslzchen: yes, need to be a `return {'verification_key': verification_key,}`.

framework/auth/views.py
cslzchen: Or we can have
 ```
if request.method == `GET`
    return {}
```

File: framework/auth/views.py [hunk 16]
```diff
@@ -51,88 +54,126 @@ def reset_password(auth, **kwargs):
...
-            auto=True,
-            username=user_obj.username,
-            verification_key=user_obj.verification_key
-        ))
-
-    forms.push_errors_to_status(form.errors)
-    return {
-        'verification_key': verification_key,
-    }
+    if request.method == 'GET':
```
icereval: can be consolidated with logic below

framework/auth/views.py
```diff
@@ -51,88 +54,126 @@ def reset_password(auth, **kwargs):
...
+            # redirect to CAS and authenticate the user with the one-time verification key.
+            return redirect(cas.get_login_url(
+                web_url_for('user_account', _absolute=True),
+                username=user_obj.username,
+                verification_key=user_obj.verification_key
+            ))
+        else:
+            forms.push_errors_to_status(form.errors)
+            # Don't go anywhere
+            return {
```
icereval: ^

framework/auth/views.py
icereval: If the intent is to leave the 'GET' method clear this is fine

framework/auth/views.py
```diff
@@ -175,33 +214,41 @@ def auth_login(auth, **kwargs):
...
     """
+
     redirect_url = redirect_url or request.args.get('redirect_url') or web_url_for('goodbye', _absolute=True)
+    # OSF log out, remove current OSF session
     logout()
-    if 'reauth' in request.args:
-        cas_endpoint = cas.get_login_url(redirect_url)
-    else:
+    # set redirection to CAS log out (or log in if 'reauth' is present)
+    if 'reauth' not in request.args:
```
icereval: the `not` can be a bit harder to follow

framework/auth/views.py
```diff
@@ -449,116 +515,35 @@ def register_user(**kwargs):
...
-                return {'form': form}
-            try:
-                send_confirm_email(user, clean_email)
-            except KeyError:  # already confirmed, redirect to dashboard
-                status_message = 'Email has already been confirmed.'
-                kind = 'warning'
-            else:
-                status_message = 'Resent email to {0}'.format(clean_email)
-                kind = 'success'
+            status_message = ('If there is an OSF account associated with this unconfirmed email {0}, '
```
icereval: TODO: add resend confirmation rate limit @brianjgeiger

tests/factories.py
```diff
@@ -827,7 +811,7 @@ def create_fake_user():
         fullname=name,
         is_registered=True,
         is_claimed=True,
-        verification_key=security.random_string(15),
+        verification_key= generate_verification_key(),
```
icereval: extra space

framework/auth/views.py
```diff
@@ -449,116 +515,35 @@ def register_user(**kwargs):
...
 def resend_confirmation():
-    """View for resending an email confirmation email.
     """
+    View for resending an email confirmation email.
+    HTTP Method: GET, POST
+    """
+
     form = ResendConfirmationForm(request.form)
+
+    if request.method == 'GET':
```
icereval: split view method into `get/post` specific methods

framework/auth/views.py
```diff
@@ -3,46 +3,49 @@
...
+
+    # If user is already logged in, log user out
     if auth.logged_in:
         return auth_logout(redirect_url=request.url)
-    verification_key = kwargs['verification_key']
+
     form = ResetPasswordForm(request.form)
 
+    # Check if request bears a valid verification_key
     user_obj = get_user(verification_key=verification_key)
```
icereval: @brianjgeiger, @cslzchen - we should require `verification_key` + `username` + `expiration` for better security, also matches existing functionality of the CAS `/login` endpoint.

website/project/views/contributor.py
```diff
@@ -593,54 +595,59 @@ def replace_unclaimed_user_with_registered(user):
...
-    if not verify_claim_token(user, token, pid):
-        return redirect(web_url_for('auth_login'))
     unclaimed_record = user.unclaimed_records[pid]
     user.fullname = unclaimed_record['name']
     user.update_guessed_names()
     # The email can be the original referrer email if no claimer email has been specified.
     claimer_email = unclaimed_record.get('claimer_email') or unclaimed_record.get('email')
     form = SetEmailAndPasswordForm(request.form, token=token)
+
+    if request.method == 'GET':
```
icereval: split view methods if possible

website/project/views/contributor.py
```diff
@@ -593,54 +595,59 @@ def replace_unclaimed_user_with_registered(user):
...
             # Clear unclaimed records
             user.unclaimed_records = {}
-            user.verification_key = security.random_string(20)
+            user.verification_key = generate_verification_key()
             user.save()
             # Authenticate user and redirect to project page
-            status.push_status_message(language.CLAIMED_CONTRIBUTOR,
-                                       kind='success',
-                                       trust=True)
+            status.push_status_message(language.CLAIMED_CONTRIBUTOR, kind='success', trust=True)
```
icereval: can `trust=False`?

website/templates/public/pages/comingsoon.mako
```diff
@@ -3,4 +3,3 @@
 <%def name="content()">
```
icereval: Can you double check this file is in the correct location.

icereval:
Second pass finished, a few notes on splitting up view methods by verb, looks good!

website/templates/public/pages/comingsoon.mako
cslzchen: It should be in `/public/` not `/public/pages`.

PR #5805, opened 2016-06-08
move clean function to cas from auth_login

billyhunt:
https://github.com/CenterForOpenScience/osf.io/pull/5000/

## Purpose

Ironically, auth_login is no longer used when logging in.  So cleaning up the old emails should happen in cas.py now.

## Ticket

### DIFF
--- a/framework/auth/cas.py
+++ b/framework/auth/cas.py
@@ -224,6 +224,7 @@ def make_response_from_ticket(ticket, service_url):
     cas_resp = client.service_validate(ticket, service_furl.url)
     if cas_resp.authenticated:
         user = User.load(cas_resp.user)
+        user.clean_email_verifications()
         # if we successfully authenticate and a verification key is present, invalidate it
         if user.verification_key:
             user.verification_key = None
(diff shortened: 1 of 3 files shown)


framework/auth/cas.py
```diff
@@ -224,6 +224,7 @@ def make_response_from_ticket(ticket, service_url):
     cas_resp = client.service_validate(ticket, service_furl.url)
     if cas_resp.authenticated:
         user = User.load(cas_resp.user)
+        user.clean_email_verifications()
```
sloria: This is ok here, but it would be good to group all of the "extra-stuff-to-do-when-a-user-logs-in" code in one place. I just found that the code to auto-affiliate a user based on email address is in the `authenticate` function: https://github.com/CenterForOpenScience/osf.io/blob/1311018158a31edcbb65b9314f1a4f52b1365d37/framework/auth/__init__.py#L47 . Let's put this code there.

In the future it'd be good to have a signal that we could subscribe, but that can be a TODO.

PR #5806, opened 2016-06-08
API File tags fixes

samchrisinger:
replaces https://github.com/CenterForOpenScience/osf.io/pull/5704/


sloria:
Added the index discussed in https://github.com/CenterForOpenScience/osf.io/pull/5704#discussion_r64812844 when I merged.

PR #5807, opened 2016-06-08
[Feature] add withdrawn before retracted registration files in search [OSF-6456]

leb2dg:
## Purpose

Currently, the user cannot tell if a file is attached to a retracted registration in search results. This PR adds "Withdrawn" before "Registration File" so that the results look similar to retracted registrations.

screen shot 2016-06-06 at 9 53 35 am

## Changes

* Add `is_retracted` to elastic search file_doc
* Add "Withdrawn" text in search results if the file is attached to a retracted node

screen shot 2016-06-08 at 11 47 46 am

## Side effects

I had to run `inv rebuild_search` on my local but I'm not sure what that means for production.

## Ticket

https://openscience.atlassian.net/browse/OSF-6456

### DIFF
--- a/website/search/elastic_search.py
+++ b/website/search/elastic_search.py
@@ -466,6 +466,7 @@ def update_file(file_, index=None, delete=False):
         'node_title': file_.node.title,
         'parent_id': file_.node.parent_node._id if file_.node.parent_node else None,
         'is_registration': file_.node.is_registration,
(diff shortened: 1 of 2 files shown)


brianjgeiger:
Looks great. 🎉 🐧 🎉 I'm reasonably certain the failing test is entirely unrelated to this.

PR #5809, opened 2016-06-08
[RFR] Make Postcommit tasks marked as celery run as a chain

cwisecarver:
## Purpose

Currently, postcommit tasks could run in any order, asynchronously. This changes them to run as a celery chain: synchronously and in order.

## Changes

There is a separate queue for celery specific postcommit tasks. This queue is run as a celery chain after the request's transaction is committed successfully.

## Side effects

Tasks decorated with run_postcommit could be delayed to the point where a user would notice.

### DIFF
--- a/framework/postcommit_tasks/handlers.py
+++ b/framework/postcommit_tasks/handlers.py
@@ -7,7 +7,6 @@ import binascii
 from collections import OrderedDict
 import os
 
-
 from celery.local import PromiseProxy
 from gevent.pool import Pool
 
--- a/framework/postcommit_tasks/handlers.py
+++ b/framework/postcommit_tasks/handlers.py
@@ -69,7 +69,7 @@ def run_postcommit(once_per_request=True, celery=False):
             return func
         @functools.wraps(func)
         def wrapped(*args, **kwargs):
-            if celery is True and isinstance(func, PromiseProxy):
+            if settings.USE_CELERY and celery is True and isinstance(func, PromiseProxy):
                 func.delay(*args, **kwargs)
             else:
                 enqueue_postcommit_task(func, args, kwargs, once_per_request=once_per_request)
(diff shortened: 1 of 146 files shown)


framework/postcommit_tasks/handlers.py
```diff
@@ -70,9 +95,6 @@ def wrapper(func):
             return func
         @functools.wraps(func)
         def wrapped(*args, **kwargs):
-            if celery is True and isinstance(func, PromiseProxy):
-                func.delay(*args, **kwargs)
-            else:
-                enqueue_postcommit_task(func, args, kwargs, once_per_request=once_per_request)
+            enqueue_postcommit_task(func, args, kwargs, celery=celery, once_per_request=once_per_request)
```
brianjgeiger: 👍

framework/postcommit_tasks/handlers.py
```diff
@@ -35,12 +48,20 @@ def postcommit_after_request(response, base_status_error_code=500):
             for func in postcommit_queue().values():
                 pool.spawn(func)
             pool.join(timeout=5.0, raise_error=True)  # 5 second timeout and reraise exceptions
+
+        if postcommit_celery_queue():
+            if settings.USE_CELERY:
+                postcommit_celery_task_wrapper.delay(postcommit_celery_queue())
```
cwisecarver: This is the top-secret bad way to get around the maximum recursion bug. Make a task that calls all the other tasks as a chain and call that, passing in the signatures of the other tasks. I read about it on the internet.

brianjgeiger: Is it worth linking to the source in a code comment for people who later want to know about this?

cwisecarver: dun

PR #5810, opened 2016-06-08
Have /help/ display the /faq/ page [#OSF-4039]

DanielSBrown:
## Purpose

- As of now, the osf.io/help/ page has not been made, and is a dead end

- Displays the /faq/ page on attempting to reach the /help/ page

## Changes

- Added to a Rule in routes.py, where going to /help/ will display the /faq/ mako template.

- Added a test case that tests to make sure the page is reached in test_views.py

## Ticket
https://openscience.atlassian.net/browse/OSF-4039

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -4366,6 +4366,9 @@ class TestStaticFileViews(OsfTestCase):
         res = self.app.get('/getting-started/')
         assert_equal(res.status_code, 302)
         assert_equal(res.location, 'http://help.osf.io/')
+    def test_help_redirect(self):
+        res = self.app.get('/help/')
+        assert_equal(res.status_code,200)
 
 
 class TestUserConfirmSignal(OsfTestCase):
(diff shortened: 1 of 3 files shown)


DanielSBrown:
I am not sure if the test case is comprehensive.  I compared the HTTP status code with 200, which may not be enough.  Please let me know if I should add other assert statements/change the status code check.

brianjgeiger:
This is good. 🎉 🐧 🎉

sloria:
I'm thinking we should redirect `/help/` to `/faq/`, rather than having `/help/` return a 200. Having multiple URLs point to the same resource is an antipattern.

Also, `/help/` should be removed from the old Rule (https://github.com/CenterForOpenScience/osf.io/blob/35514cbb6112e64ecf676213db21f6423c70ccd5/website/routes.py#L255).

sloria:
Pass finished.

PR #5816, opened 2016-06-09
Feature/Google Analytics Audit [OSF-6479]

pattisdr:
## Purpose

I did an audit of the google analytics events in the Dashboard and My Projects pages.  There are a couple items that are triggering firing of two events.

1) There is an 'x' (close) button as well as a 'Cancel' button on the Add Collection modal on the My Projects page.   When the close button is clicked, the 'cancel' button event tracking was also fired.

2) There are a few items we track in the mobile view only.  If you click on the button with a horizontal ellipsis beside a project in the mobile view to open up the information panel, the event tracking for a single project being selected is also fired. 

## Changes

Corrected firing of two events for these two actions.

## Side effects

Before the date that this gets merged in, event tracking will be incorrect. After this date, event tracking for these two items should be fixed.  If analyzing data before and after this change, corrections will need to be made.

These are minor items and easy to fix when analyzing data - when I did my GA presentation in April, I manually corrected this data for these items.

## Ticket

https://openscience.atlassian.net/browse/OSF-6479

### DIFF
--- a/website/static/js/myProjects.js
+++ b/website/static/js/myProjects.js
@@ -1114,7 +1114,6 @@ var Collections = {
             self.newCollectionName('');
             self.isValid(false);
             $('#addCollInput').val('');
-            $osf.trackClick('myProjects', 'add-collection', 'click-cancel-button');
         },
         self.deleteCollection = function _deleteCollection(){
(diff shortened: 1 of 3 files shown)


mfraezz:
LGTM. Dropping this in RTM 🌟

PR #5819, opened 2016-06-09
[Feature] fix registration preview style on narrow window [OSF-6464]

leb2dg:
## Purpose

If the window is narrow when looking at a preview of a registration, the formatting is strange.
screen shot 2016-06-09 at 3 00 57 pm

## Changes

* switch from col-md to col-xs

screen shot 2016-06-09 at 3 04 24 pm

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-6464


brianjgeiger:
Looks good. 🎉 🐧 🎉

PR #5824, opened 2016-06-10
Concatenate contributor ids

hmoco:
## Purpose
Contributors don't have guids, so it used the id of the user, beign unique for user per node. However that becomes a problem if you look for the ocntributor data universally, as missing the node information will lead to duplicate results (problems encountered in ember-osf). New ids for contributors are unique, and are concatenations of user and node ids.

## Changes
Contributor ids are now a concatenation of node id and user id, separated by `-`

## Ticket

https://openscience.atlassian.net/browse/OSF-6490


hmoco:
test broken, working on fixes.

samchrisinger:
:+1: needs review

sloria:
While I think using `<nodeid>-<userid>` is fine (unless @brianjgeiger objects), this makes the `id` field read-only, when it was previously a required field.

Rather than using a SerializerMethodField, it might make more sense to add a custom field to `api.nodes.serializers`, e.g. `ContributorIDField`, that can validate input from the client.

sloria:
Pass finished.

hmoco:
@sloria, id is still a required field that validates user input on creation, and on the detail serializer for update/delete (https://github.com/CenterForOpenScience/osf.io/pull/5824/files#diff-7a6d21a50624153d15e63a5f643401c7R383). Do you still feel it needs a custom field?

sloria:
@hmoco Does the validation work correctly for updates/creates/deletes? If you're using `IDField(source='_id', required=True)`, I'm not sure how it would know to check against `<nodeid>-<userid>`.

hmoco:
It doesnt look at the id, it checks for the user relationship. Seems to be working right

sloria:
@hmoco I'm not seeing the correct behavior.

I added the following test to `TestNodeContributorUpdate`, and it fails:

```python
    def test_change_contributor_correct_id(self):
        contrib_id = '{}-{}'.format(self.project._id, self.user_two._id)
        data = {
            'data': {
                'id': contrib_id,
                'type': 'contributors',
                'attributes': {
                    'permission': permissions.ADMIN,
                    'bibliographic': True
                }
            }
        }
        res = self.app.put_json_api(self.url_contributor, data, auth=self.user.auth, expect_errors=True)
```

hmoco:
For the purposes of documenting conversation, here is how the api docs describe the endpoint's payload expectation:
```
Adding Contributors

Method:        POST
URL:           /links/self
Query Params:  <none>
Body (JSON): {
              "data": {
                "type": "contributors",                   # required
                "attributes": {
                  "bibliographic": true|false,            # optional
                  "permission": "read"|"write"|"admin"    # optional
                },
                "relationships": {
                  "users": {
                    "data": {
                      "type": "users",                    # required
                      "id":   "{user_id}"                 # required
                    }
                }
            }
        }
    }
Success:       201 CREATED + node contributor representation
Add a contributor to a node by issuing a POST request to this endpoint. This effectively creates a relationship between the node and the user. Besides the top-level type, there are optional "attributes" which describe the relationship between the node and the user. bibliographic is a boolean and defaults to true. permission must be a valid OSF permission key and defaults to "write". A relationship object with a "data" member, containing the user type and user id must be included. The id must be a valid user id. All other fields not listed above will be ignored. If the request is successful the API will return a 201 response with the representation of the new node contributor in the body. For the new node contributor's canonical URL, see the /links/self field of the response.
```

hmoco:
Nevermind ^ it seems that is only for creating and not updaitng

hmoco:
Ok, so this is still a blocker for ember-osf. Most of the work is done, I believe, so it just needs to be pushed through code review.

hmoco:
Reason for the problem is defined in the PR description, if @brianjgeiger could assign someone to push through code review, I am available to answer questions about desired behavior

sloria:
@hmoco Was https://github.com/CenterForOpenScience/osf.io/pull/5824#issuecomment-225993504 ever addressed?

sloria:
Discussed with Henrique in person: Since Henrique has other priorities at the moment, he has moved off of this PR for the time being.

I'll take a look at a solution for the above issue.

sloria:
Closing in favor of #5920

PR #5827, opened 2016-06-10
github usernames appearing in View Only links [#OSF-6449]

erinspace:
## Purpose
Anonymous view only links that were connected to a github repo still showed their github username as a part of the connected repo storage heading - since usernames can lead to identification, they should be hidden if the files are being shown as part of a view only link. 

In fact, to anonymize a little more, no part of the file path should be shown at all on anonymous links - only the addon heading. 

## Changes
- Add private_key kwarg that can be passed when building the addon root name for the hgrid file tree
- if that private_key kwarg is passed, and checks out as anonymous, remove the file path or folder or bucket name from the name of the heading for all addons
- refactor ```is_anonymized``` in ```api.base.serializers``` to facilitate reuse elsewhere
- use new ```check_private_key_for_anonymized_link``` to check if passed view_only private key is anonymous or not
- Add rubeus tests to make sure path/bucket is removed for anon but not all private links

In short - anonymized view_only links should no longer show the exact path as part of the storage heading. Normal view_only links should stay the same and still show the path and username as appropriate.

### before -- anonymous view_only link
screen shot 2016-06-10 at 1 32 34 pm

### after -- anonymous view_only link
screen shot 2016-06-13 at 1 42 56 pm

Non-anonymous view only links, and projects should **still** show the full filepath and usernames where passed.

## Side effects
None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-6449

### DIFF
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -15,11 +15,8 @@ from api.base.exceptions import InvalidQueryStringError, Conflict, JSONAPIExcept
 from api.base.settings import BULK_SETTINGS
 from api.base.utils import extend_querystring_params
 from framework.auth import core as auth_core
-from modularodm import Q
-from modularodm.exceptions import NoResultsFound
 from website import settings
 from website import util as website_utils
-from website.models import PrivateLink
 from website.util.sanitize import strip_html
 
 
@@ -52,16 +49,8 @@ def format_relationship_links(related_link=None, self_link=None, rel_meta=None, 
 
 
 def is_anonymized(request):
-    is_anonymous = False
     private_key = request.query_params.get('view_only', None)
-    if private_key is not None:
-        try:
-            link = PrivateLink.find_one(Q('key', 'eq', private_key))
-        except NoResultsFound:
-            link = None
-        if link is not None:
-            is_anonymous = link.anonymous
-    return is_anonymous
+    return website_utils.check_private_key_for_anonymized_link(private_key)
 
 
 class HideIfRegistration(ser.Field):
--- a/website/util/rubeus.py
+++ b/website/util/rubeus.py
@@ -74,12 +74,10 @@ def build_addon_root(node_settings, name, permissions=None,
     from website.util import check_private_key_for_anonymized_link
 
     permissions = permissions or DEFAULT_PERMISSIONS
-    if name:
+    if name and not check_private_key_for_anonymized_link(private_key):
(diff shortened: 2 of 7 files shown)


brianjgeiger:
The suggestion I had made in the comments from https://openscience.atlassian.net/browse/OSF-6449?focusedCommentId=33923&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-33923 was to eliminate showing the name of the top level item altogether, because Anon VOL people don't need the info and it could be leaky. For example, your dropbox hints that you are a Center for Open Science person. I would just, for AVOL links, take off everything from the colon on. So non anon dropbox would be `Dropbox: /COS/OSF Demo add on` and AVOL would be `Dropbox`.

erinspace:
@brianjgeiger ooooooooooh I see. Sorry, I had read that but then misunderstood - by top level I thought you meant the ability to remove the username in a generic way. I get it now. Thanks!

brianjgeiger:
Gotcha. No problem, it was vague and I was hoping to hear back from QA on it, so it never made it into the main description with more details. Hopefully this only simplifies things.

brianjgeiger:
Or, at least, not be any more complicated than what you have.

website/util/rubeus.py
```diff
@@ -66,12 +66,15 @@ def build_addon_root(node_settings, name, permissions=None,
...
+    :param bool private_key: Used to check if information should be stripped from anonymous links
     :param dict kwargs: Any additional information to add to the root folder
     :return dict: Hgrid formatted dictionary for the addon root folder
 
     """
+    from website.util import check_private_key_for_anonymized_link
+
     permissions = permissions or DEFAULT_PERMISSIONS
-    if name:
+    if name and not check_private_key_for_anonymized_link(private_key):
```
brianjgeiger: 👍

brianjgeiger:
This looks great! 🎉 🐧 🎉 Please update the "after" screenshot and the description of that changes to match current functionality.

erinspace:
@brianjgeiger oooh good call. Descriptions and screenshots updated!

api/base/serializers.py
```diff
@@ -52,16 +49,8 @@ def format_relationship_links(related_link=None, self_link=None, rel_meta=None,
 
 
 def is_anonymized(request):
-    is_anonymous = False
```
sloria: 👍

PR #5829, opened 2016-06-11
Add "Views" to OSF for Meetings Conference Pages [#OSF-4999]

chennan47:
## Purpose

Change the meeting page's project download count to project view and downloads count.

## Changes

Now the meeting's page's download counts on each post and talk should include not only the download number of the file but also the file page visit count.

## Side effects

This pr relies on keen to get the data of the page view count of the file page.

## Ticket

https://openscience.atlassian.net/browse/OSF-4999


scripts/meeting_visit_count.py
```diff
@@ -0,0 +1,58 @@
...
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+
+def main(dry_run=True):
+    init_app(routes=False)
+    for conf in Conference.find():
+        count = 0
+        for tag in Tag.find(Q('lower', 'eq', conf.endpoint.lower())):
+            for node in tag.node__tagged.find(Q('is_public', 'eq', True) & Q('is_deleted', 'eq', False)):
```
mfraezz: The `tagged` backref has been removed, use `Node.find(<query> & Q('tags', 'eq', tag))`

website/conferences/views.py
```diff
@@ -158,7 +158,7 @@ def _render_conference_node(node, idx, conf):
         )
     except StopIteration:
         download_url = ''
-        download_count = 0
+        view_and_download = record.visit
```
mfraezz: `record` is out of scope in this `except` block

mfraezz: This is the main reason the build is failing, but there's also an API test that doesn't expect `visit` to be returned in `.to_storage`

scripts/meeting_visit_count.py
```diff
@@ -0,0 +1,56 @@
...
+
+                    resp = requests.get(url)
+                    record.visit = resp.json()['result']
+                    record.save()
+                    count += 1
+
+        logger.info('Get visit counts for {} projects in Conference {}'.format(count, conf.name))
+    
+
+@celery_app.task(name='scripts.meeting_visit_count')
```
mfraezz: I'm not sure that this script needs to be a celerybeat job. If it were a one-shot migration, you could set each `record.visit` to be `<keen result> + record.get_download_count()`, and increment `.visit` each time a file is viewed or downloaded. There are benefits and drawbacks to each approach further discussion may be required.

Alternatively, @felliott indicates that it might be possible to do batch requests on the front-end to populate this column when the page loads, which would obviate storing `visit` on the model. This would probably be the best option. Regardless of implementation, however, this PR is blocked until the hold on Keen is removed.

chennan47: Talk to @felliott, it's still to slow to do batch requests to keen on pageload. Each one will take a few seconds to response.

chennan47:
Changes have been made. Still need to wait for @felliott 's pr to add in the public read_key for keen in website.settings.default

scripts/meeting_visit_count.py
```diff
@@ -0,0 +1,56 @@
...
+                        Q('node', 'eq', node) &
+                        Q('is_file', 'eq', True)
+                    ).limit(1)
+                ).wrapped()
+
+                if not dry_run:
+                    url = 'https://api.keen.io/3.0/projects/{}/events/meeting_view' \
+                          '?api_key={}&event_collection=pageviews&timezone=UTC&timeframe=this_14_days' \
+                          '&filters=%5B%7B%22property_name%22%3A%22page.info.path%22%2C%22' \
+                          'operator%22%3A%22eq%22%2C%22property_value%22%3A%22{}%22%7D%5D'.format(
```
felliott: Minor: the `filters` query argument is just url-encoded json.  This will be a lot clearer if you declare the structure, jsonify it, then url-encode it.

PR #5834, opened 2016-06-13
Unregistered contributor email should have link to project that they are invited to [OSF-3211]

Johnetordoff:
## Purpose

To allow a user visit a project they were just invited to, before registering.

## Changes

modified email templates and send parameters, add comments to template explain template function

## Side effects

none that I know of.

## Ticket

https://openscience.atlassian.net/browse/OSF-3211

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -2366,7 +2366,7 @@ class TestUserInviteViews(OsfTestCase):
             auth=Auth(project.creator),
         )
         project.save()
-        send_claim_email(email=given_email, user=unreg_user, node=project)
+        send_claim_email(email=given_email, invitedUser=unreg_user, node=project)
 
         assert_true(send_mail.called)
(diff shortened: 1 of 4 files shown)


mfraezz:
It seems like a lot of these changes are either out of scope for this ticket, in violation of PEP8 naming conventions, or break some of the generalized mailing behavior. Discussion on the ticket seems to imply that the only change necessary for this is an update to the template, to include a link to the node (which should already be accessible to the template) and a note that the recipient will only be able to view it before confirming their account if the node is public.

Please resubmit this without all the unnecessary changes. If you're not sure which changes are important, feel free to ask me, either with an `@tag` or via Flowdock.

Pass complete. Back to you, @Johnetordoff :bird:

Johnetordoff:
@mfraezz I hope this is better, https://github.com/CenterForOpenScience/osf.io/pull/5868

PR #5839, opened 2016-06-14
Improvement/wiki-spell-check   System-level spell checking on wiki editing [OSF-2197]

SSJohns:
## Purpose

Adds spell checking to the wiki page so that users can get free inline spell checking. Does not currently include right click suggestions, but those can be added in a later ticket. 

## Changes

Added the libraries [spellcheck_ace](https://github.com/swenson/ace_spell_check_js) and [typo.js](https://github.com/cfinke/Typo.js/blob/master/typo/typo.js) which combine to check for misspelled words and add the red squiggle(scientific term) to the bottom of those words which are misspelled. Updated webpack to allow for the typo file to be aliased and added dictionaries that would be parsed by the added libraries in the files. Some of the ace.js files are also updated to make interacting with the typo libraries more streamlined. Most things already implemented should work the same.

### Look of Wiki edited page
screen shot 2016-06-14 at 11 56 35 am

### Additional button for toggling spellcheck
Added a button that lets the user turn spellcheck on and off. This works only for that specific page by storing the value in a cookie that is grabbed when the user goes to that page. The idea is that the user has certain projects that they will want to turn spell check off for and other that they will want it to be on for and this difference will determine how much they make use of spellcheck. The new button is right next to the B and I on the toggle bar.
screen shot 2016-07-19 at 11 57 19 am

## Side effects

Checked other editors and they do not seem to be affected by this change, but there is a chance that the editor in other parts of the webpage may be affected.

The added dictionaries are un-parseable as they are  of unique file types. This might bring up problems with the vendor and sourcemaps, however as I've seen it it only produces an error on the wiki page and not much else. The problem seems to be a problem with my node configuration, but it can be ignored now.

## Ticket

https://openscience.atlassian.net/browse/OSF-2197


SSJohns:
I'm making a ticket for improvements so that someone can later implement the right-click and word suggestions for misspelled words.

brianjgeiger:
It is unfortunate that neither typo nor spellcheck_ace are bower/npm installable.

brianjgeiger:
Request for more info on the Ace Spellchecking landscape on the jira ticket. Pass done. 🐧 

(If we do keep this solution, add licensing information into the NOTICE at the root of the repo)

SSJohns:
Hey, can I ask are you requesting more information from me about the Ace Spellchecking or from Product?

brianjgeiger:
From you in this case. Since you chose some external libraries, I want to get an idea of the process that went into that and to ensure you were thorough and thinking about the right things when choosing.

website/addons/wiki/static/spellcheck_ace.js
```diff
@@ -0,0 +1,102 @@
+// You also need to load in typo.js and jquery.js
```
sloria: It's fine to have these 3rd party modules in our repo since they're not on npm or bower, but any 3rd-party code should go in `static/vendor/`. Wouldn't hurt to add a folder for ace plugins.

sloria:
Pass finished. Behavior looks good. Did you make any modifications to the vendorized modules? If so, please make that clear using code comments. Also, move the vendorized code to `static/vendor`.

sloria:
Note: I am going to hold on merging this until my question on [the ticket](https://openscience.atlassian.net/browse/OSF-2197?focusedCommentId=35300&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-35300) re: disabling spell-check is addressed. 

@SSJohns Go ahead and make the above changes in the meantime.

sloria:
Also, would it be easy to make the red underlines dotted, to make them look more like OS-level spell-checkers? 

improvement_wiki-spell-check_system-level_spell_checking_on_wiki_editing__osf-2197__by_ssjohns_ _pull_request__5839_ _centerforopenscience_osf_io

brianjgeiger:
Looks good. Per [Steve's comment](https://github.com/CenterForOpenScience/osf.io/pull/5839#issuecomment-229958788) about clearly labelling any changes to the original code, did you make any changes and, if so, did you clearly label the changes as things that we did? Pass done. If comments were made appropriately, then this is good. 🎉 🐧 🎉

SSJohns:
@brianjgeiger I have added more comments on things I changed. All in all my changes are all marked in the code.

package.json
```diff
@@ -58,6 +58,7 @@
     "raven": "^0.7.2",
     "raw-loader": "^0.5.1",
     "share": "0.7.27",
+    "share-js": "0.0.3",
```
SSJohns: Hey, I saw that this was trying to add share-js. Did someone take out share-js or should that still be there? I'm not sure. @brianjgeiger @sloria

PR #5518, opened 2016-04-21
[feature] @ mentions and notifications [OSF-5975][OSF-5658]

leb2dg:
Contains #5452 
Replacing #5376 

## Purpose
Add functionality for @/+ mention of contributors in comments. Only confirmed contributors, both bibliographic and non-bibliographic, can be mentioned.

Related tickets: https://openscience.atlassian.net/browse/OSF-5658 & https://openscience.atlassian.net/browse/OSF-5975

## Changes
* @ symbol or + triggers dropdown where user can select the full name of a contributor to mention them in a comment
* At mention names appear as a link to the mentioned contributor
* Email notifications are sent to contributors who are mentioned but are grouped if user would receive multiple for one event i.e. new comment and mention
* An error message is shown and the submit button is disabled if a comment is over 500 characters but the validator on the model allows up to 1000 characters
    * padding is added to the model validator to account for html elements that are converted to markdown before they are saved

Mentioning a contributor:
screen shot 2016-04-19 at 2 05 06 pm

Entering a comment longer than 500 characters:
screen shot 2016-04-19 at 2 04 36 pm

## Possible additions
* Emojis PR #5418 

## Side Effects
* change comments template from textareas to ContentEditable divs 
    * support keeping track of users who have been mentioned
    * does not allow partial deletion of the mention
    * can cause some unexpected behavior with pasting/arrow keys/enter
* add dependency At.js
    * provide @/+ functionality
* add new_mentions & old_mentions to the comment model
    * support determining who is a new mention after editing so contributors are not notified multiple times for the same comment
   * new_mentions is a required field so it must be passed as an attribute or the request will fail
* add new notification types and email templates
    * mention notifications can occur on creation or on edit while comments only notify subscribers on creation
   * comments and mentions have different notification settings
   * currently the mentions are all of the 'email_transactional' notification_type
* a user could submit a comment with up to 1000 characters using the API

PR #5840, opened 2016-06-14
[feature] @ mentions and notifications [OSF-5975][OSF-5658]

leb2dg:
Cleaner version of https://github.com/CenterForOpenScience/osf.io/pull/5518

## Purpose

Related tickets: https://openscience.atlassian.net/browse/OSF-5658 & https://openscience.atlassian.net/browse/OSF-5975

## Changes
* @ mention names appear as a link to the mentioned contributor
* An error message is shown and the submit button is disabled if a comment is over 500 characters
* add dependency At.js
    * provide @/+ functionality
* add ever_mentioned to the comment model

Mentioning a contributor:

## Possible additions
* Emojis PR #5418 (this is being updated to reflect changes on this PR)

## Side Effects
* ContentEditable divs    
* a user can write in a different name for the mention if they really understand the code


whit537:
##### Investigations—for @whit537

- [x] run code
- [x] skim code
- [x] see about PR for At.js fork— dce5c22d3e42fc7393ee65a17b93c414dfab33c4 
- [x] python
 - [x] signal wiring
 - [x] emailaroo
 - [x] sup with `new_mentions` and `old_mentions`? what's the role of `new_mentions` in the api?
 - [x] 500 + 500
- [ ] javascript
 - [ ] look over `onPaste` and `onReturn`
 - [x] `foo\nbar` is yielding `foo\nbar` instead of `foo<br>bar`
- [ ] check on noncontributors
- [ ] review tests
- [ ] understand options in 991abf4e8470272dc2e8edc245ab236ac8270975
- [ ] rereview mentions processing post- 89a59eef5407791337cdf5642abda221784c2296

### Files

#### Unprocessed

##### Backend

api/comments/serializers.py
api/comments/views.py
api/nodes/views.py
website/notifications/constants.py
website/notifications/emails.py
website/project/model.py
website/project/views/comment.py
website/settings/defaults.py

##### Frontend

website/static/js/comment.js

##### Tests

api_tests/comments/views/test_comment_detail.py
api_tests/nodes/views/test_node_comments_list.py
tests/factories.py
tests/test_comments.py
tests/test_notifications.py

#### Processed

bower.json
website/project/signals.py
website/signals.py
website/static/css/commentpane.css
website/static/js/commentpane.js
website/static/js/koHelpers.js
website/templates/emails/mentions.html.mako
website/templates/emails/mentions.txt.mako
website/templates/include/comment_pane_template.mako
website/templates/include/comment_template.mako

whit537:
##### Punchlist—for @leb2dg 

- [ ] test conversion functions—https://github.com/leb2dg/osf.io/pull/14#discussion_r66808779
- [ ]  DRY up `new_mentions` processing—https://github.com/CenterForOpenScience/osf.io/pull/5518#discussion_r60930682 —mostly done in 89a59eef5407791337cdf5642abda221784c2296
 - [x] don't store `new_mentions` on comments—https://github.com/leb2dg/osf.io/pull/14#discussion_r66820552 —done in 5bdfe2be07e100d94a4f3775b6a0364876aca983
 - [ ] remove test fixture—https://github.com/leb2dg/osf.io/pull/14/files#r66844879
- [ ] fix bug in edit processing—https://github.com/leb2dg/osf.io/pull/14#discussion_r66862289
- [ ] comment `+ 1`—https://github.com/leb2dg/osf.io/pull/14#discussion_r61802712
- [ ] standardize validation between client and server—https://github.com/leb2dg/osf.io/pull/14#discussion_r61798316
- [ ] prevent HTML via shortcut—https://github.com/leb2dg/osf.io/pull/14#discussion_r66867773
- [ ] explain `initialValue` reset—https://github.com/leb2dg/osf.io/pull/14#discussion_r61803112

##### Punchlist—done

website/static/js/tests/comment.test.js
```diff
@@ -0,0 +1,133 @@
+/*global describe, it, expect, example, before, after, beforeEach, afterEach, mocha, sinon*/
```
whit537: What does this do?

whit537:
Gosh. Doesn't it feel good to [have tests](https://github.com/CenterForOpenScience/osf.io/commit/913db15a6da64379504226baf483fb019aef673c)? ☺️

website/static/js/tests/comment.test.js
leb2dg: It tells [JSHint](http://jshint.com/docs/#inline-configuration) that those variables are global and defined elsewhere.

website/notifications/emails.py
```diff
@@ -24,24 +24,35 @@ def notify(event, user, node, timestamp, **context):
...
+                        sent_users.append(target_user_id)
+                if subscriptions[notification_type]:
+                    store_emails(subscriptions[notification_type], notification_type, event_type, user, node,
                                  timestamp, **context)
-                    sent_users.append(target_user_id)
-            if subscriptions[notification_type]:
-                store_emails(subscriptions[notification_type], notification_type, event_type, user, node,
-                             timestamp, **context)
-                sent_users.extend(subscriptions[notification_type])
+                    sent_users.extend(subscriptions[notification_type])
```
whit537: This is a really daunting function. There has to be a better way to factor it.

website/project/model.py
```diff
@@ -321,6 +350,13 @@ def create(cls, auth, **kwargs):
             save=False,
         )
 
+        new_mentions = get_valid_mentioned_users_guids(comment, comment.node.contributors)
+
+        if new_mentions:
+            project_signals.mention_added.send(comment, new_mentions=new_mentions, auth=auth)
+            comment.ever_mentioned.extend(new_mentions)
+            comment.save()
+
```
whit537: Can we rework this to only save once? We should be able to make it look a lot like the `edit` case, no?

whit537: (That is, I'm seeing another `comment.save()` on line 344.)

website/static/js/notificationsTreebeard.js
```diff
@@ -155,6 +155,7 @@ function ProjectNotifications(data) {
                 });
             }
             else if (item.parent().data.kind === 'folder' || item.parent().data.kind === 'heading' && item.data.kind === 'event') {
+                var mentionsInTitle = ~item.data.event.title.indexOf('mentions');
```
whit537: Presumably `item` is the same object both here and inside the anonymous function assigned to `custom` below on line 174, but it's probably best to use the name as bound down there rather than up here. Let's move this line down into that function.

website/static/js/notificationsTreebeard.js
```diff
@@ -177,9 +178,18 @@ function ProjectNotifications(data) {
...
+                                        m('option', {value: 'none', selected : item.data.event.notificationType === 'none' ? 'selected': ''}, 'Never'),
+                                    mentionsInTitle ?
+                                        null :
+                                        m('option', {value: 'email_transactional', selected : item.data.event.notificationType === 'email_transactional' ? 'selected': ''}, 'Instantly'),
+                                    mentionsInTitle ?
+                                        null :
+                                        m('option', {value: 'email_digest', selected : item.data.event.notificationType === 'email_digest' ? 'selected': ''}, 'Daily'),
+                                    mentionsInTitle ?
+                                        m('option', {value: 'email_transactional', selected: 'email_transactional', disabled: true}, 'Instantly') :
+                                        null,
```
whit537: [Thinking](https://github.com/leb2dg/osf.io/pull/14#discussion_r65965721) about this more ... we can save ourselves some long lines and I think make this easier to follow if we factor out the array of options for mention notifications vs. other notifications. I'm envisioning something like this just prior to this big ol' `return` statement we're in:

```javascript
if (mentionsInTitle)
    var foo = [m(... 'Instantly')];
else {
    var type = item.data.event.notificationType;
    var foo = [m(...), m(...), m(...)];
}
```

website/static/js/comment.js
```diff
@@ -13,7 +13,57 @@ require('jquery-autosize');
...
+    .atwho(atjsConfig.at_config)
+    .atwho(atjsConfig.plus_config)
+    .bind('paste', atjsConfig.onPaste)
+    .on('focusin keyup', atjsConfig.lastElementBr)
+    .on('focusout', atjsConfig.onlyElementBr)
+    .keydown(function(e) {
+        if(e.which === 13 && !e.isDefaultPrevented()) {
+            atjsConfig.onReturn(e);
+        }
+    });
```
whit537: It looks really weird to me to run configuration like this outside of some sort of `init` function. Do we really want loading this file to have this side-effect?

website/static/js/comment.js
```diff
@@ -62,6 +115,58 @@ var exclusifyGroup = function() {
     }
 };
 
+ko.subscribable.fn.convertHtmlToMarkdown = function(commentContent) {
+    return ko.computed(function() {
```
whit537: Why does this function signature have to be so complicated? Why can't we just have ...

```javascript
function convertHtmlToMarkdown(html) {
    ...
    return markdown;
}
```

? Then in the tests we can have the more thorough tests (making sure that `@` works, `+` works, etc.) only once—right now in https://github.com/CenterForOpenScience/osf.io/commit/913db15a6da64379504226baf483fb019aef673c we test the same things twice—with just one or two tests to make sure that the conversion code is getting called _at all_ by the higher-order knockout wrappers. Does that make any sense?

whit537: I guess it's not a generic HTML to Markdown converter, we're just looking at the links and `<br>`s. Maybe there's a better name.

website/static/js/atjsConfig.js
```diff
@@ -0,0 +1,153 @@
...
+    highlighter: function(li, query) {
+        var regexp;
+        if (!query) {
+            return li;
+        }
+        regexp = new RegExp('>\\s*([\\w\\s]*?)(' + query.replace('+', '\\+') +
+            ')([\\w\\s]*)\\s*<', 'ig');
+        return li.replace(regexp, function(str, $1, $2, $3) {
+            return '> ' + $1 + '<strong>' + $2 + '</strong>' + $3 + ' <';
+        });
```
whit537: Gosh, do we really need to use a regex here? Can't we just add a CSS class or something?

website/static/js/atjsConfig.js
```diff
@@ -0,0 +1,153 @@
...
+        space = acceptSpaceBar ? '\ ' : '';
+        regexp = new RegExp(flag + '([A-Za-z' + _a + '-' + _y + '0-9_' + space +
+            '\'\.\+\-]*)$|' + flag + '([^\\x00-\\xff]*)$', 'gi');
+        match = regexp.exec(subtext.replace(/\s/g, ' '));
+        if (match) {
+            return match[2] || match[1];
+        } else {
+            return null;
+        }
+    }
```
whit537: What is this magic? 😳

website/static/js/atjsConfig.js
```diff
@@ -0,0 +1,153 @@
...
+};
+
+// remove br if no text
+var onlyElementBr = function() {
+    if (this.innerText.trim() === '') {
+        this.innerHTML = '';
+    }
+};
+
+// make sure br is always the lastChild of contenteditable so return works properly
```
whit537: What does return do if br is not the lastChild of contenteditable? Is that where we start seeing extra `<p>` tags and such?

website/static/js/comment.js
leb2dg: I'll move it into the `init` function. It will only impact contenteditable fields with the `atwho-input` class though (`var input = $('.atwho-input');`).

website/static/js/atjsConfig.js
leb2dg: This is overriding the default [At.js callbacks](https://github.com/ichord/At.js/wiki/Callbacks) which is [here](https://github.com/ichord/At.js/blob/070d3181174a5893788f65e34ccea93dc91b5f83/dist/js/jquery.atwho.js#L107).

website/static/js/atjsConfig.js
leb2dg: This is overriding another default [At.js callbacks](https://github.com/ichord/At.js/wiki/Callbacks) which is [here](https://github.com/ichord/At.js/blob/070d3181174a5893788f65e34ccea93dc91b5f83/dist/js/jquery.atwho.js#L46).

whit537: Right, but *why*? :-)

When overriding an upstream method, it's good practice to document the code, explaining the reason for the override—it's not obvious from a quick visual scan of our code and their code.

website/static/js/atjsConfig.js
leb2dg: If I remember correctly, the first return just adds a `<br>` to the end with no impact on the UI and then on the second return it behaves normally. The `onReturn` function deals with the `<p>` tags and such.

bower.json
```diff
@@ -19,6 +19,7 @@
     "locales": "https://github.com/citation-style-language/locales.git#d2b612f8a6f764cbd66e67238636fac3888a7736",
     "raven-js": "2.1.0",
     "zeroclipboard": "2.1.6",
-    "osf-style": "https://github.com/CenterForOpenScience/osf-style.git#5dcf5997ad2808ffc7c8564f94cde102bf7a4a96"
+    "osf-style": "https://github.com/CenterForOpenScience/osf-style.git#5dcf5997ad2808ffc7c8564f94cde102bf7a4a96",
+    "At.js": "https://github.com/CenterForOpenScience/At.js.git#d4c1c6ebce3808a275bac59740dabd9bcf76d96f"
```
whit537: Did we consider changing the behavior through an override (cf. https://github.com/CenterForOpenScience/osf.io/pull/5840#discussion_r67533780) rather than a complete fork?

website/static/js/atjsConfig.js
leb2dg: Ohh it is to allow spaces in names. I'll add a comment.

bower.json
leb2dg: Yeah, unfortunately there is no built in way to override the `insert` method of `EditableController`.

website/static/js/atjsConfig.js
whit537: Okay. Let's at least comment this as with the other override.

website/project/model.py
```diff
@@ -179,6 +179,33 @@ class MetaData(GuidStoredObject):
     date_modified = fields.DateTimeField(auto_now=datetime.datetime.utcnow)
 
 
+def validate_contributor(guid, contributors):
+    if guid:
```
whit537: The way the regex in `get_valid_mentioned_users_guids` is written, I don't believe it's possible to have `''` for a guid here. If it *is* possible, then the way this is written will mean that any empty strings will be included in the return value of `get_valid_mentioned_users_guids`, which seems non-ideal. A couple options:

 - Remove this condition, and assume that `guid` will never be empty (though if it is, what will the effect be?)
  - We could explicitly filter out empty guids in the `new_mentions` list comprehension assignment on line 202-206.
 - Raise `ValidationValueError` when guid is empty.

whit537:
Alright, @leb2dg, I'm gonna hold off on reviewing the test suite further until the remaining refactoring settles down. Ping me when you're ready! :-)

whit537:
@leb2dg Roger that! :-)

website/notifications/emails.py
```diff
@@ -24,6 +24,7 @@ def notify(event, user, node, timestamp, **context):
     subscriptions = compile_subscriptions(node, event_type, event)
     sent_users = []
     target_user = context.get('target_user', None)
+
```
whit537: Let's take this line out since we're otherwise not touching this function.

website/static/js/atjsConfig.js
```diff
@@ -0,0 +1,172 @@
...
+            return li;
+        }
+        regexp = new RegExp('>\\s*([\\w\\s]*?)(' + query.replace('+', '\\+') +
+            ')([\\w\\s]*)\\s*<', 'ig');
+        return li.replace(regexp, function(str, $1, $2, $3) {
+            return '> ' + $1 + '<strong>' + $2 + '</strong>' + $3 + ' <';
+        });
+    },
+    matcher: function(flag, subtext, should_startWithSpace, acceptSpaceBar) {
+    /* Override upsream to support names with spaces */
```
whit537: Let's fix the typos here. :)

website/project/model.py
```diff
@@ -312,6 +340,12 @@ def create(cls, auth, **kwargs):
 
         log_dict.update(comment.root_target.referent.get_extra_log_params(comment))
 
+        if comment.content:
```
whit537: Why do we have to check `comment.content` here?

leb2dg: The check prevents a TypeError if the content is `None`. A validator is used to ensure the content is not none but it doesn't get called until the node is saved.
```
ERROR: test_create_comment_content_cannot_be_none (tests.test_comments.TestCommentModel)
  File "/Users/laurenbarker/GitHub/COS/osf.io/tests/test_comments.py", line 161, in test_create_comment_content_cannot_be_none
    content=None
  File "/Users/laurenbarker/GitHub/COS/osf.io/website/project/model.py", line 344, in create
    new_mentions = get_valid_mentioned_users_guids(comment, comment.node.contributors)
  File "/Users/laurenbarker/GitHub/COS/osf.io/website/project/model.py", line 200, in get_valid_mentioned_users_guids
    new_mentions = set(re.findall(r"\[[@|\+].*?\]\(\/([a-z\d]{5})\/\)", comment.content))
  File "/Users/laurenbarker/.virtualenvs/osf/lib/python2.7/re.py", line 181, in findall
    return _compile(pattern, flags).findall(string)
TypeError: expected string or buffer
```

leb2dg:
The project level, mention subscriptions were removed since mention notifications are restricted to 'Instantly'.

api/comments/serializers.py
```diff
@@ -111,6 +113,8 @@ def update(self, comment, validated_data):
                     comment.edit(content, auth=auth, save=True)
                 except PermissionsError:
                     raise PermissionDenied('Not authorized to edit this comment.')
+                except ValidationValueError:
+                    raise ValidationError('Ensure this field has no more than 500 characters.')
```
sloria: We might not want to have such a specific error message since we aren't validating against the exact number of entered characters. Something more generic will do, like `"Comment too long"`. 

Also, there may be other validators that could raise a `ValidationValueError`, so it would be more future-proof to use the error message from the `ValidationValueError`. Something like

```python
except ValidationValueError as err:
    raise ValidationError(err.args[0])
```

framework/mongo/validators.py
```diff
@@ -11,6 +12,20 @@ def string_required(value):
     return True
 
 
+def comment_maxlength(max_length):
+    def link_repl(matchobj):
+        return matchobj.group(1)
+
+    def validator(value):
+        reduced_comment = re.sub(r"\[([@|\+].*?)\]\(\/[a-z\d]{5}\/\)", link_repl, value)
```
sloria: Minor: As per COSDev, use singlequotes for strings.

sloria: Also minor: A simple optimization:

```python
mention_re = re.compile(r'\[....)')
def validator(value):
    reduced_comment = mention_re.sub(link_repl, value)
    #...
```

framework/mongo/validators.py
```diff
@@ -11,6 +12,20 @@ def string_required(value):
...
+
+        # two characters accounts for the \r\n at the end of comments
+        if len(reduced_comment) > max_length + 2:
+            raise ValidationValueError('Value exceed maximum length of {}'.format(max_length))
```
sloria: See https://github.com/CenterForOpenScience/osf.io/pull/5840/files#r68639556

api/comments/serializers.py
sloria: Oh.. I think I understand how the `comment_maxvalidator` is actually validating against the number of input characters. So no need to be too generic with the error message. I suppose you could use the value of `settings.COMMENT_MAXLENGTH` though.

website/notifications/emails.py
```diff
@@ -44,6 +44,19 @@ def notify(event, user, node, timestamp, **context):
                 sent_users.extend(subscriptions[notification_type])
     return sent_users
 
+def notify_mentions(event, user, node, timestamp, **context):
+    event_type = utils.find_subscription_type(event)
+    sent_users = []
+    new_mentions = context.get('new_mentions', None)
+    for m in new_mentions:
```
sloria: A `TypeError` will be raised here if `new_mentions` is assigned to `None` in the line above.

website/project/model.py
```diff
@@ -204,7 +230,9 @@ class Comment(GuidStoredObject, SpamMixin, Commentable):
     # The type of root_target: node/files
     page = fields.StringField()
     content = fields.StringField(required=True,
-                                 validate=[MaxLengthValidator(settings.COMMENT_MAXLENGTH), validators.string_required])
+                                 validate=[validators.comment_maxlength(settings.COMMENT_MAXLENGTH), validators.string_required])
+    # The mentioned users
+    ever_mentioned = fields.ListField(fields.StringField())
```
sloria: I'm not sure this is correct; everywhere else in the codebase I see

```python
fields.StringField(list=True)
```

sloria: Just curious: why "ever_mentioned"? Is there any reason to distinguish it from "mentioned"?

website/project/views/comment.py
```diff
@@ -139,6 +139,32 @@ def send_comment_added_notification(comment, auth):
...
 
 
+@mention_added.connect
+def send_mention_added_notification(comment, new_mentions, auth):
+    node = comment.node
+    target = comment.target
+
+    context = dict(
+        gravatar_url=auth.user.profile_image_url(),
+        content=markdown.markdown(comment.content, ['del_ins', 'markdown.extensions.tables', 'markdown.extensions.fenced_code']),
```
sloria: Minor: there's some duplication with this line: https://github.com/CenterForOpenScience/osf.io/blob/fe718f7ef94486949c7fe996ac07ea07c46a3ea0/website/project/views/comment.py#L114 . It would be useful to hide implementation details, e.g. which markdown plugins we're using:

```python
content=render_email_markdown(comment.content)
```

website/project/model.py
leb2dg: "ever_mentioned" is to make it clear that if users have previously been mentioned in a comment they will not be notified for any future mentions in that comment. "mentioned" could be interpreted as the mentions that are added in a particular save of a comment.

sloria: Got it. Thanks.

website/static/js/atjsConfig.js
```diff
@@ -0,0 +1,172 @@
...
+        } else {
+            return null;
+        }
+    }
+};
+
+var headerTemplate = '<div class="atwho-header">Contributors<small>&nbsp;↑&nbsp;↓&nbsp;</small></div>';
+var displayTemplate = '<li>${fullName}</li>';
+
+var at_config = {
```
sloria: Minor: As per COSDev, use lowerCamelCase for variable names.

website/static/js/atjsConfig.js
```diff
@@ -0,0 +1,172 @@
...
+    at: '@',
+    headerTpl: headerTemplate,
+    insertTpl: '@${fullName}',
+    displayTpl: displayTemplate,
+    searchKey: 'fullName',
+    limit: 6,
+    callbacks: callbacks
+};
+
+var plus_config = {
```
sloria: Same here.

sloria: Minor: You can dry this up with `$.extend`.

```js
var plusConfig = $.extend({}, atconfig, {
    at: '+',
    //...
}
```

website/templates/emails/mentions.txt.mako
```diff
@@ -0,0 +1,3 @@
+${user} at ${localized_timestamp}: ${content}.
```
sloria: Is the content and UI for this email in the spec? I would think we'd want to use our existing notifications HTML email template for this.

sloria:
Just a few driveby comments. Looking good so far. 

@whit537 As a heads up, we're trying to fast-track this PR to staging, especially because Lauren is moving teams. Please let Lauren know if there are any blocking issues here.

website/project/model.py
whit537: FTR: https://github.com/leb2dg/osf.io/pull/14#discussion_r61786124

whit537: In other words, `ever_mentioned` was originally meant to contrast with `new_mentions`. Since there's now only one attribute instead of two, it could make sense to simplify to `mentions`.

whit537:
@sloria Roger that. I've updated the [punchlist](https://github.com/CenterForOpenScience/osf.io/pull/5840#issuecomment-226013705) with your comments.

api/comments/serializers.py
whit537: `ValidationValueError` is a [subclass](https://github.com/CenterForOpenScience/modular-odm/blob/d13f941e4d6c90a4840c492b7225e8e85dafeea7/modularodm/exceptions.py#L41-L44) of `ValidationError`. Why do we need to capture and reraise here at all? Why not let the `ValidationValueError` itself raise?

whit537: The docstrings over there suggest that the use of `ValidationError` elsewhere in this file is straining the design of the system.

framework/mongo/validators.py
whit537: Typo: "exceed" &rarr; "exceeds"

website/static/js/tests/comment.test.js
```diff
@@ -0,0 +1,48 @@
...
+                assert.equal(converted, returnHTML);
+            });
+            it('convert @ mention markdown link to span', () => {
+                var converted = convertMentionMarkdownToHtml(atMentionMarkdown);
+                assert.equal(converted, atMentionHTML);
+            });
+            it('convert + mention markdown link to span', () => {
+                var converted = convertMentionMarkdownToHtml(plusMentionMarkdown);
+                assert.equal(converted, plusMentionHTML);
+            });
```
whit537: I believe this block should be dedented a level.

whit537:
Alright, I'm good on the latest major commits here (skipping some things in the interest of expediency). The last thing on my list is to look over the test suite. If I don't get to that in time that consider me 👍 here.

api_tests/comments/views/test_comment_detail.py
```diff
@@ -202,7 +203,7 @@ def test_public_node_logged_out_user_cannot_update_comment(self):
 
     def test_update_comment_cannot_exceed_max_length(self):
         self._set_up_private_project_with_comment()
-        content = ''.join(['c' for c in range(osf_settings.COMMENT_MAXLENGTH + 1)])
+        content = ''.join(['c' for c in range(osf_settings.COMMENT_MAXLENGTH + 3)])
```
whit537: This can be simplified to `'c' * (osf_settings.COMMENT_MAXLENGTH + 3)`.

api_tests/nodes/views/test_node_comments_list.py
```diff
@@ -647,7 +647,7 @@ def test_create_comment_exceeds_max_length(self):
             'data': {
                 'type': 'comments',
                 'attributes': {
-                    'content': (''.join(['c' for c in range(osf_settings.COMMENT_MAXLENGTH + 1)]))
+                    'content': (''.join(['c' for c in range(osf_settings.COMMENT_MAXLENGTH + 3)]))
```
whit537: Here, too (https://github.com/CenterForOpenScience/osf.io/pull/5840/files#r68674521).

whit537:
Okay, I've skimmed the test suite. I'm 👍 here.

whit537:
Well, I don't currently have anything more to add to the [punchlist](https://github.com/CenterForOpenScience/osf.io/pull/5840#issuecomment-226013705), at least. :)

website/project/model.py
leb2dg: @sloria, do we not use the [ListField](https://github.com/CenterForOpenScience/modular-odm/blob/develop/modularodm/fields/listfield.py) from modular-odm?

api/comments/serializers.py
leb2dg: It is a little weird because the max_length validator was previously on the [serializer](https://github.com/CenterForOpenScience/osf.io/blob/develop/api/comments/serializers.py#L40) and now the length check is only on the [model](https://github.com/leb2dg/osf.io/blob/f59b9b72886868f16c993571fe99229484a56239/website/project/model.py#L233) since the content needs to be processed first. So the model is throwing the modularodm `ValidationValueError` and the serializer is catching it and raising the rest framework `ValidationError` it would normally throw.

whit537: Well, golly, would ya look at that. Indeed, this isn't ModularODM's `ValidationError` after all!

whit537: In which case, how about going back to @sloria's original suggestion?

whit537: (Namely, https://github.com/CenterForOpenScience/osf.io/pull/5840/files/f59b9b72886868f16c993571fe99229484a56239#r68639556.)

website/templates/emails/mentions.txt.mako
leb2dg: It is the same as [comment email template](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/templates/emails/comments.txt.mako). What is the notifications HTML email template? [notify_base.mako](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/templates/emails/notify_base.mako)?

leb2dg:
The ListField and email template still need to be addressed once a decision has been made about them.

website/project/model.py
sloria: Looking through the OSF, looks like we use `list=True` whenever we need an mongo array field.

sloria: Although, now that I think of it, would this make more sense as a one-to-many ForeignField to the User model?

website/templates/emails/mentions.txt.mako
sloria: Oh, didn't notice this was the same as the comment email template; this is probably fine then. Do @mention emails get sent immediately or do they get bundled together over 5-minute time periods (by send_user_email: https://github.com/CenterForOpenScience/osf.io/blob/a5752fa055516ee3dd5fd04613ddaebb723c26b6/website/notifications/tasks.py#L20) ?

leb2dg: They get stored and sent out by `send_users_email`. The email flow wasn't altered.

sloria:
I'm seeing a JS error if I hit Backspace after entering a mention:

jserror

website/static/js/atjsConfig.js
```diff
@@ -0,0 +1,167 @@
...
+        match = regexp.exec(subtext.replace(/\s/g, ' '));
+        if (match) {
+            return match[2] || match[1];
+        } else {
+            return null;
+        }
+    }
+};
+
+var headerTemplate = '<div class="atwho-header">Contributors<small>&nbsp;↑&nbsp;↓&nbsp;</small></div>';
```
sloria: Very minor: The arrows are probably unnecessary and possibly misleading; I thought they were clickable sort buttons. Not a big deal though.

sloria:
Seeing a JS error when typing some combinations of non-ASCII characters in succession.

Here, holding Alt and pressing "o" and "i":

strange

website/static/css/commentpane.css
```diff
@@ -122,3 +122,20 @@
...
+    display: block;
+    height: auto;
+    min-height: 56px;
+    overflow: hidden;
+    word-wrap: break-word;
+}
+
+.atwho-input[placeholder]:empty:before {
+    content: attr(placeholder);
+    color: #555;
```
sloria: Minor: The placeholder text should be a more muted color than input text. 

In this PR:

osf___small_project


On prod (Chrome's native placeholder styling):

osf___small_project

website/static/js/atjsConfig.js
```diff
@@ -0,0 +1,167 @@
...
+    document.execCommand('insertHTML', false, pasteText);
+};
+
+// prevent bold, italic, underline
+var preventKeyboardShortcuts = function(e) {
+    var k = e.keyCode;
+    var c = e.ctrlKey || e.metaKey;
+
+    // bold, italic, underline
+    if ((k === 66 && c) || (k === 73 && c) || (k === 85 && c)){
```
sloria: Constants would help here:

```js
if ((k === B_KEY ...)
```

website/static/js/atjsConfig.js
```diff
@@ -0,0 +1,167 @@
...
+var $ = require('jquery');
+var ko = require('knockout');
+
+require('Caret.js');
+require('At.js');
+
+// preventing unexpected contenteditable behavior
+
+// prevent pasting HTML
+var onPaste = function(e) {
```
sloria: If there is any code copied from somewhere else, it's very helpful to provide the source as it helps us distinguish copied code from custom code. Also provide any attribution if necessary.

Not a blocker, as long as we're following the terms of this code's license.

leb2dg: A few of the functions in atjsConfig.js are approximately 60% from elsewhere. Should I document those locations?

website/static/js/comment.js
```diff
@@ -14,7 +14,44 @@ require('jquery-autosize');
...
+                'id': item.id,
+                'name': item.embeds.users.data.attributes.given_name,
+                'fullName': item.embeds.users.data.attributes.full_name,
+                'link': item.embeds.users.data.links.html
+            };
+        });
+        // for any input areas that currently exist on page
+        input.atwho('load','@', data).atwho('load', '+', data).atwho('run');
+        // for future input areas so that data doesn't need to be reloaded
+        atjsConfig.atConfig.data = data;
```
sloria: Minor: It would be nice to avoid mutating this object that we're importing (`atConfig`)

```js
// this would go whereever you initialize At.js
getContributorList().then(function(contributors) {
    var config = $.extend({}, atjsConfig.atConfig, {data: contributors});
    $.atjs(config);
};
```

Not a blocking refactor, though.

website/static/js/atjsConfig.js
sloria: That would be helpful, yes; we may need to refer back to those in the future.

website/static/js/comment.js
```diff
@@ -40,6 +77,9 @@ var relativeDate = function(datetime) {
 };
 
 var notEmpty = function(value) {
+    if (value === '<br>')  {
```
sloria: Should you also handle a `<br>` with any surrounding whitespace, i.e. `'<br> '`?

sloria: You could simplify this function to

```js
var trimmed = $.trim('<br>').toLowerCase();
return !!trimmed && trimmed !== '<br>';
```

website/static/js/comment.js
```diff
@@ -63,6 +103,54 @@ var exclusifyGroup = function() {
     }
 };
 
+var convertMentionHtmlToMarkdown = function(commentContent) {
+    var content = commentContent || '';
+    var pattern = '<span[^>]*?data-atwho-guid="([a-z\\d]{5})"[^>]*?>((@|\\+)[\\w\\s]+)<\/span>',
+        regex = new RegExp(pattern),
```
sloria: Minor: As per COSDev, declare one `var` per line.
https://github.com/felixge/node-style-guide#declare-one-variable-per-var-statement

sloria: BTW, it's ok to diverge from the style guidelines iff this is copy-n-pasted code; just make clear that it was copied.

website/static/js/koHelpers.js
```diff
@@ -459,6 +459,50 @@ ko.bindingHandlers.datePicker = {
...
  /**
+ * Bind content of contenteditable to observable. Looks for maxlength attr
+ * and underMaxLength binding to limit input.
+ * Example:
+ * <div contenteditable="true" data-bind="editableHTML: <observable_name>" maxlength="500"></div>
+ */
+ko.bindingHandlers.editableHTML = {
+    init: function(element, valueAccessor, allBindings, bindingContext) {
+        var $element = $(element);
+        var initialValue = ko.utils.unwrapObservable(valueAccessor());
```
sloria: Minor: You shouldn't invoke `valueAccessor` if you're going to use `unwrapObservable`. The function is meant to handle that (https://github.com/knockout/knockout/blob/master/src/utils.js#L418).

sloria: Note: If you require valueAccessor to be an observable (you assign to it to $element.html() by invoking it a few lines down), then don't use `unwrapObservable`.

website/static/js/koHelpers.js
```diff
@@ -459,6 +459,50 @@ ko.bindingHandlers.datePicker = {
...
+        var charLimit = $element.attr('maxlength');
+        if (charLimit && bindingContext.underMaxLength) {
+            var inputTextLength = $element[0].innerText.length;
+            var errorExceedsCharacterLimit = 'Exceeds character limit. Please reduce to ' +
+                charLimit + ' characters or less.';
+            // + 1 to account for the <br> that is added to the end of the contenteditable content
+            // <br> is necessary for the return key to function properly
+            if (inputTextLength > parseInt(charLimit) + 1) {
+                bindingContext.underMaxLength(false);
```
sloria: This makes this binding handler difficult to reuse because it assumes a lot about the ViewModel that controls it (it requires `underMaxLength` and `errorMessage` to be observables on the ViewModel).

How about passing in an error handler function?

```html
<div contenteditable="true" 
    data-bind="editableHTML: {observable: comment, onError: handleError" maxlength="500">
</div>
```

```js
var ViewModel = function(...) {
    var self = this;
    self.content = ko.observable();
    self.underMaxLength(true);
    self.errorMessage = ko.observable();
    self.handleError = function(message) {
          self.errorMessage(message);
          self.underMaxLength(false);
     }
}
```

Not a blocker, but would be a nice refactor.

website/templates/emails/mentions.html.mako
```diff
@@ -0,0 +1,19 @@
+<% from website.models import User %>
```
sloria: Minor: This can be removed.

sloria:
Pass finished.

sloria:
Carrying this over to https://github.com/CenterForOpenScience/osf.io/pull/5943 , which has been updated with develop.

PR #5841, opened 2016-06-14
[RFR][Feature][OSF-6504]Add exception handling for invalid subscription targets 

TomBaxter:
## Purpose:
[OSF-6504](https://openscience.atlassian.net/browse/OSF-6504)
Fix "populate_institutions.py no longer works"

## Changes:
Add  website/notifications/exceptions.py containing class InvalidSubscriptionError(NodeError)
Update website/notifications/listeners.py to try/except and catch InvalidSubscriptionError
Update website/notifications/utils.py "def subscribe_user_to_notifications" raise InvalidSubscriptionError for institutions/deleted nodes/collections targets

## Side effects
None

[#OSF-6504]

### DIFF
--- a/website/notifications/exceptions.py
+++ b/website/notifications/exceptions.py
@@ -1,0 +1,8 @@
+from website.exceptions import NodeError
+
+class InvalidSubscriptionError(NodeError):
+    """Raised if an invalid subscription is attempted. e.g. attempt to
+    subscribe to an invalid target: institution, bookmark, deleted project etc.
+    """
+    message_short = 'Invalid Subscription'
+    message_long = 'This Subscription is not valid.'
--- a/website/notifications/exceptions.py
+++ b/website/notifications/exceptions.py
@@ -1,8 +1,7 @@
-from website.exceptions import NodeError
+from website.exceptions import OSFError
 
-class InvalidSubscriptionError(NodeError):
+class InvalidSubscriptionError(OSFError):
     """Raised if an invalid subscription is attempted. e.g. attempt to
     subscribe to an invalid target: institution, bookmark, deleted project etc.
     """
-    message_short = 'Invalid Subscription'
-    message_long = 'This Subscription is not valid.'
+    pass
--- a/website/notifications/listeners.py
+++ b/website/notifications/listeners.py
@@ -1,12 +1,27 @@
+import logging
+from website.notifications.exceptions import InvalidSubscriptionError
 from website.notifications.utils import subscribe_user_to_notifications
 from website.project.signals import contributor_added, project_created
 from website.project.views.contributor import notify_added_contributor
 
+logger = logging.getLogger(__name__)
+
 @project_created.connect
 def subscribe_creator(node):
-    subscribe_user_to_notifications(node, node.creator)
+    try:
+        subscribe_user_to_notifications(node, node.creator)
+    except InvalidSubscriptionError as err:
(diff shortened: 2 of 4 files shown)


sloria:
Good catch!

Rather than disconnecting the signal, it would be safer to modify `subscribe_user_to_notifications` to make it impossible for any user to get subscribed to an institution. We should also handle other nodes that shouldn't be subscribed to, such as Bookmark nodes. If subscribe_user_to_notifications is called with one of these nodes, you could raise a custom error.

Once that's done, the signal listener becomes something like:

```python
from website.notifications.exceptions import InvalidSubscriptionError

@project_created.connect
def subscribe_creator(node):
    try:
    except InvalidSubscriptionError as err:
        logger.warn('Skipping subscription of user {} to node {}'.format(node.creator._id, node._id)
        logger.warn('Reason: {}'.format(str(err))
```

sloria:
Pass finished.

sloria:
Deleted projects should also not be able to be subscribed to.

TomBaxter:
PR has been updated per CR. Ready for further review.

website/notifications/utils.py
```diff
@@ -404,6 +405,16 @@ def subscribe_user_to_notifications(node, user):
 
     :param user: User to subscribe to notifications
     """
+
+    if node.institution_id:
+        raise InvalidSubscriptionError('Institutions are invalid targets for subscriptions')
+
+    if node.is_bookmark_collection:
```
brianjgeiger: `is_collection` is a better choice than `is_bookmark_collection` because all bookmark collections are collections.

TomBaxter: Ok wasn't sure what the difference between bookmark_collections and collections. Are all collections invalid targets for subscriptions?

brianjgeiger: Collections are invalid targets for subscriptions, yes.

website/notifications/exceptions.py
```diff
@@ -0,0 +1,8 @@
+from website.exceptions import NodeError
+
+class InvalidSubscriptionError(NodeError):
```
sloria: I don't think this needs to subclass `NodeError`. It isn't specific to nodes (it can be raised for user subscriptions). Also, `message_short` and `message_long` aren't useful, since the error is never reraised as an HTTPError.

website/notifications/listeners.py
```diff
@@ -1,12 +1,27 @@
...
+        logger.warn('Reason: {}'.format(str(err)))
 
 @contributor_added.connect
 def subscribe_contributor(node, contributor, auth=None, *args, **kwargs):
-    subscribe_user_to_notifications(node, contributor)
-    notify_added_contributor(node, contributor, auth)
+    try:
+        subscribe_user_to_notifications(node, contributor)
+    except InvalidSubscriptionError as err:
+        user = contributor or 'None'
```
sloria: Minor: This is unnecessary; `None` already stringifies to `"None"`.

```python

In [1]: '{}'.format(None)
Out[1]: 'None'
```

sloria:
Just a few additional drive-by comments. Once those and Brian's suggestion are addressed, this is good to merge.

website/notifications/listeners.py
TomBaxter: Switched is_bookmark_collection to is_collection
removed message_short and message_long
removed user = contributor or 'None'
Ready for further review

PR #5843, opened 2016-06-15
[Feature] Move dev_only routes to regular [OSF-6457]

Ghalko:
## Purpose

This moves the list routes that are currently `dev_mode` into normal routes.

## Changes

- Add list routes to `links` in base. `metaschemas` (highlighted in second) are dev only still.

before
screen shot 2016-06-16 at 3 48 23 pm

after
screen shot 2016-06-16 at 3 51 11 pm

## Side effects

More lists.

## Ticket

https://openscience.atlassian.net/browse/OSF-6457

### DIFF
--- a/api/base/views.py
+++ b/api/base/views.py
@@ -464,11 +464,10 @@ def root(request, format=None):
         'links': {
             'nodes': absolute_reverse('nodes:node-list'),
             'users': absolute_reverse('users:user-list'),
+            'collections': absolute_reverse('collections:collection-list'),
+            'registrations': absolute_reverse('registrations:registration-list'),
         }
     }
-    if settings.DEV_MODE:
-        return_val['links']['collections'] = absolute_reverse('collections:collection-list')
-        return_val['links']['registrations'] = absolute_reverse('registrations:registration-list')
 
     return Response(return_val)
 
--- a/api/base/views.py
+++ b/api/base/views.py
@@ -8,7 +8,6 @@ from rest_framework.mixins import ListModelMixin
 
 from api.users.serializers import UserSerializer
 
-from website import settings
 from django.conf import settings as django_settings
 from .utils import absolute_reverse, is_truthy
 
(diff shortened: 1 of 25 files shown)


api/base/views.py
```diff
@@ -464,11 +463,13 @@ def root(request, format=None):
         'links': {
             'nodes': absolute_reverse('nodes:node-list'),
             'users': absolute_reverse('users:user-list'),
+            'collections': absolute_reverse('collections:collection-list'),
+            'registrations': absolute_reverse('registrations:registration-list'),
+            'institutions': absolute_reverse('institutions:institution-list'),
+            'licenses': absolute_reverse('licenses:license-list'),
+            'metaschemas': absolute_reverse('metaschemas:metaschema-list'),
```
brianjgeiger: It looks like the feature that uses this will remain in dev_mode for the moment, so could you move metaschemas down into the if block you deleted below? Everything else looks right. Pass done. 🐧

Ghalko: Done.

brianjgeiger:
Looks great. 🎉 🐧 🎉 Could you update the screenshot with the current version, and put a before screenshot of what's on production right now? (Or you can do blocks of json rather than screenshots, either way)

Ghalko:
Done.

api/base/views.py
```diff
@@ -464,11 +464,14 @@ def root(request, format=None):
...
+            'collections': absolute_reverse('collections:collection-list'),
+            'registrations': absolute_reverse('registrations:registration-list'),
+            'institutions': absolute_reverse('institutions:institution-list'),
+            'licenses': absolute_reverse('licenses:license-list'),
         }
     }
     if settings.DEV_MODE:
-        return_val['links']['collections'] = absolute_reverse('collections:collection-list')
-        return_val['links']['registrations'] = absolute_reverse('registrations:registration-list')
+        return_val['links']['metaschemas'] = absolute_reverse('metaschemas:metaschema-list')
```
sloria: metaschema-list` is now out of dev-only as well. I'll move this when I merge.

PR #5845, opened 2016-06-15
Add ability to invoke automatic maintenance notice [#OSF-6482]

erinspace:
## Purpose
To put up a maintenance notice right now, we manually commit and uncommit code that adds a specified banner to the top of the page.  This adds instead a way to do this programmatically, either through an invoke task or by calling the maintenance module directly (future uses in the admin module!)

Examples:
```
        invoke set_maintenance_state
        invoke set_maintenance_state --start 2016-03-16T15:41:00-04:00
        invoke set_maintenance_state --end 2016-03-16T15:41:00-04:00
```
## Changes
- Add a maintenance module with a few options:
    - **set_maintenance**:  make a new collection called 'maintenance' in the database that will have some info about what period to set the maintenance from. Either pass in a start and end date or have them be calculated for you. Dates default to UTC if no timezone info is specified.
    - **get_maintenance**: return the current state of the maintenance notice, if there is one
    - **unset_maintenance**: get rid of the maintenance collection altogether, removing the notice
- Update ```base-page.js``` to have display info if there is maintenance info to display, in local time converted from UTC

## Side effects
- None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-6482

### DIFF
--- a/tasks/__init__.py
+++ b/tasks/__init__.py
@@ -10,12 +10,16 @@ import json
 import platform
 import subprocess
 import logging
-from time import sleep
+import pymongo
+from datetime import datetime, timedelta
+
+from dateutil.parser import parse
 
 import invoke
 from invoke import run, Collection
 
 from website import settings
+from framework.mongo import database
 from utils import pip_install, bin_prefix
 
 logging.getLogger('invoke').setLevel(logging.CRITICAL)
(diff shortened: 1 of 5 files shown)


website/maintenance.py
```diff
@@ -0,0 +1,47 @@
...
+        start = start.replace(tzinfo=pytz.UTC)
+
+    if not end.tzinfo:
+        end = end.replace(tzinfo=pytz.UTC)
+
+    if start > end:
+        start = end - timedelta(1)
+
+    database.drop_collection('maintenance')
+    database.maintenance.insert({'maintenance': True, 'start': start.isoformat(), 'end': end.isoformat()})
```
sloria: Mongo can store datetimes, so no need to stringify the dates here.

tasks/__init__.py
```diff
@@ -958,3 +958,40 @@ def clean(verbose=False):
 @task(default=True)
 def usage():
     run('invoke --list')
+
+
+### Maintenance Tasks ###
+
+@task
+def set_maintenance_state(start=None, end=None):
```
sloria: Minor: for consistency in naming (and less typing), `set_maintenance` is fine for the command name. Same with the other task names.

tasks/__init__.py
```diff
@@ -958,3 +958,40 @@ def clean(verbose=False):
...
+
+    If a given end time results in a start that is after the end, start
+    will be changed to be 24 hours before the end time.
+
+    Examples:
+        invoke set_maintenance_state
+        invoke set_maintenance_state --start 2016-03-16T15:41:00-04:00
+        invoke set_maintenance_state --end 2016-03-16T15:41:00-04:00
+    """
+    set_maintenance(start, end)
```
sloria: Minor: Some output would be nice, but not strictly necessary.

tasks/__init__.py
```diff
@@ -958,3 +958,40 @@ def clean(verbose=False):
...
+
+
+@task
+def get_maintenance_state():
```
sloria: This task doesn't do anything (there's no output). We can probably just remove this.

website/maintenance.py
sloria: Disregard my comment; this is actually the correct thing to do. Pymongo (or at least the version we're using) only allows us to retrieve timezone-naive datetimes, so we shouldn't stringify the dates *after* they are stored. Stringifying upon storage ensures that we keep timezone info.

PR #5853, opened 2016-06-16
[Feature] Add HttpOnly flag for OSF, API and ADMIN cookie [#OSF-6339, #SEC-16]

cslzchen:
## Purpose

OSF does not set `HttpOnly` flag. Using the flag when generating a cookie helps mitigate the risk of client side script accessing the protected cookie.

## Changes

```
if response is not None:
    response.set_cookie(
        settings.COOKIE_NAME,
        value=cookie_value,
        domain=settings.OSF_COOKIE_DOMAIN,
        secure=settings.SESSION_COOKIE_SECURE,
        httponly=settings.SESSION_COOKIE_HTTPONLY)
    return response
```

[
osf-cookie-httponly
](url)

Note: The above screenshot is from local environment. `Secure` flag is disabled locally.

## Side effects

Not compatible with IE9 or earlier. We do not support IE9.
Scripts can no longer access OSF cookie.

## Staging, Staging2, Production, Test Deployment

```
SESSION_COOKIE_HTTPONLY = True
```

## Local Development

Do not change `SESSION_COOKIE_HTTPONLY` to `False`. OSF cookie should never be accessed with scripts directly. Use Cross-Domain XMLHttpRequest with correct security settings instead.

## Tickets

https://openscience.atlassian.net/browse/OSF-6339
https://openscience.atlassian.net/browse/SEC-16

## PRs

Blocked by https://github.com/CenterForOpenScience/osf.io/pull/5506


framework/auth/cas.py
```diff
@@ -208,15 +213,23 @@ def get_logout_url(*args, **kwargs):
...
+
 def make_response_from_ticket(ticket, service_url):
-    """Given a CAS ticket and service URL, attempt to the user and return a proper
-    redirect response.
     """
+    Given a CAS ticket and service URL, attempt to the user and return a proper redirect response.
+
+    :param ticket:
+    :param service_url:
+    :param user_agent:
```
mfraezz: It doesn't look like `user_agent` is actually a param for this function.

framework/sessions/__init__.py
```diff
@@ -95,7 +95,7 @@ def set_session(session):
     sessions[request._get_current_object()] = session
 
 
-def create_session(response, data=None):
+def create_session(response, data=None, user_agent=None):
```
mfraezz: It doesn't look like `user_agent` is used in this function, or ever passed in when it's called

framework/auth/cas.py
cslzchen: You are right. Working on a clean-up right now. I used to have it for IE9 compatibility.

framework/auth/cas.py
```diff
@@ -208,15 +213,23 @@ def get_logout_url(*args, **kwargs):
     """
     return get_client().get_logout_url(*args, **kwargs)
 
+
 def get_profile_url():
     """Convenience function for getting a profile URL for a user.
     """
     return get_client().get_profile_url()
 
+
```
mfraezz: Nitpicky, but we tell `flake` to ignore the PEP8 requirement of having two blank lines between methods. Not a blocker

cslzchen: Agreed. However, In code that have both 1 and 2 blank lines, do we prefer to make the style consistent or just leave them the way it is.

framework/sessions/__init__.py
```diff
@@ -128,60 +122,53 @@ def before_request():
...
         user = get_user(
             email=request.authorization.username,
             password=request.authorization.password
         )
-        # Create empty session
+        # Create an empty session
         # TODO: Shoudn't need to create a session for Basic Auth
-        session = Session()
-        set_session(session)
+        user_session = Session()
```
cslzchen: Note for CR: change shadowing variable name to better alternative.

framework/auth/cas.py
mfraezz: It's really up to personal preference, but according to [PEP8](https://www.python.org/dev/peps/pep-0008/#blank-lines):

> Surround top-level function and class definitions with two blank lines.
>
>Method definitions inside a class are surrounded by a single blank line.
>
>Extra blank lines may be used (sparingly) to separate groups of related functions. Blank lines may be omitted between a bunch of related one-liners (e.g. a set of dummy implementations).
>
>Use blank lines in functions, sparingly, to indicate logical sections.

cslzchen:
Code clean up done:
- fix imports order: python, third-party, our own
- remove unused params `user-agent`
- unify comments style and blank lines
- rename three shadowing variables `session` to respective names

framework/auth/cas.py
cslzchen: Thanks. I am for the style as well so I try to improve them on files (only related to my tickets) Changing too much may cause headaches of merge conflicts for everyone so I took this "little by little" approach. :)

mfraezz:
This seems to have a lot of extraneous formatting changes; the only thing that looks like it should affect behavior is the toggling of a `SESSION_COOKIE_HTTPONLY` flag (as expected). While the whitespace changes aren't all to my personal taste for readability, I applaud the docstring improvements. 👏 

Pass complete, pushing this ahead 👓

cslzchen:
Pull from develop to solve merge conflicts. Travis error is unrelated to this PR.

PR #5854, opened 2016-06-16
[Feature fix] global notification settings [OSF-5748]

leb2dg:
## Purpose

Newly created components should automatically be set to "Adopt settings from parent project"
Currently on staging, new components are getting the defaults from User Settings, Notifications, and not inheriting the settings from the parent project.

## Changes

* have created components default to `adopt_parent`
* only projects where the user has changed the settings appear in configured projects

## Side effects

Users have no subscription objects for components they create.

## Ticket

### DIFF
--- a/website/notifications/utils.py
+++ b/website/notifications/utils.py
@@ -175,8 +175,11 @@ def get_configured_projects(user):
         # If the user has opted out of emails skip
         node = subscription.owner
 
-        if not isinstance(node, Node) or (user in subscription.none and not node.parent_id) or node.is_bookmark_collection:
-            continue
+        if (not isinstance(node, Node) or
+                (user in subscription.none and not node.parent_id) or not
+                node.notification_settings_dirty or
+                node.is_bookmark_collection):
+                    continue
 
         while node.parent_id and not node.is_deleted:
             node = Node.load(node.parent_id)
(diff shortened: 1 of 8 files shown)


website/notifications/utils.py
```diff
@@ -426,14 +429,17 @@ def subscribe_user_to_notifications(node, user):
...
-                subscription = NotificationSubscription(_id=event_id, owner=node, event_name=event)
-            if global_subscription:
-                global_notification_type = get_global_notification_type(global_subscription, user)
-                subscription.add_user_to_subscription(user, global_notification_type)
-            else:
-                subscription.add_user_to_subscription(user, notification_type)
-            subscription.save()
+
+            # if no subscription for component and creator is the user, do not create subscription
+            if not(node and node.parent_node and not subscription and node.creator == user):
```
sloria: Minor: Checking the truthiness of `node` is unnecessary; this function assumes the `node` argument is a `Node`.

sloria: Also minor: it might be worth mentioning in a code comment that "no subscription" = "adopt from parent". As you know, this isn't immediately obvious.

website/project/model.py
```diff
@@ -3681,6 +3683,10 @@ def institutions_url(self):
     def institutions_relationship_url(self):
         return self.absolute_api_v2_url + 'relationships/institutions/'
 
+    def flip_notification_settings_dirty(self):
+        self.notification_settings_dirty = True
+        self.save()
```
sloria: We try to avoid implicit saves as much as possible because they may cause unwanted side effects--especially risky given how much logic is in `Node#save`.

```python
node.flip_notification_settings.dirty()  # implicit save
node.category = 'data'
node.save()   # another save
```

I think we can get ride of this method; setter methods in Python are usually unnecessary.

```python
node.notification_settings_dirty = True
node.category = 'data'
node.save()
```

website/notifications/utils.py
```diff
@@ -176,8 +176,11 @@ def get_configured_projects(user):
         # If the user has opted out of emails skip
         node = subscription.owner
 
-        if not isinstance(node, Node) or (user in subscription.none and not node.parent_id) or node.is_bookmark_collection:
-            continue
+        if (not isinstance(node, Node) or
+                (user in subscription.none and not node.parent_id) or not
+                node.notification_settings_dirty or
+                node.is_bookmark_collection):
```
sloria: Minor: I realize this was here before, but I think we should skip over any collection (not just Bookmarks), i.e. `node.is_collection`.

sloria:
Pass finished. Just nitpicks; I can address those when I merge.

leb2dg:
I'll be done updating the changes in approximately 1 minute if that is easier.

sloria:
@leb2dg No worries; I've already made the changes.

leb2dg:
@sloria did you get the tests?

sloria:
@leb2dg Sure did.

leb2dg:
💯  🎆  🎉

PR #5856, opened 2016-06-17
[migration] Hotfix/Prereg Metaschema Fix [OSF-6515]

pattisdr:
## Purpose

Prereg Challenge metaschema has extra spaces after q5 multiple choice responses.  This makes it tricky to answer this question in the APIv2 because this extra space is required in the request.

## Changes

Removes spaces from schema and requires migration to repair drafts and registrations that used this schema and answered q5.

## Side effects
Migration required.

## Ticket
https://openscience.atlassian.net/browse/OSF-6515


pattisdr:
Going to master

PR #5860, opened 2016-06-17
[RFR][Feature][OSF-2718]Can't see all components when permission to one is denied

TomBaxter:
## Purpose:
[OSF-2718](https://openscience.atlassian.net/browse/OSF-2718)
In project overview page and project file view page...
- Enable viewing components masked by unauthorized parent.
- Allow for breadcrumbs to to get back from masked components.
In project overview page
- Update components widget to match info in file tree widget.

## Bonus
Remove visibility of unauthorized components. "Private Component"

## Changes:
Updated tests/test_views.py
* Updated various for get_children now get_readable_descendants
* Fixed test_pointer_list_read_contributor_cannot_remove_public_component_entry now actually public
* Added def test_readable_descendants_masked_by_permissions

Updated website/project/model.py
* Added def find_readable_antecedent to find logical parent
* Added def find_readable_descendants to find logical children

Updated website/project/views/node.py "def _view_project" to find logical parent and children

Updated website/routes.py added get_readable_descendants route for use by component widget

Updated website/templates/project/project.mako to utilize new get_readable_descendants route

Updated website/util/rubeus.py "def _collect_components" to find logical children and not return unauthorized nodes

## Side effects
None

[#OSF-2718]

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -1164,6 +1164,87 @@ class TestChildrenViews(OsfTestCase):
         pointer = Pointer.find_one(Q('node', 'eq', pointed))
         assert_equal(nodes[0]['id'], pointer._primary_key)
 
+    def test_readable_descendants_masked_by_permissions(self):
+        # Users should be able to see through components they do not have
+        # permissions to.
+        # Users should not be able to see through links to nodes they do not
+        # have permissions to.
+        #
+        #                   1(AB)
+        #                  /  |  \
+        #                 *   |   \
+        #                /    |    \
+        #             2(A)  4(B)    7(A)
+        #               |     |     |    \
+        #               |     |     |     \
+        #             3(AB) 5(B)    8(AB) 9(B)
+        #                     |
+        #                     |
+        #                   6(A)
+        #
+        #
+        userA = AuthUserFactory()
+        userB = AuthUserFactory()
+
+        project1 = ProjectFactory(creator=self.user, title='One')
+        project1.add_contributor(userA, auth=Auth(self.user), permissions=['read'])
+        project1.add_contributor(userB, auth=Auth(self.user), permissions=['read'])
+
+        component2 = ProjectFactory(creator=self.user, title='Two')
+        component2.add_contributor(userA, auth=Auth(self.user), permissions=['read'])
+
+        component3 = ProjectFactory(creator=self.user, title='Three')
+        component3.add_contributor(userA, auth=Auth(self.user), permissions=['read'])
+        component3.add_contributor(userB, auth=Auth(self.user), permissions=['read'])
+
+        component4 = ProjectFactory(creator=self.user, title='Four')
+        component4.add_contributor(userB, auth=Auth(self.user), permissions=['read'])
+
+        component5 = ProjectFactory(creator=self.user, title='Five')
+        component5.add_contributor(userB, auth=Auth(self.user), permissions=['read'])
+
+        component6 = ProjectFactory(creator=self.user, title='Six')
+        component6.add_contributor(userA, auth=Auth(self.user), permissions=['read'])
+
+        component7 = ProjectFactory(creator=self.user, title='Seven')
+        component7.add_contributor(userA, auth=Auth(self.user), permissions=['read'])
+
+        component8 = ProjectFactory(creator=self.user, title='Eight')
+        component8.add_contributor(userA, auth=Auth(self.user), permissions=['read'])
+        component8.add_contributor(userB, auth=Auth(self.user), permissions=['read'])
+
+        component9 = ProjectFactory(creator=self.user, title='Nine')
+        component9.add_contributor(userB, auth=Auth(self.user), permissions=['read'])
+
+        project1.add_pointer(component2, Auth(self.user))
+        project1.nodes.append(component4)
+        project1.nodes.append(component7)
+        component2.nodes.append(component3)
+        component4.nodes.append(component5)
+        component5.nodes.append(component6)
+        component7.nodes.append(component8)
+        component7.nodes.append(component9)
+        project1.save()
+        component2.save()
+        component3.save()
+        component4.save()
+        component5.save()
+        component6.save()
+        component7.save()
+        component8.save()
+
+        url = project1.api_url_for('get_readable_descendants')
+
+        res = self.app.get(url, auth=userA.auth)
+        assert_equal(len(res.json['nodes']), 3)
+        for node in res.json['nodes']:
+            assert_in(node['title'], ['Two', 'Six', 'Seven'])
+
+        res = self.app.get(url, auth=userB.auth)
+        assert_equal(len(res.json['nodes']), 3)
+        for node in res.json['nodes']:
+            assert_in(node['title'], ['Four', 'Eight', 'Nine'])
+
     def test_get_readable_descendants_filter_for_permissions(self):
         # self.user has admin access to this project
         project = ProjectFactory(creator=self.user)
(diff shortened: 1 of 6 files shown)


TomBaxter:
Though this test has failed twice in Travis....
FAIL: test_merge (tests.models.test_user.TestUserMerging)
It is passing for me locally and has no connection that I can see with this PR.

tests/test_views.py
```diff
@@ -2848,25 +2848,6 @@ def _make_pointer_only_user_can_see(self, user, project, save=False):
         node = ProjectFactory(creator=user)
         project.add_pointer(node, auth=Auth(user=user), save=save)
 
-    def test_pointer_list_write_contributor_can_remove_private_component_entry(self):
```
mfraezz: IIRC, this test existed to prevent an edge case where a pointer was created, but the only people with permission to view it were removed from the project, leading to a state where a pointer could not be deleted from the node. I'm not sure if recent changes have obviated this test, but it might still be important.

tests/test_views.py
```diff
@@ -2903,8 +2884,9 @@ def test_pointer_list_read_contributor_cannot_remove_private_component_entry(sel
     def test_pointer_list_read_contributor_cannot_remove_public_component_entry(self):
         url = web_url_for('view_project', pid=self.project._id)
 
-        self.project.add_pointer(ProjectFactory(creator=self.user),
-                                 auth=Auth(user=self.user))
+        linked_to = ProjectFactory(creator=self.user)
+        linked_to.update(fields={'is_public':1})
```
mfraezz: You should be able to specify arbitrary fields when using factories, as though you were instantiating an object of that type -- these three lines would be equivalent to:

```python
self.project.add_pointer(ProjectFactory(creator=self.user, is_public=True),
                         auth=Auth(user=self.user))
```

That said, this isn't a blocker

mfraezz:
>Though this test has failed twice in Travis....
FAIL: test_merge (tests.models.test_user.TestUserMerging)
It is passing for me locally and has no connection that I can see with this PR.

Travis doesn't test just your code, it tests the code resultant after a merge into the targeted branch. See if that test breaks after a merge/rebase, there could be breaking but non-conflicting changes upstream.

tests/test_views.py
TomBaxter: Gave it a go, but got....

```
  File "/home/tkb/work/venv/cos/osf/osf.io/tests/test_views.py", line 2891, in test_pointer_list_read_contributor_cannot_remove_public_component_entry
    self.project.add_pointer(ProjectFactory(creator=self.user), is_public=True, auth=Auth(user=self.user))
TypeError: add_pointer() got an unexpected keyword argument 'is_public'

```

  Maybe factory needs some help.

tests/test_views.py
TomBaxter: I'm not sure where to go with this. This PR has removed most, possibly all, visibility to "Private" links/components. My understanding is that a user without at least read permissions should not know about "Private" links/components, even their existence.
  This policy may be somewhat academic, but it has a practical advantage for me in this PR as I can skip "Private" breaks in the tree and show previously hidden sub-trees. I've updated the components widget to match the tree view.
  As this PR stands, I won't be able to make this test pass. Maybe there needs to be a discussion about how we want to handle this edge case. It seems like it might be a security/user experience question. Who can delete a component/link when no one has permissions to it, and where in the interface it should happen.

TomBaxter:
I've found an issue while testing and have changed this PR to [WIP]

mfraezz:
>I've found an issue while testing and have changed this PR to [WIP]

Moving this `Additional Development` until that's resolved, then

TomBaxter:
Sorry for the confusion. It turned out to be a super quick fix. Ready for further review.

TomBaxter:
During scrum, I discussed with Brian and Nici the issue with the test I removed. test_pointer_list_write_contributor_can_remove_private_component_entry. He and Nici believe that the edge case of no one being able to delete an orphaned component was fixed in another PR.

brianjgeiger:
So @mfraezz found a problem where if you have a pointer to a node with a child you can read, and you retain access to the child but lose access to the node that the pointer points to, then you get the child node displayed. You shouldn't do this. Check for access to the node that pointers point to and, if you don't have permissions for that node, then don't check for orphaned children. The v2 api won't return pointers that you don't have access to, but v1 is not as limited.

TomBaxter:
@brianjgeiger @mfraezz Have updated logic to properly handle links. Also tested VOLs.

tests/test_views.py
mfraezz: Would https://github.com/CenterForOpenScience/osf.io/pull/5860/commits/c3b4a96d779e4f61ab0a5037c08bf87ec6c43794 make this pass again?

TomBaxter: For posterity. Per a meeting with @mfraezz @brianjgeiger and I, it was determined that in the component widget, we would not admit in any way that a private link exists if the user does not have read access to the node referenced by that link.

  As long as that is the case this test can't pass. Unless there is an alternative target for the test.

mfraezz: I spoke with product about it, and the desired functionality is to maintain current behavior in the components widget for project admins only, so that they can still remove links that no one can see. No other users should be able to see the link unless they can view the node it points to.

TomBaxter:
Visibility to links without read access added back in. Ready for review.

website/project/views/node.py
```diff
@@ -915,6 +915,21 @@ def get_summary(auth, node, **kwargs):
...
+def get_readable_descendants(auth, node, **kwargs):
+    descendants = []
+    for child in node.nodes:
+        if child.is_deleted:
+            continue
+        elif child.can_view(auth):
+            descendants.append(child)
+        else:
+            if not child.primary:
+                if node.has_permission(auth.user, 'read'):
```
mfraezz: This should be `'write'`, because you need `write` access or higher to remove a `Private link`.

Nitpick: why not do `elif not child.primary and node.has_permission(auth.user, 'read'):`?

website/templates/project/project.mako
```diff
@@ -403,7 +403,7 @@
                 <div id="containment">
                     <div mod-meta='{
                         "tpl": "util/render_nodes.mako",
-                        "uri": "${node["api_url"]}get_children/",
```
mfraezz: Was `/get_children/` being hit anywhere else? A quick grep seems to indicate that it's not.

If not, consider removing that route, method, and any tests for it. Not a blocker

TomBaxter: Done.

website/project/views/node.py
TomBaxter: Done.

website/project/views/node.py
```diff
@@ -915,25 +915,25 @@ def get_summary(auth, node, **kwargs):
...
+        if request.args.get('permissions'):
+            perm = request.args['permissions'].lower().strip()
+            if perm not in child.get_permissions(auth.user):
+                continue
+        if child.is_deleted:
+            continue
+        elif child.can_view(auth):
+            descendants.append(child)
+        elif not child.primary:
+                if node.has_permission(auth.user, 'write'):
```
mfraezz: This block should be dedented by 4 spaces

TomBaxter: Done.

website/util/rubeus.py
```diff
@@ -165,8 +165,20 @@ def to_hgrid(self):
...
-            if child.resolve()._id not in visited and not child.is_deleted and node.can_view(self.auth):
+            if child.is_deleted:
+                continue
+            elif not child.can_view(self.auth):
+                if child.primary:
+                    readable_descendants = child.find_readable_descendants(self.auth)
+                    for desc in readable_descendants:
+                        visited.append(desc.resolve()._id)
+                        rv.append(self._serialize_node(desc, visited=visited))
+                else:
```
mfraezz: Minor, but I think this `else` block is superfluous. It just triggers the next iteration of this loop, but if it gets hit there's nothing else that would occur this iteration anyway.

website/routes.py
```diff
@@ -1243,11 +1243,13 @@ def make_url_map(app):
             '/project/<pid>/get_summary/',
             '/project/<pid>/node/<nid>/get_summary/',
         ], 'get', project_views.node.get_summary, json_renderer),
-
+        # Route "get_children" is deprecated. Use get_readable_descendants.
         Rule([
             '/project/<pid>/get_children/',
             '/project/<pid>/node/<nid>/get_children/',
```
mfraezz: I highly doubt that any external parties are using this route, so it can probably be removed, as we're no longer using it. Not a blocker

mfraezz:
Other than those last two nitpicky comments, LGTM. Passing this ahead 🐎

website/routes.py
brianjgeiger: The comment should start `# TODO: [#OSF-6557] Route…`

website/util/rubeus.py
TomBaxter: Done.

website/routes.py
TomBaxter: Done.

TomBaxter:
Travis failure was the usual intermittent "returns false for date more than 4 years in the future"

website/project/model.py
```diff
@@ -1267,6 +1267,36 @@ def has_permission_on_children(self, user, permission):
...
+        while next_parent:
+            if next_parent.can_view(auth):
+                return next_parent
+            next_parent = next_parent.parent_node
+
+    def find_readable_descendants(self, auth):
+        """ Returns list of first descendant node readable by <user> in
+        each descendant branch.
+        """
+        ret = []
```
sloria: In general, it's better for methods like this to be generators rather than returning a list. Leave it up to the caller if they want a list (or some other type of collection).

```python
descendants = set(node.find_readable_descendants(auth)
# or
for node in node.find_readable_descendants(auth):  # we don't load up everything in memory at once
   #..
```

sloria:
Pass finished. Just one minor refactor to make: https://github.com/CenterForOpenScience/osf.io/pull/5860/files#r69187381. Once that's addressed, this will be good to merge for next improvements release.

PR #5867, opened 2016-06-20
Add permissions info to "My Projects" page info [OSF-6036]

Johnetordoff:
## Purpose

Allows user to see his/her permissions on a project by selecting the corresponding row. 

## Changes

Simple JS changes. CSS used to capitalize text.

## Side effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/OSF-6036


whit537:
Passes a manual functional test. 👍 

# Before

screen shot 2016-06-28 at 8 35 37 am

# After

screen shot 2016-06-28 at 8 10 21 am

whit537:
##### Punchlist

- [ ] rebase on develop

website/static/js/myProjects.js
```diff
@@ -1836,7 +1813,8 @@ var Information = {
                         m('[role="tabpanel"].tab-pane.active#tab-information',[
                             m('p.db-info-meta.text-muted', [
                                 m('', 'Visibility : ' + (item.attributes.public ? 'Public' : 'Private')),
-                                m('', 'Category: ' + categoryMap(item.attributes.category)),
+                                m('.capitalize', 'Category: ' + item.attributes.category),
+                                m('.capitalize', 'My permission: ' + permission),
```
whit537: Where is the `.capitalize` rule defined in our CSS? I'm having trouble finding it. It looks like it may be coming from a library we're using, like maybe Bootstrap, but in Bootstrap the class is [`.text-capitalize`](http://getbootstrap.com/css/#type-transformation).

website/static/js/myProjects.js
```diff
@@ -1810,6 +1786,7 @@ var Information = {
         }
         if (args.selected().length === 1) {
             var item = args.selected()[0].data;
+            var permission = item.attributes.current_user_permissions[item.attributes.current_user_permissions.length-1];
```
whit537: Why are we only showing the last permission?

website/static/js/myProjects.js
```diff
@@ -1777,31 +1777,7 @@ var Filters = {
...
-                case 'analysis':
-                    return 'Analysis';
-                case 'communication':
-                    return 'Communication';
-                case 'data':
-                    return 'Data';
-                case 'hypothesis':
-                    return 'Hypothesis';
-                case 'methods and measures':
-                    return 'Methods and Measures';
```
whit537: The `.capitalize` css class will capitilize "and."

https://jsfiddle.net/1u05m23o/

Do we care about that?

whit537:
Okay @Johnetordoff, [punchlist](https://github.com/CenterForOpenScience/osf.io/pull/5867#issuecomment-229036230) started. Over to you! :-)

website/static/js/myProjects.js
Johnetordoff: Because that's the highest permission the users has it's written as a list so it can be evaluated in statements like `if 'read' in permmissions:` the instead of `if permmission == 'read' or permmission == 'write' or permmission == 'admin'`  so the last element in the list will display the highest permission.

whit537: > the last element in the list will display the highest permission

👍

whit537: Cool. In that case, let's use [this idiom](http://stackoverflow.com/a/12099341) to avoid such a long line here. :-)

website/static/js/myProjects.js
Johnetordoff: I guess this is a case of two-birds-one-stone, changing to it text-capitalize stops it from capitalizing 'and', I think we're ok with that.

website/templates/_log_templates.mako
```diff
@@ -0,0 +1,941 @@
+## Built templates file. DO NOT MODIFY.
```
brianjgeiger: I'm pretty sure this shouldn't be checked in with this PR.

brianjgeiger:
Aside from removing the file that probably got added as part of a merge conflict, this looks good to me. 🐧

website/static/js/myProjects.js
sloria: We made an explicit decision to not automatically capitalize the category identifiers and instead use a mapping (https://github.com/CenterForOpenScience/osf.io/blob/459172cde60ebe0d5f9d93d6ac56c5cc2e47ebc1/website/settings/defaults.py#L209-L221). 

The identifiers are meant to be stable, whereas the human-readable representation is meant to be change-able. We may, for example, change "Methods and Measures" to "Methods".

Pedantry aside, using `.text-capitalize` is probably fine for the time being.

PR #5869, opened 2016-06-21
[Feature] Fix prereg download

Ghalko:
## Purpose

Currently the download button is broken for prereg Admin. This is because someone used a specifically unicode character (2019) in their title. In other fields this doesn't seem to cause an issue, but the title is more finicky it seems. 

## Changes

Go through the dictionary of serialized drafts and encode every instance of unicode into utf8 before writing to csv.

## Side effects

It fixes things.

## Ticket

None


Ghalko:
Replaced by #5945

PR #5874, opened 2016-06-22
[hotfix] [OSF-6393] move folders from osfstorage

felliott:
## Purpose

Moving a folder from osfstorage to an external provider reports an error, even though the move actually succeeds.  This is caused by the guid-preservation code not knowing how to look up the children of trashed osfstorage folders.  While fixing this, I also discovered that we weren't properly returning guids we did find, meaning they would still point to the old TrashedFileNode.

## Changes

* Taught `website.files.models.osfstorage.get_file_guids` to handle trashed osfstorage folders.
* Removed pass-by-reference code from `get_file_guids`.  It was buggy, unnecessary, and unused.  Both `get_file_guids` implementations now just return lists of guids found.
* Added tests for `osfstorage.get_file_guids`

## Side effects

None expected.

## Ticket

[#OSF-6393] https://openscience.atlassian.net/browse/OSF-6393

### DIFF
--- a/website/addons/osfstorage/tests/test_models.py
+++ b/website/addons/osfstorage/tests/test_models.py
@@ -430,6 +430,88 @@ class TestOsfstorageFileNode(StorageTestCase):
             '/'+folder._id, provider='osfstorage', node=node)
         assert [] == all_guids
 
+    def test_get_file_guids_for_live_folder_recursive(self):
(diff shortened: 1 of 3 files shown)


sloria:
Looks good. Will merge when Travis passes.

PR #5877, opened 2016-06-22
Multiple File Check Out/In via APIv2 and Files Page [#OSF-5108]

erinspace:
## Purpose

Previously, efficient multiple file check in and check out was blocked by the lack of bulk API requests. Now that bulk API requests are possible, this PR adds bulk file operations functionality so that files can be multi-selected and checked in or out.

## Changes
- Update the files serializer to handle bulk updates through the API
- Add a new route to access files lists from the API independent of nodes
- Add a new files list view to apiv2 that can only access metadata for files stored in OSF storage
    - this was to get around update complications for waterbutler, and to simplify updating a file's metadata only via the API
- Add tests for bulk updating checkout metadata via this files route
- Add multiple checkout buttons and functions to fangorn using the new bulk-enabled files list route

## Side effects
- a files list can now be accessed through the nodes route and files route.

## Questions for CR:
- The route allows entering any provider to get the file lists, but updating metadata this way should really only be for files in osfstorage (esp since you can only check out files from osfstorage). Should I make this route more restrictive, or leave it open to future changes that would allow for metadata changes via the API directly instead of thru waterbutler?

## Ticket
https://openscience.atlassian.net/browse/OSF-5108

### DIFF
--- a/api/files/serializers.py
+++ b/api/files/serializers.py
@@ -246,6 +246,8 @@ class FileSerializer(JSONAPISerializer):
 
         for attr, value in validated_data.items():
             if attr == 'checkout':
+                if value == '':
+                    value = None
                 user = self.context['request'].user
                 instance.check_in_or_out(user, value)
             else:
--- a/api/files/serializers.py
+++ b/api/files/serializers.py
@@ -246,8 +246,6 @@ class FileSerializer(JSONAPISerializer):
 
         for attr, value in validated_data.items():
             if attr == 'checkout':
-                if value == '':
-                    value = None
                 user = self.context['request'].user
                 instance.check_in_or_out(user, value)
             else:
--- a/api/files/urls.py
+++ b/api/files/urls.py
@@ -6,4 +6,5 @@ urlpatterns = [
     url(r'^(?P<file_id>\w+)/$', views.FileDetail.as_view(), name=views.FileDetail.view_name),
     url(r'^(?P<file_id>\w+)/versions/$', views.FileVersionsList.as_view(), name=views.FileVersionsList.view_name),
     url(r'^(?P<file_id>\w+)/versions/(?P<version_id>\w+)/$', views.FileVersionDetail.as_view(), name=views.FileVersionDetail.view_name),
+    url(r'^(?P<node_id>\w+)/list/(?P<provider>\w+)(?P<path>/(?:.*/)?)', views.FileList.as_view(), name=views.FileList.view_name),
 ]
--- a/api/files/views.py
+++ b/api/files/views.py
@@ -333,7 +333,6 @@ class FileDetail(JSONAPIBaseView, generics.RetrieveUpdateAPIView, FileMixin):
 
 
 class FileList(JSONAPIBaseView, generics.ListAPIView, NodeMixin, ListFilterMixin, WaterButlerMixin, bulk_views.BulkUpdateJSONAPIView):
-# class FileList(JSONAPIBaseView, NodeMixin, ListFilterMixin, WaterButlerMixin, bulk_views.BulkUpdateJSONAPIView):
     """Special case for getting lists from files on a node stored with OSFStorage.
     Used for bulk file checkout, and other simple metadata updates.
     """
--- a/api/files/views.py
+++ b/api/files/views.py
@@ -15,9 +15,12 @@ from api.base.exceptions import Gone
 from api.base.permissions import PermissionWithGetter
 from api.base.utils import get_object_or_error
 from api.base.views import JSONAPIBaseView
+from api.base.filters import ListFilterMixin
+from api.base import generic_bulk_views as bulk_views
 from api.base import permissions as base_permissions
 from api.nodes.permissions import ContributorOrPublic
 from api.nodes.permissions import ReadOnlyIfRegistration
+from api.nodes.views import WaterButlerMixin, NodeMixin
 from api.files.permissions import CheckedOutOrAdmin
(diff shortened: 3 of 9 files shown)


api/files/urls.py
```diff
@@ -6,4 +6,5 @@
     url(r'^(?P<file_id>\w+)/$', views.FileDetail.as_view(), name=views.FileDetail.view_name),
     url(r'^(?P<file_id>\w+)/versions/$', views.FileVersionsList.as_view(), name=views.FileVersionsList.view_name),
     url(r'^(?P<file_id>\w+)/versions/(?P<version_id>\w+)/$', views.FileVersionDetail.as_view(), name=views.FileVersionDetail.view_name),
+    url(r'^(?P<node_id>\w+)/list/(?P<provider>\w+)(?P<path>/(?:.*/)?)', views.FileList.as_view(), name=views.FileList.view_name),
```
mfraezz: This doesn't seem to follow standard route conventions, consider discussing it with resident routing czar @brianjgeiger

api/files/serializers.py
```diff
@@ -232,6 +232,8 @@ def update(self, instance, validated_data):
 
         for attr, value in validated_data.items():
             if attr == 'checkout':
+                if value == '':
```
mfraezz: This seems like it shouldn't be done here, and might not need doing at all. 

In `def check_in_or_out`, `action = NodeLog.CHECKED_OUT if <value> else NodeLog.CHECKED_IN` should result in `CHECKED_IN` for falsey `value`s, including `''` and `None`

api/files/urls.py
brianjgeiger: it should just be `/v2/files/` not `/v2/files/list/`

website/static/js/fangorn.js
```diff
@@ -1133,6 +1133,46 @@ function doCheckout(item, checkout, showError) {
...
+            type: 'files',
+            attributes: {
+                checkout: checkout
+            }
+        });
+    });
+
+    return $osf.ajaxJSON(
+        'PUT',
+        window.contextVars.apiV2Prefix + 'files' + items[0].data.nodeUrl + 'list/osfstorage/',
```
mfraezz: Consider using the `osfHelper` `apiV2Url` to build this url

api/files/urls.py
brianjgeiger: No, it's more complicated than that, isn't it. Okay, I'll read more closely.

brianjgeiger: Ah, okay. Ideally, this should happen in `/v2/nodes/<node_id>/files/` rather than `/v2/files/<node_id>/list/`

api/files/views.py
```diff
@@ -311,6 +315,60 @@ def get_object(self):
...
+
+        return list(files_list.children)
+
+    def get_queryset(self):
+        queryset = self.get_queryset_from_request()
+
+        # If bulk request, queryset only contains contributors in request
+        if is_bulk_request(self.request):
+            file_ids = [item['id'] for item in self.request.data]
+            queryset[:] = [file for file in queryset if file._id in file_ids]
```
mfraezz: I think the `[:]` is unnecessary

api/files/views.py
```diff
@@ -311,6 +315,60 @@ def get_object(self):
         return self.get_file()
 
 
+class FileList(JSONAPIBaseView, generics.ListAPIView, NodeMixin, ListFilterMixin, WaterButlerMixin, bulk_views.BulkUpdateJSONAPIView):
+    """Special case for getting lists from files on a node stored with OSFStorage.
```
mfraezz: Please make this follow the standard template for browsable docs

mfraezz:
>The route allows entering any provider to get the file lists, but updating metadata this way should really only be for files in osfstorage (esp since you can only check out files from osfstorage). Should I make this route more restrictive, or leave it open to future changes that would allow for metadata changes via the API directly instead of thru waterbutler?

The route shouldn't be changed for _this_ purpose, but the API should do some validation and raise an error if a user tries to check out a file in a provider that doesn't support it. I spoke briefly about this with @brianjgeiger; a `409 Conflict` would probably be appropriate.

mfraezz:
Pass complete, back to you @erinspace 🚀

api/files/urls.py
erinspace: Oh! @brianjgeiger  `/v2/nodes/<node_id>/files/` already seems to be `NodeProvidersList` ... as this was supposed to just be for OSF Storage, should it be something like `/v2/nodes/<node_id>/files/osfstorage` ??

brianjgeiger: Yes, sorry. The File List view should handle the bulk PATCH request, rather than a root file view.

api/files/serializers.py
erinspace: The problem is rather 2 lines down, `self.checkout = checkout` results in `
TypeError: Value must be StoredObject, tuple, or None`  if "checkout" is passed through as `''` - I'll change where this conversion happens though for that line instead of in the serializer

api/files/views.py
erinspace: Just to document, checked with @brianjgeiger and he said to remove docstrings as we don't document bulk API routes just yet

api/files/urls.py
mfraezz: If I'm understanding this correctly, this bulk feature should go into the existing `NodeFilesList` view, not a new one... which is a shame, because manual testing revealed this to work perfectly.

PR #5878, opened 2016-06-22
Invoke 0.13.0 Update 

caspinelli:
## Purpose

- Update all task files for 0.13.0 compatibility

## Changes

- Updated all invoke tasks to take 'ctx' as first argument
- In invoke tasks, replaced all mentions of run with ctx.run
- Bumped version number of invoke to 0.13.0 in requirements.txt

## Side effects

- Tasks no longer compatible with earlier versions of invoke

## Ticket

https://openscience.atlassian.net/browse/OSF-6492

### DIFF
--- a/scripts/populate_institutions.py
+++ b/scripts/populate_institutions.py
@@ -120,16 +120,16 @@ def main(env):
                 'domains': ['osf.usc.edu'],
                 'email_domains': [],
             },
-            # {
-            #     '_id': 'uva',
-            #     'name': 'University of Virginia',
-            #     'description': 'Projects must abide by the University <a href="http://www.virginia.edu/informationpolicy/security.html">Security and Data Protection Policies</a>',
-            #     'banner_name': 'uva-banner.png',
-            #     'logo_name': 'uva-shield.png',
-            #     'auth_url': SHIBBOLETH_SP.format(encode_uri_component('urn:mace:incommon:virginia.edu')),
-            #     'domains': ['osf.virginia.edu'],
-            #     'email_domains': [],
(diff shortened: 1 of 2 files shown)


tasks/__init__.py
```diff
@@ -346,93 +346,93 @@ def sharejs(host=None, port=None, db_url=None, cors_allow_origin=None):
...
-def rebuild_search():
+def rebuild_search(ctx):
     """Delete and recreate the index for elasticsearch"""
-    run('curl -s -XDELETE {uri}/{index}*'.format(uri=settings.ELASTIC_URI,
+    ctx.run('curl -s -XDELETE {uri}/{index}*'.format(uri=settings.ELASTIC_URI,
                                              index=settings.ELASTIC_INDEX))
-    run('curl -s -XPUT {uri}/{index}'.format(uri=settings.ELASTIC_URI,
+    ctx.run('curl -s -XPUT {uri}/{index}'.format(uri=settings.ELASTIC_URI,
                                           index=settings.ELASTIC_INDEX))
     migrate_search()
```
sloria: You will need to pass `ctx` to this function and any other task that is called like this, since `ctx` is a required argument.

tasks/__init__.py
```diff
@@ -558,7 +558,7 @@ def test_travis_osf():
     test_osf()
```
sloria: See previous comment re: passing context.

sloria:
Please go through the tasks and make sure that you are passing `ctx` to any task called within another task.

sloria:
Pass finished.

PR #5880, opened 2016-06-22
Hotfix/Remove devModeControls when not in dev mode

brianjgeiger:
## Purpose

Removes the devModeControls div when not in master so the previous fix will apply

## Changes

Removed the else branch on the mako template that added an empty div. Tested by manually removing all divs to ensure that nothing was loaded and removed the file to make sure file didn't 404 if the div didn't exist.

## Side effects

No, should be good.

## Ticket

https://openscience.atlassian.net/browse/OSF-6535


brianjgeiger:
Cancelled tests

PR #5884, opened 2016-06-23
[Feature] 'create component' modal on my project and project overview [OSF-6443]

leb2dg:
## Purpose

Ensure that both My Projects and the project overview page are using the same UI for component creation.

screen shot 2016-06-01 at 10 56 47 am
screen shot 2016-06-01 at 10 57 02 am

## Changes

* use `addProjectPlugin` on the project overview page
* change category selection on modal to dropdown
* add checkbox on modal for `Add Contributors from <parent name>`
* add check for `inherit_contributors` query param on `NodeSerializer`

screen shot 2016-06-23 at 3 27 26 pm

## Side effects

Logs do not show that a contributor was added to the project if `Add Contributors from <parent name>` is selected on creation.

## Ticket

https://openscience.atlassian.net/browse/OSF-6443


api/nodes/serializers.py
```diff
@@ -267,6 +267,18 @@ def create(self, validated_data):
             node.save()
         except ValidationValueError as e:
             raise InvalidModelValueError(detail=e.message)
+        if request.GET.get('inherit_contributors') and validated_data['parent'].has_permission(user, 'write'):
```
sloria: Minor: `?inherit_contributors=false` would actually result in contributors getting inherited. You can fix this by using `api.base.utils.is_truthy`.

sloria: How does `validated_data['parent'].has_permission` work? Isn't `validated_data['parent']` a primary key? Or does the node get loaded somewhere?

website/static/js/pages/project-dashboard-page.js
```diff
@@ -93,6 +94,32 @@ var institutionLogos = {
     }
 };
 
+/* Send with ajax calls to work with api2 */
+var xhrconfig = function (xhr) {
```
sloria: We recently added a utility function to generate xhr config functions for APIv2. If you pull from the latest develop, you'll be able to use `mithrilHelpers.apiV2Config`.

website/static/js/addProjectPlugin.js
```diff
@@ -82,7 +86,7 @@ var AddProject = {
             var data;
             self.viewState('processing');
             if(self.options.parentID) {
-                url = $osf.apiV2Url('nodes/' + self.options.parentID + '/children/', { query : {}});
+                url = $osf.apiV2Url('nodes/' + self.options.parentID + '/children/', { query : {'inherit_contributors' : self.newProjectInheritContribs()}});
```
sloria: See https://github.com/CenterForOpenScience/osf.io/pull/5884/files#r68627316 . I think this will result in contributors being inherited, no matter what value the user inputs here.

website/static/js/myProjects.js
```diff
@@ -1635,7 +1635,9 @@ var Breadcrumbs = {
                                     });
                                 },
                                 trackingCategory: 'myProjects',
-                                trackingAction: 'add-component'
+                                trackingAction: 'add-component',
+                                contributors: Array.from(linkObject.data.contributorSet),
```
sloria: Can you explain why `Array.from` is necessary here? What data type is `contributorSet`?

website/static/js/pages/project-dashboard-page.js
```diff
@@ -93,6 +94,32 @@ var institutionLogos = {
...
+    xhr.withCredentials = window.contextVars.isOnRootDomain,
+    xhr.setRequestHeader('Content-Type', 'application/vnd.api+json;');
+    xhr.setRequestHeader('Accept', 'application/vnd.api+json; ext=bulk');
+};
+var categoryList = [];
+// Load categories to pass in to create project
+var loadCategories = function _loadCategories () {
```
sloria: Minor: Rather than pushing to a "global" array (`categoryList`), it would be cleaner for `loadCategories` to return a promise that resolves to a list of categories. The resulting API would look like:

```js
loadCategories().then(function(categoriesList) {
    // do stuff with categoriesList
})
```

Mithril has a m.deferred object that would be useful for this: http://mithril.js.org/mithril.deferred.html .

Certainly not a blocker, though.

website/static/js/myProjects.js
leb2dg: linkObject.data.contributorSet
Set {"2edvq", "zfepy", "yr8n3", "9rtbc", "wpdcu"}
typeof(linkObject.data.contributorSet)
"object"

sloria:
Pass finished.

leb2dg:
Elastic search tests are failing so I don't think related to this PR.

api_tests/nodes/views/test_node_list.py
```diff
@@ -632,6 +632,28 @@ def test_creates_project_creates_project_and_sanitizes_html(self):
         assert_equal(res.json['data']['attributes']['description'], strip_html(description))
         assert_equal(res.json['data']['attributes']['category'], self.category)
 
+    def test_create_component_inherit_contributors(self):
+        parent_project = ProjectFactory(creator=self.user_one)
+        parent_project.add_contributor(self.user_two, permissions=[permissions.READ], save=True)
+        url = '/{}nodes/{}/children/{}'.format(API_BASE, parent_project._id, '?inherit_contributors=true')
```
sloria: Very minor: no need to str.format with a string literal (`'?inherit_contributors=true'`). Ain't no thang, though

scripts/populate_institutions.py
```diff
@@ -91,6 +91,16 @@ def main(env):
                 'email_domains': [],
             },
             {
+                '_id': 'nyu',
```
sloria: Not sure how these unrelated changes made it in. Was this PR branched off of master perchance?

website/static/js/myProjects.js
sloria: Got it; thanks for the clarification.

website/static/js/pages/project-dashboard-page.js
```diff
@@ -93,6 +95,29 @@ var institutionLogos = {
...
+            if(results.actions && results.actions.POST.category){
+                var categoryList = results.actions.POST.category.choices;
+                categoryList.sort(function(a, b){ // Quick alphabetical sorting
+                    if(a.display_name < b.display_name) return -1;
+                    if(a.display_name > b.display_name) return 1;
+                    return 0;
+                });
+                deferred.resolve(categoryList);
+            }
+            deferred.reject(message);
```
sloria: Minor: It might be useful to distinguish this "error" from the one a few lines down. We should never get to this point, so logging with Raven would be useful as well.

```js
errorMsg = 'API returned as success response, but no categories were fetched.';
Raven.captureMessage(message);
deferred.reject(errorMsg);
```

website/static/js/pages/project-dashboard-page.js
```diff
@@ -213,22 +238,26 @@ $(document).ready(function () {
...
+            buttonTemplate: m('.btn.btn-sm.btn-default[data-toggle="modal"][data-target="#addSubComponent"]', {onclick: function() {
+                $osf.trackClick('project-dashboard', 'add-component', 'open-add-project-modal');
+            }}, 'Add Component'),
+            modalID: 'addSubComponent',
+            title: 'Create new component',
+            parentID: window.contextVars.node.id,
+            parentTitle: window.contextVars.node.title,
+            categoryList: categoryList,
+            stayCallback: function () {
+                var ap = this; // AddProject controller
```
sloria: Is this necessary? Can `stayCallback` be set to `null`?

scripts/populate_institutions.py
sloria: Don't worry about this for now. I'll do some plumbing before I merge.

leb2dg: Shoot. It looks like it was. I can create a new branch.

sloria:
Looks good. I'll take care of the remaining changes.

sloria:
Also, we should remove any of the old create component modal code.

I'll take a look a look at this when I merge, too.

website/static/js/pages/project-dashboard-page.js
```diff
@@ -213,22 +238,26 @@ $(document).ready(function () {
...
+            },
+            trackingCategory: 'project-dashboard',
+            trackingAction: 'add-component',
+            contributors: window.contextVars.node.contributors,
+            currentUserCanEdit: window.contextVars.currentUser.canEdit
+        });
 
-    $('#newComponent').on('hidden.bs.modal', function(){
-        $('#newComponent .modal-alert').text('');
+        m.mount(document.getElementById('newComponent'), m.component(addProjectTemplate, {wrapperSelector : '#addSubComponent'}));
```
sloria: Rendering the button after we load the categories results in the button being flashed on screen:

button

This flash shouldn't be necessary. The user has to go through a few clicks to get to the Category dropdown, which should provide enough time for the OPTIONS request to finish.

I suggest we either

1. Render the button immediately, then send the OPTIONS request on page load, or
2. Render the button immediately, then send the OPTIONS request when the user clicks the "Add component" button.

This isn't necessarily a blocker to getting this merged, but it's worth looking into. @leb2dg Can you take a look?

website/static/js/pages/project-dashboard-page.js
sloria: Actually, I think we do need to refresh the page here; otherwise, the newly created component won't be displayed.

website/static/js/pages/project-dashboard-page.js
leb2dg: Yeah, I thought about that. Unfortunately, the categoryList is passed into the AddProjectPlugin which creates the button. I can play with it a little though and see if I can get the button to render when the page loads.

scripts/populate_institutions.py
sloria: Don't worry about. I'm about to create a new PR with the clean up.

sloria:
Would like to see https://github.com/CenterForOpenScience/osf.io/pull/5884#discussion_r68782106 addressed before this is merged, but it's not a hard blocker.

PR #5890, opened 2016-06-24
Feature/root dash

zamattiac:
## Purpose

Add parent and root project names to the dashboard

## Changes

- Embeds those names in the API request for each project 
- Handles private projects
- Notes if parent and root are not directly touching in the project tree

## Side effects

- Already tested for special characters
- Already tested for same named parents/children
- Already tested for periods in name
- Any failed request becomes an empty string, and strings are compared to see if projects match. Unforseen failures may result in a positive match.

## Ticket

https://openscience.atlassian.net/browse/OSF-6027

### DIFF
--- a/website/static/js/home-page/quickProjectSearchPlugin.js
+++ b/website/static/js/home-page/quickProjectSearchPlugin.js
@@ -546,9 +546,8 @@ var QuickSearchProject = {
     }
 };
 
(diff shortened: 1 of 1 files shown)


sloria:
I fixed the jshint error when I merged. Merged into `release/0.78.0`. 🐙

PR #5825, opened 2016-06-10
[feature] Project Log Feed [OSF-5980] [OSF-6454] [OSF-4582] [OSF-6386]

caseyrollins:
#### Purpose
- Updates the Recent Activity widget on the project view page to use the new log feed that appears on the My Projects page. Additionally, the logs for the project and its components are now retrieved from API v2 endpoints. 

#### Changes
- Updated log feed UI. 
- Implemented pagination on the project page log feed.
- Logs on the view project page retrieved from API v2. 
- Added a specific log text file for anonymized logs.

#### Side Effects
- The 'My Projects' page and its Activity Logs widget should be unaffected by this PR, but the page should still be tested as a minor change was made to the file that renders the log text.
- This PR makes the OSF rich in dead log code (which could be removed now in this PR, or later as part of [a maintenance/improvement JIRA ticket](https://openscience.atlassian.net/browse/OSF-6491)). 

#### QA Notes
- Everywhere on the OSF where logs for a project/component/registration are shown should be tested. Noticeable errors that would definitely be related to this PR would include receiving an endless loading icon or an 'Unable to retrieve logs' message -- there would also be console errors in both of these cases. 
- Places to check:
 - project overview page (for both projects and registrations -- with and without anonymous VOLs)
 - components widget on the project overview page, the 'Recent Activity' dropdown for components (public and private) and links (public and private)
 - profile page, 'Recent Activity' dropdown for listed public components and public projects
 - registrations tab for a project, the 'Recent Activity' dropdown for the listed registrations (pending, archiving, withdrawn)

#### Tickets
- [OSF-5980](https://openscience.atlassian.net/browse/OSF-5980)
- [OSF-6454](https://openscience.atlassian.net/browse/OSF-6454)
- [OSF-4582](https://openscience.atlassian.net/browse/OSF-4582)
- [OSF-6386](https://openscience.atlassian.net/browse/OSF-6386)

#### Screenshots
**- Recent Activity -- Project -- Before**

screen shot 2016-06-10 at 2 11 24 pm

**- Recent Activity -- Project -- After**

screen shot 2016-06-15 at 10 52 51 am

**- Recent Activity -- Component -- Before**
screen shot 2016-06-10 at 2 12 12 pm

**- Recent Activity -- Component -- After**
screen shot 2016-06-15 at 10 53 06 am

**- Recent Activity -- Anonymous VOL -- Before**

screen shot 2016-06-10 at 2 12 28 pm

**- Recent Activity -- Anonymous VOL -- After**

[...]

PR #5895, opened 2016-06-27
[feature] Remove Dead Log Code [OSF-6491]

caseyrollins:
#### Purpose
- Removes log code that is now obsolete (following #5825).

#### Changes
- Removes unused log templates as well as v1 endpoints and their associated view functions and tests.

#### Ticket
- [OSF-6491](https://openscience.atlassian.net/browse/OSF-6491)


sloria:
This is better than Christmas.

PR #5897, opened 2016-06-28
Permissions for node_links detail different from permissions for list [#OSF-6540]

chennan47:
## Purpose

Make the node_link detail consistent with the node_link api
## Changes

Now on node_link detail api even if you don't have access you should be able to see the serialized infomation.
## Ticket

https://openscience.atlassian.net/browse/OSF-6540


api_tests/nodes/views/test_node_links_detail.py
```diff
@@ -55,8 +55,8 @@ def test_returns_public_node_pointer_detail_logged_in(self):
 
     def test_returns_private_node_pointer_detail_logged_out(self):
         res = self.app.get(self.private_url, expect_errors=True)
-        assert_equal(res.status_code, 401)
-        assert_in('detail', res.json['errors'][0])
+        assert_equal(res.status_code, 200)
+        assert_in('errors', res.body)
```
brianjgeiger: Are the errors from the embedding of the node detail information failing? If so, could you do a second test for each of these conditions where it verifies the correct error is in the correct place (embeds.target_node.errors == 'you don't have permission' or whatever)?

api_tests/nodes/views/test_node_links_detail.py
```diff
@@ -125,7 +125,7 @@ def test_cannot_delete_if_registration(self):
             pointer_id,
         )
         res = self.app.delete(url, auth=self.user.auth, expect_errors=True)
-        assert_equal(res.status_code, 403)
+        assert_equal(res.status_code, 404)
```
brianjgeiger: This should probably be a 405 Method Not Allowed, because the GET request at that URL should work, but the DELETE method at the same url shouldn't, yes? Or is it 404ing for another reason?

api_tests/nodes/views/test_node_links_detail.py
```diff
@@ -68,8 +68,8 @@ def test_returns_private_node_pointer_detail_logged_in_contributor(self):
 
     def test_returns_private_node_pointer_detail_logged_in_non_contributor(self):
         res = self.app.get(self.private_url, auth=self.user_two.auth, expect_errors=True)
-        assert_equal(res.status_code, 403)
-        assert_in('detail', res.json['errors'][0])
+        assert_equal(res.status_code, 200)
+        assert_in('errors', res.body)
```
brianjgeiger: Same comment as https://github.com/CenterForOpenScience/osf.io/pull/5897/files#diff-12e9831453ea3ad6f52e01a6fda7868aR59

api_tests/registrations/views/test_withdrawn_registrations.py
```diff
@@ -71,7 +71,8 @@ def test_cannot_access_withdrawn_files_list(self):
     def test_cannot_access_withdrawn_node_links_detail(self):
         url = '/{}registrations/{}/node_links/{}/'.format(API_BASE, self.registration._id, self.public_pointer._id)
         res = self.app.get(url, auth=self.user.auth, expect_errors=True)
-        assert_equal(res.status_code, 403)
+        assert_equal(res.status_code, 200)
+        assert_in('errors', res.body)
```
brianjgeiger: Same comment as https://github.com/CenterForOpenScience/osf.io/pull/5897/files#diff-12e9831453ea3ad6f52e01a6fda7868aR59

brianjgeiger:
Pass done. A little more detail on the tests, and possibly a new error code if appropriate. 🐧

brianjgeiger:
That's a lot of failing tests.

api/nodes/views.py
```diff
@@ -1390,8 +1390,6 @@ def get_object(self):
             self.kwargs[node_link_lookup_url_kwarg],
             'node link'
         )
-        # May raise a permission denied
-        self.check_object_permissions(self.request, node_link)
```
brianjgeiger: This doesn't seem like the proper way to do this. I think modifying what permission classes are on the view would be a better idea than removing the object permissions check altogether. Same with registrations below. We do a lot of things in permission checking, and this is a little to broad of a change.

brianjgeiger:
Pass done. 🐧

api/nodes/views.py
brianjgeiger: This still doesn't seem right. If you make a token with a osf.profile_read scope, can you access this endpoint or does it give a scope error? If you try to post to the registration view derived from this, does it prevent it or let it go through to the model?

api_tests/nodes/views/test_node_links_detail.py
brianjgeiger: This wasn't addressed.

brianjgeiger:
Tests are passing, and the output looks right for those test items, but there are some items that I commented on that weren't addressed either with a change or an explanation of why it's okay, so be sure to address those. Unit tests may have to be added (or linked to) that address situations that were concerns for the major comment. Pass done. 🐧

api_tests/nodes/views/test_node_links_detail.py
chennan47: Since it is a registration, try to get the node_link of a registration using node route should raise a 404

brianjgeiger: There should be an equivalent view. /v2/registrations/`reg_id`/node_links/`link_id`/ is a valid endpoint, but DELETEing should give a 405. We may have a bug where the link is giving you /v2/nodes/ instead of /v2/registrations/ though.

brianjgeiger: Ah, yes. L122 should be `url = '/{}registrations/{}/node_links/{}/'.format(`

pattisdr: I think the reason deleting is not currently giving a 405 is because the RegistrationNodeLinksDetail view inherits from the NodeLinksDetail view, which inherits from RetrieveDestroyAPIView.  So Destroy is implemented here, however, the permission_class "ReadOnlyIfRegistration" is probably what's giving the 403.

If you want to get a 405 I would not inherit from the NodeLinksDetail view.

api/nodes/views.py
brianjgeiger: Okay, here's my problem with this. I modified your permissions classes so that they were:

```py
    permission_classes = (
        FailPermissionCheck,
        ContributorOrPublicForPointers,
        base_permissions.TokenHasScope,
        drf_permissions.IsAuthenticatedOrReadOnly,
        ExcludeWithdrawals
    )
```
Note the one at the top. It is 
```py
class FailPermissionCheck(BasePermission):

        return False
```

Which, if I did it right, means it should never get anything but a 403 no matter what. But it doesn't, because you removed the check_object_permissions for this view. 

Most or all of the stuff you did below should be done by the permissions class. Take a look at how `ContributorOrPublicForPointers` works; it loads the node object and does checks on that.

brianjgeiger:
Other than not checking the permissions, this seems to be good. 🐧 Pass done.

api/nodes/views.py
```diff
@@ -1394,7 +1397,20 @@ def get_object(self):
             self.kwargs[node_link_lookup_url_kwarg],
             'node link'
         )
-        # May raise a permission denied
+        node = get_object_or_error(
```
brianjgeiger: Very nearly. https://github.com/CenterForOpenScience/osf.io/pull/5897#discussion_r71893066 suggested "Most or all of the stuff you did below should be done by the permissions class. Take a look at how ContributorOrPublicForPointers works; it loads the node object and does checks on that." 

Rather than keeping all of this in the get object method, it should be part of the permission classes. The example of how to do that is in ContributorOrPublicForPointers.

brianjgeiger:
So close. Move stuff back into permission classes. 🐧 Pass done.

api/nodes/views.py
```diff
@@ -1385,23 +1389,24 @@ class NodeLinksDetail(JSONAPIBaseView, generics.RetrieveDestroyAPIView, NodeMixi
...
         )
-        # May raise a permission denied
         self.check_object_permissions(self.request, node_link)
         return node_link
 
     # overrides DestroyAPIView
     def perform_destroy(self, instance):
         auth = get_user_auth(self.request)
         node = self.get_node()
+        if not node.can_edit(auth):
```
brianjgeiger: This looks like something that was in the permissions classes you have removed (isAuthenticatedOrReadOnly, readOnlyIfRegistration, etc). Anything in this PR that is related to denying permission to do an action should almost certainly be inside a permissions class, and there hopefully shouldn't be a lot of removal of existing permission classes. Please do a pass and see how much you can put back into permissions classes now that you have a better idea of how they work.

brianjgeiger:
Looks good! 🎉 🐧 🎉

PR #5898, opened 2016-06-28
[WIP] Fedora storage addon

markpatton:
## Purpose

Storage addon that allows a user to interact with a Fedora 4 repository.

Currently this is just a stub created from the dropbox addon. It's only purpose is experimentation as I try to tease out how a storage addon interacts with the rest of osf.io.

@mfraezz  @felliott

## Changes

Created the addon in the addons directory. Some other shared config files were also touched.

## Side effects

There should be no side effects.


markpatton:
Redoing pull request.

PR #5900, opened 2016-06-28
Update timezone on myProject page

hmoco:
## Purpose

Update timezone and locale for users when they visit the myProjects page.

## Changes

Use the user profile endpoint to update locale and timezone if they are different than the user's once they visit the my projects page

## Side effects
more accurate timezone and locale for our users

## Ticket

https://openscience.atlassian.net/browse/OSF-6538

### DIFF
--- a/website/static/js/pages/dashboard-page.js
+++ b/website/static/js/pages/dashboard-page.js
@@ -43,7 +43,10 @@ var ensureUserTimezone = function(savedTimezone, savedLocale, id) {
 $(document).ready(function() {
     m.mount(document.getElementById('dashboard'), m.component(MyProjects, {wrapperSelector : '#dashboard'}));
     // TODO: new data does not have timezone information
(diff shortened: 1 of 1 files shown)


website/static/js/pages/dashboard-page.js
```diff
@@ -43,7 +43,10 @@ var ensureUserTimezone = function(savedTimezone, savedLocale, id) {
 $(document).ready(function() {
     m.mount(document.getElementById('dashboard'), m.component(MyProjects, {wrapperSelector : '#dashboard'}));
     // TODO: new data does not have timezone information
```
sloria: Is this TODO no longer relevant?

sloria:
Since the user is usually redirected to the Dashboard, it would make sense to call ensureUserTimezone on the Dashboard rather than My Projects (note: the name of the page modules is confusing at the moment).

sloria:
I can make the above change before I merge.

PR #5904, opened 2016-06-28
Fix how osfToggleHeight is loaded [OSF-4265]

adlius:
## Purpose

Change how the collapse/expand caret (osfToggleHeight) is loaded.

## Changes

Changed how the caret is loaded in dataverseUserConfig.js and s3UserConfig.js since Dataverse and S3 are the only two addons that are not loaded through OAuthAddonSettingsViewModel.
For the addons that are loaded through OAuthAddonSettingsViewModel, the caret is loaded for each addon. 

## Side effects

None that I can think of.

## Ticket

https://openscience.atlassian.net/browse/OSF-4265


mfraezz:
LGTM. RTM when the build passes. ⭐

PR #5906, opened 2016-06-28
Feature/create component dialogs better button [OSF-6443]

leb2dg:
## Purpose

Fix the weird `Add Component` button.

## Changes

* add `unwrap` to mithrilHelpers
* add disabled placeholder button

## Side effects

Disabled button shows until mithril renders the actual button.

## Ticket

https://openscience.atlassian.net/browse/OSF-6443

### DIFF
--- a/website/static/js/pages/project-dashboard-page.js
+++ b/website/static/js/pages/project-dashboard-page.js
@@ -268,29 +268,6 @@ $(document).ready(function () {
         }
     });
 
-
-
-    // var addProjectTemplate = m.component(AddProject, {
-    //     buttonTemplate: m('.btn.btn-sm.btn-default[data-toggle="modal"][data-target="#addSubComponent"]', {onclick: function() {
-    //         $osf.trackClick('project-dashboard', 'add-component', 'open-add-project-modal');
-    //     }}, 'Add Component'),
-    //     modalID: 'addSubComponent',
-    //     title: 'Create new component',
-    //     parentID: window.contextVars.node.id,
-    //     parentTitle: window.contextVars.node.title,
-    //     categoryList: categoryList,
-    //     stayCallback: function() {
-    //         // We need to reload because the components list needs to be re-rendered serverside
-    //         window.location.reload();
-    //     },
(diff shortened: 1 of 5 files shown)


website/static/js/pages/project-dashboard-page.js
```diff
@@ -29,6 +29,27 @@ var node = window.contextVars.node;
...
+    stayCallback: function() {
+        // We need to reload because the components list needs to be re-rendered serverside
+        window.location.reload();
+    },
+    trackingCategory: 'project-dashboard',
+    trackingAction: 'add-component',
+    contributors: window.contextVars.node.contributors,
+    currentUserCanEdit: window.contextVars.currentUser.canEdit
+});
+m.mount(document.getElementById('newComponent'), m.component(addProjectTemplate, {wrapperSelector : '#addSubComponent'}));
```
sloria: It would still be good to do all of this in `document.ready`, in order to avoid trying to mount the component before the document has loaded.

website/static/js/tests/mithrilHelpers.test.js
```diff
@@ -0,0 +1,20 @@
+/*global describe, it, expect, example, before, after, beforeEach, afterEach, mocha, sinon*/
+'use strict';
+var assert = require('chai').assert;
+var m = require('mithril');
+var mHelpers = require('../mithrilHelpers');
+
+describe('mithrilHelpers', () => {
+    describe('unwrap', () => {
```
sloria: 👍

website/static/js/pages/project-dashboard-page.js
sloria: Also, I'm a bit confused by this outer call to `m.component(addProjectTemplate, {wrapperSelector})`. Is it necessary? Is wrapperSelector used? Could this be simplified to

```js
m.mount(..., addProjectTemplate);
```

sloria:
Turns out we can use the nodeCategories that are already in contextVars. Made this change in https://github.com/CenterForOpenScience/osf.io/commit/71e7467057a80348159e29dd53c626dd41132ab6 . 🐙

PR #5907, opened 2016-06-28
Feature/admin prereg checkout [OSF-6411]

Ghalko:
## Purpose

This replaces #5775 

It allows prereg admins to checkout files while they are reviewing a submission. It also gathers guid info to make file tracking more robust if the user changes the file between submission and review.

## Changes

- Add checkin/checkout methods to relevant admin interface views.
- Add file guid request

## Side effects

More API v1 for (hopefully soon) deprecation

## Ticket


admin/pre_reg/views.py
```diff
@@ -209,3 +222,21 @@ def view_file(request, node_id, provider, file_id):
     fp = FileNode.load(file_id)
     wb_url = fp.generate_waterbutler_url()
     return redirect(wb_url)
+
+
+def get_metadata_files(draft):
+    for q in ['q7', 'q11', 'q12', 'q13', 'q16', 'q19', 'q26']:
```
sloria: Can we avoid hardcoding these? Can you inspect the metadata (i.e. questions that have `/value/uploader/extra`) instead?

website/addons/base/views.py
```diff
@@ -643,6 +643,10 @@ def addon_view_or_download_file(auth, path, provider, **kwargs):
     if action == 'download':
         return redirect(file_node.generate_waterbutler_url(**dict(extras, direct=None, version=version.identifier)))
 
+    if action == 'get_guid':
```
sloria: This exposes an endpoint to programmatically generate GUIDs. This was implemented in https://github.com/CenterForOpenScience/osf.io/pull/5836 , but we decided to put that on hold because we didn't want to expose GUID creation quite yet.

This is a potential blocker; let's continue discussion of this on the ticket.

sloria: See https://openscience.atlassian.net/browse/OSF-6411?focusedCommentId=35797&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-35797 : Can we just create guids for the selected files and check those out?

sloria:
Pass finished.

admin/pre_reg/views.py
Ghalko: Yes we can. It is going to be a bit messy with how large and weird the metadata layout is.

admin/pre_reg/views.py
```diff
@@ -212,3 +226,42 @@ def view_file(request, node_id, provider, file_id):
...
+def get_metadata_files(draft):
+    for q in get_file_questions('prereg-prize.json'):
+        for file_info in draft.registration_metadata[q]['value']['uploader']['extra']:
+            if file_info['data']['provider'] != 'osfstorage':
+                raise Http404('File does not exist in OSFStorage: {} {}'.format(
+                    q, file_info
+                ))
+            file_guid = file_info['data'].get('fileId')
+            if file_guid is None:
+                raise Http404('File in {} does not have a guid.'.format(q))
```
sloria: Minor: A 400 is more appropriate response here, since the data are invalid.

website/static/js/registrationUtils.js
```diff
@@ -981,7 +983,7 @@ RegistrationEditor.prototype.context = function(data, $root, preview) {
     });
 
     if (this.extensions[data.type]) {
-        return new this.extensions[data.type](data, $root, preview);
+        return new this.extensions[data.type](data, $root.pk, preview);
```
sloria: Why was this changed?

website/static/js/registrationEditorExtensions.js
```diff
@@ -165,10 +165,11 @@ ko.bindingHandlers.osfUploader = {
 
 
 var uploaderCount = 0;
-var Uploader = function(question) {
+var Uploader = function(question, pk) {
```
Ghalko: This is where that `$root.pk` goes.

website/static/js/registrationUtils.js
Ghalko: The root wasn't being used by either of the functions this called. I needed to get the id to trickle down to the lower reaches of the gargantuan pile, passing just root caused errors, so I passed only part of it. https://github.com/CenterForOpenScience/osf.io/pull/5907/files#r71750654

sloria:
No major blockers here. This is good to go as is, but it'd be worth looking into https://github.com/CenterForOpenScience/osf.io/pull/5907#discussion_r71747989 when you get the chance.

admin/pre_reg/views.py
```diff
@@ -102,16 +103,22 @@ class DraftDetailView(PreregAdmin, DetailView):
...
-                DraftRegistration.load(self.kwargs.get('draft_pk'))
-            )
+            return serializers.serialize_draft_registration(draft)
         except AttributeError:
             raise Http404('{} with id "{}" not found.'.format(
                 self.context_object_name.title(),
                 self.kwargs.get('draft_pk')
             ))
 
+    def checkout_files(self, draft):
```
brianjgeiger: So I think this is what you'd want to commit and begin a new transaction in if the problem ends up being on Production. I imagine the checkin view as well.

PR #5908, opened 2016-06-28
Fix API V2 filter by size [OSF-4948]

adlius:
## Purpose

Fix the problem that API v2 is not able to filter by file size

## Changes

Add SerializerMethodIntegerField, a subclass of SerializerMethodField.
Change "size" of FileSerializer to be a SerializerMethodIntegerField

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-4948

## Note

This PR is a new version of https://github.com/CenterForOpenScience/osf.io/pull/5715, which I had to close due to some improper use of git.

### DIFF
--- a/api_tests/nodes/views/test_node_files_list.py
+++ b/api_tests/nodes/views/test_node_files_list.py
@@ -321,9 +321,9 @@ class TestNodeFilesListFiltering(ApiTestCase):
             'service': 'cloud',
             osfstorage_settings.WATERBUTLER_RESOURCE: 'osf',
         }, {
-                                     'size': 1337,
-                                     'contentType': 'img/png'
-                                 }).save()
+            'size': 1337,
+            'contentType': 'img/png'
+        }).save()
         return test_file
 
     def _create_test_file_for_size_filter(self, node, user):
@@ -336,9 +336,9 @@ class TestNodeFilesListFiltering(ApiTestCase):
             'service': 'cloud',
             osfstorage_settings.WATERBUTLER_RESOURCE: 'osf',
         }, {
-                                     'size': 123,
-                                     'contentType': 'img/png'
-                                 }).save()
+            'size': 123,
+            'contentType': 'img/png'
+        }).save()
 
         test_file_2 = root_node.append_file('test_file_2')
         test_file_2.get_guid(create=True)
@@ -347,9 +347,9 @@ class TestNodeFilesListFiltering(ApiTestCase):
             'service': 'cloud',
             osfstorage_settings.WATERBUTLER_RESOURCE: 'osf',
         }, {
-                                       'size': 456,
-                                       'contentType': 'img/png'
-                                   }).save()
+            'size': 456,
+            'contentType': 'img/png'
+        }).save()
 
         return test_file, test_file_2
 
--- a/api_tests/nodes/views/test_node_files_list.py
+++ b/api_tests/nodes/views/test_node_files_list.py
@@ -10,12 +10,12 @@ from website.addons.github.tests.factories import GitHubAccountFactory
 from website.models import Node
 from website.util import waterbutler_api_url_for
 from api.base.settings.defaults import API_BASE
-from api_tests import utils as api_utils
 from tests.base import ApiTestCase
 from tests.factories import (
     ProjectFactory,
     AuthUserFactory
 )
+from website.addons.osfstorage import settings as osfstorage_settings
 
 def prepare_mock_wb_response(
         node=None,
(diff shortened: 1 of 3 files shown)


api_tests/nodes/views/test_node_files_list.py
```diff
@@ -311,6 +311,48 @@ def tearDown(self):
...
+        osfstorage = node.get_addon('osfstorage')
+        root_node = osfstorage.get_root()
+        test_file = root_node.append_file('test_file')
+        test_file.get_guid(create=True)
+        test_file.create_version(user, {
+            'object': '06d80e',
+            'service': 'cloud',
+            osfstorage_settings.WATERBUTLER_RESOURCE: 'osf',
+        }, {
+                                     'size': 1337,
```
brianjgeiger: This formatting seems weird.

api_tests/nodes/views/test_node_files_list.py
```diff
@@ -311,6 +311,48 @@ def tearDown(self):
...
+        osfstorage = node.get_addon('osfstorage')
+        root_node = osfstorage.get_root()
+        test_file = root_node.append_file('test_file')
+        test_file.get_guid(create=True)
+        test_file.create_version(user, {
+            'object': '06d80e',
+            'service': 'cloud',
+            osfstorage_settings.WATERBUTLER_RESOURCE: 'osf',
+        }, {
+                                     'size': 123,
```
brianjgeiger: Same with this formatting. A little too much whitespace.

api_tests/nodes/views/test_node_files_list.py
```diff
@@ -311,6 +311,48 @@ def tearDown(self):
...
+                                 }).save()
+
+        test_file_2 = root_node.append_file('test_file_2')
+        test_file_2.get_guid(create=True)
+        test_file_2.create_version(user, {
+            'object': '06d80e',
+            'service': 'cloud',
+            osfstorage_settings.WATERBUTLER_RESOURCE: 'osf',
+        }, {
+                                       'size': 456,
```
brianjgeiger: And here.

brianjgeiger:
I think, aside from minor formatting issues, this should be good. If we have to do many more variations of the filterable serializer method fields then what we have would become cumbersome, but I don't think we'll need to and we can cross that bridge when we come to it. For this PR, should be great. 🐧

adlius:
Okay. I'll do the cleanup. The odd formatting seems like an issue caused by copying and pasting.

adlius:
@brianjgeiger Code cleanup for the test file is done.

brianjgeiger:
Sorry about this, should have caught it weeks ago, and hopefully there won't be a lot of scope creep here. The simple thing is ensure that the test will also filter out things that don't match; if your file came up first and the filter did nothing, then the test you have would pass. So just verify that a file that does not match the criterion does not show up in the results.

Second, numeric things should be able to use other operators than just equality. So `filter[size][gt]=100` should get every file larger than 100. I thiiiiink that should just happen, in which case all you need is a test. If the test doesn't succeed, however, then we can evaluate. I'm 90% sure it'll be fine, though.

adlius:
@brianjgeiger 
So I should add another test to verify that whatever files that does not match the size would not show up, right?
For the second thing, I just tried and "filter[size][gt]=100" doesn't seem to work. In the code it it specified that
`COMPARABLE_FIELDS = NUMERIC_FIELDS + DATE_FIELDS`
and that
`NUMERIC_FIELDS = (ser.IntegerField, ser.DecimalField, ser.FloatField)`
Simply adding the `SerializerMethodIntegerFied` to the set of `NUMERIC_FIELDS` can make it work. Should I proceed and do so? Are there any other unit tests I should include?

brianjgeiger:
Yes to adding the field to NUMERIC_FIELDS. That plus the couple of tests I proposed *should* do it.

adlius:
@brianjgeiger Okay. I have added a line in the original test to make sure the data returned contains only one file. I have also added a separate test to make sure filter by size using operators works. Also added the `SerializerMethodIntegerField` to the set of `NUMERIC_FIELDS`.

sloria:
This approach has a couple of minor design issues:

- The `SerializerMethodIntegerField` adds no behavior to its parent class and exists only for type-checking. This is antipattern in OOP, esp. in the Python language, which discourages type-checking
- We should be able to generalize the implementation of a filterable SerializerMethodField to other types

How about a `TypedSerializerMethodField` which can know about its return value type?

```python
size = TypedSerializerMethodField(return_type=int)
```

The `return_type` can be used by the filter class to cast the client's input value. You could also (optionally) use it to validate against the serializer method's return value.

```python
size = TypedSerializerMethodField(return_type=int)

def get_size(self, obj):
     # this will cause an error to be raised by TypedSerializerMethodField
    return 'not an int'  
```

sloria:
Pass finished.

PR #5910, opened 2016-06-29
Fixed icon popover in "New Token" [#OSF-6321]

ccfair:
## Purpose

  The fa-icons in the New Token Detail page in settings/tokens/create/ didn't all show information when hovering over them. 

## Changes

I changed the title in the i tag so it now shows the popover, and it goes out to the right instead of bottom so it doesn't cover the other options.

## Side effects

None that I can think of, I only changed the icon, not anything else on the page.

## Ticket

https://openscience.atlassian.net/browse/OSF-6321

### DIFF
--- a/website/templates/profile/personal_tokens_detail.mako
+++ b/website/templates/profile/personal_tokens_detail.mako
@@ -37,7 +37,7 @@
                             % for scope in scope_options:
                                 <input type="checkbox" id="${scope[0]}" value="${scope[0]}" data-bind="checked: scopes">
                                 <label for="${scope[0]}">${scope[0]} </label>
-                                <i class="fa fa-info-circle text-muted" data-bind="tooltip: {title: ${scope[1] | sjson, n }, placement: 'bottom'}"></i>
+                                <i class="fa fa-info-circle text-muted" data-bind="tooltip: {title: '${scope[1]}', placement: 'right'}"></i>
                                 <br>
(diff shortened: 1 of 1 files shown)


website/templates/profile/personal_tokens_detail.mako
```diff
@@ -37,7 +37,7 @@
                             % for scope in scope_options:
                                 <input type="checkbox" id="${scope[0]}" value="${scope[0]}" data-bind="checked: scopes">
                                 <label for="${scope[0]}">${scope[0]} </label>
-                                <i class="fa fa-info-circle text-muted" data-bind="tooltip: {title: ${scope[1] | sjson, n }, placement: 'bottom'}"></i>
+                                <i class="fa fa-info-circle text-muted" data-bind="tooltip: {title: '${scope[1]}', placement: 'right'}"></i>
```
brianjgeiger: The options ` | sjson, n ` were added for security purposes. Is that preventing it from working?

ccfair: Brian, didn't know that but I do believe so. When inspecting the page, the information is there but the popover isn't working. I think those options are preventing the functionality.

brianjgeiger: See if you can figure out why. Taking out the security features would probably be bad and *shouldn't* be necessary.

PR #5913, opened 2016-06-29
[Feature] No way to distinguish between projects in Add Links modal [OSF-4046]

el1t:
## Purpose

The _Add Links_ modal does not distinguish between projects of the same name and author, which is particularly problematic for projects that have been registered multiple times (see screenshot).

Note that registrations have the same creation and modification date as their parent project, but have a separate registration date. This means that the "last modified" date could come before the registration date, which can be very confusing. Thus, only the registration date is shown for registrations.

### Before

Original
Revision 1:
With tooltip
### After (Revision 2)
registration
project
- Added tooltip conditionally displaying date created and date modified or date registered, depending on whether or not the node is a registration
- Added `dateCreated`, `dateModified`, `isRegistration`, and `dateRegistered` to `search_node` endpoint
- Removed tooltips on **+** and **-** buttons, which caused issues when two tooltips were triggered simultaneously (and are also redundant)

## Side effects

- `search_node` returns up to three additional parameters (two for registrations)
- The two removed tooltips mentioned above
- Added class `pointer-row` to node rows

## Ticket

[https://openscience.atlassian.net/browse/OSF-4046](https://openscience.atlassian.net/browse/OSF-4046)

### DIFF
--- a/website/project/views/node.py
+++ b/website/project/views/node.py
@@ -1100,6 +1100,8 @@ def _serialize_node_search(node):
         'title': title,
         'firstAuthor': first_author.family_name or first_author.given_name or first_author.full_name,
         'etal': len(node.visible_contributors) > 1,
+        'dateCreated': node.date_created.isoformat(),
+        'dateModified': node.date_modified.isoformat()
     }
 
 
--- a/website/project/views/node.py
+++ b/website/project/views/node.py
@@ -1089,20 +1089,23 @@ def _serialize_node_search(node):
     :return: Dictionary of node data
 
     """
-    title = node.title
+    data = {
+        'id': node._id,
+        'title': node.title,
+        'etal': len(node.visible_contributors) > 1,
+        'isRegistration': node.is_registration
+    }
     if node.is_registration:
-        title += ' (registration)'
+        data['title'] += ' (registration)'
+        data['dateRegistered'] = node.registered_date.isoformat()
+    else:
+        data['dateCreated'] = node.date_created.isoformat()
+        data['dateModified'] = node.date_modified.isoformat()
 
     first_author = node.visible_contributors[0]
+    data['firstAuthor'] = first_author.family_name or first_author.given_name or first_author.full_name,
 
-    return {
-        'id': node._id,
-        'title': title,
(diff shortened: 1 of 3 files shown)


whit537:
##### Punchlist

- [ ] fix tests on Travis

sloria:
@el1t The JS that are failing have since been fixed on develop, so getting your branch up to date should fix those. The failing Python test, however, is worth looking into.

website/project/views/node.py
```diff
@@ -1089,18 +1089,23 @@ def _serialize_node_search(node):
     :return: Dictionary of node data
 
     """
-    title = node.title
+    data = {
+        'id': node._id,
+        'title': node.title,
+        'etal': len(node.visible_contributors) > 1,
```
whit537: What does "etal" mean?

website/project/views/node.py
```diff
@@ -1089,18 +1089,23 @@ def _serialize_node_search(node):
...
     if node.is_registration:
-        title += ' (registration)'
+        data['title'] += ' (registration)'
+        data['dateRegistered'] = node.registered_date.isoformat()
+    else:
+        data['dateCreated'] = node.date_created.isoformat()
+        data['dateModified'] = node.date_modified.isoformat()
 
     first_author = node.visible_contributors[0]
+    data['firstAuthor'] = first_author.family_name or first_author.given_name or first_author.full_name,
```
whit537: There's a left-over comma at the end of this line.

website/project/views/node.py
el1t: et. al :wink:

whit537:
Code lgtm, made a little [punchlist](https://github.com/CenterForOpenScience/osf.io/pull/5913#issuecomment-231353985).

website/project/views/node.py
el1t: Yep pushing changes

el1t:
If you'd prefer, I can add a comment clarifying `etal` means "et. al"

whit537:
@el1t A less clever variable name would be even better, but I'm not going to block on that.

👍  from me!

PR #5914, opened 2016-06-30
Collections: when none show help text [OSF-6532]

el1t:
## Purpose

There was no help text for empty collections besides "All my projects" and "All my registrations," which left an empty box in the view. Turns out this text existed but did not show due to an oversight in using the ternary operator; adding the string to the boolean occurred first, which always returned true.

## Changes
help text

- Restored existing help text to working order
 - Altered `nonLoadTemplate` logic
- Added specific help text for bookmarks

## Side effects

None.

## Ticket

https://openscience.atlassian.net/browse/OSF-6532

### DIFF
--- a/website/static/js/myProjects.js
+++ b/website/static/js/myProjects.js
@@ -624,11 +624,11 @@ var MyProjects = {
                             'You have not made any registrations yet. Go to ',
                             m('a', {href: 'http://help.osf.io/m/registrations'}, 'Getting Started'), ' to learn how registrations work.' );
                     } else {
-                        template = 'This collection is empty.';
+                        var helpText = 'This collection is empty.';
                         if (!self.viewOnly) {
-                            template +=' To add projects or registrations, click "All my projects" or "All my registrations" in the sidebar, and then drag and drop items into the collection link.';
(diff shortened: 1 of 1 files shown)


website/static/js/myProjects.js
```diff
@@ -624,8 +624,11 @@ var MyProjects = {
                             'You have not made any registrations yet. Go to ',
                             m('a', {href: 'http://help.osf.io/m/registrations'}, 'Getting Started'), ' to learn how registrations work.' );
                     } else {
-                        template = m('.db-non-load-template.m-md.p-md.osf-box',
-                            'This collection is empty.' + self.viewOnly ? '' : ' To add projects or registrations, click "All my projects" or "All my registrations" in the sidebar, and then drag and drop items into the collection link.');
+                        template = 'This collection is empty.';
```
sloria: No need to reassign `template` here; code becomes harder to follow when variables change type and meaning.

```js
var helpText = 'This collection...';
//...
template = m('.db....', helpText);
```

sloria:
Pass finished. Just a minor change to make. Once this gets past Product Demo, this is good to merge.

PR #5916, opened 2016-06-30
#OSF-6011 Fixed bug 'Extra logs when deauthorizing addons' 

caspinelli:
## Purpose

Fix bug where, when deleting and deauthorizing addons, extra duplicate logs appear in recent activity

## Changes

Corrected code that checked if addons had external_accounts
## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-6011


brianjgeiger:
Could you create a unit test for this? Ideally, without this change, the test should fail, and with the change, the test should pass. There will be examples of tests that check for logs, and of course you mentor can give you pointers on getting started with tests if you haven't had a chance to do them yet.

PR #5917, opened 2016-06-30
Add `resolve` query param to V2 guid endpoints [OSF-6606]

samchrisinger:
## Purpose

The existing behavior of the v2 guid endpoints is to simply redirect to endpoint for the resolved guid. This is fine, but doesn't play nicely with cross-origin consumers. 

Adding a `resolve` query param changes this behavior to instead return a serialization of the guid resource like:
```json
{
    "data": {
        "relationships": {
            "referent": {
                "links": {
                    "related": {
                        "href": "https://api.osf.io/v2/nodes/ezcuj/",
                        "meta": {
                            "type": "nodes"
                        }
                    }
                }
            }
        },
        "links": {
            "self": "https://api.osf.io/v2/guids/ezcuj/",
            "html": "https://osf.io/ezcuj/"
        },
        "attributes": {},
        "type": "guids",
        "id": "ezcuj"
    }
}
```

This can be used to resolve the guid without a redirect, and will allow consumers to embed the referent resource if desired.

## Changes

- Change `GuidRedirect` to `GuidDetail`
- Add `GuidSerializer`
- Add tests for serialization behavior
- Lets Files routes resolve guids, e.g. `/v2/files/<guid>/`

## Notes

Letting the Files endpoints resolve guids seems fine-- but the links in the file serialization will reference non-guid versions. IMO in the near future we're going to need to rework how we assign guids to Files (right now there's no way for API consumers to get a guid for a file that doesn't already have one {short of making a GET to that file's detail page -- not ideal}). So I'm comfortable ignoring this until we decide to spend some more serious effort improving the Files interface.

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-6606


brianjgeiger:
You can't embed a referent. If you try, you get
```
[30/Jun/2016 13:59:04] "GET /v2/guids/qjycn/?resolve=false&embed=referent HTTP/1.1" 500 59
  File "/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/wsgiref/handlers.py", line 85, in run
  File "/Users/bgeiger/Envs/osf/lib/python2.7/site-packages/django/contrib/staticfiles/handlers.py", line 63, in __call__
  File "/Users/bgeiger/Envs/osf/lib/python2.7/site-packages/django/core/handlers/wsgi.py", line 177, in __call__
  File "/Users/bgeiger/Envs/osf/lib/python2.7/site-packages/django/core/handlers/base.py", line 230, in get_response
    response = self.handle_uncaught_exception(request, resolver, sys.exc_info())
  File "/Users/bgeiger/Envs/osf/lib/python2.7/site-packages/django/core/handlers/base.py", line 149, in get_response
  File "/Users/bgeiger/Envs/osf/lib/python2.7/site-packages/django/core/handlers/base.py", line 147, in get_response
  File "/Users/bgeiger/Envs/osf/lib/python2.7/site-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/Users/bgeiger/Envs/osf/lib/python2.7/site-packages/django/views/generic/base.py", line 68, in view
  File "/Users/bgeiger/Envs/osf/lib/python2.7/site-packages/rest_framework/views.py", line 466, in dispatch
    response = self.handle_exception(exc)
  File "/Users/bgeiger/Envs/osf/lib/python2.7/site-packages/rest_framework/views.py", line 463, in dispatch
    response = handler(request, *args, **kwargs)
  File "/Users/bgeiger/cos/osf/api/guids/views.py", line 59, in get
    return super(GuidDetail, self).get(request, **kwargs)
  File "/Users/bgeiger/Envs/osf/lib/python2.7/site-packages/rest_framework/generics.py", line 210, in get
    return self.retrieve(request, *args, **kwargs)
  File "/Users/bgeiger/Envs/osf/lib/python2.7/site-packages/rest_framework/mixins.py", line 58, in retrieve
    return Response(serializer.data)
  File "/Users/bgeiger/Envs/osf/lib/python2.7/site-packages/rest_framework/serializers.py", line 503, in data
    ret = super(Serializer, self).data
  File "/Users/bgeiger/Envs/osf/lib/python2.7/site-packages/rest_framework/serializers.py", line 239, in data
    self._data = self.to_representation(self.instance)
  File "/Users/bgeiger/cos/osf/api/base/serializers.py", line 1019, in to_representation
    result = self.context['embed'][field.field_name](obj)
  File "/Users/bgeiger/cos/osf/api/base/views.py", line 40, in partial
    v, view_args, view_kwargs = field.resolve(item, field_name)
  File "/Users/bgeiger/cos/osf/api/base/serializers.py", line 384, in resolve
    lookup_url_kwarg = lookup_url_kwarg(getattr(resource, field_name))
  File "/Users/bgeiger/cos/osf/api/guids/serializers.py", line 21, in get_related_view_kwargs
    record = guid.referent
  File "/Users/bgeiger/Envs/osf/lib/python2.7/site-packages/modularodm/storedobject.py", line 95, in wrapped
  File "/Users/bgeiger/Envs/osf/lib/python2.7/site-packages/modularodm/storedobject.py", line 935, in __getattr__
    raise AttributeError(errmsg)
AttributeError: User object has no attribute referent
```

brianjgeiger:
I think it would still be appropriate to fill out the links field with `self` and `html` entries.

brianjgeiger:
Pass done. Hopefully the embedding thing won't be too hard. As internal-only, it's not strictly necessary, but I worry that it's a symptom of something that's incorrectly implemented which may cause other problems later. 🐧

samchrisinger:
> You can't embed a referent.
I'll look at this

> I think it would still be appropriate to fill out the links field with self and html entries.
It's a pretty non-trivial amount of additional hackery to get this working. Dynamic properties of RelationshipFields are sort of tricky to add. If it's important I'll add it, but does the benefit outweigh the cost?

brianjgeiger:
The links are not vital. Again, internal-only, so should be okay. If we every wanted to expose the functionality, though, I would definitely include them.

brianjgeiger:
That being said, it's the links for the guids, so it should be `"self": "https://api.osf.io/v2/guids/ezcuj/"` and `"html": "https://osf.io/ezcuj/"`, so I wouldn't expect that to be too terrible.

brianjgeiger:
I updated the description with an ideal view of the output without embeds.

brianjgeiger:
Looks awesome. 🎉 🐧 🎉

samchrisinger:
Please hold until we figure out the correct way to generate relationship urls for files.

samchrisinger:
@brianjgeiger @sloria @pattisdr 

Does it seem reasonable to allow `api/v2/files/<file_id>` to respect guids as file ids? This could be a fallback if the regular id lookup fails.

brianjgeiger:
I would check what Comments is doing, since they have to track by file guid. Though I just looked through and didn't see anything. Hmm. 

I'm not opposed to having the file lookup work with GUIDs as well; that seems like a valid use case as long as you 404 any guids that aren't files being accessed that way. Obvs, it should be efficient and it should work, but philosophically it seems correct behavior.

samchrisinger:
@brianjgeiger File comments are fetched using a filter on the node comments endpoint. Probably not a tenable solution here.

brianjgeiger:
I meant more that file comments should have a target of a file, and that file will have to have a guid to be trackable, so I would presume that the file is directly or indirectly being acquired by its guid.

samchrisinger:
AFIK it's impossible to comment on a file until after it's guid has been assigned. This is a known complication with supporting comments in ember-osf. 

Taking a stab at this... we can revert if the solution doesn't work out.

sloria:
> Does it seem reasonable to allow api/v2/files/<file_id> to respect guids as file ids? This could be a fallback if the regular id lookup fails.

Sounds fine to me.

brianjgeiger:
Sam, I think we're talking about different things and that I am not being helpful. You go forward with what you're doing. 👍

pattisdr:
> Does it seem reasonable to allow api/v2/files/ to respect guids as file ids? This could be a fallback if the regular id lookup fails.

I also agree. I see no reason why it wouldn't be allowed, and having a fallback seems like a good idea.

brianjgeiger:
Okay, this meets all my expectations. If tests pass, I'm good with it. 🎉 🐧 🎉

api/guids/views.py
```diff
@@ -33,10 +35,29 @@ class GuidRedirect(JSONAPIBaseView):
     view_category = 'guids'
     view_name = 'guid-detail'
 
+    serializer_class = GuidSerializer
+
+    @staticmethod
+    def should_resolve(request):
+        resolve = request.query_params.get('resolve')
+        return resolve is None or is_truthy(resolve)
```
sloria: Are we "resolving" by default? I was under the impression that resolving would be opt-in, but I'm fine with this as long as @brianjgeiger is good with it.

sloria: Oh, I think I was misunderstanding the meaning of resolve. 

"resolve" = redirect to the referent's canonical URL

OK, this is good.

PR #5918, opened 2016-06-30
Feature/invoke update

caspinelli:
## Purpose

Update all task files for 0.13.0 compatibility

## Changes

- Updated all invoke tasks to take 'ctx' as first argument
- In invoke tasks, replaced all mentions of run with ctx.run
- Bumped version number of invoke to 0.13.0 in requirements.txt

## Side effects

Tasks no longer compatible with earlier versions of invoke
## Ticket

https://openscience.atlassian.net/browse/OSF-6492


tasks/__init__.py
```diff
@@ -790,25 +790,25 @@ def hotfix(name, finish=False, push=False):
     print('Renaming branch...')
 
     new_branch_name = 'hotfix/{}'.format(next_patch_version)
-    run('git checkout {}'.format(name), echo=True)
-    run('git branch -m {}'.format(new_branch_name), echo=True)
+    ctx.run('git checkout {}'.format(name), echo=True)
+    ctx.run('git branch -m {}'.format(new_branch_name), echo=True)
     if finish:
         run('git flow hotfix finish {}'.format(next_patch_version), echo=True, pty=True)
```
sloria: This should be changed to `ctx.run`. You should also be able to remove the `run` import.

sloria:
In the admin `tasks.py`, remove the unused `run` import.

sloria:
Pass finished. Just a few small changes, and this is good to merge.

tasks/__init__.py
```diff
@@ -1003,14 +1003,14 @@ def set_maintenance(start=None, end=None):
...
     print('Maintenance notice up for {} to {}.'.format(state['start'], state['end']))
 
 
 @task
-def unset_maintenance():
+def unset_maintenance(ctx):
     from website.maintenance import unset_maintenance
     print('Taking down maintenance notice...')
-    unset_maintenance()
+    unset_maintenance(ctx)
```
sloria: This is not an invoke task and there should not take `ctx`.

tasks/__init__.py
```diff
@@ -1003,14 +1003,14 @@ def set_maintenance(start=None, end=None):
         invoke set_maintenance_state --start 2016-03-16T15:41:00-04:00
         invoke set_maintenance_state --end 2016-03-16T15:41:00-04:00
     """
-    set_maintenance(start, end)
+    set_maintenance(ctx, start, end)
```
sloria: This is not an invoke task; it is an imported function. Make sure that `ctx` is only added to calls to invoke tasks.

sloria:
Pass finished.

PR #5919, opened 2016-07-01
[Fix] Log user out before resend confirmation [#OSF-6615]

cslzchen:
## Purpose

Users should be logged out when they are redirected to resending email confirmation page.

## Changes

Add `collect_auth` decorator and the following check to `resend_confirmation_get` and `resend_confirmation_post`. This behavior is consistent with forgot and resend password

```
# If user is already logged in, log user out
    if auth.logged_in:
        return auth_logout(redirect_url=request.url)
```

## Other

- Use session util function `remove_session(session)` instead of directly calling `Session.remove(Q('_id', 'eq', session._id))` when remove user session upon logout.
- Improve related docstring and import orders

## Side effects

No

## Ticket

https://openscience.atlassian.net/browse/OSF-6615

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -101,7 +101,7 @@ install:
     # - travis_retry invoke encryption
 
     - travis_retry pip install --upgrade pip
-    - travis_retry pip install invoke==0.9.0
+    - travis_retry pip install invoke==0.13.0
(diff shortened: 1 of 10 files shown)


sloria:
Looks good to merge when Travis passes.

sloria:
Test failure is unrelated.

PR #5920, opened 2016-07-01
[feature] Concatenate contributor ids

sloria:
Replaces #5824 

--------------------

## Purpose
Contributors don't have guids, so it used the id of the user, beign unique for user per node. However that becomes a problem if you look for the ocntributor data universally, as missing the node information will lead to duplicate results (problems encountered in ember-osf). New ids for contributors are unique, and are concatenations of user and node ids.

## Changes
Contributor ids are now a concatenation of node id and user id, separated by `-`

## Ticket

https://openscience.atlassian.net/browse/OSF-6490


sloria:
TODO

- [ ] Update any frontend code to send correct contributor IDs.

PR #5921, opened 2016-07-01
[WIP] Feature/linked nodes for nodes and registrations [OSF-6541]

brianjgeiger:
## Purpose

Add linked_node relationship and relationship self link to nodes and registrations

## Changes

1. Refactor linked_nodes so that it's not collection specific
2. Add to nodes (with tests)
3. Add to registrations (with tests)

## Side effects

Broken in the WIP is the related_count meta serializer test. Not sure why yet.

## Ticket

https://openscience.atlassian.net/browse/OSF-6541

### DIFF
--- a/api/base/views.py
+++ b/api/base/views.py
@@ -3,13 +3,26 @@ from django.http import JsonResponse
 from rest_framework.decorators import api_view
 from rest_framework.response import Response
 from rest_framework import generics
-# from rest_framework.serializers
+from rest_framework import status
+from rest_framework import permissions as drf_permissions
+
+from framework.auth.oauth_scopes import CoreScopes
+
 from rest_framework.mixins import ListModelMixin
+from api.base import permissions as base_permissions
+from api.base.exceptions import RelationshipPostMakesNoChanges
 
 from api.users.serializers import UserSerializer
+from api.base.parsers import JSONAPIRelationshipParser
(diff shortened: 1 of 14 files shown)


api/base/views.py
```diff
@@ -136,6 +149,112 @@ def get_serializer_context(self):
...
+        self.check_object_permissions(self.request, obj)
+        return obj
+
+    def perform_destroy(self, instance):
+        data = self.request.data['data']
+        auth = get_user_auth(self.request)
+        current_pointers = {pointer.node._id: pointer for pointer in instance['data']}
+        collection = instance['self']
+        for val in data:
+            if val['id'] in current_pointers:
```
sloria: Is `id` guaranteed to be in the payload for each datum? 

I'll write a test for this.

api/base/views.py
```diff
@@ -136,6 +149,112 @@ def get_serializer_context(self):
         })
         return context
 
+
+class LinkedNodesRelationship(JSONAPIBaseView, generics.RetrieveUpdateDestroyAPIView, generics.CreateAPIView):
```
brianjgeiger: No, collections use them too. 

.:Brian J. Geiger

sloria: OK, I see. I think this is fine here then.

PR #5924, opened 2016-07-05
Add scrolling to comment boxes for long comments [#OSF-4002]

amyshi188:
## Purpose
To allow user to know how many characters they have left in the comment pane. 

## Changes

Added a character count feature that shows the user how many characters they have left. Once it count reaches 0, if the user continues to type, the number goes into the negatives but the Comment button disables. 

## Side effects

Long comments do not appear immediately when clicking Comment, you have to refresh the page to see the comment show up. However, this was a problem even before I started working on the ticket so I feel that it might have to do with the database. Sometimes when I refresh the page and open the comment pane, the already existing comments do not show up until after it loads for a couple minutes. 

## Link to JIRA ticket

https://openscience.atlassian.net/browse/OSF-4002

### DIFF
--- a/website/templates/include/comment_pane_template.mako
+++ b/website/templates/include/comment_pane_template.mako
@@ -25,14 +25,14 @@
                     <div class="form-group">
                         <span>
                             <span>Characters Remaining: </span><span data-bind="css: {disabled: submittingReply}, text: remainingLength"></span>
-                            <textarea class="form-control" placeholder="Add a comment" data-bind="value: replyContent, valueUpdate: 'input'">
+                           <textarea class="form-control" placeholder="Add a comment" data-bind="value: replyContent, valueUpdate: 'input'"></textarea>
                         </span>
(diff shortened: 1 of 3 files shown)


amyshi188:
The code still works properly. I've fixed the merge conflicts and pushed my changes.

website/templates/include/comment_pane_template.mako
```diff
@@ -24,14 +24,15 @@
                 <form class="form">
                     <div class="form-group">
                         <span>
-                            <div class="form-control atwho-input" placeholder="Add a comment" data-bind="editableHTML: {observable: replyContent, onUpdate: handleEditableUpdate}, attr: {maxlength: $root.MAXLENGTH}" contenteditable="true"></div>
```
leb2dg: At mentions won't work if the comment pane uses a `textarea` instead of a contenteditable `div`.

PR #5928, opened 2016-07-06
[hold] [feature] Password verification

sloria:
Replaces #5585 

--------------------------------------------

## Purpose

When signing up for the OSF or changing your password, there are not very strict password strength requirements at the moment. This PR will add some stricter security requirements, and also some password strength suggestions, using Dropbox's zxcvbn library https://github.com/dropbox/zxcvbn

## Changes

- Add zxcvbn library to requirements
- Add password strength bar to front page, create account login page, forgot password page and reset password page
    - relies on the strength assigned 0-4 by zxcvbn
- CSS classes for password width added to OSF helpers based on those strength numbers
- only allows passwords with strength 2-4 as calculated by zxcvbn (yellow indicator)
- On new accounts and password changes:
    - password cannot be the same as the email address
- Redo reset password form to use KO validation over WTForms
- After password reset, user is redirected to login instead of logging in automatically

With the password "bbbbbb" --
screen shot 2016-05-23 at 1 34 54 pm

With the password "password" -- 
screen shot 2016-05-23 at 1 42 49 pm

With the password "horsesrule" --
screen shot 2016-05-23 at 1 35 08 pm

With the password asdfghjkl;'
screen shot 2016-05-23 at 1 45 23 pm

With username same as your password:
screen shot 2016-05-23 at 2 04 08 pm

## Side effects
- It'd be awesome if someone could take a look at https://github.com/dropbox/zxcvbn#performance and double check that I haven't loaded the library in a way that will cause performance issues.  For example, I require it in ```koHelpers``` to add reusable custom KO validation - will this be slow?
- Advice on how these strength indicators are placed are much welcome! Are they in the way? Is the feedback placed well? Any quirks?

NOTE: This does NOT deal with password reset in the admin. 

## Tickets
https://openscience.atlassian.net/browse/OSF-3457
https://openscience.atlassian.net/browse/OSF-6514


sloria:
UI update PR is merged. This is next in line to go to staging2.

erinspace:
Hello @sloria ! #6001 ready for review!

PR #5934, opened 2016-07-07
Contributor lists for projects under "Most Popular" (dashboard) display incorrect order. [OSF-5948]

mluke93:
## Purpose

The contributors list under the "Most Popular" nodes were displaying in the incorrect order. They were ordering randomly, instead of the intended listed order. 

## Changes

Calls get_default_queryset() instead of get_queryset_from_request(). The former grabs all of the contributors from the list of contributors on a project, therefore grabbing them in the correct order.

## Side effects

n/a

## Ticket

### DIFF
--- a/api/nodes/views.py
+++ b/api/nodes/views.py
@@ -646,7 +646,7 @@ class NodeContributorsList(JSONAPIBaseView, bulk_views.BulkUpdateJSONAPIView, bu
 
     # overrides ListBulkCreateJSONAPIView, BulkUpdateJSONAPIView
     def get_queryset(self):
-        queryset = self.get_queryset_from_request()
+        queryset = self.get_default_queryset()
 
         # If bulk request, queryset only contains contributors in request
         if is_bulk_request(self.request):


mluke93:
NOTE: There was one Python test that was failing for me. However, when I pulled from develop, it was failing for me then as well. It was unrelated to what I was working on.

sloria:
@mluke93 The failing tests do look related to this change (they aren't currently failing on develop). Can you look into those?

api/nodes/views.py
```diff
@@ -646,7 +646,7 @@ def get_serializer_class(self):
 
     # overrides ListBulkCreateJSONAPIView, BulkUpdateJSONAPIView
     def get_queryset(self):
-        queryset = self.get_queryset_from_request()
+        queryset = self.get_default_queryset()
```
sloria: I'm not sure this is the correct thing to do. `get_queryset_from_request` respects query parameters (e.g. filtering); `get_default_queryset` does not. This is making the tests (correctly) fail.

sloria:
Pass finished. Please check your approach so that the right ordering is returned *and* we respect any input from the client (incl. query params).

mluke93:
Huh. The failing tests weren't failing on my end, and the test that was isn't failing here. Yeah, I'll go back and look into this again. Sorry about that.

PR #5937, opened 2016-07-07
#OSF-6011 Fixed bug 'Extra logs when deauthorizing addons'

caspinelli:
## Purpose

Fix bug where deleting an addon would create unnecessary log messages. 

## Changes

Fixed code that checked if addon that was being deleted had attribute 'external_accounts' 
Wrote unit test that checks that deleting an addon does not make any unnecessary messages in the log. Test fails without change, passes with it. 

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-6011

### DIFF
--- a/api_tests/nodes/views/test_node_logs.py
+++ b/api_tests/nodes/views/test_node_logs.py
@@ -108,8 +108,10 @@ class TestNodeLogList(ApiTestCase):
     def test_remove_addon(self):
         self.public_project.add_addon('github', auth=self.user_auth)
         assert_equal('addon_added', self.public_project.logs[OSF_LATEST].action)
+        old_log_length = len(list(self.public_project.logs))
         self.public_project.delete_addon('github', auth=self.user_auth)
         assert_equal('addon_removed', self.public_project.logs[OSF_LATEST].action)
+        assert_equal((len(list(self.public_project.logs)) - 1), old_log_length)
         res = self.app.get(self.public_url, auth=self.user.auth)
         assert_equal(res.status_code, 200)
         assert_equal(len(res.json['data']), len(self.public_project.logs))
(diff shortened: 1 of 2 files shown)


api_tests/nodes/views/test_node_logs.py
```diff
@@ -108,8 +108,10 @@ def test_project_add_remove_contributor(self):
     def test_remove_addon(self):
         self.public_project.add_addon('github', auth=self.user_auth)
         assert_equal('addon_added', self.public_project.logs[OSF_LATEST].action)
+        old_log_length = len(list(self.public_project.logs))
         self.public_project.delete_addon('github', auth=self.user_auth)
         assert_equal('addon_removed', self.public_project.logs[OSF_LATEST].action)
+        assert_equal((len(list(self.public_project.logs)) - 1), old_log_length)
```
sloria: I'm a bit confused about this test. This fix is supposed to prevent unnecessary logs from being created, but this assertion is testing that 1 log was added.

It might be worth creating a new test that specifically tests the behavior changed in this PR.

framework/addons/__init__.py
```diff
@@ -124,7 +124,8 @@ def delete_addon(self, addon_name, auth=None, _force=False):
             if self._name in addon.config.added_mandatory and not _force:
                 raise ValueError('Cannot delete mandatory add-on.')
             if hasattr(addon, 'external_account'):
-                addon.deauthorize(auth=auth)
+                if getattr(addon, 'external_account'):
```
sloria: `getattr` takes third argument, which is the default value, if the objects doesn't have the attribute. Therefore, no need to have both `hasattr` and `getattr`.

```python
if getattr(addon, 'external_account', None):
```

sloria:
Pass finished.

api_tests/nodes/views/test_node_logs.py
caspinelli: The test asserts that only one log is made, that way if extra logs are made, the test will fail
At least, that is what it is supposed to do

caspinelli: The test fails without the change and passes with it

sloria: Ah, OK. I didn't realize that we expected 1 log to be created, and that > 1 log was created before. Thanks for the clarification.

PR #5938, opened 2016-07-07
Fix bug "new folder error persists through multiple 'create folder' attempts" [#OSF-5707]

caspinelli:
## Purpose

Fixes bug where any error message shown when creating a folder persists through multiple create folder attempts and certain error messages sometimes take a while to display or may even never display at all

## Changes

Fixes code to clear helpText whenever the dismiss toolbar button is hit and also whenever a create folder attempt is made
Also makes sure error messages appear using m.redraw() after helpText is filled with the error

## Side effects
None

## Ticket

https://openscience.atlassian.net/browse/OSF-5707


mfraezz:
LGTM, and I see that this passed a product demo. Moving this ahead ⏩

PR #5940, opened 2016-07-07
Create component, selecting "import contributors" takes u… [#OSF-6517]

baylee-d:
## Purpose

When the user goes to create a new component, they have the option to add contributors from the parent project:
  
- If they don't add contributors, they stay on the main project page
- If they choose to add contributors, they're automatically redirected to the new component's collaborators page

The user experience should be the same no matter what they select.  The purpose of this PR is to make sure that the user isn't redirected to a new page if they choose to add collaborators.

## Changes

The changes include removing code that redirects the user if they choose to add collaborators from the main project.  Now the user stays on the same page no matter what they choose when making a new component.

## Side effects

There shouldn't be any side effects.

## Ticket

https://openscience.atlassian.net/browse/OSF-6517

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -4115,7 +4115,7 @@ class TestProjectCreation(OsfTestCase):
         assert_in(self.user1, child.contributors)
         assert_in(self.user2, child.contributors)
         # check redirect url
-        assert_in('/contributors/', res.location)
+        assert_not_in('/contributors/', res.location)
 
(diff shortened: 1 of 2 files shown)


brianjgeiger:
There are a bunch of failing tests, but the code change looks good. Fix the tests and move it to Product Demo for Sara to look at, then I'll do another round of code review after. Pass done. 🐧

PR #5942, opened 2016-07-08
Share files

Johnetordoff:
## Purpose

Allows user to easily share files without having to create a project.

## Changes

Creates a dropzone widget on the front page for file drops and gives each user a new type of node (is_public_files_collection) that they can upload files to.

###  "Public Files" node:
- Dropped files go here, directly to osfstorage root directory
- Given to a user when he/she registers
- Is public
- Titled "Public Files" 
- Cannot be deleted or modified in a number of ways (functions are disabled using a decorator) 
- Has specialized merge behavior in the event of a user merge. The merge is canceled and user is prompted if they have files of conflicting versions.
- Doesn't appear in ES results.
- Creates users/<uid>/public_files endpoint for api v2, whose primary use case is to provide info for user profiles.

###  The Dropzone (fig 1-2):
- Uses dropzone.js
- When a file is dropped into the Dropzone uploads the file and creates a preview with a link to the file-view-page. (fig 2)
- Only a single file can be dropped at a time, but many can be previewed uploaded one-by-one (this is a UX decision).
- Contains a clipboard button for easy copying 
- Contains link to Public Files file browser which is typical fangorn browser modified with clipboard. (fig 3)

## Side effects

Now running the api endpoint for for the file detail generates a guid.

## Ticket

https://openscience.atlassian.net/browse/RND-61

## Screen Shots

fig 1.
screen shot 2016-06-27 at 4 45 51 pm

fig 2.
screen shot 2016-06-27 at 4 44 50 pm

fig 3.
screen shot 2016-06-27 at 4 46 55 pm


api/files/views.py
```diff
@@ -320,6 +320,9 @@ def get_node(self):
 
     # overrides RetrieveAPIView
     def get_object(self):
+        if self.request.GET and self.request.GET.get('giveGuid', False):
```
mfraezz: Prefer `underscore_method` to `camelCase` for variables, as specified in the [dev docs](http://cosdev.readthedocs.io/en/latest/style_guides/python.html)

api/users/views.py
```diff
@@ -662,3 +662,20 @@ def perform_destroy(self, instance):
...
+    required_write_scopes = [CoreScopes.USERS_WRITE, CoreScopes.NODE_REGISTRATIONS_WRITE]
+
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        base_permissions.TokenHasScope,
+    )
+
+    def get_queryset(self):
+
+        return Node.find(Q('contributors', 'eq', self.get_user()._id) & Q('is_public_files_collection', 'eq', True))
```
mfraezz: So, this is for a *user*'s public *files*, but it returns a serialized node? It seems like it should behave more like the `/files/` endpoint than like the `/nodes/` endpoint.

api/users/views.py
```diff
@@ -662,3 +662,20 @@ def perform_destroy(self, instance):
             if val['id'] in current_institutions:
                 user.remove_institution(val['id'])
         user.save()
+
+class UserPublicFiles(JSONAPIBaseView, generics.ListAPIView, UserMixin, ODMFilterMixin):
+    view_name = 'user-public-files'
+    view_category = 'users'
+    serializer_class = PublicFilesSerializer
+
+    required_read_scopes = [CoreScopes.USERS_READ, CoreScopes.NODE_REGISTRATIONS_READ]
```
mfraezz: Why should reading a node containing a users public files require `NODE_REGISTRATIONS_READ`?

api/users/urls.py
```diff
@@ -13,4 +13,5 @@
     url(r'^(?P<user_id>\w+)/institutions/$', views.UserInstitutions.as_view(), name=views.UserInstitutions.view_name),
     url(r'^(?P<user_id>\w+)/registrations/$', views.UserRegistrations.as_view(), name=views.UserRegistrations.view_name),
     url(r'^(?P<user_id>\w+)/relationships/institutions/$', views.UserInstitutionsRelationship.as_view(), name=views.UserInstitutionsRelationship.view_name),
+    url(r'^(?P<user_id>\w+)/public_files/$', views.UserPublicFiles.as_view(), name=views.UserPublicFiles.view_name),
```
mfraezz: I'm not in love with this route, but that part is up to @brianjgeiger 

I'd suggest, as an alternative, something like `/files/public/<user_id>/` to return a list of serialized public files for that user, rather than returning a nonstandard serialization of a node.

api/users/views.py
mfraezz: Also, given that a user should only ever have one node that `is_public_files_collection`, with current behavior this class should probably inherit from `generics.RetrieveAPIView`, as it's only really returning one object. 

However, I think behavior may need to change, so that it serializes the list of files, rather than wrapping that list by serializing the node.

mfraezz:
Please add api tests for the public files endpoint. Depending on what the route ends up being finalized as, the blank file `api_tests/nodes/views/test_public_files_list.py` may need to be removed.

framework/auth/core.py
```diff
@@ -574,7 +574,8 @@ def register(self, username, password=None):
         self.date_confirmed = dt.datetime.utcnow()
         self.update_search()
         self.update_search_nodes()
-
+        from website.project import new_public_files_collection
```
mfraezz: Why was this imported here? To avoid a circular import?

framework/auth/core.py
```diff
@@ -574,7 +574,8 @@ def register(self, username, password=None):
...
+        new_public_files_collection(self)
         # Emit signal that a user has confirmed
         signals.user_confirmed.send(self)
```
mfraezz: It seems like a better approach might be to create the public files collection in a function that listens for this signal. There are multiple ways that user onboarding can occur (currently just this and conferences), so that might be a more future-proof implementation.

framework/auth/core.py
```diff
@@ -1026,6 +1027,12 @@ def given_name_initial(self):
         return self.given_name[0]
 
     @property
+    def public_files_node(self):
+        from website.project.model import Node
+        if self.is_registered:
+            return Node.find_one(Q('contributors', 'eq', self._id) & Q('is_public_files_collection', 'eq', True))
+        return None
+    @property
```
mfraezz: This appears to need a newline above.

scripts/migrate_public_files_nodes.py
```diff
@@ -0,0 +1,47 @@
...
+    users = get_targets()
+    for user in users:
+        try:
+            Node.find_one(Q('is_public_files_collection', 'eq', True) & Q('contributors', 'eq', user._id))
+        except NoResultsFound:
+            new_public_files_collection(user)
+
+if __name__ == '__main__':
+    init_app(set_backends=True, routes=False)
+    dry_run = 'dry' in sys.argv
```
mfraezz: I think the current convention is `--dry`, but this isn't a blocker

scripts/migrate_public_files_nodes.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def main(dry_run):
+    users = get_targets()
+    for user in users:
+        try:
+            Node.find_one(Q('is_public_files_collection', 'eq', True) & Q('contributors', 'eq', user._id))
+        except NoResultsFound:
+            new_public_files_collection(user)
+
```
mfraezz: There should be a conditional that checks `dry_run` and raises an error at the very end of the migration if so, to roll back the transaction.

scripts/migrate_public_files_nodes.py
```diff
@@ -0,0 +1,47 @@
...
+if __name__ == '__main__':
+    init_app(set_backends=True, routes=False)
+    dry_run = 'dry' in sys.argv
+
+    # Log to file
+    if not dry_run:
+        script_utils.add_file_logger(logger, __file__)
+
+    main(dry_run=dry_run)
```
mfraezz: This should be wrapped in a `TokuTransaction` to allow a run to actually be dry.

tests/factories.py
```diff
@@ -185,6 +185,11 @@ class Meta:
 class ProjectFactory(AbstractNodeFactory):
     category = 'project'
 
+class PublicFilesFactory(AbstractNodeFactory):
+    title = "Public Files"
```
mfraezz: As per the [dev docs](http://cosdev.readthedocs.io/en/latest/style_guides/python.html#general-naming-guidelines), this should use single quotes.

tests/test_models.py
```diff
@@ -4667,6 +4667,80 @@ def test_create_from_node(self):
...
+            self.project.set_title('Look at me: I\'m the title now',auth=self.auth)
+
+    def test_forking(self):
+        with assert_raises(NodeStateError):
+            fork = self.project.fork_node(auth=self.auth)
+
+    def test_add_remove_permissions(self):
+        unauthorized_user = UserFactory()
+        with assert_raises(NodeStateError):
+            fork = self.project.add_permission(unauthorized_user,'write')
```
mfraezz: Prefer constants when they're available (see `website.utils.permissions.WRITE`)

tests/test_models.py
```diff
@@ -4667,6 +4667,80 @@ def test_create_from_node(self):
...
+        assert_raises(NodeStateError, self.project.set_privacy, 'private', self.auth)
+        assert_equal(self.project.is_public,True)
+
+    def test_citations_for_public_files(self):
+        with assert_raises(NodeStateError):
+            self.project.add_citation(self.auth)
+            self.project.edit_citation(self.auth,{})
+            self.project.remove_citation(self.auth,{})
+
+    def test_user_merge_with_other_public_files_collection(self):
```
mfraezz: Should this test have any asserts? Also, should there be one like this that has naming conflicts, and checks to make sure an appropriate error is raised?

website/files/models/base.py
```diff
@@ -497,11 +497,17 @@ def save(self):
         return self.stored_object.save()
 
     def serialize(self, **kwargs):
+        if hasattr(self.get_guid(), '_stored_key'):
```
mfraezz: Why/Is this necessary?

api/users/views.py
```diff
@@ -662,3 +662,20 @@ def perform_destroy(self, instance):
             if val['id'] in current_institutions:
                 user.remove_institution(val['id'])
         user.save()
+
+class UserPublicFiles(JSONAPIBaseView, generics.ListAPIView, UserMixin, ODMFilterMixin):
+    view_name = 'user-public-files'
+    view_category = 'users'
+    serializer_class = PublicFilesSerializer
```
mfraezz: This should just be using the `FilesSerializer`, unless there's any specific behavior you need to override. `get_queryset` should be returning a list of files contained within a user's public files collection.

api/users/views.py
```diff
@@ -662,3 +662,20 @@ def perform_destroy(self, instance):
             if val['id'] in current_institutions:
                 user.remove_institution(val['id'])
         user.save()
+
+class UserPublicFiles(JSONAPIBaseView, generics.ListAPIView, UserMixin, ODMFilterMixin):
```
mfraezz: Classes that inherit from `ODMFilterMixin` must define a `get_default_odm_query`.

```python
class ODMFilterMixin(FilterMixin):
    """View mixin that adds a get_query_from_request method which converts query params
    of the form `filter[field_name]=value` into an ODM Query object.

    Subclasses must define `get_default_odm_query()`.

    Serializers that want to restrict which fields are used for filtering need to have a variable called
    filterable_fields which is a frozenset of strings representing the field names as they appear in the serialization.
    """
```

api/users/views.py
```diff
@@ -662,3 +662,20 @@ def perform_destroy(self, instance):
...
+    view_name = 'user-public-files'
+    view_category = 'users'
+    serializer_class = PublicFilesSerializer
+
+    required_read_scopes = [CoreScopes.USERS_READ, CoreScopes.NODE_REGISTRATIONS_READ]
+    required_write_scopes = [CoreScopes.USERS_WRITE, CoreScopes.NODE_REGISTRATIONS_WRITE]
+
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
```
mfraezz: You may want to define a custom permissions class that ensures authenticated users who aren't the requested user cannot write.

Also, it seems somewhat counter-intuitive that a *public* endpoint requires authentication.

website/project/model.py
```diff
@@ -1317,6 +1337,26 @@ def has_addon_on_children(self, addon):
...
         return False
 
+    def merge_public_files(self, node):
+
+        if not self.is_public_files_collection:
+            raise NodeStateError('must be Public Files collection to merge')
+
+        from website.files.models.osfstorage import OsfStorageFile
+
+        primaryUserFiles = OsfStorageFile.find(Q('node', 'eq', self) & Q('title', 'ne', 'Public Files'))
```
mfraezz: Prefer `underscore_method` to `camelCase` for variables, as specified in the [dev docs](http://cosdev.readthedocs.io/en/latest/style_guides/python.html)

website/project/__init__.py
```diff
@@ -109,6 +109,35 @@ def new_collection(title, user):
...
 
+def new_public_files_collection(user):
+    """Create a new folder project.
+
+    :param str title: Node title
+    :param User user: User object
+    :return Node: Created node
+
+    """
+    existing_public_files_collections = Node.find(
```
mfraezz: This seems like a somewhat unnecessary variable declaration. Consider instead:
```python
if Node.find(<public_files_query>).count():
    raise NodeStateError('<message>')
```
Not a blocker.

website/project/__init__.py
```diff
@@ -109,6 +109,35 @@ def new_collection(title, user):
...
+        Q('is_public_files_collection', 'eq', True) & Q('contributors', 'eq', user._id)
+    )
+
+    if existing_public_files_collections.count() > 0:
+        raise NodeStateError('Users may only have one public files collection')
+
+    title = 'Public Files'
```
mfraezz: This also seems like an unnecessary variable declaration, as it's used on one line.

website/public_files/views.py
```diff
@@ -0,0 +1,43 @@
...
+        publicFilesCollection = Node.find_one(Q('is_public_files_collection', 'eq', True) & Q('contributors', 'eq', user._id))
+    except NoResultsFound:
+        raise HTTPError(http.NOT_FOUND)
+
+    return {
+        'node':
+            {
+                'id': publicFilesCollection._id,
+                'api_url': publicFilesCollection.api_url,
+                'ownerName': publicFilesCollection.creator.fullname,
```
mfraezz: Prefer `underscore_naming` to `camelCase`.

website/static/js/filepage/revisions.js
```diff
@@ -20,6 +20,7 @@ var model = {
     hasUser: false,
     hasDate: false,
     hasHashes: false,
+    hasURL: false,
```
mfraezz: This appears to be unused.

website/static/js/home-page/quickProjectSearchPlugin.js
```diff
@@ -494,7 +494,6 @@ var QuickSearchProject = {
         if (ctrl.eligibleNodes().length === 0 && ctrl.filter() == null) {
             return m('.row',
                 m('.col-xs-12',[
-                    headerTemplate(),
```
mfraezz: Why was this removed?

website/templates/project/project_header.mako
```diff
@@ -3,10 +3,22 @@
 %>
 
 <div id="projectBanner" >
-    <header class="subhead" id="overview">
-        <nav id="projectSubnav" class="navbar osf-project-navbar" role="navigation">
-            <div class="container">
 
+
+     % if node['category_short']  == 'share window':
```
mfraezz: There doesn't appear to be anything setting a node's category to `'share window'`

website/templates/project/share_window.mako
```diff
@@ -0,0 +1,42 @@
+<%inherit file="project/project_base.mako"/>
+<%def name="title()">${node['title']} Files</%def>
+
```
mfraezz: This file doesn't appear to be `<%include/>`'d or used anywhere.

website/templates/public_files.mako
```diff
@@ -0,0 +1,55 @@
...
+		<p class="m-t-sm fg-load-message"> Loading files...  </p>
+	</div>
+</div>
+
+</%def>
+
+<%def name="javascript_bottom()">
+
+<script src="/static/vendor/citeproc-js/xmldom.js"></script>
+<script src="/static/vendor/citeproc-js/citeproc.js"></script>
```
mfraezz: This might end up being obvious during functional review, but why/does this page need `citeproc`, `MathJax`, etc?

website/views.py
```diff
@@ -166,6 +166,11 @@ def resolve_guid(guid, suffix=None):
     guid_object = Guid.load(guid)
     if guid_object:
 
+        # redirect if user tries to visit public files wiki or add contributors ect.
+        if Node.find(Q('_id', 'eq', guid_object._storage_key) & Q('is_public_files_collection', 'eq', True)):
```
mfraezz: I'm not certain that this should be happening here, or at all. This is intended to prevent a user from seeing the error (raised by `disable_for_public_files_collection`) that happens as a result of manually navigating to a route that is 1) disabled for public files and 2) hidden from the public files page UI, correct? If that's the case, it seems unnecessary to redirect them -- it should just `404`.

Also, you shouldn't use `_storage_key`; `_id` is preferred.

api/users/views.py
mfraezz: Also, is it intended that users be able to perform CRUD actions aside from just Read on public files?

website/public_files/views.py
```diff
@@ -0,0 +1,43 @@
...
+                'isPublicFilesCol': publicFilesCollection.is_public_files_collection,
+            }
+    }
+
+def view_public_files_id(uid, **kwargs):
```
mfraezz: Could you [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself) these two views up?

api/users/serializers.py
```diff
@@ -11,6 +11,29 @@
 )
 from api.base.utils import absolute_reverse
 
+class PublicFilesSerializer(JSONAPISerializer):
```
mfraezz: As documented [elsewhere](https://github.com/CenterForOpenScience/osf.io/pull/5942/files#r71531078), this class can probably go away. However, it may be better to subclass `FileSerializer` and force guid creation there instead of how it's currently done.

Johnetordoff: Yeah you might take a look at this as well, if you haven't already https://github.com/CenterForOpenScience/osf.io/pull/5996, it's on hold awaiting Jeff's go head. Forcing guid creation is a heavily discussed issue. Making it dependent on a PublicFIles endpoint, would limit spamming guid creation, but It might effect Preprints who also need to force guids. I'll inquire what there needs are.

mfraezz: It wouldn't necessarily be dependent on a PublicFiles endpoint -- a subclass of `FileSerializer` that could optionally force guid creation (assuming the user has proper permissions) wouldn't need to be limited to association with only one view. Also, subclassing would only provide benefits over adding that in `FileSerializer` if there are certain file views that we never want to allow the forced creation of guids for. I'm not sure that there are any, but it seems probable.

And it is likely that increasing guid length for files will be a blocker for this feature, but thankfully that shouldn't be hard to do. There's a `__guid_min_length__` that can be set on `GuidStoredObject`s, and it doesn't look like it would break backwards compatibility if it was suddenly bumped up.

mfraezz:
Pass complete, back to you @Johnetordoff 🔙

 I'll start functional review after the architectural concerns have been resolved.

website/static/js/home-page/quickProjectSearchPlugin.js
Johnetordoff: This was moved to PublicFilesDropZone.js

Johnetordoff:
@mfraezz, I really appreciate you the taking the time to do this. Really thorough job, good work, thanks!

tests/test_models.py
Johnetordoff: @SSJohns ping!

website/project/model.py
Johnetordoff: Underscoring for database fields of the node object as in "self.is_public_files_collection" seems to be a convention as well though. Otherwise unsure how this doesn't fit.

mfraezz: `primaryUserFiles` should be `primary_user_files`

mfraezz:
I'm going to put review of this on hold temporarily; it's blocked by an API endpoint for generating file guids, which is probably (pending further discussion) blocked by a ticket that I'm assigned to -- verifying the security of our `osf.admin` OAuth scope.

If the discussion results in the decision that we should allow an API kwarg to create guids significantly longer than 5 characters, I'll push for getting that through. In the interim, I'm going to work on resolving the blockers. 

Please continue responding to existing CR, just note that not much more will be generated until this has a clear path for being shipped :octocat:

tests/test_models.py
SSJohns: The only assert the test should need is to check that the files were properly moved over. So an assert that the appended files are in the users public files project should be enough.

And I think John did the behavior that if there were conflicting files the whole process would stop, but I wouldn't really know.

website/files/models/base.py
```diff
@@ -497,11 +497,13 @@ def save(self):
         return self.stored_object.save()
 
     def serialize(self, **kwargs):
+
         return {
             'id': self._id,
             'path': self.path,
             'name': self.name,
             'kind': self.kind,
+            'guid': self.get_guid()._id if self.get_guid() is not None else None
```
caseyrollins: Can this be simplified to ```self.get_guid()._id if self.get_guid() else None```?

api/users/serializers.py
```diff
@@ -10,7 +10,6 @@
     JSONAPISerializer, LinksField, RelationshipField, DevOnly, IDField, TypeField
 )
 from api.base.utils import absolute_reverse
-
```
caseyrollins: Add back line pls.

scripts/migrate_public_files_nodes.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python
+# encoding: utf-8
+"""Migrate Share Windows to existing users
```
caseyrollins: "Create a public files node for existing users."

scripts/migrate_public_files_nodes.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+def main(dry_run):
+    users = core.User.find()
+
+    for user in users:
+        if user.is_registered and user.public_files_node is None:
```
caseyrollins: Will a public files node be created for existing unregistered users when they register?

website/public_files/views.py
```diff
@@ -0,0 +1,34 @@
...
+from framework.auth.decorators import must_be_logged_in
+
+@must_be_logged_in
+def view_public_files(auth, **kwargs):
+    return serialize_public_files_node(auth.user.public_files_node)
+
+def view_public_files_id(uid, **kwargs):
+
+    try:
+        publicFilesNode = Node.find_one(Q('is_public_files_node', 'eq', True) & Q('contributors', 'eq', uid))
```
caseyrollins: Prefer ```underscore_case``` to ```camelCase``` for variable names.

website/routes.py
```diff
@@ -278,6 +286,22 @@ def make_url_map(app):
...
+            [
+                '/public_files/',
+            ],
+            'get',
+            public_files_views.view_public_files,
+            OsfWebRenderer('public_files.mako', trust=False),
+        ),
+        Rule(
+            [
+                '/public_files/<uid>',
```
caseyrollins: Please add a trailing ```/``` to the end of this route.

website/routes.py
```diff
@@ -56,9 +58,14 @@ def get_globals():
     OSFWebRenderer.
     """
     user = _get_current_user()
+    try:
+        public_files_id = Node.find_one(Q('contributors', 'eq', user._id) & Q('is_public_files_node', 'eq', True))._id
```
caseyrollins: Why is it necessary that ```public_files_id``` is globally accessible? For the dashboard page?

api/nodes/views.py
```diff
@@ -2961,3 +2962,27 @@ def get_parser_context(self, http_request):
         res = super(LinkedNodesList, self).get_parser_context(http_request)
         res['is_relationship'] = True
         return res
+
+class UserPublicFiles(NodeFilesList, UserMixin):
```
caseyrollins: If the api endpoint remains ```/users/<user-id>/public_files/```, then this view should be moved to ```api/users/views.py```. Additionally, this endpoint needs tests for retriving, creating, updating, and deleting public files.

framework/auth/core.py
```diff
@@ -1034,6 +1033,16 @@ def given_name_initial(self):
         return self.given_name[0]
 
     @property
+    def public_files_node(self):
+        from website.project.model import Node
+        if self.is_registered:
+            try:
+                return Node.find_one(Q('contributors', 'eq', self._id) & Q('is_public_files_node', 'eq', True))
+            except NoResultsFound:
+                None
```
caseyrollins: ```return None``` here and remove return statement below.

website/public_files/__init__.py
```diff
@@ -0,0 +1,24 @@
+from framework.auth.signals import user_confirmed
+from website.exceptions import NodeStateError
+from website.project.model import Node
+
+@user_confirmed.connect
+def give_user_public_files_node(user):
```
caseyrollins: Please add tests for this function.

website/public_files/views.py
```diff
@@ -0,0 +1,34 @@
+import httplib as http
+
+from website.project.model import Node
+from modularodm import Q
+from modularodm.exceptions import NoResultsFound
+from framework.exceptions import HTTPError
+from framework.auth.decorators import must_be_logged_in
+
+@must_be_logged_in
+def view_public_files(auth, **kwargs):
```
caseyrollins: Please add tests for the public files views.

brianjgeiger:

PR #5943, opened 2016-07-08
[hold] [feature] @mentions and notifications [OSF-5975][OSF-5658]

sloria:
Carried over from #5840

--------

## Purpose

Related tickets: https://openscience.atlassian.net/browse/OSF-5658 & https://openscience.atlassian.net/browse/OSF-5975

## Changes
* @ mention names appear as a link to the mentioned contributor
* An error message is shown and the submit button is disabled if a comment is over 500 characters
* add dependency At.js
    * provide @/+ functionality
* add ever_mentioned to the comment model

Mentioning a contributor:

## Possible additions
* Emojis PR #5418 (this is being updated to reflect changes on this PR)

## Side Effects
* ContentEditable divs    
* a user can write in a different name for the mention if they really understand the code


sloria:
Initial round of bugs/questions found by @AllisonLBowers: 

- [ ] *In email notifications, the usernames that are @ mentioned have links that go nowhere (they are just the user GUIDs)
- [ ] *Usernames with certain characters do not work (!, @, ?, -, #, $, ~, +, `, “, (), *, ., etc). The username shows up correctly in the list of contributors when you type “@,” but if you select the name, it just shows up as regular text and does not perform the @mention.
- [ ] *I’m not getting notifications for users who have comments set to “Never”
- [ ] *Is there supposed to be a separate email template for @mentions? Currently, they only show up in regular OSF Notifications format and seem to be treated like regular comments.
- [ ] *On global notification settings, it is only possible to set Mentions to “Instantly.” However, it is possible to receive mentions daily if you have comments for a particular project set to “Daily.” Should there be multiple options listed for Mentions? And if Instantly is the only option, should it be presented in a drop-down menu? This makes it look like the user can change their settings, only to click on it and discover this isn’t the case.
- [ ] *If I’m understanding the PR correctly, if a user edits a comment and adds an @mention, the mentioned user should receive a notification. However, this behavior is not occurring.

website/project/model.py
```diff
@@ -179,6 +179,32 @@ class MetaData(GuidStoredObject):
...
+    return True
+
+def get_valid_mentioned_users_guids(comment, contributors):
+    """ Get a list of valid users that are mentioned in the comment content.
+
+    :param Node comment: Node that has content and ever_mentioned
+    :param list contributors: List of contributors on the node
+    :return list new_mentions: List of valid users mentioned in the comment content
+    """
+    new_mentions = set(re.findall(r"\[[@|\+].*?\]\(\/([a-z\d]{5})\/\)", comment.content))
```
leb2dg: I think this is why the notifications stopped working. This regular expression wasn't updated after the url was changed. Although a lot of tests should have failed.

sloria: Ah ok. My mistake. If we're going to reuse this regex, we should probably make it a constant

leb2dg: Do you mean the JavaScript one --`'\\[(@|\\+)(.*?)\\]\\(\\/([a-z\\d]{5})\\/\\)'`-- and the Python one -- `"\[[@|\+].*?\]\(\/([a-z\d]{5})\/\)"`?

sloria: No, it would only make sense to have the constant if we could reuse it in the same language. Might not be the case...

leb2dg: Oh there is another python one in the comment length validator. Where would the constant be?

PR #5944, opened 2016-07-08
[Feature] APIv2 Addons: Box and Dropbox [OSF-6120]

mfraezz:
## Purpose
Add support for folder listing through the API for more addons: Box and Dropbox

## Changes
* Refactor the view for getting folders
* Improvements to code consistency across addons
* Remove dead code
* Update test imports/mocking

## Side effects
None expected. It's possible that some other service's API (aside from just Box) will return `'All Files'` for the name / path of the root folder, but I didn't notice any during testing. If this is the case, a similar approach to Box's renaming method should be implemented.

## QA Notes
* Reuse test sessions from #5532 for Google Drive folder listing.
* Test UI: folder picker and the files page could be affected.

## Deployment Notes
```bash
python -m scripts.box.migrate_folder_language --dry
```

## Ticket
Partial fix for [OSF-6120](https://openscience.atlassian.net/browse/OSF-6120)


scripts/box/migrate_folder_language.py
```diff
@@ -0,0 +1,43 @@
...
+from website.app import init_app
+from scripts import utils as script_utils
+
+logger = logging.getLogger(__name__)
+
+def migrate(dry_run=False):
+    bns_collection = database['boxnodesettings']
+    migrated_names = {b['_id'] for b in bns_collection.find({'folder_name': 'All Files'})}
+    migrated_paths = {b['_id'] for b in bns_collection.find({'folder_path': 'All Files'})}
+    assert migrated_names == migrated_paths
```
brianjgeiger: So if `migrated_names != migrated_paths`, more development will be needed on this migration, but you're expecting a low probability of that?

brianjgeiger: Looks like because you're expecting maybe someone linked to a folder named `All Files` which is not actually the root, yes?

mfraezz: No. 

>MattFrazier
>`{b._id for b in BoxNodeSettings.find(Q('folder_name', 'eq', 'All Files'))} == {b._id for b in BoxNodeSettings.find(Q('folder_path', 'eq', 'All Files'))}`
>[...]
>Steve
>`Out[3]: True`

website/addons/base/generic_views.py
```diff
@@ -54,6 +54,7 @@ def _account_list(auth):
     return _account_list
 
 def folder_list(addon_short_name, addon_full_name, get_folders):
+    # TODO: Generalize this for API use after node settings have been refactored
```
brianjgeiger: Link to a ticket?

brianjgeiger:
This looks good. It functions well in manual testing and you have unit tested to your preferred level of rigor?

mfraezz:
>This looks good. It functions well in manual testing and you have unit tested to your preferred level of rigor?

Yes. I manually tested folder listing and linking for both addons in this PR, behavior is approximately equal to GDrive. Unit tests have been updated to ensure correct behavior.

The only other improvement I was considering putting in this PR was `folder_id` validation for other addons that don't strictly require it -- Dropbox and GDrive. That could alternatively go in the follow-follow-up ticket for clean-up / behavior normalization across all addons, once each has been migrated to be able to use APIv2 for basic functionality. I was indifferent about doing so here because some of the changes made are blocking other addons from going through.

Response when you specify an incorrect `folder_id` for Box:

screen shot 2016-07-12 at 14 40 54

website/addons/box/model.py
```diff
@@ -188,5 +247,4 @@ def after_delete(self, node=None, user=None):
 
     def on_delete(self):
         self.deauthorize(add_log=False)
-        self.clear_auth()
```
sloria: Why was this removed?

mfraezz: `deauthorize` calls `clear_auth`

PR #5952, opened 2016-07-11
Fix `inv karma` docstring

whit537:
## Purpose

The default test browser in `karma.conf.js` [is PhantomJS](https://github.com/CenterForOpenScience/osf.io/blob/ef98b0bf6eb5b520fff990306bc58ea3fcaca061/karma.conf.js#L6), not Chrome.

## Changes

s/Chrome/PhantomJS

## Side effects

n/a

## Ticket

Noticed while working on [OSF-6447](https://openscience.atlassian.net/browse/OSF-6447).


sloria:
Good catch. Thanks!

PR #5953, opened 2016-07-11
Fix unnecessary loading icon on external link popup [#OSF-6520]

caspinelli:
## Purpose

Fix bug where there is some weird behavior in which your mouse cursor continues to show a loading icon for as long as the external link pop-up on a project title page is open

## Changes
Remove CSS styling that makes cursor be styled as 'wait' on the popup and popup overlay

## Side effects
I don't know if there was a reason that the cursor should be styled as 'wait' on the popup, but if there was then this change may not be the right direction

## Ticket

https://openscience.atlassian.net/browse/OSF-6520

### DIFF
--- a/website/addons/forward/static/forwardWidget.js
+++ b/website/addons/forward/static/forwardWidget.js
@@ -31,7 +31,7 @@ var ViewModel = function(url) {
     self.queueRedirect = function() {
         self.redirecting(true);
         if (!self.isRegistration()) {
-            $.blockUI({message: $('#forwardModal')});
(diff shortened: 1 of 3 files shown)


caspinelli:
The only reason the travis build failed is the weird error: 
✗ returns false for date more than 4 years in the future

sloria:
This modifies a 3rd-party library, which we should avoid. Can you instead override the default setting you changed?

sloria:
Pass finished.

PR #5961, opened 2016-07-12
Fix the display of unrendered markdown on wiki and project page [OSF-6361]

caspinelli:
## Purpose

Fix the unwanted unrendered markdown on the wiki and project page.

## Changes

Added CSS so the unwanted unrendered markdown is not displayed until it is rendered.

## Side effects
None

## Ticket

https://openscience.atlassian.net/browse/OSF-6361


brianjgeiger:
Looks good. Did it work well in manual testing?

caspinelli:
Yeah, with the change there is no flash of unrendered markdown on the wiki page, and without it there is

brianjgeiger:
Awesome. 🎉 🐧 🎉

PR #5967, opened 2016-07-13
Add node relationship to APIv2 files detail endpoint [#OSF-6685]

abought:
## Purpose

Adds a new "relationships.node" field to APIv2 representation of each file. This field provides a link to the project (or registration) that owns the file.
screen shot 2016-07-13 at 11 34 00 am

## Changes

The following endpoints are affected:
1. NodeFileDetail `http://localhost:8000/v2/nodes/f263w/files/osfstorage/57598fb3c4f5fbf0f066a8fe`
2. NodeFilesList  `http://localhost:8000/v2/nodes/f263w/files/osfstorage/`
3. FileDetail `http://localhost:8000/v2/files/5769ef78c4f5fb37ee10ef2e/`
4. RegistrationFileDetail  `http://localhost:8000/v2/registrations/vytz8/files/osfstorage/5769ef78c4f5fb37ee10ef2d/`
5. RegistrationFileList `http://localhost:8000/v2/registrations/vytz8/files/osfstorage/`

## Side effects
None expected, but there is one known bug beyond the scope of this PR.

## Known issues
The file detail page shows the wrong relationship link for registration files. This is a [known issue](https://openscience.atlassian.net/browse/OSF-6551) that also affects comments links.

This PR assumes that every file belongs to exactly one project (consistent with current field definition). This could behave weirdly if that assumption is broken in the future.

## Ticket
https://openscience.atlassian.net/browse/OSF-6685


abought:
Requesting CI restart for unrelated failing tests.

A personal branch build did pass, but there's no guarantee that represents the final merged state.
https://travis-ci.org/abought/osf.io/builds/144504199

pattisdr:
This is super helpful Andy - I can use it for the file detail page.  We should also add the `node` relationship to the file model in the ember-osf addon.

abought:
> This is super helpful Andy - I can use it for the file detail page. We should also add the node relationship to the file model in the ember-osf addon.

Indeed. That should go in along with the project navbar PR. (I'm working through the template conversion process now)

https://github.com/CenterForOpenScience/ember-osf/pull/100/files#diff-d44f067459505b9ee4c68e51f1b0ff48R45

brianjgeiger:
Steve is looking into the failing test; don't worry about it.

api/registrations/serializers.py
```diff
@@ -314,8 +314,13 @@ class RegistrationFileSerializer(FileSerializer):
     comments = FileCommentRelationshipField(related_view='registrations:registration-comments',
                                             related_view_kwargs={'node_id': '<node._id>'},
                                             related_meta={'unread': 'get_unread_comments_count'},
-                                            filter={'target': 'get_file_guid'})
+                                            filter={'target': 'get_file_guid'}
+                                            )
 
+    node = RelationshipField(related_view='registrations:registration-detail',
```
brianjgeiger: Hmmm. I wonder what happens to sub-views on a withdrawn registration. https://api.osf.io/v2/registrations/28yxd/files/ apparently permissions work properly, so you shouldn't have access to a file that belongs to a withdrawn registration. Phew. Very good, continue on.

brianjgeiger:
Looks great. 🎉 🐧 🎉

api/registrations/serializers.py
abought: > Hmmm. I wonder what happens to sub-views on a withdrawn registration. https://api.osf.io/v2/registrations/28yxd/files/ apparently permissions work properly, so you shouldn't have access to a file that belongs to a withdrawn registration. Phew. Very good, continue on.

Interesting! Thanks for looking into that; it's an edge case I hadn't known to look for.

In the original workup, I noticed that many files endpoints appear to use permissions classes that already check the owning node, internally. The hope is that therefore node and files rules are already internally consistent.

brianjgeiger: I checked; it's fine. That was mostly a note in case anyone happened to think of the same case on a subsequent review.

abought:
Restarted CI to resolve non-deterministic failures.

Pinging to avoid PR going stale.

PR #5968, opened 2016-07-13
Update install instructions to include RHEL and CentOS

atelic:
## Purpose

The install instructions for Linux are pretty Ubuntu-centric 

## Changes
Where we mention Ubuntu in the README, supply the same RHEL command to install. Fedora users will be able to substitute `dnf` for `yum` and have the instructions work.


sloria:
Thanks

PR #5971, opened 2016-07-13
Fix bug 'user cannot scroll to top ERPC terms modal' [OSF-6635]

caspinelli:
## Purpose

Enable users to scroll to top of ERPC terms model when modal is long enough that there is overflow

## Changes
Since the default CSS for modals is found in a 3rd party library, added jquery code that overrode default CSS and made the overflow for the modal scrollable. Also made overflow of the background (body) hidden so there is only one scroll bar

Before:
screen shot 2016-07-14 at 10 29 48 am
After: (Notice scrollbar)
screen shot 2016-07-14 at 10 30 14 am

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-6635

### DIFF
--- a/website/static/css/style.css
+++ b/website/static/css/style.css
@@ -698,3 +698,14 @@ button.close {
 div.ball-scale-blue>div {
     background-color: #BED1E2;
 }
+
+/* Make registrations modal scrollable and background hidden
+-------------------------------------------------- */
+
+.background-unscrollable {
+    overflow: hidden;
+}
+
+.modal-scrollable {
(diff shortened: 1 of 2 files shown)


brianjgeiger:
Can you add a class to the modal and make the CSS for that class rather than adding the CSS via Mithril?

brianjgeiger:
Pass done. 🐧

brianjgeiger:
Looks good. 🎉 🐧 🎉

PR #5972, opened 2016-07-14
add toRelativeUrl function to logTextParser.js [OSF-6691]

adlius:
## Purpose

Currently some of the urls in the recent activity logs are absolute urls. This PR changes them to relative urls.

## Changes

toRelativeUrl() is added to LogTextParst.js

## Side effects

None

## Ticket
https://openscience.atlassian.net/browse/OSF-6691


website/static/js/logTextParser.js
```diff
@@ -25,6 +25,12 @@ function ravenMessage (message, logObject) {
     }
 }
 
+function toRelativeUrl(url) {
```
brianjgeiger: This is a good candidate for some JS unit tests.

brianjgeiger:
Pass done. Some tests would be awesome. 🐧

website/static/js/logTextParser.js
```diff
@@ -25,6 +25,12 @@ function ravenMessage (message, logObject) {
...
+    var parser = document.createElement('a');
```
brianjgeiger: Also, I'm unfamiliar with that pattern of URL manipulation. Is that a Mithril thing? Will it work on all our supported browsers?

adlius: @brianjgeiger I was referring to [http://stackoverflow.com/questions/736513/how-do-i-parse-a-url-into-hostname-and-path-in-javascript](url) when I wrote the toRelativeUrl method. It seems that it is a DOM thing. We could do some string manipulations of course but I think this is probably the most robust way as it would extract the relative path of the URL no matter what the domain name is and in what form.

adlius: @brianjgeiger Also, would any of the log links be an external link? Currently I am converting all the links to relative assuming that all the log links are internal links.

brianjgeiger: I'm not sure; depending on what gets logged if you add an external link with the forward addon, that could be. It would probably be wise to verify that the window's url matches the link url before making it relative.

brianjgeiger: Go ahead and link to that stack overflow question in a comment in your function.

adlius:
A conditional is added to check whether the link to be converted is internal or external.
Also added a JS test for toRelativeUrl().

website/static/js/tests/logTextParser.test.js
```diff
@@ -0,0 +1,25 @@
...
+            externalAbsoluteUrl: 'http://externaldomin.com/some/external/path/',
+            internalAbsoluteUrl: window.location.origin + '/mst3k/files/',
+            internalRelativeUrl: '/mst3k/forks/'
+        }
+
+        it('Does not affect external URLs', () => {
+            assert.equal(logTextParser.toRelativeUrl(testData.externalAbsoluteUrl), testData.externalAbsoluteUrl);
+        });
+
+        it('DOes not affect URLs that are already relative', () => {
```
brianjgeiger: Typo with `DOes`

website/static/js/logTextParser.js
```diff
@@ -25,6 +25,23 @@ function ravenMessage (message, logObject) {
...
 
+/**
+ * Utility function to convert absolute URLs to relative urls
+ * See:
+ *      http://stackoverflow.com/questions/736513/how-do-i-parse-a-url-into-hostname-and-path-in-javascript
+ * This method have no effect on external urls.
+ * @param url {string} url to be converted
+ * @returns {string} converted relative url
+ */
+function toRelativeUrl(url) {
```
brianjgeiger: This seems like a fairly universal function. We should probably add it to osfHelpers.js. Sorry for not noticing before.

brianjgeiger:
```js
if (window.location.hostname === parser.hostname){
    relative_url = parser.pathname + parser.search + parser.hash;
```

Your function relies on hostname, pathname, search, and hash. Probably would be good to have unit tests that handle each of these items explicitly. Also one that makes explicit if you're using a different method to access a resource (http vs https, or ftp vs https) in case someone makes a change in the future that might break something.

brianjgeiger:
Pass done. Some more tests and a little refactoring to move the function. If you can think of anything else that would be wise to test (consider your assumptions of how you expect it to work and make sure they're explicit), go for it. But overall, this is very good. 🐧

adlius:
The newest commits includes moving "toRelatvieUrl()" to osfHelpers.js. Improve the tests and move the tests to osfHelpers.test.js.

mfraezz:
The build failed due to `jshint` errors, please resolve them. 🚞

brianjgeiger:
Please resolve conflicts and this should be ready to go. 🐧

adlius:
Resolved conflicts. Waiting for Travis to pass......

brianjgeiger:
Looks great. 🎉 🐧 🎉

PR #5975, opened 2016-07-14
[OSF-3520] Registrations draggable

ccfair:
## Purpose
You shouldn't be able to drag or drop any files when you are viewing a registration because everything is supposed to be frozen. This is a fix preventing moving in treebeard while viewing a registration

## Changes
I created a function that checks the boolean value `node.isRegistration` which determines if files can be moved on the project dashboard and files page by the `allowMove` option in treebeard.

## Side effects
None that I can think of.

## Ticket

[Osf-3520](https://openscience.atlassian.net/browse/OSF-3520)

### DIFF
--- a/website/static/js/pages/files-page.js
+++ b/website/static/js/pages/files-page.js
@@ -15,19 +15,11 @@ $(document).ready(function(){
     $.ajax({
       url: nodeApiUrl + 'files/grid/'
     }).done(function(data) {
-        function register() {
-            if(node.isRegistered){
-                return false;
-            }
-            else{
-                return true;
-            }
-        }
         new Fangorn({
             placement: 'project-files',
             divID: 'treeGrid',
             filesData: data.data,
-            allowMove: register(),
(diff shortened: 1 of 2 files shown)


website/static/js/pages/files-page.js
```diff
@@ -14,10 +15,19 @@ $(document).ready(function(){
     $.ajax({
       url: nodeApiUrl + 'files/grid/'
     }).done(function(data) {
+        function register() {
+            if(node.isRegistered){
```
brianjgeiger: This could be simplified to `return !node.isRegistered`. If you have access to the data in the Fangorn setup, then `allowMove: !node.isRegistered` on those two lines will probably do what you need.

brianjgeiger:
I think this could be simplified a bit. At the very list the if statement could be simplified. Pass done. 🐧

mfraezz:
This PR can probably be closed, it appears to be replaced by #5990

PR #5976, opened 2016-07-14
[#OSF-6633]Search on prod: hitting enter multiple times causes doubling of search results

chennan47:
## Purpose

Fix the problem that hit enter multiple times causes doubling of search results.

## Ticket

https://openscience.atlassian.net/browse/OSF-6633


brianjgeiger:
Looks good. 🎉 🐧 🎉

PR #5977, opened 2016-07-15
Fix heisenbug in test for 4 year limit on embargo dates

whit537:
#6025 &rarr;

## Purpose

[OSF-6477](https://openscience.atlassian.net/browse/OSF-6447) describes a bug that we're seeing intermittently during testing at Travis. This PR fixes it.

## Status

Root cause [identified](https://github.com/CenterForOpenScience/osf.io/pull/5977#issuecomment-232963579) and [analyzed](https://github.com/CenterForOpenScience/osf.io/pull/5977#issuecomment-236922483). Surrounding area [excavated](https://github.com/CenterForOpenScience/osf.io/pull/5977#issuecomment-233401612), and partially refactored (#6025, #6050). Fix [discovered](https://github.com/CenterForOpenScience/osf.io/pull/5977#issuecomment-235797811), and implemented on this PR. Ready for review! :)

## Changes

See the [commit log](https://github.com/CenterForOpenScience/osf.io/pull/5977/commits).

## Side effects

Note for QA: verify that the embargo date functionality still works as expected.

## Ticket

[OSF-6477](https://openscience.atlassian.net/browse/OSF-6447)

### DIFF
--- a/website/static/js/registrationModal.js
+++ b/website/static/js/registrationModal.js
@@ -20,7 +20,6 @@ var MAKE_EMBARGO = {
     value: 'embargo',
     message: 'Enter registration into embargo'
 };
-var utcOffset = moment().utcOffset();
 var today = new Date();
 var todayMinimum = moment().add(2, 'days');
 var todayMaximum = moment().add(4, 'years');
@@ -47,7 +46,7 @@ var RegistrationViewModel = function(confirm, prompts, validator) {
         }
     );
     self.embargoEndDate = ko.computed(function() {
-        return moment(new Date(self.pikaday())).subtract(utcOffset, 'minutes');
+        return moment(new Date(self.pikaday()));
     });
 
     self.minimumTimeValidation = function (x, y, embargoLocalDateTime) {
--- a/website/static/js/tests/registrationModal.test.js
+++ b/website/static/js/tests/registrationModal.test.js
@@ -5,8 +5,6 @@ var utils = require('tests/utils');
 var faker = require('faker');
 var moment = require('moment');
 
-console.log('UTC offset: ' + moment().utcOffset());
-
 var RegistrationModal = require('js/registrationModal').ViewModel;
 
 // Add sinon asserts to chai.assert, so we can do assert.calledWith instead of sinon.assert.calledWith
--- a/website/static/js/tests/registrationModal.test.js
+++ b/website/static/js/tests/registrationModal.test.js
@@ -94,7 +94,7 @@ describe('registrationModal', () => {
         });
         it('returns false for date more than 4 years in the future', () => {
             var invalidFutureDate = new Date();
-            invalidFutureDate.setDate(invalidFutureDate.getDate() + 1460);
+            invalidFutureDate.setDate(invalidFutureDate.getDate() + 1462);
             vm.pikaday(invalidFutureDate);
             assert.isFalse(vm.pikaday.isValid());
         });
--- a/website/static/js/tests/registrationModal.test.js
+++ b/website/static/js/tests/registrationModal.test.js
@@ -92,7 +92,7 @@ describe('registrationModal', () => {
             vm.pikaday(invalidPastDate);
             assert.isFalse(vm.pikaday.isValid());
         });
-        it.skip('returns false for date more than 4 years in the future', () => {
+        it('returns false for date more than 4 years in the future', () => {
             var invalidFutureDate = new Date();
             invalidFutureDate.setDate(invalidFutureDate.getDate() + 1460);
             vm.pikaday(invalidFutureDate);


whit537:
Blam: [subsecond differences](https://travis-ci.org/CenterForOpenScience/osf.io/builds/145012685#L1913-L1914) when UTF offset is zero. I surmise that the occasional _successful_ test is due to the clock precision only going to three places, with the moments sometimes being the same to three places.

whit537:
Alright, so what's the right fix?

whit537:
@brianjgeiger at https://github.com/CenterForOpenScience/osf.io/pull/5708#issuecomment-221261771 articulates some confusion about UTC vs. local time in this code: a sentiment I share. Let's look into that ...

whit537:
It's also smelly that `embargoLocalDateTime` is `undefined`. Is that simply a test knob? I _think_ so.

whit537:
And what's up with `todayM{ini,axi}mum` part of the validation tests? What's the purpose of that

whit537:
Seems (be77d53c98276c8cf2d8ed05ca1f73bd6e85e252) that that's designed to prevent the "load the page and leave it open for three days" issue mentioned in https://github.com/CenterForOpenScience/osf.io/pull/5708#issuecomment-221261771.

whit537:
But wait, that's precisely backwards. We want to validate the embargo date against the moment when the user selects the date, not the date when the page loads. Hmm ...

Ultimately we should also have a server-side check on the date. Do we?

whit537:
> Fix issue where Embargos could be made in the past if time is changed

If time is changed then we've all got bigger problems. ;-)

whit537:
> Do we?

[Yes](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/project/views/drafts.py#L56-L80). Do we use it? :)

whit537:
[Yes](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/project/views/drafts.py#L126), during [`submit_draft_for_review`](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/routes.py#L1278).

whit537:
Hitting a Chrome DevTools [bug](http://stackoverflow.com/questions/23398854/chrome-developer-tools-caches-old-version-of-javascript-files) ...

whit537:
... back on track.

whit537:
> [Yes](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/project/views/drafts.py#L126), during [`submit_draft_for_review`](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/routes.py#L1278).

However, I'm not seeing where that route is hit. When I click "Continue" on the registration modal (it *is* the registration modal, correct?), I see a `POST` to this route instead:

http://localhost:5000/api/v1/project/f39dm/drafts/5789352a250c776b3c57d5cd/register/

screen shot 2016-07-15 at 5 38 06 pm

whit537:
Alright, so it looks like server-side validation of the embargo date doesn't happen until the draft registration is submitted for approval. I haven't fully traced the flow for this, but for our purposes here we shouldn't need to. The point here is that we don't need to ... well, hmm ...

I'm trying to understand the relationship between the client-side validation and the server-side validation. Are they targeting the same conditions? This [sentence-ish piece of documentation](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/project/views/drafts.py#L59-L60) is tantalizing on this point:

> If an embargo end_date that is within that window is at risk of causing validation errors down the line if the draft is approved and registered.

How to interpret ... ?

whit537:
I'm excavating to find a good point at which to begin fixing the presenting issue with the test.

whit537:
I now find an `_is_embargo_date_valid` [method](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/project/model.py#L3453-L3458) on the `Node` class. I believe the call chains look like this:

```
registrationModal.js
  /api/v1/project/<pid>/drafts/<draft_id>/register/
    website.project.views.drafts.register_draft_registration
      website.project.model.Node.embargo_registration
        website.project.model.Node._is_embargo_date_valid      
???
  /api/v1/project/<pid>/drafts/<draft_id>/submit/
    website.project.views.drafts.submit_draft_for_review
      website.project.views.drafts.validate_embargo_end_date
        website.project.model.Node._is_embargo_date_valid      
```

whit537:
My general impression is that the client-side and (both) server-side validations are supposed to be the same. It'd be clearer if they used the [config setting](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/settings/defaults.py#L537) consistently.

whit537:
This reminds me of when we looked at the comment length validation at https://github.com/leb2dg/osf.io/pull/14#discussion_r61798316. I suppose this is where Node.js gets to say, "Told ya so!"

whit537:
Why are we using timestamps at all? The input is a date picker. Why aren't we validating based on date alone, without reference to time of day?

whit537:
I find a [call](https://github.com/whit537/osf.io/blob/e62cf7a0d19323c920f12844c9d28d2a334a8383/website/static/js/registrationModal.js#L45) to a seemingly vestigial `isEmbargoEndDateValid` function.

whit537:
... unless it's a magic name from Knockout-Validation?

whit537:
> Why aren't we validating based on date alone, without reference to time of day?

http://momentjs.com/docs/#/manipulating/start-of/

```
moment().startOf('day')
```

([via](http://stackoverflow.com/a/19699447))

whit537:
I'm having trouble triggering the `onSelect` for `pikaday`. Is it because of a blockage in my assets pipeline? Or because the code is, in fact, never called?

whit537:
#### Open Questions

- [ ] What is the validation intention?
 - [ ] How do client and server sides relate?
 - [ ] How do local and UTC time relate?
- [ ] Is `isEmbargoEndDateValid` vestigial?
- [ ] What is [this sentence](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/project/views/drafts.py#L59-L60) _supposed_ to say?

whit537:
In restricted Flowdock, @abought [seconds the opinion](https://www.flowdock.com/app/cos/osf-dev/threads/q2bCXTu6SpzGOnidSqoTYcBpvwJ) that the `isEmbargeEndDateValid` call is vestigial.

whit537:
`isEmbargoEndDateValid` entered in c711f60 (no PR?) ([h/t](https://gist.github.com/shaiguitar/7867796#gistcomment-965936))

whit537:
It was a [computed observable](https://github.com/CenterForOpenScience/osf.io/blob/c711f6096035eda04c1eb50efb92c3d13820f2cc/website/static/js/registrationEmbargo.js#L66-L70). Okay! Vestigial code!

whit537:
[Relevant](http://www.commitstrip.com/en/2016/07/18/where-does-it-come-from?). :-)

strip-walking-dead-code-english-2

whit537:
Hitting [that](https://github.com/CenterForOpenScience/osf.io/pull/5977#issuecomment-233058680) Chrome bug again. 😞

whit537:
Turns out that the entire [`picker` object](https://github.com/CenterForOpenScience/osf.io/blob/e62cf7a0d19323c920f12844c9d28d2a334a8383/website/static/js/registrationModal.js#L39-L48) was vestigial. PR to remove it in #6025.

whit537:
Gosh. Alright, [now for](https://github.com/CenterForOpenScience/osf.io/pull/5977#issuecomment-234574381) the validation intent ...

whit537:
> the validation intent ...

What's the workflow for a registration? What role does the embargo end date play?

whit537:
There's a draft registration approval period which [should be more](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/settings/defaults.py#L537-L538) than the minimum embargo end date.

whit537:
That's confusing to me. I was about to think that the registration approval phase wanted to always complete _before_ the embargo started.

brianjgeiger:
http://help.osf.io/m/registrations may be useful.

whit537:
First I've seen of our user-facing help docs—I was actually about to ask for design docs or some such. Thanks, @brianjgeiger! :-)

whit537:
> First I've seen of our user-facing help docs

(That may not be strictly true. :)

whit537:
> When a project is registered, all administrators on the project are notified via email and given the opportunity to approve or disapprove the registration by clicking the appropriate link in the email. The registration goes into effect when all administrators have approved the registration or 48 hours have passed, whichever comes first. If one administrator disapproves the registration, it is cancelled. You will be notified via email if an administrator disapproves the registration.

whit537:
> Embargoed registrations undergo the same contributor approval process as immediate registrations. The registration will enter the embargo period when all project administrators approve the registration or 48 hours pass, whichever comes first.

http://help.osf.io/m/registrations/l/524207-embargoes

whit537:
I find an additional [validator](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/static/js/registrationUtils.js#L612-L617) for the case where the meta-schema of the draft registration requires approval (the configured timeouts are to be kept [manually in sync](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/static/js/registrationUtils.js#L25-L29)(!) between Python and JavaScript). On the one hand, that makes it sound like there's a case that _doesn't_ require approval, which the help docs don't seem to mention. On the other, I don't believe I've yet been able to trigger the additional validator in the UI ...

whit537:
It's weird to have two minimum date validations:

 - [one](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/static/js/registrationModal.js#L67): `'Embargo end date must be at least three days in the future.'`
 - [two](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/static/js/registrationUtils.js#L616): `'Embargo end date must be at least ' + DRAFT_REGISTRATION_MIN_EMBARGO_DAYS + ' days      in the future.'`

If I'm reading this code right, _both_ messages could be shown at the same time. No? That'd just be weird.

whit537:
Is *draft* registration approval different than (non-draft) registration approval?

whit537:
Coincidentally(?), I'm seeing this in my `inv server` console:

```
  File "/Users/whit537/workbench/cos/osf.io/env/lib/python2.7/site-packages/modularodm/storedobject.py", line 935, in __getattr__
    raise AttributeError(errmsg)
AttributeError: DraftRegistration object has no attribute requires_approval
```

whit537:
I'm beginning to wonder whether [`requiresApproval`](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/static/js/registrationUtils.js#L611-L618) is _also_ vestigial. O.O

whit537:
I find a `metaschema` collection. I haven't traced the connection from the db out to the UI, but it seems likely that this collection is the source for e.g., [`self.metaSchema`](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/static/js/registrationUtils.js#L611). I don't find a `requires{A,_a}pproval` key in any of the documents in this collection (I'd want to wipe my db and retry with a fully updated `develop` to be sure). It's possible that this key is inserted somewhere between the db and the UI (computed how?). Perhaps a missing `requiresApproval` in the db results in a `false` value in the JavaScript. But why is the key missing from *all* documents? _None_ have a truthy `requires_approval`.

screen shot 2016-07-26 at 12 58 46 pm

whit537:
It appears that [`draft`](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/templates/project/register_draft.mako#L61) in the Mako context is upstream of `metaSchema`; `draft` [comes](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/routes.py#L1016) [from](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/project/views/drafts.py#L165) [`serialize_draft_registration`](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/project/metadata/utils.py#L29-L54).

whit537:
The `draft` argument to `serialize_draft_registration` is a [`DraftRegistration`](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/project/model.py#L3859-L4060) instance (created and injected, [with](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/project/views/drafts.py#L156) [impressive gymnastics](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/project/views/drafts.py#L36-L42), by [`autoload`](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/framework/mongo/utils.py#L116-L148)).

whit537:
`DraftRegistration` has a [`requires_approval`](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/project/model.py#L3945-L3947) computed property, which [delegates](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/project/model.py#L3887) to [`MetaSchema`](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/project/model.py#L87-L161). The meta-schema in turn dynamically computes its [`requires_approval`](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/project/model.py#L102-L104) property, from the `.schema.config.requiresApproval` key in the Mongo document.

whit537:
Okay! Looks like `requiresApproval` is for the pre-reg challenge.

screen shot 2016-07-26 at 2 32 00 pm

whit537:
... and _only_ for the Prereg Challenge, at this point.

screen shot 2016-07-26 at 2 43 53 pm

whit537:
Okay, so I predict that if I try to submit a registration for the Prereg Challenge that has a date that's less than 3 days in the future, I'll get _two_ error messages:

> Embargo end date must be at least three days in the future.
> Embargo end date must be at least 10 days in the future.

whit537:
False! I get only one message, the 10 days one:

screen shot 2016-07-26 at 2 49 23 pm

Why?

whit537:
Looks like Knockout-Validation bails after the first validation failure; if I change [`unshift`](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/static/js/registrationModal.js#L73) to `push` I get the "three days" message.

Okay!

whit537:
... so `requiresApproval` is _not_ vestigial. How does it interact with the requirement for approval in non-prereg-challenge cases?

whit537:
Alright, I'm pretty sure this is the basic workflow:

workflow

whit537:
It seems that the complication here is in the variability of the approval step: ideally the end of the embargo would be expressed in terms of the end of the approval process, but ...

- approval time varies from seconds (there is one admin, they approve immediately), to 48 hours; and
- approval for the Prereg Challenge seems to want up to 10 days instead of only three.

Therefore, we validate the embargo end date when the draft is _submitted_, and, as a proxy, when the embargo end date is selected in the registration modal.

whit537:
The correct behavior is to work with local time in the UI, and store UTC. I don't know yet what our actual behavior is, but I note that the server-side validation on submit will (I expect) have UTC to look at, while the client-side validation will need to present local time to the user.

[Again](https://github.com/CenterForOpenScience/osf.io/pull/5977#issuecomment-233763531), on my read, the time component is actually irrelevant, and should be removed from the equation. We only care to compare the date component.

whit537:
Since the server-side validation on registration submission is the actual validation that counts, I'm thinking we should start by cleaning that up: making sure it is using UTC properly, and only using the date part. Perhaps that could introduce some discrepancy with the client-side, though I wouldn't be surprised if such a discrepancy already exists. On the client side, I think we should clean up the two validators in `registrationModal.js` first, and then refactor the third validator as well: it should use `moment.js` instead of `Date()`, and it should be clearer that it's _replacing_ the minimum validation instead of augmenting it.

That's a first outline of a work plan, anyway ...

whit537:
> approval for the Prereg Challenge seems to want up to 10 days instead of only three.

I think I have this wrong. `DRAFT_REGISTRATION_APPROVAL_PERIOD` is where the 10 days is defined, and it' used in `validate_embargo_end_date`, so let's see about resolving the `???` in https://github.com/CenterForOpenScience/osf.io/pull/5977#issuecomment-233406232 ...

whit537:
I find a "Submit for review" [button](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/templates/project/register_draft.mako#L29-L35) that [promises](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/static/js/registrationUtils.js#L725-L740) (eventually) [to](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/static/js/registrationUtils.js#L660) [post](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/static/js/registrationUtils.js#L696) [to](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/project/metadata/utils.py#L45) [`/submit/`](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/util/__init__.py#L82-L96) (...). How do I see it in the UI?

Sounds like this is another difference with the Prereg Challenge.

> When you're ready to preregister, select **Submit for review** if you would like to submit your preregistration to the Preregistration Challenge. If you would like to preregister without entering the Challenge, select **Register without review**.
>
A preregistration submitted for review is not immediately registered. A member of the COS review team will check your submission to assure completeness and adherence to the competition’s terms and contact you within 10 business days (typical review time is 1-3 business days).
>
Registering without review will preregister your project right away. Your preregistration may be made public immediately or embargoed for up to 4 years.

http://help.osf.io/m/registrations/l/546603-preregistrations

whit537:
Alright, so the Prereg Challenge adds an additional step: review by COS staff. Hmm ...

whit537:
Confirmed. I get "Submit for review" for the Prereg Challenge, and a `POST` to `/submit/`. 👍

---------------

screen shot 2016-07-27 at 12 08 34 pm

---------------

screen shot 2016-07-27 at 12 10 20 pm

whit537:
Alright, staring at `registrationModal.js`, trying to rein this back in.

> It's also smelly that `embargoLocalDateTime` is `undefined`. Is that simply a test knob? I _think_ so.

The [docs](https://github.com/Knockout-Contrib/Knockout-Validation/wiki/Custom-Validation-Rules#custom-rules) say that validator functions take two args: the value to validate, and an `otherValue` to check it against. Our third argument is a [hack](https://github.com/CenterForOpenScience/osf.io/pull/5708#issuecomment-223019785) to make [testing](https://github.com/CenterForOpenScience/osf.io/blob/f0ba507fff067a0597536a7364d8b08b804615a3/website/static/js/tests/registrationModal.test.js#L112-L165) easier. It seems we're not using Knockout-Validation as designed.

whit537:
Why do we need a minimum embargo length? What's wrong with an embargo that lasts one day and an approval process that takes 10 days, as long as we check _both_ conditions before displaying a registration?

whit537:
(For that matter, who cares if an embargo lasts longer than four years? What's the harm in a 100-year embargo?)

whit537:
I've added some commits to #6025, and started a second PR to clean up our Knockout-Validation usage: #6050.

whit537:
I pushed another couple commits to #6050, and rebased #6025 along the way. I'm now thinking through `todayM{in,ax}imum`.

whit537:
`todayMinimum` is the moment two days out from when the page loads. Since `min` is calculated in the same way (`.add(2, 'days')`) _at validation-time_, it is necessarily after `todayMinimum`. The test `end.isSameOrAfter(todayMinimum)` is therefore redundant with the test `end.isSameOrAfter(min)`, and the `todayMinimum` test can be safely removed.

(I suppose it's possible that the test `min` and `todayMinimum` tests are supposed to differ: the `max` tests do, one testing 4 years, the other 4 years - 1 day.)

whit537:
On the other hand, the computation of `todayMinimum` as four years does differ from that of `max`, which is four years less a day, so it's not necessary that `end` precede the latter when it precedes the former. There is a one day window after loading the page during which `end.isSameOrBefore(max)` can be false while `end.isSameOrBefore(todayMaximum)` is true. In this case, the `todayMaximum` check prevents a user from loading the page, leaving it open for a few days, and _then_ selecting a date that is more than four years in the future from when they first loaded the page. However, this is trivially worked around by simply reloading the page. The `todayMaximum` check is, afaict, pointless.

whit537:
>> Fix issue where Embargos could be made in the past if time is changed

Okay, I now find a comment in [restricted Jira](https://openscience.atlassian.net/browse/OSF-4304?focusedCommentId=34323&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-34323) that seems to make sense of be77d53c98276c8cf2d8ed05ca1f73bd6e85e252:

> When I manually change my date and time, I am able to register less than 3 days into the future. Is this an issue?
>
This also means that if I manually change my date, I can set embargoes to end in the past.
>
Here's what to do to make an embargo in the past:
>
1. Make an Open-Ended registration on staging.osf.io
1. Click continue all the way until you hit the modal that gives you the option of either making it public immediately or setting an embargo.
1. Choose to make an embargo. Click on the bar to select the end date so that the calendar shows up.
1. Go to your date and time preferences and change your date to a date in the past, such as June 1st.
1. Make sure that on the calendar on OSF, it shows that it has registered the fact that your date is now June 1st.
1. Now select June 4th. It will let you do so.

However, be77d53c98276c8cf2d8ed05ca1f73bd6e85e252 doesn't actually fix this issue, because it's possible to change one's operating system clock before loading the registration modal.

It's always going to be possible for a sufficiently motivated user to bypass the date validation on the client-side, because of the high degree of control they have in that environment. We have (or _should_ have, if what have isn't sufficient) server-side validation to prevent gaming the system. What we should be thinking about on the client-side is not so much preventing abuse and preventing confusion. Validating against the date at which the user actually selects a value is less confusing than validating against the date when the page loaded.

whit537:
I'm still able to reproduce the bug locally after refactoring (I'm using a slightly modified version of 6acc411501dbd3aa4a780ce5b047442f921bd7fa from #6050). 👍

whit537:
The crucial clue here is that the test consistently fails when UTC offset is zero or positive, and consistently passes when UTC offset is negative.

I note that our timezone tests are rather [U.S.-centric](https://github.com/whit537/osf.io/blob/e62cf7a0d19323c920f12844c9d28d2a334a8383/website/static/js/tests/registrationModal.test.js#L133-L164). Perhaps a test or two for a truly Eastern timezone would help us out here. ;-)

whit537:
Well, one thing to notice is that `max` is local time, while `end` is UTC (though converting a moment to UTC by [subtracting](https://github.com/CenterForOpenScience/osf.io/blob/6acc411501dbd3aa4a780ce5b047442f921bd7fa/website/static/js/registrationModal.js#L61) [`moment().utcOffset()`](https://github.com/CenterForOpenScience/osf.io/blob/6acc411501dbd3aa4a780ce5b047442f921bd7fa/website/static/js/registrationModal.js#L23) minutes seems weird; why not use [`moment.utc()`](http://momentjs.com/docs/#/parsing/utc/)?). Returning local time from `embargoEndDate`'s function corrects for that, but then we're _always_ in essentially the same case as we've been in with a UTC offset of zero. Now what?

whit537:
This is where [subtracting](https://github.com/CenterForOpenScience/osf.io/blob/6acc411501dbd3aa4a780ce5b047442f921bd7fa/website/static/js/registrationModal.js#L72) a day from the four years starts to look weird. Did we kludge that just to line up with the non-moment [construction](https://github.com/CenterForOpenScience/osf.io/blob/6acc411501dbd3aa4a780ce5b047442f921bd7fa/website/static/js/tests/registrationModal.test.js#L133) in the test?

whit537:
Right, well ... adding a millisecond to the value in the test makes it pass for me here. If there truly [are](https://github.com/CenterForOpenScience/osf.io/pull/5977#issuecomment-232963579) some subsecond differences going on then a millisecond may not be enough, but surely a second or two would be. A day or two would satisfy the test.

whit537:
Rebased on `develop` and fix pushed. Previous head was e62cf7a0d19323c920f12844c9d28d2a334a8383.

Let's see what Travis has to say ...

whit537:
Alright, I've rebased #6025 and #6050 on the latest here (9b46822332409beb646b65d7f6485f08c1b947e2). I'm gonna let Travis churn, maybe tomorrow I'll be able to submit this for code review. 💛

whit537:
D'oh! 🐭

screen shot 2016-07-28 at 7 36 02 am

whit537:
`TEST_BUILD="else"` got clipped at [two hours](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/147948889#L7259).

`TEST_BUILD="varnish"` hit what could be [another heisenbug](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/147948890#L1563).

whit537:
I don't find any immediately helpful error message in the failed build log, though I do recognize it from last night. I see some tracebacks during license and schema loading, but I'm not sure they're not part of normal operation. More research required.

The [build](https://travis-ci.org/CenterForOpenScience/osf.io/builds/147950531) for 9b46822332409beb646b65d7f6485f08c1b947e2 looks better—no failure on `"varnish"`—but this time it [errored](https://travis-ci.org/CenterForOpenScience/osf.io/builds/147950531) out not on a timeout but ... what? Now even the Chrome tabs for these Travis builds are erroring out on me. 😩

It [appears](https://travis-ci.org/CenterForOpenScience/osf.io/builds/147950492) that 6fb954f6a981290469c09d72cabeb2383e32af7e on #6050 also timed out on `"else"` ... can't say for sure without loading the detail page.

The [build](https://travis-ci.org/CenterForOpenScience/osf.io/builds/147950029) in the middle, for a52b9ab0aff65b0dcbd71388964fa95d00ccfe9b on #6025, failed on both else and varnish.

whit537:
The 9b46822332409beb646b65d7f6485f08c1b947e2 build errored with "[The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising the same exception over and over).](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/147950533#L54761)" Click at your own risk; you'll be loading four megs.

whit537:
It appears to have been an elasticsearch failure, though in other news I don't see tracebacks loading licenses or schemas, so maybe that _is_ a clue to the other failure after all.

whit537:
Here's the first elasticsearch [failure](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/147950533#L4097-L4111). I believe what we're seeing is `urllib3` time out after 10 seconds, which somehow leaves the connection in a bad state. Line 4097 seems to indicate that elasticsearch is no longer blocking on the timed-out request, but it appears that all subsequent requests to elasticsearch are failing with `ProtocolError: ('Connection aborted.', BadStatusLine("''",))`. Elasticsearch is reporting response times in the milliseconds in these cases.

For our purposes here we can probably chalk this one up to "Travis was busy," as with the builds hitting Travis's two-hour cap. It would be nice to find a better failure mode before we hit the urllib3/elasticsearch problem in production.

whit537:
> can't say for sure without loading the detail page.

[Confirmed](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/147950494#L6141)

whit537:
> The [build](https://travis-ci.org/CenterForOpenScience/osf.io/builds/147950029) in the middle, for a52b9ab0aff65b0dcbd71388964fa95d00ccfe9b on #6025, failed on both else and ~~varnish~~ osf.

The "osf" build exhibits the elasticsearch timeout and subsequent bad status line, but seems to recover somewhat so it doesn't hit the four meg log limit.

The "else" build failed with ... [our heisenbug](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/147950031#L8694)! 😄  While the familiarity is comforting, it seems to mean that the fix is wrong. 😞

Hmm ...

whit537:
Hmm ...

whit537:
Walking through the build results again, to make sure I read them right.

whit537:
#### Build Results

| <br>step | bugfix (eafdfa53b8aeff900c61af7a3ea3b07a77500729) | etc. (9b46822332409beb646b65d7f6485f08c1b947e2) |
|:---------|:----|:----|
| `osf` | ✅ | ✅ |
| `else` | ❗([timeout](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/147948889#L7259)) | ❗([spaceout](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/147950533#L54761)) |
| `varnish` | ❌  ([license-loading issue](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/147948890)(?)) | ✅  |

whit537:
Gosh. How did I come to believe that I was seeing the heisenbug again?

whit537:
Ah, yes: it wasn't on this PR, but rather on #6025. Hmm ...

whit537:
(Results [table](https://github.com/CenterForOpenScience/osf.io/pull/5977#issuecomment-236613218) updated.)

whit537:
Does a52b9ab0aff65b0dcbd71388964fa95d00ccfe9b indeed contain the fix, or did that get gunked up in rebasing?

whit537:
Looks like it does indeed contain the fix ([one](https://github.com/CenterForOpenScience/osf.io/compare/dd68fca70c442c7a4bf208367977104cfa39ac0e...a52b9ab0aff65b0dcbd71388964fa95d00ccfe9b#diff-9f8b0649d71a07ff6a3d91a6348fcb92L50), [two](https://github.com/CenterForOpenScience/osf.io/compare/dd68fca70c442c7a4bf208367977104cfa39ac0e...a52b9ab0aff65b0dcbd71388964fa95d00ccfe9b#diff-fc0e591a28e810591c23733b07df85b8L130)). Hmm ... I guess we're back to the drawing board! 🐉

whit537:
Closing and reopening this PR to trigger a rebuild at Travis should surface the bug again on this PR (assuming the other Travis issues don't continue to afflict us).

whit537:
Did the same with #6025 and #6050. Let's see if we can dodge the issues apparently related to Travis slowness and get a clean picture of where we stand.

whit537:
Clear enough. :)

#### Build Results

| build<br>step | bugfix<br>eafdfa53b8aeff900c61af7a3ea3b07a77500729 | #6025<br>a52b9ab0aff65b0dcbd71388964fa95d00ccfe9b | #6050<br>2aef3b57416f89c639202fb2a2ee3e13ad92bc1b |
|:----|:----|:---------|:--------|
| `osf` | ✅ | ✅ | ✅ |
| `else` | ❌ ([heisenbug](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/148926431#L8314)) | ❌ ([heisenbug](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/148918863#L8846)) | ❌ ([heisenbug](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/148918640#L8851)) |
| `varnish` | ✅ | ✅  | ✅ |

whit537:
Ooooh! I think I just fixed the wrong test ...

whit537:
Yup. Tee-hee! 😊

whit537:
New fix pushed. Let's see if this 🐶  can hunt ...

whit537:
Wherefore art thou, Travis?

whit537:
https://travis-ci.org/CenterForOpenScience/osf.io/builds/148970448 ...

whit537:
Woo-hoo! Green! Reproducible? :)

whit537:
Running at Travis again.

Hmm ... in the mean time, I'm looking at this [second](https://github.com/whit537/osf.io/blob/a6efbb4739bc476ee8a01deeba37228aedfed1dc/website/static/js/tests/registrationModal.test.js#L128-L132) test, which seems at first blush to duplicate the [first](https://github.com/whit537/osf.io/blob/a6efbb4739bc476ee8a01deeba37228aedfed1dc/website/static/js/tests/registrationModal.test.js#L95-L100). Does it? Does it fail when the first does? If not, why?

whit537:
I added some logging to `maximumTimeValidation` and captured the following results.

```javascript
console.log('loc:', moment(embargoLocalDateTime).format('YYYY-MM-DD HH:mm:ss.SSSSS'));
console.log('max:', moment(maxEmbargoMoment).format('YYYY-MM-DD HH:mm:ss.SSSSS'));
console.log('end:', moment(endEmbargoMoment).format('YYYY-MM-DD HH:mm:ss.SSSSS'));
console.log()
```


"returns false for date more than 4 years in the future":

```
LOG LOG: 'loc:', '2016-08-01 17:18:10.18600'
LOG LOG: 'max:', '2020-07-31 17:18:10.18600'
LOG LOG: 'end:', '2020-07-31 21:18:10.18500'
LOG LOG:
LOG LOG: 'loc:', '2016-08-01 17:18:10.18700'
LOG LOG: 'max:', '2020-07-31 17:18:10.18700'
LOG LOG: 'end:', '2020-07-31 21:18:10.18500'
```

"returns false for date at least 4 years in the future":

```
LOG LOG: 'loc:', '2016-08-01 16:56:31.72300'
LOG LOG: 'max:', '2020-07-31 16:56:31.72300'
LOG LOG: 'end:', '2020-07-31 16:56:31.72200'
LOG LOG:
LOG LOG: 'loc:', '2016-08-01 16:56:31.72300'
LOG LOG: 'max:', '2020-07-31 16:56:31.72300'
LOG LOG: 'end:', '2020-07-31 16:56:31.72200'
LOG LOG:
LOG LOG: 'loc:', '2016-08-01 16:56:31.72200'
LOG LOG: 'max:', '2020-07-31 16:56:31.72200'
LOG LOG: 'end:', '2020-07-31 16:56:31.72200'
LOG LOG:
```

whit537:
The heisenbug is in a test for `#pikaday.isValid`. The second test is in the suite for `#timeValidation`, and it calls `maximumTimeValidation` directly rather than going through `isValid`.

whit537:
Analyzing the [results](https://github.com/CenterForOpenScience/osf.io/pull/5977#issuecomment-236899272), I take it that the last call to `maximumTimeValidation` is the relevant one. I see a subsecond difference between `max` and `end` in the heisenbug case, but in the second case they're exactly the same. Predictably so?

whit537:
> Predictably so?

Yes.

In the pikaday [case](https://github.com/CenterForOpenScience/osf.io/blob/a6efbb4739bc476ee8a01deeba37228aedfed1dc/website/static/js/tests/registrationModal.test.js#L95-L100), `embargoLocalDateTime` is `undefined` [coming into](https://github.com/CenterForOpenScience/osf.io/blob/a6efbb4739bc476ee8a01deeba37228aedfed1dc/website/static/js/registrationModal.js#L58) `maximumTimeValidation`, and therefore [going into](https://github.com/CenterForOpenScience/osf.io/blob/a6efbb4739bc476ee8a01deeba37228aedfed1dc/website/static/js/registrationModal.js#L59) `getMaximumDate`, such that a `new Date` is constructed [under](https://github.com/CenterForOpenScience/osf.io/blob/a6efbb4739bc476ee8a01deeba37228aedfed1dc/website/static/js/registrationModal.js#L126) `getMaximumDate` ([in](https://github.com/moment/moment/blob/develop/moment.js#L2370-L2371)[si](https://github.com/moment/moment/blob/develop/moment.js#L4166)[de](https://github.com/moment/moment/blob/develop/moment.js#L2478-L2480) Moment.js). The `endEmbargoMoment`, however, is coming from the pikaday, which was populated just prior in the test with `invalidFutureDate`, a `Date` created at the start of the test. Therefore, in the heisenbuggy test case (modulo the UTC [gaffe](https://github.com/CenterForOpenScience/osf.io/pull/5977#issuecomment-235800435)), `endEmbargoMoment` is necessarily earlier in time than `maxEmbargoMoment`. That it was occasionally not [seems](https://github.com/CenterForOpenScience/osf.io/pull/5977#issuecomment-232963579) likely to be due to clock imprecision: **they sometimes _appeared_ to be the same**.

In the timeValidation [case](https://github.com/CenterForOpenScience/osf.io/blob/a6efbb4739bc476ee8a01deeba37228aedfed1dc/website/static/js/tests/registrationModal.test.js#L128-L132), `embargoLocalDateTime` is _not_ `undefined` , but rather a `Date` that [becomes](https://github.com/CenterForOpenScience/osf.io/blob/a6efbb4739bc476ee8a01deeba37228aedfed1dc/website/static/js/registrationModal.js#L126) the basis of a `moment` four years (less one day?!) hence: `maxEmbargoMoment`. Now, the self-same `Date` had _already_ been [used](https://github.com/CenterForOpenScience/osf.io/blob/a6efbb4739bc476ee8a01deeba37228aedfed1dc/website/static/js/tests/registrationModal.test.js#L130) to compute a `Date` four years hence to populate the pikaday, which is where `endEmbargoMoment` [comes](https://github.com/CenterForOpenScience/osf.io/blob/a6efbb4739bc476ee8a01deeba37228aedfed1dc/website/static/js/registrationModal.js#L60) [from](https://github.com/CenterForOpenScience/osf.io/blob/a6efbb4739bc476ee8a01deeba37228aedfed1dc/website/static/js/registrationModal.js#L49). Therefore, `endEmbargoMoment` and `maxEmbargoMoment` are **guaranteed to be the same**—as long as Moment's computation syncs with ours (perhaps leap years (cf. [private Jira](https://openscience.atlassian.net/browse/OSF-6447?focusedCommentId=35975&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-35975)) could throw this off?).

I've modified the logging [above](https://github.com/CenterForOpenScience/osf.io/pull/5977#issuecomment-236899272) to show the variance in `embargoLocalDateTime` (I'm running `inv karma --single` with `it.only` for each test).

```javascript
console.log('loc:', embargoLocalDateTime === undefined ? 'undefined' : moment(embargoLocalDateTime).format('YYYY-MM-DD HH:mm:ss.SSSSS'));
console.log('max:', moment(maxEmbargoMoment).format('YYYY-MM-DD HH:mm:ss.SSSSS'));
console.log('end:', moment(endEmbargoMoment).format('YYYY-MM-DD HH:mm:ss.SSSSS'));
console.log()
```

pikaday:

```
LOG LOG: 'loc:', 'undefined'
LOG LOG: 'max:', '2020-08-01 10:16:48.28300'
LOG LOG: 'end:', '2020-08-03 10:16:48.28200'
```

timeValidation:

```
LOG LOG: 'loc:', '2016-08-02 10:19:02.84900'
LOG LOG: 'max:', '2020-08-01 10:19:02.84900'
LOG LOG: 'end:', '2020-08-01 10:19:02.84900'
```

whit537:
(The pikaday case there is showing a two day offset because I ran with the fix, a6efbb4739bc476ee8a01deeba37228aedfed1dc. However, the subseconds are the relevant bit.)

whit537:
Ready for review! :-)

samchrisinger:
Pulitzer worthy no doubt. 

Changes lgtm; 🚢

PR #5978, opened 2016-07-15
Fix bug 'Contributors listed multiple times' [OSF-6690]

caspinelli:
## Purpose

Fix bug where contributors are listed multiple times on the My Projects page and clicking on a contributor does not show all of the projects they contributed to, only the most recent.

## Changes

Added code that made comparison between the contributors use the contributor's user id instead of the contributor id. Up until recently, those were the same, because the contributor id was incorrectly using the user's id. Now the contributor id is unique and the user's id is used instead.

## Side effects
None

## Ticket

https://openscience.atlassian.net/browse/OSF-6690

### DIFF
--- a/website/static/js/myProjects.js
+++ b/website/static/js/myProjects.js
@@ -662,6 +662,7 @@ var MyProjects = {
                 if (contributors) {
                     for(var i = 0; i < contributors.length; i++) {
                         var u = contributors[i];
+                        u.id = (u.id.indexOf('-') > -1) ? u.id.split('-')[1] : u.id;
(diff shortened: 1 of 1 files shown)


website/static/js/myProjects.js
```diff
@@ -662,6 +662,7 @@ var MyProjects = {
                 if (contributors) {
                     for(var i = 0; i < contributors.length; i++) {
                         var u = contributors[i];
+                        u.id = (u.id.indexOf('-') > -1) ? u.id.split('-')[1] : u.id;
```
sloria: This is fine. Elsewhere, we use /data/embeds/user/id, but this works too.

PR #5981, opened 2016-07-15
Fix list filter issue when field name did not match source field name [#OSF-6702]

abought:
## Purpose

Fix an APIv2 issue with 500 errors: could not filter on fields whose field name was different than the DB source field. (eg `filter[id]=<item_id>`. Affected multiple endpoints using ListFilterMixin; see JIRA ticket for list.

## Changes
Changed `parse_query_params` to avoid losing information about the modified field name.

## Side effects

None expected. Filtering should work the same as before on all ODMFilter endpoints, but now should also work for ListFilterMixin endpoints.

## Ticket
https://openscience.atlassian.net/browse/OSF-6702

### DIFF
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -171,7 +171,7 @@ class FilterMixin(object):
         return values
 
     def parse_query_params(self, query_params):
-        """Maps query params to a dict useable for filtering
+        """Maps query params to a dict usable for filtering
         :param dict query_params:
         :return dict: of the format {
             <resolved_field_name>: {
@@ -191,8 +191,9 @@ class FilterMixin(object):
                 op = match_dict.get('op') or self._get_default_operator(field)
                 self._validate_operator(field, field_name, op)
 
+                source_field_name = field_name
                 if not isinstance(field, ser.SerializerMethodField):
-                    field_name = self.convert_key(field_name, field)
+                    source_field_name = self.convert_key(field_name, field)
 
                 if field_name not in query:
                     query[field_name] = []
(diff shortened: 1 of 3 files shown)


abought:
Once this is clear to go, it would be helpful to merge this into the feature branch for osf-ember work:
https://github.com/CenterForOpenScience/osf.io/tree/feature/osf-ember

api/base/filters.py
```diff
@@ -171,12 +174,13 @@ def bulk_get_values(self, value, field):
         return values
 
     def parse_query_params(self, query_params):
-        """Maps query params to a dict useable for filtering
+        """Maps query params to a dict usable for filtering
```
brianjgeiger: `Useable` is a valid spelling.

api/base/filters.py
```diff
@@ -134,7 +134,7 @@ def _validate_operator(self, field, field_name, op):
                     detail="Field '{0}' does not support match operators in a filter.".format(field_name)
                 )
 
-    def _parse_date_param(self, field, field_name, op, value):
+    def _parse_date_param(self, field, source_field_name, op, value):
```
brianjgeiger: Can't you use convert_key() for this? https://github.com/centerforopenscience/osf.io/blob/f9e0eb63581d642956172dd3fb662e240ba334c0/api/base/filters.py#L215-L224

brianjgeiger:
Pass done. See about using existing methods and don't change valid spelling. 🐧

api/base/filters.py
abought: In this case, the spell checker default configuration prefers the change. I like to keep tooling warnings to a minimum so that the warnings I do so are worth the effort.

api/base/filters.py
abought: Yes. In fact, that's how we got the value passed into this function in the first place...

I'm kind of on the fence about this one, personally. Ideally we'd only call convert key if we needed to resolve to the model source field... but in this case, parsing *always* calls it anyway.

Resolving the serializer--> model field name requires having the field object. If we called `convert_key` from other methods (like `query_params_to_odm_query`), we'd need to repeat the calls to `_get_field_or_error` that were already done when query params first got parsed.

PR #5983, opened 2016-07-15
Add boolean field for whether user can comment [#OSF-6701]

abought:
## Purpose
We need a new field (similar to current_user_permissions) that identifies whether a user is allowed to comment on a given resource that supports comments. This will affect ember-osf work.

## Changes
Add `current_user_can_comment` boolean field to serializers for projects, registrations, wikis, and files.

## Side effects
None expected. Adds a new field without altering functionality.

## Ticket
https://openscience.atlassian.net/browse/OSF-6701


mfraezz:
LGTM :shipit:

api/files/serializers.py
```diff
@@ -5,8 +5,9 @@
 
 from modularodm import Q
 
-from website import settings
 from framework.auth.core import Auth, User
+from website import settings
```
brianjgeiger: This, incidentally, is why I discourage re-shuffling imports that don't actually affect functionality.

PR #5985, opened 2016-07-18
Replace Flash clipboards

Johnetordoff:
## Purpose

To replace current Flash based clipboard library (Zeroclipboard) with clipboard.js. Because

- Flash isn't supported on Apple mobile devices, so clipboards are broken on iPhone etc.
- People with Adbliock get  'Block...' bubbles above their clipboard buttons.
- Clipboard.js reduces the amount of code used on the OSF. 
- Clipboard.js is easier to use.

## Changes

- Deletes refference to Zeroclipboard and replaces them with functional clipboard.js

## Side effects

None.


sloria:
Is there a ticket for this?

sloria:
IIRC, we decided against using clipboard.js in the past because it does not support Safari.

PR #5987, opened 2016-07-18
[Hold][WIP] Feature/File Detail Page Porting File Editor [EOSF-96]

pattisdr:
❗ On Hold
## Purpose

Match the existing behavior on the file detail page, but use ember-osf addon.  Majority of file-detail page made with file components from ember-osf, very similar to behavior @aaxelb demonstrated in ember dummy app.  File editor functionality will need to be a component similar to existing filepage/editor.js.

## Changes

For this PR, navigating to `/<file_guid>/` or `/file_id/` will render the file-detail page containing ember components.  node guids will not work.  This is for demonstration only - routes will need to be changed.

- [x] Mimic ember-osf dummy app file detail page using components from ember-osf addon
- [x] Load both node and file on file-detail page (node needed for file browser widget to render).  Currently using RSVP.hash.  Defining in setupController not working.
- [x] Get file browser working
- [x] Get tags widget working (jquery.tagsinput needed to be in bower.json file for website-ember - thanks @abought)
- [ ] Routing (for demonstration only, I have commented out the GUID flask route and am using the ember guid route here. Needs to change. File guids should be handled by Ember app, all other guid's (node, etc) handled by Flask app.)
- [ ] Add file editing functionality (making ember component similar to filepage/editor.js
- [ ] Comment pane matching OSF 
- [ ] File navbar matching OSF 
- [ ] Styling

## Side effects

Dependent on the following PR's:
- Node relationship on files https://github.com/CenterForOpenScience/osf.io/pull/5967
- node added to file model https://github.com/CenterForOpenScience/ember-osf/pull/100

## Current Appearance of Page
screen shot 2016-07-14 at 1 07 40 pm

## Ticket

https://openscience.atlassian.net/browse/EOSF-96


abought:
In no way relevant at this moment, but I'm leaving a note here so we can remember some day...

Recently, various aspects of ember-osf have changed. 

In particular, usages of `NodeActionsMixin` and `TaggableMixin` have changed in ways that could affect widgets on this page. This affects call signatures, and return values of, various actions on these mixins, including for tagging.

PR #5988, opened 2016-07-18
[OSF-6565] Treebeard update SHA

ccfair:
## Purpose
The treebeard source was changed to fix a bug where you couldn't shift-click and select multiple files if you scrolled. [Treebeard ticket](https://openscience.atlassian.net/browse/OSF-4442).

## Changes
Changed the treebeard version that was being imported. Simple.

## Side effects

Might not work on osf, I'm not able to test because of some current waterbutler issues, but it should probably go to QA or someone who can test. It should work though.

## Ticket

[OSF-6565](https://openscience.atlassian.net/browse/OSF-6565)

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "raw-loader": "^0.5.1",
     "share": "0.7.27",
     "style-loader": "^0.8.3",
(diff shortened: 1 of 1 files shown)


brianjgeiger:
This needs merging with Develop. Did treebeard get updated again?

PR #5989, opened 2016-07-18
[CR] Keen UI updates [#OSF-6652]

felliott:
## Purpose

This PR addresses some, but not all, of the issues mentioned in [#OSF-6652].  I'm throwing this up here because the chart library was refactored muchly and will need a code re-review.  The remaining changes will be implemented in a follow-up PR, which should be much less invasive.

## Changes

- Display spinners after clicking "Update date range"
- Make "Popular pages" bar graph horizontal to avoid wrapping issues
- Bump up font size of chart labels
- Renamed "Visits over past week" to "Unique visits"
- date range form: initial version is text, click update button to reveal form
- more spacing between "from" and "until" in date range updater
- We exclude analytics for today, should reflect that in the picker: can select today, but end date is exclusive.
- not-UI related but good: refactor DateRangePicker constructor to take params
- Limit referrers/popular pages to top N results
- Referrers: rename "null" => "direct link"

## Side effects

May cause drowsiness.

## Ticket

[#OSF-6652]

https://openscience.atlassian.net/browse/OSF-6652

### DIFF
--- a/website/static/js/statistics.js
+++ b/website/static/js/statistics.js
@@ -2,234 +2,120 @@
 
 require('keen-dataviz/dist/keen-dataviz.min.css');
 
+var oop = require('js/oop');
 var $osf = require('js/osfHelpers');
 var keenDataviz = require('keen-dataviz');
 var keenAnalysis = require('keen-analysis');
 
-var ProjectUsageStatistics = function(keenProjectId, keenReadKey, params){
-    var self = this;
 
-    params = params || {};
-    self.MAX_DISPLAY_ENTRIES = params.maxDisplayEntries || 10;
+var UserFacingChart = oop.defclass({
+    constructor: function(params) {
+        var self = this;
+        params = params || {};
+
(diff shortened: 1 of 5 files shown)


website/static/js/statistics.js
```diff
@@ -2,241 +2,378 @@
...
+     * parameters of the query.  The timeframe for the analysis is be filled in by the query()
+     * method, which calls baseQuery().
+     *
+     * See: https://keen.io/docs/api/?javascript#analyses
+     *
+     * @method baseQuery
+     * @return {Object}
+     */
+    baseQuery: function() {
+        throw 'Concrete class does not provide a "baseQuery" method!';
```
sloria: Minor: As per COSDev, always throw an instance an instance of `Error`. This ensures that the error will get picked up by Raven.

```js
throw new Error('Concrete class..')
```

sloria:
Refactor looks good. No blocking issues. The nits can go in the follow up PR.

PR #5990, opened 2016-07-18
[OSF-3520] Simplified code, changed variable

ccfair:
## Purpose
Addl Dev on [[OSF-3520] Registrations draggable](https://github.com/CenterForOpenScience/osf.io/pull/5975). Make registration files not able to drag and drop.
## Changes
From the original PR, I simplified my code my removing the functions, and corrected the variable I was using to check if the node was a registration. `node.isRegistered` --> `node.isRegistration`

## Side effects
None that I can think of.

## Ticket
[OSF-3520](https://openscience.atlassian.net/browse/OSF-3520)

### DIFF
--- a/website/static/js/pages/files-page.js
+++ b/website/static/js/pages/files-page.js
@@ -3,6 +3,7 @@
 var $ = require('jquery');
 var $osf = require('js/osfHelpers');
 var Fangorn = require('js/fangorn');
+var node = window.contextVars.node;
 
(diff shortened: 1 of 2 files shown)


mfraezz:
The PR title could be a little more descriptive, but otherwise LGTM. Verified that fangorn items items on registrations were no longer draggable, and that non-registered fangorn items were. Dropping this in PCR 🏭

PR #5991, opened 2016-07-18
[Feature] APIv2 support for Amazon S3 [OSF-6656][OSF-4067][OSF-3504]

mfraezz:
## Purpose

Primary: Add support for folder (in this case, bucket) listing / setting through APIv2.
Opportunistically fixed two other tickets relating to S3:
* Expose bucket location after linking (UX delay prohibitive for showing location for every bucket before linking)
* Add link to information about bucket location when creating

## Changes

* Refactor views
* Refactor models
* Refactor serializer
* Refactor UI -- now uses folder picker
* Update tests -- more behavior is now generalizable
* Add migration script

## Side effects
None known.

## Deployment Notes
```
python -m scripts.s3.migrate_folder_language --dry
```
* Requires a small amount of downtime; makes outgoing calls to S3 for bucket locations, and renames a db column (`bucket` --> `folder_id`).

## QA Notes
* Test FolderPicker UI for S3 and at least one other addon
* Ensure that files page still works, and displays the location of the linked bucket in the form `<name> (<location>)`
* It is expected that folder listing through the API will not include `child` links -- only a list of top-level buckets

Screenshots to follow:

## Ticket
[OSF-6656](https://openscience.atlassian.net/browse/OSF-6656)
[OSF-4067](https://openscience.atlassian.net/browse/OSF-4067)
[OSF-3504](https://openscience.atlassian.net/browse/OSF-3504)


api/addons/serializers.py
```diff
@@ -19,6 +19,8 @@ class Meta:
     })
 
     def get_absolute_url(self, obj):
+        if obj['addon'] in ('s3', ):
```
brianjgeiger: Why does only s3 return in this case?

scripts/s3/migrate_folder_language.py
```diff
@@ -0,0 +1,100 @@
...
+
+def migrate(dry_run=False):
+    bucket_name_location_map = {}
+    sns_collection = database['s3nodesettings']
+    logger.info('Migrating all {} S3NodeSettings documents: {}'.format(
+        sns_collection.count(), [s['_id'] for s in sns_collection.find()]
+    ))
+    for document in sns_collection.find():
+        sns_collection.find_and_modify(
+            {'_id': document['_id']},
```
brianjgeiger: L39 looks nice and respectful of multiple runs in case of failure. I'm guessing this particular query won't actually take long to run, but would it be feasible or desirable to make it also check for documents where bucket is set or anything?

api/addons/serializers.py
mfraezz: Because S3 is currently the only addon that doesn't have nested selectable ~~buckets~~ folders. This prevents a `children` link from appearing, as following it would exclusively result in either empty data or an error.

website/addons/s3/static/s3UserConfig.js
```diff
@@ -30,13 +30,52 @@ function ViewModel(url) {
     self.accounts = ko.observableArray();
 
     ChangeMessageMixin.call(self);
+    self.options = {
```
brianjgeiger: So you got rid of s3NodeConfig and replaced it with node-cfg but you're keeping s3UserConfig?

scripts/s3/migrate_folder_language.py
mfraezz: Currently, the entire migration is rolled back in the event of a failure (one caused by an exception, at least), so this is technically "respectful of multiple runs".

>would it be feasible or desirable to make it also check for documents where bucket is set or anything?

Yes and no, respectively. What this loop is really doing is closer to the rename of a database column... but the only documented indication that pymongo could rename a column proved unsuccessful upon manual testing. Something like 
```python
list(sns_collection.find({'$eq': {'bucket': None}})) + list(sns_collection.find({'$ne': {'bucket': None}}))
```
would probably return the actual set of things that need to be iterated over... but the chances of this mattering are close to zero.

mfraezz: Also, just to confirm: it worked for me locally with a recent dump of staging.

website/addons/s3/static/s3UserConfig.js
mfraezz: I migrated existing s3-specific functionality into the generic class `OauthAddonNodeConfig`, which provides the folder picker UI and a set of automated tests for addons that use it. We eventually want all of them that can to use it for both ease of maintenance and UX consistency. While S3 isn't technically an `OauthAddon`, much of the back-end treats it as such -- our current naming convention may be a misnomer.

The user settings page has a generalized viewmodel for addons that *actually* support OAuth, but given that it worked with relatively few changes (really just changing how certain functions were accessed to allow the credentials modal to work), and that none of the tickets this was intended to solve were directly related to front-end maintenance, I thought writing another generalized viewmodel for addon_user_settings with modals that receive hard-coded credentials was out of scope of this PR. Possibly in scope of the [follow-follow-up ticket](https://openscience.atlassian.net/browse/OSF-6678), however

scripts/s3/migrate_folder_language.py
brianjgeiger: We will have to add a blocker issue to the go-nogo ticket to ensure that the migration is tested on Production data while this is going through QA. If it is at all unreliable, we may consider refactoring so that it's continuable, but otherwise should be fine.

brianjgeiger:
Seems good to me. 🎉 🐧 🎉 We'll check with QA on how long to get it tested and test this like a Feature.

sloria:
Feature branch created: `feature/api-s3`.

PR #5993, opened 2016-07-19
Importing contributors from parent does not paginate [OSF-4762]

caspinelli:
## Purpose

Fix bug where when importing contributors in a component from a parent node, the results do not paginate like a normal search, and there is some weird behavior where the results of an import from parent will appear on whatever page of the pagination from a normal search, and if you move to another page and then back to the page the import from parent appeared on, it is gone.

## Changes

Since, unlike the normal search results, the import from parent does not have API support for pagination (normal search request to API includes the page number that is wanted), had to add frontend code that splits the results of an import from parent into the page requested, and added paginator for the import from parent results.
Also moved the import from parent button from above the results header to below the search box because it makes more sense for it to be up there and also makes for an easier solution to this bug
Before:
screen shot 2016-07-19 at 10 00 38 am
After:
screen shot 2016-07-19 at 9 59 55 am

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-4762

### DIFF
--- a/website/static/js/contribAdder.js
+++ b/website/static/js/contribAdder.js
@@ -199,32 +199,37 @@ AddContributorViewModel = oop.extend(Paginator, {
         this.fetchResults();
     },
     fetchResults: function () {
-        var self = this;
-        self.doneSearching(false);
-        self.notification(false);
-        if (self.query()) {
-            return $.getJSON(
-                '/api/v1/user/search/', {
-                    query: self.query(),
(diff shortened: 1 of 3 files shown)


brianjgeiger:
Looks like it should be good. Might could be DRYd a bit, but there's a good chance this page will be redone soon, so it's not worth overengineering. 🎉 🐧 🎉

PR #5996, opened 2016-07-19
[hold] [ShareFiles/Preprints] Add query param to file detail endpoint to give file guid

Johnetordoff:
## Purpose

ShareFiles requires that files that are uploaded immediately have short, convenient guid links for easy sharing. Problematically files only get guids when the File View Page is visited. This feature allows us to add guids using the API v2 without visiting the file View Page.

## Changes

Modifies the get_object method of the view to accept query param 'giveGuid' and creates that id if true and user has admin permission. Also changes API docs slightly.

## Side effects

None that I know of.

## Discussions

https://openscience.atlassian.net/browse/OSF-6503
https://github.com/CenterForOpenScience/osf.io/pull/5836

### DIFF
--- a/api/files/views.py
+++ b/api/files/views.py
@@ -328,7 +328,7 @@ class FileDetail(JSONAPIBaseView, generics.RetrieveUpdateAPIView, FileMixin):
     def get_object(self):
         user = get_user_auth(self.request).user
 
-        if self.request.GET.get('giveGuid', False) and self.get_node().has_permission(user, 'admin'):
+        if self.request.GET.get('create_guid', False) and self.get_node().has_permission(user, 'admin'):
             self.get_file(check_permissions=True).get_guid(create=True)
 
         return self.get_file()
--- a/api/files/views.py
+++ b/api/files/views.py
@@ -300,12 +300,8 @@ class FileDetail(JSONAPIBaseView, generics.RetrieveUpdateAPIView, FileMixin):
 
     ##Query Params
 
-    For this endpoint,
-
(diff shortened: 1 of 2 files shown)


samchrisinger:
+1

api/files/views.py
```diff
@@ -300,8 +300,12 @@ class FileDetail(JSONAPIBaseView, generics.RetrieveUpdateAPIView, FileMixin):
 
     ##Query Params
 
-    For this endpoint, *none*.  Actions may permit or require certain query parameters.  See the individual action
-    documentation.
+    For this endpoint,
+
+    + giveGuid={bool} if true will give the file metadata a guid, a unique identifier for internal use. You must be an
```
sloria: Let's keep this undocumented for now. We might change this API in the future.

api/files/views.py
```diff
@@ -326,8 +330,12 @@ def get_node(self):
 
     # overrides RetrieveAPIView
     def get_object(self):
-        return self.get_file()
+        user = get_user_auth(self.request).user
+
+        if self.request.GET.get('giveGuid', False) and self.get_node().has_permission(user, 'admin'):
```
sloria: We don't typically camel-case query params in APIv2. How about just `guid` or `create_guid`?

sloria:
Can you please add a test for the added query param?

sloria:
Pass finished.

sloria:
Also, I've created a new JIRA ticket for this, since this PR doesn't directly address the issue from the original ticket.

Here is the new ticket: https://openscience.atlassian.net/browse/OSF-6727

Johnetordoff:
Closing in favor of #6019

PR #5997, opened 2016-07-19
subscribe users to global notifications when confirmed [OSF-5658]

leb2dg:
## Purpose

Subscribe users to global notifications upon confirmation.

## Changes

* check the proper users subscriptions
* add listener for `user_confirmed`

## Side effects

Users are subscribed to user subscriptions on confirmation.

## Ticket

https://openscience.atlassian.net/browse/OSF-5658


leb2dg:
Close in favor of #5998

PR #6000, opened 2016-07-19
add migration for global notification subscriptions [OSF-5658]

leb2dg:
## Purpose

Subscribe all current users to default notifications.

## Changes

* Added migration for global subscriptions

## Side effects

All users will have global notifications set to instantly unless they already have global notifications.

## Ticket

https://openscience.atlassian.net/browse/OSF-5658

### DIFF
--- a/scripts/add_global_subscriptions.py
+++ b/scripts/add_global_subscriptions.py
@@ -25,17 +25,18 @@ def add_global_subscriptions():
     user_events = constants.USER_SUBSCRIPTIONS_AVAILABLE
 
     for user in models.User.find():
-        for user_event in user_events:
-            user_event_id = to_subscription_key(user._id, user_event)
+        if user.is_active and user.is_registered:
+            for user_event in user_events:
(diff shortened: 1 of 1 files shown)


scripts/add_global_subscriptions.py
```diff
@@ -0,0 +1,44 @@
...
+
+app = init_app()
+
+
+def add_global_subscriptions():
+
+    notification_type = 'email_transactional'
+    user_events = constants.USER_SUBSCRIPTIONS_AVAILABLE
+
+    for user in models.User.find():
```
sloria: We'll probably want to exclude inactive users (i.e. any users that don't meet the criteria for the `User.is_active` property).

leb2dg: Oh yep. I forgot about unregistered users as well.

leb2dg:
@sloria, updated 🎉

sloria:
Ran locally. Seems to have worked. 🐙

PR #6001, opened 2016-07-19
Feature/password verification

erinspace:
## Purpose
Fix up some things that were broken with the refactor of the password verifications

Tested on the following pages:
- Front splash page
- Create an account page
- Account Settings page
- Forgot password page
- Claim account page

For the following actions:
- Password strength bar fills and empties
- Cannot use your email address as your password
- Cannot click submit when the password is invalid
- Cannot hit enter to submit when the password is invalid
- Password is successfully reset on completion

## Changes
- Change order of constructor call so that ```self.password``` is defined before it's checked
- Add back checking for frontend validation of using email as password on the claim account page
- Add back disabling ENTER key also when validation does not pass on the front end
- Add back success message back for signing up on the ```login``` page

## Side effects
None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-6689


website/static/js/passwordForms.js
```diff
@@ -243,15 +266,51 @@ var SignUpViewModel = oop.extend(BaseViewModel, {
...
+                        return false;
+                    }
+                }
+            }
+        );
 };
 
 var SignUp = function(selector) {
-    $osf.applyBindings(new SignUpViewModel(), selector);
+    this.SignUpViewModel = new SignUpViewModel();
```
sloria: Minor: no need to assign to `this`.

website/static/js/passwordForms.js
```diff
@@ -243,15 +266,51 @@ var SignUpViewModel = oop.extend(BaseViewModel, {
...
+    $osf.applyBindings(this.SignUpViewModel, selector);
+        // Necessary to prevent enter submitting forms with invalid frontend zxcvbn validation
+        $(selector).keypress(
```
sloria: Minor: Fix indentation.

website/static/js/passwordForms.js
```diff
@@ -243,15 +266,51 @@ var SignUpViewModel = oop.extend(BaseViewModel, {
...
+                        return false;
+                    }
+                }
+            }
+        );
 };
 
 var SetPassword = function(selector) {
-    $osf.applyBindings(new SetPasswordViewModel(), selector);
+    this.SetPasswprdViewModel = new SetPasswordViewModel();
```
sloria: Typo here. Also--as mentioned in my previous comment--no need to assign to `this`.

```js
var viewModel = new SetPasswordViewModel();
// ...
    if (viewModel.password.isValid()) {
         // ... 
    }
```

website/static/js/passwordForms.js
```diff
@@ -119,6 +119,30 @@ var ChangePasswordViewModel = oop.extend(BaseViewModel, {
 var SetPasswordViewModel = oop.extend(BaseViewModel, {
     constructor: function () {
         var self = this;
+        // Call constructor at the beginning so that self.password exists
+        self.super.constructor.call(this);
```
sloria: Are you certain that this doesn't break validation? Other ViewModels have a comment that indicate that the constructor needs to be called after the observables are declared: https://github.com/CenterForOpenScience/osf.io/blob/607e4d7ee05b4e310a7400b0183c8a68dcc7e92c/website/static/js/passwordForms.js#L107-L108

sloria: Oh, but this class doesn't override `getValidatedFields`, so I think this is OK. 👍

website/static/js/passwordForms.js
```diff
@@ -243,15 +266,51 @@ var SignUpViewModel = oop.extend(BaseViewModel, {
 
 /** Wrapper classes */
 var ChangePassword = function(selector) {
-    $osf.applyBindings(new ChangePasswordViewModel(), selector);
+    this.ChangePasswordViewModel = new ChangePasswordViewModel();
+    $osf.applyBindings(this.ChangePasswordViewModel, selector);
+        // Necessary to prevent enter submitting forms with invalid frontend zxcvbn validation
+        $(selector).keypress(
+            event => {
```
sloria: This looks invalid; I don't believe ES6 is valid in this file. Are you sure that this callback is being triggered?

sloria:
Pass finished. A few nitpicks. https://github.com/CenterForOpenScience/osf.io/pull/6001/files#r71588093 is worth looking at.

website/static/js/passwordForms.js
erinspace: It does indeed work for me locally for its intended purpose here. I'll change it though to avoid any confusion

sloria: Oooh... this is working in Chrome (and probably Firefox) because ES6 (or at least some of it) is supported natively in these browsers! 

osf___home
 
Safari looks like a no-go, though.

how_people_build_software_ _github

I wouldn't count on IE, either.

erinspace:
@sloria OK, ready for another pass!

Had to put the call to `self.super.constructor.call(this);` above the rest of the code in the ChangePasswordViewModel as well because `self.password` wasn't getting evaluated correctly there either.  Everything seems OK though in local testing. Not sure how to handle since it does seem to ovverride `getValidatedFields`. You're still not allowed to use the samenew  password though as your old one on the settings page so maybe this OK?

sloria:
@erinspace Looks like you might have merged develop into `feature/password-verification`. Can you please create a new branch and PR with just your changes? Make sure to branch off of `feature/password-verification` and keep it separate from develop.

PR #6002, opened 2016-07-20
Upgrade markdown-it-video [OSF-6130]

amyshi188:
**Purpose**

Markdown-it-video has upgraded to a new version that has Vine support and bug fixes. The version used by the OSF did not have these changes, so an upgrade was necessary. After the upgrade was implemented, testing was done to ensure everything still works. Autocomplete for Vine and Prezi has also been added. 

**Changes**

Vine and Prezi can be embedded into the Wiki editor by using the same Markdown syntax used for Youtube and Vimeo. This was possible after updating the version number in package.json.

@[youtube](url)
@[vimeo](url)
@[vine](url)
@[prezi](url)

Adding snippets to ace-markdown-snippets.js for Vine and Prezi allowed for both websites to be autocompleted as well as Youtube and Vimeo. 

image

image

**Side effects**

N/A

**Ticket**

https://openscience.atlassian.net/browse/OSF-6130

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "markdown-it-ins-del": "^0.1.1",
     "markdown-it-sanitizer": "~0.3.0",
     "markdown-it-toc": "~1.1.0",
-    "markdown-it-video": "~0.2.1",
+    "markdown-it-video": "~0.4.0",
     "mime-types": "~2.0.12",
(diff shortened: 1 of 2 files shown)


brianjgeiger:
Looks great! 🎉 🐧 🎉

PR #6008, opened 2016-07-20
Feature/at mentions contributor sorting [OSF-5948] [OSF-6734]

brianjgeiger:
## Purpose

Make default contributor sort order a) a thing; and b) match what we use on the front-end.

## Changes

1. Test
2. Add an index to allow us to know the order from the list structure
3. Sort by that list by default

## Side effects

Shouldn't be, although the index is currently read-only, so we'll need to do more work. Added to at-mentions because it's related to a problem they're having, but they may fix it another way so we could save this for an IR.

## Ticket

https://openscience.atlassian.net/browse/OSF-6734

### DIFF
--- a/api_tests/nodes/views/test_node_contributors_list.py
+++ b/api_tests/nodes/views/test_node_contributors_list.py
@@ -180,6 +180,33 @@ class TestNodeContributorList(NodeCRUDTestCase):
         assert_equal(res.json['data'][1]['embeds']['users']['data']['attributes']['full_name'], 'Robert Jackson')
         assert_equal(res.json['data'][1]['attributes'].get('unregistered_contributor'), 'Bob Jackson')
 
+    def test_contributors_order_is_the_same_over_multiple_requests(self):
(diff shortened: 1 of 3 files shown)


api_tests/nodes/views/test_node_contributors_list.py
```diff
@@ -180,6 +180,33 @@ def test_unregistered_contributors_show_up_as_name_associated_with_project(self)
...
+    def test_contributors_order_is_the_same_over_multiple_requests(self):
+        self.public_project.add_unregistered_contributor(
+            'Robert Jackson',
+            'robert@gmail.com',
+            auth=Auth(self.user),
+            save=True
+        )
+
+        for i in range(0,10):
+            new_user = AuthUserFactory()
```
sloria: Minor: use `UserFactory` if you don't need to use basic auth in the tests. `AuthUserFactory` instances take longer to instantiate because they bcrypt passwords.

PR #6010, opened 2016-07-21
Add date_modified field for Users and Institutions. [OSF-6642]

adlius:
## Purpose

Adding a date_modified field to the api of Users and Institutions.

## Changes

Added date_modified field to framework/auth/core.py for User class.
Added node_updated signal to website/project/signals.py
In website/project/utils.py, connect to the node_updated signal and do a post commit update to the date_modified field of the user.
In api/users/serializers.py, added the serializer field for date_modified.
In api/institutions/serializers.py, added the serializer field for date_modified. (Note: because Institution class is a wrapper over Node class, no new field is needed as the date_modified field is already included and updated like a node.)

## Side effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/OSF-6642


website/project/utils.py
```diff
@@ -45,3 +48,18 @@ def recent_public_registrations(n=10):
...
+
+
+@run_postcommit(once_per_request=False, celery=True)
+@app.task(max_retries=5, default_retry_delay=60)
+def update_date_modified_task(node_id):
+    node = Node.load(node_id)
+    now = datetime.utcnow()
+    for contrib in node.contributors:
+        contrib.date_modified = now
+        contrib.save()
```
brianjgeiger: Do not save the entire contributor object every time any node the contributor was attached to is updated. You are going to have to just update the one field in the most efficient manner possible. There are other examples of postcommit tasks which should give you an idea of how to do this.

brianjgeiger:
I think you're going to need to be more specific with your use case in the purpose, because I don't think this is going to do everything you want it to. You want to scrape the user profile in the case of change, yes? But you're only updating the date_modified when the nodes are changed. So if someone updates their social information or their employment or similar (really, anything about the user itself), you aren't going to be notified of a change.

brianjgeiger:
This is not quite ready. Update the purpose, be specific, and do not kill the server when updating the date_modified field.

brianjgeiger:
Failing test, but I think is otherwise good (getting a second opinion on that). Fix the test and this is probably ready to go. Pass done. 🐧

website/project/utils.py
```diff
@@ -45,3 +49,24 @@ def recent_public_registrations(n=10):
...
         n -= 1
         yield reg
+
+
+@signals.node_updated.connect
+def update_date_modified_for_users(node):
+    update_date_modified_task(node._id)
+
+
+@run_postcommit(once_per_request=False, celery=True)
```
brianjgeiger: once_per_request to True, please.

adlius:
In test_user.py, TestUserMerging.test_merge was checking whether two merged users have fields with the same values. Since the date_modified field of the user would be set when save() is called on the user object, the field may be different for two users.
Thus, a 'different_fields' list is added to the test and the test shall exclude fields in the list when checking for equality of the fields.

tests/models/test_user.py
```diff
@@ -420,16 +420,22 @@ def test_merge(self, mock_get_mailchimp_api):
...
         expected.update(calculated_fields)
         for key in default_to_master_user_fields:
             expected[key] = getattr(self.user, key)
 
         # ensure all fields of the user object have an explicit expectation
+        # except for 'different_fields'
         assert_equal(
             set(expected.keys()),
-            set(self.user._fields),
+            set(self.user._fields).difference(different_fields),
```
brianjgeiger: This isn't bad, but is there a way you could use the calculated_fields structure to get the proper expected value rather than skipping it entirely? Even if it's not much better information, I'd rather use the existing structure to do the test than to add another exception. If it's too crazy, then it'll probably be fine this way.

brianjgeiger:
Potential test update. Pass done. 🐧

tests/models/test_user.py
adlius: @brianjgeiger I thought about adding it to the calculated_field but felt it was a bit of a stretch. The main reason to add a different_fields list is for future expansion, in case we add more fields to the user that allows different value for the merged users. But I guess the possibility of adding another such field is pretty low. In hindsight it's reasonable to add to the calculated_fields rather than excluding the different_fields, which makes it a bit less readable. I'll go ahead and do what you suggested.

adlius:
Test update is done. In calculated_fields, 'date_modified' is first set to None. After the last save() call on self.user, the 'date_modified' of calculated_fields is set to self.user.date_modified for the test to pass.
Will move this ticket to CR once Travis is done.

PR #6012, opened 2016-07-21
[#OSF-6420]Fix modal message inconsistent for folder deletion

chennan47:
## Purpose

Fix the delete folder pop test for github is different from other addons.

## Changes

When you try to delete a folder with content on github,
Before the change:

image

After the change:

image
## Ticket

### DIFF
--- a/website/addons/github/static/githubFangornConfig.js
+++ b/website/addons/github/static/githubFangornConfig.js
@@ -85,7 +85,7 @@ function _removeEvent (event, items) {
         // If all items can be deleted
         if(canDelete){
             mithrilContentMultiple = m('div', [
-                    m('p', 'This action is irreversible.'),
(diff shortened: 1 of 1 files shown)


brianjgeiger:
Looks solid. 🎉 🐧 🎉

website/addons/github/static/githubFangornConfig.js
```diff
@@ -85,7 +85,7 @@ function _removeEvent (event, items) {
         // If all items can be deleted
         if(canDelete){
             mithrilContentMultiple = m('div', [
-                    m('p', 'This action is irreversible.'),
+                    m('p.text-danger',, 'This folder and ALL its contents will be deleted. This action is irreversible.'),
```
sloria: I removed the extra comma when I merged.

PR #6015, opened 2016-07-21
[WIP][hold] Demonstrate node modules + osf code in ember app [#EOSF-116]

abought:
## Purpose
To reimplement the OSF front-end, we need to be able to consume modules that are not available on bower.

This PR adds support for importing npm modules without additional per-module shims or one-off boilerplate.

Additionally, it demonstrates using this to import our markdown wrapper code in `website/static/js` in a limited testcase. (possibly relevant to @pattisdr )

## Changes
- Add a basic markdown rendering template helper 
- Connect app to browserify and evaluate whether this works as intended.

### TODO
To test this, you must comment out `require('highlight-css')` in the `js/markdown.js` file. This is an alias defined in webpack.common.config, so browserify doesn't know the module by that name. We may need to similarly evaluate other built modules.

- [ ] Allow named/aliased entities (eg `highlight-css`)
- [ ] Don't use relative paths- "relative to what" changes from file to file
- [ ] Get tests passing.  (see https://github.com/ef4/ember-browserify/issues/14 )

## Side effects
This PR is a proof of concept. It is possible, but it is for us to determine whether it is wise. In particular we would need to keep any included OSF modules lightweight and focused. We should not need, eg, knockout dependencies.

- ember-browserify comes well-recommended, but it could easily lead to bloated vendor files if a nested layer of things imports. We should monitor build file size carefully and ensure dependent libraries such as jquery are not being included twice.

## Alternatives
The other leading alternative would be custom broccoli build tree manipulation, eg `treeForVendor` (a function available to addons though not apps). This might require additional boilerplate for each separate module, however, and it could be limited wrt using OSF code directly in ember.

## Ticket
https://openscience.atlassian.net/browse/EOSF-116


brianjgeiger:
It's Autumn Cleaning time, so I'm closing this PR. Feel free to re-open when the issue is no longer on hold.

PR #6016, opened 2016-07-22
[Feature] Verify Admin Scope [OSF-6725]

mfraezz:
## Purpose
We want to use the admin scope things, and we need to verify that it's as secure as we want it to be.

Resolves (some) blockers for [OSF-6727](https://openscience.atlassian.net/browse/OSF-6727) and [OSF-6724](https://openscience.atlassian.net/browse/OSF-6724) / #5942 

## Changes

- [X] - Modify a route to have admin scope-specific behavior (`/v2/`)
- [X] - Verify that the behavior of that endpoint is correct:
- [X] ---- PATs with admin scope and cookied requests exhibit that behavior
- [X] ---- PATs without admin scope and basic auth requests do not exhibit that behavior
- [X] - Make sure that arbitrary users cannot get admin 
- [X] - Unit tests to verify the above

**TODO** (Someone with RS access after this goes in):
- [ ]  Runscope tests (Ghost inspector test of the BAPI if Runscope won't work) 

## Side effects
Until the admin scope is available outside of `DEV_MODE`, `RequestHasAdminScope` will only be useful for cookied requests.

## Ticket
[OSF-6725](https://openscience.atlassian.net/browse/OSF-6725)

### DIFF
--- a/api/base/permissions.py
+++ b/api/base/permissions.py
@@ -3,6 +3,8 @@ import operator
 from django.core.exceptions import ImproperlyConfigured
 from rest_framework import exceptions, permissions
 
+from api.base.utils import has_admin_scope
+
 from framework.auth import oauth_scopes
 from framework.auth.cas import CasResponse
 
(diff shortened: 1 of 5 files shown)


brianjgeiger:
So it looks like BAPI should be able to submit admin requests based on this. I don't know if that's a problem. Probably not?

brianjgeiger:
This seems good. 🎉 🐧 🎉

PR #6017, opened 2016-07-22
[OSF-5808] Retractions should show the date retracted in the metadata

mluke93:
## Purpose

Tracks the datetime in which a registration has been withdrawn, and then outputs the data to the Withdrawn Registration page.

## Changes

Added new field to the model and serializer to record the datetime that a registration is withdrawn.

## Side effects

n/a

## Ticket

https://openscience.atlassian.net/browse/OSF-5808


mluke93:
There's some funky stuff going on with Git. Let me fix before you look this over.

PR #6018, opened 2016-07-22
[OSF-5808] Retractions should show the date retracted in the metadata

mluke93:
## Purpose

This improvement records the date and time in which a Registration is withdrawn.

## Changes

A new field has been added to the model and serializer, to record the date and time a Registration is withdrawn, and then displays the data on the "Withdrawn Registration" page.

## Side effects

n/a

## Ticket

https://openscience.atlassian.net/browse/OSF-5808


mluke93:
The last PR was messed up due to a branching error, so I created a new, correct one.

brianjgeiger:
This looks like it will need a migration for any records that were withdrawn before this code went in, but otherwise seems good. Pass done. 🐧

PR #6019, opened 2016-07-22
[Feature] API-based file guid creation [OSF-6727]

mfraezz:
## Purpose
Allow for the creation of file guids through the API. Blocks the Public Files feature #5942 

Based on #5996 , with security checks from #6016 

## Changes
 Allow a kwarg passed to files detail `GET` to generate a guid, if this request is coming from the osf (as indicated by a valid cookie) or from an OAuth token with the (restricted) `osf.admin` scope.

## Side effects
5-char guid availability will continue to decrease, but enough restrictions are in place that this shouldn't be much worse than the current rate. Consider bumping the `__guid_min_length__` of files up to 8-12

## Ticket
[OSF-6727](https://openscience.atlassian.net/browse/OSF-6727)


brianjgeiger:
Can you merge develop into this so we get the minimized diff?

mfraezz:
@brianjgeiger, updated.

samchrisinger:
I'd like to see resolution of the [discussion here](https://openscience.atlassian.net/browse/OSF-6503) before this moves forward.
@JeffSpies @sloria

mfraezz:
This was opened primarily to allow that discussion to continue while providing a layer of security that 5996 didn't have; it's not intended to go in immediately.

brianjgeiger:
Pass finished. Looks good, but there are concerns about what happens when this goes into full effect on both Admin scoped tickets and Production via cookies. We will need to find a good way to rate limit before then. Please make a ticket to that effect. 🎉 🐧 🎉

mfraezz:
Ticket created, `TODO` added in https://github.com/CenterForOpenScience/osf.io/pull/6019/commits/3e6d5c15510be3d7b9b6692bed3312dd658ad4a6

PR #6020, opened 2016-07-22
[OSF-4670] Do not show a profile page for an unconfirmed user

DanielSBrown:
## Purpose

- Before, when an account was made but not yet confirmed, their page was still able to be accessed.  

- Makes it so when an unconfirmed user's profile page is attempted to be accessed, it displays the text "Account not yet confirmed.  The profile page could not be displayed as the user has not confirmed the account" and throws and HTTPerror.

## Changes

- Added the decorator ``must_be_confirmed`` in ``decorators.py`` that checks if the user on the page being visited has activated their account.  If so, it returns normally.  If it is not activated, it raises the error and displays the message.

- Wrapped the function ``profile_view_id`` in ``profile/views.py`` that is accessed when a page is loaded up 

 
Here is a picture of the changes on a page of an unconfirmed user:
screen shot 2016-07-22 at 12 10 29 pm

## Ticket

- https://openscience.atlassian.net/browse/OSF-4670

### DIFF
--- a/framework/auth/decorators.py
+++ b/framework/auth/decorators.py
@@ -10,8 +10,8 @@ from framework.auth import cas
 from framework.auth import signing
 from framework.flask import redirect
 from framework.exceptions import HTTPError
-
 from .core import Auth
+from .core import User
 
 
 def collect_auth(func):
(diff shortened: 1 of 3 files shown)


HalcyonChimera:
Would it make sense to have a link to claim the account here? i.e. "Is this your Account? Click here to learn how to claim/confirm it..." This might need to be a different ticket (if it's something we want to do), but it's functionality I've seen on other sites before.

brianjgeiger:
Perfect. 🎉 🐧 🎉

PR #6021, opened 2016-07-22
Feature/password verification

erinspace:
Fixing messed up #6001 

## Purpose
Fix up some things that were broken with the refactor of the password verifications

Tested on the following pages:
- Front splash page
- Create an account page
- Account Settings page
- Forgot password page
- Claim account page

For the following actions:
- Password strength bar fills and empties
- Cannot use your email address as your password
- Cannot click submit when the password is invalid
- Cannot hit enter to submit when the password is invalid
- Password is successfully reset on completion

## Changes
- Change order of constructor call so that ```self.password``` is defined before it's checked
- Add back checking for frontend validation of using email as password on the claim account page
- Add back disabling ENTER key also when validation does not pass on the front end
- Add back success message back for signing up on the ```login``` page

## Side effects
None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-6689

### DIFF
--- a/website/static/js/passwordForms.js
+++ b/website/static/js/passwordForms.js
@@ -119,7 +119,7 @@ var ChangePasswordViewModel = oop.extend(BaseViewModel, {
 var SetPasswordViewModel = oop.extend(BaseViewModel, {
     constructor: function () {
         var self = this;
-        // Call constructor at the begining so that self.password exists
+        // Call constructor at the beginning so that self.password exists
         self.super.constructor.call(this);
 
         // pick up the email from contextVars if we can't get it from first typing it in
--- a/website/static/js/passwordForms.js
+++ b/website/static/js/passwordForms.js
@@ -244,7 +244,19 @@ var SignUpViewModel = oop.extend(BaseViewModel, {
 
 /** Wrapper classes */
 var ChangePassword = function(selector) {
-    $osf.applyBindings(new ChangePasswordViewModel(), selector);
+    this.ChangePasswordViewModel = new ChangePasswordViewModel();
+    $osf.applyBindings(this.ChangePasswordViewModel, selector);
+        // Necessary to prevent enter submitting forms with invalid frontend zxcvbn validation
+        $(selector).keypress(
+            event => {
+            // If the enter key is pressed to submit a form, check if the password is valid
+            if (event.which === 13) {
+                    if (!this.ChangePasswordViewModel.password.isValid()) {
+                        return false;
+                    }
+                }
+            }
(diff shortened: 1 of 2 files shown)


website/static/js/passwordForms.js
```diff
@@ -92,6 +92,31 @@ var BaseViewModel = oop.extend(ChangeMessageMixin, {
 var ChangePasswordViewModel = oop.extend(BaseViewModel, {
     constructor: function () {
         var self = this;
+
+        // Call constructor at the beginning so that self.password exists
+        self.super.constructor.call(this);
```
sloria: Since ChangePasswordViewModel does override getValidatedObservables, you will need to make sure that `self.validatedFields`  gets populated correctly.

sloria: How about adding `password: self.password` to the return value of getValidated fields, then manually assigning `self.validatedFields` and `self.validatedObservables` in the constructor.

```js
self.validatedObservables = self.getValidatedFields();
self.validatedFields = ko.validatedObservable(self.validatedObservables);
```

By the way, you'll probably want to comment the above lines; it's a bit of a clumsy workaround.

website/static/js/passwordForms.js
```diff
@@ -243,15 +289,42 @@ var SignUpViewModel = oop.extend(BaseViewModel, {
 
 /** Wrapper classes */
 var ChangePassword = function(selector) {
-    $osf.applyBindings(new ChangePasswordViewModel(), selector);
+    ChangePasswordViewModel = new ChangePasswordViewModel();
```
sloria: This variable needs to be declared. Also, because it's an instance of `ChangePasswordViewModel`, the variable should be lowercased. `viewModel` would be a fine name.

website/static/js/passwordForms.js
```diff
@@ -243,15 +289,42 @@ var SignUpViewModel = oop.extend(BaseViewModel, {
...
+            if (!ChangePasswordViewModel.password.isValid()) {
+                return false;
+            }
+        }
+    });
 };
 
 var SetPassword = function(selector) {
-    $osf.applyBindings(new SetPasswordViewModel(), selector);
+    SetPasswordViewModel = new SetPasswordViewModel();
```
sloria: See https://github.com/CenterForOpenScience/osf.io/pull/6021/files#r71915848

sloria:
Pass finished. Just a few minor fixups, and this is good to merge.

erinspace:
@sloria ready!

PR #6028, opened 2016-07-25
[Feature] Allow DOI Creation on Public Projects [OSF-6754]

mfraezz:
## Purpose
From ticket:
>Users should be able to create a DOI for Public Projects and Components. Much like for public registrations, DOIs should only be issued upon request (link click) from the Project/Component Overview page.
>
>If a public project has a DOI and then it is registered (or forked), the project DOI should not be copied over to the registration (or fork).

## Changes
* Allow admin contributors to create DOI/ARKs on public projects

## Side effects
If a public project with a DOI becomes private again, anyone who follows that DOI but cannot view the project will be confused.

## Ticket
[OSF-6754](https://openscience.atlassian.net/browse/OSF-6754)

### DIFF
--- a/framework/exceptions/__init__.py
+++ b/framework/exceptions/__init__.py
@@ -22,7 +22,7 @@ class HTTPError(FrameworkError):
         },
         http.FORBIDDEN: {
             'message_short': 'Forbidden',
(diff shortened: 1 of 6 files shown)


icereval:
LGTM

PR #6030, opened 2016-07-25
Linkify project descriptions [OSF-6626]

Johnetordoff:
## Purpose

Allow uses to include links in there project descriptions.

## Changes

Uses linkify.js to allow html links in otherwise sanitized content. Edit: To avoid built-in sanitization this will only apply to descriptions when users can't edit them. 

## Side effects

None that I know of

## Ticket

https://openscience.atlassian.net/browse/OSF-6626


brianjgeiger:
I'm not *entirely* sure how this PR would make those tests fail, so I'm restarting Travis. When it's done, if the tests pass, it can go back into code review; otherwise it'll be good to figure out what's going on there. Also, can you do this with just a front-end change? Sanitizers might take out HTML, but I wouldn't expect them to remove URLs. It's those that should be recognized and made clickable.

Johnetordoff:
Thanks for restarting Travis, but this does appear to be a real bug. It seems the code nugget:
`if 'write' in node.get_permissions(user)`  throws errors while seemly analogous code nugget
`if node.has_permission(user, WRITE)` is fine. Might be important or I'm misunderstanding what these do.

To address the other part of your comment, the description is sanitized differently depending on the what the user's permissions are. I'll try doing this in the Mako.

brianjgeiger:
Okay. Just in case, I've updated the description of the jira ticket. If you still need to make changes to python or mako (other than just adding a script to the page), please post specifics of the effects.

Johnetordoff:
@brianjgeiger I'm getting a Phantom JS error on this and am unsure as to why. This has happened a few time with various things for me, how do I see the Phantom JS tests and what are some good resources for avoiding this?

brianjgeiger:
I hadn't seen that one before. Restarting the test. We'll keep an eye on that in the mid-term to see if we need to find a way to fix it.

brianjgeiger:
Is this enabling full markdown for this element, or just linkifying the urls? If the former, was that functionality demo'd, because it's out of scope. If the latter, I think you've added an unnecessary level of indirection.

Also, be sure to wait for tests to pass before putting into code review.

atelic:
@brianjgeiger, It was enabling full markdown for this element. I have removed this functionality. Now it just linkifys the urls.

Tests were passing when I moved it but it looks like @Johnetordoff updated this PR 3 hours ago? Are you still working on this feature?

atelic:
I've created #6302 with the change to markdown rendering and a happy karma.

brianjgeiger:
Closed in favor of 6302

PR #6034, opened 2016-07-26
File widget textboxes have incorrect highlight [OSF-4481]

Johnetordoff:
## Purpose

Text inputs for renaming and folder adding didn't focus properly, this fixes that and makes them consistent.

## Changes

one line fix makes all input fields focus properly

## Side effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/OSF-4481


brianjgeiger:
🎉 🐧 🎉

PR #6035, opened 2016-07-26
hotfix/Password strength bugfixes [#OSF-6759]

erinspace:
## Purpose
Two bugfixes and some stumbled on improvements to the new password strength checks:
- Disable submission for low strength password on password change form in settings - needs to be handled separately because the form's submit is not data-bound in knockout
- Frontend on signup with existing account showed incorrect error message
- Backend would error on password reset to same as old password - added that to frontend validation so that feedback happens quicker

## Changes
- Add frontend validation for old password matching new password
- Move oldPassword to be used in password validation extender
- Remove catching of all 400s and changing the message - email as password checks are taking place separately and don't need a special override in submit any longer.
- Disable password reset in settings explicitly - is not data-bound to the form in knockout as the others are

## Side effects
None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-6759

### DIFF
--- a/website/templates/profile/account.mako
+++ b/website/templates/profile/account.mako
@@ -160,7 +160,7 @@
                                 >
                                 <p class="help-block" data-bind="validationMessage: passwordConfirmation" style="display: none;"></p>
                             </div>
-                            <button type="submit" class="btn btn-primary">Update password</button>
+                            <button type="submit" class="btn btn-primary" data-bind="disable: !password.isValid()">Update password</button>
                         </form>
                     </div>
                 </div>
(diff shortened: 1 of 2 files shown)


website/templates/profile/account.mako
```diff
@@ -160,7 +160,8 @@
                                 >
                                 <p class="help-block" data-bind="validationMessage: passwordConfirmation" style="display: none;"></p>
                             </div>
-                            <button type="submit" class="btn btn-primary">Update password</button>
+                            ## TODO: change so that password strength submit validation happens in knockout on the form, not with this disable
```
brianjgeiger: Could you make a jira ticket for this todo, then update the comment so that it's `## TODO: [OSF-XXXX] change…`?

brianjgeiger:
No complaints about the code. Quick request for an update to the todo comment. 🎉 🐧 🎉

erinspace:
Issue added with https://openscience.atlassian.net/browse/OSF-6764 !

PR #6037, opened 2016-07-26
Remove dependencies on personal forks

sloria:
## Purpose

For security and reliable builds, all dependencies should be either pinned to a version on a package index or pinned to a SHA on a COS-owned GitHub repo.

## Changes

Remove dependencies on non-COS GitHub repos and change them to COS owned repos/forks

## Ticket

https://openscience.atlassian.net/browse/OSF-6092


whit537:
👍

For even more secure and reliable builds, we could run our own package server. That'd protect against things like left-pad, e.g.

PR #6038, opened 2016-07-26
Add template_from to node API links [OSF-4464]

Johnetordoff:
## Purpose

Allows a direct link between a node and it's template through and API link.

## Changes

Add properties to node model and link to API serializer.

## Side effects

None that I know of

## Ticket

https://openscience.atlassian.net/browse/OSF-4464


brianjgeiger:
Tests would be awesome.

brianjgeiger:
Pass complete. 🐧

Johnetordoff:
@brianjgeiger Added tests, there's still small area of contention. If a registration is withdrawn the api will hide certain fields, should this be one of them? 

Currently registrations and draft_registrations are hidden, not sure what the rational behind this is so am just checking before I alter the test to that's failing to allow for the new field.

brianjgeiger:
You might as well hide this field if it's withdrawn.

Johnetordoff:
I was mistaken, all Relationship fields are hidden during withdrawal/reaction the test just didn't reflect this, now it does. Should be ready now.

Johnetordoff:
@brianjgeiger this is failing a Phantom JS test, that shouldn't happen for something like this, right?

brianjgeiger:
I've restarted the branch. If it fails again, you should look into it. Does it fail locally?

Johnetordoff:
Yes, it also fails locally.

api/nodes/serializers.py
```diff
@@ -164,6 +164,11 @@ class NodeSerializer(JSONAPISerializer):
         related_view_kwargs={'node_id': '<forked_from_id>'}
     )
 
+    templated_from = RelationshipField(
+        related_view=lambda n: 'registrations:registration-detail' if getattr(n, 'template_node', None) else 'nodes:node-detail',
```
brianjgeiger: I'm reasonably certain the getattr there should be grabbing 'is_registration' to determine if it's a registration or a node.

api_tests/base/test_serializers.py
```diff
@@ -76,7 +76,7 @@ def test_registration_serializer(self):
         # fields that are visible for withdrawals
         visible_on_withdrawals = ['contributors', 'date_created', 'description', 'id', 'links', 'registration', 'title', 'type', 'current_user_can_comment']
         # fields that do not appear on registrations
-        non_registration_fields = ['registrations', 'draft_registrations']
+        non_registration_fields = ['registrations', 'draft_registrations', 'templated_from']
```
brianjgeiger: Although if you're not showing the field at all on registrations, then you don't need the lambda on https://github.com/CenterForOpenScience/osf.io/pull/6038/files#r72799108 at all

api/nodes/serializers.py
```diff
@@ -164,6 +164,11 @@ class NodeSerializer(JSONAPISerializer):
...
+        related_view_kwargs={'node_id': '<templated_from_id>'}
```
brianjgeiger: The gymnastics with the model are probably unnecessary. You can probably just do `'node_id': <'template_node._id>`

api_tests/base/test_serializers.py
brianjgeiger: I think if the model keeps the "template_node" on the registration, then it makes sense to display this on the registration, but it shouldn't be visible on withdrawals. So I think drop all the changes on this file and ensure it still works. You may have to change another test.

website/project/model.py
```diff
@@ -2728,6 +2728,18 @@ def forked_from_id(self):
         return None
 
     @property
+    def templated_from(self):
```
brianjgeiger: As mentioned above, I don't _think_ you need to change the model at all to make this work. You should be able to drop the changes to this file.

brianjgeiger:
Pass done. 🐧 I think this can be simplified, and I expect a registration test will fail after making some changes that I suggest and will need to be modified.

brianjgeiger:
Looks good. 🎉 🐧 🎉

PR #6039, opened 2016-07-26
Remove Download/Delete/Check Out buttons from old version view [#OSF-4544]

caspinelli:
## Purpose

Removed Download/Delete/Check Out buttons from the viewing of an old version of a file.
Those buttons actually referred to the current version of the file, not the revision, so it could be misleading to the user

## Changes
Made sure to check if in an old version view before rendering those specific buttons.

## Side effects
None

## Ticket

https://openscience.atlassian.net/browse/OSF-4544


brianjgeiger:
Everything looks good here. 🎉 🐧 🎉

PR #6041, opened 2016-07-26
[hotfix] [OSF-6745] Contributor Bulk Updates

caseyrollins:
#### Purpose
- Contributor bulk updates in API v2 would not work with the new contributor identifier format (```node-id```-```user-id```). This PR updates the ```NodeContributorsList``` view in the API so that bulk updates with the new identifiers will succeed. Additionally, this PR updates the tests for bulk updates to use the new contributor identifier format in their request bodies. 

#### Changes
- Update tests to make bulk requests with new contributor identifiers.
- Override ```get_requested_resources()``` in ```BulkDestroyJSONAPIView```
- Update ```get_queryset()``` in ```NodeContributorsList``` to still look for User ID's. 

#### Side effects
- None expected, as it doesn't appear that we are currently making bulk contributor updates via API v2 on the OSF.

#### Ticket
- [OSF-6745](https://openscience.atlassian.net/browse/OSF-6745)

### DIFF
--- a/api_tests/nodes/views/test_node_contributors_list.py
+++ b/api_tests/nodes/views/test_node_contributors_list.py
@@ -1006,22 +1006,40 @@ class TestNodeContributorBulkUpdate(NodeCRUDTestCase):
         self.private_url = '/{}nodes/{}/contributors/'.format(API_BASE, self.private_project._id)
         self.public_url = '/{}nodes/{}/contributors/'.format(API_BASE, self.public_project._id)
 
-        self.payload_one = {
-                'id': self.user_two._id,
-                'type': 'contributors',
-                'attributes': {
-                    'bibliographic': True,
-                    'permission': "admin"
-                }
+        self.payload_one_public = {
+            'id': make_contrib_id(self.public_project._id, self.user_two._id),
+            'type': 'contributors',
+            'attributes': {
+                'bibliographic': True,
+                'permission': "admin"
+            }
         }
 
-        self.payload_two = {
-                'id': self.user_three._id,
-                'type': 'contributors',
-                'attributes': {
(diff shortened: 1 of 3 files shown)


api_tests/nodes/views/test_node_contributors_list.py
```diff
@@ -1006,40 +1006,58 @@ def setUp(self):
...
     def test_bulk_update_contributors_dict_instead_of_list(self):
-        res = self.app.put_json_api(self.public_url, {'data': self.payload_one},
+        res = self.app.put_json_api(self.public_url, {'data': self.payload_one_public},
                                     auth=self.user.auth, expect_errors=True, bulk=True)
         assert_equal(res.status_code, 400)
 
     def test_bulk_update_contributors_public_project_one_not_found(self):
         invalid_id = {
-            'id': '12345',
+            'id': '12345-abcde',
```
brianjgeiger: Could you make an additional test wherever this diff occurs that tries it with the non-hyphenated id and ensure that it just fails with a 400-series error and not a 500-series error?

brianjgeiger:
Have you made sure that contributor management from the front end all still works? I don't know if any of that uses v2 bulk requests.

brianjgeiger:
Pass finished. Add some tests. Verify front-end not broken.

caseyrollins:
@brianjgeiger, verified that the front-end is not broken, and added tests ([here](https://github.com/CenterForOpenScience/osf.io/pull/6041/files#diff-a594bd7246f94b3bdb9a3ca5d4f52b80R1157) and here) to check the contributor identifier format.

brianjgeiger:
Awesome, thanks.

PR #6042, opened 2016-07-26
[Feature] Create Unregistered Users via APIv2 [OSF-6753]

mfraezz:
## Purpose
Allow (with limited availability) the creation of users through APIv2. This is required for several services currently in development.

## Changes
* Create custom scope for `USERS_CREATE`, included in the admin scope
* Add tracking mechanism to User model to see who added each user (spam mitigation)
* Create UserCreateSerializer, conditionally used on `POST` requests to `/v2/users/`
* Create permissions class to allow *only* tokened requests for unsafe methods.

### Notes
Uniqueness of users ensured by ignoring username case. This was previously a bug in one possible user-creation flow

## Side effects
`osf.users.create` scope will need to be exposed in `public_scopes` (with `public=False`) before this will work in production.

## Ticket
[OSF-6753](https://openscience.atlassian.net/browse/OSF-6753)

### DIFF
--- a/api/base/permissions.py
+++ b/api/base/permissions.py
@@ -34,6 +34,12 @@ class TokenHasScope(permissions.BasePermission):
             # Assumption: user authenticated via non-oauth means, so don't check token permissions.
             return True
 
+        return self._verify_scopes(request, view, token)
+
+    def _verify_scopes(self, request, view, token):
+        # Anything calling this method should handle the case where
+        # `token is None` before making this call.
+
         required_scopes = self._get_scopes(request, view)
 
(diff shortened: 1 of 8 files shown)


brianjgeiger:
Prooobably should fix failing tests.

mfraezz:
>Prooobably should fix failing tests.

The only failing one was `TestUserMerging.test_merge` because of a new field (the other build I cancelled due to an unused import and the desire to not waste TCI time). *Should* be fixed now

brianjgeiger:
If the tests pass, this should be good to merge. 🎉 🐧 🎉

PR #6043, opened 2016-07-26
Manage VOLs via API [OSF-6738] 

caseyrollins:
#### Purpose
- Allows view only links to be managed through the API. VOLs can be created, updated, and deleted via the ```/nodes/<node-id>/view_only_links/``` and ```/nodes/<node-id>/view_only_links/<link-id>/``` endpoints. Additionally, VOLs can be viewed via the base endpoint ```view_only_links/<link-id>/```.
- The nodes associated with a VOL are managed via the ```/view_only_links/<link-id>/relationships/nodes/``` endpoint in a similar fashion to how node/institution relationships are managed.
- VOLs can be filtered by their ```name```, ```anonymous```, and ```date_created``` attributes.
- The nodes associated with a VOL must all be part of the same project, however it is *not* necessary for a component's parent to be associated with a VOL before it can be associated.

#### Changes
- Added routes:
 - ```/nodes/<node-id>/view_only_links/```
 - ```/nodes/<node-id>/view_only_links/<link-id>/```
 - ```/registrations/<node-id>/view_only_links/```
 - ```/registrations/<node-id>/view_only_links/<link-id>/```
 - ```/view_only_links/<link-id>/```
 - ```/view_only_links/<link-id>/nodes/```
 - ```/view_only_links/<link-id>/relationships/nodes/```

- Added views:
 - ```api.nodes.views.NodeViewOnlyLinksList```
 - ```api.nodes.views.NodeViewOnlyLinkDetail```
 - ```api.registrations.views.RegistrationViewOnlyLinksList```
 - ```api.registrations.views.RegistrationViewOnlyLinkDetail```
 - ```api.view_only_links.views.ViewOnlyLinkDetail```
 - ```api.view_only_links.views.ViewOnlyLinkNodes```
 - ```api.view_only_links.views.ViewOnlyLinkNodesRelationships```

- Added serializers:
 - ```api.nodes.serializers.NodeViewOnlyLinkSerializer```
 - ```api.nodes.serializers.NodeViewOnlyLinkUpdateSerializer```
 - ```api.view_only_links.serializers.ViewOnlyLinkDetailSerializer```
 - ```api.view_only_links.serializers.ViewOnlyLinkNodesSerializer```

#### Side effects
- None expected.

#### Ticket
- [OSF-6738](https://openscience.atlassian.net/browse/OSF-6738)

### DIFF
--- a/api/nodes/serializers.py
+++ b/api/nodes/serializers.py
@@ -15,6 +15,7 @@ from website.project.metadata.utils import is_prereg_admin_not_project_admin
 from website.models import Node, User, Comment, Institution, MetaSchema, DraftRegistration
 from website.exceptions import NodeStateError
 from website.util import permissions as osf_permissions
+from website.project import new_private_link
 from website.project.model import NodeUpdateError
 
 from api.base.utils import get_user_auth, get_object_or_error, absolute_reverse, is_truthy
@@ -935,3 +936,45 @@ class DraftRegistrationDetailSerializer(DraftRegistrationSerializer):
             draft.update_metadata(metadata)
             draft.save()
         return draft
+
+
+class NodeViewOnlyLinkSerializer(JSONAPISerializer):
+    '''
+    Document pls.
+    '''
+
+    key = ser.CharField()
+    id = IDField(source='_id')
+    date_created = ser.DateTimeField()
+    nodes = JSONAPIListField(child=NodeSerializer(), required=True)
+    anonymous = ser.BooleanField(default=False)
+    name = ser.CharField(default='Shared project link')
+
+    creator = RelationshipField(
+        related_view='users:user-detail',
+        related_view_kwargs={'user_id': '<creator._id>'},
+    )
+
+    def create(self, validated_data):
+        # name = validated_data.pop('name')
+        # user = get_user_auth(self.context['request'])
+        # nodes = validated_data.pop('nodes')
+        # anonymous = validated_data.pop('anonymous')
+        #
+        # try:
+        #     private_link = new_private_link(
+        #         name=name,
+        #         user=user,
+        #         nodes=nodes,
+        #         anonymous=anonymous
+        #     )
+        # except ValidationValueError:
+        #     raise exceptions.ValidationError('Invalid link name.')
+        # return private_link
+        pass
+
+    def update(self, instance, validated_data):
+        pass
+
+    class Meta:
+        type_ = 'view_only_links'
--- a/api/nodes/serializers.py
+++ b/api/nodes/serializers.py
@@ -15,7 +15,6 @@ from website.project.metadata.utils import is_prereg_admin_not_project_admin
 from website.models import Node, User, Comment, Institution, MetaSchema, DraftRegistration, PrivateLink
 from website.exceptions import NodeStateError
 from website.util import permissions as osf_permissions
-from website.util.sanitize import strip_html
 from website.project import new_private_link
 from website.project.model import NodeUpdateError
 
--- a/api/nodes/serializers.py
+++ b/api/nodes/serializers.py
@@ -221,6 +221,12 @@ class NodeSerializer(JSONAPISerializer):
         self_meta={'count': 'get_node_links_count'}
     )
 
+    view_only_links = RelationshipField(
+        related_view='nodes:node-view-only-links',
+        related_view_kwargs={'node_id': '<pk>'},
+        related_meta={'count': 'get_view_only_links_count'},
+    )
+
     def get_current_user_permissions(self, obj):
         user = self.context['request'].user
         if user.is_anonymous():
@@ -277,6 +283,10 @@ class NodeSerializer(JSONAPISerializer):
         return {
             'node': node_comments
         }
+
+    def get_view_only_links_count(self, obj):
+        links = [link for link in obj.private_links if not link.is_deleted]
+        return len(links)
 
     def create(self, validated_data):
         request = self.context['request']
@@ -956,21 +966,27 @@ class NodeViewOnlyLinkSerializer(JSONAPISerializer):
     anonymous = ser.BooleanField(required=False)
     name = ser.CharField(required=False)
 
+    links = LinksField({
+        'self': 'get_absolute_url'
+    })
+
     creator = RelationshipField(
         related_view='users:user-detail',
         related_view_kwargs={'user_id': '<creator._id>'},
     )
 
     nodes = RelationshipField(
-        related_view='view_only_links:view-only-link-nodes',
-        related_view_kwargs={'link_id': '<_id>'}
+        related_view='view-only-links:view-only-link-nodes',
+        related_view_kwargs={'link_id': '<_id>'},
     )
 
     def get_absolute_url(self, obj):
+        node_id = self.context['request'].parser_context['kwargs']['node_id']
         return absolute_reverse(
             'nodes:node-view-only-link-detail',
             kwargs={
-                'key': obj.key
+                'link_id': obj._id,
+                'node_id': node_id
             }
         )
 
--- a/api/nodes/serializers.py
+++ b/api/nodes/serializers.py
@@ -950,9 +950,6 @@ class NodeVOL(ser.Field):
 
 
 class NodeViewOnlyLinkSerializer(JSONAPISerializer):
-    """
-    Document pls.
-    """
     key = ser.CharField(read_only=True)
     id = IDField(source='_id', read_only=True)
     date_created = ser.DateTimeField(read_only=True)
--- a/api/nodes/serializers.py
+++ b/api/nodes/serializers.py
@@ -955,6 +955,12 @@ class NodeVOL(ser.Field):
 
 
 class NodeViewOnlyLinkSerializer(JSONAPISerializer):
+    filterable_fields = frozenset([
+        'anonymous',
+        'name',
+        'date_created'
+    ])
+
     key = ser.CharField(read_only=True)
     id = IDField(source='_id', read_only=True)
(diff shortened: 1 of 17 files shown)


brianjgeiger:
You should add a relationship into the node serializer for the view only links. Also, are VOLs for registrations a thing? If so, you'll need to make sure the registration endpoints all work properly for registration VOLs (including adding /v2/registrations/* urls for the appropriate views). If not, be sure not to include the relationship in the registration serializer.

api/nodes/serializers.py
```diff
@@ -935,3 +937,120 @@ def update(self, draft, validated_data):
...
+        if name:
+            name = strip_html(name)
+            if name is None or not name.strip():
+                raise exceptions.ValidationError('Invalid link name.')
+            view_only_link.name = name
+
+        if anonymous:
+            view_only_link.anonymous = anonymous
+
+        if nodes:
```
brianjgeiger: I suspect this whole serializer is not necessary. Ideally the relationship self link (that you're learning about in your other issue) will handle managing the nodes that are related to this VOL, and ideally ideally the relationshipfield would be able to handle being given a list of items. That being said, I'm not entirely sure that all of the plumbing is in for this. Hmm.

While QA is working on the tests for this, we might want to take a look at what the Preprints team did to update the relationship link management and see if it's partially, fully, or not at all useful for this. Then we can decide if we want to make a TODO to make this work better or if we want to just go ahead and make it right now.

brianjgeiger:
Added a couple of big items, so we'll pop this over to Addl Development and then get the QA test cases to be sure everything's right. Pass done. 🐧

api/nodes/serializers.py
```diff
@@ -935,3 +942,94 @@ def update(self, draft, validated_data):
...
+
+    def update(self, link, validated_data):
+        name = validated_data.get('name')
+        anonymous = validated_data.get('anonymous')
+
+        assert isinstance(link, PrivateLink), 'link must be a PrivateLink'
+
+        if name:
+            name = strip_html(name)
+            if name is None or not name.strip():
```
brianjgeiger: Should be the only reason name is none is if it's not in validated_data, which means it wouldn't make it to here, right? Or does strip_html sometimes return none? Also, I think we automatically clean incoming data, so the strip_html might not be necessary.

api/nodes/views.py
```diff
@@ -2935,3 +2937,156 @@ def get_parser_context(self, http_request):
         res = super(LinkedNodesList, self).get_parser_context(http_request)
         res['is_relationship'] = True
         return res
+
+
+class NodeViewOnlyLinksList(JSONAPIBaseView, generics.ListCreateAPIView, NodeMixin):
```
brianjgeiger: We should go ahead and make this filterable on name, anonymous, and date_created.

api/view_only_links/serializers.py
```diff
@@ -0,0 +1,155 @@
...
+        nodes_to_add = []
+        for node_id in diff['add']:
+            node = Node.load(node_id)
+            if not node:
+                raise NotFound
+            nodes_to_add.append(node)
+
+        return nodes_to_add, diff['remove'].values()
+
+    def get_eligible_nodes(self, nodes, add):
```
brianjgeiger: Hmmm.

api/view_only_links/serializers.py
```diff
@@ -0,0 +1,155 @@
...
+            raise RelationshipPostMakesNoChanges
+
+        eligible_nodes = self.get_eligible_nodes(nodes, add)
+
+        for node in add:
+            if not node.has_permission(user, 'admin'):
+                raise PermissionDenied
+            if node not in eligible_nodes:
+                raise ValidationError(detail='The node {0} cannot be affiliated with this VOL because it is not a child of the associated VOL nodes.'.format(node._id))
+            view_only_link.nodes.append(node)
```
brianjgeiger: I *think* you might need to update your eligible nodes here? If you're adding a bunch of nodes, as you add nodes in the hierarchy, that should give you more node options, yes?

brianjgeiger: You know what, let's have a chat with Product. I don't think it's necessary to have this really complicated thing that you worked very hard on any more.

brianjgeiger: I talked with product. Go ahead and remove this. Make a VOL with something impossible from the front-end's perspective and make sure it isn't disastrous. I'm okay with the front end displaying it right but not being able to reproduce it, and probably even not letting the front end edit it but only delete it. But it should accurately represent the state of the ticket and should let you delete it.

If it is disastrous, the options are to fix it or to make this set of endpoints dev_only. I'll leave it to your better judgement about which way you want to go, but make a note for QA and we can discuss if any additional tickets will need to be made for the future.

api/registrations/views.py
```diff
@@ -424,3 +424,15 @@ def get_object(self):
         ], 'self': node}
         self.check_object_permissions(self.request, obj)
         return obj
+
+
+class RegistrationViewOnlyLinksList(NodeViewOnlyLinksList, RegistrationMixin):
+
```
brianjgeiger: I think you'll need some registrations-based scopes here (and you'll probably have to define those as well).

brianjgeiger:
Pass done. Simplify the node adding, add some registration-based scopes, add filtering, and check that one validation. 🐧

api/nodes/serializers.py
caseyrollins: Yep, it appears the incoming data is cleaned, so this can be removed.

api/view_only_links/serializers.py
```diff
@@ -108,13 +107,13 @@ def create(self, validated_data):
...
 
-        eligible_nodes = self.get_eligible_nodes(nodes, add)
+        eligible_nodes = self.get_eligible_nodes(nodes)
 
         for node in add:
             if not node.has_permission(user, 'admin'):
                 raise PermissionDenied
             if node not in eligible_nodes:
-                raise ValidationError(detail='The node {0} cannot be affiliated with this VOL because it is not a child of the associated VOL nodes.'.format(node._id))
+                raise ValidationError(detail='The node {0} cannot be affiliated with this VOL because it is not a child of the associated VOL node.'.format(node._id))
```
brianjgeiger: Yay validation error! Let's go ahead and make it a tiny bit more explicit. `The node {0} cannot be affiliated with this View Only Link because the node you're trying to affiliate is not descended from the node that the View Only Link is attached to: {1}.` Unless that doesn't make any sense, in which case we can work harder at this.

api/view_only_links/serializers.py
```diff
@@ -0,0 +1,154 @@
...
+                raise NotFound
+            nodes_to_add.append(node)
+
+        return nodes_to_add, diff['remove'].values()
+
+    def get_eligible_nodes(self, nodes):
+        return [
+            descendant
+            for node in nodes
+            for descendant in node.get_descendants_recursive()
```
brianjgeiger: This seems like it doesn't need to be this complicated. Shouldn't it just be `return node.get_descendants_recursive()`?

api/view_only_links/serializers.py
```diff
@@ -0,0 +1,154 @@
...
+        if not len(add):
+            raise RelationshipPostMakesNoChanges
+
+        eligible_nodes = self.get_eligible_nodes(nodes)
+
+        for node in add:
+            if not node.has_permission(user, 'admin'):
+                raise PermissionDenied
+            if node not in eligible_nodes:
+                raise ValidationError(detail='The node {0} cannot be affiliated with this VOL because it is not a child of the associated VOL node.'.format(node._id))
```
brianjgeiger: Yay validation error! Let's go ahead and make it a tiny bit more explicit. `The node {0} cannot be affiliated with this View Only Link because the node you're trying to affiliate is not descended from the node that the View Only Link is attached to: {1}.` Unless that doesn't make any sense or is wrong, in which case we can work harder at this. (Moved from commit comment)

api/view_only_links/serializers.py
```diff
@@ -0,0 +1,154 @@
...
+        view_only_link.save()
+
+        nodes = [Node.load(node) for node in view_only_link.nodes]
+        eligible_nodes = self.get_eligible_nodes(nodes)
+
+        for node in add:
+            if not node.has_permission(user, 'admin'):
+                raise PermissionDenied
+            if node not in eligible_nodes:
+                raise ValidationError(detail='The node {0} cannot be affiliated with this VOL because it is not a child of the associated VOL node.'.format(node._id))
```
brianjgeiger: Since this ValidationError is used in two places, maybe make a specific validation error class and use that instead for DRYness.

brianjgeiger:
A couple of small changes and fix your flake8 error. Pass done. 🐧

api/view_only_links/serializers.py
caseyrollins: ```node.get_descendants_recursive()``` returns a generator which still needs to be iterated over to get each node :/

caseyrollins: The function needs to somehow iterate over all the nodes currently associated with the VOL. Since view only links don't have a concept of a primary or parent node, checking the descendants of a single node in the list of nodes doesn't guarantee that we're actually checking all of the eligible descendants. 

Additionally, ```node.get_descendants_recursive()``` returns a generator that needs to be iterated over to add all of the descendants to the eligible nodes list.

It's a bit complicated, but the nested loops are necessary :/

brianjgeiger: Okay. At some point I need to look into this so I understand it properly, but for now I'm good. 👍

brianjgeiger:
Looks great. 🎉 🐧 🎉 RTM when tests pass.

brianjgeiger:
Hey, tests passed!

PR #6046, opened 2016-07-27
[OSF-2052] LinkedIn share button should warn about adblockers. Share button file url should be the OSF file page url.

acshi:
## Purpose
(1) If a user has ad block turned on and tries to share a project via linked in, the user is not able to share the project and get no information as to why. I think we should improve this with a tooltip that informs the user that having ad blocked enabled may prevent successful sharing of their project.
(2) The link on the file detail page used in the share/embed popover is to the rendered file, which is fine, however with the social media sharing links the link should be to the file.

## Changes
Added tooltip, changed share button file link.

screen shot 2016-07-28 at 9 57 40 am

screen shot 2016-07-27 at 2 04 00 pm

## Side effects
NO

## Ticket
https://openscience.atlassian.net/browse/OSF-2052

### DIFF
--- a/website/static/js/components/socialshare.js
+++ b/website/static/js/components/socialshare.js
@@ -22,12 +22,19 @@ var ShareButtons = {
         var facebookHref = 'https://www.facebook.com/sharer/sharer.php?u=' + url;
         var linkedinHref = 'https://www.linkedin.com/cws/share?url=' + url + '&title=' + title;
         var emailHref = 'mailto:?subject=' + title + '&body=' + url;
-        return m('div.share-buttons', {}, [
+        return m('div.share-buttons ', {
+                config: function(el, isInitialized) {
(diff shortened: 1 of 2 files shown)


brianjgeiger:
If item 1 was approved by product, great, otherwise it should be something more obvious if possible than a tooltip (for example, overlaying a 🚫 over top of the LinkedIn icon). We can change this in the future if need be. 

Overall, code seems fine. Has the functionality been demo'd and approved? 🎉 🐧 🎉

acshi:
Those were both requests from the @saradbowman and @nicipfeiffer. Detecting whether the user has an inappropriately acting adblocker seems like it would be a lot of additional work, especially since it is likely that newer adblocking programs would have fewer false positives; but yes, that could be done in the future.

The previous version before these small changes went through Q/A, so otherwise I believe they are happy with it.

PR #6049, opened 2016-07-27
New Folder Error Message Persists [#OSF-6778]

caspinelli:
## Purpose

Fix bug where new folder error persists if user clicks on OSF Storage immediately. Also fixed weird behavior where clicking inside the tb-tbody element but not on an actual file would deselect the currently selected file without dismissing the toolbar.

## Changes

made sure that whenever a new folder is selected the error text clears. 

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-6778

### DIFF
--- a/website/static/js/fangorn.js
+++ b/website/static/js/fangorn.js
@@ -91,6 +91,7 @@ function findByTempID(parent, tmpID) {
     return item;
 }
 
+
 /**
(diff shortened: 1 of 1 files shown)


brianjgeiger:
Looks good. 🎉 🐧 🎉

PR #6052, opened 2016-07-28
[OSF-5639] Warn users they cannot search by email address

mluke93:
## Purpose

Inform users that they cannot add an email address as an unregistered user. Gives the user an error message if the regular expression matches the syntax of an email address.

## Changes

Checks the regular expression of the query. If the input has the format of [text]@[text].[text], it registers it as an email address, informs the user with an error, and does NOT allow for that user to be added as an unregistered contributor.

## Side effects

n/a

## Ticket


brianjgeiger:
Looks great. 🎉 🐧 🎉

PR #6054, opened 2016-07-28
License cannot be applied to the RPP project[#OSF-5389]

chennan47:
## Purpose
Add scripts to add license to 'RPP' projects.

## Ticket

https://openscience.atlassian.net/browse/OSF-5389

### DIFF
--- a/scripts/add_rpp_license.py
+++ b/scripts/add_rpp_license.py
@@ -27,7 +27,7 @@ def main(dry_run=False):
                 license_id=node_license.id,
                 year='2016',
                 copyright_holders='',
-                auth=Auth(user)
+                auth=Auth(user=user)
         )
         node.save()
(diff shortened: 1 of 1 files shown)


brianjgeiger:
So no ideas on why it's actually failing? I'm not entirely opposed to covering up the problem, but I'd hate to have to revisit this. What happens if you fork RPP and then try to apply a license to that? Also, why would this script work if you can't change it via an api v1 view?

brianjgeiger:
Take an hour to see if you can find out what's actually happening, or if you can make a project you can reproduce this with. If not, we'll probably go ahead and use the script, but I'd feel better if we don't just mask it.

brianjgeiger:
Closing this for now. Will re-open if the other fix doesn't solve the problem.

PR #6056, opened 2016-07-28
License cannot be applied to the RPP project[#OSF-5389]

chennan47:
## Purpose

Excluded deleted and private node from update search when try to save nodes for node_license update. Can be a solution to the problem of https://openscience.atlassian.net/browse/OSF-5389. Even if it does not solve this ticket, deleted nodes and private nodes shouldn't be included in updates for node_license search update. And we can still use https://github.com/CenterForOpenScience/osf.io/pull/6054 to add license to rpp as a ultimate solution.

## Ticket

… nodes for node_license update[#OSF-5389]
https://openscience.atlassian.net/browse/OSF-5389.

### DIFF
--- a/website/project/model.py
+++ b/website/project/model.py
@@ -1692,7 +1692,11 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin, Commentable):
             ) if c.is_public and not c.is_deleted]
             # this returns generator, that would get unspooled anyways
             if children:
-                Node.bulk_update_search(children)
+                if len(children) > 100:
+                    Node.bulk_update_search(children[:99])
+                    Node.bulk_update_search(children[99:])
+                else:
+                    Node.bulk_update_search(children)
 
         # This method checks what has changed.
         if settings.PIWIK_HOST and update_piwik:
--- a/website/project/model.py
+++ b/website/project/model.py
@@ -1689,7 +1689,7 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin, Commentable):
         if 'node_license' in saved_fields:
             children = [c for c in self.get_descendants_recursive(
                 include=lambda n: n.node_license is None
-            )]
+            ) if c.is_public and not c.is_deleted]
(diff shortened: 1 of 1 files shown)


brianjgeiger:
This looks good once tests pass. 🎉 🐧 🎉

website/project/model.py
```diff
@@ -1689,10 +1689,14 @@ def save(self, *args, **kwargs):
...
             children = [c for c in self.get_descendants_recursive(
                 include=lambda n: n.node_license is None
-            )]
+            ) if c.is_public and not c.is_deleted]
             # this returns generator, that would get unspooled anyways
             if children:
-                Node.bulk_update_search(children)
+                if len(children) > 100:
+                    Node.bulk_update_search(children[:99])
+                    Node.bulk_update_search(children[99:])
```
brianjgeiger: We won't run into the problem for a while, but since we're solving this anyways, let's make this handle an arbitrary number of nodes. Perhaps a loop?

brianjgeiger:
Almost there. Pass done. 🐧

website/project/model.py
```diff
@@ -1689,10 +1689,20 @@ def save(self, *args, **kwargs):
...
+                if count == 0:
+                    Node.bulk_update_search(children)
+                else:
+                    start = 0
+                    end = 100
+                    for i in range(1, count + 1):
+                        Node.bulk_update_search(children[start:end])
+                        start = end
+                        end = 100 * (i + 1)
+                    Node.bulk_update_search(children[(100 * count):])
```
brianjgeiger: This looks like it might hit a problem on an edge case, so we'd either want to break it out into a function and unit test that function, or we can possibly simplify with something like:

```py
while len(children):
    batch = children[:99]
    children = children[99:]
    Node.bulk_update_search(batch)
```

brianjgeiger:
Pass done. Mostly me being picky because it's in the Node's save method. 🐧

brianjgeiger:
Looks good. 🎉 🐧 🎉  This is approved once tests pass.

PR #6058, opened 2016-07-28
Fix registration comment related links [OSF-6551]

TomBaxter:
## Purpose:
[OSF-6551](https://openscience.atlassian.net/browse/OSF-6551)
Fix "node" and "replies" links for registration comments. Covers comments for nodes, wikis, and files.

## Changes:
Update api/comments/serializers.py make registration and node super classes of *CommentSerializer
Update api/comments/views.py add logic to differentiate between Node and Registration
Update api/nodes/views.py switch to NodeCommentSerializer
Update api/registrations/views.py switch to RegistrationCommentSerializer
Update api_tests/comments/views/test_comment_detail.py fix several tests and add three tests
Update api_tests/utils.py add "def urlparse_drop_netloc" for converting links to URIs for app.get

## Side effects
None

https://openscience.atlassian.net/browse/OSF-6551

### DIFF
--- a/api/comments/serializers.py
+++ b/api/comments/serializers.py
@@ -43,8 +43,8 @@ class CommentSerializer(JSONAPISerializer):
 
     target = TargetField(link_type='related', meta={'type': 'get_target_type'})
     user = RelationshipField(related_view='users:user-detail', related_view_kwargs={'user_id': '<user._id>'})
-    node = RelationshipField(related_view='nodes:node-detail', related_view_kwargs={'node_id': '<node._id>'})
-    replies = RelationshipField(self_view='nodes:node-comments', self_view_kwargs={'node_id': '<node._id>'}, filter={'target': '<pk>'})
+    node = RelationshipField(related_view='registrations:registration-detail', related_view_kwargs={'node_id': '<node._id>'})
+    replies = RelationshipField(related_view='registrations:registration-comments', related_view_kwargs={'node_id': '<node._id>'}, filter={'target': '<pk>'})
     reports = RelationshipField(related_view='comments:comment-reports', related_view_kwargs={'comment_id': '<pk>'})
 
     date_created = ser.DateTimeField(read_only=True)
--- a/api/comments/serializers.py
+++ b/api/comments/serializers.py
@@ -43,8 +43,6 @@ class CommentSerializer(JSONAPISerializer):
 
     target = TargetField(link_type='related', meta={'type': 'get_target_type'})
     user = RelationshipField(related_view='users:user-detail', related_view_kwargs={'user_id': '<user._id>'})
-    node = RelationshipField(related_view='registrations:registration-detail', related_view_kwargs={'node_id': '<node._id>'})
-    replies = RelationshipField(related_view='registrations:registration-comments', related_view_kwargs={'node_id': '<node._id>'}, filter={'target': '<pk>'})
     reports = RelationshipField(related_view='comments:comment-reports', related_view_kwargs={'comment_id': '<pk>'})
 
     date_created = ser.DateTimeField(read_only=True)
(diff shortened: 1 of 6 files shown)


brianjgeiger:
Be sure to pull in the tests I sent you for this.

api_tests/comments/views/test_comment_detail.py
```diff
@@ -139,15 +139,15 @@ def test_comment_has_node_link(self):
...
         assert_equal(res.status_code, 200)
         assert_equal(urlparse(url).path, expected_url)
 
     def test_comment_has_replies_link(self):
         self._set_up_public_project_with_comment()
         res = self.app.get(self.public_url)
-        url = res.json['data']['relationships']['replies']['links']['self']['href']
-        expected_url = '/{}nodes/{}/comments/?filter[target]={}'.format(API_BASE, self.public_project._id, self.public_comment._id)
+        url = res.json['data']['relationships']['replies']['links']['related']['href']
+        expected_url = '/{}registrations/{}/comments/?filter[target]={}'.format(API_BASE, self.public_project._id, self.public_comment._id)
```
brianjgeiger: I don't think this is right. Is public_project a node or a registration? If it's a node, the expect url should go to nodes. If it's a registration, then just be sure that there is an equivalent test for nodes as well.

api/comments/serializers.py
```diff
@@ -43,8 +43,8 @@ class CommentSerializer(JSONAPISerializer):
 
     target = TargetField(link_type='related', meta={'type': 'get_target_type'})
     user = RelationshipField(related_view='users:user-detail', related_view_kwargs={'user_id': '<user._id>'})
-    node = RelationshipField(related_view='nodes:node-detail', related_view_kwargs={'node_id': '<node._id>'})
-    replies = RelationshipField(self_view='nodes:node-comments', self_view_kwargs={'node_id': '<node._id>'}, filter={'target': '<pk>'})
+    node = RelationshipField(related_view='registrations:registration-detail', related_view_kwargs={'node_id': '<node._id>'})
```
brianjgeiger: And looking here, I'm pretty sure that you're just not handling the case where it's a node. This is a tricky one, so don't be surprised if it takes another day or two to fix.

brianjgeiger:
Not quite there yet. I don't think those old tests should have been changed and you have to get both the old tests and the new tests passing. I ran into these problems when I tried to fix it as well. Pass done. 🐧

api/comments/serializers.py
TomBaxter: Fixed

api_tests/comments/views/test_comment_detail.py
TomBaxter: fixed

api/comments/views.py
```diff
@@ -159,13 +163,24 @@ class CommentDetail(JSONAPIBaseView, generics.RetrieveUpdateDestroyAPIView, Comm
...
+        comment = self.get_comment()
+
+        if isinstance(comment.target.referent, Node):
+            comment_node = comment.target.referent
+        elif isinstance(comment.target.referent, (NodeWikiPage,
+                                                  StoredFileNode)):
+            comment_node = Guid.load(comment.target.referent.node).referent
+
+        if comment_node.is_registration:
+            self.serializer_class = RegistrationCommentDetailSerializer
```
brianjgeiger: Further down you used get_serializer_class to pick the serializer. Could you do that here, or does it not have access to the right information?

brianjgeiger:
A quick question about implementation, but if the answer isn't an easy, "Oh, yeah, that would be easy to do", we'll go forward with it as is (presuming Travis passes). 🐧

api/comments/views.py
TomBaxter: I don't believe I can use get_serializer_class here. This logic is for the call to a particular comment....
e.g. http://localhost:8000/v2/comments/ehar9ypk7wt8/
get_serializer_class is used for listing comments in a node/registration...
e.g. http://localhost:8000/v2/registrations/avn9t/comments/?filter[target]=avn9t

PR #6060, opened 2016-07-28
Add date modified comparison filtering to registrations

zamattiac:
## Purpose

To filter registrations by their date modified, particularly for API scraping.

## Changes

Originally, the field had a HideIfWithdrawal serialization, which prevented the filtering. Product allowed removing this feature, so I did. It serializes normally now. 

## Side effects

- Withdrawing the registration won't hide its last date modified
- Unit tests had to be changed to allow date_modified in withdrawn registrations

## Ticket

https://openscience.atlassian.net/browse/OSF-6686


zamattiac:
I'm going back to additional development to work on unit tests

api/registrations/serializers.py
```diff
@@ -26,7 +26,8 @@ class BaseRegistrationSerializer(NodeSerializer):
     category_choices_string = NodeSerializer.category_choices_string
     category = HideIfWithdrawal(ser.ChoiceField(read_only=True, choices=category_choices, help_text='Choices: ' + category_choices_string))
 
-    date_modified = HideIfWithdrawal(ser.DateTimeField(read_only=True))
+    date_modified = ser.DateTimeField(read_only=True)
+
```
mfraezz: Superfluous whitespace, but that's not a blocker.

mfraezz:
Pass complete. Verified this now works for all cases specified on the ticket; no bugs found :star:

Regarding `Side Effects`:
> Unit tests had to be changed to allow date_modified in withdrawn registrations

Behavior changed, so this is expected.

> Withdrawing the registration won't hide its last date modified

Nor should it. In fact, in discussing this with product, `date_modified` should be updated when the registration is withdrawn. Currently, it is not.

**TODO**: This can be accomplished by adding a log to the registration when it's withdrawn, in `website.project.sactions.Retractions._on_complete`, or (perhaps more cleanly) by manually modifying this field upon retraction completion. However, this will require a migration of any existing retractions -- their `.registered_from` will have a `RETRACTION_APPROVED` `NodeLog` with the timestamp you're looking for. I can advise more on the migration, if needed. Back to you, @zamattiac :back:

zamattiac:
@mfraezz This improvement is a bit of a blocker for the mirroring project. Is there any way we can spin that off into a separate ticket (which I will take) so that the API filter can get implemented?

mfraezz:
No. Maintaining correct behavior on the OSF is important, especially when it comes to registrations. This shouldn't take long -- aside from the migration script, it should be about a one line change. And again, I can help with the migration script if needed.

scripts/migration/migrate_retraction_dates.py
```diff
@@ -0,0 +1,54 @@
...
+        for target in targets:
+            logger.info("{} {}: {}".format(target.date, target.params['registration'], target.action))
+    else:
+        do_migration(targets)
+
+if __name__ == '__main__':
+    dry = '--dry' in sys.argv
+    if not dry:
+        script_utils.add_file_logger(logger, __file__)
+    main(dry=dry)
```
mfraezz: Not a huge deal, but there should be a newline at the end of this file.

scripts/migration/migrate_retraction_dates.py
```diff
@@ -0,0 +1,54 @@
...
+    logger.info("Retractions found: {}".format(len(targets)))
+    return targets
+
+
+def main(dry=True):
+    init_app(set_backends=True, routes=False)  # Sets the storage backends on all models
+
+    targets = get_targets()
+
+    if dry:
```
mfraezz: This conditional should be removed.

scripts/migration/migrate_retraction_dates.py
```diff
@@ -0,0 +1,54 @@
...
+        for target in targets:
+            logger.info("{} {}: {}".format(target.date, target.params['registration'], target.action))
+    else:
+        do_migration(targets)
```
mfraezz: `do_migration` should be wrapped in a `TokuTransaction`. Afterwards, `if dry_run`, you should `raise RuntimeError('Dry run, transaction rolled back.')`

scripts/migration/migrate_retraction_dates.py
```diff
@@ -0,0 +1,54 @@
...
+
+    # ... perform the migration using a list of logs ...
+    for log in logs:
+        when = log.date
+        registration = Node.load(log.params['registration'])
+
+        registration.update_fields(date_modified=when)
+        registration.save()
+
+        logger.warning("{} date updated to {}".format(registration, when))
```
mfraezz: This should be `.info`

scripts/migration/migrate_retraction_dates.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def do_migration(logs):
+
+    # ... perform the migration using a list of logs ...
+    for log in logs:
+        when = log.date
+        registration = Node.load(log.params['registration'])
+
+        registration.update_fields(date_modified=when)
```
mfraezz: Why call `update_fields`?

`registration.date_modified = log.date` would be equivalent.

scripts/migration/migrate_retraction_dates.py
```diff
@@ -0,0 +1,54 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def do_migration(logs):
+
+    # ... perform the migration using a list of logs ...
+    for log in logs:
+        when = log.date
+        registration = Node.load(log.params['registration'])
```
mfraezz: On staging, only 30 of 185 NodeLogs have `registration` in the params, which would definitely break this. Investigating...

scripts/migration/migrate_retraction_dates.py
```diff
@@ -0,0 +1,54 @@
...
+
+        registration.update_fields(date_modified=when)
+        registration.save()
+
+        logger.warning("{} date updated to {}".format(registration, when))
+
+
+def get_targets():
+    # ... return the list of logs whose registrations we want to migrate ...
+    query = Q('action', 'eq', "retraction_approved")
```
mfraezz: If you're not using `query` elsewhere, you can just put it in `.find`:
```python
targets = NodeLog.find(Q('action', 'eq', 'retraction_approved'))
```

Also, as per [cosdev](http://cosdev.readthedocs.io/en/latest/style_guides/python.html), please use singlequotes in python, unless it would require lots of escaping.

mfraezz: It's apparently not guaranteed to be a 1-to-1 relationship between `retraction_approved` logs, and retracted registrations. This means that it's only the most recent log's `.date` should be used

scripts/migration/migrate_retraction_dates.py
```diff
@@ -0,0 +1,54 @@
...
+    targets = NodeLog.find(query)
+
+    logger.info("Retractions found: {}".format(len(targets)))
```
mfraezz: This isn't strictly correct (on staging, at least):
```
In [63]: models.NodeLog.find(Q('action', 'eq', 'retraction_approved')).count()
Out[63]: 185

In [64]: models.Retraction.find().count()
Out[64]: 182
```

scripts/migration/migrate_retraction_dates.py
mfraezz: Good news: it appears to mostly from a time before Retractions or NodeLogs were well-formed. Every `retraction_approved` log on prod has a `params.registration`... but please add some error handling to make sure this will succeed on staging data.

scripts/migration/migrate_retraction_dates.py
mfraezz: Consider serializing the set of targets into a list of `{<_id>: <correct_date>}` mappings.

website/project/sanctions.py
```diff
@@ -691,6 +692,10 @@ def _on_complete(self, user):
             node.set_privacy('public', auth=None, save=True, log=False)
             node.update_search()
 
+        parent_registration.date_modified = datetime.datetime.utcnow()
+        parent_registration.save()
+
+
```
mfraezz: >flake8 .
./website/project/sanctions.py:699:5: E303 too many blank lines (2)

api_tests/registrations/views/test_withdrawn_registrations.py
```diff
@@ -100,10 +100,10 @@ def test_withdrawn_registrations_display_limited_fields(self):
             'description': registration.description,
             'date_created': registration.date_created.isoformat(),
             'date_registered': registration.registered_date.isoformat(),
+            'date_modified': registration.date_modified.isoformat(),
```
mfraezz: The build failed because `registration.date_modified` isn't rounded here, but is in the API response. Calling `registration.reload()` above should fix that.

scripts/migration/migrate_retraction_dates.py
```diff
@@ -0,0 +1,50 @@
...
+logger = logging.getLogger(__name__)
+
+
+def do_migration(logs):
+    # ... perform the migration using a list of logs ...
+    for log in logs:
+        registration_id = log.params.get('registration')
+        if registration_id:
+            registration = Node.load(registration_id)
+            registration.date_modified = log.date
```
mfraezz: I added an assert just before this line, and it failed:
```
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 162, in _run_module_as_main
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/Users/matt/Devel/osf.io/scripts/migration/migrate_retraction_dates.py", line 51, in <module>
    raise RuntimeError('Dry run, rolling back transaction.')
  File "/Users/matt/Devel/osf.io/scripts/migration/migrate_retraction_dates.py", line 49, in <module>
    do_migration(targets)
  File "/Users/matt/Devel/osf.io/scripts/migration/migrate_retraction_dates.py", line 23, in do_migration
    assert log.date >= registration.date_modified, 'Tried to roll date_modified backwards from {} to {}'.format(registration.date_modified, log.date)
AssertionError: Tried to roll date_modified backwards from 2016-06-30 00:00:29.531000 to 2016-06-24 20:05:17.769000
```

mfraezz: The most recent change on the offending node is from a `made_public` `NodeLog`, presumably because it was embargoed. Please handle this case too.

scripts/migration/migrate_retraction_dates.py
```diff
@@ -0,0 +1,53 @@
...
+    for log in logs:
+        registration_id = log.params.get('registration')
+        if registration_id:
+            registration = Node.load(registration_id)
+            if registration.date_modified < log.date:
+                registration.date_modified = log.date
+                registration.save()
+                logger.info('{} date updated to {}'.format(registration, log.date))
+            else:
+                logger.warning('Date modified is more recent than retraction ' + log._id)
```
mfraezz: This should be `.info`

scripts/migration/migrate_retraction_dates.py
```diff
@@ -0,0 +1,53 @@
...
+        else:
+            logger.warning('No parent registration found for retraction log ' + log._id)
+
+
+def get_targets():
+    # ... return the list of logs whose registrations we want to migrate ...
+    targets = NodeLog.find(Q('action', 'eq', 'retraction_approved'))
+
+    logger.info('Retractions found: {}'.format(len(targets)))
```
mfraezz: Again, not strictly correct, because a 1-1 correlation between retraction approval logs and retracted nodes isn't guaranteed, but not a blocker.

mfraezz:
Pass complete, LGTM. Verified that the issue was (still) fixed and that the migration works on staging data. Moving this forward 🎊

mfraezz:
## Deployment Node ( @sloria ) :

Migration:
```
python -m scripts.migration.migrate_retraction_dates --dry
```

PR #6067, opened 2016-07-29
Feature/mailing lists [OSF-3975]

HalcyonChimera:
## Purpose
Implements mailing list integration with the OSF

## Changes
- Mailing list is created on node creation
- Archiver visibility set to private by default. (Only allow contributors to see the list)
- Archiver visibility updates when the node is changed between private and public
- Mailing list title and description updated when the node is updated
- Contributors added automatically to list when they're added to a project or component
- Allow contributors to unsubscribe from a mailing list if desired
- Users may choose which email address (that is associated with their account) to subscribe to the mailing list

## Side effects
Touches a few addon tests, but is fairly isolated as it is now.
Touches notifications moderately

## Ticket
OSF-3975


brianjgeiger:

PR #6070, opened 2016-07-30
GitHub: modal message inconsistent for folder deletion[#OSF-6420]

chennan47:
## Purpose

Fix the delete folder pop test for github is different from other addons.
## Ticket


brianjgeiger:
Looks good. 🎉 🐧 🎉

PR #6072, opened 2016-08-01
Improve Prerender functionalities [OSF-6735]

adlius:
## Purpose

To improve functionality of Prerender without affecting the experience of normal users. Prerender should wait for a page to render completely, or return a 504 status code if a page fails to render before timeout.

## Changes

Three lines are added to base.mako within a conditional so that they are only viewable to prerender. In other words, the change to base.mako can make sure there are two different set of templates, one for normal browsers and another one for robots (as identified by user agent string of the request).
A new javascript file (prerender.js) is added to be included in the template for robots. The purpose of the js file is to determine whether there are currently active XHR requests. If not, then the meta tag of "prerender-status-code" would be set to 200 (so that Prerender knows to return a 200 to the request) and the prerenderReady variable of the window will be set to true.

## Side effects

The beauty of this approach is that there is no side effects, or any effects at all on normal users. (Big thanks to @JeffSpies who came up with the idea) 
## Ticket

https://openscience.atlassian.net/browse/OSF-6735

### DIFF
--- a/website/static/js/prerender.js
+++ b/website/static/js/prerender.js
@@ -1,0 +1,25 @@
+window.activeAjaxCount = 0;
+
+function isAllXhrComplete(){
+    window.activeAjaxCount--;
+    if (window.activeAjaxCount === 0){
+        $('meta[name=prerender-status-code]').attr('content', '200');
+        window.prerenderReady = true;
+    }
+
+}
+
+(function(open) {
+    XMLHttpRequest.prototype.open = function(method, url, async, user, pass) {
+        this.addEventListener('load', isAllXhrComplete);
+        open.call(this, method, url, async, user, pass);
+    };
+})(XMLHttpRequest.prototype.open);
+
+
+(function(send) {
+    XMLHttpRequest.prototype.send = function (data) {
+        window.activeAjaxCount++;
+        send.call(this, data);
+    };
+})(XMLHttpRequest.prototype.send);
--- a/website/templates/base.mako
+++ b/website/templates/base.mako
@@ -1,4 +1,5 @@
 <% from website import settings %>
+<% from flask import request %>
 
 <!DOCTYPE html>
 <html lang="en">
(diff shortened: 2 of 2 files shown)


brianjgeiger:
There are a bunch of failing tests. Once the tests have passed, go ahead and place this in Code Review. Pass done. 🐧

adlius:
The tests seems to fail because the request sent in the tests does not contain a user-agent string, causing the server-side rendering of templates to crash. Added a conditional to check whether user agent string is contained in the request header. I'll move this to CR once tests are done.

brianjgeiger:
Two general concerns:
1) Is there a way to test this that won't significantly slow down test times;
2) My understanding is that Google SEO is heavily weighted to fast response times from pages, and this would likely slow that down. Would this affect that, or do we cache all the pre-rendered pages (or the important pre-rendered pages). If we do cache, are there potential consequences to slowing that process down at all?

brianjgeiger:
How fault-tolerant is this? Will certain kinds of javascript errors prevent the pre-rendering of the page entirely?

brianjgeiger:
🐧 Pass done. This will likely be another thing that may get set aside unless Product puts a priority on it, but it would be good to know what the consequences of the changes would be.

adlius:
@brianjgeiger The change could potentially speed up the Prerender-ing speed for some pages as the page will be returned as soon as the first-round AJAX have finished. But more importantly, it is for the purpose of making sure pages with heavy API calls are Prerendered correctly by making it wait longer. For example, the page for reproducibility project is not prerendered correctly under the current setting. This fix could potentially solve the problem if the settings are correct.
If so, the meta-question seems to be, "do we want Google to index us quickly, or correctly?". 
Also, another choice is to not Prerender for Google at all. Since last October, Google is able to index pages with AJAX contents.

adlius:
@brianjgeiger Also, Prerender itself is not fault-tolerant at all... Based on PhantomJS, Prerender seems to have all sorts of problems with speed (it is already very slow currently) and memory (to deal with memory leaks, it needs to be constantly restarting after certain number of iterations). I guess using Prerender was more like an improvement to SEO. But as to how much our SE rankings are improved as a result of using Prerender, it is still an open question, due to the (already) slow speed, crashing problem and incorrectly rendered pages.

brianjgeiger:
Unfortunately, the questions about whether we should are not mine to answer; product needs to prioritize the issue, especially if it's one like this. I recommend adding much of this information to the Jira ticket and tagging Nici and Sara to prioritize. What would also be helpful are some benchmarks about how this should change real-world usage along with some examples of the errors and which situations this should fix. It's a fairly fundamental change to something that I don't know if it's well tested, so some data would be awesome.

adlius:
I see. I'll evaluate the situation further and try to gather some data and do some benchmarking. Thanks for the input.

brianjgeiger:

PR #6074, opened 2016-08-01
[Feature] Fix guid issue and q26 [OSF-6411]

Ghalko:
## Purpose
This fixes three issues from the previous PR:
- files not loading from misplaced guid
- generating a guid for older files
- and taking care of the special case of `q26` which is different from all of the rest.

## Changes
- Add checking and different ways to handle file questions
- Add tests

## Ticket

- Also add tests


brianjgeiger:
Looks great. 🎉 🐧 🎉

PR #6075, opened 2016-08-01
[CR] ownCloud Addon

kwierman:
## Purpose
 Provides the ownCloud addon. ownCloud is a cloud based storage platform with a number of different implementations. This addon allows the settings to specify a number of different hosts to utilize and creates settings for the associated storage provider.

## Changes

These changes *should* be independent of the rest of the osf package.

## Side effects
 
None at the moment.


felliott:
Huzzah!

Quick note, @mfraezz: if you're testing against a local docker image of owncloud, make sure to pass the schema with the url in the user settings.  @kwierman is aware of this and will fix.

felliott:
@kwierman: the "Go To Files page" link displayed after you select the linked folder is being set to `/settings/undefined`:

screen shot 2016-08-01 at 3 08 45 pm

website/addons/owncloud/README.md
```diff
@@ -0,0 +1,18 @@
+OSF ownCloud Addon
+
+
+## Setup:
+1. Copy settings/local-dist.py to settings/local.py and change the necessary settings.
+2. Make sure to list your server under `trusted_servers`
```
mfraezz: Is this necessary/true? It looks like a user can specify a custom host when connecting (which makes sense, given ownCloud's model)

website/addons/owncloud/model.py
```diff
@@ -0,0 +1,148 @@
...
+
+class AddonOwnCloudUserSettings(AddonOAuthUserSettingsBase):
+    oauth_provider = OwnCloudProvider
+    serializer = OwnCloudSerializer
+
+class AddonOwnCloudNodeSettings(StorageAddonBase, AddonOAuthNodeSettingsBase):
+    oauth_provider = OwnCloudProvider
+    serializer = OwnCloudSerializer
+
+    folder_name = fields.StringField()
```
mfraezz: The preferred "primary" folder-related field is `folder_id`. This will matter for APIv2 support.

Conceptually, it's supposed to be whatever the service provider uses as a unique identifier. Some providers key by path (like GoogleDrive), others by an ID unrelated to name or path (like Box). It seems odd that ownCloud would key by name -- do they ensure uniqueness of every folder name owned by a user, regardless of hierarchical level?

website/addons/owncloud/model.py
```diff
@@ -0,0 +1,148 @@
...
+
+    @property
+    def complete(self):
```
mfraezz: This should probably be using the superclass' implementation -- it looks like it's trying to do the work of the `configured` property, currently.

website/addons/owncloud/model.py
```diff
@@ -0,0 +1,148 @@
...
+    @property
+    def folder_id(self):
+        return self.folder_name
+
+    @property
+    def folder_path(self):
+        return self.folder_name
+
+    @property
+    def nodelogger(self):
```
mfraezz: `AddonOAuthNodeSettingsBase` provides a generic `AddonNodeLogger` with a `.nodelogger` property, this (including the `OwnCloudNodeLogger` class definition) is probably unnecessary.

website/addons/owncloud/model.py
```diff
@@ -0,0 +1,148 @@
...
+                action='owncloud_folder_selected',
+                params={
+                    'project': self.owner.parent_id,
+                    'node': self.owner._id,
+                    'folder': folder,
+                },
+                auth=auth,
+            )
+
+    def _get_fileobj_child_metadata(self, filenode, user, cookie=None, version=None):
```
mfraezz: Is this override necessary? In what cases, when requesting metadata from WaterButler, would an HTTPError be raised that we wanted to suppress?

website/addons/owncloud/model.py
```diff
@@ -0,0 +1,148 @@
...
+            auth=auth
+        )
+
+    def set_folder(self, folder, auth=None):
+        if folder == '/ (Full ownCloud)':
+            folder = '/'
+        self.folder_name = folder
+        self.save()
+        if auth:
+            self.owner.add_log(
```
mfraezz: This can probably be replaced with `self.nodelogger.log(action='folder_selected', save=True)`, using the generic `nodelogger`

website/addons/owncloud/model.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def deauthorize(self, auth=None, add_log=True):
+        """Remove user authorization from this node and log the event."""
+        self.clear_settings()
+        self.clear_auth()  # Also performs a save
+
+        # Log can't be added without auth
+        if add_log and auth:
+            node = self.owner
+            self.owner.add_log(
```
mfraezz: This can probably be replaced with `self.nodelogger.log(action='node_deauthorized', extra=<dict>, save=True)`

website/addons/owncloud/model.py
```diff
@@ -0,0 +1,148 @@
...
+                    'project': node.parent_id,
+                    'node': node._id,
+                },
+                auth=auth,
+            )
+
+    def serialize_waterbutler_credentials(self):
+        if not self.has_auth:
+            raise exceptions.AddonError('Addon is not authorized')
+        converted = ExternalAccountConverter(account=self.external_account)
```
mfraezz: I'm not certain that an `ExternalAccountConverter` is necessarily the best approach for this use case. More on this later

website/addons/owncloud/serializer.py
```diff
@@ -0,0 +1,81 @@
...
+class OwnCloudSerializer(OAuthAddonSerializer):
+
+    addon_short_name = 'owncloud'
+
+    REQUIRED_URLS = []
+
+    # Include host information with more informative labels / formatting
+    def serialize_account(self, external_account):
+        ret = super(OwnCloudSerializer, self).serialize_account(external_account)
+        converted = ExternalAccountConverter(external_account)
```
mfraezz: It seems unnecessary to go through the password decryption process here. Why not just use `external_account.oauth_key` below?

website/addons/owncloud/serializer.py
```diff
@@ -0,0 +1,81 @@
+from website.addons.base.serializer import OAuthAddonSerializer
+from website.addons.owncloud.settings import DEFAULT_HOSTS
+from website.util import api_url_for, web_url_for
+from website.addons.owncloud.utils import ExternalAccountConverter
+
+class OwnCloudSerializer(OAuthAddonSerializer):
```
mfraezz: It seems like you should be using `StorageAddonSerializer`, given that ownCloud is a storage provider.

website/addons/owncloud/serializer.py
```diff
@@ -0,0 +1,81 @@
...
+            assert url in addon_urls, "addon_serilized_urls must include key '{0}'".format(url)
+        ret.update(addon_urls)
+        return ret
+
+    @property
+    def addon_serialized_urls(self):
+        node = self.node_settings.owner
+
+        return {
+            'create': api_url_for('owncloud_add_user_account'),
```
mfraezz: This key should probably be `auth`, and the one below should probably be `config`. This will allow you to use [`DRY`](https://en.wikipedia.org/wiki/Don't_repeat_yourself)er front-end code (more on that later).

website/addons/owncloud/serializer.py
```diff
@@ -0,0 +1,81 @@
...
+    def serialized_urls(self):
+        external_account = self.node_settings.external_account
+        node = self.node_settings.owner
+        ret = {
+            'settings': web_url_for('user_addons'),
+            'folders': node.api_url_for('owncloud_folder_list'),
+            'files': node.api_url_for('owncloud_folder_list'),
+            'config': node.api_url_for('owncloud_set_config'),
+        }
+        # Dataverse users do not currently have profile URLs
```
mfraezz: >Dataverse users do not currently have profile URLs

I'm assuming this is also true for ownCloud users?

website/addons/owncloud/serializer.py
```diff
@@ -0,0 +1,81 @@
...
+            'name': node_settings.folder_name,
+            'path': node_settings.folder_name
+        }
+
+    @property
+    def credentials_owner(self):
+        return self.node_settings.user_settings.owner
+
+    @property
+    def serialized_urls(self):
```
mfraezz: If you're using `StorageAddonSerializer` and don't unset `REQUIRED_URLS`, the `super` implementation of this should be fine.

website/addons/owncloud/serializer.py
```diff
@@ -0,0 +1,81 @@
...
+            'accounts': api_url_for('owncloud_account_list'),
+        }
+
+    @property
+    def serialized_node_settings(self):
+        result = super(OwnCloudSerializer, self).serialized_node_settings
+        result['hosts'] = DEFAULT_HOSTS
+
+        # Update with Dataverse specific fields
+        if self.node_settings.has_auth:
```
mfraezz: This block is likely superfluous:
1) `has_auth` doesn't (at least, shouldn't) guarantee a folder.
2) `serialize_settings` should be the only thing that needs to include a `folder` in this class, and if you use `StorageAddonSerializer` that's given to you for free
3) This class shouldn't care about `Dataverse specific fields`

website/addons/owncloud/serializer.py
```diff
@@ -0,0 +1,81 @@
+from website.addons.base.serializer import OAuthAddonSerializer
+from website.addons.owncloud.settings import DEFAULT_HOSTS
+from website.util import api_url_for, web_url_for
+from website.addons.owncloud.utils import ExternalAccountConverter
+
+class OwnCloudSerializer(OAuthAddonSerializer):
+
+    addon_short_name = 'owncloud'
+
+    REQUIRED_URLS = []
```
mfraezz: This should probably not be unset

website/addons/owncloud/serializer.py
mfraezz: This will require implementing a `credentials_are_valid` method, which may make a trivial request to the connected ownCloud to test for credential validity. Overall, though, it should simplify this class.

website/addons/owncloud/serializer.py
```diff
@@ -0,0 +1,81 @@
...
+        return ret
+
+    def serialized_folder(self, node_settings):
+        return {
+            'name': node_settings.folder_name,
+            'path': node_settings.folder_name
+        }
+
+    @property
+    def credentials_owner(self):
```
mfraezz: This should probably be using the `super` implementation

website/addons/owncloud/serializer.py
```diff
@@ -0,0 +1,81 @@
...
+    def serialized_node_settings(self):
+        result = super(OwnCloudSerializer, self).serialized_node_settings
+        result['hosts'] = DEFAULT_HOSTS
+
+        # Update with Dataverse specific fields
+        if self.node_settings.has_auth:
+            result.update({'folder': self.node_settings.folder_name})
+        return result
+
+    def serialize_settings(self, node_settings, user):
```
mfraezz: If you use `StorageAddonSerializer`, the `super` implementation of this should work.

website/addons/owncloud/settings/defaults.py
```diff
@@ -0,0 +1 @@
+DEFAULT_HOSTS = ['https://localhost/owncloud']
```
mfraezz: This should probably be removed from `defaults`, as we wouldn't want it in production. It can stay in `local-dist`, though

website/addons/owncloud/static/owncloudLogActionList.json
```diff
@@ -0,0 +1,4 @@
+{
+    "owncloud_folder_selected" : "${user} selected ownCloud folder ${folder} in ${node}",
```
mfraezz: This file should have each of the `<addon>_<action>` logs common to storage providers with a mutable file/folder structure. This includes:
```
_file_added
_file_removed
_file_updated
_folder_created
_folder_selected
_node_authorized
_node_deauthorized
_node_deauthorized_no_user
```

website/addons/owncloud/static/owncloudAnonymousLogActionList.json
```diff
@@ -0,0 +1,3 @@
+{
+    "owncloud_folder_selected" : "A user selected a folder for this node"
```
mfraezz: See [this comment](https://github.com/CenterForOpenScience/osf.io/pull/6075/files#r73200861) on other log action file.

website/addons/owncloud/static/owncloudNodeConfig.js
```diff
@@ -0,0 +1,472 @@
...
+var bootbox = require('bootbox');
+var Raven = require('raven-js');
+
+var $osf = require('js/osfHelpers');
+
+var $modal = $('#ownCloudCredentialsModal');
+var oop = require('js/oop');
+var FolderPickerViewModel = require('js/folderPickerNodeConfig');
+
+var ViewModel = oop.extend(FolderPickerViewModel,{
```
mfraezz: Consider `oop.extend`ing `OauthAddonFolderPicker`. This will dramatically reduce the amount of code needed in this file. Recent changes on develop make [`s3NodeConfig.js`](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/addons/s3/static/s3NodeConfig.js#L15) a good example of how to extend that viewmodel to support addons that don't support OAuth.

website/addons/owncloud/static/owncloudUserConfig.js
```diff
@@ -0,0 +1,226 @@
...
+    self.selector = selector;
+    self.url = url;
+    // On success, instantiate and bind the ViewModel
+    self.viewModel = new ViewModel(url);
+    osfHelpers.applyBindings(self.viewModel, self.selector);
+}
+
+module.exports = {
+    OwnCloudViewModel: ViewModel,
+    OwnCloudUserConfig: OwnCloudUserConfig    // for backwards-compat
```
mfraezz: >`// for backwards-compat`

This is new, why is backwards compatibility an issue?

website/addons/owncloud/static/user-cfg.js
```diff
@@ -0,0 +1,9 @@
+var $osf = require('js/osfHelpers');
+var OwnCloudViewModel = require('./owncloudUserConfig.js').OwnCloudViewModel;
+
+var url = '/api/v1/settings/owncloud/';
+
+var owncloudViewModel = new OwnCloudViewModel(url);
+$osf.applyBindings(owncloudViewModel, '#owncloudAddonScope');
```
mfraezz: This viewmodel initialization is what's being done by `OwnCloudUserConfig`. Why not `require` and instantiate that instead?

website/addons/owncloud/static/user-cfg.js
```diff
@@ -0,0 +1,9 @@
...
+
+owncloudViewModel.fetch();
```
mfraezz: This call should happen when either `OwnCloudUserConfig` or `OwnCloudViewModel` is instantiated, you shouldn't need to call it here.

website/addons/owncloud/static/owncloudNodeConfig.js
```diff
@@ -0,0 +1,472 @@
...
+            type: 'GET',
+            dataType: 'json'
+        }).done(function(response) {
+            // Update view model
+            self.folderName(response.result.folder);
+            self.updateFromData(response.result);
+            self.loadedSettings(true);
+        }).fail(function(xhr, textStatus, error) {
+            self.changeMessage(self.messages.userSettingsError, 'text-danger');
+            Raven.captureMessage('Could not GET dataverse settings', {
```
mfraezz: >Could not GET dataverse settings

website/addons/owncloud/tests/utils.py
```diff
@@ -0,0 +1,61 @@
...
+    mock_owncloud.logout.return_value = None
+    mock_owncloud.list.return_value =[
+        FileInfo('/Documents', file_type='dir', attributes=None),
+        FileInfo('/Pictures', file_type='dir', attributes=None),
+        FileInfo('/secrets.txt', file_type='txt', attributes=None),
+    ]
+    return mock_owncloud
+
+def create_external_account(host='foo.bar.baz', username='doremi-abc-123',password='eh'):
+    """Creates external account for Dataverse with fields populated the same
```
mfraezz: >Creates external account for Dataverse

website/addons/owncloud/utils.py
```diff
@@ -0,0 +1,45 @@
...
+                provider_name='owncloud',
+                display_name=username,
+                oauth_key=host,
+                oauth_secret=encrypt(password),
+                provider_id=host
+            )
+
+from website.addons.base.logger import AddonNodeLogger
+
+class OwnCloudNodeLogger(AddonNodeLogger):
```
mfraezz: As mentioned [elsewhere](https://github.com/CenterForOpenScience/osf.io/pull/6075/files#r73179230), this class is probably unnecessary.

website/addons/owncloud/utils.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
+from website.oauth.models import ExternalAccount
+from website.security import encrypt, decrypt
+
+
+class ExternalAccountConverter:
```
mfraezz: The behavior that this class provides isn't necessarily something unique to ownCloud -- it's something that would need to be done for any service provider that uses Basic Auth. Consider moving this to somewhere like the `website.oauth.models` module, renaming it `BasicAuthProviderMixin`, and having `OwnCloudProvider` inherit from it.

(And updating the docstring)

website/addons/owncloud/utils.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+        OwnCloudProvider from website.addons.owncloud.models hijacks the oauth
+        framework to serialize account credentials for owncloud.
+        ExternalAccountConverter provides a wrapper class for mapping between
+        the two.
+    """
+    def __init__(self, account=None, host=None, username=None, password=None):
```
mfraezz: This seems like there should be three successful initialization scenarios:
1) an account is provided, but nothing else is
2) nothing is provided
3) a host, username, and password are provided, but account is not.

An error should be raised in any other case.

I'm also tempted to recommend breaking case 3 out into its own method -- `encrypt_external_account` or some such, so as to disambiguate the intended behavior when initializing this class, and allow this to accept only `account` or nothing. Creating a new `ExternalAccount` whenever you pass this raw credentials could be dangerous, as saving them (which I know isn't being done here) may result in EA's that violate uniqueness constraints.

website/addons/owncloud/utils.py
```diff
@@ -0,0 +1,45 @@
...
+            self.username = None
+            self.password = None
+        else:
+            self.account = ExternalAccount(
+                provider='owncloud',
+                provider_name='owncloud',
+                display_name=username,
+                oauth_key=host,
+                oauth_secret=encrypt(password),
+                provider_id=host
```
mfraezz: >may result in EA's that violate uniqueness constraints

Especially when, given that `ExternalAccount` is `unique_on(['provider', 'provider_id'])`, you could only have one account linked per ownCloud instance, currently. 

Consider field changes:
* `provider_id` to store `username` 
* depending on whether or not ownCloud has public user profiles associate with any given account, `profile_url` *could* contain `host`
* it follows the standard convention best if what's stored in `oauth_key` is what one would use to actually make a request, so `encrypt(password)`. `oauth_secret`, which is generally unused, could store `host`. Not necessarily a blocker.

website/addons/owncloud/model.py
```diff
@@ -0,0 +1,148 @@
...
+        auth = None
+        if self.user_settings:
+            auth = Auth(self.user_settings.owner)
+        return OwnCloudNodeLogger(
+            node=self.owner,
+            auth=auth
+        )
+
+    def set_folder(self, folder, auth=None):
+        if folder == '/ (Full ownCloud)':
```
mfraezz: This class should have a `def fetch_folder_name` that reverses this conversion, so that the front-end doesn't display `/` as the linked folder (and for `generic_view.root_folder` support).

website/addons/owncloud/views.py
```diff
@@ -0,0 +1,173 @@
...
+owncloud_deauthorize_node = generic_views.deauthorize_node(
+    SHORT_NAME
+)
+
+owncloud_get_config = generic_views.get_config(
+    SHORT_NAME,
+    OwnCloudSerializer
+)
+
+def owncloud_root_folder(node_settings, auth, **kwargs):
```
mfraezz: With a small model change, you should be able to use `generic_views.root_folder` for this

website/addons/owncloud/views.py
```diff
@@ -0,0 +1,173 @@
...
+        name=node_settings.folder_name,
+        permissions=auth,
+        user=auth.user,
+        nodeUrl=node.url,
+        nodeApiUrl=node.api_url,
+    )
+    return [root]
+
+@must_be_logged_in
+def owncloud_user_config_get(auth, **kwargs):
```
mfraezz: If you modified the `ownCloudUserSettings.to_json` to make a super call then inject `hosts` in the return, you could be using the generic method of getting a user's config (see `website/addons/views/base.py:get_addon_user_config`). This could simplify both routing and frontend code, but isn't necessarily a blocker.

website/addons/owncloud/views.py
```diff
@@ -0,0 +1,173 @@
...
+def owncloud_add_user_account(auth, **kwargs):
+    """Verifies new external account credentials and adds to user's list"""
+    user = auth.user
+    provider = OwnCloudProvider()
+
+    host = request.json.get('host').rstrip('/')
+    username = request.json.get('username')
+    password = request.json.get('password')
+
+    oc = OwnCloudClient(host, verify_certs=False)
```
mfraezz: ...I haven't looked terribly far down this rabbit hole, but it seems like SSL certification is something we'd want to verify, for the security of user data?

mfraezz: @kwierman, do you have any guess as to how common or uncommon it would be for any given ownCloud user to not enable SSL on a server they've put up? For something that only accepts Basic Auth, it seems like using SSL would be the norm rather than the exception, but I might have some biases at play here. Regardless, we _cannot_ allow plaintext passwords to be transmitted over the network in a production environment -- it's just too much of a security risk.

It seems like the right UX here would be to:
* have SSL required
* an error raised here if it's not enabled for a `host`, causing users to be presented a message about encryption being a good thing, and possibly directing them to [ownCloud documentation about secure encryption practices](https://doc.owncloud.org/server/9.1/admin_manual/configuration_server/harden_server.html?highlight=ssl#use-https)

Pinging @saradbowman or @nicipfeiffer about UX / language

website/addons/owncloud/views.py
```diff
@@ -0,0 +1,173 @@
...
+    oc.login(username, password)
```
mfraezz: If this doesn't already present the user with a helpful error message after an unsuccessful login, this should probably also be wrapped in a `try`/`except` that can raises an error for this purpose.

website/addons/owncloud/utils.py
kwierman: Should I do this with it's own commit or PR then?

website/addons/owncloud/views.py
```diff
@@ -0,0 +1,173 @@
...
+    if provider.account not in user.external_accounts:
+        user.external_accounts.append(provider.account)
+
+    user_addon = auth.user.get_addon('owncloud')
+    if not user_addon:
+        user.add_addon('owncloud')
+    user.save()
+
+    # Need to ensure that the user has owncloud enabled at this point
+    user.get_or_add_addon('owncloud', auth=auth)
```
mfraezz: This appears to be doing the same thing as the block above.

website/addons/owncloud/views.py
```diff
@@ -0,0 +1,173 @@
...
+    user.save()
+
+    return {}
+
+@must_have_addon(SHORT_NAME, 'user')
+@must_have_addon(SHORT_NAME, 'node')
+def owncloud_folder_list(node_addon, user_addon, **kwargs):
```
mfraezz: This isn't necessarily a blocker, but we've been working on APIv2 support for much of the intended functionality for addons -- particularly folder listing. `v1` views aren't accessible outside of a `flask` request context, so the bulk of this method will (eventually) need to be moved to the `OwnCloudNodeSettings` model.

website/addons/owncloud/utils.py
mfraezz: It will certainly get in faster in a separate PR, but I don't know of any other addons nearing completion that use Basic Auth, so either is fine.

website/addons/owncloud/views.py
mfraezz: Check recent changes to S3, Box, Dropbox, and Google Drive on `develop`

website/addons/owncloud/utils.py
mfraezz: A related but less important note: APIv2 will serialize `provider`, `profile_url`, and `display_name`, so it's good to put useful but non-sensitive things there.

website/addons/owncloud/views.py
mfraezz: This could be handled by a setting, e.g. `USE_SSL`, to make it easy to disable for development.

kwierman: Sounds like a plan. I ran into a problem using SSL on my docker image of oC, and it looks like [I'm not the only one](https://github.com/owncloud/pyocclient/issues/102). I assume professional providers will have verified certs up, so this seems like a good compromise.

mfraezz:
Please update `addons.json`'s `addons_description` with language describing ownCloud, @saradbowman might be willing to provide the language.

Please also add `"owncloud": "https://owncloud.org/"` to `addons_url` in that same file.

website/addons/owncloud/utils.py
felliott: @mfraezz and I discussed this.  Please make it part of this PR, otherwise we won't have any real-world use cases to test it with.

website/addons/owncloud/static/owncloudUserConfig.js
```diff
@@ -0,0 +1,226 @@
...
+var language = require('js/osfLanguage').Addons.owncloud;
+var osfHelpers = require('js/osfHelpers');
+var addonSettings = require('js/addonSettings');
+
+var ExternalAccount = addonSettings.ExternalAccount;
+
+var $modal = $('#ownCloudCredentialsModal');
+
+
+function ViewModel(url) {
```
mfraezz: Hinted at [elsewhere](https://github.com/CenterForOpenScience/osf.io/pull/6075/files/b9fcb57ebb58841535a051209878dbcb51edbcac#r73215120), but with a small model change this could probably be simplified by `oop.extend`ing `OAuthAddonSettingsViewModel` in `website/static/js/addonSettings.js`

website/addons/owncloud/static/owncloudUserConfig.js
```diff
@@ -0,0 +1,226 @@
...
+
+var ko = require('knockout');
+var $ = require('jquery');
+var Raven = require('raven-js');
+var bootbox = require('bootbox');
+require('js/osfToggleHeight');
+
+var language = require('js/osfLanguage').Addons.owncloud;
+var osfHelpers = require('js/osfHelpers');
+var addonSettings = require('js/addonSettings');
```
mfraezz: The only thing this is used for is declaring `ExternalAccount` --  why not just 
```javascript
var ExternalAccount = require('js/addonSettings').ExternalAcccount
```

(disregard this if you implement the refactor mentioned below)

mfraezz:
Pass complete @kwierman. This is looking good. :penguin:

I only glanced over the tests because some will likely change with other requested changes, but they generally looked correct. I'll give them a more scrutinous look and do manual testing during the next pass. Feel free to ask questions or request clarification here or on Slack.

Back to you :back:

website/addons/owncloud/serializer.py
```diff
@@ -0,0 +1,65 @@
+from website.addons.base.serializer import StorageAddonSerializer
+from website.addons.owncloud.settings import DEFAULT_HOSTS
+from website.util import web_url_for
+from website.addons.owncloud.utils import ExternalAccountConverter
```
felliott: @kwierman: I think this was accidentally left in.  This module doesn't exist in the PR and keeps the OSF from starting up.  Same issue in `views.py`.

kwierman: Woops! Sorry, I'm in the middle of making changes from the last set of comments, and I accidentally pushed. I'll finish up with the refactor quickly and push again when ready

kwierman:
All right, sorry for the long delay, but I think I've responded to most/all of the comments above. The big things here are:

* Switching over the folder lists to apiv2
* Refactoring the ExternalAccountConverter to `oauth.models.BasicAuthMixin`
* Refactoring the javascript UserNodeSettings ViewModel to oop

And, on initial check, it looks like the other concerns are addressed. All right, bring on round 2. Also, if I missed anything let me know. I used this page like a checklist, so it should be covered.

website/addons/owncloud/__init__.py
```diff
@@ -0,0 +1,28 @@
+from . import model
```
mfraezz: Prefer absolute over relative module import paths, but this isn't a blocker.

kwierman: Easy fix. This might be from whichever addon I originally copypasta'd to make this one.

website/addons/owncloud/model.py
```diff
@@ -0,0 +1,153 @@
...
+from website.util import api_v2_url
+
+logger = logging.getLogger(__name__)
+
+class OwnCloudProvider(BasicAuthProviderMixin):
+    """An alternative to `ExternalProvider` not tied to OAuth"""
+
+    name = 'ownCloud'
+    short_name = 'owncloud'
+    serializer = OwnCloudSerializer
```
mfraezz: Does `OwnCloudProvider` need a serializer? It doesn't appear to be used.

kwierman: Not sure.  I used the design pattern from Dataverse, which had this declaration, and also doesn't appear to utilize the serializer via the provider anywhere. My guess was that this was needed for a base class somewhere. I'll try removing it and see what that gets us.

website/addons/owncloud/model.py
```diff
@@ -0,0 +1,153 @@
...
+        return self.folder_id
+
+    def clear_settings(self):
+        self.folder_id = None
+
+    def deauthorize(self, auth=None, add_log=True):
+        """Remove user authorization from this node and log the event."""
+        self.clear_settings()
+        self.clear_auth()
+        if add_log and auth:
```
mfraezz: I think this should be logging regardless of `auth` (which might be a superfluous kwarg, though it should currently be retained because it may be passed in). 

`AddonNodeLogger` will implicitly detect auth if it's there, and use it to try build the `user` argument for the `NodeLog` it creates. When the front-end tries to render a log, `logTextParser.js` will replace a `None` user with `'A user'`.

kwierman: Oh neat! So, the difference between the logs and the anonymous logs is done is the js layer?

Also, would it be appropriate to replace this with:

```python
def deauthorize(self, auth=None, add_log=True):
  self.clear_settings()
  self.clear_auth()
  super(OwnCloudNodeSettings, self).deauthorize(auth, add_log)
```

website/addons/owncloud/model.py
```diff
@@ -0,0 +1,153 @@
...
+            self.nodelogger.log(action='node_deauthorized', save=True)
+
+    def serialize_waterbutler_credentials(self):
+        if not self.has_auth:
+            raise exceptions.AddonError('Addon is not authorized')
+        provider = OwnCloudProvider(self.external_account)
+        return {'host': provider.host,
+                'username': provider.username,
+                'password': provider.password
+                }
```
mfraezz: Not a blocker, but we generally have multi-line iterables open and close with a curly brace on their own line, i.e.
```
return {
    'host': provider.host,
    'username': provider.username,
    'password': provider.password
}
```
However, please be consistent with what you're doing elsewhere -- `serialize_waterbutler_settings` does not have a hanging curly brace.

website/addons/owncloud/model.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def after_delete(self, node, user):
+        self.deauthorize(Auth(user=user), add_log=True)
+        self.save()
+
+    def on_delete(self):
+        self.deauthorize(add_log=False)
+        self.save()
+
+    def get_folders(self, path='/', folder_id=None):
```
mfraezz: Not blocking, but prefer `**kwargs` for this -- less likely to break if anything else needs to be added in the API call, which is intended to be consistent across addons

website/addons/owncloud/model.py
```diff
@@ -0,0 +1,153 @@
...
+
+        ret = []
+        for item in c.list(path):
+            if item.file_type is 'dir':
+                ret.append({
+                    'addon': 'owncloud',
+                    'path': item.path,
+                    'kind': 'folder',
+                    'id': item.path,
+                    'name': item.path.strip('/').split('/')[-1],
```
mfraezz: `fetch_folder_name` should probably be doing this too.

website/addons/owncloud/routes.py
```diff
@@ -0,0 +1,68 @@
+# -*- coding: utf-8 -*-
+from framework.routing import Rule, json_renderer
+
+from . import views
```
mfraezz: Prefer absolute over relative module import paths, but this isn't a blocker. (As stated elsewhere)

website/addons/owncloud/static/owncloudAnonymousLogActionList.json
```diff
@@ -0,0 +1,10 @@
+{
+    "owncloud_file_added" : "A user added a file to ownCloud in a project",
+    "owncloud_file_removed" : "A user removed a file from ownCloud in a project",
+    "owncloud_file_updated" : "A user  updated a file in ownCloud in a project",
```
mfraezz: There's an extra space here, between `user` and `updated`

website/addons/owncloud/static/owncloudLogActionList.json
```diff
@@ -0,0 +1,11 @@
+{
+    "owncloud_file_added" : "${user} added file ${path} to ownCloud in ${node}",
+    "owncloud_file_removed" : "${user} removed ${path_type} ${path} from ownCloud in ${node}",
+    "owncloud_file_updated" : "${user} updated file ${path} in ownCloud in ${node}",
+    "owncloud_folder_created" : "${user} created folder ${path} in ownCloud in ${node}",
+    "owncloud_folder_selected" : "${user} linked ownCloud folder ${folder} to ${node}",
+    "owncloud_node_authorized" : "${user} authorized the ownCloud addon for ${node}",
+    "owncloud_node_deauthorized" : "${user} deauthorized the ownCloud addon for ${node}",
+    "owncloud_node_deauthorized_no_user" : "ownCloud addon for ${node} deauthorized"
+
```
mfraezz: (Nonblocking) Superfluous line here

website/addons/owncloud/static/owncloudNodeConfig.js
```diff
@@ -0,0 +1,315 @@
...
+var language = require('js/osfLanguage').Addons.owncloud;
+
+
+var ViewModel = oop.extend(OauthAddonFolderPicker,{
+    constructor: function(addonName, url, selector, folderPicker, opts){
+        var self = this;
+        self.super.constructor.call(self, addonName, url, selector, folderPicker);
+        const otherString = 'Other (Please Specify)';
+
+        self.addonName = 'ownCloud';
```
mfraezz: I think everything between here and `self.hosts` is handled by the superclass. Please remove; there's no reason to duplicate these assignments.

website/addons/owncloud/static/owncloudNodeConfig.js
```diff
@@ -0,0 +1,315 @@
...
+                    return item.data.urls.folders;
+                },
+                decodeFolder: function(item) {
+                    return this.options.decodeFolder.call(this, item);
+                }.bind(this)
+
+            }
+        );
+
+        self.credentialsChanged = ko.pureComputed(function() {
```
mfraezz: This is duplicated from above.

website/addons/owncloud/static/owncloudNodeConfig.js
```diff
@@ -0,0 +1,315 @@
...
+                window.location.href = this.urls().auth;
+            },
+            decodeFolder: function(folder_name) {
+                return folder_name;
+            }
+        };
+        // Overrides
+        self.options = $.extend({}, defaults, opts);
+        // Treebeard config
+        self.treebeardOptions = $.extend(
```
mfraezz: Superclass is already doing this; this is unnecessary.

website/addons/owncloud/static/owncloudNodeConfig.js
```diff
@@ -0,0 +1,315 @@
...
+
+        self.credentialsChanged = ko.pureComputed(function() {
+            return self.nodeHasAuth() && !self.validCredentials();
+        });
+        self.showCredentialInput = ko.pureComputed(function() {
+            return (self.credentialsChanged() && self.userIsOwner()) ||
+                (!self.userHasAuth() && !self.nodeHasAuth() && self.loadedSettings());
+        });
+
+        self.showSettings = ko.pureComputed(function() {
```
mfraezz: This is also in a superclass.

website/addons/owncloud/static/owncloudNodeConfig.js
```diff
@@ -0,0 +1,315 @@
...
+
+        self.selectionChanged = function() {
+            self.changeMessage('','');
+        };
+    }
+});
+
+
+/** Reset all fields from OwnCloud host selection modal */
+ViewModel.prototype.clearModal = function() {
```
mfraezz: This (and other functions) should be specified in the `oop.extend` block.

mfraezz: It seems like you're doing more work in this class than you need to. Take a look at the updates to `s3NodeConfig.js`; it's doing almost exactly what you need, just with a different modal that accepts an API key rather than username and password.

website/addons/owncloud/static/owncloudNodeConfig.js
```diff
@@ -0,0 +1,315 @@
...
+        self.showCredentialCreateButton = ko.pureComputed(function() {
+            return !self.userHasAuth() && !self.nodeHasAuth() && self.loadedSettings();
+        });
+
+        // Flashed messages
+        self.message = ko.observable('');
+        self.messageClass = ko.observable('text-info');
+
+        // Update above observables with data from the server
+        $.ajax({
```
mfraezz: You should be overriding `updateFromData` rather than doing this here. All that actually needs to be done is overriding `applySettings`, but it seems to be less extensible than it should be, as that function is defined within `updateFromData`.

mfraezz: Actually, this might not need to be done at all -- I don't think it's doing anything ownCloud specific, so the superclass should handle it.

website/addons/owncloud/static/owncloudUserConfig.js
```diff
@@ -0,0 +1,197 @@
...
+        var self = this;
+        self.message('');
+        self.messageClass('text-info');
+        self.username(null);
+        self.password(null);
+        self.selectedHost(null);
+        self.customHost(null);
+    },
+
+    updateAccounts : function() {
```
mfraezz: I think this might override be superfluous. It looks like the only thing it's doing differently from the parent is assigning fields that are on the output from `<Addon>Serializer.serialize_account`... effectively re-mapping a field that gets passed to the front-end already (`providerUrl`)

website/addons/owncloud/tests/test_models.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def _node_settings_class_kwargs(self, node, user_settings):
+        return {
+            'user_settings': self.user_settings,
+            'folder': '/Documents',
+            'owner': self.node,
+            'node': self.node
+        }
+
+    def setUp(self):
```
mfraezz: I think this override might be superfluous -- the base class should already be calling `set_node_settings`

website/addons/owncloud/model.py
```diff
@@ -0,0 +1,153 @@
...
+    def create_waterbutler_log(self, auth, action, metadata):
+        url = self.owner.web_url_for('addon_view_or_download_file', path=metadata['path'], provider='owncloud')
+        self.owner.add_log(
+            'owncloud_{0}'.format(action),
+            auth=auth,
+            params={
+                'project': self.owner.parent_id,
+                'node': self.owner._id,
+                'folder': self.folder_id,
+                'filename': metadata['materialized'].strip('/'),
```
mfraezz: `path` is more consistent with other addons than `filename`, please consider renaming (and in any affected logs)

website/addons/owncloud/tests/test_models.py
```diff
@@ -0,0 +1,88 @@
...
+        super(TestNodeSettings, self).setUp()
+        self.set_node_settings(self.node_settings)
+
+    def test_create_log(self):
```
mfraezz: Please put overridden tests beneath an `## Overrides` comment, with any non-overriden but newly-defined tests above.

Also this override appears superfluous -- the only difference is that `path` has been renamed to `filename`, which should be changed

website/addons/owncloud/tests/test_models.py
```diff
@@ -0,0 +1,88 @@
...
+        assert_equal(self.node_settings.folder_id, '/')
+        # Log was saved
+        last_log = self.node.logs[-1]
+        assert_equal(last_log.action, '{0}_folder_selected'.format(self.short_name))
+
+    def test_serialize_credentials(self):
+        credentials = self.node_settings.serialize_waterbutler_credentials()
+
+        assert_is_not_none(self.node_settings.external_account.oauth_secret)
+        expected = {'host': self.node_settings.external_account.oauth_secret,
```
mfraezz: Please be consistent with indentation on iterable definition.

website/addons/owncloud/tests/test_serializer.py
```diff
@@ -0,0 +1,63 @@
...
+from website.addons.owncloud.tests.utils import (
+    create_mock_owncloud, OwnCloudAddonTestCase
+)
+from website.addons.owncloud.serializer import OwnCloudSerializer
+from website.addons.owncloud.model import OwnCloudProvider
+from website.addons.owncloud.tests.factories import OwnCloudAccountFactory
+from website.util import api_url_for
+
+
+class TestOwnCloudSerializer(OwnCloudAddonTestCase, OAuthAddonSerializerTestSuiteMixin):
```
mfraezz: Why not inherit from `StorageAddonSerializerTestSuiteMixin`?

website/addons/owncloud/tests/test_serializer.py
```diff
@@ -0,0 +1,63 @@
...
+    Serializer = OwnCloudSerializer
+    ExternalAccountFactory = OwnCloudAccountFactory
+    client =OwnCloudProvider
+
+    required_settings = ('userIsOwner', 'nodeHasAuth', 'urls', 'userHasAuth')
+    required_settings_authorized = ('ownerName', )
+
+    def setUp(self):
+        super(TestOwnCloudSerializer, self).setUp()
+        self.ser = self.Serializer(
```
mfraezz: The `super` call should already be defining `self.ser` to be this.

website/addons/owncloud/tests/test_serializer.py
```diff
@@ -0,0 +1,63 @@
...
+        provider = OwnCloudProvider(self.node_settings.external_account)
+        self.mock_api = mock.patch('website.addons.owncloud')
+        self.mock_api.return_value = create_mock_owncloud(provider.host, provider.username, provider.password)
+        self.mock_api.start()
+
+    def tearDown(self):
+        self.mock_api.stop()
+        super(TestOwnCloudSerializer, self).tearDown()
+
+    def set_node_settings(self, user):
```
mfraezz: This doesn't appear to be doing anything differently from the mixin's implementation, consider removing

website/addons/owncloud/serializer.py
```diff
@@ -0,0 +1,68 @@
...
+from website.util import web_url_for
+
+from owncloud import Client as OwnCloudClient
+
+class OwnCloudSerializer(StorageAddonSerializer):
+
+    addon_short_name = 'owncloud'
+
+    # Include host information with more informative labels / formatting
+    def serialize_account(self, external_account):
```
mfraezz: I'm not sure if this override is necessary --  it breaks some architectural generalizations without providing any more benefit than a docstring would. Consider adding a docstring noting quirks instead.

website/addons/owncloud/serializer.py
```diff
@@ -0,0 +1,68 @@
...
+        ret.update({
+            'host': external_account.profile_url,
+            'host_url': 'https://{0}'.format(external_account.profile_url),
+        })
+
+        return ret
+
+    def serialized_folder(self, node_settings):
+        return {
+            'name': node_settings.folder_id,
```
mfraezz: This should perhaps be calling `node_settings.fetch_folder_name`

website/addons/owncloud/tests/test_serializer.py
```diff
@@ -0,0 +1,63 @@
...
+        self.node = self.project
+        self.node_settings = self.node.get_or_add_addon(self.addon_short_name, auth=Auth(user))
+        self.node_settings.set_auth(self.user_settings.external_accounts[0], self.user)
+
+    def test_serialize_acccount(self):
```
mfraezz: >acccount

Thank you for bringing this typo to my attention. Also, this override *might* become unnecessary. As stated elsewhere, the `OwnCloudSerializer.serialize_account` override isn't really doing anything necessary.

website/addons/owncloud/tests/test_views.py
```diff
@@ -0,0 +1,97 @@
...
+
+from website.util import api_url_for
+from website.addons.base.testing import views
+from website.addons.owncloud.model import OwnCloudProvider
+from website.addons.owncloud.serializer import OwnCloudSerializer
+from website.addons.owncloud.tests.utils import (
+    create_mock_owncloud, OwnCloudAddonTestCase,create_external_account
+)
+
+class TestAuthViews(OwnCloudAddonTestCase):
```
mfraezz: Consider inheriting from `OAuthAddonAuthViewsTestCaseMixin`. You would need to override `test_oauth_start` and `test_oauth_finish` (this class should be testing account creation anyway), but would give you two free tests for account deactivation.

website/addons/owncloud/tests/test_views.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def test_deauthorize(self):
```
mfraezz: This test doesn't appear to be doing much of anything that the test of the same name in ownCloud's `TestNodeSettings` isn't doing, aside from sending it through flask. Consider removing

website/addons/owncloud/tests/test_views.py
```diff
@@ -0,0 +1,97 @@
...
+        self.set_node_settings(self.node_settings)
+
+    def tearDown(self):
+        self.mock_ser_api.stop()
+        super(TestConfigViews, self).tearDown()
+
+    @mock.patch('website.addons.owncloud.model.OwnCloudClient')
+    def test_folder_list(self, mock_connection):
+        #test_get_datasets
+        mock_connection.return_value = self.connection
```
mfraezz: You should just be able to call `super` after mocking this.

website/addons/owncloud/tests/test_views.py
```diff
@@ -0,0 +1,97 @@
...
+
+        url = api_url_for('owncloud_folder_list', pid=self.project._primary_key)
+        params = {'path':'/'}
+        res = self.app.get(url, params, auth=self.user.auth)
+        assert_equal(len(res.json), 2)
+        first = res.json[0]
+        assert_equal(first['path'], '/Documents')
+
+
+    def test_get_config(self):
```
mfraezz: I'm not sure why you need to override this, other than because the `super` version of this test fails.

mfraezz: Ah. `super` version fails because `credentials_are_valid` are mocked only on the first serialization call. This appears not to affect other addons because they don't have an `ExternalAccount` set on the node settings object when this is ran. `<Addon>AddonTestCase.set_node_settings` is intended to only set strings associated with a configuration -- the `OAuthEtCeteraMixin` classes are supposed to handle connecting `ExternalAccount` objects if necessary for the tests.

website/addons/owncloud/tests/utils.py
```diff
@@ -0,0 +1,61 @@
...
+    ExternalAccountFactory = OwnCloudAccountFactory
+    Provider = OwnCloudProvider
+    NodeSettingsFactory = OwnCloudNodeSettingsFactory
+    NodeSettingsClass = AddonOwnCloudNodeSettings
+    UserSettingsFactory = OwnCloudUserSettingsFactory
+
+    def set_node_settings(self, settings):
+        super(OwnCloudAddonTestCase, self).set_node_settings(settings)
+        settings.folder_id='/Documents'
+        settings.external_account = self.external_account
```
mfraezz: The other test classes should be handling this if necessary, please remove this line.

website/addons/owncloud/tests/test_views.py
mfraezz: You should just be able to mock `OwnCloudClient` with `self.connection`, as above, and this will work. Otherwise you'd get `UnmockedError`s

website/addons/owncloud/tests/utils.py
```diff
@@ -0,0 +1,61 @@
...
+    mock_owncloud.login.return_value = None
+    mock_owncloud.logout.return_value = None
+    mock_owncloud.list.return_value =[
+        FileInfo('/Documents', file_type='dir', attributes=None),
+        FileInfo('/Pictures', file_type='dir', attributes=None),
+        FileInfo('/secrets.txt', file_type='txt', attributes=None),
+    ]
+    return mock_owncloud
+
+def create_external_account(host='foo.bar.baz', username='doremi-abc-123',password='eh'):
```
mfraezz: This is only used in one place. Can it be replaced with a call to `OwnCloudAccountFactory`?

website/addons/owncloud/views.py
```diff
@@ -0,0 +1,130 @@
...
+owncloud_deauthorize_node = generic_views.deauthorize_node(
+    SHORT_NAME
+)
+
+owncloud_root_folder = generic_views.root_folder(
+    SHORT_NAME
+)
+
+@must_be_logged_in
+def owncloud_user_config_get(auth, **kwargs):
```
mfraezz: I'm not convinced this needs to exist. Dataverse does it this way, admittedly, but Dataverse hasn't been normalized very well yet. S3 has the same sort of non-standard auth mechanism and hijacked `ExternalAccount` fields and gets around it. Not necessarily a blocker, but this will need to be removed eventually.

website/addons/owncloud/views.py
```diff
@@ -0,0 +1,130 @@
...
+## Config ##
+
+@must_be_logged_in
+def owncloud_add_user_account(auth, **kwargs):
+    """Verifies new external account credentials and adds to user's list"""
+    user = auth.user
+
+    host = request.json.get('host').rstrip('/')
+    if 'https://' not in host:
+        host = 'https://' + host
```
mfraezz: This seems like it could end up making `host` something like `https://http://myawesome.owncloud.com`, etc. Consider using `furl` instead

website/addons/owncloud/views.py
```diff
@@ -0,0 +1,130 @@
...
+    password = request.json.get('password')
+    provider = OwnCloudProvider(account=None, host=host, username=username, password=password)
+
+    try:
+        oc = OwnCloudClient(host, verify_certs=settings.USE_SSL)
+        oc.login(username, password)
+        oc.logout()
+    except:
+        return {
+            'message': 'All the fields above are required.'
```
mfraezz: This message doesn't seem to fit with what could be happening here. That is, supplying all of the required fields but with any one of them incorrect would lead to a `400` telling you that you need to supply all the required fields. This check *should* be done, but probably above rather than here.

website/addons/owncloud/views.py
mfraezz: Additionally, you'd need to add `owncloud` to `api.base.settings.defaults.ADDONS_FOLDER_CONFIGURABLE` to enable API support. This will eventually become implicit after we normalize the remaining addons... This looks like it should work currently, and you'd also have to add test classes for ownCloud to `test_node_addons.py` and `test_user_addons.py`, so I stand by my previous statement that this isn't necessarily a blocker.

website/addons/owncloud/model.py
mfraezz: Yeah... Dataverse has some code rot. I plan on cleaning it up whenever I add APIv2 support for it, which is on my `TODO` list.

website/addons/owncloud/model.py
mfraezz: >So, the difference between the logs and the anonymous logs is done is the js layer?

Erm, not exactly. Anonymous logs are only shown when an "Anonymous View-Only Link` is used to authenticate and view a page (these are project-specific and generated on the Contributors page). They're intended for things like getting peer review anonymously, to prevent as much bias as possible.

It is, however, possible for a log that includes `${user}` in its JSON definition to be generated -- I think if another contributor removes the addon-authorizing contributor, one of these logs will be generated without `auth`, which would either cause `LogText.view.logText` to append `_no_user` to the log's `.action`, or cause `LogPieces.user` to evaluate to `A user` rather than the user's name (I don't recall exactly how it works).

```python
def deauthorize(self, auth=None, add_log=True):
    self.clear_settings()
    self.nodelogger.log(action='node_deauthorized')
    self.clear_auth()  # Also performs a .save()
```

mfraezz:
Pass complete, this is getting really close. Think I caught just about everything this time, but I'll give it another quick pass after the next iteration. :octocat: 

This also looks to be close enough to begin manual testing -- I'll start on that tomorrow or the next day, but feel free to update this at your pace. Back to you, @kwierman

website/addons/owncloud/model.py
kwierman: Alright, back on track with responses to the current round of updates. The provider certainly doesn't need a serializer, though I am still curious as to why the S3 provider needs one. In any case, in the current round, I'm experimenting with removing code blocks I added in to conform with the example plugins and this is one that it appears we can do without.

website/addons/owncloud/model.py
kwierman: Done and done.

Just out of curiosity, would adding `self.nodelogger.log(action='node_deauthorized')` to one of the base classes such as `StorageAddonBase.deauthorize` make sense? It seems like most of the addons use a similar behavior here.

website/addons/owncloud/model.py
kwierman: Cool. I'm applying this whenever I run into it, but if I miss a couple, let me know and I'll do another pass.

website/addons/owncloud/model.py
kwierman: Done and done. If you see anywhere else where this is appropriate let me know. In fact, if there's a general rule for when/if one should use kwargs over explicit parameters, I'd like to know so I go back and do a pass on this edit.  The literature on the subject is subjective to the context it's applied in, so you guys would know better than me.

website/addons/owncloud/model.py
kwierman: Ah, for a minute there, I thought you meant using kwargs via the above comment. In this case, I used the conversion from path to name. If that's incorrect, let me know.

website/addons/owncloud/static/owncloudNodeConfig.js
kwierman: I found out I did something stupid with inheritance which is why this wasn't working before. That being said, there's a couple of non-superclassed members here. It should be fixed now.

kwierman: Same goes for the user settings config.

website/addons/owncloud/static/owncloudNodeConfig.js
kwierman: Gotcha. You're right. The latest version should have most of this removed. On this latest edit, I think I've been fairly exhaustive  with checking against `OauthAddonFolderPickerViewModel` and `FolderPickerViewModel`. You might want to double check though. The observables and methods that correspond to the modal behavior should be all that's left. Same goes for the userconfig

website/addons/owncloud/static/owncloudNodeConfig.js
kwierman: Yup. Removed without change in behavior. Some of the fields ended up getting renamed, but that's about it.

website/addons/owncloud/static/owncloudUserConfig.js
kwierman: Yup. Same as above.

kwierman:
Hey guys, assuming the tests pass after I rebased to a single commit, I think I'm ready for the next phase of CR. I think I applied all of the changes you mentioned, so if there's something there that aught not to be, let me know. Back to you @mfraezz !

website/addons/owncloud/static/owncloudNodeConfig.js
```diff
@@ -0,0 +1,128 @@
...
+        self.selectedHost(null);
+        self.customHost(null);
+    },
+    connectAccount : function() {
+        var self = this;
+        if( !self.selectedHost() ){
+            self.changeMessage("Please select a OwnCloud repository.", 'text-danger');
+            return;
+        }
+        var url = self.urls().create;
```
mfraezz: `url` doesn't get defined here, probably because it's not being returned by the serializer:

screen shot 2016-09-13 at 10 39 02

kwierman: That's weird. Now, I'm curious as to why this worked on my local install. In any case, it's fixed. I'll consolidate the commits in a little bit.

website/addons/owncloud/static/owncloudNodeConfig.js
```diff
@@ -0,0 +1,128 @@
...
+        return $osf.postJSON(
+            url,
+            ko.toJS({
```
mfraezz: This isn't serializing username / password, but the view function that this is intended to hit expects them in the JSON body

website/addons/owncloud/views.py
```diff
@@ -0,0 +1,134 @@
...
+
+## Config ##
+
+@must_be_logged_in
+def owncloud_add_user_account(auth, **kwargs):
+    """Verifies new external account credentials and adds to user's list"""
+
+    # Ensure that ownCloud uses https
+    host = furl(request.json.get('host').rstrip('/'))
+    host.scheme = 'https'
```
mfraezz: I'm not yet certain why, but this seems to be causing some strange behavior:

```
    ipdb> l
         75     # Ensure that ownCloud uses https
         76     host = furl(request.json.get('host').strip('/'))
         77     host.scheme = 'https'
         78     import ipdb; ipdb.set_trace()
         79
    ---> 80     username = request.json.get('username')
         81     password = request.json.get('password')
         82
         83     try:
         84         oc = owncloud.Client(host.url, verify_certs=settings.USE_SSL)
         85         oc.login(username, password)
    
    ipdb> host.scheme
    'https'
    ipdb> host.url
    'https:///mycloud.<redacted>'
    ipdb> host.scheme = ''
    ipdb> host.url
    'mycloud.<redacted>'
```
>`'https:///mycloud.<redacted>'`

website/addons/owncloud/static/owncloudNodeConfig.js
kwierman: Sorry, this must have been overwritten when I was trying out the best way to remove super'd functions from this viewmodel. It should now have the correct function.

website/addons/owncloud/views.py
kwierman: Redacted? What happens to the original url a la `request.json.get('host')`?

mfraezz:
(Initially posted on the WaterButler PR because they both appear as `[CR] ownC...` in browser tabs)

There needs to be a v2 route for folder listing, otherwise you'll only be able to see the top level directories. If everything is wired up correctly, this might be as simple as adding `'owncloud'` to `api.settings.ADDONS_FOLDER_CONFIGURABLE`

screen shot 2016-09-13 at 11 11 48

website/addons/owncloud/views.py
mfraezz: It's the same as what gets entered in the UI

addons.json
```diff
@@ -23,6 +24,7 @@
         "forward": "full",
         "googledrive": "partial",
         "github": "partial",
+	    "owncloud": "partial",
```
mfraezz: Minor, but this is over-indented

addons.json
```diff
@@ -34,6 +36,7 @@
         "figshare",
         "github",
         "googledrive",
+	"owncloud",
```
mfraezz: This is under-indented and should be using spaces, not tabs

kwierman: Huh. These three items were fixed in the last commit before I rebased to a single commit. I'm getting the sneaking suspicion that I dropped a commit when I rebased....

website/addons/owncloud/model.py
```diff
@@ -0,0 +1,157 @@
...
+    def after_delete(self, node, user):
+        self.deauthorize(Auth(user=user), add_log=True)
+        self.save()
+
+    def on_delete(self):
+        self.deauthorize(add_log=False)
+        self.save()
+
+    def get_folders(self, **kwargs):
+        path = kwargs.get('path', '/')
```
mfraezz: This will probably need to be `kwargs.get('path') or '/'`. The API will include `path` regardless of whether it's `None`.

Also, this method should be serializing a "root folder", so that users can link their entire ownCloud, if they want

website/addons/owncloud/model.py
```diff
@@ -0,0 +1,157 @@
...
+    def folder_name(self):
+        return self.folder_id
+
+    def set_folder(self, folder, auth=None):
+        if folder == '/ (Full ownCloud)':
+            folder = '/'
+        self.folder_id = folder
+        self.save()
+        if auth:
+            self.nodelogger.log(action='folder_selected', save=True)
```
mfraezz: This should perhaps be called regardless of `auth`

mfraezz:
A quick run-through of the API endpoints for this seems to indicate that it's working. Please create generic test classes in `api_tests/nodes/test_node_addons.py` and `users/test_user_addons.py`. This should be behaving similarly to Box and Dropbox, so feel free to template from those

website/addons/owncloud/model.py
kwierman: Ah, I see. Ok, I guess my question here, is what would be the best way to implement this in terms of the path? In it's current iteration a GET with `/` as an argument lists the top level directories. Would it be possible to use a null path `` to  GET a list of just the root level directory?

kwierman:
Will do!

website/addons/owncloud/views.py
mfraezz: Some elaboration on this: it was a url of the form `subdomain.domain.tld`. `furl` seemed, for whatever reason, to this that this was a path when it was instantiated, rather than a host. The following snippet worked:
```
host_url = request.json.get('host').strip('/')
host = furl()
host.host = host_url
host.scheme = 'https'
```

website/addons/owncloud/model.py
mfraezz: Yes. If this is called without kwargs (or with the root path `/`), the default behavior should be to return the root directory

website/addons/owncloud/views.py
kwierman: Huh. OK. [This makes sense to me now.](https://github.com/gruns/furl/blob/b0af00f6b5d22bf5c23f7a823ad8769fd4bf915b/furl/furl.py#L1175). Expect to see this in the next commit.

kwierman: haha, alright so fun fact; this fix breaks normal operations due to bypassing the tokenizer. However, I have an idea. If we strip off the schema beforehand then this functions. It's not pretty... but it functions

kwierman: Yea, with a little thorough testing this works nicely.

website/addons/owncloud/views.py
kwierman: Alright, with a little finangling, this is now gone. You're right, most of this was being pushed through the accounts_list endpoint. Looking at Dataverse, this same tactic could be used.

website/addons/owncloud/views.py
```diff
@@ -0,0 +1,114 @@
...
+## Config ##
+
+@must_be_logged_in
+def owncloud_add_user_account(auth, **kwargs):
+    """Verifies new external account credentials and adds to user's list"""
+
+    # Ensure that ownCloud uses https
+    host_url = request.json.get('host')
+    host = furl()
+    host.host = host_url.rstrip('/').strip('https://').strip('http://')
```
mfraezz: If it works (and a bit of local testing indicates that it does). ¯\\\_(ツ)_/¯ 

Could you add a test or two, passing poorly formatted urls in the json, for this view function in `owncloud/tests/test_views.py:TestAuthViews`, to prevent any regressions? Some amount of frontend validation / helper-text to indicate proper formatting to the user would be nice as well, but isn't necessarily a blocker.

felliott:
Sorry if this has been addressed already: I added a second owncloud account to my local testing.  When I went to link my account the dropdown box only showed my usernames, not the servers they were on:

screen shot 2016-09-15 at 10 48 28 am

That box should probably show both host and username.

kwierman:
Cool cool. That's a quick fix.

kwierman:
Oh wow, thanks! I'm still a little concerned with the webDAV bug from earlier. I spun up a secure (non-self signed) OC server at kjwcloud.com this morning to do some more thorough testing, so I might come back with some additional changes

mfraezz:
I think figured that part out -- after using `litmus` to verify that WebDAV was working on my server, I realized that the `ExternalAccount` in my db was still using `http` from when I'd been testing without SSL. This was causing a redirect to a secure version of the landing page, rather than the WebDAV route. After deleting that `ExternalAccount` and going through the connection flow again, it worked perfectly.

kwierman:
Awesome! I was racking my brain against this one trying to figure it out.

PR #6079, opened 2016-08-01
[Hotfix] List more than 10 Folders through API [OSF-6830]

mfraezz:
## Purpose
List more than 10 folders for addons when using APIv2

## Changes
* Page size set to 1k
* Opportunistic Fix: `InstitutionsList` page size set to 1k:

>I do see us going over 100 institutions but we will likely remain under 1k
\- @icereval 

## Side effects
This isn't pagination. Because [where this is being called in TreeBeard](https://github.com/CenterForOpenScience/treebeard/blob/develop/dist/treebeard.js#L1064) doesn't resolve promises, this flow is difficult to hook into.

Consider adding a JSONAPI-compliant pagination feature to TreeBeard.

## Ticket
[OSF-6830](https://openscience.atlassian.net/browse/OSF-6830)

### DIFF
--- a/api/base/pagination.py
+++ b/api/base/pagination.py
@@ -128,6 +128,11 @@ class JSONAPIPagination(pagination.PageNumberPagination):
             return super(JSONAPIPagination, self).paginate_queryset(queryset, request, view=None)
 
 
+class MaxSizePagination(JSONAPIPagination):
+    page_size = 1000
+    max_page_size = None
+    page_size_query_param = None
+
 class CommentPagination(JSONAPIPagination):
 
     def get_paginated_response(self, data):
--- a/api_tests/base/test_pagination.py
+++ b/api_tests/base/test_pagination.py
@@ -7,4 +7,5 @@ from api.base.pagination import MaxSizePagination
 
 class TestMaxPagination(ApiTestCase):
     def test_no_query_param_alters_page_size(self):
-        assert_is_none(MaxSizePagination.page_size_query_param)
+        assert MaxSizePagination.page_size_query_param is None, 'Adding variable page sizes to the paginator ' +\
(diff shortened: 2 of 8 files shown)


api_tests/base/test_pagination.py
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
+from nose.tools import *  # flake8: noqa
+
+from tests.base import ApiTestCase
+
+from api.base.pagination import MaxSizePagination
+
+class TestMaxPagination(ApiTestCase):
+    def test_no_query_param_alters_page_size(self):
+        assert_is_none(MaxSizePagination.page_size_query_param)
```
brianjgeiger: Put an error message on this to say, "Adding variable page sizes to the paginator requires tests to ensure that you can't request more than the class's maximum number of values." So if they break it, they know what they need to do to fix it, and if it's removed in a PR, that the code reviewer will have a chance to see what it should have been replaced by.

brianjgeiger: Or you can add in support for the page size and make sure that it works properly. Either way.

brianjgeiger:
Looks good. 🎉 🐧 🎉

api/base/pagination.py
```diff
@@ -128,6 +128,11 @@ def paginate_queryset(self, queryset, request, view=None):
             return super(JSONAPIPagination, self).paginate_queryset(queryset, request, view=None)
 
 
+class MaxSizePagination(JSONAPIPagination):
+    page_size = 1000
```
icereval: Can we not have just made the page_size get set to the length of the queryset results (or equivalent)? Thus not much of a max but a full results set on a single page...

PR #6081, opened 2016-08-01
[#OSF-6420]now delete folder will dismiss the delete alert correctly

chennan47:
## Purpose
Dismiss the delete alert modal after hit delete for github folder and file deletion.

## Ticket


brianjgeiger:
🎉 🐧 🎉

PR #6083, opened 2016-08-02
[feature] external ember app [#PREP-87]

icereval:
external ember app & configuration to support local development via simple reverse proxy

h/t @abought

### DIFF
--- a/website/routes.py
+++ b/website/routes.py
@@ -174,11 +174,11 @@ def robots():
     )
 
 
-def ember_app(path=None):
+def external_ember_app(path=None):
     """Serve the contents of the ember application"""
     if (request.path == '/ember-cli-live-reload.js'):
         path = 'ember-cli-live-reload.js'
-    url = furl(settings.EMBER_APP_URL).add(path=path)
+    url = furl(settings.EXTERNAL_EMBER_URL).add(path=path)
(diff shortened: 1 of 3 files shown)


abought:
Should this be targeted at the `feature/osf-ember` branch instead?

## Documenting changes required for the companion ember app
1. Check out this OSF branch
2. In website/settings/local.py, specify `EXTERNAL_EMBER_BASEURL = '/preprints/'`
3. Start your ember server (on port 4200).
4. Go to `localhost:5000/preprints/' to see the preprints page.

## Evaluation checklist
1.  [x] Static assets should be served correctly (the proxy handles image, css, and js requests correctly)
2. [x] The app responds to the nested URL (eg /preprints/ ) with the top level route for the application.
3. [ ] Doesn't use any deprecated settings (eg `baseURL`)

## Other notes
Proxying to a static server makes sense for ember preprints, which has a requirement to stay as separate as possible. But unlike serving the contents of a static build directory, this method means the ember server *must* be running to view those pages, even if they are not being actively developed.

This could be inconvenient for regular OSF development, for example- I'm not sure the proxy solution is what we want to use for everything. For example, the current OSF build process (eg webpack) lets us use prebuilt static assets, which is one less thing to keep running.

Issue #5270, opened 2016-03-22
Profile - long urls cause problems in display


AndrewSallans:
![screen shot 2016-03-22 at 3 34 25 pm](https://cloud.githubusercontent.com/assets/2151933/13965400/687b16ee-f044-11e5-9247-7b054f39a9b0.png)

What you did (step by step)
Looked at user's profile: https://osf.io/dhyac/

What you expected
Expected to see short links next to social profile categories.
![screen shot 2016-03-22 at 3 41 06 pm](https://cloud.githubusercontent.com/assets/2151933/13965419/86439110-f044-11e5-95fe-7fa5fc51cba5.png)

Potential causes
Not sure. Not seeing any Javascript errors in console.

Suggest a solution
Not sure.

sambrady3:
From what I can see there are two different problems here. One is that the Google Scholar parser in profile.js wasn't able to properly parse this profile's Google Scholar link, which is why when you click on it you are not taken to the correct page. This led to the super long link, rather than the expected shortened link.

The other problem is what to do with long personal website urls, which will still cause the undesirable word-wrapping of the social labels.

GaryKriebel:
@AndrewSallans, is this still an issue?

AndrewSallans:
Yes

PR #6086, opened 2016-08-02
Profile: long Personal Website URLs run over in cause problems in display #5270 [OSF-6722]

amyshi188:
## Purpose

Having a long URL for personal websites causes the rest of the items under "Social" to be misaligned. 

## Changes

Changed the width of each column in the table so that the first column with the labels can all fit on one line. 

## Side effects

N/A

## Ticket
https://openscience.atlassian.net/browse/OSF-6722


brianjgeiger:
Looks good. If product likes it, I'm happy. 🎉 🐧 🎉

PR #6087, opened 2016-08-02
Allow bulk DELETE with query parameters [OSF-6732]

brianjgeiger:
## Purpose

The Google Cloud / Kubernetes load balancer does not allow DELETE requests with JSON bodies, and the initial spec for JSON API Bulk DELETEs has just this combination. So we have a new implementation of Bulk DELETEs that will use query parameters to perform the delete rather than JSON body. The JSON body is still acceptable for now, because it's used in places, but that will probably be phased out before we move all the servers over to the new load balancer.

A request would look like:
DELETE `https://api.osf.io/v2/nodes/?type=nodes&id=abcde,fgh4j,k8mn2`

You cannot use both the JSON body and the query parameters. if the id field is in the query parameters and there is a JSON body, the request will fail with a good error. If only the type field is in the query parameters, it will ignore the type field and will work with the JSON body.

There is a set of runscope test on the OSF Status Board that should be updated for this new method of deleting. When this makes it onto staging2, then the test that has been failing is updated and runs, it should all work again.

## Changes

1. Update bulk delete code to allow use of query parameters
2. Update node contributors bulk delete code to work with refactored code
3. Update one error message to be more useful
4. Tests

## Side effects

If any bulk requests were overridden but not tested, those requests will probably fail. 

## Ticket

https://openscience.atlassian.net/browse/OSF-6732

### DIFF
--- a/api/base/generic_bulk_views.py
+++ b/api/base/generic_bulk_views.py
@@ -71,19 +71,19 @@ class BulkDestroyJSONAPIView(bulk_generics.BulkDestroyAPIView):
     Custom BulkDestroyAPIView that handles validation and permissions for
     bulk delete
     """
-    def get_requested_resources(self, request):
+    def get_requested_resources(self, request, request_data):
         """
         Retrieves resources in request body
         """
         model_cls = request.parser_context['view'].model_class
-        requested_ids = [data['id'] for data in request.data]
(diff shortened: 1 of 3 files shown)


mfraezz:
Pass complete, LGTM.

Dropping this in PCR :shipit:

PR #6088, opened 2016-08-02
Stop side navbar on account settings scrolls behind main text of page [OSF-5125]

Johnetordoff:
## Purpose

Currently on account setting the side navbar won't be properly responsive and will slide under the panels on that page.

## Changes

Simple change to media query.

## Side effects

None that I know of

## Ticket

https://openscience.atlassian.net/browse/OSF-5125

### DIFF
--- a/website/static/css/style.css
+++ b/website/static/css/style.css
@@ -265,7 +265,7 @@ a {
 
 }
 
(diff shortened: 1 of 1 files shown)


brianjgeiger:
🎉 🐧 🎉

PR #6089, opened 2016-08-02
Make flag for non-meetings Meetings in Admin app [#OSF-5645]

caspinelli:
## Purpose

Adds functionality where it is possible to specify in the admin app when creating a meeting that the meeting should not show up on the meetings page. 
Also removes 2 already existing non-meetings meetings from the view on the meetings page

## Changes

Added input in the create meeting form in the admin app that you deselect if the meeting is actually not a meeting

## Side effects

When editing and older meeting that does not have the attribute is_meeting because if has been added in this change) the check-box for 'Is A Meeting' defaults to unchecked and if not checked before submitting, the meeting will not display on the meetings page. 

## Ticket

https://openscience.atlassian.net/browse/OSF-5645

### DIFF
--- a/admin/meetings/forms.py
+++ b/admin/meetings/forms.py
@@ -61,6 +61,11 @@ class MeetingForm(forms.Form):
         required=False,
         widget=forms.TextInput(attrs={'size': '60'}),
     )
+    is_meeting = forms.BooleanField(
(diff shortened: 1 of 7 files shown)


brianjgeiger:
Looks good. 🎉 🐧 🎉

PR #6090, opened 2016-08-02
Feature/Fix premature check if contributor is admin [OSF-6848]

pattisdr:
## Purpose

Fixes node model update_contributor bug.  If you are the only admin, a PATCH request only changing your bibliographic info would fail with a 400 - saying you were the only admin.  Cause was a premature check to see whether or not only admin was removing herself as an admin.

## Changes

 Moves check of "if admin == user and ADMIN not in permissions" to inside the conditional that checks if permissions have been changed.

## Side effects

Does this break anything? This is a node model change.  Probably was only visible from the APIv2, though. QA: Check for any weirdness with contributor modification, especially around admin permissions.

## Ticket

https://openscience.atlassian.net/browse/OSF-6848


brianjgeiger:
Looks good. 🎉 🐧 🎉

pattisdr:
Thanks for the quick review!

PR #6092, opened 2016-08-03
Draft registrations should be in reverse chronological order [#OSF-6243]

caspinelli:
## Purpose

Sorted draft registrations in reverse chronological order by last updated date.

## Changes

Added code to sort draft registrations by last updated date in API and removed code that was sorting draft registrations by created date on frontend

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-6243

### DIFF
--- a/website/project/views/drafts.py
+++ b/website/project/views/drafts.py
@@ -3,6 +3,8 @@ import httplib as http
 import datetime
 import itertools
 
+from operator import itemgetter
+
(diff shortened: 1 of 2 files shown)


brianjgeiger:
Ah, okay, this is an api/v1 view. Yes, this method should be fine for v1. 🎉 🐧 🎉

PR #6094, opened 2016-08-03
On Create Component modal, Add language to indicate Admins have read access [OSF-6826]

amyshi188:
## Purpose

Based on feedback, it seems that many OSF users do not know that admins of parent projects have read access to children projects. The purpose of this fix is to clarify that by updating the language in the Create Component modal.

## Changes

Used mithril to add italicized and bold text to Create Component modal in addProjectPlugin.js file.

image

## Side effects

N/A

## Ticket

https://openscience.atlassian.net/browse/OSF-6826

### DIFF
--- a/website/static/js/addProjectPlugin.js
+++ b/website/static/js/addProjectPlugin.js
@@ -230,7 +230,9 @@ var AddProject = {
                                     onchange : function() {
                                         ctrl.newProjectInheritContribs(this.checked);
                                     }
-                                }), ' Add contributors from ', m('b', options.parentTitle)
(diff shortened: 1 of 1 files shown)


brianjgeiger:
JS Test failing; could be related to this issue. `invoke karma` to see if it happens locally. 🐧 Pass done

amyshi188:
@brianjgeiger I just did `invoke karma` a second time on develop and the "creates new identifiers" test failed so I don't believe it is related to this ticket.

brianjgeiger:
This is good once tests pass. The one JS test seems to be intermittent. 🎉 🐧 🎉

PR #6095, opened 2016-08-04
Allow website containing unicode to be added to OSF social profile [#OSF-6831]

caspinelli:
## Purpose

Allow user to add a website to 'Your Websites' that contains unicode in either the domain or path of the website

## Changes

Added to regex that checks validity of url entered by user to allow unicode in certain spots
This changes frontend behavior. The error message 'Please enter a valid website' now doesn't show up due to unicode in the url. However, there is still an error when the user hits save and the info is sent to the server. There is another url validator there that is changed in the PR on this ticket (https://openscience.atlassian.net/browse/OSF-4842) and is a blocker for this bug to be fully fixed. The frontend validator changed in this ticket has been updated to match exactly the changes made to the backend validator in the PR mentioned above so they will both match on the same things and fail on the same things. 

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-6831


tests/test_models.py
```diff
@@ -91,35 +92,30 @@ def test_validate_social_profile_websites_empty(self):
...
-        self.user.save()
-        assert_equal(self.user.social['profileWebsites'], ['https://definitelyawebsite.com'])
-
-    def test_validate_social_valid_website_ipv4(self):
-        self.user.social = {'profileWebsites': ['http://127.0.0.1']}
-        self.user.save()
-        assert_equal(self.user.social['profileWebsites'], ['http://127.0.0.1'])
+    def test_validate_social_profile_website_many_different(self):
+        with open('../website/static/urlValidatorTest.json') as urlTestData:
+            data = json.load(urlTestData)
```
brianjgeiger: This doesn't look like it's actually doing anything other than loading test data. There are no assertions or anything.

tests/test_models.py
```diff
@@ -91,35 +92,30 @@ def test_validate_social_profile_websites_empty(self):
...
-    def test_validate_social_valid_website_portandpath(self):
-        self.user.social = {'profileWebsites': ['http://127.0.0.1:5000/hello/']}
-        self.user.save()
-        assert_equal(self.user.social['profileWebsites'], ['http://127.0.0.1:5000/hello/'])
-
-    def test_validate_social_valid_website_querystrings(self):
-        self.user.social = {'profileWebsites': ['http://definitelyawebsite.com?real=yes&page=definitely']}
-        self.user.save()
-        assert_equal(self.user.social['profileWebsites'], ['http://definitelyawebsite.com?real=yes&page=definitely'])
+        failsAtEnd = False
```
brianjgeiger: Python variables should be snake_case

brianjgeiger:
This is fairly close, but you won't finish today. I'll take it over or have someone take it over and finish it by next week's IR.

PR #6099, opened 2016-08-04
Feature/make swagger json links browsable [OSF-6163]

HalcyonChimera:
## Purpose

Make any links in JSON responses in Swagger docks clickable.
Causes docs to open/scroll to the pertinent response.

## Changes

Modify swagger-ui.js to make links in JSON clickable/browsable
Upgrade Swagger spec to 2.0 -- OpenAPI compatibility
CoreAPI compatibility
Upgrade DRF to 3.4
DRS to 2.0.4
update some OSF files to use unicode strings by default

## Side effects

Swagger docs render two separate endpoints as if they were one;
node-file-list and node-file-detail as well as registration-file-list and registration-file-detail appear as though they were a single endpoint. This is due to the fact that the way django renders uritemplates from router regexes cannot account for certain aspects of regular expressions. Both file-list and file-detail are operable from the operation 'try it out' panel in the swagger docs.

## Ticket

https://openscience.atlassian.net/browse/OSF-6163


api/base/urls.py
```diff
@@ -19,8 +20,16 @@
                 url(r'^applications/', include('api.applications.urls', namespace='applications')),
                 url(r'^collections/', include('api.collections.urls', namespace='collections')),
                 url(r'^comments/', include('api.comments.urls', namespace='comments')),
-                url(r'^docs/', include('rest_framework_swagger.urls')),
+                url(r'^nodes/', include('api.nodes.urls', namespace='nodes')),
```
brianjgeiger: See L12 and please keep this list alphabetized (unless the new version of DRS auto-alphabetizes regardless if this list order, in which remove L12).

HalcyonChimera: DRS does alphabetize on the actual API -> Core api transition, though something weird happens between there and rendering on the documentation page.

HalcyonChimera:
Not sure what I'm missing. All tests run fine locally...

brianjgeiger:
```
FAIL: test_get_absolute_url (tests.test_models.TestNode)
  File "/home/travis/build/CenterForOpenScience/osf.io/tests/test_models.py", line 1625, in test_get_absolute_url
    .format(settings.API_DOMAIN, self.node._id)
AssertionError: 'http://localhost:8000/nodes/v5q9y/' != 'http://localhost:8000/v2/nodes/v5q9y/'
```

Did you change anything in there that you shouldn't have? Such as `settings.API_DOMAIN`? And are you running the api tests?

HalcyonChimera:
I do have API_DOMAIN set to something different, only in local.py defaults is set to localhost:8000, which seems correct.

Huh. api settings is different.

brianjgeiger:
Based on the test failures, I'm guessing that the new version of DRF is causing problems. Did you `invoke test_api` or otherwise run the api tests?

HalcyonChimera:
Tests all pass locally! I'm not sure whats going on here. I have something I'm going to try though.

api/base/serializers.py
```diff
@@ -509,7 +509,10 @@ def get_url(self, obj, view_name, request, format):
                 else:
                     if callable(view):
                         view = view(getattr(obj, self.field_name))
-                    url = self.reverse(view, kwargs=kwargs, request=request, format=format)
+                    try:
+                        url = self.reverse(view, kwargs=kwargs, request=request, format=format)
+                    except:
```
mfraezz: This isn't doing anything.

api/nodes/urls.py
```diff
@@ -1,8 +1,13 @@
+from __future__ import unicode_literals
 from django.conf.urls import url
 
 from api.nodes import views
 from website import settings
 
+#nodeProviderDetailView = views.NodeProviderDetailView.as_view()
```
brianjgeiger: No commented out lines, please.

api/nodes/urls.py
```diff
@@ -20,8 +25,11 @@
     url(r'^(?P<node_id>\w+)/draft_registrations/(?P<draft_id>\w+)/$', views.NodeDraftRegistrationDetail.as_view(), name=views.NodeDraftRegistrationDetail.view_name),
     url(r'^(?P<node_id>\w+)/files/$', views.NodeProvidersList.as_view(), name=views.NodeProvidersList.view_name),
     url(r'^(?P<node_id>\w+)/files/providers/(?P<provider>\w+)/?$', views.NodeProviderDetail.as_view(), name=views.NodeProviderDetail.view_name),
-    url(r'^(?P<node_id>\w+)/files/(?P<provider>\w+)(?P<path>/(?:.*/)?)$', views.NodeFilesList.as_view(), name=views.NodeFilesList.view_name),
     url(r'^(?P<node_id>\w+)/files/(?P<provider>\w+)(?P<path>/.+[^/])$', views.NodeFileDetail.as_view(), name=views.NodeFileDetail.view_name),
+    url(r'^(?P<node_id>\w+)/files/(?P<provider>\w+)(?P<path>/.*)$', views.NodeFilesList.as_view(), name=views.NodeFilesList.view_name),
+    #url(r'^(?P<node_id>\w+)/files/(?P<provider>\w+)/?(?P<path>/.+[^/])$', views.NodeFileDetail.as_view(), name=views.NodeFileDetail.view_name),
```
brianjgeiger: Delete these

api/nodes/views.py
```diff
@@ -130,34 +131,25 @@ def get_draft(self, draft_id=None):
 
 class WaterButlerMixin(object):
 
-    path_lookup_url_kwarg = 'path'
```
brianjgeiger: You should be careful about refactoring things like this, especially on a PR that is as large as this one. This is the kind of thing that leads to merge conflicts, so the potential for trouble is a lot greater than the gain.

brianjgeiger:
What happens when the next version of django-rest-swagger or the swagger-ui comes out?

api_tests/base/test_root.py
```diff
@@ -94,12 +95,12 @@ def test_non_admin_scoped_token_does_not_have_admin(self, mock_auth):
...
             attributes={
-                'accessToken': token.token_id,
-                'accessTokenScope': [s for s in token.scopes.split(' ')]
+                'accessToken': str(token.token_id),
+                'accessTokenScope': [str(s) for s in token.scopes.split(' ')]
             }
         )
         mock_auth.return_value = self.user, mock_cas_resp
-        res = self.app.get(self.url, headers={'Authorization': 'Bearer {}'.format(token.token_id)})
+        res = self.app.get(self.url, headers={'Authorization': str('Bearer {}'.format(token.token_id))})
```
brianjgeiger: What's the chain of event that lead to these changes? The need for unicode_literals? The new version of DRF?

api_tests/files/views/test_file_detail.py
```diff
@@ -23,8 +23,7 @@
 def _dt_to_iso8601(value):
     iso8601 = value.isoformat()
     if iso8601.endswith('+00:00'):
-        iso8601 = iso8601[:-9] + 'Z'  # offset upped to 9 to get rid of 3 ms decimal points
-
+        iso8601 = iso8601[:-6] + 'Z'  # offset upped to 9 to get rid of 3 ms decimal points
```
brianjgeiger: This line is now contradicting the comment.

api_tests/nodes/views/test_node_links_detail.py
```diff
@@ -58,7 +59,7 @@ def test_returns_private_node_pointer_detail_logged_out(self):
         assert_equal(res.status_code, 200)
         target_node = res.json['data']['embeds']['target_node']
         assert_in('errors', target_node)
-        assert_equal(target_node['errors'][0]['detail'], 'Authentication credentials were not provided.')
+        assert_equal(target_node['errors'][0]['detail'], 'You do not have permission to perform this action.')
```
brianjgeiger: This should still be the first error because auth credentials were not provided.

api_tests/nodes/views/test_node_comments_list.py
```diff
@@ -59,9 +60,9 @@ def test_return_private_node_comments_logged_out_user(self):
 
     def test_return_private_node_comments_logged_in_non_contributor(self):
         self._set_up_private_project_with_comment()
-        res = self.app.get(self.private_url, auth=self.non_contributor, expect_errors=True)
-        assert_equal(res.status_code, 401)
```
brianjgeiger: Yeah, this test should never have worked.

website/static/js/_anonymousLogTexts.json
```diff
@@ -0,0 +1 @@
...file_added":"A user added a file to an Amazon S3 bucket in a project","s3_file_removed":"A user removed a file in an Amazon S3 bucket in a project","s3_file_updated":"A user updated a file in an Amazon S3 bucket in a project","s3_folder_created":"A user created a folder in an Amazon S3 in a project","s3_node_authorized":"A user authorized the Amazon S3 addon for a project","s3_node_deauthorized":"A user deauthorized the Amazon S3 addon for a project","s3_node_deauthorized_no_user":"Amazon S3 addon for a project deauthorized","googledrive_file_added":"A user added a file to Google Drive in a project","googledrive_file_removed":"A user removed a file from Google Drive in a project","googledrive_file_updated":"A user updated a file in Google Drive in a project","googledrive_folder_created":"A user created a folder in Google Drive in a project","googledrive_folder_selected":"A user linked a Google Drive folder to a project","googledrive_node_authorized":"A user authorized the Google Drive addon for a project","googledrive_node_deauthorized":"A user deauthorized the Google Drive addon for a project","googledrive_node_deauthorized_no_user":"Google Drive addon for a project deauthorized"}
```
brianjgeiger: This seems like it's a built file. I don't think it's used any more.

brianjgeiger:
Pass done. 🐧 Most stuff in this pass is simple, except for the one big question.

api_tests/base/test_root.py
HalcyonChimera: The new version of DRF requires unicode rather than bytestrings, so it was easier to change strings to unicode strings with the import unicode literals. A few things need to be explicitly bytestrings, like the (context?) dicts that are created from the headers here.

api_tests/nodes/views/test_node_links_detail.py
HalcyonChimera: This is weird. I'll look at this some more.

api/nodes/urls.py
HalcyonChimera: Not certain I remember what I was trying here...

api/base/urls.py
HalcyonChimera: Sorted out the alphabetization thing.

HalcyonChimera:
Regarding what happens when DRS and swagger-ui get updated; the code in swagger-ui just need to gets pasted into the right spot. the code inserted is contiguous, and should be able to be placed int one location. as long as too much doesn't change all should be fine. The codebase here (barring the swagger-ui js) is relatively small, so it's fairly straightforward to understand what they may have changed. the big difference in the new version of DRS was updating the version of the spec to openapi from swagger.

api_tests/nodes/views/test_node_links_detail.py
HalcyonChimera: I'd like to make this a separate PR; It seems that this was always the case that the wrong response was being provided here. The string originates from a 403 response that should be a 401.

api_tests/files/views/test_file_detail.py
HalcyonChimera: I've updated the comment and kept the decimal places. This was a change introduce in the new version of DRF. The rationale is that JSON is not a strict subset of ECMAScript which stipulates that dates look this way.

brianjgeiger:
Part of why I ask about the swagger-ui is because:

screen shot 2016-09-01 at 4 34 14 pm

So if you're adding the whole file *plus* the patch in one go, it's very difficult to see what's happening. When we talked at the beginning of the summer, you described this as a monkey-patch, which is generally a runtime patch rather than the whole file. If you haven't put your code directly into swagger-ui, then ideally it should be being brought in by another mechanism that doesn't involve a file put directly in `/static/js/`, and then you can patch that as a smaller diff? 

If we have to maintain a modification to a static file, then we'll want to fork the repo and then make changes there, then import it with one build tool or another.

And if we have to be super careful about updating DRS / swagger-ui, it would be good to include a note next to where someone might do that update describing what would need to happen during that upgrade.

HalcyonChimera:
At the beginning of the summer this was a monkey patch, but this new version isn't as nice to play around with. I can fork swagger-ui.

brianjgeiger:

PR #6102, opened 2016-08-04
[RFR][Feature][OSF-6780]Raise minimum password length to 8 characters

TomBaxter:
## Purpose:
[OSF-6780](https://openscience.atlassian.net/browse/OSF-6780)
Change minimum password length from 6 to 8

## Changes:
Change 6 to 8 for all of the following
Update framework/auth/core.py
Update framework/auth/forms.py
Update website/static/js/passwordForms.js
Update website/templates/landing.mako

## Side effects
None

[#OSF-6780]

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -1763,7 +1763,7 @@ class TestUserAccount(OsfTestCase):
             old_password='password',
             new_password='12345',
             confirm_password='12345',
-            error_message='Password should be at least six characters',
+            error_message='Password should be at least eight characters',
         )
 
     def test_password_change_invalid_blank_password(self, old_password='', new_password='', confirm_password=''):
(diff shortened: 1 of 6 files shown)


tests/test_views.py
```diff
@@ -1763,7 +1763,7 @@ def test_password_change_invalid_new_password_length(self):
             old_password='password',
             new_password='12345',
             confirm_password='12345',
```
brianjgeiger: Could you also change this so that it's '1234567' so it tests the new edge case? And is there already another test that will try out an 8 character password to verify that that's okay? Might as well get both sides of the edge.

brianjgeiger:
A little bit more testing should do it. Have you done manual testing on all the front-end places to verify that it seems to work there as well? Pass done. 🐧

tests/test_views.py
TomBaxter: There is a test for valid password change that coincidentally has 8 character password. I added one specifically for 8 characters.

TomBaxter:
Yes, have tested 7 and 8 character passwords both when adding a user and changing a password for an existing user.

brianjgeiger:
Looks great. 🎉 🐧 🎉

PR #6104, opened 2016-08-05
Fix Varnish tests

whit537:
## Purpose

Our Varnish tests are occasionally failing at Travis, obscuring test results and adding drag to our development process. Let's fix 'em!

## Changes

Switch `ensure_{licenses,schema}` from a "look before you leap" (LBYL) model to "easier to ask forgiveness than permission" (EAFP), so that each can safely be run concurrently with itself. See the [commit log](https://github.com/CenterForOpenScience/osf.io/pull/6104/commits) for details.

## Side effects

- A couple uniqueness constraints are now implemented in the database instead of in Python, such that violations now raise `KeyExistsException` instead of `ValueValidationError`.
- We now log process and thread ids on calls to `init_app`.
- The `apiserver` task sprouted an `autoreload` knob.
- `ensure_licenses` now returns a tuple (number inserted, number updated) instead of `None` (this aids in testing)
- `ensure_schema` now returns `None` instead of a `MetaSchema` instance (this simplifies the code, allowing it to be mildly more parallel with `ensure_licenses`, though asymmetry yet predominates between the two, a misfortune itself adjudged to be out of scope for this PR).

## Ticket

[OSF-6861](https://openscience.atlassian.net/browse/OSF-6861)

## QA notes

This PR changes _how_ licenses and schemas are loaded, but shouldn't change _what_ is loaded. Therefore, check that the places where these show up in the UI still list the licenses and schemas (I believe these manifest as "registration types"?) that we expect.

## Todo

- [x] consider flipping `autoreload` flag—https://github.com/CenterForOpenScience/osf.io/pull/6104#discussion_r77021869 —addressed at https://github.com/CenterForOpenScience/osf.io/pull/6104#discussion_r77381938
- [ ] move teardown to `tearDown`—https://github.com/CenterForOpenScience/osf.io/pull/6104#discussion_r77040084 —ready for review in f559d4749701b213d8a12a76057b6788498b89e5

### DIFF
--- a/tasks/__init__.py
+++ b/tasks/__init__.py
@@ -86,11 +86,13 @@ def git_logs(ctx, branch=None):
 
 
 @task
-def apiserver(ctx, port=8000, wait=True, host='127.0.0.1'):
+def apiserver(ctx, port=8000, wait=True, autoreload=True, host='127.0.0.1'):
     """Run the API server."""
     env = os.environ.copy()
     cmd = 'DJANGO_SETTINGS_MODULE=api.base.settings {} manage.py runserver {}:{} --nothreading'\
         .format(sys.executable, host, port)
+    if not autoreload:
+        cmd += ' --noreload'
     if settings.SECURE_MODE:
         cmd = cmd.replace('runserver', 'runsslserver')
         cmd += ' --certificate {} --key {}'.format(settings.OSF_SERVER_CERT, settings.OSF_SERVER_KEY)
@@ -532,10 +534,12 @@ def test_admin(ctx):
 @task
 def test_varnish(ctx):
     """Run the Varnish test suite."""
-    proc = apiserver(ctx, wait=False)
-    sleep(5)
-    test_module(ctx, module='api/caching/tests/test_caching.py')
-    proc.kill()
+    proc = apiserver(ctx, wait=False, autoreload=False)
+    try:
+        sleep(5)
+        test_module(ctx, module='api/caching/tests/test_caching.py')
+    finally:
+        proc.kill()
 
 
 @task
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -19,20 +19,23 @@ from tests.base import OsfTestCase, teardown_database
 
 class TestMetaData(OsfTestCase):
 
-    def tearDown(self):
-        super(TestMetaData, self).tearDown()
+    # For whatever reason (efficiency?), DbTestCase only wipes the database
+    # during *class* setup and teardown. This leaks database state across test
+    # cases, which a) smells pretty bad, and b) breaks the test_metaschema_
+    # tests here.
 
-        # For whatever reason, DbTestCase only wipes the database during
-        # *class* setup and teardown. This leaks database state across test
-        # cases, which a) smells pretty bad, and b) breaks the test_metaschema_
-        # tests here.
-
-        teardown_database(database=database_proxy._get_current_object())
+    def setUp(self):
+        super(TestMetaData, self).setUp()
         set_up_storage(
             website.models.MODELS,
             storage.MongoStorage,
             addons=settings.ADDONS_AVAILABLE,
         )
+
+    def tearDown(self):
+        super(TestMetaData, self).tearDown()
+        teardown_database(database=database_proxy._get_current_object())
+
 
     def test_ensure_schemas(self):
 
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -19,12 +19,12 @@ from tests.base import OsfTestCase, teardown_database
 
 class TestMetaData(OsfTestCase):
 
-    def setUp(self):
-        super(TestMetaData, self).setUp()
+    def tearDown(self):
+        super(TestMetaData, self).tearDown()
 
         # For whatever reason, DbTestCase only wipes the database during
         # *class* setup and teardown. This leaks database state across test
-        # cases, which a) smells pretty bad, and b) breaks the test_metadata_
+        # cases, which a) smells pretty bad, and b) breaks the test_metaschema_
         # tests here.
 
         teardown_database(database=database_proxy._get_current_object())
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -4,15 +4,35 @@ import unittest
 from nose.tools import *  # PEP8 asserts
 
 from framework.forms.utils import process_payload
+from framework.mongo import database as database_proxy, set_up_storage
+from modularodm import storage
+from modularodm.exceptions import KeyExistsException
 
+import website.models
+from website import settings
 from website.project.model import MetaSchema
 from website.project.model import ensure_schemas
 from website.project.metadata.schemas import OSF_META_SCHEMAS
 
-from tests.base import OsfTestCase
+from tests.base import OsfTestCase, teardown_database
 
 
 class TestMetaData(OsfTestCase):
+
+    def setUp(self):
+        super(TestMetaData, self).setUp()
+
+        # For whatever reason, DbTestCase only wipes the database during
+        # *class* setup and teardown. This leaks database state across test
+        # cases, which a) smells pretty bad, and b) breaks the test_metadata_
+        # tests here.
+
+        teardown_database(database=database_proxy._get_current_object())
+        set_up_storage(
+            website.models.MODELS,
+            storage.MongoStorage,
+            addons=settings.ADDONS_AVAILABLE,
+        )
 
     def test_ensure_schemas(self):
 
@@ -27,6 +47,18 @@ class TestMetaData(OsfTestCase):
             MetaSchema.find().count(),
             len(OSF_META_SCHEMAS)
         )
+
+    def test_metaschema_uniqueness_is_enforced_in_the_database(self):
+        # Using MongoDB's uniqueness instead of modular-odm's allows us to
+        # kludge a race-less upsert in ensure_schema.
+        MetaSchema(name='foo', schema_version=1).save()
+        assert_raises(KeyExistsException, MetaSchema(name='foo', schema_version=1).save)
+        # modular-odm's uniqueness constraint would raise ValidationValueError instead.
+
+    def test_metaschema_is_fine_with_same_name_but_different_version(self):
+        MetaSchema(name='foo', schema_version=1).save()
+        MetaSchema(name='foo', schema_version=2).save()
+        assert_equal(MetaSchema.find(name='foo').count(), 2)
 
     def test_process(self):
         processed = process_payload({'foo': 'bar&baz'})
--- a/tests/test_node_licenses.py
+++ b/tests/test_node_licenses.py
@@ -8,7 +8,7 @@ import __builtin__ as builtins
 from nose.tools import *  # flake8: noqa (PEP8 asserts)
 import mock
 from modularodm import Q
-from modularodm.exceptions import NoResultsFound
+from modularodm.exceptions import NoResultsFound, KeyExistsException
 
 from framework.auth import Auth
 
@@ -91,6 +91,13 @@ class TestNodeLicenses(OsfTestCase):
         assert_not_equal(record._id, copied._id)
         for prop in ('id', 'name', 'node_license'):
             assert_equal(getattr(record, prop), getattr(copied, prop))
+
+    def test_license_uniqueness_on_id_is_enforced_in_the_database(self):
+        # Using MongoDB's uniqueness instead of modular-odm's allows us to
+        # kludge a race-less upsert in ensure_licenses.
+        NodeLicense(id='foo', name='bar', text='baz').save()
+        assert_raises(KeyExistsException, NodeLicense(id='foo', name='buz', text='boo').save)
+        # modular-odm's uniqueness constraint would raise ValidationValueError instead.
 
     def test_ensure_licenses_existing_licenses(self):
         with mock.patch('website.project.licenses.NodeLicense.__init__', autospec=True) as MockNodeLicense:
--- a/tests/test_node_licenses.py
+++ b/tests/test_node_licenses.py
@@ -125,15 +125,6 @@ class TestNodeLicenses(OsfTestCase):
         MIT = NodeLicense.find_one(
             Q('id', 'eq', 'MIT')
         )
-        assert_equal(MIT.text, CHANGED_TEXT)
-        assert_equal(MIT.properties, CHANGED_PROPERTIES)
-
-    def test_ensure_licenses_updates_existing(self):
-        with mock.patch.object(builtins, 'open', mock.mock_open(read_data=LICENSE_TEXT)):
-            ensure_licenses()
-        MIT = NodeLicense.find_one(
-            Q('id', 'eq', 'MIT')
-        )
         assert_equal(MIT.name, CHANGED_NAME)
         assert_equal(MIT.text, CHANGED_TEXT)
         assert_equal(MIT.properties, CHANGED_PROPERTIES)
--- a/tests/test_node_licenses.py
+++ b/tests/test_node_licenses.py
@@ -99,10 +99,8 @@ class TestNodeLicenses(OsfTestCase):
         assert_raises(KeyExistsException, NodeLicense(id='foo', name='buz', text='boo').save)
         # modular-odm's uniqueness constraint would raise ValidationValueError instead.
 
-    def test_ensure_licenses_existing_licenses(self):
-        with mock.patch('website.project.licenses.NodeLicense.__init__', autospec=True) as MockNodeLicense:
-            ensure_licenses()
-        assert_false(MockNodeLicense.called)
+    def test_ensure_licenses_updates_existing_licenses(self):
+        assert_equal(ensure_licenses(), (0, 16))
 
     def test_ensure_licenses_no_licenses(self):
         before_count = NodeLicense.find().count()
--- a/tests/test_registrations/test_archiver.py
+++ b/tests/test_registrations/test_archiver.py
@@ -15,6 +15,7 @@ from mock import call
 from nose.tools import *  # noqa PEP8 asserts
 import httpretty
 from modularodm import Q
+from modularodm.exceptions import KeyExistsException
 
 from scripts import cleanup_failed_registrations as scripts
 
@@ -254,25 +255,42 @@ def generate_schema_from_data(data):
             return {
                 'qid': qid,
                 'type': 'string'
-            }                
-    schema = MetaSchema(
-        name='Test',
-        schema={
-            'name': "Test",
-            'version': 2,
-            'config': {
-                'hasFiles': True
-            },
-            'pages':  [{
-                'id': 'page1',
-                'questions': [
-                    from_question(qid, q)
-                    for qid, q in data.items()
-                ]
-            }]
+            }
+
+    _schema = {
+        'name': "Test",
+        'version': 2,
+        'config': {
+            'hasFiles': True
         },
+        'pages':  [{
+            'id': 'page1',
+            'questions': [
+                from_question(qid, q)
+                for qid, q in data.items()
+            ]
+        }]
+    }
+    schema = MetaSchema(
+        name=_schema['name'],
+        schema_version=_schema['version'],
+        schema=_schema
     )
-    schema.save()
+    try:
+        schema.save()
+    except KeyExistsException:
+
+        # Unfortunately, we don't have db isolation between test cases for some
+        # reason. Update the doc currently in the db rather than saving a new
+        # one.
+
+        schema = MetaSchema.find_one(
+            Q('name', 'eq', _schema['name']) &
+            Q('schema_version', 'eq', _schema['version'])
+        )
+        schema.schema = _schema
+        schema.save()
+
     return schema
 
 def generate_metadata(file_trees, selected_files, node_index):
--- a/website/app.py
+++ b/website/app.py
@@ -3,6 +3,7 @@
 import importlib
 import json
 import os
+import thread
 from collections import OrderedDict
 
 import django
@@ -92,6 +93,10 @@ def init_app(settings_module='website.settings', set_backends=True, routes=True,
     :param routes: Whether to set the url map.
 
     """
+    logger.info('Initializing the application from process {}, thread {}.'.format(
+        os.getpid(), thread.get_ident()
+    ))
+
     # The settings module
     settings = importlib.import_module(settings_module)
 
--- a/website/project/licenses/__init__.py
+++ b/website/project/licenses/__init__.py
@@ -32,12 +32,18 @@ def serialize_node_license_record(node_license_record):
     return ret
 
 
-@mongo_utils.unique_on(['id', '_id'])
+@mongo_utils.unique_on(['id'])
 class NodeLicense(StoredObject):
 
     _id = fields.StringField(primary=True, default=lambda: str(ObjectId()))
 
-    id = fields.StringField(required=True, unique=True, editable=False)
+    id = fields.StringField(
+        required=True,
+        unique=False,   # Skip modular-odm's uniqueness implementation, depending on MongoDB's
+                        # instead (the decorator will install the proper index), so that we can
+                        # kludge a non-racey upsert in ensure_licenses.
+        editable=False
+    )
     name = fields.StringField(required=True, unique=True)
     text = fields.StringField(required=True)
     properties = fields.StringField(list=True)
--- a/website/project/licenses/__init__.py
+++ b/website/project/licenses/__init__.py
@@ -4,7 +4,7 @@ import os
 import warnings
 
 from modularodm import fields, Q
-from modularodm.exceptions import NoResultsFound
+from modularodm.exceptions import KeyExistsException
 
 from framework.mongo import (
     ObjectId,
@@ -33,6 +33,7 @@ def serialize_node_license_record(node_license_record):
 
 
 @mongo_utils.unique_on(['id'])
+@mongo_utils.unique_on(['name'])
 class NodeLicense(StoredObject):
 
     _id = fields.StringField(primary=True, default=lambda: str(ObjectId()))
@@ -44,7 +45,10 @@ class NodeLicense(StoredObject):
                         # kludge a non-racey upsert in ensure_licenses.
         editable=False
     )
-    name = fields.StringField(required=True, unique=True)
+    name = fields.StringField(
+        required=True,
+        unique=False    # Ditto.
+    )
     text = fields.StringField(required=True)
     properties = fields.StringField(list=True)
 
@@ -84,6 +88,13 @@ class NodeLicenseRecord(StoredObject):
 
 
 def ensure_licenses(warn=True):
+    """Upsert the licenses in our database based on a JSON file.
+
+    :return tuple: (number inserted, number updated)
+
+    """
+    ninserted = 0
+    nupdated = 0
     with open(
             os.path.join(
                 settings.APP_PATH,
@@ -95,27 +106,29 @@ def ensure_licenses(warn=True):
             name = info['name']
             text = info['text']
             properties = info.get('properties', [])
-            node_license = None
             try:
-                node_license = NodeLicense.find_one(
-                    Q('id', 'eq', id)
-                )
-            except NoResultsFound:
-                if warn:
-                    warnings.warn(
-                        'License {name} ({id}) not already in the database. Adding it now.'.format(
-                            name=name,
-                            id=id
-                        )
-                    )
-                node_license = NodeLicense(
+                NodeLicense(
                     id=id,
                     name=name,
                     text=text,
                     properties=properties
+                ).save()
+            except KeyExistsException:
+                node_license = NodeLicense.find_one(
+                    Q('id', 'eq', id)
                 )
-            else:
                 node_license.name = name
                 node_license.text = text
                 node_license.properties = properties
-            node_license.save()
+                node_license.save()
+                nupdated += 1
+            else:
+                if warn:
+                    warnings.warn(
+                        'License {name} ({id}) added to the database.'.format(
+                            name=name,
+                            id=id
+                        )
+                    )
+                ninserted += 1
+    return ninserted, nupdated
--- a/website/project/model.py
+++ b/website/project/model.py
@@ -85,7 +85,7 @@ def has_anonymous_link(node, auth):
         return auth.private_link.anonymous
     return False
 
-@unique_on(['name', 'schema_version', '_id'])
+@unique_on(['name', 'schema_version'])
 class MetaSchema(StoredObject):
 
     _id = fields.StringField(default=lambda: str(ObjectId()))
--- a/website/project/model.py
+++ b/website/project/model.py
@@ -17,8 +17,7 @@ from django.core.validators import URLValidator
 from modularodm import Q
 from modularodm import fields
 from modularodm.validators import MaxLengthValidator
-from modularodm.exceptions import NoResultsFound
-from modularodm.exceptions import ValidationValueError
+from modularodm.exceptions import KeyExistsException, NoResultsFound, ValidationValueError
 
 from framework import status
 from framework.mongo import ObjectId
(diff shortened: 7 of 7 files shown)


whit537:
From restricted [Flowdock](https://www.flowdock.com/app/cos/osf-dev/threads/I5HEZ6DzZg8ouUUdBHtfbZWpTfW):

> I’m pretty sure that’s just MVCC, `ensure_schema`, and a race condition because the varnish test runs both the api server and the flask server and they’re both running `ensure_schema` relatively close together in time.
>
> Seems like we either instruct only one to run `ensure_schema` somehow, or build a check into `ensure_schema` itself, some sort of locking.

whit537:
Working on running the Varnish tests locally ...

whit537:
P.S. Here's [a sample failure](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/149210200).

whit537:
Hmm ... looks like we didn't document installing and configuring Varnish locally.

whit537:
```
$ brew install varnish
Warning: varnish-4.1.0 already installed
```

👍

whit537:
Picked out [a command](https://github.com/whit537/osf.io/blob/23fb6522c10ed9320bc4acb8fde8a2133b6284e2/.travis.yml#L90) from `.travis.yml` and modified it:

```
varnishd -n /tmp -p feature=+esi_disable_xml_check -F -f tests/test_files/varnish.vc -a *:8080
```

Looks like 8080 is conflicting with FakeCAS?

whit537:
Let's see if I can get away without CAS ...

whit537:
```
14:51:48 varnish.1       | Error: socket *:8080 didn't resolve 
```

whit537:
Specifying `0.0.0.0:8080` instead of the wildcard seems to work.

whit537:
Okay!

```
$ ps|grep varnish
81981 ttys005    0:00.00 grep varnish
81798 ttys009    0:00.10 varnishd -n /tmp -p feature=+esi_disable_xml_check -F -f tests/test_files/varnish.vcl -a 0.0.0.0:8080
81923 ttys009    0:00.08 varnishd -n /tmp -p feature=+esi_disable_xml_check -F -f tests/test_files/varnish.vcl -a 0.0.0.0:8080
```

Now for tests?

whit537:
Need to configure some [settings](https://github.com/whit537/osf.io/blob/23fb6522c10ed9320bc4acb8fde8a2133b6284e2/website/settings/defaults.py#L557-L560) ...

```
  File "/Users/whit537/workbench/cos/osf.io/env/lib/python2.7/site-packages/nose/loader.py", line 418, in load
TestsFromName
    addr.filename, addr.module)
  File "/Users/whit537/workbench/cos/osf.io/env/lib/python2.7/site-packages/nose/importer.py", line 47, in imp
ortFromPath
    return self.importFromDir(dir_path, fqname)
  File "/Users/whit537/workbench/cos/osf.io/env/lib/python2.7/site-packages/nose/importer.py", line 94, in imp
ortFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/Users/whit537/workbench/cos/osf.io/api/caching/tests/test_caching.py", line 21, in <module>
    class TestVarnish(DbTestCase):
  File "/Users/whit537/workbench/cos/osf.io/api/caching/tests/test_caching.py", line 22, in TestVarnish
    local_varnish_base_url = '{}/v2/'.format(django_settings.VARNISH_SERVERS[0])
IndexError: list index out of range
```

screen shot 2016-08-05 at 2 56 16 pm

whit537:
[Here we go](https://github.com/whit537/osf.io/blob/23fb6522c10ed9320bc4acb8fde8a2133b6284e2/website/settings/local-travis.py#L54-L57).

whit537:
Progress!

```
FAIL: test_cache_invalidation (api.caching.tests.test_caching.TestVarnish)
  File "/Users/whit537/workbench/cos/osf.io/api/caching/tests/test_caching.py", line 172, in test_cache_invalidation
    assert create_response.ok, 'Failed to create node'
AssertionError: Failed to create node
-------------------- >> begin captured logging << --------------------
framework.mongo.handlers: WARNING: Creating new client instance. 1 instances initialized.
website.util.paths: WARNING: Skipping load of "webpack-assets.json" in DEBUG_MODE.
website.addons.github.settings: WARNING: No local.py settings file found
website.addons.osfstorage.settings: WARNING: No local.py settings file found
website.addons.wiki.settings: WARNING: No local.py settings file found
root: DEBUG: Setting storage backends
root: INFO: Sentry disabled; Flask's debug mode enabled
root: DEBUG: Setting storage backends
root: INFO: Sentry disabled; Flask's debug mode enabled
root: INFO: Sentry disabled; Flask's debug mode enabled
py.warnings: WARNING: /Users/whit537/workbench/cos/osf.io/env/lib/python2.7/site-packages/modularodm/storedobject.py:940: UserWarning: Setting an attribute that is neither a field nor a protected value.
  warnings.warn('Setting an attribute that is neither a field nor a protected value.')

amqp: DEBUG: Start from server, version: 0.9, properties: {u'information': u'Licensed under the MPL.  See http://www.rabbitmq.com/', u'product': u'RabbitMQ', u'copyright': u'Copyright (C) 2007-2015 Pivotal Software, Inc.', u'capabilities': {u'exchange_exchange_bindings': True, u'connection.blocked': True, u'authentication_failure_close': True, u'basic.nack': True, u'per_consumer_qos': True, u'consumer_priorities': True, u'consumer_cancel_notify': True, u'publisher_confirms': True}, u'cluster_name': u'rabbit@Chads-MacBook-Pro', u'platform': u'Erlang/OTP', u'version': u'3.6.0'}, mechanisms: [u'AMQPLAIN', u'PLAIN'], locales: [u'en_US']
amqp: DEBUG: Open OK!
amqp: DEBUG: using channel_id: 1
amqp: DEBUG: Channel open
--------------------- >> end captured logging << ---------------------

Ran 2 tests in 11.317s

FAILED (failures=1)
osf.io| $
```

whit537:
Hmm ... I'm still seeing the apiserver running after the test suite completes. Is [`proc.kill`](https://github.com/whit537/osf.io/blob/23fb6522c10ed9320bc4acb8fde8a2133b6284e2/tasks/__init__.py#L538) not working properly?

whit537:
Yeah, in the failure case the test task exceptions out and never hits `proc.kill`. Let's try:finally that ...

whit537:
Failing test output:

https://gist.github.com/whit537/5e37a8982329510c6a346d32c9d3ad7d

whit537:
Now *that* is some interprocess goobledygook that I've never seen before. 😶

```
(Pdb) ll

168             create_response = requests.post(
169                 '{}nodes/'.format(self.local_python_base_url),
170                 json=payload,
171                 auth=self.authorization)
172             import pdb; pdb.set_trace()
173  ->         assert create_response.ok, 'Failed to create node'
174     
175             node_id = create_response.json()['data']['id']
176             new_title = '{} -- But Changed!'.format(create_response.json()['data']['attributes']['title'])
177     
178             response = requests.get(
(Pdb) ccrreeaattee__rreesspp^^^^^^^^ssppooeenn

*** NameError: name 'crespoen' is not defined
(Pdb) rreeaattee__rreessppoonnssee

*** NameError: name 'reate_response' is not defined
(Pdb) ccrreeaattee__rreessppoonnssee

<Response [401]>
(Pdb)
```

whit537:
```
(Pdb) sseellff..aauutthhoorriizzaattiioonn__^..ppaasssswwoorrdd

u'password'
(Pdb) sseellff..aauutthhoorriizzaattiioonn__^..uusseerrnnaammee

u'ce8065a2-2101-41ee-9e7d-2adf7a50f1fa@example.com'
(Pdb)
```

Hmm ...

whit537:
Alright, let's make sure all parties are talking to the same database ...

whit537:
Looks like the user is in the db. Is the apiserver talking to the same db?

```
(Pdb) sseellff..DDBB__NNAAMMEE

'osf_test'
(Pdb)
```

screen shot 2016-08-05 at 4 04 51 pm

whit537:
No! I think ...

whit537:
Yessssssssssss. 👍

Okay! Now what?

whit537:
I'm seeing license spam!

```
/Users/whit537/workbench/cos/osf.io/website/project/licenses/__init__.py:102: UserWarning: License MIT License (MIT) not already in the database. Adding it now.
  id=id

[py.warnings]  WARNING: /Users/whit537/workbench/cos/osf.io/website/project/licenses/__init__.py:102: UserWarning: License MIT License (MIT) not already in the database. Adding it now.
  id=id

/Users/whit537/workbench/cos/osf.io/website/project/licenses/__init__.py:102: UserWarning: License GNU General Public License (GPL) 3.0 (GPL3) not already in the database. Adding it now.
  id=id

[py.warnings]  WARNING: /Users/whit537/workbench/cos/osf.io/website/project/licenses/__init__.py:102: UserWarning: License GNU General Public License (GPL) 3.0 (GPL3) not already in the database. Adding it now.
  id=id

/Users/whit537/workbench/cos/osf.io/website/project/licenses/__init__.py:102: UserWarning: License GNU General Public License (GPL) 2.0 (GPL2) not already in the database. Adding it now.
  id=id

[py.warnings]  WARNING: /Users/whit537/workbench/cos/osf.io/website/project/licenses/__init__.py:102: UserWarning: License GNU General Public License (GPL) 2.0 (GPL2) not already in the database. Adding it now.
  id=id

/Users/whit537/workbench/cos/osf.io/env/lib/python2.7/site-packages/modularodm/storedobject.py:940: UserWarning: Setting an attribute that is neither a field nor a protected value.
  warnings.warn('Setting an attribute that is neither a field nor a protected value.')
```

whit537:
I'm impressed at how long it takes this single test to run.

whit537:
```
Ran 2 tests in 229.160s

OK
osf.io| $
```

Okay! Now to trigger the discombobulation ...

whit537:
> the varnish test runs both the api server and the flask server

Is this true? I see the api server starting ("Starting development server at http://127.0.0.1:8000/"), but I don't see the flask server (I'd expect "Starting development server at http://127.0.0.1:5000/").

whit537:
I'm definitely seeing double-logging following `invoke test_travis_$TEST_BUILD` in the [failure case](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/149210200#L1336), and single logging in the [successful case](https://travis-ci.org/CenterForOpenScience/osf.io/builds/150147761#L1334).

whit537:
Yet in both cases, "No handlers could be found for logger 'framework.mongo.handlers'" is only logged once: a clue?

whit537:
Found another bug with terminating the API server (using Django's reloader meant there was an orphan process). Fixed in 2b469d4.

whit537:
> the varnish test runs both the api server and the flask server

Hey @cwisecarver, do you know this for a fact? I'm definitely seeing the API server, but I think I'm not seeing the Flask server ...

cwisecarver:
[You're correct.](https://github.com/CenterForOpenScience/osf.io/blob/develop/tasks/__init__.py#L533)

I would just put a command that prints the caller of the function [here](https://github.com/CenterForOpenScience/osf.io/blob/696d3154a61b48470b202c039b21e2e55028749c/website/project/model.py#L166)

That should tell you what's calling it twice.

cwisecarver:
It's probably true that the proc.kill isn't working. @icereval and I didn't spend a lot of time trying to fix that because Travis will kill the vm/container once the job has run anyway which *should* kill the process.

whit537:
> It's probably true that the proc.kill isn't working.

It's working now in 2b469d4f61d54b5c90f8c3457bb1f5bab1796ccf! :-)

> You're correct.

Cool. I wasn't sure if there was another layer I was missing where the Flask server might be getting launched (I scoured `.travis.yml` and didn't find anything).

> I would just put a command that prints the caller of the function here
>
That should tell you what's calling it twice.

Good idea, lessee here ...

whit537:
It's not just `ensure_schemas` that's getting called twice. Based on the [log output](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/149210200#L1336), it appears to be [`init_app`](https://github.com/whit537/osf.io/blob/2b469d4f61d54b5c90f8c3457bb1f5bab1796ccf/website/app.py#L84-L133) itself. Hmm ...

cwisecarver:
That's what I figured. I know it's called in `/manage.py`. Not sure where else.

whit537:
Yeah, I see it in `manage.py` ... and in `api/base/wsgi.py`? Maybe that's the second call site?

cwisecarver:
Probable. If we're running the uwsgi server in travis it will call api/base/wsgi but I don't believe that the django dev server uses wsgi.py for anything.

whit537:
Confirmed:

```
-> return get_internal_wsgi_application()
  /Users/whit537/workbench/cos/osf.io/env/lib/python2.7/site-packages/django/core/servers/basehttp.py(47)get_internal_wsgi_application()
-> return import_string(app_path)
  /Users/whit537/workbench/cos/osf.io/env/lib/python2.7/site-packages/django/utils/module_loading.py(20)import_string()
-> module = import_module(module_path)
  /usr/local/lib/python2.7/importlib/__init__.py(37)import_module()
-> __import__(name)
  /Users/whit537/workbench/cos/osf.io/api/base/wsgi.py(53)<module>()
-> init_app(set_backends=True, routes=False, attach_request_handlers=False)
> /Users/whit537/workbench/cos/osf.io/website/app.py(97)init_app()
-> settings = importlib.import_module(settings_module)
(Pdb)
```

whit537:
> I don't believe that the django dev server uses wsgi.py for anything.

Django's dev server [does use WSGI](https://github.com/django/django/blob/1.9/django/core/servers/basehttp.py#L182-L198).

whit537:
Okay! Why does the bug only appear *sometimes*?

whit537:
I see checks inside [`ensure_schema`](https://github.com/whit537/osf.io/blob/2b469d4f61d54b5c90f8c3457bb1f5bab1796ccf/website/project/model.py#L146-L163) and [`ensure_licenses`](https://github.com/whit537/osf.io/blob/2b469d4f61d54b5c90f8c3457bb1f5bab1796ccf/website/project/licenses/__init__.py#L80-L115) for whether a {schema,license} is already loaded. Let's dig into that behavior ...

cwisecarver:
Mongo is MVCC. I bet it's just a race condition where one time it's getting a dirty read and the next time it's not.

whit537:
Well, the [failure](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/149210200#L1480) is that `find_one` is finding two:

> Query for find_one must return exactly one result; returned 2

Which means that both processes are ending up in the `NoResultsFound` [block](https://github.com/whit537/osf.io/blob/2b469d4f61d54b5c90f8c3457bb1f5bab1796ccf/website/project/model.py#L153-L159). Wait ... doesn't this mean that `ensure_schema` is getting called _three_ times? Twice to load, then a third time to fall over on `find_one`?

whit537:
I mean, the fix here is either to teach `ensure_{license,schema}` how to deal with this situation, or to teach `init_app` how to serialize itself (only running once), or to put the serialization at a higher level (don't run `init_app` more than once in the first place).

whit537:
[`--no-init-app`](https://github.com/whit537/osf.io/blob/2b469d4f61d54b5c90f8c3457bb1f5bab1796ccf/manage.py#L7-L10), eh? Hmm ...

cwisecarver:
If it's a race condition around concurrent database transactions there's not really a way to fix it other than letting the database handle uniqueness rather than allowing duplicate records to be inserted. `find_one` should fail if there are more than one because we depend on that behavior elsewhere. But inserting duplicate records should fail and the thing inserting them should accept that failure and work around it.

cwisecarver:
The `--no-init-app` thing is interesting. Probably worth a shot.

whit537:
I can suppress one call to `init_app` using `--no-init-app`, but there are three others(!) that aren't thus suppressed.

```diff
diff --git a/tasks/__init__.py b/tasks/__init__.py
index ff6613f..881f072 100755
--- a/tasks/__init__.py
+++ b/tasks/__init__.py
@@ -86,13 +86,15 @@ def git_logs(ctx, branch=None):
 
 
 @task
-def apiserver(ctx, port=8000, wait=True, autoreload=True, host='127.0.0.1'):
+def apiserver(ctx, port=8000, wait=True, autoreload=True, init_app=True, host='127.0.0.1'):
     """Run the API server."""
     env = os.environ.copy()
     cmd = 'DJANGO_SETTINGS_MODULE=api.base.settings {} manage.py runserver {}:{} --nothreading'\
         .format(sys.executable, host, port)
     if not autoreload:
         cmd += ' --noreload'
+    if not init_app:
+        cmd += ' --no-init-app'
     if settings.SECURE_MODE:
         cmd = cmd.replace('runserver', 'runsslserver')
         cmd += ' --certificate {} --key {}'.format(settings.OSF_SERVER_CERT, settings.OSF_SERVER_KEY)
@@ -534,7 +536,7 @@ def test_admin(ctx):
 @task
 def test_varnish(ctx):
     """Run the Varnish test suite."""
-    proc = apiserver(ctx, wait=False, autoreload=False)
+    proc = apiserver(ctx, wait=False, autoreload=False, init_app=False)
     try:
         sleep(5)
         test_module(ctx, module='api/caching/tests/test_caching.py')
diff --git a/website/app.py b/website/app.py
index d93a943..b216b0b 100644
--- a/website/app.py
+++ b/website/app.py
@@ -92,6 +92,7 @@ def init_app(settings_module='website.settings', set_backends=True, routes=True,
     :param routes: Whether to set the url map.
 
     """
+    open('init_app.log', 'a').write('+')
     # The settings module
     settings = importlib.import_module(settings_module)
```

whit537:
Here's the five calls:

```
1:  manage.py(13)

2:  env/lib/python2.7/site-packages/django/core/servers/basehttp.py(47)get_internal_wsgi_application()
    [...]
    api/base/wsgi.py(53)<module>()
    website/app.py(97)init_app()

3:  tests/base.py(27)<module>()
    api/base/wsgi.py(53)<module>()
    website/app.py(97)init_app()

4:  tests/base.py(27)<module>()
    api/base/wsgi.py(53)<module>()
    website/app.py(97)init_app()

5:  tests/base.py(66)<module>()
    website/app.py(97)init_app()
```

cwisecarver:
The last three appear to be the same...

whit537:
(3) and (4) are the same. (5) is different.

whit537:
The *real* fix here is to refactor so that we only call `init_app` once ever.

cwisecarver:
I'd still say, just because the team is so big and fluid, it would be nice if `init_app` was smart enough to know it had been called and either bail out or raise an exception if it's called more than once. Sometimes it's better to expect failure.

whit537:
> smart enough to know it had been called and either bail out or raise an exception if it's called more than once. 

Seems like a good approach:

1. Teach `init_app` to return on calls after the first.
1. Gradually remove calls after the first.
1. Convert the `return` from (1) into a raise.

One consideration: `init_app` provides a number of parameters. It could be that multiple calls are _necessary_ in some circumstances in order to fully initialize the app.

whit537:
> Teach `init_app` to return on calls after the first.

We need to synchronize across a process boundary ... hmm ...

whit537:
Alright, that's way more complicated. I'm flipping back to addressing the immediate issue with modifications to `ensure_{schema,licenses}`. Basically we should be able to add a uniqueness constraint to and then flip the try/except.

whit537:
I've discovered the [`__indices__`](https://github.com/whit537/osf.io/blob/2b469d4f61d54b5c90f8c3457bb1f5bab1796ccf/framework/mongo/handlers.py#L134-L136) feature. Working up a commit ...

whit537:
Gosh. Why would we not just use `upsert` here?

whit537:
I find scant usage of `upsert` in our codebase. Perhaps our MongoDB wrappers don't make it easy?

whit537:
> Perhaps our MongoDB wrappers don't make it easy?

Hrm, [not promising](https://github.com/CenterForOpenScience/modular-odm/blob/639faf59bd20166438a7629239380ea63c50677e/modularodm/storage/mongostorage.py#L242).

cwisecarver:
Looks like [we use pymongo directly when we want upsert](https://github.com/CenterForOpenScience/osf.io/search?utf8=%E2%9C%93&q=upsert).

whit537:
Blerb. ModularODM also [seems](https://github.com/CenterForOpenScience/modular-odm/blob/master/modularodm/fields/field.py#L112-L120) to implement its own uniqueness constraint, apart from MongoDB. Is that going to trip us up here?

cwisecarver:
"Implement" ... it doesn't work, at least not consistently. You can however [add an index to the model and include unique](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/project/model.py#L764).

whit537:
Yeah, I'm not sure we can avoid this race condition without dropping below ModularODM.

whit537:
I've got a couple more commits for the `NodeLicense` side. I'm waiting for Travis before pushing them, in order to show the test results helpfully. I'm hitting some test fixture issues with `MetaSchema`. It looks like the db is not getting wiped from the second test to the first:

```diff
diff --git a/tests/test_metadata.py b/tests/test_metadata.py
index 24af65c..88a30ab 100644
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -4,6 +4,7 @@ import unittest
 from nose.tools import *  # PEP8 asserts
 
 from framework.forms.utils import process_payload
+from modularodm.exceptions import KeyExistsException
 
 from website.project.model import MetaSchema
 from website.project.model import ensure_schemas
@@ -28,6 +29,18 @@ class TestMetaData(OsfTestCase):
             len(OSF_META_SCHEMAS)
         )
 
+    def test_metaschema_uniqueness_is_enforced_in_the_database(self):
+        # Using MongoDB's uniqueness instead of modular-odm's allows us to
+        # kludge a race-less upsert in ensure_schema.
+        MetaSchema(name='foo', schema_version=1).save()
+        assert_raises(KeyExistsException, MetaSchema(name='foo', schema_version=1).save)
+        # modular-odm's uniqueness constraint would raise ValidationValueError instead.
+
+    def test_metaschema_is_fine_with_same_name_but_different_version(self):
+        MetaSchema(name='foo', schema_version=1).save()
+        MetaSchema(name='foo', schema_version=2).save()
+        assert_equal(MetaSchema.find(name='foo').count(), 2)
+
     def test_process(self):
         processed = process_payload({'foo': 'bar&baz'})
         assert_equal(processed['foo'], 'bar%26baz')
```

cwisecarver:
If the db isn't getting wiped that is a much simpler explanation/fix.

whit537:
Ah! It's because [`teardown_database`](https://github.com/CenterForOpenScience/osf.io/blob/2b469d4f61d54b5c90f8c3457bb1f5bab1796ccf/tests/base.py#L93-L102) is part of _class_ [setup](https://github.com/CenterForOpenScience/osf.io/blob/2b469d4f61d54b5c90f8c3457bb1f5bab1796ccf/tests/base.py#L121-L148) and [teardown](https://github.com/CenterForOpenScience/osf.io/blob/2b469d4f61d54b5c90f8c3457bb1f5bab1796ccf/tests/base.py#L150-L161), not instance. Hmm ...

whit537:
That seems non-ideal, to only wipe the database once per test class. That means that most of our tests are not isolated from one another.

whit537:
Alright, I've got commits ready for `ensure_schema` as well.

whit537:
Had a braintstorm on vacation: maybe the autoreload is the source of the problem? Initializing the application multiple times is not the problem; initializing the application _concurrently_ is the problem. I'm almost positive we don't have any threading concurrency in play here, but rather that two _processes_ are implicated. If that's the case, then 2b469d4f61d54b5c90f8c3457bb1f5bab1796ccf should be enough to fix the bug, and if the rest of the refactoring here is worthwhile then that should probably be a new PR.

whit537:
Hmm ... with some local logging, it looks like we have multiple processes even with 2b469d4f61d54b5c90f8c3457bb1f5bab1796ccf.

```
Performing system checks...

System check identified no issues (0 silenced).
August 15, 2016 - 07:18:12
Django version 1.9, using settings 'api.base.settings'
Starting development server at http://127.0.0.1:8000/
Quit the server with CONTROL-C.
[root]  INFO: Initializing the application from process 6462, thread 140735236629248.
[root]  DEBUG: Setting storage backends
[root]  INFO: Sentry disabled; Flask's debug mode enabled
[root]  INFO: Initializing the application from process 6481, thread 140735236629248.
```

whit537:
Really weird that the thread id is the same, even though the pid is different. 😶

whit537:
Alright, so not already fixed with 2b469d4f61d54b5c90f8c3457bb1f5bab1796ccf. I guess we're back to converting `ensure_{schema,licenses}` to EAFP.

whit537:
The first process is the [apiserver](https://github.com/CenterForOpenScience/osf.io/blob/2b469d4f61d54b5c90f8c3457bb1f5bab1796ccf/tasks/__init__.py#L537), the second is [nose](https://github.com/CenterForOpenScience/osf.io/blob/2b469d4f61d54b5c90f8c3457bb1f5bab1796ccf/tasks/__init__.py#L510). There's no shortcut here. EAFP it is! Onward! 🚵

whit537:
Alright, so why did b3fce73ecdfa4a2ed2b4cc0db115b8f5c4d6cad3 _not_ fail?

whit537:
It faileth not locally. 😕

whit537:
Sike! I tested it not locally. :)

whit537:
Ah! And we're not running that test on Travis, because of 23fb6522c10ed9320bc4acb8fde8a2133b6284e2. I guess it's time to rebase that back out ...

whit537:
Rebased, previous head was b3fce73ecdfa4a2ed2b4cc0db115b8f5c4d6cad3.

whit537:
D'oh! Looks like we've got [a few regressions](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/152485009#L3768-L3872), seemingly related to the insufficient database isolation during testing that surfaced in b213fb51f87ce61984df17ed4105f32a4912764e. So close ... !

whit537:
Ready for review! 💃

tasks/__init__.py
```diff
@@ -86,11 +86,13 @@ def git_logs(ctx, branch=None):
 
 
 @task
-def apiserver(ctx, port=8000, wait=True, host='127.0.0.1'):
+def apiserver(ctx, port=8000, wait=True, autoreload=True, host='127.0.0.1'):
```
mfraezz: I'm not certain that it's possible to specify a `False` value for boolean flags for `invoke`. I remember running into this before, and a quick grep through [their docs](http://docs.pyinvoke.org/en/0.13.0/index.html) doesn't help much.
```
~/D/osf.io ❯ inv apiserver --autoreload=0                      ⏎ remotes/cos/pr/6104 ✭ ◼
No idea what '0' is!
~/D/osf.io ❯ inv apiserver --autoreload=False                  ⏎ remotes/cos/pr/6104 ✭ ◼
No idea what 'False' is!
~/D/osf.io ❯ inv apiserver --autoreload=null                   ⏎ remotes/cos/pr/6104 ✭ ◼
No idea what 'null' is!
~/D/osf.io ❯ inv apiserver --autoreload=false                  ⏎ remotes/cos/pr/6104 ✭ ◼
No idea what 'false' is!
```
Consider a default of `noreload=False` instead

tests/test_metadata.py
```diff
@@ -4,16 +4,36 @@
...
 
+    def setUp(self):
+        super(TestMetaData, self).setUp()
+
+        # For whatever reason, DbTestCase only wipes the database during
+        # *class* setup and teardown. This leaks database state across test
+        # cases, which a) smells pretty bad, and b) breaks the test_metadata_
+        # tests here.
+
+        teardown_database(database=database_proxy._get_current_object())
```
mfraezz: While I agree that this is probably necessary, it seems like it'd be more appropriate to go into the instance `tearDown` instead of `setUp`. Not blocking

mfraezz:
@whit537 two non-blocking comments, otherwise LGTM. Please address, this should be RTM afterwards :octocat:

whit537:
Thanks @mfraezz. I've added a Todo section to the PR description and will look into those ...

tasks/__init__.py
whit537: > I'm not certain that it's possible to specify a `False` value for boolean flags for `invoke`.

This is achievable with `inv apiserver --no-autoreload`.

whit537: Looks like the docs refer to this as "[Automatic boolean inverse flags](http://docs.pyinvoke.org/en/0.13.0/concepts/cli/intro.html#automatic-boolean-inverse-flags)."

tests/test_metadata.py
whit537: Done in f559d4749701b213d8a12a76057b6788498b89e5.

whit537:
Okay! Two issues addressed, back to Code Review ...

tasks/__init__.py
mfraezz: Hmm. Good to know, thanks.

tests/test_metadata.py
```diff
@@ -4,16 +4,36 @@
...
-from tests.base import OsfTestCase
+from tests.base import OsfTestCase, teardown_database
 
 
 class TestMetaData(OsfTestCase):
 
+    def tearDown(self):
+        super(TestMetaData, self).tearDown()
+
+        # For whatever reason, DbTestCase only wipes the database during
```
mfraezz: >For whatever reason

Probably efficiency, but I don't know that anyone has timed it to see if it makes a significant difference.

tests/test_metadata.py
```diff
@@ -4,16 +4,36 @@
...
+        # *class* setup and teardown. This leaks database state across test
+        # cases, which a) smells pretty bad, and b) breaks the test_metaschema_
+        # tests here.
+
+        teardown_database(database=database_proxy._get_current_object())
+        set_up_storage(
```
mfraezz: I only meant that `teardown_database` should be logically placed in `def tearDown`. Similarly, `set_up_storage` seems like it would fit best in `def setUp`. If this causes state issues with the class methods, disregard.

whit537: I yanked this code straight [from](https://github.com/CenterForOpenScience/osf.io/blob/f559d4749701b213d8a12a76057b6788498b89e5/tests/base.py#L137-L144) `DbTestCase.setUpClass`. I can try splitting the two calls across `setUp` and `tearDown` here ...

whit537: Done in 7adb8ecdd90306178261fc83fa1eb7733c12007e.

tests/test_metadata.py
whit537: Nodded to in 7adb8ecdd90306178261fc83fa1eb7733c12007e. :)

whit537:
More spurious [test](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/157158642#L1379-L1382) [failures](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/157158643#L1377-L1380)? :-/

```
osf.io| $ flake8 .
osf.io| $ flake8 framework/
osf.io| $ flake8 framework/auth/views.py 
osf.io| $
```

Maybe that's a merge commit and the flake8iness is on `develop`?

whit537:
Yorp.

```
osf.io| $ flake8 .
./framework/auth/views.py:361:10: E111 indentation is not a multiple of four
./framework/auth/views.py:366:10: E111 indentation is not a multiple of four
./framework/auth/views.py:367:10: E111 indentation is not a multiple of four
osf.io| $ git show
commit b9e724f6aac57aac9857033d3cf9b7de74544e77
Merge: cbe3a5c 29875c7
Author: icereval <icereval@gmail.com>
Date:   Fri Sep 2 15:29:47 2016 -0400

    Merge branch 'hotfix/0.91.19' into develop

osf.io| $
```

whit537:
PR to fix indentation in #6273.

PR #6117, opened 2016-08-09
[#OSF-6351]change the dropzone route

chennan47:
## Purpose
Change the dropzone to be based on commits instead of branch.

## Side effects
Have a updated custome version of dropzone rdy to go in. This pr should be merged after https://github.com/CenterForOpenScience/dropzone/pull/2 is merged.

## Ticket

https://openscience.atlassian.net/browse/OSF-6351


brianjgeiger:
Well, that didn't go as well as we might hope.

PR #6122, opened 2016-08-09
[Feature] Configure External Link addon via APIv2 [OSF-6653]

mfraezz:
## Purpose
Allow users to configure forward via APIv2

## Changes
* Refactored AddonNodeSettingsSerializer
* Added `url` and `label` as serializer fields for forward
* Tests

## Side effects
None expected

## Ticket
[OSF-6653](https://openscience.atlassian.net/browse/OSF-6653)

### DIFF
--- a/api/nodes/serializers.py
+++ b/api/nodes/serializers.py
@@ -349,16 +349,20 @@ class NodeSerializer(JSONAPISerializer):
         return node
 
 
-class NodeAddonSettingsSerializer(JSONAPISerializer):
+class NodeAddonSettingsSerializerBase(JSONAPISerializer):
     class Meta:
         type_ = 'node_addons'
 
     id = ser.CharField(source='config.short_name', read_only=True)
+    node_has_auth = ser.BooleanField(source='has_auth', read_only=True)
+    configured = ser.BooleanField(read_only=True)
     external_account_id = ser.CharField(source='external_account._id', required=False, allow_null=True)
     folder_id = ser.CharField(required=False, allow_null=True)
     folder_path = ser.CharField(required=False, allow_null=True)
-    node_has_auth = ser.BooleanField(source='has_auth', read_only=True)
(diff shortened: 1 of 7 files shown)


api/nodes/serializers.py
```diff
@@ -374,6 +378,67 @@ def get_absolute_url(self, obj):
...
+        elif instance and not instance.url:
+            instance.url = url
+            instance.label = label
+            url_changed = True
+
+        instance.save()
+
+        if url_changed:
+            # add log here because forward architecture isn't great
+            # TODO [OSF-6678]: clean this up
```
brianjgeiger: 👍 You're the best, Matt!

api_tests/nodes/views/test_node_addons.py
```diff
@@ -831,3 +829,166 @@ def test_settings_detail_incomplete_PUT_raises_error(self):
...
+        res = self.app.get(
+            self.setting_list_url,
+            auth=noncontrib.auth)
+        addon_data = self.get_response_for_addon(res)
+
+        assert_equal(self.node_settings.url, addon_data['url'])
+        assert_equal(self.node_settings.label, addon_data['label'])
+
+    def test_settings_detail_PATCH_to_add_folder_without_auth_conflict(self):
+        pass
```
brianjgeiger: Since this sort of thing is generally discouraged when we make tests, it's probably worth putting in a comment about it. I presume it's because these tests don't make sense for this addon but you need them for your testing suite?

brianjgeiger:
A comment on the 'pass'ed tests to explain why it's okay in this case, but otherwise looks great. 🐧

api_tests/nodes/views/test_node_addons.py
mfraezz: That is correct.

brianjgeiger:
Looks great. 🎉 🐧 🎉

PR #6124, opened 2016-08-10
Remove routes for updating registration subviews [OSF-5486]

Johnetordoff:
## Purpose

Currently, subviews for registrations (comments, contributors, node_links) are being subclassed from the node views for maximal DRY. Unfortunately, those views are based on write mixins, so the documentation for the registration views include put, post, patch, and deletes. This PR breaks up removes thoses unsupported routes by breaking up the Node classes

## Changes

Split apart the Contributor Detail/List and Node Links Detail/List for nodes and registrations into base views that only retrieve the information.

## Side effects

A test had to be changed slightly due to the way permissions are used for withdrawn registrations.

## Ticket

https://openscience.atlassian.net/browse/OSF-5486


api/registrations/views.py
```diff
@@ -282,25 +283,70 @@ def get_object(self):
...
-class RegistrationChildrenList(NodeChildrenList, RegistrationMixin):
+    permission_classes = (
+        ContributorDetailPermissions,
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        ReadOnlyIfRegistration,
+        base_permissions.TokenHasScope,
+    )
+
+class RegistrationChildrenList(JSONAPIBaseView, generics.ListAPIView, ODMFilterMixin, RegistrationMixin):
+    """Lists the children of a registration."""
```
brianjgeiger: Everything that's appropriate from the original view docstring should still be moved over to this view so that the swagger/BAPI docs are complete. It's not going to be super-DRY, but we are stuck with what we have for docs right now.

brianjgeiger: The same goes for all of the other views, of course.

api/registrations/views.py
```diff
@@ -282,25 +283,70 @@ def get_object(self):
...
+        return contributors
+
+
+class RegistrationContributorDetail(BaseContributorDetail, RegistrationMixin, UserMixin):
+    """Detail of a contributor for a registration."""
     view_category = 'registrations'
     view_name = 'registration-contributor-detail'
     serializer_class = RegistrationContributorsSerializer
 
+    required_read_scopes = [CoreScopes.NODE_CONTRIBUTORS_READ]
```
brianjgeiger: Anything for a registration should have a CoreScope based on `REGISTRATION_` rather than `NODE_`, as that's one of the major reason why we've broken registrations away from nodes in the API.

api/registrations/views.py
```diff
@@ -282,25 +283,70 @@ def get_object(self):
...
 
+    permission_classes = (
+        ContributorOrPublic,
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        ReadOnlyIfRegistration,
+        base_permissions.TokenHasScope,
+        ExcludeWithdrawals
+    )
+
+    required_read_scopes = [CoreScopes.NODE_CHILDREN_READ]
```
brianjgeiger: Same for here on scopes for registrations

api/registrations/views.py
```diff
@@ -282,25 +283,70 @@ def get_object(self):
...
+    required_write_scopes = [CoreScopes.NODE_CHILDREN_WRITE]
```
brianjgeiger: This should probably be the, what is it , CoreScopes.Null? Something like that. We can't edit a registrations children, so it shouldn't have a write scope. (If this is all how it is now, I apologize, but now's as good of a time to catch it as any)

api/registrations/views.py
```diff
@@ -282,25 +283,70 @@ def get_object(self):
...
-class RegistrationContributorsList(NodeContributorsList, RegistrationMixin):
+class RegistrationContributorsList(BaseContributorList, RegistrationMixin, UserMixin):
+    """List of contributors for a registration."""
     view_category = 'registrations'
     view_name = 'registration-contributors'
+    serializer_class = RegistrationContributorsSerializer
 
-    def get_serializer_class(self):
-        return RegistrationContributorsSerializer
+    required_read_scopes = [CoreScopes.NODE_CONTRIBUTORS_READ]
```
brianjgeiger: `REGISTRATIONS_`

brianjgeiger:
Pass done. Good by and large with a few additional changes. In general, make sure the scopes are all correct (registration_read and null for the write), make sure all the docstrings are in any appropriate for generally read-only views, and also please get me before and after screenshots of the swagger docs list of methods for registrations. 🐧

caseyrollins:
@brianjgeiger or @johnetordoff, can you close this in favor of #6174?

PR #6126, opened 2016-08-10
Prereg admin should not be able to approve registrations more than once[#OSF-6845]

chennan47:
## Purpose
Disable the "Save" button after it is clicked to prevent multi click duplicate more than one pending registration. 

## Changes
Now after the button clicked and if the respond is slow, you will see the button text changed to "Sending. Please wait" and the button is diabled

## Ticket

https://openscience.atlassian.net/browse/OSF-6845


brianjgeiger:
This looks good to me. 🎉 🐧 🎉

PR #6129, opened 2016-08-10
[Hotfix]Make an analytics task to determine the number of older wiki entries [OSF-6752]

TomBaxter:
## Purpose:
[OSF-6752](https://openscience.atlassian.net/browse/OSF-6752)
Create analytics script to count legacy wikis

## Changes:
Add scripts/analytics/legacy_wikis.py counts legacy wikis and average size
Update  scripts/analytics/tasks.py to reference new script

## Side effects
None

[#OSF-6752]

### DIFF
--- a/scripts/analytics/legacy_wikis.py
+++ b/scripts/analytics/legacy_wikis.py
@@ -1,0 +1,62 @@
+# -*- coding: utf-8 -*-
+
+import os
+import sys
+from tabulate import tabulate
+
+from framework.mongo import database
+from website import settings
+from website.addons.wiki import settings as wiki_settings
+
+from .utils import mkdirp
+
+wiki_collection = database['nodewikipage']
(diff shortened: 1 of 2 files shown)


brianjgeiger:
I think I'll restart travis on this. I'm not sure who stopped it, but it will let us know if, for example, we created a problem with the install or a import loop or similar. Maybe I'll just do one branch.

brianjgeiger:
Okay, this looks good to me. 🎉 🐧 🎉

TomBaxter:
Sample output. Average size and largest are in bytes.

  total    legacy    average_size    largest
-------  --------  --------------  ---------
     13         0               0          0

sloria:
Got this data to @brianjgeiger , and we don't need to run this weekly, so closing this for now.

PR #6130, opened 2016-08-10
Feature/Create Users Via Cookie Auth [OSF-6879]

pattisdr:
## Purpose

Changes permissions under which a user can create an unregistered user via the API v2.  Previously, a user could only create another user if they had a properly scoped token (users-create).  Creating them with cookie auth was not allowed.  This allows you to create users if you have a cookie.

Needed to create unregistered contributors for preprints and will eventually be needed for OSF unreg contrib functionality as well. 

## Changes

Removes RequiresScopedRequestOrReadOnly from permissions classes

## Ticket

https://openscience.atlassian.net/browse/OSF-6879

### DIFF
--- a/api/users/views.py
+++ b/api/users/views.py
@@ -103,6 +103,7 @@ class UserList(JSONAPIBaseView, generics.ListCreateAPIView, ODMFilterMixin):
     """
     permission_classes = (
         drf_permissions.IsAuthenticatedOrReadOnly,
+        base_permissions.RequiresScopedRequestOrReadOnly,
         base_permissions.TokenHasScope,
     )
 
--- a/api/users/views.py
+++ b/api/users/views.py
@@ -103,7 +103,6 @@ class UserList(JSONAPIBaseView, generics.ListCreateAPIView, ODMFilterMixin):
     """
     permission_classes = (
         drf_permissions.IsAuthenticatedOrReadOnly,
-        base_permissions.RequiresScopedRequestOrReadOnly,
         base_permissions.TokenHasScope,
     )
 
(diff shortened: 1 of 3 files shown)


api/users/views.py
```diff
@@ -103,7 +103,6 @@ class UserList(JSONAPIBaseView, generics.ListCreateAPIView, ODMFilterMixin):
     """
     permission_classes = (
         drf_permissions.IsAuthenticatedOrReadOnly,
-        base_permissions.RequiresScopedRequestOrReadOnly,
```
brianjgeiger: I don't think removing this is correct; I think we just want to remove the cookie requirement for this permission class. We do want it to be either scoped, readonly, or cookie.

api_tests/users/views/test_user_list.py
```diff
@@ -122,27 +122,38 @@ def test_user_can_not_create_other_user_or_send_mail(self, mock_mail):
             expect_errors=True
         )
 
-        assert_equal(res.status_code, 403)
+        assert_equal(res.status_code, 201)
```
brianjgeiger: Hmmm. Probably cookie and appropriately-scoped token are good, maybe no on the basic auth.

brianjgeiger:
A couple minor changes should do for this. Pass done. 🐧

api/users/views.py
pattisdr: So this class, `RequiresScopedRequestOrReadOnly`, needs to check if there's a cookie or appropriate token.  If so, allow to go through. If just basic auth, the request should fail?

brianjgeiger: Yes, that all sounds correct.

brianjgeiger:
Looks good. 🎉 🐧 🎉

PR #6135, opened 2016-08-10
[hotfix] limit max number of popular projects to show

felliott:
## Purpose

Keen doesn't limit how many entries it returns.  Set a maximum limit to keep explore activity page at a  reasonable size.


felliott:
I'm dumb, this should be against master.  Closing.

PR #6137, opened 2016-08-10
[OSF-5360] Search in 'add link' should show all your projects by default

mluke93:
## Purpose

Allow for users to more easily search through their projects, even if they're unsure of the project names. Also to update to V2, and deal with any changes between V1 and V2.

## Changes

V1 API has been discarded. Instead V2 API has taken its place. If query is left blank, then it will search without a filter (ie. searches everything). The title informs you if you are currently searching "All" or "My" Projects. Due to V2 separating Projects and Registrations, two tabs have been added to distinguish which type you are looking for. Now, when clicking the "+" button next to a node, it adds it straight to the project, instead of a "node cart" for you to have to click save to add. The nodes become toggleable instead of removed.

## Side effects

BLOCKER: Needs /registration_links/ endpoint to be added to the API before this will work.

## Ticket

https://openscience.atlassian.net/browse/OSF-5360

### DIFF
--- a/website/static/css/style.css
+++ b/website/static/css/style.css
@@ -715,6 +715,12 @@ div.ball-scale-blue>div {
     overflow: scroll;
 }
 
+/* Add Links Modal
+----------------------- */
+.node-dates {
+    white-space: pre;
+}
+
 /* Password strength check component colors for darker backgrounds  */
 .pv-darkbg .text-danger {
     color: #f97773;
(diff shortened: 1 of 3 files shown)


mluke93:
Tests seemed to work locally, so I want to make sure they pass here too. Also, I am still in need of a Product Demo.

mluke93:
Met with Sara and did a Product Demo. We decided on a few Visual changes. These are now updated in the PR.

PR #6138, opened 2016-08-10
Feature/api reorder contributors

brianjgeiger:
## Purpose

Allow API v2 consumers to re-order contributors.
 
## Changes

1. Tests
2. Add a method on the node model to do single-contributor moves
3. Modify the API so that it can accept a re-order request and pass that along to the node.

## Side effects

Should be fairly isolated.

## Ticket

https://openscience.atlassian.net/browse/OSF-6832

### DIFF
--- a/api/nodes/serializers.py
+++ b/api/nodes/serializers.py
@@ -543,6 +543,7 @@ class NodeContributorsSerializer(JSONAPISerializer):
 
     id = ContributorIDField(read_only=True, source='_id')
     type = TypeField()
+    index = ser.IntegerField(required=False, read_only=True)
 
     bibliographic = ser.BooleanField(help_text='Whether the user will be included in citations for this node or not.',
                                      default=True)
@@ -579,6 +580,7 @@ class NodeContributorsSerializer(JSONAPISerializer):
         if unclaimed_records:
             return unclaimed_records.get('name', None)
 
+
 class NodeContributorsCreateSerializer(NodeContributorsSerializer):
     """
     Overrides NodeContributorsSerializer to add target_type and required id field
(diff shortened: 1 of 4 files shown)


api/nodes/serializers.py
```diff
@@ -602,21 +604,33 @@ def create(self, validated_data):
...
     Overrides node contributor serializer to add additional methods
     """
     id = ContributorIDField(required=True, source='_id')
+    index = ser.IntegerField(required=False, read_only=False)
+    # index = ser.IntegerField(required=False)
 
     def update(self, instance, validated_data):
+        index = None
+        if 'index' in validated_data:
+            index = validated_data.pop('index')
```
mfraezz: Why `.pop`? `index = validated_data.get('index')` should do the same as these three lines.

website/project/model.py
```diff
@@ -2935,6 +2935,14 @@ def remove_contributors(self, contributors, auth=None, log=True, save=False):
 
         return all(results)
 
+    def move_contributor(self, user, auth, index, save=False):
```
mfraezz: Currently, `Node.manage_contributors` is used to reorder contributors, which also adds a `contributors_reordered` `NodeLog`. This should probably also add one of those logs.

However, `manage_contributors` allows for bulk reordering. Should this endpoint also?

brianjgeiger: We can have an improvement to do a relationship self link for managing just the order of contributors. No reason to enforce that here.

Definitely need a log, though.

api/nodes/serializers.py
brianjgeiger: Yeah, it's not strictly necessary here. Philosophically, it's because it's not a real value on the contributor, so if someone just passed validated_data as kwargs to update the contributor, it could break. As it stands, it's fine, since we aren't doing that.

mfraezz:
Changes look good, dropping this in PCR ⛄

pattisdr:
@brianjgeiger thanks for this - easy to get working with ember-preprints. Sending in the new index is a good way to solve this!

PR #6140, opened 2016-08-11
Old meetings' is_meeting field default to true [#OSF-5645]

caspinelli:
## Purpose
Make it so that old meetings' is_meeting field defaults to true so that when editing an older meeting you don't accidently submit the form without setting is_meeting to true

## Changes
Changed serializer to set is_meeting to true is it is none

## Side effects
None

## Ticket

https://openscience.atlassian.net/browse/OSF-5645

### DIFF
--- a/admin/meetings/serializers.py
+++ b/admin/meetings/serializers.py
@@ -2,7 +2,7 @@
 
 def serialize_meeting(meeting):
     is_meeting = True
-    if hasattr(meeting, 'is_meeting'):
+    if meeting.is_meeting is not None:
         is_meeting = meeting.is_meeting
     return {
         'endpoint': meeting.endpoint,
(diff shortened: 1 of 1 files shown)


admin/meetings/serializers.py
```diff
@@ -2,7 +2,7 @@
 
 def serialize_meeting(meeting):
     is_meeting = True
-    if hasattr(meeting, 'is_meeting'):
+    if meeting.is_meeting is not None:
```
brianjgeiger: Are there any situations in which this would cause an exception (such as if meeting doesn't have an is_meeting attr)? Should it be `if hasattr(meeting, 'is_meeting') and meeting.is_meeting is not None:`?

brianjgeiger:
Pass done. One potential change. Write a test for the scenario and change it if it needs to be (unless that situation can't come up, in which case it's fine). 🐧

brianjgeiger:
Looks great. 🎉 🐧 🎉 I'll merge after tests pass.

brianjgeiger:
The only failing test was the js test on the second branch and this is a simple enough change. I'm gonna live on the edge and merge it.

PR #6144, opened 2016-08-11
Unable to access all pages of Meetings in Admin App[#OSF-6475]

chennan47:
## Purpose
Fix the problem that admin for OSF for meetings cannot visit the second page.

## Changes
Remove unconfirmed user from OSF for meetings admin serializer.

## Ticket

https://openscience.atlassian.net/browse/OSF-6475

### DIFF
--- a/admin/meetings/serializers.py
+++ b/admin/meetings/serializers.py
@@ -10,7 +10,7 @@ def serialize_meeting(meeting):
         'info_url': meeting.info_url,
         'logo_url': meeting.logo_url,
         'active': meeting.active,
-        'admins': ', '.join([u.emails[0] for u in meeting.admins]),
+        'admins': ', '.join([u.emails[0] for u in meeting.admins if u.is_confirmed]),
         'public_projects': meeting.public_projects,
         'poster': meeting.poster,
         'talk': meeting.talk,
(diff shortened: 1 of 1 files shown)


brianjgeiger:
Code looks fine. It would be good to have a little more documentation that this was the problem (link to the sentry error or whatever) so that QA knows how to test properly (and so I know what was wrong).

 It looks like, ultimately, the problem is that an unconfirmed user was made an admin and that made problems for the serializer. Questions related are:

- Why are unconfirmed users able to be added as admins?
- _Should_ unconfirmed users be able to added as admins?
- If not, should we fix that or have a ticket for it?
- If so, should we serialize those users properly?

brianjgeiger:
Pass done. Just some questions that need answering. 🐧

chennan47:
The sentry log for this issue is https://sentry.cos.io/sentry/osf/issues/6522/.  An unconfirmed user is an admin of a meeting which breaks the admin-meeting-serializer and caused the admin meeting's second page return a 502. 
For Brian's question, I believe those admins not necessary are OSF users before those meetings which means they don't necessary have an account and I believe they are added as admin by community managers. For more correctly serialize those user, I add back those unconfirmed user to the serializer and show their primary email for the admins who want to view those meetings have a better sense of the more accurate information.

PR #6145, opened 2016-08-11
API Multi Field Filtering [OSF-6851]

caseyrollins:
#### Purpose
- Allows for multi-field filtering in the API, i.e. ```/v2/users/?filter[given_name, family_name]=roll``` will filter users whose ```given_name``` **OR** ```family_name``` contains 'roll'.

#### Changes
- Updated ```QUERY_PATTERN``` regex in the ```FilterMixin``` class (note: I'm not 100% confident in this regex). 
- Changes to ```parse_query_params()``` in ```FilterMixin``` and the format of the query dict that is returned. 

#### Notes
- The multiple fields included in the filter must be the same type, i.e. ```/v2/nodes/?filter[title, public]=science``` will give an error. 
- Multi-field filters can be combined with other multi-field or single-field filters.
 - ```/v2/nodes/?filter[title,description]=music&filter[public]=true``` will filter by:
   - ```(title contains 'music' OR description contains 'music') AND (public equals true)```
 - ```/v2/nodes/?filter[title,description]=music&filter[category,description]=hypothesis``` will filter by:
   - ```(title contains 'music' OR description contains 'music') AND (category contains 'hypothesis' OR description contains 'hypothesis')```

#### Ticket
- [OSF-6851](https://openscience.atlassian.net/browse/OSF-6851)

### DIFF
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -60,7 +60,7 @@ class ODMOrderingFilter(OrderingFilter):
 class FilterMixin(object):
     """ View mixin with helper functions for filtering. """
 
-    QUERY_PATTERN = re.compile(r'^filter\[(?P<field>\w+)\](\[(?P<op>\w+)\])?$')
+    QUERY_PATTERN = re.compile(r'^filter\[(?P<fields>((?:,*\s*\w+)*))\](\[(?P<op>\w+)\])?$')
 
     MATCH_OPERATORS = ('contains', 'icontains')
     MATCHABLE_FIELDS = (ser.CharField, ser.ListField)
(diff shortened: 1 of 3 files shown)


api/base/filters.py
```diff
@@ -60,7 +60,7 @@ def filter_queryset(self, request, queryset, view):
 class FilterMixin(object):
     """ View mixin with helper functions for filtering. """
 
-    QUERY_PATTERN = re.compile(r'^filter\[(?P<field>\w+)\](\[(?P<op>\w+)\])?$')
+    QUERY_PATTERN = re.compile(r'^filter\[(?P<fields>((?:,*\s*\w+)*))\](\[(?P<op>\w+)\])?$')
```
brianjgeiger: I'm not a regex master, so maybe isolate this regex and run it through some unit tests. Or we can just make sure that the view/filter tests cover it, but if we want to be super-thorough about this regex, the fastest way is to make a non-osf test case series of tests that just gets thrown at this regex. I think that would be a wise plan.

brianjgeiger:
This looks great. Might as well do some regex tests in the super-fast TestCase (rather than OSFTestCase), and let's do one set of integrated view tests with in user search. Pass done. 🐧

brianjgeiger:
Awesome. RTM when tests pass. 🎉 🐧 🎉

PR #6149, opened 2016-08-12
Fix admin app meetings form 'Homepage link text' [#OSF-6854]

caspinelli:
## Purpose

Fix bug where even if a meeting has a homepage link text it does not show up in the form and you cannot submit the form without entering another one.

## Changes

Fixed bug mentioned above so now if a meeting has a homepage link text it will appear in the field.
Also made it so that if you do not enter a homepage link text it will default to 'Conference homepage' so that field is no longer required to submit the form.

## Side effects

None that I can see 

## Ticket

https://openscience.atlassian.net/browse/OSF-6854


brianjgeiger:
🎉 🐧 🎉

PR #6152, opened 2016-08-12
Add check for too meny requests when adding contributors via API [OSF-6880]

TomBaxter:
## Purpose:
[OSF-6880](https://openscience.atlassian.net/browse/OSF-6880)
Limit contributor adds to 1 per 6 seconds. Email spam mitigation.

## Changes:
Update  api/nodes/serializers.py add catch for TooManyRequests exception
Update  website/exceptions.py  add TooManyRequests exception
Update website/project/model.py add def throttle_add_contributor for API add contributor request
Update website/settings/defaults.py add API_SEND_EMAIL_THROTTLE = 6

## Side effects
None

[#OSF-6880]

### DIFF
--- a/website/project/model.py
+++ b/website/project/model.py
@@ -42,10 +42,12 @@ from website.util import web_url_for
 from website.util import api_url_for
 from website.util import api_v2_url
 from website.util import sanitize
+from website.util.time import throttle_period_expired
 from website.exceptions import (
     NodeStateError,
     InvalidTagError, TagNotFoundError,
     UserNotAffiliatedError,
+    TooManyRequests
 )
 from website.institutions.model import Institution, AffiliatedInstitutionsList
 from website.citations.utils import datetime_to_csl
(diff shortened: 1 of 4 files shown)


website/project/model.py
```diff
@@ -3070,6 +3072,20 @@ def manage_contributors(self, user_dicts, auth, save=False):
             if to_remove or permissions_changed and ['read'] in permissions_changed.values():
                 project_signals.write_permissions_revoked.send(self)
 
+    def throttle_add_contributor(self, contributor, permissions=None,
+                                 visible=True, auth=None, log=True, save=False):
+
+        if not throttle_period_expired(auth.user.email_last_sent,
```
brianjgeiger: What happens here if auth isn't sent in? Probably need to add an `auth is not None` to the beginning of that or decorate the function with whatever decorators we have to ensure that the user has permission to do this.

PR #6154, opened 2016-08-13
[#OSF-6892][Hotfix] Delete, Check Out, and Download buttons missing from certain file types

chennan47:
## Purpose
Fix the problem that "delete", "check out", and "download" buttons are missing from certain file types. Now those buttons will show up on all files on latest version and won't show up for any old version files.

## Changes
Before changes:
screen shot 2016-08-13 at 5 29 45 pm

After change:
for latest version of file cannot be edited on OSF, the delete, checkout, download button still show up
screen shot 2016-08-13 at 5 30 33 pm

for previous version of file cannot be edited on OSF, the delete, checkout, download button will NOT show up
screen shot 2016-08-13 at 5 30 43 pm

## Side effects

I think the download button should still be there but with changes to make it to download the current version instead of the most recent version. This can be another improvement ticket.

## Ticket

https://openscience.atlassian.net/browse/OSF-6892

### DIFF
--- a/website/static/js/filepage/index.js
+++ b/website/static/js/filepage/index.js
@@ -108,6 +108,7 @@ var FileViewPage = {
         self.editorMeta = self.context.editor;
         self.file.checkoutUser = null;
         self.requestDone = false;
+        self.isLatestVersion = false;
         self.isCheckoutUser = function() {
             $.ajax({
                 headers: {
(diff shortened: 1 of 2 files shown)


brianjgeiger:
Looks great. Thanks, Nan! 🎉 🐧 🎉

PR #6157, opened 2016-08-15
[WIP] Add Unregistered Contributors via API [OSF-6855]

caseyrollins:
#### Ticket
- [OSF-6855](https://openscience.atlassian.net/browse/OSF-6855)


api/nodes/serializers.py
```diff
@@ -586,27 +586,55 @@ def get_unregistered_contributor(self, obj):
...
-            raise exceptions.ValidationError('{} is already a contributor'.format(contributor.fullname))
-
-        bibliographic = validated_data['bibliographic']
+        auth = Auth(self.context['request'].user)
+        full_name = validated_data.get('full_name')
+        bibliographic = validated_data.get('bibliographic')
         permissions = osf_permissions.expand_permissions(validated_data.get('permission')) or osf_permissions.DEFAULT_CONTRIBUTOR_PERMISSIONS
-        node.add_contributor(contributor=contributor, auth=auth, visible=bibliographic, permissions=permissions, save=True)
+
+        if not id and not full_name:
```
brianjgeiger: If you want to clean this up a bit, I'd lean towards:

1. Make a validation function that ensures that you have some combination of the correct information and, if not, raises the appropriate error.
2. Make a model method that will create a contributor regardless of which way it gets made and have that do the creation logic. Something like `add_contributor_registered_or_not(user_id, full_name, email, auth, bibliographic, permissions, save)`. Then most of this function gets reduced to "validate; create;" rather than heavy logic. The validation tests can happen as TestCase unit tests (which are fast) rather than as view tests, and you can do at most one view test to verify that your validation function is being used. 

Your model method will probably be a little easier to write like that is my guess. Maybe not a *lot* easier, but.

brianjgeiger:
I think you're going to have to roll https://openscience.atlassian.net/browse/OSF-6880 and https://openscience.atlassian.net/browse/OSF-6895 into this to do it right. It doesn't make sense to do these in isolation. Plus Tom's on vacation for the week, so might as well. 

If we do that, we should make sure that you have to have the special user-create scope (CoreScopes.USERS_CREATE) attached to this endpoint.

- [ ] Ensure CoreScopes.USERS_CREATE is part of the POST endpoint
- [ ] Add Tom's PR for OSF-6880 to this, fix the merge conflict, and respond to the code review
- [ ] Take a look at Tom's WIP PR for OSF-6895 and see about adding/finishing that. If we do `?send_email=` with one of `false`, `default`, or `preprint`, then we match the send email query parameter we're using for user create but with some extra options.
- [ ] If `send_email=false`, then don't throttle the user creation

I think this is all we need to finish this out, probably.

brianjgeiger:
Pass done. 🐧

api_tests/nodes/views/test_node_contributors_list.py
```diff
@@ -811,15 +804,263 @@ def test_adds_contributor_private_project_not_logged_in(self):
...
+                }
+            }
+        }
+        res = self.app.post_json_api(self.public_url, payload, auth=self.user.auth, expect_errors=True)
+        self.public_project.reload()
+        assert_equal(res.status_code, 400)
+        assert_equal(res.json['errors'][0]['detail'], 'Alphabet is already a contributor.')
+
+
+class TestNodeContributorCreateValidation(TestCase):
```
brianjgeiger: Look at all these speedy tests!

api_tests/nodes/views/test_node_contributors_list.py
```diff
@@ -811,15 +804,263 @@ def test_adds_contributor_private_project_not_logged_in(self):
...
+                            'id': self.user_two._id
+                        }
+                    }
+                }
+            }
+        }
+
+        with capture_signals() as mock_signal:
+            res = self.app.post_json_api(url, payload, auth=self.user.auth)
+        assert_in(contributor_added, mock_signal.signals_sent())
```
brianjgeiger: You miiiight be able to use assert_called_with to ensure that the signal was sent with the right info. https://docs.python.org/3/library/unittest.mock.html#quick-guide  Worth a quick check.

website/project/model.py
```diff
@@ -3219,11 +3221,50 @@ def add_unregistered_contributor(self, fullname, email, auth,
...
         self.add_contributor(
             contributor, permissions=permissions, auth=auth,
-            log=True, save=False,
+            send_email=send_email, log=True, save=False
         )
         self.save()
         return contributor
 
+    def add_contributor_registered_or_not(self, auth, user_id=None, full_name=None, email=None,
+                                          send_email='default', permissions=None, bibliographic=True, save=False):
```
brianjgeiger: If this is only being called by the API, probably best to `send_email='false'` by default.

brianjgeiger:
This looks great, and Dawn seems pretty happy with it as well. Check out that one potential update to the mock assert, but I wouldn't be surprised if it didn't work the way we want. Awesome work. 🐧

pattisdr:
Yes @caseyrollins, good work. One question I have, this puts limits on rate of any type of contributor I want to add, registered or not, correct?  If a contributor is already a registered user, should we be putting a limit on this?

brianjgeiger:
It should only rate limit if an email is sent. If it sends an email, it needs to be rate limited. If it doesn't, then it shouldn't be.

brianjgeiger:
Also, we haven't considered the case of bulk adding of contributors, but that's a question for another day, hopefully. If you're going to do bulk contributor adding, let me know.

caseyrollins:
@pattisdr, @brianjgeiger is correct. There is no limit for adding contributors, but there is a rate limit for sending the emails that will notify added contributors. If the emails aren't necessary, you can add `send_email=false` as a query parameter and avoid the limit.

pattisdr:
Ah, I see, thanks.  Still need to figure out what is happening with emails in preprints.  When I am adding users to my preprints, I probably don't want them to get an email until my preprint is finalized, but that is an issue on our end.  I will use the query parameter.

Also, not doing bulk contributor adding in preprints, so not an issue at this time.

brianjgeiger:
Looks great! 🎉 🐧 🎉

PR #6164, opened 2016-08-17
[WIP][Feature] preprints

icereval:


api/preprints/serializers.py
```diff
@@ -0,0 +1,147 @@
...
+
+        try:
+            node.set_preprint_file(primary_file, auth, save=False)
+        except PermissionsError:
+            raise exceptions.PermissionDenied('Not authorized to create a preprint from this node.')
+        except ValueError as e:
+            raise exceptions.ValidationError(detail=e.message)
+
+        if node._id != validated_data.pop('_id'):
+            raise exceptions.ValidationError('The node id in the URL does not match the id in the request body.')
```
mfraezz: This is a red flag that maybe we shouldn't be architecting these endpoints this way. Typically, `List` endpoints accept `POST`s to create, not `Detail` endpoints, so there shouldn't be a node id in the url for a call to `.create`

api/preprints/views.py
```diff
@@ -0,0 +1,197 @@
...
+            self.kwargs[self.node_lookup_url_kwarg],
+            display_name='preprint'
+        )
+        if not node.is_preprint and self.request.method != 'POST':
+            raise NotFound
+
+        return node
+
+
+class PreprintList(JSONAPIBaseView, generics.ListAPIView, ODMFilterMixin):
```
mfraezz: As mentioned above, this should perhaps be a `ListCreateAPIView`, rather than having the `Detail` view perform the create. If this is too big of an architectural change at this point, please include something like `"Under no circumstances should anyone expect this route to be stable"` in the docstring (of bot this and `PreprintDetail`).

api/preprints/views.py
```diff
@@ -0,0 +1,197 @@
...
+    ordering = ('-date_created')
+    view_category = 'preprints'
+    view_name = 'preprint-list'
+
+    # overrides ODMFilterMixin
+    def get_default_odm_query(self):
+        return (
+            Q('preprint_file', 'ne', None) &
+            Q('is_deleted', 'ne', True) &
+            Q('preprint_file', 'ne', None) &
```
mfraezz: Duplicate query part (see two lines up).

api/preprints/views.py
```diff
@@ -0,0 +1,197 @@
...
+    Most are string fields and will be filtered using simple substring matching.
+
+    #This Request/Response
+    """
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        base_permissions.TokenHasScope,
+    )
+
+    required_read_scopes = [CoreScopes.NODE_BASE_READ]
```
mfraezz: Open question (possibly at scope-master @abought ) -- While I know Preprints are just a kind of node, should these have their own specific `NODE_PREPRINTS_READ` or `NODE_PREPRINTS_WRITE` scope (much like institutions and registrations)? In other words, are there cases where we would want to grant a third-party access to only preprints or only nodes, rather than both?

api/taxonomies/serializers.py
```diff
@@ -0,0 +1,23 @@
...
+    filterable_fields = frozenset([
+        'text',
+        'parent_ids',
+        'type',
+        'id'
+    ])
+
+    type = ser.CharField(max_length=200)
+    text = ser.CharField(max_length=200)
+    parent_ids = ser.ListField(child=ser.CharField(max_length=50))
```
mfraezz: I haven't gotten that far yet, but based on the way taxonomies are usually formed, I'm assuming that A) a `Taxonomy` object is intended to represent a leaf in a tree-structured representation of possible taxonomic classifications for Preprints or files therein and that B) this field will be pointing to `_id` fields of other `Taxonomy` objects? 

If so, either this field is about 150 characters to short or the definition on the line below is 150 characters too long.

api/preprints/views.py
abought: That's a good question- and a mite tricky, as at last check you could see the full data for a preprint by asking for `v2/nodes/<preprint_guid>`.

(Eg preprints were under both endpoints, so having `nodes` alone would give access to data about the preprint node too)

My instincts are as follows:
- Scopes are about data, and preprints have a different serializer. Give them their own CoreScope to reflect this difference.
- For now bundle them in existing endpoints
- *The biggest thing to protect separately would be the preprints write scope*: letting someone edit a page on your behalf should not automatically allow them to submit that for irrevocable(ish) publication.
- As for read scope, most data overlaps with nodes. The main edge case I see where we might want to protect preprint data separately as confidential is if we create a category of preprints that were submitted, but not accepted.
  - Such functionality has been marked explicitly out of scope for the 1.0 release at the end of the month, so protecting read this way is not an immediate priority.

Regardless, all these permissions will end up in osf.full_read/write... at some point it's worth revisiting more granular scopes.

api/users/views.py
```diff
@@ -505,6 +506,31 @@ def get_queryset(self):
         return Node.find(self.get_query_from_request())
 
 
+class UserPreprints(UserNodes):
+    required_read_scopes = [CoreScopes.USERS_READ]
```
mfraezz: Why is this overriding to not require some sort of node or preprint read scope?

abought: Oh, yeah. This says "if you can see profile data, you can also see data about preprints".

If we do reuse scopes, generally we should use the scope of a route that has the same serializer (same kind of data).

api/users/views.py
```diff
@@ -505,6 +506,31 @@ def get_queryset(self):
...
+            Q('contributors', 'eq', user._id) &
+            Q('preprint_file', 'ne', None) &
+            Q('is_public', 'eq', True)
+        )
+
+        return query
+
+    def get_queryset(self):
+        nodes = Node.find(self.get_query_from_request())
+        return [node for node in nodes if node.is_preprint]
```
mfraezz: Not a big deal, but this looks inefficient -- evaluating all the nodes takes a lot longer than returning a queryset, and you'll have to do so again to serialize each one. Is it not possible to construct a query guaranteed to return only preprints?

api_tests/nodes/views/test_node_list.py
```diff
@@ -249,6 +249,19 @@ def test_filtering_tags(self):
         assert_in(self.project_one._id, ids)
         assert_not_in(self.project_two._id, ids)
 
+    def test_filter_no_tags(self):
```
mfraezz: I'm not one to complain about improved test coverage, but this seems out of scope. Are preprints identified by tags in the same way that meetings are, or something similar?

api_tests/preprints/views/test_preprint_detail.py
```diff
@@ -0,0 +1,271 @@
...
+    def test_preprint_detail_success(self):
+        assert_equal(self.res.status_code, 200)
+        assert_equal(self.res.content_type, 'application/vnd.api+json')
+
+    def test_preprint_top_level(self):
+        assert_equal(self.data['type'], 'preprints')
+        assert_equal(self.data['id'], self.preprint._id)
+
+
+def create_file(node, filename):
```
mfraezz: This seems like it should go elsewhere, e.g. `utils.py`, and maybe implemented as a `PreprintFileFactory` instead of a helper? Not a blocker.

Also, just discovered that there's already a `create_test_file` helper in `api_tests/utils.py`

website/project/model.py
```diff
@@ -1149,6 +1177,31 @@ def has_active_draft_registrations(self):
...
 
+    @property
+    def is_preprint(self):
+        if not self.preprint_file:
+            return False
+        if self.preprint_file.node == self:
+            return True
+        else:
+            self.preprint_file = None
+            self._is_preprint_orphan = True
```
mfraezz: Unless you save after checking `.is_preprint`, these field changes won't persist. You don't typically want to be calling `self.save` inside of a property, but this might be an exception to that rule.

website/project/taxonomies/__init__.py
```diff
@@ -0,0 +1,18 @@
+from modularodm import fields
+
+from framework.mongo import (
+    ObjectId,
+    StoredObject,
+    utils as mongo_utils
+)
+
+
+@mongo_utils.unique_on(['id', '_id'])
```
mfraezz: ...does `Subject` have a `.id` field? This is a reserved word in python, which is why we use `_id` everywhere

website/project/model.py
```diff
@@ -1493,6 +1546,40 @@ def set_node_license(self, license_id, year, copyright_holders, auth, save=False
...
+
+        if preprint_file.node != self or preprint_file.provider != 'osfstorage':
+            raise ValueError('This file is not a valid primary file for this preprint.')
+
+        # there is no preprint file yet! This is the first time!
+        if not self.preprint_file:
+            self.preprint_file = preprint_file
+            self.preprint_created = datetime.datetime.utcnow()
+            self.add_log(action=NodeLog.PREPRINT_INITIATED, params={}, auth=auth, save=False)
+        else:
```
chrisseto: `elif`

website/project/taxonomies/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+class Subject(StoredObject):
+    _id = fields.StringField(primary=True, default=lambda: str(ObjectId()))
+    type = fields.StringField(required=True)
+    text = fields.StringField(required=True)
+    parent_ids = fields.StringField(list=True)
+
+    def get_absolute_url(self):
+        return '{}taxonomies/?filter[id]={}'.format(self.absolute_api_v2_url, self._id)
```
chrisseto: Shouldn't this just be `taxonomies/{}/`

website/project/model.py
```diff
@@ -755,6 +760,20 @@ def __init__(self, reason, key, *args, **kwargs):
...
+    subject = Subject.load(value)
+    if not subject:
+        raise ValidationValueError('Subject with id <{}> could nor be found.'.format(value))
+    return True
+
+
+def validate_doi(value):
+    # DOI must start with 10 and have a slash in it - avoided getting too complicated
+    if not re.match('10\\.\\S*\\/', value):
+        raise ValidationValueError('')
```
chrisseto: Add an actual errormessage

api/taxonomies/views.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Only the top three levels of the PLOS taxonomy are included.
+    '''
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        base_permissions.TokenHasScope
+    )
+
+    DEFAULT_OPERATOR_OVERRIDES = {
+        ser.CharField: 'icontains',
```
chrisseto: Is this being used by anything thing? If not I'd prefer not to allow random regexes.

api/preprints/views.py
```diff
@@ -0,0 +1,197 @@
...
+            Q('preprint_file', 'ne', None) &
+            Q('is_public', 'eq', True)
+        )
+
+    # overrides ListAPIView
+    def get_queryset(self):
+        query = self.get_query_from_request()
+        nodes = Node.find(query)
+
+        return [node for node in nodes if node.is_preprint]
```
chrisseto: At the very least this should be a generator. This will load all of the user's preprints into memory every request, regardless of page size.

api/preprints/serializers.py
```diff
@@ -0,0 +1,147 @@
...
+from api.base.utils import absolute_reverse, get_user_auth
+from api.nodes.serializers import NodeTagField
+from framework.exceptions import PermissionsError
+from website.models import StoredFileNode
+
+
+class PrimaryFileRelationshipField(RelationshipField):
+    def get_object(self, file_id):
+        try:
+            return StoredFileNode.find_one(Q('_id', 'eq', file_id))
```
chrisseto: Any reason not to use `load`?

website/project/taxonomies/__init__.py
mfraezz: Detail view didn't exist, just added it.

api/taxonomies/views.py
mfraezz: That's already the default. It's the `'eq'` below that's actually overriding

PR #6166, opened 2016-08-17
PreprintProvider Model and API Endpoints for Preprints [#PREP-108]

erinspace:
## Script to Run!
Run `python -m scripts.populate_preprint_providers` to populate the `PreprintProviders` model with data you can use to affiliate providers

## Purpose
Previously, the 'preprint_provider' field on the Node was a string field, which stores a simple name for the provider. This PR will instead allow us to use a more flexible foreign key to a `PreprintProvider` model that stores more metadata about the preprint provider for branding.

## Changes
- PreprintProvider model in `model.py` along with a foreignkey relationship on the Node with `preprint_provider` replaces the existing stringfield
- API v2 views and serializers for the PreprintProvider model
- API v2 serializer and views for initiating the Preprint > PreprintProvider Relationship
- API v2 URLs for preprnt provider and preprint provider relationships
- Script to populate PreprintProvider model with relevant IDs (shorttened lowercase names of the providers) and logo path information
- Updates to `create_fakes` and `PreprintFactory` to now take a new pre-existing `PreprintProvider`, or use `PreprintProviderFactory` to create fakes for testing
- Tests for the relationship and detail views for preprint provider as well as updates to existing tests

## Side effects
- Will need to run script `populate_preprint_providers.py` to populate the model correctly before using anything

## Questions for CR or anyone, really
Does anyone have a better name than "Preprint Providers"?? It makes the Relationship Serializers and some views have really terrible names if I try to follow what I found as convention (See `PreprintPreprintProviderRelationshipSerializer` and the like...)

## Ticket
https://openscience.atlassian.net/browse/PREP-108


api/preprints/serializers.py
```diff
@@ -143,5 +149,41 @@ class PreprintDetailSerializer(PreprintSerializer):
...
     subjects = JSONAPIListField(required=False, source='preprint_subjects')
 
+
 class PreprintDetailRetrieveSerializer(PreprintDetailSerializer):
     subjects = JSONAPIListField(required=False, source='get_preprint_subjects')
+
+
+class PreprintPreprintProviderRelationshipSerializer(ser.Serializer):
+    id = ser.CharField(source='_id', required=False, allow_null=True)
+    type = TypeField(required=False, allow_null=True)
```
mfraezz: This isn't JSONAPI compliant; they should both be required and not `allow_null` 

>A resource object **MUST** contain at least the following top-level members:
>
> * id
> * type

\- http://jsonapi.org/format/#document-resource-objects

api/preprint_providers/views.py
```diff
@@ -0,0 +1,189 @@
...
+
+    #This Request/Response
+
+    """
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        base_permissions.TokenHasScope,
+    )
+
+    required_read_scopes = [CoreScopes.INSTITUTION_READ]
```
mfraezz: This read scope seems incorrect. Others use `NODE_BASE_READ` or `NODE_FILE_READ`, but I'm not certain that either of those are correct either. If we want to restrict read access to these views, consider adding a new read scope, e.g. `PREPRINT_READ`. If there isn't a reason to, `CoreScopes.ALWAYS_PUBLIC` is probably the right scope. I don't know why we would want to not tell people what preprint providers we have, but there might be a good reason.

api/preprint_providers/views.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+class PreprintProviderMixin(object):
+    institution_lookup_url_kwarg = 'institution_id'
+
+    def get_preprintprovider(self):
+        provider = get_object_or_error(
+            PreprintProvider,
+            self.kwargs[self.institution_lookup_url_kwarg],
+            display_name='institution'
```
mfraezz: Are we planning on deleting preprint providers, and if so should the response for requesting one be `410 GONE - "The requested institution is no longer available."?

api/preprint_providers/views.py
```diff
@@ -0,0 +1,189 @@
...
+from api.base.views import JSONAPIBaseView
+from api.base.utils import get_object_or_error
+from api.base.pagination import MaxSizePagination
+
+from api.preprint_providers.serializers import PreprintProviderSerializer
+from api.preprints.serializers import PreprintSerializer
+
+
+class PreprintProviderMixin(object):
+    institution_lookup_url_kwarg = 'institution_id'
```
mfraezz: This should probably be `preprintprovider_lookup_url_kwarg = 'preprintprovider_id'`

api/preprint_providers/views.py
```diff
@@ -0,0 +1,189 @@
...
+
+    pagination_class = MaxSizePagination
+    serializer_class = PreprintProviderSerializer
+    view_category = 'preprint_providers'
+    view_name = 'preprint_providers-list'
+
+    ordering = ('name', )
+
+    def get_default_odm_query(self):
+        return Q('_id', 'ne', None)
```
mfraezz: `Q('is_deleted', 'ne', True)`?

api/preprint_providers/views.py
```diff
@@ -0,0 +1,189 @@
...
+    view_category = 'preprints'
+    view_name = 'preprints-list'
+
+    # overrides ODMFilterMixin
+    def get_default_odm_query(self):
+        provider = PreprintProvider.find_one(Q('_id', 'eq', self.kwargs['provider_id']))
+        return (
+            Q('preprint_file', 'ne', None) &
+            Q('is_deleted', 'ne', True) &
+            Q('preprint_file', 'ne', None) &
```
mfraezz: Duplicate query part

website/project/model.py
```diff
@@ -891,7 +891,7 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin, Commentable):
     # Preprint fields
     preprint_file = fields.ForeignField('StoredFileNode')
     preprint_created = fields.DateTimeField()
-    preprint_provider = fields.StringField()
+    preprint_provider = fields.ForeignField('PreprintProvider')
```
mfraezz: As you mentioned earlier, nodes should be able to have more than one affiliated `PreprintProvider`, so this `ForeignField` should be passed a `list=True` kwarg

api/preprint_providers/views.py
```diff
@@ -0,0 +1,189 @@
...
+
+    #This Request/Response
+
+    """
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        base_permissions.TokenHasScope,
+    )
+
+    required_read_scopes = [CoreScopes.NODE_BASE_READ]
```
mfraezz: See above edited https://github.com/CenterForOpenScience/osf.io/pull/6166#discussion_r75343106

This should be `ALWAYS_PUBLIC`

api/preprint_providers/views.py
```diff
@@ -0,0 +1,189 @@
...
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        base_permissions.TokenHasScope,
+    )
+
+    ordering = ('-date_created')
+
+    serializer_class = PreprintSerializer
+
+    required_read_scopes = [CoreScopes.NODE_FILE_READ]
```
mfraezz: This should be the recently-added `CoreScopes.NODE_PREPRINTS_READ`

api/preprint_providers/views.py
```diff
@@ -0,0 +1,189 @@
...
+            Q('is_public', 'eq', True) &
+            Q('preprint_provider', 'eq', provider)
+        )
+
+    # overrides ListAPIView
+    def get_queryset(self):
+        query = self.get_query_from_request()
+        nodes = Node.find(query)
+
+        return [node for node in nodes if node.is_preprint]
```
mfraezz: Please make this a generator:
```
return (node for node in nodes if node.is_preprint)
```

This increases efficiency and doesn't break pagination without necessitating an architectural overhaul

api/preprints/views.py
```diff
@@ -190,8 +195,71 @@ def perform_create(self, serializer):
...
+                       }
+        Success:       200
+
+    This requires admin permissions in the node.
+    """
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        base_permissions.TokenHasScope,
+    )
+    required_read_scopes = [CoreScopes.NODE_BASE_READ]
```
mfraezz: These should be updated to `NODE_PREPRINT_[READ|WRITE]`

api/preprints/serializers.py
```diff
@@ -143,5 +149,41 @@ class PreprintDetailSerializer(PreprintSerializer):
...
+
+    def update(self, instance, validated_data):
+        node = instance['self']
+        auth = get_user_auth(self.context['request'])
+
+        if node.preprint_provider:
+            raise ValueError('Preprint provider is already assigned to this preprint')
+
+        preprint_provider = PreprintProvider.load(validated_data['_id'])
+        node.set_preprint_provider(preprint_provider, auth, save=True)
```
mfraezz: Is there any reason not to include this in `PreprintSerializer`'s `.create` or `.update`?

website/project/model.py
```diff
@@ -4241,3 +4250,33 @@ def validate_metadata(self, *args, **kwargs):
         Validates draft's metadata
         """
         return self.registration_schema.validate_metadata(*args, **kwargs)
+
+
+class PreprintProvider(StoredObject):
```
mfraezz: To avoid further clutter to this file, consider moving model to `website/preprints/`. Not a blocker

api/preprint_providers/serializers.py
```diff
@@ -0,0 +1,34 @@
...
+    })
+
+    class Meta:
+        type_ = 'preprint_providers'
+
+    def get_absolute_url(self, obj):
+        return obj.absolute_api_v2_url
+
+    def preprint_links(self, obj):
+        return '{}{}preprint_providers/{}/preprints/'.format(API_DOMAIN, API_BASE, obj._id)
```
mfraezz: Consider using `website.util.api_v2_url` or 
`absolute_reverse('<view_class>:<view_name>', kwargs={'<routing_key>': obj._id})`

Also, following convention, this function should be named `get_preprint_links`

api/preprint_providers/serializers.py
```diff
@@ -0,0 +1,34 @@
...
+from api.base.utils import absolute_reverse
+from api.base.serializers import JSONAPISerializer, LinksField
+
+
+class PreprintProviderSerializer(JSONAPISerializer):
+
+    filterable_fields = frozenset([
+        'name',
+        'description',
+        'id'
```
mfraezz: `id`'s shouldn't really be filterable fields - that's what a detail view is for - unless there's strong use case for it. Does `ember-preprints` need this?

api/preprint_providers/views.py
```diff
@@ -0,0 +1,176 @@
...
+
+    pagination_class = MaxSizePagination
+    serializer_class = PreprintProviderSerializer
+    view_category = 'preprint_providers'
+    view_name = 'preprint_providers-list'
+
+    ordering = ('name', )
+
+    def get_default_odm_query(self):
+        return Q('_id', 'ne', None) & Q('is_deleted', 'ne', True)
```
mfraezz: I think `_id` should never be `None`... if so, the first part of this `Q` is superfluous

api/preprints/serializers.py
```diff
@@ -147,3 +154,88 @@ def set_node_field(self, func, val, auth):
             raise exceptions.PermissionDenied('Not authorized to update this node.')
         except ValueError as e:
             raise exceptions.ValidationError(detail=e.message)
+
+
+class PreprintDetailSerializer(PreprintSerializer):
```
mfraezz: This class and `PreprintDetailRetrieveSerializer` are now unused, please remove them

api/preprints/views.py
```diff
@@ -227,3 +232,145 @@ class PreprintContributorsList(NodeContributorsList, PreprintMixin):
...
     serializer_class = NodeContributorsSerializer
+
+
+class PreprintPreprintProvidersList(JSONAPIBaseView, generics.ListAPIView, ODMFilterMixin, NodeMixin):
+    """ Detail of the preprint providers a preprint has, if any. Returns [] if the preprint has no
+    preprnt providers.
+
+    ##Attributes
+
+    OSF Institutions have the "institutions" `type`.
```
mfraezz: These docs should be updated

api/preprints/views.py
```diff
@@ -227,3 +232,145 @@ class PreprintContributorsList(NodeContributorsList, PreprintMixin):
...
+    required_write_scopes = [CoreScopes.NULL]
+    serializer_class = PreprintProviderSerializer
+
+    model = PreprintProvider
+    view_category = 'preprints'
+    view_name = 'preprint-preprint_providers'
+
+    def get_queryset(self):
+        node = self.get_node()
+        return node.preprint_providers or []
```
mfraezz: `[]` is the default for a `ForeignField(list=True)`; this ternary is unnecessary

api/preprints/views.py
```diff
@@ -227,3 +232,145 @@ class PreprintContributorsList(NodeContributorsList, PreprintMixin):
...
+                       }
+        Success:       204
+
+    All of these methods require admin permissions in the preprint.
+    """
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        base_permissions.TokenHasScope,
+    )
+    required_read_scopes = [CoreScopes.ALWAYS_PUBLIC]
```
mfraezz: I'm not sure that this should be `ALWAYS_PUBLIC` -- it contains mutable and possibly personally identifiable information. Consider `NODE_PREPRINTS_READ` instead

api/preprints/views.py
```diff
@@ -227,3 +232,145 @@ class PreprintContributorsList(NodeContributorsList, PreprintMixin):
...
+    required_write_scopes = [CoreScopes.NODE_BASE_WRITE]
```
mfraezz: This seems like it should be using `NODE_PREPRINTS_WRITE`

website/project/model.py
mfraezz: Also, should this class have a `url` or `external_url` that points to, e.g. `socarxiv.org`?

website/project/taxonomies/__init__.py
```diff
@@ -10,6 +11,16 @@
 
 @mongo_utils.unique_on(['text'])
 class Subject(StoredObject):
+
+    __indices__ = [
```
mfraezz: This can be removed; the `unique_on` decorator does this

tests/test_preprints.py
```diff
@@ -117,3 +118,19 @@ def test_non_admin_update_file(self):
...
+    def setUp(self):
+        super(TestPreprintProviders, self).setUp()
+        self.preprint = PreprintFactory(provider=None)
+
+    def test_set_provider(self):
+        new_provider = PreprintProviderFactory(name='WWEArxiv')
+
+        assert_equal(self.preprint.preprint_provider, None)
+
+        self.preprint.preprint_provider = new_provider
```
mfraezz: This field name changed. Also, I think it should be calling `add_preprint_provider(new_provider, self.user, save=True)`. Please add a test for `remove_preprint_provider` too

api/preprint_providers/views.py
```diff
@@ -0,0 +1,176 @@
...
+        name           type               description
+        =========================================================================
+        name           string             title of the preprint provider
+        id             string             unique identifier in the OSF
+        logo_path      string             a path to the preprint provider's static logo
+        banner_path    string             a path to the preprint provider's static banner
+
+    ##Links
+
+        self:  the canonical api endpoint of this institution
```
mfraezz: These docs should also be updated

api/preprint_providers/views.py
```diff
@@ -0,0 +1,176 @@
...
+from api.preprints.serializers import PreprintSerializer
+
+
+class PreprintProviderList(JSONAPIBaseView, generics.ListAPIView, ODMFilterMixin):
+    """
+    Paginated list of verified PreprintProviders available
+
+    ##PreprintProvider Attributes
+
+    OSF Institutions have the "institutions" `type`.
```
mfraezz: These docs should also be updated

website/project/model.py
```diff
@@ -1585,6 +1585,23 @@ def set_preprint_file(self, preprint_file, auth, save=False):
...
+        if not self.has_permission(user, ADMIN):
+            raise PermissionsError('Only admins can update a preprint provider.')
+        if not preprint_provider:
+            raise ValueError('Must specify a provider to set as the preprint_provider')
+        self.preprint_providers.append(preprint_provider)
+        if save:
+            self.save()
+
+    def remove_preprint_provider(self, provider, user, save=False):
+        if provider in self.preprint_providers:
```
mfraezz: Should this also check permissions? Otherwise passing `user` in is superfluous

scripts/populate_preprint_providers.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def main():
+    PREPRINT_PROVIDERS = [
+        {
+            '_id': 'socarxiv',
+            'name': 'SocArXiv',
+            'description': 'Open archive of the social sciences',
+            'banner_name': 'socarxiv-banner.png',
+            'logo_name': 'socarxiv-logo.png',
```
mfraezz: These should include a key for `external_url`.

I just discovered that none of the external PrePriPros have valid SSL certs, so `http://socarxiv.org` is fine for now.

api/preprint_providers/serializers.py
```diff
@@ -0,0 +1,34 @@
...
+    ])
+
+    name = ser.CharField(required=True)
+    description = ser.CharField(required=False)
+    id = ser.CharField(max_length=200, source='_id')
+
+    logo_path = ser.CharField(read_only=True)
+    banner_path = ser.CharField(read_only=True)
+
+    links = LinksField({
```
mfraezz: This serializer should also give `external_url`, probably as an additional key in `links`

api/preprints/views.py
```diff
@@ -227,3 +232,148 @@ class PreprintContributorsList(NodeContributorsList, PreprintMixin):
...
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        base_permissions.TokenHasScope,
+    )
+
+    required_read_scopes = [CoreScopes.NODE_PREPRINTS_READ]
+    required_write_scopes = [CoreScopes.NULL]
+    serializer_class = PreprintProviderSerializer
+
+    model = PreprintProvider
```
mfraezz: Is this used?

api/preprints/views.py
```diff
@@ -82,7 +87,7 @@ class PreprintList(JSONAPIBaseView, generics.ListCreateAPIView, ODMFilterMixin):
 
     + `filter[<fieldname>]=<Str>` -- fields and values to filter the search results on.
 
-    Preprints may be filtered by their `id`, `title`, `public`, `tags`, `date_created`, `date_modified`, `provider`, and `subjects`
+    Preprints may be filtered by their `id`, `title`, `public`, `tags`, `date_created`, `date_modified`, and `subjects`
```
mfraezz: Preprints can no longer be filtered by their `provider`? Will this be a problem for `ember-preprints`? 

@chrisseto seems to think so

erinspace: Filtering by provider name was broken, I removed it instead of fixing it for speed because this was being handled by `/v2/preprint_providers/<provider_id>/preprints/` but happy to fix it instead

chrisseto: Logic train `->`:

The front page gets most recent preprints.

Right now they are not filtered by provider as we only have osf preprints and haven't thought about branding for other.

But it stands to reason that socarxiv would only want socarxiv preprints on their front page.

mfraezz: If the individual providers can access only their preprints at `/v2/preprint_providers/<provider_id>/preprints/`, this shouldn't be a blocker

erinspace: Yes, individual providers get their preprints through `/v2/preprint_providers/<provider_id>/preprints/` but I can put filtering back too if that's desired, no problem at all

mfraezz: No, this is fine. The only thing filter would add would be `/v2/preprints/?filter[provider]=rxiv` to get non-osf providers, but I'm pretty sure that's not a required feature right now.

PR #6169, opened 2016-08-18
Feature/mark users as spammy [OSF-6878]

brianjgeiger:
## Purpose

Allow admin users to remove Spam users and mark all of their projects as Private. This is a fairly safe a quick way to deal with individuals who create a lot of spam projects on a small number of accounts. If an overzealous admin accidentally marks a non-spam project as private, that is easily changed.

## Changes

1. Add a 'Remove Spam User' button to the admin user editor
2. When clicked and confirmed, mark all public non-registration nodes owned by that user as private, then call the existing Delete User function
3. Hidden logs on the nodes about them being marked as private, and admin logs about marking the nodes private on the admin app
4. Test

## Side effects

If a user is a contributor to a real project that is public, that project will be marked private. Any contributor to the project with sufficient permission will still be able to make the project public again. The spam user will not, of course.

## Ticket

https://openscience.atlassian.net/browse/OSF-6876


package.json
```diff
@@ -66,7 +66,7 @@
     "typeahead.js": "^0.10.5",
     "url-loader": "^0.5.5",
     "uuid": "^2.0.1",
-    "webpack": "^1.7.2",
+    "webpack": "1.13.0",
```
mfraezz: This seems... out of scope?

admin/users/views.py
```diff
@@ -76,6 +78,51 @@ def get_object(self, queryset=None):
...
+                        user=None,
+                        params={
+                            'node': node._id,
+                        },
+                        date=datetime.utcnow(),
+                        should_hide=True,
+                    )
+                    osf_log.save()
+
+        return super(SpamUserDeleteView, self).delete(request, *args, **kwargs)
```
mfraezz: By no means a blocker, but it would be nice if the admin log generated by this call could be configured with a reason for deletion.

admin/templates/users/remove_spam_user.html
```diff
@@ -0,0 +1,17 @@
+<form class="well" method="post" action="{% url 'users:spam_disable' guid=guid %}">
+    <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal">x</button>
+        <h3>Are you sure you want to delete this user and mark all their nodes as private? {{ guid }}</h3>
+    </div>
+    <div class="modal-body">
+        Parts of this action will be reversible after the fact.
```
mfraezz: This is... cryptic. Perhaps add information about which parts of this action will be irreversible (marking nodes private?). Also not a blocker.

admin/users/urls.py
```diff
@@ -12,6 +12,9 @@
         name='reset_password'),
     url(r'^(?P<guid>[a-z0-9]+)/disable/$', views.UserDeleteView.as_view(),
         name='disable'),
+    url(r'^(?P<guid>[a-z0-9]+)/disable_spam/$', views.SpamUserDeleteView.as_view(),
+        name='spam_disable'
+        ),
```
mfraezz: Not blocking, but prefer consistency over correctness with indentation.

mfraezz:
Pass complete, no blockers. I would like some feedback on why the `webpack` version was bumped up, but otherwise LGTM. No bugs found ⭐

package.json
brianjgeiger: The latest version is 1.13.1, which ^1.7.2 happily matches to. 1.13.1 will not build our assets because of some Dependency dependency error.

admin/users/views.py
brianjgeiger: We can add that as an improvement if product wants it. I don't disagree, but we don't really mark why users are deleted otherwise, and that might become a bigger thing.

brianjgeiger:
I fixed the indentation and the cryptic message while I was here.

PR #6175, opened 2016-08-19
Append VOL to Relationship Links [OSF-6733]

caseyrollins:
#### Purpose
- Currently when a route is viewed on the API with a view only token, the view only token is not passed to the generated links of each relationship field. This PR adds the view only token to each relationship link, making the API browsable with a view only token. 

#### Changes
- `view_only` query parameter is retrieved and added to the relationship urls returned from `get_url()` in the `RelationshipField` class.

#### Side effects
- None expected.

#### Ticket
- [OSF-6733](https://openscience.atlassian.net/browse/OSF-6733)

### DIFF
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -516,7 +516,11 @@ class RelationshipField(ser.HyperlinkedIdentityField):
                             url = '{}?filter{}'.format(url, formatted_filter)
                         else:
                             url = None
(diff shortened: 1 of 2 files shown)


brianjgeiger:
Well, that's delightfully simple. Approved if tests pass. 🎉 🐧 🎉

PR #6176, opened 2016-08-19
Fix Filter Operator Error [OSF-6483]

caseyrollins:
#### Purpose
- Using an invalid operator to filter a `RelationshipField` resulted in a 500 error because `get_valid_operators()` was returning `None`. 
- This PR changes `get_valid_operators()` to return `DEFAULT_OPERATORS` instead.

#### Ticket
- [OSF-6483](https://openscience.atlassian.net/browse/OSF-6483)


brianjgeiger:
Looks great. 🎉 🐧 🎉

PR #6177, opened 2016-08-20
[WIP] [Feature] ORCID Login OSF Part [##OSF-5162, 6881, 6885]

cslzchen:
## Purpose

TBD

## Changes

TBD

## Side effects

TBD

## Ticket

https://openscience.atlassian.net/browse/OSF-6881
https://openscience.atlassian.net/browse/OSF-5162

## Related CAS PR

https://github.com/CenterForOpenScience/cas-overlay/pull/22

### DIFF
--- a/framework/auth/cas.py
+++ b/framework/auth/cas.py
@@ -9,7 +9,8 @@ from lxml import etree
 import requests
 
 from framework.auth import User
-from framework.auth import authenticate
+from framework.auth import authenticate, oauth_first_time_authenticate
+from framework.auth.core import get_user
 from framework.flask import redirect
 from framework.exceptions import HTTPError
 from website import settings
@@ -99,7 +100,7 @@ class CasClient(object):
         url.args['ticket'] = ticket
         url.args['service'] = service_url
 
-        resp = requests.get(url.url)
+        resp = requests.get(url.url, verify=False)
         if resp.status_code == 200:
             return self._parse_service_validation(resp.content)
         else:
@@ -244,11 +245,58 @@ def make_response_from_ticket(ticket, service_url):
     client = get_client()
     cas_resp = client.service_validate(ticket, service_furl.url)
     if cas_resp.authenticated:
-        user = User.load(cas_resp.user)
-        # if we successfully authenticate and a verification key is present, invalidate it
-        if user.verification_key:
-            user.verification_key = None
-            user.save()
-        return authenticate(user, access_token=cas_resp.attributes['accessToken'], response=redirect(service_furl.url))
-    # Ticket could not be validated, unauthorized.
+        user, action = get_user_from_cas_resp(cas_resp)
+        if user and action == 'authenticate':
+            # if we successfully authenticate and a verification key is present, invalidate it
+            if user.verification_key:
+                user.verification_key = None
+                user.save()
+            return authenticate(
+                user,
+                cas_resp.attributes['accessToken'],
+                redirect(service_furl.url)
+            )
+        if not user and action == 'oauth_first_login':
+            from website.util import web_url_for
+            oauth_user = {
+                'fullname': cas_resp.attributes['name'],
+                'provider': cas_resp.attributes['oauthProvider'],
+                'id': cas_resp.attributes['oauthId'],
+                'access_token': cas_resp.attributes['accessToken']
+            }
+            return oauth_first_time_authenticate(
+                oauth_user,
+                redirect(web_url_for('oauth_user_email_get'))
+            )
+    # Ticket could not be validated, or user does not exisit: unauthorized.
     return redirect(service_furl.url)
+
+
+def get_user_from_cas_resp(cas_resp):
+    """
+    Given a CAS service validation response, attempt to retrieve user information and next action.
+
+    :param cas_resp: the cas service validation response
+    :return: the user and action
+    """
+
+    # cas returns the user id
+    if cas_resp.user:
+        user = User.load(cas_resp.user)
+        # TODO: justify the this redundant user existence check
+        # We have encountered an issue before that service validation passes while the user does not exist.
+        # This happened probably because CAS and OSF are listening to different OSF database when switching servers.
+        if user:
+            return user, 'authenticate'
+
+    # cas returns the oauth credential
+    if cas_resp.attributes['oauthProvider'] and cas_resp.attributes['oauthId']:
+        user = get_user(oauth_provider=cas_resp.attributes['oauthProvider'], oauth_id=cas_resp.attributes['oauthId'])
+        # user found
+        if user:
+            return user, 'authenticate'
+        # user first time login through oauth
+        else:
+            return None, 'oauth_first_login'
+
+    return None, None
--- a/framework/auth/cas.py
+++ b/framework/auth/cas.py
@@ -260,6 +260,7 @@ def make_response_from_ticket(ticket, service_url):
         # first time login from remote oauth provider, user does not exist or is not linked
         if not user and action == 'external_first_login':
             from website.util import web_url_for
+            # TODO: when cas part is ready, change the attribute names accordingly
             user = {
                 'fullname': cas_resp.attributes['name'],
                 'external_id_provider': cas_resp.attributes['oauthProvider'],
--- a/framework/auth/cas.py
+++ b/framework/auth/cas.py
@@ -9,7 +9,7 @@ from lxml import etree
 import requests
 
 from framework.auth import User
-from framework.auth import authenticate, oauth_first_time_authenticate
+from framework.auth import authenticate, external_first_login_authenticate
 from framework.auth.core import get_user
 from framework.flask import redirect
 from framework.exceptions import HTTPError
@@ -258,17 +258,17 @@ def make_response_from_ticket(ticket, service_url):
                 redirect(service_furl.url)
             )
         # first time login from remote oauth provider, user does not exist or is not linked
-        if not user and action == 'oauth_first_login':
+        if not user and action == 'external_first_login':
             from website.util import web_url_for
-            oauth_user = {
+            user = {
                 'fullname': cas_resp.attributes['name'],
-                'provider': cas_resp.attributes['oauthProvider'],
-                'id': cas_resp.attributes['oauthId'],
+                'external_id_provider': cas_resp.attributes['oauthProvider'],
+                'external_id': cas_resp.attributes['oauthId'],
                 'access_token': cas_resp.attributes['accessToken'],
             }
-            return oauth_first_time_authenticate(
-                oauth_user,
-                redirect(web_url_for('oauth_user_email_get'))
+            return external_first_login_authenticate(
+                user,
+                redirect(web_url_for('external_login_email_get'))
             )
     # Unauthorized: ticket could not be validated, or user does not exist.
     return redirect(service_furl.url)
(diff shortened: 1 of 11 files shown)


website/templates/get_oauth_email.mako
```diff
@@ -0,0 +1,24 @@
+<%inherit file="base.mako"/>
+<%def name="title()">OSF ORCID Login</%def>
```
mfraezz: All instances of `ORCID` in this file should probably be replaced with something like `${oauth_provider}`, to generalize this template

framework/auth/core.py
```diff
@@ -481,14 +502,19 @@ def create(cls, username, password, fullname):
...
+    def create_unconfirmed(cls, username, password, fullname, oauth=None, do_confirm=True,
                            campaign=None):
         """Create a new user who has begun registration but needs to verify
         their primary email address (username).
         """
         user = cls.create(username, password, fullname)
         user.add_unconfirmed_email(username)
         user.is_registered = False
+        if oauth:
+            if user.oauth:
```
mfraezz: You shouldn't need this conditional -- `User.oauth` is a `DictionaryField`. It will default to `{}`, so you can call `.update` on it before actually setting it.

framework/auth/cas.py
```diff
@@ -99,7 +100,7 @@ def service_validate(self, ticket, service_url):
         url.args['ticket'] = ticket
         url.args['service'] = service_url
 
-        resp = requests.get(url.url)
+        resp = requests.get(url.url, verify=False)
```
mfraezz: Is this necessary, or left over from mocking the CAS response?

cslzchen: this is a left over from shibboleth-cas institution login local development, not sure how this ends up here, probably from stash, should be removed

framework/auth/core.py
```diff
@@ -118,9 +118,16 @@ def get_user(email=None, password=None, verification_key=None):
                              'be provided.')
 
     query_list = []
+    if oauth_provider and oauth_id:
+        # TODO: this query is not working, need ot be fixed
```
mfraezz: As discussed, `query_list.append(Q('oauth.{}'.format(oauth_provider), 'eq', oauth_id))` should work, with a slight change to how the `oauth` dict is stored

cslzchen: Just tested, this works! Thanks.

framework/auth/views.py
```diff
@@ -637,3 +639,80 @@ def resend_confirmation_post(auth):
...
+        oauth = {
+            oauth_provider: {
+                'oauth_id': oauth_id,
+            },
+        }
+        if user:
+            # TODO: link user's OSF account with ORCID
+            # 1. notify the user that OSF account will be linked with ORCID profile
+            # 2. user can login through both ORCID and OSF
+            # 3. send confirmation email (no need to reply)
```
mfraezz: As discussed, we do need to verify here, to be sure that the person authorizing with ORCID actually has access to this email address

framework/auth/views.py
```diff
@@ -637,3 +639,80 @@ def resend_confirmation_post(auth):
...
+            # 4. redirect to home page
+            if user.oauth:
```
mfraezz: Same as https://github.com/CenterForOpenScience/osf.io/pull/6177#discussion_r75585033 , unnecessary conditional

framework/auth/core.py
cslzchen: current `User` model does not have `oauth`, we need a migration to remove this conditional
and I think this name is a little bit confusing, could be `external_idp` or `external_login`

mfraezz: A migration shouldn't be necessary -- if it's not present in the database document, the default value will be assumed. 
On `develop`:
```
In [1]: User.find()[0].oauth
<ipython-input-1-798447f85e8d> in <module>()
----> 1 User.find()[0].oauth

/Users/matt/.virtualenvs/osf/lib/python2.7/site-packages/modularodm/storedobject.pyc in wrapped(this, *args, **kwargs)
     93         if '_detached' in this.__dict__ and this._detached:
     94             warnings.warn('here')
---> 95         return func(this, *args, **kwargs)
     96     return wrapped
     97

/Users/matt/.virtualenvs/osf/lib/python2.7/site-packages/modularodm/storedobject.pyc in __getattr__(self, item)
    933             return ForeignList(ids, literal=True, base_class=base_class)
    934
--> 935         raise AttributeError(errmsg)
    936
    937     @warn_if_detached

AttributeError: User object has no attribute oauth
```
On this PR branch:
```
In [1]: User.find()[0].oauth
Out[1]: {}
```

mfraezz: Regarding a different name for the field: consider `oauth_logins` or `external_logins`, pluralized because it's a dictionary that can represent an id from multiple auth providers

framework/auth/core.py
```diff
@@ -350,6 +366,15 @@ class User(GuidStoredObject, AddonModelMixin):
     family_name = fields.StringField()
     suffix = fields.StringField()
 
+    # OAuth2 Authorization and Authentication
+    # TODO: what are proper parameters for `DictionaryField()`
+    # TODO: what else can we store for each oauth provider?
```
mfraezz: These `TODO`s can probably be removed. This may eventually be replaced with an `AbstractForeignField` pointing to some sort of OAuth provider model, but this should be able to be migrated if and when that happens

framework/auth/views.py
```diff
@@ -637,3 +639,79 @@ def resend_confirmation_post(auth):
...
+    oauth_provider = session.data['oauth_user_provider']
+    oauth_id = session.data['oauth_user_id']
+    oauth_fullname = session.data['oauth_user_fullname']
+
+    if form.validate():
+        clean_email = form.email.data
+        user = get_user(email=clean_email)
+        oauth = {oauth_provider: oauth_id}
+        if user:
+            # TODO: link user's OSF account with ORCID
```
mfraezz: The ordering for this `TODO` is off -- as discussed, this could lead to ORCID logins gaining access to any OSF account, so the confirmation email needs to be step 1, with following that link in that email triggering the rest.

framework/auth/views.py
```diff
@@ -637,3 +639,79 @@ def resend_confirmation_post(auth):
...
+                username=clean_email,
+                password=str(uuid.uuid4()),
+                fullname=oauth_fullname,
+                oauth=oauth,
+                campaign=None
+            )
+            # TODO: update social fields
+            user.save()
+            framework_auth.signals.user_registered.send(user)
+            send_confirm_email(user, email=user.username)
```
mfraezz: Consider modifying `send_confirm_email` to determine that these events are OAuth related, and to change the email template accordingly

website/templates/get_oauth_email.mako
```diff
@@ -0,0 +1,23 @@
+<%inherit file="base.mako"/>
+<%def name="title()">OSF ${oauth_provider} Login</%def>
+<%def name="content()">
+<h1 class="page-header text-center">OSF ${oauth_provider} Login | Register Email</h1>
+
+<div class="row">
+    ## Center the form
+    <div class="col-md-6 col-md-offset-3">
+        <p class='help'>
+            Please enter you email to finalize the login.
```
mfraezz: *your

website/templates/get_oauth_email.mako
```diff
@@ -0,0 +1,23 @@
...
+            If you already have an OSF account, this will link your ORCID profile with OSF.
```
mfraezz: "`link your ${oauth_provider} profile`"

website/templates/get_oauth_email.mako
```diff
@@ -0,0 +1,23 @@
...
+            If not, this will create an new account for you with your ORCID profile.
```
mfraezz: "`with your ${oauth_provider} profile`"

framework/auth/cas.py
```diff
@@ -244,11 +245,57 @@ def make_response_from_ticket(ticket, service_url):
...
+                redirect(service_furl.url)
+            )
+        # first time login from remote oauth provider, user does not exist or is not linked
+        if not user and action == 'oauth_first_login':
+            from website.util import web_url_for
+            oauth_user = {
+                'fullname': cas_resp.attributes['name'],
+                'provider': cas_resp.attributes['oauthProvider'],
+                'id': cas_resp.attributes['oauthId'],
+                'access_token': cas_resp.attributes['accessToken'],
```
mfraezz: Is this a token to talk to CAS, or an auth credential returned by ORCID? If the latter, I don't think we need to store it in the session, as it appears unused

cslzchen: as discussed, this access token is returned by CAS during service validate and it is never used.
we will keep it here for now, will remove in the future.

framework/auth/views.py
```diff
@@ -637,3 +644,116 @@ def resend_confirmation_post(auth):
...
+            # 0. check if this user is already linked with other profile
+            # 1. update user oauth
+            # 2. send confirmation email
+            # 3. notify user
+            # 4. remove session and osf cookie
+            external_status = ''
+            if user.external_identity:
+                if external_id_provider in user.external_identity:
+                    if user.external_identity[external_id_provider]:
+                        external_status = user.external_identity[external_id_provider]['status']
```
mfraezz: Is `status` guaranteed to be here? if not use `.get('status', <default>)`

framework/auth/views.py
```diff
@@ -637,3 +644,116 @@ def resend_confirmation_post(auth):
...
+            if external_status == 'VERIFIED':
+                message = language.EXTERNAL_LOGIN_EMAIL_LINK_FAIL.format(
+                    external_id_provider=external_id_provider,
+                    email=user.username
+                )
+                kind = 'warn'
+            else:
+                # TODO: link user's OSF account with ORCID
+                external_identity[external_id_provider]['status'] = 'LINK'
+                user.external_identity.update(external_identity)
```
mfraezz: As discussed, this shouldn't be done until they confirm it via email, because it would allow users to gain unauthorized access to other accounts

framework/auth/views.py
cslzchen: Yes, it is guaranteed. But `.get('status', <default>)` seems better, I will change this.

framework/auth/views.py
cslzchen: Only when `status=='VERIFIED'` is this external identity valid. The `status` is updated through the confirm process. `get_user()` checks `status`.

framework/auth/views.py
```diff
@@ -637,3 +644,116 @@ def resend_confirmation_post(auth):
...
+            # 2. send confirmation email
+            # 3. notify user
+            # 4. remove session and osf cookie
+            external_status = ''
+            if user.external_identity:
+                if external_id_provider in user.external_identity:
+                    if user.external_identity[external_id_provider]:
+                        external_status = user.external_identity[external_id_provider]['status']
+
+            # TODO: handle pending status: the current or another user also claimed this account but not confirmed
```
mfraezz: I think this would only be hit if the current user had already claimed, but if so shouldn't this flow be avoided (because `get_user_from_cas_resp` would return `user, 'authenticate'` instead of `None, 'oauth_first_login'`)?

mfraezz:
Pulled into `feature/orcid-login`, consider this merged :octocat: 

Any additional updates should be sent there.

PR #6180, opened 2016-08-21
[#OSF-6893]Show "in common" users on contributor search if query para…

chennan47:
## Purpose

Add to v2 api that how many projects users have in common.

## Changes
before change:
screen shot 2016-08-21 at 4 20 56 pm
after change:
screen shot 2016-08-21 at 4 21 25 pm

## Ticket

https://openscience.atlassian.net/browse/OSF-6893

### DIFF
--- a/api/users/serializers.py
+++ b/api/users/serializers.py
@@ -91,7 +91,7 @@ class UserSerializer(JSONAPISerializer):
     def get_projects_in_common(self, obj):
         user = get_user_auth(self.context['request']).user
         if obj == user:
-            return user.contributed
+            return len(user.contributed)
         return len(obj.get_projects_in_common(user, primary_keys=True))
 
     def absolute_url(self, obj):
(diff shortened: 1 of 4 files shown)


api/users/serializers.py
```diff
@@ -87,6 +88,12 @@ class UserSerializer(JSONAPISerializer):
     class Meta:
         type_ = 'users'
 
+    def get_projects_in_common(self, obj):
```
mfraezz: Consider adding a `TODO` for a follow-up that returns the users approximately sorted by `projects_in_common`, but that functionality isn't necessary now. See `website.search.elastic_search:search_contributor` for how this could be done, but that's not necessarily an efficient way of doing it.

mfraezz:
Pass complete, LGTM. Please make a follow-up ticket about modifying the ordering. 

Dropping this in RTM 🏭

PR #6189, opened 2016-08-22
[feature] Validate Relationships Resource Key [OSF-5268]

caseyrollins:
#### Purpose
- Currently, there is no validation for the relationships resource key that is required in the request body when creating a relationship. This PR adds validation by requiring that the resource key matches the target type, which is followed by validation that the target type is correct.

#### Changes
- Added validation in `JSONAPIParser.flatten_relationships()` 

#### Side effects
- None expected.

#### QA Note
To verify this, just add a contributor via the v2 API (we should have tests already for that) and try adding a contributor with a user type of something incorrect to see it fail.

#### Ticket
- [OSF-5268](https://openscience.atlassian.net/browse/OSF-5268)

### DIFF
--- a/api/base/parsers.py
+++ b/api/base/parsers.py
@@ -42,6 +42,10 @@ class JSONAPIParser(JSONParser):
         if not target_type:
             raise JSONAPIException(source={'pointer': 'data/relationships/{}/data/type'.format(related_resource)}, detail=NO_TYPE_ERROR)
 
+        if target_type != related_resource:
+            raise JSONAPIException(source={'pointer': 'data/relationships/{}'.format(related_resource)},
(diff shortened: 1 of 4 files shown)


brianjgeiger:
Is this the only problem? What about the case where you have two relationships with the same type? For example, you want to create a node with a particular parent and also with a particular node_link in the same request? Will it still work?

caseyrollins:
Based on @pattisdr's comments in the JIRA ticket, I don't believe the API currently supports multiple relationships in one request. So that would be a separate implementation issue where this validation would certainly need to be improved.

brianjgeiger:
Ah, okay. Take a look at https://openscience.atlassian.net/browse/OSF-6846 and see if the id issue there can be solved now as well; I don't think I moved that detail over to this ticket when I closed the other.

PR #6190, opened 2016-08-22
Add ability to filter to NodeList with `?filter[preprint]=true|false`

erinspace:
## Purpose
Since preprints are normal nodes, `v2/nodes` shows all preprints as well, without the ability to filter. This poses a problem when selecting a node you'd like to convert to a preprint, as trying to convert an already existing preprint.

Now you can filter on `preprint` with `?filter[preprint]=true` or `?filter[preprint]=false`

## Changes
- Add special case check in `filters.py` and convert to appropriate ODM query
- Preprint field on node serializer, added to filterable fields
- Node view documentation updates

## Side effects
None except the new ability to filter out preprints when needed from `v2/nodes/` with a filter. Any place we'd like to distinguish preprints from nodes is possible. 

## Ticket
None

### DIFF
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -219,7 +219,8 @@ class FilterMixin(object):
                             }
                         })
                     elif source_field_name == 'is_preprint':
-                        op = 'eq' if value == 'false' else 'ne'
+                        # TODO: Make this also include _is_preprint_orphan when value is false [#PREP-129]
+                        op = 'ne' if utils.is_truthy(value) else 'eq'
                         query.get(key).update({
                             field_name: {
                                 'op': op,
@@ -227,13 +228,14 @@ class FilterMixin(object):
                                 'source_field_name': 'preprint_file'
                             }
                         })
-                        query['_is_preprint_orphan'] = {
-                            field_name: {
-                                'op': 'ne',
-                                'value': True,
-                                'source_field_name': '_is_preprint_orphan'
+                        if utils.is_truthy(value):
+                            query['_is_preprint_orphan'] = {
+                                field_name: {
+                                    'op': 'ne',
+                                    'value': True,
+                                    'source_field_name': '_is_preprint_orphan'
+                                }
                             }
-                        }
                     else:
                         query.get(key).update({
                             field_name: {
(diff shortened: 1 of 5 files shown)


api/base/filters.py
```diff
@@ -218,7 +218,22 @@ def parse_query_params(self, query_params):
...
+                    elif source_field_name == 'is_preprint':
+                        op = 'eq' if value == 'false' else 'ne'
+                        query.get(key).update({
+                            field_name: {
+                                'op': op,
+                                'value': None,
+                                'source_field_name': 'preprint_file'
+                            }
+                        })
+                        query['_is_preprint_orphan'] = {
```
mfraezz: I'm not certain this part works -- `_is_preprint_orphan` is set when a node is loaded into memory and the `is_preprint` property is checked, but I think it's currently not persisting because it's not being saved.

mfraezz: A test would be a good way to verify that (e.g. `test_filter_preprint_true_excludes_orphans`, maybe `false_includes` as well if we want them to show up in that case).

erinspace: True! Manually setting a breakpoint and checking the ODM query at the end showed that it had indeed persisted. But a test would be the best way.

api/base/filters.py
```diff
@@ -218,7 +218,22 @@ def parse_query_params(self, query_params):
                                 'source_field_name': source_field_name
                             }
                         })
-
+                    elif source_field_name == 'is_preprint':
+                        op = 'eq' if value == 'false' else 'ne'
```
mfraezz: There's a util called `is_truthy` that would allow things like `?filter[preprint]=0` to work in this case

api/base/filters.py
```diff
@@ -218,7 +218,22 @@ def parse_query_params(self, query_params):
...
+                        query.get(key).update({
+                            field_name: {
+                                'op': op,
+                                'value': None,
+                                'source_field_name': 'preprint_file'
+                            }
+                        })
+                        query['_is_preprint_orphan'] = {
+                            field_name: {
+                                'op': 'ne',
```
mfraezz: This should perhaps be changed to `'op': op,` if there are any cases where we want to include orphans with this filter

mfraezz:
LGTM :fire:

PR #6194, opened 2016-08-23
[feature] remove throttle test that will randomly fail

caseyrollins:
#### Purpose
- Remove a test that will randomly fail because of a 0.1 second throttle limit. 

#### Note for QA
This does not require any testing.


brianjgeiger:
Perfect. 🎉 🐧 🎉 RTM once tests pass.

PR #6184, opened 2016-08-22
Add tests for private search [OSF-6898]

whit537:
&larr; #6197

## Purpose

This adds a suite of failing tests for the behavior we'd like to see from search over private data.

## Changes

tbd

## Side effects

tbd

## Ticket

https://openscience.atlassian.net/browse/OSF-6898

## Todo

- [x] fix basic registration test
- [x] test files appended to same
- [x] test registration states (approval, embargo)
- [x] ~~improve `mock_archive` comprehensively in a separate PR~~ https://github.com/CenterForOpenScience/osf.io/pull/6184#issuecomment-252656190
- [x] address https://github.com/CenterForOpenScience/osf.io/pull/6184#issuecomment-249879871
  - [x] wiki data
  - [x] used-to-be-embargoed
  - [x] component with parent in opposite publicity state

PR #6197, opened 2016-08-23
Refactor search tests

whit537:
#6184 &rarr;

## Purpose

Clean up our search test suite ahead of extending it with tests for private search (#6184).

## Changes

There's nothing for QA to test

## Side effects

n/a

## Ticket

https://openscience.atlassian.net/browse/OSF-6928


tests/test_search/test_elastic.py
```diff
@@ -78,152 +58,149 @@ def wrapped(*args, **kwargs):
...
         search.delete_index(elastic_search.INDEX)
         search.create_index(elastic_search.INDEX)
-        self.user = UserFactory(fullname='David Bowie')
+        self.user = factories.UserFactory(fullname='David Bowie')
 
     def test_new_user(self):
         # Verify that user has been added to Elastic Search
         docs = query_user(self.user.fullname)['results']
-        assert_equal(len(docs), 1)
+        assert len(docs) == 1
```
brianjgeiger: I'm afraid this change is not approved. Aside from the stylistic difference, the nose_asserts give more information than the bare asserts unless we are using pytest (which we do not). Please revert these back.

brianjgeiger:
Pass done. I'll do more once I have a significantly smaller diff to compare. 🐧 I didn't see anything else that went against our existing standards in a serious way, although we obviously won't want to have this PR hanging out too long or it will cause so many merge conflicts.

tests/test_search/test_elastic.py
whit537: Okay, rebased out in f0eec535715c757695fb4cd5a0bcf7cefcfeb156.

whit537:
> I'll do more once I have a significantly smaller diff to compare.

I made a PR for the first commit, since that was the bulk of the diff in terms of LOC though it's really a simple change. I've updated the description on this PR with a link to the diff between that and this. The individual commits are also suitable for viewing separately. ;-)

brianjgeiger:
I wasn't super concerned about moving the search views to a new file, it was mostly all the asserts that were interspersed between things. This looks like it should be fine once tests pass. It's not going to get merged for at least a day or two, though, even if all is well.

brianjgeiger:
Oh, and don't forget to use gitflow naming for branches: items for develop should be in a `feature/feature-name branch`, and items for master should be in a `hotfix/hotfix-name`. Don't change it for this, but in the future. Thanks!

brianjgeiger:
The `Changes` section of the PR is in good part so QA can have a focused idea of where to poke at things. Please go ahead and summarize the types of changes for a non-developer technical audience on future PRs rather than pointing to another location. No need for this one because there shouldn't be anything for QA to do related to this PR.

whit537:
> I wasn't super concerned about moving the search views to a new file, it was mostly all the asserts that were interspersed between things. This looks like it should be fine once tests pass.

Okay, I've closed #6199 and the associated Jira [ticket](https://openscience.atlassian.net/browse/OSF-6930).

> The `Changes` section of the PR is in good part so QA can have a focused idea of where to poke at things.

Ah, gotcha. So instead of a separate "QA notes" section ([cf.](https://openscience.atlassian.net/browse/OSF-6861?focusedCommentId=37632&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-37632)), I will use "Changes" for that ... and I take it from your edit of https://github.com/CenterForOpenScience/osf.io/pull/5977#issue-165784169 that "Side effects" is for the same audience, QA. 👍

whit537:
> Oh, and don't forget to use gitflow naming for branches: items for develop should be in a `feature/feature-name` branch, and items for master should be in a `hotfix/hotfix-name`.

I just slowed down and actually read all of "[A successful Git branching model](http://nvie.com/posts/a-successful-git-branching-model/)" for the first time. Sounds like the original model was looser on naming conventions ("anything except [...]"), with the [git-flow](https://github.com/nvie/gitflow) scripts enforcing a slightly modified convention, the one we seem to be using. I will endeavor to observe this convention with all diligence in the future!

brianjgeiger:
Cool, thanks. We're not 100% gitflow, and it's more important with hotfixes, but it's a nice double-check even for feature branches.

PR #6198, opened 2016-08-23
[Feature] [feature/orcid-login] ORCID Login Handle CAS Response [#OSF-5162, 6881, 6885]

cslzchen:
## Purpose

Update keys from CAS

## Changes

Update keys from CAS

## Side effects

No

## Ticket

https://openscience.atlassian.net/browse/OSF-6881
https://openscience.atlassian.net/browse/OSF-5162

### DIFF
--- a/framework/auth/cas.py
+++ b/framework/auth/cas.py
@@ -245,7 +245,7 @@ def make_response_from_ticket(ticket, service_url):
     client = get_client()
     cas_resp = client.service_validate(ticket, service_furl.url)
     if cas_resp.authenticated:
-        user, action = get_user_from_cas_resp(cas_resp)
+        user, external_credential, action = get_user_from_cas_resp(cas_resp)
         # user found and authenticated
         if user and action == 'authenticate':
             # if we successfully authenticate and a verification key is present, invalidate it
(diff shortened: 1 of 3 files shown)


framework/auth/cas.py
```diff
@@ -280,24 +279,57 @@ def get_user_from_cas_resp(cas_resp):
...
+
+    :param external_credential: the external credential string
+    :return: provider and id
+
+    """
+    # wrong format
+    if not external_credential or '#' not in external_credential:
+        return False
+
+    profile_name, technical_id = external_credential.rsplit('#', 1)
```
mfraezz: This should be using normal `split` -- it's trivially likely, but more probable that some provider would include an `#` in the `id` than in their service name:
```
In [8]: 'OCRID#1234-5678#9012-3456'.split('#', 1)
Out[8]: ['OCRID', '1234-5678#9012-3456']

In [9]: 'OCRID#1234-5678#9012-3456'.rsplit('#', 1)
Out[9]: ['OCRID#1234-5678', '9012-3456']
```

framework/auth/cas.py
```diff
@@ -257,14 +257,13 @@ def make_response_from_ticket(ticket, service_url):
...
+        if not user and external_credential and action == 'external_first_login':
             from website.util import web_url_for
-            # TODO: when cas part is ready, change the attribute names accordingly
             user = {
-                'fullname': cas_resp.attributes['fullname'],
-                'external_id_provider': cas_resp.attributes['oauthProvider'],
-                'external_id': cas_resp.attributes['oauthId'],
+                'external_id_provider': external_credential['provider'],
+                'external_id': external_credential['id'],
+                'fullname': cas_resp.attributes['given-names'] + ' ' + cas_resp.attributes['family-name'],
```
mfraezz: What happens if the user hasn't set both names on the service? Does CAS include an empty `<cas:given-names>` tag, or omit it? (Curiosity, not blocker)

cslzchen: Very good case. I am not sure. Need to take a look at OAuth specs. It is safe for ORCID for now, but we may want to handle this in CAS or OSF when we have other external IDPs. I will make a TODO here.

mfraezz:
Pulling this locally and updating tests / field structure, as discussed. Consider this merged ⏩

PR #6200, opened 2016-08-23
[Feature] [feature/orcid-login] Bug Fix [#OSF6885, 6881, 5162]

cslzchen:
## Purpose

Bug Fix for refactor `User.external_identity`.
Enable user to add more than one external identity.

## Changes

Same as above.

## Side effects

No

## Ticket

### DIFF
--- a/framework/auth/views.py
+++ b/framework/auth/views.py
@@ -542,7 +542,7 @@ def unconfirmed_email_add(auth=None):
     }, 200
 
 
-def send_confirm_email(user, email, external_id_provider=None):
+def send_confirm_email(user, email, external_id_provider=None, external_id=None):
     """
     Sends a confirmation email to `user` to a given email.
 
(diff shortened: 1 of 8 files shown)


framework/auth/views.py
```diff
@@ -757,47 +757,43 @@ def external_login_email_post():
...
-                send_confirm_email(user, user.username, external_id_provider=external_id_provider)
-                # 3. notify user
-                message = language.EXTERNAL_LOGIN_EMAIL_LINK_SUCCESS.format(
-                    external_id_provider=external_id_provider,
-                    email=user.username
-                )
-                kind = 'success'
-                # 4. remove session and osf cookie
-                remove_session(session)
+            user.save()
```
mfraezz: Duplicate `save`? See three lines below

cslzchen: Looks like it is.

PR #6205, opened 2016-08-24
Message when making Preprint Private

erinspace:
## Purpose

Since a preprint is just a project on the OSF, users can make them private and perhaps not realize the concequences.

This adds at least a warning message to give a bit of a heads up
screen shot 2016-08-24 at 1 41 15 pm

## Changes
- Change to the model to make sure `is_preprint` also looks at Privacy of a node (`is_public=False > is_preprint=False)
- Add check in `nodesPrivacy.js` for when making private to add special preprints message.
- Small wording change from "your preprint" to "this preprint"

## Side effects
- Anywhere that uses the node's `is_preprint` property  will now accurately exclude former preprints that are now private (to match what the API does...)

### DIFF
--- a/website/project/model.py
+++ b/website/project/model.py
@@ -1174,7 +1174,7 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin, Commentable):
 
     @property
     def is_preprint(self):
-        if not self.preprint_file:
+        if not self.preprint_file or not self.is_public:
             return False
         if self.preprint_file.node == self:
             return True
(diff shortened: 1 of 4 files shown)


mfraezz:
>Anywhere that uses the node's is_preprint property will now accurately exclude former preprints that are now private (to match what the API does...)

That seems like the right behavior anyway. When a private once-preprint node is made public again, does it go back to being a circulated preprint?

erinspace:
It does indeed! While we'll certainly need lots more OSF side additional messaging, I think adding too much at this very minute is getting a little out of scope for the moment (not to mention has no specs from product) - just wanted to cover this basic scenario after I thought of it in the meeting.

mfraezz:
> I think adding too much at this very minute is getting a little out of scope for the moment (not to mention has no specs from product)

Agreed. RTM when the build is happy. 🌟

PR #6206, opened 2016-08-24
Feature/projects in common

chennan47:
## Purpose
Exclude deleted projects from projects_in_common count.

## Ticket

https://openscience.atlassian.net/browse/OSF-6893


brianjgeiger:
Looks great. RTM when tests pass. 🎉 🐧 🎉

PR #6207, opened 2016-08-24
[hotfix] make github wb logger less brittle [OSF-6947]

felliott:
## Purpose

WB will soon start returning the applicable branch name in all Github action metadata.  The current `create_waterbutler_log` implementation for GH cannot handle this, as it expects a commit sha if any extra metadata is returned at all.

## Changes

Test directly for `extra.commit.sha` instead of just `extra`.

## Side effects

None expected.  This should have no visible effects.  It's purpose is to not break new functionality.

## Ticket

[OSF-6947]

https://openscience.atlassian.net/browse/OSF-6947


brianjgeiger:
This seems like it'll work well. 🎉 🐧 🎉

PR #6211, opened 2016-08-24
[feature fix] Set Bibliographic for Unregistered Contributors via API [OSF-6855]

caseyrollins:
#### Purpose
- When adding an unregistered contributor via the API, the visibility specified in the request was being ignored as it was not being passed to `add_unregistered_contributor`. It works now and tests have been updated.

#### Changes
- Add `visible` as a key word argument to `add_unregistered_contributor`. 
- Add and update tests.

#### Side effects
- None expected.

#### Ticket
- [OSF-6855](https://openscience.atlassian.net/browse/OSF-6855)


brianjgeiger:
Looks good when tests pass. 🎉 🐧 🎉

PR #6214, opened 2016-08-25
[rfr][preprints] Add custom signup/ campaign page for OSF Preprints service

abought:
## Purpose

Add a new campaign, so that users of the preprints service (through the OSF) will see a custom signup page and receive unique signup emails.

Sample URL: `http://staging.osf.io/login?campaign=osf-preprints`

If they click "use an existing account", then when they are done with the login process, they will be taken to `staging.osf.io/preprints` (the preprints landing page)

Upon confirming an account, they should be taken to the landing page for the campaign (eg `staging.osf.io/preprints`)

screen shot 2016-08-25 at 1 33 06 pm

## Changes

Add new `osf-preprints` campaign: custom signup page, and track users who came in this way via metrics.

## Side effects
None expected.

## Ticket
https://openscience.atlassian.net/browse/PREP-131

# TODO
- [x] Language approval for email and landing page

### DIFF
--- a/framework/auth/campaigns.py
+++ b/framework/auth/campaigns.py
@@ -21,7 +21,15 @@ CAMPAIGNS = ImmutableDict({
         'system_tag': 'erp_challenge_campaign',
         'redirect_url': lambda: furl.furl(DOMAIN).add(path='erpc/').url,
         'confirmation_email_template': mails.CONFIRM_EMAIL_ERPC,
-    }})
+    },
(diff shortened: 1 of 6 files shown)


brianjgeiger:
This seems like it'll be fine. If it works, it's good. 🎉 🐧 🎉

mfraezz:
LGTM. Tested and confirmed :shipit:

PR #6217, opened 2016-08-25
Alert that user cannot delete project on "Delete" click [OSF-6840]

Johnetordoff:
## Purpose

This stops the using from entering an the confirmation string before they get an error message telling them they can't delete a nested component.

## Changes

Changes the project base mako so contextvars include info on wheater a node has children, then uses that info to pre-validate on clicking the delete button.

## Side effects

None that I know of

## Ticket

https://openscience.atlassian.net/browse/OSF-6840


brianjgeiger:
Looks good. 🎉 🐧 🎉

PR #6218, opened 2016-08-25
Feature/preprint emails

erinspace:
**note** Relies on:
https://github.com/CenterForOpenScience/ember-osf/pull/125
and
https://github.com/CenterForOpenScience/ember-preprints/pull/88

## Purpose

Preprints needed the ability to send an email to preprint authors after the whole preprint process has been completed, which does not happen until after a normal project has been created and contributors added to it.

After changes proposed in https://github.com/CenterForOpenScience/ember-osf/pull/125 and https://github.com/CenterForOpenScience/ember-preprints/pull/88 go through, the add_contributor signal will only be sent after the POST request is made creating a new preprint, and the newly added authors get a notification to that effect.

## Changes
- Send an email to all other contributors on the project through the create method of `preprints/serializers.py`
- Test to make sure the `contributor_added` signal is sent on receiving a POST through the preprints endpoint
- Template wording as proposed by https://openscience.atlassian.net/browse/PREP-130

## Side effects

If the changes in the other two PRs are not made, the contributors will receive two emails - one through the normal contributor being added in the preprint process, and one a short time later 

## Ticket
https://openscience.atlassian.net/browse/PREP-130

### DIFF
--- a/api_tests/preprints/views/test_preprint_list.py
+++ b/api_tests/preprints/views/test_preprint_list.py
@@ -244,3 +244,10 @@ class TestPreprintCreate(ApiTestCase):
             assert_equal(res.status_code, 201)
             assert_equal(mock_signals.signals_sent(), set([project_signals.contributor_added]))
 
+    def test_preprint_contributor_signal_not_sent_one_contributor(self):
+        with capture_signals() as mock_signals:
+            private_project_payload = build_preprint_create_payload(self.private_project._id, self.subject._id,
+                                                                   self.file_one_private_project._id)
+            res = self.app.post_json_api(self.url, private_project_payload, auth=self.user.auth)
+            assert_equal(res.status_code, 201)
+            assert_not_equal(mock_signals.signals_sent(), set([project_signals.contributor_added]))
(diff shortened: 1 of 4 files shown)


api_tests/preprints/views/test_preprint_list.py
```diff
@@ -230,3 +235,12 @@ def test_file_not_osfstorage(self):
...
         assert_equal(res.json['errors'][0]['detail'], 'This file is not a valid primary file for this preprint.')
 
+    def test_preprint_contributor_signal_sent_on_creation(self):
+        with capture_signals() as mock_signals:
+            public_project_payload = build_preprint_create_payload(self.public_project._id, self.subject._id,
+                                                                   self.file_one_public_project._id)
+            res = self.app.post_json_api(self.url, public_project_payload, auth=self.user.auth)
+
+            assert_equal(res.status_code, 201)
+            assert_equal(mock_signals.signals_sent(), set([project_signals.contributor_added]))
```
mfraezz: Is it possible to check that this signal was sent for `self.other_user` and not sent for `self.user`?

mfraezz:
Minor (non-blocking) test improvement, otherwise LGTM 🌃 

Will merge when linked PRs go in.

api_tests/preprints/views/test_preprint_list.py
erinspace: I do not think so... I think just that the signal was sent at all. I can add another test that makes sure the signal is NOT sent when just the creator is on the project though?

chrisseto:
@mfraezz Other PRs merged. pls3merge5preprints.

PR #6219, opened 2016-08-26
[HOLD] add keenio read key to preprints serializer

felliott:
## Purpose / Changes

Make the Keen.io public analytics read token available through the nodes serializer, so that preprints can display download and view counts for the file.  Document that this is a private attribute and that we can change it without notice.
## Side effects

None expected.
## Ticket

No ticket.


icereval:
We'll need to discuss if this is the proper design going forward, attaching analytics keys to models seems impractical as these could end up all over the place...

Do we need a standard endpoint for these? I just don't see how the added attribute enriches the object.

icereval:
Can we change `keenio_read_key` to `analytics_read_key` and update the text to indicate it as an unsupported api attribute.

icereval:
@binoculars will be working on the `write` capabilities on the preprints side, some help will be needed to determine how to hand out the `public/private` scoped keys for each type of object on the API. <- @brianjgeiger

brianjgeiger:
I think the major options are "proxy" or "serialize". We can do some scope restriction or only allow it to happen if you're, for example, using a cookie. But yeah, we can go over options.

sloria:
Closing, as this is now stale.

PR #6221, opened 2016-08-26
[feature] API Rate Limiting [OSF-6799]

caseyrollins:
#### Purpose
- Adds rate limiting to the API.
- Authenticated user rate limit = 10,000/day
- Unauthenticated user rate limit = 100/hour
 - Note: unauthenticated requests are not limited if cookie auth is included in the request.
- Rate limit for sending add contributor emails = 10/second

#### Changes
- Adds custom `AddContributorThrottle` class for limiting add contributor requests that send emails. 
- Adds a private (hidden from swagger docs) endpoint (`/v2/test/throttle/`) with extremely low rate limits for basic testing purposes.

#### CR Notes
- http://www.django-rest-framework.org/api-guide/throttling/ and https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/throttling.py may be useful. 

#### Ticket
- [OSF-6799](https://openscience.atlassian.net/browse/OSF-6799)


website/project/model.py
```diff
@@ -3342,9 +3340,6 @@ def add_unregistered_contributor(self, fullname, email, auth, send_email='defaul
     def add_contributor_registered_or_not(self, auth, user_id=None, full_name=None, email=None, send_email='false',
                                           permissions=None, bibliographic=True, index=None, save=False):
 
-        if send_email != 'false' and not throttle_period_expired(auth.user.email_last_sent, settings.API_SEND_EMAIL_THROTTLE):
```
brianjgeiger: I think keep the API throttle in the API and don't use it in the model. You can modify this method to allow you to tell it that you're throttling outside, or that you just don't want throttling perhaps, so it doesn't try to double-throttle.

api/base/throttling.py
```diff
@@ -0,0 +1,69 @@
...
+        while self.history and self.history[-1] <= self.now - self.duration:
+            self.history.pop()
+        if len(self.history) >= self.num_requests:
+            return self.failure(request)
+        return self.success()
+
+
+class CookieAuthThrottle(AnonRateThrottle, BaseThrottle):
+
+    rate = '100/hour'
```
brianjgeiger: It would be better if all of these rates were configurable.

api/base/throttling.py
```diff
@@ -0,0 +1,69 @@
...
+
+    rate = '10/second'
+
+    def failure(self, request):
+        if request.method == 'POST' and request.query_params.get('send_email') != 'false':
+            return False
+        return True
+
+    def wait(self):
+        return 0.1
```
brianjgeiger: What's this do?

website/project/model.py
caseyrollins: @brianjgeiger, I'm removing the previously added throttling from the model here.

api/base/throttling.py
caseyrollins: Implementing the `wait()` method results in the `Retry-After` header being included (with the recommended number of seconds to wait) in the response of a throttled request. 

http://www.django-rest-framework.org/api-guide/throttling/#custom-throttles

website/project/model.py
brianjgeiger: Yes, the text is all red, isn't it? I expect I was the one slightly confused. Carry on.

api/base/settings/defaults.py
```diff
@@ -104,6 +104,16 @@
         'api.base.authentication.drf.OSFSessionAuthentication',
         'api.base.authentication.drf.OSFCASAuthentication'
     ),
+    'DEFAULT_THROTTLE_CLASSES': (
+        'rest_framework.throttling.UserRateThrottle',
+        'api.base.throttling.NonCookieAuthThrottle',
+        'api.base.throttling.AddContributorThrottle',
```
brianjgeiger: Should this be a default throttle class? I think this only gets applied to a small number of routes.

api/base/settings/defaults.py
```diff
@@ -104,6 +104,16 @@
...
+    ),
+    'DEFAULT_THROTTLE_RATES': {
+        'user': '10000/day',
+        'non-cookie-auth': '100/hour',
+        'add-contributor': '10/hour',
```
brianjgeiger: Based on what I've seen with this, I going to go on a limb and say I think you believe that the rates are tied to the default classes, and I don't *think* that's true. I think the default throttle rates are just a dictionary used by the throttle classes so that, if there's a scope, it'll see if there's a default rate in this dictionary that's keyed to that scope. I think you could have two default throttle classes and 200 default throttle rates and there proooobably wouldn't be a problem. 

Unless you've tried that and I'm completely wrong. But that was how I was imagining it.

caseyrollins: I believe you are correct. I'll test that to double-check and then make that change.

api/base/settings/defaults.py
caseyrollins: Ah, right again. Will remove 👍

api/base/views.py
```diff
@@ -1,36 +1,41 @@
...
-from api.base.utils import is_bulk_request, get_user_auth
 
+from api.users.serializers import UserSerializer
 
 CACHE = weakref.WeakKeyDictionary()
 
 
 class JSONAPIBaseView(generics.GenericAPIView):
 
+    throttle_classes = (UserRateThrottle, NonCookieAuthThrottle,)
```
brianjgeiger: Does this need to be here if it's in default throttle classes?

brianjgeiger:
Okay, you're going to hate this idea, but I think you should make some tests that mock the throttle classes and ensures that the right ones are called on the right endpoints. Pass done. 🐧

api/base/settings/local-travis.py
```diff
@@ -0,0 +1,46 @@
...
+    'DEFAULT_AUTHENTICATION_CLASSES': (
+        # Custom auth classes
+        'api.base.authentication.drf.OSFBasicAuthentication',
+        'api.base.authentication.drf.OSFSessionAuthentication',
+        'api.base.authentication.drf.OSFCASAuthentication'
+    ),
+    'DEFAULT_THROTTLE_CLASSES': (
+        'rest_framework.throttling.UserRateThrottle',
+        'api.base.throttling.NonCookieAuthThrottle',
+        'api.base.throttling.AddContributorThrottle',
```
brianjgeiger: Probably want to take this one out, too.

api/base/views.py
```diff
@@ -1,30 +1,31 @@
...
-from rest_framework.mixins import ListModelMixin
 from api.base import permissions as base_permissions
+from api.base import utils
 from api.base.exceptions import RelationshipPostMakesNoChanges
 from api.base.filters import ListFilterMixin
-
-from api.users.serializers import UserSerializer
-from api.base.parsers import JSONAPIRelationshipParser
-from api.base.parsers import JSONAPIRelationshipParserForRegularJSON
+from api.base.parsers import JSONAPIRelationshipParser, JSONAPIRelationshipParserForRegularJSON
```
brianjgeiger: So the advantage of having two lines for this import is that, if you add or delete a new item, there is only a single line change in the comparison, rather than an add and delete. Also, be careful about doing minor changes like this for stylistic reasons; this is the sort of thing that causes merge conflicts (if someone adds one parser and someone else removes another parser, that would cause trouble).

api_tests/base/test_throttling.py
```diff
@@ -0,0 +1,77 @@
...
+    def setUp(self):
+        super(TestUserRateThrottle, self).setUp()
+        self.user = AuthUserFactory()
+        self.url = '/{}'.format(API_BASE)
+
+    @mock.patch('rest_framework.throttling.UserRateThrottle.allow_request')
+    def test_user_rate_allow_request_called(self, mock_allow):
+        res = self.app.get(self.url, auth=self.user.auth)
+        assert_equal(res.status_code, 200)
+        assert_not_equal(len(mock_allow.mock_calls), 0)
```
brianjgeiger: Two things:

1. In general, it's better to be specific with the number of calls, because knowing that something was called too many times is at least as good as knowing it wasn't called at all. Having the mocks properly configured to be able to catch multiple calls to any throttling would have caught the problem with the "too many defaults", which leads me to…
2. If we can catch something that happens for all of the throttle calls (base_throttle.get_ident maybe?) as well, then you could be sure that you're only getting requests for two different throttles on default routes and maybe three on others.

website/project/model.py
```diff
@@ -42,12 +42,10 @@
 from website.util import api_url_for
 from website.util import api_v2_url
 from website.util import sanitize
-from website.util.time import throttle_period_expired
 from website.exceptions import (
     NodeStateError,
     InvalidTagError, TagNotFoundError,
-    UserNotAffiliatedError,
-    TooManyRequests
+    UserNotAffiliatedError
```
brianjgeiger: See? Look at all this diff. So messy.

brianjgeiger:
I'm being super-fussy about tests because, well, how would we know if they're broken otherwise. 

Don't change the import things, but try to avoid reformatting imports in the future (except on really small, going to be merged quickly types of tasks), and preferably only in the way that makes me happy. :) Pass done. 🐧

brianjgeiger:
Oh, and one of your tests is failing.

brianjgeiger:
I would restart the failed elastic test, but apparently there are merge conflicts, so I'll let you merge develop and then it should be ready to merge. 🐧

caseyrollins:
@brianjgeiger, up to date 🐱

brianjgeiger:
Looks great. One JS error that isn't really a thing, but trying to get it to pass anyways. Should be good to merge. 🎉 🐧 🎉

PR #6222, opened 2016-08-26
[WIP] Add preprint primary files to the intra-only move list

erinspace:
## Purpose
Preprints are normal nodes, so users can currently do anything to those nodes and files that one can do with normal nodes and files.

This is a small fangorn update that adds preprint primary files to the class of file that is "mustBeIntra," or files that can be moved around inside the current component, including inside files, but cannot be actually moved outside the current component or file storage system. If a user tries to move the file, a dialog opens that explains that it's a primary preprint file and so can't be moved, but offers to copy it to the new location instead.

move_file

Files that are "copied" to a location with a  file that already has the same name will show the same dialog as before, but now with an extra line about how you can't move a primary preprint file.

move_file_already_exists

### KNOWN BUG!
When multiselecting files, it still gives you the warning, but hitting the copy button does not actually copy the file. It moves it, even though it's very much not supposed to.  :(

## Changes

- update checkConflicts method in fangorn to combine message modals
- add preprint primary file to mustBeIntra list of things that shouldn't be allowed to be moved out of the current component

### DIFF
--- a/website/static/js/fangorn.js
+++ b/website/static/js/fangorn.js
@@ -427,19 +427,42 @@ function _fangornToggleCheck(item) {
 }
 
 function checkConflicts(tb, item, folder, cb) {
+    var isConflict = false;
+    var titleArray = [];
+    var buttonArray = [];
+    var message = m('');
+
+    // Check the status of named conflicts
     for(var i = 0; i < folder.children.length; i++) {
         var child = folder.children[i];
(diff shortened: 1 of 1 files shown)


erinspace:
Closing in favor of https://github.com/CenterForOpenScience/osf.io/pull/6225

PR #6225, opened 2016-08-27
Warn users when moving or replacing a preprint file

erinspace:
replaces https://github.com/CenterForOpenScience/osf.io/pull/6222

## Purpose

Preprints are normal nodes, so users can currently do anything to those nodes and files that one can do with normal nodes and files.

This is a fangorn update that warns a user when they try to move a file out of a preprint project, but then ultimately allows them to do it anyway if they really want to. If they move it, the preprint will become orphaned.

move_file_warning

### Things that should happen:
- Moving a preprint file out of a component should trigger a preprint warning
- Moving a preprint file into a folder should be OK
- Moving a preprint file into a component with a file that has the same name should trigger a move preprint warning modal, and then a same filename choice modal afterwards.
- Moving a file into the component should trigger a replace warning that mentions that if it was replaced, it will remove the preprint. 
- Trying to rename a file to the same name as the preprint file should trigger a preprint warning that it shouldn't be renamed

In all of these cases, the warning shows, but they can still complete the action.

## Changes

- Update `checkConflictsRename` and `checkConflicts` to check for preprint files and add an appropriate warning in
- On a normal move, check to see if the primary file is being moved out of the project, and 

## Side effects

Tested a bunch of things related to moving and renaming non preprint related files, but more tests really cannot hurt.

## Ticket
https://trello.com/c/pNvkyoEF/41-no-warning-when-moving-a-primary-file-to-another-component

### DIFF
--- a/website/static/js/fangorn.js
+++ b/website/static/js/fangorn.js
@@ -2255,6 +2255,20 @@ function _dropLogic(event, items, folder) {
     }
 
     $.each(items, function(index, item) {
+        if (window.contextVars.node.preprintFileId === item.data.path.replace('/', '') && item.data.nodeId === window.contextVars.node.id && folder.data.nodeId != window.contextVars.node.id) {
+            tb.modal.update(m('', [
+                m('p', 'The file "' + item.data.name + '" is the primary file for a preprint and so should not be moved.'),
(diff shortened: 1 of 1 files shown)


mfraezz:
LGTM 📂

PR #6229, opened 2016-08-27
[OSF-6951] Search Facet for Institutions not Displaying

Johnetordoff:
## Purpose

Migrating the search only includes non-institution nodes, and erases institution index making them disappear from ES and thus making the facet disappear.

## Changes

This re-populates institutions every time the search is migrated.

## Side effects

This is probably not a long-term solution to this problem, but will stop us from wiping institution's indexes inadvertently. Also worth noting populate_institution.py is the only way of adding institutions at the moment.
## Ticket

https://openscience.atlassian.net/browse/OSF-6951

### DIFF
--- a/website/search_migration/migrate.py
+++ b/website/search_migration/migrate.py
@@ -14,7 +14,7 @@ from framework.auth import User
 from website.models import Node
 from website.app import init_app
 import website.search.search as search
-from scripts import utils as script_utils
+from scripts import utils as script_utils, populate_institutions
 from website.search.elastic_search import es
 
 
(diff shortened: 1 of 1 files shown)


website/search_migration/migrate.py
```diff
@@ -71,6 +71,7 @@ def migrate(delete, index=None, app=None):
         delete_old(new_index)
 
     ctx.pop()
+    populate_institutions.main('prod')
```
brianjgeiger: This is going to break staging, staging2, test, etc. Make sure this populates the proper institution for the environment it's on.

brianjgeiger:
Also, this will need to be a hotfix, I think.

PR #6234, opened 2016-08-28
Feature/Reordering Contributors via APIv2 was not updating visible_contributor_ids

pattisdr:
## Purpose

Reordering contributors via the APIv2 was not updating the visible_contributor_ids which are used to populate the contributors on the project view page.  So reordering contributors on a preprint would have no effect on the order of contributors displayed on the project view page.

## Changes

Saves the order of visible contributor ids.

## Side effects

Just passed save=save into update_visible_ids.  Cannot think of any issues.

## Ticket
https://trello.com/c/Du2ygimQ/108-order-of-authors-on-view-page-not-consistent-with-what-was-saved-and-what-appears-on-project


pattisdr:
I don't think travis failures are mine?

mfraezz:
>I don't think travis failures are mine?

They weren't 🌻

PR #6235, opened 2016-08-28
Hotfix/send new index back with contributor update response

pattisdr:
## Purpose

If user updates index of contributor via API v2, the updated index is not returned in the response.  This was causing issues in ember-preprints where requests weren't being sent because the indices were incorrect ember-side.

## Changes
Returns updated index in PATCH contributor response.

## Ticket

https://trello.com/c/Du2ygimQ/108-order-of-authors-on-view-page-not-consistent-with-what-was-saved-and-what-appears-on-project


pattisdr:
oops wrong branch

PR #6237, opened 2016-08-29
Preprints - Save the children/documentation explainations

erinspace:
## Purpose
A complicated fix for a seemingly easy problem! ember-preprints subjects on the discover page were showing a caret even if there were no children there to load - the request for more children wasn't being made until the carat was clicked, so there was no way to know weather or not to load the carrot or not.

This change adds children as another foreign field, which is set in the `update_taxonomies` script, so that ember-preprints can know how many children there are before making the request, thus weather or not to show the carrot in advance

h/t to @sloria for much help

Also at request of @brianjgeiger add message to preprint related API routes that the API routes are in flux!

## Changes
- add child count to the taxonomies serializer  (added here in attributes instead of in extra for simplicity when reading from ember for speed sake)
- add child_count property to Subjects
- add children as foreign field to Subjects cause why not be proper
- use update_taxonomies script to also set the children
- update preprints realted views with caveat message

## Side effects
Must run `update_taxonomies` again to make sure all children are saved, orelse all children counts will be 0
An ember-osf PR will rely on this to get accurate carrts or not

## Ticket
https://trello.com/c/GETV8yQ1/111-preprints-discover-the-caret-does-no-action

### DIFF
--- a/website/project/taxonomies/__init__.py
+++ b/website/project/taxonomies/__init__.py
@@ -1,4 +1,4 @@
-from modularodm import fields
+from modularodm import fields, Q
 
 from framework.mongo import (
     ObjectId,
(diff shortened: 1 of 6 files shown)


website/project/taxonomies/__init__.py
```diff
@@ -13,10 +13,22 @@ class Subject(StoredObject):
...
 
     @property
     def absolute_api_v2_url(self):
         return api_v2_url('taxonomies/{}/'.format(self._id))
 
+    @property
+    def child_count(self):
+        return len(self.children)
+
+    def set_children(self):
```
mfraezz: This seems like something that should go into the update script, not on the model, given that it's a noop after being called once.

mfraezz:
Closing in favor of #6238 ♨️

PR #6238, opened 2016-08-29
[Hotfix] Save children

mfraezz:
## Purpose
Updates #6237 

## Changes
From parent PR:
>- add child count to the taxonomies serializer (added here in attributes instead of in extra for simplicity when reading from ember for speed sake)
- add child_count property to Subjects
- add children as foreign field to Subjects cause why not be proper
- use update_taxonomies script to also set the children
- update preprints realted views with caveat message

* minor update to model / script

An ember-osf PR will rely on this to get accurate carrts or not

### DIFF
--- a/scripts/update_taxonomies.py
+++ b/scripts/update_taxonomies.py
@@ -41,7 +41,7 @@ def update_taxonomies(filename):
 
             try:
                 subject = Subject.find_one(Q('text', 'eq', text))
-            except (NoResultsFound):    
+            except (NoResultsFound):
                 # If subject does not yet exist, create it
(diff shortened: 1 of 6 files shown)


scripts/update_taxonomies.py
```diff
@@ -75,7 +75,7 @@ def main():
         # the 'children' field for every subject
         logger.info('Setting "children" field for each Subject')
         for subject in Subject.find():
-            subject.set_children()
+            subject.children = Subject.find(Q('parents', 'eq', subject))
```
sloria: 👍

PR #6240, opened 2016-08-29
Non admins get 403 Forbidden instead of 409 Conflict when Wrongly POSTing to existing preprint

erinspace:
## Purpose
When a read-write user tries to POST to create a preprint that already exists, the API returned a 409 error before the 403 forbidden. While 409 is techncally correct, 403 is way more appropraite for the user because they'll never be able to make a node a preprint. 

## Changes

- check permissions of auth user before 409 conflict

## Side effects

no?

## Ticket
https://trello.com/c/vc41eU5E/124-api-read-write-and-read-only-users-get-a-409-error-instead-of-a-403


erinspace:
close for #6241

PR #6242, opened 2016-08-29
[OSF-6968] Removed spurious png file with backslash in name

acshi:
## Purpose
The \\.png file cannot be cloned by git on windows. Because the file is unused, it is safe to remove, enabling better platform compatibility.

## Changes
Removed the website/static/img/ \\.png file.

## Side effects
If the seemingly blank file is actually used somewhere, it would not be found. But could easily be recreated with a less offensive name,

## Ticket
https://openscience.atlassian.net/browse/OSF-6968


acshi:
Failed tests all seem to do with preprints. Have nothing to do with the extraneous file being removed.

brianjgeiger:
Suuuuper complicated change, but I think I follow the details. :) I've restarted the second test, since it really shouldn't be failing based on this change. Once it passes, this is ready to go. 🎉 🐧 🎉

PR #6249, opened 2016-08-30
Finish user serialization[#OSF-6917]

chennan47:
## Purpose

Fix user's social serialization by adding social infomation.

## QA notes:
1. make a user
2. update some social information of the user
3. view the user in API
Before change:
screen shot 2017-05-02 at 9 59 01 am

After change:
screen shot 2016-11-05 at 1 56 29 pm

## Ticket

https://openscience.atlassian.net/browse/OSF-6917


api_tests/users/views/test_user_detail.py
```diff
@@ -300,14 +300,16 @@ def setUp(self):
...
-                    'scholar': 'newScholar',
-                    'personal_website': 'http://www.newpersonalwebsite.com',
-                    'twitter': 'http://www.newpersonalwebsite.com',
-                    'linkedin': 'newLinkedIn',
-                    'impactstory': 'newImpactStory',
-                    'orcid': 'newOrcid',
-                    'researcherid': 'newResearcherId',
+                },
+                'social': {
+                    'github': ['http://github.com/even_newer_github/'],
```
brianjgeiger: This should be `'github': ['http://github.com/new_github/'],` so that it gets updated in the PATCH test below.

api/users/serializers.py
```diff
@@ -32,36 +61,27 @@ class UserSerializer(JSONAPISerializer):
...
-    baiduscholar = DevOnly(HideIfDisabled(AllowMissing(ser.CharField(required=False, source='social.baiduScholar',
-                                                           allow_blank=True, help_text='Baidu Scholar Account'), required=False, source='social.baiduScholar')))
-    ssrn = DevOnly(HideIfDisabled(AllowMissing(ser.CharField(required=False, source='social.ssrn',
-                                                           allow_blank=True, help_text='SSRN Account'), required=False, source='social.ssrn')))
     timezone = HideIfDisabled(ser.CharField(required=False, help_text="User's timezone, e.g. 'Etc/UTC"))
     locale = HideIfDisabled(ser.CharField(required=False, help_text="User's locale, e.g.  'en_US'"))
 
+    social = HideIfDisabled(ListLinksField(
+        {
+            'personal_website': 'personal_website_url',
```
brianjgeiger: This is a lot of duplication. If you're defining methods based on the field name, it would be easier to do like the SerializerMethodField and assume that you know the name of the method if you know the name of the field. That being said, this might not be the best way to go. 

If you were going to define something like this, it might be better to do field_name: format_string, so you'd have `'github': 'http://github.com/{}/'`.  There'll be more thoughts below.

api/users/serializers.py
```diff
@@ -1,18 +1,47 @@
...
+    return account_list
+
+
+def update_social_id(instance, val, key, social_strip_url):
+    if not isinstance(val, list):
+        social_url = ast.literal_eval(val)
+    else:
+        social_url = val
+    # TODO: when social data remodel to be a list change this update function
+    social_id = re.sub(social_strip_url, '', social_url[0][:-1] if social_url[0].endswith('/') else social_url[0])
```
brianjgeiger: This is a lot of going back and forth with urls and things just to support our weird model. I think we can simplify this.

api/users/serializers.py
brianjgeiger: Okay, what if we do this:

```json
{
    "data": {
            "attributes": {
                "social": {
                    "github": ["brianjgeiger","centerforopenscience"],
                    "twitter": ["thefoodgeek","awesomeography"],
                    "personal_website":["https://thefoodgeek.com/","https://awesomeography.com/"],
                    [...]
                }
            }
        }
}
```
This way we don't have to do a lot of validation on things. If we want, we could go crazy and give a usage string, or we could just do it in the docs. If we did it inline, it would be something like, `'github': {'items':['brianjgeiger','centerforopenscience'], 'usage': 'https://github.com/{}/},`. This way it can just get updated in one place if, say, linkedin changes their URLs for profiles or something.

brianjgeiger:
@chennan47 Could you update the screenshot (or replace it with just the text version of the json would be better) with the output of the serialization?

brianjgeiger:
Ping @chennan47: Please update the PR description's screenshot (or replace with json) with the updated response. Probably sample JSON for a PATCH or POST request would be good as well.

api/base/serializers.py
```diff
@@ -1039,7 +1061,9 @@ def to_representation(self, obj, envelope='data'):
         data = {
             'id': '',
             'type': type_,
-            'attributes': {},
+            'attributes': {
+                'social': {},
```
brianjgeiger: Is this necessary? I would have expected, if you don't explicitly mention the social object, it would naturally be put in attributes. Please check.

api/users/serializers.py
```diff
@@ -116,7 +90,10 @@ def update(self, instance, validated_data):
         for attr, value in validated_data.items():
             if 'social' == attr:
                 for key, val in value.items():
-                    instance.social[key] = val
+                    if key == 'profileWebsites':
+                        instance.social[key] = val
+                    else:
+                        instance.social[key] = val[0]
```
brianjgeiger: So in the case of profileWebsites, we'll replace the list with what's passed in here, but for everything else, we take the first item from the list and discard the rest silently? I think this needs to be documented in the view as to which items are currently accepting multiple items, and if there are more than one items passed in to a social field that is currently only accepting one item, that an error should be raised telling them that those fields are not currently accepting more than one item.

brianjgeiger: And please add a test for this situation.

brianjgeiger:
@chennan47 This will need some merge-conflict resolution. 

As an aside, I'm going to update the picture in the PR description, because you're showing the DEV_MODE version of the serialization, which would require versioning, rather than the production version, which doesn't have any social links at all.

PR #6250, opened 2016-08-30
[Hotfix] Search Facet for Institutions not Displaying [OSF-6951]

Johnetordoff:
## Purpose

Migrating the search only includes non-institution nodes, and erases institution index making them disappear from ES and thus making the facet disappear.
## Changes

This re-populates institutions every time the search is migrated.
## Side effects

This is probably not a long-term solution to this problem, but will stop us from wiping institution's indexes inadvertently. Also worth noting populate_institution.py is the only way of adding institutions at the moment.
## Ticket

https://openscience.atlassian.net/browse/OSF-6951


website/search_migration/migrate.py
```diff
@@ -49,7 +49,7 @@ def migrate_users(index):
     logger.info('Users iterated: {0}\nUsers migrated: {1}'.format(n_iter, n_migr))
 
 
-def migrate(delete, index=None, app=None):
+def migrate(delete, env=None, index=None, app=None, ):
```
brianjgeiger: We don't normally include a comma at the end of the list of arguments.

tests/test_search/test_elastic.py
```diff
@@ -817,32 +817,46 @@ def setUp(self):
...
+        count_query['aggregations'] = {
+            'counts': {
+                'terms': {
+                    'field': '_type',
+                }
+            }
+        }
+
+        res = self.es.search(index=settings.ELASTIC_INDEX, doc_type=None, search_type='count', body=count_query)
+        assert res['aggregations']['counts']['buckets'][0]['key'] == 'institution'
```
brianjgeiger: Is 'institution' guaranteed to be in position 0, or should this be a more generic check? Also, `assert_equal` would be better here in any case even if it does match.

website/search_migration/migrate.py
```diff
@@ -71,6 +71,7 @@ def migrate(delete, index=None, app=None):
         delete_old(new_index)
 
     ctx.pop()
+    populate_institutions.main(env)
```
brianjgeiger: Are there any problems running populate_institutions multiple times on the same environment? It looks like it should handle it, but at the same time it also looks like it's going to update every node tagged with an institution, as populate_institutions is for adding or changing institution data.

If the only problem is that the institutions keys are being dropped from elastic, maybe there's a simpler way of iterating over only the institutions that we have already configured and migrating them into search, so that populate_institutions can be a separate step for its specific purpose.

brianjgeiger:
Two things:

- [ ] Super minor remove a comma;
- [ ] Find a way to just update the search info without potentially updating all the nodes that populate_institutions might do.

Pass done. 🐧

tests/test_search/test_elastic.py
```diff
@@ -834,7 +837,7 @@ def test_first_migration_with_delete(self):
 
     def test_multiple_migrations_with_delete(self):
         for n in xrange(1, 21, 2):
-            migrate(delete=True, index=settings.ELASTIC_INDEX, app=self.app.app)
+            migrate(delete=True,  index=settings.ELASTIC_INDEX, app=self.app.app)
```
brianjgeiger: Too many spaces

tests/test_search/test_elastic.py
```diff
@@ -843,6 +846,24 @@ def test_multiple_migrations_with_delete(self):
...
+            'counts': {
+                'terms': {
+                    'field': '_type',
+                }
+            }
+        }
+        res = self.es.search(index=settings.ELASTIC_INDEX, doc_type=None, search_type='count', body=count_query)
+        for bucket in res['aggregations']['counts']['buckets']:
+            if bucket['key'] == u'institution':
+                assert True
```
brianjgeiger: Better, but we lean towards using nose asserts rather than bare asserts. If you set `institution_bucket_found = False` above this loop, you can `assert_true(institution_bucket_found)` after. This will give a better error if tests go wrong.

brianjgeiger:
A little more on the test, but everything else looks great. 🐧

brianjgeiger:
Looks good. 🎉 🐧 🎉

brianjgeiger:
@sloria This is not exactly a migration, but we'll need to run migrate_search when this is deployed.

Johnetordoff:
@sloria I'm sorry, but not sure I fully understand the problem, but I've tweaked this so it migrates institutions before other nodes and users, so all institutions will be available while the long migration takes place.

Is this a feasible solution?

sloria:
The migration will need to be completed before the frontend change is deployed. Otherwise, Knockout will throw an error when trying to access `logo_path_rounded_corners`.

The simple solution is to revert the change to the template. Then we'll merge and migrate search. Once that's done, we can put the template change in.

README.md
```diff
@@ -365,11 +365,12 @@ $ sudo rpm -ivh elasticsearch-1.2.1.noarch.rpm
...
 SEARCH_ENGINE = 'elastic'
 ```
-- Start the Elasticsearch server and migrate the models.
+- Start the Elasticsearch server and migrate the models. Migrating search must include the production enviorment, to run 
+- locally use 'test'.
 
 ```bash
 $ invoke elasticsearch
-$ invoke migrate_search
+$ invoke migrate_search 'test'
```
brianjgeiger: Hmm. So specifically this needs to have institutions set up, right? Or is there something else? I know we talked about this, but maybe we can catch the condition in migrate_search and handle it automatically. Alternately, perhaps a better flag, such as 'dev' or 'local'?

website/search/elastic_search.py
```diff
@@ -505,6 +505,7 @@ def update_institution(institution, index=None):
             'id': id_,
             'url': '/institutions/{}/'.format(institution._id),
             'logo_path': institution.logo_path,
+            'logo_path_rounded_corners': institution.logo_path_rounded_corners,
```
brianjgeiger: Are you still using this in search?

brianjgeiger:
Ignore CR, got confused with the two open PRs.

PR #6252, opened 2016-08-30
Hotfix/LoginNextUrl [OSF-6972]

cslzchen:
## Purpose

Please refer to relevant ticket for detailed information.

## Side effects

No

## Ticket

https://openscience.atlassian.net/browse/OSF-6972
https://openscience.atlassian.net/browse/SEC-50

### DIFF
--- a/framework/auth/views.py
+++ b/framework/auth/views.py
@@ -245,13 +245,8 @@ def auth_login(auth, **kwargs):
         next_url = request.args.get('redirect_url')
         must_login_warning = False
 
-    if next_url:
-        # Only allow redirects which are relative root or full domain, disallows external redirects.
-        if not (next_url[0] == '/'
-                or next_url.startswith(settings.DOMAIN)
-                or next_url.startswith(settings.CAS_SERVER_URL)
-                or next_url.startswith(settings.MFR_SERVER_URL)):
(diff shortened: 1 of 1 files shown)


framework/auth/views.py
```diff
@@ -245,13 +245,8 @@ def auth_login(auth, **kwargs):
         next_url = request.args.get('redirect_url')
         must_login_warning = False
 
-    if next_url:
```
brianjgeiger: Since you've removed this check, what are the possible values for next_url? If next_url is None, then your validate_next_url() function will raise an attribute error that you won't be checking for.

mfraezz: It won't raise an `AttributeError`; the first line in `validate_next_url` is this conditional

cslzchen: `validate_next_url(next_url)` will check `None` and return `True` if there is no `next_url`.
This is because in the other PR, `next_url` always have a default value if `None`

brianjgeiger: No, I'm wrong. The if is totally where it should be in that function. Ignore me.

brianjgeiger:
Looks good. This should be ready to go when tests pass. 🎉 🐧 🎉

PR #6253, opened 2016-08-30
[OSF-6831] Javascript URL Validation should support Unicode in host name

acshi:
## Purpose
Valid urls like http://davidhaberthür.ch/ where being causing unicode encoding errors because the host name contains unicode.

## Changes
The regex used for validation was updated to support unicode in the host name. Additional tests were written to confirm the correct behaviour of the validation.

## Side effects
An extra small bug was also found and fixed where a url like htps:////acshi@osf.io would pass validation because of too lenient rules governing usernames.

## Ticket
https://openscience.atlassian.net/browse/OSF-6831

## Depends on/See also
https://openscience.atlassian.net/browse/OSF-4842 which is very similar, but for back-end validation. Both of these tickets use the same test set. This PR will not pass tests until that PR (for the ModularODM) is merged first, because the tests on the OSF also implicitly test ModularODM's behavior.


brianjgeiger:
Please make sure tests are passing before moving to Code Review column. 🐧

acshi:
Well considering that it won't be able to pass tests until the modular odm
PR is merged, I thought it might be worth reviewing early anyway.

We could also mark it in some way as blocked by the other issue if we like.

PR #6256, opened 2016-08-31
[WIP] Akismet spam checking [OSF-6977]

samchrisinger:
## Purpose

@saradbowman is being attacked full force by spammers. She needs help.

## Changes
This hooks into the `website.search.elastic_search` code to spawn a celery task to check if the Node is spam or not. This is accomplished by making a POST to the Akismet API and checking the response. If the Node is identified as probable spam that node is flagged and it (and it's children) are made private. Attempts to make a node that's been flagged as spam public will fail.

Adds:
    - basic Akismet client
    - spam celery tasks for querying akismet API
    - SpamMixin to Node
    - disallow making spammy Nodes public
screen shot 2016-08-31 at 2 25 49 pm
   - Admin app interface for listing flagged spam and known spam nodes

## Side effects

When a Node gets flagged as spam it and its children are make private. If we later mark these Nodes as non-spam, we'll need to communicate to users that the will need to manually change the privacy back.

Also some changes made to the nodesPrivacy error callback. Before it was just using a hard-coded error message, now it's actually using the message returned by the APIv2.

## Ticket
https://openscience.atlassian.net/browse/OSF-6977


website/static/js/nodesPrivacy.js
```diff
@@ -242,17 +242,22 @@ NodesPrivacyViewModel.prototype.confirmChanges =  function() {
...
-            $osf.growl('Error', 'Unable to update project privacy');
+            var errorMessage = 'Unable to update project privacy';
+            if (xhr.responseJSON && xhr.responseJSON.errors) {
+                errorMessage = xhr.responseJSON.errors[0].detail;
+            }
+            $osf.growl('Problem changing privacy', errorMessage);
             Raven.captureMessage('Could not PATCH project settings.');
             self.clear();
-            window.location.reload();
+            $('#nodesPrivacy').modal('hide');
```
samchrisinger: Note to reviewer: I'm not 100% sure of the implication of this. I _believe_ the reload added [here](https://github.com/CenterForOpenScience/osf.io/commit/ff7bf4c884acb5d559604034f82d85d78215d2c4#diff-4f5e70a2bdc38e3df54149befd145224) might not be needed. I think it was in place for non-transactional bulk updates.

tests/test_models.py
```diff
@@ -3282,6 +3282,12 @@ def test_set_privacy_requests_embargo_termination_on_embargoed_registration(self
                     registration.set_privacy('public', auth=self.auth)
                     assert_equal(mock_request_embargo_termination.call_count, 1)
 
+    def test_set_privacy_on_spammy_node(self):
+        with mock.patch.object(settings, 'HIDE_SPAM_NODES', return_value=True):
+            with mock.patch.object(Node, 'is_spammy', mock.PropertyMock(return_value=True)):
```
sloria: Minor: Why not just set

```python
self.project.is_spam = True
```

?

PR #6258, opened 2016-08-31
[WIP] Let preprints fetch comments (via the nodes endpoint) [#PREP-135]

abought:
## Purpose

Support fetching comments for preprints.
## Changes
- Add a `comments` relationship field to preprint representation. It should point to `nodes/<nid>/comments` (with appropriate filter parameter).
## Side effects

None expected. However, since preprints are implemented as nodes with an extra flag, this re-uses the `/nodes/comments` endpoint. This design choice can and should be revisited if the underlying implementation changes.
## Ticket

https://openscience.atlassian.net/browse/PREP-135

### DIFF
--- a/api/preprints/serializers.py
+++ b/api/preprints/serializers.py
@@ -2,19 +2,18 @@ from modularodm.exceptions import ValidationValueError
 from rest_framework import exceptions
 from rest_framework import serializers as ser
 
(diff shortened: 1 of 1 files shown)


abought:
Not sure on current state of requirements; this Pr is very out of date and may be worth closing. Will let maintainer make final decision.

PR #6260, opened 2016-08-31
[Hotfix] Index ORCiD-registered unreg contribs [#OSF-6979]

mfraezz:
## Purpose
Unregistered contributors should be searchable after registering, regardless of whether they set a password or signed in with ORCiD for the first time.

## Changes
Ensure `user.is_active` when `user.register` calls `user.update_search`

## Side effects
None

## Ticket
[#OSF-6979](https://openscience.atlassian.net/browse/OSF-6979)

### Note
Verified with @icereval that only QA accounts were affected. They were manually migrated; no script is necessary.


icereval:
lgtm

PR #6259, opened 2016-08-31
[hotfix] Block Spam Email Domains [OSF-6976]

caseyrollins:
#### Purpose
- Prevents users from creating an account with a spam email domain. Additionally, talks/posters cannot be submitted to conferences/meetings from a blacklisted email address, and users cannot add blacklisted email addresses as alternate emails.

#### Changes
- Adds an _extensive_ list of blacklisted domains to `defaults.py`. 

#### Side effects
- None expected.

#### Ticket
- [OSF-6976](https://openscience.atlassian.net/browse/OSF-6976)

PR #6261, opened 2016-08-31
[Hotfix] Update email blacklist PR [OSF-6976]

mfraezz:
Parent PR: #6259

> #### Purpose
> - Prevents users from creating an account with a spam email domain. Additionally, talks/posters cannot be submitted to conferences/meetings from a blacklisted email address, and users cannot add blacklisted email addresses as alternate emails.
> 
> #### Changes
> - Adds an _extensive_ list of blacklisted domains to `defaults.py`. 
> 
> #### Side effects
> - None expected.
> 
> #### Ticket
> - [OSF-6976](https://openscience.atlassian.net/browse/OSF-6976)
>


framework/auth/utils.py
```diff
@@ -26,6 +28,9 @@ def validate_email(email):
     if not email or '@' not in email:
         raise ValidationError('Invalid Email')
 
+    if email.split('@')[1] in settings.BLACKLISTED_DOMAINS:
```
sloria: Might want to lowercase here to be safe.

PR #6268, opened 2016-09-01
[feature fix] Append VOL Key to API Links [OSF-6733]

caseyrollins:
#### Purpose
- Fixes incorrectly formatted comment relationship links by using `furl` to build relationship links with multiple query parameters (`view_only` and `filter`)
- Appends VOL key to generated `self` and `html` links.

#### Changes
- Adds generic `extend_querystring_if_key_exists` helper to `api.base.utils.py` to append the VOL key query parameter if present.

#### Ticket
- [OSF-6733](https://openscience.atlassian.net/browse/OSF-6733)

### DIFF
--- a/api_tests/nodes/views/test_node_detail.py
+++ b/api_tests/nodes/views/test_node_detail.py
@@ -161,19 +161,19 @@ class TestNodeDetail(ApiTestCase):
         assert_equal(res.status_code, 200)
         assert_equal(res.json['data'][0]['type'], 'comments')
 
-    def test_node_comments_link_query_params_formattedg(self):
+    def test_node_comments_link_query_params_formatted(self):
         CommentFactory(node=self.public_project, user=self.user)
         self.private_project_link = PrivateLinkFactory(anonymous=False)
         self.private_project_link.nodes.append(self.private_project)
         self.private_project_link.save()
 
-        res = self.app.get('{}?view_only={}'.format(self.private_url, self.private_project_link.key))
-        assert_equal(res.status_code, 200)
-        assert_in('comments', res.json['data']['relationships'].keys())
+        res = self.app.get(self.private_url, auth=self.user.auth)
         url = res.json['data']['relationships']['comments']['links']['related']['href']
-        res = self.app.get(url)
-        assert_equal(res.status_code, 200)
-        assert_equal(res.json['data'][0]['type'], 'comments')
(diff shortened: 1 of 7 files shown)


brianjgeiger:
For the tests that you've modified, could you verify that the requests you made definitely got the requested resource (check the type field or something else that would not be there if someone, say, replaced the url with `/v2/`, for example). I was going to let that go, but it would be good if you added a specific test for one or more of the cases that QA found problematic to ensure that the problems are fixed. 🐧

api_tests/nodes/views/test_node_detail.py
```diff
@@ -151,11 +152,28 @@ def test_node_has_files_link(self):
...
         res = self.app.get(self.public_url)
         assert_equal(res.status_code, 200)
         assert_in('comments', res.json['data']['relationships'].keys())
         url = res.json['data']['relationships']['comments']['links']['related']['href']
-        assert_equal(self.app.get(url).status_code, 200)
+        res = self.app.get(url)
+        assert_equal(res.status_code, 200)
+        assert_equal(res.json['data'][0]['type'], 'comments')
+
+    def test_node_comments_link_query_params_formattedg(self):
```
brianjgeiger: Typo (I mean, I presume `formattedg` isn't what you meant there)

api_tests/nodes/views/test_node_detail.py
```diff
@@ -151,11 +152,28 @@ def test_node_has_files_link(self):
...
+        self.private_project_link.nodes.append(self.private_project)
+        self.private_project_link.save()
+
+        res = self.app.get('{}?view_only={}'.format(self.private_url, self.private_project_link.key))
+        assert_equal(res.status_code, 200)
+        assert_in('comments', res.json['data']['relationships'].keys())
+        url = res.json['data']['relationships']['comments']['links']['related']['href']
+        res = self.app.get(url)
+        assert_equal(res.status_code, 200)
+        assert_equal(res.json['data'][0]['type'], 'comments')
```
brianjgeiger: Could you verify in this test that the private link is included everywhere that private links are expected but weren't the first time through?

brianjgeiger:
Super-close. Pass done. 🐧

brianjgeiger:
Looking good. 🎉 🐧 🎉

PR #6269, opened 2016-09-01
Feature/quick institutions sales stats [OSF-6952]

brianjgeiger:
## Purpose

Analytics gathering for getting user and node counts grouped by institutions.

## Changes

Add a script to the analytics module.

## Side effects

1. If the environment doesn't have a private keen project or private keen write key set, the script should output the dictionary rather than trying to send the data to keen
2. I'm not entirely sure how the analytics are run, so a celery beat task or cron job may need to be updated somewhere?

## Ticket

https://openscience.atlassian.net/browse/OSF-6952

### DIFF
--- a/scripts/analytics/institutions.py
+++ b/scripts/analytics/institutions.py
@@ -19,13 +19,48 @@ def get_count_by_institutions():
         user_query = Q('_affiliated_institutions', 'eq', institution.node)
         node_query = (
             Q('is_deleted', 'ne', True) &
-            Q('is_folder', 'ne', True) &
-            Q('parent_node', 'eq', None)
+            Q('is_folder', 'ne', True)
         )
+        registration_query = node_query & Q('is_registration', 'eq', True)
+        non_registration_query = node_query & Q('is_registration', 'eq', False)
+        project_query = non_registration_query & Q('parent_node', 'eq', None)
+        registered_project_query = non_registration_query & Q('parent_node', 'eq', None)
+        public_query = Q('is_public', 'eq', True)
+        private_query = Q('is_public', 'eq', False)
+        node_public_query = non_registration_query & public_query
+        node_private_query = non_registration_query & private_query
+        project_public_query = non_registration_query & public_query
+        project_private_query = non_registration_query & private_query
+        registered_node_public_query = registration_query & public_query
+        registered_node_private_query = registration_query & private_query
+        registered_project_public_query = registered_project_query & public_query
+        registered_project_private_query = registered_project_query & private_query
         count = {
+            'id': institution._id,
             'institution': institution.name,
-            'users': User.find(user_query).count(),
-            'nodes': Node.find_by_institutions(institution, node_query).count(),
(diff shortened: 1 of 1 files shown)


scripts/analytics/institutions.py
```diff
@@ -0,0 +1,53 @@
...
+
+    for institution in institutions:
+        user_query = Q('_affiliated_institutions', 'eq', institution.node)
+        node_query = (
+            Q('is_deleted', 'ne', True) &
+            Q('is_folder', 'ne', True) &
+            Q('parent_node', 'eq', None)
+        )
+        count = {
+            'institution': institution.name,
```
felliott: You probably want to save the id, in case the name gets updated.  Perhaps:
````
'institution': {
    'id': institution._id,
    'name': institution.name,
},
````

File: scripts/analytics/institutions.py [hunk 2]
```diff
@@ -0,0 +1,53 @@
...
+            'users': User.find(user_query).count(),
+            'nodes': Node.find_by_institutions(institution, node_query).count(),
```
felliott: For these two, we might want to add breakdowns later, so I would suggest an extra layer of nesting:
````

File: scripts/analytics/institutions.py [hunk 3]
```diff
@@ -0,0 +1,53 @@
...
+def main():
+    counts_by_institutions = get_count_by_institutions()
+    keen_project = keen_settings['private']['project_id']
+    write_key = keen_settings['private']['write_key']
+    read_key = keen_settings['private']['read_key']
+    if keen_project and write_key:
+        client = KeenClient(
+            project_id=keen_project,
+            write_key=write_key,
+            read_key=read_key
```
felliott: I think it's okay to omit the read key if you're not using it.

felliott:
Looks good so far!  I need to build some fake institutions, then I'll test it locally.

felliott:
:+1:  Make the structure changes, then this is ready to go!

felliott:
...err, the ticket asked for counts of public/private projects per institution.  Will that be covered by this PR?

scripts/analytics/institutions.py
```diff
@@ -0,0 +1,86 @@
...
+        node_private_query = non_registration_query & private_query
+        project_public_query = non_registration_query & public_query
+        project_private_query = non_registration_query & private_query
+        registered_node_public_query = registration_query & public_query
+        registered_node_private_query = registration_query & private_query
+        registered_project_public_query = registered_project_query & public_query
+        registered_project_private_query = registered_project_query & private_query
+        count = {
+            'id': institution._id,
+            'institution': institution.name,
```
felliott: Instead of `id` and `institution` as same level keys, group them together:
````
{
  'institution': {
    'id': institution._id,
    'name': institution.name,
  }
}
````
Later we might expand what we capture about the institutions (e.g. type: higher_ed|professional|nonprofit) and grouping properties makes reading queries more intuitive.

PR #6282, opened 2016-09-06
[WIP] Use the https URL scheme when resolving requirements, instead of the git scheme.

emetsger:
## Purpose

Update requirements URLs that use the `git` scheme to use the `https` scheme.

Some networks do not allow outbound tcp connections to port `9418` (the port for the `git` protocol).  When attempting to retrieve requirements that use the `git` URL scheme on such networks, dependency resolution (and the build) fails.  Updating requirements URLs to use `https` instead of `git` allow builds on such networks to succeed.
## Changes

Replace URLs using the `git` scheme with their `https` equivalent.
## Side effects

HTTPS transfer is probably slightly slower, but it probably isn't noticeable.


sloria:
Closing this, as it is now stale. Feel free to re-open this if you want to pick this back up!

PR #6284, opened 2016-09-06
Update links to cos.io

whit537:
## Purpose

A bad SSL certificate on https://centerforopenscience.org/ is breaking links from OSF to cos.io. This PR works around this problem by updating the links to point directly to https://cos.io/, which is the terminal redirect.

## Changes

The main link to check is the one in the footer.

## Side effects

n/a

## Ticket

https://openscience.atlassian.net/browse/OSF-7043


icereval:
I believe @mattclark has corrected the cert issue, but this is certainly a good step towards consistency, Thanks!

PR #6288, opened 2016-09-07
Replace GPLv3-only django-queryset-csv library[#OSF-6982]

chennan47:
## Purpose
Replace GPLv3-only django-queryset-csv library.

## Changes
Remove django-queryset-scv and put in a replacement.

## Ticket

https://openscience.atlassian.net/browse/OSF-6982


brianjgeiger:
Two things:

1. Are there any tests of the CSV export functionality right now? If not, we might need to figure something out with that;
2. Did you see the comment about adding .iterator() http://palewi.re/comments/cr/15/114/#c8109 - I don't know if it's necessary or wise, just worth considering.

Aside from that, looks good. 🐧

admin_tests/metrics/test_utils.py
```diff
@@ -181,3 +183,34 @@ def test_get_all_user_count(self):
...
+        last_time = time_now - timedelta(days=2)
+        NodeFactory(category='project', date_created=last_time)
+        NodeFactory(category='project', date_created=last_time)
+        get_days_statistics(last_time + timedelta(seconds=1))
+        self.time = time_now + timedelta(seconds=1)
+
+        self.FULL_PERSON_CSV_NO_VERBOSE = ['id,users,delta_users,unregistered_users,projects,delta_projects,'
+                                           'public_projects,delta_public_projects,registered_projects,'
+                                           'delta_registered_projects,date\r',
+                                           '1,0,0,0,0,0,0,0,0,0,2016-09-25 00:00:01.000001\r', '']
```
brianjgeiger: I'm not entirely sure what's going on in this test. I would expect that, since you're generating data with dynamic dates and such, that at least some of this data would need to be pulled from the models that you built in the factories.

Also, if you were to add creators to the nodes, would you get some different values for projects, public_projects, etc?

admin_tests/metrics/test_utils.py
```diff
@@ -181,3 +183,51 @@ def test_get_all_user_count(self):
...
+        NodeFactory(category='project', date_created=last_time)
+        initial_time = last_time + timedelta(seconds=1)
+        get_days_statistics(initial_time)
+        midtime = last_time + timedelta(days=1, seconds=1)
+        self.time = time_now + timedelta(seconds=1)
+
+        self.initial_static = [
+            'id,users,delta_users,unregistered_users,projects,delta_projects,public_projects,'
+            'delta_public_projects,registered_projects,delta_registered_projects,date\r',
+            '1,0,0,0,0,0,0,0,0,0,' + initial_time.strftime('%Y-%m-%d %H:%M:%S.%f') + '\r', '']
```
brianjgeiger: Prefer `'{}'.format` to string concatenation.

admin_tests/metrics/test_utils.py
```diff
@@ -181,3 +183,51 @@ def test_get_all_user_count(self):
...
+        self.latest_static = [
+            'id,users,delta_users,unregistered_users,projects,delta_projects,public_projects,'
+            'delta_public_projects,registered_projects,delta_registered_projects,date\r',
+            '3,0,0,0,0,0,0,0,0,0,' + self.time.strftime('%Y-%m-%d %H:%M:%S.%f') + '\r',
```
brianjgeiger: You also didn't respond to the part about having values for things like `public_projects` in https://github.com/CenterForOpenScience/osf.io/pull/6288#discussion_r80723828.

PR #6289, opened 2016-09-07
Hotfix/analytics running and no user create api

brianjgeiger:
## Purpose

Make the new analytics script run and try to prevent spam.

## Changes

1. Disable v2/users create
2. Add the institutions task to celery beat

## Side effects

Shouldn't be any


mfraezz:
LGTM 🌟

PR #6291, opened 2016-09-08
[hotfix] fix build: flake8 correction + fix invoke wheel to support constraints

felliott:
*This is a tag-team PR!  @mfraezz has fixed a flake8 issue in current master that was making travis cry.  @felliott's code fixes dependency installation when wheels are involved.  By our powers combined!*

## Purpose

@emetsger encountered a problem when trying to build the OSF with JHU's custom Dockerfile.  All of the requirements were built as wheels, then `invoke req -q` was told to use those wheels.  `oauthlib` failed to install because v1.1.2 was required by the constraints file, but only v0.7.2 (required by Mendely) and v2.* (the latest) were available.

## Changes

`pip wheel` accepts a constraints file just like `pip install` does.  This PR updates our `invoke wheel` task to pass the constraints to the wheelbuilder.  While investigating this, I also found that the OSF (without addons) is missing an explicit dependency on `oauthlib` even though we use it in several modules.  I've added that, as well.

## Side effects

None expected.

## Ticket

No ticket.

### DIFF
--- a/website/project/sanctions.py
+++ b/website/project/sanctions.py
@@ -22,7 +22,6 @@ from website import (
     settings,
     tokens,
 )
-from website.project.spam.model import SpamStatus
 from website.exceptions import (
     InvalidSanctionApprovalToken,
     InvalidSanctionRejectionToken,
--- a/requirements.txt
+++ b/requirements.txt
@@ -59,6 +59,7 @@ keyring==9.1
 
 requests==2.5.3
 urllib3==1.10.4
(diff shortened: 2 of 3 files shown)


website/project/sanctions.py
```diff
@@ -22,7 +22,6 @@
     settings,
     tokens,
 )
-from website.project.spam.model import SpamStatus
```
sloria: Sorry about that. Thanks for fixing.

PR #6294, opened 2016-09-08
[feature] API Search [OSF-4254]

caseyrollins:
#### Purpose
- Adds several search endpoints to API v2. Each endpoint can accept a search query via query parameters, e.g. `/v2/search/projects/?q=science` will search for projects with the word science in their title, description, tags, etc. 
 - `/v2/search/`
    - searches across files, projects, components, registrations, and users.
    - all of the items returned from this search should be serialized appropriately in a single response (users serialized as users, registrations serialized as registrations, etc.)
    - Additionally, this endpoint includes a `search_fields` dictionary in the response which contains the appropriate links and counts of results for each doc type. 
 - `/v2/search/files/`
   - searches across components -- query parameters will search within a file's name. 
 - `/v2/search/projects/`
   - searches across public projects -- query parameters will search within a project's title, description, tags, and contributor names. 
 - `/v2/search/components/`
   - searches across public components -- query parameters will search within a component's title, description, tags, and contributor names. 
 - `/v2/search/registrations/`
   - searches across public registrations -- query parameters will search within a registration's title, description, tags, and contributor names. 
 - `/v2/search/users/`
   - searches across public registrations -- query parameters will search within a user's given name, middle names, family name, jobs, and schools. 

#### Changes
- views for each search type.
- `SearchSerializer` to appropriately serialize the mixed models that are returned from the `/v2/search/` endpoint. 
- `SearchModelPaginator`, `SearchPaginator`, and `SearchPagination` classes to appropriately load the search results and paginate the endpoints since ES only returns 10 results at a time. 

#### Example JSON outputs
- [`/v2/search/`](https://gist.github.com/caseyrollins/50119512eed999c122a1a4141231ea5c)
- [`/v2/search/files/`](https://gist.github.com/caseyrollins/97c7ff90f3ce72c896239b1020d3aa12)
- [`/v2/search/users/`](https://gist.github.com/caseyrollins/c28c34728aa47026930ce0fe6c957809)
- [`/v2/search/projects/`](https://gist.github.com/caseyrollins/1b3849923fb5483945a9553a29505f99)
- [`/v2/search/components/`](https://gist.github.com/caseyrollins/1690273a61f968fd45a81ebc133044bc)
- [`/v2/search/registrations/`](https://gist.github.com/caseyrollins/2d1b1e8ee65354214cf41f0297a99c5c)

#### Side Effects
- None expected -- this PR contains additional endpoints for API v2 search, but does not alter existing ES functionality. 

#### Ticket
- [OSF-4254](https://openscience.atlassian.net/browse/OSF-4254)

### DIFF
--- a/api/base/pagination.py
+++ b/api/base/pagination.py
@@ -175,3 +175,56 @@ class NodeContributorPagination(JSONAPIPagination):
         total_bibliographic = len(node.visible_contributor_ids)
         response_dict['links']['meta']['total_bibliographic'] = total_bibliographic
         return Response(response_dict)
+
+
+class SearchPaginator(DjangoPaginator):
+
+    def __init__(self, object_list, per_page, model):
+        super(SearchPaginator, self).__init__(object_list, per_page)
+        self.model = model
+
+    def _get_count(self):
+        self._count = self.object_list['hits']['total']
+        return self._count
+    count = property(_get_count)
+
+    def page(self, number):
+        # what if results is None?
+        # what if self.model.load returns None?
+        number = self.validate_number(number)
+        results = self.object_list['hits']['hits']
+        items = [
+            self.model.load(result.get('_id'))
+            for result in results
+        ]
+        return self._get_page(items, number, self)
+
+
+class SearchPagination(JSONAPIPagination):
+
+    def paginate_queryset(self, queryset, request, view=None):
+        page_size = self.get_page_size(request)
+        if not page_size:
+            return None
+
+        model = request.parser_context['view'].model_class
+        paginator = SearchPaginator(queryset, page_size, model)
+
+        page_number = request.query_params.get(self.page_query_param, 1)
+        if page_number in self.last_page_strings:
+            page_number = paginator.num_pages
+
+        try:
+            self.page = paginator.page(page_number)
+        except InvalidPage as exc:
+            msg = self.invalid_page_message.format(
+                page_number=page_number, message=six.text_type(exc)
+            )
+            raise NotFound(msg)
+
+        if paginator.num_pages > 1 and self.template is not None:
+            # The browsable API should display pagination controls.
+            self.display_page_controls = True
+
+        self.request = request
+        return list(self.page)
--- a/api/base/pagination.py
+++ b/api/base/pagination.py
@@ -11,6 +11,7 @@ from rest_framework.utils.urls import (
 )
 from api.base.serializers import is_anonymized
 from api.base.settings import MAX_PAGE_SIZE
+from api.base.utils import absolute_reverse
 
 from framework.auth.core import User
 from framework.guid.model import Guid
@@ -193,15 +194,13 @@ class SearchPaginator(DjangoPaginator):
             return FileNode.load(obj_id)
 
     def _get_count(self):
-        self._count = self.object_list['hits']['total']
+        self._count = self.object_list['aggs']['total']
         return self._count
     count = property(_get_count)
 
     def page(self, number):
-        # what if results is None?
-        # what if self.load_obj returns None?
         number = self.validate_number(number)
-        results = self.object_list['hits']['hits']
+        results = self.object_list['results']
         items = [
             self.load_obj(result.get('_id'), result.get('_type'))
             for result in results
@@ -216,10 +215,8 @@ class SearchModelPaginator(SearchPaginator):
         self.model = model
 
     def page(self, number):
-        # what if results is None?
-        # what if self.model.load returns None?
         number = self.validate_number(number)
-        results = self.object_list['hits']['hits']
+        results = self.object_list['results']
         items = [
             self.model.load(result.get('_id'))
             for result in results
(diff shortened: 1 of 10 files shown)


api/base/pagination.py
```diff
@@ -175,3 +178,136 @@ def get_paginated_response(self, data):
...
+
+    def get_response_dict(self, data, url):
+        if isinstance(self.paginator, SearchModelPaginator):
+            return super(SearchPagination, self).get_response_dict(data, url)
+        else:
+            query = self.request.query_params.get('q', '*')
+            return OrderedDict([
+                ('data', data),
+                ('search_fields', OrderedDict([
+                    ('files', OrderedDict([
```
brianjgeiger: I forgot to warn you this morning, but the links here are going to need to be the same general structure as our relationship links. We're doing the pagination links wrong, but we can't change them until we have API Versioning in. But it'll probably look something like:

```
links:
    first
    last
    prev
    next
    meta:
        total
        per_page
    users:
        related:
            href
            meta:
                total
    projects:
        related:
            href
            meta:
                total
    components:
        related:
            href
            meta:
                total
```

api/search/views.py
```diff
@@ -0,0 +1,110 @@
...
+    def get_queryset(self):
+        query = self.request.query_params.get('q', '*')
+        page = int(self.request.query_params.get('page', '1'))
+        start = (page - 1) * 10
+        return search.search(build_query(query, start=start), doc_type=self.doc_type, raw=True)
+
+
+class Search(BaseSearchView):
+
+    required_read_scopes = [ComposedScopes.FULL_READ]
```
brianjgeiger: Two things: 1) scope definitions here have to be CoreScopes; 2) go ahead and make a CoreScope.SEARCH and use that for all of these.

api/base/pagination.py
```diff
@@ -175,3 +178,131 @@ def get_paginated_response(self, data):
...
         return Response(response_dict)
+
+
+class SearchPaginator(DjangoPaginator):
+
+    def __init__(self, object_list, per_page):
+        super(SearchPaginator, self).__init__(object_list, per_page)
+
+    def load_obj(self, obj_id, obj_type):
+        if obj_type in ['project', 'component', 'registration']:
```
brianjgeiger: I wonder if it's worth having a "search type to model" function that makes this generic. As we add more search types (preprints, meetings) it would be nice to have it centralized.

brianjgeiger:
Could you put some sample JSON into the PR description? Especially the heterogeneous combined view.

api/search/serializers.py
```diff
@@ -0,0 +1,41 @@
...
+            serializer = NodeSerializer(data, context=self.context)
+            return NodeSerializer.to_representation(serializer, data)
+
+        if isinstance(data, User):
+            serializer = UserSerializer(data, context=self.context)
+            return UserSerializer.to_representation(serializer, data)
+
+        if isinstance(data, FileNode):
+            serializer = FileSerializer(data, context=self.context)
+            return FileSerializer.to_representation(serializer, data)
```
brianjgeiger: 👍 So fancy!

api/search/views.py
```diff
@@ -0,0 +1,107 @@
...
+    pagination_class = SearchPagination
+
+    def __init__(self):
+        super(BaseSearchView, self).__init__()
+        self.doc_type = getattr(self, 'doc_type', None)
+
+    def get_queryset(self):
+        query = self.request.query_params.get('q', '*')
+        page = int(self.request.query_params.get('page', '1'))
+        start = (page - 1) * 10
```
brianjgeiger: Page size seems like something that could change. This probably either is configured somewhere or it should be configurable. #noMagicNumbers

api_tests/search/serializers/test_serializers.py
```diff
@@ -0,0 +1,64 @@
...
+from website.project.model import ensure_schemas
+from website.project.metadata.schemas import LATEST_SCHEMA_VERSION
+from website.search import search
+
+
+class TestSearchSerializer(DbTestCase):
+
+    def setUp(self):
+        super(TestSearchSerializer, self).setUp()
+        search.delete_all()
```
brianjgeiger: This seems like something that should happen in the tearDown(). Is it clearing out things from the super() call?

api_tests/search/views/test_views.py
```diff
@@ -0,0 +1,583 @@
...
+from website.search import search
+
+
+class ApiSearchTestCase(ApiTestCase):
+
+    def setUp(self):
+        super(ApiSearchTestCase, self).setUp()
+
+        # actually tear down the database and delete elastic indices for each test
+        teardown_database(database=database_proxy._get_current_object())
```
brianjgeiger: Much like the other case, is this a thing that happens because you can't trust, essentially, all of the other tests to have torn down properly?

caseyrollins: Actually, this is no longer necessary since I'm not creating users with specific usernames. And since I'm clearing out the ES index for each test, it doesn't matter that the database isn't being torn down.

api_tests/search/serializers/test_serializers.py
caseyrollins: Ah, yes it would be better there. Will move.

api/search/views.py
caseyrollins: Fixed in 046981e21de3632946f5f17c654969fc5626a4bc

caseyrollins:
@brianjgeiger, example outputs (in gist form since they're super long) added!

api/base/pagination.py
caseyrollins: Added: 307ea3e

api/search/views.py
```diff
@@ -0,0 +1,109 @@
...
+        start = (page - 1) * page_size
+        try:
+            results = search.search(build_query(query, start=start, size=page_size), doc_type=self.doc_type, raw=True)
+        except MalformedQueryError as e:
+            raise ValidationError(e.message)
+        return results
+
+
+class Search(BaseSearchView):
+
```
brianjgeiger: Docs please

caseyrollins: 860f729 💐

brianjgeiger:
🎉 🐧 🎉

PR #1710, opened 2015-02-03
[feature] Add rich citations widget

lyndsysimon:
This PR is the first in a series that will include richer citations for projects, as well as implement addons for Zotero and Mendeley.

This PR adds CSL-formatted citations. The user must enter the format used to display the citation each time for now, but this choice will be persistent in a forthcoming PR.

PR #6295, opened 2016-09-08
Citations added to apiV2 [OSF-7004]

hmoco:
## Purpose

Adds citations to the node endpoint. 

#### New endpoints:
 - ` /v2/nodes/<node_id>/citation ` : provides the nodes csl object
     - Response:
      ```
{

    "data": {
        "links": {
            "self": "localhost:5000/2dwgz"
        },
        "attributes": {
            "publisher": "Open Science Framework",
            "author": [
                {
                    "given": "Henrique",
                    "family": "Harman"
                }
            ],
            "type": "webpage",
            "title": "Preprint!"
        },
        "type": "node-citations",
        "id": "2dwgz"
    }
}
      ```
 - ` /v2/nodes/<node_id>/citaiton/<citation_format>`: provides the properly formatted citation in the format specified
     - Response:
      ```      

{
    "data": {
        "links": {},
        "attributes": {
            "citation": "Harman, H..  (2016, September 12). Preprint!. Retrieved  from localhost:5000/2dwgz"
        },
        "type": "styled-citations",
        "id": "apa"
    }
}
      ```
 - Same endpoints as above for registrations
 - ` /v2/citations/styles/ `: provides a list of citations possible
     - Response:
      ```
       "data": [

        {
            "links": {},
            "attributes": {
                "date_parsed": "2016-09-09T16:40:46.937000",
                "summary": null,
                "short_title": "AMR",
                "title": "Academy of Management Review"
            },
            "type": "citation-styles",
            "id": "academy-of-management-review"
        }, ...
      ```

## Ticket
https://openscience.atlassian.net/browse/OSF-7004


hmoco:
Side Note: I can respond to code review in this PR, however would like @erinspace's to take a look at the PR description to see if the routes I outlined match the desired behavior (not included in the ticket).

erinspace:
@hmoco yes, looks good!

icereval:
@brianjgeiger, please review

api/citations/utils.py
```diff
@@ -0,0 +1,29 @@
...
+    bibliography = CitationStylesBibliography(bib_style, bib_source, formatter.plain)
+
+    citation = Citation([CitationItem(node._id)])
+
+    bibliography.register(citation)
+
+    def warn(citation_item):
+        pass
+
+    bibliography.cite(citation, warn)
```
mfraezz: Trivial, but if you're passing a noop as a error-handling callback, `lambda x: None` should work just as well, and obviates the need to define one above.
```
In [213]: w = lambda x: None

In [214]: def p(x):
   ....:     pass
   ....:

In [215]: p('') == w('')
Out[215]: True
```

api/citations/utils.py
```diff
@@ -0,0 +1,29 @@
...
+    return unicode(bibliography.bibliography()[0])
```
mfraezz: I'm unfamiliar with `citeproc` internals; is this guaranteed to have a zero-th element even if the `warn` callback is triggered?

api/nodes/serializers.py
```diff
@@ -47,6 +47,30 @@ class NodeLicenseSerializer(RestrictedDictSerializer):
...
+    publisher = ser.CharField(allow_blank=True, read_only=True)
+    type = ser.CharField(allow_blank=True, read_only=True)
+    doi = ser.CharField(allow_blank=True, read_only=True)
+
+    links = LinksField({'self': 'get_absolute_url'})
+
+    def get_absolute_url(self, obj):
+        return obj['URL']
+#
+# class NodeCitationSerializer(ser.Serializer):
```
mfraezz: Please remove this if it's not going to be used

api/nodes/views.py
```diff
@@ -1228,6 +1232,78 @@ def perform_create(self, serializer):
...
+        type                     string               type of citation - web
+        doi                      string               doi of the resource
+
+    """
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        base_permissions.TokenHasScope,
+    )
+
+    required_read_scopes = [CoreScopes.NODE_BASE_READ]
```
mfraezz: I know this isn't a dangerous endpoint, but prefer scope granularity -- something like `NODE_CITATIONS_READ` would be good.

api/nodes/views.py
```diff
@@ -1228,6 +1232,78 @@ def perform_create(self, serializer):
...
+        =================================================================================
+        citation                string               complete citation for a node in the given style
+
+    """
+    permission_classes = (
+        drf_permissions.IsAuthenticatedOrReadOnly,
+        base_permissions.TokenHasScope,
+    )
+
+    required_read_scopes = [CoreScopes.NODE_BASE_READ]
```
mfraezz: Ditto [above comment](https://github.com/CenterForOpenScience/osf.io/pull/6295/files#r78129716)

api/nodes/serializers.py
```diff
@@ -47,6 +47,30 @@ class NodeLicenseSerializer(RestrictedDictSerializer):
...
+    title = ser.CharField(allow_blank=True, read_only=True)
+    author = ser.ListField(read_only=True)
+    publisher = ser.CharField(allow_blank=True, read_only=True)
+    type = ser.CharField(allow_blank=True, read_only=True)
+    doi = ser.CharField(allow_blank=True, read_only=True)
+
+    links = LinksField({'self': 'get_absolute_url'})
+
+    def get_absolute_url(self, obj):
+        return obj['URL']
```
mfraezz: I was going to ask if this works, because we usually use lower-case keys, but it does:
```
{'URL': 'localhost:5000/26ynf',
 'author': [{'family': u'Frazier', 'given': u'Matt'}],
 'id': u'26ynf',
 'issued': {'date-parts': [[2015, 11, 16]]},
 'publisher': 'Open Science Framework',
 'title': u'OSFO4',
 'type': 'webpage'}
```

mfraezz:
Drive-by pass complete, looking good 🌴 

Only real concern is that there might be a [`KeyError` due to `citeproc` oddities](https://github.com/CenterForOpenScience/osf.io/pull/6295#discussion_r78129284)

erinspace:
All citeproc code by Lyndsy btw! https://github.com/CenterForOpenScience/osf.io/pull/1710

api/citations/utils.py
hmoco: Not familiar with it as well, will prevent the possibility of an index error here though.

api/nodes/views.py
```diff
@@ -1228,6 +1232,78 @@ def perform_create(self, serializer):
...
+    required_write_scopes = [CoreScopes.NULL]
+
+    serializer_class = NodeCitationStyleSerializer
+    view_category = 'nodes'
+    view_name = 'node-citation'
+
+    def get_object(self):
```
brianjgeiger: The get_object() method always needs to check permissions.

api/citations/views.py
```diff
@@ -0,0 +1,78 @@
...
+        base_permissions.TokenHasScope
+    )
+
+    required_read_scopes = [CoreScopes.ALWAYS_PUBLIC]
+    required_write_scopes = [CoreScopes.NULL]
+    serializer_class = CitationSerializer
+    view_category = 'citations'
+    view_name = 'citation-detail'
+
+    def get_object(self):
```
brianjgeiger: This should run check_object_permissions()

api/nodes/views.py
hmoco: I was under the impression that `self.get_node()`, which is called in the next line, does permission checking:
```
 def get_node(self, check_object_permissions=True):
        node = get_object_or_error(
            Node,
            self.kwargs[self.node_lookup_url_kwarg],
            display_name='node'
        )
        # Nodes that are folders/collections are treated as a separate resource, so if the client
        # requests a collection through a node endpoint, we return a 404
        if node.is_collection or node.is_registration:
            raise NotFound
        # May raise a permission denied
        if check_object_permissions:
            self.check_object_permissions(self.request, node)
        return node
```

api/citations/views.py
hmoco: Does this apply to this endpoint? `check_object_permissions` seems to take in a node, while this endpoint does not talk to any nodes.

api/nodes/views.py
brianjgeiger: You're right; I thought I had checked for the NodeMixin, but apparently I only did that on the other instance.

api/citations/views.py
brianjgeiger: check_object_permissions() is what runs the permission classes defined above. Without that, we don't, for example, check to see if the token has the appropriate scope.

brianjgeiger:
And, of course, tests should pass.

hmoco:
@brianjgeiger , @mfraezz, have responded to comments!

api_tests/base/test_serializers.py
```diff
@@ -76,7 +76,7 @@ def test_registration_serializer(self):
         # fields that are visible for withdrawals
         visible_on_withdrawals = ['contributors', 'date_created', 'description', 'id', 'links', 'registration', 'title', 'type', 'current_user_can_comment', 'preprint']
         # fields that do not appear on registrations
-        non_registration_fields = ['registrations', 'draft_registrations']
+        non_registration_fields = ['registrations', 'draft_registrations', 'citation']
```
brianjgeiger: Registrations don't get citations? That seems improper.

brianjgeiger:
Pass finished. Why shouldn't registrations get citations? I would think people would want citations for registrations even more than for node. 🐧

brianjgeiger:
Work on this ticket links to OSF-7004, so be sure to move that ticket when it gets updated.

api_tests/base/test_serializers.py
hmoco: Registrations don't get a citations relationship endpoint. I can add that, but for the purposes of pushing this forward, could that go into a follow up PR (to be worked on today)? @brianjgeiger

hmoco: @brianjgeiger, nvm adding the registration endpoints to this PR

hmoco:
This is ready for another look!

brianjgeiger:
Once tests pass, please [change the status of 7004](https://github.com/CenterForOpenScience/osf.io/pull/6295#issuecomment-246363525).

api/registrations/views.py
```diff
@@ -540,6 +545,78 @@ def get_queryset(self):
...
+    )
+
+    required_read_scopes = [CoreScopes.NODE_REGISTRATIONS_READ]
+    required_write_scopes = [CoreScopes.NULL]
+
+    serializer_class = NodeCitationSerializer
+    view_category = 'registrations'
+    view_name = 'registration-citation'
+
+    def get_object(self):
```
brianjgeiger: Why didn't you just override the NodeCitationDetail and change the differences?

api/registrations/views.py
```diff
@@ -540,6 +545,78 @@ def get_queryset(self):
...
+    )
+
+    required_read_scopes = [CoreScopes.NODE_REGISTRATIONS_READ]
+    required_write_scopes = [CoreScopes.NULL]
+
+    serializer_class = NodeCitationStyleSerializer
+    view_category = 'registrations'
+    view_name = 'registration-style-citation'
+
+    def get_object(self):
```
brianjgeiger: Same here. You could probably save a bit of duplication but subclassing the node version.

brianjgeiger:
Nearly there. A little DRYing of the registration views. Also, could you put some sample JSON responses in the PR description so we can see what the output is when it's all working properly?

hmoco:
@brianjgeiger should be good for another look!

brianjgeiger:
Okay, seeing your output, it would be better if all the responses were JSON-API Compliant. So wrap everything in `data` and have ids, types, and attributes. The attributes can have a field called `citation` which is just the raw json of the citation.

brianjgeiger:
Pass done. 🐧

hmoco:
@brianjgeiger, back to you, now with updated responses in the description too!

brianjgeiger:
How about `/v2/nodes/<node_id>/citaiton/<citation_format>` returns `styled-citations` as a type and `/v2/citations/styles/` returns `citation-styles` as a type?

brianjgeiger:
Otherwise looks good. Pass Done. 🐧

hmoco:
@brianjgeiger one more time, back to yooou!!!! :fire: :fire: :fire:

brianjgeiger:
Looks good when tests pass. 🎉 🐧 🎉

hmoco:
@brianjgeiger, tests pass!

brianjgeiger:
Yay!

icereval:
FYI, @caneruguz we will need to hold the preprints release just a bit, as I'd like to not block @brianjgeiger's 0.93 release which goes out tomorrow. My current thought is we can release this as a hotfix (which its currently pinned against master) soonish after 0.93 is up and QA is satisfied with any post-release testing.

PR #6301, opened 2016-09-12
[feature fix] Adjust Base Route Throttle

caseyrollins:
#### Purpose
- Adds a custom throttle class to the root API view with a higher throttle rate for non-authenticated requests.


brianjgeiger:
Looks great once tests pass.

brianjgeiger:
🎉 🐧 🎉

PR #6302, opened 2016-09-12
Linkify project descriptions [OSF-6626] 

atelic:
## Purpose

Replaces #6030 
Allow uses to include links in their project descriptions.

## Changes

Uses `linkify.js` to allow html links in otherwise sanitized content. 

To avoid built-in sanitization this will only apply to descriptions when users can't edit them.

## Side effects

None apparent

## Ticket

https://openscience.atlassian.net/browse/OSF-6626

### DIFF
--- a/website/static/js/nodeControl.js
+++ b/website/static/js/nodeControl.js
@@ -125,6 +125,9 @@ var ProjectViewModel = function(data, options) {
                 newValue = response.newValue; // Update display to reflect changes, eg by sanitizer
                 self.description(newValue);
                 return {newValue: newValue};
+            },
+            display: function(value){
+                $('#nodeDescriptionEditable').html($osf.markdownAcceptsLinks(value));
             }
         }));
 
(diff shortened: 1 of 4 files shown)


website/static/js/nodeControl.js
```diff
@@ -144,8 +144,9 @@ var ProjectViewModel = function(data, options) {
                 return {newValue: newValue};
             }
         }));
+    } else {
+      $('#nodeDescriptionEditable').html($osf.markdownAcceptsLinks(self.description()));
```
brianjgeiger: Could you rename this function so that it's just "linkifyText" or similar, since it doesn't do anything with Markdown?

brianjgeiger:
One more change, then should be ready. When it's done and tests pass, just move the jira ticket to Code Review. Pass done. 🐧

brianjgeiger:
Looks great! 🎉 🐧 🎉

PR #6304, opened 2016-09-13
[HOLD] Preprint Provider migration and default

hmoco:


hmoco:
@chrisseto, I assume this is not going to be used, should I close it?

chrisseto:
@hmoco We should still probably run the migration on prod.
I think with dawn's changes to the submit page and changes to spam this may no longer be an issue.

@sloria @icereval would you prefer to one off this or have it as a script? I think there are only 2 of these cases in prod, @hmoco can confirm.

hmoco:
Last time I checked (last week) there were only 2 of those cases.

icereval:
@sloria can you update/close if necessary?

PR #6310, opened 2016-09-14
Make logos on project page not a circle [#OSF-7026]

erinspace:
## Purpose
Institution logos on the project page were being forced into  a circle, making the non circular logos look a little odd. Instead of using the circular versions of the logos, go ahead and let the logos not be circles if they aren't.

### before
screen shot 2016-09-14 at 2 58 52 pm

### after
screen shot 2016-09-14 at 2 45 20 pm

## Changes

- remove the img-circle css class from project page institution images

## Side effects
Logos look a little larger next to project titles

## Ticket
https://openscience.atlassian.net/browse/OSF-7026


brianjgeiger:
🎉 🐧 🎉

PR #6296, opened 2016-09-09
[Hotfix] Autoban spam users [#OSF-7020]

mfraezz:
## Purpose
Improve spam handling for users

## Changes
* Add Admin-module utilities for deleting spam (and restoring ham) users
* Minor updates to node spam-checking
* Autoban users detected as spam
* Migration to flag existing spam users as spam
* Tests

## Side effects
None expected. Possible angry users if Akismet has too many false-positives

## Migration
```
python -m scripts.flag_disabled_spammers --dry
```

## Ticket
[OSF-7020](https://openscience.atlassian.net/browse/OSF-7020)

PR #6312, opened 2016-09-15
[Hotfix] Autoban spam users [#OSF-7020]

icereval:
### Purpose

Improve spam handling for users

Closes #6296


website/util/akismet.py
```diff
@@ -66,11 +69,15 @@ def check_comment(self, user_ip, user_agent, **kwargs):
...
-        )
+        try:
+            res = requests.post(
+                '{}{}.{}/1.1/comment-check'.format(self.API_PROTOCOL, self.apikey, self.API_HOST),
+                data=data,
+                headers=self._default_headers,
+                timeout=5
+            )
+        except RequestException as e:
+            raise AkismetClientError(reason=e.message)
```
sloria: `message` doesn't appear to be an attribute of `RequestException` objects. source: https://github.com/kennethreitz/requests/blob/master/requests/exceptions.py#L12 . 

I think you will need something like:

```python
e.response.text
```

website/util/akismet.py
```diff
@@ -66,11 +69,15 @@ def check_comment(self, user_ip, user_agent, **kwargs):
         data['user_ip'] = user_ip
         data['user_agent'] = user_agent
 
-        res = requests.post(
-            '{}{}.{}/1.1/comment-check'.format(self.API_PROTOCOL, self.apikey, self.API_HOST),
-            data=data,
-            headers=self._default_headers
-        )
+        try:
+            res = requests.post(
```
sloria: `requests.post` will not raise an exception for error responses unless you call `res.raise_for_status()`.

sloria:
A `NodeStateError` will need to be handled here; https://github.com/CenterForOpenScience/osf.io/blob/e74f86ce8d071aa0c47cbcdec1a1b403b12cab7c/website/project/views/node.py#L75

PR #6314, opened 2016-09-15
Include index on node contributor detail [#OSF-7028]

erinspace:
## Purpose
The list view for BaseContributors grabs the index, but the detail view didn't return it.

## Changes
- assign the index in `get_object` method of `BaseContributorDetail`
- test to make sure index is there and it increments

## Side effects

- none anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-7028


brianjgeiger:
🎉 🐧 🎉

PR #6318, opened 2016-09-15
[feature] Add Prereg Banner to Dashboard [OSF-7009]

caseyrollins:
#### Purpose
- Adds a banner for the Preregistration Challenge to the OSF dashboard page.

#### Changes
- Adds a background to the Preprints banner because it didn't have one. 
- Makes all of the banner buttons have the same fixed width.

#### Side effects
- None.

#### Final Screenshot
- (with "Preregister" text on button as opposed to "Begin a Preregistration")
screen shot 2016-09-23 at 10 10 09 am

#### Product Notes
- @saradbowman, I chose the background for the Prereg banner just based on what I thought looked nice (repeating the three backgrounds that were already being used on the page), but I can change the background for this banner (and the others) to be whatever your ❤️  desires. Also, I can fit "Begin a Preregistration" on the button, but it does look a bit long compared to the other banners buttons (screenshots for both options shown below so you can which you prefer).

#### Ticket
- [OSF-7009](https://openscience.atlassian.net/browse/OSF-7009)

### DIFF
--- a/website/static/css/meetings-and-conferences.css
+++ b/website/static/css/meetings-and-conferences.css
@@ -3,3 +3,7 @@
         text-align: center;
     }
 }
+
+.btn {
+    width: 70%
+}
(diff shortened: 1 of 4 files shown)


caseyrollins:
@JeffSpies, nice catch. Fixing now.

PR #6322, opened 2016-09-16
[Feature] Workshop Admin user check up [OSF-6619]

Ghalko:
## Purpose
Update a workshop's csv with relevant information.

## Ticket
https://openscience.atlassian.net/browse/OSF-6619


admin/users/views.py
```diff
@@ -263,6 +265,66 @@ def get_object(self, queryset=None):
...
+            writer.writerow(row)
+        return response
+
+    @staticmethod
+    def parse(csv_file):
+        final = []
+        for i, temp_line in enumerate(csv_file):
+            try:
+                line = temp_line.strip().split(',')[:-1]
+            except UnicodeDecodeError as e:
```
brianjgeiger: So something like http://stackoverflow.com/a/17246997 doesn't work? I spent like 15 minutes a month or two ago trying to do CSV with unicode and ended up being distracted before I finished, but I remember coming across a couple of things.

admin/users/views.py
```diff
@@ -263,6 +265,66 @@ def get_object(self, queryset=None):
...
+        final = self.parse(csv_file)
+        response = HttpResponse(content_type='text/csv')
+        response['Content-Disposition'] = 'attachment; filename="workshop.csv"'
+        writer = csv.writer(response)
+        for row in final:
+            writer.writerow(row)
+        return response
+
+    @staticmethod
+    def parse(csv_file):
```
brianjgeiger: This method looks like it could use a little more testing. Especially in the case of Unicode, since you had to put so much work into it. If you can't come up with a good solution, at least a test that shows that unicode-filled lines get skipped properly would be fine for now.

admin/users/views.py
Ghalko: Even parsing the csv file directly doesn't work. I think I will stick to catching errors and testing.

brianjgeiger:
Closed in favor of #6400

PR #6325, opened 2016-09-19
Fix OSF4i Project Stats [OSF-7041]

brianjgeiger:
## Purpose

Make the project (i.e. top-level node) stats for Institutions actually correct. There should have been 4 incorrect columns (top-level projects that were: registered and not, public and private)

## Changes

1. Make the registered_project queries work off of registered nodes instead of non-registered nodes
2. Make the non-registered project queries work off of the project queries instead of the all-node queries

## Side effects

Shouldn't be. There's not really a lot that's reasonably QA-able here.

## Ticket

https://openscience.atlassian.net/browse/OSF-7041

### DIFF
--- a/scripts/analytics/institutions.py
+++ b/scripts/analytics/institutions.py
@@ -21,16 +21,17 @@ def get_count_by_institutions():
             Q('is_deleted', 'ne', True) &
             Q('is_folder', 'ne', True)
         )
+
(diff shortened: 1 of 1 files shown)


brianjgeiger:
Passed local travis https://travis-ci.org/brianjgeiger/osf.io/builds/161040894

PR #6328, opened 2016-09-19
Use JS rendering for all wiki markdown regardless of age [OSF-6412] [OSF-7038]

erinspace:
## Purpose

While wikis nowadays are rendered in Javascript, wikis created before Febuary 12, 2015 we
    - saved the wiki to the database as markdown
    - rendered the markdown in Python
    - saved the rendered markdown as html into the database for use later

If it was an old wiki, we were using that pre-rendered wiki html put that into the view pane and wiki widget instead of rendering the plain markdown with JS as we do with newer wikis.

We no longer want to trust html, and want to use JS to render all markdown from now on. **But** we want to keep the look of the old python markdown! Which has [some key differences from javascript markdown](http://pythonhosted.org/Markdown/#differences).  Of these differences, the most notable is list behavior.

For example! The following raw markdown:

    1. first
    2. second
    * third

... would render differently in javascript and python.

new, javascript:

1. first
2. second

* third

old, python:

1. first
2. second
3. third

## Changes
- Add new version of markdown-it list parsing that will happen before the "hr" rule - thus going before the "list" rule in markdown-it parsing. Replacing the list rule entirely results in unwanted behavior!
- Make a new mdOld instance of MarkdownIt for more customization if we want
- Change variables to `renderedBeforeUpdate` or `rendered_before_update` because they were more general because there's no more python renderer
- Update tests to look for `rendered_before_update` as `True` when it's an older wiki
- Add some "old" markdown rendering JS tests that test for mimicking how python was rendering content

## Side effects
All old wiki markdown will be rendered with javascript and MarkdownIt instead of py-markdown, and there will be a few differences:

1. Words emphasized with * hello * will not be italicized in javascript rendering, where as they were in python, only \*hello\* will be italicized
2. HTML will all be rendered in plain text, where as with python, link attributes (aka \<a href=http://osf.io>\link</\a>) would be rendered as an actual link. 
 
## Ticket
https://openscience.atlassian.net/browse/OSF-6412
https://openscience.atlassian.net/browse/OSF-7038

## Notes for QA --
This might be kind of tricky to test! My thoughts are it might involve a few steps:

1. **Before this gets merged to staging!** Make a wiki that has lots of markdown in it including lists. [Here's the wiki I've been working with](https://staging.osf.io/uwbsc/wiki/home/) - lemmie know who's working on it and I'll make you a contributor!
2. A dev will need to temporarially change the variable `WIKI_CHANGE_DATE` in [website/addons/wiki/settings/defaults.py](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/addons/wiki/settings/defaults.py#L13) on staging to a timestamp that's after the wiki was last saved. I found [this website](http://www.epochconverter.com/) incredibly helpful for getting the  Unix epoch time needed to be used for that variable. This will make the wiki render in old python!
3. Make note of how all of the lists are rendered and whatnot in python. It should look like this!
python_md_list_rendering
4. **After this gets merged to staging!** Go back to the same wiki, which will now be rendered using javascript. It should look pretty darn similar to the python one, with a few changes noted above in the "side effects" section

This is a lot of info! Please let me know if I can help clarify this at all!

### DIFF
--- a/website/addons/wiki/views.py
+++ b/website/addons/wiki/views.py
@@ -151,7 +151,7 @@ def wiki_widget(**kwargs):
             more = True
         else:
             wiki_html = BeautifulSoup(wiki_html)
-            rendered_before_update = wiki_page.rendered_before_update
+        rendered_before_update = wiki_page.rendered_before_update
     else:
         wiki_html = None
 
--- a/website/addons/wiki/views.py
+++ b/website/addons/wiki/views.py
@@ -143,7 +143,7 @@ def wiki_widget(**kwargs):
     # Show "Read more" link if there are multiple pages or has > 400 characters
     more = len(node.wiki_pages_current.keys()) >= 2
     MAX_DISPLAY_LENGTH = 400
-    use_python_render = False
+    rendered_before_update = False
     if wiki_page and wiki_page.html(node):
         wiki_html = wiki_page.html(node)
         if len(wiki_html) > MAX_DISPLAY_LENGTH:
@@ -151,7 +151,7 @@ def wiki_widget(**kwargs):
             more = True
         else:
             wiki_html = BeautifulSoup(wiki_html)
-        use_python_render = wiki_page.rendered_before_update
+            rendered_before_update = wiki_page.rendered_before_update
     else:
         wiki_html = None
 
@@ -159,7 +159,7 @@ def wiki_widget(**kwargs):
         'complete': True,
         'wiki_content': unicode(wiki_html) if wiki_html else None,
         'wiki_content_url': node.api_url_for('wiki_page_content', wname='home'),
-        'use_python_render': use_python_render,
+        'rendered_before_update': rendered_before_update,
         'more': more,
         'include': False,
     }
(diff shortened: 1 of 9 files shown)


website/addons/wiki/views.py
```diff
@@ -143,23 +143,23 @@ def wiki_widget(**kwargs):
...
+    rendered_before_update = False
     if wiki_page and wiki_page.html(node):
         wiki_html = wiki_page.html(node)
         if len(wiki_html) > MAX_DISPLAY_LENGTH:
             wiki_html = BeautifulSoup(wiki_html[:MAX_DISPLAY_LENGTH] + '...', 'html.parser')
             more = True
         else:
             wiki_html = BeautifulSoup(wiki_html)
-        use_python_render = wiki_page.rendered_before_update
+            rendered_before_update = wiki_page.rendered_before_update
```
mfraezz: Should this be in the `else` (or `len(wiki_html) <= MAX_DISPLAY_LENGTH`) block? Seems like it should be dedented to be always happen if there's wiki content, and perhaps put above this conditional for clarity.

erinspace: Excellent catch, thank you! Updated!

PR #6329, opened 2016-09-20
[feature] API Versioning [OSF-6473][OSF-6330]

caseyrollins:
#### Purpose
- Adds the ability to version the API. 
- Creates version `2.1`
 - Depreciates the `/node_links/` endpoints.
 - Changes pagination links to be JSON API compliant.
    - `2.0` pagination:
        
        ```
        links
            first
            last
            next
            prev
            meta
                total
                per_page
        ```
    
    - `2.1` pagination:

        ```
        meta
            total
            per_page
        links
            self
            first
            last
            next
            prev
        ```

- Versioning can be specified by the client in three ways:

1. URL Path Versioning, e.g. `api.osf.io/v2/` or `api.osf.io/v3/` (**REQUIRED**)
 - Only a major version can be specified via the URL path, i.e. `api.osf.io/v2.0.6/` is invalid, additionally, paths such as `api.osf.io/v2.0/` are invalid.
 - If the default version is within the major version specified in the URL path, the default version will be applied (i.e. if the default version is `2.3` and the URL path is `api.osf.io/v2/`, then `request.version` will be `2.3`).
 - If the default version is not within the major version specified in the URL path, the URL path version will be applied (i.e. if the default version is `3.0` and the URL path is `api.osf.io/v2/`, then `request.version` will be `2.0`)
2. Query Parameter Versioning, e.g. `api.osf.io/v2/nodes/?version=2.1.6` (**OPTIONAL**)
 - A specific version (major, minor, or patch) for a single request can be specified via the `version` query parameter, as long as it is an allowed version.
 - If the version specified in the query parameter does not fall within the same major version specified in the URL path, i.e `api.osf.io/v2/nodes/?version=3.1.4` a `409 Conflict` response will be returned. 
3.  Header Versioning, e.g. `Accept-Header=application/vnd.api+json;version=3.0.1` (**OPTIONAL**)
 - A specific version (major, minor, or patch) for a single request can be specified via the `Accept Header` of the request, as long as it is an allowed version.
 - If the version specified in the header does not fall within the same major version specified in the URL path a `409 Conflict` response will be returned. 
 - If both a header version and query parameter version are specified, the versions must match exactly or a `409 Conflict` response will be returned (i.e. one does not take precedence over the other).

#### Changes
- Adds a custom `BaseVersioning` class.
- Adds a `version` key word argument to the base API url, and as a result adds `version` to the `kwargs` passed to `absolute_reverse()` in the serializers.

#### Side effects
- None expected, although QA should be aware that the addition of the `version` key word argument to the base URL has the _ability_ to mess up every single endpoint, but with tests passing, this shouldn't be a problem.

#### Ticket
- [OSF-6473](https://openscience.atlassian.net/browse/OSF-6473) (general versioning)
- [OSF-6330](https://openscience.atlassian.net/browse/OSF-6330) (pagination fix)


api_tests/base/test_versioning.py
```diff
@@ -0,0 +1,133 @@
...
+            'Version {} specified in "Accept" header does not match version {} specified in query parameter'.format(
+                self.valid_header_version,
+                self.valid_query_parameter_version
+            )
+        )
+
+    def test_header_version_bad_format(self):
+        headers = {'accept': 'application/vnd.api+json;version=not_at_all_a_version'}
+        res = self.app.get(self.valid_url_path_version_url, headers=headers, expect_errors=True)
+        assert_equal(res.status_code, 406)
```
brianjgeiger: Look at you and your fancy status code.

api/base/serializers.py
```diff
@@ -58,6 +58,40 @@ def is_anonymized(request):
     return website_utils.check_private_key_for_anonymized_link(private_key)
 
 
+class ShowIfVersion(ser.Field):
+    """
+    Skips the field if the specified request version is not within the field's allowed versions.
```
caseyrollins: I can certainly make the comment here more explicit. And actually, I'm not sure why I made `min_version` inclusive and `max_version` exclusive... they don't need to be at all. If they are both exclusive, do you feel that `min_version` and `max_version` are still appropriate names? Or would parameters like `earliest_supported_version`/`last(latest?)_supported_version` be better?

api/base/pagination.py
```diff
@@ -41,6 +41,10 @@ def page_number_query(self, url, page_number):
 
         return paginated_url
 
+    def get_self_real_link(self, url):
+        page_number = self.page.number
```
caseyrollins: `self` just links to the current "page"

screen shot 2016-09-27 at 1 43 48 pm

caseyrollins:
@brianjgeiger, I believe I've addressed all of the CR comments, with the exception of the [undecided](https://github.com/CenterForOpenScience/osf.io/pull/6329#discussion_r80749808) `min_version`/`max_version` parameter naming change. 🔙  to you.

api/base/serializers.py
brianjgeiger: If they're symmetrical, min and max are fine.

caseyrollins:
🐧 🎉 🐧

PR #6339, opened 2016-09-23
Feature/Make current_version a file serializer field [OSF-7067] [PREP-144]

pattisdr:
## Purpose

The version of the primary file for the preprint needs to be exposed in the view preprint page.

## Changes

Exposes current_version as a serializer field on file. Determined by the length of the history.

## Ticket
https://openscience.atlassian.net/browse/OSF-7067
Needed for https://openscience.atlassian.net/browse/PREP-144

### DIFF
--- a/api_tests/files/views/test_file_detail.py
+++ b/api_tests/files/views/test_file_detail.py
@@ -398,6 +398,20 @@ class TestFileView(ApiTestCase):
         res = self.app.get(url, auth=self.user.auth, expect_errors=True)
         assert_equal(res.status_code, 404)
 
+    def test_current_version_is_equal_to_length_of_history(self):
+        res = self.app.get(self.file_url, auth=self.user.auth)
+        assert_equal(res.json['data']['attributes']['current_version'], 1)
+        for version in range(2, 4):
+            self.file.create_version(self.user, {
+                'object': '06d80e' + str(version),
+                'service': 'cloud',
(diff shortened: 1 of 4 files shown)


api_tests/files/views/test_file_detail.py
```diff
@@ -71,6 +71,7 @@ def test_get_file(self):
         assert_equal(attributes['last_touched'], None)
         assert_equal(attributes['provider'], self.file.provider)
         assert_equal(attributes['size'], self.file.versions[-1].size)
+        assert_equal(attributes['current_version'], len(self.file.history))
```
caneruguz: @pattisdr is this foolproof? Is there any way the file history count is not the latest version?

caneruguz:
@pattisdr can you get this PR through by checking in with @brianjgeiger about what it needs?

api_tests/files/views/test_file_detail.py
brianjgeiger: This test is a little self-referencing, which is fine to prevent regressions but not the best for verifying the behavior. Could you make sure that in the setup that the file has a non-trivial number of versions, and that whatever you do to make those versions (`for` loop, for example) is the thing that gets compared?

pattisdr: I left this base test as is.

Added another test that went through a for loop creating versions of files and then requesting that file, checking the version was the latest version.

PR #6342, opened 2016-09-25
[feature fix] Fixed Width Banner Buttons on Dashboard [OSF-7009]

caseyrollins:
#### Purpose
- Adds `.banner` CSS class for fixed width, which is applied only to the banner buttons on the dashboard.
- Removes `.btn` class from `meetings-and-conferences.css` which caused the fixed width to be incorrectly applied to the "Create new project" button.

#### Ticket
- [OSF-7009](https://openscience.atlassian.net/browse/OSF-7009)

### DIFF
--- a/website/static/css/pages/home-page.css
+++ b/website/static/css/pages/home-page.css
@@ -25,9 +25,6 @@ h3, h4 {
     background-size: cover;
     color: #fff;
 }
-.meetings .btn {
-    box-shadow: 0 0 9px -4px #000;
-}
 
 .prereg {
     background: #C5D6E6 url("/static/img/bg6.jpg") no-repeat center top;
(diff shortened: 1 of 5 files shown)


website/static/css/pages/home-page.css
```diff
@@ -39,6 +36,11 @@ h3, h4 {
     color: #FFF;
 }
 
+.banner {
```
brianjgeiger: Since this is for a button, perhaps we could call it `btn-banner` to match with the naming scheme.

website/static/css/pages/home-page.css
```diff
@@ -25,9 +25,6 @@ h3, h4 {
     background-size: cover;
     color: #fff;
 }
-.meetings .btn {
```
brianjgeiger: And this isn't being used anywhere else? (Just double-checking).

caseyrollins: Nope, pretty sure it was just a duplicate of [this styling](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/home-page/meetingsAndConferencesPlugin.js#L25).

brianjgeiger:
🎉 🐧 🎉 (When tests pass)

PR #6346, opened 2016-09-27
[Feature] Preprint guids [OSF-7049]

mfraezz:
## Purpose

Make preprints addressable with guids.
## Changes
- Models:
  - `Node` loses most `preprint`-type fields. `file` and `doi` remain, but `doi` is renamed
  - `PreprintService` model created, represents a many-to-one relationship with nodes and one-to-one with preprint providers
- API:
  - Any preprint now only has one provider, simplifying requests around that.
  - Update serialization, requests based on model changes
  - Make `flatten_relationships` flatten multiple relationships (h/t @pattisdr )
- Routing to make preprint guids go to the ember app
- Migration to create `PreprintService` objects for existing preprints
- `on_preprint_updated` webhook for SHARE
- Tests 
- Update preprint links on node pages
## Deployment Notes

Migrations:

```
python -m scripts.migration.migrate_node_preprints_to_preprintservices --dry --minutes ## --hours ##
```
- `--minutes` and `--hours` are used to specify the max `timedelta` between `preprint.date_created` and `node.date_created` to determine whether or not the preprint should retain the node's GUID. Defaults to 1 hour.

```
python -m scripts.migration.migrate_share_preprint_data

invoke migrate_search
```
- Makes outgoing requests to update SHARE data. To dry run, use a non-whitelisted `SHARE_API_TOKEN`
## Side effects

Breaks almost everything about `ember-preprints`.
## Ticket

[OSF-7049](https://openscience.atlassian.net/browse/OSF-7049)
[PREP-129](https://openscience.atlassian.net/browse/PREP-129)
## Example Requests/Responses

Turn a node into a preprint:
`POST /v2/preprints/`

```
{
  "data": {
    "type": "preprints",
    "attributes": {},
    "relationships": {
      "primary_file": {
        "data": {
          "type": "primary_files",
          "id": "575992518ca57e01f03d20b3"
        }
      },
      "node": {
        "data": {
          "type": "nodes",
          "id": "tfdc5"
        }
      },
      "provider": {
        "data": {
          "type": "providers",
          "id": "osf"
        }
      }
    }
  }
}
```

`201 Created`

```
{
  "data": {
    "relationships": {
      "node": {
        "links": {
          "related": {
            "href": "http://localhost:8000/v2/nodes/tfdc5/",
            "meta": {}
          }
        }
      },
      "primary_file": {
        "links": {
          "related": {
            "href": "http://localhost:8000/v2/files/575992518ca57e01f03d20b3/",
            "meta": {}
          }
        }
      },
      "provider": {
        "links": {
          "related": {
            "href": "http://localhost:8000/v2/preprint_providers/osf/",
            "meta": {}
          }
        }
      }
    },
    "links": {
      "self": "http://localhost:8000/v2/preprints/3yzdp/",
      "html": "http://localhost:5000/3yzdp/"
    },
    "attributes": {
      "date_created": "2016-09-27T05:12:41.434699",
      "is_published": false,
      "subjects": [],
      "doi": null,
      "date_published": null
    },
    "type": "preprints",
    "id": "3yzdp"
  }
}
```

Add subjects to a preprint:
`POST /v2/preprints/3yzdp/`

```
    {
      "data": {
        "type": "preprints",
        "id": "3yzdp",
        "attributes": {
          "subjects": [
            [
              "57e978b2c2babb0aa9d7ea39",
              "57e978b2c2babb0aa9d7ea48",
              "57e978a2c2babb0aa9d7e1c4"
            ],
            [
              "57e978a2c2babb0aa9d7e1c1",
              "57e978a2c2babb0aa9d7e1c2",
              "57e978a2c2babb0aa9d7e1c4"
            ]
          ]
        }
      }
    }
```

`200 OK`

```
{
  "data": {
    "relationships": {
      "node": {
        "links": {
          "related": {
            "href": "http://localhost:8000/v2/nodes/tfdc5/",
            "meta": {}
          }
        }
      },
      "primary_file": {
        "links": {
          "related": {
            "href": "http://localhost:8000/v2/files/575992518ca57e01f03d20b3/",
            "meta": {}
          }
        }
      },
      "provider": {
        "links": {
          "related": {
            "href": "http://localhost:8000/v2/preprint_providers/osf/",
            "meta": {}
          }
        }
      }
    },
    "links": {
      "self": "http://localhost:8000/v2/preprints/3yzdp/",
      "html": "http://localhost:5000/3yzdp/"
    },
    "attributes": {
      "date_created": "2016-09-27T05:12:41.434000",
      "is_published": false,
      "subjects": [
        [
          {
            "text": "Social and behavioral sciences",
            "id": "57e978b2c2babb0aa9d7ea39"
          },
          {
            "text": "Economics",
            "id": "57e978b2c2babb0aa9d7ea48"
          },
          {
            "text": "Agricultural economics",
            "id": "57e978a2c2babb0aa9d7e1c4"
          }
        ],
        [
          {
            "text": "Biology and life sciences",
            "id": "57e978a2c2babb0aa9d7e1c1"
          },
          {
            "text": "Agriculture",
            "id": "57e978a2c2babb0aa9d7e1c2"
          },
          {
            "text": "Agricultural economics",
            "id": "57e978a2c2babb0aa9d7e1c4"
          }
        ]
      ],
      "doi": null,
      "date_published": null
    },
    "type": "preprints",
    "id": "3yzdp"
  }
}
```


api/preprints/parsers.py
```diff
@@ -3,11 +3,12 @@
...
     def flatten_relationships(self, relationships):
-        ret = super(PreprintsJSONAPIParser, self).flatten_relationships(relationships)
-        related_resource = relationships.keys()[0]
-        if ret.get('target_type') and ret.get('id'):
-            return {related_resource: ret['id']}
-        return ret
+        rel = {}
+        for resource in relationships:
+            ret = super(PreprintsJSONAPIParser, self).flatten_relationships({resource: relationships[resource]})
+            if ret.get('target_type') and ret.get('id'):
```
brianjgeiger: As I recall, you're very limited with the information you have here, but this would be so much better if it got the relationship name rather, or in addition to, the target_type. This way you could have two relationships with the same type.

Also, this would probably close another ticket if it were the base flatten_relationships, but we can just make a note on that ticket to do what we did here if we need to.

api/preprints/serializers.py
```diff
@@ -75,184 +83,105 @@ class PreprintSerializer(JSONAPISerializer):
...
-        return 'https://dx.doi.org/{}'.format(obj.preprint_doi) if obj.preprint_doi else None
+        return 'https://dx.doi.org/{}'.format(obj.preprint_article_doi) if obj.preprint_article_doi else None
 
-    def create(self, validated_data):
-        node = Node.load(validated_data.pop('_id', None))
-        if not node:
-            raise exceptions.NotFound('Unable to find Node with specified id.')
+    def update(self, preprint, validated_data):
+        from website.models import Node
+        assert isinstance(preprint, PreprintService), 'You must specify a valid preprint to be updated'
```
brianjgeiger: I don't _think_ assert is the proper way to do this here. Don't they get stripped out in production code? I believe if statements are preferred.  http://www.django-rest-framework.org/api-guide/validators/#writing-custom-validators

Or is this just a thing that is meant to catch developers and not API consumers.

api/preprints/serializers.py
```diff
@@ -75,184 +83,105 @@ class PreprintSerializer(JSONAPISerializer):
...
     def get_doi_url(self, obj):
-        return 'https://dx.doi.org/{}'.format(obj.preprint_doi) if obj.preprint_doi else None
+        return 'https://dx.doi.org/{}'.format(obj.preprint_article_doi) if obj.preprint_article_doi else None
 
-    def create(self, validated_data):
-        node = Node.load(validated_data.pop('_id', None))
-        if not node:
-            raise exceptions.NotFound('Unable to find Node with specified id.')
+    def update(self, preprint, validated_data):
+        from website.models import Node
```
brianjgeiger: Circular import problem?

api/users/views.py
```diff
@@ -529,11 +529,14 @@ def get_default_odm_query(self):
...
         nodes = Node.find(self.get_query_from_request())
+        preprints = []
         # TODO: Rearchitect how `.is_preprint` is determined,
         # so that a query that is guaranteed to return only
-        # preprints can be contructed. Use generator in meantime.
-        return (node for node in nodes if node.is_preprint)
-
+        # preprints can be contructed.
+        for node in nodes:
+            for preprint in node.preprints:
```
brianjgeiger: I know there's the todo, and it's not your doing, but this makes me very unhappy. Don't change it obvs, but we do need a Critical-priority ticket for it and label the TODO with the ticket info. This will kill us in the near future unless it all just goes away with DjangOSF.

scripts/migration/migrate_node_preprints_to_preprintservices.py
```diff
@@ -0,0 +1,515 @@
...
+            if not tree[0].parents:
+                hierarchical_subjects.append([s._id for s in tree])
+            else:
+                trees.extend([(p, ) + tree for p in tree[0].parents if p._id in flat_subjects])
+    assert set(flat_subjects) == set(sum(hierarchical_subjects, [])), \
+        'Flat subject set `{}` not equal to hierarchical subject set `{}`'.format(flat_subjects, hierarchical_subjects)
+    preprint.subjects = hierarchical_subjects
+    preprint.save()
+
+def migrate(dry_run=True):
```
brianjgeiger: Not gonna lie: this is an impressive looking migration script.

tests/test_models.py
```diff
@@ -1613,17 +1613,17 @@ def test_validate_categories(self):
...
-            Node(preprint_doi='doi:10.10.1038/nwooo1170').save()  # should save without doi: prefix
+            Node(preprint_article_doi='doi:10.10.1038/nwooo1170').save()  # should save without doi: prefix
 
     def test_validate_good_doi(self):
         doi = '10.10.1038/nwooo1170'
-        self.node.preprint_doi = doi
+        self.node.preprint_article_doi = doi
         self.node.save()
-        assert_equal(self.node.preprint_doi, doi)
+        assert_equal(self.node.preprint_article_doi, doi)
```
brianjgeiger: I know we are all "test the view not the model", but since you have a fancy new preprint model, it might be good to add some tests for that. Especially since you have permissions and stuff on it. This could be put on a new ticket if we're in a rush, but make it a high-priority Maintenance task and bump it to like spot 3-5 in the Maintenance backlog.

website/preprints/model.py
```diff
@@ -0,0 +1,183 @@
...
+            for subj_id in subj_list:
+                subj = Subject.load(subj_id)
+                if subj:
+                    subj_hierarchy += ({'id': subj_id, 'text': subj.text}, )
+            if subj_hierarchy:
+                ret.append(subj_hierarchy)
+        return ret
+
+    def set_preprint_subjects(self, preprint_subjects, auth, save=False):
+        if not self.node.has_permission(auth.user, ADMIN):
```
brianjgeiger: So fancy!

website/preprints/tasks.py
```diff
@@ -0,0 +1,129 @@
...
+
+import requests
+
+from framework.celery_tasks import app as celery_app
+
+from website import settings
+
+
+@celery_app.task(ignore_results=True)
+def on_preprint_updated(preprint_id):
```
brianjgeiger: In side effects, could you sketch out an integration test (like Runscope) that could verify if there might be any async problems with the celery tasks?

website/project/model.py
```diff
@@ -760,7 +757,7 @@ def __init__(self, reason, key, *args, **kwargs):
 
 def validate_doi(value):
     # DOI must start with 10 and have a slash in it - avoided getting too complicated
-    if not re.match('10\\.\\S*\\/', value):
+    if value and not re.match('10\\.\\S*\\/', value):
```
brianjgeiger: There should have been a ticket on Preprints that added yesterday based on talk in the Preprints flow. Could you see if you think this will take care of it and, if so, link the PR to that ticket for testing?

mfraezz: That was relating to the regex being not-great. This was added to allow someone to unset `preprint_article_doi` -- the validator didn't allow `''`

website/preprints/model.py
```diff
@@ -0,0 +1,189 @@
...
+
+    # This is a list of tuples of Subject id's. MODM doesn't do schema
+    # validation for DictionaryFields, but would unsuccessfully attempt
+    # to validate the schema for a list of lists of ForeignFields.
+    #
+    # Format: [[root_subject._id, ..., child_subject._id], ...]
+    subjects = fields.DictionaryField(list=True)
+
+    @property
+    def preprint_file(self):
```
icereval: having `preprint_` is duplicative w/ the service class name, any reason to not make it like the other attributes?

api/preprints/serializers.py
mfraezz: `nose.tools`-style asserts (`assert_equal`, etc.) won't work in production because `nose` isn't installed. Normal `assert` statements will because it's a python built-in. From [stackoverflow -- Best practice for Python Assert](http://stackoverflow.com/questions/944592/best-practice-for-python-assert#945135):

>Asserts should be used to test conditions _that should never happen_. The purpose is to crash early in the case of a corrupt program state.

That being said, `PreprintMixin` should catch this early if it's not a `PreprintService` object, and the call to `check_object_permissions` in it should do so for `User`/`Node` permissions. This is just an extra verification, and could be removed.

api/preprints/serializers.py
mfraezz: No. Dunno how that got there, removing.

api/users/views.py
mfraezz: [Ticket](https://openscience.atlassian.net/browse/OSF-7090) created and linked.

website/preprints/model.py
mfraezz: Updated in https://github.com/CenterForOpenScience/osf.io/pull/6346/commits/6669c1f9fb06ed41aa848299786875685671f0d5

website/preprints/tasks.py
mfraezz: I'm fairly certain that this would not be a suitable async task for the job. This should be ran when preprints are created or modified, but the only way to verify that it is would be to check SHARE.

Assuming that everything is working fine, this would take on average about 5 minutes to be surfaced in SHARE results (longer if SHARE is slammed, as this call is also handled async on the SHARE side), which isn't great turnaround for an automated test. Also, this would need to "publish" a preprint in order to be observable, and we try to keep bad data out of SHARE.

tests/test_models.py
mfraezz: Added in https://github.com/CenterForOpenScience/osf.io/pull/6346/commits/7c4c0c7a1027e2064e47d5c4e24c3ac65dbc3b58:
* model tests for permissions
* view tests for `NodePreprintsList`
* validator tests for `.set_subjects`

mfraezz:
>I think this just needs to prevent setting is_published from `True` to `False`.

Added in https://github.com/CenterForOpenScience/osf.io/pull/6346/commits/8a8998c424dec4b61cbc06c12d9444a1ad5c31c8

mfraezz:
Actually, because something that should've been blocked by this made it into production, the migration script needs one further tweak.

api/preprints/views.py
```diff
@@ -108,20 +97,26 @@ class PreprintList(JSONAPIBaseView, generics.ListCreateAPIView, ODMFilterMixin):
...
-                                "tags":          [{tag1}, ...],       # optional
-                                "provider":      {provider}           # optional
-                            },
+                            "attributes": {},
                             "relationships": {
-                                "primary_file": {                     # required
+                                "node": {                           # required
                                     "data": {
-                                        "type": "primary",
+                                        "type": "node",
```
pattisdr: We're not validating this right now, but aren't types plural?

brianjgeiger: Yeah, type should be plural but the relationship can be singular.

mfraezz: Good catch. Updated in https://github.com/CenterForOpenScience/osf.io/pull/6346/commits/5e6feee9b0c1fb6a8457ff4502605ad74bd2d9a2

tests/test_preprints.py
```diff
@@ -39,10 +38,10 @@ def setUp(self):
         self.preprint.save()
 
     def test_is_preprint(self):
-        assert_true(self.preprint.is_preprint)
+        assert_true(self.preprint.node.is_preprint)
```
brianjgeiger: I feel like this should be `has_preprint` if it's on the node, but.

website/preprints/tasks.py
```diff
@@ -0,0 +1,129 @@
...
+        'url': urlparse.urljoin(settings.DOMAIN, preprint.url),
+        'type': 'provider'
+    }]] + [
+        format_contributor(preprint, user) for user in preprint.node.contributors
+    ] + [
+        format_subjects(preprint)
+    ] + [[{
+        '@id': '_:{}-{}'.format(preprint._id, institution._id),
+        '@type': 'association',
+        'institution': {
```
icereval: `institution` --> `entity` h/t @chrisseto

website/preprints/tasks.py
```diff
@@ -0,0 +1,129 @@
...
+    from website.models import PreprintService
+    preprint = PreprintService.load(preprint_id)
+
+    if settings.SHARE_URL and settings.SHARE_API_TOKEN:
+        requests.post('{}api/normalizeddata/'.format(settings.SHARE_URL), json={
+            'created_at': datetime.datetime.utcnow().isoformat(),
+            'normalized_data': {
+                '@graph': format_preprint(preprint)
+            },
+        }, headers={'Authorization': 'Bearer {}'.format(settings.SHARE_API_TOKEN)}).raise_for_status()
```
mfraezz: Added in https://github.com/CenterForOpenScience/osf.io/pull/6346/commits/69bec39d6be765f48320b57d08c41de7e68844b1

pattisdr:
View preprint on project overview page still links to the old preprints/node_id instead of the new /preprint_guid.

pattisdr:
@mfraezz, are we taking orphans out in this branch, or will that come alongside edit?

brianjgeiger:
That should be with edit.

pattisdr:
OK. If we still have orphans, then I don't think loading an orphan is working.Currently an orphaned preprint loads, but there's no file.  

Should we redirect to a not-found if there's not primaryFile for an orphan, or should we load the orphan and say that the file has been removed?  Just a temporary solution until orphans are no more.


screen shot 2016-10-25 at 2 01 32 pm

mfraezz:
@saradbowman indicates that the preprint should load and display an error message in place of the file, as this still gives other metadata about what was here.

saradbowman:
Maybe the message could say:

"The user has removed this file."

pattisdr:
ok! I will make the change here https://github.com/CenterForOpenScience/ember-preprints/pull/184

website/routes.py
```diff
@@ -172,25 +171,29 @@ def robots():
         mimetype='text/plain'
     )
 
-
-def external_ember_app(_=None):
+def ember_app(_=None):
     """Serve the contents of the ember application"""
-    external_app_url = None
+    if _ and Guid.load(_):
+        return redirect(_)
```
icereval: no longer needed

website/routes.py
```diff
@@ -172,25 +171,29 @@ def robots():
...
 
+    ember_app_folder = None
+    file = _ or 'index.html'
```
icereval: reserved word

website/preprints/tasks.py
```diff
@@ -0,0 +1,153 @@
...
+            '@id': '_:{}'.format(user._id),
+            '@type': 'person',
+        }
+    }] + format_institution(institution) for institution in user.affiliated_institutions], [])
+
+def format_contributor(preprint, user):
+    return [{
+        '@id': '_:{}-{}'.format(preprint._id, user._id),
+        '@type': 'contributor',
+        'person': {'@id': '_:{}'.format(user._id), '@type': 'person'},
```
chrisseto: `'cited_name': user.fullname,`

mfraezz: Added in https://github.com/CenterForOpenScience/osf.io/pull/6346/commits/268a7a9a7db6b7ea5a5c23d41c154507519f714f

chrisseto:
@mfraezz Can a check be added to on_node_update to call on_preprint_update for any preprints?

mfraezz:
@chrisseto, added in https://github.com/CenterForOpenScience/osf.io/pull/6346/commits/854a70781d7bf750356bd66d1afb6d1850d898a5

PR #6351, opened 2016-09-28
[feature] Hide Institution Logos with < 5 Affiliated Nodes [OSF-6980]

caseyrollins:
#### Purpose
- Hides institution logos from the dashboard institution carousel until there are at least 5 projects (non-registered, public, top-level nodes) affiliated with the institution.

#### Changes
- Replaces `all_institutions` global with `dashboard_institutions`

#### Side effects
- Users may not know a given institution exists on the OSF until there are enough affiliated projects to display the logo. 

#### Ticket
- [OSF-6980](https://openscience.atlassian.net/browse/OSF-6980)

### DIFF
--- a/website/routes.py
+++ b/website/routes.py
@@ -63,7 +63,9 @@ def get_globals():
     dashboard_institutions = [
         {'id': inst._id, 'name': inst.name, 'logo_path': inst.logo_path_rounded_corners}
         for inst in all_institutions
-        if len(Node.find_by_institutions(inst, query=Q('is_public', 'eq', True))) >= settings.INSTITUTION_DISPLAY_NODE_THRESHOLD
+        if len(
+            Node.find_by_institutions(inst, query=(Q('is_public', 'eq', True) & Q('is_registration', 'eq', False)))
+        ) >= settings.INSTITUTION_DISPLAY_NODE_THRESHOLD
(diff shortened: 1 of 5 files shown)


website/routes.py
```diff
@@ -58,7 +58,20 @@ def get_globals():
     user = _get_current_user()
 
     user_institutions = [{'id': inst._id, 'name': inst.name, 'logo_path': inst.logo_path_rounded_corners} for inst in user.affiliated_institutions] if user else []
-    all_institutions = [{'id': inst._id, 'name': inst.name, 'logo_path': inst.logo_path_rounded_corners} for inst in Institution.find().sort('name')]
+    all_institutions = [inst for inst in Institution.find().sort('name')]
+    dashboard_institutions = [
+        {'id': inst._id, 'name': inst.name, 'logo_path': inst.logo_path_rounded_corners}
+        for inst in all_institutions
+        if len(
+            Node.find_by_institutions(inst, query=(
```
sloria: @caseyrollins Why does this need to go in `get_globals`. This is relatively expensive query that will happen on every request. Can this be moved to the view for the Dashboard?

sloria: @caseyrollins Thanks a lot!

PR #6356, opened 2016-09-28
Anonymous links should not provide comment information [OSF-7062]

acshi:
## Purpose

Anonymous view-only links are supposed to remove all identifying information for a blind peer review of a private project, but comments have the potential for leaking information about contributors by nature of the discursive nature, including explicit @mentions and other more casual mentions.
## Changes

Comment API returns forbidden error when using an anonymous link.
Comment pane is not shown when using an anonymous link.
Comment API test checks for 403 error when using an anonymous link.
Removed dead code changing comment author to "A User."

Before:
anonymouswithcomments

After:
anonymouswithoutcomments
## Side effects

Comment API will be inaccessible from anonymous links.
## Ticket

https://openscience.atlassian.net/browse/OSF-7062

### DIFF
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -359,12 +359,13 @@ class AnonymizedRegexField(AuthorizedCharField):
     def get_attribute(self, obj):
         value = super(AnonymizedRegexField, self).get_attribute(obj)
 
-        user = self.context['request'].user
-        auth = auth_core.Auth(user)
-        if 'view_only' in self.context['request'].query_params:
-            auth.private_key = self.context['request'].query_params['view_only']
-            if has_anonymous_link(obj.node, auth):
-                value = re.sub(self.regex, self.replace, value)
+        if value:
+            user = self.context['request'].user
+            auth = auth_core.Auth(user)
+            if 'view_only' in self.context['request'].query_params:
+                auth.private_key = self.context['request'].query_params['view_only']
+                if has_anonymous_link(obj.node, auth):
+                    value = re.sub(self.regex, self.replace, value)
 
         return value
 
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -334,6 +334,8 @@ class AuthorizedCharField(ser.CharField):
     def get_attribute(self, obj):
         user = self.context['request'].user
         auth = auth_core.Auth(user)
+        if 'view_only' in self.context['request'].query_params:
+            auth.private_key = self.context['request'].query_params['view_only']
         field_source_method = getattr(obj, self.source)
         return field_source_method(auth=auth)
 
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -327,8 +327,7 @@ class AuthorizedCharField(ser.CharField):
         content = AuthorizedCharField(source='get_content')
     """
 
-    def __init__(self, source=None, **kwargs):
-        assert source is not None, 'The `source` argument is required.'
+    def __init__(self, source, **kwargs):
         self.source = source
         super(AuthorizedCharField, self).__init__(source=self.source, **kwargs)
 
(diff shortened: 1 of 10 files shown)


api/comments/permissions.py
```diff
@@ -29,6 +31,8 @@ def has_object_permission(self, request, view, obj):
         auth = get_user_auth(request)
         comment = obj
         node = obj.node
+        if has_anonymous_link(node, auth):
```
brianjgeiger: I talked with Nici, and this is probably not the best way of dealing with this problem, because it'll be a breaking change to the API. Instead, if the view is hit with an anonymous view only link, have it go through the `content` field and look for all the things that look like `[@xxxxxx](xxxxxx)` with stuff instead of just xxxxxx and replace them with something like _Anonymous User Mention_. Ensure that this isn't being hit by the OSF in the case of an anonymous view only link (so make sure the calls to v2/nodes/nid/comments/ is in the `comment_pane_template.mako` and not anywhere above that, which I'm sure it is but just to be sure).

Alternately, you could play around with making a custom text field that can check for if it's anonymous and run a regular expression on the field. That would be better but I don't know how difficult it would be. But it's worth exploring at least some.

acshi:
@brianjgeiger Is this what you think should be good? The comment pane is still hidden from the user for an anonymous view, but the API will still allow queries for comments, but will replace mentions with "@A User". Also, there is one spurious test failing, but everything related to the PR passes.

api/base/serializers.py
```diff
@@ -292,6 +292,8 @@ def __init__(self, source=None, **kwargs):
     def get_attribute(self, obj):
         user = self.context['request'].user
         auth = auth_core.Auth(user)
+        if 'view_only' in self.context['request'].query_params:
+            auth.private_key = self.context['request'].query_params['view_only']
```
brianjgeiger: What's this for? It doesn't seem to be used and I don't _think_ it should be necessary.

website/project/model.py
```diff
@@ -282,7 +282,11 @@ def get_content(self, auth):
                                 or (auth and not auth.user.is_anonymous() and self.user._id != auth.user._id)):
             return None
 
-        return self.content
+        content = self.content
+        if has_anonymous_link(self.node, auth):
+            content = re.sub('\[@[^\]]*\]\([^\) ]*\)', '@A User', content)
```
brianjgeiger: I'm not necessarily opposed to having this in the model, but I think you should be able to handle this in the serializer. You'd make a field type like AnonymizedRegexField (or you might just make a field composition, depending).

api/base/serializers.py
acshi: @brianjgeiger Normally the Auth object has the private_key set when a request is made with a view_only link, however in the case that it is made directly from the user like this, the private_key is always null, so calling has_anonymous_link(self.node, auth) will always be false with such an auth object. In order for that call to later succeed, we need to make sure the auth object has that information.

api/base/serializers.py
```diff
@@ -337,6 +338,36 @@ def get_attribute(self, obj):
...
+    """
+    Performs a regex replace on the content of the authorized object's
+    source field when an anonymous view is requested.
+
+    Example:
+        content = AnonymizedRegexField(source='get_content', regex='\[@[^\]]*\]\([^\) ]*\)', replace='@A User')
+    """
+
+    def __init__(self, source=None, regex=None, replace=None, **kwargs):
+        assert source is not None, 'The `source` argument is required.'
```
brianjgeiger: Can't you just not put default values in to have them be required?

acshi: I followed what the AuthorizedCharField had done, but yeah, it didn't seem necessary at all to have that.

brianjgeiger:
Closed in favor of #7170

PR #6361, opened 2016-09-29
[feature/django-osf] Fix Timezone [#CAS-9]

cslzchen:
### Issue

@sloria @chris

Python's `datetime.datetime.utcnow()` is not timezone aware, when assigned to a Django-Postgres `DateTimeField`, it is treated as local time. Date of type `Timestamp with Time Zone` stored in Postgres has UTC timestamp but local timezone.
### Fix

Use Django's `django.util.timezone.now()` instead.

UTC timestamp in Django: `2016-09-29 19:03:21.439057`
Date stored in DB (before fix): `2016-09-29 19:03:21.439057-04`
Date stored in DB (after fix): `2016-09-29 15:03:21.439057-04`
### Changes

Replace AMOST EVERY occurrence of `dt.datetime.utcnow()`, `datetime.utcnow()` and `datetime.datetime.utcnow()` with `timezone.now()`. Add `from django.util import timezone` if necessary.
- Exceptions
  - `\admin\` and `\admin_tests\`
  - `utcnow = timezone.now()` where `utcnow` is used as a var
  - `website.util.time.py` where `datetime` and `timezone` are handled separately
  - deprecated model declaration `fields.DateTimeField(default=datetime.datetime.utcnow)`
### Side effects
- [ ] What about timestamp in `DateTimeAwareJSONField()`?
- [x] What about those in `osf-models`: see https://github.com/CenterForOpenScience/osf-models/pull/142
### Related PR

OSF-Models: https://github.com/CenterForOpenScience/osf-models/pull/142
## Ticket

https://openscience.atlassian.net/browse/CAS-9


cslzchen:
@cwisecarver 
this PR duplicates your PR for fixing `datetime` issue, closed

PR #6362, opened 2016-09-30
[feature] Handle Registered and Logged-out User Using Unregistered Claim Link [OSF-6857]

caseyrollins:
#### Purpose
- In the workflow of adding unregistered contributors and claiming accounts, there is an uncovered condition of a logged-out user using an unregistered contributor claim link after they have created an account with the same email they claimed to be a contributor with. 
- Steps to reproduce:
 - As an admin on a public project, add an unregistered contributor by just their name, not email. 
 - As a logged-out user visiting said public project, click on the unregistered contributor name and enter an email to claim yourself. 
 - Note that two emails are sent: one sent to the project admin who added the unregistered contributor (with a claim link), and one sent to the email of the logged-out user who just claimed themselves.
 - *do not click the claim link yet* 
 - As a logged-out user, sign up for an account, using the same email the logged-out user used to claim themselves earlier.
 - Confirm the newly created account via the account confirmation email.
 - As a _**logged-out**_ user, click the unregistered contributor claim link that was sent to the project admin.
    - Current outcome:
      - The claim link goes to the "set a password" page. (This is incorrect because an account has been created and confirmed with this email and a password has already been set -- attempting to complete the form in this page will error because the user already exists.)
    - Outcome after this PR:
      - The claim link automatically redirects to the create account/login page, where the user can login and then will be redirected to the "verify password" page where after password verification, they will successfully be added as a contributor to the project.

#### Changes
- Adds a redirect to `claim_user_registered` if the "unregistered" email exists as a given registered username in the database.

#### Side effects
- None expected.

#### Ticket
- [OSF-6857](https://openscience.atlassian.net/browse/OSF-6857)

### DIFF
--- a/website/project/views/contributor.py
+++ b/website/project/views/contributor.py
@@ -3,7 +3,6 @@
 import httplib as http
 
 from flask import request
-from modularodm import Q
 from modularodm.exceptions import ValidationError, ValidationValueError
 
 from framework import forms, status
--- a/website/project/views/contributor.py
+++ b/website/project/views/contributor.py
@@ -671,8 +671,8 @@ def claim_user_form(auth, **kwargs):
     claimer_email = unclaimed_record.get('claimer_email') or unclaimed_record.get('email')
 
     # If there is a registered user with this email, redirect to 're-enter password' page
-    user = User.find_by_email(claimer_email)
-    user_from_email = user[0] if user else None
+    found_by_email = User.find_by_email(claimer_email)
+    user_from_email = found_by_email[0] if found_by_email else None
     if user_from_email and user_from_email.is_registered:
         return redirect(web_url_for('claim_user_registered', uid=uid, pid=pid, token=token))
 
--- a/website/project/views/contributor.py
+++ b/website/project/views/contributor.py
@@ -3,6 +3,7 @@
 import httplib as http
 
 from flask import request
+from modularodm import Q
 from modularodm.exceptions import ValidationError, ValidationValueError
 
 from framework import forms, status
@@ -669,6 +670,13 @@ def claim_user_form(auth, **kwargs):
     user.update_guessed_names()
     # The email can be the original referrer email if no claimer email has been specified.
     claimer_email = unclaimed_record.get('claimer_email') or unclaimed_record.get('email')
+
+    # If there is a registered user with this email, redirect to 're-enter password' page
+    user = User.find_by_email(claimer_email)
+    user_from_email = user[0] if user else None
+    if user_from_email and user_from_email.is_registered:
+        return redirect(web_url_for('claim_user_registered', uid=uid, pid=pid, token=token))
+
     form = SetEmailAndPasswordForm(request.form, token=token)
     if request.method == 'POST':
         if not form.validate():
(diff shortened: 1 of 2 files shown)


website/project/views/contributor.py
```diff
@@ -669,6 +670,12 @@ def claim_user_form(auth, **kwargs):
     user.update_guessed_names()
     # The email can be the original referrer email if no claimer email has been specified.
     claimer_email = unclaimed_record.get('claimer_email') or unclaimed_record.get('email')
+
+    # If there is a registered user with this email, redirect to 're-enter password' page
+    user_from_email = User.find_one(Q('username', 'eq', claimer_email))
```
mfraezz: What about the edge case to this edge case where a user has confirmed this address but it's not their primary? i.e. `Q('emails', 'eq', claimer_email)`

caseyrollins: More edgy: 	9830a93

cslzchen: @mfraezz The above [commit](https://github.com/CenterForOpenScience/osf.io/commit/b885376a04cd40ad995716ea04536a06d1819350) is not in develop (not merged)?

cslzchen: @brianjgeiger the current [dev branch](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/project/views/contributor.py#L674) looks different: 

In https://github.com/CenterForOpenScience/osf.io/commit/b885376a04cd40ad995716ea04536a06d1819350:

```python
# If there is a registered user with this email, redirect to 're-enter password' page
    user = User.find_by_email(claimer_email)
    user_from_email = user[0] if user else None
    if user_from_email and user_from_email.is_registered:
        return redirect(web_url_for('claim_user_registered', uid=uid, pid=pid, token=token))
```

In develop:
```python
 # If there is a registered user with this email, redirect to 're-enter password' page
    found_by_email = User.find_by_email(claimer_email)
    user_from_email = found_by_email[0] if found_by_email else None
    if user_from_email and user_from_email.is_registered:
        return redirect(web_url_for('claim_user_registered', uid=uid, pid=pid, token=token))
```

brianjgeiger: @cslzchen  So what it looks like to me based on the [blame](https://github.com/CenterForOpenScience/osf.io/blame/develop/website/project/views/contributor.py#L674) is that they're both in there but [540aaac](https://github.com/CenterForOpenScience/osf.io/pull/6362/commits/540aaac5a719c8e099cb97489b58a5e16a8e7502) is overwriting [b885376](https://github.com/CenterForOpenScience/osf.io/pull/6362/commits/b885376a04cd40ad995716ea04536a06d1819350) because it came later, which tracks. Is there a problem with how it is right now? @caseyrollins is going to get with you to see if there needs to be changes.

caseyrollins: @cslzchen, it looks like develop does align with the most recent commit for this PR (540aaac). The commit you're referencing (b885376) wasn't the last commit for this PR and was overridden by changes in 7df1d9d and 540aaac.

cslzchen: @brianjgeiger @caseyrollins thanks a lot, totally my fault, I had thought https://github.com/CenterForOpenScience/osf.io/commit/b885376a04cd40ad995716ea04536a06d1819350 was the more edgy commit without realizing it was an old one.

PR #6371, opened 2016-10-04
[hotifx]Inheriting unclaimed users in new component seems not to work	[#OSF-7072]

chennan47:
## Purpose

Fix the problem that when you have a unregistered contributor in parent node you cannot create a component and inherit contributor from parent project.
## QA Notes
Before fix:
1. create a project and add an unconfirmed contributor to the project
2. add a component to that project and inherit contributor from the project
3. you will see an error and the component is not successfully created
After fix:
1. create a project and add an unconfirmed contributor to the project
2. add a component to that project and inherit contributor from the project
3. there will be a normal pop up to ask you if you want to stay on the project or go to the component created
4. go to the component created. You should see all contributors from the parent project being successfully inherited including the unregistered contributor

## Ticket

https://openscience.atlassian.net/browse/OSF-7072

### DIFF
--- a/website/project/model.py
+++ b/website/project/model.py
@@ -3326,7 +3326,7 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin, Commentable, Spam
         if save:
             self.save()
 
-    def add_unregistered_contributor(self, fullname, email, auth, send_email='default', visible=True, permissions=None, save=False):
+    def add_unregistered_contributor(self, fullname, email, auth, send_email='default', visible=True, permissions=None, save=False, user_created=None):
         """Add a non-registered contributor to the project.
 
         :param str fullname: The full name of the person.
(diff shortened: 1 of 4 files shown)


website/project/model.py
```diff
@@ -3335,8 +3335,8 @@ def add_unregistered_contributor(self, fullname, email, auth, send_email='defaul
         :returns: The added contributor
```
brianjgeiger: Update the parameters to have the full set for the current function. It's your new parameter plus probably some of the others.

website/project/model.py
```diff
@@ -3326,7 +3326,7 @@ def add_contributors(self, contributors, auth=None, log=True, save=False):
         if save:
             self.save()
 
-    def add_unregistered_contributor(self, fullname, email, auth, send_email='default', visible=True, permissions=None, save=False):
+    def add_unregistered_contributor(self, fullname, email, auth, send_email='default', visible=True, permissions=None, save=False, user_created=None):
```
brianjgeiger: Let's call this `existing_user` rather than `user_created`, since you're passing in a user object or None, right?

website/project/model.py
```diff
@@ -3335,8 +3335,8 @@ def add_unregistered_contributor(self, fullname, email, auth, send_email='defaul
         :returns: The added contributor
         :raises: DuplicateEmailError if user with given email is already in the database.
         """
-        # Create a new user record
-        contributor = User.create_unregistered(fullname=fullname, email=email)
+        # Create a new user record if the user haven't been created
```
brianjgeiger: "Create a new user record if you weren't passed an existing_user"

PR #6373, opened 2016-10-04
[WIP] Port comment tests for osf_models

erinspace:
## Purpose
Move model tests from `tests/test_comments.py` into `osf_models_tests/test_comment.py`, and fix a few things along the way

## Changes
- port tests
- move `test_find_unread_includes_edited_comments` into `tests/test_views.py`
- Add `NotificationDigest` to app.py
- Check for users to send to in `emails.py`, as [] does not have `.all()` method
- refactor `validate_contributor` method a bit to get user from GUID and check after the fact

## Side effects
None anticipated

## Ticket
https://github.com/CenterForOpenScience/osf-models/issues/132

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -17,6 +17,7 @@ from django.apps import apps
 
 from modularodm import Q
 from modularodm.exceptions import ValidationError
+from osf_models.models import Comment
 
 from framework import auth
 from framework.auth import User, Auth
@@ -63,6 +64,7 @@ from osf_models_tests.factories import (
     PrivateLinkFactory,
     ProjectFactory,
     NodeFactory,
+    CommentFactory,
     InstitutionFactory,
     RegistrationFactory,
     ApiOAuth2ApplicationFactory,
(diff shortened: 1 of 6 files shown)


tests/test_views.py
```diff
@@ -4359,5 +4363,92 @@ def test_confirm_user_signal_called_when_user_confirms_email(self):
...
+    def test_view_project_comments_updates_user_comments_view_timestamp(self):
+        url = self.project.api_url_for('update_comments_timestamp')
+        res = self.app.put_json(url, {
+            'page': 'node',
+            'rootId': self.project._id
+        }, auth=self.user.auth)
+        self.user.reload()
+
+        user_timestamp = self.user.comments_viewed_timestamp[self.project._id]
+        view_timestamp = dt.datetime.utcnow()
```
sloria: Minor: we should use django.utils.timezone.now instead of utcnow to mimic how datetimes are stored in app code. Not a big deal here though.

PR #6374, opened 2016-10-05
Update metadate for the 0.95.0 release [OSF-7045]

brianjgeiger:
## Purpose

There's a new release. Let's make sure people know that it's new and what's in it

## Changes

1. Entries added to CHANGELOG
2. Version number increased in package.json

## Side effects

That would be super-impressive, but no.

## Ticket

https://openscience.atlassian.net/browse/OSF-7045

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,26 @@
 Changelog
 *********
 
+0.95.0 (2016-09-22)
+===================
+- Replace GPL-only django-queryset-csv library (the last of the GPL)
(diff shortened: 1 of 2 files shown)


CHANGELOG
```diff
@@ -2,6 +2,26 @@
 Changelog
 *********
 
+0.95.0 (2016-09-22)
```
mfraezz: >2016-09-22

This seems incorrect.

PR #6375, opened 2016-10-05
[Feature] Add extra search terms [OSF-7071]

alexschiller:
## Purpose

Allow search queries with dashes, periods, and underscores to be searchable. Specifically, files and nodes ares now searchable ignoring - _ and .

Previously the project named Bohemian_Rhapsody would only be found by searching the full name with the underscore. Now  Bohemian_Rhapsody can be found via searching Bohemian or Rhapsody. 

## Changes

Add extra_search_terms elasticdoc and filedoc in elastic_search. extra_search_terms is just a version of the file name or node name with - _ and . replaced with spaces allowing all values to be indexed. If no - _ or . are in the file/node name then no extra search terms are added.

## Side effects

1. This change should only increase the number of search results and not remove any results.
2. It is possible that files and nodes with - _ and . will slightly increase in search ranking due to more matching terms
3. This will increase the size of the elasticsearch index, but the amount is limited by only adding extra search terms for those containing - _ or .
4. Will need to `invoke migrate_search` when deploying.

## Ticket
https://openscience.atlassian.net/browse/OSF-7071

## QA Considerations
1. Files/Nodes containing - _ and . should be tested to ensure they are findable via search
2. This should not affect contributors searching, although it is possible they already handle - for hyphenated names. Testing contributors is probably overkill.
3. Test a file/node can be found both ways: Bohemian_Rhapsody should be found via searching Bohemian_Rhapsody, Bohemian, and Rhapsody.

### DIFF
--- a/tests/test_search/test_elastic.py
+++ b/tests/test_search/test_elastic.py
@@ -391,6 +391,27 @@ class TestPublicNodes(SearchTestCase):
         docs = query('category:component AND ' + self.title)['results']
         assert_equal(len(docs), 0)
 
+    def test_search_node_partial(self):
+        self.project.set_title('Blue Rider-Express', self.consolidate_auth)
+        with run_celery_tasks():
+            self.project.save()
+        find = query('Blue')['results']
+        assert_equal(len(find), 1)      
+
+    def test_search_node_partial_with_sep(self):
+        self.project.set_title('Blue Rider-Express', self.consolidate_auth)
+        with run_celery_tasks():
+            self.project.save()        
+        find = query('Express')['results']
+        assert_equal(len(find), 1)
+
+    def test_search_node_not_name(self):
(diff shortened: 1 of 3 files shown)


tests/test_search/test_elastic.py
```diff
@@ -391,6 +391,27 @@ def test_make_private(self):
...
+            self.project.save()        
+        find = query('Express')['results']
+        assert_equal(len(find), 1)
+
+    def test_search_node_not_name(self):
+        self.project.set_title('Blue Rider-Express', self.consolidate_auth)
+        with run_celery_tasks():
+            self.project.save()        
+        find = query('Green Flyer-Slow')['results']
+        assert_false(len(find), 1)
```
brianjgeiger: This would be better as `assert_equal(len(find), 0)`

brianjgeiger:
🎉 🐧 🎉 (Presuming tests pass)

PR #6376, opened 2016-10-05
[feature] Return Affiliated Components in Institution's Node List [OSF-7081]

caseyrollins:
#### Purpose
- Allows affiliated components to be returned in an institution's node list (`/v2/institutions/<inst-id>/nodes/`), even if their parent node is unaffiliated.

#### Changes
- Removes the `Q('parent_node', 'eq', None)` check in the `InstitutionNodesList` API view.

#### Side effects
- None expected.

#### QA Notes
- This changes the list of nodes affiliated with an institution returned from the API, which by default also changes the list of nodes displayed on an institution's dashboard (projects page). 

#### Ticket
- [OSF-7081](https://openscience.atlassian.net/browse/OSF-7081)

### DIFF
--- a/api/institutions/views.py
+++ b/api/institutions/views.py
@@ -157,7 +157,6 @@ class InstitutionNodeList(JSONAPIBaseView, ODMFilterMixin, generics.ListAPIView,
         Q('is_deleted', 'ne', True) &
         Q('is_folder', 'ne', True) &
         Q('is_registration', 'eq', False) &
-        Q('parent_node', 'eq', None) &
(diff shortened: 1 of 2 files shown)


brianjgeiger:
The change implemented seems different from what was described on the ticket. On the ticket it sounds like they want something akin to how projects and components have been differentiated (only show the component if the parent/root aren't in the list), which is significantly more difficult. When you demo, please have a range of examples from one node on its own to a series of nested components in a project that have all of the nodes attached to the institution to one where the parent and grandchild are affiliated but the child is not.

PR #6377, opened 2016-10-05
[Feature] Fix file seach result path [OSF-7078]

alexschiller:
## Purpose
Files without guids in search lead to file not found page. This fixes the broken url.

## Changes

Wraps the file_ object with .wrapper() to properly update the file_.path for OsfStorage files.

## Side effects

None known

## Deployment notes
```
invoke migrate_search
```

## Ticket

https://openscience.atlassian.net/browse/OSF-7078

## QA Considerations
1. Test that files found through search with and without guids lead to the file when clicked.
2. Test that files on OsfStorage and at least one other provider properly redirect.

### DIFF
--- a/website/search/elastic_search.py
+++ b/website/search/elastic_search.py
@@ -355,7 +355,7 @@ def update_node(node, index=None, bulk=False, async=False):
 
     from website.files.models.osfstorage import OsfStorageFile
     for file_ in paginated(OsfStorageFile, Q('node', 'eq', node)):
-        update_file(file_, index=index)
+        update_file(file_.wrapped(), index=index)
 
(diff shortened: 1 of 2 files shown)


website/search/elastic_search.py
```diff
@@ -355,7 +355,7 @@ def update_node(node, index=None, bulk=False, async=False):
 
     from website.files.models.osfstorage import OsfStorageFile
     for file_ in paginated(OsfStorageFile, Q('node', 'eq', node)):
-        update_file(file_, index=index)
+        update_file(file_.wrapped(), index=index)
```
brianjgeiger: Can you make a test or two for this?

PR #6378, opened 2016-10-05
[feature] Check User Permissions When Serializing File Moved Log [OSF-6521]

caseyrollins:
#### Purpose
- Currently, when a file is moved from a private component to a component a user has access to, the title of the private component is exposed in the API via the `/v2/nodes/<node-id>/logs/` endpoint. This PR checks user permissions before serializing the titles of source nodes in a file move log. 

#### Changes
- Adds a serializer method field to determine if this user has permission to see the source node title for a file moved log. 

#### Side effects
- None expected.

#### JSON Responses 
- `/v2/nodes/<node-id>/logs/`
- Before:
 - `params['source']['node_title']` ("Just Me") should not be visible to this contributor:
 - screen shot 2016-10-05 at 3 02 51 pm

- After 
 - `params['source']['node_title']` has changed to "Private Component":
 - screen shot 2016-10-05 at 3 02 22 pm

#### Ticket
- [OSF-6521](https://openscience.atlassian.net/browse/OSF-6521)

### DIFF
--- a/api/logs/serializers.py
+++ b/api/logs/serializers.py
@@ -8,6 +8,7 @@ from api.base.serializers import (
     is_anonymized
 )
 from website.project.model import Node
+from website.util import permissions as osf_permissions
 from framework.auth.core import User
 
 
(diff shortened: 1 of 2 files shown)


caseyrollins:
@brianjgeiger, actually this scenario only applies to "source" components. The log is added to the destination node, not the source node, so if the contributor doesn't have access to the destination node, they wouldn't be able to view the logs at all (they would receive a 403 when visiting `/v2/nodes/<dest-node>/logs/`). In this case, the contributor can access the destination node (and so they can view the logs at `/v2/nodes/<dest-node>/logs/` which exposes info about the source node on which they are not a contributor).

PR #6380, opened 2016-10-06
[feature] Fix Wiki Serializer Relationships [OSF-7074]

caseyrollins:
#### Purpose
- Fixes the `node` and `comment` relationship links for registration wikis. 

#### Changes
- Separated current wiki serializers into a base `WikiSerializer`, `NodeWikiSerializer`, and `RegistrationWikiSerializer`.

#### Example JSON Reponses
- BEFORE `/v2/wikis/<wiki_id>/` (of a registration wiki)
screen shot 2016-10-11 at 3 53 09 pm

- AFTER `/v2/wikis/<wiki_id>/` (of a registration wiki)
screen shot 2016-10-11 at 3 51 17 pm

- BEFORE `/v2/registrations/<node_id>/wiki/`
screen shot 2016-10-11 at 3 53 17 pm

- AFTER `/v2/registrations/<node_id>/wiki/`
screen shot 2016-10-11 at 3 52 19 pm

#### Side effects
- None expected.

#### Ticket
- [OSF-7074](https://openscience.atlassian.net/browse/OSF-7074)

### DIFF
--- a/api/wikis/serializers.py
+++ b/api/wikis/serializers.py
@@ -17,6 +17,14 @@ class WikiSerializer(JSONAPISerializer):
         serializer = NodeWikiSerializer(data, context=self.context)
         return NodeWikiSerializer.to_representation(serializer, data)
 
+    def get_absolute_url(self, obj):
+        return absolute_reverse(
+            view_name='wikis:wiki-detail',
+            kwargs={
+                'version': self.context['request'].parser_context['kwargs']['version']
+            }
+        )
+
 
 class NodeWikiSerializer(JSONAPISerializer):
 
--- a/api/wikis/serializers.py
+++ b/api/wikis/serializers.py
@@ -10,24 +10,6 @@ from framework.auth.core import Auth
 
 class WikiSerializer(JSONAPISerializer):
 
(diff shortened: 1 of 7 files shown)


api/wikis/serializers.py
```diff
@@ -7,8 +7,27 @@
 
 from framework.auth.core import Auth
 
+
 class WikiSerializer(JSONAPISerializer):
 
+    def to_representation(self, data, envelope='data'):
+        if data.node.is_registration:
+            serializer = RegistrationWikiSerializer(data, context=self.context)
```
brianjgeiger: Similar to the last "make registrations behave properly" PR you did, is this going to just serve up the information on a /nodes/ link even if it's a registration? I think last time we had to have a base view/serializer and then derive the node and registrations versions from there.

caseyrollins: Good catch. In this case we aren't leaking data between the two endpoints, but I've added [tests](https://github.com/CenterForOpenScience/osf.io/pull/6380/commits/f8e7e935fc78d683bb6fba0eb61caa783bd51295#diff-93502c9a04d7a3055b84e71a66469bfeR140) for confirmation anyways.

PR #6387, opened 2016-10-10
[Feature] Handle smart quotes in registration withdraw form [OSF-7082]

alexschiller:
## Purpose

Project titles with smart quotes currently need to copy/pasted to withdraw a registration as they are not easily typeable.

## Changes
Retooled to match project deletion script. Now requires typing of a scientists name. 

## Side effects

None known.

## Ticket

https://openscience.atlassian.net/browse/OSF-7082

## QA Considerations
1. Test that you can withdraw registrations with the new text.
2. Test that you can delete projects (this change touched a tiny portion of that code).

### DIFF
--- a/website/static/js/registrationRetraction.js
+++ b/website/static/js/registrationRetraction.js
@@ -18,7 +18,7 @@ var Raven = require('raven-js');
 var RegistrationRetractionViewModel = oop.extend(
     ChangeMessageMixin,
     {
-        constructor: function(submitUrl, registrationTitle) {
+        constructor: function(submitUrl) {
             this.super.constructor.call(this);
 
             ko.validation.registerExtenders();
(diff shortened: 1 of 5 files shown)


website/static/js/registrationRetraction.js
```diff
@@ -26,7 +26,9 @@ var RegistrationRetractionViewModel = oop.extend(
             var self = this;
 
             self.submitUrl = submitUrl;
-            self.registrationTitle = $osf.htmlDecode(registrationTitle);
+            // Strips smart quotes and replaces them with typeable quotes
+            self.registrationTitle = $osf.htmlDecode(registrationTitle).replace(/[\u2018\u2019]/g, '\'').replace(/[\u201C\u201D]/g, '"');
```
brianjgeiger: This covers this case, but might not cover the opposite and a few other similar cases. I would lean towards doing a comparison function and adding a bunch of unit tests for that comparison function. It will probably need to handle this case, general unicode, em-dashes vs en-dashes vs hyphens, etc. It might be worth seeing if someone's done some sort of comparison function like this already. If it weren't for unicode, I'd say just strip out all non alphanumeric characters in both cases and compare.

Alternately, the functionality of this is to have them type in the name of the registration in order to withdraw the registration, right? Maybe use the existing functionality of typing in a scientist's name or similar for deleting a node, or use recaptcha, or something.

PR #6388, opened 2016-10-10
Fix incorrect envvar name in gpg error

felliott:
## Purpose / Changes

An error message in `website/security.py` incorrectly referred to the `GNUPG_BINARY` settings key as `GNUPGBINARY` (missing underscore)

## Side effects

None expected.

## Ticket

No ticket


mfraezz:
This is obviated by #6391 , which should be shipped during the next release.

PR #6394, opened 2016-10-11
Make public activity page load faster [OSF-7021]

erinspace:
## Purpose
Right now, the public activity page loads very slowly, because API calls to Keen are being made every time the view is accessed. This PR speeds up the process by moving those keen calls to a script that can be run once daily that populates a storage node with node links, much like New and Noteworthy nodes are stored.

## Changes
- Seperate new and noteworthy from popular projects
    - Create new script, `populate_popular_projects_and_registrations.py` to be run daily
    - Remove popular project aspects from `populate_new_and_noteworthy_projects.py` script
    - Remove popular project calls from `test_populate_new_and_noteworthy_projects.py` 
- Move call to keen from the view into utils, to be accessed from the populate new analytics script. (I chose to move it there because the original view was already accessing one function there!)
- Remove hits from mako and from analytics returns [(for now!)](https://openscience.atlassian.net/browse/OSF-7096)
- Add tests for the new `populate_popular_projects_and_registrations.py` script
- Modify old webtest test to just add node pointers for popular nodes manually

## Side effects
- The script `populate_popular_projects_and_registrations.py` will need to be run once a day
- Popular projects and registrations no longer have a View count on the right hand side -- this should be reimplemented when we have the capacity to add a "project analytics storage" model which we can use to save these values to the database. [Here's a link to the Jira ticket for that future improvement](https://openscience.atlassian.net/browse/OSF-7096)

## Questions
`populate_popular_projects_and_registrations.py` should be run daily to make sure we have the newest popular projects. Anything else I need to do to ensure that happens?

## Notes for QA!
- https://staging.osf.io/explore/activity/ should load a whole lot faster
- View count for popular projects are gone (for now!)
- View count replaced with `date_created` for projects and `registered_date` for registrations
- The popular projects and registrations should only update once a day

## Ticket
https://openscience.atlassian.net/browse/OSF-7021


website/settings/defaults.py
```diff
@@ -281,6 +281,7 @@ def parent_dir(path):
 
 # Most Popular and New and Noteworthy Nodes
 POPULAR_LINKS_NODE = None  # TODO Override in local.py in production.
+POPULAR_LINKS_NODE_REGISTRATIONS = None  # TODO Override in local.py in production.
```
brianjgeiger: Minor, but let's change references to `node_registrations` just to `registrations`, so that this becomes `POPULAR_LINKS_REGISTRATIONS` and in the other 25ish places where similar things are done. Since all registrations are of nodes, I think this is less redundant? Unless I'm missing something.

erinspace: So much less redundant! I hated that name and meant to go back and change it. So thanks for the reminder!

PR #6395, opened 2016-10-11
[Feature] Use IE compatible patch [OSF-6157]

alexschiller:
## Purpose
Settings page uses IE incompatible patch function.

## Changes

Replace IE incompatible patch with $3 patch.

## Side effects
None known.

## Ticket
https://openscience.atlassian.net/browse/OSF-6157
Also fixes: https://openscience.atlassian.net/browse/OSF-6765
[#OSF-6157]

## QA Notes
1. Test to make sure you can update the title, description, and category from the settings page. 
2. Test IE and another browser to confirm no side effects, though none are expected as this fix is used elsewhere.

### DIFF
--- a/website/static/js/projectSettings.js
+++ b/website/static/js/projectSettings.js
@@ -74,7 +74,7 @@ var ProjectSettings = oop.extend(
                     url: self.updateUrl,
                     type: 'PATCH',
                     dataType: 'json',
-                    contentType: 'application/json',
+                    contentType: 'application/vnd.api+json',
                     crossOrigin: true,
                     xhrFields: {withCredentials: true},
                     processData: false,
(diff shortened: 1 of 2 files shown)


website/static/js/projectSettings.js
```diff
@@ -68,11 +69,17 @@ var ProjectSettings = oop.extend(
...
-            var request = $osf.ajaxJSON('patch',
-                self.updateUrl,
-                { data: requestPayload,
-                isCors: true });
+            var requestPayload = JSON.stringify(self.serialize());
+            var request = $3.ajax({
+                    url: self.updateUrl,
+                    type: 'PATCH',
+                    dataType: 'json',
+                    contentType: 'application/json',
```
brianjgeiger: If this is to a v2 api url, it should probably be 'application/vnd.api+json' otherwise this is good.

brianjgeiger:
:tada::penguin::tada:

brianjgeiger:
Well, except the test is still failing. Other than that it's good.

PR #6399, opened 2016-10-12
[RFR] Analytics

cwisecarver:
## Purpose

Enable pagecounters and useractivitycounters in django.

## Changes

Added pagecounter model and useractivitycounter model. Updated functions to use these. Moved functionality from functions to model methods.

## Side effects

Potential row-locking in SQL. We'll see.

Will have a separate PR for migrating the current data.


osf/models/analytics.py
```diff
@@ -1,25 +1,151 @@
...
+                if 'total' not in model_instance.date[date_string].keys():
+                    model_instance.date[date_string].update(total=0)
+                model_instance.date[date_string]['total'] += 1
+            else:
+                model_instance.date[date_string] = dict(total=1)
+
+            # if a download counter is being updated, only perform the update
+            # if the user who is downloading isn't a contributor to the project
+            page_type = cleaned_page.split(':')[0]
+            if page_type == 'download' and node_info:
```
cwisecarver: It appears that we only call this for file downloads so I'm not sure why this is here but I left it because I wasn't 100% sure.

osf/models/analytics.py
```diff
@@ -1,25 +1,151 @@
...
+
+    @classmethod
+    def update_counter(cls, page, node_info):
+        cleaned_page = cls.clean_page(page)
+        date = timezone.now()
+        date_string = date.strftime('%Y/%m/%d')
+        visited_by_date = session.data.get('visited_by_date', {'date': date_string, 'pages': []})
+
+        with transaction.atomic():
+            model_instance, created = cls.objects.select_for_update().get_or_create(_id=cleaned_page)
```
cwisecarver: `select_for_update` inside `transaction.atomic` should cause the row to be locked until the transaction is committed (when the context manager is left) this should make sure our increments and updates are actually increments and updates instead of atomically overwriting data that was written in between reads and writes.

osf/models/analytics.py
```diff
@@ -1,25 +1,151 @@
...
+        except cls.DoesNotExist:
+            return 0
+
+    @classmethod
+    def increment(cls, user_id, action, date_string):
         date = parser.parse(date_string).strftime('%Y/%m/%d')
-        date = date
+        with transaction.atomic():
+            # select_for_update locks the row but only inside a transaction
+            uac, created = cls.objects.select_for_update().get_or_create(_id=user_id)
```
cwisecarver: `select_for_update` inside `transaction.atomic` should force a row lock until the transaction is committed, when the context manager is left. This should make our increments/updates isolated and safe.

tests/factories.py
```diff
@@ -198,7 +199,7 @@ class Meta:
     creator = SubFactory(AuthUserFactory)
 
 
-class ProjectFactory(AbstractNodeFactory):
+class ProjectFactory(NodeFactory):
```
cwisecarver: I changed this because I was getting an error about saving a `AbstractNode` without the type field being set.

tests/framework_tests/test_analytics.py
```diff
@@ -24,20 +27,20 @@ def test_get_total_activity_count(self):
         date = datetime.utcnow()
 
         assert_equal(analytics.get_total_activity_count(user._id), 0)
-        assert_equal(analytics.get_total_activity_count(user._id), user.get_activity_points(db=self.db))
+        assert_equal(analytics.get_total_activity_count(user._id), user.get_activity_points(db=None))
```
cwisecarver: I should probably just remove the db kwarg from all the functions, I was just trying to keep the API as similar as possible to lower the chance of mistakes.

tests/framework_tests/test_analytics.py
```diff
@@ -66,108 +69,102 @@ def setUp(self):
...
         assert_equal(count, (1, 1))
 
-        page = 'download:{0}:{1}'.format(self.node, self.fid)
+        page = 'download:{0}:{1}'.format(self.node._id, self.fid)
 
         session.data['visited'].append(page)
         download_file_(node=self.node, fid=self.fid)
 
-        count = analytics.get_basic_counters('download:{0}:{1}'.format(self.node, self.fid), db=self.db)
+        count = analytics.get_basic_counters('download:{0}:{1}'.format(self.node._id, self.fid), db=None)
```
cwisecarver: Not sure why this was not a problem before but the `__str__` of the Node django model is `<Node: {title} ({_id})>` or something and previous it was just getting the _id.

tests/framework_tests/test_analytics.py
```diff
@@ -66,108 +69,102 @@ def setUp(self):
...
         session.data['visited'].append(page)
         download_file_version_(node=self.node, fid=self.fid, vid=self.vid)
 
-        count = analytics.get_basic_counters('download:{0}:{1}:{2}'.format(self.node, self.fid, self.vid), db=self.db)
+        count = analytics.get_basic_counters('download:{0}:{1}:{2}'.format(self.node._id, self.fid, self.vid), db=None)
         assert_equal(count, (1, 2))
 
     def test_get_basic_counters(self):
         page = 'node:' + str(self.node._id)
+        PageCounter.objects.create(_id=page, total=5, unique=3)
```
cwisecarver: This replaces the next 7 lines.

osf/models/oauth.py
```diff
@@ -129,7 +132,7 @@ class ApiOAuth2PersonalToken(base.ObjectIDMixin, base.BaseModel):
     # /TODO DELETE ME POST MIGRATION
     # Name of the field being `token_id` is a CAS requirement.
     # This is the actual value of the token that's used to authenticate
-    token_id = models.CharField(max_length=70, default=functools.partial(random_string, length=70),
+    token_id = models.CharField(max_length=70, default=generate_random_string,
```
cwisecarver: TIL Can't use functools partial or lambda as a field default.

cwisecarver:
@sloria Fixed the conflicts. We named one function differently.

cwisecarver:
Fixed migrations again.

tests/framework_tests/test_analytics.py
```diff
@@ -5,17 +5,20 @@
...
 from flask import Flask
 
 from datetime import datetime
 
 from framework import analytics, sessions
 from framework.sessions import session
+from osf.models import PageCounter
 
 from tests.base import OsfTestCase
 from tests.factories import UserFactory, ProjectFactory
```
sloria: This should be changed to import from osf_tests.factories. Then [this change](https://github.com/CenterForOpenScience/osf.io/pull/6399/files#diff-bf9417725d38abe6abab2a649c39bed2R202) becomes unnecessary.

tests/framework_tests/test_analytics.py
sloria: Yeah, would be good to remove the unnecessary kwarg. I understand why you kept it though. It's up to you; it can be removed in a follow-up PR.

tests/framework_tests/test_analytics.py
sloria: Yeah, modm's defaults `__str__` returns the string _id, so this change is appropriate. 👍

osf/models/analytics.py
```diff
@@ -1,25 +1,145 @@
...
+                if date_string in model_instance.date.keys():
+                    # increment the number of unique visitors for today
+                    model_instance.date[date_string]['unique'] += 1
+                else:
+                    # set the number of unique visitors to 1
+                    model_instance.date[date_string] = dict(unique=1)
+
+            # update their sessions
+            visited_by_date['pages'].append(cleaned_page)
+            session.data['visited_by_date'] = visited_by_date
```
sloria: Does `session` need to be saved?

framework/analytics/__init__.py
```diff
@@ -29,49 +23,14 @@
...
-    }
-    collection.update(
-        {'_id': user_id},
-        query,
-        upsert=True,
-        manipulate=False,
-    )
-    return True
+    from osf.models import UserActivityCounter
+    # NOTE maybe this should act differently in DEBUG_MODE, I disagree
```
sloria: I also disagree; that DEBUG_MODE check was there just so I could get the app to run locally while this stuff wasn't implemented yet. So this is good as is.

sloria:
Fixed minor things before I merged.

PR #6404, opened 2016-10-13
[RFR] Fix osf.utils.base import error.

cwisecarver:
## Purpose

Stop sentry errors

## Changes

Update migrations to point to things that actually exist.

## Side effects

Less sentry errors.


cwisecarver:
@sloria Could you merge this when you get a chance?

PR #6407, opened 2016-10-14
[RFR] Fix migrations after repo merge

cwisecarver:
## Purpose

Updates migrations to work with new table naming scheme and moves missing changes from osf_models to osf.io.

## Changes

Updates migrations with correct table names, move NodeRelation migration from osf_models repo to osf.io. Add necessary fields to analytics models for migration.

### DIFF
--- a/osf/management/commands/migraterelations.py
+++ b/osf/management/commands/migraterelations.py
@@ -246,9 +246,9 @@ class Command(BaseCommand):
                                     continue
                                 if value.__class__.__name__ in ['Node', 'Registration', 'Collection',
                                                                 'Preprint']:
-                                    gfk_model = apps.get_model('osf_models', 'AbstractNode')
+                                    gfk_model = apps.get_model('osf', 'AbstractNode')
                                 else:
-                                    gfk_model = apps.get_model('osf_models', value.__class__.__name__)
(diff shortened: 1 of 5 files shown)


cwisecarver:
@sloria This should fix node relation migrations and let us give osf_models repo a viking funeral.

osf/management/commands/migraterelations.py
```diff
@@ -238,9 +246,9 @@ def format_lookup_string(modm_obj):
                                     continue
                                 if value.__class__.__name__ in ['Node', 'Registration', 'Collection',
                                                                 'Preprint']:
-                                    gfk_model = apps.get_model('osf', 'AbstractNode')
+                                    gfk_model = apps.get_model('osf_models', 'AbstractNode')
```
sloria: Why was this changed?

sloria:
Looks good to merge; just the one issue noted above.

osf/management/commands/migraterelations.py
cwisecarver: I think I missed it in the rebase. Grabbed the wrong line. I'll change it.

cwisecarver:
Fixed, merging.

PR #6414, opened 2016-10-14
Fix (most) failing views tests for django-osf

erinspace:
**note** A few views tests are still failing because the `osfstorage` addon is acting up -- which @sloria is currently workin on!

## Purpose
Make existing tests from `tests/test_views.py` pass!  Some changes to the tests (timezone assurances, generator/queryset fixes ) and a few updates to the views themselves to make sure they work with the new models and are passing along the right things!

## Changes
- Make sure a few things are in UTC when tests failed because of not being able to compare aware and naive datetimes
- Some things are generators now, update checking methods
- A few methods have new signatures, fix those when called
- Use queries in replace of the old `User. find_by_email`
- Changes in `website/project/views/node.py`:
    - Use a `Guid` object to get nodes, cause Guids are passed and not Node ids
    - Pass a `NodeRelation` to `rm_node_link` and a `Node` to `add_node_link` in new mixins

## Side effects
Most everything should stay the same, hopefully!

## Ticket
https://github.com/CenterForOpenScience/osf-models/issues/147

### DIFF
--- a/website/project/views/node.py
+++ b/website/project/views/node.py
@@ -1304,8 +1304,10 @@ def fork_pointer(auth, node, **kwargs):
     or not present in `nodes`.
 
     """
+    from osf.models import NodeRelation
+
     pointer_id = request.json.get('pointerId')
-    pointer = Node.load(pointer_id)
+    pointer = NodeRelation.load(pointer_id)
 
     if pointer is None:
         # TODO: Change this to 404?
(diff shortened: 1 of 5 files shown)


website/project/views/node.py
```diff
@@ -1185,18 +1192,18 @@ def move_pointers(auth):
 
     for pointer_to_move in pointers_to_move:
         try:
-            pointer_node = from_node.linked_nodes.get(_id=pointer_to_move)
+            node_relation = NodeRelation.objects.get(_id=pointer_to_move)
         except Node.DoesNotExist:
```
sloria: This `except` block needs to be updated to `NodeRelation.DoesNotExist`.

PR #6420, opened 2016-10-17
Allow prereg splash page to load for django-osf

erinspace:
## Purpose

Fix query for a user's nodes on which they are admins!

## Changes

- Update node query for getting a user's nodes, then filter by admin

## Side effects
One query and a pairing down of results vs one modular ODM

## Ticket
https://trello.com/c/tA2FBEWb/16-prereg-splash-page-leads-to-unable-to-resolve

### DIFF
--- a/website/prereg/utils.py
+++ b/website/prereg/utils.py
@@ -9,14 +9,13 @@ PREREG_CAMPAIGNS = {
 def drafts_for_user(user, campaign):
     from website import models  # noqa
 
-    all_user_projects = models.Node.objects.filter(is_deleted=False, _contributors=user)
-    user_projects = [node for node in all_user_projects if node.has_permission(user, 'admin')]
+    user_projects = models.Node.objects.filter(is_deleted=False, contributor__admin=True, contributor__user=user)
     PREREG_CHALLENGE_METASCHEMA = get_prereg_schema(campaign)
-    return models.DraftRegistration.find(
-        Q('registration_schema', 'eq', PREREG_CHALLENGE_METASCHEMA) &
-        Q('approval', 'eq', None) &
-        Q('registered_node', 'eq', None) &
-        Q('branched_from', 'in', [p._id for p in user_projects])
+    return models.DraftRegistration.objects.filter(
+        registration_schema=PREREG_CHALLENGE_METASCHEMA,
+        approval=None,
+        registered_node=None,
+        branched_from__in=[p.id for p in user_projects]
     )
 
 def get_prereg_schema(campaign='prereg'):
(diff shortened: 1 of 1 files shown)


erinspace:
@cwisecarver ready for a look over! Question tho -- should we be re-doing modular ODM queries when it's simple enough to do? Thought it was weird here to have a Django ORM query and then a Modular ODM one on the next line, but left it in the spirit of making as few changes as necessary. Thoughts?

website/prereg/utils.py
```diff
@@ -11,10 +9,8 @@
 def drafts_for_user(user, campaign):
     from website import models  # noqa
 
-    user_projects = models.Node.find(
-        Q('is_deleted', 'eq', False) &
-        Q('permissions.{0}'.format(user._id), 'in', [permissions.ADMIN])
-    )
+    all_user_projects = models.Node.objects.filter(is_deleted=False, _contributors=user)
+    user_projects = [node for node in all_user_projects if node.has_permission(user, 'admin')]
```
cwisecarver: You can do this in one query without the python loop. Django magics.

```
user_projects = Node.objects.filter(is_deleted=False, contributor__admin=True, contributor__user=user)
```

cwisecarver:
@erinspace I'd say if you see one that's right next to what you're working on, doesn't hurt to change it. But I wouldn't kill yourself. We're going to have to go back through and rewrite/optimize them eventually so every little bit helps. Plus, it gets people in the habit of doing them and gives everyone some examples to learn from.

erinspace:
@cwisecarver ready for another look!

website/prereg/utils.py
```diff
@@ -9,14 +9,13 @@
...
-    return models.DraftRegistration.find(
-        Q('registration_schema', 'eq', PREREG_CHALLENGE_METASCHEMA) &
-        Q('approval', 'eq', None) &
-        Q('registered_node', 'eq', None) &
-        Q('branched_from', 'in', [p._id for p in user_projects])
+    return models.DraftRegistration.objects.filter(
+        registration_schema=PREREG_CHALLENGE_METASCHEMA,
+        approval=None,
+        registered_node=None,
+        branched_from__in=[p.id for p in user_projects]
```
erinspace: That is totally awesome.

cwisecarver:
lgtm

cwisecarver:
Thanks @erinspace

erinspace:
Also checking with QA to see what can be Draft Registrations! Will update `Node` vs `AbstractNode` if necessary!

erinspace:
Checked with Amanda, can't make a registration from a registration so Node seems to work here!

cwisecarver:
Sweet! Thanks @erinspace

PR #6421, opened 2016-10-17
[Feature] Add error catch for no file [OSF-6911]

alexschiller:
## Purpose

Dropzone on Firefox and IE on Windows silently fail when trying to upload folders. This, along with a corresponding dropzone patch emit an error and handle it. 
## Changes

Dropzone: emit error when invalid upload
Fangorn: catch error and growl
## Side effects

In a Chrome dropzone, if you drag a folder and a file, the file will still upload. On windows ie/firefox both will fail, but will throw the ~'cannot upload folders' error message
## Ticket

https://openscience.atlassian.net/browse/OSF-6911
[#OSF-6911]
## QA Considerations

Error should emit properly on attempting to drag a folder into dropzone on windows. All files will get rejected, since windows does not detect -any- of them as being dropped. Dropping non-folders should work properly.


alexschiller:
Fixing this conflict right now

PR #6423, opened 2016-10-17
Adjust calls to iterate through a user's external accounts for citations Django-osf

erinspace:
## Purpose
Sentry errors happening on iterating through a user's external accounts!

## Changes
- Add `.all()` to two calls to `user.external_accounts` in `citations/providers.py`

## Side effects
None anticipated

## Note
The specific issue mentioned in the trello ticket wasn't solved for me locally until I'd re-run the script `python -m scripts.parse_citation_styles`

## Ticket
https://trello.com/c/q0aNJC9W/20-citations-widget-cannot-search-for-citations
https://sentry.cos.io/sentry/osf-back-end/issues/256168/

### DIFF
--- a/website/citations/providers.py
+++ b/website/citations/providers.py
@@ -105,8 +105,7 @@ class CitationsProvider(object):
                 self.serializer(
                     user_settings=user.get_addon(self.provider_name) if user else None
                 ).serialize_account(each)
-                for each in user.external_accounts.all()
-                if each.provider == self.provider_name
+                for each in user.external_accounts.filter(provider=self.provider_name)
             ]
         }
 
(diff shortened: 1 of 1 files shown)


erinspace:
@cwisecarver another small one!

website/citations/providers.py
```diff
@@ -105,7 +105,7 @@ def user_accounts(self, user):
                 self.serializer(
                     user_settings=user.get_addon(self.provider_name) if user else None
                 ).serialize_account(each)
-                for each in user.external_accounts
+                for each in user.external_accounts.all()
```
cwisecarver: @erinspace I probably led you astray with the `.all()`. You could probably use a `.filter(provider=provider_name)` to avoid the python loops. It'll work for now. If you want to fix it and test you're a rockstar. If you want to add a TODO and leave it for the optimization pass that's okay too. Sorry for the inadequate advice.

erinspace:
@cwisecarver filter updated! However, having some trouble figuring out where to but a test -- would it be in [CitationAddonProviderTestSuiteMixin](https://github.com/CenterForOpenScience/osf.io/blob/feature/django-osf/addons/base/tests/models.py#L572)?

cwisecarver:
Nice! Thanks! Looks like a good place to me.

mfraezz:
This has developed conflicts from #6425

erinspace:
Looks like this was fixed in #6425, moving on I suppose

PR #6425, opened 2016-10-18
[WIP][Django-OSF] Addons

mfraezz:
## Purpose
Implement addons for django-osf

## Changes
* Base model improvments
* Semi-complete functionality for
  * Box
  * Dropbox

### TODO:
Lots of things

## Side effects
No more than anything else DjOSF-related

## Ticket
None known.


addons/base/models.py
```diff
@@ -101,10 +105,10 @@ def nodes_authorized(self):
         """Get authorized, non-deleted nodes. Returns an empty list if the
         attached add-on does not include a node model.
         """
-        model = self._meta.app_config.node_settings
+        model = self.config.node_settings
         if not model:
             return []
-        return model.objects.filter(user_settings=self, owner__is_deleted=False)
+        return [obj.owner for obj in model.objects.filter(user_settings=self, owner__is_deleted=False)]
```
cwisecarver: `model.objects.filter(user_settings=self, owner__is_deleted=False).values_list('owner', flat=True)`

website/oauth/models/__init__.py
```diff
@@ -489,7 +491,7 @@ class ApiOAuth2Application(StoredObject):
     name = fields.StringField(index=True, required=True, validate=[string_required, MaxLengthValidator(200)])
     description = fields.StringField(required=False, validate=MaxLengthValidator(1000))
 
-    date_created = fields.DateTimeField(auto_now_add=datetime.datetime.utcnow,
+    date_created = fields.DateTimeField(auto_now_add=timezone.now,
```
cwisecarver: In django,`auto_now_add` mean's it's immutable. Just FYI.

mfraezz: Means the same thing for MODM, though Toku doesn't do anything to enforce that. Uncertain if psql will.

cwisecarver: psql doesn't enforce it but the django ORM will. You'll get an exception if you try and set it to anything, ever.

mfraezz: `date_created` shouldn't ever be set other than on creation, unless there's a migration to a new kind of model and we want to keep the old data (e.g. the `PreprintService` migration, which set it based on the `node.preprint_created` field). This was accomplished with pymongo's `database.find_and_modify(<selector>, {'date_created': node.preprint_created})`.

Is there a similar "this shouldn't ever happen but just in case"-type workaround for manipulating psql data?

cwisecarver: You can get around django's model validation a few ways:

- Use `bulk_create` or `bulk_update`, these don't run any of the signals or validation on the models.
- Use straight SQL and go around the ORM. This bypasses the ORM completely and won't run any validation other than making sure things pass the constraints that are actually in the database (unique, not null, type (varchar, int, jsonb), etc)

There might be others that I don't know about or aren't thinking of right now. Those are the main ones though. So there's always another way. Even if the constraint was in SQL we could drop the constraint at the beginning of the migration, do our dirt, and then put it back on at the end.

website/citations/providers.py
```diff
@@ -205,7 +205,7 @@ def citation_list(self, node_addon, user, list_id, show='all'):
 
         node_account = node_addon.external_account
         user_accounts = [
-            account for account in user.external_accounts
+            account for account in user.external_accounts.all()
```
sloria: Same here: `user.external_accounts.filter(provider=self.provider_name)`?

website/oauth/models/__init__.py
sloria: FWIW, `default=timezone.now` has the same effect as `auto_add_now` without making the field immutable. That said, it seems that `auto_add_now` is the right thing to do in this case.

sloria:
Discussed on Flowdock with @mfraezz : it would be great to get this merged before we get started on all the addons because this fixes a lot of the base addon code, which will also affect my work on the wiki addon.

Just a few nitpicks from Chris and I above. No blockers, but I'll wait to merge this until @mfraezz gives me the go-ahead.

addons/base/models.py
mfraezz: As discussed on Flowdock, this quickly becomes more trouble than it's worth.

sloria: `values_list` only returns PKs for foreign objects, not the objects themselves, so this might need to stay as a listcomp.

sloria: I guess you could at least add `select_related('owner')` to reduce the number of queries.

cwisecarver: Not 100% sure, but I think since he's filtering on owner that it'll be forced to do the join on the first query anyway. Can't hurt to add `select_related('owner')` though.

website/oauth/models/__init__.py
cwisecarver: It will, however, make any migration be inaccurate/fail.

cwisecarver: Also, I didn't notice this the first time. [`auto_now_add` should be a bool.](https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.DateField.auto_now_add)

addons/base/models.py
```diff
@@ -108,7 +108,7 @@ def nodes_authorized(self):
         model = self.config.node_settings
         if not model:
             return []
-        return [obj.owner for obj in model.objects.filter(user_settings=self, owner__is_deleted=False)]
+        return [obj for obj in model.objects.filter(user_settings=self, owner__is_deleted=False).select_related('owner')]
```
sloria: You will still need to do `obj.owner` if you want to return the nodes rather than the node settings objects.

website/addons/github/model.py
```diff
@@ -76,10 +76,7 @@ def revoke_remote_oauth_access(self, external_account):
     # Assumes oldest connected account is primary.
     @property
     def public_id(self):
-        gh_accounts = [
-            a for a in self.owner.external_accounts.all()
-            if a.provider == self.oauth_provider.short_name
-        ]
+        gh_accounts = self.owner.external_accounts.filter(provider=self.oauth_provider.short_name)
         if gh_accounts:
             return gh_accounts[0].display_name
```
sloria: Have you verified that this works? I forget the rules that Django has about indexing querysets...in any case, it's safest to do `.first()` instead.

mfraezz: GH doesn't work yet, so I have not verified, no.

cwisecarver: You probably want `.exists()` which will return a bool. `.first()` will cause a sort, which depending on the table and its indexes could be really really slow.

addons/base/models.py
mfraezz: Hmm. You're right, but this somehow doesn't break the links to nodes on `/settings/addons`, which is what I thought this populated.

website/citations/providers.py
```diff
@@ -204,10 +204,7 @@ def citation_list(self, node_addon, user, list_id, show='all'):
                 folder['parent_list_id'] = 'ROOT'
 
         node_account = node_addon.external_account
-        user_accounts = [
-            account for account in user.external_accounts.all()
-            if account.provider == self.provider_name
-        ] if user else []
+        user_accounts = user.external_accounts.filter(provider=self.provider_name) if user else []
```
sloria: I think you can replace this and the next line with

```python
user_is_owner = user.external_accounts.filter(id=node_account.id).exists()
```

addons/base/models.py
cwisecarver: [`select_related` just makes fetching the related objects not lazy.](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#select-related) Normally, django would only get the properties directly in the table of the model you're querying. When you try to access a FK or M2M it would do another sql query to get the relationship. `select_related` (fk) and `prefetch_related` (m2m) do the query or queries eagerly.

website/addons/github/model.py
sloria: @cwisecarver But this method returns the display_name of the first result in the queryset a few lines down.


The result set should be very small (# of Github accounts attached to a user), so I don't think the sort will be an issue.

sloria: Makes sense. In this case, the result sets should not exceed tens of rows.

That said, I just tried doing `u.external_accounts.filter(provider='github')[0]` and that does work, so I think this is ok as is.

cwisecarver: Yeah, the django queryset's are indexable and sliceable. They will be evaluated when sliced though so you'll have a python object, not a queryset anymore.

You can't do:
```python
Node.objects.all()[0:20].order_by('title')
```
But you can do:
```python
Node.objects.all().order_by('title')[0:20]
```

cwisecarver:
@mfraezz You think this is gtg? Want to get it merged sooner rather than later so we can build upon it instead of merge-conflicting about it.

sloria:
Got the go from Matt. This looks good.

PR #6433, opened 2016-10-19
[Feature] Addon Files

mfraezz:
## Purpose
Fix files for addons

## Changes
* Set up Django AppConfigs before importing DjangOSF models elsewhere, which led to `ImproperlyConfigured` errors
* Add addon-specific `FileNode` subclasses
* Misc bug fixes / updates to Django-style queries

## Side effects
None expected

## Ticket
None known

screen shot 2016-10-18 at 21 33 34

### DIFF
--- a/website/addons/base/views.py
+++ b/website/addons/base/views.py
@@ -15,6 +15,7 @@ import jwt
 from modularodm import Q
 from modularodm.exceptions import NoResultsFound
 
+from addons.base.models import BaseStorageAddon
 from framework import sentry
 from framework.auth import Auth
 from framework.auth import cas
@@ -27,10 +28,9 @@ from framework.transactions.context import TokuTransaction
 from framework.transactions.handlers import no_auto_transaction
 from website import mails
 from website import settings
-from website.addons.base import StorageAddonBase
 from website.addons.base import exceptions
 from website.addons.base import signals as file_signals
-from website.files.models import FileNode, StoredFileNode, TrashedFileNode
+from osf.models.files import FileNode, StoredFileNode, TrashedFileNode
 from website.models import Node, NodeLog, User
 from website.profile.utils import get_gravatar
 from website.project import decorators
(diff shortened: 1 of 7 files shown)


website/addons/base/views.py
```diff
@@ -27,10 +28,9 @@
 from framework.transactions.handlers import no_auto_transaction
 from website import mails
 from website import settings
-from website.addons.base import StorageAddonBase
 from website.addons.base import exceptions
 from website.addons.base import signals as file_signals
-from website.files.models import FileNode, StoredFileNode, TrashedFileNode
+from osf.models.files import FileNode, StoredFileNode, TrashedFileNode
```
cwisecarver: Would you import `FileNode` (and anything else defined in `osf.models.files` and not explicitly imported into `osf.models`) into `osf.models` and then import all of these from there? I'm trying to keep the pattern where everything comes from `osf.models` but nothing is actually in `osf.models`.

website/addons/base/views.py
```diff
@@ -577,7 +577,7 @@ def addon_deleted_file(auth, node, error_type='BLAME_PROVIDER', **kwargs):
         'provider': file_node.provider,
         'materialized_path': file_node.materialized_path or file_path,
         'private': getattr(node.get_addon(file_node.provider), 'is_private', False),
-        'file_tags': [tag._id for tag in file_node.tags],
+        'file_tags': file_node.tags.all().values_list('id', flat=True),
```
cwisecarver: I can't remember if files could have system tags prior, but system tags and tags are all in the tags bucket now, so they can now. To get only non system tags `file_node.tags.filter(system=False).values_list('id', flat=True)`

website/addons/base/views.py
```diff
@@ -715,7 +715,7 @@ def addon_view_file(auth, node, file_node, version):
         'extra': version.metadata.get('extra', {}),
         'size': version.size if version.size is not None else 9966699,
         'private': getattr(node.get_addon(file_node.provider), 'is_private', False),
-        'file_tags': [tag._id for tag in file_node.tags],
+        'file_tags': file_node.tags.all().values_list('id', flat=True),
```
cwisecarver: Same thing about system tags here. And also, `id` is going to be an integer primary key, not the guid string. Just wanted to make sure that the caller was accounting for that difference.

cwisecarver:
@sloria anything to add?

website/addons/base/views.py
sloria: Shuoldn't this be `file_node.tags.filter(system=False).values_list('name', flat=True)`?

sloria:
Other than the fix for tags, this looks good to go.

cwisecarver:
Closing in favor of #6438 so I can take credit for @mfraezz hard work.

PR #6436, opened 2016-10-19
[Feature] Allow read/write to modify all nested affiliations [OSF-7105]

alexschiller:
## Purpose

Currently only admins are given the dialogue option to add/remove affiliations to all nested components. This was a side effect of the data provided by the API V1 route used. Read/Write contributors are now given the add/remove all dialogue box.
## Changes
1. Strip out api v1 functionality
2. Add 3 osf helper functions to get all paginated data, concat the data, and return all children of a node from a paginated response.
3. Using API v2 calls, allows read/write users to modify nested nodes.
## Side effects

The logic of when to show the modal (add/remove from node or node and all children) was simplified to avoid some fun/edgy cases (e.g., extremely edgy case if a node has more than 10 institutions) and generally reduce complexity. Now, so long as the node you are on has at least one child, the modal will appear. 

Previously, it only appeared if at least one child would be affected by the change (or a few weird edge cases like if you don't have permission over one of the children). 

This also makes the action of pressing the add/remove button more consistent. For example, the user likely does not know the affiliation status of all of their nodes children, and might wonder why clicking the button to add the institution to the parent did not show a modal, but then clicking it again does show the modal. And then clicking it again shows the modal again, despite seeming like the same action as the first click. 
## Ticket

https://openscience.atlassian.net/browse/OSF-7105
[#OSF-7105]
## QA Considerations
1. Test that admins and read/write users get the dialogue option at the appropriate times
2. Test both root (highest level project settings page) and nested components with further nested components work properly. 
3. Test situations where nested status is mixed (some affiliated, some not) and that the nodes affiliation is what you would expect.
4. Test that removing an institution as an admin when you are no longer affiliated with the institution correctly has the warning `If you remove it from your project, you cannot add it back.`

### DIFF
--- a/website/static/js/osfHelpers.js
+++ b/website/static/js/osfHelpers.js
@@ -112,13 +112,13 @@ var ajaxJSON = function(method, url, options) {
 /**
  * Returns a promise of an array of ajaxJSON response objects
  */
-function getAllPagesAjaxJSON(method, url, options) {
+var getAllPagesAjaxJSON = function(method, url, options) {
     var responses = [];
     var fetch = function(method, url, options) {
         var request = ajaxJSON(method, url, options);
         request.done(function(response) {
             deferred.notify(response);
-            responses.push(response);
+            responses = responses.concat(response);
             if (response.links.next !== null) {
                 fetch(method, response.links.next, options);
             } else {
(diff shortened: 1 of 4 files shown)


website/static/js/osfHelpers.js
```diff
@@ -110,6 +110,83 @@ var ajaxJSON = function(method, url, options) {
...
+    return deferred.promise();
+};
+
+/**
+ * Takes an array of response objects and returns a single object
+ * @param {Array of Objects}
+ * @return {Object data}
+ */
+var mergePagesAjaxJSON = function(pages) {
+    var nodeList = {};
```
brianjgeiger: These don't have to be nodes.

PR #6441, opened 2016-10-20
[Hotfix] Encrypt unicode correctly

mfraezz:
## Purpose
Fix issue with encrypting unicode

## Changes
* `jwe` [wants bytes](https://github.com/chrisseto/pyjwe/blob/master/README.md), so give it `bytes`

## Migration
`python -m scripts.encrypt_external_accounts_unicode --dry`

## Side effects
None expected.

## Ticket
None

## Before
```
In [4]: name = 'mfräezz'

In [5]: jwe.encrypt(name.encode('utf-8'), SENSITIVE_DATA_KEY)
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-5-ca1156250faa> in <module>()
----> 1 jwe.encrypt(name.encode('utf-8'), SENSITIVE_DATA_KEY)

UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 3: ordinal not in range(128)
```
## After
```
In [6]: jwe.encrypt(bytes(name), SENSITIVE_DATA_KEY)
Out[6]: 'eyJhbGciOiAiZGlyIiwgImVuYyI6ICJBMjU2R0NNIn0.._6_kjjv7JIsqlcy_ZfRkqg.hSZpb3ymal0.z-3k--QrsHW1ry_pmTlMHg'

In [7]: jwe.decrypt(bytes(_), SENSITIVE_DATA_KEY) == name
Out[7]: True
```

### DIFF
--- a/scripts/encrypt_external_accounts_unicode.py
+++ b/scripts/encrypt_external_accounts_unicode.py
@@ -1,0 +1,64 @@
+import logging
+import sys
+
+import jwe
+
+from framework.encryption import SENSITIVE_DATA_KEY
+from framework.mongo import database
+from framework.transactions.context import TokuTransaction
+
+from website.app import init_app
+from scripts import utils as script_utils
+
+logger = logging.getLogger(__name__)
+
+
+def verify_external_account(document):
+    assert '_id' in document, 'ExternalAccount {} has no _id'.format(document)
+    assert 'provider' in document, 'ExternalAccount {} has no provider'.format(document['_id'])
+    assert 'oauth_key' in document, 'ExternalAccount {} has no oauth_key'.format(document['_id'])
+    assert 'oauth_secret' in document, 'ExternalAccount {} has no oauth_secret'.format(document['_id'])
+    assert 'display_name' in document, 'ExternalAccount {} has no display_name'.format(document['_id'])
+
+def encrypt_key(document, key):
+    database['externalaccount'].find_and_modify(
+        {'_id': document['_id']},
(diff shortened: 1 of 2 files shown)


scripts/encrypt_external_accounts_unicode.py
```diff
@@ -0,0 +1,64 @@
...
+    assert 'provider' in document, 'ExternalAccount {} has no provider'.format(document['_id'])
+    assert 'oauth_key' in document, 'ExternalAccount {} has no oauth_key'.format(document['_id'])
+    assert 'oauth_secret' in document, 'ExternalAccount {} has no oauth_secret'.format(document['_id'])
+    assert 'display_name' in document, 'ExternalAccount {} has no display_name'.format(document['_id'])
+
+def encrypt_key(document, key):
+    database['externalaccount'].find_and_modify(
+        {'_id': document['_id']},
+        {'$set': {
+            key: jwe.encrypt(bytes(jwe.decrypt(document[key].encode('utf-8'), SENSITIVE_DATA_KEY)), SENSITIVE_DATA_KEY)
```
mfraezz: This is the only non-import line that's different from the [original script](https://github.com/CenterForOpenScience/osf.io/blob/develop/scripts/encrypt_external_accounts.py). 

First it decrypts the stored value using the old decryption method, and then encrypts it with the new method before writing to the database.

sloria:
Change is correct and verified migration on a recent prod dump. 👍

PR #6443, opened 2016-10-20
[feature] Safari back button fangorn fix [OSF-5743]

alexschiller:
## Purpose

Events persist with forward and back buttons on safari. This causes fangorn to hang and not allow scrolling when the back button is used.
## Changes

Adds a check if the events persisted on fangorn pages and forces a refresh if they did. 
## Side effects

None known.
## Ticket

https://openscience.atlassian.net/browse/OSF-5743

[#OSF-5743]
## QA Considerations
1. Test all browsers to ensure the page refresh only occurs on safari and not others. 
2. Test both forward and back actions.


alexschiller:
💣 Travis passed on my fork💣

alexschiller:
Closing because it does not occur in the current version of safari and the fix is less than ideal.

PR #6447, opened 2016-10-20
[Feature] Fix VOL nesting [OSF-5507]

alexschiller:
## Purpose

VOL view on the contributors page nests very poorly. This flattens the nesting.
## Changes

Nodes no longer nest in the list of associated nodes. Set to a static 100px width, trims overflow and adds ellipsis. 
Before:
before
After:
after
## Side effects

None known.
## Ticket

https://openscience.atlassian.net/browse/OSF-5507

[#OSF-5507]
## QA Considerations
1. Should show a flat list
2. Long titles should be truncated.
3. The modal that used to create a VOL should still indent properly (related code)

### DIFF
--- a/website/static/css/pages/contributor-page.css
+++ b/website/static/css/pages/contributor-page.css
@@ -18,6 +18,13 @@
 .more-link-node{
     margin-top: 5px;
 }
+.private-link-list-node {
+    width: 100px;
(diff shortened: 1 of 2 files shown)


alexschiller:
Added before/after.

PR #6453, opened 2016-10-21
Keen Analytics Daily Script [OSF-6954]

hmoco:
## Purpose
Add analytics scripts that allow for sending daily statistics to keen.

There are two types of scripts: 
   - snapshot scripts that capture the current state of things, and can't be backfilled. These are analytics that are based on a mutable timestamp. Right now, there's only one script in this category at the moment, `addon_snapshot`, because the number of nodes and users connected to addons fluctuates (and doesn't have a date attached to it).
   - summary scripts that can be run with a date, and so can be backfilled. This includes user counts, node counts, as well as gathering node_logs for a particular day.

## Changes
- 4 analytics scripts added:
    -  `python -m scripts.analytics.addon_snapshot`. Counts how many nodes and users have connected to each addon (does not support date input)
   -  `python -m scripts.analytics.user_summary -d <date argument optional>`. Counts how many users were active, deactivated and unconfirmed in any given date. Default date is UTC today.
   -  `python -m scripts.analytics.node_summary -d <date argument optional>`. Counts how many overall nodes, projects and registrations existed up to that date. Default date is UTC today. If it's run on the same day (and not being used to backfil data), it will also include info about private and public projects, as well as embargoed and retracted registrations.
   -  `python -m scripts.analytics.node_logs_events -d <date argument optional>`. Makes keen events for logs within that date range. Default is starting yesterday and ending today. 
- 2 scripts to run the different types of analytics scripts:
   - ` python -m scripts.analytics.run_keen_summaries -s <start_date argument optional> -e <end_date argument optional> -as <scripts to run argument optional>`. Default is starting yesterday and ending today, in UTC. Runs once for each day in the range provided for the summary scripts. By default runs the user, node count, and node_log scripts, but will accept scripts via command line as well. Scripts to run should be the filename of the script you'd like to run, without .py, located inside `scripts/analytics`
   - ` python -m scripts.analytics.run_keen_snapshots -as <scripts to run argument optional>`. Get a snapshot of the state of addons right now by default. Can also accept other snapsht scripts to run. Scripts to run should be the filename of the script you'd like to run, without .py, located inside `scripts/analytics`
- Added a daily run of `run_keen_snapshots` and `run_keen_summaries` scripts for the celery schedule.

### User Count Summary
```
{
  "keen": {
    "timestamp": "2016-10-26T12:28:01.645899"
  },
  "deactivated_users": 2,
  "active_users": 4,
  "unconfirmed_users": 2
}
```
### Node Logs
```
[  
   {  
      "keen":{  
         "timestamp": "2016-10-25T17:31:54.089654"
      },
      "date": "2016-10-25T17:31:54.063000",
      "user_id": "hsytt",
      "action": "node_created"
   },
   {  
      "keen":{  
         "timestamp": "2016-10-25T17:31:54.089654"
      },
      "date": "2016-10-25T17:31:54.076000",
      "user_id": "hsytt",
      "action": "file_added"
   }
]
```
### Node Count Summary
```
{  
   "nodes": {  
      "total": 8,
      "public": 1,
      "private": 2
   },
   "registered_projects": {  
      "embargoed": 1,
      "total": 4,
      "withdrawn": 1,
      "public": 3
   },
   "registered_nodes": {  
      "embargoed": 1,
      "total": 5,
      "withdrawn": 1,
      "public": 4
   },
   "projects": {  
      "total": 2,
      "public": 1,
      "private": 1
   }
}
```
### Addon Counts - many events
```
[  
   {  
      'provider': {
          'name': 'figshare'
      },
      'users': {
          'total': 300
      },
      'nodes': {
          'total': 6,
          'connected': 4,
          'deleted': 1,
          'disconnected': 1
      }
   },
   {  
      'provider': {
          'name': 'github'
      },
      'users': {
          'total': 300
      },
      'nodes': {
          'total': 6,
          'connected': 4,
          'deleted': 1,
          'disconnected': 1
      }
  }
]
```
## Ticket:
https://openscience.atlassian.net/browse/OSF-6954


scripts/analytics/addon_count.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def count():
+    counts = []
+    addons_available = {k: v for k, v in [(addon.short_name, addon) for addon in ADDONS_AVAILABLE]}
+    for short_name, addon in addons_available.iteritems():
+        user_count = addon.settings_models['user'].find().count() if addon.settings_models.get('user') else 0
+        node_count = addon.settings_models['node'].find().count() if addon.settings_models.get('node') else 0
+        counts.append({
+            'provider': short_name,
```
brianjgeiger: Note from @felliott that we could use some more nesting here, with possibly some extra info as well. For example:
```
{
  'provider': {
    ‘id’: 'googledrive’,
  },
  'users': {
    'total': 123,
  },
  'nodes': {
    'total': 240,
    'connected': 123,
    'deleted': 5,
    'unconnected': 112,
  },
  'extra': {
    'dataverse': {
       'hosts': {
          'dataverse.harvard.edu': 20,
          'dataverse.virginia.edu': 10,
          'demo.dataverse.org': 50
        },
    },
    'github': {
       'users': {
         'uniq': 10,
       },
       'repos': {
         'uniq': 15,
       },
       'user_repos': {
         'uniq': 20,
       },
    },
  },
},
```

scripts/analytics/run_keen_counts.py
```diff
@@ -0,0 +1,64 @@
...
+        client.add_events(get_events_for_day(day))
+        day = day + timedelta(1)
+    for chunk in get_node_log_events(start_date, end_date):
+        logger.info('NodeLogs counted. {} NodeLogs.'.format(len(chunk)))
+        client.add_events({'node_log_analytics': chunk})
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Populate keen counts!')
+    parser.add_argument('-e', '--end', dest='end_date', required=False)
+    parser.add_argument('-s', '--start', dest='start_date', required=False)
```
brianjgeiger: I would add in the ability to pass in a script or list of scripts in case we need to backfill something in the future that doesn't currently exist.

brianjgeiger: Also, consider (but don't if it's too complicated) having a replacement mode where you can delete a keen event that matches one you'd like to add (for example, with the same script and date). If you can't do that without too much trouble, we should probably file a ticket to add it, but we will almost certainly need a method of removing the right event, which may involve some sort of composite field that includes the date and the script name, for example.

felliott:
Minor: I *think* this is consistent about timestamp timezones, but it would be good to be explicit about that in the code and in the help messages.  Make sure to mention that a 'day' is always 00:00UTC - 23:59:59UTC.

scripts/analytics/node_count.py
```diff
@@ -0,0 +1,113 @@
...
+        })
+
+        totals['registered_projects'].update({
+            'public': Node.find(registered_project_public_query).count(),
+            'embargoed': Node.find(registered_project_private_query).count(),
+            'withdrawn': Node.find(registered_project_retracted_query).count(),
+        })
+
+    if end_date:
+        totals['keen'] = {'timestamp': end_date.isoformat()}
```
brianjgeiger: Since we have the option of backdating many of these scripts (or parts of all the scripts), does it make sense to add the timestamp to each of them, or just the ones you've done already?

felliott:
For the `users` collection:

scripts/analytics/run_keen_counts.py
```diff
@@ -0,0 +1,64 @@
...
+def main(start_date, end_date):
+    keen_project = keen_settings['private']['project_id']
+    write_key = keen_settings['private']['write_key']
+    if keen_project and write_key:
+        client = KeenClient(
+            project_id=keen_project,
+            write_key=write_key,
+        )
+    assert(client)
+    client.add_events({'addon_count_analytics': addon_count()})
```
hmoco: This should not be run if arguments were passed to the script (i.e. if an end_date and/or start_date is specified).

felliott:
It occurs to me that we're dealing with a couple different types of data here.  `addons` is a snapshot collection; we can collect new data each day, but we can't meaningfully backfill historical data. This is irreplaceable (mostly).  `users` is a status summary collection, it summarizes data that could be regenerated at any time.  We put it into Keen to save ourselves expensive DB queries, but it's always available.

I think it might be useful to distinguish these, both in the naming of the collection, and how they're invoked.  The snapshots should be run every day around the same time by one script and get the most attention.  Snapshot analytics should not take a timestamp, because they only apply to right now.
Summaries should be run separately, and at a lower priority.  Summary scripts would take a date argument and generate the counts/tallies for that day.

I think `addons` could be renamed to `addons_snapshot` and `users` to `users_summary`. It would be good to have a separate `users_snapshot` collection that would track things like affiliated institution counts, system_tag counts, social fields filled in, locale counts, etc.  Whether that's in scope for this ticket is up to @brianjgeiger.

I'm not sure about nodes or node_logs yet.  I think the current implementation of nodes has both snapshot and summary aspects.

I'm open to alternative suffixes for `_snapshot` and `_summary` (esp. `_summary`) so please throw them out if you got them!

scripts/analytics/addon_snapshot.py
```diff
@@ -0,0 +1,57 @@
+import logging
+from modularodm import Q
+
+from website.settings import ADDONS_AVAILABLE
```
felliott: When I try to run this script standalone, it fails on this line.  `ADDONS_AVAILABLE` isn't computed until after `init_app()` is run.

scripts/analytics/addon_snapshot.py
```diff
@@ -0,0 +1,57 @@
...
+def main():
+    addon_count = get_events()
+    keen_project = keen_settings['private']['project_id']
+    write_key = keen_settings['private']['write_key']
+    if keen_project and write_key:
+        client = KeenClient(
+            project_id=keen_project,
+            write_key=write_key,
+        )
+        client.add_event('addon_count_analytics', addon_count)
```
felliott: I think this should be `client.add_events({'addon_count_analytics': addon_count})`.  Consider renaming `get_events()` to something more appropriate like `get_addon_usage_snapshot()`.  Please also rename the event collection to `addon_snapshot`.

scripts/analytics/addon_snapshot.py
```diff
@@ -0,0 +1,57 @@
...
+        counts.append({
+            'provider': {
+                'name': short_name
+            },
+            'users': {
+                'total': user_count
+            },
+            'nodes': {
+                'total': node_count,
+                'connected': node_count,
```
felliott: Why are `total` and `connected` the same?  I would think `total` would be `connected + deleted + disconnected`.

scripts/analytics/addon_snapshot.py
felliott: Derp. I get why its called `get_events`. Ignore that bit. The other two stand.

scripts/analytics/run_keen_summaries.py
```diff
@@ -0,0 +1,98 @@
...
+            try:
+                script_events = importlib.import_module('scripts.analytics.{}'.format(script))
+                keen_events.update({script: script_events.get_events(day)})
+            except ImportError as e:
+                logger.error(e)
+                logger.error('Error importing script - make sure the script specified is inside of scripts/analytics')
+    else:
+        keen_events.update({'user_summary': user_events(day)})
+        keen_events.update({'node_summary': node_events(day)})
+        keen_events.update({'node_log_events': node_log_events(day)})
```
felliott: Make the collection name an attribute of the script, so it's not hardcoded in two places.

scripts/analytics/user_summary.py
```diff
@@ -0,0 +1,88 @@
...
+from website.settings import KEEN as keen_settings
+from keen.client import KeenClient
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+
+def get_events(date):
+    """Count how many nodes exist.
+    If no date is given, include all users up until the point when the script was called.
```
felliott: Looking at the current state of the code, `date` will never be not given.  Maybe delete this comment and add an assertion?

scripts/analytics/node_log_events.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def get_events(date):
+    """ Get all node logs from a given date for a 24 hour period.
+    Defaults to starting yesterday and gathering through today (both in UTC).
+    """
+
+    # Make sure default is yesterday and not today
+    if date.day == datetime.utcnow().day:
+        date = date - timedelta(1)
```
felliott: Since this may end up getting called from different places, I wouldn't provide a default.  Instead raise an error if the caller doesn't pass in a date, or if the date is unsatisfactory (i.e. in the future or today when UTC is still < 23:59:59).  Force the caller to think about what they're asking for.

felliott:
I had an IRL discussion with @erinspace, capturing it here for posterity:

1. We should be stricter about the `date` parameter passed to summaries and events.  Since these functions can be called from different places (the individual script and the harness), they shouldn't default or DWIM.  Instead they should accept only date objects (not datetimes), and explicitly construct datetime ranges from 00:00:00Z to 23:59:59Z on that date.  It's up to the calling script to determine what the user wants and convert that to the appropriate date object.  If the given date is in the future or corresponds to a date that has not experienced 23:59:59Z yet, it should raise an error.  If no date is given, it should raise an error.

2. It's a bit pedantic, but we should distinguish between _summary and _event collections.  While they are very structurally similar, they are semantically different.  events are distinct objects with no overlap.  The value of the events doesn't change with the date given, just the subset.  Summaries are aggregations over a particular time period, so they are dependent on the definition of a date.  ex. If we change our definition of a date to be 00:00:00EST - 23:59:59EST and rerun the event and summary scripts, the event collection will be the same, but the summary data will change.  Also, the `keen.timestamp` field of a summary must be set to the given `date`, but events carry their own timestamp. Distinguishing the two will also make it less confusing if we later decide to have a `node_log_summary` along with `node_log_event`.

3. Not discussed, but something that occurred to me afterwords: if we're stricter about the `date` parameter for summaries, then that can become the "primary key" for deletion/reupload purposes.  If the `keen.timestamp` field always has the format `2016-11-01:00:00:00Z` for the Nov. 1 summary, it's very easy to query for and delete that.

scripts/analytics/node_log_events.py
```diff
@@ -0,0 +1,92 @@
...
+        date.isoformat(), (date + timedelta(1)).isoformat())
+    )
+
+    node_log_query = Q('date', 'lt', date + timedelta(1)) & Q('date', 'gte', date)
+
+    node_logs = NodeLog.find(node_log_query)
+    node_log_events = []
+    for node_log in node_logs:
+        event = {
+            'keen': {'timestamp': date.isoformat()},
```
felliott: I mentioned this in the summary comment: since this is an event, the timestamp should be the timestamp of the event itself rather than the given date object.

scripts/analytics/node_summary.py
```diff
@@ -0,0 +1,132 @@
...
+        },
+        'registered_nodes': {
+            'total': Node.find(registration_query).count(),
+        },
+        'registered_projects': {
+            'total': Node.find(registered_project_query).count(),
+        }
+    }
+
+    if is_today:
```
felliott: It looks like this is building summaries for dates that are not today, and snapshots for dates that are today.  Can we split these out into two scripts or methods and store them in separate collections?

scripts/analytics/user_summary.py
```diff
@@ -0,0 +1,88 @@
...
+            Q('date_confirmed', 'lt', date)
+        ).count(),
+        'unconfirmed_users': User.find(
+            Q('date_registered', 'lt', date) &
+            Q('date_confirmed', 'eq', None)
+        ).count(),
+        'deactivated_users': User.find(
+            Q('date_disabled', 'ne', None) &
+            Q('date_disabled', 'lt', date)
+        ).count()
```
felliott: I'd like to see `active_users`, `unconfirmed_users`, and `deactivated_users` restructured into something like:
````
{
  'status': {
    'active': ...,
    'unconfirmed': ...,
    'deactivated': ...,
  }
}
````

erinspace:
Significant refactor and response to code review in new hotfix PR in #6510 !

PR #6462, opened 2016-10-24
[WIP] Fedora addon

markpatton:
## Purpose

Storage addon that allows a user to interact with a Fedora 4 repository using HTTP basic authentication in a basic way. Fedora 4 is exposed as a hierarchy of files and folders, not LDP objects. 

This is not suitable for production usage, but is useful to explore how Fedora 4 and OSF might interact.

@mfraezz @felliott 

## Changes

Created the addon in the addons directory. Some other shared config files were also touched.

## Side effects

None.

## Ticket

None.


felliott:
Quick testing notes:  to run a local Fedora instance go [here](http://www.fedora-commons.org/download) and download the "One-Click-Run Application".  Run it on the command line with:

````
$ java -Dfcrepo.dynamic.stomp.port=61614 -Dfcrepo.home=/Users/fitz/tmp/fedora -Djetty.users.file=/Users/fitz/tmp/fedora/jetty-users.properties -jar fcrepo-webapp-4.6.0-jetty-console.jar
````

The first `-D` option is necessary for OSF devs b/c rabbitmq already binds the default port 61613.  The second option tells Fedora where to write its files.  The third tells Fedora where to find user accounts.  That file should be formatted [as described here](https://wiki.duraspace.org/display/FEDORA46/Authentication+and+Authorization#AuthenticationandAuthorization-Jetty).

**Comment by felliott** [comment]
Something else that bit me: if you see error messages about decoding a JWT, run this in your root osf directory: `python -m scripts.encrypt_external_accounts`

**Comment by markpatton** [comment]
@felliott 

Thanks for looking into this. I hope to finish up the Water Butler side by the end of the week and then start on the addon next.

**Comment by markpatton** [comment]
@felliott 

Here's an update. Things seem to be working generally well with the exceptions noted below. 

# Known issues

## Fedora account dialog

In the file tree, fedora always shows as "fedora: / (Full fedora)" even if it was passed a url to a non-root container.

It seems to be the case that the get_folders method in website/addons/fedora/model.py determines that label. As far as I can tell, get_folders is meant to retrieve the folders from a storage provider which a user has access to. In order to implement get_folders, a substantial amount of code from the water butler fedora provider would have to be copied. But actually in this case we do not need the folder list. The user specifies the folder in the repository url.

Is there a way to remove the folder selection from the dialog? Or would it be better to simply display "fedora: /user/chosen/path"? The "Full fedora" text doesn't really make sense.

## Infinite recursion bug

Fedora uses the same workaround as ownCloud to deal with a infinite recursion bug in a constructor. See website/addons/fedora/static/FedoraNodeConfig.js. But, the bug still triggers randomly.

## Revisions

The Fedora provider doesn't support revisions. Fedora actually does have support for versioning, but because that feature is in a bit of flux, it didn't seem worth implementing right now. The Fedora addon just copies another addon and always shows a boiler plate "latest" revision. I copied what another provider did, I think figshare, but it results in not useful information being displayed about the file. The date is empty etc. At the very least this revision metdata can be filled in.

## Testing

There is no unit testing of the Fedora addon. I intend to at least port the owncloud addon tests over.

**Comment by felliott** [comment]
Hey @mpatton,

Some responses:

#### Fedora account dialog:

Yes, you can remove the folder picker if it doesn't make sense.  GitHub has a repository picker, but no sub-folder pickers.  I'm not familiar enough with Fedora in practice to know if subfolder navigation is needed.  How would you expect a JHU user to use it?  If I link my Fedora account to a project, all my collaborators would have full access to it.  Setting the base folder gives the addon account owner the ability to restrict what their collaborators have access to.

As far as needing to copy over the Waterbutler code to support it, that's true, but there are a couple of things that make this easier to do in the OSF addon. If there's already an existing python client, you can use that. If not, you can build the request manually, but you don't have to worry about conforming to WB's interface.  You can just operate on the response directly.

Assuming you ditch the folder picker, using the path is probably a good first step.  If the connections are based off an individual user's id, it might also make sense to have them labeled something like "Fitz Elliott's Fedora workspace" or something like that.

#### Infinite recursion bug

I'll look into this.

#### Revisions

:+1: to "latest" if their revision support is still in flux.

#### Testing

:+1:

I'll start on a review now.

Cheers,
@felliott

**Comment by markpatton** [comment]
Latest commit to water butler Fedora provider has better latest revision support.

**Comment by markpatton** [comment]
@felliott 

Since the last update, I've made some minor cleanups so flake is happy, added some unit testing copied from ownCloud, and made the revision support a little better. At this point I've implemented everything that we want on our side. There are some issues remaining, but addressing them either takes greater knowledge of the codebase than I have or the direction to proceed is unclear.


# Known issues

## Fedora account dialog

In the file tree, fedora always shows as "fedora: / (Full fedora)" even if it was passed a url to a non-root container.

The get_folders method in website/addons/fedora/model.py seems to determine that label. The get_folders is meant to retrieve the folders from a storage provider which a user has access to. In order to implement get_folders, a substantial amount of code from the water butler fedora provider would have to be copied. But actually in this case we do not need the folder list. The user specifies the folder in the repository url. If folder selection is not supported, it should be removed.

Some of this comes down to how we expect OSF users to interact with a Fedora repository. The existing solution is more developer oriented than end user oriented.

## Infinite recursion bug

Fedora uses the same workaround as ownCloud to deal with a infinite recursion bug in a constructor. See website/addons/fedora/static/FedoraNodeConfig.js. But, the bug still triggers randomly. It may be related to activating both the Fedora provider and ownCloud addon at the same time.

## Credentials

The addon doesn't do anything to verify credentials in the account dialog. This is a complicated topic that involves the rest of whatever application stack is using Fedora. See https://wiki.duraspace.org/display/FEDORA471/Authorization.

File: website/addons/fedora/static/FedoraUserConfig.js [hunk 1]
```diff
@@ -0,0 +1,125 @@
...
+var osfHelpers = require('js/osfHelpers');
+var oop = require('js/oop');
+var $modal = $('#fedoraCredentialsModal');
+
+var FedoraViewModel = oop.extend(OAuthAddonSettingsViewModel,{
+    constructor: function(url){
+        var self = this;
+        self.super.constructor.call(self, 'fedora', 'Fedora');
+
+        const otherString = 'Other (Please Specify)';
```
felliott: `s/const/var/`, see: https://github.com/CenterForOpenScience/osf.io/commit/3362d15f6238ba862764b980459c2fd61942cd18

website/addons/fedora/static/FedoraUserConfig.js
```diff
@@ -0,0 +1,125 @@
...
+            return (self.selectedHost() === otherString || !self.hasDefaultHosts());
+        });
+        self.showCredentialInput = ko.pureComputed(function() {
+            return true;
+        });
+        self.hasDefaultHosts = ko.pureComputed(function() {
+            return Boolean(self.hosts().length);
+        });
+    },
+    updateAccounts: function(){
```
felliott: I don't think you need this. `updateAccounts()` should be installed by oop.js

website/addons/fedora/static/FedoraUserConfig.js
```diff
@@ -0,0 +1,125 @@
...
+            url: self.url,
+            type: 'GET',
+            dataType: 'json'
+        }).done(function (response) {
+            self.hosts(response.hosts);
+            self.loaded(true);
+            self.updateAccounts();
+        }).fail(function (xhr, textStatus, error) {
+            self.setMessage(language.userSettingsError, 'text-danger');
+            Raven.captureMessage('Could not GET fedora settings', {
```
felliott: Minor: "fedora" should probably be capitalized here and in the other messages.

website/addons/fedora/static/FedoraUserConfig.js
```diff
@@ -0,0 +1,125 @@
...
+        self.customHost(null);
+    },
+    connectAccount : function() {
+        var self = this;
+        // Selection should not be empty
+        if( self.hasDefaultHosts() && !self.selectedHost() ){
+            if (self.useCustomHost()){
+                self.setMessage('Please enter an fedora server.', 'text-danger');
+            } else {
+                self.setMessage('Please select an fedora server.', 'text-danger');
```
felliott: "an fedora" => "a Fedora"

website/addons/fedora/static/FedoraUserConfig.js
```diff
@@ -0,0 +1,125 @@
...
+                self.setMessage('Please select an fedora server.', 'text-danger');            
+            }
+            return;
+        }
+        if ( !self.useCustomHost() && !self.username() && !self.password() ){
+            self.setMessage('Please enter a username and password.', 'text-danger');
+            return;
+        }
+        if ( self.useCustomHost() && ( !self.customHost() || !self.username() || !self.password() ) )  {
+            self.setMessage('Please enter an fedora host and credentials.', 'text-danger');
```
felliott: "an fedora" => "a Fedora"

website/addons/fedora/static/FedoraNodeConfig.js
```diff
@@ -0,0 +1,130 @@
...
+    clearModal : function() {
+        var self = this;
+        self.selectedHost(null);
+        self.customHost(null);
+    },
+    connectAccount : function() {
+        var self = this;
+        if( self.hasDefaultHosts() && !self.selectedHost() ){
+            if (self.useCustomHost()){
+                self.changeMessage('Please enter an fedora server.', 'text-danger');
```
felliott: "an fedora" => "a Fedora"

website/addons/fedora/static/FedoraNodeConfig.js
```diff
@@ -0,0 +1,130 @@
...
+            } else {
+                self.changeMessage('Please select an fedora server.', 'text-danger');
```
felliott: "an fedora" => "a Fedora"

website/addons/fedora/static/FedoraNodeConfig.js
```diff
@@ -0,0 +1,130 @@
...
+                self.changeMessage('Please select an fedora server.', 'text-danger');            
+            }
+            return;
+        }
+        if ( !self.useCustomHost() && !self.username() && !self.password() ){
+            self.changeMessage('Please enter a username and password.', 'text-danger');
+            return;
+        }
+        if ( self.useCustomHost() && ( !self.customHost() || !self.username() || !self.password() ) )  {
+            self.changeMessage('Please enter an fedora host and credentials.', 'text-danger');
```
felliott: "an fedora" => "a Fedora"

website/addons/fedora/static/FedoraNodeConfig.js
```diff
@@ -0,0 +1,130 @@
...
+                    self.changeMessage(self.messages.connectAccountSuccess(), 'text-success', 3000);
+                }
+                catch(err){
+                    self.changeMessage(self.messages.connectAccountDenied(), 'text-danger', 6000);
+                }
+            });
+        }).fail(function(xhr, textStatus, error) {
+            var errorMessage = (xhr.status === 401) ? language.authInvalid : language.authError;
+            self.changeMessage(errorMessage, 'text-danger');
+            Raven.captureMessage('Could not authenticate with fedora', {
```
felliott: "fedora" => "Fedora"

website/addons/fedora/serializer.py
```diff
@@ -0,0 +1,53 @@
...
+    addon_short_name = 'fedora'
+
+    def serialized_folder(self, node_settings):
+        return {
+            'name': node_settings.fetch_folder_name(),
+            'path': node_settings.folder_id
+        }
+
+    def credentials_are_valid(self, user_settings, client=None):
+        return True
```
felliott: Will follow-up offline.

website/addons/fedora/model.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def after_delete(self, node, user):
+        self.deauthorize(Auth(user=user), add_log=True)
+        self.save()
+
+    def on_delete(self):
+        self.deauthorize(add_log=False)
+        self.save()
+
+    def get_folders(self, **kwargs):
```
felliott: Will follow-up offline.

api/base/settings/defaults.py
```diff
@@ -230,7 +230,7 @@
 VARNISH_SERVERS = osf_settings.VARNISH_SERVERS
 ESI_MEDIA_TYPES = osf_settings.ESI_MEDIA_TYPES
 
-ADDONS_FOLDER_CONFIGURABLE = ['box', 'dropbox', 's3', 'googledrive', 'figshare', 'owncloud']
+ADDONS_FOLDER_CONFIGURABLE = ['box', 'dropbox', 's3', 'googledrive', 'figshare', 'owncloud', 'fedora']
 ADDONS_OAUTH = ADDONS_FOLDER_CONFIGURABLE + ['dataverse', 'github', 'mendeley', 'zotero', 'forward']
```
felliott: No action needed right now, but if we decide not to go with the folder browser, `'fedora'` will need to be moved from `ADDONS_FOLDER_CONFIGURABLE` to here.

felliott:
Oops, another thing I just noticed.  You need to add fedora to [this section of addons.json](https://github.com/DataConservancy/osf.io/blob/d10102a3840e2bda9feb4c95513da200128ce4bb/addons.json#L19).  Otherwise, Fedora won't get registered. Ditto to `addons_commentable` if you want file comments to work.

felliott:
Need to fix the connect/disconnect/reconnect case in views.py (fedora_add_user_account).

felliott:
Check the validity of the root url with a HEAD request during account linking process.

felliott:
Remove folder picker for now when connecting to a project.

felliott:
Trying to upload a file called `afl-3.0 (3).txt` succeeds, but reports failure.  The waterbutler provider complains with the following message:

````

PR #6467, opened 2016-10-25
[RFR] addons/OsfStorage

cwisecarver:
## Purpose

Make osfstorage work with the django models.

## Changes

Reimplement osfstorage using django models. Fix tests. Fix codes. 

## Side effects

It can't be any worse than not working at all.


cwisecarver:
@sloria If you could review this whenever you have a few minutes.

addons/osfstorage/tests/test_views.py
```diff
@@ -1,203 +1,739 @@
...
+
+def create_record_with_version(path, node_settings, **kwargs):
+    version = factories.FileVersionFactory(**kwargs)
+    node_settings.get_root().append_file(path)
+    record.versions.append(version)
+    record.save()
+    return record
+
+
+@pytest.mark.django_db
```
sloria: This is fine, but you can also put 

```python
pytestmark = pytest.mark.django_db
```

At the top of the file to mark all tests in the file.

sloria:
Code looks correct; works locally.

PR #6468, opened 2016-10-25
[rfr] [djangosf] Fix setting root

sloria:
Now, root will be set upon first save of a Node. It needs
to be set in post_save because the NodeRelation to parent
must already exist.

We use `.filter(id=...).update(...)` in order to prevent signals
from being sent.

One side effect of this is that newly-created Nodes will
need to call node.reload() or node.refesh_from_db() if they
need to access `root`.


sloria:
Also added a small change to fix https://trello.com/c/hrDXt7zs/17-extra-project-showing-on-user-profile

PR #6472, opened 2016-10-25
[feature] Bypass API Throttle Via Request Header Token [OSF-7114]

caseyrollins:
#### Purpose
- Allows for the API throttle to be bypassed by including the correct token in the `X-THROTTLE-TOKEN` header of the request. 
#### Changes
- Changes the `BaseThrottle` class to check for an `X-THROTTLE-TOKEN` in the request header. 
- Adds `BYPASS_THROTTLE_TOKEN` to [defaults.py](https://github.com/caseyrollins/osf.io/blob/f868a80fea911da392c9278619e1894e179fc46a/api/base/settings/defaults.py#L234).
- Adds a check in [**init**.py](https://github.com/caseyrollins/osf.io/blob/f868a80fea911da392c9278619e1894e179fc46a/api/base/settings/__init__.py#L26) that the `BYPASS_THROTTLE_TOKEN` has been changed from the default in local.py.
#### Notes
- `BYPASS_THROTTLE_TOKEN` needs to be defined in local.py on production.
- Prerender needs to include the correct token in an `X-THROTTLE-TOKEN` header for all requests to bypass the API's throttle limit. 
#### Side effects
- None expected.
#### Ticket
- [OSF-7114](https://openscience.atlassian.net/browse/OSF-7114)


api/base/throttling.py
```diff
@@ -1,15 +1,40 @@
...
 
+from api.base import settings
+
 
 class BaseThrottle(SimpleRateThrottle):
 
     def failure(self, request):
         return False
 
+    def get_ident(self, request):
```
mfraezz: Not a blocker, but this could be rewritten to check for `HTTP_X_THROTTLE_TOKEN` first and return the `super` call if it's absent, as this has a three line diff from the [base implementation](https://github.com/tomchristie/django-rest-framework/blob/a68b37d8bc432fae37ef5880aec500002b59f565/rest_framework/throttling.py#L26). Something like:

```python
def get_ident(self, request):
    if request.META.get('HTTP_X_THROTTLE_TOKEN'):
        return request.META['HTTP_X_THROTTLE_TOKEN']
    return super(BaseThrottle, self).get_ident(request)
```

api/base/settings/__init__.py
```diff
@@ -11,19 +11,27 @@
...
 from .defaults import *  # noqa
+from website import settings
 
 try:
     from .local import *  # noqa
 except ImportError as error:
     warnings.warn('No api/base/settings/local.py settings file found. Did you remember to '
                   'copy local-dist.py to local.py?', ImportWarning)
 
+if not settings.DEV_MODE:
```
mfraezz: This might break the admin module -- it should probably be checking `os.environ.get('DJANGO_SETTINGS_MODULE')` as below.

api/base/settings/__init__.py
```diff
@@ -11,19 +11,27 @@
...
+    if not local.BYPASS_THROTTLE_TOKEN:
```
mfraezz: Does this work? `*` is imported from `.local` above, but `local` is imported below. I'd expect it to be undefined. Ditto for `defaults`

api/base/throttling.py
```diff
@@ -1,15 +1,40 @@
...
+            client_addr = addrs[-min(num_proxies, len(addrs))]
+            return client_addr.strip()
+
+        return ''.join(xff.split()) if xff else remote_addr
+
     def allow_request(self, request, view):
         """
         Implement the check to see if the request should be throttled.
         """
+        throttle_token = settings.BYPASS_THROTTLE_TOKEN
```
mfraezz: Superfluous var assignment -- the app should kill itself with the `assert` in `/api/settings/base/__init__.py` `if not settings.BYPASS_THROTTLE_TOKEN`. Just check `self.get_ident(request) == settings.BYPASS_THROTTLE_TOKEN` below.

api/base/settings/__init__.py
```diff
@@ -11,19 +11,27 @@
...
+        warnings.warn('Local throttle token should be set.')
+    else:
+        assert local.BYPASS_THROTTLE_TOKEN != defaults.BYPASS_THROTTLE_TOKEN, 'Local throttle token should be changed from the default value.'
```
mfraezz: You could probably just add `'BYPASS_THROTTLE_TOKEN'` to the tuple in the `for` loop below.

api/base/settings/__init__.py
```diff
@@ -11,17 +11,18 @@
...
+from website import settings
 
 try:
     from .local import *  # noqa
 except ImportError as error:
     warnings.warn('No api/base/settings/local.py settings file found. Did you remember to '
                   'copy local-dist.py to local.py?', ImportWarning)
 
-if not DEBUG and os.environ.get('DJANGO_SETTINGS_MODULE') == 'api.base.settings':
+if not settings.DEV_MODE and os.environ.get('DJANGO_SETTINGS_MODULE') == 'api.base.settings':
```
mfraezz: Minor: `DEV_MODE` appears to be the only thing all of `website.settings` is imported for.

PR #6475, opened 2016-10-26
[WIP] Feature/OSF Changes needed for Edit Preprint Functionality [OSF-7168][PREP-137]

pattisdr:
❗️ Waiting on https://github.com/CenterForOpenScience/osf.io/pull/6346
## Purpose

Small changes in addition to Matt's guid work required for 'Edit' preprint functionality to behave correctly.
## Changes
- Log preprint_initiated when preprint is published, not when it is created (text of log is still "Dawn Pattison made {project_name} a preprint")
- If user is admin, show 'Edit' button on project overview page for a preprint.
- Increase the number of emails that can be sent when contributors are added to a project from 10/hr to 1000/hr
## Side effects

I needed to increase the number of contributors that could be added to a project so the throttling limit wouldn't be hit.  I think I changed this in the right place?  The main thing, I don't want contributor operations to be limited when users manually add them.
## Ticket

https://openscience.atlassian.net/browse/PREP-137
https://openscience.atlassian.net/browse/OSF-7168


website/addons/osfstorage/views.py
```diff
@@ -101,6 +101,11 @@ def osfstorage_move_hook(source, destination, name=None, **kwargs):
         raise HTTPError(httplib.METHOD_NOT_ALLOWED, data={
             'message_long': 'Cannot move file as it is checked out.'
         })
+    except exceptions.FileNodeIsPrimaryFile:
+        raise HTTPError(httplib.METHOD_NOT_ALLOWED, data={
```
brianjgeiger: This should be FORBIDDEN because you can PATCH the file in other circumstances.

website/addons/osfstorage/views.py
```diff
@@ -217,6 +222,10 @@ def osfstorage_delete(file_node, payload, node_addon, **kwargs):
 
     except exceptions.FileNodeCheckedOutError:
         raise HTTPError(httplib.FORBIDDEN)
+    except exceptions.FileNodeIsPrimaryFile:
+        raise HTTPError(httplib.FORBIDDEN, data={
```
brianjgeiger: This should be METHOD_NOT_ALLOWED, because the DELETE method is not allowed for the file in this instance.

website/files/models/osfstorage.py
```diff
@@ -108,14 +108,15 @@ def is_checked_out(self):
         return self.checkout is not None
 
     def delete(self, user=None, parent=None):
-        if self.node.preprint_file == self:
-            self.node._is_preprint_orphan = True
-            self.node.save()
+        if self.node.preprint_file == self and not self.node._has_abandoned_preprint:
```
brianjgeiger: We should have a property named something like "delete_allowed" that does the calculation here. That way we can serialize it and make it available to Fangorn and the API. Then this method would just be `if self.delete_allowed:`

brianjgeiger: And let's go ahead and add the field to the API v2 File Serializer when this is updated.

website/addons/osfstorage/views.py
hmoco: @brianjgeiger, making those changes. However, that is the behavior used for checked-out files in both cases (move and delete).

brianjgeiger:
@pattisdr If we could pull out @hmoco's preprint-not-deleting stuff into its own ticket, that might make this easier to deal with.

pattisdr:
@brianjgeiger - so remove Henrique's work off of this branch?

brianjgeiger:
Yeah, unless it's holding up some other code. We'll get it in on a separate PR.

pattisdr:
@brianjgeiger okay finished.  This PR is very small now!

brianjgeiger:
@pattisdr The old staging-2 branch was in a weird state, so I replaced it with a copy of develop. Could you rebase off the current feature/staging-2?

brianjgeiger:
Hmmm. Hold off on that for a minute. Something weird is happening.

brianjgeiger:
Hmmm. A rebase should be fine. Let me know if it is weird for any reason, please.

pattisdr:
I am getting conflicts - not sure how to resolve.

PR #6476, opened 2016-10-26
[Feature Fix] Use promise to stop Editor ajax double call [OSF-6883]

alexschiller:
## Purpose

Reworks my previous changes to [OSF-6883]
## Changes

Instead of doing a check if created, this extracts the ajax call into a separate object. The FileEditor then checks for a previous call if reload=false, or simply reloads if reload=true;

## Side effects

None known.

## Ticket

https://openscience.atlassian.net/browse/OSF-6883
[#OSF-6883]

## New QA considerations.
Since this change has more of a chance of breaking things, the basic functionality of the file editor should be checked to ensure editing/saving/reverting works properly.

### DIFF
--- a/website/static/js/filepage/editor.js
+++ b/website/static/js/filepage/editor.js
@@ -26,11 +26,11 @@ var FileFetcher = {
                 url: url,
                 dataType: 'text',
                 beforeSend: $osf.setXHRAuthorization,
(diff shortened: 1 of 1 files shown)


alexschiller:
Yes.

PR #6477, opened 2016-10-27
[WIP][Feature] Enable Swagger with API Versioning

mfraezz:
## Purpose

API Versioning (#6329) went out recently, and broke Swagger. This fixes Swagger.
## Changes
- Update DR(F|S) to versions allow Swagger to support Versioning
- Update templates based on the DRS version bump 0.3 to 2.0
- Copy/Modify helper function in `rest_framework_swagger/views.py` to allow kwargs through
- Add default value for `kwargs[user_lookup_url_kwarg]` so that Swagger doesn't through a `KeyError` when building views.
## Side effects

**WIP:** Bug noticed (possibly in DRS) with how API version is specified. The regex in `urls.py` is `'^(?P<version>(v2))/`, but this appears mishandled on the frontend, causing any input make a request to `<base>/<input>)/`, e.g. `http://api.osf.io/v2)/`. Fixing this may require a slight modification to our versioning system or a monkey patch.
## Ticket

None known


api/base/urls.py
```diff
@@ -1,17 +1,17 @@
...
 from . import settings
 from . import versioning
+from api.base.swagger import get_swagger_view
 
 default_version = versioning.decimal_version_to_url_path(settings.REST_FRAMEWORK['DEFAULT_VERSION'])
 
 # Please keep URLs alphabetized for auto-generated documentation
 urlpatterns = [
-    url('^(?P<version>(v2))/',
+    url(r'^(?P<version>(v2))/',
```
mfraezz: The `WIP` note can be resolved by replacing the parenthesized option with the standard url param convention (`r'^(?P<version>\w+)/'`), but this may have unintended consequences.

api/base/swagger.py
```diff
@@ -0,0 +1,40 @@
...
+from rest_framework.views import APIView
+
+from rest_framework_swagger import renderers
+
+
+def get_swagger_view(title=None, url=None):
+    """
+    Returns schema view which renders Swagger/OpenAPI.
+
+    (Replace with DRF get_schema_view shortcut in 3.5)
```
mfraezz: TODO: Investigate

brianjgeiger:
Also, alternate implementation: The only thing I really like about DRS is the generation of the json file. Having it handle the creation and serving of the swagger-ui is inefficient and inconvenient. Better would be to have it create a static file either as an invoke task or on startup, then serving that information separately, so that we can upgrade or modify the swagger-ui independently of DRS.

brianjgeiger:
And I talked with Chrïs. Having had to go through the last Django/DRF pain, he did not upgrade those for DjangOSF, so this can't happen until DjangOSF is out on production and we have a spare cycle for QA to test.

mfraezz:
Improvements in API documentation obviate this (h/t @caseyrollins ), though we should still update Django/DRF in the near future. Closing this.

PR #6482, opened 2016-10-27
[djangosf] GoogleDrive Add-on

mfraezz:
## Purpose
Add support for GoogleDrive to DjangOSF

## Changes
* Migrate models/settings/tests: `website/addons/googledrive/` --> `addons/googledrive/`

## Side effects
None expected

## Ticket
None known


addons/googledrive/tests/factories.py
```diff
@@ -0,0 +1,35 @@
...
+from addons.googledrive.models import NodeSettings
+from addons.googledrive.models import UserSettings
+
+
+class GoogleDriveAccountFactory(ExternalAccountFactory):
+    provider = 'googledrive'
+    provider_id = factory.Sequence(lambda n: 'id-{0}'.format(n))
+    oauth_key = factory.Sequence(lambda n: 'key-{0}'.format(n))
+    oauth_secret = factory.Sequence(lambda n: 'secret-{0}'.format(n))
+    expires_at = datetime.datetime.now() + relativedelta(days=1)
```
cwisecarver: We should be using `django.utils.timezone.now()` for non-naive datetimes with django. Anything that's naive that goes into the db will throw a warning and potentially screw someone later if they try to compare a naive datetime to a non-naive datetime. That will throw an exception.

cwisecarver: https://docs.djangoproject.com/en/1.10/topics/i18n/timezones/#naive-and-aware-datetime-objects

addons/googledrive/tests/test_models.py
```diff
@@ -0,0 +1,123 @@
...
+            'id': 'fake-folder-id',
+            'name': 'fake-folder-name',
+            'path': 'fake_path'
+        }
+        self.node_settings.set_folder(folder, auth=Auth(self.user))
+        self.node_settings.save()
+        # Folder was set
+        assert_equal(self.node_settings.folder_id, folder['id'])
+        # Log was saved
+        last_log = self.node.logs.first()  # Django RelatedManagers prepend new objects
```
cwisecarver: I think you want `self.node.logs.latest()` amirite @sloria?

addons/googledrive/models.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def clear_settings(self):
+        self.folder_id = None
+        self.folder_path = None
+
+    def get_folders(self, **kwargs):
+        node = self.owner
+
+        #  Defaults exist when called by the API, but are `None`
+        path = kwargs.get('path') or ''
```
mfraezz: No. The API can add `path=None` to kwargs, causing `.get('path', '')` to return `None`, which it shouldn't.

cwisecarver: Guess I could have read that comment right above it... 👍

addons/googledrive/models.py
```diff
@@ -0,0 +1,243 @@
...
+        ]
+        return contents
+
+    def set_folder(self, folder, auth):
+        """Configure this addon to point to a Google Drive folder
+
+        :param dict folder:
+        :param User user:
+        """
+        self.folder_id = folder['id']
```
mfraezz: Theoretically, no. They're validated either in the API or on the frontend before this is hit.

cwisecarver: Sometimes it's better to let the `KeyError` blow up if it means someone broke something so bad and they otherwise wouldn't know it. I was just checking.

addons/googledrive/tests/factories.py
cwisecarver: Yeah, I new it was a factory. I just wanted to practice like we play.

addons/googledrive/tests/test_models.py
mfraezz: You are correct.

addons/googledrive/models.py
```diff
@@ -0,0 +1,243 @@
...
+class UserSettings(BaseStorageAddon, BaseOAuthUserSettings):
+    oauth_provider = GoogleDriveProvider
+    serializer = GoogleDriveSerializer
+
+
+class NodeSettings(BaseStorageAddon, BaseOAuthNodeSettings):
+    oauth_provider = GoogleDriveProvider
+    provider_name = 'googledrive'
+
+    folder_id = models.TextField(default=None, null=True, blank=True)
```
mfraezz: Probably not, but Box does it.

cwisecarver: Django used to give a warning about that, if it doesn't anymore then being explicit might be better.

addons/googledrive/models.py
mfraezz: That's fair. I'll leave it the way it is.

sloria:
LGTM

PR #6484, opened 2016-10-27
[djangosf] Modify transaction handlers for djangosf compat

sloria:
## Purpose

Make views atomic in the Flask app.

## Notes
django.db.transaction.atomic() provides high-level interface for
starting, rolling back, and committing transactions. It can only
be used as either a decorator or context manager, so I call
__enter__ and __exit__

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -16,7 +16,6 @@ import pytest
 from nose.tools import *  # noqa PEP8 asserts
 from django.utils import timezone
 from django.apps import apps
-from flask import jsonify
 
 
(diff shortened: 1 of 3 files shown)


tests/test_views.py
```diff
@@ -12,13 +12,15 @@
 import unittest
 
 import mock
+import pytest
 from nose.tools import *  # noqa PEP8 asserts
 from django.utils import timezone
 from django.apps import apps
+from flask import jsonify
```
cwisecarver: This doesn't appear to be used anywhere.

PR #6485, opened 2016-10-27
[RFR] Make node.root do magic SQL

cwisecarver:
## Purpose

Remove reverse field for node root and switch it to use a SQL queries.
## Changes

Rewrote SQL to get the top parent of a node. Wrote overly complicated test to make sure it works.
Added custom manager to AbstractNode with `get_children(root=the_root_node)` method that gets a queryset of children, recursively, of the root node.
## Side effects

SQL queries could need to be optimized later.


cwisecarver:
@sloria I think I got it. Have a look when you get a chance.

cwisecarver:
I just added a thing that if the root comes back null it returns self.

PR #6486, opened 2016-10-28
[RFR] Encrypted fields for django

cwisecarver:
## Purpose

Port encrypted field functionality to django.

## Changes

Added an EncryptedTextField field, added some tests, implemented new field on ExternalAccount model.

## Side effects

None expected.


osf/utils/fields.py
```diff
@@ -7,3 +9,25 @@ def get_prep_value(self, value):
...
         return value
+
+
+class EncryptedTextField(models.TextField):
+    '''
+    This field transparently encrypts data in the database. It should probably only be used with PG unless
+    the user takes into account the db specific trade-offs with TextFields.
+    '''
+    prefix = 'jwe:::'
+    SENSITIVE_DATA_KEY = jwe.kdf(settings.SENSITIVE_DATA_SECRET.encode('utf-8'), settings.SENSITIVE_DATA_SALT.encode('utf-8'))
```
mfraezz: Consider making this a module-level variable so the key doesn't need to be derived every time the class is instantiated

PR #6487, opened 2016-10-28
[WIP] Fix some sentry bugs and make naive datetimes illegal

cwisecarver:
## Purpose

Bugs:
- https://sentry.cos.io/sentry/osf-back-end/issues/258283/
- https://sentry.cos.io/sentry/osf-back-end/issues/257147/

I also added a NonNaiveDatetimeField and changed the DatetimeAwareJSONField to disallow naive datetimes. They will both throw an exception if someone tries to give them a naive datetime. I also changed all usages throughout the models to use NonNaiveDatetimeField instead of django's DatetimeField.

### DIFF
--- a/osf/utils/datetime_aware_jsonfield.py
+++ b/osf/utils/datetime_aware_jsonfield.py
@@ -7,13 +7,9 @@ from dateutil import parser
 from django.contrib.postgres import lookups
 from django.contrib.postgres.fields.jsonb import JSONField
 from django.core.serializers.json import DjangoJSONEncoder
-
-from osf.exceptions import ValidationError
 from psycopg2.extras import Json
 
-
-class NaiveDatetimeException(BaseException):
-    pass
+from osf.exceptions import ValidationError, NaiveDatetimeException
 
 
 class DateTimeAwareJSONEncoder(DjangoJSONEncoder):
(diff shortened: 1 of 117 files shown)


cwisecarver:
@sloria I think this is ready for a review/merge.

cwisecarver:
Fixing tests broken by Timezone aware datetime changes.

cwisecarver:
@sloria This is RFR again.

osf/utils/datetime_aware_jsonfield.py
```diff
@@ -12,13 +12,23 @@
 from psycopg2.extras import Json
 
 
+class NaiveDatetimeException(BaseException):
```
sloria: Custom exceptions should inherit from `Exception` not `BaseException`. Source: https://docs.python.org/3/library/exceptions.html?highlight=baseexception#BaseException

sloria: Also, it might make sense to move this to `osf.exceptions`. That's minor, though.

cwisecarver: I know you've told me that `BaseException` thing before. I always forget and then assume that it would be the one you'd think it would be... Fixed both of those, merging.

PR #6490, opened 2016-10-29
[Feature] Fix fangorn move actions [OSF-7164]

alexschiller:
## Purpose

There are a number of issues with fangorn moving and dropping that lead to some strange WB issues that should not be able to occur at all. This PR helps block a few of those by locking down files while waiting for the WB ajax to respond before releasing the file. In addition, this includes a number of other fixes and brand new fangorn unit test file testing every bit of the move/copy/forbid logic. 

## Changes
1. Blocks clicking on files that have a pending waterbutler ajax (move/rename/copy) action. This releases when a response is returned. This means that `weird things` can still occur if the 202 'taking longer than expected' modal pops up, but was done this way to avoid infinite lockups that require a page refresh. [OSF-7164]
2. Blocks dragging of folders that have a child with a pending wb action. Folders will just not have any droppable targets allowed. [OSF-7164]
3. Parent folders can no longer be dragged into own children folders [OSF-7162]
4. Fangorn data undefined error on folder moves was occuring in the move/copy/forbidden logic. Returns forbidden if data is undefined. So rarely all drop actions will be disabled if folders are not fully populated when dragging starts [OSF-7162]
5. Add a boat load of unit tests for move/copy/forbidden drag and drop logic.
   
## Side effects
1. Figshare file drop logic was wrong and attempted to allow files in public figshare projects to be copied out. Waterbutler denied these calls. Now drops should be prevented. Private figshare projects might allow dropping out. If this is tested and throws an error, please let me know and I'll disable it.
2. Dropping on files was allowed and actually defaulted to moving the object to the parent of the child. After conferring with Nici and Sara this behavior was removed as it was strange, and possibly lead to some fun edge cases. 
   
## Ticket

https://openscience.atlassian.net/browse/OSF-7162
https://openscience.atlassian.net/browse/OSF-7163
https://openscience.atlassian.net/browse/OSF-7164
[#OSF-7164]

## QA Considerations

This is unfortunately a beast of a changelog for file/folder dragging and dropping. This will probably require testing some addons that have weird move/copy/forbidden behavior. 

The ones that should be tested are figshare, dataverse, and github. Only the actions that result in different behaviors need to be tested for these specific addons (e.g., draft versus published figshare).

### DIFF
--- a/website/addons/dataverse/static/dataverseFangornConfig.js
+++ b/website/addons/dataverse/static/dataverseFangornConfig.js
@@ -5,7 +5,7 @@ var URI = require('URIjs');
 var $ = require('jquery');
 var Raven = require('raven-js');
 
-var Fangorn = require('js/fangorn');
+var Fangorn = require('js/fangorn').Fangorn;
 var waterbutler = require('js/waterbutler');
 var $osf = require('js/osfHelpers');
 
--- a/website/addons/figshare/static/figshareFangornConfig.js
+++ b/website/addons/figshare/static/figshareFangornConfig.js
@@ -2,7 +2,7 @@
 
 var m = require('mithril');
 
-var Fangorn = require('js/fangorn');
+var Fangorn = require('js/fangorn').Fangorn;
 
 
 // Define Fangorn Button Actions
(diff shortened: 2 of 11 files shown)


alexschiller:
Closing in favor of #6525 which merged all of these changes and more.

PR #6493, opened 2016-10-31
[Feature] Admin tool DOI removal for spam [OSF-7141]

alexschiller:
## Purpose
Currently admin marking spam does not change any DOI status. This allows spam that has a DOI/ARK id to be be marked as unavailable (spam) or to remark the DOI as public if accidentally marked as spam.

## Changes

1. Add change_status_identifiers to EzidClient to allow changes to status
2. Add call on confirm spam and confirm not spam to change the status to `unavailable | spam` or `public`
3. Cleanup admin docs

## Side effects
None known.

## Ticket
[#OSF-7141]
https://openscience.atlassian.net/browse/OSF-7141

## QA Notes
Do not test DOI/ARK creation production!! DOIs are supposed to be persistent, permanent identifiers. Staging/local use their testing API, which creates DOI/ARK ids that delete at 2 weeks.

You can confirm the status of the DOI by pasting the DOI from the project page into the "resolve doi" on https://www.doi.org/. Creation of the DOI and changes via the admin panel do not resolve automatically. It sometimes took 5-10 minutes for changes to populate there, so make a couple projects and get DOIs to test with. Quickly marking as spam and then marking as not-spam will lead to weirdness (not sure which state it will finally resolve to). So when testing marking not-spam after it has been marked spam, please wait for the DOI to resolve to the proper state in between.

Steps:
1. Test that for a public project with a DOI/ARK id: using the doi on doi.org resolves to the right project
2. Mark the project as spam
3. Test that the spam-marked project resolves to the spam page
4. Test that marking not spam resolves back to the original DOI


website/project/views/register.py
```diff
@@ -252,6 +252,13 @@ def _get_or_create_identifiers(node):
         }
 
 
+def osf_admin_change_status_identifier(node, status):
+    if node.get_identifier_value('doi') and node.get_identifier_value('ark'):
```
brianjgeiger: I don't know enough about identifiers to know the answer to: why `and` here?

alexschiller: The logic that creates identifiers uses an `or` and does not allow making of identifiers if one or the other exists. 
In theory it should not be possible to have one without the other if I understand ezid. Swapping to an `or` or even just checking one or the other is probably sufficient. This was mostly just to be overly explicit about what you should have access to at the time.

With that info, is there a preference for how to write it?

PR #6498, opened 2016-11-01
[feature] Fix Dashboard Institution Display [OSF-6980]

caseyrollins:
#### Purpose
- Fixes #6351 / #6474 

#### Changes
- Adds the `institution_dashboard_display` boolean field to the `Node` model (`dashboard_display` on the `Institution` "model").
- Moves the expensive query to a celery task which is called by the institution dashboard view.
- Adds a script to set/update the `dashboard_display` field for existing institutions, which should be run to update the institutions on staging & production, and could be run periodically to update which institutions should be shown on the dashboard.

#### Side effects
- The "display on dashboard" boolean for each institution will only be updated when the dashboard for that institution is viewed or when the `institution_dashboard_display.py` script is run. 
- If an institution is displayed on the dashboard, it will remain on the dashboard (even if it later has < 5 affiliated nodes) unless the `institution_dashboard_display.py` script is run (i.e. once an institution meets the threshold for being displayed, we will continue to display it even if some nodes remove their affiliation). 

#### Ticket
- [OSF-6980](https://openscience.atlassian.net/browse/OSF-6980)


scripts/institution_num_nodes.py
```diff
@@ -0,0 +1,63 @@
...
+logger = logging.getLogger(__name__)
+
+
+def set_institution_affiliated_nodes_count():
+    all_institutions = Institution.find()
+    for inst in all_institutions:
+        logger.info(
+            'Finding affiliated nodes for Institution with id <{}> and title <{}>'.format(inst._id, inst.name)
+        )
+        affiliated_nodes = Node.find_by_institutions(
```
caseyrollins: @sloria, should I be using `framework.mongo.utils.paginated()` here, or will this be okay running on production?

website/institutions/views.py
```diff
@@ -6,11 +6,27 @@
...
+
 def view_institution(inst_id, **kwargs):
     try:
         inst = Institution.find_one(Q('_id', 'eq', inst_id) & Q('is_deleted', 'ne', True))
     except NoResultsFound:
         raise HTTPError(http.NOT_FOUND)
+
+    # The total number of affiliated nodes for each institution is calculated and cached
+    # in the Institution.num_nodes (aka Node.institution_num_nodes) field.
+    num_nodes = len(Node.find_by_institutions(inst, query=(
```
brianjgeiger: A couple of potential improvements since I'm somewhat cautious now:

- This should check to see if num_nodes is already set on the model and if it's below the threshold. If it's above the threshold, then there's no need to run this. 
- We should make a model method that does this setting, and that method should do a post-commit celery task. The first part of the celery task should check the threshold as well and not bother running the query if it's above the threshold.

So this way, we can't have 300 requests all at once, and once it gets past that point, further queries will be bypassed and eventually not made at all. Also, this will DRY up the code a bit.

website/project/tasks.py
```diff
@@ -49,3 +51,19 @@ def on_node_updated(node_id, user_id, first_save, saved_fields, request_headers=
                     }]
                 },
             }, headers={'Authorization': 'Bearer {}'.format(settings.SHARE_API_TOKEN)}).raise_for_status()
+
+
+@celery_app.task(ignore_results=True)
+def institution_dashboard_display(inst):
```
brianjgeiger: Minor, but let's call this `institution_set_dashboard_display()` to keep it from looking like the property.

brianjgeiger:
@caseyrollins Merge conflicts.

caseyrollins:
Closing in favor of #7174

PR #6499, opened 2016-11-01
[Feature] handle &, <, > in dashboard and po [OSF-6853]

alexschiller:
## Purpose
& become &amp; on the Dashboard and Project Organizer.
The ideal solution to this would handle amps better on the backend. This is a cosmetic front end hack/fix to make amps show up properly. Where possible I caught the titles early, but a lot of the time.

This also now handles < and >

## Changes
Replace &amp; with & on PO and Dashboard

## Side effects

This is a hack, so manually making the name contain &amp; will convert it to & in these displays.

## Ticket
[#OSF-6853]
https://openscience.atlassian.net/browse/OSF-6853

## QA Considerations
Test a file name that contains &, <, and >'s
On the Dashboard project widget:
1. Make sure titles in the dashboard handle amps
2. Make sure each level of parent-grandparent handles amps (picture)
screen shot 2016-11-01 at 10 26 54 am
3. Make sure searching matches amp properly (e.g., title = cat&dog, quick search cat&d matches; cat&amp;dog will still match as well iirc)
4. Make sure projects loaded by pressing the 'more' down arrow handle amps

Project Organizer:
1. Make sure titles handle amps
2. Make sure child nodes handle amps
3. Make sure the side bar (information/activity area) handles amps (NOTE: Node logs in Activity do not handle amps)
4. Make sure newly created collections handle amps
5. Make sure collections handle amps on refresh (e.g., previously created ones, not new ones)
6. Make sure renaming collections properly populates the amp
7. Make sure renamed collections handle the amp
8. Check that dragging a project into a collection, the text being dragged handles the amps
9. Check that filtering filters on & not &amp;

I think that is all of the places this changed.


website/static/js/home-page/quickProjectSearchPlugin.js
```diff
@@ -52,6 +52,7 @@ var QuickSearchProject = {
         promise.then(function(result) {
             self.countDisplayed(result.data.length);
             result.data.forEach(function (node) {
+                node.attributes.title = node.attributes.title.replace(/&lt;/g, '<').replace(/&gt;/g, '>').replace(/&amp;/g, '&');
```
brianjgeiger: Seeing it all together like this makes me think we should just have a helper function. This would also make it easier if we find this in other places to keep everything consistent. Plus you'd be able to test the function.

brianjgeiger:
Please merge with develop

website/static/js/osfHelpers.js
```diff
@@ -361,7 +361,12 @@ var mapByProperty = function(list, attr) {
     });
 };
 
-
+/** 
+  * Replaces encoded <, >, and & with decoded values
+  */
+var decodeText = function(text){
+    return text.replace(/&lt;/g, '<').replace(/&gt;/g, '>').replace(/&amp;/g, '&');
```
abought: A note:

In the past, we very intentionally did not decode `<` and `>` on the front end. I'd be happy to walk you through the logic for that in person, but suffice to say it involves an edge case of bleach. Has the rationale for that design decision changed?

alexschiller:
Closing in favor of #6925 which points to master and no longer handles <> per security concerns

PR #6505, opened 2016-11-02
API should not return null log when the component inherits license from parent[OSF-7186]

chennan47:
## Purpose
Make the node api return node.license property instead of node_license field. node.license will return node.license if there is one or its parent's license.

## QA notes
1. create a project
2. create a component in the project
3. add a license to the project
4. view the component's node detail api
Before the change: 
The node_license should show up as "null" and there is no license field in realtionship field, while you can actually see a license on the component when visiting component overview page.
screen shot 2016-11-02 at 10 45 25 am
screen shot 2016-11-02 at 11 29 53 am

After the change:
The node_license should show up as its parent project's node_license , and there is a license field in relationship which links to the same license as its parent.
screen shot 2016-11-02 at 11 19 07 am
screen shot 2016-11-02 at 11 19 35 am

## Ticket

https://openscience.atlassian.net/browse/OSF-7186


brianjgeiger:
@chennan47 Can you rebase this off of COS's feature/staging-2? We're going to try to get this into their next release. Thanks!

brianjgeiger:
Never mind, I figured out what the problem was.

PR #6506, opened 2016-11-02
Block preprint file deletion or move, remove warnings [OSF-7156]

hmoco:
## Purpose
Prevent the deletion of preprints (or making it an orphan preprint) by not allowing the primary file of preprints to be deleted or moved to another folder, component or addon.

## Changes
Remove all the preprint related file operation warnings. As a replacement, when such actions fail, a growl message informs the user that such action is prohibited due to preprint status.

## Side effects
Preprint files can no longer be moved into folders or deleted, from apiV2 or the OSF web client.

## Ticket

https://openscience.atlassian.net/browse/OSF-7156

### DIFF
--- a/api/files/permissions.py
+++ b/api/files/permissions.py
@@ -25,7 +25,7 @@ class IsPreprintFile(permissions.BasePermission):
     def has_object_permission(self, request, view, obj):
         assert isinstance(obj, FileNode), 'obj must be a FileNode, got {}'.format(obj)
 
-        if request.method not in permissions.SAFE_METHODS and len(Node.find(Q('preprint_file', 'eq', obj))):
+        if request.method not in permissions.SAFE_METHODS and len(Node.find(Q('preprint_file', 'eq', obj) & Q('_has_abandoned_preprint', 'eq', False))):
             return False
 
         return True
--- a/api/files/permissions.py
+++ b/api/files/permissions.py
@@ -1,7 +1,9 @@
+from modularodm import Q
 from rest_framework import permissions
 
 from api.base.utils import get_user_auth
 from website.files.models import FileNode
+from website.models import Node
 
 class CheckedOutOrAdmin(permissions.BasePermission):
     def has_object_permission(self, request, view, obj):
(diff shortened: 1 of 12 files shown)


brianjgeiger:
Speaking of tests, looks like there are a couple of failing tests in this version.

hmoco:
@brianjgeiger tests are now there!

brianjgeiger:
👍 If it's ready, please move the Jira ticket back to Code Review

api/files/permissions.py
```diff
@@ -17,3 +19,13 @@ def has_object_permission(self, request, view, obj):
         return obj.checkout is None \
             or obj.checkout == auth.user \
             or obj.node.has_permission(auth.user, 'admin')
+
+
+class IsPreprintFile(permissions.BasePermission):
+    def has_object_permission(self, request, view, obj):
+        assert isinstance(obj, FileNode), 'obj must be a FileNode, got {}'.format(obj)
+
+        if request.method not in permissions.SAFE_METHODS and len(Node.find(Q('preprint_file', 'eq', obj) & Q('_has_abandoned_preprint', 'eq', False))):
```
mfraezz: This could be more efficient by checking the related keys from the object rather performing a query. Something like:

`[...] obj.node.preprint_file == obj and not obj.node._has_abandoned_preprint_file`

website/addons/osfstorage/views.py
```diff
@@ -217,6 +222,10 @@ def osfstorage_delete(file_node, payload, node_addon, **kwargs):
 
     except exceptions.FileNodeCheckedOutError:
         raise HTTPError(httplib.FORBIDDEN)
+    except exceptions.FileNodeIsPrimaryFile:
+        raise HTTPError(httplib.METHOD_NOT_ALLOWED, data={
```
mfraezz: After discussion with @brianjgeiger, this should be a `403`

website/files/models/osfstorage.py
```diff
@@ -107,15 +107,27 @@ def path(self):
...
+            return self._delete_allowed
+        except:
+            return False
+
+    def delete(self, user=None, parent=None):
+        return super(OsfStorageFileNode, self).delete(user=user, parent=parent) if self._delete_allowed else None
 
     def move_under(self, destination_parent, name=None):
+        if self.node.preprint_file == self and not self.node._has_abandoned_preprint:
+            raise exceptions.FileNodeIsPrimaryFile()
```
mfraezz: From ticket:
>If the user attempts to move the file, the action will be a copy action

Does this ensure that behavior? It looks like it might prevent it

PR #6507, opened 2016-11-02
[Feature] Add banner when preprint exists but private node [OSF-7118]

alexschiller:
## Purpose

Preprints that have been made private do not give any indication that they are not discoverable.

## Changes

Add banner to warn user.

## Side effects

None known

## Ticket

https://openscience.atlassian.net/browse/OSF-7118

[#OSF-7118]

## QA Considerations
1. Test that banner does not show up on public projects with preprints
2. Test that banner does show on private projects with a preprint on that node
3. Test that banner does now show up on private project for read-only contributor
4. Test that banner does not show up on private projects without preprints


tests/test_preprints.py
```diff
@@ -140,6 +140,29 @@ def test_non_admin_update_file(self):
             self.preprint.set_primary_file(self.file_two, auth=self.read_write_user_auth, save=True)
         assert_equal(self.project.preprint_file._id, self.file._id)
 
+    def test_public_private_preprint_banner_on_file_exist(self):
```
brianjgeiger: Should this be in webtests?

alexschiller: @brianjgeiger IIRC, I put it here because the webtests didn't have any of the preprint relevant boilerplate needed. I can transfer move it if it belongs there.

brianjgeiger:
Just needs conflicts resolved and it should be ready to go.

brianjgeiger:
Merge conflicts.

alexschiller:
@brianjgeiger For once a reasonable test fail! (failed locally; looks like from django file system change). Dropping to AD and looking at it now.

Issue #6116, opened 2016-08-09
Timestamps exposed by the V2 API do not have timezone information, and may be reasonably interpreted in local time.


emetsger:
See example timestamps in issue #5712; timestamps absent timezone information [may be interpreted](http://www.cl.cam.ac.uk/~mgk25/iso-time.html#zone) in the local timezone.  

Please consider putting timezone information on timestamps, or at least documenting the timezone in the API docs.  Thanks!

brianjgeiger:
This is fixed via https://github.com/CenterForOpenScience/osf.io/pull/6508 - use the API versioning to set the version to 2.2 and you will have normalized timestamps. Note that upgrading your API version will have other effects. See https://osf.io/y9jdt/wiki/Changelog/ for details.

emetsger:
🍗🐔🎉

Issue #5712, opened 2016-05-20
Slight variations on date formats used in OSF /v2/ API


karenhanson:
At least 4 slightly different date formats are used in the OSF API v2, making it harder to parse the dates consistently.  Some examples:

Node:
https://api.osf.io/v2/nodes/6n2re/:
"date_modified": "2016-05-20T20:40:32.206000"

Different node:
https://api.osf.io/v2/nodes/bygau/
"date_modified": "2016-05-23T14:40:15"

File:
https://api.osf.io/v2/files/573f742a594d9001f37be1ee/
"last_touched": "2016-05-20T20:39:26.704000",
"date_modified": "2016-04-15T12:05:01Z",
"date_created": "2016-04-15T12:05:01Z",

Different File:
https://api.osf.io/v2/files/5730aa679ad5a101e8877f18/
"date_modified": "2016-05-10T19:37:49.150Z",
"date_created": "2016-05-09T15:19:03.322Z",

It would be useful if the exact same date format was used for all. Thanks!

emetsger:
It is worth noting that many of the above examples do not have timezone information, when in fact they are UTC.  Some timestamp libraries, when encountering timestamps with no timezone, will interpret them as being in local timezone.

brianjgeiger:
This is fixed via https://github.com/CenterForOpenScience/osf.io/pull/6508 - use the API versioning to set the version to 2.2 and you will have normalized timestamps. Note that upgrading your API version will have other effects. See https://osf.io/y9jdt/wiki/Changelog/ for details.

PR #6508, opened 2016-11-02
[feature] Normalize API Timestamps [OSF-6925]

caseyrollins:
#### Purpose
- Currently, there are slightly different date formats being used in the API (see #5712 for examples). 
- This PR makes it so that when using version 2.2 of the API, all dates will be forced into a consistent ISO-8601 format with timezone information upon serialization 
(i.e. `YYYY-MM-DDTHH:MM:SS.mmmmmmZ`).
- Dates will remain unchanged when using API versions < 2.2, even after django-osf makes all dates non-naive. 

#### Changes
- Adds a `DateByVersion` serializer field which subclasses DRF's `DateTimeField` in order to check the request version and appropriately format the datetime value.
  - Changes all serializer instances of `DateTimeField` to use the `DateByVersion` field.  
- Updates the API to version **2.2**. 

#### QA Notes
- Every date field on every serializer needs to be checked to confirm that:
  - when version=2.2, **ALL** dates match the `YYYY-MM-DDTHH:MM:SS.mmmmmmZ` format
  - when using version < 2.2, dates are unchanged
    - most dates matching either: 
`YYYY-MM-DDTHH:MM:SS` or `YYYY-MM-DDTHH:MM:SS.mmmmmm`
    - FileSerializer `date_modified` and `date_created` dates matching either: 
`YYYY-MM-DDTHH:MM:SSZ` or `YYYY-MM-DDTHH:MM:SS.mmmZ`

#### Side Effects
- None expected.

#### Ticket
- JIRA: [OSF-6925](https://openscience.atlassian.net/browse/OSF-6925)
- GH: #5712 & #6116


brianjgeiger:
:tada::penguin::tada:

PR #6509, opened 2016-11-02
[RFR] [django-osf] File api tests

cwisecarver:
## Purpose

Get file api tests passing.

## Changes

- Update code to use `.all()`, `.last()`, and `.latest()` appropriately.
- Change filtering to chain django filter queries
- Change all references of .pk (integer in django) to ._id, which is what we really want in the url
- Added a `.prefetch_related('guid')` to the MODMCompatibilityQueryset so that we're doing a non-lazy join for the guid. It might actually be worth revisiting this with `.join()` or `.extra()` to do it in one query instead of the multiple queries it does now.

## Side effects

There's a small chance that filtering could be completely broken for things that aren't tags or logs.

## Ticket

https://github.com/CenterForOpenScience/osf-models/issues/133


cwisecarver:
@sloria Assuming the appropriate amount of tests pass this is RFR. There's a function that is now disused that I left in there underneath of one I refactored. I'm happy to pull it, I was mainly leaving it there for ease of comparison with the current version and the past one.

addons/osfstorage/models.py
```diff
@@ -294,6 +294,9 @@ def remove_tag(self, tag, auth, save=True, log=True):
         if self.node.is_registration:
             # Can't perform edits on a registration
             raise NodeStateError
+        # TODO does modm let you do this???!
```
sloria: We have tags stored as tuples?

api/base/filters.py
```diff
@@ -433,14 +433,15 @@ def get_queryset_from_request(self):
...
-
+        queryset = default_queryset
         if filters:
             for key, field_names in filters.iteritems():
                 for field_name, data in field_names.iteritems():
-                    queryset = queryset.intersection(set(self.get_filtered_queryset(field_name, data, default_queryset)))
-
-        return list(queryset)
+                    query_field_name = data['source_field_name']
+                    if query_field_name == 'tags':
```
sloria: I added a hook for this a while back, `postprocess_query_param`, for this type of use case. See https://github.com/CenterForOpenScience/osf.io/blob/be40da61bb738b9c81d7b4d0f3cc2a81abff2463/api/nodes/views.py#L302-L311 for an example. Can you use that instead?

osf/models/node.py
```diff
@@ -735,6 +723,20 @@ def comment_level(self, value):
     def get_absolute_url(self):
         return self.absolute_api_v2_url
 
+    @property
+    def permissions(self):
```
sloria: We shouldn't need this anymore: we already have `AbstractNode#has_permission`, `AbstractNode#get_permissions`, and `AbstractNode#is_contributor`. Was this added to make checks for `user in node.permissions` work? If so, can you change those to use `node.is_contributor(user)` instead?

addons/osfstorage/models.py
cwisecarver: No, I'm stupid. I was getting a `.values_list('name')` instead of `.values_list('name', flat=True)`.

Fixed.

cwisecarver:
@sloria I made the changes you'd asked for and then forgot to merge. This is gtg now. Rebased on django-osf just now.

PR #6511, opened 2016-11-03
[Hotfix] Preprint script fixes

mfraezz:
## Purpose
Correct small issues with original script

## Changes
* Upsert documents that might not be there
* `.reload` to prevent saving stale data
* set `date_modified`

## Side effects
None.

### DIFF
--- a/scripts/migration/migrate_node_preprints_to_preprintservices.py
+++ b/scripts/migration/migrate_node_preprints_to_preprintservices.py
@@ -103,6 +103,8 @@ def create_preprint_service_from_node(document, swap_cutoff):
                     'preprint_created': ''
                 }}
             )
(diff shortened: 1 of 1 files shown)


scripts/migration/migrate_node_preprints_to_preprintservices.py
```diff
@@ -103,6 +103,7 @@ def create_preprint_service_from_node(document, swap_cutoff):
                     'preprint_created': ''
                 }}
             )
+            node.reload()
```
icereval: preprint.reload() - https://github.com/CenterForOpenScience/osf.io/pull/6511/files#diff-a5cc311950e11a78fabbfc8ce6fe201fR80

PR #6514, opened 2016-11-03
[Hotfix] Set MyProjects preprints routes [OSF-7188]

alexschiller:
## Purpose

After the migration, guid routes changed for preprints. This PR fixes makes the preprints collection tab grab the working route to the preprint.

## Changes

Pull in preprint data to the nodes that have a preprint. Populate the route with that data.

## Side effects

Not a direct side effect of this PR, but a possible further improvement/bug:
Currently each line of the my projects page is populated from a single node that has a preprint. This does not check if the node has multiple preprints (e.g., more than one provider that has a preprint file). This means that only the first preprint value will be displayed here. It seemed sufficiently out of scope of this ticket to attempt a fix at that.

## Ticket
[#OSF-7188]

https://openscience.atlassian.net/browse/OSF-7188

## QA Considerations
1. Make sure that the link created properly goes to the preprint page (and not the node). 
2. Test both newly created preprints and migrated preprints of the various guid translation types (project kept the guid vs preprint kept the guid) if possible.


website/static/js/project-organizer.js
```diff
@@ -36,12 +36,11 @@ function _poTitleColumn(item) {
     };
     var node = item.data; // Where actual data of the node is
     var css = ''; // Keep for future expandability -- Remove: item.data.isSmartFolder ? 'project-smart-folder smart-folder' : '';
-    var preprintLinkPre = '/preprints/';
     var isMypreprintsCollection = tb.options.currentView().collection.data.nodeType === 'preprints';
     if (item.data.archiving) { // TODO check if this variable will be available
         return  m('span', {'class': 'registration-archiving'}, node.attributes.title + ' [Archiving]');
     } else if (node.attributes.preprint && isMypreprintsCollection){
-        return [ m('a.fg-file-links', { 'class' : css, href : preprintLinkPre + node.id, 'data-nodeID' : node.id, 'data-nodeTitle': node.attributes.title,'data-nodeType': node.type, onclick : function(event) {
```
brianjgeiger: This is what people get when they build their own urls.

icereval:
Is this still needed?

alexschiller:
@icereval yes.

alexschiller:
Closing in favor of #6748

PR #6519, opened 2016-11-04
Disallow changing primary preprint file after publishing [#OSF-7190]

erinspace:
## Purpose
You may no longer entirely change the primary file of a preprint, only update the version by directly dealing with the file stored on the node. Therefore, don't let the API change the primary file of the preprint either!

## Changes
- Update preprint detail docstring
- Raise validation error in serializer if the preprint file is updated
- Update test

## Side effects
QA Automated runscope tests trying to update preprint files might fail if there are any!

## Ticket
https://openscience.atlassian.net/browse/OSF-7190

### DIFF
--- a/api/preprints/serializers.py
+++ b/api/preprints/serializers.py
@@ -112,8 +112,9 @@ class PreprintSerializer(JSONAPISerializer):
 
         primary_file = validated_data.pop('primary_file', None)
         if primary_file:
-            self.set_field(preprint.set_primary_file, primary_file, auth)
-            save_node = True
+            raise exceptions.ValidationError(
+                detail='You may not change the primary file of a preprint. Please update the version of the file instead.'
+            )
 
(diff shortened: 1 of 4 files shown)


api/preprints/serializers.py
```diff
@@ -112,8 +112,9 @@ def update(self, preprint, validated_data):
 
         primary_file = validated_data.pop('primary_file', None)
         if primary_file:
-            self.set_field(preprint.set_primary_file, primary_file, auth)
-            save_node = True
+            raise exceptions.ValidationError(
```
mfraezz: With Edit coming out soon, it may be possible for a user to start a preprint and decide that they uploaded the wrong file before publishing -- this would prevent them from correcting that. 

Instead, could you add a check for `is_published` in `PreprintService.set_primary_file`?

PR #6522, opened 2016-11-04
[djangosf] Forward Addon

mfraezz:
## Purpose
Add Forward addon to DjangOSF

## Changes
* Check templates for existence before determining their `dirname`
* Move `before_save` validation to `.clean`

## Side effects
None expected.

## Ticket
None known.


framework/mongo/utils.py
```diff
@@ -39,7 +40,8 @@ def from_mongo(item):
 sanitize_pattern = re.compile(r'<\/?[^>]+>')
 def sanitized(value):
     if value != sanitize_pattern.sub('', value):
-        raise ValidationValueError('Unsanitary string')
+        raise ValidationError('Unsanitary string')
+    return True
 
 
 def unique_on(*groups):
```
cwisecarver: uhhh. mongo doesn't work in postgres.

website/addons/forward/views/config.py
```diff
@@ -43,12 +43,13 @@ def forward_config_put(auth, node_addon, **kwargs):
 
     # Save settings and get changed fields; crash if validation fails
     try:
-        saved_fields = node_addon.save()
+        dirty_fields = node_addon.get_dirty_fields()
+        node_addon.save()
     except ValidationError:
```
cwisecarver: Make sure this is the right validationerror

framework/mongo/utils.py
mfraezz: Not used here, this should be cleaned up later

PR #6523, opened 2016-11-04
Add provider specific docs in NodeFilesList view [OSF-4731]

erinspace:
## Purpose
Different addon storage providers reliably don't return values for some file attributes, and that's not documented in the API at the moment. So add those!

## Changes
- Add a " Provider Specific File Attributes and Quirks" section to `NodeFilesList` view docstring

screen shot 2016-11-04 at 4 14 26 pm

## Side effects
None anticipated, but there might be some file attributes that I missed. These are compiled  from the ticket (and verified locally), and some were gathered from notes in Waterbutler.

## Ticket
https://openscience.atlassian.net/browse/OSF-4731

### DIFF
--- a/api/nodes/views.py
+++ b/api/nodes/views.py
@@ -1789,9 +1789,12 @@ class NodeFilesList(JSONAPIBaseView, generics.ListAPIView, WaterButlerMixin, Lis
     ###ownCloud
     ownCloud returns `null` for the following fields:
 
+    - `date_modified`
+    - `date_created`
     - hashes
         - `md5`
         - `sha256`
+    - `size`
 
     ##Links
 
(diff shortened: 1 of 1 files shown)


brianjgeiger:
@felliott Is there a source for the expected box and owncloud differences?

felliott:
Sadly, nothing that summarizes like this.  Best bet is to check the provider metadata classes in WB.

Box: https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/providers/box/metadata.py
ownCloud: https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/providers/owncloud/metadata.py

PR #6525, opened 2016-11-04
[Feature] Preprint fangorn delete ui and backend update[OSF-7157]

alexschiller:
## Purpose
Preprints are no longer deletable or moveable cross node. These changes update the Fangorn File widget UI and backend to fully match this spec.

## Changes
1. This PR contains changes from https://github.com/CenterForOpenScience/osf.io/pull/6490 . These changes were included as 6490 included the testing file and major cleanup of the move/copy/forbid functionality. Those changes will need to be tested in addition to the other changes mentioned here.
2. Preprints can move into and out of folders within their parent node.
3. Attempts to move preprints outside of the current node will revert to a copy
4. Attempting to move a preprint in a folder that contains a preprint (loaded) will copy that folder and preprint 
5. Attempting to move a preprint in a folder that contains a preprint (unloaded) will block the action and growl a move error (bad ui, update next)
6. Clicking on a preprint will show a disabled delete button that shows a tooltip explaining why it is disabled
7. Clicking on a folder that contains a preprint (loaded) will show a disabled delete button and tooltip
8. Clicking on an folder that contains a preprint (unloaded) will show a normal delete button, but attempts to delete will growl an error (bad ui, update next)
9. Multiselecting a preprint will disable the delete button and show a tooltip
10. Multiselecting a folder that contains a preprint (loaded) will disable the delete button
11. Multiselecting a folder that contains a preprint (not loaded) will show the button but growl and fail.
12. All other actions should be the same (e.g., renaming, etc)
13. Delete logic fangorn tests added

## Side effects
Technically all the things in https://github.com/CenterForOpenScience/osf.io/pull/6490 are side effects since they needed to be included to complete the current work.

## Ticket
[#OSF-7157]

https://openscience.atlassian.net/browse/OSF-7157

## QA Considerations
Dragging between folders and nodes needs to be tested thoroughly that no clear routes to orphaning the preprint exist.


website/files/models/osfstorage.py
```diff
@@ -307,6 +312,14 @@ def is_checked_out(self):
                 return True
         return False
 
+    @property
+    def is_preprint_primary(self):
```
brianjgeiger: This should be more like "contains_undeletable_item()" and should check for any reason that it's not deletable.

website/static/js/fangorn.js
```diff
@@ -1742,12 +1743,22 @@ var FGItemButtons = {
...
-                        m.component(FGButton, {
-                            onclick: function(event) {_removeEvent.call(tb, event, [item]); },
-                            icon: 'fa fa-trash',
-                            className : 'text-danger'
-                        }, 'Delete Folder'));
+                    if (preprintPath && folderContainsPreprint(item, preprintPath)) {
+                        rowButtons.push(
+                            m.component(FGButton, {
+                                icon: 'fa fa-trash',
+                                tooltip: 'This folder contains a Preprint. You cannot delete Preprints, but you can upload a new version.',
```
brianjgeiger: This should be more vague. "This folder contains an un-deletable item such as a primary Preprint file or a checked-out file. You cannot delete blah blah"

website/static/js/fangorn.js
```diff
@@ -1974,31 +1997,32 @@ var FGToolbar = {
...
-                if (each.data.permissions.edit && !each.data.isAddonRoot && !each.data.nodeType) {
-                    showDelete = true;
-                    break;
+            if (showDeleteMultiple(items)) {
+                var preprintPath = getPreprintPath(window.contextVars.node.isPreprint, window.contextVars.node.preprintFileId);
+                if (preprintPath && multiselectContainsPreprint(items, preprintPath)) {
+                    generalButtons.push(
+                        m.component(FGButton, {
+                            icon: 'fa fa-trash',
+                            tooltip: 'One of these files is a Preprint. You cannot delete Preprints, but you can upload a new version.',
```
brianjgeiger: If this also covers folders, then it should have the vague folder method. Also if it has a checked out file, it should all be handled here, right? Try to make this super-generic for any reason why a folder/file would not be deletable.

website/static/js/tests/fangorn.test.js
```diff
@@ -0,0 +1,404 @@
...
+            });
+
+            it('valid copy drop', () => {
+                folder = getItem('folder', 2);
+                item = getItem('file', 3);
+                item.data.provider = 'github';
+                assert.equal(Fangorn.getCopyMode(folder, [item]), 'copy');
+            });
+
+            it('invalid drop if folder.data undefined', () => {
```
brianjgeiger: So ideally these should be named more like it('cannot be dropped if folder.data is undefined'). That being said, I wouldn't hold up the release for it.

PR #6527, opened 2016-11-07
Allow node title to be shown for public node logs [OSF-7192]

erinspace:
## Purpose
Accessing public node logs as anonymous user via the API prompted a permissions check when getting the title, which would fail! 

## Changes
- Display the node title if the node is public
- Two unrelated changes stumbled upon:
    - fix rendering in node log docstring
    - remove `expect_errors` from a test expecting a 200 success

## Side effects
None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-7192


erinspace:
Is now fixed, no longer needed!

PR #6528, opened 2016-11-07
[djangosf] Citation Addons

mfraezz:
## Purpose
Add Mendeley/Zotero to DjangOSF

## Changes

## Side effects
None expected

## Ticket
None known
```
~/D/osf.io ❯ DJANGO_SETTINGS_MODULE=api.base.settings py.test addons/ -s
[...]
=============== 975 passed, 1 skipped in 1857.37 seconds ===============
```


addons/mendeley/models.py
```diff
@@ -0,0 +1,270 @@
...
+import time
+
+from django.db import models
+import mendeley
+from mendeley.exception import MendeleyApiException
+from oauthlib.oauth2 import InvalidGrantError
+
+from addons.base.models import BaseOAuthUserSettings, BaseCitationsNodeSettings
+from website.addons.mendeley.serializer import MendeleySerializer
+from website.addons.mendeley import settings
```
cwisecarver: I've been trying to move the settings from their settings files into the app config. It can probably wait since we can't totally get rid of the old add-on anyway yet. Maybe add a todo?

website/citations/providers.py
```diff
@@ -138,7 +138,7 @@ def set_config(self, node_addon, user, external_list_id, external_list_name, aut
     def add_user_auth(self, node_addon, user, external_account_id):
         """Adds authorization to a node
         if the user has authorization to grant"""
-        external_account = ExternalAccount.load(external_account_id)
+        external_account = ExternalAccount.objects.get(_id=external_account_id)
```
cwisecarver: This will throw `ExternalAccount.DoesNotExist()` if it's not there. MODM's load didn't used to do that. If you're looking for the same behavior you can do `.filter().first()` and the retval will be `None` if the criteria isn't met. (h/t @chrisseto This is also slower because it implies a sort but ... we can fix that later.)

mfraezz: Ah, right. I'll just have it use `BaseModel.load` for the same behavior

PR #6535, opened 2016-11-08
Filter URL serialization in Relationship Field [#OSF-7194]

erinspace:
## Purpose
Filters added to the query params by the `RelationshipSeralizer` were using the normal way to add arguments, with `filter` as the name and the entire filter as the value: `filter=[field_name]=example`. This is done using the `extend_querystring_params` function which adds query string params with `furl`. So, pass the right thing to `extend_querystring_params`, and make sure it works with multiple filters!

### Before
screen shot 2016-11-08 at 1 27 59 pm

### After
screen shot 2016-11-08 at 1 27 29 pm

## Changes
- Update `format_filters` method so that it returns a list of dicts with checked filter params, instead of a string with all the checked filters bundled together
- Add each of those checked filters to the url via `extend_querystring_params`
- Test to make sure adding multiple filters via the `RelationshipSerializer` looks as we expect

## Side effects
Any use of `format_filters` (couldn't find any others) will need to expect a list of dicts with formatted values instead of a pre-formatted string.

## Ticket
https://openscience.atlassian.net/browse/OSF-7194

### DIFF
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -4,7 +4,6 @@ import re
 import furl
 from django.core.urlresolvers import resolve, reverse, NoReverseMatch
 from django.core.exceptions import ImproperlyConfigured
-from django.http.request import QueryDict
 from rest_framework import exceptions
 from rest_framework import serializers as ser
 from rest_framework.fields import SkipField
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -583,6 +583,7 @@ class RelationshipField(ser.HyperlinkedIdentityField):
                         formatted_filter = self.format_filter(obj)
                         if formatted_filter:
                             url = extend_querystring_params(url, {'filter': formatted_filter})
+                            url = url.replace('?filter=', '?filter')
                         else:
                             url = None
 
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -580,10 +580,13 @@ class RelationshipField(ser.HyperlinkedIdentityField):
                     kwargs.update({'version': request.parser_context['kwargs']['version']})
                     url = self.reverse(view, kwargs=kwargs, request=request, format=format)
                     if self.filter:
-                        formatted_filter = self.format_filter(obj)
-                        if formatted_filter:
-                            url = extend_querystring_params(url, {'filter': formatted_filter})
-                            url = url.replace('?filter=', '?filter')
+                        formatted_filters = self.format_filter(obj)
+                        if formatted_filters:
+                            for filter in formatted_filters:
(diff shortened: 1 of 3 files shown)


api_tests/base/test_serializers.py
```diff
@@ -310,6 +317,15 @@ def test_field_with_two_kwargs(self):
         field = data['relationships']['two_url_kwargs']['links']
         assert_in('/v2/nodes/{}/node_links/{}/'.format(node._id, node._id), field['related']['href'])
 
+    def test_field_with_two_filters(self):
+        req = make_drf_request_with_version(version='2.0')
+        project = factories.ProjectFactory()
+        node = factories.NodeFactory(parent=project)
+        data = self.BasicNodeSerializer(node, context={'request': req}).data['data']
+        field = data['relationships']['field_with_filters']['links']
+        assert_in(urllib.quote('filter[target]=hello', safe='?='), field['related']['href'])
```
brianjgeiger: So `assert_in('filter[target]=hello', field['related']['href'])` doesn't work? What happens?

api/base/serializers.py
```diff
@@ -566,9 +565,13 @@ def get_url(self, obj, view_name, request, format):
...
                     if self.filter:
-                        formatted_filter = self.format_filter(obj)
-                        if formatted_filter:
-                            url = extend_querystring_params(url, {'filter': formatted_filter})
+                        formatted_filters = self.format_filter(obj)
+                        if formatted_filters:
+                            for filter in formatted_filters:
+                                url = extend_querystring_params(
+                                    url,
+                                    {'filter[{}]'.format(filter['field_name']): filter['value']}
```
brianjgeiger: Very nice. Not terrible at all.

brianjgeiger:
Oh could you add sample json for before and after to the PR Description, please?

api_tests/base/test_serializers.py
erinspace: `AssertionError: 'filter[target]=hello' not found in 'http://localhost:8000/v2/nodes/7yy4n/?filter%5Btarget%5D=hello&filter%5Bwoop%5D=yea'`

brianjgeiger: Ah, right.

api/base/serializers.py
```diff
@@ -596,8 +599,8 @@ def to_esi_representation(self, value, envelope='data'):
                 return '<esi:include src="{}"/>'.format(esi_url)
 
     def format_filter(self, obj):
```
brianjgeiger: Could you do a docstring for format_filter so people know what is expected here?

PR #6536, opened 2016-11-08
[WIP] Comment detail tests

cwisecarver:

### DIFF
--- a/addons/osfstorage/models.py
+++ b/addons/osfstorage/models.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 import logging
 import os
 
@@ -365,6 +367,7 @@ class NodeSettings(BaseStorageAddon, BaseNodeSettings):
         if self.root_node:
             return
 
+
         # A save is required here to both create and attach the root_node
         # When on_add is called the model that self refers to does not yet exist
         # in the database and thus odm cannot attach foreign fields to it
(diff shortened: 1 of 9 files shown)


addons/osfstorage/models.py
```diff
@@ -365,6 +367,7 @@ def on_add(self):
         if self.root_node:
```
cwisecarver: @sloria, here's the problem. Modm would have nulled out `root_node` upon clone. So, getting here, `self.root_node` would have been falsely. So it would have made a new one based off the `self.owner` of the `NodeSettings`, which would have correctly been set to the `Registration` in `after_register()`, seen below. 

I was thinking about comparing the types of `self.owner` and `self.root_node` and then nulling out `root_node` if the types mismatched, but I don't know if that's sufficient.

@chrisseto, other than nulling out foreign fields on create, which @sloria was hoping to avoid, can you think of a better way to do this?

api_tests/comments/views/test_comment_detail.py
```diff
@@ -147,6 +148,7 @@ def test_comment_has_node_link(self):
         assert_equal(urlparse(url).path, expected_url)
 
     def test_registration_comment_has_node_link(self):
+        import ipdb;ipdb.set_trace()
```
cwisecarver: @sloria This is the actual test that's failing with, well, before I broke it all, something along the lines of `'/v2/nodes/123ab/' != '/v2/registrations/123ab/'`.

PR #6553, opened 2016-11-11
[hotfix] Add preprints search and view to admin [OSF-7198]

erinspace:
## Purpose
Since preprints have their own GUIDs now, you can't currently view them in the admin module. Let's fix that!

This somewhat limited preprint view shows  preprint specific information, but then encourages the user to head to the node to perform operations such as deleting users. Thought it best to do this to avoid repetition, since most operations happen on the node anyhow.

screen shot 2016-11-15 at 5 04 18 pm

#### Change preprint provider

screen shot 2017-04-03 at 3 59 31 pm

## Changes
- Add preprint view, serializer, search, templates, and links to these in the sidebar
- bonus, fix `create_fakes` to work for preprint generation
- **Now updated to work with new Admin app and permissions!**

## Side effects
None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-7198


admin/preprints/serializers.py
```diff
@@ -0,0 +1,29 @@
...
+        'is_published': preprint.is_published,
+        'date_published': preprint.date_published,
+        'subjects': serialize_subjects(preprint.subjects),
+    }
+
+
+def serialize_subjects(subjects):
+    serialized_subjects = []
+    for subject in subjects:
+        if len(subject) == 1:
```
brianjgeiger: What's going on here? Is there a mess on the back end causing this?

erinspace: They were stored this way for me, not entirely sure if it was an artifact of how my preprints were created or an oddity in general. I can try to make preprints again through the API and see if this persists.

erinspace: Yep that was totally weird. @brianjgeiger fixed now!

admin/templates/preprints/preprint.html
```diff
@@ -8,138 +8,160 @@
...
+                        <td>Public</td>
+                        <td>{{ serialized_preprint.node.is_public }}</td>
+                    </tr>
+                    <tr>
+                        <td>Provider</td>
+                        <td>
+                            {{ serialized_preprint.provider.name }}
+                            <span class="provider_form_link" style="margin-left: 100px;">
+                                <a class="btn btn-link" role="button" data-toggle="collapse" href="#collapseChangeProvider">
+                                    Change prepint provider
```
brianjgeiger: How are the permissions for this handled? There's a separate permission for editing than for viewing, yes?

erinspace: Yes! Good call! Missed that one. Will update!

sloria:
@erinspace This should be targeted towards develop, yes?

admin/preprints/views.py
```diff
@@ -46,6 +46,7 @@ def get_object(self, queryset=None):
 
     def get_context_data(self, **kwargs):
         preprint = PreprintService.load(self.kwargs.get('guid'))
+        # TODO - we shouldn't need this serialized_preprint value -- https://openscience.atlassian.net/browse/OSF-7743
```
sloria: 👍

erinspace:
@sloria -- @brianjgeiger asked me to make this a hotfix

sloria:
@erinspace OK, looks like this was branched off of develop though. Please create a different branch off of master if this is to be a hotfix.

erinspace:
Goooooooood call.

Closing in favor of #7064!

PR #6557, opened 2016-11-14
[feature] Remove and Redirect "Explore" Page [OSF-7166]

caseyrollins:
#### Purpose
- The `/explore/` page on the OSF contains a static graph that is no longer being updated and is never referenced on the OSF. This PR removes that page, and adds necessary redirects to the "Public Activity" page.

#### Changes
- Adds an `/activity/` route that goes to the "Public Activity" page
- Redirects `/explore/` -> `/activity/`
- Redirects `/explore/activity` -> `/activity/`
- Links the "New Projects" selection from the "Browse" dropdown in the navbar to `/activity/` instead of `/explore/activity/` to avoid unnecessary redirects.
- Removes the now obsolete `explore.mako`

#### Ticket
- [OSF-7166](https://openscience.atlassian.net/browse/OSF-7166)


brianjgeiger:
@caseyrollins Merge conflicts. Can go to Passed CR when fixed.

PR #6560, opened 2016-11-15
[Feature Fix] Fix fangorn tests and minor update [OSF-6319]

alexschiller:
## Purpose
Fangorn tests needed to have updated names and added minor fixes along the way.

## Changes

Rename fangorn tests

## Side effects

1. Add check for no rows on multiselect
2. Disable all fangorn figshare item drag outs

## Ticket
[#OSF-6319]
https://openscience.atlassian.net/browse/OSF-6319

## QA Notes
1. Test that you ctrl-clicking when you have only a single item selected, and you select the time you are already on does not throw an error in the console
2. Check that attempting to drag figshare items (both published and draft) out of figshare shows disabled the drops

### DIFF
--- a/website/static/js/tests/fangorn.test.js
+++ b/website/static/js/tests/fangorn.test.js
@@ -238,20 +238,20 @@ describe('fangorn', () => {
                 assert.equal(Fangorn.isInvalidDropItem(folder, item, false, true), true);
             });
 
-            it('not a valid drop if item provider is figshare and private', () => {
+            it('cannot be dropped if item provider is figshare and private', () => {
                 folder = getItem('folder', 2);
                 item = getItem('folder', 3);
                 item.data.provider = 'figshare';
                 item.data.extra = {'status' : 'private'};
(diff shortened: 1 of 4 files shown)


website/static/js/tests/fangorn.test.js
```diff
@@ -376,23 +369,23 @@ describe('fangorn', () => {
...
+            it('do not show multi delete if no edit permissions', () => {
                 folder = getItem('folder', 2);
                 folder.data.permissions.edit = false;
                 item = getItem('file', 3);
                 item.data.permissions.edit = false;
                 assert.equal(Fangorn.showDeleteMultiple([folder, item]), false);
             });
 
-            it('one edit permissions returns true', () => {
+            it('do show multi delete if edit permissions for at least one selected', () => {
```
brianjgeiger: 'It does show multi delete' rather than 'it do show multi delete'. Also the similar ones.

PR #6562, opened 2016-11-15
Add more analytics to Keen! [OSF-7206]

erinspace:
## Purpose
We did a cool first round of sending stats to Keen, but let's make that more awesome by making it more closely match the other stats that we've been gathering regarding projects and nodes and addons and whatnot.

Here are what the new keen analytics events look like:

### User domain events
```
[
    {
        "date": "2016-11-13T09:21:13.988000+00:00",
        "keen": {
            "timestamp": "2016-11-13T09:21:13.988000+00:00"
        },
        "domain": "schuppe.com"
    },
    {
        "date": "2016-11-13T09:21:13.988000+00:00",
        "keen": {
            "timestamp": "2016-11-13T09:21:13.988000+00:00"
        },
        "domain": "mail.com"
    }
]
```

### Addon snapshot
```
{
    "users": {
        "enabled": 5,
        "authorized": 3,
        "linked": 2
    },
    "nodes": {
        "deleted": 0,
        "total": 2,
        "connected": 2,
        "disconnected": 0
    },
    "provider": {
        "name": "github"
    }
}
```

### User summary
```
{
    "status": {
        "active": 5,
        "deactivated": 2,
        "depth": 1,
        "unconfirmed": 1
    },
    "keen": {
        "timestamp": "2016-11-14T00:00:00+00:00"
    }
}
```

## Changes
- Add a `yesterday` argument to the main parsers that will automatically harvest from the day before when the script was called
- Use `yesterday` kwarg when calling from celery
- Add new `user_domain_events` collection that sends an event for every user created in a 24 hour period with what email domain they used for their username (e.x a user who signed up with my_awesome_email@woohoo.com would send an event for the domain woohoo.com)
- For addon snapshot, add `enabled`, `authorized` and `linked` instead of just `total` for `users`, which mirrors current analytics. This captures the number of users who have addons in their different potential stages of use (clicking the checkbox vs actually authorizing with the app vs actually linking it successfully to a node), and might help track improvements in usability.
- For the user summary, add depth users (using the same threshold of those with 11+ logs as we're currently using)
- Add tests for this stuff!

## Side effects
- Celery task updated! Will it have to be loaded separately?
- Addon Snapshot analytics are different for the user - remove `total` and report instead on `enabled`, `authorized` and `linked` -- enabled now is essentially the same as total. Chose to replace the language to match the analytics we were gathering in the past.

## Ticket
https://openscience.atlassian.net/browse/OSF-7206

### DIFF
--- a/scripts/analytics/institution_summary.py
+++ b/scripts/analytics/institution_summary.py
@@ -111,9 +111,9 @@ if __name__ == '__main__':
     institution_summary = InstitutionSummary()
     args = institution_summary.parse_args()
     yesterday = args.yesterday
-    if not yesterday:
-        date = parse(args.date).date() if args.date else None
-    else:
+    if yesterday:
         date = (datetime.today() - timedelta(1)).date()
+    else:
+        date = parse(args.date).date() if args.date else None
     events = institution_summary.get_events(date)
     institution_summary.send_events(events)
--- a/scripts/analytics/institution_summary.py
+++ b/scripts/analytics/institution_summary.py
@@ -110,6 +110,10 @@ if __name__ == '__main__':
     init_app()
     institution_summary = InstitutionSummary()
     args = institution_summary.parse_args()
-    date = parse(args.date).date() if args.date else None
+    yesterday = args.yesterday
+    if not yesterday:
+        date = parse(args.date).date() if args.date else None
+    else:
+        date = (datetime.today() - timedelta(1)).date()
     events = institution_summary.get_events(date)
     institution_summary.send_events(events)
--- a/scripts/analytics/node_summary.py
+++ b/scripts/analytics/node_summary.py
@@ -99,9 +99,9 @@ if __name__ == '__main__':
     node_summary = NodeSummary()
     args = node_summary.parse_args()
     yesterday = args.yesterday
-    if not yesterday:
-        date = parse(args.date).date() if args.date else None
-    else:
+    if yesterday:
         date = (datetime.today() - timedelta(1)).date()
+    else:
+        date = parse(args.date).date() if args.date else None
     events = node_summary.get_events(date)
     node_summary.send_events(events)
--- a/scripts/analytics/node_summary.py
+++ b/scripts/analytics/node_summary.py
@@ -98,6 +98,10 @@ if __name__ == '__main__':
     init_app()
     node_summary = NodeSummary()
     args = node_summary.parse_args()
-    date = parse(args.date).date() if args.date else None
+    yesterday = args.yesterday
+    if not yesterday:
+        date = parse(args.date).date() if args.date else None
+    else:
+        date = (datetime.today() - timedelta(1)).date()
     events = node_summary.get_events(date)
     node_summary.send_events(events)
--- a/scripts/analytics/user_summary.py
+++ b/scripts/analytics/user_summary.py
@@ -98,9 +98,9 @@ if __name__ == '__main__':
     user_summary = UserSummary()
     args = user_summary.parse_args()
     yesterday = args.yesterday
-    if not yesterday:
-        date = parse(args.date).date() if args.date else None
-    else:
+    if yesterday:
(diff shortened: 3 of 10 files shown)


scripts/analytics/institution_summary.py
```diff
@@ -110,6 +110,10 @@ def get_class():
     init_app()
     institution_summary = InstitutionSummary()
     args = institution_summary.parse_args()
-    date = parse(args.date).date() if args.date else None
+    yesterday = args.yesterday
+    if not yesterday:
```
brianjgeiger: Probably should be `if yesterday` and swap the conditions. That way if we wanted to put any more special conditions in, it would just be `elif`s.

scripts/analytics/user_summary.py
```diff
@@ -6,12 +6,36 @@
...
 logging.basicConfig(level=logging.INFO)
 
+LOG_THRESHOLD = 11
+
+
+# Modified from scripts/analytics/depth_users.py
+def count_user_logs(user):
+    logs = NodeLog.find(Q('user', 'eq', user._id))
+    length = logs.count()
+    if length > 0:
```
brianjgeiger: This should be `if length == LOG_THRESHOLD`

scripts/analytics/user_summary.py
```diff
@@ -50,8 +79,9 @@ def get_events(self, date):
             }
         }
         logger.info(
-            'Users counted. Active: {}, Unconfirmed: {}, Deactivated: {}'.format(
+            'Users counted. Active: {}, Depth: {}, Unconfirmed: {}, Deactivated: {}'.format(
                 counts['status']['active'],
+                counts['status']['depth'],
                 counts['status']['unconfirmed'],
```
brianjgeiger: Let's throw in a `merged_users` count since we're here, 'cause why not?

brianjgeiger: Unless it looks like more than 30 minutes work including tests, then maybe not.

scripts/analytics/user_summary.py
```diff
@@ -67,6 +97,10 @@ def get_class():
     init_app()
     user_summary = UserSummary()
     args = user_summary.parse_args()
-    date = parse(args.date).date() if args.date else None
+    yesterday = args.yesterday
+    if not yesterday:
```
brianjgeiger: As above: swap.

scripts/analytics/node_summary.py
```diff
@@ -98,6 +98,10 @@ def get_class():
     init_app()
     node_summary = NodeSummary()
     args = node_summary.parse_args()
-    date = parse(args.date).date() if args.date else None
+    yesterday = args.yesterday
+    if not yesterday:
```
brianjgeiger: As above: swap.

scripts/analytics/user_summary.py
```diff
@@ -6,12 +6,36 @@
...
+    length = logs.count()
+    if length > 0:
+        item = logs[0]
+        if item.action == 'project_created' and item.node.is_bookmark_collection:
+            length -= 1
+    return length
+
+
+# Modified from scripts/analytics/depth_users.py
+def get_number_of_depth_users(active_users):
```
brianjgeiger: For the analytics requirements ticket you have on the side, make sure there's something in the requirements that ensures that it will work for user analytics as well. Iterating over 40k users every night is less than ideal, hence my uncharacteristic optimization above. Once we have the caching, we can eliminate most of the work from this task.

scripts/analytics/user_summary.py
```diff
@@ -19,7 +19,7 @@
 def count_user_logs(user):
     logs = NodeLog.find(Q('user', 'eq', user._id))
     length = logs.count()
-    if length > 0:
+    if length >= LOG_THRESHOLD:
```
brianjgeiger: No, just `==`. If it's greater than, subtracting one won't make a difference to the function that uses it, so we only need to do the check under one very specific instance.

PR #6564, opened 2016-11-16
[Feature] Add str value conversion to encrypt and decrypt [OSF-7200]

alexschiller:
## Purpose

Names on Dropbox and Google drive break when we try to encrypt them.

## Changes

1. Add helper function to ensure incoming strings to be encoded are in the proper utf-8 format
2. Add tests for helper function and encryption in general since it was untested.

## Side effects

None known

## Ticket
https://openscience.atlassian.net/browse/OSF-7200

[#OSF-7200]

## QA Considerations
1. Test auth on at least 1 addon. I did dropbox locally and it worked. Per MattF if it works for one, there is no reason it shouldn't work for all of them. Worth testing at least one or two for a sanity check, but probably not worth checking all of them (especially if you can't easily change your name on the service).
2. Change your name on <addon> to contain unicode madness and try to authorize the addon on the osf
3. Ensure it also works with non-unicode madness


framework/encryption/__init__.py
```diff
@@ -6,13 +6,22 @@
...
+
 def encrypt(value):
     if value:
+        value = ensure_str(value)
         return jwe.encrypt(bytes(value), SENSITIVE_DATA_KEY)
     return None
 
 def decrypt(value):
     if value:
+        value = ensure_str(value)
```
mfraezz: Here, `value` is going to be an encrypted string of the form
```
r'\w{43}\.\.\w{22}\.\w*\.\w{22}'
```
I'm not sure that this needs to be called here.

tests/framework_tests/test_encryption.py
```diff
@@ -0,0 +1,74 @@
...
+        assert_true(isinstance(my_value_encrypted, str))
+
+        my_value_decrypted = decrypt(my_value_encrypted)
+        assert_equal(my_value_decrypted, ensure_str(my_value))
+
+        my_value = '찦차КЛМНО💁◕‿◕｡)╱i̲̬͇̪͙n̝̗͕v̟̜̘̦͟o̶̙̰̠kè͚̮̺̪̹̱̤  ǝɹol'
+        my_value_encrypted = encrypt(my_value)
+        my_value_decrypted = decrypt(my_value_encrypted)
+        assert_true(isinstance(my_value_decrypted, str))
+        assert_equal(my_value_decrypted, ensure_str(my_value))
```
mfraezz: Why do you need to `ensure_str` here, `_in_string_type_str`?

framework/encryption/__init__.py
alexschiller: @chrisseto recommended to mirror them if I understood his suggestion correctly.

alexschiller:
Closing in favor of #6766

PR #6567, opened 2016-11-16
[hotfix] Populate User Names From Institution Account Creation [OSF-7212]

caseyrollins:
#### Purpose
- When signing up for an OSF account via institution login, the `fullname` field of the user is automatically populated, but the `given`/`middle`/`family` name fields are not. 

#### Changes
- These fields are guessed when the user is created, but the guessed name fields were being overwritten by the names provided by the institution (i.e. being overwritten with `None` when the institution didn't provide `given`/`middle`/`family` names).
- This PR changes things to use the institution provided names, if they are provided. Otherwise, the names that are guessed from the `fullname` will be used.

#### Ticket
- [OSF-7212](https://openscience.atlassian.net/browse/OSF-7212)

### DIFF
--- a/api/institutions/authentication.py
+++ b/api/institutions/authentication.py
@@ -55,10 +55,10 @@ class InstitutionAuthentication(BaseAuthentication):
         user, created = get_or_create_user(fullname, username, reset_password=False)
 
         if created:
-            user.given_name = provider['user'].get('givenName')
-            user.middle_names = provider['user'].get('middleNames')
-            user.family_name = provider['user'].get('familyName')
-            user.suffix = provider['user'].get('suffix')
+            user.given_name = provider['user'].get('givenName') or user.given_name
+            user.middle_names = provider['user'].get('middleNames') or user.middle_names
+            user.family_name = provider['user'].get('familyName') or user.family_name
+            user.suffix = provider['user'].get('suffix') or user.suffix
             user.date_last_login = datetime.utcnow()
             user.save()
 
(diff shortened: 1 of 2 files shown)


api/institutions/authentication.py
```diff
@@ -54,10 +54,10 @@ def authenticate(self, request):
         user, created = get_or_create_user(fullname, username, reset_password=False)
 
         if created:
-            user.given_name = provider['user'].get('givenName')
-            user.middle_names = provider['user'].get('middleNames')
-            user.family_name = provider['user'].get('familyName')
-            user.suffix = provider['user'].get('suffix')
+            user.given_name = provider['user'].get('givenName') or user.given_name
```
icereval: @cslzchen thoughts?

cslzchen: @icereval  Looks good. 🔥 

Currently we don't have any issue with `fullname` being `None`, which will happen for OK State University. (See https://github.com/CenterForOpenScience/osf.io/pull/6754/commits/69e04b5711e3886ff4cfe578d4e095c574b1b7ef) I can fix the conflicts when this gets merged. Mine code are not as concise as @caseyrollins's and I forgot to add `middle_names` and `suffix` which are also guessed by `update_guessed_names`: [link to the code](https://github.com/CenterForOpenScience/osf.io/blob/develop/framework/auth/core.py#L616).

brianjgeiger:
@caseyrollins Could you rebase onto Master please? They'd like this as a hotfix. Let me know if there are any conflicts with develop after you do; that will complicate things.

caseyrollins:
@brianjgeiger Rebased 🌶 🦊

PR #6572, opened 2016-11-17
[feature] Tap chrisseto/homebrew-tokumx [OSF-7224]

alexschiller:
## Purpose

invoke setup is broken because tokumx uses sha1 instead of sha256. Unlikely to be updated.

## Changes

Tap chrisseto's fork instead

## Side effects
None

## Ticket

https://openscience.atlassian.net/browse/OSF-7224
[#OSF-7224]

## No QA needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,8 +192,8 @@ $ invoke setup
 - Temporarily, tokumx needs to be installed separately from the invoke setup script:
 
 ```bash
-$ tap chrisseto/homebrew-tokumx
-$ install tokumx-bin
+$ brew tap chrisseto/homebrew-tokumx
+$ brew install tokumx-bin
 ```
 
 To verify that your installation works, follow the instructions to [start the OSF](#running-the-osf) and
--- a/README.md
+++ b/README.md
@@ -189,6 +189,13 @@ $ pip install invoke==0.13.0
 $ invoke setup
 ```
 
+- Temporarily, tokumx needs to be installed separately from the invoke setup script:
+
+```bash
+$ tap chrisseto/homebrew-tokumx
+$ install tokumx-bin
+```
+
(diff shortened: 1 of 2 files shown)


tasks/__init__.py
```diff
@@ -776,7 +776,7 @@ def packages(ctx):
         'install elasticsearch',
         'install rabbitmq',
         'install node',
-        'tap tokutek/tokumx',
+        'tap chrisseto/homebrew-tokumx',
```
brianjgeiger: Can this be pinned to a commit or anything? New policy is to not just link to an arbitrary branch or repo, especially if it's not under centerforopenscience,

README.md
```diff
@@ -189,6 +189,13 @@ $ pip install invoke==0.13.0
 $ invoke setup
 ```
 
+- Temporarily, tokumx needs to be installed separately from the invoke setup script:
+
+```bash
+$ tap chrisseto/homebrew-tokumx
```
sloria: Shouldn't these be `brew tap..` and `brew install..`?

alexschiller: Er, yeah. Bad copy paste from tasks/__init__ where it does the brew part. Will fix that quick.

tasks/__init__.py
```diff
@@ -776,8 +776,6 @@ def packages(ctx):
         'install elasticsearch',
         'install rabbitmq',
         'install node',
-        'tap tokutek/tokumx',
```
alexschiller: See my original commit and briang's comment. That's what I originally had more or less.

binoculars:
Thanks. I got burned by this in the new mac setup. All fixed now. Docs update in CenterForOpenScience/COSDev#47

PR #6573, opened 2016-11-17
[Feature] Fix myProjects Uncategorized [OSF-6907]

alexschiller:
## Purpose
Uncategorized was being explicitly changed on item.attributes.category to be 'Uncategorized' which broke the icon when lines were clicked.

## Changes
No longer explicitly change the value since icons rely on it.

## Side effects
Also cleaned up the text in the side bar when you multi-click

## Ticket
https://openscience.atlassian.net/browse/OSF-6907

## QA Considerations
1. Make sure uncategorized projects show the right icon after being clicked.
2. Make sure the data listed in the multi-click for public and private projects shows up properly. 
Previously for public projects it would say: "Public, Last Modified on 2016-11-16 04:24 PM" Now it should say "Public project, Last Modified on 2016-11-16 04:24 PM" Test that both public projects (and uncategorized projects) show the proper data.


brianjgeiger:
@alexschiller This test branch keeps failing. Could you look into it?

PR #6578, opened 2016-11-18
Fix OSF Storage name bug

cwisecarver:
## Purpose

https://trello.com/c/rRi3ZPz5/34-osf-storage-is-written-as-osfstorage-in-files-widget


cwisecarver:
I deem it safe.

PR #6532, opened 2016-11-07
Feature/branding populate providers

binoculars:
## Purpose

Needed for branded preprints (see Jira ticket)

## Changes

- changes from @hmoco under CenterForOpenScience/osf.io#6428
- Updates to `scripts/populate_preprint_providers.py`

## Side effects

N/A

## Ticket

https://openscience.atlassian.net/browse/PREP-216

PR #6581, opened 2016-11-21
Feature/branded preprint nodelogs [PREP-232]

binoculars:
Needs:
- [ ] #6532 (Updates provider names and external urls)

## Purpose

In the node log for PREPRINT_INITIATED and PREPRINT_FILE_UPDATED, Add link to branded service and to the preprint itself.

## Changes

- Update to the node log serializer, model
- Updates the log actions json and log text parser
- Migration script

## Side effects

- Run the migration script, there should be no skipped files
- Be sure to run this before the branded migration MattF alludes to

## Ticket

https://openscience.atlassian.net/browse/PREP-232

## Notes for QA
screen shot 2016-11-21 at 15 55 18
It should look like this in the node logs after creating a preprint and updating the primary file
Note that the provider names and external url are pulled from the database and will be updated in #6532


scripts/migrate_logs_branded_preprints.py
```diff
@@ -0,0 +1,57 @@
...
+    logger.info('Preparing to migrate {} NodeLogs'.format(len(node_logs)))
+
+    count = 0
+
+    for log in node_logs:
+        preprint = None
+        node_id = log.params.get('node')
+
+        try:
+            preprint = PreprintService.find_one(Q('node', 'eq', node_id))
```
mfraezz: As long as this goes out before the branded migration, this shouldn't raise an exception. Otherwise it will definitely raise `MultipleResultsFound`. No action required, just as a note.

api/logs/serializers.py
```diff
@@ -11,6 +11,7 @@
 from website.project.model import Node
 from website.util import permissions as osf_permissions
 from framework.auth.core import User
+from website.preprints.model import PreprintProvider, PreprintService
```
mfraezz: It looks like the `PreprintProvider` import is unused, which will break flake

brianjgeiger:
@binoculars Could you merge this with develop and fix conflicts?

binoculars:
@brianjgeiger should be good to go now

PR #6595, opened 2016-11-28
[Improvement] Update campaign tests [OSF-7236]

cslzchen:
## Purpose

- Add unit tests for campaign initialization/refresh and campaign helper methods.
- Fixe the bug that `LAST_TIME_REFRESHED` is updated every time `get_campaigns()` is called.
  - The expected behavior should be when `CAMPAIGNS` are re-calculated.
  - It does not lead to any functional failure.

## Changes

See **Purpose** above.

## Side effects

No.

## Deployment Notes

No.

## QA Notes

No QA tests required.

## Ticket

https://openscience.atlassian.net/browse/OSF-7236


framework/auth/campaigns.py
```diff
@@ -74,7 +74,8 @@ def get_campaigns():
         except (NoResultsFound or QueryException or ImproperConfigurationError) as e:
             logger.warn('An error has occurred during campaign initialization: {}', e)
 
-    CAMPAIGNS_LAST_REFRESHED = datetime.utcnow()
+        CAMPAIGNS_LAST_REFRESHED = datetime.utcnow()
```
cslzchen: `CAMPAIGNS_LAST_REFRESHED` should be only updated when `CAMPAIGNS` are recalculated.

tests/test_campaigns.py
```diff
@@ -1,20 +1,172 @@
...
+        for campaign in self.campaign_lists:
+            proxy = campaigns.is_proxy_login(campaign)
+            if campaign.endswith('-preprints'):
+                assert_true(proxy)
+            else:
+                assert_false(proxy)
+        proxy = campaigns.is_proxy_login(self.invalid_campaign)
+        assert_true(proxy is None)
+
+    def test_system_tag_for_campaign(self):
```
cslzchen: For the next four tests, only test if the return value is `None` or `not None`. The `value` itself is not tested in order to avoid redundancy. Or we may want to initialize a test dictionary contains all detailed data and check the equality?

PR #6609, opened 2016-12-01
[#OSF-6790]Verify that Forked Nodes on Prod Don't have Wikis pointing to Source Nodes

chennan47:
## Purpose
Verify that Forked Nodes on Prod Don't have Wikis pointing to Source Nodes

## Changes
Nothing to be tested for QA. This is a script to be run to checkout if there is any forked node's wiki points to the wrong node.

## Ticket
https://openscience.atlassian.net/browse/OSF-6790


scripts/check_fork_node_wiki.py
```diff
@@ -0,0 +1,47 @@
...
+from website.app import init_app
+from website.project.model import Node
+from website.addons.wiki.model import NodeWikiPage
+from scripts import utils as script_utils
+
+logger = logging.getLogger(__name__)
+
+
+def get_fork_projects():
+    return Node.find(
```
mfraezz: This would run _a lot_ faster if you used `framework.mongo.database` instead of modm

```python
return database.node.find({'is_fork': True, 'is_deleted': False}, {'wiki_page_versions: 1})
```

scripts/check_fork_node_wiki.py
```diff
@@ -0,0 +1,47 @@
...
+        Q('is_fork', 'exists', True) &
+        Q('is_deleted', 'eq', False)
+    )
+
+
+def main():
+    counts = 0
+    for node in get_fork_projects():
+        if node.wiki_pages_versions:
```
mfraezz: It would be nice to have some indication of script progress here. Running it locally, my computer appeared to freeze up for about 10 minutes before any output was displayed.

scripts/check_fork_node_wiki.py
```diff
@@ -0,0 +1,47 @@
...
+            for key in node.wiki_pages_versions:
+                for wiki_id in node.wiki_pages_versions[key]:
+                    wiki = NodeWikiPage.load(wiki_id)
```
mfraezz: Extension of above:
```python
wiki = database.nodewikipage.find_one({'_id': wiki_id}, {'node': 1})
```

`wiki['node']` would be the current `wiki.node._id`. Everything else would be the same, just with dict-style accesses.

PR #6610, opened 2016-12-01
[#OSF-5809]Failed registrations with 100+ nodes

chennan47:
## Purpose
Fix the problem that projects with 100 + nodes registration will fail.

## QA notes:
Before change:
Steps to Reproduce:
1.On staging, go to https://staging.osf.io/yxzrn/
2.Create a fork of this project
3.Or make your own with 100+ nodes
4.Go to the registrations tab of the 100+ node project
5.Make a new registration
6.See that it fails
After change:
6. the registration should success

## Ticket
https://openscience.atlassian.net/browse/OSF-5809


tests/test_registrations/test_views.py
```diff
@@ -40,6 +40,16 @@ def test_node_register_page_registration(self, mock_archive):
         res = self.app.get(url, auth=self.user.auth)
         assert_equal(res.status_code, http.OK)
 
+    @mock.patch('website.archiver.tasks.archive')
+    def test_node_register_over_100_nodes(self, mock_archive):
```
mfraezz: This test takes ~30 seconds to run, and doesn't fail on `develop` without the changes to `register_node`, meaning that it's not useful as a regression test. Please ensure this test can fail.

```
~/D/osf.io ❯ g diff                                    remotes/cos/develop 
diff --git a/tests/test_registrations/test_views.py b/tests/test_registrations/test_views.py
index b932585..1513710 100644
--- a/tests/test_registrations/test_views.py
+++ b/tests/test_registrations/test_views.py
@@ -28,6 +28,16 @@ from tests.base import get_default_metaschema

 class TestRegistrationViews(RegistrationsTestBase):

+    def test_node_register_over_100_nodes(self, mock_archive):
+        for i in xrange(110):
+            node = NodeFactory(parent=self.node)
+            node.save()
+        reg = self.node.register_node(get_default_metaschema(), self.auth, '', None)
+        url = reg.web_url_for('node_register_page')
+        assert_equal(res.status_code, http.OK)
+
     def test_node_register_page_not_registration_redirects(self):
         url = self.node.web_url_for('node_register_page')
         res = self.app.get(url, auth=self.user.auth)
```
```
~/D/osf.io ❯ nosetests tests/test_registrations/test_views.py:TestRegistrationViews.test_node_register_over_100_nodes         remotes/cos/develop
[logging omitted...]
.
Ran 1 test in 30.840s

OK
```

website/project/model.py
```diff
@@ -2444,21 +2444,26 @@ def register_node(self, schema, auth, data, parent=None):
...
+            for node_contained in batch:
+                if not node_contained.is_deleted:
+                    child_registration = node_contained.register_node(
+                        schema=schema,
+                        auth=auth,
+                        data=data,
+                        parent=registered,
+                        recur=True,
+                    )
+                    if child_registration and not child_registration.primary:
```
brianjgeiger: If this is the check to see if this is a node pointer, that's different now in DjangOSF. There is, I think, a better way to get the node hierarchy in the node model at this point. To verify, make a pointer from a child to a parent, from a parent to a different child, and a child to itself, then see if this works or if it just goes forever / hits the maximum recursion depth.

This might not be the right time to try to eliminate this recursive function, but at least verify that it doesn't break with node pointers as described above.

PR #6615, opened 2016-12-01
Fix Broken APIv1 License Updates [OSF-7250]

caseyrollins:
#### Purpose
- Handle both `copyrightHolders` and `copyright_holders` casing for node license update.

#### Ticket
- [OSF-7250](https://openscience.atlassian.net/browse/OSF-7250)

### DIFF
--- a/website/project/model.py
+++ b/website/project/model.py
@@ -1604,7 +1604,7 @@ class Node(GuidStoredObject, AddonModelMixin, IdentifierMixin, Commentable, Spam
                     {
                         'id': value.get('id'),
                         'year': value.get('year'),
-                        'copyrightHolders': value.get('copyrightHolders')
+                        'copyrightHolders': value.get('copyrightHolders') or value.get('copyright_holders', [])
                     },
                     auth,
                     save=save


website/project/model.py
```diff
@@ -1604,7 +1604,7 @@ def update(self, fields, auth=None, save=True):
                     {
                         'id': value.get('id'),
                         'year': value.get('year'),
-                        'copyrightHolders': value.get('copyrightHolders')
+                        'copyrightHolders': value.get('copyrightHolders') or value.get('copyright_holders', [])
```
brianjgeiger: What's this thing's story?

caseyrollins: The APIv1 (current OSF front-end) update returns `copyright_holders` from [`licensePicker.js`](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/licensePicker.js#L219) whereas APIv2 will directly return `copyrightHolders` [here](https://github.com/CenterForOpenScience/osf.io/blob/8cb2fb682504c488232d7117645811ba3ce9fe97/api/nodes/serializers.py#L103).

APIv2 returns `copyrightHolders` because in [`set_license()`](https://github.com/CenterForOpenScience/osf.io/blob/8cb2fb682504c488232d7117645811ba3ce9fe97/website/project/licenses/__init__.py#L164) the license details are checked against the required properties of the license (which is literally `NodeLicense.properties.copyrightHolders`). So, the key passed to `set_license()` needs to end up being `copyrightHolders`.

Alternatively, I could change the line in licensePicker.js to return `copyrightHolders` in its payload.

brianjgeiger: Makes sense, thanks. This is good.

caseyrollins: It's essentially just a bit of a mess between JS, python, and a model property that's in camel case...

PR #6619, opened 2016-12-01
Migration updates: Logs, Targets, Settings

mfraezz:
## Changes
* Disable logging
* Exclude deleted nodes and nodes tagged with multiple providers
* Unset SHARE_URL to prevent `on_preprint_updated`


mfraezz:
Ran and validated the result of this against local data overnight. Nothing unexpected.

PR #6623, opened 2016-12-02
[WIP] Feature/API citation improvement

pattisdr:

### DIFF
--- a/api/citations/utils.py
+++ b/api/citations/utils.py
@@ -27,4 +27,15 @@ def render_citation(node, style='apa'):
 
     bibliography.cite(citation, warn)
     bib = bibliography.bibliography()
-    return unicode(bib[0] if len(bib) else '')
+
+    if len(bib):
+        if style == 'apa':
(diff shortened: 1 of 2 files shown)


api/citations/utils.py
```diff
@@ -27,4 +27,15 @@ def warn(citation_item):
 
     bibliography.cite(citation, warn)
     bib = bibliography.bibliography()
-    return unicode(bib[0] if len(bib) else '')
+
+    if len(bib):
+        if style == 'apa':
```
brianjgeiger: Could you put in tests for these cases? I'd like to make sure that if someone fixes CiteProc that what you have here won't go weird.

pattisdr: Good idea.

brianjgeiger:
@pattisdr It's not a big deal, but there's one extra commit that doesn't need to be there, and that occasionally causes trouble. Could you rebase this on master, please?

PR #6631, opened 2016-12-06
[#OSF-7232]Adding affiliations from components does not affiliate nested components (grandchildren)

chennan47:
## Purpose
Fix the problem that adding affiliations from components does not affiliate nested components (grandchildren)

## QA notes:
What to test:
1. Create a project (1) and add a component (2)
2. Within the component, create a nested component (3)
3. Within 3, create another nested component (4), and then another within 4 (5), and another within 5 (6)
4. Go to Component 2 and go to Settings
5. Click to Add Affiliation and in the modal that appears, select to add the affiliation to 2 and every component within it
6. After affiliation has been added, check the components and nested components
Before change:
Only a few of components child nodes has the correct affiliation information. Usually it will be component 2, 3, 5 has but 4,6 doesn't
After change:
All child nodes of component 2 should have the correct affiliation information no matter it is add or remove affiliation.

## Ticket

https://openscience.atlassian.net/browse/OSF-7232


website/static/js/osfHelpers.js
```diff
@@ -179,7 +179,7 @@ var getAllNodeChildrenFromNodeList = function(parent, nodeList) {
         var node = remaining.pop();
         for (var c in tree[node]){
             var child = tree[node][c];
-            remaining.push(tree[child]);
+            remaining.push([child]);
```
brianjgeiger: Since this is an isolated method and not part of a view model, this should be fairly easy to test. Could you add a basic test suite for this including this case? Single item, parent with child, parent with child and grandchild. If there's a couple of other variations that might be troublesome, go ahead and do those as well, but especially the current case. There's some good samples in fangorn.test.js.

brianjgeiger:
@chennan47 Can you rebase onto master, please? They'd like this as a hotfix.

chennan47:
closed in favor of https://github.com/CenterForOpenScience/osf.io/pull/6982

PR #6633, opened 2016-12-06
[hotfix] Fix Duplicate Preprint Licensing Logs [OSF-7260]

caseyrollins:
#### Purpose
- When editing a preprint, multiple PATCH requests are sent to the API, which results in duplicate `perprint_license_updated` logs being added to the preprint node.
- This PR prevents duplicate logs by checking that the update license record differs from the current license record before adding a log to the node. 
- Additionally, this PR updates the language of both node and preprint license logs to include the license that was applied (screenshots below).

#### Changes
- Adds a check to ensure a log is only added if the updated license record is different from the current license record. 
- Adds tests to ensure duplicate logs are not added.
- Adds a migration to remove the duplicate `preprint_license_updated` logs.
- Update `license_changed` and `preprint_license_updated` log language.
- Add `license` to logTextParser.js and `NodeLogParamsSerializer`.

#### Migration
```
python -m scripts.remove_duplicate_preprint_logs --dry 
```

#### Screenshots 
- "preprint_license_updated" (logs created before this PR -- without license information)
screen shot 2016-12-07 at 10 30 28 am

- "preprint_license_updated" (logs created after this PR -- will have license information)
screen shot 2016-12-07 at 10 30 21 am

- "node_license_changed" (all project license updated logs)
screen shot 2016-12-07 at 10 30 09 am

#### Ticket
- [OSF-7260](https://openscience.atlassian.net/browse/OSF-7260)

### DIFF
--- a/scripts/remove_duplicate_preprint_logs.py
+++ b/scripts/remove_duplicate_preprint_logs.py
@@ -1,0 +1,73 @@
+import sys
+import logging
+
+from modularodm import Q
+
+from framework.transactions.context import TokuTransaction
+from scripts import utils as script_utils
+from website.app import init_app
+from website.project.model import Node, NodeLog
+
+
+logger = logging.getLogger(__name__)
+
+
+# This is where all your migration log will go
+def do_migration():
+    dupe_nodes = [n for n in Node.find(Q('_id', 'in', list(set([l.node._id for l in NodeLog.find(Q('action', 'eq', 'preprint_license_updated'))])))) if NodeLog.find(Q('action', 'eq', 'preprint_license_updated') & Q('node', 'eq', n._id)).count() > 1]
+    logger.info('Found {} nodes with multiple preprint_license_updated logs'.format(len(dupe_nodes)))
+
+    for node in dupe_nodes:
+        preprint_license_updated_logs = [log for log in node.logs if log.action == 'preprint_license_updated']
+
+        log = preprint_license_updated_logs.pop()
+        while(preprint_license_updated_logs):
+            next_log = preprint_license_updated_logs.pop()
+            timedelta = log.date - next_log.date
+            if timedelta.seconds < 1:
+                logger.info(
+                    'Hiding duplicate preprint_license_updated log with ID {} from node {}, timedelta was {}'.format(
+                        log._id, node._id, timedelta
+                    )
+                )
+                log.should_hide = True
+                log.save()
+            else:
+                logger.info(
+                    'Skipping preprint_license_updated log with ID {} from node {}, timedelta was {}'.format(
+                        log._id, node._id, timedelta
+                    )
+                )
+
+            log = next_log
+
+
+def main(dry=True):
+    init_app(set_backends=True, routes=False)  # Sets the storage backends on all models
+
+    # Start a transaction that will be rolled back if any exceptions are un
+    with TokuTransaction():
+        do_migration()
+        if dry:
+            # When running in dry mode force the transaction to rollback
+            raise Exception('Abort Transaction - Dry Run')
+
+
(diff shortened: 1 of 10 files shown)


scripts/remove_duplicate_preprint_logs.py
```diff
@@ -0,0 +1,62 @@
...
+        preprint_license_updated_logs = [log for log in node.logs if log.action == 'preprint_license_updated']
+
+        log = preprint_license_updated_logs.pop()
+        while(preprint_license_updated_logs):
+            next_log = preprint_license_updated_logs.pop()
+            if (log.date - next_log.date).seconds < 1:
+                logger.info('Removing duplicate preprint_license_updated log with ID {} from node {}'.format(log._id, node._id))
+                log.should_hide = True
+                log.save()
+
```
mfraezz: Could you increase logging here? Important things to log tend to be what you're doing and why -- the time delta and logs that you've found but don't migrate would be good to know.

website/preprints/model.py
```diff
@@ -178,16 +178,17 @@ def set_published(self, published, auth, save=False):
...
-            save=False
-        )
+        license_record, license_changed = set_license(self, license_detail, auth, node_type='preprint')
+
+        if license_changed:
+            self.node.add_log(
+                action=NodeLog.PREPRINT_LICENSE_UPDATED,
+                params={
+                    'preprint': self._id
+                },
```
mfraezz: Discuss: Would it make sense to store the `license_record` here? It seems like the sort of thing we should be logging -- `"<User> updated <preprint> license to <license>"` instead of just `"<User> updated <preprint> license"`

It wouldn't be necessary to display it yet; until `NodeLogParamsSerializer` and `logActionsList.json` are updated it won't be apparent that it's being stored.

See if @saradbowman or anyone else from Product has opinions about how this log should be displayed.

caseyrollins: Discussed with @saradbowman/Product and agreed upon new log language for license updates, added [here](https://github.com/CenterForOpenScience/osf.io/pull/6633/commits/46d4326636eaa31840780603b070b12c95caa1c7)

scripts/remove_duplicate_preprint_logs.py
```diff
@@ -0,0 +1,73 @@
...
+
+# This is where all your migration log will go
+def do_migration():
+    dupe_nodes = [n for n in Node.find(Q('_id', 'in', list(set([l.node._id for l in NodeLog.find(Q('action', 'eq', 'preprint_license_updated'))])))) if NodeLog.find(Q('action', 'eq', 'preprint_license_updated') & Q('node', 'eq', n._id)).count() > 1]
+    logger.info('Found {} nodes with multiple preprint_license_updated logs'.format(len(dupe_nodes)))
+
+    for node in dupe_nodes:
+        preprint_license_updated_logs = [log for log in node.logs if log.action == 'preprint_license_updated']
+
```
sloria: It would be useful to log the _id of the node being migrated.

scripts/remove_duplicate_preprint_logs.py
```diff
@@ -0,0 +1,73 @@
...
+        log = preprint_license_updated_logs.pop()
+        while(preprint_license_updated_logs):
```
sloria: Why not a `for` loop?

caseyrollins: Honestly no reason, I guess I just thought to do it this way first yesterday. I could change it to use a `for` though.

scripts/remove_duplicate_preprint_logs.py
sloria: Ah ok, missed that. That's fine.

PR #6643, opened 2016-12-08
[fix] Update query in campaign initialization [OSF-7265] 

cslzchen:
### Purpose & Changes

As requested: `preprint_providers = PreprintProvider.find(Q('_id', 'ne', None))` is replaced by ~~`preprint_providers = PreprintProvider.find()`~~.

Further replaced by `PreprintProvider.objects.all()` for Django-OSF.

### Side Effects

No

### QA and Deployment Notes

No

### Ticket

https://openscience.atlassian.net/browse/OSF-7265

### DIFF
--- a/framework/auth/campaigns.py
+++ b/framework/auth/campaigns.py
@@ -1,9 +1,6 @@
 import furl
-import logging
 
 from django.utils import timezone
-from modularodm import Q
-from modularodm.exceptions import NoResultsFound, QueryException, ImproperConfigurationError
 
 from website import mails
 from website.models import PreprintProvider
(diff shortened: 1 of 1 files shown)


cslzchen:
Back to AD

framework/auth/campaigns.py
```diff
@@ -50,30 +45,27 @@ def get_campaigns():
...
-                        'system_tag': system_tag,
-                        'redirect_url': furl.furl(DOMAIN).add(path=url_path).url,
-                        'confirmation_email_template': mails.CONFIRM_EMAIL_PREPRINTS(template, name),
-                        'login_type': 'proxy',
-                        'provider': name,
-                    }
-                })
-        except (NoResultsFound or QueryException or ImproperConfigurationError) as e:
-            logger.warn('An error has occurred during campaign initialization: {}', e)
+        preprint_providers = PreprintProvider.objects.all()
```
cslzchen: `try ...  except ..` is removed. `.objects.all()` will return empty if no provider is found, which is safe for the `for` loop below. No exception will be thrown.

PR #6645, opened 2016-12-08
Fix notification digest and it's tests

cwisecarver:
## Purpose

This switches some map reduce mongo to SQL aggregations.

## Side effects

The user _id was being used for querying in mongo, I switched that to user instance because the other requires a join. I searched a bunch of different ways and only found two usages outside of tests. If I missed updating any of them they won't work.

### DIFF
--- a/osf_tests/factories.py
+++ b/osf_tests/factories.py
@@ -1,6 +1,10 @@
 # -*- coding: utf-8 -*-
 import functools
+
+import datetime
 import mock
+from factory import SubFactory
+from factory.fuzzy import FuzzyDateTime, FuzzyAttribute, FuzzyChoice
 from mock import patch, Mock
 
 import factory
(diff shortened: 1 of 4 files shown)


osf_tests/factories.py
```diff
@@ -664,7 +670,22 @@ class Meta:
...
+
+    return [node1._id, node2._id, node3._id, node4._id]
+
+
 class NotificationDigestFactory(DjangoModelFactory):
+    timestamp = FuzzyDateTime(datetime.datetime(1970, 1, 1, tzinfo=pytz.UTC))
+    node_lineage = FuzzyAttribute(fuzzer=make_node_lineage)
+    user = factory.SubFactory(UserFactory)
+    send_type = FuzzyChoice(choices=NOTIFICATION_TYPES.keys())
+    message = fake.text(max_nb_chars=2048)
```
sloria: Minor: to generate a new message for each instance, you could do

```python
message = factory.Faker('text', max_nb_chars=2048)
```

PR #6655, opened 2016-12-12
Fix spurious test failure with "creates new identifiers" [OSF-6970]

whit537:
## Purpose

Debug and fix a [spurious test case failure](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/156100208#L2556-L2557) with 37ae2bbc27de10bdc9cb3331aae408f719fbb881 under `TEST_BUILD="varnish"`, where `nodeControl > ViewModels > ProjectViewModel > create new identifiers` dies with a `TypeError` inside `contribAdder`:

```
TypeError: Fake XHR onreadystatechange handler threw exception:
'undefined' is not an object (evaluating 'response.data.map')
(/home/travis/build/CenterForOpenScience/osf.io/website/static/js/tests/tests.webpack.js:131102:0
<- webpack:///website/static/js/contribAdder.js:257:0)
```

screen shot 2016-12-12 at 5 34 33 am

## Changes

n/a

## Side effects

n/a

## Ticket

https://openscience.atlassian.net/browse/OSF-6970

PR #6656, opened 2016-12-12
Clean up Travis wiring [OSF-7270]

whit537:
## Purpose

On #6655 I was confused why the JS tests were being run under `TEST_BUILD="varnish"`. Turns [out](https://github.com/CenterForOpenScience/osf.io/pull/6655#issuecomment-266419210) that is intentional, so this PR fixes up names to avoid future confusion. It also DRYs up the wiring a bit.

## Changes

No user-facing changes.

## Side effects

None.

## Ticket

https://openscience.atlassian.net/browse/OSF-7270


brianjgeiger:
The Django change has done a larger refactor of Travis, so this is no longer applicable.

PR #6657, opened 2016-12-12
[CR] Figshare interface improvements [OSF-7272]

felliott:
## Purpose

The update to the Figshare v2 API means that a bunch of common fangorn features that weren't supportable before now work.

## Changes

* Allow copying files and folders out of Figshare.
* Enable folder creation in Projects.
* Enable download button for all files & folders.
* Update Figshare delete logic. All private files and folders can be deleted.
* Update supported features modal.

## Side effects

None expected.

## Ticket

[#OSF-7272] - [JIRA](https://openscience.atlassian.net/browse/OSF-7272)

### DIFF
--- a/framework/addons/data/addons.json
+++ b/framework/addons/data/addons.json
@@ -67,15 +67,15 @@
             },
             "View / download file versions": {
                 "status": "partial",
-                "text": "figshare content can be viewed and downloaded via OSF provided it is \"published\" on figshare."
+                "text": "figshare files can be viewed/downloaded via OSF, but version history is not supported."
             },
             "Add / update files": {
                 "status": "partial",
                 "text": "Files can be added but not updated."
             },
             "Delete files": {
-                "status": "none",
-                "text": "figshare files cannot be deleted via OSF."
+                "status": "partial",
+                "text": "Private files and filesets can be deleted via OSF."
             },
             "Logs": {
                 "status": "partial",
--- a/website/static/js/fangorn.js
+++ b/website/static/js/fangorn.js
@@ -2343,8 +2343,6 @@ function _dropLogic(event, items, folder) {
         folder = folder.parent();
     }
 
-    // if (items[0].data.kind === 'folder' && ['github', 'figshare', 'dataverse'].indexOf(folder.data.provider) !== -1) { return; }
-
     if (!folder.open) {
         return tb.updateFolder(null, folder, _dropLogic.bind(tb, event, items, folder));
     }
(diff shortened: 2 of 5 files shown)


framework/addons/data/addons.json
```diff
@@ -67,15 +67,15 @@
...
+                "text": "figshare files can be viewed/downloaded via OSF, but version history is not supported."
             },
             "Add / update files": {
                 "status": "partial",
                 "text": "Files can be added but not updated."
             },
             "Delete files": {
-                "status": "none",
-                "text": "figshare files cannot be deleted via OSF."
+                "status": "partial",
```
brianjgeiger: This looks like it will affect the API docs as well. Could you check and update as appropriate?

felliott: Do you mean in Erin's PR? (#6523)  That does not mention delete capabilities, only metadata.  Figshare doesn't seem to be mentioned anywhere else in the API.

website/static/js/fangorn.js
```diff
@@ -2441,9 +2439,7 @@ function isInvalidDropItem(folder, item, cannotBeFolder, mustBeIntra) {
         // no dropping if waiting on waterbutler ajax
         item.inProgress ||
         (cannotBeFolder && item.data.kind === 'folder') ||
-        (mustBeIntra && item.data.provider !== folder.data.provider) ||
-        //Disallow moving OUT of figshare
-        item.data.provider === 'figshare'
```
alexschiller: Is copying working? The logic looks like getCopyMode will return 'move' for figshare unless I missed something. 

`!(item.data.provider === 'figshare') ` should be added to `allowedToMove` to make getCopyMode return 'copy'

website/static/js/tests/fangorn.test.js
```diff
@@ -209,21 +209,6 @@ describe('fangorn', () => {
                 item.data.provider = 'github';
                 assert.equal(Fangorn.isInvalidDropItem(folder, item, false, true), true);
             });
-
-            it('cannot be dropped if item provider is figshare and private', () => {
```
alexschiller: Adding a test back into getCopyMode to ensure it is copying and not moving would be nice. 

These two tests could be reworded and modified to pass for regression sake.

website/static/js/fangorn.js
```diff
@@ -2441,9 +2439,7 @@ function isInvalidDropItem(folder, item, cannotBeFolder, mustBeIntra) {
         // no dropping if waiting on waterbutler ajax
         item.inProgress ||
         (cannotBeFolder && item.data.kind === 'folder') ||
```
alexschiller: Should you be able to drag folders into figshare now? If so, this line prevents that.

felliott: Folders can be dragged into the figshare root if the root is a Project.  Otherwise, it should be disallowed.  Code and tests updated to reflect this.

website/static/js/tests/fangorn.test.js
felliott: These test are no longer appropriate, but new tests have been added to `getCopyMode` and `allowedToMove` to update new behavior.

PR #6659, opened 2016-12-13
[djangosf] More core test fixes

mfraezz:
## Purpose
Fix tests

### Note
Some tests are only failing on Travis, not locally:
```
tests/test_addons.py::TestAddonAuth::test_auth_bad_cookie
tests/test_auth.py::TestAuthUtils::test_confirm_email
tests/test_auth_basic_auth.py::TestAuthBasicAuthentication::test_valid_cookie
```

## Changes
1. Fix some tests
2. Unskip tests in `osf_tests/` and `tests/` that seem like they should not be skipped
36. Fix some of the unskipped tests.


tests/test_preprints.py
```diff
@@ -256,6 +256,13 @@ class TestOnPreprintUpdatedTask(OsfTestCase):
     def setUp(self):
         super(TestOnPreprintUpdatedTask, self).setUp()
         self.user = AuthUserFactory()
+        if len(self.user.fullname.split(' ')) > 2:
```
sloria: Is this still necessary, given the way this PR changes how fullname is generated?

mfraezz: No, but it won't break anything. I wrote this before realizing that the new way of name generation could cause any number of tests to break.

PR #6670, opened 2016-12-15
[Feature] Rework fangorn folder and unavailable provider error handling [OSF-7285]

alexschiller:
## Purpose

Fangorn Error handling defaults to saying cannot upload folders for everything but chrome. This updates error handling to work properly.

## Changes
1. Rework how folders are handled for firefox and safari
2. Update safari error message

## Side effects
None known.

## Ticket
[#OSF-7285]
https://openscience.atlassian.net/browse/OSF-7285

## QA Considerations
1. Test uploading folders via drop on all browsers
2. Test other error messages related to dropping for files to ensure you do not get the cannot upload folders message for browsers
3. Safari is the only weird case. It has an error message that says it is either a folder or unavailable provider. 
4. I've requested that this ticket goes out with an IE fix for folders (linked in jira), so these two tickets can have a single test plan.


brianjgeiger:
Merge conflicts

PR #6671, opened 2016-12-15
[djangosf] Figshare Addon

mfraezz:
## Purpose
Add Figshare addon to DjangOSF

## Changes
* Reorganize TCI jobs, add `addons/` to modules tested.

## Side effects
1. 14 more tests are failing because an extra thousand are now being ran. 
1. More jobs are failing (3/5 instead of 1/5) due to test redistribution. 
1. No tests are failing due to `429`'s, but this is not a long-term solution to that issue.

## Ticket
None known.


sloria:
👍

PR #6685, opened 2016-12-27
[fix] [django-osf] Fix Default ODM Query for Preprint Provider [no ticket]

cslzchen:
@sloria @cwisecarver 

There is no `is_deleted` field for `osf_preprints_provider`. Mongo does not complain in OSF but Postgres fails in Django-OSF.


cwisecarver:
lgtm

PR #6689, opened 2017-01-03
[feature] [django-osf] update model `osf_institution` [CAS-15]

cslzchen:
@cwisecarver 

for `osf_institution` model:
- [x] add `delegation_protocol`
- [x] ~~remove `null=True` and `blank=True` for `logo_name` and `banner_name`~~
- [x] update `TODO` comments
- [x] update `populate_institutions`
- [x] update migration


osf/models/institution.py
```diff
@@ -18,18 +18,37 @@ class Institution(Loggable, base.ObjectIDMixin, base.BaseModel):
...
+    logout_url = models.URLField(null=True, blank=True)
+
     domains = fields.ArrayField(models.CharField(max_length=255), db_index=True, null=True, blank=True)
     email_domains = fields.ArrayField(models.CharField(max_length=255), db_index=True, null=True, blank=True)
-    logo_name = models.CharField(max_length=255, null=True)  # TODO: Could this be a FilePathField?
-    logout_url = models.URLField(null=True, blank=True)
-    name = models.CharField(max_length=255)
 
-    description = models.TextField(blank=True, default='', null=True)
+    # TODO what is this field?
```
cslzchen: Just curious, what is this new field? All users that are associated with this institution?

osf/models/institution.py
```diff
@@ -18,18 +18,37 @@ class Institution(Loggable, base.ObjectIDMixin, base.BaseModel):
...
+    contributors = models.ManyToManyField(
+        settings.AUTH_USER_MODEL,
+        through=InstitutionalContributor,
+        related_name='institutions'
+    )
+
+    # TODO Should we remove this is_deleted field given Institution now has its own model
```
cslzchen: We should remove or rename this field some time later, similar to `osf_preprintprovider`.

scripts/populate_institutions.py
```diff
@@ -71,6 +72,7 @@ def main(env):
                 'logout_url': SHIBBOLETH_SP_LOGOUT.format(encode_uri_component('https://osf.io/goodbye')),
                 'domains': [],
                 'email_domains': [],
+                'delegation_protocol': 'SAML_SHIBBOLETH',
```
cslzchen: The protocol is `SAML` and the implementation we use is `SHIBBOLETH`.
I reserve `SAML` for using `SAML` through `pac4j` and use `SAML_SHIBBOLETH` for our own Shibboleth Server.

osf/models/institution.py
```diff
@@ -18,18 +18,37 @@ class Institution(Loggable, base.ObjectIDMixin, base.BaseModel):
...
-    banner_name = models.CharField(max_length=255, null=True, blank=True)
+
+    name = models.CharField(max_length=255)
+    description = models.TextField(blank=True, null=True, default='')
+
+    # TODO Could `banner_name` and `logo_name` be a FilePathField?
+    # Both banner_name and logo_name are the name of the file under the `banners/` or `shields/` folder in
+    # `website/static/img/institutions/`. They cannot be null or empty, otherwise OSF dashboard page and institution
+    # project page won't finish loading.
+    banner_name = models.CharField(max_length=255, blank=True, default='')
```
cslzchen: Both `banner_name` and `logo_name` must be a non-empty string. I remove `null=True` but set default as empty string. The banner and logo files of the name must exist for each institutions.

I had thought `blank` is for non-empty at first but it is not at database level.

cwisecarver: During the migrations all fields **MUST** be `null=True`. If you want this to change after the migrations are finished you need to add a TODO that indicates that the `null=True` should be removed.

Also, `blank=True` implies `default=''`. It will not set it at the database level but django will set values left blank to `''` during the model save.

osf/models/institution.py
```diff
@@ -18,18 +18,37 @@ class Institution(Loggable, base.ObjectIDMixin, base.BaseModel):
...
+    logo_name = models.CharField(max_length=255, blank=True, default='')
+
+    # the protocol used to delegate authentication: `CAS`, `SAML`, `OAuth`, e.t.c
+    # only institutions with a delegation protocol (not null and not empty) shows up on institution login page
+    delegation_protocol = models.CharField(max_length=255, null=True, blank=True)
```
cwisecarver: If this is essentially an enum you should use the [choices attribute](https://docs.djangoproject.com/en/1.10/ref/models/fields/#choices) to provide options/validation.

`CharFields` should **never** be `null=True` (after the migrations are complete). If you want to have an option where there is no delegation protocol you should add a `'None'` option to the choices field that's database value is `''`.

osf/models/institution.py
cwisecarver: We probably left this field in to keep backwards compatibility with Institutions in modm that were built off of the Node model.

osf/models/institution.py
cwisecarver: This is not a new field. It holds a description of the institution. Out of the 16 institutions on production 14 have string values and 2 have null values.

cwisecarver: Oh, I thought you were talking about description.

It holds the list of contributors to the institution. People with permissions to contribute to the institution, assign projects to it, etc. It also isn't new. There is a contributors on Node that is what was used in modm.

scripts/populate_institutions.py
cwisecarver: I would just make sure that you make clear the difference in the choices tuple referenced in the comment above.

cwisecarver:
👍

osf/models/institution.py
```diff
@@ -17,19 +17,44 @@ class Institution(Loggable, base.ObjectIDMixin, base.BaseModel):
...
+    # Only institutions with a delegation protocol shows up on institution login page
+    DELEGATION_PROTOCOL_CHOICES = (
+        ('cas-pac4j', 'CAS by pac4j'),
+        ('saml-pac4j', 'SAML by pac4j'),
+        ('oauth-pac4j', 'OAuth by pac4j'),
+        ('saml-shib', 'SAML by shibboleth'),
+        ('', 'No Delegation Protocol'),
+    )
+    delegation_protocol = models.CharField(max_length=15, choices=DELEGATION_PROTOCOL_CHOICES, blank=True, default='')
+
```
cslzchen: Use `enum` and `choices` for delegation protocol. The "`<protocol>-<implementation>`" format makes it easier for us to understand. Use empty string `''` for non-delegation institutions.

scripts/populate_institutions.py
```diff
@@ -137,17 +144,19 @@ def main(env):
...
+                'delegation_protocol': 'saml-shib',
             },
             {
                 '_id': 'okstate',
                 'name': 'Oklahoma State University',
                 'description': '<a href="http://www.library.okstate.edu/research-support/research-data-services/">OSU Library Research Data Services</a>',
                 'banner_name': 'okstate-banner.png',
                 'logo_name': 'okstate-shield.png',
-                'auth_url': None,  # https://stwcas.okstate.edu/cas/login?service=...
+                'login_url': None,  # https://stwcas.okstate.edu/cas/login?service=...
```
cslzchen: Change `auth_url` to `login_url` (a few occurrences). It seems Django-OSF keeps receiving `auth_url` from new OSF merge.

cslzchen:
@cwisecarver PR ready again, thanks

- use `enum` and `choices`
- update TODO and comments for `null=True`
- replace `auth_url` (from merge) with `login_url`
- combine migrations

cslzchen:
@cwisecarver 

This is replaced #6754, which handle all OSF side updates including this model change.

PR #6690, opened 2017-01-03
Update Metrics on OSF Admin [OSF-7253]

erinspace:
## Purpose

OSF Admin Metrics changes, according to the needs and requests of the community team.  Redo the gathering of keen analytics in a new style and broken into new sections for clarity, speed, and usability.

screen shot 2017-01-03 at 1 45 20 pm

## Changes

- Revamp all calls to keen - now use the newer Keen dataviz  and analytics libraries
- Add a wide variety of new charts
- New chart organization
- add metrics-page that does the keen initializing and calls seperate analytics sections when a new tab is clicked
- All new layout, colors and organization.
- Removed link to old OSF Sales Analytics, as those charts aren't working at the moment.

## Side effects

Removed sidebar link to OSF Sales Analytics, so any documentation that says to look for that will be wrong. I left the page itself intact for now though. 

## Ticket
https://openscience.atlassian.net/browse/OSF-7253

### DIFF
--- a/admin/static/css/keen-dashboards.css
+++ b/admin/static/css/keen-dashboards.css
@@ -96,3 +96,7 @@ body.application .return-link:hover,
   font-size: 12px;
   padding: 8px 10px 5px;
 }
+
+.keen-dataviz-table-dataset td{
+  min-width: 100px !important;
+}
--- a/admin/static/css/keen-dashboards.css
+++ b/admin/static/css/keen-dashboards.css
@@ -98,7 +98,7 @@ body.application .return-link:hover,
 }
 
 .keen-dataviz-table-dataset td{
-  min-width: 100px !important;
+  min-width: 60px !important;
 }
 
 #dropdownMenuButton {
--- a/admin/static/css/keen-dashboards.css
+++ b/admin/static/css/keen-dashboards.css
@@ -106,7 +106,7 @@ body.application .return-link:hover,
   margin-right: 20px;
 }
 
-thead {
+.institution-tables thead {
   display: none;
 }
 
--- a/admin/static/css/keen-dashboards.css
+++ b/admin/static/css/keen-dashboards.css
@@ -103,6 +103,7 @@ body.application .return-link:hover,
 
 #dropdownMenuButton {
   margin-top: 20px;
+  margin-right: 20px;
 }
 
 thead {
@@ -112,4 +113,8 @@ thead {
 .hr-menu {
   background-color: lightgray;
   height: 1px;
+}
+
+#reload-node-logs {
+  margin-left: 20px;
 }
--- a/admin/static/css/keen-dashboards.css
+++ b/admin/static/css/keen-dashboards.css
@@ -100,3 +100,11 @@ body.application .return-link:hover,
 .keen-dataviz-table-dataset td{
   min-width: 100px !important;
 }
+
+#dropdownMenuButton {
+  margin-top: 15px;
+}
+
+thead {
+  display: none;
+}
--- a/admin/static/css/keen-dashboards.css
+++ b/admin/static/css/keen-dashboards.css
@@ -102,9 +102,14 @@ body.application .return-link:hover,
 }
 
 #dropdownMenuButton {
-  margin-top: 15px;
+  margin-top: 20px;
 }
 
 thead {
   display: none;
+}
+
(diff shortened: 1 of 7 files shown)


admin/static/css/keen-dashboards.css
```diff
@@ -96,3 +96,25 @@ body.application .return-link:hover,
...
+.keen-dataviz-table-dataset td{
+  min-width: 60px !important;
+}
+
+#dropdownMenuButton {
+  margin-top: 20px;
+  margin-right: 20px;
+}
+
+thead {
```
brianjgeiger: Have you verified that none of the css you added affects any of the rest of the admin app?

admin/static/js/metrics/metrics.js
```diff
@@ -1,187 +1,499 @@
...
+    readKey: keenReadKey
 });
 
+var publicClient = new keenAnalysis({
+    projectId: keenPublicProjectId,
+    readKey: keenPublicReadKey
+});
 
-Keen.ready(function () {
+var defaultHeight = 200;
```
brianjgeiger: I'm not super-concerned about responsiveness, but have you checked the site on various browser widths to make sure it basically works?

admin/static/js/metrics/metrics.js
```diff
@@ -1,187 +1,499 @@
...
+var bigMetricHeight = 350;
+var institutionTableHeight = "auto";
 
-    // Active user count!
-    var active_user_count = new Keen.Query("sum", {
-        eventCollection: "user_summary",
-        targetProperty: "status.active",
-        timeframe: "previous_1_days",
-        timezone: "UTC"
+var defaultColor = '#00BBDE';
```
brianjgeiger: Any way these can be pulled from CSS? If not, not a big deal, but it might also be better to make these configurable. You know, just in case we change our color scheme or something on the admin app.

admin/static/js/metrics/metrics.js
```diff
@@ -1,187 +1,499 @@
...
+    .catch(function(err){
+        differenceMetric.message(err.message);
     });
+};
+
+var getWeeklyUserGain = function() {
+    var queries = [];
+    var timeframes = [];
+
+    for (var i = 3; i < 12; i++) {
```
brianjgeiger: It is unclear why the boundary conditions on the loop are set the way they are.

admin/static/js/metrics/metrics.js
```diff
@@ -200,27 +512,434 @@ Keen.ready(function () {
...
+                event_collection: "node_summary",
+                target_property: key,
+                timeframe: "previous_1_days",
+                timezone: "UTC"
+            }))
+        }
+    }
+
+    var results = {};
+    var privateColor = '#F20066';
```
brianjgeiger: Same as the comment above about colors.

admin/static/js/metrics/metrics.js
erinspace: Agreed! I made sure they were all at the top - the reason it's nice to configure them this way was that keen configures its colors when the chart is instantiated in the js - and then I can also have conditional colors based on the names of classes and such, that I couldn't do with CSS selectors as all of the charts are generated on the fly. This gives us more control on the keen side!

admin/static/js/metrics/metrics.js
erinspace: Works in my versions of firefox and chrome, had to update to the latest version of Safari for some of the fancier javascript to work properly...

admin/static/js/metrics/metrics.js
```diff
@@ -236,7 +237,11 @@ var getWeeklyUserGain = function() {
 
 };
 
-var renderKeenMetric = function(element, type, query, height, colors=null, keenClient=client) {
+var renderKeenMetric = function(element, type, query, height, colors, keenClient) {
+
+    if (!keenClient) {
```
brianjgeiger: Is this going to work if nothing is passed in? My memory of JS is that you have to `typeof keenClient != "undefined"` or some other nonsense.

erinspace: It does indeed work! There are lots of places where this is called without specifying a client. Maybe didn't even have to change it as I changed it to start to accommodate older safari, but there ya go....

PR #6695, opened 2017-01-05
docker-sync

icereval:
cc @cwisecarver

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -61,9 +61,30 @@
 #       - ../waterbutler/:/code
 
 
+###############
+# Docker Sync #
+###############
+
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -204,19 +206,19 @@ services:
     stdin_open: true
```
cwisecarver: These two lines (stdin_open and tty) were only there to force docker to rebuild the mfr/wb containers. They're no longer needed after the PRs Fitz merged yesterday. Plus, they just make your terminal break.

icereval: leaving this on to facilitate ipdb debugging w/ attach

PR #6698, opened 2017-01-05
Fix api user filtering

cwisecarver:
Fixes: https://sentry.cos.io/sentry/osf-back-end/issues/256022/

### DIFF
--- a/api/users/views.py
+++ b/api/users/views.py
@@ -1,33 +1,32 @@
 
-from rest_framework import generics
-from rest_framework import permissions as drf_permissions
-from rest_framework.exceptions import NotAuthenticated, NotFound
-from django.contrib.auth.models import AnonymousUser
-
-from modularodm import Q
-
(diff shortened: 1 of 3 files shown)


cwisecarver:
@brianjgeiger This is the issue I was talking to you about earlier. From @pattisdr's comment about her commit, not filtering out merged users prior to 2.3, I think this will fix the problem. 

is_merged in django osf is a computed property so you cannot filter by it. I **believe** that this only worked in modm because the key didn't exist and the assertion in the query was negative. Since the key was missing the value of the missing key was, in fact, never equal to `True` and the query returned all results.

brianjgeiger:
Gotcha. Yeah, the pre-2.3 behavior what whatever messed up thing that we were just trying to maintain backwards compatibility in case anyone was relying on a side effect. Your description makes sense, so the changes look fine. Well, once the failing tests work, of course.

PR #6700, opened 2017-01-05
Creating nodes via the API with tags didn't work

cwisecarver:
Now it does and this is more DRFy.


cwisecarver:
@brianjgeiger Here's the new stuff for node creation w/ tags.

PR #6707, opened 2017-01-06
Update node serializer to accept tags.

cwisecarver:
## Purpose

Tags are a model now, not just a ListField. This updates the node serializer to allow creation and update of nodes in the v2 API with get_or_create of tags.


cwisecarver:
Merging. Verified on staging3 that it works. All tests past. lgtm

PR #6709, opened 2017-01-09
[WIP] Fix more migration bugs [#OSF-7422]

cwisecarver:
## Purpose

This fixes a bunch of bugs found when running migrations against production data. It also adds validation mgmt commands for confirming the migration ran correctly.

## Changes

- Add order field to node_relations
- Force node titles to be <= 200char
- Switch from print to logger.info
- Fix dropbox repr
- Stop indexing the location field, production data is too big
- Add code to try and recover guids with None referents, guids with inaccurate referents, etc

## Side effects

None are obvious.


cwisecarver:
@sloria I'm pretty sure this is safe but now that we're in develop, if you wanted to look it over before it's merged that would be appreciated.

osf_tests/settings.py
```diff
@@ -22,7 +22,6 @@
 
 REST_FRAMEWORK['ALLOWED_VERSIONS'] = (
         '2.0',
-        '2.0.1',
```
sloria: Why was this removed?

osf/management/commands/migratedata.py
```diff
@@ -4,6 +4,7 @@
 import importlib
 import itertools
 import logging
+import pprint
```
sloria: This doesn't appear to be used.

osf/management/commands/migraterelations.py
```diff
@@ -437,79 +445,87 @@ def migrate_node_through_models(self):
...
+                            if count % page_size == 0 or count == total:
+                                Contributor.objects.bulk_create(contributors)
+                                logger.info('Through {} nodes and {} contributors, '
+                                      'saved {} contributors'.format(count, contributor_count, len(contributors)))
+                                contributors = []
+                                modm_obj._cache.clear()
+                                modm_obj._object_cache.clear()
+                                logger.info('Took out {} trashes'.format(gc.collect()))
+
+                            node_relations = []
```
felliott: I think this should be declared in the outer scope with `contributors`.  Otherwise, only `node_relations` for the last page of nodes (if there are any) will be created.

osf/management/commands/migratedata.py
sloria: Oh woops, this is outdated... ignore

osf/management/commands/migraterelations.py
cwisecarver: You're right. I'll fix it. Seems like a very strange loop structure. I know it saves iterations but doing two individual things inside one loop iteration is confusing. I might pull this out into methods or something just to make it easier to not be very confused by.

osf_tests/settings.py
cwisecarver: Not sure. Either a bad merge or a quick trigger finger. It's still in local_travis so I'll put it back here.

cwisecarver:
closing in favor of #6818

PR #6710, opened 2017-01-09
[OSF-7324] Add Registries, share hook, migration and more!

hmoco:
## Purpose
Add a link and config settings for registries ember-app.
Add docker-information for ember-osf-registries
Add hooks to update share with new registrations.

## Ticket

https://openscience.atlassian.net/browse/OSF-7324


scripts/migration/migrate_share_registration_data.py
```diff
@@ -0,0 +1,55 @@
...
+    logger.info('Preparing to migrate {} registrations.'.format(registrations_count))
+    for registration in registrations:
+        if not registration.is_public or registration.is_deleted:
+            pass
+        count += 1
+        logger.info('{}/{} - {}'.format(count, registrations_count, registration._id))
+        try:
+            on_registration_updated(registration)
+        except Exception as e:
+            # TODO: This reliably fails for certain nodes with
```
chrisseto: Is this still the case or just copied from the preprints migration?

scripts/migration/migrate_share_registration_data.py
```diff
@@ -0,0 +1,55 @@
...
+            logger.warn('Encountered exception {} while posting to SHARE for registration {}'.format(e, registration._id))
+        else:
+            successes.append(registration._id)
+
+    logger.info('Successes: {}'.format(successes))
+    logger.info('Failures: {}'.format(failures))
+
+
+def main():
+    dry_run = '--dry' in sys.argv
```
chrisseto: Doesn't look like dry run will actually do anything here. Same with the transaction. Should dry mode just not send registrations to SHARE?

scripts/migration/migrate_share_registration_data.py
```diff
@@ -0,0 +1,55 @@
...
+from website.app import init_app
+from website.project.tasks import on_registration_updated
+
+logger = logging.getLogger(__name__)
+
+def migrate():
+    assert settings.SHARE_URL, 'SHARE_URL must be set to migrate.'
+    assert settings.SHARE_API_TOKEN, 'SHARE_API_TOKEN must be set to migrate.'
+    registrations = Registration.objects.all()
+    registrations_count = len(registrations)
```
chrisseto: Use `.count()` not `__len__`. IIRC `__len__` actually evaluates the query set, loads it into memory, and then returns the `len` of that list.

scripts/migration/migrate_share_registration_data.py
```diff
@@ -0,0 +1,55 @@
...
+    successes = []
+    failures = []
+    count = 0
+
+    logger.info('Preparing to migrate {} registrations.'.format(registrations_count))
+    for registration in registrations:
```
chrisseto: @cwisecarver should this be done with the whatever the iter method is? Do you remember if this also just loads everything or not?

website/project/tasks.py
```diff
@@ -56,3 +57,57 @@ def on_node_updated(node_id, user_id, first_save, saved_fields, request_headers=
             }, headers={'Authorization': 'Bearer {}'.format(settings.SHARE_API_TOKEN), 'Content-Type': 'application/vnd.api+json'})
             logger.debug(resp.content)
             resp.raise_for_status()
+            if node.is_registration:
```
chrisseto: This should be performed instead of the above request if the node is a registration.

website/project/tasks.py
```diff
@@ -56,3 +57,57 @@ def on_node_updated(node_id, user_id, first_save, saved_fields, request_headers=
...
+        }
+    }, headers={'Authorization': 'Bearer {}'.format(settings.SHARE_API_TOKEN), 'Content-Type': 'application/vnd.api+json'})
+    logger.debug(resp.content)
+    resp.raise_for_status()
+
+def format_registration(node):
+    registration_graph = GraphNode('registration', **{
+        'title': node.title,
+        'description': node.description or '',
+        'is_deleted': node.retraction or not node.is_public or node.is_preprint_orphan or 'qatest' in (node.tags.all() or []) or node.is_deleted,
```
chrisseto: Is `is_preprint_orphan` relevant here?

scripts/migration/migrate_share_registration_data.py
cwisecarver: I'd use `registrations.iterator():` if you only need the registration inside the loop and won't be accessing it again. It skips putting the records in the query cache and uses one query per record instead of paging them. Saving memory at the expense of # of queries and time.

scripts/migration/migrate_share_registration_data.py
hmoco: Think this is a leftover from copying the other script. Will remove it.

scripts/migration/migrate_share_registration_data.py
hmoco: It seems like a transaction rollback is being used for dry, but that wouldn't affect the requests sent to SHARE, so yes, dry run should just not make the requests.

website/project/tasks.py
hmoco: Nope!

scripts/migration/migrate_share_registration_data.py
```diff
@@ -0,0 +1,42 @@
...
+    assert settings.SHARE_URL, 'SHARE_URL must be set to migrate.'
+    assert settings.SHARE_API_TOKEN, 'SHARE_API_TOKEN must be set to migrate.'
+    registrations = Registration.objects.all()
+    registrations_count = registrations.count()
+    count = 0
+
+    logger.info('Preparing to migrate {} registrations.'.format(registrations_count))
+    for registration in registrations.iterator():
+        if not registration.is_public or registration.is_deleted:
+            pass
```
chrisseto: continue?

hmoco: Yep :/

hmoco: That seems important

scripts/migration/migrate_share_registration_data.py
```diff
@@ -0,0 +1,42 @@
...
+    assert settings.SHARE_API_TOKEN, 'SHARE_API_TOKEN must be set to migrate.'
+    registrations = Registration.objects.all()
+    registrations_count = registrations.count()
+    count = 0
+
+    logger.info('Preparing to migrate {} registrations.'.format(registrations_count))
+    for registration in registrations.iterator():
+        if not registration.is_public or registration.is_deleted:
+            continue
+        count += 1
```
chrisseto: This should be above the `continue` otherwise you'll end up with some add looking counts. 
Or, better idea, just filter them out in the query.

hmoco:
@brianjgeiger , so the changes to django need to be reversed?

brianjgeiger:
I'm not sure. @sloria will have to do some dealing with this on the hotfix merge, so it might be useful for him to have this handy either as reference or for extra commits? No idea. Chat with him and come up with a plan, please. But once this is good, we can merge it in and have everything ready for Registries, hopefully.

PR #6720, opened 2017-01-12
Fix Node preprint_url Property 

caseyrollins:
#### Purpose
- `Node.preprint_url` currently attempts to access an invalid field (`preprint`), resulting in an `AttributeError: 'Node' object has no attribute ‘preprint’` whenever `preprint_url` is called.

#### Changes
- Return `node.preprints.first().url` instead

### DIFF
--- a/osf/models/node.py
+++ b/osf/models/node.py
@@ -393,8 +393,8 @@ class AbstractNode(DirtyFieldsMixin, TypedModel, AddonModelMixin, IdentifierMixi
     def preprint_url(self):
         if self.is_preprint:
             try:
-                return self.preprint.url
-            except IndexError:
+                return self.preprints.first().url
+            except AttributeError:
                 pass
 
     @property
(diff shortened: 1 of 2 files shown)


osf/models/node.py
```diff
@@ -393,7 +393,7 @@ def is_preprint_orphan(self):
     def preprint_url(self):
         if self.is_preprint:
             try:
-                return self.preprint.url
+                return self.preprints.first().url
             except IndexError:
```
sloria: I *think* this needs to be changed to `AttributeError` because IIRC, `.first()` can return `None`.

sloria: Would be nice to have a test for that, just to make sure.

caseyrollins:
Closing as this was fixed in https://github.com/CenterForOpenScience/osf.io/commit/17ad92aa6424d58247b28e6f9f6131f23b0fd991

PR #6724, opened 2017-01-13
Move addons code (views, static, etc) to addons/ directory

sloria:
## Purpose

Colocate all addons code to the addons/ directory.

## Changes

- Move views, static files, and configuration from website/addons/ to addons/
- Addon configuration was moved from `website/addons/<addon_name>/__init__.py` to `addons/<addon_name>/constants.py` in order to prevent import side effects when initializing the Django apps.
- Remove redundant Flask view for serving addons' static assets. NOTE: static assets are expected to be served by nginx when `DEV_MODE=False`.

## Side effects

- `website/addons/` still exists and contains the legacy MODM models. The legacy models are necessary for the mongo -> postgres migration


sloria:
Doh; I'll also need to move `website.addons.base.AddonConfig` over to /addons for tests to work and to make things cleaner. I'll do that now.

addons/base/signals.py
```diff
@@ -0,0 +1,4 @@
+import blinker
```
sloria: Good idea.

addons/base/views.py
```diff
@@ -0,0 +1,742 @@
...
+    'delete': NodeLog.FILE_REMOVED,
+    'create_folder': NodeLog.FOLDER_CREATED,
+}
+
+
+@must_be_signed
+@no_auto_transaction
+@must_be_valid_project
+def create_waterbutler_log(payload, **kwargs):
+    with TokuTransaction():
```
sloria: No, it isn't; I've been switching these to `atomic.transaction()`, but I missed this one. I'll change this over, but it wouldn't be a bad idea to patch it. I'll do this in a separate PR.

PR #6729, opened 2017-01-17
[OSF-7354] Add staging preprint providers: SciELO + AgriXiv

hmoco:
## Purpose
Add the SciELO, AgriXiv preprint provider and add environment options for preprint providers. Needs: https://github.com/CenterForOpenScience/ember-preprints/pull/251

## Changes
Current states:
screen shot 2017-01-24 at 10 01 54 am
screen shot 2017-01-30 at 4 12 22 pm

## Ticket
https://openscience.atlassian.net/browse/OSF-7354


scripts/populate_preprint_providers.py
```diff
@@ -360,14 +361,666 @@ def main():
...
+                    (['Social and Behavioral Sciences', 'Science and Technology Studies', 'Work, Economy and Organizations'], False),
+                    (['Social and Behavioral Sciences', 'Science and Technology Studies', 'Other Sociology'], False),
+                    (['Social and Behavioral Sciences', 'Other Social and Behavioral Sciences'], False)
+                ]
+            }
+        ]
+        PREPRINT_PROVIDERS = PREPRINT_PROVIDERS + STAGING_PREPRINT_PROVIDERS
+
     with transaction.atomic():
         for provider_data
```
brianjgeiger: Typo. Should be, 'A specified…'

scripts/populate_preprint_providers.py
```diff
@@ -360,14 +361,666 @@ def main():
                 (['Law'], True),
                 (['Social and Behavioral Sciences'], True),
             ],
-        },
+        }
     ]
 
+    if env == 'stage':
+        STAGING_PREPRINT_PROVIDERS = [
```
brianjgeiger: Because of the size of the preprint provider data, it might be better to just have a dictionary of all possible preprint providers, then make a list of all of the ids of the preprint providers we want on any given box. Then, when the script is run, it can just build the lists of dictionaries based on the list of ids. Error if any of the ids don't match.

hmoco: The only problem is that some providers require the ` use_plos ` flag to be set to properly create their attribute dict. So those provider dictionaries need to be created at runtime.

docker-sync.yml
```diff
@@ -25,14 +25,14 @@ syncs:
...
 
-#  preprints-sync:
-#    src: '../ember-preprints'
-#    dest: '/code'
-#    sync_strategy: 'unison'
-#    sync_args: [ '-prefer newer' ]
-#    sync_excludes_type: 'Name'
-#    sync_excludes: ['.DS_Store', '*.map', '*.pyc', '*.tmp', '.git', '.idea', 'bower_components', 'node_modules', 'tmp', 'dist']
-#    watch_excludes: ['.*\.DS_Store', '.*\.map', '.*\.pyc', '.*\.tmp', '.*/\.git', '.*/\.idea', '.*/bower_components', '.*/node_modules', '.*/tmp', '.*/dist']
+  preprints-sync:
```
brianjgeiger: Is this supposed to be active now, or is that something that gets activated for individuals based on if they're doing preprints work? (i.e. should this change be checked in?)

scripts/tests/test_add_preprint_providers.py
```diff
@@ -0,0 +1,29 @@
...
+        populate_main('prod')
+        assert_equal(PreprintProvider.find().count(), 4)
+
+    def test_add_default_providers(self):
+        populate_main(None)
+        assert_equal(PreprintProvider.find().count(), 4)
+
+    def test_add_staging_providers(self):
+        populate_main('stage')
+        assert_equal(PreprintProvider.find().count(), 6)
```
brianjgeiger: This is a little fragile because of the magic numbers. Can you maybe get the list of preprint_providers_to_add for each environment then count that and ensure that, when they're added, the numbers match?

docker-compose.override.yml
```diff
@@ -54,3 +54,9 @@ services:
   api:
     volumes_from:
       - container:osf-sync
+
+  preprints:
+    build:
+      context: ../ember-osf2/ember-preprints
```
sloria: Remove/modify anything that is specific to your local dev environment.

scripts/tests/test_populate_preprint_providers.py
```diff
@@ -0,0 +1,12 @@
+from nose.tools import *  # noqa
+
```
sloria: Remove this file if there will be no tests in it.

scripts/update_taxonomies.py
```diff
@@ -18,7 +18,7 @@
 logging.basicConfig(level=logging.INFO)
 
 
-def update_taxonomies(filename):
+def update_taxonomies(filename, warn=True):
```
sloria: Flags like `warn` are unnecessary when using the `logging` module. You can just do 

```python
logger.warn('....')
```

Then expose a way to change the logging level from the CLI:

```python
if '--warn' in sys.argv:
    logger.setLevel(logging.WARNING)
```

(Note: the scripts [README](https://github.com/CenterForOpenScience/osf.io/blob/8019d1510b79619a59b1818a212fe5c8cc311342/scripts/README.md#L40-L57) has a more complete example)

hmoco: The purpose for the warning argument was to prevent warnings when using the function on a test. Would I be able to set a logger level above warning on my test?

sloria: Yes, you should be able to do something like

```python
logging.getLogger('scripts.update_taxonomies').setLevel(logging.CRITICAL)
```

hmoco: Totally found a way to do it. Thank you

hmoco:
@sloria, good catches, have responded.

sloria:
👍  Since this will need to be released before DjangoOSF, I'll squash-merge this and make it into a hotfix.

PR #6731, opened 2017-01-17
Fix Preprints api taxonomy [OSF-7352]

alexschiller:
## Purpose

Taxonomy api is broken

## Changes

Changes the query to work with django

## Side effects

None known
## Ticket
[#OSF-7352]
https://openscience.atlassian.net/browse/OSF-7352

### DIFF
--- a/api/preprint_providers/views.py
+++ b/api/preprint_providers/views.py
@@ -226,7 +226,7 @@ class PreprintProviderSubjectList(JSONAPIBaseView, generics.ListAPIView):
             #  Calculate this here to only have to do it once.
             allowed_parents = [id_ for sublist in provider.subjects_acceptable for id_ in sublist[0]]
             allows_children = [subs[0][-1] for subs in provider.subjects_acceptable if subs[1]]
-            return [sub for sub in Subject.find(Q('parents', 'eq', parent)) if provider.subjects_acceptable == [] or self.is_valid_subject(allows_children=allows_children, allowed_parents=allowed_parents, sub=sub)]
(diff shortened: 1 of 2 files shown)


api/preprint_providers/views.py
```diff
@@ -226,7 +226,7 @@ def get_queryset(self):
             #  Calculate this here to only have to do it once.
             allowed_parents = [id_ for sublist in provider.subjects_acceptable for id_ in sublist[0]]
             allows_children = [subs[0][-1] for subs in provider.subjects_acceptable if subs[1]]
-            return [sub for sub in Subject.find(Q('parents', 'eq', parent)) if provider.subjects_acceptable == [] or self.is_valid_subject(allows_children=allows_children, allowed_parents=allowed_parents, sub=sub)]
+            return [sub for sub in Subject.find(Q('parents___id', 'eq', parent)) if provider.subjects_acceptable == [] or self.is_valid_subject(allows_children=allows_children, allowed_parents=allowed_parents, sub=sub)]
```
mfraezz: This would also work with `Subject.objects.filter(parents___id=parent)`, but that's not necessary until phase 2

alexschiller: IIRC I tried doing it that way and it failed. @cwisecarver pointed out the reason, pretty sure it was a rats nest and decided to do it the easier way until phase 2.

mfraezz: You may have only had two underscores? I tried the above and relevant tests passed.

cwisecarver:
glam

PR #6733, opened 2017-01-17
Basic admin app functionality working for DjangoOSF [OSF-7338]

erinspace:
## Purpose

The OSF Admin app needed some updates to get working with the brand new DjangOSF. This is the first iteration of that, getting the basic containers for the app and the app's assets configured, and a few of the imported models and database calls updated.

**NOTE!** See side effects below - still perhaps needs a bit of configuration to run properly. 

## Changes

- docker-compose and settings for `admin` and `admin-assets`
- Remove the `MyUser` model - all OSFAdmin users are now `OSFUser`s
- Migration for this ^
- Fix the desk tokens to work with new user
- Update desk frontend form to show tokens if they're entered
- Update node view and registration view
- Update node and contributor serializers

## Side effects

Issues encountered:
- Running the app through docker, the `--no-init-app` arg was getting stripped away when running the server through `manage.py`, so the OSF app was initiating early. Tried for a long while to fix this, but thought it better get the whole thing in first. 
- Had some trouble with running `adminassets` because of some javascript webpack errors from osf javascript includes - namely:
    - [osf-style](https://github.com/CenterForOpenScience/osf.io/blob/develop/admin/static/js/pages/prereg-admin-page.js#L4)
    - [clipboard.js](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/clipboard.js#L4)
    - [growlBox.js](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/growlBox.js#L4)

Still not fixed:
- Admin's admin panel, for adding new users with group permissions, etc
- Admin tests

## Ticket

https://openscience.atlassian.net/browse/OSF-7338


admin/nodes/views.py
```diff
@@ -192,10 +194,7 @@ class RegistrationListView(OSFAdmin, ListView):
     context_object_name = '-node'
 
     def get_queryset(self):
-        query = (
-            Q('is_registration', 'eq', True)
-        )
-        return Node.find(query).sort(self.ordering)
```
sloria: Does this break ordering?

admin/base/utils.py
```diff
@@ -40,10 +40,14 @@ class PreregAdmin(OSFAdmin):
...
 
 
 def reverse_qs(view, urlconf=None, args=None, kwargs=None, current_app=None, query_kwargs=None):
     base_url = reverse(view, urlconf=urlconf, args=args, kwargs=kwargs, current_app=current_app)
     if query_kwargs:
         return '{}?{}'.format(base_url, urlencode(query_kwargs))
+
+
+def osf_admin_check(user):
+    return user.is_authenticated() and user.groups.filter(name='osf_admin')
```
erinspace: I have tried it and does indeed return falsy:

```
In [6]: me
Out[6]: <OSFUser: erin@cos.io>

In [7]: me.groups.filter(name='nopenotathing')
Out[7]: []
````

admin/nodes/views.py
erinspace: Good call! Updated to fix this!

PR #6734, opened 2017-01-18
[WIP] Feature/limit flask django sql connections

cwisecarver:
## Purpose

Close django connections after request and reset queries before every request for flask app.

## Changes

Added a new set of handlers for flask that use internal django methods to reset queries and close old connections.

## Side effects

@alexschiller might have a more productive day tomorrow.

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -4,8 +4,8 @@ import abc
 import datetime as dt
 import functools
 import logging
-import re
 import os
+import re
 import shutil
 import unittest
 
@@ -13,12 +13,9 @@ import blinker
 import httpretty
 import mock
 import pytest
+from addons.wiki.models import NodeWikiPage
 from django.test import TestCase as DjangoTestCase
 from faker import Factory
-from nose.tools import *  # noqa (PEP8 asserts); noqa (PEP8 asserts)
-from pymongo.errors import OperationFailure
-from webtest_plus import TestApp
-
 from framework.auth import User
 from framework.auth.core import Auth
 from framework.celery_tasks.handlers import celery_before_request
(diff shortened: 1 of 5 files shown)


sloria:
We'll have to be mindful of Python 2's less-than-ideal import behavior. Specifically, anything that imports from `django` in the `framework` package will need to do `from __future__ import absolute_import` to avoid accidentally importing `framework.django`.

Got bit by a similar issue in my addons PR, so I thought I'd pass a word of warning.

cwisecarver:
@chrisseto @sloria Could you review at least the last commit and see if you can think of a less kludgy way to accomplish the same thing? If those handlers are attached while tests are running then the collections are closed twice ... which doesn't work. I looked into the way the django db test case does it and the way pytest django does it and they both appear to be based mostly on magic and hope.

tests/base.py
```diff
@@ -92,7 +94,7 @@ def teardown_database(client=None, database=None):
     client.drop_database(database)
 
 
-@pytest.mark.django_db
+@pytest.mark.django_db(transaction=True)
```
sloria: Why is this necessary? Does this slow down the tests?

cwisecarver: It isn't. I forgot to commit a file. Fixed nao.

chrisseto:
What part of this looks kludgy? It's pretty much exactly what Django does just shoved down flasks throat.

How are the connections closed twice? Wouldn't the closed one get skipped over when `close_outdated_more_words` is called?

cwisecarver:
@chrisseto I was thinking the `attach_django_handlers` kwarg that changes behavior only during testing was kludgy. Not so kludgy that I didn't write it out an put in a PR though. 

I think they're being closed twice because the testing stuff closes them with a different method than the `close_outdated_more_words` thing. The trace is [here](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/192890561#L1608).

cwisecarver:
💩

PR #6736, opened 2017-01-18
[Fix] Fix Scripts for Populating OAuth Scopes [#OSF-7332]

cslzchen:
@sloria 

Update scripts to populate `ApiOAuth2Scope`.

This is part of https://openscience.atlassian.net/browse/OSF-7332.

### DIFF
--- a/scripts/register_oauth_scopes.py
+++ b/scripts/register_oauth_scopes.py
@@ -1,24 +1,23 @@
 """
-Register the list of OAuth2 scopes that can be requested by third parties. This populates the Mongo collection
-    referenced by CAS when responding to authorization grant requests.
-
-The database class is minimal; the exact specification for what a scope contains lives in the
-    python module from which this collection is drawn.
+Register the list of OAuth2 scopes that can be requested by third parties. This populates the Postgres collection
+referenced by CAS when responding to authorization grant requests. The database class is minimal; the exact
+specification for what a scope contains lives in the python module from which this collection is drawn.
 """
+
 import sys
 import logging
 
-from modularodm import Q
-from modularodm.exceptions import NoResultsFound
+import django
+from django.db import transaction
+
+django.setup()
 
 from scripts import utils as script_utils
 
 from framework.auth import oauth_scopes
-from framework.mongo import set_up_storage, storage
-from framework.transactions.context import TokuTransaction
-
+from osf.models import ApiOAuth2Scope
 from website.app import init_app
-from website.oauth.models import ApiOAuth2Scope
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


scripts/register_oauth_scopes.py
```diff
@@ -27,43 +26,38 @@ def get_or_create(name, description, save=True):
...
+    Given a dictionary of scope definitions, {name: scope_namedtuple}, load the
+    resulting data into a database collection
     """
+
     scope_dict = oauth_scopes.public_scopes
 
-    # Clear the scope collection and populate w/ only public scopes,
-    # nothing references these objects other than CAS in name only.
-    ApiOAuth2Scope.remove()
+    if clear:
```
cslzchen: The previous script clears the collection every time the scripts is run, which is preserved here as an option.

scripts/register_oauth_scopes.py
```diff
@@ -74,12 +68,10 @@ def do_populate():
 
 
 def main(dry=True):
-    init_app(set_backends=True, routes=False)  # Sets the storage backends on all models
 
-    with TokuTransaction():
-        # Set storage backends for this model
-        set_backend()
```
cslzchen: I removed `set_backend()`.

scripts/register_oauth_scopes.py
```diff
@@ -1,24 +1,23 @@
...
+
+django.setup()
 
 from scripts import utils as script_utils
 
 from framework.auth import oauth_scopes
-from framework.mongo import set_up_storage, storage
-from framework.transactions.context import TokuTransaction
-
+from osf.models import ApiOAuth2Scope
```
cslzchen: Is there a difference between `from website.oauth.models import ApiOAuth2Scope` and `from osf.models import ApiOAuth2Scope`?

cwisecarver: Unless you're checking `isinstance`,`issubclass`, or `==` then no. They should have the same behavior for querying.

scripts/register_oauth_scopes.py
```diff
@@ -27,43 +26,38 @@ def get_or_create(name, description, save=True):
...
+    :return: the scope object
     """
+
     if name != name.lower():
         raise ValueError('Scope names are case-sensitive, and should always be lower-case.')
 
     try:
-        scope_obj = ApiOAuth2Scope.find_one(Q('name', 'eq', name))
-    except NoResultsFound:
+        scope_obj = ApiOAuth2Scope.objects.get(name=name)
```
cwisecarver: @cslzchen https://docs.djangoproject.com/en/1.10/ref/models/querysets/#get-or-create

cwisecarver: nm, that might not be as effective as what you're already doing.

cwisecarver:
lgtm

PR #6739, opened 2017-01-18
Fix datetimeawarejsonfield decoding and add tests

cwisecarver:
## Purpose

https://sentry.cos.io/sentry/osf-back-end/issues/269425/

## Changes

Switched from calling to_python which only runs when forms are used to using from_db_value which is called every time data is retrieved from the database.


cwisecarver:
Tests are happy, merging.

PR #6741, opened 2017-01-18
[Fix] Fix Personal Access Token Authentication [#OSF-7332]

cslzchen:
This issue is caused by CAS returns `username` instead of `GUID` for OSF during PAT authentication. 

This is the second (and the last) part of https://openscience.atlassian.net/browse/OSF-7332.

### DIFF
--- a/api/base/authentication/drf.py
+++ b/api/base/authentication/drf.py
@@ -93,17 +93,22 @@ class OSFBasicAuthentication(BasicAuthentication):
 
 
 class OSFCASAuthentication(authentication.BaseAuthentication):
-    """Check whether the user provides a valid OAuth2 bearer token"""
 
     def authenticate(self, request):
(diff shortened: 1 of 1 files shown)


api/base/authentication/drf.py
```diff
@@ -112,9 +117,11 @@ def authenticate(self, request):
         if cas_auth_response.authenticated is False:
             raise exceptions.NotAuthenticated(_('CAS server failed to authenticate this token'))
 
-        user_id = cas_auth_response.user
-        user = User.load(user_id)
-        if user is None:
+        # TODO: @cslzchen use GUID when switching from OSF-Postgres-CAS to OSF-API-CAS
```
cslzchen: I choose a simple fix here sine `username` is an unique property.

However, this will be further improved when CAS talks to Database through OSF API instead of direct access.

PR #6748, opened 2017-01-19
[hotfix] Properly embed preprint url [OSF-7188]

alexschiller:
## Purpose

After the migration, guid routes changed for preprints. This PR fixes makes the preprints collection tab grab the working route to the preprint.

## Changes

Pull in preprint data to the nodes that have a preprint. Populate the route with that data.

## Side effects

Not a direct side effect of this PR, but a possible further improvement/bug:
Currently each line of the my projects page is populated from a single node that has a preprint. This does not check if the node has multiple preprints (e.g., more than one provider that has a preprint file). This means that only the first preprint value will be displayed here. It seemed sufficiently out of scope of this ticket to attempt a fix at that.

## Ticket
[#OSF-7188]

https://openscience.atlassian.net/browse/OSF-7188

## QA Considerations
1. Make sure that the link created properly goes to the preprint page (and not the node). 
2. Test both newly created preprints and migrated preprints of the various guid translation types (project kept the guid vs preprint kept the guid) if possible.

### DIFF
--- a/website/static/js/myProjects.js
+++ b/website/static/js/myProjects.js
@@ -359,7 +359,7 @@ var MyProjects = {
         self.systemCollections = options.systemCollections || [
             new LinkObject('collection', { nodeType : 'projects'}, 'All my projects'),
             new LinkObject('collection', { nodeType : 'registrations'}, 'All my registrations'),
-            new LinkObject('collection', { nodeType : 'preprints', link: $osf.apiV2Url('users/me/nodes/', { query : { 'filter[preprint]': true, 'related_counts' : 'children', 'embed' : 'contributors'}})}, 'All my preprints')
+            new LinkObject('collection', { nodeType : 'preprints', link: $osf.apiV2Url('users/me/nodes/', { query : { 'filter[preprint]': true, 'related_counts' : 'children', 'embed' : ['contributors', 'preprints']}})}, 'All my preprints')
         ];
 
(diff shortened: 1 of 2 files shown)


website/static/js/myProjects.js
```diff
@@ -359,7 +359,7 @@ var MyProjects = {
         self.systemCollections = options.systemCollections || [
             new LinkObject('collection', { nodeType : 'projects'}, 'All my projects'),
             new LinkObject('collection', { nodeType : 'registrations'}, 'All my registrations'),
-            new LinkObject('collection', { nodeType : 'preprints', link: $osf.apiV2Url('users/me/nodes/', { query : { 'filter[preprint]': true, 'related_counts' : 'children', 'embed' : 'contributors'}})}, 'All my preprints')
+            new LinkObject('collection', { nodeType : 'preprints', link: $osf.apiV2Url('users/me/nodes/', { query : { 'filter[preprint]': true, 'related_counts' : 'children', 'embed' : ['contributors', 'preprints']}})}, 'All my preprints')
```
sloria: Embeds are relatively expensive; can we simply build the URLs on the client-side?

sloria: Discussed in person--since we're following a pattern of using embeds, this is fine for now.

alexschiller: @sloria willdo.

PR #6749, opened 2017-01-20
[Hotfix] Update admin groups and permissions to allow more layers of access [OSF-7359]

erinspace:
## Purpose

Right now there is no way to grant access to the OSF Admin app without granting access to everything. This PR adds a few new groups and makes use of new permissions to instead grant a few layers of access:

basic - can log in and see OSF Metrics
read only - can view information about nodes, users, and meetings but cannot perform any actions
admin - can view and modify things
prereg - can view prereg submissions but not perform any actions
prereg_admin - can view and provide feedback on prereg submissions
superuser - has access to all groups and modifications

This will use a combination of groups and specific edit permissions for each user. The groups a user is in will define which pages they can visit, while if a user specifically is given the `auth.admin` or `auth.prereg_admin` permission will be able to make changes on those pages.

In summary:
- Those in the `osf_admin` group can log in and access metrics
- Those in the `nodes_and_users` group can visit Meetings, Node and User summary pages, but not make changes.
- Those in the `nodes_and_users` group AND whose users have the `auth.admin` permission can make changes 
- Those in the `prereg` group can visit the prereg submissions page
- Those in the `prereg` group AND whose users have the `auth.prereg_admin` permission can make changes on the prereg submissions.
- Superusers can do everything.

## Changes

- Migration that:
    - adds the `nodes_and_users` group 
    - changes the `prereg_group` name to `prereg`
- Use the `nodes_and_users` group to restrict access to Nodes and Users altogether
- Update edit views to check user permissions
- Update templates to check for users with specific permissions set. Users will need to have `auth.admin` set to change node, user and meeting information, and will need to have `auth.prereg_admin` set to change prereg information. See the side effects section for more info.
- Superusers can now access and modify all things in the admin app, including prereg.

## Side effects

1. Migrations will need to be run to add the new groups necessary.

2. Permissions are usually tied to a specific model, and in this case I didn't want to tie any permissions to `MyUser` because that is going away with the next OSF release. Instead I added some permissions via the admin interface, namely `auth.admin` and `auth.prereg_admin`. Those should be created manually with the admin app and then permissions added to the appropriate users also with the admin's admin app via an admin app superuser.

screen shot 2017-01-19 at 11 09 30 pm

## Ticket
https://openscience.atlassian.net/browse/OSF-7359

### DIFF
--- a/admin/base/migrations/0002_groups.py
+++ b/admin/base/migrations/0002_groups.py
@@ -24,6 +24,14 @@ def add_groups(*args):
             logger.info('prereg group created')
 
 
+def remove_groups(*args):
+    Group.objects.filter(name='nodes_and_users').delete()
+
+    group = Group.objects.get(name='prereg')
+    group.name = 'prereg_group'
+    group.save()
+
+
 class Migration(migrations.Migration):
 
     dependencies = [
@@ -31,5 +39,5 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RunPython(add_groups),
+        migrations.RunPython(add_groups, remove_groups),
     ]
--- a/admin/base/migrations/0002_groups.py
+++ b/admin/base/migrations/0002_groups.py
@@ -1,0 +1,35 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+from django.contrib.auth.models import Group
+import logging
+
+logger = logging.getLogger(__file__)
+
+
+def add_groups(*args):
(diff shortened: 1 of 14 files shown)


erinspace:
@cwisecarver totally, totally agree. As per discussion, going to to a major overhaul of permissions and groups in the changes I am making to make the admin module play well with DjangOSF.

cwisecarver:
lgtm

sloria:
Merged this into `develop-backup` and ran the migration there for QA purposes. Ping @erinspace and Amanda for next QA steps.

When QA gives the go ahead, I'll merge this into master.

sloria:
bitmoji

PR #6750, opened 2017-01-20
Exclude contribs when incrementing download count [#OSF-7360]

mfraezz:
## Purpose
Self-explanatory

## Changes
Compare strings to strings, not complex objects.

## Side effects
None expected

## Ticket
[#OSF-7360](https://openscience.atlassian.net/browse/OSF-7360)

### DIFF
--- a/osf/models/analytics.py
+++ b/osf/models/analytics.py
@@ -125,9 +125,7 @@ class PageCounter(BaseModel):
             # if the user who is downloading isn't a contributor to the project
             page_type = cleaned_page.split(':')[0]
             if page_type == 'download' and node_info:
-                contributors = node_info['contributors']
-                current_user = session.data.get('auth_user_id')
-                if current_user and current_user in contributors:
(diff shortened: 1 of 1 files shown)


osf/models/analytics.py
```diff
@@ -125,7 +125,7 @@ def update_counter(cls, page, node_info):
             # if the user who is downloading isn't a contributor to the project
             page_type = cleaned_page.split(':')[0]
             if page_type == 'download' and node_info:
-                contributors = node_info['contributors']
+                contributors = node_info['contributors'].values_list('guids___id', flat=True)
```
sloria: I believe you can use a query here; something like:

```python
current_user_id = session.data.get('auth_user_id')
if node_info['contributors'].filter(guids__id=current_user_id).exists():
     # ...
```

mfraezz: It does, updated.

PR #6752, opened 2017-01-20
[RFR] Limit Postgres connections outside of testing environment

cwisecarver:
## Purpose

pytest django handles database connections itself, as does django, flask does not handle django database connections. This adds database connection handling to flasks handler framework and disable this handling during unit tests. Based on `framework.mongo.handlers`. h/t @sloria @chrisseto 

## Changes

- Added handlers to to reset queries and close database connections during the flask request cycle
- Added rm_handlers and rm_handler following the `framework.flask.add_handlers` arch.
- Added handlers to the list of running flask handlers when `settings.USE_POSTGRES is True`
- Made rm_handlers run when unit tests are running

## Side effects

None observed during testing.


cwisecarver:
@sloria @chrisseto The tests are passing now. I did have to restart a couple of the test matrices a couple of times. They failed the first time. This doesn't give me a great deal of confidence. They run fine locally consistently though.

framework/flask/__init__.py
```diff
@@ -19,6 +19,30 @@
...
+    :param app: Flask app
+    :param handler_name: Name of handler type, e.g. 'before_request'
+    :param func: Handler function to attach
+    :param key: Blueprint name
+    """
+    handler_funcs_name = '{0}_funcs'.format(handler_name)
+    handler_funcs = getattr(app, handler_funcs_name)
+    try:
+        handler_funcs.get(key, []).remove(func)
+    except ValueError:
```
cwisecarver: @sloria I had to put this in because `rm_handlers` gets called repeatedly during unit tests and this exception makes all the tests fail.

PR #6766, opened 2017-01-23
[hotfix] Handle unicode in encryption [OSF-7200]

alexschiller:
## Purpose

Names on Dropbox and Google drive break when we try to encrypt them.

## Changes

1. Add helper function to ensure incoming strings to be encoded are in the proper utf-8 format
2. Add tests for helper function and encryption in general since it was untested.

## Side effects

None known

## Ticket
https://openscience.atlassian.net/browse/OSF-7200

[#OSF-7200]

## QA Considerations
1. Test auth on at least 1 addon. I did dropbox locally and it worked. Per MattF if it works for one, there is no reason it shouldn't work for all of them. Worth testing at least one or two for a sanity check, but probably not worth checking all of them (especially if you can't easily change your name on the service).
2. Change your name on <addon> to contain unicode madness and try to authorize the addon on the osf
3. Ensure it also works with non-unicode madness

### DIFF
--- a/framework/encryption/__init__.py
+++ b/framework/encryption/__init__.py
@@ -6,13 +6,22 @@ from website import settings
 
 SENSITIVE_DATA_KEY = jwe.kdf(settings.SENSITIVE_DATA_SECRET.encode('utf-8'), settings.SENSITIVE_DATA_SALT.encode('utf-8'))
 
+def ensure_str(value):
+    """Helper function to ensure all inputs are encoded to the proper value utf-8 value regardless of input type"""
+    try:
+        return value.encode('utf-8')
+    except UnicodeDecodeError:
+        return value.decode('utf-8').encode('utf-8')
+
 def encrypt(value):
     if value:
+        value = ensure_str(value)
         return jwe.encrypt(bytes(value), SENSITIVE_DATA_KEY)
     return None
 
 def decrypt(value):
     if value:
+        value = ensure_str(value)
         return jwe.decrypt(bytes(value), SENSITIVE_DATA_KEY)
     return None
 
--- a/framework/encryption/__init__.py
+++ b/framework/encryption/__init__.py
@@ -6,22 +6,21 @@ from website import settings
 
 SENSITIVE_DATA_KEY = jwe.kdf(settings.SENSITIVE_DATA_SECRET.encode('utf-8'), settings.SENSITIVE_DATA_SALT.encode('utf-8'))
 
-def ensure_str(value):
+def ensure_bytes(value):
     """Helper function to ensure all inputs are encoded to the proper value utf-8 value regardless of input type"""
-    try:
-        return value.encode('utf-8')
-    except UnicodeDecodeError:
-        return value.decode('utf-8').encode('utf-8')
+    if isinstance(value, bytes):
+        return value
+    return value.encode('utf-8')
 
 def encrypt(value):
     if value:
-        value = ensure_str(value)
+        value = ensure_bytes(value)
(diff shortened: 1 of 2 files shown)


sloria:
@alexschiller If this is a hotfix, can you please resend this to master? Can you also verify that this was branched off of master?

framework/encryption/__init__.py
```diff
@@ -6,13 +6,22 @@
 
 SENSITIVE_DATA_KEY = jwe.kdf(settings.SENSITIVE_DATA_SECRET.encode('utf-8'), settings.SENSITIVE_DATA_SALT.encode('utf-8'))
 
+def ensure_str(value):
```
chrisseto: Just for clarity you may want to name this `ensure_bytes`.
In python 2.7 `str == bytes` but that is often confused with `unicode`.
It's fairly difficult to determine what a string literal will end up being.

```python
isinstance(b'example text', bytes)  # True
isinstance(b'example text', str) # True
isinstance(b'example text', unicode) # False
```

```python
isinstance('example text', bytes)  # True
isinstance('example text', str) # True
isinstance('example text', unicode) # False
```

```python
isinstance(u'example text', bytes)  # False
isinstance(u'example text', str) # False
isinstance(u'example text', unicode) # True
```

```python
from __future__ import unicode_literals
isinstance('example text', bytes)  # False
isinstance('example text', str) # False
isinstance('example text', unicode) # True
```

```python
from __future__ import unicode_literals
isinstance(b'example text', bytes)  # True
isinstance(b'example text', str) # True
isinstance(b'example text', unicode) # False
```

But wait there's more

```python

In [43]: x = u'x'

In [44]: x.encode?
Docstring:
S.encode([encoding[,errors]]) -> string or unicode

Encodes S using the codec registered for encoding. encoding defaults
to the default encoding. errors may be given to set a different error
handling scheme. Default is 'strict' meaning that encoding errors raise
a UnicodeEncodeError. Other possible values are 'ignore', 'replace' and
'xmlcharrefreplace' as well as any other name registered with
codecs.register_error that can handle UnicodeEncodeErrors.
Type:      builtin_function_or_method

In [45]: x = b'x'

In [46]: x.encode?
Docstring:
S.encode([encoding[,errors]]) -> object

Encodes S using the codec registered for encoding. encoding defaults
to the default encoding. errors may be given to set a different error
handling scheme. Default is 'strict' meaning that encoding errors raise
a UnicodeEncodeError. Other possible values are 'ignore', 'replace' and
'xmlcharrefreplace' as well as any other name registered with
codecs.register_error that is able to handle UnicodeEncodeErrors.
Type:      builtin_function_or_method
```

`unicode.encode` -> `bytes`
`bytes.decode` -> `unicode`

I think it would be best to check if the value is already in bytes and `encode` it if not.

Final thought, throw some `assert isinstance`s into the tests to make sure that the value your attempting to send into this function is actually the type you think it is.


PS `# -*- coding: utf-8 -*-` Does weird stuff too. I don't remember what, however. 👍

cwisecarver: What I think @chrisseto is saying is that he's going to upgrade the osf to python 3 while he's in GA.

alexschiller: Popped back into AD and will take a look tomorrow when I bring my old machine that has *vintage* osf still running.

alexschiller: @chrisseto adding  `-*- coding: utf-8 -*-`seemed to be required for python not to get mad on parsing the tests. 🤷‍♂️

PR #6768, opened 2017-01-23
Add back necessary addon code for modm->django migration

sloria:
## Purpose

Add back necessary legacy code to make modm->django migrations work. These changes allow `init_app` to run when `USE_POSTGRES = False`.

## Changes

Adds:

- minimal set of config variables (e.g. `MODELS`, `SHORT_NAME`) to the legacy `website/<addon>/__init__.py` files
- init_addon
- a fix to ensure_licenses to make it run when USE_POSTGRES = False


sloria:
Not quite working; will send a new PR.

PR #6769, opened 2017-01-24
Add back necessary addon code for modm->django migration

sloria:
## Purpose

Add back necessary legacy code to make modm->django migrations work. These changes allow `init_app` to run when `USE_POSTGRES = False`.

## Changes

Adds:

- legacy addon code back to website/addons/
- init_addon
- a fix to ensure_licenses to make it run when USE_POSTGRES = False


cwisecarver:
@sloria I think you broke travis. It's been running for > 3hrs.

cwisecarver:
I restarted the build because I want to merge this.

PR #6770, opened 2017-01-24
[Feature] fix node contributors filter [OSF-7371]

alexschiller:
## Purpose
Filtering nodes by contributors is broken.

## Changes
Fixes filtering nodes by contributors

## Side effects
Changes filtering nodes by tags to icontains instead of iexact, adds this explicitly to all post processors for filters

## Ticket
[#OSF-7371]
https://openscience.atlassian.net/browse/OSF-7371

## QA Notes
On the api:
- Test filtering nodes by contributors guid exact match
`https://staging3-api.osf.io/v2/nodes/?filter[contributors]=ptfb7`
- Test filtering nodes by contributors guid non exact match does not work
`https://staging3-api.osf.io/v2/nodes/?filter[contributors]=ptf`
- Test that filtering nodes by tags works on the api
`https://staging3-api.osf.io/v2/nodes/?filter[tags]=sand`
- Test that filtering nodes by partial tags works on the api
`https://staging3-api.osf.io/v2/nodes/?filter[tags]=san`


api/nodes/views.py
```diff
@@ -1920,7 +1925,7 @@ def postprocess_query_param(self, key, field_name, operation):
         if field_name == 'tags':
             if operation['value'] not in (list(), tuple()):
                 operation['source_field_name'] = 'tags__name'
-                operation['op'] = 'iexact'
+                operation['op'] = 'icontains'
```
mfraezz: Why was this changed? I'm fairly certain that without changes to the model this will be egregiously slow, but benchmarking it would be a good idea. (There are [ways](https://docs.djangoproject.com/en/1.10/ref/contrib/postgres/search/) of making this faster, but the best way requires upgrading to Django 1.10)

Also, this seems out of scope. This ticket is for fixing node/contributor filtering, not node/tag filtering, and this isn't current prod behavior:

https://api.osf.io/v2/nodes/?filter[tags]=sci
https://api.osf.io/v2/nodes/?filter[tags]=science

PR #6774, opened 2017-01-25
Handle user error when patching licenses [#OSF-7383]

mfraezz:
## Purpose
Provide a useful error when `PATCH` an invalid `license`

## Changes
* Improve error handling
* Add regression tests

## Side effects
None expected

## Ticket
[#OSF-7383](https://openscience.atlassian.net/browse/OSF-7383)

### DIFF
--- a/api_tests/nodes/views/test_node_detail.py
+++ b/api_tests/nodes/views/test_node_detail.py
@@ -1197,6 +1197,21 @@ class TestNodeUpdateLicense(ApiTestCase):
     def make_request(self, url, data, auth=None, expect_errors=False):
         return self.app.patch_json_api(url, data, auth=auth, expect_errors=expect_errors)
 
+    def test_admin_update_license_with_invalid_id(self):
+        data = self.make_payload(
+            node_id=self.node._id,
+            license_id='thisisafakelicenseid'
+        )
+
+        assert_equal(self.node.node_license, None)
+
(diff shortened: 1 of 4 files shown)


api_tests/nodes/views/test_node_detail.py
```diff
@@ -1197,6 +1197,21 @@ def make_payload(self, node_id, license_id=None, license_year=None, copyright_ho
     def make_request(self, url, data, auth=None, expect_errors=False):
         return self.app.patch_json_api(url, data, auth=auth, expect_errors=expect_errors)
 
+    def test_admin_update_license_with_invalid_id(self):
```
mfraezz: Verified that this and the `preprints` version produced the expected error on `develop`:
```
api/nodes/serializers.py:99: in get_license_details
    license_id = validated_data['license_type'].license_id
E   AttributeError: 'unicode' object has no attribute 'license_id'
```

cwisecarver:
lgtm. Was this broken on master or did we lose something in the translation? Should this be a hotfix?

mfraezz:
>Was this broken on master

I haven't tested to see but I assume so, given that both of the [affected](https://github.com/CenterForOpenScience/osf.io/blob/master/api/nodes/serializers.py#L54-L56) [methods](https://github.com/CenterForOpenScience/osf.io/blob/master/api/preprints/serializers.py#L51-L53) are the same on `master`.

mfraezz:
Closing in favor of #6775

PR #6781, opened 2017-01-25
Fix Node.objects.get_children [OSF-7395]

cwisecarver:
## Purpose

Make node.objects.get_children stop killing the postgres server.

## Changes

Switch from recursive CTE to python.

## Side effects

Node links will not be followed or returned by this implementation. This was my understanding of the correct behavior. That makes it so that you can never create at circular loop that will run until maximum recursion depth.

Places where this is likely to affect:

- Anything in the API using node filters on root
- Probably the bulk change privacy for node and its children that was updated a year or so ago
- Showing orphaned children in the file viewer and component list
- Institution affiliation settings for projects

## Ticket
[#OSF-7395](https://openscience.atlassian.net/browse/OSF-7395)

### DIFF
--- a/osf/models/node.py
+++ b/osf/models/node.py
@@ -15,6 +15,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError
 from django.core.urlresolvers import reverse
 from django.db import connection, models, transaction
+from django.db.models import Model
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 from django.utils import timezone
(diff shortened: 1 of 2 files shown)


cwisecarver:
Dis? https://openscience.atlassian.net/browse/OSF-7395?filter=16800

cwisecarver:
I tried several ways to include Jira in on the secret of my pull request. It's not listening to me.

cwisecarver:
I see.

sloria:
Implementation looks correct.

Looks like the only place we use get_children is for filtering by `root` in APIv2. @brianjgeiger Do you know which OSF features do this? Would be good to know where QA should look for any performance problems.

cwisecarver:
These are also not intentionally ordered. So ... someone should say that's okay.

sloria:
Found one usage of `root` filtering: https://github.com/CenterForOpenScience/osf.io/blob/63669eab5fcc10e8ebd2d5353f32defc8ce52038/website/static/js/institutionProjectSettings.js#L12 , which is used on the Institutions affiliation settings on project "Settings" pages.

cwisecarver:
Wait for tests to pass and ask for forgiveness rather than permission?

brianjgeiger:
I've updated the PR description with things to test.

PR #6790, opened 2017-01-30
[feature fix] Prevent duplicates from appearing in node lists [OSF-7400]

sloria:
## Purpose

When a node is both a component of a project and linked to from
another project, the node would *sometimes* appear twice when
calling AbstractNode#get_nodes() on either the parent project or the
linking project.

## Changes

This filters out duplicates in `AbstractNode.get_nodes()`, but does so in Python, because there
is no way to both (1) ensure uniqueness of returned nodes and
(2) maintain ordering by NodeRelation._order in the same query.

## Side effects

- `AbstractNode.get_nodes()` and `AbstractNode.nodes` now return a list rather than a QuerySet. I verified that no application code treats the return values as a QuerySet.
- This is an ugly solution--I tried many different queries to do the filtering in SQL rather than in Python, but none produced the desired behavior 😞 
- The test added here is not sufficient to reproduce the issue. In other words, it still passes without the code changes. I was unable to reproduce the issue in an automated test; I could only reproduce it when creating projects manually 😞  . Here are the steps:
  - Create a project A and a component B
  - Make B public
  - Create a separate public project C.
  - Create a separate project D.
  - Add link from project D to component B.
  - Add link from project D to component D.
  - On project D, reorder the nodes so that D appears before B.
  - View project D and project A.

Expected: component B appears once
Actual: component B appears twice

## Ticket

https://openscience.atlassian.net/browse/OSF-7400

### DIFF
--- a/api_tests/nodes/views/test_node_children_list.py
+++ b/api_tests/nodes/views/test_node_children_list.py
@@ -158,7 +158,7 @@ class TestNodeChildCreate(ApiTestCase):
         assert_equal(res.status_code, 401)
 
         self.project.reload()
-        assert_equal(self.project.nodes.count(), 0)
+        assert_equal(len(self.project.nodes), 0)
 
     def test_creates_child_logged_in_owner(self):
         res = self.app.post_json_api(self.url, self.child, auth=self.user.auth)
@@ -168,8 +168,8 @@ class TestNodeChildCreate(ApiTestCase):
         assert_equal(res.json['data']['attributes']['category'], self.child['data']['attributes']['category'])
 
         self.project.reload()
-        assert_equal(res.json['data']['id'], self.project.nodes.first()._id)
-        assert_equal(self.project.nodes.first().logs.latest().action, NodeLog.PROJECT_CREATED)
+        assert_equal(res.json['data']['id'], self.project.nodes[0]._id)
+        assert_equal(self.project.nodes[0].logs.latest().action, NodeLog.PROJECT_CREATED)
 
     def test_creates_child_logged_in_write_contributor(self):
         self.project.add_contributor(self.user_two, permissions=[permissions.READ, permissions.WRITE], auth=Auth(self.user), save=True)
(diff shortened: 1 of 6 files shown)


sloria:
Here were the [failed] implementations I tried. Most of them still returned duplicates. Doing an aggregation filters out duplicates, but it is incorrect: the correct order will not be maintained.


```python
AbstractNode.objects.filter(id__in=child_ids).order_by('id', 'noderelation___order')
# [<Node: B : (s72xf)>, <Node: B : (s72xf)>]
```

```python
AbstractNode.objects.filter(id__in=child_ids).order_by('noderelation__id', 'noderelation___order')
# [<Node: B : (s72xf)>, <Node: B : (s72xf)>]
```

```python
AbstractNode.objects.filter(id__in=child_ids).order_by('noderelation__parent', 'noderelation___order')
# [<Node: B : (s72xf)>, <Node: B : (s72xf)>]
```

```python
AbstractNode.objects.filter(id__in=nr_ids).order_by('noderelation__parent', 'noderelation__child', 'noderelation___order')
# [<Node: B : (s72xf)>, <Node: B : (s72xf)>]
```

```python
# http://django.cowhite.com/blog/ordering-by-relational-field-without-duplicates/
AbstractNode.objects.filter(id__in=child_ids).annotate(order_nr=Max('noderelation___order')).order_by('order_nr')
# [<Node: B : (s72xf)>]
# This succesfully filters duplicates, but the node orders will be incorrect
```

```python
AbstractNode.objects.filter(id__in=child_ids).order_by('noderelation___order').distinct('noderelation___order', 'id')
# [<Node: B : (s72xf)>, <Node: B : (s72xf)>]
```

```python
list(AbstractNode.objects.raw('''
SELECT "osf_abstractnode"."id"
FROM "osf_abstractnode"
LEFT OUTER JOIN "osf_noderelation"
ON ("osf_abstractnode"."id" = "osf_noderelation"."child_id")
WHERE "osf_abstractnode"."id" IN %s
GROUP BY "osf_abstractnode"."id", "osf_noderelation"."_order"
ORDER BY "osf_noderelation"."_order" ASC''', [tuple(child_ids)]))
# [<Node: B : (s72xf)>, <Node: B : (s72xf)>]
```

```python
list(AbstractNode.objects.raw('''
SELECT "osf_abstractnode"."id"
FROM "osf_abstractnode"
LEFT OUTER JOIN "osf_noderelation"
ON ("osf_abstractnode"."id" = "osf_noderelation"."child_id")
WHERE "osf_abstractnode"."id" IN %s
GROUP BY "osf_noderelation"."_order", "osf_abstractnode"."id"
ORDER BY "osf_noderelation"."_order" ASC
''', [tuple(child_ids)]))
# [<Node: Child : (np9jk)>, <Node: Child : (np9jk)>]
```

@cwisecarver Let me know if you have any other ideas.

PR #6791, opened 2017-01-30
[feature fix] Don't show logs from linked nodes [OSF-7378]

sloria:
## Purpose

Prevent logs from linked nodes from appearing. 

## Changes

- Pass `primary_only=True` when querying for a node's logs
- Passing a function to get_descendants_recursive was removed.
This fixes a few calls to use the newer primary_only parameter

## Ticket

Comment in https://openscience.atlassian.net/browse/OSF-7378


sloria:
Turns out this approach might be incorrect; looking into this a bit deeper.

PR #6793, opened 2017-01-30
[OSF-7411] Branded Preprints Phase 2

binoculars:
Companion to CenterForOpenScience/ember-preprints#201

❗️ Needs: #6827 for production, but this PR can be tested separately

Documentation update: CenterForOpenScience/developer.osf.io#46

## Purpose

* Adds support for domains in branded preprints
* Proxies `ember-cli-live-reload.js` asset for live reload capability

## Changes
* Updates the preprint provider serializer, model, and views with new `domain` field
* Migration script for the new field
* Adds a 301 redirect for GUIDs whose provider has a domain. If the provider is OSF or the provider doesn't have a domain, the index.html is served (like before)
* Local prefix/suffix for development (See changes in CenterForOpenScience/ember-preprints#201)

## Side effects

N/A

## Notes for Reviewers & Deployment
Domains are currently disabled, so this can be merged without affecting anything. Enable this, set two change are required:

### Enable providers in `website/settings/local.py`
```python
PREPRINT_PROVIDER_DOMAINS = {
    'enabled': True,
    'prefix': 'http://local.',
    'suffix': ':4200'
}
```

### For testing with preprints, In `docker-compose.override.yml`
```yml
  preprints:
    volumes_from:
      - container:preprints-sync
    environment:
      - ENABLE_PROVIDER_DOMAINS=true
```

## Ticket

https://openscience.atlassian.net/browse/OSF-7411


website/settings/local-dist.py
```diff
@@ -22,6 +22,11 @@
 #WATERBUTLER_URL = 'http://localhost:7777'
 #WATERBUTLER_INTERNAL_URL = WATERBUTLER_URL
 
+LIVE_RELOAD_DOMAIN = 'http://localhost:4200'
```
abought: With Retraction Watch and Registries, we're very close to having multiple `EXTERNAL_EMBER_APPS`.

How will this handle that scenario- do we just advise developers to change the `LIVE_RELOAD_DOMAIN` setting based on their current project?

(Documenting `EXTERNAL_EMBER_APPS` is outside scope of this PR, but it's worth thinking ahead about this feature. It would be great if we could avoid having to break dev setups that depend on this feature later)

website/routes.py
```diff
@@ -1705,7 +1705,15 @@ def make_url_map(app):
...
+
         @app.route('/static/addons/<addon>/<path:filename>')
         def addon_static(addon, filename):
             addon_path = os.path.join(addon_base_path, addon, 'static')
             return send_from_directory(addon_path, filename)
+
+        @app.route('/ember-cli-live-reload.js')
+        def ember_cli_live_reload():
+            req = requests.get('{}/ember-cli-live-reload.js'.format(settings.LIVE_RELOAD_DOMAIN), stream=True)
+            return Response(stream_with_context(req.iter_content()), content_type=req.headers['content-type'])
```
abought: I apologize for not trying this myself yet (my docker setup needs rebuilding again). 

Normally I think the local dev mode flask app runs on a single thread, which would imply that streaming this response would block other requests. Can you confirm that someone using this feature would be able to, say, view an OSF project locally while a tab with `/preprints` was using this to monitor livereload?

osf/models/preprint_service.py
```diff
@@ -74,11 +74,25 @@ def deep_url(self):
...
+            domain_settings = settings.PREPRINT_PROVIDER_DOMAINS
+            return ''.join((domain_settings['prefix'], self.provider.domain, domain_settings['suffix']))
+    else:
+        def get_provider_domain(self):
+            return settings.PROTOCOL + self.provider.domain
 
     @property
     def absolute_url(self):
-        return urlparse.urljoin(settings.DOMAIN, self.url)
+        use_osf_domain = self.provider._id == 'osf' or not hasattr(self.provider, 'domain')
```
abought: In this brave new Django + migrations world, under what circumstances would a provider not have a `domain` attribute?

website/settings/local-dist.py
binoculars: The live reload asset from the root path can only be served from one place at a time. It's possible to run ember serve on a different port then adjust the port here in local.py. This is overridden by the environment variable with Docker anyhow.

website/routes.py
binoculars: This is just a proxy to the live reload javascript asset. The asset itself points the client to the correct host with the actual live reload websocket requests.

binoculars: This is just a proxy to the live reload javascript asset. The asset itself points the client to the correct host with the actual live reload websocket requests.

website/views.py
```diff
@@ -214,10 +215,10 @@ def resolve_guid(guid, suffix=None):
         if not referent.deep_url:
             raise HTTPError(http.NOT_FOUND)
         if isinstance(referent, PreprintService):
-            return send_from_directory(
-                os.path.abspath(os.path.join(os.getcwd(), EXTERNAL_EMBER_APPS['preprints']['path'])),
-                'index.html'
-            )
+            if referent.provider._id == 'osf' or not referent.provider.domain:
```
abought: I'm not 100% clear how time-sensitive this PR is, but GUIDs are complex, and this code adds some new special cases. Normally, we would encourage unit tests for such functionality as time permits. 

Let me know if I can be of assistance in finding the right place to test that and move forward. I might consider some tests on the model (where url behavior is conditional), but that's not as high a priority as testing the GUID functionality.

api/preprint_providers/views.py
```diff
@@ -28,12 +28,21 @@ class PreprintProviderList(JSONAPIBaseView, generics.ListAPIView, ODMFilterMixin
...
 
-        name           type               description
-        =========================================================================
-        name           string             name of the preprint provider
-        logo_path      string             a path to the preprint provider's static logo
-        banner_path    string             a path to the preprint provider's banner
-        description    string             description of the preprint provider
+        name                 type                description
+        ==========================================================================================================
+        name                 string              name of the preprint provider
```
abought: Thanks for keeping this documentation up to date- having chased down "ghost fields" before, this is hugely appreciated!!

osf/migrations/0030_preprintprovider_domain.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
abought: Hurray for migrations!!

One request: typically when a PR calls for a migration, we call it out very explicitly as a separate PR section, and arrange for the migration to be tracked in the go/nogo ticket. Basically, the goal is to 100% ensure we take care of it. 

It's possible that policy has changed since I last checked, but repo maintainer can advise on current preference.

osf/models/preprint_provider.py
```diff
@@ -21,6 +21,7 @@ class PreprintProvider(ObjectIDMixin, BaseModel):
     header_text = models.TextField(default='', blank=True)
     description = models.CharField(null=True, blank=True, max_length=256)  # max length on prod: 56
     banner_name = models.CharField(null=True, blank=True, max_length=128)  # max length on prod: 19
+    domain = models.CharField(null=True, blank=True, max_length=253)  #
```
abought: I actually had to look this one up; nice catch. For future generations, this was informative:

http://blog.sacaluta.com/2011/12/dns-domain-names-253-or-255-bytesoctets.html

osf/models/preprint_service.py
binoculars: When the provider doesn't have a domain (yet). This is preserving the current functionality of having /preprints/:provider

website/views.py
binoculars: I could definitely use some help here, as I haven't found the tests for GUID views

osf/migrations/0030_preprintprovider_domain.py
binoculars: I was told to put it in the PR comments and it will go out with the release notes when the release happens

osf/models/preprint_service.py
abought: Hruh, interesting. I assumed it would have the attribute in that case, but with a null value. Ok then.

(`hasattr` checks are something we used to do more often with schemaless data, where mongo sometimes literally did not have the field)

website/views.py
abought: That's a fair question! 

In truth, GUIDs (and related tests) are wound through many parts of the OSF, which makes it hard to find a single obvious place to put the test.

Here are some general GUID tests:
https://github.com/abought/osf.io/blob/5ad6ebe27e63649213801cee9c395f665fb84056/tests/test_guids.py#L36-L36

Or sometimes GUID-related functionality will get tested with the module for that specific feature. Here are a few examples; there are 279 total references in the `tests` folder, but it's not at all centralized.
Addons: https://github.com/abought/osf.io/blob/33eff35d7fadc322dc4fcd1f9c885e48366a9681/tests/test_addons.py#L656-L656

Comments: https://github.com/abought/osf.io/blob/41484ecfd66a69ebaeeeb5db8183867c28c896d1/tests/test_comments.py#L137-L137

Models: https://github.com/abought/osf.io/blob/e17666d666beb60dbabfb4eaf652efe523300460/tests/test_models.py#L1051-L1051

binoculars: Latest commits now test for this case. Thanks @abought and @pattisdr

osf/models/preprint_service.py
```diff
@@ -74,11 +74,25 @@ def deep_url(self):
 
     @property
     def url(self):
-        return '/{}/'.format(self._id)
+        if self.provider._id == 'osf' or self.provider.domain:
+            return '/{}/'.format(self._id)
+
+        return '/preprints/{}/{}/'.format(self.provider._id, self._id)
+
+    if settings.DEV_MODE:
```
cslzchen: All staging environments are using `DEV_MODE`. May revisit this part when deploying.

brianjgeiger:
Some failing tests

```
________ TestViewsAreAtomic.test_error_response_rolls_back_transaction _________
[gw0] linux2 -- Python 2.7.9 /home/travis/virtualenv/python2.7.9/bin/python
tests/test_views.py:128: in test_error_response_rolls_back_transaction
    assert_equal(User.objects.count(), 0)
E   AssertionError: 1 != 0
__________ TestAuthLogout.test_logout_with_invalid_next_url_logged_in __________
[gw3] linux2 -- Python 2.7.9 /home/travis/virtualenv/python2.7.9/bin/python
tests/test_views.py:3916: in test_logout_with_invalid_next_url_logged_in
    assert_equal(cas.get_logout_url(self.goodbye_url), resp.headers['Location'])
E   AssertionError: 'http://localhost:8080/logout?service=http://localhost:5000/goodbye/' != 'http://localhost:8080/logout?service=http://localhost:5000/logout/?next_url=http%253A%252F%252Flocalhost%253A1234%252Fabcde'
_________ TestAuthLogout.test_logout_with_invalid_next_url_logged_out __________
[gw3] linux2 -- Python 2.7.9 /home/travis/virtualenv/python2.7.9/bin/python
tests/test_views.py:3922: in test_logout_with_invalid_next_url_logged_out
    assert_equal(cas.get_logout_url(self.goodbye_url), resp.headers['Location'])
E   AssertionError: 'http://localhost:8080/logout?service=http://localhost:5000/goodbye/' != 'http://localhost:1234/abcde'
============= 3 failed, 1225 passed, 19 skipped in 609.61 seconds ==============
LOG:  unexpected EOF on client connection with an open transaction
LOG:  unexpected EOF on client connection with an open transaction
```

cslzchen:
@brianjgeiger This failure is caused by my test/code. My own travis fails with the same errors, but local tests passed. I am looking into this.

api/preprint_providers/views.py
abought: Since this PR was started, `developer.osf.io` has come into existence. Will this documentation need to be captured in the new source of record?

binoculars: It will. Good catch. I will need another PR against that repo.

binoculars: CenterForOpenScience/developer.osf.io#46 is RFR

cslzchen:
As suspected, it fails again with the following error:
```
    ________ TestViewsAreAtomic.test_error_response_rolls_back_transaction _________
    [gw0] linux2 -- Python 2.7.9 /home/travis/virtualenv/python2.7.9/bin/python
    tests/test_views.py:128: in test_error_response_rolls_back_transaction
        assert_equal(User.objects.count(), 0)
    E   AssertionError: 1 != 0
    ============= 1 failed, 1227 passed, 19 skipped in 839.41 seconds ==============
```
It is probably due to Travis Cache. Here is the passing from my own Travis after cache cleared: https://travis-ci.org/cslzchen/osf.io/builds/211043064.

osf/models/preprint_service.py
cslzchen: @brianjgeiger @binoculars How about we add a `LOCAL_MODE` in settings. In this case, the staging and production environment can both use the `else` statement below.

brianjgeiger: If we add another setting that sounds semantically the same as DEV_MODE, that will cause confusion. My preference would be not to require a new mode, but if we do, then it should be something like "ENABLE_PROVIDER_DOMAINS".

cslzchen:
@brianjgeiger For the `CORS_ORIGIN_WHITELIST`, do we want to code it in the settings, which will be different for each environment. Or we can build it dynamically when the app starts, which is similar to how we handle campaigns?

brianjgeiger:
There will need to be a dynamic component to the whitelist that can be updated with the database. If we can't start with that, then on launch is fine. It should be based on the domains attached to the preprint_provider model.

binoculars:
The front-end will accept anything with the provider ID, e.g. `socarxiv`, provider domain e.g. `socarxiv.org`, or the provider domain with dashes instead of periods, e.g. `socarxiv-org`.

All of the following domains would render socarxiv:
* socarxiv.staging.osf.io
* socarxiv.org.staging.osf.io
* socarxiv-org.staging.osf.io
* socarxiv.org

brianjgeiger:
Since anything that works with domains is going to have to work across OSF, possibly CAS, and the front-end, should we just make the domains that should be used explicit and stored on the preprint providers rather than fuzzy? @mattclark @cslzchen @binoculars

cslzchen:
Yep. We do have a dedicated field in the preprint provider model for this `domain` and we build full URL from it. We will also dynamically build the whitelists from the database as well. However, we cannot continue further until the exact `domain` for these providers on each environment are finalized.

binoculars:
The bottleneck is really just the naming scheme for staging and test. Local and prod are ready to go.

osf/models/preprint_service.py
cslzchen: FYI, it is much simpler than I thought. No need for checking settings. Staging and Prod will have the same code. Only need to change `local.py` for local development.

brianjgeiger:
But why are we doing extrapolations of the domains instead of just setting them in the DB per environment? Is it just because it's not easy to set right now?

osf/models/preprint_provider.py
cwisecarver: If this will be a fully qualified url django has a [URLField](https://docs.djangoproject.com/en/1.11/ref/models/fields/#urlfield) that will do validation for you.

scripts/populate_preprint_providers.py
```diff
@@ -65,16 +65,14 @@ def update_or_create(provider_data):
         return new_provider, True
 
 def main(env):
-    use_plos = '--plos' in sys.argv
-
-
     PREPRINT_PROVIDERS = {
```
cwisecarver: This should eventually be moved to either a [django fixture](https://docs.djangoproject.com/en/1.10/howto/initial-data/) or a django migration that uses the loaddata functionality.

website/preprints/model.py
```diff
@@ -63,15 +64,25 @@ def deep_url(self):
...
-        if self.provider._id != 'osf':
-            # Note that this will change with Phase 2 of branded preprints.
-            return '/preprints/{}/{}/'.format(self.provider._id, self._id)
+        if self.provider._id == 'osf' or self.provider.domain:
+            return '/{}/'.format(self._id)
 
-        return '/{}/'.format(self._id)
+        return '/preprints/{}/{}/'.format(self.provider._id, self._id)
+
+    if settings.DEV_MODE:
```
cwisecarver: Why put the conditional around the function instead of inside it?

website/views.py
```diff
@@ -21,11 +21,12 @@
...
 from website.models import Guid
 from website.models import Institution, PreprintService
-from website.settings import EXTERNAL_EMBER_APPS
+from website.settings import EXTERNAL_EMBER_APPS, PREPRINT_PROVIDER_DOMAINS
 from website.project.model import has_anonymous_link
 
 logger = logging.getLogger(__name__)
-
+preprints_dir = os.path.abspath(os.path.join(os.getcwd(), EXTERNAL_EMBER_APPS['preprints']['path']))
+domains_disabled = not PREPRINT_PROVIDER_DOMAINS['enabled']
```
cwisecarver: This is already in the model, why not check that instead of redefining it here?

binoculars:
Just refactored to make everything based on the populate_preprint_providers script, and the local.py controls the prefix and suffix for the domains. No more DEV_MODE or anything. If you want to change the domain for the redirects, just add it to the domain field in the database. This is also now a full url field. 

@brianjgeiger @mattclark

brianjgeiger:
Not an official review note, but some things that may turn into changes as we get this running. I'll update this comment as I think of things.

- [ ] We have to ensure that we can handle a state where we accepts requests for the branded domain but do not redirect requests to osf/guid or osf/preprints/provider to provider_domain. This is to handle the case of "they are supposed to have their domain pointing soon but we don't know exactly when and we haven't verified it works yet" or, ultimately, "they lost, hopefully temporarily, control of their domain."
- [ ] It looks like we have to maintain three separate lists of the domains: inside the database, in the config for CORS, and inside nginx for the initial "send requests for these domains to preprints". I'd like to have a task that runs on deployment that will build the CORS config from what's stored in the DB (which isn't ideal but), and I don't know for sure if we're going to keep the nginx redirect or go to a slower OSF app handling of those URLs. If we keep with the nginx thing, I'd *like* a way to do like I suggested with the CORS where we can just bounce the app and have it rebuild from the currently active database.
- [ ] But what I'd really like is to not have to bounce the app every time we update a provider, but from what Michael's said, that's somewhat tricky right now, so that may be a thing for the Django/LE team to deal with some time in the future.

binoculars:
If you don't want the redirects on for a single provider, all you have to do is put a null value in for the domain field in the database. Nginx and the ember-preprints app will still handle the rendering.

The DNS shouldn't be an issue for us to monitor. If the provider's A record doesn't point to our IP address, the page won't load. There's no actual redirect. It's just resolving the domain.

We can probably just run a query to gather the list of allowed preprint domains for CORS on app startup, but I wouldn't run this for every request. That would kill the DB and the list rarely changes.

cwisecarver:
Thank you for thinking of our database @binoculars 😍

brianjgeiger:
Definitely not every request. Michael is pretty sure we can't handle CORS dynamically in any case currently (if I've understood properly), but even when we get it all the way handled, I'd cache the info and only update it whenever we either start up or update a preprint provider.

brianjgeiger:
If we populate nginx from the db somehow, then we have to have another way of differentiating the trusted domains from the less trusted ones. I'm not particularly advocating for a solution, as long as a) it's handled; and b) it works with the other optimizations we want to do.

brianjgeiger:
CORS?

binoculars:
Updated for CORS

api/preprint_providers/serializers.py
```diff
@@ -19,6 +19,7 @@ class PreprintProviderSerializer(JSONAPISerializer):
     email_contact = ser.CharField(required=False, allow_null=True)
     email_support = ser.CharField(required=False, allow_null=True)
     example = ser.CharField(required=False, allow_null=True)
+    domain = ser.CharField(required=False, allow_null=True)
```
cwisecarver: You probably don't want this allow null anymore.

website/templates/public/register.mako
```diff
@@ -109,7 +109,7 @@
                 %elif campaign == "psyarxiv-preprints":
                     <table style="border-collapse: separate; border-spacing: 30px 0; margin-top: 20px;  margin-bottom: 10px;">
                         <tr>
-                            <td><img src="/static/img/preprint_providers/psyarxiv-login.png" style="width: 150px; background: #a9a9a9; padding: 0 10px 0 10px;" /></td>
+                            <td><img src="/static/img/preprint_providers/psyarxiv-login.png" style="width: 150px; background: #062F4F; padding: 0 10px 0 10px;" /></td>
```
cwisecarver: Inline css. Gross.

binoculars: Blame @cslzchen

cwisecarver: Inline css is the devil @cslzchen. You can't override it outside the html document, not even with `!important`.

api/preprint_providers/serializers.py
binoculars: Good catch here--also forgot to add the `domain_redirect_enabled` field, but fixed with the latest commit and updated views.py to match.

cwisecarver: :high-five:

website/templates/public/register.mako
cslzchen: @binoculars 😞 😠 😡 
@cwisecarver 👍  The devil it is.
We made the decision that this page will be gone "soon" so I just patch it without making changes.

PR #6802, opened 2017-01-31
Add permissions to django-updated OSF Admin app [OSF-7417]

erinspace:
## Purpose

The final (hopefully) update in a long string of OSF Admin updates getting it all ready for DjangOSF. This one redoes the admin permissions system to better make use of Django's permissions and groups.

This PR takes some pre-existing groups and permissions, changes some names, and hopefully makes some logical sense.

Groups:

`read_only`:
- view_node
- view_user
- view_conference
- view_spam
- view_metrics
- view_desk

`osf_admin`
All of the above plus:
- change_node
- delete_node
- change_user
- change_conference
- mark_spam

`prereg_admin`
- view_prereg
- administer_prereg

`prereg_view`
- view_prereg

`metrics_only`
- view_metrics

## Changes

- Changes to templates checking for specific permissions instead of group membership
- Changes to views checking for permissions
- Remove groups for auth mixins
- Migration to create groups with permissions
- Migration to create special view permissions for some OSF models
- Change's admin's password hashing to match the OSF's
- Update the API's (and the Admin's) `ODMBackend` to inheret from Django's `ModelBackend`
 - Add `blank=True` to Institutions and External Accounts for OSFUsers so that permissions can successfully be added via the admin

## Side effects
1. Migrations will need to be run to add new groups along with permissions in those new groups. There is also a migration to add special view permissions to the OSF models that need them, mainly in Node, Registration, Conference, and field updates for OSFUser

2. *NONE* of the old OSF Admin Accounts will work any longer! We'll have to instead mark each of the OSF Admin users' OSF accounts through the admin app. Sara can still do this, as she's a superuser. 

# How to test locally!!
pre-setup - run all migrations and whatnot
1. Create a superuser
- open the shell with `docker-compose run --rm web python manage.py shell_plus` and get your user
- set `is_superuser` and `is_staff` to True and save
2. log in to the osf admin with that superuser
3. upgrade different already existing OSF users with http://localhost:8001/account/register/
- Select their permissions, mentioned above

notes:
- metrics might not work if you don't have API keys locally, but this PR does not deal with these as the're js only and not python

## Ticket
https://openscience.atlassian.net/browse/OSF-7417


admin/meetings/forms.py
```diff
@@ -37,10 +48,9 @@ class MeetingForm(forms.Form):
         required=True,
         widget=forms.TextInput(attrs={'size': '40'}),
     )
-    info_url = forms.CharField(
+    info_url = forms.URLField(
         label='Info url',
-        required=False,
-        widget=forms.TextInput(attrs={'size': '60'}),
+        required=True
```
mfraezz: Why'd this go from not required to required?

admin/nodes/serializers.py
```diff
@@ -24,7 +24,7 @@ def serialize_node(node):
         'children': map(serialize_simple_node, node.nodes),
         'deleted': node.is_deleted,
         'pending_registration': node.is_pending_registration,
-        'creator': node.creator._id,
+        'creator': node.creator._id if node.creator else None,
```
mfraezz: Minor, but shouldn't a `node` always have a `creator`?

admin/users/views.py
```diff
@@ -90,21 +91,20 @@ def delete(self, request, *args, **kwargs):
...
-        context.setdefault('guid', kwargs.get('object').pk)
+        context.setdefault('guid', kwargs.get('object')._id)
         return super(UserDeleteView, self).get_context_data(**context)
 
     def get_object(self, queryset=None):
         return OSFUser.load(self.kwargs.get('guid'))
 
 
-class SpamUserDeleteView(UserDeleteView, PermissionRequiredMixin):
+class SpamUserDeleteView(UserDeleteView):
```
mfraezz: Should this class not have any required permission?

admin/nodes/serializers.py
erinspace: You're right! I added this cause I had some bad data in my DB I was trying to account for. Will fix :star:

admin/meetings/forms.py
erinspace: Oh hmm good Q. I thought it was required in the OSF Conferences model, but I was mistaken there. I'll revert!

admin/users/views.py
erinspace: It inherets from the `UserDeleteView`, which has the `PermissionRequiredMixin` and requires the `osf.change_user` permission already. Though I do suppose this view in particular should also require the `osf.delete_user` permission as well.

erinspace: On second thought, not gonna include those extra permissions I don't think cause it's not really a delete, persay, but rather a disable. So should just stick with `DeleteUser`'s `osf.change_user` permission

erinspace:
@mfraezz updated and ready!

 bitmoji

mfraezz:
LGTM. RTM when Travis is happy.

bitmoji

PR #6784, opened 2017-01-26
[WIP] Stopping point; Included guids on subqueries, some tests still failin…

cwisecarver:
…g on updates

PR #6803, opened 2017-01-31
[Feature] Optimize GUID queries

mfraezz:
## Purpose
Extends #6784 

## Changes
* Override Django's `_for_write` to conditionally add or remove query annotations
* Ensure `GuidMixin` models have correct `_id` after cloning
* Add `.subselect` manager for subqueries that don't require annotation

## Side effects
None Expected

## Ticket
None known

### DIFF
--- a/osf/models/base.py
+++ b/osf/models/base.py
@@ -47,13 +47,6 @@ def generate_object_id():
     return str(bson.ObjectId())
 
 class MODMCompatibilityQuerySet(models.QuerySet):
-    def __init__(self, model=None, query=None, using=None, hints=None):
-        super(MODMCompatibilityQuerySet, self).__init__(model=model, query=query, using=using, hints=hints)
-        if issubclass(self.model, (GuidMixin)):
-            # import ipdb; ipdb.set_trace()
-            if not self._for_write:
-                # automatically include guids___id for models that have a guid
-                self.query.add_annotation(F('guids___id'), 'guids___id', is_summary=False)
 
     def __getitem__(self, k):
         item = super(MODMCompatibilityQuerySet, self).__getitem__(k)
@@ -88,6 +81,27 @@ class MODMCompatibilityQuerySet(models.QuerySet):
 
     def limit(self, n):
         return self[:n]
+
+
+class GuidMODMCompatibilityQuerySet(MODMCompatibilityQuerySet):
+    """ Hijacks Django internals, adding annotations to read queries
+        to limit the number of DB calls when serializing objects with Guids.
+    """
+
+    def __init__(self, model=None, query=None, using=None, hints=None):
+        self.__for_write = False
+        return super(GuidMODMCompatibilityQuerySet, self).__init__(model=model, query=query, using=using, hints=hints)
+
+    @property
+    def _for_write(self):
+        return self.__for_write
+
+    @_for_write.setter
+    def _for_write(self, value):
+        self.query.annotations.pop('guids___id', None)
+        if not value:
+            self.query.add_annotation(F('guids___id'), 'guids___id', is_summary=False)
+        self.__for_write = value
 
 
 class BaseModel(models.Model):
@@ -177,12 +191,6 @@ class BaseModel(models.Model):
     def _primary_name(self):
         return '_id'
 
-    def reload(self):
-        return self.refresh_from_db()
-
-    def _natural_key(self):
-        return self.pk
-
     def clone(self):
         """Create a new, unsaved copy of this object."""
         copy = self.__class__.objects.get(pk=self.pk)
@@ -193,11 +201,18 @@ class BaseModel(models.Model):
         for field_name in fk_field_names:
             setattr(copy, field_name, None)
 
-        try:
-            copy._id = bson.ObjectId()
-        except AttributeError:
-            pass
+        return self._on_clone(copy)
+
+    def _on_clone(self, copy):
+        """ Override hook to add subclass-specific clone behavior
+        """
         return copy
+
+    def reload(self):
+        return self.refresh_from_db()
+
+    def _natural_key(self):
+        return self.pk
 
     def save(self, *args, **kwargs):
         # Make Django validate on save (like modm)
@@ -378,6 +393,12 @@ class ObjectIDMixin(BaseIDMixin):
     def _natural_key(self):
         return self._id
 
+    def _on_clone(self, copy):
+        try:
+            copy._id = bson.ObjectId()
+        except AttributeError:
+            pass
+        return copy
 
 class InvalidGuid(Exception):
     pass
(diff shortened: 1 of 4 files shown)


addons/wiki/models.py
```diff
@@ -244,6 +244,8 @@ def clone_wiki(self, node_id):
         clone.node = node
         clone.user = self.user
         clone.save()
+        # Find and set Guid created during clone/save
+        clone._id = clone.guids.first()._id
```
sloria: Rather than setting this in multiple places, you might be able to add this to the `ensure_guid` function, which runs after every save: https://github.com/CenterForOpenScience/osf.io/blob/c67b78a0d2b3b08c6cc228f5bbff6ed144dc843d/osf/models/base.py#L519

osf/utils/fields.py
```diff
@@ -12,7 +12,7 @@ class LowercaseCharField(models.CharField):
     def get_prep_value(self, value):
         value = super(models.CharField, self).get_prep_value(value)
         if value is not None:
-            value = value.lower()
+            value = str(value).lower()
```
sloria: Does this run before validation happens? If so, then we run the risk of putting bad data in the database. For example, if an `int` is passed to this field, it will silently be cast to a `str`.

website/prereg/views.py
```diff
@@ -24,7 +24,7 @@ def prereg_landing_page(auth, **kwargs):
         if node.has_permission(user=auth.user, permission='admin')
     ]
     has_projects = bool(registerable_nodes)
-    has_draft_registrations = bool(utils.drafts_for_user(auth.user, campaign).count())
+    has_draft_registrations = utils.drafts_for_user(auth.user, campaign).exists()
```
sloria: Good catch.

osf/utils/fields.py
cwisecarver: It's the last thing that happens before the SQL is written.

sloria: Ok. I see now that `super()` is called before this line, so I assume invalid types are handled there?

osf/models/base.py
```diff
@@ -86,14 +83,31 @@ def limit(self, n):
...
+class GuidMODMCompatibilityQuerySet(MODMCompatibilityQuerySet):
+    __for_write = False
+
+    @property
+    def _for_write(self):
+        return self.__for_write
+
+    @_for_write.setter
+    def _for_write(self, value):
+        if issubclass(self.model, (GuidMixin)):
```
cwisecarver: We'd need to do it on OptionalGuidMixin too.

osf/utils/fields.py
mfraezz: Seemingly not; `AttributeError`s were being thrown from this call to `.lower`

cwisecarver: Yeah, super calls [`to_python`](https://github.com/django/django/blob/1.9/django/db/models/fields/__init__.py#L1105), which checks to see if it's a string and then makes it a string if it isn't ([smart_text](https://github.com/django/django/blob/1.9/django/utils/encoding.py#L32) which calls [force_text](https://github.com/django/django/blob/1.9/django/utils/encoding.py#L58)). This is django's default behavior. If you want to prevent integers being cast you'd have to write your own validator and put it on the field. I'm not 100% sure but I'm fairly confident that if you did a SQL update with an integer into a varchar PostgreSQL would coerce it for you.

cwisecarver: Hmm. Anyone see why, following that path, they wouldn't come out the other side as strings? You remember what type of thing you were testing with @mfraezz?

cwisecarver:
Just in general, while this does cut the query count down, it doesn't eliminate all the extra queries. As discussed with @mfraezz last night, I think we should hold off on merging. Either revisit this approach trying to remove redundant queries or start again with a new technique.

Some examples:
- https://github.com/ionelmc/django-prefetch
- https://github.com/podhmo/django-aggressivequery
- https://github.com/coleifer/django-generic-aggregation
Alternatively, I suppose we could just ask @jmcarp to tell us how to fix it. [He seems to have some experience in this area.](https://github.com/jmcarp/nplusone)

osf/utils/fields.py
mfraezz: I do not, only that you said we should probably be casting this `value`.

cwisecarver: I confess.

mfraezz:
>As discussed with [me] last night, I think we should hold off on merging

Actually, that was because `AbstractNodeQueryset` was still inheriting from `ModmCompatibilityQuerySet`, not the `Guid*` version. Confirmed that the right number of queries are happening now:

```
In [3]: Node.objects.all()[:10]
Out[3]: SELECT "osf_abstractnode"."id",
       "osf_abstractnode"."guid_string",
       "osf_abstractnode"."content_type_pk",
       "osf_abstractnode"."type",
       "osf_abstractnode"."spam_status",
       "osf_abstractnode"."spam_pro_tip",
       "osf_abstractnode"."spam_data",
       "osf_abstractnode"."date_last_reported",
       "osf_abstractnode"."reports",
       "osf_abstractnode"."category",
       "osf_abstractnode"."child_node_subscriptions",
       "osf_abstractnode"."creator_id",
       "osf_abstractnode"."date_created",
       "osf_abstractnode"."date_modified",
       "osf_abstractnode"."deleted_date",
       "osf_abstractnode"."description",
       "osf_abstractnode"."file_guid_to_share_uuids",
       "osf_abstractnode"."forked_date",
       "osf_abstractnode"."forked_from_id",
       "osf_abstractnode"."is_fork",
       "osf_abstractnode"."is_public",
       "osf_abstractnode"."is_deleted",
       "osf_abstractnode"."node_license_id",
       "osf_abstractnode"."piwik_site_id",
       "osf_abstractnode"."public_comments",
       "osf_abstractnode"."suspended",
       "osf_abstractnode"."template_node_id",
       "osf_abstractnode"."title",
       "osf_abstractnode"."wiki_pages_current",
       "osf_abstractnode"."wiki_pages_versions",
       "osf_abstractnode"."wiki_private_uuids",
       "osf_abstractnode"."preprint_file_id",
       "osf_abstractnode"."preprint_article_doi",
       "osf_abstractnode"."_is_preprint_orphan",
       "osf_abstractnode"."_has_abandoned_preprint",
       "osf_abstractnode"."keenio_read_key",
       "osf_guid"."_id" AS "guids___id"
FROM "osf_abstractnode"
LEFT OUTER JOIN "osf_guid" ON ("osf_abstractnode"."id" = "osf_guid"."object_id"
                               AND ("osf_guid"."content_type_id" = 27))
WHERE "osf_abstractnode"."type" = 'osf.node' LIMIT 10

Execution time: 0.004986s [Database: default]

[<Node: Churn inquiry ethical sphere ancestor bacteria. : (2xa5t)>, <Node: Lethal constellation aven load amplitude dampness. : (nbq67)>, <Node: бабушка : (xzsur)>, <Node: child1 : (6vbj5)>, <Node: parent : (dx7yv)>, <Node: Fissure stalagmites climate unpredictable authority propagation. : (df5cq)>, <Node: Satellite vacuole lethal redwoods capacity decay. : (gb289)>, <Node: родитель : (n7vab)>, <Node: ребенок : (y5nmr)>, <Node: asdf : (mt8ge)>]
```

osf/utils/fields.py
sloria: I'm concerned about letting invalid input pass. How about we revert this change for now so we can diagnose where the invalid input is coming from?

PR #6804, opened 2017-02-01
Update fix for infinite recursion [OSF-7396]

alexschiller:
## Purpose
Get children returned all nested children in a flat list, which caused duplicates to occur when component nesting was more than one level deep.

## Changes
Update and use get_nodes instead. 

## Side effects
Refactored notifications, wikis, and privacy tree creating functions to take lists  of node objects instead of ids. Relevant tests updated and such. This should function the same, but be more efficient/clean.

## Ticket
[OSF-7396]
https://openscience.atlassian.net/browse/OSF-7396


osf/models/node.py
```diff
@@ -274,7 +274,12 @@ def get_nodes(self, **kwargs):
         children.
         """
         # Prepend 'child__' to kwargs for filtering
-        filter_kwargs = {'child__{}'.format(key): val for key, val in kwargs.items()}
+        filter_kwargs = {}
+        if 'is_node_link' in kwargs.keys():
```
sloria: Minor: Calling `.keys()` is unnecessary:

```python
if 'is_node_link' in kwargs:
```

Also, the below two lines can be simplified to:

```python
filter_kwargs['is_node_link'] = kwargs.pop('is_node_link')
```

PR #6808, opened 2017-02-01
[Feature] Encrypt unicode properly for EncryptedTextFields [#OSF-7200]

mfraezz:
## Purpose
#6766, but for Django

h/t @alexschiller 

## Side effects
None expected

## Ticket
[#OSF-7200](https://openscience.atlassian.net/browse/OSF-7200)

### DIFF
--- a/osf_tests/test_external_accounts.py
+++ b/osf_tests/test_external_accounts.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 import string
 import random
 
@@ -7,11 +8,11 @@ from django.db import connection
 from psycopg2._psycopg import AsIs
 
 from osf.models import ExternalAccount
-from osf.utils.fields import EncryptedTextField, SENSITIVE_DATA_KEY
+from osf.utils.fields import EncryptedTextField, SENSITIVE_DATA_KEY, ensure_bytes
 from .factories import ExternalAccountFactory
 
 @pytest.mark.django_db
-class TestEncryptedStringField(object):
+class TestEncryptedExternalAccountFields(object):
     def setup_class(self):
         self.magic_string = ''.join(random.choice(string.hexdigits) for _ in range(25))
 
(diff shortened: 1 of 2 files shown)


osf_tests/test_external_accounts.py
```diff
@@ -39,3 +42,88 @@ def test_database_is_encrypted(self):
...
+    def setUpClass(cls):
+        super(TestEncryptedTextField, cls).setUpClass()
+        cls.field = EncryptedTextField(null=True, blank=True)
+
+
+    def test_ensure_bytes_encodes_no_unicode_in_string_type_str(self):
+        my_value = 'hello'
+        assert_true(isinstance(my_value, bytes))
+        my_str = ensure_bytes(my_value)
+        assert_true(isinstance(my_str, bytes))
```
sloria: Minor: we aren't using nose asserts in the `osf_tests`; were these tests copied from somewhere else?

osf_tests/test_external_accounts.py
```diff
@@ -39,3 +42,88 @@ def test_database_is_encrypted(self):
             row = cursor.fetchone()
             for blicky in row:
                 assert jwe.decrypt(bytes(blicky[len(EncryptedTextField.prefix):]), SENSITIVE_DATA_KEY) == self.magic_string
+
+
+class TestEncryptedTextField(TestCase):
```
sloria: Minor: We aren't using `unittest.TestCase` for `osf_tests`, since it is unnecessary with pytest.

```python
class TestEncryptedTextField:
    @pytest.fixture
    def field(self):
        return EncryptedTextField(null=True, blank=True)

    def test_ensure_bytes...(self, field):
        # ...
```

osf_tests/test_external_accounts.py
mfraezz: They were, see the referenced PR. I'll convert all of these

mfraezz:
TCI passed for [my build](https://travis-ci.org/mfraezz/osf.io/builds/197457906), merging.

PR #6809, opened 2017-02-02
[WIP] Add first pass at db router [#OSF-7406]

cwisecarver:

### DIFF
--- a/osf/db/router.py
+++ b/osf/db/router.py
@@ -1,36 +1,33 @@
 from django.conf import settings
 import psycopg2
 
-CACHED_MASTER = None
-
 
 class PostgreSQLFailoverRouter(object):
     """
-    1. CHECK MASTER_SERVER_DSN @ THREAD LOCAL
-    2. THERE?, GOTO 9
-    3. GET RANDOM_SERVER FROM `settings.DATABASES`
-    4. CONNECT TO RANDOM_SERVER
-    5. IS MASTER SERVER?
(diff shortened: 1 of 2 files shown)


osf/db/router.py
```diff
@@ -0,0 +1,70 @@
...
+        self._get_dsns()
+        global CACHED_MASTER
+        if not CACHED_MASTER:
+            CACHED_MASTER = self._get_master()
+
+    def _get_master(self):
+        for name, dsn in self.DSNS.iteritems():
+            conn = self._get_conn(dsn)
+            cur = conn.cursor()
+            cur.execute('SELECT pg_is_in_recovery();')
```
icereval: This is good but lets stick with the JDBC implementation if at all possible.

https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/core/v3/ConnectionFactoryImpl.java#L692

cwisecarver: They're probably the same. My only concern would be a pedantic one. Yours checks if the server is read only. Mine checks to see if one is master vs one is slave. You can technically have a master that's read only. It's going to error either way though if you try to write to it.

cwisecarver:
@icereval @chrisseto  Updated the query, added some comments, added setting to enable router, still running migrations so haven't actually tested running django with it.

mfraezz:
Closing in favor of #6820

PR #6812, opened 2017-02-02
Update Docker config for use in development of MFR and WB

TomBaxter:
## Purpose:
[SVCS-245](https://openscience.atlassian.net/browse/SVCS-245)
Update local development Docker config and doc for WB and MFR

## Changes:
Update  README-docker-compose.md 
- docker/docker-compose install
- loopback alias
- alternative to docker-sync
Add docker-compose.ubuntu.yml
- Used in place of docker-compose.override.yml

## Side effects
None

## This PR is tied to WaterButler PR#   . Though they should not have to be merged at the same time.

[SVCS-245]


docker-compose.ubuntu.yml
```diff
@@ -7,13 +7,25 @@ services:
 #    image: quay.io/centerforopenscience/fakecas:latest
 #    command: fakecas -host=0.0.0.0:8080 -dbaddress=172.19.0.7:27017
 
-#  wb:
-#    volumes:
-#      - ../waterbutler:/code
+# wb:
+#   volumes:
+#     - ../../wb/waterbutler:/code
```
felliott: Minor: can you set these and the other volume relative paths back to `../waterbutler` and `../modular-file-renderer`?

TomBaxter: Done

README-docker-compose.md
```diff
@@ -93,6 +93,22 @@
       wb:
         volumes_from:
           - container:wb-sync
+
+      wb_worker:
```
icereval: no need to duplicate the config, these can be left in the override file commented out, further clarification can be added above if I did not explain it well enough.

icereval: maybe just add an ellipse?

README-docker-compose.md
```diff
@@ -106,6 +122,15 @@
         sync_excludes_type: 'Name'
         sync_excludes: ['.DS_Store', '*.pyc', '*.tmp', '.git', '.idea']
         watch_excludes: ['.*\.DS_Store', '.*\.pyc', '.*\.tmp', '.*/\.git', '.*/\.idea']
+
+      mfr-sync:
```
icereval: same as above comment about an ellipse

website/util/__init__.py
```diff
@@ -123,7 +123,7 @@ def api_v2_url(path_str,
     return str(base_url)
 
 
-def web_url_for(view_name, _absolute=False, _guid=False, *args, **kwargs):
+def web_url_for(view_name, _absolute=False, _internal=False, _guid=False, *args, **kwargs):
```
icereval: I see we have a couple duplicative `web_url_for` methods, do these need to be vetted?

addons/base/views.py
```diff
@@ -289,6 +289,7 @@ def get_auth(auth, **kwargs):
             'callback_url': node.api_url_for(
                 ('create_waterbutler_log' if not node.is_registration else 'registration_callbacks'),
                 _absolute=True,
+                _internal=settings.DEV_MODE,
```
icereval: same question as below "can you elaborate on why is this correct only in DEV_MODE?"

website/addons/base/views.py
```diff
@@ -289,6 +289,7 @@ def get_auth(auth, **kwargs):
             'callback_url': node.api_url_for(
                 ('create_waterbutler_log' if not node.is_registration else 'registration_callbacks'),
                 _absolute=True,
+                _internal=settings.DEV_MODE,
```
TomBaxter: Sounds like I don't have a full understanding of DEV_MODE. My assumption was that it meant local deployment, now specifically local docker deployment. It is required for local docker deployment, but I'm not sure what other scenarios we might need to cover. If production is also using the localhost aliasing idiom, then perhaps this should always be set to True. Or maybe I need to a new variable to specifically denote local docker deployment?

TomBaxter:
RE: "share rabbitmq between OSF and WB"
  I've been using this setup for a while and done quite a bit of cross provider testing in WB. I would say that is tested. I can't say I've done much on the OSF side as far as exercising celery/rabbitmq.

docker-compose.yml
```diff
@@ -152,11 +154,33 @@ services:
     restart: unless-stopped
     ports:
       - 7777:7777
+    environment:
+      RABBITMQ_PORT_5672_TCP_ADDR: 192.168.168.167
+    env_file:
+      - .docker-compose.wb.env
+    volumes:
```
TomBaxter: I've got the celery wb_worker depending on rabbitmq, does the the wb service need to be dependent on it as well?

docker-compose.yml
```diff
@@ -152,11 +154,33 @@ services:
     restart: unless-stopped
     ports:
       - 7777:7777
+    environment:
+      RABBITMQ_PORT_5672_TCP_ADDR: 192.168.168.167
```
TomBaxter: Complete.

docker-compose.yml
```diff
@@ -152,11 +154,33 @@ services:
...
+
+  wb_worker:
+    image: quay.io/centerforopenscience/waterbutler:develop
+    command: invoke celery
+    restart: unless-stopped
+    depends_on:
+      - wb
+      - rabbitmq
+    environment:
+      RABBITMQ_PORT_5672_TCP_ADDR: 192.168.168.167
```
TomBaxter: Complete.

README-docker-compose.md
TomBaxter: Complete.

README-docker-compose.md
TomBaxter: Complete.

website/util/__init__.py
TomBaxter: `osf/models/node.py:    def web_url_for`  and
`website/project/model.py:    def web_url_for`  call
`website/util/__init__.py:def web_url_for` with kwargs so I think we're good here.

addons/base/views.py
TomBaxter: Complete.

icereval:
@TomBaxter, I apologize I was confusing the use of rabbitmq with the need to run the celery tasks for parity and glacier backup, you've enlightened/reminded me of the need for celery for cross provider file movies which is certainly a valid use case for this configuration.

I'll put up a PR to your branch here soon with a few small changes (mainly need to update rabbitmq for vhosts - https://stackoverflow.com/questions/12209652/multi-celery-projects-with-same-rabbitmq-broker-backend-process)

Good work on this.

docker-compose.yml
```diff
@@ -375,4 +395,5 @@ services:
       - osf_node_modules_vol:/code/node_modules
       - preprints_dist_vol:/preprints
       - registries_dist_vol:/registries
+      - osfstoragecache_vol:/code/website/osfstoragecache
```
icereval: this does not need to be here, osf merely directs wb where to store its temp files relative to the wb process

icereval: fyi, removed in merge

TomBaxter: Coolio, I only left it there so that those working on OSF container would find the files where they expected to.

PR #6814, opened 2017-02-02
[SVCS-5] Fix GitHub addon so you can rename with a percent sign

Johnetordoff:
## Purpose

Allow the user to rename a folder or file using the % sign.

## Changes

Short line encoding the url for github properly.

## Side effects

None that I know of

## Ticket

https://openscience.atlassian.net/browse/SVCS-5


felliott:
This will work, but is just a band-aid for a larger issue @Johnetordoff uncovered.  See [SVCS-256](https://openscience.atlassian.net/browse/SVCS-256).  Recommend closing for now.

PR #6816, opened 2017-02-03
[SVCS-234][WIP] Support slashes in google drive file names

Johnetordoff:
## Purpose

Allows user to rename files using the googledrive addon with a slash in the title.

## Changes

Add line that encodes url for gdrive

## Side effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/SVCS-234

### DIFF
--- a/website/static/js/fangorn.js
+++ b/website/static/js/fangorn.js
@@ -530,6 +530,10 @@ function doItemOp(operation, to, from, rename, conflict) {
         moveSpec.branch = from.data.branch;
     }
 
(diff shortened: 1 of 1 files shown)


felliott:
I don't think this fixes the underlying issue.  Recommend closing until confirmed.

Johnetordoff:
@felliott Yes I was just testing this allows you to rename files, but then they can't be retrieved (downloaded or deleted) without changing back to a non-slash containing file. I can see what's causing the error now.

PR #6830, opened 2017-02-07
Add Institutions matching query to API search results [OSF-7370]

erinspace:
## Purpose

Search results from the API were not equipped to deal with showing found Institutions matching the query search terms. Add the means for Institutions to show up in search results now without throwing a 500 error!

## Changes

- Update search views, urls, and serializers to also show Institution information in results
- Add Institution results in the pagination view for search results
- Add key to `elastic_search.py`'s `DOC_TYPE_TO_MODEL`
- Update elastic search when an institution is saved
- Add tests

## Side effects

None internally, but external uses of the search endpoint of the URL will now have to account for another key, and another type of serialized search result.

## Ticket
https://openscience.atlassian.net/browse/OSF-7370

### DIFF
--- a/api/base/pagination.py
+++ b/api/base/pagination.py
@@ -323,6 +323,7 @@ class SearchPagination(JSONAPIPagination):
                     ('components', self.get_search_field('component', query)),
                     ('registrations', self.get_search_field('registration', query)),
                     ('users', self.get_search_field('user', query)),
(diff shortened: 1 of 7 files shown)


brianjgeiger:
@erinspace QA says this doesn't seem to be happening any more. Can you see if other changes may have been made related to this? If need be, we can just merge the test if the test passes on current develop.

PR #6834, opened 2017-02-08
[Feature] Add sitemap generator [OSF-7322]

alexschiller:
## Purpose
No osf sitemap.

## Changes
Adds a sitemap generator and beat job.

## Side effects
None known

## Notes
1. This requires beat and will require prod modifications to work.
2. This will also require a later PR to add a ping to google on prod when the sitemap is updated. 
3. After CR, I would like to add a few tests, but I want at least one pass of CR first in case things need to drastically change.
4. This will require some product changes to priorities (not currently generated) and changefreqs.

## QA Notes
I will confirm that it is working, no normal QA at the moment.

## Ticket
[#OSF-7322]
https://openscience.atlassian.net/browse/OSF-7322

### DIFF
--- a/scripts/generate_sitemap.py
+++ b/scripts/generate_sitemap.py
@@ -30,9 +30,6 @@ logging.basicConfig(level=logging.INFO)
 class Sitemap(object):
     def __init__(self):
         self.sitemap_count = 0
-        # ajs: This url_max was chosen to be sufficiently lower than the 5mb per file limit given in spec. 
-        # It was not clear whether 5mb was pre or post gzip. This could be revisited if limits are reached.
-        self.url_max = 24999
         self.new_doc()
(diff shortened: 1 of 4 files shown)


scripts/generate_sitemap.py
```diff
@@ -0,0 +1,195 @@
...
+            self.add_url(**{
+                'loc': urlparse.urljoin(settings.DOMAIN, url),
+                'changefreq': freq,
+            })
+
+        # User urls
+        for obj in OSFUser.objects.filter(is_active=True):
+            self.add_url(**{
+                'loc': urlparse.urljoin(settings.DOMAIN, obj.url),
+                'changefreq': 'yearly',
```
brianjgeiger: Depending on how fancy we get with this, we'd likely want to make this the configurable part.
```
'changefreq`: settings.SITEMAP_FREQUENCY.OSF_USER
```
Or similar.

Issue #6062, opened 2016-07-29
502 Bad Gateway from /files/github/ endpoint after updating file in GitHub


emetsger:
I have a [project](https://osf.io/xug4a/) in the OSF with a GitHub addon configured to expose this [repository](https://github.com/emetsger/osf-test/).

To recreate the issue:
1. Visit the [OSF v2 API /files/github/](https://api.osf.io/v2/nodes/xug4a/files/github/) endpoint.  Receive a 200 response.
2. Go to GitHub, modify the [README.md](https://github.com/emetsger/osf-test/blob/master/README.md) file.
3. Refresh the [files endpoint](https://api.osf.io/v2/nodes/xug4a/files/github/), receive a 502 Bad Gateway response.
4. View the [content of README.md](https://osf.io/8pdf3/) via the OSF UI, confirm it contains edits from step 2.
5. Refresh [files endpoint](https://api.osf.io/v2/nodes/xug4a/files/github/), receive a 200.

I'm not expecting to receive the 502 Bad Gateway response in step 3.

felliott:
Hey @emetsger,

I tagged you into a similar issue that was reported over on the WB queue.  I just tried replicating this locally and on prod with a fork of your repo, but no dice.  Do you mind trying this again and letting me know if it still errors for you?  I'm wondering if this issue affects a bunch of different GH endpoints and if they're fixing them one at a time.  I working on adding some logging code to WB that'll let us ID the troublesome endpoints, but it's not yet done.  Thank you and have a happy new year!

Cheers,
Fitz

felliott:
Summary: this occurs because WB returns different metadata for a file depending on if the file or the file's parent's endpoint is used.  Specifically, the parent metadata doesn't include the `modified` timestamp for the file.  When the OSF tries to update the its file metadata cache, it fails, expecting `modified` to be there.  If you then access the *file*'s Node File endpoint directly, that will succeed, the metadata will be updated, and subsequent requests to the parent endpoint will succeed, since the metadata is now up-to-date.  This will need to be fixed OSF-side.

emetsger:
Hey @felliott,

So I confirmed that I can still re-create the issue described above by following the original steps.

I could also re-create the issue by replacing the original Step 4 with:
[...]

PR #6835, opened 2017-02-09
Handle incomplete metadata from WB when addon's parent directory accessed [OSF-7148]

erinspace:
## Purpose

Waterbutler returns different metadata for a file when its accessed from its parent's endpoint or the file endpoint directly. Accessing the file endpoint directly (`/files/github/filename.txt`) includes "modified" information, and then editing and then going back to the parent (`/files/github/`) did not have the "modified" field for the new update. The OSF was breaking when trying to insert this edited entry with missing modified information into the file history.

As a fix, if there is no modified metadata available, append to the end of the history.

h/t to @felliott for much help and amazingly helpful github comments!

## Changes

- Update URL for accessing Waterbutler file info to use the internal waterbutler URL
- If there is no modified information included in the metadata, append to the end of the file history.

## Side effects

- If a file is accessed for metadata via the `/files/github/` endpoint after it's been edited, it might not be in the correct point of the file's history, if the commits that updated the file were from the past. 

## Ticket
Github issue: #6062 
Jira ticket: https://openscience.atlassian.net/browse/OSF-7148

### DIFF
--- a/osf/models/files.py
+++ b/osf/models/files.py
@@ -361,7 +361,11 @@ class BaseFileNode(TypedModel, CommentableMixin, OptionalGuidMixin, Taggable, Ob
                 break
         else:
             # Insert into history if there is no matching etag
-            utils.insort(self.history, data, lambda x: x['modified'])
+            if data.get('modified'):
+                utils.insort(self.history, data, lambda x: x['modified'])
+            # If modified not included in the metadata, insert at the end of the history
+            else:
+                self.history.append(data)
 
         # Finally update last touched
         self.last_touched = timezone.now()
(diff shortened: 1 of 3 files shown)


osf/models/files.py
```diff
@@ -754,7 +754,11 @@ def update(self, revision, data, user=None):
                 break
         else:
             # Insert into history if there is no matching etag
-            utils.insort(self.history, data, lambda x: x['modified'])
+            if data.get('modified'):
```
brianjgeiger: Could you make a model test for this?

brianjgeiger:
@erinspace I just restarted tests on this. Could you check on it around 3:30 and see if they're still failing and fix if so? If they pass, move the ticket to Passed Code Review. Thanks!

PR #6845, opened 2017-02-13
Prefix Admin App Table Names w/ admin_  [OSF-7489]

erinspace:
## Purpose

Table names for apps under the admin umbrella didn't have the admin_ prefix.  Add appropritate app configs so that things are labeled properly, much like `addons` have done.

## Changes

- Add admin app configs in `apps.py` in `/base/`, `/metrics/`, and `/common_auth/`
- Call those app configs in their respective `__init__.py`
- Update migrations to use new app names, account for "old" databases with potentially duplicate data in migrations.

## Side effects

This will create alternate sets of Permissions and models on an older database, as the app label has changed so all models are named differently.  Everything should work fine, your DB will just be cluttered.

## Ticket

https://openscience.atlassian.net/browse/OSF-7489


admin/common_auth/migrations/0002_auto_20170212_2151.py
```diff
@@ -46,22 +44,15 @@ def add_group_permissions(*args):
...
-        logger.info('nodes_and_users renamed to read_only')
-    except Group.DoesNotExist:
-        group = Group.objects.get_or_create(name='read_only')
+    # Create and add permissions for the read only group
+    group, created = Group.objects.get_or_create(name='read_only')
+    if created:
         logger.info('read_only group created')
-
-    # Read only for nodes, users, meetings, and spam
     [group.permissions.add(perm) for perm in get_read_only_permissions()]
```
sloria: Minor: You can do

```python
group.permissions.add(*get_read_only_permissions())
```

PR #6849, opened 2017-02-14
Fix Node deletion behavior wrt NodeLinks [OSF-7444]

mfraezz:
## Purpose
Fix behavior around deleting Nodes with with NodeLinks or components

## Changes
* Only `.load` actual models
* `.filter` correctly, if less efficiently

## Side effects
None expected

## Tickets
[[#OSF-7444]](https://openscience.atlassian.net/browse/OSF-7444)
[~~[#OSF-7492]~~](https://openscience.atlassian.net/browse/OSF-7492)

### DIFF
--- a/osf/models/node.py
+++ b/osf/models/node.py
@@ -72,7 +72,7 @@ class AbstractNodeQueryset(GuidMODMCompatibilityQuerySet):
             where=['"osf_abstractnode".id in (SELECT id FROM osf_abstractnode WHERE id NOT IN (SELECT child_id FROM '
                    'osf_noderelation WHERE is_node_link IS false))'])
 
-    def get_children(self, root, primary_keys=False):
+    def get_children(self, root, primary_keys=False, active=False):
         sql = """
             WITH RECURSIVE descendants AS (
               SELECT
@@ -80,7 +80,8 @@ class AbstractNodeQueryset(GuidMODMCompatibilityQuerySet):
                 child_id,
                 1 AS LEVEL
               FROM %s
-              WHERE is_node_link IS FALSE
+              %s
+              WHERE is_node_link IS FALSE %s
               UNION ALL
               SELECT
                 s.parent_id,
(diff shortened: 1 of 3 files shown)


website/project/views/node.py
```diff
@@ -692,7 +692,7 @@ def _view_project(node, auth, primary=False):
             'date_created': iso8601format(node.date_created),
             'date_modified': iso8601format(node.logs.latest().date) if node.logs.exists() else '',
             'tags': list(node.tags.values_list('name', flat=True)),
-            'children': bool(node.nodes_active),
+            'children': bool(set(Node.objects.get_children(node)) & set(node.nodes_active)),
```
mfraezz: This change appears to have broken [tests](https://travis-ci.org/CenterForOpenScience/osf.io/jobs/201399990#L1904-L1919). Previously, it was causing the [frontend](https://github.com/CenterForOpenScience/osf.io/blob/914a6e64890a7d69dbf7dbb757c85e480f9d9eea/website/templates/project/project_base.mako#L106) to [display](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/pages/project-settings-page.js#L86-L87) a "`Any child components must be deleted prior to deleting this project`" warning without checking to see if those children were pointers, as  described in [the ticket](https://openscience.atlassian.net/browse/OSF-7492). Further investigation confirmed that said behavior is present on prod.

osf/models/node.py
```diff
@@ -2475,7 +2475,7 @@ def remove_node(self, auth, date=None):
                 '{0!r} does not have permission to modify this {1}'.format(auth.user, self.category or 'node')
             )
 
-        if self.nodes_primary.filter(is_deleted=False).exists():
+        if set(Node.objects.get_children(self)) & set(self.nodes_active):
```
icereval: django query?

cc @cwisecarver

sloria:
Looks good

PR #6861, opened 2017-02-16
[Feature] Convert UTC to local, remove kludge [OSF-7509]

alexschiller:
## Purpose
At some point fangorn stopped using proper local time but instead was reporting UTC time.

## Changes
Convert UTC to local time.

## Side effects
Remove a kludge that allowed dropbox v1 api dates to work. We use v2 now so it should be fine.

## Ticket
[#OSF-7509]
https://openscience.atlassian.net/browse/OSF-7509

## QA Notes
1. Note that timestamps for previously uploaded files will not be accurate (see https://openscience.atlassian.net/browse/SVCS-259)
2. Test this with multiple service providers to ensure that timestamps work across the board (in particular dropbox since there was a change with that one). 
3. Test sorting by modified date.


brianjgeiger:
Test failures can't be related to code, I'm pretty sure. Going to merge.

PR #6872, opened 2017-02-19
[RFR] Add counting to notification subscription fks

cwisecarver:
## Purpose

See title.

## Changes

Increment fk count for notification subscription fks.

## Side effects

nah


cwisecarver:
Merging because addition has never hurt anyone.

PR #6873, opened 2017-02-19
[RFR] Work around case sensitive node id

cwisecarver:
## Purpose

JYznd is a cased node id and the permissions migration for contributors was failing on it.

## Changes

Created a lowercased, unicode version of node permissions and visible contributor ids and used those to do lookups against for contributor migration.

## Side effects

Unlikely.


cwisecarver:
Merging because it can only be self-harm.

PR #6881, opened 2017-02-21
[feature] Log errors in sitemap generator [OSF-7322]

alexschiller:
## Purpose
Previous had no error handling.

## Changes
Now catches all errors in xml serialization and logs them, throws sentry errors when too much is breaking.

## Side effects
NA

## Ticket
[#OSF-7322]
https://openscience.atlassian.net/browse/OSF-7322

##QA Notes
No QA, if it gets moved to QA put back in AD and send it back to me. I'll verify that everything is working with staging data.

### DIFF
--- a/scripts/generate_sitemap.py
+++ b/scripts/generate_sitemap.py
@@ -111,7 +111,7 @@ class Sitemap(object):
         logger.info('Error on {}, {}:'.format(obj, obj_id))
         logger.exception(error)
         if self.errors == 1000:
-            sentry.log_message('WARNING: generate_sitemap stopped execution after reaching 1000 errors.')
+            sentry.log_message('ERROR: generate_sitemap stopped execution after reaching 1000 errors. See logs for details.')
             raise Exception('Too many errors generating sitemap.')
 
     def generate(self):
@@ -187,11 +187,11 @@ class Sitemap(object):
         # TODO: server side cursor query wrapper might be useful as the index gets larger.
         # Sitemap indexable limit check
         if self.sitemap_count > settings.SITEMAP_INDEX_MAX * .90: # 10% of urls remaining
-            sentry.log_message('WARNING: Max sitemaps nearly reached: {} of 50000 max sitemaps for index file'.format(str(self.sitemap_count)))
+            sentry.log_message('WARNING: Max sitemaps nearly reached.')
         print('Total url_count = {}'.format((self.sitemap_count - 1) * settings.SITEMAP_URL_MAX + self.url_count))
         print('Total sitemap_count = {}'.format(str(self.sitemap_count)))
         if self.errors:
-            sentry.log_message('WARNING: Generate sitemap encountered {} error(s). See logs for details.'.format(self.errors))
+            sentry.log_message('WARNING: Generate sitemap encountered errors. See logs for details.')
             print('Total errors = {}'.format(str(self.errors)))
         else:
             print('No errors')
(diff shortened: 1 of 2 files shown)


scripts/generate_sitemap.py
```diff
@@ -160,10 +187,14 @@ def generate(self):
         # TODO: server side cursor query wrapper might be useful as the index gets larger.
         # Sitemap indexable limit check
         if self.sitemap_count > settings.SITEMAP_INDEX_MAX * .90: # 10% of urls remaining
-            logger.exception('WARNING: Max sitemaps nearly reached: {} of 50000 max sitemaps for index file'.format(str(self.sitemap_count)))
-            log_exception()
+            sentry.log_message('WARNING: Max sitemaps nearly reached: {} of 50000 max sitemaps for index file'.format(str(self.sitemap_count)))
         print('Total url_count = {}'.format((self.sitemap_count - 1) * settings.SITEMAP_URL_MAX + self.url_count))
         print('Total sitemap_count = {}'.format(str(self.sitemap_count)))
+        if self.errors:
+            sentry.log_message('WARNING: Generate sitemap encountered {} error(s). See logs for details.'.format(self.errors))
```
brianjgeiger: The title of the sentry message for this should be static so that it doesn't create a new event each time. The details can differ, but the titles need to be the same. Also, this one should be lower priority than the one above where we had to cancel building the sitemap.

scripts/generate_sitemap.py
```diff
@@ -160,10 +187,14 @@ def generate(self):
         # TODO: server side cursor query wrapper might be useful as the index gets larger.
         # Sitemap indexable limit check
         if self.sitemap_count > settings.SITEMAP_INDEX_MAX * .90: # 10% of urls remaining
-            logger.exception('WARNING: Max sitemaps nearly reached: {} of 50000 max sitemaps for index file'.format(str(self.sitemap_count)))
-            log_exception()
+            sentry.log_message('WARNING: Max sitemaps nearly reached: {} of 50000 max sitemaps for index file'.format(str(self.sitemap_count)))
```
brianjgeiger: Ensure that the title for this one is the same from error to error as in the message below.

scripts/generate_sitemap.py
```diff
@@ -102,6 +104,16 @@ def write_sitemap_index(self):
...
             f.write(xml_str)
 
+    def log_errors(self, obj, obj_id, error):
+        if not self.errors:
+            script_utils.add_file_logger(logger, __file__)
+        self.errors += 1
+        logger.info('Error on {}, {}:'.format(obj, obj_id))
+        logger.exception(error)
+        if self.errors == 1000:
+            sentry.log_message('WARNING: generate_sitemap stopped execution after reaching 1000 errors.')
```
brianjgeiger: This should be higher priority than the one where the errors are anywhere from 1-999

scripts/generate_sitemap.py
```diff
@@ -111,45 +123,60 @@ def generate(self):
...
+                            'osfstorage',
+                            obj.primary_file._id, # Preprint file deep_url
+                            '?action=download'
+                        )
+                    )
+                    file_config['lastmod'] = preprint_date
+                    self.add_url(file_config)
+                except Exception as e:
+                    self.log_errors(obj.primary_file, obj.primary_file._id, e)
+            except Exception as e:
```
brianjgeiger: This might be a little grabby, but it's probably fine. I'm expecting the first few rounds of work post-release will be tightening this up a bit.

alexschiller: Intentionally grabby here-- in this case if there is a problem with the obj.primary_file (as could be the case in an orphan) it is will raise further and get caught on the preprint itself. The preprint info should still be accessible and the error won't stop completely break the sitemap generation.

PR #6883, opened 2017-02-21
[OSF-7511] Put admin models in osf database

sloria:
## Purpose

Put admin-related models in the same database as osf models so that they can have relationships with each other (e.g. AdminProfile relates to OSFUser). Also, prevent migrations from being run from the admin server to avoid overwriting data in the osf database (e.g. the django_migrations table).

## Changes

- Move AdminProfile and OSFLogEntry from admin project to osf app
- Rename OSFLogEntry -> AdminLogEntry
- Prevent migrations from being run by the admin app
- Fix changing nodes in Django admin

## Side effects

- `python manage.py migrate` needs to be run on app server
- `python manage.py migrate` should not be run on admin servers. The command will fail

## Ticket

https://openscience.atlassian.net/browse/OSF-7511


admin/base/settings/defaults.py
```diff
@@ -102,8 +96,27 @@
...
+    'admin.spam',
+    'admin.metrics',
+    'admin.nodes',
+    'admin.users',
+    'admin.desk',
+    'admin.meetings',
+
 )
 
+MIGRATION_MODULES = {
```
cwisecarver: This doesn't appear to get used anywhere, at least not in this PR.

osf/admin.py
```diff
@@ -6,48 +6,33 @@
 
 class NodeAdmin(ForeignKeyAutocompleteAdmin):
     fields = (
-        'title',
-        'description',
+        'affiliated_institutions',
```
cwisecarver: This might be shorter: https://docs.djangoproject.com/en/1.10/ref/contrib/admin/#django.contrib.admin.ModelAdmin.exclude

admin/base/settings/defaults.py
sloria: It is a Django setting: https://docs.djangoproject.com/en/1.10/ref/settings/#migration-modules .

> When you supply None as a value for an app, Django will consider the app as an app without migrations regardless of an existing migrations submodule. 

I also made `python manage.py migrate` crash if run from the admin server and used a custom router to make migrations in the admin project be noops. These are redundant, but I think the safeguards are worthwhile.

osf/admin.py
sloria: True, I only modified this because `fields` was already there, and a whitelist is safer than a blacklist.

PR #6884, opened 2017-02-22
[OSF-7440] Change Project Description Prompt text

Johnetordoff:
## Purpose

This change encourages users to add a description to their project by changing the prompt text from "No Description" to "'Add a brief description of your [node category]"

## Changes

minor textual change to JS.

## Side effects

None that I know of.

## Ticket 

https://openscience.atlassian.net/browse/OSF-7440

## Figure
screen shot 2017-03-01 at 9 23 47 am


Johnetordoff:
@brianjgeiger I tested for both read-only contributors and non-contributors and it works in the proper way. The description element disappears for these users when empty.

brianjgeiger:
I feel fairly certain that the proposed change can't have caused that massive test failure. Going to merge it in.

PR #6892, opened 2017-02-22
Minor optimizations

sloria:
## Purpose

Add a few minor optimizations.

## Changes

- Prevent redundant query for Institutions on user home page
- Prevent unnecessary Institution and User query on index page when not logged in
- Simplify implementation of `User.update_affiliated_institutions_by_email_domain`

## Ticket

No ticket, just found these inefficiencies while browsing New Relic.

### DIFF
--- a/website/views.py
+++ b/website/views.py
@@ -8,7 +8,6 @@ import urllib
 
 from django.apps import apps
 from modularodm import Q
-from modularodm.exceptions import NoResultsFound
 from flask import request, send_from_directory
 
 from framework import utils, sentry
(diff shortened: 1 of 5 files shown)


website/views.py
```diff
@@ -88,30 +88,36 @@ def _render_nodes(nodes, auth=None, show_path=False, parent_node=None):
...
-    all_institutions = Institution.find().sort('name')
-    dashboard_institutions = [
-        {'id': inst._id, 'name': inst.name, 'logo_path': inst.logo_path_rounded_corners}
-        for inst in all_institutions
-    ]
-
-    return {
-        'home': True,
-        'dashboard_institutions': dashboard_institutions
-    }
```
sloria: This looks like a large change, but all I added was a check for a logged in user and indented everything.

website/views.py
```diff
@@ -88,30 +88,36 @@ def _render_nodes(nodes, auth=None, show_path=False, parent_node=None):
...
+    user_id = get_current_user_id()
+    if user_id:  # Logged in: return either institution page or user home page
```
icereval: needs to move down to L107

PR #6893, opened 2017-02-22
Deprecate TokuTransactions [OSF-7543]

mfraezz:
## Purpose
Make TokuTransactions ~behave as expected wrt the current DB~ raise an exception

## Changes
* ~Conditionally use pre- or post-django code~
* Raise an exception

## Side effects
Any existing uses will error.

## Ticket
[[OSF-7543]](https://openscience.atlassian.net/browse/OSF-7543)

### DIFF
--- a/framework/transactions/context.py
+++ b/framework/transactions/context.py
@@ -1,15 +1,4 @@
 # -*- coding: utf-8 -*-
-import logging
-import functools
-
-from django.db import transaction as dj_transaction
-from pymongo.errors import OperationFailure
-
-from framework.mongo import database as proxy_database
-from framework.transactions import commands, messages, utils
-from website.settings import USE_POSTGRES
-
-logger = logging.getLogger(__name__)
 
 
 class TokuTransaction(object):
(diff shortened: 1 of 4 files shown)


cwisecarver:
After discussing with @sloria and @icereval and all of your hard work we decided that the `TokuTransaction` context manager should just throw an exception instructing the user to use django's `transaction.atomic()` and then immediately die not running any of the code inside it. That should encourage people to switch things over and also not accidentally write things to a database they didn't mean to. 🤦‍♂️

framework/transactions/context.py
```diff
@@ -17,17 +10,7 @@ class TokuTransaction(object):
...
-        self.pending = True
-        return self.atomic.__enter__()
-
-    def __exit__(self, exc_type, exc_val, exc_tb):
-        self.pending = False
-        return self.atomic.__exit__(exc_type, exc_val, exc_tb)
+        raise Exception('This functionality is deprecated, use `django.db.transaction.atomic`')
 
 
 def transaction(database=None):
```
icereval: delete ❓

PR #6900, opened 2017-02-23
Fix tests broken by correct GH behavior [No Ticket]

mfraezz:

### DIFF
--- a/api_tests/nodes/views/test_node_files_list.py
+++ b/api_tests/nodes/views/test_node_files_list.py
@@ -112,8 +112,11 @@ class TestNodeFilesList(ApiTestCase):
         self.user.external_accounts.add(oauth_settings)
         self.user.save()
         addon.user_settings = self.user.get_addon('github')
+        addon.external_account = oauth_settings
         addon.save()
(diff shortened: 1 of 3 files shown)


api_tests/nodes/views/test_node_files_list.py
```diff
@@ -432,14 +441,17 @@ def add_github(self):
...
         self.project.save()
+        addon.user_settings.oauth_grants[self.project._id] = {oauth_settings._id: []}
+        addon.user_settings.save()
 
     def check_file_order(self, resp):
         previous_file_name = 0
         for file in resp.json['data']:
             int_file_name = int(file['attributes']['name'])
-            assert(int_file_name > previous_file_name, 'Files were not in order')
+            assert int_file_name > previous_file_name, 'Files were not in order'
```
mfraezz: Opportunistic fix: this was causing a `pytest.warning`. It was previously just asserting the truthiness of one of two tuples -- either `(True, 'Files were not in order')` or `(False, 'Files were not in order')`, both of which are `True`

PR #6903, opened 2017-02-24
[RFR] Refactor user/me/nodes to hold onto the queryset as long as possible

cwisecarver:
## Purpose

Speedup dashboard page load by reducing number of queries needed for /user/me/nodes api requests.

## Changes

Used SQL annotations to gather necessary data in one query.

## Side effects

Possible that the query is only working and the tests passing because there is a limited amount of data in the database during testing and the `SELECT CASE WHEN` `JOIN` could be working randomly.

### DIFF
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -472,7 +472,6 @@ class ListFilterMixin(FilterMixin):
             operation['value'] = operation['value'] == 'file'
         if field_name == 'bibliographic':
             operation['op'] = 'exact'
-            operation['source_field_name'] = 'contributor__visible'
 
     def get_filtered_queryset(self, field_name, params, default_queryset):
         """filters default queryset based on the serializer field type"""
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -472,6 +472,8 @@ class ListFilterMixin(FilterMixin):
             operation['value'] = operation['value'] == 'file'
         if field_name == 'bibliographic':
             operation['op'] = 'exact'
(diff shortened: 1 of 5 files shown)


cwisecarver:
- Not 100% confident about the `LIMIT 1` in the subquery being the right solution.
   - If the tests pass it might be because of a limited amount of data in the database and it just happening to work out.
- Concerned that the subquery references to columns outside the subquery aren't working

cwisecarver:
Looks like everything is working but bulk. I'll keep digging.

cwisecarver:
Fixed bulk. Tests passed locally, should pass on travis.

http://localhost:8000/v2/users/me/nodes/?related_counts=children&embed=contributors&version=2.2
Before: 11847.62 ms (1919 queries including 1914 duplicates ) 
After: 1793.67 ms (522 queries including 515 duplicates )

Still not good but 15% of what it used to be.

@sloria @icereval

cwisecarver:
@icereval @sloria [I added a test for a large number of users.](https://github.com/CenterForOpenScience/osf.io/pull/6903/commits/22e4c0f5d38dad22e9fca54d128a5e68663c142e) If you wouldn't mind checking my logic but it passes locally so... gtg?

sloria:
The added test looks good to me.

sloria:
Doh, I accidentally closed and reopened. This is the relevant build: https://travis-ci.org/CenterForOpenScience/osf.io/builds/205867302

PR #6904, opened 2017-02-24
[Hotfix] Bookmark collection uniqueness [OSF-7551]

mfraezz:
## Purpose
Ensure/Enforce a 1-1 relationship between Users and bookmark collection Nodes

## Changes
* Add migration to fix data
* Create users bookmark collection only when user is created
* Add commented-out uniqueness constraint

## Side effects
Uniqueness won't be enforced until #6905 is merged.

## Ticket

### DIFF
--- a/scripts/migration/ensure_bookmark_uniqueness.py
+++ b/scripts/migration/ensure_bookmark_uniqueness.py
@@ -1,0 +1,62 @@
+import argparse
+import logging
+
+from modularodm import Q
+
+from framework.auth.core import User
+from framework.transactions.context import TokuTransaction
+from scripts import utils as script_utils
+from website.app import init_app
+from website.project import Node
+
+
+logger = logging.getLogger(__name__)
+
+def get_targets():
+    logger.info('Acquiring targets...')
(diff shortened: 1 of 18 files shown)


sloria:
Since we're now creating a bookmark node whenever a user is created, a lot of warning output is generated:

```
[website.notifications.listeners]  WARNING: Reason: Collections are invalid targets for subscriptions
```

It might make prevent the call to subscribe_user_to_notifications here: https://github.com/CenterForOpenScience/osf.io/blob/2c83ed2343bf9c18e4992425fea33e4b782fb3b2/website/notifications/listeners.py#L10 . 

```python

@project_created.connect
def subscribe_creator(node):
    if node.institution_id or node.is_collection or node.is_deleted:
        return None
    try:
    #...
```

scripts/migration/ensure_bookmark_uniqueness.py
```diff
@@ -0,0 +1,62 @@
...
+from scripts import utils as script_utils
+from website.app import init_app
+from website.project import Node
+
+
+logger = logging.getLogger(__name__)
+
+def get_targets():
+    logger.info('Acquiring targets...')
+    targets = [u for u in User.find() if Node.find(Q('is_bookmark_collection', 'eq', True) & Q('contributors', 'eq', u._id)).count() > 1]
```
sloria: Add `Q('is_deleted', 'ne', True)` to make this script idempotent.

PR #6905, opened 2017-02-25
[Hotfix][Follow-up] Bookmark Collection Uniqueness [OSF-7551]

mfraezz:
## Purpose
Follow up to #6904 

## Changes
* Uncomment uniqueness constraint

## Side effects
None expected

## Ticket


mfraezz:
Constraint wouldn't have worked as intended -- mongo has no support for what we wanted this to do.

PR #6906, opened 2017-02-27
Add node-specific entries to gitignore

sloria:
Copied from https://github.com/github/gitignore/blob/master/Node.gitignore


.gitignore
```diff
@@ -44,6 +44,66 @@ mathenv
...
+bower_components
+
+# node-waf configuration
+.lock-wscript
+
+# Compiled binary addons (http://nodejs.org/api/addons.html)
+build/Release
+
+# Dependency directories
+node_modules/
```
icereval: cannot recall off hand, does `node_modules/` match nested sub directories? asking because I've also seen `**/node_modules` before but maybe that was just another way of saying the same thing?

Example: admin has its own package.json -> `/admin/node_modules`

sloria: Confirmed: nested `node_modules` will be ignored.

sloria:
Removed a couple duplicate/unnecessary entries. Should be good to go.

PR #6908, opened 2017-02-27
Add log when user adds or removes a private link [OSF-7309]

erinspace:
## Purpose
Add logs whenever a view-only link is created on a project or component. Currently, a user could create a view-only link, share that with others, allowing them to access the project. The user could then remove the VOL and his collaborators would never know that the project's contents had been shared. 

Create new logs that say when a normal or an anonymous view only link is added or removed.

screen shot 2017-02-27 at 12 06 32 pm

## Changes

- Add `VIEW_ONLY_LINK_ADDED` and `VIEW_ONLY_LINK_REMOVED` to the `NodeLog` model
- Add `anonymous_link` to the NodeLog serializer
- Add log when node link is added or removed
- Tests to make sure logs are added
- Log text added to log actions list
- Log kwarg `anonymous_link` parsed in `LogTextParser.js`

## Side effects

None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-7309

### DIFF
--- a/website/project/views/node.py
+++ b/website/project/views/node.py
@@ -14,7 +14,7 @@ from framework import status
 from framework.utils import iso8601format
 from framework.flask import redirect
 from framework.auth.decorators import must_be_logged_in, collect_auth
-from framework.exceptions import HTTPError, PermissionsError
+from framework.exceptions import HTTPError
 from osf.models.nodelog import NodeLog
 
 from website import language
--- a/website/project/views/node.py
+++ b/website/project/views/node.py
@@ -14,7 +14,8 @@ from framework import status
 from framework.utils import iso8601format
 from framework.flask import redirect
 from framework.auth.decorators import must_be_logged_in, collect_auth
-from framework.exceptions import HTTPError
+from framework.exceptions import HTTPError, PermissionsError
+from osf.models.nodelog import NodeLog
 
 from website import language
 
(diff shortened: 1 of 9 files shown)


brianjgeiger:
Fix merge conflicts when you have a moment, please.

website/project/views/node.py
```diff
@@ -605,6 +606,21 @@ def remove_private_link(*args, **kwargs):
         link = PrivateLink.load(link_id)
         link.is_deleted = True
         link.save()
+
+        for node in link.nodes.all():
```
erinspace: As for the django-y ness, normally I'd be worried about this here, but I do need to add a log to each node, so can't use something like `values` or `values_list` I don't think....

website/project/views/node.py
```diff
@@ -596,6 +597,21 @@ def remove_private_link(*args, **kwargs):
...
+            log_dict = {
+                'project': node.parent_id,
```
sloria: Isn't `parent_id` the pk of the parent, not the GUID?

erinspace: [Seems to be the _id?](https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/models/node.py#L967). Also, I was following the example of [other node logs that use node.parent_id](https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/models/node.py#L1081)

sloria: Ah, cool cool. Just making sure.

erinspace:
@brianjgeiger merge conflict free!

PR #6913, opened 2017-02-27
[WIP][Feature] Fix dupe comment notifications [OSF-6528]

alexschiller:
## Purpose
Comments send dupe emails if they are replies or at mentions. This fixes that.

## Changes
1. Fix bug where user object was being checked against a list of string guids.
2. Add an exclude context variable to stop dupes.

## Side effects
1. Greatly refactored/tidied code (no more 3 nested deep if statements in a for loop)
2. Added line comments.

## Code Notes
Exclude includes the user._id, which means checks for target_user._id and checks if user._id are no longer necessary as they are filtered in the list comprehension.

## Ticket
[#OSF-6528]
https://openscience.atlassian.net/browse/OSF-6528

## QA Notes

pending

### DIFF
--- a/osf/models/comment.py
+++ b/osf/models/comment.py
@@ -162,6 +162,7 @@ class Comment(GuidMixin, SpamMixin, CommentableMixin, BaseModel):
 
         log_dict.update(comment.root_target.referent.get_extra_log_params(comment))
 
+        new_mentions = []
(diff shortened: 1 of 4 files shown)


alexschiller:
Isn't updating on pushes. Closing and reopening new one.

PR #6914, opened 2017-02-27
[Django] Bookmark collection uniqueness [OSF-7551]

mfraezz:
## Purpose
Port #6904 for Django

## Changes
* Port script
* Port model / utils changes
* Add migration to create a partial uniqueness index

## Side effects
None expected

## Deployment Notes
`python -m scripts.migration.ensure_bookmark_uniqueness` 
must be ran before 
`python manage.py migrate`

This can be done while avoiding the autodeployment with #6918 

## Ticket


osf/migrations/0037_auto_20170227_1522.py
```diff
@@ -0,0 +1,24 @@
...
+        cursor.execute(sql)
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('osf', '0036_merge'),
+    ]
+
+    operations = [
+        migrations.RunPython(add_partial_uniqueness_index)
```
cwisecarver: The second parameter will allow you to apply a reverse migration to remove the index. Always a good idea to write reverse migrations.

scripts/migration/ensure_bookmark_uniqueness.py
```diff
@@ -1,20 +1,22 @@
...
 from website.app import init_app
-from website.project import Node
 
 
 logger = logging.getLogger(__name__)
 
 def get_targets():
     logger.info('Acquiring targets...')
-    targets = [u for u in User.find() if Node.find(Q('is_bookmark_collection', 'eq', True) & Q('is_deleted', 'eq', False) & Q('contributors', 'eq', u._id)).count() > 1]
+    targets = [u for u in OSFUser.find() if Collection.find(Q('is_bookmark_collection', 'eq', True) & Q('is_deleted', 'eq', False) & Q('creator', 'eq', u)).count() > 1]
```
cwisecarver: You should be able to do something like this: `OSFUser.objects.filter(collection__is_bookmark_collection=True, collection__is_deleted=False, collection__creator_id=F('id')).aggregate(col_count=Count('collection'))`

But since it's just a migration it's probably fine.

osf/models/user.py
```diff
@@ -746,6 +748,8 @@ def save(self, *args, **kwargs):
             self.update_search()
             # TODO
             # self.update_search_nodes_contributors()
+        if first_save:
```
cwisecarver: If you did this in a signal instead of in save you'd get a `created` function param telling you that it was created. Check the bottom of `osf/models/node.py` for an example.

cwisecarver: That's where we create the parent_node node relation of a node.

website/views.py
```diff
@@ -120,7 +120,7 @@ def index():
 
 def find_bookmark_collection(user):
     Collection = apps.get_model('osf.Collection')
-    return Collection.objects.get(creator=user, is_deleted=False)
+    return Collection.objects.get(creator=user, is_deleted=False, is_bookmark_collection=True)
```
cwisecarver: This won't ever fail to return, right? Because you'll get an exception if it does. But from the diff, you would always.

mfraezz: It should never fail to return, assuming that a user is actually passed in.

osf/models/user.py
```diff
@@ -1393,3 +1396,10 @@ class Meta:
         permissions = (
             ('view_user', 'Can view user details'),
         )
+
+@receiver(post_save, sender=OSFUser)
+def create_bookmark_collection(sender, instance, created, **kwargs):
+    if created:
+        new_bookmark_collection(instance)
```
mfraezz: For some reason, creating a bookmark collection when a user is created leads to incorrect contributor permissions for just about everything. It could be that writes aren't happening on the contributor table, or the MRM is caching outdated values, or something else entirely.
```
#TODO: Investigate further tomorrow
```

cwisecarver:
Something!

PR #6923, opened 2017-02-28
[RFR] Add custom caching decorator and implement on various properties

cwisecarver:
## Purpose

Make less guid queries

## Changes

Add custom cached_property decorator that supports setters. Implement on GuidMixin._id and Node.parent_node and Node.root.

## Side effects

Could cause setting guids on existing objects to not work good.

### DIFF
--- a/osf/utils/caching.py
+++ b/osf/utils/caching.py
@@ -17,7 +17,7 @@ class _CachedProperty(property):
 
     def __init__(self, fget, fset=None, fdel=None, doc=None):
         """Initializes the cached property."""
-        self._cache_name = "_{name}_cache".format(
+        self._cache_name = '_{name}_cache'.format(
             name=fget.__name__,
         )
         # Wrap the accessors.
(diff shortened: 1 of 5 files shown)


cwisecarver:
Passed on my local TCI.

cwisecarver:
https://travis-ci.org/cwisecarver/osf.io/builds/207507731

osf/utils/caching.py
```diff
@@ -0,0 +1,63 @@
...
+The cache is stored on the model as a protected attribute. Expensive
+property lookups, such as database access, can therefore be sped up
+when accessed multiple times in the same request.
+The property can also be safely set and deleted without interference.
+"""
+from __future__ import unicode_literals
+
+from functools import wraps
+
+# from https://github.com/etianen/django-optimizations/blob/master/src/optimizations/propertycache.py
```
sloria: This code is BSD-3 licensed. Might need to put the license text in our `NOTICE` file: https://github.com/etianen/django-optimizations/blob/master/LICENSE

cwisecarver: Will do.

PR #6924, opened 2017-02-28
Store root in the DB

sloria:
## Purpose

Optimize accessing node's roots

## Changes

Calculate and store a Node's root when it saves

## Side effects

We should be able to optimize getting all of a node's descendants by using the `related_name` (`descendants`). Any relevant methods should be updated.


osf/migrations/0038_abstractnode_root.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9 on 2017-02-28 20:20
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
```
icereval: best to make a django data migration that runs the root cte on all nodes, this will fill in dev's local environments and staging w/o a full migration.

icereval: safe either way 👍

sloria: Good idea. Alternatively, I could add a check in the post save handler to ensure that root is always set:

```python
if created or not instance.root:
     instance.get_root()
```

icereval: i'd prefer not to see the performance hit we are trying to fix on newrelic randomly

icereval: but do agree it might be worth while to add a last minute fail safe

sloria: K; will do both

sloria:
@icereval Changes made.

osf/models/node.py
```diff
@@ -2961,3 +2961,6 @@ def set_parent(sender, instance, created, *args, **kwargs):
             del instance.__dict__['parent_node']
         except KeyError:
             pass
+    if created or not instance.root:
```
icereval: nitpick: why not just `if not instance.root:  # should only trigger on create`

sloria:
@icereval Made the changes and squashed.

sloria:
@icereval I've fixed the failing test that is relevant to this change. The remaining failing test is unrelated and exists on master and develop (I'll look into that).

PR #6929, opened 2017-03-01
[EOSF-470] Fix twitter link and add sign-up logo

hmoco:
## Purpose 
 Fix twitter link and add sign-up logo

## Ticket
https://openscience.atlassian.net/browse/EOSF-470


abought:
Changes look good! There is one CI failure but it may be unrelated.

abought:
One side note for future reference: in looking at the folder of preprint provider logos, it seems there are no consistent conventions for image size.

For example, if the provider logo were constrained to be 120px tall, then a 1080px tall logo (BITSS) or 3,586 px tall logo (socarxiv-login) would both be overkill, to different degrees. Sending more data than the user needs for the image increases download times.

Also, why is the BITSS logo 600 KB? That seems like an unusually large file size for what is basically a black-and-white image.

brianjgeiger:
Yes, please shrink the images. 

.:Brian J. Geiger

hmoco:
@abought image is now 27kb!

sloria:
NOTE: The commits from this PR were cherry-picked into a hotfix branch and merged in b84cd8f291

PR #6930, opened 2017-03-01
Use c lib for datetime parsing

cwisecarver:
## Purpose

dateutil.parser was 48% of the profile

## Changes

Replace dateutil.parser with C library

## Side effects

Speed.


cwisecarver:
This passed on my local travis.

PR #6933, opened 2017-03-02
YAJL JSON renderer instead of UJSON

cwisecarver:
## Purpose

[Make JSON rendering faster.](https://github.com/esnme/ultrajson#benchmarks)

## Changes

Switch JSON rendering to use YAJL 

## Side effects

Lists could be misordered.

### DIFF
--- a/api/base/renderers.py
+++ b/api/base/renderers.py
@@ -1,8 +1,8 @@
 import re
 
+import yajl as yajl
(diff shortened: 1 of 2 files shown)


cwisecarver:
yajl doesn't work

icereval:
:-(

PR #6935, opened 2017-03-03
API log list optimizations and .root fix

sloria:
## Purpose

Make APIv2 logs/ endpoints faster.

## Changes

* Fixes bug where forked nodes' `root` would be incorrect
* Simplify and optimize `AbstractNodeQuerySet.get_children` by using the the `descendants` related name rather than a recursive query
* Use `.values` in NodeLogParamsSerializer to avoid instantiation and select overhead
* Use `select_related` to eagerly select NodeLogSerializer's relationship fields

## Side effects

~2 second speedup for the `/v2/nodes/<nid>/logs/` endpoint

Before

node_log_list_ _django_rest_framework

After

node_log_list_ _django_rest_framework

### DIFF
--- a/api/logs/serializers.py
+++ b/api/logs/serializers.py
@@ -99,15 +99,15 @@ class NodeLogParamsSerializer(RestrictedDictSerializer):
     def get_params_node(self, obj):
         node_id = obj.get('node', None)
         if node_id:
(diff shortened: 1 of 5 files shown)


sloria:
The TCI build on my fork passed.

PR #6937, opened 2017-03-06
[HOLD FOR DEVOPS] Add django-cacheops -- UPDATED

cwisecarver:
## Purpose

Reduce queries to postgres by caching querysets in redis.

## Changes

- Installed django-cacheops
- Removed GuidMixinQuerySet annotations that were conflicting with cacheops
- Added redis to docker-compose.yml

## Side effects

- Faster things? 
- Developers angry and confused.

@icereval If you could check the docker stuff. I tried but unrelated (accidentally upgraded postgres and fell into a docker rabbit hole of failure) errors made me give up after an hour of trying.

# This does not technically require redis to be running. It's disabled in defaults.py by default.


cwisecarver:
@icereval New PR, cherry picked just the relevant commits.

docker-compose.yml
```diff
@@ -90,6 +90,13 @@ services:
     stdin_open: true
     tty: true
 
+  redis:
```
icereval: can you merge this in https://github.com/cwisecarver/osf.io/pull/10

osf/models/base.py
```diff
@@ -506,129 +478,159 @@ def remove_guid_annotations(self):
             if table_name in self.query.tables:
                 del self.query.tables[self.query.tables.index(table_name)]
 
-    def _clone(self, annotate=False, **kwargs):
-        query = self.query.clone()
-        if self._sticky_filter:
-            query.filter_is_sticky = True
-        if annotate:
+    if not settings.CACHEOPS_ENABLED:
```
icereval: Mind making this more like a pythonic preprocessor definition? e.g. https://stackoverflow.com/questions/482014/how-would-you-do-the-equivalent-of-preprocessor-directives-in-python

```python
if not settings.CACHEOPS_ENABLED:
    class GuidMixinQuerySet(object):

        def annotate_query_with_guids(self):
            return self

        # ...
else:
    class GuidMixinQuerySet(object):
        # ...
```

icereval: reviewing this a bit closer maybe some of the fetch/clone code is used?

docker-compose.yml
cwisecarver: There are conflicts too complicated to resolve on GitHub, you want me to do it or you want to?

PR #6940, opened 2017-03-06
Fix Parent Filtering on Nodes [OSF-7578]

caseyrollins:
#### Purpose
- Filtering nodes by parent node is broken on the user nodes list view (`/v2/users/me/nodes/`) and institution nodes list view (`/v2/institutions/<id>/nodes/`). 
  - Both routes currently return a `502` on staging.
- Filtering by a parent node works as expected on the nodes list view (`/v2/nodes/`) because of the ODM filter logic [here](https://github.com/CenterForOpenScience/osf.io/blob/develop/api/nodes/views.py#L297).
- Filtering registrations by parent is also broken.

#### Changes
- Adds a `NodesListFilterMixin` class that contains the [existing logic](https://github.com/CenterForOpenScience/osf.io/blob/develop/api/nodes/views.py#L297) for the parent filter to work correctly. 
- Adds the `NodesListFilterMixin` class to views that return a serialized list of nodes/registrations (nodes list, institution nodes list, user nodes list, registrations list).
- Adds `NodesListFilteringMixin` and `RegistrationListFilteringMixin` classes to test filtering on the views mentioned above.

#### Side effects
- None expected.

#### Ticket
- [OSF-7578](https://openscience.atlassian.net/browse/OSF-7578)


api/nodes/filters.py
```diff
@@ -3,10 +3,36 @@
...
+        # a query from that
+        if operation['source_field_name'] == 'root':
+            child_pks = []
+            for root_guid in operation['value']:
+                if not root_guid:
+                    raise InvalidFilterValue(value=root_guid)
+                root = utils.get_object_or_error(Node, root_guid, display_name='root', prefetch_fields=self.serializer_class().model_field_names)
+                child_pks.extend(Node.objects.get_children(root=root, primary_keys=True))
+            return Q('id', 'in', child_pks)
+        elif operation['source_field_name'] == 'parent_node':
```
cwisecarver: `parent_nodes` is a reverse relationship on Node. If you have the guid of the parent you're looking for you can do `Node.objects.filter(parent_nodes__guids___id=parent_guid)` and that will get you a queryset of all the nodes that's parent guid is `parent_guid`. I don't know how to translate that into a MODM Q object.

api/nodes/filters.py
```diff
@@ -3,10 +3,36 @@
...
+from osf.models import AbstractNode as Node
+
+
+class NodesListFilterMixin(ODMFilterMixin):
+
+    def _operation_to_query(self, operation):
+        # We special case filters on root because root isn't a field; to get the children
+        # of a root, we use a custom manager method, Node.objects.get_children, and build
+        # a query from that
+        if operation['source_field_name'] == 'root':
```
cwisecarver: Root is a field now so this has been obviated.

api_tests/registrations/filters/test_filters.py
```diff
@@ -0,0 +1,57 @@
...
+
+from osf.models import Node
+
+
+class RegistrationListFilteringMixin(object):
+
+    def setUp(self):
+        super(RegistrationListFilteringMixin, self).setUp()
+
+        self._setUp()
```
sloria: Minor: this shouldn't be necessary.

```python
class RegistrationListFilteringMixin(object):

    def setUp(self):
        super(..).setUp(...)
        assert self.url, 'Subclasses of RegistrationListFilteringMixin must define self.url'
        assert self.node_B2, '...'
        # enforce necessary attributes
        self.parent_url = '{}filter[parent]='.format(self.url)
        self.root_url ='{}filter[root]='.format(self.url)

class TestSomething(RegistrationListFilteringMixin, ...):

    url = '/api/v2/registrations/'

    def setUp(self):
        super(..).setUp(...)
        #... fixtures, etc.
```

api/nodes/filters.py
```diff
@@ -3,10 +3,36 @@
...
+            for root_guid in operation['value']:
+                if not root_guid:
+                    raise InvalidFilterValue(value=root_guid)
+                root = utils.get_object_or_error(Node, root_guid, display_name='root', prefetch_fields=self.serializer_class().model_field_names)
+                child_pks.extend(Node.objects.get_children(root=root, primary_keys=True))
+            return Q('id', 'in', child_pks)
+        elif operation['source_field_name'] == 'parent_node':
+            if operation['value']:
+                parent = utils.get_object_or_error(Node, operation['value'], display_name='parent', prefetch_fields=self.serializer_class().model_field_names)
+                return Q('parent_nodes', 'eq', parent.id)
```
sloria: `parent_nodes` includes parents that have a node link to the node. These should be excluded.

sloria: Please add a test for this case (excluding node link parents).

api/nodes/filters.py
sloria: Maybe `Q('parent_nodes__guids___id', 'eq', parent_guid)`?

api_tests/registrations/filters/test_filters.py
alexschiller: The setup for the preprint_provider tests is different since all nodes must be the same provider, which is also why one of the tests is stubbed to pass.

caseyrollins:
@sloria @cwisecarver Updated.

api/nodes/filters.py
```diff
@@ -13,23 +13,18 @@
...
-        elif operation['source_field_name'] == 'parent_node':
-            if operation['value']:
-                parent = utils.get_object_or_error(Node, operation['value'], display_name='parent', prefetch_fields=self.serializer_class().model_field_names)
-                return Q('parent_nodes', 'eq', parent.id)
-            else:
-                return Q('parent_nodes', 'isnull', True)
+            if None in operation['value']:
+                raise InvalidFilterValue()
+            return functools.reduce(operator.or_, [
+                Q('root__guids___id', 'eq', value)
```
cwisecarver: Did you try: `Q('root__guids___id', 'in', operation['value'])`? Assuming that `operation['value']` is a list of strings I think that would give the same result without the list comp or the reduce.

caseyrollins: Ah, hadn't tried this. It does work and I made that change in 915b8ea. 

Also updated the `parent` case to (properly) exclude linked nodes. 

Tests will (maybe, hopefully) pass this time.

PR #6941, opened 2017-03-06
[feature] Fix api post process [OSF-7584]

alexschiller:
## Purpose
Api preprint filter routes are 503ing

## Changes
Fixes the broken filters

## Side effects
None known

## Ticket
[#OSF-7584]

https://openscience.atlassian.net/browse/OSF-7584

## Qa Notes:
test these routes:

https://staging-api.osf.io/v2/preprints/?filter[id]=<id>
https://staging-api.osf.io/v2/nodes/<node_id>/preprints/?filter[id]=<id>
https://staging-api.osf.io/v2/preprint_providers/preprints/?filter[id]=<id>
https://staging-api.osf.io/v2/nodes/<node_id>/preprints/?filter[provider]=<provider_id>
https://staging-api.osf.io/v2/preprint_providers/preprints/?filter[provider]=<provider_id>

### DIFF
--- a/api/nodes/views.py
+++ b/api/nodes/views.py
@@ -3362,7 +3362,7 @@ class NodePreprintsList(JSONAPIBaseView, generics.ListAPIView, NodeMixin, NodePr
             operation['source_field_name'] = 'provider___id'
 
         if field_name == 'id':
-            operation['source_field_name'] = '_guids___id'
+            operation['source_field_name'] = 'guids___id'
 
     # overrides ODMFilterMixin
     def get_default_odm_query(self):
--- a/api/nodes/views.py
+++ b/api/nodes/views.py
@@ -3357,6 +3357,13 @@ class NodePreprintsList(JSONAPIBaseView, generics.ListAPIView, NodeMixin, NodePr
     view_category = 'nodes'
     view_name = 'node-preprints'
 
+    def postprocess_query_param(self, key, field_name, operation):
+        if field_name == 'provider':
+            operation['source_field_name'] = 'provider___id'
+
+        if field_name == 'id':
+            operation['source_field_name'] = '_guids___id'
+
     # overrides ODMFilterMixin
     def get_default_odm_query(self):
         return (
--- a/api/preprint_providers/views.py
+++ b/api/preprint_providers/views.py
@@ -181,7 +181,7 @@ class PreprintProviderPreprintList(JSONAPIBaseView, generics.ListAPIView, ODMFil
             operation['source_field_name'] = 'provider___id'
 
         if field_name == 'id':
-            operation['source_field_name'] = '_guids___id'
+            operation['source_field_name'] = 'guids___id'
 
     # overrides ODMFilterMixin
     def get_default_odm_query(self):
--- a/api/preprint_providers/views.py
+++ b/api/preprint_providers/views.py
@@ -176,6 +176,13 @@ class PreprintProviderPreprintList(JSONAPIBaseView, generics.ListAPIView, ODMFil
     view_category = 'preprint_providers'
     view_name = 'preprints-list'
 
+    def postprocess_query_param(self, key, field_name, operation):
+        if field_name == 'provider':
+            operation['source_field_name'] = 'provider___id'
+
+        if field_name == 'id':
+            operation['source_field_name'] = '_guids___id'
+
     # overrides ODMFilterMixin
     def get_default_odm_query(self):
         # TODO: this will return unpublished preprints so that users
--- a/api/preprints/views.py
+++ b/api/preprints/views.py
@@ -160,7 +160,7 @@ class PreprintList(JSONAPIBaseView, generics.ListCreateAPIView, ODMFilterMixin):
         if field_name == 'provider':
             operation['source_field_name'] = 'provider___id'
         if field_name == 'id':
-            operation['source_field_name'] = '_guids___id'
+            operation['source_field_name'] = 'guids___id'
 
     def get_serializer_class(self):
         if self.request.method == 'POST':
--- a/api/preprints/views.py
+++ b/api/preprints/views.py
@@ -157,18 +157,10 @@ class PreprintList(JSONAPIBaseView, generics.ListCreateAPIView, ODMFilterMixin):
 
     # overrides FilterMixin
     def postprocess_query_param(self, key, field_name, operation):
-        # tag queries will usually be on Tag.name,
-        # ?filter[tags]=foo should be translated to Q('tags__name', 'eq', 'foo')
-        # But queries on lists should be tags, e.g.
-        # ?filter[tags]=foo,bar should be translated to Q('tags', 'isnull', True)
-        # ?filter[tags]=[] should be translated to Q('tags', 'isnull', True)
-        if field_name == 'tags':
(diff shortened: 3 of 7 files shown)


api_tests/preprints/filters/test_filters.py
```diff
@@ -0,0 +1,113 @@
...
+from nose.tools import *  # flake8: noqa
+import pytest
+from mock import MagicMock
+
+class PreprintsListFilteringMixin(object):
+
+    def setUp(self):
+        super(PreprintsListFilteringMixin, self).setUp()
+
+        self._setUp()
```
sloria: See https://github.com/CenterForOpenScience/osf.io/pull/6940/files#r106475339 

Not a blocker, but could be refactored.

api/nodes/views.py
```diff
@@ -3357,6 +3357,13 @@ class NodePreprintsList(JSONAPIBaseView, generics.ListAPIView, NodeMixin, NodePr
     view_category = 'nodes'
     view_name = 'node-preprints'
 
+    def postprocess_query_param(self, key, field_name, operation):
+        if field_name == 'provider':
+            operation['source_field_name'] = 'provider___id'
+
+        if field_name == 'id':
+            operation['source_field_name'] = '_guids___id'
```
cwisecarver: `_guids` doesn't exist. I think you want `guids___id`.

api/preprint_providers/views.py
```diff
@@ -176,6 +176,13 @@ class PreprintProviderPreprintList(JSONAPIBaseView, generics.ListAPIView, ODMFil
     view_category = 'preprint_providers'
     view_name = 'preprints-list'
 
+    def postprocess_query_param(self, key, field_name, operation):
+        if field_name == 'provider':
+            operation['source_field_name'] = 'provider___id'
+
+        if field_name == 'id':
+            operation['source_field_name'] = '_guids___id'
```
cwisecarver: Again, just `guids___id`

api/preprints/views.py
```diff
@@ -157,18 +157,10 @@ class PreprintList(JSONAPIBaseView, generics.ListCreateAPIView, ODMFilterMixin):
...
-        # ?filter[tags]=[] should be translated to Q('tags', 'isnull', True)
-        if field_name == 'tags':
-            if operation['value'] not in (list(), tuple()):
-                operation['source_field_name'] = 'tags__name'
-                operation['op'] = 'iexact'
-
         if field_name == 'provider':
             operation['source_field_name'] = 'provider___id'
+        if field_name == 'id':
+            operation['source_field_name'] = '_guids___id'
```
cwisecarver: Again, `guids___id`

api_tests/preprint_providers/views/test_preprint_provider_preprints_list.py
```diff
@@ -0,0 +1,48 @@
...
+        self.preprint_three = PreprintFactory(creator=self.user, filename='darn.txt', provider=self.provider, subjects=[[self.subject._id], [self.subject_two._id]])
+        self.preprint_three.date_created = '2013-12-11 10:09:08.070605+00:00'
+        self.preprint_three.date_published = '2013-12-11 10:09:08.070605+00:00'
+        self.preprint_three.is_published = False
+        self.preprint_three.save()
+
+        self.url = '/{}preprint_providers/{}/preprints/?version=2.2&'.format(API_BASE, self.provider._id)
+
+    def test_provider_filter_equals_returns_one(self):
+      pass
```
sloria: Finish this or remove?

api_tests/preprints/filters/test_filters.py
```diff
@@ -0,0 +1,113 @@
+# -*- coding: utf-8 -*-
+from nose.tools import *  # flake8: noqa
+import pytest
+from mock import MagicMock
+
+class PreprintsListFilteringMixin(object):
```
sloria: Why is this mixin necessary? It's only used in one class.

api_tests/preprints/filters/test_filters.py
sloria: Actually, I don't think this class is necessary: see https://github.com/CenterForOpenScience/osf.io/pull/6941/files#r106481230

api_tests/preprints/filters/test_filters.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def test_provider_filter_equals_returns_one(self):
+        expected = [self.preprint_two._id]
+        res = self.app.get('{}{}'.format(self.provider_url, self.provider_two._id), auth=self.user.auth)
+        actual = [preprint['id'] for preprint in res.json['data']]
+        assert_equal(expected, actual)
+
+    def test_provider_filter_equals_returns_multiple(self):
+        expected = set([self.preprint._id, self.preprint_three._id])
+        res = self.app.get('{}{}'.format(self.provider_url, self.provider._id), auth=self.user.auth)
```
sloria: I'm not sure I understand this test. Why do you expect to receive multiple results if you are only filtering on one _id?

api_tests/preprints/filters/test_filters.py
```diff
@@ -0,0 +1,113 @@
...
+        actual = set([preprint['id'] for preprint in res.json['data']])
+        assert_equal(expected, actual)
+
```
sloria: Can you add a test for filtering on multiple ids, e.g `filter[id]=abc12,def34`

api/nodes/views.py
sloria: I think Chris is right, though I'm not sure why the tests would be passing if this field is incorrect. Are the tests testing what we think they are testing? @alexschiller Please investigate.

api_tests/preprint_providers/views/test_preprint_provider_preprints_list.py
alexschiller: The tests are from the mixin. This test doesn't apply but all the others do. Is there a better way to exclude it?

api/nodes/views.py
alexschiller: per wisecarver: "Those fields aren’t real fields they’re annotated onto the query specifically to reduce extra queries and are not guaranteed to be there." 

They do work and the filters do work properly locally. I'll swap to guids___id to be correct.

api_tests/preprints/filters/test_filters.py
alexschiller: Replied in wrong place, adding back here: The setup for the preprint_provider tests is different since all nodes must be the same provider, which is also why one of the tests is stubbed to pass.

api_tests/preprints/filters/test_filters.py
alexschiller: These are provider tests not id tests, so yes, you would expect a provider to return more than one preprint. Note the id tests below only filter for one.

api_tests/preprint_providers/views/test_preprint_provider_preprints_list.py
alexschiller: I'll actually pull out the tests that can't pass generally and implement versions in the subclass.

api_tests/preprints/filters/test_filters.py
sloria: Oh, I see; I misread the test.

PR #6943, opened 2017-03-07
[RFR] Make it more faster pls (API) v2

cwisecarver:
## Purpose

But Faster. 

## Changes

- Pass data we already had to embeds so no extra queries.
- Add the ability to eagerly load fk and m2m fields passed to `.eager(['field', 'field1'])`
- Pycharm reformatted some things for me.
## Side effects

Nothing whatsoever could work.

### DIFF
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -248,6 +248,7 @@ class DateByVersion(ser.DateTimeField):
     """
     Custom DateTimeField that forces dates into the ISO-8601 format with timezone information in version 2.2.
     """
+
     def to_representation(self, value):
         request = self.context.get('request')
         if request:
@@ -274,7 +275,9 @@ class IDField(ser.CharField):
             if request.method in utils.UPDATE_METHODS and not utils.is_bulk_request(request):
                 id_field = self.get_id(self.root.instance)
                 if id_field != data:
-                    raise Conflict(detail=('The id you used in the URL, "{}", does not match the id you used in the json body\'s id field, "{}". The object "{}" exists, otherwise you\'d get a 404, so most likely you need to change the id field to match.'.format(id_field, data, id_field)))
+                    raise Conflict(detail=(
+                    'The id you used in the URL, "{}", does not match the id you used in the json body\'s id field, "{}". The object "{}" exists, otherwise you\'d get a 404, so most likely you need to change the id field to match.'.format(
+                        id_field, data, id_field)))
         return super(IDField, self).to_internal_value(data)
 
     def get_id(self, obj):
@@ -301,7 +304,9 @@ class TypeField(ser.CharField):
             type_ = self.root.Meta.type_
 
         if type_ != data:
-            raise Conflict(detail=('This resource has a type of "{}", but you set the json body\'s type field to "{}". You probably need to change the type field to match the resource\'s type.'.format(type_, data)))
+            raise Conflict(detail=(
+            'This resource has a type of "{}", but you set the json body\'s type field to "{}". You probably need to change the type field to match the resource\'s type.'.format(
+                type_, data)))
         return super(TypeField, self).to_internal_value(data)
 
 
@@ -318,7 +323,9 @@ class TargetTypeField(ser.CharField):
 
     def to_internal_value(self, data):
         if self.target_type != data:
-            raise Conflict(detail=('The target resource has a type of "{}", but you set the json body\'s type field to "{}".  You probably need to change the type field to match the target resource\'s type.'.format(self.target_type, data)))
+            raise Conflict(detail=(
+            'The target resource has a type of "{}", but you set the json body\'s type field to "{}".  You probably need to change the type field to match the target resource\'s type.'.format(
+                self.target_type, data)))
         return super(TargetTypeField, self).to_internal_value(data)
 
 
@@ -484,7 +491,8 @@ class RelationshipField(ser.HyperlinkedIdentityField):
             # Ignore related_counts for these fields
             fetched_field = self.parent.fields.get(count_field)
 
-            hidden = fetched_field and isinstance(fetched_field, HideIfWithdrawal) and getattr(value, 'is_retracted', False)
+            hidden = fetched_field and isinstance(fetched_field, HideIfWithdrawal) and getattr(value, 'is_retracted',
+                                                                                               False)
 
             if not hidden and count_field not in countable_fields:
                 raise InvalidQueryStringError(
@@ -508,11 +516,13 @@ class RelationshipField(ser.HyperlinkedIdentityField):
                 field_counts_requested = self.process_related_counts_parameters(show_related_counts, value)
 
                 if utils.is_truthy(show_related_counts):
-                    meta[key] = website_utils.rapply(meta_data[key], _url_val, obj=value, serializer=self.parent, request=self.context['request'])
+                    meta[key] = website_utils.rapply(meta_data[key], _url_val, obj=value, serializer=self.parent,
+                                                     request=self.context['request'])
                 elif utils.is_falsy(show_related_counts):
                     continue
                 elif self.field_name in field_counts_requested:
-                    meta[key] = website_utils.rapply(meta_data[key], _url_val, obj=value, serializer=self.parent, request=self.context['request'])
+                    meta[key] = website_utils.rapply(meta_data[key], _url_val, obj=value, serializer=self.parent,
+                                                     request=self.context['request'])
                 else:
                     continue
             elif key == 'projects_in_common':
@@ -520,9 +530,11 @@ class RelationshipField(ser.HyperlinkedIdentityField):
                     continue
                 if not self.context['request'].query_params.get('show_projects_in_common', False):
                     continue
-                meta[key] = website_utils.rapply(meta_data[key], _url_val, obj=value, serializer=self.parent, request=self.context['request'])
+                meta[key] = website_utils.rapply(meta_data[key], _url_val, obj=value, serializer=self.parent,
+                                                 request=self.context['request'])
             else:
-                meta[key] = website_utils.rapply(meta_data[key], _url_val, obj=value, serializer=self.parent, request=self.context['request'])
+                meta[key] = website_utils.rapply(meta_data[key], _url_val, obj=value, serializer=self.parent,
+                                                 request=self.context['request'])
         return meta
 
     def lookup_attribute(self, obj, lookup_field):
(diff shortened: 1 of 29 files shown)


cwisecarver:
Tests passed on my travis.

api/base/serializers.py
```diff
@@ -1003,7 +1009,19 @@ def is_valid(self, clean_html=True, **kwargs):
         return ret
 
 
-class JSONAPISerializer(ser.Serializer):
+class PrefetchRelationshipsSerializer(ser.Serializer):
+    relationship_field_names = list()
```
sloria: This probably shouldn't be defined on the class--it should not be shared across subclasses.

api/base/serializers.py
```diff
@@ -1003,7 +1009,19 @@ def is_valid(self, clean_html=True, **kwargs):
...
+
+    def __init__(self, *args, **kwargs):
+        super(PrefetchRelationshipsSerializer, self).__init__(*args, **kwargs)
+        for field_name, field in self.fields.iteritems():
+            if field.source == '*':
+                self.relationship_field_names.append(field_name)
```
sloria: Just so I understand this correctly: `relationship_field_names` is actually *all* field names? And the non-relationship fields just get ignored by `eager`? If that's the case, `relationship_field_names` is a misnomer; probably should be `field_names`. Also, `PrefetchRelationshipsSerializer` doesn't have any prefetching behavior, so we can probably get rid of it and just put this code on `JSONAPISerializer`.

sloria: Also, is it necessary to store all the field names on the serializer?

Wherever we would want to fetch an object with prefetching, couldn't we do something like

```python
prefetch_fields=Node.get_fk_field_names() + Node.get_m2m_field_names()
```

Of course, you could refactor that using a cached property, e.g. `Node.relationship_field_names`.

api/base/serializers.py
chrisseto: 👍 

```python
In [1]: class TestA:
   ...:     my_list = []
   ...:

In [2]: class TestB(TestA):
   ...:     pass
   ...:

In [3]: TestB.my_list.append('Foo')

In [4]: TestA.my_list
Out[4]: ['Foo']
```

api/base/utils.py
```diff
@@ -70,15 +71,31 @@ def absolute_reverse(view_name, query_kwargs=None, args=None, kwargs=None):
...
+                query = {model_cls.primary_identifier_name: query_or_pk}
+            else:
+                obj = model_cls.load(query_or_pk)
     else:
+        if hasattr(model_cls, 'primary_identifier_name'):
+            query = to_django_query(query_or_pk, model_cls=model_cls)
+        else:
+            obj = model_cls.find_one(query_or_pk)
+
+    if not obj:
```
chrisseto: It might be a touch more understandable if you short circuit upon getting `obj`

osf/models/base.py
```diff
@@ -69,6 +69,13 @@ def __iter__(self):
             else:
                 yield item
 
+    def eager(self, fields):
```
cwisecarver: 👍

api/base/serializers.py
chrisseto: @sloria I think `self.fields` is the subset of fields on the serializer. There's probably a fair about of relations that aren't exposed via the API that we wouldn't want to join/prefetch.

osf/models/base.py
```diff
@@ -659,10 +677,8 @@ def _id(self, value):
     @classmethod
     def load(cls, q):
         try:
-            content_type = ContentType.objects.get_for_model(cls)
-            # if referent doesn't exist it will return None
-            return Guid.objects.get(_id=q, content_type=content_type).referent
-        except Guid.DoesNotExist:
+            return cls.objects.filter(guids___id=q)[0]
+        except (IndexError, cls.DoesNotExist):
```
chrisseto: This can probably just be `except IndexError:`

cwisecarver: It can.

api/base/serializers.py
```diff
@@ -508,21 +513,25 @@ def get_meta_information(self, meta_data, value):
                 field_counts_requested = self.process_related_counts_parameters(show_related_counts, value)
 
                 if utils.is_truthy(show_related_counts):
-                    meta[key] = website_utils.rapply(meta_data[key], _url_val, obj=value, serializer=self.parent, request=self.context['request'])
+                    meta[key] = website_utils.rapply(meta_data[key], _url_val, obj=value, serializer=self.parent,
```
chrisseto: Not digging the formatting changes.

api/base/serializers.py
cwisecarver: `relationship_field_names` is named poorly. It's the list of fields on the serializer if their source is unset then it's the field name, otherwise it's the source. Further down in the PR this list gets turned into a set and intersected with the fk fields and the m2m fields in order to get the list of relationship fields that are also on the serializer.

sloria: > There's probably a fair about of relations that aren't exposed via the API that we wouldn't want to join/prefetch.

Ah ok, that makes sense. In that case, I suggest we just rename `relationship_field_names` to `field_names` or `model_field_names` (as opposed to serializer field names).

sloria: Also, minor nitpicky refactor:

```python
self.model_field_names = [name if field.source == '*' else field.source
                          for name, field in self.fields.iteritems()]
```

cwisecarver: > Also, PrefetchRelationshipsSerializer doesn't have any prefetching behavior, so we can probably get rid of it and just put this code on JSONAPISerializer.

Happy to rename it but there are a relatively large amount of serializers that don't inherit from the JSONAPISerializer and I didn't want to deal with the reasons/repercussions of changing that.

api/base/utils.py
sloria: I don't think that would quite work, because we still need to do the `is_disabled`, `not is_active` , etc. checks.

api/base/serializers.py
sloria: Yep, I realized that after my first comment, hence the 'EDIT:'. Sorry for the noise.

cwisecarver: Sorry, didn't see that until now.

api/base/utils.py
```diff
@@ -70,15 +71,31 @@ def absolute_reverse(view_name, query_kwargs=None, args=None, kwargs=None):
     return url
 
 
-def get_object_or_error(model_cls, query_or_pk, display_name=None, **kwargs):
+def get_object_or_error(model_cls, query_or_pk, display_name=None, prefetch_fields=list()):
```
cwisecarver: 👍

api/base/utils.py
cwisecarver: Too bad python doesn't have `GOTO`.

cwisecarver: If we could find a way around the metaclass magic in FileNode this wouldn't be necessary. Just time constraints what they are I'm copping out.

cwisecarver: I'll add some comments for clarity.

api/nodes/serializers.py
```diff
@@ -336,9 +338,36 @@ def get_logs_count(self, obj):
 
     def get_node_count(self, obj):
         auth = get_user_auth(self.context['request'])
-        return len([node_relation
-                    for node_relation in obj.node_relations.filter(child__is_deleted=False, is_node_link=False)
-                    if node_relation.child.can_view(auth)])
+        user_id = getattr(auth.user, 'id', None)
+        with connection.cursor() as cursor:
```
cwisecarver: @chrisseto You said there were tests that initially did not pass and they did pass after that, correct? In running the full suite I found that occasionally `auth` will be `None` and you would get a `AttributeError` which is why lino 341 was added.

api/base/serializers.py
cwisecarver: ⁉️

api/base/serializers.py
```diff
@@ -1006,15 +1006,11 @@ def is_valid(self, clean_html=True, **kwargs):
...
-    relationship_field_names = list()
 
     def __init__(self, *args, **kwargs):
         super(PrefetchRelationshipsSerializer, self).__init__(*args, **kwargs)
-        for field_name, field in self.fields.iteritems():
-            if field.source == '*':
-                self.relationship_field_names.append(field_name)
-            else:
-                self.relationship_field_names.append(field.source)
+        self.model_field_names = [name if field.source == '*' else field.source
```
sloria: You'll need to change this in the places you pass this to prefetch_fields

cwisecarver: 🤦‍♂️

api/base/serializers.py
cwisecarver: 🤢

api/nodes/serializers.py
chrisseto:
```
api_tests/base/test_serializers.py
153:    def test_counts_included_in_link_fields_with_related_counts_query_param(self):
155:        res = self.app.get(self.url, params={'related_counts': True})
167:    def test_related_counts_excluded_query_param_false(self):
169:        res = self.app.get(self.url, params={'related_counts': False})
177:    def test_invalid_related_counts_value_raises_bad_request(self):
179:        res = self.app.get(self.url, params={'related_counts': 'fish'}, expect_errors=True)
192:    def test_counts_included_in_children_field_with_children_related_counts_query_param(self):
194:        res = self.app.get(self.url, params={'related_counts': 'children'})
208:    def test_counts_included_in_children_and_contributors_fields_with_field_csv_related_counts_query_param(self):
210:        res = self.app.get(self.url, params={'related_counts': 'children,contributors'})
224:    def test_error_when_requesting_related_counts_for_attribute_field(self):
226:        res = self.app.get(self.url, params={'related_counts': 'title'}, expect_errors=True)
228:        assert_equal(res.json['errors'][0]['detail'], "Acceptable values for the related_counts query param are 'true', 'false', or any of the relationship fields; got 'title'")

api_tests/files/views/test_file_detail.py
126:        res = self.app.get('/{}files/{}/?related_counts=True'.format(API_BASE, self.file._id), auth=self.user.auth)

api_tests/nodes/views/test_node_comments_list.py
98:        url = '{}?filter[target]={}&related_counts=False'.format(self.public_url, self.public_project._id)
105:        url = '{}?filter[target]={}&related_counts=False&version=2.1'.format(self.public_url, self.public_project._id)
1000:    def test_filtering_by_target_no_results_with_related_counts(self):
1001:        url = '{}?filter[target]=fakeid&related_counts=True'.format(self.base_url)

api_tests/nodes/views/test_node_detail.py
183:        res = self.app.get(self.public_url + '?related_counts=True', auth=self.user.auth)

api_tests/registrations/views/test_registration_detail.py
89:    def test_registration_shows_specific_related_counts(self):
90:        url = '/{}registrations/{}/?related_counts=children'.format(API_BASE, self.private_registration._id)

api_tests/registrations/views/test_registration_linked_nodes.py
57:    def test_get_linked_nodes_related_counts(self):
59:            '/{}registrations/{}/?related_counts=linked_nodes'.format(API_BASE, self.linking_node._id),

api_tests/registrations/views/test_withdrawn_registrations.py
140:    def test_field_specific_related_counts_ignored_if_hidden_field_on_withdrawn_registration(self):
141:        url = '/{}registrations/{}/?related_counts=children'.format(API_BASE, self.registration._id)
147:    def test_field_specific_related_counts_retrieved_if_visible_field_on_withdrawn_registration(self):
148:        url = '/{}registrations/{}/?related_counts=contributors'.format(API_BASE, self.registration._id)
```

PR #6944, opened 2017-03-07
[Hotfix] Fix reg withdraw hook [OSF-7545]

alexschiller:
## Purpose
Withdrawing does not manually call on node updated actions on the registration (only on the node), causing withdrawals to not hit the share hook properly.

## Changes
Add a manual call to the hook in the retraction model at the end of on complete

## Side effects
None known

## Ticket
[#OSF-7545]

## QA Notes

Make sure retracting a registration populates the withdraw tag on Registries and that it generally doesn't break anything :)


alexschiller:
Not sure why this one is failing when the nearly identical develop fix is passing (https://github.com/CenterForOpenScience/osf.io/pull/6945), will explore in the AM.

tests/test_registrations/test_retractions.py
```diff
@@ -28,11 +28,18 @@
 class RegistrationRetractionModelsTestCase(OsfTestCase):
     def setUp(self):
         super(RegistrationRetractionModelsTestCase, self).setUp()
+        self.mock_registration_update = mock.patch('website.project.tasks.on_registration_updated')
```
brianjgeiger: And you've verified that these tests function properly, yes? (I ask this of any tests that include mocks now, since we've had a bit of a history with mocking tests not actually doing anything).

alexschiller:
Old master. The other PR only applies now. Closing.

PR #6945, opened 2017-03-07
[Feature] Fix reg withdraw hook for django [OSF-7545]

alexschiller:
## Purpose
Withdrawing does not manually call on node updated actions on the registration (only on the node), causing withdrawals to not hit the share hook properly.

## Changes
Add a manual call to the hook in the retraction model at the end of on complete

## Side effects
None known

## Ticket
[#OSF-7545]

## QA Notes

Make sure retracting a registration populates the withdraw tag on Registries and that it generally doesn't break anything :)

from https://github.com/CenterForOpenScience/osf.io/pull/6944


alexschiller:
There are a lot of mocks in here, but most exist to prevent test failures due to either setting or not setting SHARE_URL etc.

PR #6949, opened 2017-03-08
[feature] parallel tests

icereval:
Optimize tests for parallel runs with local (docker) and travis-ci.

### DIFF
--- a/osf_tests/settings.py
+++ b/osf_tests/settings.py
@@ -1,6 +1,5 @@
 # Use API defaults. This allows these settings to work with API tests
-import uuid
-from api.base.settings.defaults import *  # noqa
+from api.base.settings import *  # noqa
 
 DEBUG_PROPAGATE_EXCEPTIONS = True
 #DATABASES = {
@@ -31,3 +30,13 @@ REST_FRAMEWORK['ALLOWED_VERSIONS'] = (
     '3.0',
     '3.0.1',
 )
+REST_FRAMEWORK['DEFAULT_THROTTLE_CLASSES'] = ()
+REST_FRAMEWORK['DEFAULT_THROTTLE_RATES'] = {
+    'user': '1000000/second',
+    'non-cookie-auth': '1000000/second',
+    'add-contributor': '1000000/second',
+    'create-guid': '1000000/second',
+    'root-anon-throttle': '1000000/second',
+    'test-user': '2/hour',
+    'test-anon': '1/hour',
+}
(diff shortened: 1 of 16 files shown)


osf_tests/settings.py
```diff
@@ -22,11 +22,21 @@
...
-    )
+    '2.0',
+    '2.0.1',
+    '2.1',
+    '2.2',
+    '2.3',
+    '3.0',
+    '3.0.1',
+)
+REST_FRAMEWORK['DEFAULT_THROTTLE_CLASSES'] = ()
```
sloria: Are these lines necessary, since we're already setting the high throttle rates in `local-travis.py`?

tests/base.py
```diff
@@ -238,17 +239,17 @@ def tearDownClass(cls):
     httpretty.DELETE,
 ]
 def kill(*args, **kwargs):
-    raise httpretty.errors.UnmockedError
+    print('httppretty.kill', args, kwargs)
```
sloria: Minor: Was this just for debugging? Should this be removed?

website/search/elastic_search.py
```diff
@@ -66,9 +66,11 @@ def client():
     global CLIENT
     if CLIENT is None:
         try:
+            logger.info('!!!!!!!!! NEW ES CONNECTION !!!!!!!!!!!!')
```
sloria: Remove debugging statement?

osf_tests/settings.py
icereval: :+1:

website/search/elastic_search.py
icereval: :+2:

tests/base.py
icereval: switch over to logger call, w/o we get nothing from where the error occurred during tests

tests/base.py
```diff
@@ -239,7 +240,7 @@ def tearDown(self):
     httpretty.DELETE,
 ]
 def kill(*args, **kwargs):
-    print('httppretty.kill', args, kwargs)
+    logger.error('httppretty.kill: %s - %s', args, kwargs)
```
sloria: Needs to be 

```python
logger.error('httppretty.kill: %s - %s' % (args, kwargs))
```

or 

```python
logger.error('httppretty.kill: {} - {}'.format(args, kwargs))
```

icereval: passing as args should str them, just verified in ipython, plus it lazy evals based on the set log level, where `.format` can format and then potentially never display the string.

sloria: Ah...that's right. I forgot about that behavior. Ignore me.

PR #6950, opened 2017-03-09
Add migration for index abstractnode is_public, is_deleted, and type

cwisecarver:
## Purpose

Make abstractnode queries faster.

## Changes

Added index_together on is_public, is_deleted, and type

## Side effects

Queries could be faster.


icereval:
:+1:

PR #6951, opened 2017-03-09
Add migration for abstractnode index

cwisecarver:


cwisecarver:
It's good. I promise.

PR #6953, opened 2017-03-09
[RFR] Make explore/new projects page faster [#OSF-7600]

cwisecarver:
## Purpose

Make explore/new projects page faster

## Changes

Replaced modm find and iterative filtering in python with analogous django queryset.

## Side effects

Unlikely. Left the previous code in and compared the two result sets to make sure SQL matched the python+modm.


website/project/utils.py
```diff
@@ -25,21 +25,16 @@ def serialize_node(*args, **kwargs):
...
-        n -= 1
-        yield reg
+
+    return Registration.objects.filter(
+        is_public=True,
+        is_deleted=False,
+    ).filter(
+        DQ(DQ(embargo__isnull=True) | ~DQ(embargo__state='unapproved')) &
+        DQ(DQ(retraction__isnull=True) | ~DQ(retraction__state='approved'))
+    ).get_roots().order_by('-registered_date').limit(n)
```
icereval: do we need an index on the desc order by?

cwisecarver: I'll take a look as soon as the db restore is finished.

icereval: btw, don't let this be much of a blocker, its certainly an improvement and we'll see any nuance in newrelic.

website/project/utils.py
```diff
@@ -25,21 +25,16 @@ def serialize_node(*args, **kwargs):
...
-    )
-    for reg in registrations:
-        if not n:
-            break
-        if reg.is_retracted or reg.is_pending_embargo:
-            # Filter based on calculated properties
-            continue
-        n -= 1
-        yield reg
+
```
icereval: might be able to remove `CONTENT_NODE_QUERY` and L24 `PROJECT_QUERY` variables.

website/project/utils.py
cwisecarver: The desc on registered_date didn't help much.

osf/migrations/0042_add_registration_registered_date_desc_index.py
```diff
@@ -0,0 +1,29 @@
...
+
+    dependencies = [
+        ('osf', '0041_auto_20170308_1932'),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            [
+                'CREATE INDEX osf_abstractnode_registered_date_index ON public.osf_abstractnode (registered_date DESC);',
+                "CREATE INDEX osf_abstractnode_registration_pub_del_type_index ON public.osf_abstractnode (is_public, is_deleted, type) WHERE is_public=TRUE and is_deleted=FALSE and type = 'osf.registration';"
```
cwisecarver: @icereval Then I added these partial indexes on the things we filter on all the time because the query I'd written was doing a seq scan. This seems to have made things much faster. ~350ms -> 66ms. Hopefully the nodes one will help other things.

icereval: :+1:

cwisecarver:
Failed because of probabilistic dupe username. [Passed on my travis.](https://travis-ci.org/cwisecarver/osf.io/builds/209518500)

website/project/utils.py
cwisecarver: PROJECT_QUERY gets used everywhere. I removed CONTENT_NODE_QUERY.

PR #6956, opened 2017-03-10
Fix number of subject children if not allowed by preprint provider [OSF-7615]

sheriefvt:
## Purpose
Subjects might have children that the provider does not allow. There is an API response for any single subject listing its children. There is a need to fix the API side so that the serializer show the real children count based on the current provider.

## Changes

- Update preprint_provider view. 
   - Count number of children for each subject, if only allowed by the current preprint provider.
   - Store the count into a subject attribute **'child_count_provider'**.
- Update serializer.
   - Use a new method that checks if the subject has the **'child_count_provider'** and returns its value. Otherwise, returns all the children.

## Side effects
The response time for the API has been recorded. The extra step needed for searching and counting the children added an extra time to the API response (approx. 3-4x times slower at worst case scenario). This effect is noted with large lists (e.g. Arts and Humanities subject under SocArXiv)

## Ticket

https://openscience.atlassian.net/browse/OSF-7615

### DIFF
--- a/api/preprint_providers/views.py
+++ b/api/preprint_providers/views.py
@@ -206,6 +206,19 @@ class PreprintProviderSubjectList(JSONAPIBaseView, generics.ListAPIView):
 
     serializer_class = TaxonomySerializer
 
+    def preprare_subjects(self, provider):
+        final_subjects = set()
+        for i in provider.all_subjects:
+            final_subjects.add(i._id)
+        return final_subjects
+
+    def check_child_count(self, sub, final_form):
+        num_children = 0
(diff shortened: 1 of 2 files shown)


brianjgeiger:
Hey, @mfraezz, I expect your custom taxonomy stuff is going to intersect here, yes? Will you have a better way of ensuring that the count is right on your stuff?

mfraezz:
Yes, on both accounts. This will conflict, and could be done more optimally for any provider that has confirmed their desired taxonomy mapping.

PR #6952, opened 2017-03-09
[EOSF-498] Add ember-osf link for local development

hmoco:
## Purpose
Allow ember-osf local development by using `yarn link` in the preprints container.

## Changes
Docker override now has (commented out) code to share a local ember-osf directory as a volume with the preprints container. 

## Ticket
https://openscience.atlassian.net/browse/EOSF-498

PR #6957, opened 2017-03-10
[rfr] Update dockerfiles to support running registries locally [#OSF-7616]

abought:
## Purpose
Allow running registries from within docker compose for ongoing development.

This is a developer-facing change and is not expected to affect production environments. Users should not notice any difference.

## Changes
Update docker compose and local-dist settings examples with the changes required to get registries running locally. Uncomments and slightly updates code as needed.

Also fixes an issue running mfr sync locally.

## Side effects
None expected.

## Ticket
https://openscience.atlassian.net/browse/OSF-7616


docker-compose.yml
```diff
@@ -216,21 +216,21 @@ services:
...
-  #     - 4300
-  #   ports:
-  #     - 4300:4300
-  #   volumes:
-  #     - registries_dist_vol:/code/dist
-  #   stdin_open: true
-  #   tty: true
+  registries:
+    image: quay.io/centerforopenscience/ember-osf-registries:develop-development-local
+    build: ../ember-osf-registries
```
alexschiller: This line needs to be removed as it defaulted to my local ember-osf-registries instead of the image

abought: cc @binoculars as I understand he's working on a similar PR.

binoculars: Should be fixed after my PR to @hmoco's PR goes in, hmoco/osf.io#11 → #6952

You don't need the build property here. If anything, it should go in the override file. It's only to specify a different docker container (i.e. you wanted to update something in the registries Dockerfile)

abought:
Pushed requested changes. Per discussion with @binoculars , we'll both keep our respective PRs open as there is not total overlap.

Thanks for the catch, Alex!

brianjgeiger:
There are merge conflicts due to similar changes for EmberOSF. Could you fix?

abought:
Thanks. I'm in a final round of manual testing, at which point I will push two more small changes discovered while integrating Barrett's changes.

abought:
Sorry for the delay. I've pushed the changes; Alex has confirmed that the `volumes` change was helpful for him as well.

docker-compose.yml
```diff
@@ -212,8 +212,6 @@ services:
     volumes:
       - registries_dist_vol:/code/dist
     stdin_open: true
-    tty: true
```
alexschiller: 👍

docker-compose.override.yml
```diff
@@ -1,9 +1,9 @@
 
 version: '2'
 
-volumes:
-  emberosf-sync:
-    external: true
+#volumes:
+#  emberosf-sync:
+#    external: true
```
alexschiller: @brianjgeiger FYI: Anyone who pulled today will fail on `docker-compose up` until this change goes in.

PR #6962, opened 2017-03-13
[OSF-7624] Allow preprints sitemap code to be used pre-Django

hmoco:
## Purpose
Add preprints sitemap to Production. 

## Changes
Add the celery task to calculate preprints sitemap.

## Ticket
https://openscience.atlassian.net/browse/OSF-7624


scripts/generate_sitemap.py
```diff
@@ -0,0 +1,203 @@
...
+            raise Exception('Too many errors generating sitemap.')
+
+    def generate(self):
+        print('Generating Sitemap')
+
+        # Progress bar
+        progress = Progress()
+
+        # Preprint urls
+        objs = PreprintService.find(Q('is_published', 'eq', True))
```
brianjgeiger: This does not look like it's using the MODM-iterator to handle large data sets. Can you check in migrate_search for the example and make sure you're doing that? @erinspace has also done that recently with analytics.

website/settings/defaults.py
```diff
@@ -1782,3 +1787,22 @@ def parent_dir(path):
 
 # refresh campaign every 5 minutes
 CAMPAIGN_REFRESH_THRESHOLD = 5 * 60  # 5 minutes in seconds
+
+# sitemap default settings
+
+SITEMAP_URL_MAX = 25000
+SITEMAP_INDEX_MAX = 50000
+SITEMAP_STATIC_URLS = [
```
alexschiller: Not sure it matters, but the part that iterates over these was removed.  It is in the django version.

scripts/generate_sitemap.py
```diff
@@ -0,0 +1,210 @@
...
+        # Can't do a proper percentage bar, as can't count total ahead
+        progress.start(100, 'PREP: ')
+        for obj in objs:
+            progress.increment()
+            if progress.count >= 100:
+                progress.restart()
+            if not obj.node or not obj.primary_file or obj.node.is_deleted or not obj.node.is_public:
+                continue
+            try:
+                preprint_date = obj.date_modified.isoformat() if obj.date_modified else obj.date_created.isoformat()
```
alexschiller: @brianjgeiger The date thing I mentioned is handled here by using date_modified but defaulting to date_created. For this version I think that is fine. If it is a common issue on prod data, it is probably a bigger problem than just the sitemap.

hmoco:
Closed and reopen to trigger rebuild, failure seems to not have anything to do with changes made in the last commit.

.gitignore
```diff
@@ -69,6 +69,7 @@ website/mfrcache/
 website/mfrtemp/
 website/osfstoragecache/
 website/gpg/
+website/static/sitemaps/
```
sloria: Just a drive-by comment: It might make sense to generate the sitemaps to `website/static/built/sitemaps`. The `/built/` directory is used for generated assets and is already gitignored.

alexschiller: If this is done, another PR will need to update the routes on develop, and check with @mattclark if that changes how he is handling the sitemap routes on staging (I recall he did something a little differently, but that could be false). Happy to make the develop changes, if this change is desirable.

hmoco:
This was run, and further runs will be using post-django script,

PR #6966, opened 2017-03-14
[do not merge] postgres does not order rows, order with id by default [#OSF-7607]

icereval:
Force `ORDER BY 'id'` in `BaseModel` for queries since Postgres does not order rows.

https://docs.djangoproject.com/en/1.10/topics/db/models/#meta-inheritance
https://www.postgresql.org/docs/9.1/static/queries-order.html

#### Performance Consideration

Its possible this will decrease performance of queries, though all PK's have a default index and sort so in theory they will remain optimal, we'll just need to keep an eye on newrelic.


icereval:
Researching this a bit more I believe this approach is flawed after [reading the Django Pagination docs](https://docs.djangoproject.com/en/1.10/topics/pagination/#required-arguments).

e.g. adding an optional `order_by('id')` -  https://github.com/CenterForOpenScience/osf.io/commit/d4644e94109cbb6f2858186058704e91126b7dd5

PR #6977, opened 2017-03-16
Feature/remove old api docs [OSF-7627]

fabmiz:
## Purpose
Replace old API docs with links to new API docs.

## Changes
For each API endpoint, removes old documentation from old API docs (api.osf.io/v2/) and 
adds a link pointing to corresponding documentation in the new API docs (develop.osf.io).

## Ticket
https://openscience.atlassian.net/browse/OSF-7627


api/base/views.py
```diff
@@ -368,357 +368,7 @@ def create(self, *args, **kwargs):
...
-        box          Box.com
-        dataverse    Dataverse
-        dropbox      Dropbox
-        figshare     figshare
-        github       GitHub
-        googledrive  Google Drive
-        osfstorage   OSF Storage
-        s3           Amazon S3
-
+    """The documentation for this endpoint can be found [here](https://developer.osf.io/#Base_base_read).
```
caseyrollins: Since this is the root endpoint, can you add a general docs link here as well?

```
The documentation for the Open Science Framework API can be found at https://developer.osf.io. 

The documentation for this endpoint can be found [here](https://developer.osf.io/#Base_base_read).
```

api/base/views.py
```diff
@@ -368,357 +368,11 @@ def create(self, *args, **kwargs):
...
-        box          Box.com
-        dataverse    Dataverse
-        dropbox      Dropbox
-        figshare     figshare
-        github       GitHub
-        googledrive  Google Drive
-        osfstorage   OSF Storage
-        s3           Amazon S3
 
+    The documentation for the Open Science Framework API can be found at `https://developer.osf.io`.
```
caseyrollins: Can you move this line above the other, so that the link to developer.osf.io comes before the link for this endpoint? 

Also, I think the link in this line should be `[developer.osf.io](https://developer.osf.io)`, so that it's rendered as a link and not a code snippet.

caseyrollins:
Closing in favor of #8147

PR #6987, opened 2017-03-16
Asynchronize user-related elasticsearch updates [OSF-7643]

mfraezz:
## Purpose
Make `.save` calls faster for users with many nodes

## Changes
* Update elastic with queued tasks

## Side effects
None expected

## Ticket
[[OSF-7643]](https://openscience.atlassian.net/browse/OSF-7643)


website/search/search.py
```diff
@@ -60,11 +60,25 @@ def delete_node(node, index=None):
 def update_contributors(nodes):
```
icereval: cleanup, remove?

PR #6995, opened 2017-03-20
[OSF-7662] Filter out system tags when accessing Node.tags

sloria:
## Purpose

System tags were being exposed through the v2 node detail endpoint.

## Changes

- Override the default Manager for Tag, so that `node.tags` filters out system tags
- Add the `all_tags` manager for Tag. Usage: `Tag.all_tags.filter(osfuser=user, system=False)`
- Add the `all_tags` property to OSFUser and AbstractNode. 

## Side effects

When filtering tags, you must call `all_tags` if you want to include system tags, e.g. `node.all_tags.filter(name=name, system=True)`.

## Ticket

https://openscience.atlassian.net/browse/OSF-7662

## Other notes

In an attempt to address [OSF-7606](https://openscience.atlassian.net/browse/OSF-7606), I tried a couple optimizations on this branch, but neither of them improved performance or reduced the number of queries.

1. Making `tags` a SerializerMethodField in NodeSerializer

```python
# Before
    tags = JSONAPIListField(child=NodeTagField(), required=False)

# After
    tags = ser.SerializerMethodField()
    def get_tags(self, node):
        return node.tags.values_list('name', flat=True)
```

This did not reduce the number of queries or the load time.

2. Eager-loading _contributors in the UserNodes endpoint

```python
# Before
    def get_queryset(self):
        return Node.find(self.get_query_from_request()).select_related('root')


# After
    def get_queryset(self):
        return Node.find(self.get_query_from_request()).select_related('root').eager('_contributors')
```

This actually increased the number of queries and increased load time.

### DIFF
--- a/osf_tests/test_node.py
+++ b/osf_tests/test_node.py
@@ -685,9 +685,9 @@ class TestTagging:
         node.add_system_tag('FoO')
         node.save()
 
-        tag = Tag.objects.get(name='FoO')
-        assert node.tags.count() == 1
-        assert tag in node.tags.all()
+        tag = Tag.all_tags.get(name='FoO', system=True)
+        assert node.all_tags.count() == 1
+        assert tag in node.all_tags.all()
 
         assert tag.system is True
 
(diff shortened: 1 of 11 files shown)


sloria:
Thanks. Looking into the failing tests now.

osf_tests/test_node.py
```diff
@@ -685,9 +685,9 @@ def test_add_system_tag(self, node):
         node.add_system_tag('FoO')
         node.save()
 
-        tag = Tag.objects.get(name='FoO')
-        assert node.tags.count() == 1
-        assert tag in node.tags.all()
+        tag = Tag.all_tags.get(name='FoO')
```
icereval: to be correct this needs `system=True`

sloria:
Fixed the incorrect tests. Also removed some unnecessary logging from the tests.

sloria:
Build passed on my fork's TCI.

sloria:
Build is failing due to a missing requirement on develop.

PR #6994, opened 2017-03-20
Update user.date_last_login every 8 hours [OSF-7660]

mfraezz:
## Purpose
Update `user.date_last_login` less frequently than every request to invalidate cache less frequently

## Changes
* Fatten model, update after 8 hour threshold

## Side effects
None expected. `NO_LOGIN_WAIT_TIME` and `NO_LOGIN_OSF4M_WAIT_TIME` are compared against `date_last_login`, but are `4` and `6` weeks, respectively.

## Ticket
[[OSF-7660]](https://openscience.atlassian.net/browse/OSF-7660)

PR #6998, opened 2017-03-20
Add throttle for date_last_login

sloria:
Includes #6994

## Purpose

Prevents potential locking errors and unnecessary updates to User records.

## Changes

Only update `date_last_login` after 60 seconds have elapsed after last update.

## Side effects

We used to do this for requests that used basic auth: https://github.com/CenterForOpenScience/osf.io/commit/c35b12f9de9cb209eb649234708adfb3199f997f , but we must have forgotten to add the throttle for cookie auth.

### DIFF
--- a/framework/sessions/__init__.py
+++ b/framework/sessions/__init__.py
@@ -1,9 +1,11 @@
 # -*- coding: utf-8 -*-
+import datetime as dt
 import httplib as http
 import urllib
 import urlparse
 
 from django.utils import timezone
+from django.db.models import Q
 from django.apps import apps
 import bson.objectid
 import itsdangerous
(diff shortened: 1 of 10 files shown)


framework/sessions/__init__.py
```diff
@@ -158,9 +159,14 @@ def before_request():
...
             return
         if not util_time.throttle_period_expired(user_session.date_created, settings.OSF_SESSION_TIMEOUT):
+            # Update date last login when making non-api requests
             if user_session.data.get('auth_user_id') and 'api' not in request.url:
                 OSFUser = apps.get_model('osf.OSFUser')
-                OSFUser.objects.filter(guids___id=user_session.data['auth_user_id']).invalidated_update(date_last_login=timezone.now())
+                query = OSFUser.objects.filter(guids___id=user_session.data['auth_user_id']).only('date_last_login')
+                user = query.get()
+                # Throttle updates to date_last_login (60 seconds)
+                if user.date_last_login is None or (timezone.now() - user.date_last_login) > dt.timedelta(seconds=60):
```
brianjgeiger: https://github.com/CenterForOpenScience/osf.io/blob/develop/website/util/time.py#L14

sloria:
Updated with `settings.DATE_LAST_LOGIN_THROTTLE` and `utils.throttle_period_expired`.

osf/models/user.py
```diff
@@ -1134,6 +1135,12 @@ def update_search_nodes(self):
         for node in self.contributed:
             node.update_search()
 
+    def update_date_last_login(self, save=False):
+        if not self.date_last_login or throttle_period_expired(self.date_last_login, website_settings.DATE_LAST_LOGIN_THROTTLE):
```
icereval: The throttle can be removed from the property wrapper - this is only called during explicit one time authentication events and it seems safe to perform the update action.

PR #6999, opened 2017-03-20
Squash Migrations [OSF-7651]

mfraezz:
## Purpose
Limit size of `osf/migrations` directory and optimize migration runtime

## Changes
* Create squash migration
* Manually add migrations that couldn't be auto-squashed

## Side effects
None expected

## TODO
Follow up migration removing old migration files after this has been ran everywhere, as per [the docs](https://docs.djangoproject.com/en/1.8/topics/migrations/#squashing-migrations)

## Ticket
[[OSF-7651]](https://openscience.atlassian.net/browse/OSF-7651)


mfraezz:
This will need to be redone after @cwisecarver's file changes go out, closing

PR #7004, opened 2017-03-21
[hotfix] migrate specific preprints script [OSF-7649]

alexschiller:
## Purpose
There is no easy way to add individual (or a list) of preprints to share. 

## Changes
This script allows a list to be passed to the migrate specific preprints instead of all preprints
e.g., for a dry run: `scripts.migration.migrate_specific_share_preprint_data --dry --targets '{"data":["rnizy"]}'`

## Side effects
N/A, though the script will need to be modified for django

## Ticket
https://openscience.atlassian.net/browse/OSF-7649

## QA Considerations
Can be tested by whoever runs the script, no standard QA

### DIFF
--- a/scripts/migration/migrate_specific_share_preprint_data.py
+++ b/scripts/migration/migrate_specific_share_preprint_data.py
@@ -1,0 +1,68 @@
+import argparse
+import json
+import logging
+import sys
+
+from framework.mongo import database
+from framework.transactions.context import TokuTransaction
+from scripts import utils as script_utils
+from website.app import init_app
+from website.models import PreprintService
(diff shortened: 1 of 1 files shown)


brianjgeiger:
I'm not suggesting to change anything, but nothing actually happens in the on_preprint_update() method locally, so running in a transaction won't actually do anything, right? Other than stopping the script if one of the requests fails?

alexschiller:
@brianjgeiger I believe that is correct-- copypasta from every other similar migration script (including the other share preprint migration script).

alexschiller:
Old PR that I'm fairly certain has been subsumed by admin actions I wrote. Closing.

PR #7005, opened 2017-03-21
Only embed public projects and components in the HTML view

sloria:
## Purpose

No need to embed public projects and components in the `/api/v1/profile/` endpoint; it just slows things down on the Dashboard and Account Settings pages.

## Changes

Don't embed nodes in the /api/v1/profile/ view.


sloria:
Tests passed on my fork's TCI.

PR #7008, opened 2017-03-22
Feature/fix add system tag

sloria:
## Purpose

`add_system_tag` would allow adding non-system tags if passing a Tag instance.

## Changes

- Raise an error if passing a non-system Tag instance to add_system_tag.
- Silence a few loud logs during tests.


sloria:
Tests have passed on my fork's TCI.

PR #7012, opened 2017-03-23
Optimize license property [No Ticket]

mfraezz:
## Purpose
Optimize license property

## Changes
Replace recursive python with recursive SQL

## Side effects
Haven't profiled, this could be slower in all but the edgiest of cases

## Ticket
None known


chrisseto:
@icereval it doesn't make it noticeably slower or faster. That being said, on average it was around 60ms slower. Overall this is probably a good change to have. It just happened to hit `parent_node` a little bit earlier in the request cycle. Combining this with an optimization of `parent_node` should be a very significant change.

tl;dr
Merge it. Might be couple ms slower but is a good change in the long run.

PR #7014, opened 2017-03-23
[feature] PreprintProvider MetaSchema [OSF-7672]

caseyrollins:
❗️ Requires migration to add `extra` field to the `PreprintProvider` model.
❗️ Requires running `populate_preprint_providers` to populate the `extra` field for each provider.

## Purpose
From the ticket:
> Various PreprintProvider fields (both existing ones and some in development) would make more sense contained within an `extra` JSON field.
Add a PreprintProvider metaschema to define the structure of the supplementary fields (about, description_link, blog_url, social links, advisory_board, header_text, description, example, etc). 
Add some sort of `extra` JSON field to the PreprintProvider model to contain said information.
For now, don't remove the existing fields.

- **Existing** fields that are part of the preprint provider JSON schema (_and which should eventually be removed from the model_):
  - `advisory_board` (now as an object, not an HTML string! :woo:)
  - `banner_name`
  - `description`
  - `emails` (an object containing the contact and support emails)
  - `example`
  - `header_text` (**NOTE**: this field is blank for every provider, can it be removed completely?)
  - `social` (an object containing facebook/instagram/twitter handles)
- **New** fields that are being placed in `extra`, as opposed to on the model:
  - `about_link` ([OSF-7409](https://openscience.atlassian.net/browse/OSF-7409))
  - `blog_url` ([OSF-7405](https://openscience.atlassian.net/browse/OSF-7405))
  - `description-link` ([OSF-7623](https://openscience.atlassian.net/browse/OSF-7623))

## Changes
- `PreprintProvider` model
  - Added `extra` DateTimeAwareJSONField.
  - Added `metaschema` property.
- Web
  - Added a Preprint Provider JSON schema ([preprint-provider.json](https://github.com/CenterForOpenScience/osf.io/pull/7014/files#diff-eb484f2a2ca91e559eba7bf06ad7826c)).
  - Added the `category` field to the JSON schema for all existing registration metaschemas.
  - Updated `ensure_schemas()` to set the `category` field for each metaschema. 
- API
  - Added `extra` DictField to the `PreprintProviderSerializer`
  - Added `category` CharField to the `MetaSchemaSerializer`
  - Added the ability to filter `/v2/metaschemas/` by category (categories are "registration" and "preprintprovider").
  - Added tests for filtering the metaschemas list.

## Side Effects
- None expected.

## Ticket


osf/models/preprint_provider.py
```diff
@@ -79,3 +83,7 @@ def banner_path(self):
             return '/static/img/preprint_providers/{}'.format(self.logo_name)
         else:
             return None
+
+    @property
+    def metaschema(self):
+        return MetaSchema.objects.get(name='Preprint Provider')
```
brianjgeiger: How is this linked to the extra field? Also are there tests (maybe in registrations) that verify that having the metaschema'd field will error appropriately if bad data is submitted? Can you add a test to verify that this field is being appropriately restricted? If there are thorough tests elsewhere, then we just need the one to ensure that the mechanism is active on this field.

caseyrollins: Ah, they currently aren't linked at all. 

I think I need to add a validator to the `extra` field that should call the `MetaSchema` class' [`validate_metadata`](https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/models/metaschema.py#L69).

It doesn't appear that [`validate_metadata`](https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/models/metaschema.py#L69) is tested at all, so I'll add tests for that as well.

api/metaschemas/views.py
```diff
@@ -50,11 +51,17 @@ class MetaSchemasList(JSONAPIBaseView, generics.ListAPIView):
     view_category = 'metaschemas'
     view_name = 'metaschema-list'
 
+    # implement ODMFilterMixin
+    def get_default_odm_query(self):
+        return (
+            Q('name', 'in', ACTIVE_META_SCHEMAS) &
+            Q('schema_version', 'eq', LATEST_SCHEMA_VERSION) |
+            Q('name', 'eq', 'Preprint Provider')
```
caseyrollins: The first two lines query for MetaSchemas that are in the [`ACTIVE_META_SCHEMAS`](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/project/metadata/schemas.py#L41) list **AND** that are [version 2](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/project/metadata/schemas.py#L4). Essentially, it just excludes all of the old v1 registration schemas, and any registration schemas that aren't active yet (which I think is just the [EGAP Project](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/project/metadata/egap-project-2.json)). 

Since the Preprint Provider metaschema is version 1, it would be left out, hence the additional `OR name="Preprint Provider"`. This could also be `OR category="preprintprovider"`.

brianjgeiger: Hmmm. I don't like that. Not your fault, I think it's the old thing that I don't like, but your addition brings out the weirdness. This will probably be easy to fix in Django when we have real SQL. But there's not, for example, a 'latest_version' property or anything useful like that on the metaschemas as things stand?

osf/models/preprint_provider.py
caseyrollins: Added validation and tests in 41fd81f

caseyrollins:
Closing in favor of #7156

PR #6965, opened 2017-03-13
[WIP] Refactor files to use TypedModels [OSF-7632]

cwisecarver:
## Purpose

Refactor files to be django models.

PR #7016, opened 2017-03-23
[OSF-7632] Make files typed models (fixed)

sloria:
Same as #6965 ,

- with tests fixed
- tests added
- `BaseFileNode.active` manager, for filtering out trashed file nodes
- A safety check in the /api/v2/nodes/files/<provider>/ endpoint to ensure it doesn't return `TrashedFileNodes`

https://openscience.atlassian.net/browse/OSF-7632


sloria:
Fork's Travis showed all passing tests, except for that probabilistic count test we've been seeing...

PR #7017, opened 2017-03-23
Improve object reprs [OSF-7668]

mfraezz:
## Ticket
[[OSF-7668]](https://openscience.atlassian.net/browse/OSF-7668)


sloria:
Fixed the failing test when I merged.

PR #7026, opened 2017-03-27
[hold] [Hotfix][SHARE-674][Fix] Don't use DOIs as preprint identifiers

chrisseto:
## Purpose

DOIs have been incorrectly processed as the DOI of the preprint it is attached to. `article_doi` is actually a descriptor of a related article/preprint.

## Changes

DOIs are now used to describe a related object.

## Ticket

https://openscience.atlassian.net/browse/SHARE-674


sloria:
@chrisseto I'll hold on merging this until you give me the go-ahead.

PR #7027, opened 2017-03-27
Allow unregistered contribs to be imported across components [OSF-7697]

mfraezz:
## Purpose
The `Contributor` model was added in the rewrite, and a small edge case was missed. This fixes

## Changes
* Fix bad merge from ~October

## Side effects
None expected

## Ticket
[[OSF-7697]](https://openscience.atlassian.net/browse/OSF-7697)

### DIFF
--- a/api/nodes/serializers.py
+++ b/api/nodes/serializers.py
@@ -439,7 +439,7 @@ class NodeSerializer(JSONAPISerializer):
                 if not contributor.user.is_registered:
                     node.add_unregistered_contributor(
                         fullname=contributor.user.fullname, email=contributor.user.email, auth=auth,
-                        permissions=parent.get_permissions(contributor.user)
+                        permissions=parent.get_permissions(contributor.user), existing_user=contributor.user
                     )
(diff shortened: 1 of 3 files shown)


api/nodes/serializers.py
```diff
@@ -437,10 +437,14 @@ def create(self, validated_data):
...
-                    node.add_unregistered_contributor(
-                        fullname=contributor.user.fullname, email=contributor.user.email, auth=auth,
-                        permissions=parent.get_permissions(contributor.user)
-                    )
+                    if contributor.user.has_usable_username():
+                        node.add_unregistered_contributor(
+                            fullname=contributor.user.fullname, email=contributor.user.email, auth=auth,
+                            permissions=parent.get_permissions(contributor.user)
+                        )
+                    elif node._primary_key not in contributor.user.unclaimed_records:
```
cwisecarver: Seems like this logic would be better handled in `user.add_unclaimed_record` but it might be a little late in the game to be changing model functionality. @icereval thoughts?

PR #7028, opened 2017-03-27
[Hotfix] Use `creator` to distinguish valid bookmarks [OSF-7696]

mfraezz:
## Purpose
Fix bad data

## Changes
* Use `creator` to distinguish valid bookmarks

## Side effects
~~It might be possible for this to break existing valid bookmark collections with links.~~

~~TODO: investigate if there are any~~

Update: not a problem. See ticket for details

## Deployment Notes
```
python -m scripts.migration.ensure_bookmark_existence --dry
python -m scripts.migration.ensure_bookmark_uniqueness --dry 
```

## Ticket
[[OSF-7696]](https://openscience.atlassian.net/browse/OSF-7696)

### DIFF
--- a/scripts/migration/ensure_bookmark_existence.py
+++ b/scripts/migration/ensure_bookmark_existence.py
@@ -14,7 +14,7 @@ logger = logging.getLogger(__name__)
 
 def get_targets():
     logger.info('Acquiring targets...')
-    targets = [u for u in User.find() if Node.find(Q('is_bookmark_collection', 'eq', True) & Q('is_deleted', 'eq', False) & Q('contributors', 'eq', u._id)).count() == 0]
+    targets = [u for u in User.find(Q('merged_by', 'eq', None)) if Node.find(Q('is_bookmark_collection', 'eq', True) & Q('is_deleted', 'eq', False) & Q('creator', 'eq', u._id)).count() == 0]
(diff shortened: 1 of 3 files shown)


scripts/migration/ensure_bookmark_existence.py
```diff
@@ -14,7 +14,7 @@
 
 def get_targets():
     logger.info('Acquiring targets...')
-    targets = [u for u in User.find() if Node.find(Q('is_bookmark_collection', 'eq', True) & Q('is_deleted', 'eq', False) & Q('contributors', 'eq', u._id)).count() == 0]
+    targets = [u for u in User.find(Q('merged_by', 'eq', None)) if Node.find(Q('is_bookmark_collection', 'eq', True) & Q('is_deleted', 'eq', False) & Q('creator', 'eq', u._id)).count() == 0]
```
sloria: Would it make sense for `OSFUser.merge_user` to skip transferring the merged user's bookmark collection?

mfraezz: Doesn't it already?

https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/models/user.py#L665-L667

sloria: Ah, sure does. 👍

PR #7036, opened 2017-03-29
Display lowercase tags [OSF-7713]

mfraezz:
## Ticket
[[OSF-7713]](https://openscience.atlassian.net/browse/OSF-7713)


sloria:
We decided this morning that we should be displaying the same casing that the users' input for their project. @cwisecarver modified the data migration to ensure this happens.

PR #7037, opened 2017-03-29
[Hotfix] Post-migration cleaning

mfraezz:


osf/migrations/0002_auto_20170329_1729.py
```diff
@@ -0,0 +1,99 @@
...
+        ),
+        migrations.AlterField(
+            model_name='abstractnode',
+            name='date_created',
+            field=osf.utils.fields.NonNaiveDateTimeField(auto_now_add=True),
+        ),
+        migrations.AlterField(
+            model_name='abstractnode',
+            name='date_modified',
+            field=osf.utils.fields.NonNaiveDateTimeField(auto_now=True, db_index=True, default=django.utils.timezone.now),
```
cwisecarver: This shouldn't need a default if it's auto_now is set.

osf/migrations/0002_auto_20170329_1729.py
```diff
@@ -0,0 +1,99 @@
...
+        ),
+        migrations.AlterField(
+            model_name='guid',
+            name='created',
+            field=osf.utils.fields.NonNaiveDateTimeField(auto_now_add=True, db_index=True),
+        ),
+        migrations.AlterField(
+            model_name='nodelog',
+            name='date',
+            field=osf.utils.fields.NonNaiveDateTimeField(auto_now_add=True, db_index=True, null=True),
```
cwisecarver: I don't think this null=True is necessary if auto_now_add is True unless we want to be able to unset it after it's been created.

osf/migrations/0002_auto_20170329_1729.py
mfraezz: It's because it was previously nullable. Django uses this to populate empty cells when migrating:

```
You are trying to change the nullable field 'date_modified' on abstractnode to non-nullable without a default; we can't do that (the database needs something to populate existing rows).
Please select a fix:
 1) Provide a one-off default now (will be set on all existing rows)
 2) Ignore for now, and let me handle existing rows with NULL myself (e.g. because you added a RunPython or RunSQL operation to handle NULL values in a previous data migration)
 3) Quit, and let me add a default in models.py
Select an option: 1
Please enter the default value now, as valid Python
The datetime and django.utils.timezone modules are available, so you can do e.g. timezone.now()
>>> timezone.now
```

osf/migrations/0002_auto_20170329_1729.py
mfraezz: This actually shouldn't be an `auto_now.*` field; it's possible for logs to be retroactively created by webhooks, which sets a different `date`. Fixing

osf/migrations/0002_auto_20170329_1729.py
cwisecarver: Okay. So it's not set in the model that way?

PR #7038, opened 2017-03-30
Add Institutions and Preprint Providers to the Admin [OSF-7556]

erinspace:
## Purpose
Add the ability to create and edit `preprint_providers` and `institutions` in the admin app. 

Also, add the ability to export the json of records and then import those records to a new database. 

examples:

#### Institution Detail
screen shot 2017-03-29 at 9 09 18 pm

#### Preprint Provider Change Form
screen shot 2017-03-29 at 9 11 57 pm

## Changes
- Institution and Preprint provider List, Detail, and Create views
- Forms for updating preprint providers and institutions
- Update preprint provider from Detail view
- Import and Export Preprint Provider and Institution JSON metadata
- Institution and Preprint Provider model changes to add permissions
- Helper methods for processing Preprint Provider subjects acceptable into rules
- Frontend UI for selecting individual subjects, that get converted to proper subject rules
- Tests for Institution and Preprint Provider view methods

unrelated change:
- Add dropdown menus on sidebar to deal with ever increasing length of menu options!

## Side effects
- New migration to add new permissions and add those permissions to the appropriate groups

## General Notes
This only allows you to type in a filename for the banner or logos, but does *not* allow you to upload an image -- this is because the images are currently stored in our staticfiles, and I believe there's a push to change where they're stored somewheres. So the adding of the actual images will still need to be done manually via PR

Editing of preprintprovider subjects accepted will *only* work until a provider defines a custom taxonomy. After that happens, a new admin formfield will need to be developed to deal with those.

## Notes for QA
- Permission Groups should function as usual (after the migration has been applied). Only those in the admin group should be able to edit, while read only can view details. 
- The institutions list should show all current institutions 
- The PreprintProviders List should show all preprint providers
- Only admins should see the "delete" button on the detail page, or the "Create" button in the menu
- The "delete" button should only work for admins if there are 0 nodes affiliated with that Institution, or 0 Preprints associated with that Preprint Provider
- For new providers and institutions, this will only create a new entry in the database, and will NOT work for new logos. It will also not do any of the authentication setup that needs to happen. A new logo still needs to be added separately via a github pull request, and ember work needs to be done for the provider to actually show up on the OSF based Preprints workflow.
- The new items **will** show up in the API in the list of preprint providers and institutions (`v2/preprint_providers/` and `v2/institutions/`)
- Editing of preprintprovider subjects accepted will *only* work until a provider defines a custom taxonomy. After that happens, a new admin formfield will need to be developed to deal with those. (and the formfield should go away)

## Ticket
https://openscience.atlassian.net/browse/OSF-7556


admin/preprint_providers/forms.py
```diff
@@ -0,0 +1,32 @@
...
+        widgets = {
+            'licenses_acceptable': CheckboxSelectMultiple(),
+            'subjects_acceptable': HiddenInput(),
+        }
+
+    def __init__(self, *args, **kwargs):
+        toplevel_choices = get_toplevel_subjects()
+        nodelicense_choices = get_nodelicense_choices()
+        super(PreprintProviderForm, self).__init__(*args, **kwargs)
+        self.fields['toplevel_subjects'].choices = toplevel_choices
```
sloria: Drive-by comment: you can pass a callable to `choices` to avoid doing this.

erinspace: Passing a callable (even uncalled) here still called the tests to error when importing the form, sadly...

admin/preprint_providers/views.py
```diff
@@ -0,0 +1,253 @@
...
+    permission_required = 'osf.view_preprintprovider'
+    raise_exception = True
+
+    def get_object(self, queryset=None):
+        return PreprintProvider.objects.get(id=self.kwargs.get('preprint_provider_id'))
+
+    def get_context_data(self, *args, **kwargs):
+        preprint_provider = self.get_object()
+        all_subjects = preprint_provider.all_subjects
+        subject_ids = [subject.id for subject in all_subjects]
```
sloria: Another drive-by: 

```python
preprint_provider.all_subjects.values_list('id', flat=True)
```

admin/preprint_providers/views.py
```diff
@@ -0,0 +1,253 @@
...
+            NodeLicense.objects.get(id=identifier).name.encode('utf-8') for identifier in preprint_provider_attributes['licenses_acceptable']
+        ]
+        preprint_provider_attributes['licenses_acceptable'] = '; '.join(preprint_licenses)
+
+        subject_html = '<ul class="three-cols">'
+        for parent in preprint_provider.top_level_subjects:
+            subject_html += '<li>{}</li>'.format(parent.text)
+            child_html = '<ul>'
+            for child in parent.children.all():
+                if child in all_subjects:
```
sloria: And another: always filter in the db when possible:

```python
for child in parent.children.filter(id__in=subject_ids):
```

sloria:
Just a few drive-by comments--I haven't reviewed this in full.

admin/preprint_providers/forms.py
sloria: What error?

admin/base/utils.py
```diff
@@ -13,3 +17,102 @@ def reverse_qs(view, urlconf=None, args=None, kwargs=None, current_app=None, que
...
+            else:
+                new_rules += potential_children_rules
+
+            subjects_done.append(parent)
+
+    return new_rules
+
+
+def get_nodelicense_choices():
+    return [(sub.id, sub.name) for sub in NodeLicense.objects.all()]
```
sloria: Use `.values` or `.values_list` when possible.

admin/base/utils.py
```diff
@@ -13,3 +17,102 @@ def reverse_qs(view, urlconf=None, args=None, kwargs=None, current_app=None, que
...
+
+
+def get_toplevel_subjects():
+    subjects = Subject.objects.filter(parents__isnull=True)
+    return [(sub.id, sub.text) for sub in subjects]
```
sloria: Same here: use `.values` or `.values_list`.

admin/preprint_providers/forms.py
erinspace: While it did avoid the "these tests need db access" error when importing any of the views (which later import the form), passing an uncalled callable to choices:

```
class PreprintProviderForm(ModelForm):
    toplevel_subjects = MultipleChoiceField(widget=CheckboxSelectMultiple(choices=get_toplevel_subjects))
```
 in the field threw the error:

`TypeError: 'function' object is not iterable`

sloria: The tests already should have db access. You shouldn't call the function because that would result in database access upon *importing* the file.

Have you tried passing choices to `MultipleChoiceField`? According to the [docs](https://docs.djangoproject.com/en/1.10/ref/forms/fields/#django.forms.ChoiceField.choices), this is allowed.

admin/preprint_providers/views.py
```diff
@@ -93,15 +92,13 @@ def get_context_data(self, *args, **kwargs):
...
-            for child in parent.children.all():
-                if child in all_subjects:
-                    child_html += '<li>{}</li>'.format(child.text)
-                    grandchild_html = '<ul>'
-                    for grandchild in child.children.all():
-                        if grandchild in all_subjects:
-                            grandchild_html += '<li>{}</li>'.format(grandchild.text)
-                    grandchild_html += '</ul>'
-                    child_html += grandchild_html
+            for child in parent.children.filter(id__in=subject_ids):
```
sloria: If you only need the `text` field, use `.values('text')` or `.values_list('text', flat=true)`.

admin/preprint_providers/views.py
```diff
@@ -93,15 +92,13 @@ def get_context_data(self, *args, **kwargs):
...
+                child_html += '<li>{}</li>'.format(child.text)
+                grandchild_html = '<ul>'
+                for grandchild in child.children.filter(id__in=subject_ids):
```
sloria: Same here: use .values or .values_list

admin/templates/base.html
```diff
@@ -98,31 +100,69 @@
...
+                <li><a role="button" data-toggle="collapse" href="#collapseNodes">
+                  <i class='fa fa-caret-down'></i> OSF Nodes
+              </a></li>
             <li>
-              <ul class="sidebar-menu" style="position:relative; left: 20px;">
-                <li><a href="{% url 'nodes:flagged-spam' %}"><i class="fa fa-exclamation-triangle"></i><span> Flagged Spam</span> </a></li>
-                <li><a href="{% url 'nodes:known-spam' %}"><i class="fa fa-cutlery"></i><span> Known Spam</span> </a></li>
-                <li><a href="{% url 'nodes:known-ham' %}"><i class="fa fa-star"></i><span> Known Ham</span> </a></li>
-              </ul>
+                <div class="collapse" id="collapseNodes">
```
brianjgeiger: Is this change noted in the PR description?

brianjgeiger:
Oh, and update migrations, please.

admin/templates/base.html
erinspace: it is now!

erinspace:
@brianjgeiger all updated!

Oh and admin tests all totally pass:

screen shot 2017-05-15 at 11 30 58 am

PR #7046, opened 2017-03-31
[hotfix] Fix Prereg Banner on Mobile [OSF-7644]

caseyrollins:
#### Purpose
- The button on the new prereg banner overlapped with the banner text on mobile screens, and also made the mobile dropdown menu disappear. Oops.

#### Screenshots 
- Still good on large screens:
screen shot 2017-03-31 at 10 40 46 am

- Better on mobile screens:
screen shot 2017-03-31 at 10 40 30 am

#### Ticket
- [OSF-7644](https://openscience.atlassian.net/browse/OSF-7644)

### DIFF
--- a/website/templates/home.mako
+++ b/website/templates/home.mako
@@ -33,6 +33,7 @@
 
 <%def name="stylesheets()">
   <link rel="stylesheet" href="/static/css/pages/home-page.css">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1, minimum-scale=1">
 </%def>
 
 <%def name="javascript_bottom()">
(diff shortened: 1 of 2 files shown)


website/templates/home.mako
```diff
@@ -33,6 +33,7 @@
 
 <%def name="stylesheets()">
   <link rel="stylesheet" href="/static/css/pages/home-page.css">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1, minimum-scale=1">
```
brianjgeiger: What's this do? Is it the first time we've had to use it?

caseyrollins: Prevents zooming on this page (essentially the same as an `overflow-x:hidden` style, but for mobile). There's a bit of hackery involved in getting the banner to have a wider width than the element it's contained by, making this necessary. 

We use a less restrictive version of this in base.mako: https://github.com/CenterForOpenScience/osf.io/blob/develop/website/templates/base.mako#L11.

brianjgeiger: Gotcha. I'm guessing this will cause us a little trouble later, but we can deal with that in a future improvement if so.

PR #7048, opened 2017-03-31
[feature] Update Dropbox Client [OSF-7384]

caseyrollins:
#### Purpose
- Updates the Dropbox client we're using so that requests will be made to Dropbox's APIv2.

#### Changes
- Upgrades [dropbox client](https://github.com/dropbox/dropbox-sdk-python) from v2.0.0 to v7.2.1
- Replaces instances of the deprecated `DropboxClient` class with `Dropbox` instances, per [docs](https://github.com/dropbox/dropbox-sdk-python/blob/master/dropbox/client.py#L48).
  - [`DropboxClient.account_info()`](https://github.com/dropbox/dropbox-sdk-python/blob/master/dropbox/client.py#L145) -> [`Dropbox.users_get_current_account()`](https://github.com/dropbox/dropbox-sdk-python/blob/master/dropbox/base.py#L3181)
  - [`DropboxClient.disable_access_token()`](https://github.com/dropbox/dropbox-sdk-python/blob/master/dropbox/client.py#L157) -> [`Dropbox.auth_token_revoke()`](https://github.com/dropbox/dropbox-sdk-python/blob/master/dropbox/base.py#L58)
  - [`DropboxClient.metadata(path)`](https://github.com/dropbox/dropbox-sdk-python/blob/master/dropbox/client.py#L789) -> [`Dropbox.files_list_folder(path)`](https://github.com/dropbox/dropbox-sdk-python/blob/master/dropbox/base.py#L649)

#### QA Notes
- As Tom noted in the ticket, most file actions involving Dropbox files and folders should be unaffected by this change as they are handled by WB (which has already been updated). 
- For this ticket, please test authorizing/deauthorizing a dropbox account from both the project settings and user settings pages, as well as the folder picker functionality when configuring the addon.

#### Ticket
- [OSF-7384](https://openscience.atlassian.net/browse/OSF-7384)


addons/dropbox/models.py
```diff
@@ -190,43 +192,38 @@ def get_folders(self, **kwargs):
...
-        except MaxRetryError:
-            raise max_retry_error
-
-        # Raise error if folder was deleted
-        if metadata.get('is_deleted'):
-            raise file_not_found
+            folder_id = '' if folder_id == '/' else folder_id
+            list_folder = client.files_list_folder(folder_id)
+            contents = [x for x in list_folder.entries]
+            while list_folder.has_more:
```
mfraezz: This function is only supposed  to return metadata about folders beneath an optional `folder_id` kwarg. Does the new DropBox client not have an way to get only folders, to limit the number of outgoing requests?

caseyrollins: @mfraezz , feel free to [check](https://www.dropbox.com/developers/documentation/http/documentation#files-list_folder), but I don't think there is. There's a [`get_metadata`](https://www.dropbox.com/developers/documentation/http/documentation#files-get_metadata) endpoint (which one would think would be the replacement for the old metadata endpoint...) but it only returns metadata for a specific file/folder, not its contents. 

I also can't find anywhere in the documentation how many entries will be returned before `has_more` is set to true and the /list_folder/continue endpoint needs to be used, so I can't say how many requests could potentially be made here. I'm *hoping* it's a very small number since the v1 metadata endpoint returned 10,000 files (see [`file_limit`](https://www.dropbox.com/developers-v1/core/docs#metadata)).

PR #7053, opened 2017-04-03
[OSF-7636] Add ability to inherit tags from parent node

Johnetordoff:
## Purpose

Make it so users can, upon creation of a new component, choose to import all tags from the parent or no tags from the parent. 

## Changes

Minor changes to add component modal.

## Side effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/OSF-7636


api/nodes/serializers.py
```diff
@@ -396,8 +397,14 @@ def create(self, validated_data):
             raise InvalidModelValueError(detail=e.messages[0])
         if len(tag_instances):
             node.tags.add(*tag_instances)
+
+        if is_truthy(request.GET.get('inherit_tags')) and validated_data['parent'].has_permission(user, 'write'):
+            parent = validated_data['parent']
```
caseyrollins: Can you define `parent` outside of this conditional since it's used here and in the `if` below (and remove it from there as well)?

api/nodes/serializers.py
```diff
@@ -396,8 +397,14 @@ def create(self, validated_data):
...
+            for tag in parent.tags.filter():
+                node.tags.add(tag)
```
caseyrollins: You should just be able to use `node.add_tag(...)` here. It handles the addition and creates the log for you.

api/nodes/serializers.py
caseyrollins: Similarly, `validated_data['parent'].has_user_permission(user, 'write')` is repeated.

api_tests/nodes/views/test_node_list.py
```diff
@@ -819,6 +819,29 @@ def test_create_component_inherit_contributors(self):
...
+                }
+            }
+        }
+        res = self.app.post_json_api(url, component_data, auth=self.user_one.auth)
+        assert_equal(res.status_code, 201)
+        json_data = res.json['data']
+
+        new_component_id = json_data['id']
+        new_component = Node.load(new_component_id)
+        assert_in(tag, new_component.tags.filter())
```
caseyrollins: I don't think it really matters, but `new_component.tags.all()` might make more sense than a blank filter here.

website/static/js/addProjectPlugin.js
```diff
@@ -233,6 +234,18 @@ var AddProject = {
...
                                 m('br'),
                                 m('i', ' Admins of ', m('b', options.parentTitle), ' will have read access to this component.')
+                            ),
+                            m('br'),
+                            m('label.f-w-md',
+
+                                m('input', {
+                                    type: 'checkbox',
+                                    name: 'inherit_tags',
+                                    value: true,
```
caseyrollins: Shouldn't the default value be false?

Johnetordoff: this actually didn't appear to have any effect, so I deleted it.

Johnetordoff:
@brianjgeiger right you are, this should work better.

PR #7054, opened 2017-04-03
[Feature] Unpublished preprints filter [OSF-7433]

alexschiller:
## Purpose
Unpublished and private preprints show up via the api.

## Changes
Add DjangoFilterMixin to be used by all preprints routes. Refactor all base querysets to return proper preprints.

Adds a boatload of tests.

## Side effects
None known.

## Ticket
https://openscience.atlassian.net/browse/OSF-7433

## QA Consierations
This makes changes to these API routes:
1. users/<user_id>/preprints/
2. nodes/<node_id>/preprints/
3. preprint_providers/<provider_id>/preprints/
4. /preprints/

A. Admins should see unpublished preprints (exception: the users route only shows public projects for that user if you are not the user being viewed)
B. Non-Admins should not see the unpublished preprint
C. Filters should work properly


api/base/filters.py
```diff
@@ -400,6 +401,102 @@ def convert_special_params_to_odm_query(self, field_name, query_params, key, dat
         """
         pass
 
+class DjangoFilterMixin(FilterMixin):
```
alexschiller: Most of this is copied from the odmfiltermixin, the main things that changed is the _operation_to_query

api_tests/preprint_providers/views/test_preprint_provider_preprints_list.py
```diff
@@ -1,9 +1,8 @@
 from nose.tools import *  # flake8: noqa
```
alexschiller: Note for reviewers: I know we are switching to pytest, but all these other tests were still in nose format, so I wanted to keep it consistent. I imagine switching all of them will be easier than having mixed files

api/nodes/filters.py
```diff
@@ -25,7 +25,7 @@ def _operation_to_query(self, operation):
         return super(NodesListFilterMixin, self)._operation_to_query(operation)
 
 
-class NodePreprintsFilterMixin(ODMFilterMixin):
+class NodeODMFilterMixin(ODMFilterMixin):
```
alexschiller: Because NodeForksList (etc) having NodePreprintsFilterMixin does not make sense. Imho it is a generic mixin that happens to contain preprint specific behavior. In addition, I'm not 100% sure it actually is required anymore but left it in as a precaution. Seemed silly to me.

osf/models/preprint_service.py
```diff
@@ -206,5 +206,7 @@ def set_preprint_license(self, license_detail, auth, save=False):
 
     def save(self, *args, **kwargs):
         saved_fields = super(PreprintService, self).save(*args, **kwargs)
-        if saved_fields:
+        if not saved_fields:
+            saved_fields = []
+        if 'is_published' in saved_fields or self.is_published:
```
mfraezz: `saved_fields` will always be an empty list here; Django doesn't return saved fields like MODM did. See `DirtyFieldsMixin`

alexschiller:
Botched this branch. New PR as hotfix #7094 7094

PR #7059, opened 2017-04-03
[Hotfix] Optimize BaseLinkedList [OSF-7738]

mfraezz:
## Purpose
Optimize. 

`c` is a collection with 15 links. Old way:
```
In [26]: %timeit -n 10 list(c.linked_nodes.filter(id__in=[e.id for e in c.linked_nodes.filter(is_deleted=False).exclude(type='osf.collection').order_by('-date_modified') if e.can_view(Auth(u))]))
[...]
10 loops, best of 3: 864 ms per loop
```
New way:
```
In [27]: %timeit -n 10 list(c.linked_nodes.filter(is_deleted=False).exclude(type='osf.collection').can_view(user=u, private_link=None).order_by('-date_modified'))
[...]
10 loops, best of 3: 261 ms per loop
```
## Changes
* Use `AbstractNodeQuerySet.can_view`
* Incidental fix: use `.exclude` instead of python filtering
* Incidental fix: explicitly check view permissions in `AbstractNodeQuerySet.can_view`

## Side Effects
None expected

## Ticket
[[OSF-7738]](https://openscience.atlassian.net/browse/OSF-7738)

### DIFF
--- a/osf/models/node.py
+++ b/osf/models/node.py
@@ -105,6 +105,7 @@ class AbstractNodeQuerySet(MODMCompatibilityQuerySet, IncludeQuerySet):
             if not isinstance(user, int):
                 raise TypeError('"user" must be either {} or {}. Got {!r}'.format(int, OSFUser, user))
 
+            qs |= self.filter(contributor__user_id=user, contributor__read=True)
             qs |= self.extra(where=['''
                 "osf_abstractnode".id in (
                     WITH RECURSIVE implicit_read AS (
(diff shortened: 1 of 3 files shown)


osf/models/node.py
```diff
@@ -105,6 +105,7 @@ def can_view(self, user=None, private_link=None):
             if not isinstance(user, int):
                 raise TypeError('"user" must be either {} or {}. Got {!r}'.format(int, OSFUser, user))
 
+            qs |= self.filter(contributor__user_id=user, contributor__read=True)
```
sloria: Shouldn't this be `contributor__user=user`?

mfraezz: No; several lines up it assigns `user = user.pk`

sloria: Ah ok, thanks. This is good to go then.

PR #7062, opened 2017-04-04
Update `inv shell` for Django

sloria:
## Purpose

Updates and improves the `inv shell` (which is equivalent to `python manage.py osf_shell`.

## Side effects

This depends on a fork of `django-extensions` that has a modification to allow adding additional imports at runtime (https://github.com/django-extensions/django-extensions/pull/1039). This change has been merged upstream, so we can remove the fork when django-extensions cuts another release.

Devs will need to update their requirements: `docker-compose up requirements` and/or `pip install -r requirements.txt`

## Notes

- I'm sending this as a hotfix so that the behavior of `inv shell` is consistent on master and develop. It's a low-risk PR, anyway.
- I had issues with garbage characters showing up when entering characters into IPython. Seems to be an issue with invoke<=0.13.0. Upgrading invoke locally (0.15.0) fixed the issue. I'll send another PR to develop upgrading our pinned version.

### DIFF
--- a/tasks/__init__.py
+++ b/tasks/__init__.py
@@ -109,10 +109,14 @@ def adminserver(ctx, port=8001, host='127.0.0.1', pty=True):
     ctx.run(cmd, echo=True, pty=pty)
 
 @task
-def shell(ctx, quiet=False):
-    cmd = 'DJANGO_SETTINGS_MODULE="api.base.settings" python manage.py shell_plus'
-    if not quiet:
+def shell(ctx, transaction=True, print_sql=False, notebook=False):
+    cmd = 'DJANGO_SETTINGS_MODULE="api.base.settings" python manage.py osf_shell'
+    if print_sql:
         cmd += ' --print-sql'
+    if notebook:
+        cmd += ' --notebook'
+    if not transaction:
+        cmd += ' --no-transaction'
     return ctx.run(cmd, pty=True, echo=True)
 
 @task(aliases=['mongo'])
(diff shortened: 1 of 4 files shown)


tasks/__init__.py
```diff
@@ -109,152 +108,12 @@ def adminserver(ctx, port=8001, host='127.0.0.1', pty=True):
         cmd += ' --certificate {} --key {}'.format(settings.OSF_SERVER_CERT, settings.OSF_SERVER_KEY)
     ctx.run(cmd, echo=True, pty=pty)
 
-
-SHELL_BANNER = """
```
icereval: :+1: keep banner and put in shell/shell_plus

tasks/__init__.py
```diff
@@ -109,152 +108,12 @@ def adminserver(ctx, port=8001, host='127.0.0.1', pty=True):
...
-            embed(banner1=banner, user_ns=context)
-        return
-    except ImportError:
-        pass
-    # fallback to basic python shell
-    code.interact(banner, local=context)
-    return
-
+def shell(ctx, quiet=False):
+    cmd = 'DJANGO_SETTINGS_MODULE="api.base.settings" python manage.py shell_plus'
```
icereval: :+1: add transaction opt-out to shell/shell_plus

sloria:
Added back the automatic transaction behavior, added back the banner, and better formatting of the imports. Also added a `OSF_SHELL_USER_IMPORTS` setting so that devs can add more imports.

2__tmux

PR #7064, opened 2017-04-04
[hotfix] Add preprints search and view to admin [OSF-7198]

erinspace:
## Purpose
Since preprints have their own GUIDs now, you can't currently view them in the admin module. Let's fix that!

This somewhat limited preprint view shows  preprint specific information, but then encourages the user to head to the node to perform operations such as deleting users. Thought it best to do this to avoid repetition, since most operations happen on the node anyhow.

screen shot 2016-11-15 at 5 04 18 pm

#### Change preprint provider

screen shot 2017-04-03 at 3 59 31 pm

## Changes
- Add preprint view, serializer, search, templates, and links to these in the sidebar
- Ability to change preprintprovider
- While we're at it, also make the "Add an admin user" form able to update permissions for a user as well!
- Update the `__unicode__` method for Subjects to use correct parameters

## Side effects
- Will need a migration to update permissions on `PreprintService` objects, and also add permissions to the read only and osf admin groups

## Ticket
https://openscience.atlassian.net/browse/OSF-7198


admin/preprints/serializers.py
```diff
@@ -0,0 +1,29 @@
...
+        'date_published': preprint.date_published,
+        'subjects': serialize_subjects(preprint.subjects),
+    }
+
+
+def serialize_subjects(subject_chains):
+    serialized_subjects = []
+    for subject_chain in subject_chains:
+        for subject in subject_chain:
+            subject = Subject.load(subject)
```
sloria: This is doing a lot of queries. Could you see if you can refactor it into one? Something like

```python
subject_ids = [_id for subject_chain in subject_chains for _id in subject_chain]
for subject in Subject.objects.filter(_id__in=subject_ids):
```

admin/preprints/serializers.py
```diff
@@ -18,12 +18,10 @@ def serialize_preprint(preprint):
...
-    for subject_chain in subject_chains:
-        for subject in subject_chain:
-            subject = Subject.load(subject)
-            if subject:
-                serialized_subjects.append({
-                    'id': subject._id,
-                    'text': subject.text
-                })
+    subject_ids = [_id for subject_chain in subject_chains for _id in subject_chain]
+    for subject in Subject.objects.filter(_id__in=subject_ids):
```
sloria: Minor: If you only need a few attributes, you can use .only

```python
Subjects.objects.filter(...).only('_id', 'text')
```

https://docs.djangoproject.com/en/1.10/ref/models/querysets/#only

Definitely not a blocker though.

sloria:
Looks good to me. I'll merge and deploy this this afternoon.

PR #7065, opened 2017-04-04
[HAUTFIX] [RFR] Fix IntegrityError on duplicate file upload [OSF-7741]

cwisecarver:
## Purpose

Stop IntegrityErrors from being logged in postgres logs for duplicate file uploads.

## Changes

Reversed the logic to try and find a file before creating one.

## Side effects

Small possibility the file could change in between it being found and it being updated but that existed before and could not be mitigated without a fairly large refactor.

### DIFF
--- a/addons/osfstorage/views.py
+++ b/addons/osfstorage/views.py
@@ -228,7 +228,6 @@ def osfstorage_create_child(file_node, payload, node_addon, **kwargs):
 
     # Create a save point so that we can rollback and unlock
     # the parent record
-    import ipdb; ipdb.set_trace()
     with transaction.atomic():
         try:
             created, file_node = False, parent.find_child_by_name(name, kind=int(not is_folder))
--- a/addons/osfstorage/views.py
+++ b/addons/osfstorage/views.py
@@ -11,6 +11,7 @@ from django.db.models.expressions import RawSQL
 from modularodm import Q
 
 from flask import request
+from modularodm.exceptions import NoResultsFound
 
 from framework.auth import Auth
 from framework.sessions import get_session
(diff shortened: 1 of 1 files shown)


addons/osfstorage/views.py
```diff
@@ -225,16 +226,16 @@ def osfstorage_create_child(file_node, payload, node_addon, **kwargs):
...
-    try:
-        # Create a save point so that we can rollback and unlock
-        # the parent record
-        with transaction.atomic():
+    # Create a save point so that we can rollback and unlock
+    # the parent record
+    with transaction.atomic():
+        try:
+            created, file_node = False, parent.find_child_by_name(name, kind=int(not is_folder))
+        except NoResultsFound:
```
chrisseto: Is this the correct exception?

cwisecarver:
```
ipdb> n
> /Users/cwisecarver/Dropbox/cos-dev/osf.io-migrations/addons/osfstorage/views.py(233)osfstorage_create_child()
    231     import ipdb; ipdb.set_trace()
    232     with transaction.atomic():
--> 233         try:
    234             created, file_node = False, parent.find_child_by_name(name, kind=int(not is_folder))
    235         except NoResultsFound:

ipdb> n
> /Users/cwisecarver/Dropbox/cos-dev/osf.io-migrations/addons/osfstorage/views.py(234)osfstorage_create_child()
    232     with transaction.atomic():
    233         try:
--> 234             created, file_node = False, parent.find_child_by_name(name, kind=int(not is_folder))
    235         except NoResultsFound:
    236             if is_folder:

ipdb> n
NoResultsFound: NoResultsFound()
> /Users/cwisecarver/Dropbox/cos-dev/osf.io-migrations/addons/osfstorage/views.py(234)osfstorage_create_child()
    232     with transaction.atomic():
    233         try:
--> 234             created, file_node = False, parent.find_child_by_name(name, kind=int(not is_folder))
    235         except NoResultsFound:
    236             if is_folder:

ipdb> n

cwisecarver: I think there was some change that made the django exceptions be caught and rethrow the modm version.

chrisseto: 👍

icereval:
lets hold on this, i'll just remember to ignore the noise, its more correct the way it is if many transactions are passing through this endpoint at once...

PR #7066, opened 2017-04-04
[hotfix] [OSF-7739] Prevent duplicate join clause when calling .load

sloria:
## Purpose

Prevent duplicate joins from happening when calling `.load`.

## Changes

In `GuidMixin.load`, re-annotate the queryset after filtering, so the join with `guid` can be re-used
(source: https://stackoverflow.com/questions/40663638/removing-a-superfluous-inner-join-from-a-django-query)

## Ticket

https://openscience.atlassian.net/browse/OSF-7739

### DIFF
--- a/osf/models/base.py
+++ b/osf/models/base.py
@@ -114,11 +114,7 @@ class BaseModel(models.Model):
     @classmethod
     def load(cls, data):
         try:
-            if issubclass(cls, GuidMixin):
-                return cls.objects.get(guids___id=data)
-            elif issubclass(cls, ObjectIDMixin):
-                return cls.objects.get(_id=data)
-            elif isinstance(data, basestring):
+            if isinstance(data, basestring):
                 # Some models (CitationStyle) have an _id that is not a bson
                 # Looking up things by pk will never work with a basestring
                 return cls.objects.get(_id=data)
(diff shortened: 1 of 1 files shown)


osf/models/base.py
```diff
@@ -114,11 +114,7 @@ def get_m2m_field_names(cls):
     @classmethod
     def load(cls, data):
         try:
-            if issubclass(cls, GuidMixin):
```
cwisecarver: I assume this is all being handled by these classes' load methods?

sloria: Yes

sloria:
Edited the description with the SQL queries before and after this change.

sloria:
Seems to make a slight improvement to the query time, and only on the first load of an object, since django will cache.

```python
# Before (dupe join, first load)
In [1]: %time OSFUser.load('qrgl2')
CPU times: user 19.5 ms, sys: 3.39 ms, total: 22.9 ms
Wall time: 42.8 ms
Out[1]: <OSFUser(u'sloria1@gmail.com') with guid u'qrgl2'>

# After (no dupe join, first load)
In [1]: %time OSFUser.load('qrgl2')
CPU times: user 13.6 ms, sys: 3.01 ms, total: 16.6 ms
Wall time: 32 ms
Out[1]: <OSFUser(u'sloria1@gmail.com') with guid u'qrgl2'>
```

PR #7067, opened 2017-04-04
[Hotfix] Optimize view project

mfraezz:
## Purpose
Reduce number of queries required to view a project with lots of contributors

New:
```
In [7]: %timeit -n 10 utils.serialize_visible_contributors(included_n)
10 loops, best of 3: 25.6 ms per loop
```
Old:
```
In [8]: %timeit -n 10 utils.serialize_contributors(n.visible_contributors, node=n)
10 loops, best of 3: 14.9 s per loop
```

## Changes
* `.include` related models
* Create new utils for this case

## Side effects
None expected

## Ticket
None known

### DIFF
--- a/website/profile/utils.py
+++ b/website/profile/utils.py
@@ -3,7 +3,7 @@ from framework import auth
 
 from website import settings
 from website.filters import gravatar
-from osf.models import Node
+from osf.models import Node, Contributor
 from website.util.permissions import reduce_permissions
 
 
@@ -24,6 +24,10 @@ def serialize_user(user, node=None, admin=False, full=False, is_profile=False):
     :param bool full: Include complete user properties
     """
     from website.project.utils import PROJECT_QUERY
+    contrib = None
+    if isinstance(user, Contributor):
+        contrib = user
+        user = contrib.user
     fullname = user.display_full_name(node=node)
     ret = {
         'id': str(user._primary_key),
@@ -45,7 +49,7 @@ def serialize_user(user, node=None, admin=False, full=False, is_profile=False):
             }
         else:
             flags = {
-                'visible': node.contributor_set.filter(user=user, visible=True).exists(),
+                'visible': contrib.visible if isinstance(contrib, Contributor) else node.contributor_set.filter(user=user, visible=True).exists(),
                 'permission': reduce_permissions(node.get_permissions(user)),
             }
         ret.update(flags)
@@ -107,6 +111,11 @@ def serialize_contributors(contribs, node, **kwargs):
         for contrib in contribs
     ]
 
+def serialize_visible_contributors(node):
+    # This is optimized when node has .include('contributor__user__guids')
+    return [
+        serialize_user(c, node) for c in node.contributor_set.all() if c.visible
+    ]
 
 def add_contributor_json(user, current_user=None):
     """
--- a/website/project/views/node.py
+++ b/website/project/views/node.py
@@ -654,6 +654,7 @@ def _view_project(node, auth, primary=False,
     """Build a JSON object containing everything needed to render
     project.view.mako.
     """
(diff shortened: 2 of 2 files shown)


website/profile/utils.py
```diff
@@ -45,7 +49,7 @@ def serialize_user(user, node=None, admin=False, full=False, is_profile=False):
             }
         else:
             flags = {
-                'visible': node.contributor_set.filter(user=user, visible=True).exists(),
+                'visible': getattr(contrib, 'visible', node.contributor_set.filter(user=user, visible=True).exists()),
```
icereval: should this match on `instanceof` for `Contributor`?

@cwisecarver, thoughts?

cwisecarver: Yeah, based on the change above if we get a contributor instance as the user variable then it will reassign contrib to the contributor and set user to the user. Once we're here if it's not a contributor instance then it won't have a visible and it'll do the extra query but if it is a contributor then it skips that query.

website/profile/utils.py
```diff
@@ -107,6 +111,11 @@ def serialize_contributors(contribs, node, **kwargs):
         for contrib in contribs
     ]
 
+def serialize_visible_contributors(node):
+    # This is optimized when node has .include('contributor__user__guids')
+    return [
+        serialize_user(c, node) for c in node.contributor_set.all() if c.visible
```
cwisecarver: This could be a little faster, the difference will depend on the size of the list and the number that are or are not visible. `[serialize_user(c, node) for c in node.contributor_set.filter(visible=True)]`

website/project/views/node.py
```diff
@@ -791,7 +792,7 @@ def _view_project(node, auth, primary=False,
         ]
     }
     if embed_contributors and not anonymous:
-        data['node']['contributors'] = utils.serialize_contributors(node.visible_contributors, node=node)
+        data['node']['contributors'] = utils.serialize_visible_contributors(node)
```
cwisecarver: If this wasn't for a mako view I'd say we should put this method on the model but since it's probably all going to be thrown away anyway, for either django views or ember views, I won't say anything.

website/profile/utils.py
cwisecarver: Second read it should probably have an `isinstance()` in front of it.

website/profile/utils.py
mfraezz: That won't work if `node` has contibutors/users/guids `.include`'d:

```
In [2]: included_n = Node.objects.filter(pk=n.pk).include('contributor__user__guids').get()
[...]
Execution time: 0.250131s [Database: default]

In [3]: included_n.contributor_set.count()
Out[3]: 274

In [4]: included_n.contributor_set.filter(visible=True).count()
Out[4]: 0

In [5]: len([c for c in included_n.contributor_set.all() if c.visible])
Out[5]: 271

In [6]: n.contributor_set.filter(visible=True).count()
SELECT COUNT(*) AS "__count"
FROM "osf_contributor"
WHERE ("osf_contributor"."node_id" = %s
       AND "osf_contributor"."visible" = true)

Execution time: 0.058040s [Database: default]

Out[6]: 271
```

cwisecarver: It seems like` include` is broken if that's the outcome. Without `include` filtering on visible works, right?

mfraezz: Yes, but without it there's no reduction in query count.

I can't confirm this, but think that it's likely broken due to our monkey patching of QuerySets.

cwisecarver: It shouldn't reduce the query count. Just the data coming back from the server and the number of iterations in that comp. Nothing that's not visible would be transferred or iterated through.

cwisecarver: Not a big deal for the 80%. It's probably not worth the time we talked about it. Feel free to just leave it as written.

PR #7069, opened 2017-04-05
[RFR] Speed up search indexing [OSF-7748]

cwisecarver:
## Purpose

Speed up search indexing without refactoring the whole thing.

## Changes

- Use django magic to filter in sql instead of filtering in python.
- Use values_list to only pull back data we're actually going to use.
- Make Nodes modified during the first pass of indexing be indexed again afterwards.

## Side effects

None expected.

## Ticket

https://openscience.atlassian.net/browse/OSF-7748

### DIFF
--- a/website/search/elastic_search.py
+++ b/website/search/elastic_search.py
@@ -341,7 +341,7 @@ def serialize_node(node, category):
                 'url': '/{}/'.format(x['guids___id']) if x['is_active'] else None
             }
             for x in node._contributors.filter(contributor__visible=True)
-                .values('fullname', 'guids___id', 'is_active')
+            .values('fullname', 'guids___id', 'is_active')
         ],
         'title': node.title,
         'normalized_title': normalized_title,
--- a/website/search/elastic_search.py
+++ b/website/search/elastic_search.py
@@ -417,9 +417,9 @@ def serialize_contributors(node):
         'contributors': [
             {
                 'fullname': x['fullname'],
-                'url': '/{}/'.format(x['guids___id'])
+                'url': '/{}/'.format(x['user__guids___id'])
             } for x in
-            node.contributor_set.filter(visible=True, user__is_active=True).values('user__fullname', 'user__guids___ids')
+            node.contributor_set.filter(visible=True, user__is_active=True).values('user__fullname', 'user__guids___id')
         ]
     }
 
--- a/website/search/elastic_search.py
+++ b/website/search/elastic_search.py
@@ -419,7 +419,7 @@ def serialize_contributors(node):
                 'fullname': x['fullname'],
                 'url': '/{}/'.format(x['guids___id'])
             } for x in
-            node.contributor_set.filter(visible=True, user__is_active=True).values('fullname', 'guids___ids')
+            node.contributor_set.filter(visible=True, user__is_active=True).values('user__fullname', 'user__guids___ids')
         ]
     }
 
--- a/website/search/elastic_search.py
+++ b/website/search/elastic_search.py
@@ -340,7 +340,7 @@ def serialize_node(node, category):
                 'fullname': x['fullname'],
                 'url': '/{}/'.format(x['guids___id']) if x['is_active'] else None
             }
-            for x in node._contributors.filter(contributor__visible=True).order_by('contributor___order')
+            for x in node._contributors.filter(contributor__visible=True).order_by('_order')
             .values('fullname', 'guids___id', 'is_active')
         ],
         'title': node.title,
@@ -419,7 +419,7 @@ def serialize_contributors(node):
                 'fullname': x['user__fullname'],
                 'url': '/{}/'.format(x['user__guids___id'])
             } for x in
-            node.contributor_set.filter(visible=True, user__is_active=True).order_by('contributor___order').values('user__fullname', 'user__guids___id')
+            node.contributor_set.filter(visible=True, user__is_active=True).order_by('_order').values('user__fullname', 'user__guids___id')
         ]
     }
 
--- a/website/search/elastic_search.py
+++ b/website/search/elastic_search.py
@@ -340,7 +340,7 @@ def serialize_node(node, category):
                 'fullname': x['fullname'],
                 'url': '/{}/'.format(x['guids___id']) if x['is_active'] else None
             }
-            for x in node._contributors.filter(contributor__visible=True).order_by('_order')
+            for x in node._contributors.filter(contributor__visible=True).order_by('contributor___order')
             .values('fullname', 'guids___id', 'is_active')
         ],
         'title': node.title,
--- a/website/search/elastic_search.py
+++ b/website/search/elastic_search.py
@@ -1,3 +1,4 @@
+
 # -*- coding: utf-8 -*-
 
 from __future__ import division
@@ -336,11 +337,11 @@ def serialize_node(node, category):
         'id': node._id,
         'contributors': [
             {
-                'fullname': x.fullname,
-                'url': x.profile_url if x.is_active else None
+                'fullname': x['fullname'],
+                'url': '/{}/'.format(x['guids___id']) if x['is_active'] else None
             }
-            for x in node.visible_contributors
-            if x is not None
+            for x in node._contributors.filter(contributor__visible=True)
+                .values('fullname', 'guids___id', 'is_active')
         ],
         'title': node.title,
         'normalized_title': normalized_title,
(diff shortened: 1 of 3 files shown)


sloria:
Was this branched off of master? Can you change the base of this PR to `master` and check that?

website/search/elastic_search.py
```diff
@@ -336,11 +337,11 @@ def serialize_node(node, category):
...
         'contributors': [
             {
-                'fullname': x.fullname,
-                'url': x.profile_url if x.is_active else None
+                'fullname': x['fullname'],
+                'url': '/{}/'.format(x['guids___id']) if x['is_active'] else None
             }
-            for x in node.visible_contributors
-            if x is not None
+            for x in node._contributors.filter(contributor__visible=True)
```
sloria: Will these return contributors in the right order? You might need `.order_by('contributor___order')`.

website/search_migration/migrate.py
```diff
@@ -73,6 +77,9 @@ def migrate(delete, index=None, app=None):
 
     set_up_alias(index, new_index)
 
+    # migrate nodes modified since start
+    migrate_nodes(new_index, query=Q('date_modified', 'gte', start_time))
```
sloria: I like this idea, but shouldn't you delete the other call to `migrate_nodes`?

website/search_migration/migrate.py
```diff
@@ -5,29 +5,32 @@
...
-    query = Q('is_public', 'eq', True) & Q('is_deleted', 'eq', False)
-    total = Node.find(query).count()
+    node_query = Q('is_public', 'eq', True) & Q('is_deleted', 'eq', False)
+    if query:
+        node_query = query & node_query
+    total = Node.find(node_query).count()
     increment = 200
     total_pages = (total // increment) + 1
-    pages = paginated(Node, query=query, increment=increment, each=False)
+    pages = paginated(Node, query=node_query, increment=increment, each=False)
```
sloria: I wonder if we can do a `.include('contributor__user__guids')`. I'll play around to see if it will work.

website/search/elastic_search.py
cwisecarver: Shouldn't it be ordered by that by default? I'll add it just to be sure.

website/search_migration/migrate.py
cwisecarver: That would probably help. I'll add it and give it a try.

website/search_migration/migrate.py
cwisecarver: This is to catch any nodes that were modified while `migrate_nodes` was running.

1. Start the clock
2. Do all the institutions
3. Do all the nodes
4. Do all the users
5. Do any nodes that have been modified since the clock started.

If I removed the other one anything modified outside of the running time wouldn't be migrated to ES.

sloria: Ah ok.

cwisecarver: Could the comment be clearer?

website/search/elastic_search.py
```diff
@@ -418,11 +416,10 @@ def serialize_contributors(node):
...
             {
-                'fullname': user.fullname,
-                'url': user.profile_url if user.is_active else None
-            } for user in node.visible_contributors
-            if user is not None
-            and user.is_active
+                'fullname': x['user__fullname'],
+                'url': '/{}/'.format(x['user__guids___id'])
+            } for x in
+            node.contributor_set.filter(visible=True, user__is_active=True).order_by('contributor___order').values('user__fullname', 'user__guids___id')
```
sloria: Is this working? I just discovered that filtering on included relations doesn't quite work (https://github.com/chrisseto/django-include/issues/2).

```
In [1]: qs_include = Node.objects.filter(is_public=True).include('contributor__user')

In [2]: node = qs_include[0]

In [3]: node.contributor_set.all()
Out[3]: [<Contributor(user=saskia.vanderoord@ppw.kuleuven.be, read=True, write=True, admin=True, visible=True)>, <Contributor(user=tom.beckers@kuleuven.be, read=True, write=True, admin=True, visible=True)>]

In [4]: node.contributor_set.filter(visible=True)
Out[4]: []
```

If you wanted to keep the include, you could filter out in Python. It might be worth it.

cwisecarver: I would imagine that doing the extra queries would be faster than filtering in python in almost all circumstances. Regardless, we should fix include, or stop using it, if this doesn't work. Chaining is a very highly used feature of the ORM.

sloria: In this case, I think `.include` will be faster. The savings in number of queries would outweigh the cost of skipping non-visible and inactive users in Python.

I'll do some comparisons to verify this, though.

I do agree that we need to fix that bug in `include` though. Could lead to some very hard-to-find bugs.

sloria: `.include()` is an order-of-magnitude speedup:

Without include:

```
In [1]: from django.core.paginator import Paginator

In [2]: paginator = Paginator(Node.objects.all(), 20)

In [3]: page = paginator.page(1)

In [4]: node = page[0]

In [5]: %time list(node.contributor_set.all())
CPU times: user 3.32 ms, sys: 1.26 ms, total: 4.57 ms
Wall time: 6.69 ms
Out[5]:
[<Contributor(user=jeswes@gmail.com, read=True, write=True, admin=True, visible=True)>,
 <Contributor(user=livsolis@utexas.edu, read=True, write=True, admin=True, visible=True)>,
 <Contributor(user=n.hutcheson@austin.utexas.edu, read=True, write=True, admin=True, visible=True)>]
```

With include:

```python
In [1]: from django.core.paginator import Paginator

In [2]: paginator = Paginator(Node.objects.all().include('contributor__user'), 20)

In [3]: page = paginator.page(1)

In [4]: node = page[0]

In [5]: %time list(node.contributor_set.all())
CPU times: user 143 µs, sys: 13 µs, total: 156 µs
Wall time: 152 µs
Out[5]:
[<Contributor(user=jeswes@gmail.com, read=True, write=True, admin=True, visible=True)>,
 <Contributor(user=livsolis@utexas.edu, read=True, write=True, admin=True, visible=True)>,
 <Contributor(user=n.hutcheson@austin.utexas.edu, read=True, write=True, admin=True, visible=True)>]
```

sloria: That said, using .include with a queryset that includes every public node will result in one huuuge query. Maybe we should just do the .include on individual pages.

sloria: Running .include on a `Page.object_list` doesn't seem to work too well:

```python
In [5]: page.object_list.include('contributor__user__guids')
[<Node: (title=u'Reproducibility Project: Psychology', category=u'project') with guid u'ezcuj'>]

In [6]: node = page[0]
<ipython-input-6-044c114f3a4e> in <module>()
----> 1 node = page[0]

/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/django/core/paginator.pyc in __getitem__(self, index)
    123         # it won't be a database hit per __getitem__.
    124         if not isinstance(self.object_list, list):
--> 125             self.object_list = list(self.object_list)
    126         return self.object_list[index]
    127

/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/django/db/models/query.pyc in __iter__(self)
    256                - Responsible for turning the rows into model objects.
    257         """
--> 258         self._fetch_all()
    259         return iter(self._result_cache)
    260

/Users/sloria/miniconda/envs/osf/lib/python2.7/site-packages/django/db/models/query.pyc in _fetch_all(self)
   1072     def _fetch_all(self):
   1073         if self._result_cache is None:
-> 1074             self._result_cache = list(self.iterator())
   1075         if self._prefetch_related_lookups and not self._prefetch_done:
   1076             self._prefetch_related_objects()

/Users/sloria/projects/cos-projects/osf/osf/utils/manager.pyc in __iter__(self)
     96     def __iter__(self):
     97         for instance in super(IncludeModelIterable, self).__iter__():
---> 98             self.parse_includes(instance, self.queryset._includes)
     99
    100             yield instance

/Users/sloria/projects/cos-projects/osf/osf/utils/manager.pyc in parse_includes(cls, instance, fields)
     90     def parse_includes(cls, instance, fields):
     91         for field, nested in fields.items():
---> 92             data = getattr(instance, '__' + field.name)
     93             delattr(instance, '__' + field.name)
     94             cls.parse_nested(instance, field, nested, data)

AttributeError: 'Node' object has no attribute '__contributor'
```

website/search_migration/migrate.py
sloria: The comment is fine, I was just misunderstanding the code; my bad.

sloria:
Based on [this thread](https://github.com/CenterForOpenScience/osf.io/pull/7069#discussion_r110002365), I think we should hold off on adding the `.include` for now--sorry for leading you astray.

Once that's done, I think this is good to go.

sloria:
On second thought, the .include query might not be too huge. So it's your call if you want to include (NPI) it in this PR.

cwisecarver:
@sloria New branch off master, cherry-picked in the commits.

PR #7071, opened 2017-04-05
Update Admin  OSFUser permissions to correct codename [OSF-7752]

erinspace:
## Purpose

I accidentally earlier named the new `OSFUser`'s view permission as `view_user` instead of the appropriate `view_osfuser` -- and added a few nonexistent permissions to groups like `change/dete_user` instead of `change/delete_osfuser`

So here's a migration to fix that! Along with changed templates and views when checking for those permissions.

## Changes
- Migration to update permissions on the `OSFUser` model, and `osf_admin` group
- Updates in views when checking for permissions
- Updates in templates when checking for permissions

## Side effects

- A small migration for permissions will need to be run. This also affects the user model, but I have no idea if this is super intrusive (my guess is no?)

## Ticket
https://openscience.atlassian.net/browse/OSF-7752


erinspace:
closing in favor of hotfix #7092

PR #7073, opened 2017-04-05
[#OSF-7684]Moving files between components breaks links in logs

chennan47:
## Purpose
Fix the breaking links in logs after files got moved.

## QA notes:
Steps:
1.Create a project and a component
2.Upload a file to OSF Storage in the component
3.Move the file to OSF Storage in the top level project
4.View the logs for "[contributor] added [filename] to OSF Storage..."
5.Click the link the logs for that file

Before Change:
You will be taken to a file not found page
After Change:
You will be taken to the moved file

## Ticket

https://openscience.atlassian.net/browse/OSF-7684


website/static/js/logTextParser.js
```diff
@@ -446,7 +446,10 @@ var LogPieces = {
                     'github_file_added', 'github_file_updated', 'box_file_added', 'box_file_updated', 'dropbox_file_added', 'dropbox_file_updated',
                     's3_file_added', 's3_file_updated', 'figshare_file_added', 'checked_in', 'checked_out'];
                     if (acceptableLinkedItems.indexOf(action) !== -1 && logObject.attributes.params.urls) {
-                         return logObject.attributes.params.urls.view;
+                        if (logObject.attributes.params.params_file){
```
brianjgeiger: We get a lot of errors in Sentry where we try to deep access objects and they aren't complete before they're there. https://sentry.cos.io/sentry/osf-front-end/issues/276034/ is an example where we are trying `if (!ctrl.node.anonymous && item.embeds.user && item.embeds.user.data)` but item.embeds doesn't exist and so it gives a TypeError. You should use lodash.get for this, I think. https://github.com/CenterForOpenScience/osf.io/search?utf8=✓&q=lodash.get&type

api_tests/logs/views/test_log_detail.py
```diff
@@ -141,3 +141,12 @@ def test_title_hidden_from_non_contributor_in_file_move(self):
         assert_equal(res.status_code, 200)
         assert_not_in(self.component.title, res.json['data'])
         assert_equal(res.json['data'][0]['attributes']['params']['source']['node_title'], 'Private Component')
+
+    def test_file_params(self):
+        res = self.app.get(self.node_logs_url, auth=self.user_one.auth)
+        assert_equal(res.status_code, 200)
+        assert_not_equal(
```
brianjgeiger: This seems like only half a test. Shouldn't it equal something predictable as well?

website/static/js/logTextParser.js
```diff
@@ -446,7 +446,10 @@ var LogPieces = {
                     'github_file_added', 'github_file_updated', 'box_file_added', 'box_file_updated', 'dropbox_file_added', 'dropbox_file_updated',
                     's3_file_added', 's3_file_updated', 'figshare_file_added', 'checked_in', 'checked_out'];
                     if (acceptableLinkedItems.indexOf(action) !== -1 && logObject.attributes.params.urls) {
-                         return logObject.attributes.params.urls.view;
+                        if (lodashGet(logObject,'attributes.params.params_file', null)){
```
brianjgeiger: Not a big deal, but this could probably be reduced to:

```js
viewUrl = lodashGet(logObject, 'attributes.params.urls.view', null);
return lodashGet(logObject, 'attributes.params.params_file', viewURL)
```

brianjgeiger:
@chennan47 Merge conflict. Could you get that cleared up? I'm going to try to get this merged today.

chennan47:
@brianjgeiger conflicts solved in 0f9624c

PR #7074, opened 2017-04-05
Update local figshare instructions to reflect model changes [#OSF-7755]

abought:
## Purpose
There are special instructions required to use the Figshare addon locally. These instructions should be brought up to date with recent model changes, and we should confirm that the instructions work as written.

## Changes
This is a developer facing change to docs, and should not affect user or site functionality.

## Side effects
None expected. This is my first PR with the new models, so worth verifying for accuracy. Worked when I ran locally in `shell_plus`.

## Ticket
https://openscience.atlassian.net/browse/OSF-7755

### DIFF
--- a/addons/figshare/README.md
+++ b/addons/figshare/README.md
@@ -33,7 +33,7 @@ has instructions for use with ngrok, but there are other methods.
 from addons.figshare.client import FigshareClient
 from website.oauth.models import ExternalAccount
 
-me = User.find_one(Q('username', 'eq', '<your username>'))
+me = OSFUser.load('<osf_guid>')
 token = '<personal token id from figshare>'
 client = FigshareClient(token)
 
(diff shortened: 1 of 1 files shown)


addons/figshare/README.md
```diff
@@ -33,7 +33,7 @@ has instructions for use with ngrok, but there are other methods.
 from addons.figshare.client import FigshareClient
 from website.oauth.models import ExternalAccount
 
-me = User.find_one(Q('username', 'eq', '<your username>'))
+me = OSFUser.load('<osf_guid>')
```
Johnetordoff: OSFuser needs to be imported 'from osf.models.user import OSFUser', otherwise  works fine

abought: Thanks, @Johnetordoff . Updated per your suggestion. Let me know if additional changes are needed.

PR #7076, opened 2017-04-05
[Hotfix] Let related filtering work correctly

chrisseto:
Addresses the issues found/discussed [here](https://github.com/chrisseto/django-include/issues/2)


sloria:
Fixed in https://github.com/chrisseto/django-include/issues/2

PR #7078, opened 2017-04-06
[feature] Fix Failing Registrations with &, <, > in File Names [OSF-7613]

caseyrollins:
#### Purpose
- Fixes `ArchivedFileNotFound` error being incorrectly raised for files with <, >, and & characters in their name that have been attached to a registration schema.

#### Changes
- Unescape selected file names for comparison to file map in archival process.
- Adds regression test for file names with special characters.
- Adds failing test case for `different_file_names_same_shas` (mostly unrelated, just seemed like a good test to add).
- Allows additional safe characters to be passed to `unescape_entities()`.
- Adds test for `unescape_entities()` with additional characters.

#### Side effects
- None expected.

#### Ticket
- [OSF-7613](https://openscience.atlassian.net/browse/OSF-7613)

### DIFF
--- a/website/archiver/utils.py
+++ b/website/archiver/utils.py
@@ -14,6 +14,7 @@ from website import (
     mails,
     settings
 )
+from website.util import sanitize
 
 def send_archiver_size_exceeded_mails(src, user, stat_result):
     mails.send_mail(
@@ -205,9 +206,14 @@ def get_file_map(node, file_map):
 
 def find_registration_file(value, node):
     from osf.models import AbstractNode as Node
-
     orig_sha256 = value['sha256']
-    orig_name = value['selectedFileName']
+    orig_name = sanitize.unescape_entities(
+        value['selectedFileName'],
+        {
+            '&lt;': '<',
+            '&gt;': '>'
+        }
+    )
     orig_node = value['nodeId']
     file_map = get_file_map(node)
     for sha256, value, node_id in file_map:
--- a/website/archiver/utils.py
+++ b/website/archiver/utils.py
@@ -209,7 +209,7 @@ def find_registration_file(value, node):
     orig_sha256 = value['sha256']
     orig_name = sanitize.unescape_entities(
         value['selectedFileName'],
-        {
+        safe={
             '&lt;': '<',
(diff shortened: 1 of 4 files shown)


website/util/sanitize.py
```diff
@@ -78,7 +78,7 @@ def _ensure_clean(value):
 
 
 # TODO: Remove unescape_entities when mako html safe comes in
-def unescape_entities(value):
+def unescape_entities(value, safe=None):
```
abought: Can you confirm that no result of filenames unescaped using this parameter is *ever* shown to the user / passed to the front end?

See comments on this ticket about bleach for context on why we've so carefully avoided adding this functionality. If the output of this version of the function is ever sent back to the front end, it could be problematic.
 https://openscience.atlassian.net/browse/EOSF-202?jql=project%20%3D%20EOSF%20AND%20text%20~%20ampersand 

Because it's such a common and regrettable sticking point, I highly recommend adding a description of this parameter (and a warning not to use it freely) to the docstring for this function.

website/archiver/utils.py
```diff
@@ -205,9 +206,14 @@ def get_file_map(node, file_map):
 
 def find_registration_file(value, node):
     from osf.models import AbstractNode as Node
-
     orig_sha256 = value['sha256']
-    orig_name = value['selectedFileName']
+    orig_name = sanitize.unescape_entities(
+        value['selectedFileName'],
+        {
```
abought: A small suggestion: for all uses of the function here and in tests, please include the keyword argument where appropriate. (eg `safe={}`)

Being explicit with kwargs always helps avoid future bugs (like if you added or changed params later), but this particular kwarg has a deeper meaning, and it's especially worth being explicit about the intended behavior in this case.

website/util/sanitize.py
caseyrollins: @abought, I can confirm that the unescaped file name here will never reach the user / front end. Only the escaped versions of file names are displayed in the registration form and in the admin module. I will definitely add a description of this parameter with a warning to use with caution though, thanks for the suggestion!

PR #7083, opened 2017-04-06
[RFR] Speed up profile page [OSF-7750]

cwisecarver:
## Purpose

Take cdi38's profile page from 58s locally to 32s.
Remove custom guid annotation code and replace with `include()`

## Changes

Refactored to remove unnecessary queries.

## Side effects

Locally, I was seeing missing contributors and project paths. Since the data structure didn't change and the queries should have just become eager instead of lazy if the data wasn't there they should have gone and fetched it lazily. If it occurs on test we should pull this and debug. @sloria any ideas on a cause?
screen shot 2017-04-06 at 4 26 24 pm
screen shot 2017-04-06 at 4 26 38 pm

## Ticket

https://openscience.atlassian.net/browse/OSF-7750?filter=17500


sloria:
Looks good to merge when tests pass.

PR #7087, opened 2017-04-07
[feature] Profile Page Use APIv2 + Pagination [OSF-7259]

caseyrollins:
#### Purpose
- Display only the first 5 public projects and components on a user's profile page; paginate the rest.
- Removes the recent activity dropdown from the projects displayed on this page (this feature is going away soon anyways, see: [OSF-7442](https://openscience.atlassian.net/browse/OSF-7442)).
- Removes the display of the contributions count on each node.
  - It's unclear what is actually counted here. See [my profile](https://osf.io/q7fts) as an example. I have never contributed to Courtney's "Demo Project" and the project has 47 logs, yet "14 contributions" is displayed on my profile. 

#### Changes
- Adds the `NodesFilterMixin` class that allows direct filtering on a queryset as opposed to returning `Q` objects for filtering. 
  - The `NodesFilterMixin` class replaces the `NodesODMFilterMixin` and `NodesListFilterMixin` classes. 
  - All views that inherit from those two mixins **_should_** inherit from `NodesFilterMixin` instead. However, to avoid too many changes in this hotfix, I've only replaced the two mixins with `NodesFilterMixin` on the `UserNodes` view.
- Adds the `publicNodes` mithril component for displaying a users's public projects and components. 
- Allows nodes returned from the API to be filtered with `?filter[parent][ne]=null`.

#### Side Effects
- ~More spinners.~
- There is **a lot** of duped code between this component and the existing [logFeed](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/components/logFeed.js) component that should be DRY'ed up in the future.

#### QA Notes
- Complete testing of the user profile page.
- Runscope tests of the `users/<uid>/nodes` endpoint (testing filters especially) 
- Quick check of the "All my projects/registrations/preprints" functionality on the My Projects page as this PR made changes to the endpoints used for those views.

#### Screenshots
- No public projects/components, user viewing their own profile page:
screen shot 2017-04-07 at 12 38 22 pm

- No public projects/components, user viewing a profile page that is not theirs:
screen shot 2017-04-07 at 12 38 44 pm

- Public projects/components, user viewing their own profile page
screen shot 2017-04-07 at 2 38 59 pm
screen shot 2017-04-07 at 2 40 06 pm

- Public projects/components, user viewing a page that is not theirs (note this user does not have access to "Rain Drop, Dropbox" so it is displayed as "-- private project --")
screen shot 2017-04-07 at 2 43 56 pm
screen shot 2017-04-07 at 2 44 31 pm

#### Ticket
- [OSF-7259](https://openscience.atlassian.net/browse/OSF-7259)


sloria:
osf___steven_loria


Apparently the API is returning dupes. This is a blocking issue.

sloria:
There should be some feedback while waiting for a page to load, as these requests can be slow.

pagination

sloria:
Related to the above comment: button-mashing shouldn't result in multiple requests.

requests

Note: the user should be able to request a different page from the one that is currently loading. But they should not be able to send multiple requests for the same page.

api/users/views.py
```diff
@@ -495,19 +511,22 @@ class UserNodes(JSONAPIBaseView, generics.ListAPIView, UserMixin, NodeODMFilterM
...
         if user != self.request.user:
             query &= default_node_permission_query(self.request.user)
-        return query
+        return Node.find(query)
 
     # overrides ListAPIView
     def get_queryset(self):
-        return Node.find(self.get_query_from_request()).select_related('node_license').include('guids', 'contributor__user__guids', 'root__guids', limit_includes=10)
+        return (
+            Node.objects.filter(id__in=set(self.get_queryset_from_request().values_list('id', flat=True)))
```
sloria: Why the extra query here? Can you just chain calls to `self.get_queryset_from_request()`?

```python
self.get_queryset_from_request()
.select_related(...)
...
```

sloria:
Discussed with @caneruguz : the logo spinner is no longer recommended. Caner will update the osf-style package with the recommended spinners (https://openscience.atlassian.net/browse/OSF-7837).

This is not a hard blocker to getting this merged, but I anticipate Caner will have that done before this goes to prod anyway.

api/nodes/filters.py
```diff
@@ -57,14 +60,59 @@ def convert_special_params_to_odm_query(self, field_name, query_params, key, dat
...
+                if operation['value']:
+                    # filter[parent]=<nid>
+                    parent = utils.get_object_or_error(Node, operation['value'], display_name='parent')
+                    return queryset.filter(guids___id__in=[node._id for node in parent.get_nodes(is_node_link=False)])
+                # else filter[parent]=null
+                return queryset.get_roots()
+            elif operation['op'] == 'ne':
+                if not operation['value']:
+                    # filter[parent][ne]=null
+                    child_ids = [each.child.id for each in (
```
sloria: You should be able to do `.values_list('child_id', flat=True)` here (see website/profile/views.py)

api/nodes/filters.py
```diff
@@ -57,14 +60,59 @@ def convert_special_params_to_odm_query(self, field_name, query_params, key, dat
...
+                        NodeRelation.objects.filter(is_node_link=False, child___contributors=self.get_user())
+                        .exclude(parent__type='osf.collection')
+                        .exclude(child__is_deleted=True)
+                    )]
+                    return queryset.filter(id__in=set(child_ids))
+                # else filter[parent][ne]=<nid>
+                raise InvalidFilterValue()
```
sloria: This should be supported in the future. Better add a `TODO`. 

In the meantime, pass a useful message, e.g.

```python
raise InvalidFilterValue(detail='Only "null" is accepted as valid input to "filter[parent][ne]"')
```

api/nodes/filters.py
```diff
@@ -57,14 +60,59 @@ def convert_special_params_to_odm_query(self, field_name, query_params, key, dat
...
+            else:
+                # filter[parent][gte]=''
+                raise InvalidFilterOperator()
```
sloria: Pass `value` and `value_operators` here so the error message will be more useful.

api/nodes/filters.py
```diff
@@ -57,14 +60,59 @@ def convert_special_params_to_odm_query(self, field_name, query_params, key, dat
...
+
+        if field_name == 'root':
+            if None in operation['value']:
+                raise InvalidFilterValue()
```
sloria: Pass `value='null'`.

api/nodes/filters.py
```diff
@@ -57,14 +60,59 @@ def convert_special_params_to_odm_query(self, field_name, query_params, key, dat
...
+            return queryset.filter(root__guids___id__in=operation['value'])
+
+        if field_name == 'preprint':
```
sloria: Why was this added?

api/users/views.py
```diff
@@ -577,7 +596,7 @@ def get_queryset(self):
         return user.affiliated_institutions.all()
 
 
-class UserRegistrations(UserNodes):
+class UserRegistrations(JSONAPIBaseView, generics.ListAPIView, UserMixin, NodeODMFilterMixin, NodesListFilterMixin):
```
sloria: Why is this change necessary?

api_tests/nodes/filters/test_filters.py
```diff
@@ -32,7 +32,7 @@ def setUp(self):
 
     def test_parent_filter_null(self):
         expected = [self.node_A._id]
-        res = self.app.get('{}null'.format(self.parent_url), auth=self.user.auth)
+        res = self.app.get('{}null'.format(self.parent_url), auth=self.user.auth, expect_errors=True)
```
sloria: This test is confusing; expect_errors is `True`, but we assert that the returned data are valid?

api_tests/nodes/views/test_node_list.py
```diff
@@ -535,6 +536,7 @@ def test_filtering_on_parent(self):
         assert_not_in(parent._id, guids)
         assert_not_in(parent2._id, guids)
 
+    @unittest.skip('Fix filter[parent]=null for NodesList')
```
sloria: Isn't this supported?

api_tests/registrations/filters/test_filters.py
```diff
@@ -36,7 +36,7 @@ def setUp(self):
 
     def test_parent_filter_null(self):
         expected = [self.node_A._id, self.node_B2._id]
-        res = self.app.get('{}null'.format(self.parent_url), auth=self.user.auth)
+        res = self.app.get('{}null'.format(self.parent_url), auth=self.user.auth, expect_errors=True)
```
sloria: Same here; this is confusing--do we actually expect errors?

website/static/js/components/publicNodes.js
```diff
@@ -0,0 +1,332 @@
...
+            m('a', {'href': person.embeds.users.data.links.html}, name)
+        ]);
+    });
+}
+
+var PublicProjects = {
+
+    controller: function(options) {
+        var self = this;
+        self.user = options.user._id;
```
sloria: This doesn't appear to be used.

website/static/js/components/publicNodes.js
```diff
@@ -0,0 +1,332 @@
...
+        self.is_profile = options.user.is_profile;
```
sloria: As per COSDEv, use `lowerCamelCase` for attribute names

website/static/js/components/publicNodes.js
```diff
@@ -0,0 +1,332 @@
...
+
+            // Display each project
+            [
+                ctrl.publicProjects().length !== 0 ? ctrl.publicProjects().map(function(item) {
+
+                    var parent = ctrl.nodeType === 'components' && item.embeds.parent.data ? item.embeds.parent.data.attributes : null;
+                    var category = item.attributes.category;
+                    var icon =  iconmap.projectComponentIcons[category];
+
+                    return m('div', [
```
sloria: This should probably be split into a separate component.

website/static/js/components/publicNodes.js
```diff
@@ -0,0 +1,332 @@
...
+                        m('p', {}, [
+                            'Find out how to make your ' + ctrl.nodeType + ' ',
+                            m('a', {'href': 'http://help.osf.io/m/gettingstarted/l/524048-control-your-privacy-settings', 'target': '_blank'}, 'public'),
+                            '.'
+                        ])
+                    ])
+                : m('div.help-block', {}, 'This user has no public ' + ctrl.nodeType + '.'),
+
+                // Pagination
+                m('.db-activity-nav.text-center', {style: 'margin-top: 5px; margin-bottom: -10px;'}, [
```
sloria: Separating this into a component will make it easier to reuse (e.g. for the log feed) in the future.

brianjgeiger: I told Casey not to be super-concerned about reuse in this case. Getting it so that it can't take down the OSF is more important than reusability, and honestly it really needs to be done in Ember so we can get towards a single framework. My recommendation was to write it as a one-off for now, and then if we *have* to use it somewhere else, we'd make it better then.

sloria: That is fine for the time being. I still think that [factoring out a node item](https://github.com/CenterForOpenScience/osf.io/pull/7087#discussion_r112044897) into a separate component (since it's new code) is still worth it.

I anticipate that this code will still be used by the time we get to [OSF-7442](https://openscience.atlassian.net/browse/OSF-7442), so we can't ignore maintainability too much. =)

api/users/views.py
caseyrollins: IIRC this was your solution to the API returning dupes (because we're using `.order_by()` without `.distinct()` since that doesn't work with `.include()`). 

[`get_public_projects()`](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/profile/views.py#L45) does the same sort of thing.

api_tests/nodes/filters/test_filters.py
caseyrollins: I think this is just bad clean-up on my part. Oops.

api_tests/nodes/views/test_node_list.py
caseyrollins: I think in some iteration of working on this I removed support for it... but yes, this test currently passes and shouldn't be skipped. Removing.

api_tests/registrations/filters/test_filters.py
caseyrollins: More bad clean-up.

website/static/js/components/publicNodes.js
caseyrollins: Passed to `_buildUrl()` on LL146.

caseyrollins: And referenced several times in the view.

api/nodes/filters.py
caseyrollins: Ported from `NodesListFilterMixin`, which this mixin class replaces on the `UserNodes` view.

api/users/views.py
caseyrollins: `UserNodes` has been updated to handle all of the filtering necessary for the profile page (addition of the `NodesFilterMixin` class). I didn't want to change `UserRegistrations` in this PR too, so this class now inherits from what `UserNodes` used to.

caseyrollins:
@sloria RFR#2. All comments addressed (or discussed), button mashing shouldn't send multiple requests, no more spinners, and loading indicator should show while paginating now.

website/static/js/components/publicNodes.js
```diff
@@ -0,0 +1,344 @@
...
+        return m('ul.list-group m-md', [
+            // Error message if the request fails
+            ctrl.failed ? m('p', [
+                'Unable to retrieve public ' + ctrl.nodeType + ' at this time. Please refresh the page or contact ',
+                m('a', {'href': 'mailto:support@osf.io'}, 'support@osf.io'),
+                ' if the problem persists.'
+            ]) :
+
+            // Show laoding icon while there is a pending request
+            ctrl.requestPending() ?  m('.ball-scale.ball-scale-blue.text-center', m('')) :
```
sloria: I believe `ball-scale` is meant to be used for a single large area rather than multiple widgets on the page. `ball-pulse` might be more appropriate here. Is this correct, @caneruguz ?

sloria:
The nodes aren't rendering until all the requests are done:

profile

By default, mithril waits to render until all pending requests are finished. IIRC, you can pass `background: true` somehow to m.request to override this behavior. It would be better to render as quickly as possibl.

api/users/views.py
sloria: Yeah, totally. Thanks for the reminder.

api/users/views.py
sloria: Ah, ok, thanks. A code comment and a `TODO` would be helpful then.

website/static/js/components/publicNodes.js
```diff
@@ -0,0 +1,344 @@
...
+    }
+    else {
+        query['filter[parent][ne]'] = 'null';
+        query.embed.push('parent');
+    }
+
+    return $osf.apiV2Url('users/' + user +  '/nodes/', { query: query});
+};
+
+function _formatContributors(item) {
```
sloria: If we're not going to DRY this up now, then could you please create a follow-up ticket to do so?

website/static/js/components/publicNodes.js
sloria: ping @caneruguz .

For context:

so-much-ball-scale

website/static/js/components/publicNodes.js
```diff
@@ -0,0 +1,344 @@
...
+            ]);
+        }
+        return m('span', {}, [
+            comma,
+            m('a', {'href': person.embeds.users.data.links.html}, name)
+        ]);
+    });
+}
+
+var PublicNode = {
```
sloria: 👍  Good improvement

website/static/js/components/publicNodes.js
```diff
@@ -0,0 +1,344 @@
...
+                self.prevPage(result.links.prev);
+
+                var params = $osf.urlParams(url);
+                var page = params.page || 1;
+
+                self.currentPage(parseInt(page));
+                self.totalPages(Math.ceil(result.meta.total / result.meta.per_page));
+            }
+
+            var promise = m.request({method : 'GET', url : url, config: mHelpers.apiV2Config({withCredentials: window.contextVars.isOnRootDomain})});
```
sloria: I believe this is where you can pass `background: true`.

website/static/js/components/publicNodes.js
caneruguz: Yes, @sloria, this looks more ideal for ball-pulse.

website/static/js/components/publicNodes.js
caseyrollins: [OSF-8030](https://openscience.atlassian.net/browse/OSF-8030) created.

sloria:
I wonder if we can align these widgets better, since we know we'll always be showing 5 items in each widget:

osf___steven_loria

Some possible solutions:

- Render more than 5 "public projects" so that the height is closer to the "Public Components" widget
- Make the "Public Projects" list items larger, e.g. increasing the font size or adding more padding.

Pinging @caneruguz for input.

All that said, I don't think this needs to be a blocker, unless @nicipfeiffer or @saradbowman disagree.

sloria:
The loading indicators don't seem to have been changed-- @caseyrollins can you double-check this?

sloria:
Also, https://github.com/CenterForOpenScience/osf.io/pull/7087#issuecomment-303183954 is still an issue. Double-check that too, please.

sloria:
Mmm.. disregard my previous two comments--might have a local issue... looking into it

sloria:
Clicking a page button results in a jarring change to the widget size:

loading1

Can we avoid this? Something like:

loading

Thoughts, @caneruguz ?

Again, this isn't a hard blocker to getting this merged.

sloria:
I don't see any remaining hard blockers to this. There are three improvements that I think should be made (2 of them noted above):

- [Aligning widgets better](https://github.com/CenterForOpenScience/osf.io/pull/7087#issuecomment-303759188)
- [Improving the loading UI](https://github.com/CenterForOpenScience/osf.io/pull/7087#issuecomment-303788927)
- Using [sparse fieldsets](https://openscience.atlassian.net/browse/OSF-7208), which is planned for release on Tuesday.

We are planning to release this after 0.111.0, so late next week. Would be great if one or more the above could be merged before the release date.

sloria:
Also, in my local testing, Brian Nosek's profile is still taking >30s for *first load*. I'm wondering if there are still some queries being done in the Flask app that are unnecessary-- @caseyrollins can you look into this?

caneruguz:
@sloria; to this comment: https://github.com/CenterForOpenScience/osf.io/pull/7087#issuecomment-303788927

Yes absolutely, your solution is spot on. There will be some movement, it;s hard to avoid height changes, but page numbers should stay where they are and view should change when new data can be drawn, so overlaying the loading indicator on top is a good idea. Not a trivial thing to do though.

caneruguz:
@sloria 
on this comment: https://github.com/CenterForOpenScience/osf.io/pull/7087#issuecomment-303759188

Keeping heights the same is one of the visual alignment aspects that I am personally most willing to let go. Yes we know how many items but the content inside is dynamic, meaning project titles, contributor last names can be different heights, adding to that viewport widths may be different. So it will be a time consuming and difficult to maintain problem. 

If that's something we want to pursue our best shot for this case would be to use flex box that would essentially spread the shorter stack and equally distribute the height dynamically.

caseyrollins:
@sloria, re [this comment](https://github.com/CenterForOpenScience/osf.io/pull/7087#issuecomment-303793907): 

Nice catch. The expensive v1 calls to `get_public_projects` and `get_public_components` were still being made. Fixed in 64e90a3

website/profile/views.py
```diff
@@ -314,20 +314,20 @@ def profile_view_id_json(uid, auth):
...
+    return _profile_view(auth.user, True, embed_nodes=False, include_node_counts=True)
 
 @collect_auth
 @must_be_confirmed
 def profile_view_id(uid, auth):
     user = User.load(uid)
     is_profile = auth and auth.user == user
     # Embed node data, so profile node lists can be rendered
-    return _profile_view(user, is_profile, embed_nodes=True)
+    return _profile_view(user, is_profile, embed_nodes=False)
```
sloria: Aha, yep, that will make a difference. 👍

sloria:
Thanks for the feedback Caner. So let's defer aligning the widgets for now and focus on the other two improvements ([2 and 3 above](https://github.com/CenterForOpenScience/osf.io/pull/7087#issuecomment-303791099)). @caseyrollins If you have time to work on those before this gets released, go for it. I'll plan to merge this in after the next IR (0.110.0), so you can do those improvements as follow-up PRs.

website/profile/views.py
```diff
@@ -319,15 +319,15 @@ def profile_view_id_json(uid, auth):
...
+    return _profile_view(auth.user, True, embed_nodes=False, include_node_counts=True)
 
 @collect_auth
 @must_be_confirmed
 def profile_view_id(uid, auth):
     user = User.load(uid)
     is_profile = auth and auth.user == user
     # Embed node data, so profile node lists can be rendered
-    return _profile_view(user, is_profile, embed_nodes=True)
+    return _profile_view(user, is_profile, embed_nodes=False, include_node_counts=True)
```
sloria: After this change, first load for Brian N's profile is ~3s. Much better 💯

sloria:
Sparse fieldsets change looks good. I've created a follow-up ticket for improving the loading UI: https://openscience.atlassian.net/browse/OSF-8078 . Good to go!

PR #7091, opened 2017-04-07
Remove dead code [No Ticket]

mfraezz:
## Purpose
Remove dead code

## Changes
* Delete `website/addons`
* Remove references to deleted code
* Refactor addon `AppConfig`s to not touch models until they're loaded
* Update template routing
* Deprecate misc. mongo-related code that was affected by `website/addon` removal

## Side effects
None expected. Tested addon widgets locally to ensure routing still worked.

## Ticket
None known


website/citations/models.py
```diff
@@ -1,131 +0,0 @@
...
-        self.save()
-
-    def after_delete(self, node=None, user=None):
-        self.deauthorize(Auth(user=user), add_log=True)
-
-    def on_delete(self):
-        self.deauthorize(add_log=False)
-
-
-class CitationStyle(StoredObject):
```
cwisecarver: This doesn't need to be here for the monkey patching?

website/citations/views.py
```diff
@@ -5,7 +5,7 @@
 
 from framework.auth.decorators import must_be_logged_in
 
-from website.models import CitationStyle
+from osf.models.citation import CitationStyle
```
cwisecarver: I guess it doesn't if you fixed the imports.

website/citations/models.py
mfraezz: Old models have to be phased out eventually. This one was being troublesome, so it got the axe. 
Also,
> cwisecarver
> I guess it doesn't if you fixed the imports.

website/app.py
```diff
@@ -137,8 +124,6 @@ def init_app(settings_module='website.settings', set_backends=True, routes=True,
     app.config['SESSION_COOKIE_SECURE'] = settings.SESSION_COOKIE_SECURE
     app.config['SESSION_COOKIE_HTTPONLY'] = settings.SESSION_COOKIE_HTTPONLY
 
-    if set_backends:
-        do_set_backends(settings)
```
sloria: Might be good to raise a DeprecationWarning if `set_backends` is set, but it's not completely necessary.

mfraezz: It's still needed below for the `if set_backends and fixtures` block -- lots of tests break if that logic changes

PR #7094, opened 2017-04-10
[hotfix] Fix Preprint is_published [OSF-7433]

alexschiller:
# Previous conversation can be found here where I botched my branch https://github.com/CenterForOpenScience/osf.io/pull/7054

## Purpose
Unpublished and private preprints show up via the api.

## Changes
Add DjangoFilterMixin to be used by all preprints routes. Refactor all base querysets to return proper preprints.

Adds a boatload of tests.

## Side effects
None known.

## Ticket
https://openscience.atlassian.net/browse/OSF-7433

## QA Consierations
This makes changes to these API routes:
1. users/<user_id>/preprints/
2. nodes/<node_id>/preprints/
3. preprint_providers/<provider_id>/preprints/
4. /preprints/

A. Admins should see unpublished preprints (exception: the users route only shows public projects for that user if you are not the user being viewed)
B. Non-Admins should not see the unpublished preprint
C. Filters should work properly


api_tests/preprints/views/test_preprint_list_mixin.py
```diff
@@ -0,0 +1,156 @@
...
+    AuthUserFactory,
+    SubjectFactory,
+)
+
+from api_tests import utils as test_utils
+
+class PreprintIsPublishedListMixin(object):
+
+    def setUp(self):
+        super(PreprintIsPublishedListMixin, self).setUp()
```
sloria: This is fine for now, but in the future, we should be using pytest fixtures for this sort of thing. Using fixtures is more composable and can potentially lead to faster tests (because you only set up the objects that you need for a particular test).

osf/models/node.py
```diff
@@ -438,6 +438,11 @@ def is_preprint_orphan(self):
         return False
 
     @property
+    def has_published_preprint(self):
+        PreprintService = apps.get_model('osf.PreprintService')
+        return PreprintService.objects.filter(node_id=self.id, is_published=True).exists()
```
sloria: Always use the `related_name` (`preprints` in this case) for relationships when possible. Doing so allows Django to use queryset caching.

api_tests/preprints/views/test_preprint_list_mixin.py
alexschiller: @sloria I had a note on the previous version of this ticket commenting on that -- none of the other api tests have been ported and it drew from that functionality heavily. I figured it would be faster to port everything over at once when I understand pytest better than to do one-off tests now.

osf/models/node.py
alexschiller: @sloria Can you say more or point to an example, I'm not sure I understand the change.

sloria: Use the `related_name` (aka backref or reverse accessor) rather than querying on the related model.

```python
class PreprintService(GuidMixin, BaseModel):
    node = models.ForeignKey('osf.AbstractNode',
                             related_name='preprints',...)


# Yes
self.preprints.filter(...)

# No
PreprintService.objects.filter(node=self, ...)
```

PR #7096, opened 2017-04-10
[WIP] Add tests to check for adding unregistered contributors [OSF-7777]

erinspace:
## Purpose

There are django errors when saving a user via the v1 API as a POST request...

## Changes

- Tests for sending a POST request for unconfirmed and unregistered users via the API

## Side effects
None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-7777


sloria:
Closing this for now, since the underlying issue can't be reproduced in a test (see discussion on the ticket for an explanation).

PR #7097, opened 2017-04-10
[Hotfix] Optimize get_root [OSF-7779]

mfraezz:
## Purpose
Optimize `get_root` to speed up profile, other pages.

Before:
```
In [4]: %timeit -n 3 get_public_projects(user=u)
3 loops, best of 3: 12.4 s per loop
```
After:
```
In [4]: %timeit -n 3 get_public_projects(user=u)
3 loops, best of 3: 8.51 s per loop
```
Note: most of the remaining 8.5s is contributor selects or nodelog counts. Also, `u` has ~250 projects, 149 public.

## Changes
* Utilize `root_id` field
* Add test

## Side effects
None expected

## Ticket
[[OSF-7779]](https://openscience.atlassian.net/browse/OSF-7779)

### DIFF
--- a/osf/models/node.py
+++ b/osf/models/node.py
@@ -71,9 +71,7 @@ logger = logging.getLogger(__name__)
 class AbstractNodeQuerySet(MODMCompatibilityQuerySet, IncludeQuerySet):
 
     def get_roots(self):
-        return self.extra(
-            where=['"osf_abstractnode".id in (SELECT id FROM osf_abstractnode WHERE id NOT IN (SELECT child_id FROM '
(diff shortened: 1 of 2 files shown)


cwisecarver:
@icereval @sloria This makes a pretty big improvement to the profile page among other things. If we could get it reviewed/hotfixed the profile page, institutions list api, usage_audit, and recent_public_registrations will all speed up.

cwisecarver:
Plus, it's pretty easy to confirm it does or doesn't do what it says it does.

sloria:
Awesome

PR #7103, opened 2017-04-11
[RFR] Bring Folder.update back from the dead [OSF-7783]

cwisecarver:
## Purpose

Updating folders was failing because the File.update method was brought over from modm but the Folder.update method was not.

## Changes

Moved the BaseFileNode.update method to File. Added Folder.update method from modm. Added a test to make sure calling Folder.update calls Folder.update.

## Side effects

None expected.

## Ticket

https://openscience.atlassian.net/browse/OSF-7783?filter=17500

### DIFF
--- a/osf_tests/test_files.py
+++ b/osf_tests/test_files.py
@@ -1,11 +1,12 @@
 import pytest
+from django.utils import timezone
 
 from addons.osfstorage import settings as osfstorage_settings
+from osf.models import BaseFileNode, Folder
 from osf_tests.factories import (
     UserFactory,
     ProjectFactory,
 )
-from osf.models import BaseFileNode
 
 pytestmark = pytest.mark.django_db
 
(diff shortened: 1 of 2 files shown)


cwisecarver:
These finally passed. First it was bad migrations from master and then it was never-ending probabilistic failures.

osf_tests/test_files.py
```diff
@@ -51,3 +52,16 @@ def test_active_manager_does_not_return_trashed_file_nodes(project, create_test_
     assert BaseFileNode.objects.filter(node=project).count() == 3
     # root folder + file = 2 BaseFileNodes
     assert BaseFileNode.active.filter(node=project).count() == 2
+
+def test_folder_update_calls_folder_update_method(project, create_test_file, monkeypatch):
```
sloria: This test is a bit strange--why not just test the behavior of Folder.update rather than monkeypatching it?

That said, this is not a blocker.

cwisecarver: I don't disagree. What I really wanted to do was assert that the method on the child class was the same as the method on the Folder class. I couldn't find a way to do that. The suggested alternative on "The Internet" was to mock the method to do something obvious and simple to check and then call it like normal and assert the return was as expected. 

The test is basically to make sure that the Folder.update method continues to exist and that the Folder subclasses in addons, like OsfStorageFolder.update, continue to call Folder.update.

All that said, if you've got a better way to test that the method I'm trying to call is the method I'm trying to call I'm happy to change the implementation. Or I can put a comment in the test explaining why it seems a bit off.

sloria: I'm fine with a comment.

osf_tests/test_files.py
```diff
@@ -53,15 +52,12 @@ def test_active_manager_does_not_return_trashed_file_nodes(project, create_test_
     # root folder + file = 2 BaseFileNodes
     assert BaseFileNode.active.filter(node=project).count() == 2
 
-def test_folder_update_calls_folder_update_method(project, create_test_file, monkeypatch):
+def test_folder_update_calls_folder_update_method(project, create_test_file):
```
cwisecarver: @sloria Better? @chrisseto suggested matching on unbound methods, which seems rather obvious now.

sloria: lgtm

PR #7105, opened 2017-04-12
[feature] Add migrate specific registrations [OSF-7545]

alexschiller:
## Purpose
Current share hook script sends all registrations to share. This allows for just a few to be sent at a time.

## Changes
Add migration file.

## Side effects
N/A

## Ticket
OSF-7545

## SCRIPT NOTES
If any guid provided is not a registration it will fail an assert.

Running it locally:
`docker-compose run --rm web python -m scripts.send_specific_registration_data_to_share --targets '{"data":["vfxg2"]}'`

Getting the guids that need migration for this current use case (retracted registrations after 2017-2-17):
```
import pytz; import datetime
script_date = datetime.datetime(2017, 2, 17, 0, 0, 0, 0, pytz.UTC)
list(AbstractNode.objects.filter(retraction_id__isnull=False, date_modified__gte=script_date).values_list('guids___id',flat=True))
```


alexschiller:
@brianjgeiger There are three other similar scripts (two merged, one lingering as a hotfix pre-django that was run as a one-off and probably won't be merged) that do the same thing. Should those be moved as well?

brianjgeiger:
No; we'll do a general cleanup task for scripts at some point, but for now let's just make sure your new script is right.

sloria:
@alexschiller Why have the two steps for this migration instead of just computing the relevant registration IDs in the script?

scripts/send_specific_registration_data_to_share.py
```diff
@@ -0,0 +1,55 @@
...
+    parser.add_argument(
+        '--targets',
+        action='store',
+        dest='targets',
+        help='List of targets, of form {"data": ["registration_id", ...]}',
+    )
+    pargs = parser.parse_args()
+    script_utils.add_file_logger(logger, __file__)
+    init_app(set_backends=True, routes=False)
+    with transaction.atomic():
```
sloria: No need for a transaction (we don't want to accidentally lock records).

scripts/send_specific_registration_data_to_share.py
```diff
@@ -0,0 +1,55 @@
...
+from scripts import utils as script_utils
+from website import settings
+from website.app import init_app
+from website.project.tasks import on_registration_updated
+
+
+logger = logging.getLogger(__name__)
+
+def migrate(registrations):
+    print registrations
```
sloria: Remove prints

sloria:
I'll make the above changes.

scripts/send_specific_registration_data_to_share.py
```diff
@@ -0,0 +1,55 @@
...
+        '--targets',
+        action='store',
+        dest='targets',
+        help='List of targets, of form {"data": ["registration_id", ...]}',
+    )
+    pargs = parser.parse_args()
+    script_utils.add_file_logger(logger, __file__)
+    init_app(set_backends=True, routes=False)
+    with transaction.atomic():
+        migrate(json.loads(pargs.targets)['data'])
```
sloria: The `data` envelope is unnecessary.

alexschiller: @sloria Was following convention from https://github.com/CenterForOpenScience/osf.io/blob/develop/scripts/migration/migrate_preprints_across_providers.py

sloria: Fair enough. It's not an established convention, so omit it in the future.

brianjgeiger: If you have a gold standard for migration scripts in the new world, it would be nice to have that somewhere for people to use.

sloria: `scripts/README.md` has a basic template to use as a starting point. Given that `scripts/` is for one-off scripts, though, each script might have its own tweaks (most scripts don't need to parametrize logging level, for example). So there's no "gold standard". This isn't a big deal though, because one-off scripts are just that...one-offs.

PR #7116, opened 2017-04-13
Add subject children to subjects in preprint provider taxonomies [EOSF-491]

sheriefvt:
## Purpose
There are some cases where there is a need to retrieve subjects' children in the same request to save additional server calls from the front end. An example here is the need to show a tooltip of example subdisciplines for each top level subject in the preprint taxonomy. Currently, the subdisciplines are hard-coded in the preprint front-end, causing the display of one or more subjects that are not used (allowed) by the provider. This PR is related to PRs [326](https://github.com/CenterForOpenScience/ember-preprints/pull/326) in preprints and [202](https://github.com/CenterForOpenScience/ember-osf/pull/202) in ember-osf. The three PRs present solution 1. Another solution is presented in PR [312](https://github.com/CenterForOpenScience/ember-preprints/pull/312)

## Changes
- Update the preprint view
- Update the taxonomy serializer

## Side effects

In Figure 1, The average response time of API has been recorded using soapUI tool for both the old and new version.

**Figure 1**
image

In Figure 2, The average (finish) time to fulfill the API requests by the chrome browser based on the developer tools has been recorded

**Figure 2**
image

**JSON output of the API after including the children**
image

### DIFF
--- a/api/taxonomies/serializers.py
+++ b/api/taxonomies/serializers.py
@@ -25,6 +25,13 @@ class TaxonomySerializer(JSONAPISerializer):
     text = ser.CharField(max_length=200)
     parents = JSONAPIListField(child=TaxonomyField())
     child_count = ser.IntegerField()
+    children = ser.SerializerMethodField('get_child_num')
+
+    def get_child_num(self, obj):
+        if hasattr(obj, 'child_list_provider'):
+            return getattr(obj, 'child_list_provider')
+        else:
+            return []
 
     links = LinksField({
         'parents': 'get_parent_urls',
--- a/api/preprint_providers/views.py
+++ b/api/preprint_providers/views.py
@@ -221,6 +221,19 @@ class PreprintProviderSubjectList(JSONAPIBaseView, generics.ListAPIView):
 
     serializer_class = TaxonomySerializer
 
+    def preprare_subjects(self, provider):
+        final_subjects = set()
+        for i in provider.all_subjects:
+            final_subjects.add(i.text)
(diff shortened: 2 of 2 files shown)


brianjgeiger:
@sheriefvt Could you update this ticket with what the JSON output of the API will be before and after this change, please?

api/taxonomies/serializers.py
```diff
@@ -25,6 +25,13 @@ class TaxonomySerializer(JSONAPISerializer):
     text = ser.CharField(max_length=200)
     parents = JSONAPIListField(child=TaxonomyField())
     child_count = ser.IntegerField()
+    children = ser.SerializerMethodField('get_child_num')
+
+    def get_child_num(self, obj):
```
brianjgeiger: If you just make this method `get_children` then you don't have to set the method on the serializer method field.

brianjgeiger:
Also, since you're trying to optimize this endpoint, I need to know the relative speed differences. It looks like each request to Production for a taxonomies endpoint takes 100-200ms. What's the time difference on this? How long does the request take to get the children? I would compare locally to see, one at a time, how much the old endpoint took and the new endpoint. Also, would this be fixed by caching instead? If we just cached everything under the taxonomies endpoint and cleared it our every half an hour or so, would that just make everything better?

At what point is the API taxed? Is this really a problem, or is it a hypothetical problem? All good things to think about when you're optimizing.

laurenbarker:
Closing since this improvement was resolved by removing tooltips on subjects.

PR #7117, opened 2017-04-14
[OSF-7756] Use external django-include package

sloria:
Use the external django-include package rather than the vendorized code.


icereval:
@sloria plz, pinn to a release

sloria:
@icereval Now pinning to a PyPI release. 

I recommend that we *don't* use git flow for django-include. git flow makes sense for applications but not for packages. Development can just continue on master, since we're using tags/PyPI releases for stable releases.

PR #7122, opened 2017-04-17
[hotfix] [OSF-7811] Don't quote page_name in absolute_url and url

sloria:
## Purpose

An error was being raised by NodeWikiPage.absolute_url if the
NodeWikiPage's page_name contained special characters.
This error caused GUID resolution to fail for these NodeWikiPages.

## Changes

Don't quote `absolute_url` and `url` properties. Allow the browser to do that.

## Ticket

https://openscience.atlassian.net/browse/OSF-7811

### DIFF
--- a/addons/wiki/models.py
+++ b/addons/wiki/models.py
@@ -2,7 +2,6 @@
 import datetime
 import functools
 import logging
-import urllib
 
 import markdown
 import pytz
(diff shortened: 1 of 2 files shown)


sloria:
Build passing on my fork's TCI.

PR #7127, opened 2017-04-18
[Feature] Allow Custom Taxonomies [OSF-7788]

mfraezz:
## Purpose
Allow third-party preprint providers to submit custom taxonomies.

## Changes
* Add migrations / models changes for `Subject` and `PreprintService`
* Add conditional logic for using custom mapping if available
* Fix broken tests
* Ensure the API doesn't have breaking changes

## Side effects
None expected

## Ticket
[[OSF-7788]](https://openscience.atlassian.net/browse/OSF-7788)


osf/models/subject.py
```diff
@@ -31,6 +37,31 @@ def get_absolute_url(self):
 
     @property
     def hierarchy(self):
-        if self.parents.exists():
-            return self.parents.first().hierarchy + [self._id]
+        if self.parent:
+            return self.parent.hierarchy + [self._id]
         return [self._id]
+
+    @property
```
sloria: Drive-by comment: Can this be a `cached_property`?

osf/models/subject.py
```diff
@@ -12,7 +16,9 @@ class Subject(ObjectIDMixin, BaseModel):
     modm_query = None
 
     text = models.CharField(null=False, max_length=256, unique=True)  # max length on prod: 73
-    parents = models.ManyToManyField('self', symmetrical=False, related_name='children')
+    parent = models.ForeignKey('self', related_name='children', null=True, blank=True, on_delete=models.SET_NULL)
+    bepress_subject = models.ForeignKey('self', related_name='aliases', null=True, blank=True)
```
sloria: Drive-by: `on_delete` should be explicitly set on all ForeignKey fields. It will be a required argument in new versions of Django.

cwisecarver:
Please add reverse migrations.

osf/migrations/0015_migrate_subject_parents_to_parent.py
```diff
@@ -0,0 +1,75 @@
...
+    sql = """
+        ALTER TABLE osf_subject
+        ADD CONSTRAINT customs_must_be_mapped
+        CHECK (bepress_subject_id IS NOT NULL OR provider_id = %s);
+    """
+    try:
+        osf_id = state.get_model('osf', 'preprintprovider').objects.get(_id='osf').id
+    except models.ObjectDoesNotExist:
+        # Allow test / local dev DBs to pass
+        pass
```
mfraezz: This should perhaps give a `logger.warn`message

osf/models/subject.py
```diff
@@ -31,6 +37,31 @@ def get_absolute_url(self):
...
+
+    @classmethod
+    def create(cls, text, provider, parent=None, bepress_subject=None):
+        validate_subject_hierarchy_length(parent)
+        validate_subject_provider_mapping(provider, bepress_subject)
+        subject = cls(text=text, provider=provider, parent=parent, bepress_subject=bepress_subject)
+        subject.save()
+        return subject
+
+    # TODO: uncomment after PPS model changes
```
mfraezz: `TODO:` this

osf/models/subject.py
mfraezz: Yes. `hierarchy` probably should be too.

cwisecarver:
🙏

osf/models/preprint_provider.py
```diff
@@ -34,24 +33,34 @@ def __unicode__(self):
 
     @property
     def top_level_subjects(self):
```
sloria: Can you add a test for this property?

osf/models/preprint_provider.py
```diff
@@ -34,24 +33,34 @@ def __unicode__(self):
...
-        if len(self.subjects_acceptable) == 0:
-            return Subject.find(Q('parents', 'isnull', True))
-        tops = set([sub[0][0] for sub in self.subjects_acceptable])
-        return [Subject.load(sub) for sub in tops]
+        if self.subjects.exists():
```
sloria: Where does `subjects` come from?

osf/models/preprint_provider.py
```diff
@@ -34,24 +33,34 @@ def __unicode__(self):
...
-            if rule[1]:
-                q.append(Q('parents', 'eq', Subject.load(rule[0][-1])))
-                if len(rule[0]) == 1:
-                    potential_parents = Subject.find(Q('parents', 'eq', Subject.load(rule[0][-1])))
-                    for parent in potential_parents:
-                        q.append(Q('parents', 'eq', parent))
-            for sub in rule[0]:
-                q.append(Q('_id', 'eq', sub))
-        return Subject.find(reduce(lambda x, y: x | y, q)) if len(q) > 1 else (Subject.find(q[0]) if len(q) else Subject.find())
+        if self.subjects.exists():
```
sloria: Same here: Where is `subjects` defined?

Also please add a test for this.

osf/models/subject.py
```diff
@@ -1,9 +1,14 @@
...
 from django.db import models
+from django.core.exceptions import ValidationError
 
 from website.util import api_v2_url
 
 from osf.models.base import BaseModel, ObjectIDMixin
+from osf.models.preprint_provider import PreprintProvider
+from osf.models.preprint_service import PreprintService
+from osf.models.validators import validate_subject_hierarchy_length, validate_subject_provider_mapping
+from osf.utils.caching import cached_property
```
sloria: Use  django's cached_property. You only need osf.utils.caching.cached_property if you need to invalidate the cached property when it returns `None` or you need a setter.

osf/models/preprint_service.py
```diff
@@ -71,6 +64,12 @@ def is_preprint_orphan(self):
         return self.node.is_preprint_orphan
 
     @property
```
sloria: This can be a cached_property.

osf/models/subject.py
```diff
@@ -29,8 +36,32 @@ def child_count(self):
...
         return [self._id]
+
+    @cached_property
+    def object_hierarchy(self):
+        if self.parent:
+            return self.parent.object_hierarchy + [self]
+        return [self]
+
+    @classmethod
+    def create(cls, text, provider, parent=None, bepress_subject=None):
```
sloria: Please add a test for this method.

sloria: Also, this doesn't appear to be called. Can this be removed?

osf/models/subject.py
```diff
@@ -29,8 +36,32 @@ def child_count(self):
...
+        validate_subject_hierarchy_length(parent)
```
sloria: Can this be added as a validator to the parent field?

api_tests/preprint_providers/views/test_preprint_provider_subjects_list.py
```diff
@@ -11,9 +12,9 @@ def create_subject_rules(self):
         '''
         Subject Hierarchy
         +-----------------------------+
-        |                    +-->E    |
-        |      +-------->B+--+        |
```
sloria: Why was this hierarchy changed? Aren't we still supporting subjects with multiple parents?

osf/models/subject.py
```diff
@@ -12,7 +17,9 @@ class Subject(ObjectIDMixin, BaseModel):
     modm_query = None
 
     text = models.CharField(null=False, max_length=256, unique=True)  # max length on prod: 73
-    parents = models.ManyToManyField('self', symmetrical=False, related_name='children')
+    parent = models.ForeignKey('self', related_name='children', null=True, blank=True, on_delete=models.SET_NULL)
```
sloria: Are we not supporting multiple parents any more?

osf/models/preprint_provider.py
mfraezz: Reverse FK, from `Subject.provider`

osf/models/preprint_provider.py
mfraezz: This behavior is implicitly tested by [these tests](https://github.com/mfraezz/osf.io/blob/cb114cd94c1c56eb1dbcfda24b804a490ce14c53/api_tests/preprint_providers/views/test_preprint_provider_subjects_list.py#L79-L94), but I can add a couple model tests if you want

osf/models/preprint_provider.py
mfraezz: See [(comment)](https://github.com/CenterForOpenScience/osf.io/pull/7127/files/4ebf70540530dc021826f8caf90dceccc3af3ab3#r112549120) and [(comment)](https://github.com/CenterForOpenScience/osf.io/pull/7127/files/4ebf70540530dc021826f8caf90dceccc3af3ab3#r112550075) , respectively

osf/models/subject.py
mfraezz: No. That functionality was obviated when we moved to BePress, it just wasn't fixed before.

osf/models/subject.py
mfraezz: It's not currently used, as we never plan on exposing this functionality. It's here to be used to simplify the insertion process, when third-parties give us a mapping

osf/models/subject.py
mfraezz: ...Probably.

api_tests/preprint_providers/views/test_preprint_provider_subjects_list.py
mfraezz: No. That functionality was obviated when we moved to BePress, it just wasn't fixed before.

osf/models/subject.py
sloria: This seems to only add validation. Can we move this validation around so that we don't need to use this special method? I'm concerned that a developer would accidentally bypass validation by constructing an object the normal way.

osf/models/subject.py
```diff
@@ -29,8 +36,32 @@ def child_count(self):
...
+    @classmethod
+    def create(cls, text, provider, parent=None, bepress_subject=None):
+        validate_subject_hierarchy_length(parent)
+        validate_subject_provider_mapping(provider, bepress_subject)
+        subject = cls(text=text, provider=provider, parent=parent, bepress_subject=bepress_subject)
+        subject.save()
+        return subject
+
+    def save(self, *args, **kwargs):
+        if PreprintService.objects.filter(subjects=self).exists():
```
sloria: Can't this be

```
if self.provider:
```

?

sloria: Better to use a `related_name` rather than query on the model in order to use Django's query cache.

```
if self.preprint_services.exists():
```

osf/models/subject.py
mfraezz: Fair point

osf/models/preprint_provider.py
mfraezz: Added in https://github.com/CenterForOpenScience/osf.io/pull/7127/commits/a431d82ec0fe7ac1e8310766961ba86ea73ff9e5

osf/models/preprint_provider.py
mfraezz: Added in https://github.com/CenterForOpenScience/osf.io/pull/7127/commits/a431d82ec0fe7ac1e8310766961ba86ea73ff9e5

api/taxonomies/serializers.py
```diff
@@ -19,20 +19,29 @@ class TaxonomySerializer(JSONAPISerializer):
     filterable_fields = frozenset([
         'text',
         'parents',
+        'parent',
         'id'
     ])
     id = ser.CharField(source='_id', required=True)
     text = ser.CharField(max_length=200)
-    parents = JSONAPIListField(child=TaxonomyField())
+    parents = ser.SerializerMethodField()
```
sloria: `parents` should be deprecated. Maybe just add a comment for now, unless @brianjgeiger has a different way to mark fields as deprecated.

mfraezz: It can't be removed entirely until we remove support for older API versions -- this would break backwards compatibility.

brianjgeiger: We can deprecate and version it, but it has to continue working the same as it did in earlier versions of the API. There are examples of deprecating fields in the API; I think probably node_links is a good example.

sloria: Thanks Brian. @mfraezz Can you look into that?

mfraezz: :+1:

sloria:
LGTM. Feature branch created: `feature/custom-taxonomies`.

mfraezz:
Build broke due to filtering on deprecated field, fixed on `feature/custom-taxonomies`

PR #7130, opened 2017-04-19
Allow postgres data volume to be overridden with an envvar

sloria:
This makes it easier to use local database dumps

The compose file version was bumped to 2.1 in order to add support
for ${VARIABLE:-default}

https://docs.docker.com/compose/compose-file/#variable-substitution

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -1,5 +1,5 @@
 
-version: '2'
+version: '2.1'
 
 services:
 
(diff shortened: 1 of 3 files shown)


sloria:
Anyone who uses a local database dump might find this useful, @mfraezz @cwisecarver .

Just use the following envvar:

```
export POSTGRES_DATA_VOL="../path/to/backup"
```

PR #7133, opened 2017-04-20
Update PreprintProvider to add Identifiers [OSF-7772]

erinspace:
## Purpose

We will now automatically create new DOIs (and ARK IDs while we're at it) for new preprints that are created via the OSF. This PR is to rearrange and add new methods that will make that possible.

When a preprint is published, a new DOI and ARK will be minted and saved as an Identifier. When that preprint is updated, updated metadata about the preprint will be sent to EZID and datacite.

## Changes

- Add the `IdentifierMixin` to `PreprintService` model, along with methods to set and create identifiers
- Tests for getting preprint identifiers
- Break out methods previously found in `project/views/register.py` and 
- Modify `get_identifier` method in the `IdentifierMixin` to work for both nodes and preprints
- Migration to add Identifiers to existing preprints
- Add mocking to `PreprintService` factory so that real IDs aren't created from the EZID service when the publish method is called
- When an update is made via the API and a DOI and ARK has already been minted, send those changes to the EZID service. This includes: 
    - privacy changes 
    - adding/removing contributors
    - deleting a preprint's node
- Test to make sure update method is called with preprints
- New `datacite_metadata_for_preprint` method to format metadata for version 4 of datacite's format
- Re-write `PrerpintFactory` so that most work happens in the `_create` method so that `_build` is truly without saving

## Side effects
Script that will be run to add DOIs and ARK ids to every existing Preprint. Without the script, everything should still work fine, but all new preprints will be issued a DOI and ARK id

## QA Considerations
- All new preprints should have 2 new identifiers associated with them
- These identifiers will be visible via the API, but *not* yet visible on the preprint detail page until Ember frontend work is done
- The Identifier on the API should appear as a relationship link
- The relationship should show two Identifiers - one DOI and one ARK
- The DOI will also appear in URL form under the "links" section on the detail (and list) view for preprints
- The doi's value should be in the format `add_me/osf.io/guid` and the ARK's value should be in the format ` `
- There is a script that needs to be run to back-fill DOIs and ARK ids for existing prerpints.
- Nothing should break before that script is run

## Ticket
Main ticket
https://openscience.atlassian.net/browse/OSF-7772

Other tickets:
https://openscience.atlassian.net/browse/OSF-7807
https://openscience.atlassian.net/browse/OSF-7817


osf/migrations/0015_add_identifiers_to_preprints.py
```diff
@@ -0,0 +1,34 @@
...
+
+from osf.models import PreprintService, Identifier
+from website.identifiers.utils import get_or_create_identifiers
+
+
+def add_identifiers_to_preprints(*args):
+    preprints_without_identifiers = PreprintService.objects.filter(identifiers__isnull=True)
+
+    for preprint in preprints_without_identifiers:
+        new_identifiers = get_or_create_identifiers(preprint)
```
sloria: Drive-by: I'm not quite sure we want to do this in a Django migration. For one, this is is sending requests to an external service, EZID, which we can't always assume is available. Also, developers will not need this migration to be run in their local environment. This might be better off as a one-off script in scripts/.

Thoughts?

osf/models/preprint_service.py
```diff
@@ -43,6 +46,8 @@ class PreprintService(DirtyFieldsMixin, GuidMixin, BaseModel):
     # Format: [[root_subject._id, ..., child_subject._id], ...]
     subjects = DateTimeAwareJSONField(default=list, null=True, blank=True)
 
+    identifiers = GenericRelation(Identifier, related_query_name='preprintservices')
```
sloria: Is this ever used?

osf/models/preprint_service.py
```diff
@@ -182,6 +191,9 @@ def set_published(self, published, auth, save=False):
                     log=True
                 )
 
+            new_identifiers = get_or_create_identifiers(self)
```
sloria: This should probably triggered as a celery task.

api/preprints/serializers.py
```diff
@@ -70,7 +71,7 @@ class PreprintSerializer(JSONAPISerializer):
     date_created = DateByVersion(read_only=True)
     date_modified = DateByVersion(read_only=True)
     date_published = DateByVersion(read_only=True)
-    doi = ser.CharField(source='article_doi', required=False, allow_null=True)
```
sloria: This breaks backwards compatibility.

osf/models/preprint_service.py
erinspace: Just one place -- in the script to add identifiers to preprints that don't have them -- https://github.com/CenterForOpenScience/osf.io/pull/7133/files#diff-12a91336af7651551db5ec264ee2eee2R14

website/identifiers/utils.py
```diff
@@ -48,3 +56,80 @@ def from_anvl(data):
 
 def merge_dicts(*dicts):
     return dict(sum((each.items() for each in dicts), []))
+
+
+def get_top_level_domain(preprint):
```
sloria: It's a bit unintuitive that this function takes a `preprint` as input. A URL would make this more intuitive and reusable.

website/identifiers/metadata.py
```diff
@@ -57,3 +63,60 @@ def format_contrib(contributor):
...
+            if contributor.external_identity['ORCID'][orcid_value] == 'VERIFIED':
+                creator.append(E.nameIdentifier(orcid_value, nameIdentifierScheme='ORCID', schemeURI='http://orcid.org/'))
+
+        creators.append(creator)
+
+    return creators
+
+
+def format_subjects(preprint):
+    subject_names = set()
```
sloria: Does order matter for the subjects? Or will EZID sort them automatically?

website/identifiers/metadata.py
```diff
@@ -57,3 +63,60 @@ def format_contrib(contributor):
...
+def format_creators(preprint):
+    creators = []
+    for contributor in preprint.node.visible_contributors:
+        creator = CREATOR(E.creatorName(u'{}, {}'.format(contributor.family_name, contributor.given_name)))
+        creator.append(E.givenName(contributor.given_name))
+        creator.append(E.familyName(contributor.family_name))
+
+        # contributor.external_identity = {'ORCID': {'1234-1234-1234-1234': 'VERIFIED'}}
+        if contributor.external_identity.get('ORCID'):
+            orcid_value = contributor.external_identity['ORCID'].keys()[0]
```
sloria: Minor: This can be simplified to

```python
verified = contributor.external_identity['ORCID'].values()[0] == 'VERIFIED'
if verified:

File: osf/models/node.py [hunk 8]
```diff
@@ -1484,15 +1485,21 @@ def set_privacy(self, permissions, auth=None, log=True, save=True, meeting_creat
...
             else:
                 self.is_public = False
                 self.keenio_read_key = ''
+                status = 'unavailable'
         else:
             return False
 
+        for preprint in self.preprints.all():
+            if preprint.is_published:
+                update_ezid_metadata_on_change(preprint, status=status)
```
sloria: This should be done in celery. We shouldn't allow an error from EZID prevent the project from becoming public/private.

osf/models/node.py
```diff
@@ -2599,6 +2606,9 @@ def remove_node(self, auth, date=None):
 
         log_date = date or timezone.now()
 
+        for preprint in self.preprints.all():
+            update_ezid_metadata_on_change(preprint, status='unavailable')
```
sloria: Same here: this should be done in celery.

scripts/add_identifiers_to_existing_preprints.py
```diff
@@ -0,0 +1,36 @@
...
+from website.identifiers.utils import get_or_create_identifiers, get_top_level_domain
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+
+def add_identifiers_to_preprints():
+    from osf.models import PreprintService
+
+    preprints_without_identifiers = PreprintService.objects.filter(identifiers__isnull=True)
```
sloria: Will the API still return preprints successfully before and during this migration?

scripts/add_identifiers_to_existing_preprints.py
```diff
@@ -0,0 +1,36 @@
...
+    logger.info('About to add identifiers to {} preprints.'.format(preprints_without_identifiers.count()))
+
+    for preprint in preprints_without_identifiers:
+        new_identifiers = get_or_create_identifiers(preprint)
+        logger.info('Saving identifier for preprint {} from source {}'.format(preprint.node.title, preprint.provider.name))
```
sloria: Logging the preprint _id would be more useful than the node title here.

tests/test_identifiers.py
```diff
@@ -66,6 +68,87 @@ def test_metadata_for_node_has_correct_structure(self):
         pub_year = root.find('{%s}publicationYear' % metadata.NAMESPACE)
         assert_equal(pub_year.text, str(self.node.registered_date.year))
 
+    def test_metadata_for_preprint_has_correct_structure(self):
+        ensure_licenses()
+
+        provider = PreprintProviderFactory()
+        license = NodeLicense.objects.get(name="CC-By Attribution 4.0 International")
+        license_details = {
+                'id': license.license_id,
```
sloria: Minor: Overindentation.

scripts/add_identifiers_to_existing_preprints.py
erinspace: Yes! Preprints without identifiers show up just fine in the preprint detail and list in the API without identifiers as far as I can tell

website/settings/local-travis.py
```diff
@@ -87,3 +87,6 @@
 NEW_AND_NOTEWORTHY_LINKS_NODE = 'helloo'
 POPULAR_LINKS_NODE = 'hiyah'
 POPULAR_LINKS_REGISTRATIONS = 'woooo'
+
+EZID_USERNAME = 'testfortravisnotreal'
```
sloria: Drive-by: Why does travis need these set? Would it make sense to just mock `request_identifiers_from_ezid` instead?

website/identifiers/metadata.py
erinspace: I couldn't find any info at all about order of subjects on the [datacite metadata schema PDF](https://schema.datacite.org/meta/kernel-4.0/doc/DataCite-MetadataKernel_v4.0.pdf), so I'm guessing not? I could be wrong though.

api/preprints/views.py
```diff
@@ -350,3 +354,72 @@ def get_object(self):
...
+        )
+
+        # May raise a permission denied
+        if check_object_permissions:
+            self.check_object_permissions(self.request, preprint)
+        return preprint
+
+    # overrides ODMFilterMixin
+    def get_default_odm_query(self):
+        return MODMQ('pk', 'in', self.get_preprint().identifiers.values_list('pk', flat=True))
```
sloria: This isn't ideal--this is doing at least 3 queries (`get_preprint`, `.identifiers`, `pk__in`). I'm wondering if we can rid of the outer query by doing:

```python
def get_queryset(self):
    return self.get_preprint().identifiers.all()
```

That said, I see this code is copied from `IdentifierList`, so this is fine for now. I've created a follow-up ticket to look into this: https://openscience.atlassian.net/browse/OSF-7932

No action necessary for now.

api/preprints/views.py
```diff
@@ -350,3 +354,72 @@ def get_object(self):
             return {'citation': citation, 'id': style}
 
         raise PermissionDenied if auth.user else NotAuthenticated
+
+
+class PreprintIdentifierList(JSONAPIBaseView, generics.ListAPIView, ODMFilterMixin, PreprintMixin):
```
sloria: Hm, the inheritance doesn't feel quite right--I would expect this to inherit from `IdentifierList` (you'll need to generalize IdentifierList a bit to make this work).

```python
# api/identifiers/views.py
class IdentifierList(...):
    def get_object(self...):
        raise NotImplementedError

# api/preprints/views.py
class PreprintIdentifierList(..., IdentifierList, PreprintMixin):
     # Override IdentifierList
     def get_object(self, check_object_permissions=True):
         return self.get_preprint(check_object_permissions=check_object_permissions)

# do something simpliar with NodeIdentifierList
```

osf_tests/factories.py
```diff
@@ -505,43 +507,68 @@ class Meta:
...
+def sync_set_identifiers(preprint):
+    ezid_return_value ={
+        'response': {
+            'success': '{doi}osf.io/{guid} | {ark}osf.io/{guid}'.format(
+                doi=settings.DOI_NAMESPACE, ark=settings.ARK_NAMESPACE, guid=preprint._id
+            )
+        },
+        'already_exists': False
+    }
+    preprint.set_preprint_identifiers(ezid_return_value)
```
sloria: It's not intuitive for `set_preprint_identifiers` to receive the EZID response data. The model shouldn't really need to know anything about EZID at all. I'd suggest either getting rid of or modifying set_preprint_identifiers.

```python
identifiers = parse_identifiers(ezid_response)
preprint.set_identifier_value('doi', identifiers['doi'])
preprint.set_identifier_value('ark', identifiers['ark'])
```

Then you could avoid having to hand-write the response data in this test, and you get rid of EZID-parsing code in the model.

sloria: If you want to save some typing, then change `set_preprint_identifers` to take values.

```
preprint.set_identifier_values(doi=..., ark=...)
```

That way, you could do:

```python
identifiers = parse_identifiers(ezid_response)
preprint.set_identifier_values(**identifiers)
```

tests/test_preprints.py
```diff
@@ -234,6 +231,12 @@ def setUp(self):
         self.preprint = PreprintFactory(provider=None, is_published=False)
         self.provider = PreprintProviderFactory(name='WWEArxiv')
 
+        self.patch_toplevel = mock.patch('website.identifiers.utils.get_top_level_domain')
```
sloria: Why is this patched?

api/preprints/serializers.py
```diff
@@ -150,17 +160,17 @@ def update(self, preprint, validated_data):
             preprint.node.preprint_article_doi = validated_data['article_doi']
             save_node = True
 
+        if 'license_type' in validated_data or 'license' in validated_data:
```
erinspace: The datacite metadata has a spot for license - since the call to create a new EZID happens in publish, I wanted to set the license if it's there before the call to publish so that datacite and EZID could store the license information in the metadata

sloria: Ah OK. It would be worth adding a comment by the EZID update below that it should be called after all the updates are finished so that the metadata are correct.

website/settings/local-travis.py
erinspace: Without this set in travis, two tests that I have not added or modified  -- [this test here and the one below it](https://github.com/CenterForOpenScience/osf.io/blob/develop/tests/test_identifiers.py#L129) fail because of a check that I did add on `request_identifiers_from_ezid` - namely that if `if settings.EZID_USERNAME and settings.EZID_PASSWORD` aren't set, then nothing in that method is run and no errors are thrown, so that a preprint can still be created just without any identifiers. The two tests linked above are testing methods that are called from within `request_identifiers_from_ezid` that are already mocked using httpretty in the tests themselves.  So this is just to make two tests that were already there pass with an extra check on `request_identifiers_from_ezid` that wasn't really needed before because identifiers weren't automatically created.

erinspace:
@sloria alright! I ended up cherry-picking that lost commit, cause I have no idea what happened there... all the code that's here now is intentional tho, and tests pass on my local travis (still waitin on OSF travis to catch up) so should be ready for another look?

scripts/add_identifiers_to_existing_preprints.py
```diff
@@ -0,0 +1,36 @@
+import logging
+import time
+
+from website.app import init_app
+from website.identifiers.utils import request_identifiers_from_ezid, parse_identifiers
+
+logger = logging.getLogger(__name__)
```
sloria: Use `scripts.add_file_logger` so that logs get recorded to a file.

sloria: See scripts/README.md for an example.

website/identifiers/metadata.py
```diff
@@ -57,3 +63,60 @@ def format_contrib(contributor):
         publication_year=getattr(node.registered_date or node.date_created, 'year'),
         pretty_print=pretty_print
     )
+
+
+def format_creators(preprint):
+    creators = []
+    for contributor in preprint.node.visible_contributors:
+        creator = CREATOR(E.creatorName(u'{}, {}'.format(contributor.family_name, contributor.given_name)))
```
sloria: Minor: You can use `format_contributor` here.

erinspace:
@sloria, fixed!

scripts/add_identifiers_to_existing_preprints.py
```diff
@@ -31,6 +34,30 @@ def add_identifiers_to_preprints():
     logger.info('Finished Adding identifiers to {} preprints.'.format(preprints_without_identifiers.count()))
 
 
+def main(dry=True):
+    # Start a transaction that will be rolled back if any exceptions are un
+    with transaction.atomic():
```
sloria: This might not be a good idea...even when `--dry` is passed, a request to create the DOIs and ARKs will still be made. Also, this script will take multiple minutes, during which we'd be locking all of the preprints without identifiers.

sloria: Instead of using a transaction, how about preventing the request and database change?

```python
if not dry:
     .. request_identifiers_from_ezid...
     ...set_identifier_values(...)

Comment (erinspace) [inline scripts/add_identifiers_to_existing_preprints.py]: goooood call. Fixed now!

File: scripts/add_identifiers_to_existing_preprints.py [hunk 22]
```diff
@@ -20,27 +20,29 @@ def add_identifiers_to_preprints():
...
-        assert preprint._id.upper() in doi.value
+            doi = preprint.get_identifier('doi')
+            assert preprint._id.upper() in doi.value
 
-        logger.info('Created DOI {} for Preprint with guid {} from service {}'.format(doi.value, preprint._id, preprint.provider.name))
-        time.sleep(1)
+            logger.info('Created DOI {} for Preprint with guid {} from service {}'.format(doi.value, preprint._id, preprint.provider.name))
+            time.sleep(1)
+        else:
+            logger.info('Dry run - would have created identifier for preprint {} from service {}'.format(preprint._id, preprint.provider.name))
```
sloria: 👍

scripts/add_identifiers_to_existing_preprints.py
```diff
@@ -20,27 +20,29 @@ def add_identifiers_to_preprints():
...
 
     logger.info('Finished Adding identifiers to {} preprints.'.format(preprints_without_identifiers.count()))
 
 
 def main(dry=True):
     # Start a transaction that will be rolled back if any exceptions are un
-    with transaction.atomic():
-        add_identifiers_to_preprints()
+        add_identifiers_to_preprints(dry)
```
sloria: This needs to be dedented.

scripts/add_identifiers_to_existing_preprints.py
```diff
@@ -20,27 +20,29 @@ def add_identifiers_to_preprints():
...
         if dry:
```
sloria: This block is no longer necessary, since there's no transaction to roll back.

erinspace:
@sloria womp womp

scripts/add_identifiers_to_existing_preprints.py
```diff
@@ -20,27 +20,29 @@ def add_identifiers_to_preprints():
...
 
 def main(dry=True):
     # Start a transaction that will be rolled back if any exceptions are un
-    with transaction.atomic():
-        add_identifiers_to_preprints()
-        if dry:
-            # When running in dry mode force the transaction to rollback
-            raise Exception('Abort Transaction - Dry Run')
+    add_identifiers_to_preprints(dry)
+    if dry:
```
sloria: This block is unnecessary, since there's no transaction to roll back any more.

sloria:
Feature branch created: `feature/preprint-dois.` Consider this merged.

PR #7140, opened 2017-04-21
[Hotfix] Outlook still sometimes breaks password links [OSF-7665]

alexschiller:
## Purpose
It seems some custom Outlook setups are prefetching or navigating to urls (possibly for security reasons). The user doing the fetching always seems to be IE9.

## Changes
Block IE9 from viewing the secure pages to prevent autoclicking of routes. Throw a custom error message saying BingPreview and IE9 cannot view that page.
screen shot 2017-04-21 at 1 22 17 pm

## Side effects
People using IE9 cannot see those pages. People using IE9 already can't use the osf (very well) so this shouldn't be a big problem. 

## Ticket
[#OSF-7665]
https://openscience.atlassian.net/browse/OSF-7665

## QA NOTES:
This change affects emails for:
1. Resetting Passwords
2. External login email confirmations
3. Normal signup email confirmation
4. Unclaimed user project registration confirmations

The change should affect all of these equally, so checking each of those on a different browser should be sufficient (i.e., no need to test all 4 on every browser).

Please test password reset with IE9 to ensure it does get blocked  properly (see screenshot; per discussions this should be possible to do), all other browsers should work as normal.

### DIFF
--- a/framework/auth/decorators.py
+++ b/framework/auth/decorators.py
@@ -24,7 +24,7 @@ def block_bing_preview(func):
     def wrapped(*args, **kwargs):
         user_agent = request.headers.get('User-Agent')
         if user_agent and ('BingPreview' in user_agent or 'MSIE 9.0' in user_agent):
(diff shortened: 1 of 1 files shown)


saradbowman:
@alexschiller can we add a statement to ask the user to try another browser?

saradbowman:
"Forbidden
Internet Explorer 9 and BingPreview cannot be used to access this page for security reasons. Please use another browser. If this should not have occurred and the issue persists, please report it to support@osf.io."

PR #7153, opened 2017-04-25
[OSF-7876] Bluebook citations defined.

hmoco:
## Purpose
Allow users to make bluebook citations, as they are currently show on our list of available citation types, both via api and the project page.

## Changes
Added bluebook.csl file, used in place of the given bluebook csls, as they do not defined `bibliography` section needed by citeproc.

## Side effects
This implementation creates a very simple interpretation of a bluebook citation as defined by http://guides.libraries.uc.edu/c.php?g=222561&p=1472886. It could contain small errors and inconsistencies. It also defines all three bluebook citations available (Bluebook Law Review (2), Bluebook Law Review and Bluebook Inline) as having the same citations for the purpose of citing an OSF project.

## Ticket
https://openscience.atlassian.net/browse/OSF-7876

## Notes for QA
Would be good to test this with different kind of projects and preprints, with one and multiple authors. This can be tested by going to the project page and selecting one of the three results that show up when you search for bluebook. This can also be tested by navigating to the preprints' endpoint for a bluebook citation (those are: `v2/preprints/<guid>/citation/bluebook2/`, `v2/preprints/<guid>/citation/bluebook-law-review/` and `v2/preprints/<guid>/citation/bluebook-inline/`


api/citations/utils.py
```diff
@@ -48,18 +48,15 @@ def render_citation(node, style='apa'):
 
     bib_source = CiteProcJSON(data)
 
-    bib_style = CitationStylesStyle(os.path.join(CITATION_STYLES_PATH, style), validate=False)
+    path = os.path.join(BASE_PATH, 'static', 'bluebook') if 'bluebook' in style else os.path.join(CITATION_STYLES_PATH, style)
```
brianjgeiger: I have a feeling that we're going to want to do this again. Should we just make an array of objects with name and filename so it'll be easy in the future?

hmoco:
Seems like this broke something else, or at least changed the behavior. Will take a look at what caused the test to fail tomorrow morning. Should not impact demo.

PR #7156, opened 2017-04-26
[feature] Update PreprintProvider [OSF-7672]

caseyrollins:
❗️ needs migrations
❗️ needs `scripts.populate_preprint_providers` run

#### Purpose
- Adds supporting models so the PreprintProvider model doesn't need to be altered whenever there is a request to add a new social field, email, or link to a provider. 
  - two new social fields requested in this ticket: [OSF-7672](https://openscience.atlassian.net/browse/OSF-7672)
  - three additional URL fields requested in these tickets: [OSF-7409](https://openscience.atlassian.net/browse/OSF-7409), [OSF-7405](https://openscience.atlassian.net/browse/OSF-7405), [OSF-7623](https://openscience.atlassian.net/browse/OSF-7623)
- This PR makes it so that social accounts, emails, and links to be associated with a preprint provider can be added to the populate_preprint_providers script without requiring additional changes to the PreprintProvider model and API serializer.

#### Changes
- Versions the API to **2.4**
- New models:
  - `Email`
  - `SocialAccount` (ideally, this will replace the social JSON field on the User model in the near future)
  - `SocialNetwork` 
  - `PreprintProviderLink`
- Changes to `PreprintProvider` model:
  - `emails = ManyToManyField(Email, ...)`
    - Replaces existing email_contact and email_support fields
  - `social_accounts = ManyToManyField(SocialAccount, ...)`
    - Replaces existing social_facebook, social_instagram, social_twitter fields
  - `links = ManyToManyField(PreprintProviderLink, ...)`
    - Replaces existing external_url field
- Update `scripts.populate_preprint_providers`:
  - Sets `emails`, `social_accounts`, `links` for each preprint provider.
  - Adds an "about" link to `links` ([OSF-7409](https://openscience.atlassian.net/browse/OSF-7409))
  - Adds a "blog" link to `links` ([OSF-7405](https://openscience.atlassian.net/browse/OSF-7405))
  - Adds a "description" link, with linked text, to `links`([OSF-7623](https://openscience.atlassian.net/browse/OSF-7623))

#### Side Effects
- None expected.

#### QA Notes
- The changes in this PR shouldn't be noticeable on the front-end or to an end user.
- The `/v2/preprint_providers/` list and detail endpoints are versioned and should be tested:
  - v2.4: The attributes section of each preprint provider should contain an emails list, social_accounts list, and links list. The attributes section should **NOT** contain email_contact, email_support, social_facebook, social_instagram, or social_twitter.
  - v2.0 -  v2.3: The attributes section of each preprint provider should contain an emails list, social_accounts list, and links list. The attributes section should also contain email_contact, email_support, social_facebook, social_instagram, or social_twitter.

#### Ticket


osf/models/preprint_provider.py
```diff
@@ -20,16 +30,25 @@ class PreprintProvider(ObjectIDMixin, BaseModel):
...
-    external_url = models.URLField(null=True, blank=True, max_length=200)  # max length on prod: 25
-    email_contact = models.CharField(null=True, blank=True, max_length=200)  # max length on prod: 23
-    email_support = models.CharField(null=True, blank=True, max_length=200)  # max length on prod: 23
     example = models.CharField(null=True, blank=True, max_length=20)  # max length on prod: 5
     access_token = EncryptedTextField(null=True, blank=True)
     advisory_board = models.TextField(null=True, blank=True)
+
+    emails = models.ManyToManyField(Email, blank=True)
+    # TODO: Remove email fields in favor of emails
+    email_contact = models.CharField(null=True, blank=True, max_length=200)  # max length on prod: 23
```
cwisecarver: No null=True on CharFields.

screen shot 2017-04-28 at 1 02 22 pm

osf/models/user.py
```diff
@@ -64,6 +65,30 @@ def get_default_mailing_lists():
     ),
 }
 
+def get_default_mailing_lists():
+    return {'Open Science Framework Help': True}
+
+
+class Email(ObjectIDMixin, BaseModel):
+    email = models.EmailField(validators=[validate_email])
```
cwisecarver: Looks like `EmailField` automatically includes a validator: https://docs.djangoproject.com/en/1.11/ref/models/fields/#emailfield

osf/models/user.py
```diff
@@ -64,6 +65,30 @@ def get_default_mailing_lists():
...
+    email_type = models.CharField(choices=EMAIL_TYPES, null=True, blank=True, max_length=50)
```
cwisecarver: Null CharField again

osf/utils/social.py
```diff
@@ -0,0 +1,38 @@
...
+        '_id': 'youtube',
+        'name': 'YouTube',
+        'base_url': 'https://www.youtube.com/'
+    }
+]
+
+def ensure_social_networks():
+    SocialNetwork = apps.get_model('osf.SocialNetwork')
+    for network in SOCIAL_NETWORKS:
+        try:
```
cwisecarver: Convenience method that does the same thing: https://docs.djangoproject.com/en/1.11/ref/models/querysets/#get-or-create

scripts/populate_preprint_providers.py
```diff
@@ -36,36 +44,70 @@ def get_subject_id(name):
...
     try:
         license = NodeLicense.find_one(Q('name', 'eq', name))
     except NoResultsFound:
         raise Exception('License: "{}" not found'.format(name))
     return license
 
+def set_or_update_m2m_fields(provider, emails, licenses, links, social_accounts):
+    for email in emails:
+        try:
+            provider_email = Email.objects.get(email=email['email'], email_type=email['email_type'])
```
cwisecarver: You could use `Email.objects.get_or_create` here too.

scripts/populate_preprint_providers.py
```diff
@@ -36,36 +44,70 @@ def get_subject_id(name):
...
+            provider_email = Email(**email)
+            provider_email.save()
+        provider.emails.add(provider_email)
+
+    if licenses:
+        provider.licenses_acceptable.add(*licenses)
+
+    for link in links:
+        try:
+            provider_link = PreprintProviderLink.objects.get(url=link['url'], description=link['description'])
```
cwisecarver: get_or_create here too.

scripts/populate_preprint_providers.py
```diff
@@ -36,36 +44,70 @@ def get_subject_id(name):
...
+        except PreprintProviderLink.DoesNotExist:
+            provider_link = PreprintProviderLink(**link)
+            provider_link.save()
+        provider.links.add(provider_link)
+
+    for account in social_accounts:
+        try:
+            social_account = SocialAccount.objects.get(network=account['network'], username=account['username'])
```
cwisecarver: get_or_create

osf/utils/social.py
```diff
@@ -0,0 +1,38 @@
...
+        'base_url': 'https://www.linkedin.com/'
+    },
+    {
+        '_id': 'youtube',
+        'name': 'YouTube',
+        'base_url': 'https://www.youtube.com/'
+    }
+]
+
+def ensure_social_networks():
```
cwisecarver: I would do this as a data migration, so it only runs when it needs to. We have a bad habit of "ensure"ing things which means we read a json file full of licenses and shove them into the database every time the application starts.

osf/models/user.py
caseyrollins: Yep, but our email validator also checks against blacklisted domains (for spam).

https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/models/validators.py#L89

cwisecarver: :+2:

osf/models/preprint_provider.py
caseyrollins: @cwisecarver, email_contact is an existing field, not one that I'm adding. Should I still go ahead and remove null=True here? Also, there's 8 other nullable CharFields on this model, should I update those as well?

cwisecarver: Sorry, looked like you were adding it. During the migration all fields had to be nullable because of the way the migrations worked and the fact that our data was, previously, inconsistent. There was supposed to be a story to go back through and fix them all. Changing it and it's 8 friends is probably out of scope for your story. I'll bug @sloria to see if that story exists.

osf/models/user.py
caseyrollins: This one was definitely me. Will remove.

api/base/settings/defaults.py
```diff
@@ -151,6 +151,7 @@
         '2.1',
         '2.2',
         '2.3',
+        '2.4',
```
caseyrollins: Nope. Although this reminds me that the versioning tests really need to be revisited...

osf/models/preprint_provider.py
```diff
@@ -6,12 +6,22 @@
...
 from osf.models.licenses import NodeLicense
 from osf.models.subject import Subject
+from osf.models.user import Email, SocialAccount
 from osf.utils.datetime_aware_jsonfield import DateTimeAwareJSONField
 from osf.utils.fields import EncryptedTextField
 
 from website.util import api_v2_url
 
 
+class PreprintProviderLink(ObjectIDMixin, BaseModel):
```
mfraezz: This doesn't seem specific to `PreprintProvider` in anything but name. Consider renaming to, e.g. `ExternalLink`, or creating a base class to make this more easily extensible for other places where we have links, such as `User.social.profileWebsites`

cwisecarver: :+1: for making a generic model and naming the preprint field that references it something specific to preprints.

caseyrollins: Renamed/generalized in 33846248ef29550da4d687b52092908bd5004c4c

scripts/populate_preprint_providers.py
```diff
@@ -60,7 +59,13 @@ def set_or_update_m2m_fields(provider, emails, licenses, links, social_accounts)
         provider.licenses_acceptable.add(*licenses)
 
     for link in links:
-        obj, created = PreprintProviderLink.objects.get_or_create(**link)
+        obj, created = provider.links.update_or_create(
```
cwisecarver: Just FYI:
If you have a `{url:'Facebook.com', description:'Facebook', linked_text:'Go to the face space!'}` in the database and someone tries to add `{url:'Facebook.com', description:'Facebook', linked_text:'FACESPACE IS THE BEST'}` it's going to replace the first one since linked_text is in defaults.

cwisecarver: Actually, no. It will put `'FACESPACE IS THE BEST'` in linked_text if it's empty. It'll be thrown away otherwise.

caseyrollins: https://docs.djangoproject.com/en/1.11/ref/models/querysets/#update-or-create
> The update_or_create method tries to fetch an object from database based on the given kwargs. If a match is found, it updates the fields passed in the defaults dictionary.

The situation you described in your first comment is the desired behavior here (and based on my interpretation of the django docs, that's what will happen).

cwisecarver: Oh, okay. I think I was getting it confused with get_or_create which doesn't update using the defaults kwarg, I don't believe. 

I just thought that if two providers happened to create the same external link they might want the linked text to be different. It won't necessarily be unique across providers. If two providers put in the same url and description then one will update the other's linked_text.

caseyrollins: Hmmm, you have a point there. 

Originally, `ExternalLink` was unique on url and description, which would avoid this problem entirely, but in making the model more generic I removed that since we don't enforce any sort of uniqueness on user profile websites (which will use `ExternalLink`). 

And in this case since `update_or_create` is called on the links for a given preprint provider, one provider won't  be able to update another's external links.

cwisecarver:
@caseyrollins @brianjgeiger After talking to @icereval about this PR this afternoon maybe we can take 10m in a room tomorrow and chat about it in person? I think I might have steered you in the wrong direction or I'm confused about what we're trying to solve.

PR #7179, opened 2017-05-02
[OSF-7925] Add default_license field to preprint providers.

hmoco:
## Purpose
To allow preprint_providers to have a default license. This was done previously by just adding the items to the relationship in order, but, since the change to Django, and Django's many to many field being unordered, there is no way to indicate which of the acceptable_licenses from a provider is the default with just that field, without at least a through field. This adds a new field, default_license, to indicate which of the licenses is the default, and makes it so the licenses endpoint respects that as the first item in the array, allowing for the consuming applications to not have to change how they detect the default (first item).

## Changes
Add the `default_license` field, and change the behavior or get_queryset on the preprint_providers/licenses endpoint, so that the default_license, if it exists, is the first item in the array.

## Side effects
Migration required to add field. 

## Ticket

https://openscience.atlassian.net/browse/OSF-7925


brianjgeiger:
Update migrations please.

PR #7182, opened 2017-05-03
[feature] Allow cancelling of draft reg with deleted node [OSF-7768]

alexschiller:
## Purpose
Deleting a node before a registration is approved/canceled makes it so that the registration cannot be cancelled. 

## Changes
Changes the order of the decorators to allow the token to execute.

## Side effects
None known, permissions are checked via the token as well, so shouldn't cause any problems.

## Ticket
https://openscience.atlassian.net/browse/OSF-7768

## QA Notes
Steps to reproduce:
1. Create Node
2. Create Draft Registration (but do not approve)
3. Open draft in a new tab for easy access (to not lose the link)
4. Delete Node
5. Go to draft and click note at top to delete draft
6. Draft should now delete (although it will send you to a resource not found page with a drop down saying draft registration was deleted).

Test:
1. Approving and Rejecting a registration with a deleted and intact node
2. Test the above with embargoes as well.
3. Because permissions checking orders changed, also test using the approve/reject registration email link with a non-contributor user to ensure it fails.

### DIFF
--- a/tests/test_registrations/test_embargoes.py
+++ b/tests/test_registrations/test_embargoes.py
@@ -503,17 +503,8 @@ class RegistrationEmbargoApprovalDisapprovalViewsTestCase(OsfTestCase):
     def setUp(self):
         super(RegistrationEmbargoApprovalDisapprovalViewsTestCase, self).setUp()
         self.user = AuthUserFactory()
-        self.registration = RegistrationFactory(creator=self.user)
-
-    # node_registration_embargo_approve tests
-    def test_GET_from_unauthorized_user_raises_HTTPForbidden(self):
-        unauthorized_user = AuthUserFactory()
-        res = self.app.get(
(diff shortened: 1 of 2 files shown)


tests/test_registrations/test_embargoes.py
```diff
@@ -503,17 +503,8 @@ class RegistrationEmbargoApprovalDisapprovalViewsTestCase(OsfTestCase):
     def setUp(self):
         super(RegistrationEmbargoApprovalDisapprovalViewsTestCase, self).setUp()
         self.user = AuthUserFactory()
-        self.registration = RegistrationFactory(creator=self.user)
-
-    # node_registration_embargo_approve tests
-    def test_GET_from_unauthorized_user_raises_HTTPForbidden(self):
```
alexschiller: note, the two deleted tests were consolidated in the added tests below

PR #7088, opened 2017-04-07
Generate account confirmation and password reset links [OSF-7123]

erinspace:
## Purpose

If an OSF user's account confirmation email or password reset email is swallowed by an Institutional spam filter, Sara wants a way to quickly get these links to manually send them via the help desk!

Add buttons that display a brand new shiny account confirmation link or password reset link, with a token that expires 48 hours from generation

screen shot 2017-04-06 at 4 25 10 pm

#### Make the user action buttons into an orderly grid while we're at it
screen shot 2017-04-06 at 4 25 19 pm

## Changes

- Two new modals, urls, and views that display password and confirmation links
- Tests for those views to make sure links are generated as expected

#### unrelated changes:
- Reorganize the links at the top of the user detail page to use responsive bootstrap columns
- Organize modal HTML in its own section instead of mixed in with the top buttons
- Fix logic for which button to display for disabling and spam disabling
- Add system tag when disable as spam button is clicked
- Add note about TODO to actually fix removing the system tag (currently not possible) [Link to resulting jira ticket](https://openscience.atlassian.net/browse/OSF-7760)

## Ticket
https://openscience.atlassian.net/browse/OSF-7123

PR #7186, opened 2017-05-04
Also generate links for Unclaimed users in admin [OSF-7123]

erinspace:
## Purpose

Addendum to https://github.com/CenterForOpenScience/osf.io/pull/7088

If a user is unclaimed, generate claim links for every node that they've been added as an unregistered contributor on

screen shot 2017-05-04 at 2 43 34 pm

## Changes

- urls for unclaimed user route
- view that parses unclaimed records and turns to links
- changes to structure of modal to show list of links if necesary
- add is_claimed to user serializer
- test for link generation from unclaimed user record

## Side effects

None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-7123

### DIFF
--- a/admin/templates/users/get_link.html
+++ b/admin/templates/users/get_link.html
@@ -3,7 +3,16 @@
     <h3>{{ title }} for user {{ username }}</h3>
 </div>
 <div class="modal-body">
+    {% if user_link %}
     <p>{{ user_link }}</p>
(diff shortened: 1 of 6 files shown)


brianjgeiger:
Can you show this to Product or QA really quickly to make sure that does what they want?

PR #7187, opened 2017-05-04
[EOSF-339] Subject Filtering

hmoco:
## Purpose
To allow api filtering of preprints by the id of their subjects again. Additionally, allow filtering by text in the subject.
 
## Changes
Make subjects a filterable field on preprints.

## Ticket

https://openscience.atlassian.net/browse/EOSF-339

## QA Considerations:
Should be able to do two things:
`/v2/preprints/?filter[subjects]=<subject_id>`
`/v2/preprints/?filter[subjects]=<subject_text>`
and it should return all matching preprints.

### DIFF
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -18,6 +18,7 @@ from modularodm import Q as MQ
 from modularodm.query import queryset as modularodm_queryset
 from rest_framework import serializers as ser
 from rest_framework.filters import OrderingFilter
+from osf.models import Subject
 
 
 def lowercase(lower):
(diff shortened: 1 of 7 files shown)


api/preprints/serializers.py
```diff
@@ -63,6 +63,7 @@ class PreprintSerializer(JSONAPISerializer):
         'date_published',
         'provider',
         'is_published',
+        'subjects'
```
mfraezz: Minor, but convention is a trailing `,` here.

hmoco:
Lot's of failing tests seem to be unrelated. Would you mind taking a quick look @mfraezz

api_tests/preprints/filters/test_filters.py
```diff
@@ -137,3 +139,25 @@ def test_multiple_filters_returns_one(self):
...
+    def test_subject_filter_using_text(self):
+        expected = set([self.preprint._id, self.preprint_three._id])
+        res = self.app.get('{}{}'.format(self.has_subject, self.subject.text),
+            auth=self.user.auth
+        )
+        actual = set([preprint['id'] for preprint in res.json['data']])
+        assert_equal(expected, actual)
+
+    def test_unknows_subject_filter(self):
+        res = self.app.get('{}notActuallyASubjectIdOrTestMostLikely'.format(self.has_subject),
```
mfraezz: Unimportant, but `Text`*, probably

api/base/filters.py
```diff
@@ -636,3 +637,24 @@ def get_serializer_method(self, field_name):
...
+    def postprocess_query_param(self, key, field_name, operation):
+        if field_name == 'provider':
+            operation['source_field_name'] = 'provider___id'
+
+        if field_name == 'id':
+            operation['source_field_name'] = 'guids___id'
+
+        if field_name == 'subjects':
+            try:
+                Subject.objects.get(_id=operation['value'])
```
sloria: Drive-by comment: It would be slightly more efficient to do 

```python
if Subject.objects.filter(_id=operation['value']).exists():
```

Since you don't need the actual subject objects, you save the overhead of deserializing the data to a `Subject`.

PR #7193, opened 2017-05-08
Unclaimed vs Unregistered users fix [OSF-7123]

erinspace:
## Purpose
Better way to distinguish which set of links should be shown on a user's detail page based on their status

## Changes
- fix URL name for claim links
- change field in serializer to be more clear

## Side effects

none anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-7123


admin/users/serializers.py
```diff
@@ -17,7 +17,7 @@ def serialize_user(user):
         'two_factor': user.has_addon('twofactor'),
         'osf_link': user.absolute_url,
         'system_tags': user.system_tags,
-        'is_claimed': user.is_claimed,
+        'unclaimed': True if user.unclaimed_records else False
```
brianjgeiger: I think you could `bool(user.unclaimed_records)` here to simplify a bit.

PR #7194, opened 2017-05-08
[hotfix] stop preprint deletions [OSF-7938]

alexschiller:
## Purpose
Code preventing preprints from being deleted did not get ported over when osfstorage was moved to addons. Currently only fangorn is preventing deletions of preprint files and bad moves, but fangorn has some known flaws (lazy loaded nested folders are supposed to be caught by the backend)

## Changes
Add back in the code blocking bad preprint actions

## Side effects
None known; someone should check how many orphans were created.

## Ticket
https://openscience.atlassian.net/browse/OSF-7938

## QA Notes
This is a regression of https://github.com/CenterForOpenScience/osf.io/pull/6525. However, all of the fangorn logic is still in place so the things that need thorough testing are the areas where fangorn is known to fail: when a preprint is inside a folder that has not been expanded. Test this situation thoroughly. A light regression of the fangorn blocking of bad actions (e.g., it copies instead of moves when it has a preprint; the button is grayed out when you click a preprint) would not be a bad idea.

### DIFF
--- a/addons/osfstorage/models.py
+++ b/addons/osfstorage/models.py
@@ -128,8 +128,7 @@ class OsfStorageFileNode(BaseFileNode):
     def is_checked_out(self):
         return self.checkout is not None
 
-    @property
-    def _delete_allowed(self):
+    def _check_delete_allowed(self):
         if self.is_preprint_primary:
             raise exceptions.FileNodeIsPrimaryFile()
         if self.is_checked_out:
@@ -143,11 +142,11 @@ class OsfStorageFileNode(BaseFileNode):
     def delete(self, user=None, parent=None, **kwargs):
         self._path = self.path
         self._materialized_path = self.materialized_path
-        return super(OsfStorageFileNode, self).delete(user=user, parent=parent) if self._delete_allowed else None
+        return super(OsfStorageFileNode, self).delete(user=user, parent=parent) if self._check_delete_allowed() else None
 
     def move_under(self, destination_parent, name=None):
         if self.is_preprint_primary:
-            if self.node._id != destination_parent.node._id or self.provider != destination_parent.provider:
+            if self.node != destination_parent.node or self.provider != destination_parent.provider:
                 raise exceptions.FileNodeIsPrimaryFile()
         if self.is_checked_out:
             raise exceptions.FileNodeCheckedOutError()
@@ -355,7 +354,7 @@ class OsfStorageFolder(OsfStorageFileNode, Folder):
     @property
     def is_preprint_primary(self):
         if self.node.is_preprint:
-            for child in self.children:
+            for child in self.children.all():
                 if child.is_preprint_primary:
                     return True
         return False
--- a/addons/osfstorage/models.py
+++ b/addons/osfstorage/models.py
@@ -128,17 +128,27 @@ class OsfStorageFileNode(BaseFileNode):
     def is_checked_out(self):
         return self.checkout is not None
 
-    def delete(self, user=None, parent=None, **kwargs):
-        if self.node.preprint_file and self.node.preprint_file.pk == self.pk:
-            self.node._is_preprint_orphan = True
-            self.node.save()
+    @property
+    def _delete_allowed(self):
+        if self.is_preprint_primary:
+            raise exceptions.FileNodeIsPrimaryFile()
         if self.is_checked_out:
             raise exceptions.FileNodeCheckedOutError()
+        return True
+
+    @property
+    def is_preprint_primary(self):
+        return self.node.preprint_file == self and not self.node._has_abandoned_preprint
+
+    def delete(self, user=None, parent=None, **kwargs):
         self._path = self.path
         self._materialized_path = self.materialized_path
-        return super(OsfStorageFileNode, self).delete(user=user, parent=parent, **kwargs)
+        return super(OsfStorageFileNode, self).delete(user=user, parent=parent) if self._delete_allowed else None
 
     def move_under(self, destination_parent, name=None):
+        if self.is_preprint_primary:
+            if self.node._id != destination_parent.node._id or self.provider != destination_parent.provider:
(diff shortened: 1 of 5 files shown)


addons/osfstorage/models.py
```diff
@@ -342,6 +352,14 @@ def is_checked_out(self):
                 pass
         return False
 
+    @property
+    def is_preprint_primary(self):
```
alexschiller: The first time I wrote this I didn't like it. Having to put it in again still makes me sad. I'm sure there is a better Django-y way to do this (and the function that lives above it related to checked out files).

addons/osfstorage/tests/test_views.py
```diff
@@ -649,6 +665,30 @@ def test_move_checkedout_file(self):
         )
         assert_equal(res.status_code, 405)
 
+    def test_within_node_move_while_preprint(self):
```
alexschiller: These three tests were what I had included originally. There are some more edge cases that would ideally be covered. A followup ticket for that (and the note about is_preprint_primary) is probably needed.

addons/osfstorage/models.py
mfraezz: A recursive CTE would work, but I don't know of any Django magic for this. Check out `AbstractNodeQuerySet.get_children` if you're feeling adventurous -- you'd want something similar.

addons/osfstorage/tests/test_views.py
brianjgeiger: Go ahead and make that. Might as well add an API test to disallow deleting a preprint file as well, since that's a case that also would have been caught if this had disappeared.

addons/osfstorage/models.py
```diff
@@ -128,17 +128,27 @@ def path(self):
...
+        return self.node.preprint_file == self and not self.node._has_abandoned_preprint
+
+    def delete(self, user=None, parent=None, **kwargs):
         self._path = self.path
         self._materialized_path = self.materialized_path
-        return super(OsfStorageFileNode, self).delete(user=user, parent=parent, **kwargs)
+        return super(OsfStorageFileNode, self).delete(user=user, parent=parent) if self._delete_allowed else None
 
     def move_under(self, destination_parent, name=None):
+        if self.is_preprint_primary:
```
sloria: Better to move this below the `is_checked_out` block, so we can avoid this expensive operation for checked out files.

addons/osfstorage/models.py
sloria: @alexschiller For now, just make a DjangoLE ticket to investigate optimizing this.

addons/osfstorage/models.py
```diff
@@ -342,6 +352,14 @@ def is_checked_out(self):
                 pass
         return False
 
+    @property
+    def is_preprint_primary(self):
+        if self.node.is_preprint:
+            for child in self.children:
```
sloria: You should be able to do `.select_related('node')`

addons/osfstorage/models.py
alexschiller: @sloria Don't worry, checked out files make an equally expensive query... Might be able to combine them. 

https://github.com/alexschiller/osf.io/blob/a202b3c04fcff401e0ada1b3887dc9e6c6a21b6e/addons/osfstorage/models.py#L347

addons/osfstorage/models.py
```diff
@@ -128,17 +128,27 @@ def path(self):
     def is_checked_out(self):
         return self.checkout is not None
 
-    def delete(self, user=None, parent=None, **kwargs):
-        if self.node.preprint_file and self.node.preprint_file.pk == self.pk:
-            self.node._is_preprint_orphan = True
-            self.node.save()
+    @property
+    def _delete_allowed(self):
```
sloria: Minor: It feels a bit unintuitive for a property to do validation and raise validation errors. Perhaps make this a method, e.g. `_check_delete_allowed()`. I don't feel strongly about it though.

addons/osfstorage/models.py
```diff
@@ -128,17 +128,27 @@ def path(self):
...
+
+    def delete(self, user=None, parent=None, **kwargs):
         self._path = self.path
         self._materialized_path = self.materialized_path
-        return super(OsfStorageFileNode, self).delete(user=user, parent=parent, **kwargs)
+        return super(OsfStorageFileNode, self).delete(user=user, parent=parent) if self._delete_allowed else None
 
     def move_under(self, destination_parent, name=None):
+        if self.is_preprint_primary:
+            if self.node._id != destination_parent.node._id or self.provider != destination_parent.provider:
```
sloria: You can avoid some queries by not accessing `_id`:

```python
if self.node != destination_parent.node:

# OR

if self.node.pk != destination_parent.node.pk:
```

addons/osfstorage/models.py
alexschiller: @sloria https://openscience.atlassian.net/browse/OSF-7958

addons/osfstorage/models.py
sloria: Ah, I see. That would be worth mentioning in the ticket suggested in https://github.com/CenterForOpenScience/osf.io/pull/7194/files?diff=split#r115499174

sloria: Also, minor thing I noticed: `children.all()` is called in the code you linked; is that also necessary for `is_preprint_primary`?

addons/osfstorage/models.py
alexschiller: Deferring this change to https://openscience.atlassian.net/browse/OSF-7954 to not hold up this hotfix while waiting for travis.

brianjgeiger:
```
jshint /home/travis/build/CenterForOpenScience/osf.io/website/static/js
/home/travis/build/CenterForOpenScience/osf.io/website/static/js/filepage/index.js: line 514, col 17, Confusing use of '!'.
```

PR #7199, opened 2017-05-09
[Feature] Replace zeroclipboardjs with clipboardjs [OSF-7771]

fabmiz:
## Purpose

Replace the current module (zeroclipboard) used for copy-to-clipboard buttons
with clipboard which does not rely on Flash. Removing the flash requirement allows copying
to clipboard to work effortlessly in modern browsers.

## Changes
- Adds clipboard as bower dependency
- Removes zeroclipboard as bower dependency
- Modifies the clipboard abstract file to use clipboard
- Modifies AddonNodeConfig and CitationsNodeConfig tests to use clipboard

## QA Considerations
- This fix affects all the places on the osf that have copy-to-clipboard button. So far, I know:
files view page under **revisions**, files views page under **share** (must be public project), and view only links table in **Contributors**.

- In all the above mentioned places: please make sure the tooltip 'Copied' shows upon hitting the copy button.

## Ticket
https://openscience.atlassian.net/browse/OSF-7771


brianjgeiger:
@fabmiz Could you resolve conflicts, please?

website/static/js/clipboard.js
```diff
@@ -1,26 +1,33 @@
...
 
-    $elm.on('mouseover', function() {
-        $elm.addClass('active');
-    });
-    $elm.on('mouseout', function() {
-        $elm.removeClass('active');
+    var client = new Clipboard(elm);
+
+    client.on('success', function(e){
+        setTooltip(e.trigger, 'Copied!');
```
brianjgeiger: Since you refactored the tooltips here, you should mention in your changes so QA will be sure to test it.

PR #7202, opened 2017-05-10
[hotfix] Add keen key fix script [OSF-7408]

alexschiller:
## Purpose
Old preprint nodes are using the preprint guid to generate the keen id when it needs to be from the node guid. This is working properly inside of the osf now, but there is old broken data.

## Changes
This script will clean up the broken ones.

## Side effects
NA

## Ticket
https://openscience.atlassian.net/browse/OSF-7408

## QA
Make sure analytics work on preprints after the script is run-- the script will output at least 10 guids that were updated for testing.

### DIFF
--- a/scripts/fix_keen_preprint_keys.py
+++ b/scripts/fix_keen_preprint_keys.py
@@ -19,8 +19,8 @@ def valid_keen_key(keenio_read_key, node_id):
         ret = scoped_keys.decrypt(settings.KEEN['public']['master_key'], keenio_read_key)
         return ret['filters'][0]['property_value'] == node_id
     except Exception as error:
-        logger.info('Error on {}:'.format(node_id))
-        logger.exception(error)
+        logger.exception('Error on {}:'.format(node_id))
+        # Returns True to bypass updating of the key -- failed nodes should be investigated separately
         return True
 
 def update():
--- a/scripts/fix_keen_preprint_keys.py
+++ b/scripts/fix_keen_preprint_keys.py
@@ -1,0 +1,50 @@
+import logging
+import sys
+import django
+from django.db import transaction
+django.setup()
+
+from osf.models import Node
+from scripts import utils as script_utils
+from website import settings
+from website.app import init_app
+
+from keen import scoped_keys
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+def valid_keen_key(keenio_read_key, node_id):
+    try:
+        ret = scoped_keys.decrypt(settings.KEEN['public']['master_key'], keenio_read_key)
+        return ret['filters'][0]['property_value'] == node_id
+    except Exception as error:
+        logger.info('Error on {}:'.format(node_id))
+        logger.exception(error)
+        return True
+
+def update():
+    queryset = Node.objects.filter(preprints__isnull=False, is_public=True, is_deleted=False)
+    print('Checking {} preprint nodes'.format(queryset.count()))
+    count = 0
+    updated_nodes = []
+    for node in queryset:
+        if not valid_keen_key(node.keenio_read_key, node._id):
+            count += 1
+            node.keenio_read_key = node.generate_keenio_read_key()
+            node.save()
+            updated_nodes.append(node._id)
+    print('Updated {} node keen key(s), subset for QA = {}'.format(count, str(updated_nodes[0:10])))
+
+def main():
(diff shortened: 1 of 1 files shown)


scripts/fix_keen_preprint_keys.py
```diff
@@ -0,0 +1,50 @@
...
+    queryset = Node.objects.filter(preprints__isnull=False, is_public=True, is_deleted=False)
+    print('Checking {} preprint nodes'.format(queryset.count()))
+    count = 0
+    updated_nodes = []
+    for node in queryset:
+        if not valid_keen_key(node.keenio_read_key, node._id):
+            count += 1
+            node.keenio_read_key = node.generate_keenio_read_key()
+            node.save()
+            updated_nodes.append(node._id)
```
alexschiller: Should this write a log of every node it updates just in case?

scripts/fix_keen_preprint_keys.py
```diff
@@ -0,0 +1,50 @@
...
+logging.basicConfig(level=logging.INFO)
+
+def valid_keen_key(keenio_read_key, node_id):
+    try:
+        ret = scoped_keys.decrypt(settings.KEEN['public']['master_key'], keenio_read_key)
+        return ret['filters'][0]['property_value'] == node_id
+    except Exception as error:
+        logger.info('Error on {}:'.format(node_id))
+        logger.exception(error)
+        return True
```
sloria: Shouldn't `False` be returned here?

scripts/fix_keen_preprint_keys.py
```diff
@@ -0,0 +1,50 @@
...
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+def valid_keen_key(keenio_read_key, node_id):
+    try:
+        ret = scoped_keys.decrypt(settings.KEEN['public']['master_key'], keenio_read_key)
+        return ret['filters'][0]['property_value'] == node_id
+    except Exception as error:
+        logger.info('Error on {}:'.format(node_id))
+        logger.exception(error)
```
sloria: Minor: `logger.exception` emits the exception info of the current exception in context, so it's argument is a message, not an `Exception` object.

scripts/fix_keen_preprint_keys.py
alexschiller: False will reroll the key -- if it can't decrypt the key in the first place it should be investigated outside of the context of this ticket. The only instance I ran into was a preprint that was created prior to setting up keen. It shouldn't happen on staging/prod, but wanted to add a check to make sure the script doesn't fail.

scripts/fix_keen_preprint_keys.py
alexschiller: @sloria This was the pattern I saw everywhere when searching for logger.exception-- should a ticket be made to update those 40ish other places?

scripts/fix_keen_preprint_keys.py
sloria: @alexschiller That's a good idea.

scripts/fix_keen_preprint_keys.py
sloria: It's not a huge deal. The output will just be a bit duplicative. I'd say leave them be for now.

scripts/fix_keen_preprint_keys.py
```diff
@@ -0,0 +1,50 @@
...
+            node.keenio_read_key = node.generate_keenio_read_key()
+            node.save()
+            updated_nodes.append(node._id)
+    print('Updated {} node keen key(s), subset for QA = {}'.format(count, str(updated_nodes[0:10])))
+
+def main():
+    dry_run = '--dry' in sys.argv
+    if not dry_run:
+        script_utils.add_file_logger(logger, __file__)
+    init_app(set_backends=True, routes=False)
```
sloria: Minor: `init_app` should no longer be necessary, since you're using `django.setup()`.

scripts/fix_keen_preprint_keys.py
```diff
@@ -0,0 +1,50 @@
+import logging
+import sys
+import django
+from django.db import transaction
+django.setup()
```
sloria: Minor: You can use `website.app.setup_django` which will set `DJANGO_SETTINGS_MODULE` in addition to calling `django.setup()`, but this is fine.

PR #7204, opened 2017-05-11
Remove usages of StoredFileNode and FileNode [OSF-7854]

erinspace:
## Purpose

Currently, StoredFileNode and FileNode are aliases of BaseFileNode, and they only exist for backwards compatibility.

## Changes

- Remove aliases of `StoredFileNode` and `FileNode` from new OSF models
- Switch usages in tests, views, serializers, etc to use `BaseFileNode`

## Side effects

none anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-7854

### DIFF
--- a/addons/base/views.py
+++ b/addons/base/views.py
@@ -31,7 +31,7 @@ from website import mails
 from website import settings
 from addons.base import exceptions
 from addons.base import signals as file_signals
-from osf.models import FileNode, StoredFileNode, TrashedFileNode
+from osf.models import BaseFileNode, TrashedFileNode
 from website.models import Node, NodeLog, User
 from website.profile.utils import get_gravatar
 from website.project import decorators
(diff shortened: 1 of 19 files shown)


addons/base/views.py
```diff
@@ -439,7 +439,7 @@ def create_waterbutler_log(payload, **kwargs):
 
 @file_signals.file_updated.connect
 def addon_delete_file_node(self, node, user, event_type, payload):
-    """ Get addon StoredFileNode(s), move it into the TrashedFileNode collection
+    """ Get addon BsaeFileNode(s), move it into the TrashedFileNode collection
```
alexschiller: stupid driveby comment-- typo :)

erinspace: thank yoooooou

PR #7206, opened 2017-05-11
[feature] Add share_source to Preprint Provider [OSF-7947]

caseyrollins:
#### Purpose
- Allows the preprints app to retrieve preprints from the SHARE source in the corresponding environment.

#### Changes
- Adds `share_source` to the `PreprintProvider` model. 
- `share_source` can be set in the admin app, but cannot be imported/exported.

#### Ticket
- [OSF-7947](https://openscience.atlassian.net/browse/OSF-7947)


brianjgeiger:
When https://openscience.atlassian.net/browse/OSF-7556 is added, we'll need to update the migrations and also add this field to the "do not import/export" constant in the admin app. More details on the second bit soon.

scripts/populate_preprint_providers.py
```diff
@@ -96,7 +96,8 @@ def main(env):
             'social_instagram': '',
             'licenses_acceptable': ['CC0 1.0 Universal', 'CC-By Attribution 4.0 International', 'No license'],
             'header_text': '',
-            'subjects_acceptable': [],
+            'share_source': 'OSF' if env == 'prod' else 'OSF-{}'.format(env),
```
brianjgeiger: This file should be going away in your other PR, so probably ditch the changes in here.

osf/models/preprint_provider.py
```diff
@@ -28,6 +28,7 @@ class PreprintProvider(ObjectIDMixin, BaseModel):
     social_twitter = models.CharField(null=True, blank=True, max_length=200)  # max length on prod: 8
     social_facebook = models.CharField(null=True, blank=True, max_length=200)  # max length on prod: 8
     social_instagram = models.CharField(null=True, blank=True, max_length=200)  # max length on prod: 8
+    share_source = models.CharField(blank=True, max_length=200)
```
brianjgeiger: There should be a list or similar in the Admin to keep certain fields from being exported with the preprint_provider records. Let's update the admin to make sure that this field is not exported.

caseyrollins:
Migration updated, changes to populate script removed, and admin add updates made 👍

brianjgeiger:
@mfraezz Can I sneak this in without causing you too much trouble? It looks like it's not hitting things, but there is the migration.

PR #7211, opened 2017-05-11
[#OSF-7821]Citations leave off suffixes

chennan47:
## Purpose
1.Add suffix to citation
2.When a user only set his fullname in osf but leave family name and given name blank, make the citation show the correct citation format of his fullname instead of just leave it blank

## Changes
1.Create a new user with suffix
2.Create a new project
Before change:
screen shot 2017-05-11 at 4 33 41 pm
After change:
screen shot 2017-05-11 at 4 34 06 pm

The citation doesn't show suffix on production. But with this change, suffix show up on citation.

## Ticket
https://openscience.atlassian.net/browse/OSF-7821


tests/test_auth.py
```diff
@@ -42,6 +42,13 @@
 
 class TestAuthUtils(OsfTestCase):
 
+    def test_generate_csl_given_name(self):
+        given_name = 'Cause'
+        middle_names = 'Awesome'
+        suffix = 'Jr.'
+        csl_given_name = generate_csl_given_name(given_name, middle_names, suffix)
+        assert_equal(csl_given_name, 'Cause A ,Jr.')
```
brianjgeiger: This doesn't look right. Shouldn't it be `'Cause A, Jr.'`?

Also, there should probably be a more comprehensive set of these tests to cover all of the different combos of given, middle, and suffix, since we're here. Maybe ask someone from QA or Product what they would expect for various name combinations to return so we can make sure everyone is happy.

tests/test_auth.py
```diff
@@ -42,6 +42,13 @@
 
 class TestAuthUtils(OsfTestCase):
```
brianjgeiger: Especially since I'm asking for several more tests, it would be best not to use a test case that initializes the database for these tests.

PR #7213, opened 2017-05-12
Use custom JSON encoder to implement DateTimeAwareJSONField [OSF-7874]

erinspace:
[#OSF-7874]

## Purpose

Since upgrading to Django 1.11, we can now modify DatetimeAwareJSONField to use a custom JSON encoder in the definitoin

## Changes

Modify `DateTimeAwareJSONField` to accept a default encoder instead of using Json explicity to dump and load for encoding

## Side effects

none anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-7874

### DIFF
--- a/osf/utils/datetime_aware_jsonfield.py
+++ b/osf/utils/datetime_aware_jsonfield.py
@@ -4,7 +4,6 @@ import datetime as dt
 import json
 import logging
 from decimal import Decimal
-from functools import partial
 
 import pytz
 import ciso8601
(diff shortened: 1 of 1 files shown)


osf/utils/datetime_aware_jsonfield.py
```diff
@@ -76,19 +74,17 @@ def decode_datetime_objects(nested_value):
     return nested_value
 
 
-# TODO: This might not be necessary in Django 1.11, which allows JSONField to use
-# custom JSON encoders.
 class DateTimeAwareJSONField(JSONField):
+
+    def __init__(self, verbose_name=None, name=None, encoder=None, **kwargs):
+        encoder = DateTimeAwareJSONEncoder
```
sloria: Drive-by: Why not just make `DateTimeAwareJSONEncoder` the default for the `encoder` kwarg, i.e. `encoder= DateTimeAwareJSONEncoder`?

erinspace: that makes a whole lot of sense....

erinspace:
Unrelated to anything -- I just noticed this PR number was the same as the address of a house I lived in as a kid. That is all!

PR #7218, opened 2017-05-14
[OSF-7974] Remove old models

mfraezz:
## Purpose
Remove unused code.

## Changes
* Delete old models
* Move mixins / validators to expected locations
* Fix most imports
* Fix tests 

## Side effects
Even though the build passes, there's a chance that invalid imports still exist inside of functions that aren't being tested. This may require reasonably-intense QA

## Ticket

https://openscience.atlassian.net/browse/OSF-7974


tests/factories.py
```diff
@@ -437,25 +422,13 @@ def _create(cls, *args, **kwargs):
         return fork
 
 
-class PointerFactory(ModularOdmFactory):
-    class Meta:
-        model = Pointer
-    node = SubFactory(NodeFactory)
-
-
 class NodeLogFactory(ModularOdmFactory):
```
cwisecarver: Should `ModularOdmFactory` be changed or renamed?

tests/test_addons.py
```diff
@@ -25,10 +24,11 @@
 from addons.github.exceptions import ApiError
 from addons.github.models import GithubFolder, GithubFile, GithubFileNode
 from addons.github.tests.factories import GitHubAccountFactory
+from osf.models import Session, MetaSchema
 from osf.models import files as file_models
 from osf.models.files import StoredFileNode, TrashedFileNode
 from website.project import new_private_link
-from website.project.model import MetaSchema, ensure_schemas
+from website.project.model import ensure_schemas
```
cwisecarver: This should probably be a class method on MetaSchema.

cwisecarver: Or better yet a data migration.

tests/test_node_licenses.py
```diff
@@ -15,10 +15,10 @@
 from tests.base import OsfTestCase
 from tests.utils import assert_logs, assert_not_logs
 from website import settings
-from website.project.licenses import (NodeLicense, ensure_licenses,
-                                      serialize_node_license,
-                                      serialize_node_license_record)
-from website.project.model import NodeLog, NodeStateError
+from osf.models.licenses import NodeLicense, serialize_node_license_record, serialize_node_license
+from website.project.licenses import ensure_licenses
```
cwisecarver: This would also be good as a data migration, but it might be out of scope for this PR.

website/project/model.py
```diff
@@ -736,9 +115,10 @@ def validate_title(value):
 
 
 def validate_user(value):
```
cwisecarver: I'd probably move most of these into class methods on the models that they relate to. Django prefers fat models over helper methods.

website/project/views/node.py
```diff
@@ -465,69 +465,75 @@ def project_set_privacy(auth, node, **kwargs):
...
 def watch_post(auth, node, **kwargs):
-    user = auth.user
-    watch_config = WatchConfig(node=node,
-                               digest=request.json.get('digest', False),
-                               immediate=request.json.get('immediate', False))
-    try:
-        user.watch(watch_config)
-    except ValueError:  # Node is already being watched
-        raise HTTPError(http.BAD_REQUEST)
+    # # This behavior is deprecated, consider removing route
```
cwisecarver: I would just delete this and the route. We have source control if we need to get it back, which we won't.

cwisecarver:
Your PR was so big that safari was having trouble scrolling.

framework/auth/core.py
```diff
@@ -2,40 +2,16 @@
...
-from framework.mongo import get_cache_key
 from framework.mongo.validators import string_required
-from framework.sentry import log_exception
 from framework.sessions import session
-from framework.sessions.model import Session
-from framework.sessions.utils import remove_sessions_for_user
-from modularodm import Q, fields
-from modularodm.exceptions import (NoResultsFound, QueryException,
-                                   ValidationError, ValidationValueError)
+from modularodm import Q
```
mfraezz: Presumably the same one to [remove MODMCompatibilityQuerySet](https://openscience.atlassian.net/browse/OSF-7863)

framework/sessions/__init__.py
```diff
@@ -77,6 +76,7 @@ def prepare_private_key():
 
 
 def get_session():
+    Session = apps.get_model('osf.Session')
```
mfraezz: IIRC, importing it at the top of this file broke app initialization. I'll test that to confirm.

tests/factories.py
mfraezz: Eventually, these factories should also be removed. Usages should be replaced with the ones in `osf_tests/factories.py`

tests/test_addons.py
mfraezz: Agreed, but out of scope.

tests/test_node_licenses.py
mfraezz: >Agreed, but out of scope.

website/project/model.py
mfraezz: Wouldn't it make more sense to put them in `osf/models/validators.py`, given that they... validate?

framework/auth/core.py
cwisecarver: :+1:

tests/factories.py
cwisecarver: 👍

framework/sessions/__init__.py
cwisecarver: So long as there's a reason. I'd just rather do things the python way than the Django way if we can.

tests/test_addons.py
cwisecarver: :+1:

tests/test_node_licenses.py
cwisecarver: > :+1: todo? or additional ticket maybe?

framework/sessions/__init__.py
sloria: I agree with @cwisecarver . I'll take the blame for this, though--I think I started using apps.get_model during Django phase 1. Thinking it through now, though, using imports seems better so we don't break static analysis.

That said, I think this is not a blocker for merging this.

sloria:
Please double check any of the files in `scripts/` that we actually use. On quick glance, looks like the imports in `populate_preprint_providers` needs to be updated.

tests/test_addons.py
mfraezz: [OSF-7980](https://openscience.atlassian.net/browse/OSF-7980) created

tests/test_node_licenses.py
mfraezz: >[OSF-7980](https://openscience.atlassian.net/browse/OSF-7980) created

framework/sessions/__init__.py
cwisecarver:
```
  File "/Users/matt/Devel/osf.io/website/app.py", line 23, in <module>
    from framework.sentry import sentry
```
This actually appears to be the cause. If we localize the sentry import this should go away.,

osf/models/licenses.py
```diff
@@ -8,11 +8,11 @@
 
 def _serialize(fields, instance):
     return {
-        field: getattr(instance, field)
+        field: getattr(instance, field if field != 'id' else 'license_id')
         for field in fields
     }
 
-serialize_node_license = functools.partial(_serialize, ('_id', 'name', 'text'))
+serialize_node_license = functools.partial(_serialize, ('id', 'name', 'text'))
```
sloria: Minor: Why not just change `id` to `license_id` and get rid of the special case in `_serialize`?

osf_tests/test_queued_mail.py
```diff
@@ -62,32 +64,31 @@ def test_new_public_project_presend_for_no_project(self, mock_mail, user):
...
-    #     mail = self.queue_mail(
-    #         mail=mails.WELCOME_OSF4M,
-    #         user=user,
-    #         conference='Buttjamz conference',
-    #         fid=''
-    #     )
-    #     assert bool(mail.send_mail()) is True
-    #     assert mail.data['downloads'] == 0
+    @mock.patch('osf.models.queued_mail.send_mail')
+    def test_welcome_osf4m_presend(self, mock_mail, user):
```
sloria: Good catch  👍

website/project/model.py
cwisecarver: If they're validators, sure.  There are a few others that look like utility methods that could just be  pushed onto their respective classes. Again, might be out of scope.

osf/models/licenses.py
mfraezz: Frontend code expects `license_id` to be serialized as `id`. I can't say exactly why, but it does.

scripts/analytics/addon_snapshot.py
```diff
@@ -4,7 +4,8 @@
 from modularodm import Q
 
 from website.app import init_app
-from website.models import Node, User
+from website.models import Node
```
sloria: This should be changed

scripts/analytics/institution_summary.py
```diff
@@ -4,8 +4,9 @@
 from dateutil.parser import parse
 from datetime import datetime, timedelta
 
+from osf.models import OSFUser as User
 from website.app import init_app
-from website.models import User, Node, Institution
+from website.models import Node, Institution
```
sloria: This should be changed too

sloria:
An import in `website/templates/emails/digest.txt.mako` needs to be changed.

sloria:
An import in website/templates/emails/transactional.txt.mako needs to be changed.

framework/sessions/__init__.py
mfraezz: It does, and is replaced by the same error caused by importing `archiver.listeners`.

cwisecarver: :sadfitz:

mfraezz:
Rather than spending the entire next week incrementally resolving conflicts whenever anything goes in, I'm going to hold off until `develop` is ready for it.

PR #7223, opened 2017-05-15
[feature] Update Preprint Provider [OSF-7672]

caseyrollins:
❗️ Migration

#### Purpose
- Makes the `PreprintProvider` model a bit easier to maintain by replacing certain fields (emails, social handles, provider links) with a more general `TextField` that contains HTML (like the existing `advisory_board` field).

#### Changes
- Update PreprintProvider model:
  - `footer_links` added (`TextField`)
  - `description` changed from `CharField` to `TextField`
  - `allow_submissions` added (`BooleanField`)
  - `additional_providers` added (`ArrayField(CharField(), ...)`)
  - unused fields **removed** from the model:
    - `header_text`
    - `logo_name`
    - `banner_name`
- Update API:
  - Adds `footer_links` to preprint provider serializer.
  - Adds `allow_submissions` and `additional_providers` to preprint provider serializer (`DevOnly` fields).
  - Versioned to **2.4**
    - Removes email_support, email_contact, social_facebook, social_twitter, and social_instagram from preprint provider response.
- Update admin app:
  - Adds TinyMCE editors for the `description`, `footer-links`, and `advisory_board` fields.
- Adds management command to populate a few different types of preprint providers for local development. 

#### Deployment Notes
- Admin app will need a TinyMCE api key (although I have one that we could use). Sign up [here](https://www.ephox.com/?add-to-cart=2881) and see config instructions [here](https://www.ephox.com/blog/how-to-get-tinymce-cloud-up-in-less-than-5-minutes/).

#### QA Notes
- The changes in this PR shouldn't be noticeable on the front-end or to an end user (at least on the OSF or the preprints app -- the admin app was changed).
- The `/v2/preprint_providers/` list and detail endpoints are versioned and should be tested:
  - v2.4: The attributes section of each preprint provider should contain a footer_links field. The attributes section should **NOT** contain email_contact, email_support, social_facebook, social_instagram, or social_twitter.
  - v2.0 -  v2.3: The attributes section of each preprint provider should contain a footer_links field. The attributes section should also contain email_contact, email_support, social_facebook, social_instagram, or social_twitter.

#### Ticket 


osf/models/preprint_provider.py
```diff
@@ -15,7 +15,7 @@ class PreprintProvider(ObjectIDMixin, BaseModel):
     name = models.CharField(null=False, max_length=128)  # max length on prod: 22
     logo_name = models.CharField(null=True, blank=True, max_length=128)  # max length on prod: 17
     header_text = models.TextField(default='', blank=True)
-    description = models.CharField(null=True, blank=True, max_length=256)  # max length on prod: 56
+    description = models.TextField(null=True, blank=True)
```
cwisecarver: TextField and CharFields should be `null=False` unless there's a very good reason. `blank=True` is the preferred way and it will keep an empty string in the field if it's not provided.

cwisecarver: Previously, things were nullable for the migration.

osf/models/preprint_provider.py
```diff
@@ -28,6 +28,7 @@ class PreprintProvider(ObjectIDMixin, BaseModel):
     social_twitter = models.CharField(null=True, blank=True, max_length=200)  # max length on prod: 8
     social_facebook = models.CharField(null=True, blank=True, max_length=200)  # max length on prod: 8
     social_instagram = models.CharField(null=True, blank=True, max_length=200)  # max length on prod: 8
+    footer_links = models.TextField(null=True, blank=True)
```
cwisecarver: `null=False`

api/preprint_providers/serializers.py
```diff
@@ -13,21 +13,38 @@ class PreprintProviderSerializer(JSONAPISerializer):
     ])
 
     name = ser.CharField(required=True)
-    description = ser.CharField(required=False)
+    description = ser.CharField(required=False, allow_null=True)
```
cwisecarver: Once these are not nullable at the model level it should probably change here too.

osf/models/preprint_provider.py
```diff
@@ -28,6 +28,7 @@ class PreprintProvider(ObjectIDMixin, BaseModel):
     social_twitter = models.CharField(null=True, blank=True, max_length=200)  # max length on prod: 8
```
cwisecarver: Is there any strategy for keeping these in sync with the html fields? Since they're versioned, if someone changes the html version this won't change and then will be incorrect. Or if someone changes this version via the old API then the html version will be incorrect.

brianjgeiger: No. I think that, _technically_ the API won't be breaking, but if anyone is using the older version of the preprint_provider information, they'll quickly become sad.

osf/models/preprint_provider.py
caseyrollins: Removed. Forgot that you pointed that out last time since nearly everything in this model is still nullable.

caseyrollins:
@brianjgeiger, I don't think it would be a bad idea to keep the populate_preprint_providers script for local development.

brianjgeiger:
If we're doing that, we should make it a management command (if I remember the advice that @cwisecarver gave) and it doesn't need to be so complete.

cwisecarver:
Two choices:
- Management command: Good: Everything is contained in one file; Bad: people need to know they have to run it.
- Custom data migration: Good: Runs automatically; Bad: Every time you add/subtract to it you have to make another migration so things are spread across files.

My recommendation would be both. Make a management command that reads from a dictionary or JSON file and safely makes sure that all the providers are there. Then call that with a migration. The next time someone needs to add one change the dictionary or JSON and make another migration that calls the mgmt command.

Thoughts? @brianjgeiger @caseyrollins

osf/models/preprint_provider.py
cwisecarver: Good thing we're keeping the old versions around 😜

brianjgeiger:
I lean towards a management command with a note in cosdev or the preprint docs about running it. It's not the only thing they'd have to do to get preprints running, so adding that command to a list probably wouldn't be terrible.

cwisecarver:
I was just trying to eliminate those steps completely and push everything towards migrations. Then it's full integrated into Django so anyone who knows Django would pull the project and run migrate and expect things to work because that's how all Django projects are built.

brianjgeiger:
If it's the same work long-term, then I'm fine with doing it the way you suggest. And also so it's the standard way of doing things, I _guess_.

cwisecarver:
Technically, [the standard way of doing things is fixtures](https://docs.djangoproject.com/en/1.11/howto/initial-data/#providing-initial-data-for-models). But this gets out of hand fairly quickly if there are any relationships to this data and a primary key changes.

caseyrollins:
@cwisecarver I'm also leaning towards a management command since preprint providers vary across environments (staging v. prod). I'm not entirely sure how that would work with a data migration, or if it would being a lot messier than a single management command. Thoughts?

sloria:
We'll need a way to specify which set of providers to use. Right now we do that with a positional argument: `python -m scripts.populate_preprint_providers stage`.

With a mgmt command, just use a required option: 

```
python manage.py populate_preprint_providers --env=stage
```

With a migration, I guess we could use an envvar...

```
ENV=stage python manage.py migrate
```

Haven't thought about this enough to have a strong preference either way...

brianjgeiger:
We will no longer be using this for staging environments. The admin module will be used to configure all staging environments. The management command is only for local development.

sloria:
@brianjgeiger But we'll still have multiple environments, right? Prod vs dev?

brianjgeiger:
No, we are not using that management command except for local development. If it's been deployed, they should use the admin app.

caseyrollins:
@brianjgeiger You're right. I forgot admin existed on all environments and not just prod. A data migration or @cwisecarver's hybrid recommendation will work for local dev.

sloria:
Ah, I see. I think a mgmt command is fine. I think they hybrid approach could be kinda nice, but not really a blocker.

caseyrollins:
I'm going to make this a management command unless there are objections (@cwisecarver?).

This is really just a convenience tool for devs to populate a few preprint providers with different settings -- think of it as create_fakes but for preprint providers. The JSON file won't be in sync with the provider data on staging and production (since those models will be updated by the admin app) and running a migration would reset those models.

brianjgeiger:
@caseyrollins For the general breakout, I think we'll want 4 providers:

1. Super-simple, straight bepress taxonomy, no custom domain, two licenses
2. Basic change to taxonomy showing some simple hierarchy, a custom domain but no redirect, three licenses
3. Custom taxonomy of the type that we just had a release to support, no custom domain, all the licenses
4. Straight except custom domain that redirects to that domain, one license.

Something like that.

brianjgeiger:
Based on the description for https://github.com/CenterForOpenScience/osf.io/pull/7247 we should definitely have a complete provider for OSF.

admin/templates/preprint_providers/update_preprint_provider_form.html
```diff
@@ -1,3 +1,72 @@
...
+                plugins: 'link lists preview',
+                toolbar: 'bold italic underline | undo redo | link | bullist | preview'
+            }, TINYMCE_BASE_CONFIG);
+
+            var FOOTER_LINKS_CONFIG = $.extend({
+                selector:'textarea[name=footer_links]',
+                content_css: [
+                    'https://netdna.bootstrapcdn.com/font-awesome/4.6.3/css/font-awesome.min.css',
+                ],
+                valid_elements : "@[id|class|style|title|dir<ltr?rtl|lang|xml::lang|onclick|ondblclick|"
```
caseyrollins: Ah, no. I can narrow down this list. 
Essentially, this widget just couldn't accept everything (`"*[*]"`) like the others.

brianjgeiger: Not for this release, but I'm probably going to soon update the other two to me *much* more restrictive.

admin/preprint_providers/forms.py
```diff
@@ -30,3 +32,27 @@ def clean_subjects_acceptable(self, *args, **kwargs):
         subjects_selected = [Subject.objects.get(id=ident) for ident in subject_ids]
```
cwisecarver: `Subject.objects.filter(id__in=subject_ids)` ?

caseyrollins: Wasn't added in this PR, but I can go ahead and update that.

osf/models/preprint_provider.py
```diff
@@ -14,21 +14,19 @@
 
 class PreprintProvider(ObjectIDMixin, BaseModel):
```
brianjgeiger: Please add two fields to the model:

- `allow_submissions`, boolean, default to true, all existing providers should be set to true;
- `additional_providers`, list of strings, will be used to allow the provider to mix in other SHARE-harvested preprint providers into their whitelist.

The API fields should be dev_only for the moment. This is to support https://openscience.atlassian.net/browse/EOSF-680?filter=18001

admin/preprint_providers/forms.py
cwisecarver: Thanks! Once people get more familiar with the ORM they'll stop doing that stuff.

osf/models/preprint_provider.py
cwisecarver: There isn't a ListField in SQL. Depending on how `additional_providers` will be used it could either be a `CharField`, `JSONField` or an `ArrayField(CharField())`. Not sure which would be better based on the above. If we're not going to query inside it or pull out individual strings from inside it then I'd just make it a CharField that's comma-delimited.

brianjgeiger: It's going to be sent to SHARE as part of a query. We won't be querying inside of it, but we might need to use the individual strings. Comma delimiting will be easy to parse, though.

cwisecarver: If you did the array field you could validate each string individually, you'd just have to concatenate the results when you need the full string. The more I think about the comma-delimited string the more I feel bad about myself for suggesting it.

brianjgeiger: Array field works for me.

osf/models/preprint_provider.py
```diff
@@ -14,22 +15,22 @@
...
     access_token = EncryptedTextField(null=True, blank=True)
-    advisory_board = models.TextField(null=True, blank=True)
+    advisory_board = models.TextField(default='', blank=True)
     social_twitter = models.CharField(null=True, blank=True, max_length=200)  # max length on prod: 8
     social_facebook = models.CharField(null=True, blank=True, max_length=200)  # max length on prod: 8
     social_instagram = models.CharField(null=True, blank=True, max_length=200)  # max length on prod: 8
+    footer_links = models.TextField(default='', blank=True)
     share_source = models.CharField(blank=True, max_length=200)
+    allow_submissions = models.BooleanField(default=True)
+    additional_providers = fields.ArrayField(models.CharField(max_length=200), default=list, blank=True)
```
caseyrollins: @cwisecarver +1?

osf/models/preprint_provider.py
caseyrollins: New fields added in 0909c3f

osf/models/preprint_provider.py
cwisecarver: +at-least-2. I didn't know that you should set the default on an `ArrayField` to `list`, nice.

caseyrollins: Benefits of having the django docs bookmarked... and organized into multiple folders...

https://docs.djangoproject.com/en/1.11/ref/contrib/postgres/fields/#django.contrib.postgres.fields.ArrayField
> If you give the field a default, ensure it’s a callable such as list (for an empty default) or a callable that returns a list (such as a function). Incorrectly using default=[] creates a mutable default that is shared between all instances of ArrayField.

cwisecarver: Yep, I went and found the same thing because I'd never seen that before, or at least I didn't remember. You should checkout [dash](https://kapeli.com/dash). Then you can get them when the power is out and the internet doesn't work. :high-five:

PR #7224, opened 2017-05-15
Optimize checkout and is_preprint_primary checks for folders [OSF-7958]

erinspace:
[#OSF-7958]

## Purpose

On `OsfStorageFolder`, the properties `is_preprint_primary` and `is_checked_out` have a few recursive calls to see if a folder contains a preprint or checked out file. This can be optimized!

## Changes

- optimize check for `is_preprint_primary`
- optimize check for `is_checked_out`

## Side effects

None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-7958

### DIFF
--- a/addons/osfstorage/models.py
+++ b/addons/osfstorage/models.py
@@ -342,21 +342,15 @@ class OsfStorageFolder(OsfStorageFileNode, Folder):
                 return True
         except AttributeError:
             return False
-        # TODO this should be one query
-        for child in self.children.all():
-            try:
-                if child.is_checked_out:
-                    return True
-            except AttributeError:
-                pass
+        if self.children.filter(checkout=True).exists():
+            return True
         return False
 
     @property
     def is_preprint_primary(self):
         if self.node.preprint_file:
-            for child in self.children.all().select_related('node'):
-                if child.is_preprint_primary:
-                    return True
+            if self.children.filter(node__preprint_file__isnull=False).filter(node___has_abandoned_preprint=False).exists():
+                return True
         return False
 
     def serialize(self, include_full=False, version=None):
--- a/addons/osfstorage/models.py
+++ b/addons/osfstorage/models.py
@@ -337,20 +337,14 @@ class OsfStorageFolder(OsfStorageFileNode, Folder):
 
     @property
     def is_checked_out(self):
-        try:
-            if self.checkout:
-                return True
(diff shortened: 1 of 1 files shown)


addons/osfstorage/models.py
```diff
@@ -342,21 +342,15 @@ def is_checked_out(self):
...
+            return True
         return False
 
     @property
     def is_preprint_primary(self):
         if self.node.preprint_file:
-            for child in self.children.all().select_related('node'):
-                if child.is_preprint_primary:
-                    return True
+            if self.children.filter(node__preprint_file__isnull=False).filter(node___has_abandoned_preprint=False).exists():
```
alexschiller: This appears to be block the deletion of any folder that has a file in it (if the node is a preprint), not just folders that contain preprints.

Additionally, you cannot move any folder with a file in it into a component (if the node is a preprint).

erinspace:
Putting this one on hold for now until tests from another ticket are in! Will make things easier and avoid duplicate effort!

addons/osfstorage/models.py
```diff
@@ -337,26 +337,13 @@ class OsfStorageFolder(OsfStorageFileNode, Folder):
...
-            return False
-        # TODO this should be one query
-        for child in self.children.all():
-            try:
-                if child.is_checked_out:
-                    return True
-            except AttributeError:
-                pass
-        return False
+        return self.children.filter(checkout__isnull=False).exists()
```
alexschiller: I looked this over and decided to add two more tests to my blocking ticket (in CR). I'm pretty sure this is going to allow deep nested checked-out files to be moveable to another component. Sorry for missing that in my tests the first time around.

addons/osfstorage/models.py
```diff
@@ -337,26 +337,13 @@ class OsfStorageFolder(OsfStorageFileNode, Folder):
...
 
     @property
     def is_preprint_primary(self):
-        if self.node.preprint_file:
```
alexschiller: I think the `self.node.preprint_file` check prevents the query from ever having to happen if it isn't a preprint. Removing it might slow down checks on nodes that are not preprints.

erinspace:
Matt F helped out with profiling the new SQL query on production, ran in an average of 30ms or so!

sloria:
This is rad.

PR #7226, opened 2017-05-16
[noticket] Swap FileNode - BaseFileNode

alexschiller:
## Purpose
One missed file with FileNode changes borked admin.

## Changes
Fixes that file

## Side effects
NA

## Ticket
NA


sloria:
Thanks!

PR #7229, opened 2017-05-16
[OSF-7942] Automatic register campaign for preprint providers

hmoco:
## Purpose
Preventing the need to manually add handling of new providers campaigns in the registration page.

## Changes
Use mako template syntax and fetch the providers on the view to add handling of all providers' campaigns.

##Notes for QA
Just make sure that all of the providers work properly on the register page.

## Ticket
https://openscience.atlassian.net/browse/OSF-7942


website/templates/public/register.mako
```diff
@@ -92,57 +58,20 @@
...
-                %else:
+                %elif campaign not in preprint_campaigns.keys():
                     <h3 class="m-b-lg"> Create a free account </h3>
+                %else:
+                    %for provider in preprint_campaigns.keys():
+                        %if campaign == provider:
+                            <table style="border-collapse: separate; border-spacing: 30px 0; margin-top: 20px;  margin-bottom: 10px;">
+                                <tr>
+                                    <!-- Path to be changed to /wherever preprint assets are/id/square_color_no_transparent.png -->
+                                    <td><img src="${"/static/img/preprint_providers/" + preprint_campaigns[provider]['id'] + "-login.png"}" style="width: 200px; margin-top: 15px" /></td>
```
brianjgeiger: I should probably make these images 200px, then.

Also, call these images `/static/preprint_assets/<provider_name>/square_color_no_transparent.png` and we'll need to move files into there. I'll try to make these while you update (but you'll have your changes done sooner).

brianjgeiger: Also, make a PREPRINTS_ASSETS in defaults.py that is `/static/`and use that for the beginning of that path. So when the assets are in, we can just replace that variable.

hmoco:
@brianjgeiger Using the new files, some look good (better), and some look a little not so good (little worse) and don't seem to use the same files as the ones currently used:
## Before:
screen shot 2017-05-16 at 1 51 01 pm
screen shot 2017-05-16 at 1 50 45 pm
screen shot 2017-05-16 at 1 50 36 pm

## After:
screen shot 2017-05-16 at 1 32 17 pm
screen shot 2017-05-16 at 1 32 09 pm
screen shot 2017-05-16 at 1 52 15 pm

brianjgeiger:
I think probably we should go back to 100 x 100 logos or we could do the 300 x 140 logos.

framework/auth/views.py
```diff
@@ -29,7 +29,7 @@
 from framework.sessions import get_session
 
 from website import settings, mails, language
-from website.models import User
+from website.models import User, PreprintProvider
```
mfraezz: Please import the real model to minimize conflicts with #7218

PR #7231, opened 2017-05-16
[hotfix] Override CheckoutField get_choices method to use pk  [OSF-7967]

erinspace:
## Purpose
In a recent django rest framework update, they [changed the choices](https://github.com/encode/django-rest-framework/commit/f1a2eeb818366b11c44c86d2ccfa1d6822d216b9) for a field (in a model form) to be derived from the result of `get_representation` instead of just coercing the value to a string - this method for choices is called by [the base class for our browsable docs renderer](https://github.com/CenterForOpenScience/osf.io/blob/develop/api/base/renderers.py#L28) (even tho we don't really use it), and so needs to be changed to use a pk instead of the return value of `to_representation`, which is a dict due to our JSON API compliant API.

This was causing our file detail endpoint to error when instantiating the renderer for the browsable API docs, even tho we were removing those forms after the fact.

## Changes

- override the `get_choices` method for `CheckoutField` to use the item's pk instead of the return value from `get_representation`

## Side effects

None anticipated

## QA Considerations
To test this, visit any file's direct detail page on the API at `/v2/files/<file_id>/`
-- I got there by first going to my node's osfstorage endpoint (or other provider I suppose, this should be provider agnostic) at `/v2/nodes/<guid>/files/osfstorage/` and then clicking on the `versions.related.href` link -- and **then** removing `/versions/`. Without this fix, doing so will result in a 500 error, with this fix the detail page should render just fine and all info should be visible.

## Ticket
https://openscience.atlassian.net/browse/OSF-7967

### DIFF
--- a/api/files/serializers.py
+++ b/api/files/serializers.py
@@ -1,4 +1,5 @@
 from datetime import datetime
+from collections import OrderedDict
 
 from django.core.urlresolvers import resolve, reverse
 from modularodm import Q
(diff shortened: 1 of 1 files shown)


sloria:
Ref https://github.com/encode/django-rest-framework/issues/5141

PR #7233, opened 2017-05-17
[OSF-7965] Allow the use of iexact on django filter operations

hmoco:
## Purpose
Allow the use of iexact on django filter operations and use it for subject text filtering

## Changes
`iexact` and `exact` can now be used for DjangoFilter queries.

## QA Considerations:
This will need testing to make sure it doesn't blow up api when querying on a large amount of preprints


hmoco:
A potential alternative to `icontains` would be `iexact`, as substring filtering is probably not needed

sloria:
`iexact` seems more correct. Users may want, e.g. `engineering` without returning the results for `chemical engineering`.

brianjgeiger:
So default to iexact and allow icontains as an option?

hmoco:
Changed it to just iexact, since there is no current usecase for contains but we can have both

brianjgeiger:
I would think a contains type search would be more likely, personally (find all of the `bio` things, for example), but we can change it later.

brianjgeiger:
@hmoco There's a jira ticket for this, right?

hmoco:
@brianjgeiger, could you take a look at the failing test to make sure it is not a probabilistic one? As this change (https://github.com/CenterForOpenScience/osf.io/pull/7233/commits/fdf82ef8564b497372c6ebc5db80b9c0a86faaa4) shouldn't break 
```
_________________ ERROR collecting osf_tests/test_archiver.py __________________
ImportError while importing test module '/home/travis/build/CenterForOpenScience/osf.io/osf_tests/test_archiver.py'.
Original error message:
'cannot import name cleanup_failed_registrations'
Make sure your test modules/packages have valid Python names.
```

sloria:
The failing test is unrelated to this PR and has since been fixed on develop.

PR #7242, opened 2017-05-18
Add docs for debugging services

Johnetordoff:
## Purpose

Some of the documentation surrounding debugging services in a docker environment were unclear or incomplete. This adds some notes about how to best debug with docker-sync.

## Changes

Adds and reorganizes to README-docker-compose.md

## Side effects

None

### DIFF
--- a/README-docker-compose.md
+++ b/README-docker-compose.md
@@ -86,48 +86,6 @@
 
       _NOTE: Similar docker-compose.\<name\>.env environment configuration files exist for services._
 
-3. OPTIONAL (skip if you do not need to modify services, e.g. mfr and waterbutler): Mounting Service Code
-  - By modifying docker-compose.override.yml and docker-sync.yml you can specify the relative path to your service code directories. e.g.
-    - This makes it so your local changes will be reflected in the docker containers. Until you do this none of your changes will have any effect.
-
-  - In `docker-compose.override.yml`:
-
-    ```yml
-    services:
-      wb:
-        volumes_from:
-          - container:wb-sync
-
-    ...
-    ```
-
-  - In `docker-sync.yml`:
-
-    ```yml
-    syncs:
-      wb-sync:
-        src: '../waterbutler'
-        dest: '/code'
-        sync_strategy: 'unison'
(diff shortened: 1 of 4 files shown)


README-docker-compose.md
```diff
@@ -229,6 +187,79 @@ Ubuntu: Skip install of docker-sync, fswatch, and unison. instead...
...
+    volumes_from:
+      - container:wb-sync
+  ```
+  becomes
+  ```yml
+    wb:
+    volumes_from:
+      - container:wb-sync
+    environment:
+      - PYTHONUNBUFFERED=0
```
abought: Rather than editing the dockerfile, might I suggest instructing users to put this in the separate user-specific `<wb|mfr>.env` files instead? This would reduce the odds that a personal setting might sneak into a generic docker-related PR in the future.

README-docker-compose.md
```diff
@@ -229,6 +187,79 @@ Ubuntu: Skip install of docker-sync, fswatch, and unison. instead...
...
+  git update-index --skip-worktree docker-compose.override.yml docker-sync.yml
+  ```
+  
+  To be able to commit changes to these files again, run:
+  
+  ```bash
+  git update-index --no-skip-worktree docker-compose.override.yml docker-sync.yml
+  ```
+
+  This will allow local changes to be reflected in your Docker environment, but you'll have to run `docker-compose up --force-recreate <container name>` every time you modify code in order to see your changes reflected in the application. But fear not! Setting `DEBUG=1` and `SERVER_CONFIG_DEBUG=1` in `docker-compose.wb.env` or `docker-compose.mfr.env` will enable live reload for those services, so your changes will take effect automatically in a few seconds.
```
abought: Small note: `--force-recreate` is only required once, when you first changes the volumes associated with docker sync. After that it will always know to look for the correct storage volume when the docker container is "turned on".

I suggest the following wording, to place greater emphasis on the recommended solution:
> The first time that sync settings are changed, you will need to run `docker-compose up --force-recreate <container name>`. To see the effect of code changes as you work (without needing to restart the container), you will need to separately turn on debug mode for the service: ...

PR #7248, opened 2017-05-19
[Feature] Add common actions menu to component widget [OSF-7442]

fabmiz:
## Purpose

Add quick actions menu to component widget on project overview page.
Menu items include:
- Settings link to the component's settings page
- Manage contributors link to the component's contributors page
- Delete option to allow quick delete of the component.

## Changes
- Add html for the menu and css consistent with other osf styles to render_node.mako
- Add component-settings-page.js for click events handling
- Modifies projectSettings.js to allow api url to be passed as argument instead of contextVars
- Modifies project-settings-page.js so it does not break because of changes to projectSettings.js

## Ticket
https://openscience.atlassian.net/browse/OSF-7442

## QA Considerations
Some edge cases QA will want to consider
1. Test deleting both public and private preprints
2. Test that the common actions dropdown button is not visible to people with proper permission
3. Confirm that the delete component button does not show up if the user does not have (admin) permission  to delete the component. 
4. On a registration's component widget, the menu should only have one option (Manage contributors).
5. The button should not show up for non contributors even if parent project is public.
5. Make sure deleting components through the settings page works
6. Make sure nav bar drop down style looks the same (there were some early interactions here)


website/templates/util/render_node.mako
```diff
@@ -1,10 +1,32 @@
 <%namespace name="contributor_list" file="./contributor_list.mako" />
 ## TODO: Rename summary to node
-<%def name="render_node(summary, show_path)">
+<%def name="render_node(index, summary, show_path)">
 ## TODO: Don't rely on ID
+<style>
```
alexschiller: Belongs in style.css (ideally in osf-style, but that can be delayed). Add a comment to the style section where this is added that states it should be moved to osf-style.

website/templates/util/render_node.mako
```diff
@@ -58,7 +80,18 @@
...
                 % endif
-                <i id="icon-${summary['id']}" class="pointer fa fa-angle-down" onclick="NodeActions.openCloseNode('${summary['id']}');" style="font-weight:bold;"></i>
+                <div class="dropdown pull-right">
+                  <button class="btn btn-default dropdown-toggle quickActions" type="button" data-toggle="dropdown">
+                    <span class="glyphicon glyphicon-option-horizontal"></span>
+                  </button>
+                  <ul class="dropdown-menu dropdown-menu-right" id="componentQuickActions">
+                    <li><a tabindex="-1" href="${domain}${summary['id']}/settings/">Settings</a></li>
+                    <li><a tabindex="-1" href="${domain}${summary['id']}/contributors/">Manage Contributors</a></li>
+                    % if 'admin' in user['permissions']:
```
alexschiller: Was admin in spec or did it include write contrib? This could get tricky as you may need to assess if the person has write permissions on the component, not on the current project?

website/static/js/pages/component-settings-page.js
```diff
@@ -0,0 +1,38 @@
...
+
+$(document).ready(function() {
+
+  //Project overview >> Component Widget >> Event/Click handler for componentQuickActions menu item: Delete.
+  $('.deleteComponent').each( function() {
+    $(this).off().on('click', function(e) {
+      var componentIndex = $(this).data('index');
+      var component = window.contextVars.nodes[componentIndex].node;
+
+      var {node_type, api_url, isPreprint, childExists} = component;
```
alexschiller: Probably can just access via `component.api_url`, etc. I don't believe the `{var_a, var_B} = obj` pattern is used at all in the osf (though I don't know if there was a particular reason why).

website/views.py
```diff
@@ -88,6 +88,8 @@ def serialize_node_summary(node, auth, primary=True, show_path=False):
             'api_url': node.api_url,
             'title': node.title,
             'category': node.category,
+            'isPreprint': node.is_preprint,
```
alexschiller: This will need to check if the file has a preprint id instead, `is_preprint` becomes false if the preprint is made private, which can cause problems.

website/templates/util/render_node.mako
```diff
@@ -127,23 +160,31 @@
...
+                    <i class="fa fa-times remove-pointer pointer" data-id="${summary['id']}"
+                    data-toggle="tooltip" title="Remove link"></i>
                 </span>
             % endif
         </p>
     </li>
-
 % endif
 </div>
 <script type="text/javascript">
```
alexschiller: Are there other places where js is being embedded like this? Might consider moving to its own file.

website/templates/util/render_node.mako
alexschiller: Another thing is, the button you are adding will show up for read contributors and not-logged in users-- you probably need to hide the button

website/static/js/pages/component-settings-page.js
fabmiz: Done.

website/templates/util/render_node.mako
fabmiz: Yes, most project mako template include js

website/static/js/pages/component-settings-page.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+var $ = require('jquery');
+var bootbox = require('bootbox');
+var Raven = require('raven-js');
+var $osf = require('js/osfHelpers.js');
+var getConfirmationCode = require('js/projectSettings.js').getConfirmationCode;
```
alexschiller: This should be extracted from projectSettings and imported separately by the projectSettings file and component-settings-page. Console is throwing an error because the project overview page is now loading the whole projectSettings page and running JS that it doesn't have the correct data to execute.

website/static/js/projectSettings.js
```diff
@@ -152,7 +152,7 @@ request.fail(function(xhr, textStatus, err) {
  * Pulls a random name from the scientist list to use as confirmation string
  *  Ignores case and whitespace
  */
-var getConfirmationCode = function(nodeType, isPreprint) {
+var getConfirmationCode = function(nodeType, isPreprint, nodeApiUrl=nodeApiUrl) {
```
alexschiller: I don't believe the `nodeApiUrl=nodeApiUrl` is doing anything. Probably can be just `nodeApiUrl`

website/templates/util/render_node.mako
```diff
@@ -51,16 +51,26 @@
...
 
-            <!-- Show/Hide recent activity log -->
-            % if not summary['archiving']:
             <div class="pull-right">
                 % if not summary['primary'] and 'write' in user['permissions'] and not node['is_registration']:
                     <i class="fa fa-times remove-pointer" data-id="${summary['id']}" data-toggle="tooltip" title="Remove link"></i>
                     <i class="fa fa-code-fork" onclick="NodeActions.forkPointer('${summary['id']}', '${summary['primary_id']}');" data-toggle="tooltip" title="Fork this ${summary['node_type']} into ${node['node_type']} ${node['title']}"></i>
                 % endif
-                <i id="icon-${summary['id']}" class="pointer fa fa-angle-down" onclick="NodeActions.openCloseNode('${summary['id']}');" style="font-weight:bold;"></i>
+                % if summary['logged_in'] and summary['is_contributor']:
```
alexschiller: 👍

website/views.py
```diff
@@ -88,6 +89,11 @@ def serialize_node_summary(node, auth, primary=True, show_path=False):
             'api_url': node.api_url,
             'title': node.title,
             'category': node.category,
+            'isPreprint': bool(node.preprint_file and node.preprint_file._id),
+            'childExists': bool(node.nodes_active),
+            'is_admin': node.has_permission(user, permissions.ADMIN),
+            'is_contributor': node.is_contributor(user),
+            'logged_in': auth.logged_in,
```
alexschiller: This seems a bit out of place-- you can probably get this information from contextVars.currentUser.

website/templates/util/render_node.mako
```diff
@@ -51,16 +51,26 @@
             % endif
             </span>
 
-            <!-- Show/Hide recent activity log -->
-            % if not summary['archiving']:
```
alexschiller: I'm not sure this line should go away. Can you look into what this is doing more? I'm fairly certain certain actions are disabled if a node is mid archive job (when a node is registered).

website/templates/util/render_node.mako
```diff
@@ -51,16 +51,26 @@
...
             <div class="pull-right">
                 % if not summary['primary'] and 'write' in user['permissions'] and not node['is_registration']:
                     <i class="fa fa-times remove-pointer" data-id="${summary['id']}" data-toggle="tooltip" title="Remove link"></i>
                     <i class="fa fa-code-fork" onclick="NodeActions.forkPointer('${summary['id']}', '${summary['primary_id']}');" data-toggle="tooltip" title="Fork this ${summary['node_type']} into ${node['node_type']} ${node['title']}"></i>
                 % endif
-                <i id="icon-${summary['id']}" class="pointer fa fa-angle-down" onclick="NodeActions.openCloseNode('${summary['id']}');" style="font-weight:bold;"></i>
```
alexschiller: I just noticed this old line has an onclick function. Can you look into doing something similar to replace your class level onclick binding?

alexschiller: Also, if the code this is referencing is only used here, that code should be removed by your PR.

website/views.py
fabmiz: The data in contextVars.currentUser pertains to the project not its components, reason why I added these to the summary of a component. That way we can get the user's permissions on the component.

alexschiller: I only mean for logged_in, the others are fine here

alexschiller: Ah it is only used in the render_node mako, this is probably fine then.

fabmiz: Ok. Got it. Sorry about the confusion :)

website/static/js/projectSettings.js
fabmiz: Right, this is not needed. Forgot to clean that up.

website/static/js/pages/component-settings-page.js
fabmiz: Nice. Where do you suggest I add the new file? Can I just add it to osfHelpers? 
Funny enough, the code in projectSettings (getting 5 contributors for display in delete modal) that errors is not used at all though I understand the purpose. That request is not needed since we can pass component/project contributors to getConfirmationCode.

alexschiller: I'm not sure it is general enough to be added to helpers. A separate file similar to projectSettings.js or the helpers would probably work fine.

website/templates/util/render_node.mako
fabmiz: I'm probably wrong but it does not seem to be related to archiveJob. It seems to be for compatibility reasons. https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/models/node.py#L530

website/templates/util/render_node.mako
```diff
@@ -51,16 +51,30 @@
...
+                % if summary['logged_in'] and summary['is_contributor']:
```
alexschiller: Node links also get this button at the moment. My guess is that should be restricted. You may need to check with product on how they want node links handled. 
screen shot 2017-05-26 at 10 00 18 am

website/templates/util/render_node.mako
alexschiller: It is always false for non-registration nodes, but if it is a registration it can be true while it is archiving.

website/templates/util/render_node.mako
```diff
@@ -51,16 +51,30 @@
...
 
-            <!-- Show/Hide recent activity log -->
-            % if not summary['archiving']:
             <div class="pull-right">
                 % if not summary['primary'] and 'write' in user['permissions'] and not node['is_registration']:
                     <i class="fa fa-times remove-pointer" data-id="${summary['id']}" data-toggle="tooltip" title="Remove link"></i>
                     <i class="fa fa-code-fork" onclick="NodeActions.forkPointer('${summary['id']}', '${summary['primary_id']}');" data-toggle="tooltip" title="Fork this ${summary['node_type']} into ${node['node_type']} ${node['title']}"></i>
                 % endif
-                <i id="icon-${summary['id']}" class="pointer fa fa-angle-down" onclick="NodeActions.openCloseNode('${summary['id']}');" style="font-weight:bold;"></i>
+                % if summary['node_type'] == 'component' and summary['logged_in'] and summary['is_contributor']:
```
alexschiller: Speaking of Registrations (re archiving), you'll need to check with product whether they want these buttons to show up at all for registrations. At a minimum, the delete button should be removed as you cannot delete registrations and it fails when you try.

website/views.py
```diff
@@ -88,6 +89,11 @@ def serialize_node_summary(node, auth, primary=True, show_path=False):
             'api_url': node.api_url,
             'title': node.title,
             'category': node.category,
+            'isPreprint': bool(node.preprint_file and node.preprint_file._id),
```
alexschiller: https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/models/node.py#L445
I think you can just do `preprint_file_id` and not `preprint_file` and `preprint_file._id`. The last one is particularly undesirable as the guid (_id) is not on that table

PR #7253, opened 2017-05-22
[feature] Pytest Conversion: Registrations (linked nodes, forks, detail) [OSF-8010]

Nesiehr:
## Purpose

Increase in performance

## Changes

Pytest conversion of test_registration_detail, test_registration_forks, test_registration_linked_nodes
Grouping of tests within test_registration_detail, test_registration_linked_nodes

## Side effects

None known.

## Ticket

https://openscience.atlassian.net/browse/OSF-8010

## QA Considerations

None known.

## Test Validation

Change: api_tests/registrations/views/test_registration_linked_nodes.py
Modification: res = self.app.get(self.url, auth=self.user2.auth)
Modification Rational: The node was assigned to user, so user2 shouldn't get the same response code
Expected Result: A 400 error message instead of 200 Ok
Result: 403 Forbidden instead of 200 Ok
Test(s) Broken:
test_node_relationship_node_links

Change: api_tests/registrations/views/test_registration_detail.py
Modification: assert res.status_code == 404
Modification Rational: original status code was 200, so if the test is testing what it should be this change should break it and show that it is in fact testing what it is supposed to.
Expected Result: 404 != 200
Result: 200 != 404
Test(s) Broken:
test_registration_detail

Change: api_tests/registrations/views/test_registration_forks.py
Modification: assert len(res.json['data']) > 0
Modification Rational: The purpose of the test is to make sure no data is sent when unauthenticated, so by asserting that the response data is > 0 it should break and show that it is testing what it is supposed to.
Expected Result: 0 isn't greater than 0
Result: E   assert 0 > 0
Test(s) Broken:
test_can_access_public_registration_forks_list_when_unauthenticated


api_tests/registrations/views/test_registration_forks.py
```diff
@@ -19,269 +16,327 @@
...
+        assert res.json['data']['embeds']['node_links']['data'][1]['embeds']['target_node']['data']['id'] == pointer._id
+        assert res.json['data']['embeds']['node_links']['links']['meta']['total'] == 2
 
-    def test_cannot_fork_retractions(self):
+    def test_cannot_fork_retractions(self, app, user, private_registration, fork_data):
         with mock.patch('osf.models.AbstractNode.update_search'):
-            retraction = WithdrawnRegistrationFactory(registration=self.private_registration, user=self.user)
-        url = '/{}registrations/{}/forks/'.format(API_BASE, self.private_registration._id) + '?embed=forked_from'
+            retraction = WithdrawnRegistrationFactory(registration=private_registration, user=user)
+        url = '/{}registrations/{}/forks/'.format(API_BASE, private_registration._id) + '?embed=forked_from'
```
alexschiller: This should be either a format or a  `+` (probably format)

api_tests/registrations/views/test_registration_linked_nodes.py
```diff
@@ -1,395 +1,457 @@
...
+        assert private_node_two._id in nodes_returned
+        assert public_node._id in nodes_returned
 
-    def test_attempt_to_return_linked_nodes_logged_out(self):
-        res = self.app.get(
-            self.url, auth=None,
-            expect_errors=True
-        )
+    def test_attempt_to_return_linked_nodes_logged_out(self, app, url):
+        res = app.get(url, auth=None, expect_errors=True)
```
alexschiller: is there a difference between `auth=None` and leaving it off entirely?

api_tests/registrations/views/test_registration_forks.py
```diff
@@ -19,269 +16,327 @@
...
-        assert_equal(res.status_code, 201)
+        assert data['id'] == public_registration.forks.first()._id
+        assert data['attributes']['title'] == 'Fork of ' + public_registration.title
+        assert data['attributes']['registration'] == False
+        assert data['attributes']['fork'] == True
+
+    def test_cannot_fork_public_registration_logged_out(self, app, public_registration_url, fork_data):
+        res = app.post_json_api(public_registration_url, fork_data, expect_errors=True)
+        assert res.status_code == 401
+        assert res.json['errors'][0]['detail'] == 'Authentication credentials were not provided.'
```
sloria: Just driving by: You can use `rest_framework.exceptions.NotAuthenticated.default_detail`.

sloria: ^ might be outside the scope of this PR, though. This change can be made in a follow-up PR.

api_tests/registrations/views/test_registration_detail.py
```diff
@@ -1,297 +1,293 @@
...
+    def public_registration(self, user, public_project):
+        return RegistrationFactory(project=public_project, creator=user, is_public=True)
 
-class TestRegistrationDetail(ApiTestCase):
+    @pytest.fixture()
+    def private_registration(self, user, private_project):
+        return RegistrationFactory(project=private_project, creator=user)
 
-    def setUp(self):
-        self.maxDiff = None
```
Nesiehr: Nope, I looked for this and I couldn't find it. Definitely a bit troubling.

api_tests/registrations/views/test_registration_detail.py
```diff
@@ -1,297 +1,293 @@
...
+@pytest.mark.django_db
+class TestRegistrationDetail:
+
+    @pytest.fixture()
+    def public_project(self, user):
+        return ProjectFactory(title='Project One', is_public=True, creator=user)
+
+    @pytest.fixture()
+    def private_project(self, user):
+        return ProjectFactory(title='Project Two', is_public=False, creator=user)
```
Nesiehr: I believe you are right

api_tests/registrations/views/test_registration_detail.py
```diff
@@ -1,297 +1,293 @@
...
+    def unapproved_registration(self, registration_approval):
+        return Registration.find_one(Q('registration_approval', 'eq', registration_approval))
+
+    @pytest.fixture()
+    def unapproved_url(self, unapproved_registration):
+        return '/{}registrations/{}/'.format(API_BASE, unapproved_registration._id)
+
+    @pytest.fixture()
+    def public_project(self, user):
+        return ProjectFactory(title='Project One', is_public=True, creator=user)
```
Nesiehr: Nope

api_tests/registrations/views/test_registration_detail.py
```diff
@@ -1,297 +1,293 @@
...
+
+    @pytest.fixture()
+    def private_project(self, user):
+        return ProjectFactory(title='Project Two', is_public=False, creator=user)
```
Nesiehr: Nope

api_tests/registrations/views/test_registration_detail.py
```diff
@@ -1,297 +1,293 @@
...
+
+    #   test_id_field_must_match
+        mismatch_id_payload = make_payload(id='12345')
+
+        res = app.put_json_api(private_url, mismatch_id_payload, auth=user.auth, expect_errors=True)
+        assert res.status_code == 409
+
+    def test_turning_private_registrations_public(self, app, user, make_payload):
+        private_project = ProjectFactory(creator=user, is_public=False)
+        private_registration = RegistrationFactory(project=private_project, creator=user, is_public=False)
```
sloria: Keep it for now--there were some old registrations that were allowed to be made private, so we do want to support this functionality.

api_tests/registrations/views/test_registration_forks.py
```diff
@@ -19,269 +16,327 @@
...
         }
 
-        self.fork_data_with_title = {
+    @pytest.fixture()
+    def fork_data_with_title(self):
+        return {
             'data': {
                 'type': 'nodes',
                 'attributes':
                     {'title': 'My Forked Project'}
```
Nesiehr: Could you clarify here? The reason that they appear different seems to be tied to what they are used for in the tests.

api_tests/registrations/views/test_registration_detail.py
sloria: The title's aren't necessary but can help when debugging tests. Better to make them more descriptive, though--e.g. `Private Project`.

api_tests/registrations/views/test_registration_linked_nodes.py
sloria: Not a big deal either way, but it's ok to explictly pass `auth=None` here to communicate what the test is testing.

api_tests/registrations/views/test_registration_forks.py
Nesiehr: I like this change, I'll work this change into the conversions as they appear

api_tests/registrations/views/test_registration_forks.py
alexschiller: The above expands out all {} pairs and this one has a single line with {} wrapping title.

api_tests/registrations/views/test_registration_linked_nodes.py
alexschiller: I mostly was looking for consistency since there are other places in this file that don't put it explicitly.

api_tests/registrations/views/test_registration_detail.py
Nesiehr: So keep these around?

api_tests/registrations/views/test_registration_forks.py
Nesiehr: Got it 👍

api_tests/registrations/views/test_registration_forks.py
Nesiehr: I'll change the url strings to all use format

api_tests/registrations/views/test_registration_detail.py
alexschiller: @Nesiehr Yes, with more descriptive names per Steve's comment.

api_tests/registrations/views/test_registration_linked_nodes.py
sloria: Right; I think it's ok for this test since it specifically tests a "logged_out" request.

api_tests/registrations/views/test_registration_detail.py
```diff
@@ -1,297 +1,293 @@
...
 from urlparse import urlparse
-from nose.tools import *  # flake8: noqa
 from modularodm import Q
 
+from rest_framework import exceptions
 from api.base.settings.defaults import API_BASE
 from website.util import permissions
-
 from osf.models import Registration
 from tests.base import ApiTestCase
```
sloria: Although, when I tried removing the import, I got an error:

```
AttributeError: 'module' object has no attribute 'ADDONS_AVAILABLE'
```

This is because `init_app` is no longer being called. You can fix this by adding the following to `conftest.py`:

```python
...

# NOTE: autouse so that ADDONS_REQUESTED gets set on website.settings
@pytest.fixture(autouse=True, scope='session')
def app_init():
    init_app(routes=False, set_backends=False)
```

PR #7256, opened 2017-05-23
[hotfix] registration files id [OSF-8015]

alexschiller:
## Purpose
the id for a registration is wrong in certain places

## Changes
Uses _id instead of id; add a test, h/t @Nesiehr

## Side effects
Swapped two other places to _id (that currently worked) for consistency

## Ticket
https://openscience.atlassian.net/browse/OSF-8015

## QA Considerations
Test a registrations `https://api.osf.io/v2/registrations/b6psa/files/osfstorage/` route and make sure the guid is populated in the href (see ticket for an example)

### DIFF
--- a/api_tests/registrations/views/test_registration_files_list.py
+++ b/api_tests/registrations/views/test_registration_files_list.py
@@ -1,0 +1,35 @@
+import pytest
+
+from tests.json_api_test_app import JSONAPITestApp
+from api.base.settings.defaults import API_BASE
+from api_tests import utils as api_utils
+from tests.base import ApiTestCase
+from osf_tests.factories import (
+    AuthUserFactory,
+    ProjectFactory,
+    RegistrationFactory
+)
+
+@pytest.mark.django_db
+class TestRegistrationFilesList(object):
+
+    @pytest.fixture(autouse=True)
+    def setUp(self):
+        self.app = JSONAPITestApp()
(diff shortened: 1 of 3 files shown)


api_tests/registrations/views/test_registration_files_list.py
```diff
@@ -0,0 +1,35 @@
...
+    AuthUserFactory,
+    ProjectFactory,
+    RegistrationFactory
+)
+
+@pytest.mark.django_db
+class TestRegistrationFilesList(object):
+
+    @pytest.fixture(autouse=True)
+    def setUp(self):
```
sloria: Drive-by: you can use `setup_method` here: https://docs.pytest.org/en/latest/xunit_setup.html#method-and-function-level-setup-teardown

alexschiller: Per discussion, this throws a weird error that we haven't been able to diagnose yet: `Failed: Database access not allowed, use the "django_db" mark, or the "db" or "transactional_db" fixtures to enable it.`

sloria: Ah, looks like it's a known issue: https://github.com/pytest-dev/pytest-django/issues/297 , and that your workaround is correct: https://stackoverflow.com/questions/34089425/django-pytest-setup-method-database-issue

PR #7257, opened 2017-05-23
[feature] Pytest Conversion: Application (detail, list, reset) [OSF-8018]

Nesiehr:
## Purpose

Increase in performance

## Changes

Pytest conversion of test_application_detail.py, test_application_list.py, test_application_reset.py
Grouping of tests within test_application_detail.py
Added assert_dict_contains_subset to api_tests/utils.py

## Side effects

None known.

## Ticket

https://openscience.atlassian.net/browse/OSF-8018

## QA Considerations

None known.

## Test Validation

Change: api_tests/applications/test_application_detail.py
Modification: assert res.status_code == 404
Modification Rational: Shows that the test is testing the right thing by verifying that it returns the right 4** error message
Expected Result: returned 409 not 404
Result: E   assert 409 == 404
Test(s) Broken:
test_update_application_incorrect_type(self)

Change: api_tests/applications/test_application_list.py
Modification: assert len(res.json['data']) == len(self.user_one_apps) + len(self.user_two_apps)
Modification Rational: Shows that the test is testing the right thing by making sure a user can't see more than their number of apps.
Expected Result: # != # + #
Result: E   assert 3 == (3 + 2)
Test(s) Broken:
test_user_one_should_see_only_their_applications

Change: api_tests/applications/test_application_reset.py
Modification: assert res.status_code == 404
Modification Rational: Shows that the test is testing the right thing by verifying that it returns the right 4** error message
Expected Result: returned 403 not 404
Result: E   assert 403 == 404
Test(s) Broken:
test_other_user_cannot_reset


api_tests/applications/views/test_application_detail.py
```diff
@@ -1,33 +1,33 @@
...
-        self.user1 = AuthUserFactory()
-        self.user2 = AuthUserFactory()
-
-        self.user1_app = ApiOAuth2ApplicationFactory(owner=self.user1)
-        self.user1_app_url = _get_application_detail_route(self.user1_app)
+        self.app = JSONAPITestApp()
+        self.user_one = AuthUserFactory()
+        self.user_two = AuthUserFactory()
+        self.user_one_app = ApiOAuth2ApplicationFactory(owner=self.user_one)
+        self.user_one_app_url = _get_application_detail_route(self.user_one_app)
```
sloria: Drive-by comment: We should be moving away from the xunit style (setUp/tearDown) of tests and use pytest fixtures instead. This will allow individual tests to instantiate the minimum subset of fixtures that are needed.

sloria: Something along the lines of : 

```python
# in conftest.py

def client(...):
    return APITestApp(app)


class Test....:

    @pytest.fixture()
    def user(self):
        return AuthUserFactory()

    @pytest.fixture()
    def oauth_app(self, user):
        return ApiOAuth2ApplicationFactory(owner=user)

    @pytest.fixture()
    def oauth_app_url(self, oauth_app):
        return _get_application_detail_route(oauth_app)

    def test_non_owner_cant_view(self, client, oauth_app_url):
        non_owner = AuthUserFactory()

        res = client.get(oauth_app_url, auth=non_owner.auth, expect_errors=True)
        assert res.status_code == 403
````

File: api_tests/applications/views/test_application_detail.py [hunk 2]
```diff
@@ -1,35 +1,49 @@
...
+
+    @pytest.fixture()
+    def user_one(self):
+        return AuthUserFactory()
 
-class TestApplicationDetail(ApiTestCase):
-    def setUp(self):
-        super(TestApplicationDetail, self).setUp()
+    @pytest.fixture()
+    def user_two(self):
```
sloria: I would like to avoid creating a pattern of using "_one", "_two", etc. when multiple objects of a type are needed.

Instead, we can just instantiate objects in the test methods:

```python
class Test...:

    @pytest.fixture()
    def user(self):
        return AuthUserFactory()

    @pytest.fixture()
    def node(self, user):
        return NodeFactory(creator=user)

    def test_that_needs_another_user(self, node):
        non_owner = AuthUserFactory()
        res = self.app.get(..., auth=non_owner.auth)
```

This allows us to use more names that are more concise ("user") *and* more descriptive ("non_owner").

As an added bonus, this makes it much easier to refactor class-level fixture into module-level fixtures:

```python
@pytest.fixture()
def user():
    return AuthUserFactory()


class Test...:

    @pytest.fixture()
    def node(self, user):
        return NodeFactory(creator=user)

    def test_that_needs_another_user(self, node):
        non_owner = AuthUserFactory()
        res = self.app.get(..., auth=non_owner.auth)
```

api_tests/applications/views/test_application_detail.py
```diff
@@ -1,35 +1,49 @@
...
 def _get_application_list_url():
-    path = "applications/"
+    path = 'applications/'
     return api_v2_url(path, base_route='/')
 
+@pytest.mark.django_db
+class TestApplicationDetail:
+
+    @pytest.fixture()
+    def app(self):
```
sloria: This will be used in almost every test, so let's make this a "global" fixture (i.e. move it to `conftest.py`).

api_tests/applications/views/test_application_detail.py
```diff
@@ -39,19 +53,25 @@ def setUp(self):
                 }
             }
         }
+        return missing_id
 
-        self.missing_type = {
+    @pytest.fixture()
+    def missing_type(self, user_one_app):
```
sloria: There's a lot of duplication in the payload fixtures, and the names aren't always obvious. A useful pattern would be to create a "function fixture", i.e. a fixture that returns a function. Something like:

```python
class TestApplication...:

    @pytest.fixture()
    def make_payload(self, user_app):
        """Function that returns a valid application payload by default
        Can pass type and ID to modify the payload.
        """
        def func(type='applications', id=user_app.client_id):
            return {
                'data': {
                    'id': id,
                    'type': type,
                    'attributes': {
                        #...
                    }
                }
            }
        return func

    def test_update_application(self, make_payload, ...):
        # make_payload returns a valid payload by default
        valid = make_payload()
        # ...

        invalid_type = make_payload(type='invalid')
        # ...

        missing_type = make_payload()
        del missing_type['type']
        # ...
```

api_tests/applications/views/test_application_list.py
```diff
@@ -1,45 +1,64 @@
...
+    @pytest.fixture()
+    def app(self):
+        return JSONAPITestApp()
+
+    @pytest.fixture()
+    def user_one(self):
+        return AuthUserFactory()
+
+    @pytest.fixture()
+    def user_two(self):
```
sloria: See previous comment re: "_one", "_two", etc.

api_tests/applications/views/test_application_detail.py
```diff
@@ -1,35 +1,41 @@
...
 
+@pytest.fixture()
+def user():
+    return AuthUserFactory()
 
-class TestApplicationDetail(ApiTestCase):
-    def setUp(self):
-        super(TestApplicationDetail, self).setUp()
+@pytest.mark.django_db
+@pytest.mark.usefixtures('app')
```
sloria: I think this mark is unnecessary, since you're declaring `app` in the method signatures.

api_tests/applications/views/test_application_detail.py
```diff
@@ -1,35 +1,41 @@
...
 
-        self.user1_app = ApiOAuth2ApplicationFactory(owner=self.user1)
-        self.user1_app_url = _get_application_detail_route(self.user1_app)
+    @pytest.fixture()
+    def user_app_url(self, user_app):
+        return _get_application_detail_route(user_app)
 
-        self.missing_id = {
+    @pytest.fixture()
+    def missing_id(self):
```
sloria: I would still like to see https://github.com/CenterForOpenScience/osf.io/pull/7257#discussion_r118514062 implemented, unless you have any objections.

api_tests/applications/views/test_application_detail.py
```diff
@@ -86,131 +104,134 @@ def setUp(self):
...
-        res = self.app.put_json_api(self.user1_app_url, self.correct, auth=self.user1.auth, expect_errors=True)
-        assert_equal(res.status_code, 200)
+    def test_update_application(self, app, user, user_app, user_app_url, missing_id, missing_type, incorrect_id, incorrect_type, correct):
+        res = app.put_json_api(user_app_url, correct, auth=user.auth, expect_errors=True)
+        assert res.status_code == 200
 
-    def test_update_application_incorrect_type(self):
-        res = self.app.put_json_api(self.user1_app_url, self.incorrect_type, auth=self.user1.auth, expect_errors=True)
-        assert_equal(res.status_code, 409)
+    #   test_update_application_incorrect_type
```
sloria: Were these tests combined to make them faster? In general, it's better to have focused tests that test one thing. Do we get a lot of time savings if we combine these?

api_tests/applications/views/test_application_list.py
```diff
@@ -1,45 +1,47 @@
...
-        self.user1 = AuthUserFactory()
-        self.user2 = AuthUserFactory()
+@pytest.mark.django_db
+@pytest.mark.usefixtures('app')
+class TestApplicationList:
 
-        self.user1_apps = [ApiOAuth2ApplicationFactory(owner=self.user1) for i in xrange(3)]
-        self.user2_apps = [ApiOAuth2ApplicationFactory(owner=self.user2) for i in xrange(2)]
+    @pytest.fixture()
+    def user_apps(self, user):
```
sloria: Do we actually need to create multiple apps? I think it would make sense to just have a singular "user_app" fixture, then create more ApiOauth2Application objects in any test that requires more than one.

api_tests/applications/views/test_application_reset.py
```diff
@@ -1,29 +1,36 @@
...
 
-class TestApplicationReset(ApiTestCase):
-    def setUp(self):
-        super(TestApplicationReset, self).setUp()
+@pytest.fixture()
+def user():
+    return AuthUserFactory()
+
+@pytest.mark.django_db
+@pytest.mark.usefixtures('app')
```
sloria: See previous comment--can this be removed?

api_tests/utils.py
```diff
@@ -25,3 +25,6 @@ def urlparse_drop_netloc(url):
     if url[4]:
         return url[2] + '?' + url[4]
     return url[2]
+
+def assert_dict_contains_subset(a, b):
+    assert set(b.keys()) >= a.keys() and {k: b[k] for k in a if k in b} == a
```
sloria: A slightly more straightforward implementation was posted on StackOverflow: https://stackoverflow.com/questions/21058230/how-to-assert-a-dict-contains-another-dict-without-assertdictcontainssubset-in-p

api_tests/applications/views/test_application_detail.py
```diff
@@ -86,131 +104,134 @@ def setUp(self):
...
             }
         }
+        return correct
 
-    def test_owner_can_view(self):
-        res = self.app.get(self.user1_app_url, auth=self.user1.auth)
-        assert_equal(res.status_code, 200)
-        assert_equal(res.json['data']['attributes']['client_id'], self.user1_app.client_id)
+    def test_owner_can_view(self, app, user, user_app, user_app_url):
+        print user_app_url
```
sloria: Remove any debugging statements

api_tests/applications/views/test_application_list.py
```diff
@@ -1,45 +1,46 @@
...
-        self.user1 = AuthUserFactory()
-        self.user2 = AuthUserFactory()
+@pytest.mark.django_db
+class TestApplicationList:
 
-        self.user1_apps = [ApiOAuth2ApplicationFactory(owner=self.user1) for i in xrange(3)]
-        self.user2_apps = [ApiOAuth2ApplicationFactory(owner=self.user2) for i in xrange(2)]
+    @pytest.fixture()
+    def user_app(self, user):
+        return [ApiOAuth2ApplicationFactory(owner=user) for i in xrange(1)]
```
sloria: No need for this to be a list. It should just return the AO2A object.

api_tests/applications/views/test_application_list.py
```diff
@@ -48,69 +49,67 @@ def setUp(self):
...
+        res = app.get(user_list_url, auth=user.auth)
+        assert len(res.json['data']) == len(user_app)
 
-    def test_user2_should_see_only_their_applications(self):
-        res = self.app.get(self.user2_list_url, auth=self.user2.auth)
-        assert_equal(len(res.json['data']),
-                     len(self.user2_apps))
+    def test_other_user_should_see_only_their_applications(self, app, user_list_url):
+        other_user = AuthUserFactory()
+        other_user_app = [ApiOAuth2ApplicationFactory(owner=other_user) for i in xrange(2)]
```
sloria: nit: should be `other_user_apps`

api_tests/applications/views/test_application_list.py
```diff
@@ -1,45 +1,46 @@
...
-        self.user1_apps = [ApiOAuth2ApplicationFactory(owner=self.user1) for i in xrange(3)]
-        self.user2_apps = [ApiOAuth2ApplicationFactory(owner=self.user2) for i in xrange(2)]
+    @pytest.fixture()
+    def user_app(self, user):
+        return ApiOAuth2ApplicationFactory(owner=user)
 
-        self.user1_list_url = _get_application_list_url()
-        self.user2_list_url = _get_application_list_url()
+    @pytest.fixture()
+    def user_list_url(self):
```
sloria: Minor: could just be `url` since the test class name makes it clear we're testing the ApplicationList endpoint, but this is fine.

api_tests/applications/views/test_application_list.py
```diff
@@ -1,45 +1,46 @@
...
+        return _get_application_list_url()
 
-        self.sample_data = {
+    @pytest.fixture()
+    def sample_data(self):
+        sample_data = {
```
sloria: Minor: Can just `return` here.

api_tests/applications/views/test_application_list.py
```diff
@@ -48,69 +49,67 @@ def setUp(self):
...
         }
+        return sample_data
 
-    def test_user1_should_see_only_their_applications(self):
-        res = self.app.get(self.user1_list_url, auth=self.user1.auth)
-        assert_equal(len(res.json['data']),
-                     len(self.user1_apps))
+    def test_user_should_see_only_their_applications(self, app, user, user_app, user_list_url):
+        res = app.get(user_list_url, auth=user.auth)
+        assert len(res.json['data']) == ApiOAuth2Application.objects.count()
```
sloria: This assertion isn't consistent with the test method name. The length of the returned data should be equal to the length of the user's application list, not the count of all applications.

api_tests/applications/views/test_application_list.py
```diff
@@ -49,68 +50,65 @@ def setUp(self):
             }
         }
 
-    def test_user1_should_see_only_their_applications(self):
-        res = self.app.get(self.user1_list_url, auth=self.user1.auth)
-        assert_equal(len(res.json['data']),
-                     len(self.user1_apps))
+    def test_user_should_see_only_their_applications(self, app, user, user_app, url):
+        res = app.get(url, auth=user.auth)
+        assert len(res.json['data']) == ApiOAuth2Application.objects.count()
```
sloria: Ah, my previous comment got lost, so i'll repost: This assertion is inconsistent with the method name. The length of the returned data should be equal to the number of applications owned by `user`, not the count of *all* applications.

api_tests/utils.py
```diff
@@ -25,3 +25,6 @@ def urlparse_drop_netloc(url):
     if url[4]:
         return url[2] + '?' + url[4]
     return url[2]
+
+def assert_dict_contains_subset(a, b):
```
sloria: This should probably go in `tests.base`, since there's nothing api-specific about it.

sloria:
Looks good! The test failure is unrelated to these changes.

PR #7258, opened 2017-05-24
Proof of concept: Refactoring UserListView an NodeListView to use django-filter [OSF-7962] [OSF-8090]

erinspace:
## Note:
This is no longer an actual PR that would be merged, rather a proof of concept. Main focus is document linked on below JIRA ticket!

## Purpose

Investigate using the `DjangoFilterBackend` for APIv2. This is just a proof-of-concept, not necessarily for merging, but just for reviewing and seeing what sorts of modifications using the 3rd party `DjangoFilterBackend` could entail.

## Changes

- Use `DjangoFilterBackend` for the `UserList` view and `NodeList` views
- Create `JSONAPIFilterSet` using some of the same filter param logic to break down filter syntax
- Custom filter properties that will filter with OR queries for two different fields when using filter syntax such as `/?filter[given_name,family_name]=Doe`
- Also add OR functionality with one field, such as `/?filter[family_name]=Jane,Dog`
- `UserFilterSet` and `NodeFilterSet` that facilitate filtering on node and user fields, consistent with what we have in the API currently

## Side effects

No real ones at the moment,  since it's only a proof of concept!

## Other thoughts 
A lot of messing around with standard functionality seemed to be needed to be done for our filter syntax to work properly with variable OR filters (see changes above). I'm not sure how well the built in solutions will work given the modifications we've done to get these things to work, but it could be alright! Just not sure if we'd need to alter the basics so much that we lose the benefits of using a maintained 3rd party library.

## Ticket
https://openscience.atlassian.net/browse/OSF-7962
https://openscience.atlassian.net/browse/OSF-8090

### DIFF
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -3,8 +3,6 @@ import functools
 import operator
 import re
 
-from django.utils import six
-from django import forms
 import django_filters
 from django.db import models
 import pytz
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -90,6 +90,7 @@ class JSONAPIFilterSet(django_filters.FilterSet):
 
     def __init__(self, data=None, *args, **kwargs):
         self.or_fields = {}
+        field_names = []
         if data:
             new_data = {}
             for key, value in data.iteritems():
@@ -105,6 +106,10 @@ class JSONAPIFilterSet(django_filters.FilterSet):
                         new_data.update({field: value})
             data = new_data
         super(JSONAPIFilterSet, self).__init__(data=data, *args, **kwargs)
+
+        for field in field_names:
+            if field not in self.form.fields.keys():
+                raise InvalidFilterFieldError(parameter='filter', value=field)
 
     @property
     def qs(self, *args, **kwargs):
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -2,7 +2,6 @@ import datetime
 import functools
 import operator
 import re
-from operator import or_
 
 from django.utils import six
 from django import forms
@@ -77,7 +76,7 @@ class NewDjangoFilterMixin(django_filters.FilterSet):
     QUERY_PATTERN = re.compile(r'^filter\[(?P<fields>((?:,*\s*\w+)*))\](\[(?P<op>\w+)\])?$')
     FILTER_FIELDS = re.compile(r'(?:,*\s*(\w+)+)')
 
-    or_fields = {}
+    or_fields = []
 
     def __init__(self, data=None, *args, **kwargs):
         if data:
@@ -88,8 +87,14 @@ class NewDjangoFilterMixin(django_filters.FilterSet):
                     match_dict = match.groupdict()
                     fields = match_dict['fields']
                     field_names = re.findall(self.FILTER_FIELDS, fields.strip())
-                    if len(field_names) > 1:
-                        self.or_fields[frozenset(field_names)] = value
+
+                    self.or_fields = []
+                    multiple_values = value.split(',')
+                    if len(multiple_values) > 1 or len(field_names) > 1:
+                        for field_name in field_names:
+                            for value in multiple_values:
+                                self.or_fields.append({'field': field_name, 'value': value})
+
                     for field in field_names:
                         new_qd.update({field: value})
             data = new_qd
@@ -115,7 +120,8 @@ class NewDjangoFilterMixin(django_filters.FilterSet):
             qs = self.queryset.all()
             for name, filter_ in six.iteritems(self.filters):
                 value = self.form.cleaned_data.get(name)
-                if name in reduce(or_, self.or_fields.keys(), set()):
+                or_keys = [entry['field'] for entry in self.or_fields]
+                if name in or_keys:
                     qs = self.filter_groups(qs)
                 else:
                     qs = filter_.filter(qs, value)
@@ -124,13 +130,10 @@ class NewDjangoFilterMixin(django_filters.FilterSet):
             return self._qs
 
     def filter_groups(self, qs):
-        for group, value in self.or_fields.iteritems():
-            group_q = Q()
-            for field in group:
-                if field in self.form.fields.keys():
-                    group_q |= Q(**{field: value})
-
-            qs = qs.filter(group_q)
+        group_q = Q()
+        for field_dict in self.or_fields:
+            group_q |= Q(**{field_dict['field']: field_dict['value']})
+        qs = qs.filter(group_q)
 
         return qs
 
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -86,8 +86,6 @@ class JSONAPIFilterSet(django_filters.FilterSet):
     QUERY_PATTERN = re.compile(r'^filter\[(?P<fields>((?:,*\s*\w+)*))\](\[(?P<op>\w+)\])?$')
     FILTER_FIELDS = re.compile(r'(?:,*\s*(\w+)+)')
 
-    or_fields = {}
-
     def __init__(self, data=None, *args, **kwargs):
         self.or_fields = {}
         field_names = []
@@ -98,12 +96,12 @@ class JSONAPIFilterSet(django_filters.FilterSet):
                 if match:
                     match_dict = match.groupdict()
                     fields = match_dict['fields']
-                    field_names = re.findall(self.FILTER_FIELDS, fields.strip())
+                    field_names = self.FILTER_FIELDS.findall(fields.strip())
                     values = value.split(',')
                     if len(field_names) > 1:
                         self.or_fields[frozenset(field_names)] = values
-                    for field in field_names:
-                        new_data.update({field: value})
+                    else:
+                        new_data.update({field_names[0]: value})
             data = new_data
         super(JSONAPIFilterSet, self).__init__(data=data, *args, **kwargs)
 
@@ -114,9 +112,6 @@ class JSONAPIFilterSet(django_filters.FilterSet):
     @property
     def qs(self, *args, **kwargs):
         self.form.is_valid()
-        or_keys = reduce(operator.or_, self.or_fields.keys(), set())
-        for key in or_keys:
-            self.form.cleaned_data.pop(key, None)
         qs = super(JSONAPIFilterSet, self).qs
         return self.filter_groups(qs)
 
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -144,6 +144,8 @@ class JSONAPIFilterSet(django_filters.rest_framework.FilterSet):
         for field_name, value in data.iteritems():
             # to filter on a relationship field, values must be in a list
             if field_name in self.MANY_TO_MANY_FIELDS and value != 'null':
+                if field_name == 'contributors':
+                    field_name = '_contributors'
                 value = value if type(value) == list else value.split(',')
             if value == 'true' or value == 'false':
                 value = value.title()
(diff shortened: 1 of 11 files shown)


api/base/filters.py
```diff
@@ -61,6 +66,83 @@ def filter_queryset(self, request, queryset, view):
...
+                    match_dict = match.groupdict()
+                    fields = match_dict['fields']
+                    field_names = re.findall(self.FILTER_FIELDS, fields.strip())
+
+                    self.or_fields = []
+                    multiple_values = value.split(',')
+                    if len(multiple_values) > 1 or len(field_names) > 1:
+                        for field_name in field_names:
+                            for value in multiple_values:
+                                self.or_fields.append({'field': field_name, 'value': value})
```
sloria: You'll need to handle multiple groups. Maybe it makes sense to use a dictionary, where the keys are sets of field names, and the value is the value.

```python
{
    set('given_name', 'family_name'): 'Erin',
    ...
}
```

api/users/views.py
```diff
@@ -63,7 +64,18 @@ def get_user(self, check_permissions=True):
...
-class UserList(JSONAPIBaseView, generics.ListAPIView, ODMFilterMixin):
+class UserFilter(NewDjangoFilterMixin):
+
+    full_name = django_filters.CharFilter(name='fullname', lookup_expr='icontains')
+    id = django_filters.CharFilter(name='guids___id')
+
+    class Meta:
+        model = OSFUser
+        fields = ['id', 'full_name', 'given_name', 'middle_names', 'family_name']
+        strict = django_filters.constants.STRICTNESS.RAISE_VALIDATION_ERROR
```
sloria: Would it make sense to define `strict` on the base class (currently NewDjangoFilterMixin)?

api/users/views.py
```diff
@@ -138,10 +153,10 @@ def get_default_odm_query(self):
 
     # overrides ListCreateAPIView
     def get_queryset(self):
-        # TODO: sort
-        query = self.get_query_from_request()
-        return User.find(query)
-
+        queryset = OSFUser.objects.filter(is_registered=True, date_disabled=None).order_by('-date_registered')
+        if self.request.version >= '2.3':
+            queryset = queryset.filter(merged_by__isnull=True)
```
sloria: Why was this added?

api/users/views.py
sloria: Also, is there a way for us to raise the same error that we're doing now if the client passes an invalid field name? https://github.com/CenterForOpenScience/osf.io/blob/7e41c06588a5d1e370f2ac2a4b1b6e57e2a6c159/api/base/filters.py#L116

api_tests/users/views/test_user_list.py
```diff
@@ -261,6 +261,27 @@ def test_users_list_filter_multiple_fields_with_bad_filter(self):
         res = self.app.get(url, expect_errors=True)
         assert_equal(res.status_code, 400)
 
+    def test_users_list_or_filter(self):
+        self.john_doe = UserFactory(fullname='John Doe')
```
sloria: Assigning to `self` is unnecessary.

api/base/filters.py
```diff
@@ -61,6 +66,83 @@ def filter_queryset(self, request, queryset, view):
...
+            for key, value in data.iteritems():
+                match = self.QUERY_PATTERN.match(key)
+                if match:
+                    match_dict = match.groupdict()
+                    fields = match_dict['fields']
+                    field_names = re.findall(self.FILTER_FIELDS, fields.strip())
+
+                    self.or_fields = []
+                    multiple_values = value.split(',')
+                    if len(multiple_values) > 1 or len(field_names) > 1:
```
erinspace: the len of `multiple_values` is checked here cause if there are more than one, those also need to be sent to the or_fields list for a built `OR` query, independent if there are multiple field names

sloria: Ah, I think I see. But don't filters with multiple values need to be handled separately from filters with multiple field names anyway? Also, doesn't django-filter have a built in way to handle multiple values? Intuitively, `?full_name=Erin,Steve` should be translated to `.filter(full_name__in=['Erin', 'Steve'])`; I wonder if that's supported out-of-the-box.

erinspace: I had a whole lot of trouble getting this to work out of the box -- theoretically, from reading through a few issues and their resolutions, the only mention I can find on how to handle those natively was through the [CSV Widget] -- http://django-filter.readthedocs.io/en/develop/ref/widgets.html#csv-widget -- though applying that in my testing of things didn't get me anywhere for a long while so I decided to just deal with it while dealing with other `OR` cases at the same time. I can dive back in tho and really try to figure out what's happening....

sloria: Cool cool, looks like it might not be supported OOB. [MultipleChoiceFilter](https://github.com/carltongibson/django-filter/blob/7e949076a5dec69811d081fd78838999c8abf374/django_filters/filters.py#L229) does something similar to what we want, except without a constrained set of "choices".  So maybe we can at least steal some code from that class.

sloria: Also came across this issue, which might be worth looking into: https://github.com/carltongibson/django-filter/issues/727

api/users/views.py
erinspace: This was added for [this test here](https://github.com/CenterForOpenScience/osf.io/blob/develop/api_tests/users/views/test_user_list.py#L60) that checks for version 2.3. Since django-filter is based on the view's `get_queryset` method (rather than our filter's relying on `get_default_odm_query`) -- I added the same checks on `get_queryset` as [already existed](https://github.com/CenterForOpenScience/osf.io/blob/develop/api/users/views.py#L135) on the odm query method

sloria: Ah, ok. Thanks for the clarification.

api/base/filters.py
```diff
@@ -61,6 +63,78 @@ def filter_queryset(self, request, queryset, view):
...
+        for field in field_names:
+            if field not in self.form.fields.keys():
+                raise InvalidFilterFieldError(parameter='filter', value=field)
+
+    @property
+    def qs(self, *args, **kwargs):
+        self.form.is_valid()
+        or_keys = reduce(operator.or_, self.or_fields.keys(), set())
+        for key in or_keys:
+            self.form.cleaned_data.pop(key, None)
```
sloria: Rather than removing these keys from `form.cleaned_data`, would it make sense to prevent these fields from going through the form validation in the first place, i.e. preventing all the or_fields from going into `self.form.fields`?

api/base/filters.py
```diff
@@ -61,6 +63,78 @@ def filter_queryset(self, request, queryset, view):
...
+                    field_names = re.findall(self.FILTER_FIELDS, fields.strip())
+                    values = value.split(',')
+                    if len(field_names) > 1:
+                        self.or_fields[frozenset(field_names)] = values
+                    for field in field_names:
+                        new_data.update({field: value})
+            data = new_data
+        super(JSONAPIFilterSet, self).__init__(data=data, *args, **kwargs)
+
+        for field in field_names:
```
sloria: Actually, is this code necessary, since this check is happening in `filter_groups`?

api/base/filters.py
```diff
@@ -61,6 +63,78 @@ def filter_queryset(self, request, queryset, view):
...
+            name = self.name
+            q |= Q(**{'{}__icontains'.format(name): value})
+        qs = qs.filter(q)
+
+        return qs
+
+
+class JSONAPIFilterSet(django_filters.FilterSet):
+
+    QUERY_PATTERN = re.compile(r'^filter\[(?P<fields>((?:,*\s*\w+)*))\](\[(?P<op>\w+)\])?$')
```
erinspace: Not yet! This was used when parsing out the other operators used in other filters, such as `eq`, `ne`, etc, which aren't used in this endpoint that I picked. Perhaps this proof of concept should include porting a second endpoint that uses operators?

api/base/filters.py
erinspace: I at first tried to do this check inside of the loop itself, but the only way I could access the fields that were supposed to be there was to access the form fields -- which is inaccessible until `super(JSONAPIFilterSet, self).__init__(...)` is called -- since `self.form.fields` isn't set until `self.filters` is set, which happens in the `super`'s init. 

The check is necessary here (or somewhere, at least), as the check in `filter_groups` only checks for invalid fields that are included in a group that will be part of an OR query (eg `?filter[wrong,field]`) and not filters that will go through normal filtering later (eg `?filter[wrong_field]=whateves`)

There might be a way to instead deal with this using [django-filter's strict validation](http://django-filter.readthedocs.io/en/develop/ref/filterset.html#strict) -- I tried for quite a long time to get this to work, and concluded that it was checking for an invalid filter **value** rather than an invalid filter itself -- which gets overlooked as far as I can tell by django-filter by default.

api/users/views.py
```diff
@@ -70,14 +71,25 @@ def get_user(self, check_permissions=True):
...
-        obj = get_object_or_error(User, key, 'user')
+        obj = get_object_or_error(OSFUser, key, 'user')
         if check_permissions:
             # May raise a permission denied
             self.check_object_permissions(self.request, obj)
         return obj
 
 
-class UserList(JSONAPIBaseView, generics.ListAPIView, ODMFilterMixin):
+class UserFilterSet(JSONAPIFilterSet):
```
sloria: `FilterSets` and filter field classes should probably be separated into a separate file. Let's move this to `api/users/filters.py`.

api/users/views.py
```diff
@@ -29,7 +30,7 @@
 from rest_framework import permissions as drf_permissions
 from rest_framework import generics
 from rest_framework.exceptions import NotAuthenticated, NotFound
-from osf.models import Contributor, ExternalAccount, AbstractNode as Node, PreprintService, OSFUser as User
+from osf.models import Contributor, ExternalAccount, AbstractNode as Node, PreprintService, OSFUser
```
sloria: 👍

api/base/filters.py
```diff
@@ -61,6 +64,163 @@ def filter_queryset(self, request, queryset, view):
...
+    def filter(self, qs, value):
+        q = Q()
+        values = value or []
+        for value in values:
+            q |= Q(**{'{}__icontains'.format(self.name): value.strip()})
+        qs = qs.filter(q)
+
+        return qs
+
+class NullModelMultipleChoiceCaseInsensitiveField(forms.ModelMultipleChoiceField):
```
mfraezz: Minor: I think the Django convention is to define custom fields in a `fields.py` file, but we haven't yet done that anywhere else in the API ¯\\\_(ツ)_/¯

api/base/filters.py
```diff
@@ -61,6 +64,163 @@ def filter_queryset(self, request, queryset, view):
...
+            q |= Q(**{self.name: value})
+        return qs.filter(q)
+
+
+class JSONAPIFilterSet(django_filters.rest_framework.FilterSet):
+
+    QUERY_PATTERN = re.compile(r'^filter\[(?P<fields>((?:,*\s*\w+)*))\](\[(?P<op>\w+)\])?$')
+    FILTER_FIELDS = re.compile(r'(?:,*\s*(\w+)+)')
+
+    # TODO - there must be a better way to recognize these
```
mfraezz: You can do something like `if <class>._meta.get_field('<name>').many_to_many` to check for M2M. 
For date...
```
from django.db import connection

if 'timestamp' in <class>._meta.get_field('<name>').db_type(connection)
```
Not sure how hard it is to get the class in scope here, though

erinspace: Ah, that's good to know! I tried to get the class in scope here, but it's really tough, as far as I can tell my only args here are the field name and value being passed in by the filter....

sloria:
Closing. See https://openscience.atlassian.net/browse/PLAT-492 for our analysis

PR #7259, opened 2017-05-25
Turn ensure_licenses and ensure_schemas into django migrations [OSF-7980]

erinspace:
## Purpose

Every time the app is run, and several times during tests, `ensure_licenses` and `ensure_schemas` are called - let's make those django migrations so that we don't have to do those every time!

## Changes

- Add `ensure_licenses` migration
- Add `ensure_schemas` migration
- Add `osf/utils/migrations.py` with the functions to be accessed outside the migrations id needbe
- Remove most usages of those two functions from the app
- Remove most usages of those two functions from the tests
- Remove usages from old models as well

## Side effects
While the migration should also take care of updating licenses and schemas with new information, this will not automatically happen anymore whenever the app is run - so therefore would have to be run manually. It would have to happen by manually running the `ensure_schemas` and `ensure_licenses` functions from `osf/utils/migrations.py`

Current side effect: if the test `api_tests/metaschemas/views/test_metaschemas_detail.py::TestMetaSchemaDetail::test_inactive_metaschema_returned` fails locally with a NotFound error,  its because the migration was run in your tests before a new metaschema was added with new active information. Temporarially remove `--reuse-db` from `pytest.ini` to re-run migrations for tests and fix the problem.

## Ticket
https://openscience.atlassian.net/browse/OSF-7980


osf/utils/migrations.py
```diff
@@ -0,0 +1,116 @@
...
+            logger.info('License {name} ({id}) added to the database.'.format(name=name, id=id))
+
+    logger.info('{} licenses inserted into the database, {} licenses updated in the database.'.format(
+        ninserted, nupdated
+    ))
+
+    return ninserted, nupdated
+
+
+def remove_licenses(*args):
```
sloria: I'm thinking this function (and `remove_schemas`) should just remove all license/metaschema records. If we only remove the ones defined in files, there'd be no way to remove a license/metaschema in a migration.

erinspace: Done!

osf/utils/migrations.py
```diff
@@ -0,0 +1,88 @@
...
+    pre_count = NodeLicense.objects.all().count()
+    NodeLicense.objects.all().delete()
+
+    logger.info('{} licenses removed from the database.'.format(pre_count))
+
+
+def ensure_schemas(*args):
```
sloria: This differs slightly from the implementation on the latest develop, specifically in the use of `get_or_create` vs. `update_or_create`.

On `develop`:

```python
def ensure_schema(schema, name, version=1, active=True):
    MetaSchema = apps.get_model('osf.MetaSchema')
    MetaSchema.objects.update_or_create(
        name=name,
        schema_version=version,
        defaults={
            'schema': schema,
            'active': active
        }
    )

def ensure_schemas():
    """Import meta-data schemas from JSON to database if not already loaded
    """
    for schema in OSF_META_SCHEMAS:
        ensure_schema(schema, schema['name'], version=schema.get('version', 1), active=schema.get('active', True))
```

Is there a reason to change this?

website/app.py
```diff
@@ -128,9 +126,6 @@ def init_app(settings_module='website.settings', set_backends=True, routes=True,
         sentry.init_app(app)
         logger.info("Sentry enabled; Flask's debug mode disabled")
 
-    if set_backends and fixtures:
```
sloria: 🎉

sloria: Now that `fixtures` is unused, it can be removed from the function signature. Be sure to remove any usages of that param, e.g. in manage.py.

osf/utils/migrations.py
erinspace: Ah, no reason! Missed that update somehow, and using that makes a lot more sense.

website/app.py
erinspace: Done!

erinspace: Done!

osf/utils/migrations.py
```diff
@@ -63,7 +63,7 @@ def ensure_schemas(*args):
     """
     schema_count = 0
     for schema in OSF_META_SCHEMAS:
-        schema_obj, created = MetaSchema.objects.get_or_create(
+        schema_obj, created = MetaSchema.objects.update_or_create(
```
sloria: Minor: With this change, L74 and L75 are unnecessary.

osf/migrations/0038_ensure_schemas.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('osf', '0037_ensure_licenses'),
+    ]
+
+    operations = [
+        migrations.RunPython(ensure_schemas, remove_schemas),
```
sloria: In the future, we can probably put ensure_schemas and ensure_licenses in a single migration, but this is fine for now

sloria:
Merged in `feature/next-djangole`.

PR #7262, opened 2017-05-25
[OSF-8057] OwnCloud incorrectly logs folder deletions as "file" in logs

Johnetordoff:
## Purpose

OSF wrongly logged every item deleted as a file (even if it's a folder)

## Changes

The osf infers whether something is a folder by checking if it has a trailing slash, but our logging function was stripping the slashes, presumably for aesthetic reasons. This fix stops that. 

## Side effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/OSF-8057


Johnetordoff:
Unfortunately I don't think so, the path is the only indication the logger gets to determine whether the item is a file or a folder. The only possibility would be if OwnCould also kept logs and you probably know more about that then I do.

mfraezz:
I'm not sure that this is the correct approach:
```

In [11]: [{'folder': l.params['folder'], 'path': l.params['path']} for l in n.logs.filter(action='owncloud_file_removed')]
Out[11]:
[{'folder': u'/Globally Shared/New Folder/', 'path': u'empty'},
 {'folder': u'/Globally Shared/New Folder/', 'path': u'new folder'}]
```
`n` is the node from the ticket.

Johnetordoff:
I'm not sure I understand the objection, it does say the action is "owncloud_file_removed" when it's actually a folder being removed, but that's true for every provider. maybe it would be best to refactor all the log maps to say "{provder}_item_removed"?

mfraezz:
I thought I'd found a way: we only keep information about trashed files for third party addons, not folders. That is to say, if `TrashedFileNode.objects.filter(provider='owncloud', node=<log.node>, name=<log.params['path']>.split('/')[0]).exists()` is `True`, you know it's a file. Unfortunately, there are some false negatives, making that method unusable.

However, looking at the actual results, checking for file extension existence with `bool('.' in log.params['path'] and len(log.params['path'].split('.')[-1]) <= 5)` is 100% accurate at distinguishing.

addons/owncloud/models.py
```diff
@@ -129,7 +129,7 @@ def create_waterbutler_log(self, auth, action, metadata):
                 'project': self.owner.parent_id,
                 'node': self.owner._id,
                 'folder': self.folder_id,
-                'path': metadata['materialized'].strip('/'),
+                'path': metadata['materialized'],
```
mfraezz: Minor: This should probably still `lstrip('/')`, if the original intent was aesthetics.

Johnetordoff:
OwnCloud can do extension-less file names, so I'm not you can know for sure.

mfraezz:
> OwnCloud can do extension-less file names, so I'm not you can know for sure

I agree -- theoretically, you can't. However, there aren't very many of these logs, and currently they all behave as expected.

PR #7277, opened 2017-05-31
[Feature] Prevent redraw grey screen when add component is clicked quickly [OSF-8053]

fabmiz:
## Purpose
Prevent grey screen (modal backdrop) which occurs when the addComponent is clicked before the 
treebeard/fangorn filebrowser is fully added to the DOM.

## Changes

Modifies the order in which Mithril draws mounted components by using
m.startComputation/m.endComputation to only allow one redraw. As a result,
the addCompnent button is not clickable till filebrowser is fully drawn.

## Ticket

https://openscience.atlassian.net/browse/OSF-8053

### DIFF
--- a/website/static/js/pages/project-dashboard-page.js
+++ b/website/static/js/pages/project-dashboard-page.js
@@ -121,16 +121,15 @@ $(document).ready(function () {
         contributors: window.contextVars.node.contributors,
         currentUserCanEdit: window.contextVars.currentUser.canEdit
     });
-    var newComponentElem = document.getElementById('newComponent');
-
-    m.startComputation();
-
-    if (ctx.node.institutions.length && !ctx.node.anonymous && !ctx.node.isRetracted) {
-        m.mount(document.getElementById('instLogo'), m.component(institutionLogos, {institutions: window.contextVars.node.institutions}));
-    }
-    $('#contributorsList').osfToggleHeight();
 
     if (!ctx.node.isRetracted) {
+        m.startComputation();
+
+        if (ctx.node.institutions.length && !ctx.node.anonymous) {
+            m.mount(document.getElementById('instLogo'), m.component(institutionLogos, {institutions: window.contextVars.node.institutions}));
+        }
+        $('#contributorsList').osfToggleHeight();
+
         // Recent Activity widget
         m.mount(document.getElementById('logFeed'), m.component(LogFeed.LogFeed, {node: node}));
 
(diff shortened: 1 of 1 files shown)


website/static/js/pages/project-dashboard-page.js
```diff
@@ -201,7 +200,16 @@ $(document).ready(function () {
                 }
             };
             var filebrowser = new Fangorn(fangornOpts);
+            if (newComponentElem) {
+                m.mount(newComponentElem, AddComponentButton);
+            };
+            m.endComputation();
         });
+    } else {
```
alexschiller: This else (and the mount/endcomp after) isn't needed as it only occurs when the node is retracted, in which case you don't need to have an addComponent button.

website/static/js/pages/project-dashboard-page.js
```diff
@@ -122,9 +122,8 @@ $(document).ready(function () {
         currentUserCanEdit: window.contextVars.currentUser.canEdit
     });
     var newComponentElem = document.getElementById('newComponent');
-    if (newComponentElem) {
-        m.mount(newComponentElem, AddComponentButton);
-    }
+
+    m.startComputation();
 
     if (ctx.node.institutions.length && !ctx.node.anonymous && !ctx.node.isRetracted) {
```
alexschiller: I think this mount can be moved inside the `if (!ctx.node.isRetracted) {` block (and remove the isRetracted check)

website/static/js/pages/project-dashboard-page.js
```diff
@@ -122,9 +122,8 @@ $(document).ready(function () {
         currentUserCanEdit: window.contextVars.currentUser.canEdit
     });
     var newComponentElem = document.getElementById('newComponent');
-    if (newComponentElem) {
-        m.mount(newComponentElem, AddComponentButton);
-    }
+
+    m.startComputation();
```
alexschiller: With the institutions mount moved inside the isRetracted block, the start Computation can also be moved inside the isRetracted block (see next comment)

website/static/js/pages/project-dashboard-page.js
```diff
@@ -122,9 +122,8 @@ $(document).ready(function () {
         currentUserCanEdit: window.contextVars.currentUser.canEdit
     });
     var newComponentElem = document.getElementById('newComponent');
```
alexschiller: This can also be moved down (see next two comments)

website/static/js/pages/project-dashboard-page.js
```diff
@@ -201,7 +199,18 @@ $(document).ready(function () {
                 }
             };
             var filebrowser = new Fangorn(fangornOpts);
+            var newComponentElem = document.getElementById('newComponent');
+            if (window.contextVars.node.isPublic) {
+                m.mount(document.getElementById('shareButtonsPopover'),
+                        m.component(SocialShare.ShareButtonsPopover,
```
alexschiller: Over indented

website/static/js/pages/project-dashboard-page.js
```diff
@@ -201,7 +199,18 @@ $(document).ready(function () {
...
+                            {title: window.contextVars.node.title, url: window.location.href}));
+            }
+            if (newComponentElem) {
+                m.mount(newComponentElem, AddComponentButton);
+            };
```
alexschiller: jshint ;

PR #7282, opened 2017-05-31
[feature] Email Contributors Added to Forks and Templates [OSF-7699]

caseyrollins:
#### Purpose
- Send emails to contributors when they are added to a project that was created by forking or templating. 

#### Changes
- Send the `contributor_added` signal in `fork_node()` and `use_as_template()` functions (in order to subscribe the user that created the project to notifications) but with `email_template=false` so that the contributor added email is not sent when a user forks/templates a project. 
- Remove the check in `notify_contributor` that prevented contributor emails from being sent for nodes where `is_fork` or `template_node` was truthy. 

#### Side Effects
- None expected. 

#### QA 
- Fun fact: from what I can tell, this functionality has never worked and we've never sent emails to contributors that were added to forked/templated nodes. Both should work now.
- Additionally, if a user templated a project, that user was not subscribed to notifications for the new templated project. This should work now (it wasn't broken for forks).

#### Ticket
- [OSF-7699](https://openscience.atlassian.net/browse/OSF-7699)


caseyrollins:
Added 🦄

PR #7286, opened 2017-06-01
[feature] Prevent Unconfirmed User Contributor Additions (API) [OSF-7975]

caseyrollins:
#### Purpose
- Prevent unconfirmed users from being added as contributors via the API.

#### Changes
- Check `user.is_registered` in `node.add_contributors_registered_or_not`

#### Side Effects
- None expected

#### Ticket
- [OSF-7975](https://openscience.atlassian.net/browse/OSF-7975)

### DIFF
--- a/osf/models/node.py
+++ b/osf/models/node.py
@@ -1240,6 +1240,11 @@ class AbstractNode(DirtyFieldsMixin, TypedModel, AddonModelMixin, IdentifierMixi
             contributor = OSFUser.load(user_id)
             if not contributor:
                 raise ValueError('User with id {} was not found.'.format(user_id))
+            if not contributor.is_registered:
+                raise ValueError(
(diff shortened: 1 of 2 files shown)


osf/models/node.py
```diff
@@ -1240,6 +1240,8 @@ def add_contributor_registered_or_not(self, auth, user_id=None,
             contributor = OSFUser.load(user_id)
             if not contributor:
                 raise ValueError('User with id {} was not found.'.format(user_id))
+            if not contributor.is_registered:
+                raise ValueError('Cannot add unconfirmed user {} to node {}'.format(user_id, self._id))
```
brianjgeiger: I think this is not entirely correct. You can add an unconfirmed user, but you have to do it another way. Can you indicate the proper method in this error response?

PR #7299, opened 2017-06-05
[OSF-8099] API citations show wrong names

baylee-d:
## Purpose

For some unregistered contributors, the API was showing the wrong name.  When there is a new unregistered contributor being added to a project, the API will automatically check if the email being used is already in use by another contributor.  If it is, it'll add the new unregistered contributor's name to a new field under the contributor that already exists.  

For citations it will get the first and last name fields of the contributor.  If the contributor is an unregistered contributor with an email that already exists, this can be an issue if the names aren't exactly the same.  

## Changes

The changes include adding a conditional to the API's logic when making citations.  If the user isn't registered, it will get the unregistered_contributor field instead of the first/last name fields.  

For example, the name 'Pasquale Anselmi' in the contributor list shows up correctly.
screen shot 2017-05-16 at 6 24 41 pm

But in the citation it only shows up as 'P.'
screen shot 2017-05-16 at 6 29 18 pm

The reason why it wasn't showing up correctly is because it was using the name of the previously entered in contributor name. 

In the test example after the changes, the unregistered user 'Pasquale Anselmi' was created using an email that's also being used by another test user:
screen shot 2017-06-05 at 8 09 06 am
And the citations show up correctly:
screen shot 2017-06-05 at 8 09 20 am

## Ticket

https://openscience.atlassian.net/browse/OSF-8099

## Notes to QA

The process for testing will be pretty much the same for https://github.com/CenterForOpenScience/ember-preprints/pull/373, but instead of checking the meta tags you'll just be checking the citations:

1.) Create a new preprint
2.) Add an unregistered contributor
3.) Save the preprint and view it.  Make sure that the meta tags are showing the right information for the contributors currently added
4.) Edit the preprint, this time removing the old unregistered contributor and adding a new unregistered contributor with the same email
5.) Save the changes and view the preprint.  Check the contributor list and the names inside the citations to make sure they reflect the current contributors

### DIFF
--- a/osf/models/node.py
+++ b/osf/models/node.py
@@ -638,7 +638,7 @@ class AbstractNode(DirtyFieldsMixin, TypedModel, AddonModelMixin, IdentifierMixi
             'id': self._id,
             'title': sanitize.unescape_entities(self.title),
             'author': [
-                contributor.csl_name  # method in auth/model.py which parses the names of authors
+                contributor.csl_name(self._id)  # method in auth/model.py which parses the names of authors
                 for contributor in self.visible_contributors
             ],
             'publisher': 'Open Science Framework',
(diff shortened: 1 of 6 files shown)


brianjgeiger:
@baylee-d Could you fix the conflicts on this, please?

brianjgeiger:
@baylee-d One of the tests is failing and seems like it is related to this PR.

PR #7300, opened 2017-06-05
[feature] Pytest Conversion: Node Part 1 [OSF-8038]

Nesiehr:
## Purpose

Testing Performance Increase

## Changes

Pytest conversion ~13 api_tests/nodes files
Small addition to tests/utils.py

## Side effects

None known.

## Ticket

https://openscience.atlassian.net/browse/OSF-8038

## QA Considerations

N/A


api_tests/nodes/views/test_view_only_query_parameter.py
```diff
@@ -1,256 +1,316 @@
...
+            'view_only': public_node_one_private_link.key,
         }, expect_errors=True)
-        assert_equal(res.status_code, 401)
+        assert res.status_code == 401
 
-    def test_private_project_logs_with_anonymous_link_does_not_expose_user_id(self):
-        res = self.app.get(self.private_node_one_url+'logs/', {
-            'view_only': str(self.private_node_one_anonymous_link.key),
+    #   test_private_project_logs_with_anonymous_link_does_not_expose_user_id
+        res = app.get(private_node_one_url+'logs/', {
```
alexschiller: spaces around operator, else, format

api_tests/nodes/views/test_node_sparse_fieldsets.py
```diff
@@ -1,76 +1,88 @@
...
         ids = [each['id'] for each in node_json]
-        assert self.public._id in ids
-        assert self.deleted._id not in ids
-        assert self.private._id not in ids
+        assert public_project._id in ids
+        assert deleted_project._id not in ids
+        assert private_project._id not in ids
 
         assert len(node_json) == 1
         node_json = node_json[0]
```
alexschiller: rename the first node_json to `resp_json` or something, this reassignment is unpleasant.

api_tests/nodes/views/test_node_sparse_fieldsets.py
```diff
@@ -115,61 +132,95 @@ def test_embed_sparse_same_type(self):
...
+
+    @pytest.fixture()
+    def contributing_write_user(self):
+        return AuthUserFactory()
+
+    @pytest.fixture()
+    def private_node_one(self, creation_user, contributing_read_user, contributing_write_user):
+        private_node_one = ProjectFactory(is_public=False, creator=creation_user, title='Private One')
+        private_node_one.add_contributor(contributing_read_user, permissions=[permissions.READ], save=True)
+        private_node_one.add_contributor(contributing_write_user, permissions=[permissions.WRITE], save=True)
```
alexschiller: This gives only write permission, which does not implicitly give read. While it probably won't cause a problem since model level things will get it implicitly, it is best to add READ anywhere you explicitly add WRITE.

api_tests/view_only_links/views/test_view_only_link_nodes.py
```diff
@@ -1,111 +1,164 @@
+import pytest
 from nose.tools import *  # flake8: noqa
```
alexschiller: nose needed?

api_tests/view_only_links/views/test_view_only_link_nodes.py
```diff
@@ -1,111 +1,164 @@
...
 
+from website.util import permissions
 from api.base.settings.defaults import API_BASE
+from tests.base import ApiTestCase
+from osf_tests.factories import (
+    ProjectFactory,
+    AuthUserFactory,
+    PrivateLinkFactory,
+    NodeFactory
```
alexschiller: comma

api_tests/nodes/serializers/test_serializers.py
```diff
@@ -1,64 +1,62 @@
...
         parent_link = relationships['parent']['links']['related']['href']
-        assert_equal(
-            urlparse(parent_link).path,
-            '/{}nodes/{}/'.format(API_BASE, parent._id)
-        )
-        assert_in('registrations', relationships)
+        assert (
+            urlparse(parent_link).path ==
+            '/{}nodes/{}/'.format(API_BASE, parent._id))
+        assert 'registrations' in relationships
```
Nesiehr: You're saying to change the assert 'registrations' in relationships to assert * in relationships?

alexschiller: @Nesiehr No, I mean  there are bunch of other assert x in relationships a few lines up and it didn't make sense for this one to be apart from those

alexschiller: https://github.com/CenterForOpenScience/osf.io/pull/7300/files/228573a0f6495f1cda595cc47e100f9e13a0d232#diff-c9b86918299f4661da979f26c652adb1R45

Nesiehr: Ahh I see, I put it with its family

api_tests/nodes/serializers/test_serializers.py
```diff
@@ -67,13 +65,12 @@ def test_fork_serialization(self):
...
-        )
+        assert (
+            urlparse(forked_from).path ==
+            '/{}nodes/{}/'.format(API_BASE, node._id))
 
-    def test_template_serialization(self):
-        node = NodeFactory(creator=self.user)
+    #   test_template_serialization
+        node = NodeFactory(creator=user)
         fork = node.use_as_template(auth=Auth(user=node.creator))
```
Nesiehr: I get this error when I try to do that:
E   AttributeError: 'tuple' object has no attribute 'user'

Nesiehr: Auth(user) does work though

tests/utils.py
```diff
@@ -73,6 +70,16 @@ def wrapper(self, *args, **kwargs):
     return outer_wrapper
 
 @contextlib.contextmanager
+def assert_latest_log(log_action, node_key, index=-1):
+    node =  node_key
+    last_log = node.logs.latest()
+    node.reload()
+    yield
+    new_log = node.logs.order_by('-date')[-index - 1]
```
alexschiller: This looks like a roundabout way of getting to 0, is it needed?

api_tests/nodes/serializers/test_serializers.py
```diff
@@ -109,28 +106,25 @@ def test_serialization(self):
...
+            urlparse(registered_from).path ==
+            '/{}nodes/{}/'.format(API_BASE, reg.registered_from._id))
         for relationship in relationship_urls:
             if relationship in should_not_relate_to_registrations:
-                assert_not_in('/{}registrations/'.format(API_BASE), relationship_urls[relationship])
+                assert '/{}registrations/'.format(API_BASE) not in relationship_urls[relationship]
             else:
-                assert_in('/{}registrations/'.format(API_BASE), relationship_urls[relationship],
-                          'For key {}'.format(relationship))
+                assert '/{}registrations/'.format(API_BASE) in relationship_urls[relationship], 'For key {}'.format(relationship)
```
Nesiehr: This would be identical to the nose test correct?
assert '/{}registrations/'.format(API_BASE) in relationship_urls[relationship]
assert '/{}registrations/'.format(API_BASE) in 'For key {}'.format(relationship)

api_tests/nodes/views/test_node_citations.py
```diff
@@ -1,80 +1,95 @@
...
-
-class TestNodeCitationsStyle(NodeCitationsMixin, ApiTestCase):
-
-	def setUp(self):
-		super(TestNodeCitationsStyle, self).setUp()
-		self.public_url = '/{}nodes/{}/citation/apa/'.format(API_BASE, self.public_project._id)
-		self.private_url = '/{}nodes/{}/citation/apa/'.format(API_BASE, self.private_project._id)
+@pytest.fixture()
+def admin_contributor():
+	return AuthUserFactory()
```
Nesiehr: Looks like an indentation of four spaces, is that incorrect?

api_tests/nodes/views/test_node_citations.py
```diff
@@ -1,80 +1,95 @@
...
+
+@pytest.fixture()
+def public_project(admin_contributor):
+	return ProjectFactory(creator=admin_contributor, is_public=True)
+
+@pytest.fixture()
+def private_project(admin_contributor, rw_contributor, read_contributor):
+	private_project = ProjectFactory(creator=admin_contributor)
+	private_project.add_contributor(rw_contributor, auth=Auth(admin_contributor))
```
Nesiehr: Doesn't work here either

api_tests/nodes/views/test_node_citations.py
```diff
@@ -1,80 +1,95 @@
...
+	private_project.add_contributor(read_contributor, permissions=['read'], auth=Auth(admin_contributor))
+	private_project.save()
+	return private_project
+
+@pytest.mark.django_db
+class NodeCitationsMixin:
+
+	def test_node_citations(self, app, admin_contributor, rw_contributor, read_contributor, non_contributor, private_url, public_url):
```
Nesiehr: How do you mean exactly?

api_tests/nodes/views/test_node_embeds.py
```diff
@@ -1,100 +1,129 @@
...
+        root_node.add_contributor(contrib_two, ['read', 'write'], auth=auth, save=True)
+        return root_node
 
-    def setUp(self):
-        super(TestNodeEmbeds, self).setUp()
+    @pytest.fixture()
+    def child_one(self, auth, contrib_one, contrib_two, make_public_node, root_node):
+        child_one = make_public_node(parent=root_node)
+        child_one.add_contributor(contrib_one, ['read', 'write'], auth=auth, save=True)
+        child_one.add_contributor(contrib_two, ['read', 'write'], auth=auth, save=True)
```
Nesiehr: My first instinct is to say that one save would be faster, but I don't think it would be easy to accurately test that and I think the difference would likely be pretty marginal

api_tests/nodes/serializers/test_serializers.py
alexschiller: @Nesiehr Works for me-- though store the  `'/{}registrations/'.format(API_BASE)` so you don't have to do the format twice

api_tests/nodes/views/test_node_citations.py
alexschiller: screen shot 2017-06-08 at 10 07 23 am

api_tests/nodes/views/test_node_citations.py
alexschiller: https://docs.pytest.org/en/latest/parametrize.html

api_tests/nodes/views/test_node_embeds.py
```diff
@@ -1,100 +1,129 @@
...
+        url = '/{0}nodes/{1}/?embed=children'.format(API_BASE, root_node._id)
 
-        res = self.app.get(url, auth=self.user.auth)
+        res = app.get(url, auth=user.auth)
         embeds = res.json['data']['embeds']
-        ids = [self.child1._id, self.child2._id]
+        ids = [child_one._id, child_two._id]
         for child in embeds['children']['data']:
-            assert_in(child['id'], ids)
+            assert child['id'] in ids
```
Nesiehr: Hmm... I agree that checking the len of the response is a valid thing to make sure that children aren't somehow being duplicated, I'm not sure if there is a benefit to using a set in this case - what benefit does that provide in this instance?

Nesiehr: Especially if you check the length before running that original for loop - that would show that there aren't duplicates

alexschiller: @Nesiehr set protects against order being different:
screen shot 2017-06-08 at 10 15 22 am
Since the first version was iterating and checking `in` there is no assumptions about order being made, `set` keeps that assumption.
Also it needs to be set([child_one._id, child_two._id]) (nested in a list, whoops)

api_tests/nodes/views/test_node_registrations_list.py
```diff
@@ -1,64 +1,77 @@
...
 
+from api.base.settings.defaults import API_BASE
 from tests.base import ApiTestCase
 from osf_tests.factories import (
     ProjectFactory,
     RegistrationFactory,
     AuthUserFactory,
 )
 
-
```
Nesiehr: I don't think most files follow that rule - should that be added to the checklist?

alexschiller: yes, https://stackoverflow.com/questions/33466860/expected-two-blank-lines-pep8-warning-in-python, https://www.python.org/dev/peps/pep-0008/#blank-lines doesn't

tests/utils.py
sloria: Drive-by: Agreed, the `index` parameter is probably unnecessary; I don't believe we override the default value (`-1`) anywhere. So it should suffice to do:

```python
new_log = node.logs.latest()
```

api_tests/nodes/views/test_node_citations.py
Nesiehr: Yea I see what it looks like on here, it just doesn't look the same in my text editor

alexschiller: Might be tabs not spaces?

api_tests/nodes/views/test_node_citations.py
```diff
@@ -1,80 +1,95 @@
...
+	private_project.save()
+	return private_project
+
+@pytest.mark.django_db
+class NodeCitationsMixin:
+
+    def test_node_citations(self, app, admin_contributor, write_contrib, read_contrib, non_contrib, private_url, public_url):
+
+	#   test_admin_can_view_private_project_citations
+		res = app.get(private_url, auth=admin_contributor.auth)
```
sloria: These are indented with tabs. Use 4 spaces instead of tabs.

api_tests/nodes/views/test_node_citations.py
```diff
@@ -1,80 +1,95 @@
...
+	private_project.add_contributor(read_contrib, permissions=['read'], auth=Auth(admin_contributor))
+	private_project.save()
+	return private_project
+
+@pytest.mark.django_db
+class NodeCitationsMixin:
+
+    def test_node_citations(self, app, admin_contributor, write_contrib, read_contrib, non_contrib, private_url, public_url):
+
+	#   test_admin_can_view_private_project_citations
```
sloria: This line is showing a flake error due to the indentation. Give these the correct indentation and spacing , e.g.

```python
def test_node_citations(self, app, admin_contributor, write_contrib, read_contrib, non_contrib, private_url, public_url):
    # test_blah_blah
    assert ....
```

PR #7313, opened 2017-06-07
[#OSF-7977]Forking a project with a long name fails

chennan47:
## Purpose
Fix the problem when fork a project with a very long name will fail.

## Changes
Now when you fork or template a new project and the title is going above 200 characters, it will truncate the end of the title and only has 200 characters for the title.

## Ticket
https://openscience.atlassian.net/browse/OSF-7977


brianjgeiger:
What happens if you do a chain of forking projects? If you start with "This project has a very long name and will be fine to start with but if we keep on doing things and adding words it will quickly become ridiculous and cause trouble" and fork that then fork that then fork it that and so on, will it ever fail? Or do we have something somewhere that keeps us for having "Fork of Fork of This project has a very long name and will be fine to start with but if we keep on doing things and adding words it will quickly become ridiculous and cause trouble"?

PR #7315, opened 2017-06-07
[#OSF-7684]Moving files between components breaks links in logs

chennan47:
## Purpose
Improve the logic to handle url from different file storage sometimes ends with'/' sometimes not

## Ticket
https://openscience.atlassian.net/browse/OSF-7684

### DIFF
--- a/api/logs/serializers.py
+++ b/api/logs/serializers.py
@@ -105,7 +105,8 @@ class NodeLogParamsSerializer(RestrictedDictSerializer):
         urls = obj.get('urls', None)
         if urls:
             view = urls.get('view', None)
-            if view:
+            #TODO: remove the check for '//' when the urls for folder is removed from api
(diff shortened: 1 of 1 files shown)


api/logs/serializers.py
```diff
@@ -105,11 +105,13 @@ def get_params_file(self, obj):
         urls = obj.get('urls', None)
         if urls:
             view = urls.get('view', None)
-            if view:
-                file_id = view.split('/')[-2]
-                provider = view.split('/')[-3]
+            #TODO: remove the check for '//' when the urls for folder is removed from api
```
alexschiller: This definitely needs a jira ticket linked to the original-- make sure you mention this TODO to remove when it is fixed.

chennan47: did in https://openscience.atlassian.net/browse/OSF-8121

PR #7316, opened 2017-06-07
Remove unused APIv1 profile endpoints [OSF-8107]

erinspace:
## Purpose

It appears that the following endpoints are unused by the frontend and can therefore be removed:
PUT /api/v1/profile/<uid>/edit/
GET /api/v1/profile/<user_id>/summary/

So let's get rid of them!

## Changes

- Remove `edit_profile` and `get_profile_summary` routes
- Remove those views as well
- Remove test that references `edit_profile`

## Side effects

None anticipated

## Ticket

https://openscience.atlassian.net/browse/OSF-8107

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -1048,12 +1048,6 @@ class TestUserProfile(OsfTestCase):
         super(TestUserProfile, self).setUp()
         self.user = AuthUserFactory()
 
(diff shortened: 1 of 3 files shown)


sloria:
👍  Merged into `feature/next-djangole`(which will be merged into develop after IR 0.114.0). Consider this merged.

PR #7317, opened 2017-06-07
Remove unused APIv1 pointer endpoints [OSF-8108]

erinspace:
## Purpose

It appears that the following endpoints are unused by the frontend and can therefore be removed:
- POST /api/v1/pointers/move/
- POST /api/v1/project/<pid>/pointer/
- DELETE /api/v1/folder/<pid>/pointer/<pointer_id>

## Changes
- Remove those three routes
- Remove views associated with those routes
- Remove tests accessing those routes or views

## Side effects

Doesn't look like any other code is accessing those pointer views, so none anticipated

## Ticket

https://openscience.atlassian.net/browse/OSF-8108


sloria:
Merged into `feature/next-djangole`.

PR #7318, opened 2017-06-07
[Feature] Make sentry error less annoying [OSF-7823]

fabmiz:
## Purpose

Make sentry aggregate errors 

## Changes

- Modify sentry (flask client)  config to take extra data as optional keyword arg
- Modify website view to call sentry with guid as extra data

## Ticket

https://openscience.atlassian.net/browse/OSF-7823

### DIFF
--- a/framework/sentry/__init__.py
+++ b/framework/sentry/__init__.py
@@ -35,13 +35,11 @@ def log_exception():
     })
 
 
-def log_message(message):
+def log_message(message, extra={}):
     if not enabled:
         logger.warning(
             'Sentry called to log message, but is not active: %s' % message
         )
         return None
 
-    return sentry.captureMessage(message, extra={
-        'session': get_session_data(),
-    })
+    return sentry.captureMessage(message, extra=extra)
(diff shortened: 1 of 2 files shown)


framework/sentry/__init__.py
```diff
@@ -35,13 +35,16 @@ def log_exception():
     })
 
 
-def log_message(message):
+def log_message(message, extra_data={}):
```
sloria: Drive-by comment: Don't use mutable default arguments: http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

website/views.py
```diff
@@ -246,7 +246,7 @@ def resolve_guid(guid, suffix=None):
         #   expected.
         if not hasattr(guid_object.referent, 'deep_url'):
             sentry.log_message(
-                'Guid `{}` resolved to an object with no deep_url'.format(guid)
+                'Guid resolved to an object with no deep_url', dict(guid = guid)
```
sloria: Drive-by: remove extra spacing around `=`, or do `{'guid': guid}`.

framework/sentry/__init__.py
fabmiz: Thanks a lot for the tip!

PR #7320, opened 2017-06-07
Port maintenance code to use Postgres [OSF-7737]

erinspace:
## Purpose

The code for the backend to put up the maintenance notice was solidly mongo related - make that use django and postgres instead!

## Changes

- new `MaintenanceState` model that will hold details of the maintenance state
- updated maintenance tasks and methods for setting maintenance state
- tests for maintenance state methods

## Side effects

A migration needs to be run to add the `MaintenanceState` model

## Ticket
https://openscience.atlassian.net/browse/OSF-7737


website/maintenance.py
```diff
@@ -40,27 +29,31 @@ def set_maintenance(start=None, end=None):
...
-    Return None for start and end if there is no maintenance state
+    """Get the current start and end times for the current maintenance state.
+    Return None for start and end if there is no current maintenance state.
+    Return None if there is a maintenance state set for the future.
     """
-    if not database:
-        return None
-    maintenance_state = database.maintenance.find_one({'maintenance': True})
+    now = timezone.now()
+    maintenance_state = MaintenanceState.objects.filter(start__lte=now, end__gte=now).first()
```
erinspace: I wanted to be able to set maintenance periods in the future -- this way there can be a state set for the future, and then `get_maintenance` will still return the right thing if there isn't maintenance at that exact moment.  I can change it tho so that you can't set periods for the future if that'll be better!

sloria: I'm not sure I understand..can you clarify your intent? We certainly need to be able to set maintenance periods in the future (they will *always* be in the future), but why is filtering here necessary to achieve that?

osf_tests/test_maintenance.py
```diff
@@ -0,0 +1,65 @@
...
+from django.utils import timezone
+
+from website import maintenance
+from osf.models import MaintenanceState
+
+
+pytestmark = pytest.mark.django_db
+
+
+class TestMaintenance(unittest.TestCase):
```
sloria: Since we're using pytest now, we no longer need to subclass `TestCase`. You  can still use a [teardown method](https://docs.pytest.org/en/latest/xunit_setup.html?highlight=teardown#method-and-function-level-setup-teardown).

```python
class TestMaintenance:
    def teardown_method(self, method):
         #...
```

sloria: Erm, looks like `teardown_method` won't quite work due to a bug in pytest-django: https://github.com/pytest-dev/pytest-django/issues/297. I suppose you could use an autoused fixture that had a teardown:


```python
class TestMaintenance:

    @pytest.fixture(autouse=True)
    def clear_maintenance(self):
        # Clear out all maintenance records after each test
        yield
        MaintenanceState.objects.all().delete()
```

sloria: But the fixture isn't necessarily better than `tearDown`; either way is fine.

PR #7325, opened 2017-06-08
[Feature] Enforce prepopulating user emails [OSF-7790]

fabmiz:
## Purpose

Enforce prepopulating user emails in admin: user password reset form

## Changes

Override formView get_initial data to set default user emails 

## Ticket

https://openscience.atlassian.net/browse/OSF-7790


admin/users/views.py
```diff
@@ -514,17 +514,26 @@ class ResetPasswordView(PermissionRequiredMixin, FormView):
...
                     self.context_object_name.title(),
                     self.kwargs.get('guid')
                 ))
-        kwargs.setdefault('guid', user._id)
+        return super(ResetPasswordView, self).dispatch(request, *args, **kwargs)
+
+    def get_initial(self):
+        user = self.user
+        self.initial = {
+            'emails': [(r, r) for r in user.emails.values_list('address', flat=True)],
```
alexschiller: You can just use `self.user` instead of `user` and remove the `user = self.user`.

I see three places (this one and two in tests) where you do this, change all three.

fabmiz: Thanks. The flake8 is unrelated...it seems to be happening on all my PRs.

PR #7326, opened 2017-06-09
[feature] Special case checkout banner for preregistration checkouts [OSF-8071]

alexschiller:
## Purpose
Prereg checks out files when they are viewed. The files show whoever last viewed the prereg as the person who checked it out to the user without notifying the user as to why. Who is David Mellor and why is he in my project checking out my files?

## Changes
Adds a special case to the banner for prereg checkouts
screen shot 2017-06-09 at 3 20 38 pm

## Side effects
The prereg checkout banner will not show up if the prereg admin who checked out the file is a contributor on the project. The prereg banner will show up if a prereg admin was a contributor on a project that has a prereg draft and checked out a file before being removed as a contributor. That just might be the edgiest of edge cases. 

In other words, people with prereg (admin) view access might have some weird views, but they are weird anyway.

The original code appears to have a small chance of a race condition on an ajax request. This code moves that function call that used to occur at the end of the ajax request to the end of the controller setup. In theory bad things should be less likely to happen now, but I still think there might be a weird race condition that could occur with the fangorn instance and mithril. I saw one mithril error while testing, though I don't know if it was caused by this or my server resetting. I was unable to replicate it through 50+ page refreshes, so it is worth keeping an eye out for, but I don't think it is too big of a worry.

## Ticket
https://openscience.atlassian.net/browse/OSF-8071

## QA Considerations
1. Check that checkouts for prereg show the proper banner. (Make sure you create the prereg and view the prereg on admin with different users!!)
2. Confirm that one or two addon files load fine and don't break anything. There was some touching of related code. I checked github locally and looked fine.
3. Make sure revisions and editing load properly on both checkouts and files that are not checked out since this moved some of the code around
4. Make sure normal checkouts also work properly and don't accidentally show the banner.
5. Make sure all the normal checkout buttons appear (download, checkout/checkin/force checkin/delete) 
6. Make sure non-admin users see the proper buttons

### DIFF
--- a/addons/base/views.py
+++ b/addons/base/views.py
@@ -748,12 +748,23 @@ def addon_view_file(auth, node, file_node, version):
         'file_tags': list(file_node.tags.filter(system=False).values_list('name', flat=True)) if not file_node._state.adding else [],  # Only access ManyRelatedManager if saved
         'file_guid': file_node.get_guid()._id,
         'file_id': file_node._id,
-        'allow_comments': file_node.provider in settings.ADDONS_COMMENTABLE
+        'allow_comments': file_node.provider in settings.ADDONS_COMMENTABLE,
+        'checkout_user': file_node.checkout._id if file_node.checkout else None,
+        'pre_reg_checkout': is_pre_reg_checkout(node, file_node),
     })
 
     ret.update(rubeus.collect_addon_assets(node))
     return ret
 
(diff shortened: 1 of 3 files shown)


alexschiller:
@brianjgeiger I discussed the special cases with product and they were fine with them. I'll fill in Prereg people.

PR #7330, opened 2017-06-12
[feature] Shorten Travis Logs [OSF-8102]

Nesiehr:
## Purpose

Decrease Travis Log Output

## Changes

1. Silent loggers weren't working, set the logging level for all logs to critical
2. Set pytest to ignore user warnings, and only alert for deprecation and pending deprecation warnings once

## Side effects

Less verbose log output could make tracking things down more difficult

## Ticket

https://openscience.atlassian.net/browse/OSF-8102

### DIFF
--- a/api_tests/conftest.py
+++ b/api_tests/conftest.py
@@ -18,6 +18,9 @@ SILENT_LOGGERS = [
     'website.search.elastic_search',
     'website.search_migration.migrate',
     'website.util.paths',
+    'osf.migrations',
+    'osf.models',
+    'addons.osfstorage.models',
 ]
 for logger_name in SILENT_LOGGERS:
     logging.getLogger(logger_name).setLevel(logging.CRITICAL)
--- a/api_tests/conftest.py
+++ b/api_tests/conftest.py
@@ -30,7 +30,7 @@ from tests.json_api_test_app import JSONAPITestApp
 # for logger_name in SILENT_LOGGERS:
 #     logging.getLogger(logger_name).setLevel(logging.CRITICAL)
 
-logging.getLogger().setLevel(logging.CRITICAL)
+logging.disable(logging.WARNING)
 
 @pytest.fixture()
 def app():
--- a/api_tests/conftest.py
+++ b/api_tests/conftest.py
@@ -21,6 +21,8 @@ SILENT_LOGGERS = [
     'osf.migrations',
     'osf.models',
     'addons.osfstorage.models',
+    'rest_framework.pagination',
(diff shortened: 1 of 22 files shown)


api_tests/conftest.py
```diff
@@ -5,23 +5,32 @@
...
+#     'addons.osfstorage.models',
+#     'rest_framework.pagination',
+#     'tests.base',
+#     'website.project',
+#     'django.db.models.fields',
+# ]
+# for logger_name in SILENT_LOGGERS:
+#     logging.getLogger(logger_name).setLevel(logging.CRITICAL)
+
+logging.disable(logging.CRITICAL)
```
sloria: From the docs:

>  Its effect is to disable all logging calls of severity lvl and below, so that if you call it with a value of INFO, then all INFO and DEBUG events would be discarded, whereas those of severity WARNING and above would be processed according to the logger’s effective level.

This means that we would be ignoring CRITICAL messages, which we don't want to do. I think `logging.WARNING` would makes sense here.

sloria: That said, does this result in a lot of savings? Could we be missing out on some legitimate warnings if we do this? It might be better to stick with the explicit list of loggers to silence if that's the case.

pytest.ini
```diff
@@ -1,3 +1,7 @@
 [pytest]
 DJANGO_SETTINGS_MODULE=osf_tests.settings
 addopts = --tb=short --reuse-db
+filterwarnings = 
+    ignore::UserWarning
```
sloria: Again, I'm concerned that we might miss out on legitimate warnings if we do this; maybe `once` would be better?

osf_tests/conftest.py
```diff
@@ -11,27 +11,35 @@
...
+#     'addons.osfstorage.models',
+#     'rest_framework.pagination',
+#     'tests.base',
+#     'website.project',
+#     'django.db.models.fields',
+# ]
+# for logger_name in SILENT_LOGGERS:
+#     logging.getLogger(logger_name).setLevel(logging.CRITICAL)
 
+logging.disable(logging.WARNING)
```
sloria: We don't want to disable *all* warnings. Some of these are useful. For example, we would not have found https://openscience.atlassian.net/browse/OSF-8136 if not for the warnings from Django.

We only want to disable *expected* warnings, which is why we had the list of `SILENT_LOGGERS`. Would it make sense to revert this change? I think your configuration change in `pytest.ini` looks correct and may be enough to close the ticket.

PR #7339, opened 2017-06-13
QuickFilesNode model and API Views, methods, and tests [OSF-8116] [OSF-8117]

erinspace:
## Purpose

From the [Modeling section of the requirements doc](https://docs.google.com/document/d/1pEmGrsLlpklwaNkbyY8uGzmZX4jjLuDftDI2mr1l4PE/edit#heading=h.qqejb3a5h72):
New QuickFiles model that is an AbstractNode subclass. Here are the parts that are model-specific:
- Cannot delete
- Public upon creation; cannot make private
- Cannot be forked
- Cannot be registered
- Cannot have VOLs
- Cannot have addons other than osfstorage
- GUIDs longer for these nodes (10 chars) 
- When an OSFUser is created, create a QuickFiles node (similar to how a Bookmark collection is created for every user)
- QuickFiles nodes have the title "<<user's fullname>>'s Quick Files"
- If a user updates their name, the title of Quickfiles will change as well.

Also includes API updates: `/v2/users/<uid>/files/` and links on the user detail page to view the files in quickfiles and upload a new one.

## Changes

- New `QuickFilesNode` model and method overrides that match the requirements described above.
- Create a new `QuickFilesNode` for a user when they are created
- Tests for `QuickFilesNode` methods and creation
- Migration to add `QuickFilesNode` model
- Script to add a `QuickFilesNode` for every user that doesn't yet have one
- If a user changes their name, the title of the QuickFiles project changes as well

API side:
- new `/v2/users/<uid>/files/` url and views (also supports `/v2/users/me/files/`)
- files relationship to the user serializer including an upload link for the quickfiles node on waterbutler

## Side effects
- A *long* Migration to add `QuickFilesNode` model and to add a `QuickFilesNode` for all existing users will need to be run
- this PR doesn't include associated frontend work, so all testing for now will need to happen via API.

## Ticket
backend:
https://openscience.atlassian.net/browse/OSF-8116
api:

## QA Considerations
- Most of this is backend and API work, so you shouldn't have any user facing UI things to test yet, only API things
- There is a script that needs to be run before existing users will have a quickfiles node, so API things will not work with existing users on staging until that happens
- When testing via the API, posting to the WaterButler `upload` link found at `v2/users/<uid>/` under the files relationship should indeed create new files in your quickfiles "project"
- a GET request on `v2/users/<uid>/files/` should show the quickfiles for that user
- a GET request on `v2/users/me/files/` should show the files for the currently logged in user


osf/models/node.py
```diff
@@ -71,7 +71,7 @@
 class AbstractNodeQuerySet(IncludeQuerySet):
 
     def get_roots(self):
-        return self.filter(id__in=self.exclude(type='osf.collection').values_list('root_id', flat=True))
+        return self.filter(id__in=self.exclude(type='osf.collection').exclude(type='osf.QuickFiles').values_list('root_id', flat=True))
```
sloria: `osf.quickfiles` should be lowercased.

scripts/add_quickfiles_to_osfusers.py
```diff
@@ -0,0 +1,48 @@
...
+from website.project import new_quickfiles_project
+
+setup_django()
+logger = logging.getLogger(__name__)
+
+
+def main(dry=True):
+    from osf.models import OSFUser
+
+    users_without_quickfiles = OSFUser.objects.exclude(created__is_quickfiles_node=True)
```
sloria: This query needs to be updated, since `is_quickfiles_node` is not a field.

website/project/__init__.py
```diff
@@ -92,6 +92,31 @@ def new_bookmark_collection(user):
     return collection
 
 
+def get_quickfiles_project_title(user):
```
sloria: This file is an artifact of v1. To keep the code more cohesive, it might make sense to put these helpers in `osf.models.QuickFiles`.

Also, instead of `new_quickfiles_project(user)`, it would be more intuitive to implement a manager method, e.g. `QuickFiles.objects.create_for_user(user)`. It should error if the user already has quickfiles node.

osf/models/user.py
```diff
@@ -1416,3 +1416,21 @@ class Meta:
...
+@receiver(post_save, sender=OSFUser)
+def update_quickfiles_project_title(sender, instance, created, **kwargs):
+    from website.project import get_quickfiles_project_title
+    from osf.models.quickfiles import QuickFiles
+
+    if not created:
+        quickfiles = QuickFiles.objects.filter(creator=instance).first()
+        if quickfiles:
+            quickfiles.title = get_quickfiles_project_title(instance)
+            quickfiles.save()
```
sloria: It'd be nice if we could avoid doing this on every OSFUser.save call.

Could you instead do this `OSFUser.save` and check if the user's name has changed (in `dirty_fields`)?

osf/models/quickfiles.py
```diff
@@ -0,0 +1,71 @@
...
+from osf.models.node import AbstractNode, AbstractNodeManager
+from website.exceptions import NodeStateError
+
+
+logger = logging.getLogger(__name__)
+
+
+class QuickFilesManager(AbstractNodeManager):
+
+    def create_for_user(self, user):
```
erinspace: Added!

addons/osfstorage/views.py
```diff
@@ -208,6 +208,9 @@ def osfstorage_create_child(file_node, payload, node_addon, **kwargs):
     if not (name or user) or '/' in name:
         raise HTTPError(httplib.BAD_REQUEST)
 
+    if file_node.node.is_quickfiles and is_folder:
+        raise HTTPError(httplib.FORBIDDEN, data={'message': 'You may not create a folder for QuickFiles'})
```
sloria: v1 errors should be passed `message_long` (not `message`).

website/search/elastic_search.py
```diff
@@ -489,7 +495,7 @@ def update_file(file_, index=None, delete=False):
     index = index or INDEX
 
     # TODO: Can remove 'not file_.name' if we remove all base file nodes with name=None
-    if not file_.name or not file_.node.is_public or delete or file_.node.is_deleted or file_.node.archiving:
+    if not file_.name or not file_.node.is_public or delete or file_.node.is_deleted or file_.node.archiving or 'spam_confirmed' in file_.node.creator.system_tags:
```
sloria: This is adding a lot of joins (on `node`, `creator`, and `tags`) every time we update a file in ES. For now, it might make more sense to just delete the files in the `update_user` function:

```python
def update_user(..):

    ...
    for quickfile in BaseFileNode.objects.filter(...).only('_id'):
         client().delete(
             index=index,
             doc_type='file',
             id=file_._id,
             refresh=True,
             ignore=[404]
        )
```

osf_tests/test_node.py
```diff
@@ -3951,7 +3951,7 @@ def test_empty_is_public(self):
         qs = AbstractNode.objects.can_view()
 
         assert proj in qs
-        assert qs.count() == 1
+        assert qs.count() == 2  # proj and user's quickfiles node
```
sloria: Minor: It might make more sense to use `Node` rather than `AbstractNode` in these tests, so we can avoid dealing with QuickFiles.

osf/migrations/0041_add_quickfiles.py
```diff
@@ -0,0 +1,63 @@
...
+from django.core.management.sql import emit_post_migrate_signal
+
+from osf.models import OSFUser, QuickFiles
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+
+def add_quickfiles(*args, **kwargs):
+    emit_post_migrate_signal(2, False, 'default')
```
sloria: Can you explain why this is necessary?

api/users/views.py
```diff
@@ -532,6 +533,37 @@ def get_queryset(self):
...
+
+    required_read_scopes = [CoreScopes.ALWAYS_PUBLIC]
+    required_write_scopes = [CoreScopes.NODE_FILE_WRITE]
+
+    serializer_class = UserQuickFilesSerializer
+    view_category = 'users'
+    view_name = 'user-files'
+
+    def get_node(self, check_object_permissions):
+        return QuickFiles.objects.get(creator=self.get_user(check_permissions=False))
```
sloria: Can use `get_for_user`.

api/users/serializers.py
```diff
@@ -13,6 +15,19 @@
 from api.base.utils import absolute_reverse, get_user_auth
 
 
+class QuickFilesUploadRelationshipField(RelationshipField):
+
+    def to_representation(self, value):
+        relationship_links = super(QuickFilesUploadRelationshipField, self).to_representation(value)
+        quickfiles_guid = QuickFiles.objects.get(creator=value)._id
```
sloria: You can query for the `_id` only here.

```python
quickfiles_guid = QuickFiles.objects.filter(creator=value).values_list('_id', flat=True).get()
```

osf/migrations/0041_add_quickfiles.py
erinspace: I thought this was necessary so that the models created earlier in the very same migration were available in the next function, as I had this problem in a previous PR where I was creating permissions and then wanted to assign them in groups in one migration -- found the [solution in this django ticket](https://code.djangoproject.com/ticket/23422) -- though I'm realizing that that might have been Permissions specifically created... so I'll remove it! (having trouble dropping my local DB completely, but reversing the migration seems to be OK without it...)

sloria: I think that is only pertinent to permissions because permissions are automatically created from `class Meta` options.

erinspace: Makes sense! Thank you!

tests/test_quickfiles.py
```diff
@@ -0,0 +1,60 @@
...
+def flask_app():
+    return TestApp(test_app)
+
+
+@pytest.fixture()
+def post_to_quickfiles(quickfiles, user, flask_app, **kwargs):
+    def func(name, *args, **kwargs):
+        osfstorage = quickfiles.get_addon('osfstorage')
+        root = osfstorage.get_root()
+        url = '/api/v1/project/{}/osfstorage/{}/children/'.format(quickfiles._id, root._id)
```
erinspace: For clarification on the review, I posted here to v1 to mimic the structure of [other osfstorage upload tests](https://github.com/CenterForOpenScience/osf.io/blob/develop/addons/osfstorage/tests/test_views.py#L146) without re-doing or re-implementing the structure of the Hook test cases. Simply posting to the v1 route seemed more straightforward

osf/migrations/0042_add_quickfiles.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def add_quickfiles(*args, **kwargs):
+    users_without_quickfiles = OSFUser.objects.exclude(created__type='osf.quickfiles')
+    logger.info('About to add QuickFiles for {} users.'.format(users_without_quickfiles.count()))
+
+    total_quickfiles_to_create = users_without_quickfiles.count()
+    quickfiles_to_create = []
+    for user in users_without_quickfiles:
+        quickfiles_to_create.append(
```
sloria: There are > 70K users on prod, so this list will take up quite a bit of memory. It would be a good idea to paginate. The page size can be pretty large (> 1000).

osf/models/quickfiles.py
```diff
@@ -0,0 +1,74 @@
...
+        kwargs['is_public'] = True
+        super(QuickFiles, self).__init__(*args, **kwargs)
+
+    def remove_node(self, auth, date=None):
+        raise NodeStateError('A QuickFiles node may not be deleted.')
+
+    def set_privacy(self, permissions, *args, **kwargs):
+        if permissions == 'private':
+            raise NodeStateError('A QuickFiles node cannot be made private.')
+        return super(QuickFiles, self).set_privacy(permissions, *args, **kwargs)
```
sloria: I think this method should error regardless of the value of `permissions`. `set_privacy` should never be called on a quickfiles node.

osf/models/user.py
```diff
@@ -7,6 +7,7 @@
 from copy import deepcopy
 from flask import Request as FlaskRequest
 from framework import analytics
+from os.path import splitext
```
sloria: Minor: Follow pep8 conventions for import order/grouping: stdlib, then 3rd party, then internal.

osf/models/quickfiles.py
```diff
@@ -0,0 +1,74 @@
...
+
+        quickfiles.add_addon('osfstorage', auth=None, log=False)
+
+        return quickfiles
+
+    def get_for_user(self, user):
+        return QuickFiles.objects.get(creator=user)
+
+
+class QuickFiles(AbstractNode):
```
mfraezz: Should this also override `use_as_template` to error? Alternatively, doing so with `clone` would prevent registering, forking, and templating.

Also, should `add_addon` be overridden to break for anything other than `'osfstorage'`?

osf/migrations/0043_add_quickfiles.py
```diff
@@ -0,0 +1,90 @@
...
+            total_created += 1
+
+        all_quickfiles = QuickFiles.objects.bulk_create(quickfiles_to_create)
+
+        contributors_to_create = []
+        for quickfiles in all_quickfiles:
+            quickfiles.add_addon('osfstorage', auth=None, log=False)
+            quickfiles.save()
+
+            contributors_to_create.append(
```
erinspace: It totally does when you use the shell! It's the post-save signal that adds the creator as contributor for QuickFiles, and that doesn’t get called in the migration. Found this little bug when working with Henrique who was using the current state of the QuickFiles backend. So this does add a step but as far as I can tell is actually needed

osf/migrations/0043_add_quickfiles.py
```diff
@@ -0,0 +1,90 @@
...
+
+from osf.models import OSFUser, QuickFiles, Contributor
+from osf.models.quickfiles import get_quickfiles_project_title
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+
+def add_quickfiles(*args, **kwargs):
+    ids_without_quickfiles = list(OSFUser.objects.exclude(created__type='osf.quickfiles').values_list('id', flat=True))
```
erinspace: Yes! They will be the exact same! I originally had exactly what you suggest here, but ran into a lot of trouble when re-writing the migration to use pagination. It turns out that for pagination, the query was re-run every time, leading to a weird moving target issue where `users_without_quickfiles` would shrink every page, and therefore never complete all available pages.

I went over this with @sloria for quite awhile, and this was the solution (although definitely weird and verbose) solution we came up with.  I tried running it several different times with different methods (I have the data for using timeit if you wanna see) and this was the fastest that actually was able to get through all the pages.

osf/models/quickfiles.py
erinspace: Done!

osf/models/user.py
```diff
@@ -680,14 +682,45 @@ def merge_user(self, user):
...
         for file_node in BaseFileNode.files_checked_out(user=user):
             file_node.checkout = self
             file_node.save()
 
+        # - move files in the merged user's quickfiles node, checking for name conflicts
+        from osf.models import QuickFiles
+        primary_quickfiles = QuickFiles.objects.get(creator=self)
+        merging_user_quickfiles = QuickFiles.objects.get(creator=user)
+
+        files_in_merging_user_quickfiles = BaseFileNode.objects.filter(node=merging_user_quickfiles, type='osf.osfstoragefile')
```
mfraezz: Minor: could do `merging_user_quickfiles.files.filter(type='osf.osfstoragefile')`. This should execute the same query, so it's no more efficient; it just reads nicer.

osf/models/user.py
```diff
@@ -680,14 +682,45 @@ def merge_user(self, user):
...
+        for merging_user_file in files_in_merging_user_quickfiles:
```
mfraezz: Could this fall prey to the queryset modifying issue that was present in the migration? You're filtering on `node`, but modifying `node`.

addons/osfstorage/tests/utils.py
```diff
@@ -63,3 +64,26 @@ def recursively_create_folder(settings, path):
     for subpath in path:
         current = current.append_folder(subpath)
     return current.append_file(final)
+
+
+def make_payload(user, name, **kwargs):
```
erinspace: It's also used in [Quickfiles tests](https://github.com/CenterForOpenScience/osf.io/pull/7339/files#diff-46cd1bab6ae49e96cd33eb4457cce3dbR18), I figured it was better to abstract this method to utils instead of duplicating the exact same code in two different places

osf/models/user.py
erinspace: The specific problem in the migration was only happening because of the pagination aspect -- without the pagination the queryset wasn't being re-evaluated every time.  I haven't caught a problem yet [with the merging user files tests I have](https://github.com/CenterForOpenScience/osf.io/pull/7339/files#diff-bfa630527bd745d1d1d5fe97eb1653d4R101) but let me know if there's a specific case this might be a problem for!

sloria:
This migration for this will require downtime, yes?

erinspace:
Yes! @mfraezz ran the migration on prod data and said it had a runtime of about 45 minutes.

api/users/urls.py
```diff
@@ -13,5 +13,6 @@
     url(r'^(?P<user_id>\w+)/nodes/$', views.UserNodes.as_view(), name=views.UserNodes.view_name),
     url(r'^(?P<user_id>\w+)/preprints/$', views.UserPreprints.as_view(), name=views.UserPreprints.view_name),
     url(r'^(?P<user_id>\w+)/registrations/$', views.UserRegistrations.as_view(), name=views.UserRegistrations.view_name),
+    url(r'^(?P<user_id>\w+)/files/$', views.UserQuickFiles.as_view(), name=views.UserQuickFiles.view_name),
```
sloria: When I view a user detail endpoint, I don't see a link to this endpoint in the response:

user_detail_ _django_rest_framework

Is this expected? Shouldn't there be a `related` link?

osf/models/quickfiles.py
```diff
@@ -0,0 +1,76 @@
...
```
erinspace: @sloria happy to do this -- just want to clarify as originally I did have `QuickFilesNode` and was asked to change it [earlier in the process](https://github.com/CenterForOpenScience/osf.io/pull/7339#discussion_r123606542) both in the model name and in methods.  Should I change back all of the Model name and all methods as well?

api/users/urls.py
erinspace: This format was discussed with @brianjgeiger over the best way to deal with the upload link existing on a user page over a file detail page for ease of access. This is the way he suggested I format it, happy to change it if opinions are now different

tests/test_quickfiles.py
```diff
@@ -0,0 +1,60 @@
...
+
+        assert res.status_code == 201
+        assert res.json['status'] == 'success'
+        assert BaseFileNode.objects.filter(type='osf.osfstoragefile', node__creator=user, name=name).exists()
+
+    def test_create_folder_throws_error(self, flask_app, user, quickfiles, post_to_quickfiles):
+        name = 'new_illegal_folder'
+        res = post_to_quickfiles(name, kind='folder', expect_errors=True)
+
+        assert res.status_code == 403
```
erinspace: I used a 403 instead of a 400 because it seemed to match the [RFC definitions]() -- 

400 Bad Request - The request could not be understood by the server due to malformed syntax.
vs
403 Forbidden
The server understood the request, but is refusing to fulfill it. Authorization will not help and the request SHOULD NOT be repeated.

You can't create a folder on quickfiles, and change of syntax won't help. If I'm misunderstanding or if we use 400 for consistency I'll definitely change it

api/users/urls.py
erinspace: OH. Ok I think I know what you were saying. Yes, there should be related. My bad!

osf/models/quickfiles.py
sloria: Sorry for the confusion! The comment you linked to still stands--`node.is_quickfiles` is correct. The model name is what I am proposing to change--properties and methods can stay the same.

That said, I don't feel strongly about changing it. I'll let you have final say on whether or not it's worth it.

osf/models/user.py
```diff
@@ -679,14 +681,45 @@ def merge_user(self, user):
...
             file_node.save()
 
+        # - move files in the merged user's quickfiles node, checking for name conflicts
+        from osf.models import QuickFiles
+        primary_quickfiles = QuickFiles.objects.get(creator=self)
+        merging_user_quickfiles = QuickFiles.objects.get(creator=user)
+
+        files_in_merging_user_quickfiles = merging_user_quickfiles.files.filter(type='osf.osfstoragefile')
+        for merging_user_file in files_in_merging_user_quickfiles:
+            if BaseFileNode.objects.filter(node=primary_quickfiles, name=merging_user_file.name, type='osf.osfstoragefile').exists():
```
cwisecarver: It removes magic strings :+1:

tests/test_quickfiles.py
erinspace: I understand now, will change it

sloria: 403 is used specifically for authorization/permissions.

RFC7231 has more complete definitions:

> If authentication credentials were provided in the request, the server considers them insufficient to grant access. 

https://tools.ietf.org/html/rfc7231#section-6.5.3

In this case, the client is sending a semantically invalid request--not an unauthorized request--and therefore a 400 is appropriate.

api/users/urls.py
sloria: Yes, the other links are correct; just need to add the one to `/v2/<userid>/files/`

tests/test_quickfiles.py
sloria: woops, posted after your previous comment. 👍

api/users/urls.py
erinspace: fixed!

osf/models/user.py
```diff
@@ -709,10 +711,14 @@ def merge_user(self, user):
...
-                # check if new name conflicts, update til it does not
-                while BaseFileNode.objects.filter(node=primary_quickfiles, name=new_name, type='osf.osfstoragefile').exists():
+                # check if new name conflicts, update til it does not (try up to 1000 times)
+                rename_count = 0
+                while OsfStorageFileNode.objects.filter(node=primary_quickfiles, name=new_name).exists():
                     digit += 1
                     new_name = new_name_format.format(name_without_extension, digit, extension)
+                    rename_count += 1
+                    if rename_count >= MAX_QUICKFILES_MERGE_RENAME_ATTEMPTS:
+                        raise ValueError('Maximum number of rename attempts has been reached')
```
erinspace: AH that's perfect! I'll do that. I was looking for a better error, put Value as kind of a placeholder and then forgot to change it...

sloria:
When visiting `/v2/users/`, 10 queries for the `quickfilesnode` are being done *for each user* (100 total additional queries).

user_list_ _django_rest_framework_and_node_py____projects_cos-projects_osf_website_project_views__-_vim

Can we avoid these?

sloria:
This looks like it *might* work...

```diff
diff --git a/api/users/serializers.py b/api/users/serializers.py
index 22d6b82271..e90cb463f7 100644
--- a/api/users/serializers.py
+++ b/api/users/serializers.py
@@ -19,7 +19,7 @@ class QuickFilesRelationshipField(RelationshipField):

     def to_representation(self, value):
         relationship_links = super(QuickFilesRelationshipField, self).to_representation(value)
-        quickfiles_guid = QuickFilesNode.objects.get_for_user(value)._id
+        quickfiles_guid = value.created.filter(type='osf.quickfilesnode').values_list('guids___id',
         upload_url = website_utils.waterbutler_api_url_for(quickfiles_guid, 'osfstorage')
         relationship_links['links']['upload'] = {
             'href': upload_url,
diff --git a/api/users/views.py b/api/users/views.py
index 9edff281ef..ab717b71a2 100644
--- a/api/users/views.py
+++ b/api/users/views.py
@@ -155,8 +155,8 @@ class UserList(JSONAPIBaseView, generics.ListAPIView, ODMFilterMixin):
     # overrides ListCreateAPIView
     def get_queryset(self):
         # TODO: sort
-        query = self.get_query_from_request()
-        return OSFUser.find(query)
+        # query = self.get_query_from_request()
+        return OSFUser.objects.prefetch_related('created').filter(is_registered=True)


 class UserDetail(JSONAPIBaseView, generics.RetrieveUpdateAPIView, UserMixin):
```

This isn't a complete solution (the filtering in `get_queryset` isn't correct), but it might be starting point.
```

**Comment by erinspace** [comment]
Oh, hah, @sloria I just pushed a commit that's very similar and gets the # of queries from 24 to 14!

File: api/users/serializers.py [hunk 27]
```diff
@@ -19,7 +19,7 @@ class QuickFilesRelationshipField(RelationshipField):
 
     def to_representation(self, value):
         relationship_links = super(QuickFilesRelationshipField, self).to_representation(value)
-        quickfiles_guid = QuickFilesNode.objects.get_for_user(value)._id
+        quickfiles_guid = QuickFilesNode.objects.filter(creator=value).values_list('guids___id').get()
```
sloria: You can use the reverse relation here, which will take advantage of Django's query caching:

```python
value.created.filter(type='osf.quickfilesnode').values_list('guids___id').get()
```

erinspace: I really need to get better at reverse relations. I'll work on that! (also, fixed, thank you)

sloria:
The `/v2/users/me/` endpoint went from 2 queries (~50ms) to 6 queries (~225ms). @erinspace Can you investigate if we can avoid these queries?

sloria:
Ah, so it looks like the numbers above are misleading. For some reason, the browse-able API calls `to_representation` on `QuickFilesRelationshipField` more than once for the `UserDetail` endpoint. When I render the endpoint as JSON (http://localhost:8000/v2/users/me/?format=json) and enable query logging in postgres, I only see one additional query for the QuickFilesNode.

erinspace:
Ok! I've been trying to figure out why `to_representation` is called so many times, and maybe a way to do it on another method, but still trying at it.

sloria:
For recordkeeping:

Paired with @erinspace to see if we can avoid those extra queries. We tried

- .prefetch_related('created')
- Making OSFUserManager inherit from IncludManager then doing `.include('created__guids')`. 

Neither of these prevented the additional quickfilesnode query.

This isn't a problem for now, so went ahead and merged.

cwisecarver:
@sloria It might be nice to pull the type off of the model class for queries instead of hardcoding those strings everywhere. There's also a bunch of places in here we're using BaseFile and filtering by type where we could just be using the typed class.

sloria:
@cwisecarver Good catches. `QuickFilesNode._typedmodels_type` can be used, and the concrete classes should be used instead of BaseFile whenever possible. @erinspace I'll put the ticket in Additl Dev; can you send a PR for these changes, please?

erinspace:
Ah such a good point, that makes a ton of sense -- sorry I messed that up. Thanks @sloria and @cwisecarver! I'm on it.

PR #7341, opened 2017-06-14
Remove old search and atom feed util functions for old SHARE [OSF-8153]

erinspace:
## Purpose

There are a few methods in website/search/util.py that were used by a very old iteration of SHARE – those are no longer needed.

## Changes

- remove atom feed parsing and cleanup methods and constants
- remove old SHARE search graph color parsing methods and constants
- remove webcolors from requirements

## Side effects

None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-8153


website/search/util.py
```diff
@@ -1,27 +1,8 @@
-import re
-import copy
 import logging
-import webcolors
```
sloria: Can this be removed from `requirements.txt`, too?

PR #7343, opened 2017-06-15
[hotfix] Fix Filtering by Date [OSF-8155]

caseyrollins:
#### Purpose
- Fix filtering nodes by date in the API.

#### Changes
- Add tests to check filtering nodes by dates.
- Handle lists of operations in `ListFilterMixin` as `filter[date][eq]=day` becomes `[filter[date][gt]=day-1, filter[date][lt]=day+1]`.

#### Side Effects
- None expected.

#### Ticket
- [OSF-8155](https://openscience.atlassian.net/browse/OSF-8155)

### DIFF
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -541,10 +541,12 @@ class ListFilterMixin(FilterMixin):
         if filters:
             for key, field_names in filters.iteritems():
                 for field_name, data in field_names.iteritems():
-                    if isinstance(queryset, list):
-                        queryset = self.get_filtered_queryset(field_name, data, queryset)
-                    else:
-                        queryset = self.filter_by_field(queryset, field_name=field_name, operation=data)
+                    operations = data if isinstance(data, list) else [data]
+                    for operation in operations:
+                        if isinstance(queryset, list):
+                            queryset = self.get_filtered_queryset(field_name, operation, queryset)
+                        else:
+                            queryset = self.filter_by_field(queryset, field_name, operation)
         return queryset
 
     def filter_by_field(self, queryset, field_name, operation):
@@ -581,7 +583,7 @@ class ListFilterMixin(FilterMixin):
             if operation['value'] not in (list(), tuple()):
                 operation['source_field_name'] = '_contributors__guids___id'
                 operation['op'] = 'iexact'
-        if operation['source_field_name'] == 'kind':
+        if field_name == 'kind':
             operation['source_field_name'] = 'is_file'
             # The value should be boolean
             operation['value'] = operation['value'] == 'file'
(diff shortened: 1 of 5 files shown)


api_tests/nodes/filters/test_filters.py
```diff
@@ -94,3 +94,103 @@ def test_contributor_filter(self):
         res = self.app.get('{}{}'.format(self.contributors_url, self.user_two._id), auth=self.user.auth)
         actual = [node['id'] for node in res.json['data']]
         assert_equal(expected, actual)
+
+
+class NodesListDateFilteringMixin(object):
```
brianjgeiger: Probably this should be at the top of the file rather than the bottom based on today's discussion. Mixins at the top make sense. Perhaps we'll add that to the cosdev docs one day.

api/base/filters.py
```diff
@@ -541,10 +541,12 @@ def param_queryset(self, query_params, default_queryset):
         if filters:
             for key, field_names in filters.iteritems():
                 for field_name, data in field_names.iteritems():
-                    if isinstance(queryset, list):
-                        queryset = self.get_filtered_queryset(field_name, data, queryset)
-                    else:
-                        queryset = self.filter_by_field(queryset, field_name=field_name, operation=data)
+                    operations = [item for item in data] if isinstance(data, list) else [data]
```
brianjgeiger: This seems weird to me. If data is a list, you list iterate over the list items to cast it to a list? If it were a generator, that makes sense. Wouldn't it be `operations = data if isinstance(data, list) else [data]`?

caseyrollins: It certainly would be. Please excuse my 6 PM brain yesterday...

brianjgeiger: Chances were also good that I was about to learn something interesting about a quirk in our system, but it's nice to know I can make a contribution every now and then. ;)

api_tests/nodes/filters/test_filters.py
caseyrollins: Both of the classes in this file are mixins. Do you mean that the mixins should be at the top of the file in `test_node_list.py`?

brianjgeiger: Nope, that's cool. I didn't scroll up. Feel free to ignore that comment.

sloria: Is this mixin necessary if there's only one subclass of it? We're trying to move away from these.

caseyrollins: Ah, to be as thorough as possible this should be subclassed on all views where nodes can be filtered (so user nodes and institution nodes as well). Looks like I forgot to do that though.

PR #7352, opened 2017-06-16
Use latest docker-sync, docker-compose version

birdbrained:
## Purpose

Update docker-compose and docker-sync files to use the latest versions

## Changes

- Remove volumes_from and replace with volumes in dockercompose.override.yml
- Remove dest in docker-sync.yml
- Update docker-compose.yml to version 3
- Add named volume "code"
- Remove commented out code

## Side effects

May interfere with production deploy?

## Ticket

No Ticket

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -26,8 +26,11 @@ services:
   api:
     volumes:
       - code:/code:nocopy
-
 # Additional configuration for development on services linked with osf.io
+
+# beat:
+#    volumes:
+#      - code:/code:nocopy
 
 #  fakecas:
 #    image: quay.io/centerforopenscience/fakecas:latest
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -33,41 +33,45 @@ services:
 #    image: quay.io/centerforopenscience/fakecas:latest
 #    command: fakecas -host=0.0.0.0:8080 -dbaddress=172.19.0.7:27017
 
-#  wb:
-#    volumes:
-#      - container:wb-sync
+  wb:
+    volumes:
+      - wb-sync:/code:nocopy
 
 #  wb_worker:
 #    volumes:
-#      - container:wb-sync
+#      - wb-sync:/code:nocopy
 
 #  wb_requirements:
 #    volumes:
-#      - container:wb-sync
+#      - wb-sync:/code:nocopy
 
 #  mfr:
 #    volumes:
-#      - container:mfr-sync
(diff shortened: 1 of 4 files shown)


docker-compose.override.yml
```diff
@@ -1,100 +1,33 @@
...
-#      - container:mfr-sync
-
-#  mfr_requirements:
-#    volumes_from:
-#      - container:mfr-sync
-
-#  preprints:
-#    volumes_from:
-#      - container:preprints-sync
-#########    Use this for ember-osf linked development (with docker-sync): #############
```
sloria: We should still included commented blocks to enable local development for ember-osf, waterbutler, mfr, etc.

sloria:
Ticket here: https://openscience.atlassian.net/browse/DEV-235?jql=text%20~%20%22docker-sync%22

sloria:
Just a drive-by comment for now. I don't have time to manually test this out yet; @birdbrained Can you test out local development with any of the linked projects locally, e.g. ember-osf, waterbutler?

sloria:
Why was waterbutler added as a submodule?

sloria:
Any reason we shouldn't use the default strategy, `native_osx`? https://github.com/EugenMayer/docker-sync/wiki/1.2-Upgrade-Guide#040

sloria:
Other than my 2 concerns above, this is looking good to go.

sloria:
I assume we'll need to update `docker-compose.ubuntu.yml` if this is merged.  

@TomBaxter , Would you be willing to update that after this is merged?

birdbrained:
Waterbutler submodule: oops.
We were using unison before. Idk if using native_osx would break linux (it seems it would detect? not sure) and also there seems to be some bugs with it still https://github.com/EugenMayer/docker-sync/wiki/8.1.-native_osx-sync-strategy-debugging-guide. From what I can tell unison has been plenty for what we've been doing with it so far. It might be something to explore separately.

sloria:
> We were using unison before. 

@icereval would know better, but I think we were using unison because it used to be the default. In any case, I don't think switching to native_osx. This can be merged without switching strategies.

TomBaxter:
I'll work on the ubuntu/override file. I'm wishing now I had called it linux instead of ubuntu.

sloria:
It's your file, man. Call it whatever you want.

TomBaxter:
@sloria @birdbrained 
https://github.com/CenterForOpenScience/osf.io/pull/7382  docker-compose.linux.yml is ready for review.

PR #7353, opened 2017-06-19
[OSF-7636] Fix bug in auto populating tags

Johnetordoff:
## Purpose

The feature to inherit tags from the parent node didn't account for tags that had just been added (tags which added on the backend, but not yet loaded in the frontend contextVars), this changes makes sure the contextVars always reflect reality.

## Changes

On a successful Ajax request adding or removing tags, the contextVars are changed to reflect this.

## Side effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/OSF-7636


Johnetordoff:
Yes

PR #7362, opened 2017-06-20
Highlighted Subjects [OSF-8041]

mfraezz:
## Purpose
Allow PreprintProviders to specify Subjects showing up on their landing page.

## Changes
* Add BooleanField to Subject
* Add property to PreprintProvider
* Add APIv2 route/view
* Add Admin module feature for managing Subjects
* Add tests

## Side effects
Requires change to `ember-preprints` before this feature is visible -- landing pages will need to hit `/v2/preprint_providers/<id>/taxonomies/highlighted` instead of `?filter[parent]=null`

## Ticket
[[OSF-8041]](https://openscience.atlassian.net/browse/OSF-8041)


osf/models/preprint_provider.py
```diff
@@ -46,6 +46,13 @@ def __unicode__(self):
         return '{} with id {}'.format(self.name, self.id)
 
     @property
+    def highlighted_subjects(self):
+        if self.subjects.filter(highlighted=True).exists():
+            return self.subjects.filter(highlighted=True).order_by('text')[:10]
+        else:
+            return sorted(self.top_level_subjects, key=lambda s: s.text)[:10]
```
sloria: Will this work with if `self.top_level_subjects` returns a queryset instead of a list?

mfraezz: Yes.
```
In [7]: bool(sorted(r.logs.all(), key=lambda x: x.date)[:10])
Out[7]: True
```

api/preprint_providers/views.py
```diff
@@ -263,6 +263,25 @@ def get_queryset(self):
...
+    view_name = 'highlighted-taxonomy-list'
+
+    required_read_scopes = [CoreScopes.ALWAYS_PUBLIC]
+    required_write_scopes = [CoreScopes.NULL]
+
+    serializer_class = TaxonomySerializer
+
+    def get_queryset(self):
+        provider = get_object_or_error(PreprintProvider, self.kwargs['provider_id'], display_name='PreprintProvider')
+        return provider.highlighted_subjects
```
cwisecarver: I think if filtering happens after `get_queryset` and filtering is written to expect a queryset then this could esplode because `provider.highlighted_subjects` might return a list and `list.filter(thing=stuff)` is going to throw an exception. Most places in the api I forced returning a queryset by doing a new query for the ids of the things after a `.values_list('id', flat=True)`. It's kind of bad but it's less bad than an exception.

brianjgeiger:
Conflicts

admin/subjects/views.py
```diff
@@ -0,0 +1,35 @@
...
+
+class SubjectListView(PermissionRequiredMixin, ListView):
+    model = Subject
+    permission_required = 'osf.view_subject'
+    paginate_by = 100
+    raise_exception = True
+
+    def get_queryset(self):
+        request = self.request.GET
+        qs = super(SubjectListView, self).get_queryset().order_by('text')
```
mfraezz: Currently, there are about 1200, and that will only grow with additional "custom taxonomies"

admin/subjects/views.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class SubjectListView(PermissionRequiredMixin, ListView):
+    model = Subject
+    permission_required = 'osf.view_subject'
+    paginate_by = 100
+    raise_exception = True
+
+    def get_queryset(self):
+        request = self.request.GET
```
mfraezz: Consistent variable name, but I'll change it

PR #7363, opened 2017-06-20
[OSF-8387] Add preprint_word field to preprint providers

hmoco:
## Purpose
Add a field that allows preprint providers to have language in their app that uses a different word than preprint to describe their documents. Current options are `paper` and `preprint` with additional language needed to be defined in the ember-osf app (defaulting to ` preprint `).

## Changes
Adding preprint_word field to preprint_provider, defaulting to using preprint.

## Ticket
https://openscience.atlassian.net/browse/OSF-8387


osf/models/preprint_provider.py
```diff
@@ -31,6 +31,7 @@ class PreprintProvider(ObjectIDMixin, BaseModel):
     share_source = models.CharField(blank=True, max_length=200)
     allow_submissions = models.BooleanField(default=True)
     additional_providers = fields.ArrayField(models.CharField(max_length=200), default=list, blank=True)
+    preprint_word = models.TextField(default='preprint')
```
caseyrollins: Just a passing comment: Seems like the `choices` option could be used here? See https://docs.djangoproject.com/en/1.11/ref/models/fields/#choices.

Also, `CharField` over `TextField` since this field won't be holding large amounts of text?

PR #7366, opened 2017-06-21
[Feature] Fix social profile long url wrapping [OSF-6722]

fabmiz:
## Purpose

Fix social profile rendering of long urls

## Changes

1) Make personal website url to span 2 table cols
2) Add css to fix long url wrapping issue on preexisting urls.

## Side effects

N/A

## Ticket

[OSF-6722](https://openscience.atlassian.net/browse/OSF-6722)

## QA considerations

- With a user with long social profile urls such as [dhyac](https://osf.io/dhyac/), please check the responsiveness of this table across screen sizes and browsers; not very sure how it renders in IE.

### DIFF
--- a/website/static/js/profile.js
+++ b/website/static/js/profile.js
@@ -676,6 +676,19 @@ var SocialViewModel = function(urls, modes, preventUnsaved) {
         return false;
     });
 
+    self.unmatchedUrls = ko.computed(function() {
+        var values = self.values();
+        var unmatched = [];
+        for (var i=0; i<self.values().length; i++) {
+            if (values[i].value) {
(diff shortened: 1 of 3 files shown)


website/static/js/profile.js
```diff
@@ -676,6 +676,19 @@ var SocialViewModel = function(urls, modes, preventUnsaved) {
         return false;
     });
 
+    self.unmatchedUrls = ko.computed(function() {
+        var values = self.values();
```
alexschiller: It doesn't look like values is mutated, but there is a mix-and-match of self.values() and values. Probably just pick one (I saw there was some old code that matched this style that it looks like you pulled from).

PR #7369, opened 2017-06-21
[Feature] Prevent component create with empty title [OSF-8162]

fabmiz:
## Purpose

Fix issue where a user submits empty (create component) 
form through enter key press.

## Changes

- Prevent addProjectPlugin controller from sending request if project/component
title is not provided.

## Side effects

N/A

## Ticket

https://openscience.atlassian.net/browse/OSF-8162

##QA Considerations
- In add component form, try submitting empty form by pressing enter. The form should not submit, check the network tab in console to make sure no request is made.
- Type spaces in the component title and submit form with enter key press. The form should not submit
- Type spaces followed by letters say "      awesome title", should save and submit just fine. The spaces in the title will be trimmed, the title "awesome title".

### DIFF
--- a/website/static/js/addProjectPlugin.js
+++ b/website/static/js/addProjectPlugin.js
@@ -178,8 +178,8 @@ var AddProject = {
                                 onkeyup: function(ev){
                                     var val = ev.target.value;
                                     var validTitle = val.trim().length > 0;
-                                    if (ev.which === 13) {
-                                         validTitle && ctrl.add();
+                                    if (ev.which === 13 && validTitle) {
+                                        ctrl.add();
                                     }
                                     ctrl.newProjectName(val);
                                     ctrl.isValid(validTitle);
--- a/website/static/js/addProjectPlugin.js
+++ b/website/static/js/addProjectPlugin.js
@@ -79,6 +79,10 @@ var AddProject = {
 
 
         self.add = function _add () {
+            var validTitle = self.newProjectName().trim().length > 0;
+            if (! validTitle) {
+              return;
+            }
             if (self.isAdding()) {
                 return;
             }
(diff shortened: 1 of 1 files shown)


alexschiller:
Also be sure to add QA notes for the various conditions you find.

website/static/js/addProjectPlugin.js
```diff
@@ -176,12 +176,13 @@ var AddProject = {
                             m('label[for="projectName].f-w-lg.text-bigger', 'Title'),
                             m('input[type="text"].form-control.project-name', {
                                 onkeyup: function(ev){
-                                    if (ev.which === 13) {
-                                         ctrl.add();
-                                    }
                                     var val = ev.target.value;
+                                    var validTitle = val.trim().length > 0;
+                                    if (ev.which === 13 && validTitle) {
```
brianjgeiger: I think this works, but I think it could probably be a tiny bit better. I would lean towards setting `ctrl.isValid()` early, then doing the `ctrl.add()` if it's return after that. Instead (or possibly in addition to) the validity check here for whether to `add()`, I would definitely put at the top of `add()` the check for title validity. If it's there, then as long as validity is being verified properly here, then no matter what rout the add gets called, it will have to have a valid title in order to add, and if it's close enough to to the top of the function, then the overhead for checking there vs. here is negligible. This future-proofs in case someone wants to put a new UI control that will call `add()`.

fabmiz: Got it! I will add the title validity check on top of `add()` and leave `onkeyup` event handler as was.

website/static/js/addProjectPlugin.js
```diff
@@ -79,6 +79,10 @@ var AddProject = {
 
 
         self.add = function _add () {
+            var validTitle = self.newProjectName().trim().length > 0;
```
fabmiz: Yes, that should work as well, I had not completely understood CR.

PR #7376, opened 2017-06-22
[feature] Fix preprint date modified [EOSF-709]

laurenrevere:
## Purpose

Currently the `date_modified` field of a preprint only updates if preprint specific fields are changed such as subjects or license. We want preprints to also count as modified if their nodes are modified (titles, files, abstracts, etc).

## Changes

This PR has two parts, one is making sure that preprints save when their nodes save, so that the date modified will be correct moving forward. The other is a migration to update all existing preprint `date_modified` fields to be the greater of the node's date modified or the preprint's.

## Side effects

Any side effects of saving a preprint will occur. 

## Ticket

https://openscience.atlassian.net/browse/EOSF-709


osf/models/node.py
```diff
@@ -2317,6 +2317,7 @@ def on_update(self, first_save, saved_fields):
             # .preprints wouldn't return a single deleted preprint
             for preprint in PreprintService.objects.filter(node_id=self.id, is_published=True):
                 enqueue_task(on_preprint_updated.s(preprint._id, update_share=update_share))
+                preprint.save()
```
alexschiller: Driveby comment; IIRC preprint.save() is going to call on_preprint_updated, so this will call on_preprint_updated twice every time a node is modified.

brianjgeiger: Could you verify that and, if need be, skip the enqueue task?

PR #7378, opened 2017-06-22
[#OSF-7930]Sorting OSF Meetings instances in Admin app

chennan47:
## Purpose
Sorting OSF Meetings instances in Admin app

## Changes
Now all the meetings in admin app should appear in alphabet order

## Ticket
https://openscience.atlassian.net/browse/OSF-7930


admin/meetings/views.py
```diff
@@ -23,7 +23,7 @@ class MeetingListView(PermissionRequiredMixin, ListView):
     raise_exception = True
 
     def get_queryset(self):
-        return Conference.find()
+        return sorted(Conference.find(), key=lambda meeting: meeting['name'])
```
brianjgeiger: Don't use python for this; you can do the ordering on the query itself.

admin/meetings/views.py
```diff
@@ -23,7 +23,7 @@ class MeetingListView(PermissionRequiredMixin, ListView):
     raise_exception = True
 
     def get_queryset(self):
-        return Conference.find()
+        return Conference.find().order_by('name')
```
alexschiller: Driveby: `.find` is a modm style wrapper for django functions: https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/models/base.py#L98

Just use `Conference.objects.all().order_by('name')`

chennan47: Done in d9db1be

PR #7379, opened 2017-06-22
Add default API view ordering to get rid of UnorderedObjectListWarning [#OSF-8136]

erinspace:
## Purpose

Travis logs for API tests (and local tests with the `-s` flag added) were throwing quite a few `UnorderedObjectListWarning` errors, which clued us in to the problem that many API views are indeed unordered, and so could return inconsistent results when paginating through them.

Let's add ordering to API views!

## Changes
- Add default `('-date_modified',)` ordering to many subclasses of`JSONAPIBaseView` 
- Add other orderings, mostly `('-id',)`, for other Views that aren't for a model with a `date_modified` field
- Update one taxonomies test to return results in the same way as the new view does for consistency

## Side effects
Pagination should now be consistent! So, none anticipated, but if anything was relying on inconsistent pagination (why would it?) this would go wonky.

## Ticket
https://openscience.atlassian.net/browse/OSF-8136

### DIFF
--- a/api/base/views.py
+++ b/api/base/views.py
@@ -798,6 +798,8 @@ class BaseContributorDetail(JSONAPIBaseView, generics.RetrieveAPIView):
 
 class BaseContributorList(JSONAPIBaseView, generics.ListAPIView, ListFilterMixin):
 
+    ordering = ('-date_modified',)
+
     def get_default_queryset(self):
         node = self.get_node()
 
@@ -824,6 +826,8 @@ class BaseNodeLinksDetail(JSONAPIBaseView, generics.RetrieveAPIView):
 
 
 class BaseNodeLinksList(JSONAPIBaseView, generics.ListAPIView):
+
+    ordering = ('-date_modified',)
 
     def get_queryset(self):
         auth = get_user_auth(self.request)
@@ -853,6 +857,8 @@ class BaseLinkedList(JSONAPIBaseView, generics.ListAPIView):
     serializer_class = None
     view_category = None
     view_name = None
+
+    ordering = ('-date_modified',)
 
     # TODO: This class no longer exists
     # model_class = Pointer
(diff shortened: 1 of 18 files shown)


api/base/views.py
```diff
@@ -31,6 +31,8 @@
 
 class JSONAPIBaseView(generics.GenericAPIView):
 
+    ordering = ('-date_modified',)
```
sloria: The base view class should not make assumptions about what model fields exist on the models; that's the subclasses' responsibility. So I think this can be removed.

api/nodes/views.py
```diff
@@ -2382,6 +2382,8 @@ class NodeProvidersList(JSONAPIBaseView, generics.ListAPIView, NodeMixin):
     view_category = 'nodes'
     view_name = 'node-providers'
 
+    ordering = ('-pk',)
```
sloria: Just curious: any reason this is different than the other views that do `('-id',)`?

erinspace: This view is for [NodeProviders](https://github.com/CenterForOpenScience/osf.io/blob/develop/api/nodes/views.py#L2246), which don't have an `id` field!

sloria: Ah, thanks. Looks like `NodeProvider.pk == Node._id`  though, which is probably not what we want to sort on. What if we store `NodeProvider.id` which is the equal to `node.id` and sort on that instead?

erinspace: That makes a lot of sense! Done!

api/users/views.py
```diff
@@ -395,6 +397,8 @@ class UserAddonAccountList(JSONAPIBaseView, generics.ListAPIView, UserMixin, Add
     view_category = 'users'
     view_name = 'user-external_accounts'
 
+    ordering = ('-date_last_refreshed',)
```
sloria: This should be changed to `date_modified` after https://openscience.atlassian.net/browse/OSF-7968 is implemented, but this is fine for now.

PR #7380, opened 2017-06-22
[feature] Replace '&amp;' with '&' in Profile Tabs [OSF-3945]

Nesiehr:
## Purpose

When users added '&' to their employment / education / social personal website link it resulted in being displayed as '&amp;'.
This replaces any '&amp;' with '&' in the personal website links and employment and education tabs on the profile page.

## Changes

Modified profile.js to pick up on where '&amp;' was being displayed on the employment, education, and social tabs where applicable.

## Side effects

If a user were to have '&amp;' in their personal website social link it would likely break the link by replacing that with '&'

## QA Considerations

There may be other links under social where a user could put  an ampersand, so if that is the case this will need to be updated.

## Ticket

https://openscience.atlassian.net/browse/OSF-3945

### DIFF
--- a/website/static/js/profile.js
+++ b/website/static/js/profile.js
@@ -731,6 +731,11 @@ SocialViewModel.prototype.unserialize = function(data) {
     var self = this;
     var websiteValue = [];
     $.each(data || {}, function(key, value) {
+        if (key === 'profileWebsites') {
+            value = value.map(function(website) {
+                return website.replace(/&amp;/gi, '&');
+            });
+        }
         if (ko.isObservable(self[key]) && key === 'profileWebsites') {
             if (value && value.length === 0) {
                 value.push(ko.observable('').extend({
@@ -924,6 +929,11 @@ ListViewModel.prototype.unserialize = function(data) {
         self.editable(false);
     }
     self.contents(ko.utils.arrayMap(data.contents || [], function (each) {
+        for (var attr in each) {
+            if (typeof each[attr] === 'string') {
+                each[attr] = each[attr].replace(/&amp;/gi, '&');
+            }
+        }
         return new self.ContentModel(self).unserialize(each);
     }));
 
--- a/website/static/js/profile.js
+++ b/website/static/js/profile.js
@@ -731,6 +731,9 @@ SocialViewModel.prototype.unserialize = function(data) {
     var self = this;
     var websiteValue = [];
     $.each(data || {}, function(key, value) {
+        if (key === 'scholar') {
+            value = value.replace(/&amp;/gi, '&');
+        }
         if (key === 'profileWebsites') {
             value = value.map(function(website) {
                 return website.replace(/&amp;/gi, '&');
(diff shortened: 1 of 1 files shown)


website/static/js/profile.js
```diff
@@ -731,6 +731,11 @@ SocialViewModel.prototype.unserialize = function(data) {
     var self = this;
     var websiteValue = [];
     $.each(data || {}, function(key, value) {
+        if (key === 'profileWebsites') {
```
fabmiz: Looks pretty good. One thing, it may not be in the ticket statement, in addition to personal website (profileWebsites), we should also unserialize all other social profile links. See this user, [osf.io/dhyac/](https://osf.io/dhyac/).  @alexschiller what do you think?

website/static/js/profile.js
```diff
@@ -924,6 +929,11 @@ ListViewModel.prototype.unserialize = function(data) {
         self.editable(false);
     }
     self.contents(ko.utils.arrayMap(data.contents || [], function (each) {
+        for (var attr in each) {
```
fabmiz: Nice! One thing, it's preferable to use for-in loops only for objects with string keys not arrays. There are other drawbacks of concern. Please check some good practices [here](https://hacks.mozilla.org/2015/04/es6-in-depth-iterators-and-the-for-of-loop/).

Nesiehr: each in this case is an object, so I'm using a for-in loop to loop over the attributes of an object. Is there a better way to do that? I believe for-each loops only work with arrays

alexschiller: That's probably fine then, I mentioned to @fabmiz to look into that in this case since I know weird things can happen with for...in.

website/static/js/profile.js
```diff
@@ -924,6 +929,11 @@ ListViewModel.prototype.unserialize = function(data) {
...
+            if (typeof each[attr] === 'string') {
+                each[attr] = each[attr].replace(/&amp;/gi, '&');
```
alexschiller: Use the $osf helper function `each[attr] = $osf.decodeText(each[attr])`

website/static/js/profile.js
```diff
@@ -731,6 +731,14 @@ SocialViewModel.prototype.unserialize = function(data) {
     var self = this;
     var websiteValue = [];
     $.each(data || {}, function(key, value) {
+        if (key === 'scholar') {
```
alexschiller: @fabmiz You mentioned scholar specifically-- are there others beyond just scholar and profileWebsites? I imagine all should be handled.

fabmiz: @alexschiller that change was done based on the example osf user I had shown him, but you are right. It should be done for all social profile links.

website/static/js/profile.js
```diff
@@ -731,6 +731,14 @@ SocialViewModel.prototype.unserialize = function(data) {
     var self = this;
     var websiteValue = [];
     $.each(data || {}, function(key, value) {
+        if (key === 'profileWebsites') {
+            value = value.map(function(website) {
+                return $osf.decodeText(website);
+            });
+        } else if (typeof value === 'string') {
```
alexschiller: decode text string checks

fabmiz: @Nesiehr, I think @alexschiller wanted to point out that this string check is also done in the $osf.decodeText helper, hence not needed here. See [osfHelpers.js#L369](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/osfHelpers.js#L369)

PR #7384, opened 2017-06-26
Feature/Add email template for adding contributors to a preprint OSF node [EOSF-412]

laurenrevere:
## Purpose

Currently, when there is a preprint associated with a node and someone is added to that node as a contributor, there is no indication in the email that the node also contains a preprint. 

## Changes

Added an email template to send when adding contributors to a node that contains a preprint.

## Side effects

All nodes with preprints will now receive this email instead of the default email when contributors are added. Both templates will need to be changed if the default template needs to be changed.

## Ticket

https://openscience.atlassian.net/browse/EOSF-412

### DIFF
--- a/website/templates/emails/contributor_added_preprint_node_from_osf.txt.mako
+++ b/website/templates/emails/contributor_added_preprint_node_from_osf.txt.mako
@@ -1,0 +1,23 @@
+<%!
+    from website import settings
+%>
+
+Hello ${user.fullname},
+
+${referrer_name + ' has added you' if referrer_name else 'You have been added'} as a contributor to the project "${node.title}" on the Open Science Framework: ${node.absolute_url}
+
+This project also has a public preprint, discoverable at: ${node.preprints.get_queryset()[0].absolute_url}
+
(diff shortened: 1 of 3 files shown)


website/templates/emails/contributor_added_preprint_node_from_osf.txt.mako
```diff
@@ -0,0 +1,23 @@
+<%!
+    from website import settings
+%>
+
+Hello ${user.fullname},
+
+${referrer_name + ' has added you' if referrer_name else 'You have been added'} as a contributor to the project "${node.title}" on the Open Science Framework: ${node.absolute_url}
+
+This project also has a public preprint, discoverable at: ${settings.DOMAIN.rstrip('/') + node.preprint_url}
```
binoculars: This should use the preprint's absolute URL: https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/models/preprint_service.py#L93 (branded domains)

laurenrevere: Oh. True. 👍

PR #7389, opened 2017-06-27
[OSF-8156] Change exception for updating a registration

Johnetordoff:
## Purpose

When attempting to PATCH/PUT a registration user receives a 400 error code with detail: "Registrations can only be turned from private to public." A better error code should be 403 Forbidden with a clearer message.

## Changes

Changes exception and message.

## Side effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/OSF-8156


brianjgeiger:
@Johnetordoff More merge conflicts if you have some time.

Johnetordoff:
@brianjgeiger became obsolete through the conflicting changes, however I think it's good to pick the associated unit test.

PR #7393, opened 2017-06-27
[feature] Use mako def for navbar [OSF-7562]

alexschiller:
## Purpose
HOME and MEETINGS flash when loading. Also writes urls for HOME and MEETINGS backend side

## Changes
No more flashy home 🔦 📸 💥 

## Side effects
na

## Ticket
https://openscience.atlassian.net/browse/OSF-7562

## QA Notes
Check the home page, and various other pages to ensure the following data works properly
1. The text of title (HOME, MEETINGS)
2. Clicking HOME or MEETINGS goes to "/" or "/meetings/" respectively
3. The support button goes to the proper support location

Check both the base meetings page and a specific meeting page as these were modified separately

### DIFF
--- a/website/templates/base.mako
+++ b/website/templates/base.mako
@@ -89,7 +89,7 @@
 
     <%namespace name="nav_file" file="nav.mako"/>
     <%block name="nav">
-        ${nav_file.nav()}
+        ${nav_file.nav(self.service_title, self.service_href, self.service_support)}
     </%block>
      ## TODO: shouldn't always have the watermark class
     ${self.content_wrap()}
@@ -245,6 +245,12 @@
 <%def name="alert()">
     <%include file="alert.mako"/>
 </%def>
+
+## Default nav bar service values
+<%def name="service_title()">HOME</%def>
+<%def name="service_href()">/</%def>
+<%def name="service_support()">/support/</%def>
+
 
 <%def name="content_wrap()">
     <div class="watermarked">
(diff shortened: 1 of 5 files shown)


website/templates/base.mako
```diff
@@ -246,6 +246,12 @@
     <%include file="alert.mako"/>
 </%def>
 
+## Default nav bar service values
+<%def name="service_title()">HOME</%def>
+<%def name="service_href()">/</%def>
+<%def name="service_support()">/support/</%def>
```
sloria: These names could be more consistent.

How about `service_name`, `service_url`, and `service_support_url`?

website/templates/nav.mako
```diff
@@ -13,9 +13,9 @@
             </button>
         <a class="navbar-brand" href="/" aria-label="Go home"><span class="osf-navbar-logo"></span></a>
         <div class="service-name">
-            <a data-bind="attr: {href: osfServices[currentService].href}">
+            <a href="${service_href()}">
                 <span class="hidden-xs"> OSF </span>
-                <span><strong data-bind="text: osfServices[currentService].name"></strong></span>
+                <span><strong>${service_title()}</strong></span>
```
sloria: I think I've figured out a way we can avoid this weird function-passing. It involves changing this nav file from using a `block` to a `def`. I'll send a PR shortly.

PR #7405, opened 2017-06-29
[feature] Embargoed Resources Custom Forbidden Page [OSF-5640]

Nesiehr:
## Purpose

Inform unauthenticated users of when the embargo on a registration will be lifted by customizing the forbidden page.

## Changes

Modified decorators.py to include a check for embargos that aren't pending. If it is an embargo it then provides a custom forbidden response detailing that the page is an embargo and when the embargo will end.

Before:
before
After:
after

Added tests to ensure that pending embargo projects still return the normal forbidden response and that embargoed projects return the custom forbidden response.

## Side effects

None known.

## QA Considerations

1. Pending embargo (does not show the error message)
2. Components of an embargo (show the error message)
3. When the embargo ends (becomes public as expected)

## Ticket

https://openscience.atlassian.net/browse/OSF-5640


tests/test_registrations/test_embargoes.py
```diff
@@ -1096,3 +1097,46 @@ def test_non_contributor_GET_disapproval_returns_HTTPError(self):
...
+            auth=self.non_contrib.auth,
+            expect_errors=True
+        )
+
+        assert_true(self.registration_embargo.is_pending_embargo)
+        assert_equal(res.status_code, 403)
+        assert_in('Forbidden', res.body)
+
+    def test_embargo_non_contrib_returns_resource_under_embargo(self):
+        approval_token = self.registration_embargo.embargo.approval_state[self.user._id]['approval_token']
```
fabmiz: @alexschiller, correct me if I am wrong; for performance sake, I think it's safe to assume that the embargo is already approved when a non_contrib attempts to view it. If so, both tests can be merged into one and the embargo approving part can be either added to setUp() or in the beginning of the new test.

website/project/decorators.py
```diff
@@ -249,6 +257,7 @@ def wrapped(*args, **kwargs):
 
     return wrapper
 
+
```
fabmiz: Minor thing; extra line.

tests/test_registrations/test_embargoes.py
alexschiller: @fabmiz It looks like it is needed to me-- the first test is for pending embargo, which should throw Forbidden, the second test is for after the embargo is approved, which then throws the custom embargo error.

website/project/decorators.py
```diff
@@ -179,6 +179,14 @@ def check_can_access(node, user, key=None, api_node=None):
     if not node.can_view(Auth(user=user)) and api_node != node:
         if key in node.private_link_keys_deleted:
             status.push_status_message('The view-only links you used are expired.', trust=False)
+        elif node.embargo and node.embargo_end_date is not False:
+            raise HTTPError(
+                http.FORBIDDEN,
+                data={
```
alexschiller: Instead of raising the HTTPError here, define `error_data = {'message_long': ('User... blah blah blah')}` around line 180, then overwrite the data here.

website/project/decorators.py
```diff
@@ -179,6 +179,14 @@ def check_can_access(node, user, key=None, api_node=None):
     if not node.can_view(Auth(user=user)) and api_node != node:
         if key in node.private_link_keys_deleted:
             status.push_status_message('The view-only links you used are expired.', trust=False)
+        elif node.embargo and node.embargo_end_date is not False:
```
alexschiller: Per discussion, this can probably ignore date and use `is_pending_embargo` instead

alexschiller:
Can you add QA considerations to recommend testing:
1. Pending embargo (does not show the error message)
2. Components of an embargo (show the error message)
3. When the embargo ends (becomes public as expected)

website/project/decorators.py
Nesiehr: What is the reason for this and how does that work?

website/project/decorators.py
Nesiehr: Not extra, pretty sure that is needed for flake8

website/project/decorators.py
alexschiller: @Nesiehr HTTPError is called regardless with the only thing that is changing being the data input into it, just a bit cleaner when done like below:
```
def check_can_access(node, user, key=None, api_node=None):
    """View helper that returns whether a given user can access a node.
    If ``user`` is None, returns False.

    :rtype: boolean
    :raises: HTTPError (403) if user cannot access the node
    """
    if user is None:
        return False
    if not node.can_view(Auth(user=user)) and api_node != node:
        error_data = {
            'message_long': ('User has restricted access to this page. '
            'If this should not have occurred and the issue persists, please report it to '
            '<a href="mailto:support@osf.io">support@osf.io</a>.')
        }
        if key in node.private_link_keys_deleted:
            status.push_status_message('The view-only links you used are expired.', trust=False)
        elif node.embargo and node.embargo_end_date is not False:
            error_data['message_short'] = 'Resource under embargo'
            error_data['message_long'] = 'This resource is currently under embargo, please check back when it opens {}.'.format(node.embargo_end_date.strftime('%A, %b. %d, %Y'))
            }
        raise HTTPError(http.FORBIDDEN, data=error_data)
    return True
```

brianjgeiger:
@Nesiehr Could you add before/after screenshots please?

PR #7409, opened 2017-06-29
[Feature] Bitbucket Addon [OSF-8192]

mfraezz:
Updates #6954 

> ## Purpose
> 
> Add support for a read-only bitbucket OSF add-on.  Still targets `develop-backup` and needs to be upgraded to work with DjangOSF.
> 
> ## Changes
> 
> Adds a whole new provider in `website/addons/bitbucket`.  All changes outside that directory go directly towards supporting the new addon.
> 
> ## Side effects
> 
> Unavoidable grief.
> 
> ## Ticket
> 
> [#SVCS-268] - [JIRA](https://openscience.atlassian.net/browse/SVCS-268)
> [#OSF-8192] - [JIRA](https://openscience.atlassian.net/browse/OSF-8192)
> 
> Entangled with WB PR: https://github.com/CenterForOpenScience/waterbutler/pull/198

# MIGRATION WARNING
Not only does this share a migration number (`0042`) with #7418 , that migration has a conflicting action -- altering possible types for `BaseFileNode`. 
### **Running `makemigrations --merge` will NOT be sufficient.** 
Instead, whichever PR goes in last will need the migration recreated before being merged.

PR #5929, opened 2016-07-06
[WIP] GitLab addon

danielneis:
## Purpose

Implement the GitLab addon according to the GitLab Integration Grant Spec at https://gist.github.com/felliott/cc3d48a492c174545b7f
## Changes

"This add-on will allow users to specify a GitLab instance to communicate with, and then connect their GitLab account to their OSF account via an OAuth token. Users may then connect a project for which they have admin rights to one of their GitLab-hosted repositories. Users should also be able to change which repository is linked to a given project at any time and revoke permissions from any or all projects as desired."
## Side effects

```
$ cat website/addons/gitlab/requirements.txt
cachecontrol==0.10.2
pyapi-gitlab==7.8.5
python-magic==0.4.6
```
## Roadmap
- [x] Allow user to connect GitLab.com account to project
- [x] Fetch repositories from GitLab.com
- [x] Associate a repository from GitLab.com to a project
- [x] Browse files from default branch from repository
- [x] Browse files from any branch from repository
- [x] Create repository
- [x] List repository branches
- [x] Download default repository branch as tar file
- [x] Download any repository branch as tar file
- [x] Open repository on GitLab (link to correct branch)
- [x] File action: Download after change repository branch
- [x] File action: Download for default branch
- [x] File action: Delete from default branch
- [x] File action: Delete after change repository branch
- [x] File action: View
- [x] File action:  View on GitLab.com
- [ ] File action: Rename
- [x] Create folder on default branch (on root folder, and inside other folders)
- [x] Create folder on other branches (on root folder, and inside other folders)
- [x] Upload file to root folder on any branch
- [x] Upload file to any folder on non-default branch
- [ ] Upload file to any folder on default branch
- [x] Allow user use another GitLab instance in projects
- [ ] Tests
- [ ] Documentation

PR #7418, opened 2017-07-03
GitLab Addon [#OSF-8169]

mfraezz:
Updates #5929 

> ## Purpose
> 
> Implement the GitLab addon according to the GitLab Integration Grant Spec at https://gist.github.com/felliott/cc3d48a492c174545b7f
> ## Changes
> 
> "This add-on will allow users to specify a GitLab instance to communicate with, and then connect their GitLab account to their OSF account via an OAuth token. Users may then connect a project for which they have admin rights to one of their GitLab-hosted repositories. Users should also be able to change which repository is linked to a given project at any time and revoke permissions from any or all projects as desired."
> 

## Side effects
None expected

~~# MIGRATION WARNING~~
~~Not only does this share a migration number (`0042`) with #7409 , that migration has a conflicting action -- altering possible types for `BaseFileNode`.~~
~~### **Running `makemigrations --merge` will NOT be sufficient.**~~
~~Instead, whichever PR goes in last will need the migration recreated before being merged.~~

~~This has been fixed.~~ This is true again for #7653 

## Ticket
[[OSF-8169]](https://openscience.atlassian.net/browse/OSF-8169)


addons.json
```diff
@@ -25,7 +26,9 @@
         "forward": "full",
         "googledrive": "partial",
         "github": "partial",
+        "gitlab": "partial",
         "owncloud": "partial",
+        "gitlab": "partial",
```
sloria: This is a duplicate entry.

PR #7428, opened 2017-07-06
[Feature]admin prod style [OSF-8140]

alexschiller:
## Purpose
Take 2. First time I was dumb. Make prod look different.

## Changes
It is ugly and I don't like it. Special case checking for staging, localhost, and test. If not found in the URL colorize it. Positively special casing admin is more frustrating since django don't give you access to functions. SAD.

screen shot 2017-07-06 at 12 10 51 pm
screen shot 2017-07-06 at 12 10 15 pm
screen shot 2017-07-06 at 12 11 40 pm
screen shot 2017-07-06 at 12 12 00 pm

## Side effects
NA

## Ticket
https://openscience.atlassian.net/browse/OSF-8140?devStatusDetailDialog=pullrequest


admin/templates/base.html
```diff
@@ -209,3 +209,10 @@
 
   </body>
 </html>
+
+{% if 'test' not in request.META.HTTP_HOST and 'staging' not in request.META.HTTP_HOST and 'localhost' not in request.META.HTTP_HOST %}
```
brianjgeiger: Hmmmmm. Concerns: 

- Supposedly request.META.HTTP_HOST went away between 1.5 and 1.6, though it clearly seems to exist in 1.11, so it's prooobably fine. 
- This is less brittle than the last implementation, but more brittle than I'd like. How about something like `contains 'admin.osf.io' and not contains '-admin.osf.io'`?
- Proxying may be a problem, but we'll find out if that's a real problem in integration testing or when it goes live.

PR #7430, opened 2017-07-07
[OSF-8178] Multi-field filtering is broken on staging and production

Johnetordoff:
## Purpose

OR searches (using multiple fields in a filter statement like "?filter[title,description]=Octopus") haven't been working for the APIs NodeList view, this fix aims to make OR statements work.

## Changes

Changes inheritance so NodeList inherits the ODMFilterMixin behavior allowing OR statements.

## Side effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/OSF-8178


api/base/filters.py
```diff
@@ -540,13 +540,27 @@ def param_queryset(self, query_params, default_queryset):
...
-                    for operation in operations:
-                        if isinstance(queryset, list):
-                            queryset = self.get_filtered_queryset(field_name, operation, queryset)
-                        else:
-                            queryset = self.filter_by_field(queryset, field_name, operation)
+                match = self.QUERY_PATTERN.match(key)
+                fields = match.groupdict()['fields']
+                if len(re.findall(self.FILTER_FIELDS, fields)) > 1:  # This indicates an OR statement
+                    sub_query = Node.objects.none()
+                    for field_name, data in field_names.iteritems():
```
brianjgeiger: This section looks very similar to L557-L563. Can this be dried and possibly independently tested?

Johnetordoff:
@brianjgeiger I've bumped into another problem because of this line: https://github.com/CenterForOpenScience/osf.io/blob/develop/api/addons/views.py#L98. get_default_queryset returns a list and not a queryset I can't filter it properly without special casing it, can I just remove filtering for AddonsList? Or is it worth creating a new model for the addons so it can have a proper queryset? or what?

api/addons/views.py
```diff
@@ -99,3 +101,26 @@ def get_default_queryset(self):
...
+        if filters:
+            for key, field_names in filters.iteritems():
+                match = self.QUERY_PATTERN.match(key)
+                fields = match.groupdict()['fields']
+                statement = len(re.findall(self.FILTER_FIELDS, fields)) > 1  # This indicates an OR statement
+                sub_query = set() if statement else set(default_queryset)
+                for field_name, data in field_names.iteritems():
+                    operations = data if isinstance(data, list) else [data]
+                    for operation in operations:
+                        print self.get_filtered_queryset(field_name, operation, list(default_queryset))
```
alexschiller: `print`

api_tests/nodes/views/test_node_list.py
```diff
@@ -728,6 +726,21 @@ def test_unpublished_preprint_in_preprint_false_filter_results(self, app, user_o
         assert preprint.node._id not in ids
         assert unpublished.node._id in ids
 
+    def test_nodes_list_filter_multiple_field(self, app, public_project_one, public_project_two, public_project_three, user_one):
+
+        url = "/{}nodes/?filter[title,description]=One".format(API_BASE)
```
alexschiller: Single quotes

api_tests/nodes/views/test_node_list.py
```diff
@@ -1055,6 +1068,9 @@ def test_create_project_errors(self, app, user_one, title, description, category
         assert res.json['errors'][0]['detail'] == 'Title cannot exceed 200 characters.'
 
 
+
```
alexschiller: Bunch of extra lines here

PR #7437, opened 2017-07-10
[Feature] Allow delete atmention when editing old comments [OSF-8133]

fabmiz:
## Purpose

Allow delete atmention when editing old comments

## Changes

Set `contenteditable` to `true` when displaying old comments from
database.

## Ticket

https://openscience.atlassian.net/browse/OSF-8133


brianjgeiger:
The PR description seems to be the opposite of what was changed, but it might just be that I don't fully get it, so could you clarify so I understand what's happening?

PR #7443, opened 2017-07-10
[Feature] api_tests/guids to pytest [OSF-8256]

fabmiz:
## Purpose

Convert api_tests/guids from nose to pytest

## Changes

Re-write tests for guids `detail` api endpoint with pytest.

## Ticket

[OSF-8255](https://openscience.atlassian.net/browse/OSF-8256)


api_tests/guids/views/test_guid_detail.py
```diff
@@ -1,53 +1,70 @@
...
 import pytest
 
 from addons.osfstorage.models import OsfStorageFile
-from website.settings import API_DOMAIN
-
-from tests.base import ApiTestCase
-from osf_tests.factories import (AuthUserFactory, ProjectFactory, RegistrationFactory,
-                                 CommentFactory, CollectionFactory, PrivateLinkFactory)
 from addons.wiki.tests.factories import NodeWikiFactory
+from api.base.settings.defaults import API_BASE
```
Nesiehr: I don't think you need a line break here

api_tests/guids/views/test_guid_detail.py
```diff
@@ -1,53 +1,70 @@
...
+    @pytest.fixture()
+    def user(self):
+        return AuthUserFactory()
+
+    @pytest.fixture()
+    def project(self):
+        return ProjectFactory()
+
+    @pytest.fixture()
+    def _add_private_link(self, project):
```
Nesiehr: I typically try to avoid starting a name with a "_", but I don't think it matters that much

Nesiehr:
LGTM

api_tests/guids/views/test_guid_detail.py
```diff
@@ -56,36 +72,36 @@ def test_redirect_to_file_view(self):
...
+        res = app.get(url, auth=user.auth, expect_errors=True)
+        assert res.status_code == 404
 
-    def test_redirect_when_viewing_private_project_through_view_only_link(self):
-        project = ProjectFactory()
-        view_only_link = self._add_private_link(project)
+    def test_redirects_through_view_only_link(self, app, project, add_private_link, user):
+
+        #test_redirect_when_viewing_private_project_through_view_only_link
+        view_only_link = add_private_link()
```
alexschiller: add_private_link seems to only be used here, just do it inline:

```
view_only_link = PrivateLinkFactory(anonymous=False)
view_only_link.nodes.add(project)
view_only_link.save()
```

api_tests/guids/views/test_guid_detail.py
```diff
@@ -1,53 +1,69 @@
...
+    @pytest.fixture()
+    def user(self):
+        return AuthUserFactory()
+
+    @pytest.fixture()
+    def project(self):
+        return ProjectFactory()
+
+    @pytest.fixture()
+    def add_private_link(self, project):
```
alexschiller: This seems overly complicated. Is there any reason it isn't inline (see line 97 below)

api_tests/guids/views/test_guid_detail.py
```diff
@@ -1,53 +1,69 @@
...
+        def private_link(anonymous=False):
+            view_only_link = PrivateLinkFactory(anonymous=anonymous)
+            view_only_link.nodes.add(project)
+            view_only_link.save()
+            return view_only_link
+        return private_link
+
+    @pytest.fixture()
```
alexschiller: Is anything gained by this being a fixture? It looks like it could either be a simple function (or just inline since it isn't that complicated)

```
def format_url(id):
    return '/{}guids/{}/'.format(API_BASE, id)
```

api_tests/guids/views/test_guid_detail.py
alexschiller: @Nesiehr Really the only time I use _method_naming is to signify that something is "protected" or not really intended to be called externally. Nothing --stops-- you from calling it, just as sort of a "Hey, might want to see why this is _ named." So in this case, I don't think there is a good reason. 👍

api_tests/guids/views/test_guid_detail.py
fabmiz: I agree, I was overusing fixtures.

api_tests/guids/views/test_guid_detail.py
fabmiz: Not really, it can be left a simple function. overusing fixtures.

Issue #479, opened 2014-04-07
Wiki pages downloadable


TimErrington:
It would be nice to have wiki pages downloadable (in at least a pdf format).

And if a component has multiple wiki pages associated with it, there could be an option so they can all be downloaded (if the user wanted) into a multiple page pdf document

PR #7445, opened 2017-07-11
[Feature] Add Export Wiki to PDF

sonnenschein123:
## Purpose
Users might like:
( Uploading project data, enhancing it with additional information, finish a project, end of the story? )

Sometimes it's nice to offer the option to download an overview of all surrounding wiki information, just by one klick.
With this dialog it's possible to export current wiki page or all wiki project pages via PDF. To save, to share, to print and add handwritten notes or annotations ...

## Changes
Add WeasyPrint==0.36 to requirements.txt and set html5lib==0.9999999. 

## Side effects
WeasyPrint uses html5lib==0.999999999 but it OSF has some problems with it. So I chose 0.9999999 and it work.

## Ticket
[Ticket 479](https://github.com/CenterForOpenScience/osf.io/issues/479)

### DIFF
--- a/addons/wiki/models.py
+++ b/addons/wiki/models.py
@@ -18,6 +18,8 @@ from addons.wiki import utils as wiki_utils
 from website.exceptions import NodeStateError
 from website.util import api_v2_url
 
+from weasyprint import HTML
+
 from .exceptions import (
     NameEmptyError,
     NameInvalidError,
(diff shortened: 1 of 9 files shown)


brianjgeiger:
Thank your for the Pull Request! Unfortunately, I don't think we can merge this, because it has the potential to cause a lot of strain on our main server process. We might be able to do something like this in the Modular File Renderer https://github.com/CenterForOpenScience/modular-file-renderer but that is still somewhat difficult as we have a similar load problem there with Word docs and similar. It's better in the MFR because it's a separate service, so even if it overtaxes itself, the entire site won't go down.

Ideally, we would use a purely javascript solution which we could integrate into our markdown-it renderer. Something like that would keep with our existing infrastructure and be scalable.

sonnenschein123:
Hello @brianjgeiger,
I understand that you are worry about your workload. My experience with javascript solutions on the client site is, that they are often get blocked by adblocker.

PR #7447, opened 2017-07-11
[OSF-7784] Admin app: accessing list of user's permissions 502s

Johnetordoff:
## Purpose

It was difficult to see what permission groups admins had and changing them required a somewhat circuitous route to the correct menu. This fix makes everything accessible from the admin page.

## Changes

Created a new "AdminAdmin" model which is an Admin model for the Admin site that subclasses the AdminProfile Model with the admin.ModelAdmin model so admin data can be formatted properly for the admin page (admin/osf/adminprofile/). 

## Side effects

Incidentally migrations had lead to fields being represented in the AdminNode model and the OSFUser model to become out of sync leading to 502's and other problems caused Django trying to generate html for fields which no longer exsisted. I decided to add some code the should "migration-proof" the models.

## Tests

It's difficult to write meaningful tests for these changes because despite the fact that they are back end models they directly effect the front-end design choices and nothing else and so much of the admin site is automatically generated by Django I think that's good coverage.

### Before
screen shot 2017-07-28 at 9 35 05 am
Clicking the user links causes a 502 or timesout, with no other results.

### After

admin_gif

Permissions listed and clicking them allows an overwrite.

## Ticket

https://openscience.atlassian.net/browse/OSF-7784


Johnetordoff:
@brianjgeiger added the gifs, should this be demoed too?

PR #7454, opened 2017-07-12
[feature] Add context manager for disabling auto_now updates [OSF-8222]

caseyrollins:
#### Purpose
- Adds a context manager to prevent `auto_now` fields from being updated during migrations and scripts.

#### Changes
- Adds `disable_auto_now_fields` context manager and test in `test_utils.py`. 

#### Side Effects
- None expected.

#### Ticket
- [OSF-8222](https://openscience.atlassian.net/browse/OSF-8222)

### DIFF
--- a/osf/utils/migrations.py
+++ b/osf/utils/migrations.py
@@ -2,11 +2,30 @@ import os
 import json
 import logging
 
+from contextlib import contextmanager
+
 from website import settings
 from osf.models import NodeLicense, MetaSchema
 from website.project.metadata.schemas import OSF_META_SCHEMAS
 
 logger = logging.getLogger(__file__)
+
+
+@contextmanager
+def disable_auto_now_fields(model):
+    """
+    Context manager to disable updates of all auto_now fields for a given model.
+
+    """
+    for field in model._meta.get_fields():
+        if hasattr(field, 'auto_now') and field.auto_now:
+            field.auto_now = False
+    try:
+        yield
+    finally:
+        for field in model._meta.get_fields():
+            if hasattr(field, 'auto_now') and not field.auto_now:
+                field.auto_now = True
 
 
 def ensure_licenses(*args, **kwargs):
--- a/osf_tests/test_utils.py
+++ b/osf_tests/test_utils.py
@@ -1,0 +1,33 @@
+import pytest
+
+from osf.models import Node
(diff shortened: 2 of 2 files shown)


osf/utils/migrations.py
```diff
@@ -2,13 +2,32 @@
...
 logger = logging.getLogger(__file__)
 
 
+@contextmanager
+def disable_auto_now_fields(model):
+    """
+    Context manager to disable updates of all auto_now fields for a given model.
+
+    """
+    for field in model._meta.get_fields():
```
caseyrollins: I wasn't sure if this should disable all `auto_now` fields, or just specified fields. I went with disabling all fields, since that would be simplest to use, but this func could also be:
```
@contextmanager
def disable_auto_now_fields(model, fields):
    for field in model._meta.get_fields():
        if field.name in fields:
            ...
```

osf/utils/migrations.py
```diff
@@ -2,13 +2,32 @@
...
+        if hasattr(field, 'auto_now') and field.auto_now:
```
caseyrollins: Similar to above: I could also disable `auto_now_add` fields here, but wasn't sure.

osf/utils/migrations.py
sloria: It would be nice to optionally pass fields:

```python
with disable_auto_now_fields(SomeModel):  # disable all auto_now fields

with disable_auto_now_fields(SomeModel, ['date_modified']):  # disable specified fields
```

osf/utils/migrations.py
sloria: No, let's only disable `auto_now` fields for now.

osf/utils/migrations.py
sloria: That said, that probably isn't necessary for the time being (YAGNI).

osf/utils/migrations.py
```diff
@@ -2,13 +2,32 @@
...
+            field.auto_now = False
+    try:
+        yield
+    finally:
```
sloria: 👍

sloria:
👍 Documented in [the scratchpad](https://openscience.atlassian.net/wiki/spaces/IN/pages/56365109/The+Django+LE+Scratchpad+Of+Useful+Learnings)

PR #7455, opened 2017-07-12
[feature] Fix email_read Core Scope [No Ticket]

caseyrollins:
#### Purpose
- Use the correct string when checking request scopes... h/t @cwisecarver @icereval 

#### Ticket
- None

### DIFF
--- a/api/users/serializers.py
+++ b/api/users/serializers.py
@@ -156,7 +156,7 @@ class UserDetailSerializer(UserSerializer):
     id = IDField(source='_id', required=True)
 
 
-class ReadEmailUserSerializer(UserDetailSerializer):
+class ReadEmailUserDetailSerializer(UserDetailSerializer):
 
     email = ser.CharField(source='username', read_only=True)
 
--- a/api/users/serializers.py
+++ b/api/users/serializers.py
@@ -8,7 +8,7 @@ from osf.models import OSFUser as User
 
 from api.base.serializers import (
     JSONAPISerializer, LinksField, RelationshipField, DevOnly, IDField, TypeField, ListDictField,
-    DateByVersion, EmailScopeRequired,
+    DateByVersion,
 )
 from api.base.utils import absolute_reverse, get_user_auth
 
@@ -34,7 +34,6 @@ class UserSerializer(JSONAPISerializer):
     timezone = HideIfDisabled(ser.CharField(required=False, help_text="User's timezone, e.g. 'Etc/UTC"))
     locale = HideIfDisabled(ser.CharField(required=False, help_text="User's locale, e.g.  'en_US'"))
     social = ListDictField(required=False)
-    email = EmailScopeRequired(ser.CharField(source='username', read_only=True))
 
     links = HideIfDisabled(LinksField(
         {
@@ -155,6 +154,11 @@ class UserDetailSerializer(UserSerializer):
     Overrides UserSerializer to make id required.
     """
     id = IDField(source='_id', required=True)
+
+
+class ReadEmailUserSerializer(UserDetailSerializer):
+
+    email = ser.CharField(source='username', read_only=True)
 
 
 class RelatedInstitution(JSONAPIRelationshipSerializer):
--- a/api/users/views.py
+++ b/api/users/views.py
@@ -22,9 +22,9 @@ from api.users.serializers import (UserAddonSettingsSerializer,
                                    UserDetailSerializer,
                                    UserInstitutionsRelationshipSerializer,
                                    UserSerializer,
-                                   ReadEmailUserSerializer,)
+                                   ReadEmailUserDetailSerializer,)
 from django.contrib.auth.models import AnonymousUser
-from framework.auth.oauth_scopes import CoreScopes, ComposedScopes
+from framework.auth.oauth_scopes import CoreScopes, normalize_scopes
 from modularodm import Q as MQ
 from django.db.models import Q
 from rest_framework import permissions as drf_permissions
@@ -256,8 +256,9 @@ class UserDetail(JSONAPIBaseView, generics.RetrieveUpdateAPIView, UserMixin):
     def get_serializer_class(self):
         if self.request.auth:
             scopes = self.request.auth.attributes['accessTokenScope']
-            if ComposedScopes.USER_EMAIL_READ in scopes and self.request.user == self.get_user():
-                return ReadEmailUserSerializer
+            if (CoreScopes.USER_EMAIL_READ in normalize_scopes(scopes)
+                and self.request.user == self.get_user()):
+                return ReadEmailUserDetailSerializer
         return UserDetailSerializer
 
     # overrides RetrieveAPIView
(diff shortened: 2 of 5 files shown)


cwisecarver:
This is another way to do it. https://stackoverflow.com/questions/19128793/per-field-permission-in-django-rest-framework?answertab=votes#tab-top

```
class UserSerializer(stuff):
    # all the stuff

class ReadEmailUserSerializer(UserSerializer):
   email = Email
```

caseyrollins:
Went with @cwisecarver's proposed solution, and also removed the `users_email/user_email/read_email` magic string in favor of a constant (also per @cwisecarver...).

api/users/views.py
```diff
@@ -252,6 +253,13 @@ class UserDetail(JSONAPIBaseView, generics.RetrieveUpdateAPIView, UserMixin):
 
     serializer_class = UserDetailSerializer
 
+    def get_serializer_class(self):
+        if self.request.auth:
+            scopes = self.request.auth.attributes['accessTokenScope']
+            if ComposedScopes.USER_EMAIL_READ in scopes and self.request.user == self.get_user():
```
cwisecarver: h/t @icereval for pointing me to `normalize_scopes` which [protects us from future changes](https://github.com/caseyrollins/osf.io/blob/54876e3cb804a3295e64d54abfa1215224bde782/framework/auth/oauth_scopes.py#L272)

api_tests/base/test_auth.py
```diff
@@ -248,26 +248,26 @@ def test_full_write_scope_can_read_guid_view_and_user_cannot_view_project(self,
 
     @mock.patch('framework.auth.cas.CasClient.profile')
     def test_user_email_scope_can_read_email(self, mock_user_info):
-        mock_user_info.return_value = self._scoped_response(['osf.users.user_email', 'osf.users.profile_read'])
+        mock_user_info.return_value = self._scoped_response([oauth_scopes.ComposedScopes.USER_EMAIL_READ, 'osf.users.profile_read'])
```
cwisecarver: This is pretending to be the CAS side so I'd use strings here.

api/users/serializers.py
```diff
@@ -157,6 +156,11 @@ class UserDetailSerializer(UserSerializer):
     id = IDField(source='_id', required=True)
 
 
+class ReadEmailUserSerializer(UserDetailSerializer):
```
cwisecarver: I'd call this `ReadEmailUserDetailSerializer` just to pick nits.

framework/auth/oauth_scopes.py
```diff
@@ -210,9 +210,9 @@ class ComposedScopes(object):
 
 if settings.DEV_MODE:
     public_scopes.update({
-        'osf.users.email_read': scope(parts_=frozenset(ComposedScopes.USERS_EMAIL),
-                                    description='Read your primary email address.',
-                                    is_public=True),
+        CoreScopes.USER_EMAIL_READ: scope(parts_=frozenset(ComposedScopes.USER_EMAIL_READ),
```
cwisecarver: This should also be the string literal so it's prefixed with `osf.`

api_tests/base/test_auth.py
caseyrollins: I got a bit aggressive with the string removal...

framework/auth/oauth_scopes.py
```diff
@@ -124,7 +124,7 @@ class ComposedScopes(object):
     USERS_READ = (CoreScopes.USERS_READ, CoreScopes.ALWAYS_PUBLIC, )
     USERS_WRITE = USERS_READ + (CoreScopes.USERS_WRITE,)
     USERS_CREATE = USERS_READ + (CoreScopes.USERS_CREATE, )
-    USERS_EMAIL = (CoreScopes.USERS_EMAIL, )
+    USER_EMAIL_READ =  (CoreScopes.USERS_READ, CoreScopes.USER_EMAIL_READ, )
```
icereval: Please keep this `(CoreScopes.USERS_EMAIL, )` for the time being.

Documentation should be updated to indicate this is additive and requires `USERS_READ` aka `osf.users.profile_read`.

The reason here is we won't get an explicit bullet point when the user authorizes the scopes on CAS and it should be spelled out you are releasing your email address and profile information.

caseyrollins: I'm fine with reverting this, but couldn't [`description`](https://github.com/caseyrollins/osf.io/blob/8577473cda024defad5fb37ae52abfede05a36b4/framework/auth/oauth_scopes.py#L214) just be updated to be more explicit (i.e. "Read your profile data and primary email address")?

icereval: its easier for us to justify adding permissions later if we decide that is the correct case, rather than retracting them and breaking existing integrations...

cwisecarver: I'd say it's easier to put things together later than to pull them apart. 

Either way, if it’s going to include USER_READ it needs to include the composed one like USERS_WRITE and USERS_CREATE above it does, right? So ALWAYS_PUBLIC comes too?

caseyrollins: @cwisecarver, seems like it. The comments around `ALWAYS_PUBLIC` seem a bit conflicting to me though. Every scope that CAS recognizes ends up getting `ALWAYS_PUBLIC`?

```
# NOTE: Use with extreme caution.
# This should NEVER be assigned to endpoints:
#    - with mutable data,
#    - that might contain *anything* that could be personally-identifiable,
#    - as a write scope
ALWAYS_PUBLIC = 'always_public'
```

```
class scope(namedtuple('scope', ['parts_', 'description', 'is_public'])):
    """ Patch to add `ALWAYS_PUBLIC` scope to every selectable scope,
        ensuring that public endpoints are accessible with any token.
    """
    @property
    def parts(self):
        return frozenset((CoreScopes.ALWAYS_PUBLIC, )).union(self.parts_)
```

icereval: This was also confusing to me, after some discussions w/ @mfraezz its a bit clearer. 

The `scope` helper was added and the `ComposedScope`'s w/ `ALWAYS_PUBLIC` were left over from the first implementation. (I'll clean this up when I merge the PR)

This functionality basically allows any view marked `ALWAYS_PUBLIC` (very few, where making a specific scope was not warranted) to be called as long as the user has any valid Bearer token w/ scope.

The thing to remember (and the naming certainly is not helping) is `CoreScopes` are attached to Views. Where `ComposedScopes` are merely tuples of `CoreScopes` and only attached to the `public_scopes` below.

I like to think of these as each serves the other side of the authorization spectrum. Where the `public_scopes` are granted by the user at the beginning during the OAauth2 flow, and the `CoreScopes` lock down the views, so when a view is called the system verifies it's list of `CoreScopes` against the capabilities granted to the `public_scopes` -> `ComposedScope` on behalf of CAS.

caseyrollins: Thank you for explaining! 👍

PR #7456, opened 2017-07-12
[feature] Parse Full Name in Profile Settings for Citations Preview [OSF-8237]

Nesiehr:
## Purpose

Generate a citation preview with the full name (if possible) when a user hasn't filled in their first name, middle name, last name, or suffix.

## Changes

Changes to website/static/js/profile.js:
-- Added a parseName function to osfHelpers
-- Added conditions to citeApa and citeMla that use parseName as a last resort for generating a citation preview

Before:
before

After:
after

## Side effects

None known.

## QA Considerations

Parsing data from an open input field into a structure is interesting.
What if the person has more than a single first, middle, and last name?
What if the person has special characters in their name?
What if the person has a suffix, but no middle name?
What if the person has a suffix and a middle name?

This implementation will seperate out names by spaces. The first item it finds will always become the first name. If the second item it finds looks like a suffix (contains a period or is in a list of common suffixes) it will become a suffix. If it doesn't look like a suffix it will become a last name. Any names found after that are assigned to be a middle name.

Current list of suffixes:
'jr','jnr','sr','snr','dr','phd','i','ii','iii','iv','v','md','esq'

## Ticket

https://openscience.atlassian.net/browse/OSF-8237


website/static/js/profile.js
```diff
@@ -468,7 +468,44 @@ var NameViewModel = function(urls, modes, preventUnsaved, fetchCallback) {
...
+        fullNameObj.firstName = '';
+        fullNameObj.middleName = '';
+        fullNameObj.lastName = '';
+        fullNameObj.suffix = '';
+
+        fullName = fullName.split(/(\s+)/).filter( function(name) {
+            return name.trim().length > 0;
+        });
+
+        if (fullName.length === 4 && fullName[3].includes('.')) {
```
alexschiller: The `.` check is too fragile. Might consider other ways to check for suffixes including string matching a "complete" list of suffixes.

website/static/js/profile.js
```diff
@@ -468,7 +468,44 @@ var NameViewModel = function(urls, modes, preventUnsaved, fetchCallback) {
         return suffix;
     };
 
+    self.fullNameParsed = function() {
```
alexschiller: This should be a module on its own, either as a function in osfHelpers or completely separate, and fully tested. There are too many edge cases to hope it handles them all properly.

website/static/js/profile.js
alexschiller: What is there are more than 4?

website/static/js/profile.js
alexschiller: Here's my recommendation to simplify: 
1. Check the last part to see if it is a suffix (new function). If it is, pop it off and assign to suffix
2. Pop off the first piece and assign to first name (if len === 0 return nameobj)
3. Pop off last piece and assign to last name (if len === 0 return nameobj)
4. Assign remaining pieces (however many there are) to middle name, space separated

website/static/js/osfHelpers.es6.js
```diff
@@ -372,6 +372,50 @@ var decodeText = function(text) {
     return text;
 };
 
+var parseName = function(text) {
+    const suffixList = ['jr','jnr','sr','snr','dr','phd','i','ii','iii','iv','v','md','esq'];
```
alexschiller: While not actually an issues, suffixList is kind of a humorous name since this is JS and it is an array. 😆 

Please add a brief comment that explains the rationale for this list (e.g., it is difficult to know if something without a period in it is a suffix, so catch most by using data)

Nesiehr: Ahh, yea in java an array is an implementation of list so I got used to naming them like that haha, I'll add a comment describing its purpose

sloria:
Drive-by: since `osfHelpers.js` is already using ES5, and the new code isn't using ES6 features other than `let` and `const`, let's keep the file consistent in ES5 and remove the extension. No need to add extra compilation work if it's not providing a lot of value.

website/static/js/osfHelpers.es6.js
```diff
@@ -372,6 +372,50 @@ var decodeText = function(text) {
...
+        middleName: '',
+        lastName: '',
+        suffix: ''
+    };
+
+    text = text.split(/(\s+)/).filter( function(name) {
+        return name.trim().length > 0;
+    });
+
+    if (text.length >= 1) {
```
alexschiller: Because each  `if` sub section is doing something very specific for a specific reason in a specific order, it might be nice to have a one liner explaining the logic for maintainability purposes.

website/static/js/osfHelpers.es6.js
```diff
@@ -372,6 +372,50 @@ var decodeText = function(text) {
...
+            fullName.lastName = name;
+        }
+    }
+
+    if (!fullName.lastName && text.length >= 1) {
+        fullName.lastName = text.pop();
+    } 
+
+    if (text.length >= 1) {
+        text.forEach(function(name) {
```
alexschiller: pretty sure you can just do `fullName.middleName = text.join(' ')`

PR #7463, opened 2017-07-14
Feature/Add donate banner[OSF-8220]

laurenrevere:
## Purpose

Add temporary donate banner to the OSF homepage.

## Changes

Donate banner added on OSF home and OSF meetings page.

## Side effects

On dashboard, Prereg banner now below it.

## Tests
No tests needed: completely frontend.

## Ticket
https://openscience.atlassian.net/browse/OSF-8220

week1
week 2
week 3
week 4
week 5


website/static/css/donate-banner.css
```diff
@@ -0,0 +1,34 @@
...
+}
+
+.donate-text {
+    color: #a8dcec;
+    font-weight: bold;
+}
+
+/*Backgrounds change based on the week*/
+
+/*TODO: Change photos once they are recieved*/
```
brianjgeiger: I'm told that these are a thing, now. Could you update with the new banners?

website/static/js/home-page/donateBannerPlugin.js
```diff
@@ -0,0 +1,65 @@
+var m = require('mithril');
+var $osf = require('js/osfHelpers');
+
+// CSS
+require('css/donate-banner.css');
+require('css/meetings-and-conferences.css');
+
+var currentDate = new Date();
+
+//TODO: Change these dates when we know
```
brianjgeiger: Hopefully we know this as well by now?

website/static/js/home-page/donateBannerPlugin.js
```diff
@@ -0,0 +1,65 @@
...
+    new Date('July 21, 2017')
+];
+
+var bannerText = 'The Center for Open Science (COS) created the OSF and a suite of of free products to advance the ' +
+    'work of the research community. If you value these tools, please make a gift to support COS’s efforts to improve' +
+    ' and scale these services.';
+
+var background = '.donate-banner';
+
+if (currentDate > changeDates[0]){
```
brianjgeiger: This loop is never going to get past the first date, since the dates are ascending and you check the earliest one first. Lean towards `currentDate >= changeDates[x] and currentDate < changeDates[x+1]` for these, and have the last date be the ending date of the whole campaign where there won't be a banner at all.

brianjgeiger: Also put this in a function and unit test it, please.

alexschiller:
@brianjgeiger Styles here should also probably go in osf-style. Are we using osf-assets for anything yet? Banners might belong there.

laurenrevere:
Moving these to osf-style planned for later. Ticket here: https://openscience.atlassian.net/browse/OSF-8348

website/static/js/home-page/donateBannerPlugin.js
```diff
@@ -0,0 +1,94 @@
...
+    bannerOptions.sort(function (a,b){
+        return a.startDate - b.startDate;
+    });
+
+    var i;
+    for (i = 0; i < bannerOptions.length -1; i++) {
+        if (bannerOptions[i].startDate <= date && bannerOptions[i + 1].startDate > date)
+            return i;
+    }
+    return i;
```
brianjgeiger: The only downside of this is that it will never turn off the banner, so sometime in September we'll need to pull this out. Not a big deal, just noting.

laurenrevere: True, currently this does not automatically put up or take down the banner in any way, outside the specified dates it used the last banner text and options. Should I find a way to change this?

website/static/js/home-page/donateBannerPlugin.js
```diff
@@ -0,0 +1,94 @@
...
+
+// CSS
+require('css/donate-banner.css');
+require('css/meetings-and-conferences.css');
+
+var currentDate = new Date();
+
+var bannerOptions = [
+    {
+        startDate: new Date('August 7, 2017'),
```
sloria: Drive-by: Better to construct a date with numbers rather than relying on parsing, which might break in certain browsers.

```js
new Date(2017, 8, 7)
```

sloria:
Drive by: Please make sure the images are optimized; they should only be a few Kb.

brianjgeiger:
*Apparently* they are specially designed to cause problems with compression. I think we can get them lower, but not much lower without apparent JPG artifacts. Maybe in the dozens of k range or more. Possibly we can shrink them a bit as well and not just compress them more.

brianjgeiger:
(For the curious, there's a faded photo that is intersected by a mostly transparent gradient that goes the height of the image. That… is very unfriendly to both PNG and JPG compression schemes. GIF as well, but we'd never use GIF for that. But still.)

PR #7466, opened 2017-07-14
[feature] Expand Maintenance Notice Functionality [OSF-8215]

caseyrollins:
#### Purpose
- Expands `MaintenanceState` model to allow for custom messages and levels.

#### Changes
- Adds `level`, and `message` fields to `MaintenanceState` model.
- Adds basic `MaintenanceStateSerializer` for APIv2 (`/v2/status/`) and for returning maintenance notice information to the OSF.

#### Side Effects
- None expected

#### Ticket
- [OSF-8215](https://openscience.atlassian.net/browse/OSF-8215)

#### Screenshots 
screen shot 2017-08-03 at 5 16 06 pm

screen shot 2017-08-03 at 5 13 20 pm

screen shot 2017-08-03 at 5 15 25 pm


sloria:
Not sure if adding support for multiple maintenance notices is worth the added code complexity. I don't foresee a situation where we'd want to display multiple banners. Why the change? @caseyrollins

sloria:
Also, this is a backwards-incompatible change in the API; ember-osf would need to be updated for this.

caseyrollins:
@sloria, I'm honestly not sure why I focused on displaying multiple maintenance notices. I think something in the ticket led me to interpret it that way?

I could see a case for multiple banners (Rackspace goes down while we have a planned maintenance banner already on the site), although it's highly unlikely and if it did happen one notice could be removed in favor of the other. So I agree that it probably isn't worth the additional code complexity.

And I can revert the API changes so things are backwards compatible.

sloria:
@caseyrollins OK, let's go with just one maintenance alert at a time for now.

api/base/serializers.py
```diff
@@ -1539,3 +1539,15 @@ def create(self, validated_data):
...
 
         return self.make_instance_obj(collection)
+
+
+class MaintenanceStateSerializer(ser.ModelSerializer):
+
+    level = ser.SerializerMethodField()
+
+    def get_level(self, obj):
+        return obj.get_level_display()
```
sloria: Might make sense to just serialize as an integer; that way the client can easily compare against it and decide what they want to do with it. Our API docs should make the semantics clear:

```
1 = Informational. Used most commonly for scheduled maintenance.
2 = Used for service outages
```

tasks/__init__.py
```diff
@@ -992,11 +992,16 @@ def usage(ctx):
 ### Maintenance Tasks ###
 
 @task
-def set_maintenance(ctx, start=None, end=None):
+def set_maintenance(ctx, message, level=1, start=None, end=None):
```
sloria: `message` should be optional. If it isn't provided, the behavior should be the same as before.

website/maintenance.py
```diff
@@ -4,11 +4,14 @@
 from dateutil.parser import parse
 from django.utils import timezone
 
+from api.base.serializers import MaintenanceStateSerializer
 from osf.models.maintenance_state import MaintenanceState
 
 
-def set_maintenance(start=None, end=None):
-    """Set the time period for the maintenance notice to be displayed.
+def set_maintenance(message, level=1, start=None, end=None):
```
sloria: As previously mentioned, `message` should be optional.

caseyrollins:
Made the front-end changes so that custom messages are displayed in the banner and the banner CSS now depends on the level (screenshots in description). Also made `message` optional.

PR #7473, opened 2017-07-17
Subject permissions for view/change in admin [#OSF-8284]

erinspace:
## Purpose
We can now see custom taxonomies in the Admin App as OSF Subject
We need to add a permissions scope for View OSF Subject
There should also be a test to make sure correct permissions exist

## Changes

- Add custom `view_subject` Permission in the Subject model's Meta
- Fix permission in subject udpdate to be auto-generated permission `change_subject`
- Migration to add Permissions and add new permissions to relevant groups

## Side effects
Migration will need to be run for this to take effect!

## Ticket
https://openscience.atlassian.net/browse/OSF-8284


osf/migrations/0045_add_view_subject_permissions.py
```diff
@@ -0,0 +1,53 @@
...
+    # this is to make sure that the permissions created in an earlier migration exist!
+    emit_post_migrate_signal(2, False, 'default')
+
+    view_permission = Permission.objects.get(codename='view_subject')
+    edit_permission = Permission.objects.get(codename='change_subject')
+
+    osf_admin = Group.objects.get(name='osf_admin')
+    read_only = Group.objects.get(name='read_only')
+
+    [osf_admin.permissions.add(perm) for perm in [view_permission, edit_permission]]
```
sloria: Minor: You can do `osf_admin.permissions.add(view_permission, edit_permissions)`. Not a big deal though.

PR #7476, opened 2017-07-18
[Feature] Resend share data and cleanup [OSF-8144]

alexschiller:
## Purpose
Data that fails to send to share never retries

## Changes
Add retries for 500 errors, mail to desk for 400 errors and 500 errors that fail 5 more times (with increasing retry delay). 

## Side effects
This ticket also fixes OSF-8070 by calling the preprint update function for nodes when contributors are updated. 

Cleaned up and added a bunch of tests and code. Removed a bunch of mocks that were not necessary. Standardized the code between project/tasks and preprint/tasks to be consistent for easier future maintenance.

Also, desk might get a lot of mail...

## Ticket

https://openscience.atlassian.net/browse/OSF-8144

## Tests
I added tests and updated a great deal of tests where possible. The main thing that is missing are tests for the asynchronous delayed celery tasks. Locally the delay incremented properly and was called the correct number of times, but the best I could do was check that the async delay task was called when the response was >500.

## QA Notes
There really isn't an easy test for QA. At a minimum, hit the admin reindex share buttons for a node, registration, and preprint to ensure they still work after my changes since I touched that code.


website/project/tasks.py
```diff
@@ -34,51 +35,76 @@ def on_node_updated(node_id, user_id, first_save, saved_fields, request_headers=
...
-                    '@id': '_:789',
-                    '@type': 'project',
-                    'is_deleted': not node.is_public or node.is_deleted or node.is_spammy,
-                }]}
-            }
-        }
-    }, headers={'Authorization': 'Bearer {}'.format(settings.SHARE_API_TOKEN), 'Content-Type': 'application/vnd.api+json'})
+@celery_app.task(bind=True, max_retries=4, acks_late=True)
+def _async_update_share(self, node_id):
+    # Any modifications to this function may need to change _async_update_share
```
brianjgeiger: This function is `_async_update_share`. Did you mean `_update_share` or `_async_update_preprint_share`?

website/preprints/tasks.py
```diff
@@ -30,22 +32,66 @@ def on_preprint_updated(preprint_id, update_share=True, old_subjects=None):
...
-        logger.debug(resp.content)
+        _update_preprint_share(preprint, old_subjects)
+
+def _update_preprint_share(preprint, old_subjects):
+    data = serialize_share_preprint_data(preprint, old_subjects)
+    resp = send_share_preprint_data(preprint, data)
+    try:
+        assert False
+        resp.raise_for_status()
+    # any exception gets sent to deskk
```
alexschiller: Was a missed placeholder before I added that functionality, removing

website/preprints/tasks.py
```diff
@@ -30,22 +32,66 @@ def on_preprint_updated(preprint_id, update_share=True, old_subjects=None):
...
+    except Exception:
+        if resp.status_code >= 500:
+            _async_update_preprint_share.delay(preprint._id, old_subjects)
+        else:
+            send_desk_share_preprint_error(preprint, resp, 0)
+
+@celery_app.task(bind=True, max_retries=4, acks_late=True)
+def _async_update_preprint_share(self, preprint_id, old_subjects):
+    # Any modifications to this function may need to change _async_update_share
```
alexschiller: @brianjgeiger More a defensive/paranoia thing. I think I separated the  _update and _async_update functions well enough that any changes should only have to occur to the functions both call. 

A few minutes thought and I couldn't come up with a good way to guard against them changing.

website/project/tasks.py
alexschiller: I went through and renamed all of the _update_share functions to be _update_node_share to be less ambiguous. Pushing in just a moment. Should clarify this.

alexschiller: Also, I made sure each of these comments referenced the proper function, derrr

brianjgeiger:
@alexschiller Please update with develop and let me know when tests pass.

PR #7477, opened 2017-07-18
[feature] Add DirtyFieldsMixin to Institution Model [OSF-8244]

caseyrollins:
#### Purpose
- Optimizes `scripts.populate_institutions` 
- Removes the call to `update_node` for every affiliated node of every institution updated/created in the script. Instead, nodes will be updated in search if the institution is actually changed on save, i.e. if the name of an institution is changed in the populate script.

#### Changes
- Add `DirtyFieldsMixin` to `Institution` model. 
- Update `Insitution.save()` to check for changed fields before calling `update_node` for all affiliated nodes. 

#### Side Effects
- Whenever an Institution is saved (with a changed field, such as `name`), `update_node` will be called for every node affiliated with that institution. In theory, this shouldn't be a problem because `scripts.populate_institutions` is the only place where institutions are updated.

#### Ticket
- [OSF-8244](https://openscience.atlassian.net/browse/OSF-8244)

### DIFF
--- a/osf/models/institution.py
+++ b/osf/models/institution.py
@@ -1,4 +1,7 @@
 import logging
+
+from dirtyfields import DirtyFieldsMixin
+
 from django.conf import settings
 from django.contrib.postgres import fields
 from django.core.urlresolvers import reverse
@@ -10,7 +13,7 @@ from osf.models.mixins import Loggable
 logger = logging.getLogger(__name__)
 
 
-class Institution(Loggable, base.ObjectIDMixin, base.BaseModel):
+class Institution(DirtyFieldsMixin, Loggable, base.ObjectIDMixin, base.BaseModel):
 
     # TODO Remove null=True for things that shouldn't be nullable
     # e.g. CharFields should never be null=True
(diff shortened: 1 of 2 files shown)


osf/models/institution.py
```diff
@@ -103,13 +106,22 @@ def banner_path(self):
...
         try:
             update_institution(self)
         except SearchUnavailableError as e:
             logger.exception(e)
 
+        saved_fields = self.get_dirty_fields()
+        if saved_fields and bool(self.pk):
+            for node in self.nodes.filter(is_deleted=False):
+                try:
+                    update_node(node, async=False)
```
sloria: Wouldn't it also make sense to call `update_institution` here?

caseyrollins: Ah, do you just mean calling `update_institution` after the `if saved_fields:`? It's called a few lines above (LL113) and I wouldn't want to call it within this loop.

sloria: Yeah, I mean that `update_institution` should only be called if the node was updated.

caseyrollins: Hmmm, why not update an institution that doesn't have any affiliated nodes though? Institutions still appear in search.

sloria: Woops, I had a typo in my above comment (now corrected). I think we only need to update the institution in ES if it's been changed, i.e. `saved_fields` is truthy.

caseyrollins: Yep, I agree with that. I'll make that change.

sloria: Thanks

sloria:
Made this a hotfix in https://github.com/CenterForOpenScience/osf.io/commit/bc558edd7dd7220785055bc63f13fd60de0a4cf3 in preparation for https://github.com/CenterForOpenScience/osf.io/pull/7514

PR #7478, opened 2017-07-18
[RFR][OSF-8210][Migration] Update OSFUser validation to catch invalid SOCIAL_FIELDS

TomBaxter:
## This PR requires running a migration

osf/migrations/0051_remove_invalid_social_entries.py  removes any invalid SOCIAL_FIELDS data from current users.

Without running this, current users with invalid data will not be able to update OSFUser.social and may not be able to make other OSFUser changes as well.

## Purpose

Update OSFUser social_validator to catch OSFUser.social keys not in OSFUser.SOCIAL_FIELDS

## Changes

Update osf/models/validators.py to catch fields not in SOCIAL_FIELDS
Update api/users/serializers.py to return useful error message
Add scripts/remove_invalid_social_entries.py for existing bad entries
Fixed tests broken by change

## Side effects

If third party OSF v2 API users are currently taking advantage of un-validated SOCIAL_FIELDS they will no longer be able to do so.

## Ticket
[OSF-8210 Able to create made-up fields in the OSF API](https://openscience.atlassian.net/browse/OSF-8210)


api/users/serializers.py
```diff
@@ -102,6 +102,8 @@ def update(self, instance, validated_data):
                 setattr(instance, attr, value)
         try:
             instance.save()
+        except ValidationError as e:
+            raise InvalidModelValueError(detail=e.args)
```
sloria: `detail` needs to be a string; isn't `e.args` a `tuple`?

scripts/remove_invalid_social_entries.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
+"""
```
sloria: This should be a proper Django migration. See examples of custom migrations in the `osf/migrations/` directory

scripts/remove_invalid_social_entries.py
```diff
@@ -0,0 +1,40 @@
...
+from osf.models import OSFUser
+
+logger = logging.getLogger(__name__)
+
+
+def do_migration(dry):
+    logger.warn('Removing invalid social entries!')
+    if dry:
+        logger.warn('Dry_run mode')
+    for user in OSFUser.find():
```
sloria: Prefer django syntax: `OSFUser.objects.filter()`

scripts/remove_invalid_social_entries.py
```diff
@@ -0,0 +1,40 @@
...
+        for item in user.social.copy():
```
sloria: Minor: Rather than copying the whole dict, you can do a set difference:

```python
for invalid_key in set(user.social.keys()) - set(settings.SOCIAL_FIELDS.keys()):
    user.social.pop(invalid_key)
```

scripts/remove_invalid_social_entries.py
sloria: Use `only` to optimize queries (so you don't query more data than you need.

```python
OSFUser.objects.filter().only('social')
```

api/users/serializers.py
TomBaxter: Fixed with kludge. Unfortunately the exception raised has no message attribute. To change this would require updating osf/models/base.save() which I am loath to do.

sloria: @TomBaxter How about `e.args[0]`? That is a typical way to get the error message for a given exception.

TomBaxter: I can do that, what I had planned to do was...
```
if 'social' in e.args[0]:
    raise InvalidModelValueError(detail=e.args[0]['social'][0][0])
```
That way only the message comes trough. However, I see in the production sentry errors that most ValidationErrors are using e.args[0] . What do you think? Want I should just go with e.args[0] ?

sloria: Oh, I didn't realize `e.args[0]` was a `dict`. Why is it a `dict` if you're passing a string [here](https://github.com/CenterForOpenScience/osf.io/pull/7478/files#diff-816963fa1deefc5403af7aadd46d0898R88)?

scripts/remove_invalid_social_entries.py
TomBaxter: I'm having trouble getting .only() to work. I haven't been able to find an example in the code base. I found the Django Doc on it but am receiving an error I haven't been able to resolve.
```
OSFUser.objects.only("social")
...
/usr/local/lib/python2.7/site-packages/django/db/models/query_utils.pyc in __get__(self, instance, cls)
    114             val = self._check_parent_chain(instance, self.field_name)
    115             if val is None:
--> 116                 instance.refresh_from_db(fields=[self.field_name])
    117                 val = getattr(instance, self.field_name)
    118             data[self.field_name] = val

TypeError: refresh_from_db() got an unexpected keyword argument 'fields'
```

scripts/remove_invalid_social_entries.py
TomBaxter: Complete.

scripts/remove_invalid_social_entries.py
TomBaxter: Complete.

TomBaxter:
@sloria Pushed updates.

website/settings/defaults.py
```diff
@@ -1832,3 +1832,19 @@ def parent_dir(path):
 }
 
 PREPRINTS_ASSETS = '/static/img/preprints_assets/'
+
+SOCIAL_FIELDS = {
```
sloria: Was it necessary to move this? This isn't really a setting, so putting it in settings seems inappropriate.

TomBaxter: I don't think it is necessary. I was having a little trouble with the importing OSFUser into the validator. I'm sure I could have worked around that trouble, but when I saw what I took to be hard coded configuration in OSFUser, I figured it belonged in a config file. This is the first PR I've done working with django models, so not sure what's best practice.
  Want I should move it back into OSFUser?

sloria: I think it should be either (1) a class variable on `OSFUser` (`OSFUser.SOCIAL_FIELDS`) or (2) a module-level variable (`osf.models.user.SOCIAL_FIELDS`). 

(1) seems slightly preferable so that it's easy to access whenever you have `OSFUser` imported.

If you're running into circular import errors, you can import OSFUser in the validator function.

TomBaxter:
@sloria  PR rebased and pointing to release/next-djangole. SOCIAL_FIELDS moved back to OSFUser. Ready for further review.

scripts/migration/remove_invalid_social_entries.py
```diff
@@ -0,0 +1,43 @@
+"""
```
sloria: This still needs to be made into a proper migration. It should be located in `osf/migrations` and follow our conventions for custom Django migrations (see other custom migrations in that folder for examples).

osf/models/validators.py
```diff
@@ -81,10 +81,12 @@ def validate_profile_websites(profile_websites):
...
             raise ValidationError('Invalid personal URL.')
 
-
 def validate_social(value):
     validate_profile_websites(value.get('profileWebsites'))
-
+    from osf.models import OSFUser
+    for soc_key in value.keys():
+        if soc_key not in OSFUser.SOCIAL_FIELDS:
+            raise ValidationError('{}: is not a valid key for social.'.format(soc_key))
```
sloria: The ":" shouldn't be there. Better to quote the key, e.g. `'"{}" is not valid key...'`

TomBaxter: Ok, I had it that way before, but the response comes back \"badkey\"  not sure why it escapes it, but I thought it was ugly so I changed to colon. I'll just remove the colon.

scripts/migration/remove_invalid_social_entries.py
TomBaxter: Ugh, sorry. I missed that, thought I had read move to scripts/migration. Ok, back to the drawing board.

api/users/serializers.py
```diff
@@ -103,6 +103,9 @@ def update(self, instance, validated_data):
                 setattr(instance, attr, value)
         try:
             instance.save()
+        except ValidationError as e:
+            if 'social' in e.args[0]:
```
sloria: This deserves a code comment. Something along the lines...

```python
# When a ValidationError is raised for 'social', e.args[0] will a 
# dict of the form {'social': .....}
```

TomBaxter: Complete.

osf/models/validators.py
TomBaxter: Complete.

scripts/migration/remove_invalid_social_entries.py
TomBaxter: Complete.
now ``` osf/migrations/0043_remove_invalid_social_entries.py ```
and retested.

website/settings/defaults.py
TomBaxter: Complete.

osf/migrations/0044_remove_invalid_social_entries.py
```diff
@@ -0,0 +1,35 @@
...
+
+from django.db import migrations
+
+from osf.models import OSFUser
+
+import logging
+
+logger = logging.getLogger(__file__)
+
+def remove_invalid_social_entries(state, junk):
```
sloria: Instead of `junk`, you can do `*args, **kwargs`.

osf/migrations/0044_remove_invalid_social_entries.py
```diff
@@ -0,0 +1,35 @@
...
+    osfuser = state.get_model("osf", "osfuser")
```
sloria: As per PEP8, use uppercase for classes, i.e. `OSFUser`.

sloria: Minor: Also, as per our style guidelines, use singlequotes.

osf/migrations/0044_remove_invalid_social_entries.py
```diff
@@ -0,0 +1,35 @@
...
+
+    logger.info('Invalid social entry removal completed.')
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('osf', '0043_merge_20170725_1328'),
+    ]
+
+    operations = [
```
sloria: Need to include `RunPython` here.

osf/migrations/0044_remove_invalid_social_entries.py
TomBaxter: Complete.

osf/migrations/0044_remove_invalid_social_entries.py
TomBaxter: Complete.

osf/migrations/0044_remove_invalid_social_entries.py
TomBaxter: Complete.

TomBaxter:
@sloria  CR addressed and migration re-tested.

api/users/serializers.py
```diff
@@ -102,6 +102,11 @@ def update(self, instance, validated_data):
                 setattr(instance, attr, value)
         try:
             instance.save()
+        # When a ValidationError is raised for 'social', e.args[0] will a
+        # dict of the form {'social': [<Exception>]}
+        except ValidationError as e:
+            if 'social' in e.args[0]:
+                raise InvalidModelValueError(detail=e.args[0]['social'][0][0])
```
sloria: There's a chance we might be swallow a `ValidationError` here. It probably makes sense to add an `else:` block here that just raises the ValidationError that just raises.

Also, since `ValidationValueError` is a subclass of `ValidationError`, there is no way to get to the below `except ValidationValueError` block. Would it make sense to just catch `ValidationError`?

```python
try:
    ...
except ValidationError as e:
    if 'social' ...:
        raise ...
   else:
       raise ...
```

TomBaxter: Ugh, I didn't catch the subclass issue. I've flipped the if so that subclass is caught first.

As to swallowing ```ValidationError``` s, I had done that on purpose. I only wanted to catch the error I fully understood. In any case I've now added more elaborate logic that should catch any other ```ValidationError``` s who's args take the same form as the social error. Anything that doesn't match gets re-raised as is.

  This will likely have the effect of [many current errors](https://sentry.cos.io/sentry/osf/?query=is%3Aunresolved+ValidationError) returning 400 with an error message instead of 500.

api/users/serializers.py
```diff
@@ -104,6 +104,15 @@ def update(self, instance, validated_data):
...
         except ValidationValueError as e:
             raise InvalidModelValueError(detail=e.message)
+        # ValidationError s raised by instance.save() are likely to have e.args[0] be
+        # dict of the form {'field validated': [<Exception>]}
+        except ValidationError as e:
+            if isinstance(e.args[0], dict):
+                value = e.args[0].popitem()[1]
+                if len(value) == 1:
+                    if isinstance(value[0], Exception):
+                        raise InvalidModelValueError(detail=value[0][0])
```
sloria: OK, I see now; `ValidationErrors` are usually re-raised with a dictionary of errors.

It might be worth adding 

```python
elif isinstance(e.args[0], basestring):
    raise InvalidModelValueError(detail=e.args[0])
```

TomBaxter: Will do. Hot on another ticket, so will make change in a bit.

osf/migrations/0045_remove_invalid_social_entries.py
```diff
@@ -0,0 +1,36 @@
...
+
+from osf.models import OSFUser as osfuser
+
+import logging
+
+logger = logging.getLogger(__file__)
+
+def remove_invalid_social_entries(state, *args, **kwargs):
+    OSFUser = state.get_model('osf', 'osfuser')
+    targets = OSFUser.objects.filter()
```
sloria: In order to avoid iterating through every user, you can just iterate through the ones that have anything in `social`:

```python
 OSFUser.objects.exclude(social={})
```

On prod, this is ~7,600 vs ~75,000.

api/users/serializers.py
TomBaxter: Complete and tested.

osf/migrations/0045_remove_invalid_social_entries.py
TomBaxter: Complete and tested.

api/users/serializers.py
```diff
@@ -104,6 +104,17 @@ def update(self, instance, validated_data):
             instance.save()
         except ValidationValueError as e:
             raise InvalidModelValueError(detail=e.message)
+        # ValidationError s raised by instance.save() are likely to have e.args[0] be
+        # dict of the form {'field validated': [<Exception>]}
+        except ValidationError as e:
+            if isinstance(e.args[0], dict):
+                value = e.args[0].popitem()[1]
+                if len(value) == 1:
+                    if isinstance(value[0], Exception):
```
sloria: What if `value[0]` isn't an exception? Are there other cases to handle, i.e. a string? Should there be an `else`?

api/users/serializers.py
```diff
@@ -104,6 +104,17 @@ def update(self, instance, validated_data):
             instance.save()
         except ValidationValueError as e:
             raise InvalidModelValueError(detail=e.message)
+        # ValidationError s raised by instance.save() are likely to have e.args[0] be
+        # dict of the form {'field validated': [<Exception>]}
+        except ValidationError as e:
+            if isinstance(e.args[0], dict):
+                value = e.args[0].popitem()[1]
+                if len(value) == 1:
```
sloria: Can `len(value)` be greater than 1? Should there be an `else`?

sloria:
I don't think my above concerns are necessarily blockers, so I've approved this. If you have some time, look into improving the error handling. Otherwise, please create a follow-up ticket.

api/users/serializers.py
TomBaxter: Per our conversation yesterday EOB. After further review, I learned that these exceptions are handled by a custom handler that takes all these possibilities into consideration. I'm not thrilled with the formating of the message it produces, but the pertinent information is available to API user.
  Therefore, I have updated to simply re-raise the correct Exception.

api/users/serializers.py
TomBaxter: Complete.

api/users/serializers.py
```diff
@@ -104,7 +104,8 @@ def update(self, instance, validated_data):
             instance.save()
         except ValidationValueError as e:
             raise InvalidModelValueError(detail=e.message)
-        return instance
+        except ValidationError as e:
+            raise InvalidModelValueError(e)
```
sloria: Much better!

PR #7479, opened 2017-07-18
Remove aliasing of OSFUser and AbstractNode [OSF-8223]

erinspace:
## Purpose

Some model names have changed, and lots os usages are aliased to their old names. For consistency and clarity, get rid of the aliases and use their concrete classes instead. 

## Changes

- Remove aliasing in import statements 
- Change previously aliased usages 
- Remove some unused imports

## Side effects

None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-8223


sloria:
Rebased and merged into develop!

PR #7484, opened 2017-07-19
[Feature] Convert api_tests search serializers, views to pytest [OSF-8303]

fabmiz:
## Purpose

Convert api_tests/search from nose to pytest

## Changes

Re-write tests for search search` serializers`, `views` with pytest.

## Ticket

[OSF-8303](https://openscience.atlassian.net/browse/OSF-8303)


api_tests/search/views/test_views.py
```diff
@@ -135,485 +165,496 @@ def test_search_fields_links_with_query(self):
...
-        assert_equal(res.status_code, 200)
+        #test_search_public_component_auth(self, app, all, url_component_search, user, component, component_public):
+        res = app.get(url_component_search, auth=user)
+        assert res.status_code == 200
         num_results = len(res.json['data'])
         total = res.json['links']['meta']['total']
-        assert_equal(num_results, total, 2)
-        assert_in(self.component.title, res)
-        assert_in(self.component_two.title, res)
+        assert num_results == total, 2
```
alexschiller: Driveby comment:
I asked the person who wrote these tests originally and this is a case of them misunderstanding how assert_equal worked. The intended conversion is:

```
assert num_results ==  2
assert total == 2
```

assert_equal compares the first two values, and if they fail, returns the third (msg) as a message with the failure. The original code intended for all of them to be equal.

alexschiller: A quick regex showed this is the only file that uses that pattern, I think it was 38 in here that will need to be converted.

fabmiz: oh wow, good catch, thanks. I will replace all occurrences in PR update.

api_tests/search/serializers/test_serializers.py
```diff
@@ -1,65 +1,48 @@
...
+        context = {'request': make_drf_request_with_version(version='2.0')}
+        schema = MetaSchema.objects.filter(
+            name='Replication Recipe (Brandt et al., 2013): Post-Completion',
+            schema_version=LATEST_SCHEMA_VERSION).first()
 
-    def test_search_serializer_mixed_model_project(self):
-        req = make_drf_request_with_version(version='2.0')
-        result = SearchSerializer(self.project, context={'request': req}).data
-        assert_equal(result['data']['type'], 'nodes')
+        #test_search_serializer_mixed_model_project(self, project, context):
```
Nesiehr: remove (self, ...) in all of the comments

PR #7490, opened 2017-07-19
[Feature] Add provider-distinguishing field on TaxonomySerializer [OSF-8308]

mfraezz:
## Purpose
Include a boolean attribute for Subjects in API responses. Indicates to the frontend whether to use `bepress` or `<preprint_provider._id>` for SHARE path construction

## Changes
* Add `SerializerMethodField`
* Add tests

## Side effects
None expected, though listing large number of subjects might be slightly slower

## Ticket
[[OSF-8308]](https://openscience.atlassian.net/browse/OSF-8308)

### DIFF
--- a/api_tests/preprint_providers/views/test_preprint_provider_subjects_list.py
+++ b/api_tests/preprint_providers/views/test_preprint_provider_subjects_list.py
@@ -252,3 +252,20 @@ class TestPreprintProviderHighlightedSubjects(ApiTestCase):
         assert res.status_code == 200
         assert len(res.json['data']) == 1
         assert res.json['data'][0]['id'] == self.subj_aa._id
+
+class TestCustomTaxonomy(ApiTestCase):
+    def setUp(self):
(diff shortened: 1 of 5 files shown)


api_tests/preprint_providers/views/test_preprint_provider_subjects_list.py
```diff
@@ -252,3 +252,20 @@ def test_mapped_subjects_filter_wrong_provider(self):
         assert res.status_code == 200
         assert len(res.json['data']) == 1
         assert res.json['data'][0]['id'] == self.subj_aa._id
+
+class TestCustomTaxonomy(ApiTestCase):
+    def setUp(self):
+        super(TestCustomTaxonomy, self).setUp()
+        self.osf_provider = PreprintProviderFactory(_id='osf')
```
sloria: Non-blocker: New tests should use fixtures rather than `setUp`.

PR #7491, opened 2017-07-20
[OSF-8056] Move "Add-ons" to own tab on Node Overview, out of Project Settings

Johnetordoff:
## Purpose

This creates a new page where the user can easily find their addons and allows then to begin connecting their account with one click.

## Changes

Adds a new addons page which loads the configurations template for all available addons, when the js for these templates load they will check to see if their addons are enabled for their node if not they will receive a 404 and give the user the option to enable the addon, if the addon is configured everything behaves normally. 

Originally these panels weren't suppose to be loaded if the user didn't have the addon enabled for that template, these changes are mostly centered around making it possible to load these panels without having any error behavior when the addon's aren't enabled and ensuring addons enabled on the node before they try to connect with the addon's provider.

## Side effects

None that I know of.

## Tests

Mostly front end changes, backend changes don't appear to have any new test cases associated with them. 

## Before
before_addon_page

## After

new_addon_page

## Ticket

https://openscience.atlassian.net/browse/OSF-8056


Johnetordoff:
Instead of resolve conflicts and update this I've started a new PR here: https://github.com/CenterForOpenScience/osf.io/pull/7560

PR #7492, opened 2017-07-20
[Feature] PreprintProvider/Subject `share_title` field

mfraezz:
## Purpose
Add field for front-end / SHARE support

## Changes
* Add field to `PreprintProvider`
* Proxy field through `Subject` serializer
* Deprecate unnecessary endpoint

## Side effects
None expected. Fields will need to be populated correctly

## Ticket
Unknown


jamescdavis:
Need to replace test_taxonomy_is_bepress with test for share_title

PR #7496, opened 2017-07-21
[#OSF-8189]Unregistered users added to OSF projects -- new names should overwrite old names

chennan47:
## Purpose
Unregistered users added to OSF projects -- new names should overwrite old names

## Changes
1. go to a project and add a new unreg contributor with an email
2. remove that unreg contributor just added
3. add that unreg contributor again with the same email but a new name
Before change:
the old name of the unreg contributor will show up
After change:
the ner name of the unreg contributor will show up

## Ticket
https://openscience.atlassian.net/browse/OSF-8189


osf_tests/test_user.py
```diff
@@ -492,6 +492,33 @@ def test_display_full_name_unregistered(self):
         u.reload()
         assert u.display_full_name(node=project) == name
 
+    def test_repeat_add_same_unreg_user_with_diff_name(self):
+        u = UnregUserFactory()
```
alexschiller: Minor, not blocking suggestion; We're trying to be more consistent and descriptive in test variable names, putting them in the format <what it is>_<descriptive features> (e.g., `user_one`, `user_two`, `user_admin`, `user_read` etc)

For this one, `user_unregistered` would be more ideal.

Look over some of Rheisen's pytest conversions for more examples

chennan47: solved in 60468d8
--

PR #7497, opened 2017-07-21
Improve Query performance for NodeLog [#OSF-8288]

erinspace:
## Purpose

Getting NodeLogs is super expensive, mainly because the query also involves gathering all node logs for the node and the node's children. 

## Changes
Tried to optimize this many many different ways, including replacing the call to `list()` in `get_aggregate_logs_query()` with a SQL subquery, which resulted in a much much slower result (see Jira ticket for more details). In the end the fastest optimization turned out to be adding indices and truncating the number of queries needed by the API from 49 to 19 using `includes()`.

- Use `IncludeManager` for nodelog querysets
- Remove MODM Q objects
- Migration to add indices on `should_hide` field
- Also Index on `node`, `should_hide`, and `-date` 

## Side effects

- A migration needs to be run!

## Ticket
https://openscience.atlassian.net/browse/OSF-8288


cwisecarver:
My $.02: Someone should run this migration against prod data before release. My guess is that it's going to take a loooooooong time and might benefit from [CREATE INDEX CONCURRENTLY](https://www.postgresql.org/docs/current/static/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY)

erinspace:
I ran this against staging and took a long time once, I think because of an unrelated issue.  I also wrote a version that used [CREATE INDEX CONCURRENTLY](https://www.postgresql.org/docs/current/static/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY) -- worked a bit with @mfraezz on that -- and it didn't significantly make things run faster. Prod might be different though!

cwisecarver:
The CONCURRENTLY thing will only help if the table is being written to during the index creation. Probably wasn't, at least not heavily, in staging. Definitely will be in prod unless we take downtime for this release.

erinspace:
Makes sense! I'll totally put back the concurrently thing if the lack of transaction isn't a problem. Lemmie know!

cwisecarver:
I would try and add each index individually, in it's own migration, with `CONCURRENTLY`. If one of them passes and one of them fails then there's no need to rollback both. Each index will make things faster independently. Doing multiple schema changes inside one migration, on production, is generally an anti-pattern. 

Regardless, this is one of our biggest tables so however we do it we should test it against prod data. If we use CONCURRENTLY then we should try it on someone's machine while trying to access the osf and doing things that would create nodelogs. That way we can get an ballpark of the potential time-cost with locking before we roll this out.

If we don't create the indexes concurrently then I would at least run it against prod data locally without trying to create nodelogs, because you probably won't be able to, those tables will likely be locked; and accept that however long it takes we'll have to take that much downtime, no celery, no osf, no v2 api, no admin.

If both methods take longer that we're willing to invest in downtime then I would suggest that we create the migrations, either with or without concurrently, and then `python manage.py migrate --fake`-them and then ninja them onto production with `CONCURRENTLY` through psql or datagrip. IIRC this has been the best solution for creating a few gigantic indexes on SHARE. This way they don't block the deployment, lock tables, and everyone using the osf for development or third-party still gets the index migrations. @icereval @chrisseto Thoughts?

osf/models/node.py
```diff
@@ -778,15 +779,19 @@ def can_edit(self, auth=None, user=None):
...
+            (
+                Q(node_id__in=list(Node.objects.get_children(self).can_view(user=auth.user, private_link=auth.private_link).values_list('id', flat=True))) |
+                Q(node_id=self.id)
+            ) & Q(should_hide=False)
         )
 
     def get_aggregate_logs_queryset(self, auth):
         query = self.get_aggregate_logs_query(auth)
-        return NodeLog.find(query).order_by('-date')
+        return NodeLog.objects.filter(query).order_by('-date').select_related(
```
sloria: Does `select_related` actually saving any queries now that you're using `include`?

cwisecarver: Not unless something changed. Include used to bring in the entire lineage of things you asked for.

sloria: 👍 That's what I thought. @erinspace Can you verify that removing `select_related` doesn't increase the number of queries?

erinspace:
Going to close this in favor of #7526 which just does the query improvement aspects. I created [a new jira ticket](https://openscience.atlassian.net/browse/OSF-8344) with the discussion from this PR for when we add the appropriate DB indexes.

PR #7498, opened 2017-07-21
[OSF-8213] Track Number of Preprints by Provider in Admin App

Johnetordoff:
## Purpose

Tracks preprint provider data and displays it in the admin app.

## Changes

Adds a script which runs with other analytic scripts to report the number preprints grouped by provider to Keen which is then queried by js in the Admin App and formatted into a pretty chart.

## Side effects

To get data about preprints prior to deployment of this PR someone with production access will have to run the preprint_summary script for those dates.

## Tests

One unit test is included made difficult by the face Keen can only query things that happened yesterday or earlier. I have to "mock" preprint date_created fields so they appear older then a day. (I left a short comment in the test).

## Ticket

https://openscience.atlassian.net/browse/OSF-8213

## Screenshot

screen shot 2017-07-21 at 10 07 58 am

_still needs real data_


admin/static/js/metrics/metrics.es6.js
```diff
@@ -941,12 +941,32 @@ var AddonMetrics = function() {
...
+//   preprints   |
+// ><+><+><+<+><+
+
+var PreprintMetrics = function() {
+    var preprint_created = new keenAnalysis.Query("sum", {
+        eventCollection: "preprint_summary",
+        targetProperty: "provider.total",
+        groupBy: ["provider.name"],
+        interval: "daily",
+        timeframe: "this_365_days",
```
brianjgeiger: 365 days is probably going to get expensive. It should happen too frequently, but unless Product really is hoping for a year long graph, we probably don't need it. It's worth checking to see if a week or a month view would be acceptable.

scripts/analytics/preprint_summary.py
```diff
@@ -0,0 +1,69 @@
...
+    def get_events(self, date):
+        super(PreprintSummary, self).get_events(date)
+        from osf.models import PreprintService, PreprintProvider
+
+        # Convert to a datetime at midnight for queries and the timestamp
+        timestamp_datetime = datetime(date.year, date.month, date.day).replace(tzinfo=pytz.UTC)
+        query_datetime = timestamp_datetime + timedelta(1)
+
+        counts = []
+        for preprint_provider in PreprintProvider.objects.all():
```
brianjgeiger: This information should all be available via SHARE and could probably be aggregated by elastic very easily. Alternately, you might be able to do this in SQL, but I think a SHARE query would be better. No sense stressing out the box if we don't need to.

Johnetordoff: I would prefer to stay consistent with the other analytics gathering scripts and query this way, I think it's also better if we want to include more nuanced infomation in the future,

brianjgeiger: But the canonical information for preprints is stored in SHARE. Also, these scripts take a lot of resources to run. And we may change how preprints work in the near future to not be dependent on a node. So there are good reasons to go with SHARE.

What are you thinking by way of future nuance, and what information would not be available via SHARE?

Johnetordoff: I didn't know the canonical preprint information was kept in SHARE, it makes sense if Preprints aren't going to be dependent on a node. I'll look into it.

scripts/tests/test_preprint_summary.py
```diff
@@ -0,0 +1,37 @@
+import datetime
```
alexschiller: 1. scripts/tests are not run as part of travis (there purpose is to test that one-off scripts do what is intended). There isn't a great place for celery task tests atm, but Steve's recommendation for sitemap stuff was to add it to a new file in `osf_tests`

2. New test files should use pytest instead of nose

PR #7502, opened 2017-07-21
[Feature] Rework prereg page [OSF-7291]

alexschiller:
## Purpose
Prereg page needed streamlining. 

## Changes
1. Change main text per specs
2. Make page work with not logged in users, prompt to create account if click `Preregister` button

BEFORE:
screen shot 2017-07-21 at 7 07 33 pm
AFTER:
screen shot 2017-07-21 at 6 51 24 pm
AFTER Not logged in:
 
screen shot 2017-07-21 at 6 51 39 pm

BEFORE Mobile:
screen shot 2017-07-21 at 7 05 31 pm
AFTER Mobile: 
screen shot 2017-07-21 at 7 05 52 pm

 

## Side effects
1. Give buttons max width, single button on page took up width of the page before change
2. Make button language between small and large views the same
3. Make mobile buttons scale better and not cut off.
4. Fix button action drop downs to scale better with buttons (note they are not perfectly aligned in mobile, but are better than they were previously
5. ERPC mobile view still showed buttons to make a registration, which lead to a broken page. I removed those buttons. 
6. Furthermore, I split Prereg and ERPC from using the same template, since there was way too much jerryrigging occurring to make it work.

## Ticket
https://openscience.atlassian.net/browse/OSF-8291

## QA Notes
1. Make sure ERPC and Prereg pages work properly
2. Check all side effects to ensure they look right
~3. Make sure scaling works better than it did before, still not perfect for some things~
3. Scaling should work 100% now. All buttons and drops from buttons should align.
4. Ensure all buttons still work
5. Ensure clicking `Preregister` when not logged in redirects back to prereg (both for logging in and making a new account)

## Tests
I fixed some failing tests, removed one that was no longer relevant, and added one for not-logged-in users


website/templates/prereg_landing_page.mako
```diff
@@ -134,18 +95,25 @@
...
             <tbody>
               <tr>
-                ## Always displayed
+                %if is_logged_in:
                 <td class="col-sm-${ num_cols } prereg-button-col">
-                  <div class="prereg-button m-b-md p-md osf-box-lt p-md box-round prereg" data-qtoggle-group="prereg" data-qtoggle-target="#newPrereg">Start a new ${kind()}</div>
+                  <div class="prereg-button prereg-button-qtoggle m-b-md p-md osf-box-lt p-md box-round prereg" data-qtoggle-group="prereg" data-qtoggle-target="#newPrereg">Start a new pregistration</div>
                 </td>
+                %else:
+                <td class="col-sm-1 prereg-button-col">
```
alexschiller: Was going to include { num_cols } here but missed it. Will add if there are requested changes, but not a big enough deal since it can only ever be 1

PR #7504, opened 2017-07-24
Remove MODM from osf models and osf_tests [OSF-8314]

erinspace:
## Purpose
Since we've officially moved away from using modular ODM, let's take steps to also remove the backwards-compatible modm-like queries.

## Changes

- Default to using Django queries from OSF models in `osf/`
- Default to using Django from osf models tests in `osf_tests/`
- Residual updates in API views and serializers that were necessary to go along with these backend models changes

## Side effects
None anticipated, but when merged has the potential for many merge conflicts as this touches a lot of files

## Ticket

https://openscience.atlassian.net/browse/OSF-8314

### DIFF
--- a/osf/models/files.py
+++ b/osf/models/files.py
@@ -5,11 +5,10 @@ import os
 
 import requests
 from dateutil.parser import parse as parse_date
-from django.core.exceptions import ObjectDoesNotExist
 from django.db import models
 from django.db.models import Manager
+from django.core.exceptions import ObjectDoesNotExist
 from django.utils import timezone
-from modularodm.exceptions import NoResultsFound
 from typedmodels.models import TypedModel, TypedModelManager
 from include import IncludeManager
 
@@ -18,7 +17,6 @@ from osf.models.base import BaseModel, OptionalGuidMixin, ObjectIDMixin
 from osf.models.comment import CommentableMixin
 from osf.models.mixins import Taggable
 from osf.models.validators import validate_location
-from osf.modm_compat import Q
 from osf.utils.datetime_aware_jsonfield import DateTimeAwareJSONField
 from osf.utils.fields import NonNaiveDateTimeField
 from website.files import utils
@@ -184,9 +182,7 @@ class BaseFileNode(TypedModel, CommentableMixin, OptionalGuidMixin, Taggable, Ob
         materialized_path = '/' + materialized_path.lstrip('/')
         if materialized_path.endswith('/'):
             # it's a folder
-            folder_children = cls.find(Q('provider', 'eq', provider) &
-                                       Q('node', 'eq', node) &
-                                       Q('_materialized_path', 'startswith', materialized_path))
+            folder_children = cls.objects.filter(provider=provider, node=node, _materialized_path__startswith=materialized_path)
             for item in folder_children:
                 if item.kind == 'file':
                     guid = item.get_guid()
@@ -195,9 +191,10 @@ class BaseFileNode(TypedModel, CommentableMixin, OptionalGuidMixin, Taggable, Ob
         else:
             # it's a file
             try:
-                file_obj = cls.find_one(
-                    Q('node', 'eq', node) & Q('_materialized_path', 'eq', materialized_path))
-            except NoResultsFound:
+                file_obj = cls.objects.get(
+                    node=node, _materialized_path=materialized_path
+                )
+            except cls.DoesNotExist:
                 return guids
             guid = file_obj.get_guid()
             if guid:
@@ -219,7 +216,7 @@ class BaseFileNode(TypedModel, CommentableMixin, OptionalGuidMixin, Taggable, Ob
         :param user: The user with checked out files
         :return: A queryset of all FileNodes checked out by user
         """
-        return cls.find(Q('checkout', 'eq', user))
+        return cls.objects.filter(checkout=user)
 
     @classmethod
     def resolve_class(cls, provider, type_integer):
@@ -532,10 +529,7 @@ class Folder(models.Model):
         return child
 
     def find_child_by_name(self, name, kind=2):
-        return self.resolve_class(self.provider, kind).find_one(
-            Q('name', 'eq', name) &
-            Q('parent', 'eq', self)
-        )
+        return self.resolve_class(self.provider, kind).objects.get(name=name, parent=self)
 
     def serialize(self):
         return self._serialize()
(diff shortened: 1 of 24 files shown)


osf/utils/auth.py
```diff
@@ -30,27 +30,27 @@ def get_user(email=None, password=None, verification_key=None):
     query_list = []
     if email:
         email = email.strip().lower()
-        query_list.append(Q('emails__address', 'eq', email) | Q('username', 'eq', email))
+        query_list.append(Q(emails__address=email) | Q(username=email))
     if password:
         password = password.strip()
         try:
             query = query_list[0]
             for query_part in query_list[1:]:
```
caseyrollins: Am I correct in that `query_list` could only contain one query at this point (the `Q` from `if email:`) and this for loop can be removed?

website/project/utils.py
```diff
@@ -15,10 +16,7 @@ def serialize_node(*args, **kwargs):
     return _view_project(*args, **kwargs)  # Not recommended practice
 
 PROJECT_QUERY = (
-    # Can encompass accessible projects, registrations, or forks
-    # Note: is_bookmark collection(s) are implicitly assumed to also be collections; that flag intentionally omitted
-    Q('is_deleted', 'eq', False) &
-    Q('type', 'ne', 'osf.collection')
+    DjangoQ(is_deleted=False) & ~DjangoQ(type='osf.collection')
```
caseyrollins: I would actually make this just Q and change the MODM Q import since it will be removed later.

website/routes.py
```diff
@@ -60,9 +58,9 @@ def get_globals():
     location = geolite2.lookup(request.remote_addr) if request.remote_addr else None
     if request.host_url != settings.DOMAIN:
         try:
-            inst_id = (Institution.find_one(Q('domains', 'eq', request.host.lower())))._id
+            inst_id = Institution.objects.get(domains__contains=[request.host.lower()]).only('_id')._id
```
caseyrollins: I really really wanted this to be a thing but sadly it doesn't appear to be.

```
In [159]: Institution.objects.get(domains__contains=['test-osf.cos.io']).only('_id')._id
<ipython-input-159-0246ce42e76a> in <module>()
----> 1 Institution.objects.get(domains__contains=['test-osf.cos.io']).only('_id')._id

AttributeError: 'Institution' object has no attribute 'only'
```

caseyrollins: Also `__icontains` can be used in favor of `.lower()` here.

osf/models/preprint_provider.py
```diff
@@ -87,11 +85,11 @@ def rules_to_subjects(rules):
...
-                potential_parents = Subject.find(Q('parent', 'eq', Subject.load(rule[0][-1])))
+                potential_parents = Subject.objects.filter(parent=Subject.load(rule[0][-1]))
                 for parent in potential_parents:
-                    q.append(Q('parent', 'eq', parent))
+                    q.append(models.Q(parent=parent))
         for sub in rule[0]:
-            q.append(Q('_id', 'eq', sub))
-    return Subject.find(reduce(lambda x, y: x | y, q)) if len(q) > 1 else (Subject.find(q[0]) if len(q) else Subject.find())
+            q.append(models.Q(_id=sub))
+    return Subject.objects.filter(reduce(lambda x, y: x | y, q)) if len(q) > 1 else (Subject.objects.filter(q[0]) if len(q) else Subject.objects.all())
```
erinspace: We talked about this a bit in person, but adding it here for completeness!

This method used in the admin app tests and subject methods used right now for creating/editing Preprint Providers that don't have custom taxomomies. Since there are current tests that expect the objects instead of the ids, and this method and those tests will be removed as soon as everything has custom taxonomies and subjects to rules goes away entirely, leaving this for now.

osf/utils/auth.py
```diff
@@ -30,27 +30,21 @@ def get_user(email=None, password=None, verification_key=None):
...
-        query_list.append(Q('emails__address', 'eq', email) | Q('username', 'eq', email))
+        query_list.append(Q(emails__address=email) | Q(username=email))
     if password:
         password = password.strip()
         try:
-            query = query_list[0]
-            for query_part in query_list[1:]:
-                query = query & query_part
-            user = User.find_one(query)
+            user = User.objects.get(query_list[0])
```
sloria: Is this equivalent to the old code? Looks like the old code was `&`-ing all the query parts.

osf_tests/test_archiver.py
```diff
@@ -282,7 +281,7 @@ def from_question(qid, question):
     )
     try:
         schema.save()
-    except KeyExistsException:
+    except IntegrityError:
```
sloria: Good catch.

osf_tests/test_framework_utils.py
```diff
@@ -1,52 +0,0 @@
...
-from tests.base import DbTestCase
-from osf_tests import factories
-
-from framework.database import get_or_http_error, autoload
-from framework.exceptions import HTTPError
-
-from osf.models import Node
-
-
-class FrameworkUtilsTestCase(DbTestCase):
```
sloria: Why were these tests deleted?

website/routes.py
```diff
@@ -126,13 +124,11 @@ def get_globals():
...
+    # Avoid circular import
+    from osf.models import PrivateLink
     try:
-        # Avoid circular import
-        from osf.models import PrivateLink
-        return PrivateLink.find_one(
-            Q('key', 'eq', request.args.get('view_only'))
-        ).anonymous
-    except QueryException:
+        return PrivateLink.objects.get(key=request.args.get('view_only'))
```
sloria: This needs to return the value of `anonymous`:

```python
PrivateLink.objects.filter(key=....).values_list('anonymous', flat=True).get()
```

osf/models/files.py
```diff
@@ -673,11 +667,10 @@ def _find_matching_archive(self, save=True):
...
-            Q('_id', 'ne', self._id) &
-            Q('metadata.sha256', 'eq', self.metadata['sha256']) &
-            Q('metadata.archive', 'ne', None) &
-            Q('metadata.vault', 'ne', None)
+        other = self.__class__.objects.filter(
+            metadata__sha256=self.metadata['sha256']
+        ).exclude(
+            _id=self._id, metadata__archive__is_null=True, metadata__vault__is_null=True
         ).first()
         if not other:
```
erinspace: Oops! Totally missed this in the first CR somehow, my bad!

osf_tests/test_framework_utils.py
erinspace: This is totally a mistake on my part I think -- meant to undo this and add to ticket for final pass of removing odm. Sorry bout that.

sloria:

PR #7506, opened 2017-07-24
[OSF-8228] OSF-side Navbar cleanup (firefox username alignment, long username fix)

Johnetordoff:
## Purpose

Current there are two problems with the OSF navbar:

1. The username name text and gravtar are silghtly out of vertical alignment
2. Very long usernames (even though they are wrapped properly) can break the navbar into two unsightly rows.
Alignment problem:
screen shot 2017-07-05 at 3 19 15 pm
After:
screen shot 2017-08-16 at 1 54 34 pm

Row problem:
screen shot 2017-07-05 at 3 24 23 pm
After:
screen shot 2017-08-16 at 1 56 53 pm

## Changes

Simple CSS and HTML changes, now the navbar will collapse into a dropdown sooner and the alignment is fixed with style tags.

## Side effects

Should be limited to aesthetics. I used style tags where possible to avoid any unforeseeable interaction.

## Tests

For aesthetics changes I can't think of any meaningful tests.

## Ticket

https://openscience.atlassian.net/browse/OSF-8228


Johnetordoff:
@brianjgeiger Yes, but I don't know how we'll handle that, I'm unclear how OSF-style css gets into osfio.

brianjgeiger:
It's in the package.json. https://github.com/CenterForOpenScience/osf.io/blob/276805f69f70ac534b3a67a47eb3ff832a069e54/package.json#L59-L59 Basically the process will be, we update osf-style with the new style, then we update the version of osf-style we're using while simultaneously removing the style from the OSF.

Johnetordoff:
osf-style changes have already been merged, are css changes here still necessarily or should I remove them?

brianjgeiger:
If we're using the proper version of osf-style, then you should be able to remove the styles in the OSF and it should work. If not, we need to upgrade which version of osf-style we're using. But yes, I would like these not to be in the osf project if they're unnecessary.

website/static/css/style.css
```diff
@@ -189,6 +189,49 @@ a {
     }
 }
 
+
+@media (max-width: 992px) {
```
brianjgeiger: Since we're linking to a newer version of osf style, we don't need this, right?

package.json
```diff
@@ -58,7 +58,7 @@
     "mongodb": "^1.4.40",
     "morgan": "^1.5.1",
     "object-assign": "^2.0.0",
-    "@centerforopenscience/osf-style": "^1.3.2",
+    "@centerforopenscience/osf-style": "git://github.com/CenterForOpenScience/osf-style.git#6e1b6fb35a5dfd03530f2b8e55e0a27e01b9d080",
```
jamescdavis: This commit has been merged in to osf-style as of 1.4.0

website/templates/nav.mako
```diff
@@ -48,12 +48,12 @@
             <li class="navbar-donate-button"><a data-bind="click: trackClick.bind($data, 'Donate')" href="https://cos.io/donate">Donate</a></li>
             % if user_name and display_name:
             <li class="dropdown">
-            <button class="dropdown-toggle nav-user-dropdown btn-link" data-toggle="dropdown" role="button" aria-expanded="false" aria-label="Toggle auth dropdown">
+            <a class="dropdown-toggle nav-user-dropdown btn-link" data-toggle="dropdown" role="button" aria-expanded="false" aria-label="Toggle auth dropdown">
```
jamescdavis: This causes the nav bar to be too high because bootstrap adds 15px top and bottom padding around `.navbar-nav > li > a`. One solution will be to adjust the line height in osf-style to 27px (the height of the gravatar image) and reduce top and bottom padding accordingly.

package.json
jamescdavis: This should be updated to osf-style ^1.5.1 to include the latest padding fixes for the navbar

website/templates/nav.mako
jamescdavis: this is fixed in osf-style 1.5.1

Johnetordoff:
@jamescdavis was there a resolution to this?

jamescdavis:
Change `<button>` to `<a>` and use osf-style@1.5.1

brianjgeiger:
It's set to use `1.5.1`, so should just be the `button` to `a`, yes?

Johnetordoff:
@jamescdavis I changed the button and fixed the alignment as well. They navbars and now pixel-for-pixel exactly the same.

website/templates/nav.mako
jamescdavis: You can actually just drop the `nav-user-dropdown` class and let osf-style set the padding. That's what we're doing in Ember.

brianjgeiger:
@jamescdavis Does this meet your approval?

PR #7507, opened 2017-07-25
[DEV-239] Update to node.js version 8

binoculars:
## Purpose

Updates Node.js to a maintained version

## Changes

* Add .npmrc to ignore optional dependencies - would use yarn, but yarnpkg/yarn#2666 is needed before we can use it. `livedb`'s optional dependency, `hiredis` will not build on newer node.js versions.
* Update Dockerfile to use node.js version 8
  * `npm cache clean` requires `--force` flag

## Side effects

* May affect the wiki / ShareJS, so we'll need to regression test that
* Assets should be unaffected, but again, something to watch out for

## Ticket

https://openscience.atlassian.net/browse/DEV-239

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,10 +44,59 @@ RUN apt-get update \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
-RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - \
-    && apt-get install -y nodejs \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/*
+# https://github.com/nodejs/docker-node/blob/9c25cbe93f9108fd1e506d14228afe4a3d04108f/8.2/Dockerfile
+# gpg keys listed at https://github.com/nodejs/node#release-team
+RUN set -ex \
+  && for key in \
+    9554F04D7259F04124DE6B476D5A82AC7E37093B \
+    94AE36675C464D64BAFA68DD7434390BDBE9B9C5 \
+    FD3A5288F042B6850C66B31F09FE44734EB7990E \
+    71DCFD284A79C3B38668286BC97EC7A07EDE3FC1 \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -34,37 +35,18 @@ RUN apt-get update \
...
-    && gpg --keyserver pool.sks-keyservers.net --recv-keys 7937DFD2AB06298B2293C3187D33FF9D0246406D 114F43EE0176B71C7BC219DD50A3051F888C628D \
-    && curl -SLO "http://nodejs.org/dist/v$NODE_VERSION/node-v$NODE_VERSION-linux-x64.tar.gz" \
-    && curl -SLO "http://nodejs.org/dist/v$NODE_VERSION/SHASUMS256.txt.asc" \
-    && gpg --verify SHASUMS256.txt.asc \
-    && grep " node-v$NODE_VERSION-linux-x64.tar.gz\$" SHASUMS256.txt.asc | sha256sum -c - \
-    && tar -xzf "node-v$NODE_VERSION-linux-x64.tar.gz" -C /usr/local --strip-components=1 \
-    && rm "node-v$NODE_VERSION-linux-x64.tar.gz" SHASUMS256.txt.asc \
-    && npm install -g npm@"$NPM_VERSION" \
-    && npm cache clear \
+RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - \
```
icereval: `shell`'ing scripts directly from the internet is a bit scary. If this script is really worth using rather than the [docker hub image](https://github.com/nodejs/docker-node/blob/9c25cbe93f9108fd1e506d14228afe4a3d04108f/8.2/Dockerfile), please add a sha256 check directly to the Dockerfile, there should be examples in the docker-library repo.

binoculars: Changed it to use that--we can optimize later with multi-stage builds later on once jenkins with modern docker is ready.

PR #7515, opened 2017-07-26
[OSF-8247] Send users an email about fork fail/success

Johnetordoff:
## Purpose

Sometimes making a fork can take a really long time, this PR give the user notice via an email when their for is complete.

## Changes

Adds email template and adds to API so it sends email whenever an when ever a fork is complete or fails, also displays growl box alerting the user about the email and get rid of the old modals.

## Side effects

None that I know of.

## Ticket


alexschiller:
@Johnetordoff Driveby comment: OSF-8247 not OSF-8248

api/nodes/views.py
```diff
@@ -1652,7 +1653,14 @@ def get_queryset(self):
 
     # overrides ListCreateAPIView
     def perform_create(self, serializer):
-        serializer.save(node=self.get_node())
+        user = get_user_auth(self.request).user
+        node = self.get_node()
+        try:
+            fork = serializer.save(node=node)
+            mails.send_mail(user.email, mails.FORK_COMPLETED, title=fork.title, guid=fork._id, mimetype='html')
```
alexschiller: This is going to send an email every time someone makes a fork. Can you confirm with product this is intentional? That seems incorrect.

website/templates/emails/fork_failed.html.mako
```diff
@@ -0,0 +1,19 @@
...
+  <td style="border-collapse: collapse;">
+
+    <h3 class="text-center" style="padding: 0;margin: 30px 0 0 0;border: none;list-style: none;font-weight: 300;text-align: center;">
+        An error has occurred, and the fork of <b>${title}</b> on The Open Science Framework was not successfully created. Please log in and try this action again. If the problem persists, please email support@osf.io.
+    </h3>
+  </td>
+</tr>
+<tr>
+  <td style="border-collapse: collapse;">
+    You can view the fork here <a href="${settings.DOMAIN + guid}">here.</a>
```
alexschiller: You don't pass guid to this template so this is going to fail... which means you need tests.

website/static/js/project.js
```diff
@@ -61,38 +60,8 @@ NodeActions.forkNode = function() {
...
-            } else if (response.status === 504) {
-                $osf.growl('Sorry:', 'This is taking longer than normal. </br>' +
-                    'Please check back later to access your new fork and if still unavailable, contact support@cos.io');
-            } else {
-                $osf.growl('Error:', 'Forking failed');
-                Raven.captureMessage('Error occurred during forking');
-            }
-        });
+        );
+        $osf.growl('Fork status', 'Your fork is being forked you\'ll get an email when it\'s complete.', 'info');
```
alexschiller: This is a pretty significant user experience change. Why did this bypass product demo?

website/static/js/project.js
```diff
@@ -61,38 +60,8 @@ NodeActions.forkNode = function() {
...
-                    },
-                    cancel: {
-                        label: 'Keep working here'
-                    }
-                },
-                closeButton: false
-            });
-        }).fail(function(response) {
-            $osf.unblock();
-            if (response.status === 403) {
```
alexschiller: By removing these checks, the user will never know why the forking failed if they don't have permission. Seems less than ideal.

api/nodes/views.py
```diff
@@ -1652,7 +1653,14 @@ def get_queryset(self):
...
+        except Exception, exc:
```
alexschiller: Very minor; the standard pattern used in the osf is except <exception> as <error>

website/static/js/project.js
Johnetordoff: I don't know, does everything product proposed get a demo? This seems mostly as though it's removing stuff, instead of modals you get a growl box then an email, I need more clarity on what gets demoed because I am waiting 1 to 2 weeks every time something is demoed and so I try to avoid it unless it's specifically in the ticket.

website/static/js/project.js
Johnetordoff: It's assumed the user isn't going to be hanging around for the response, if we make AJAX wait on a response it won't complete the action. Also it's redundant client-side stuff because someone with out permission shouldn't even be able to click that button.

It's consistent if you imagine the forks taking a relatively long time to complete (like 5 minutes) nobody's going to want to stick around for 5 minutes to discover they got an error. But the failure email should have that info.

api/nodes/views.py
Johnetordoff: We could try making this something that happens with a query parameter, but that would essentially be special casing it, more than it is already. 

It's uncomfortable implementing this type of notification without a more generalized system which doesn't just apply to only forks, but that's not the scope of the ticket.

website/static/js/project.js
alexschiller: From BrianG in the B/I flow on July 4: "Also, don’t forget: if it’s not a purely API/Backend change, then you’re probably going to need to go through a Product Demo. Especially if there’s UI changes they requested."

api_tests/nodes/views/test_node_forks_list.py
```diff
@@ -318,3 +322,18 @@ def test_read_only_contributor_can_fork_private_registration(self, app, private_
...
+        assert mails.send_mail.called_with(user.email, mails.FORK_COMPLETED, title= res.json['data']['attributes']['title'], guid=res.json['data']['id'], mimetype='html')
+
+    def test_send_email_failed(self, app, user, public_project_url, fork_data_with_title, public_project):
+        NodeForksSerializer.save = mock.MagicMock()
+        NodeForksSerializer.save.side_effect = Exception()
+
+        with pytest.raises(Exception):
+            app.post_json_api(public_project_url, fork_data_with_title, auth=user.auth)
+
+        assert mails.send_mail.called_with(user.email, mails.FORK_COMPLETED, title= fork_data_with_title['data']['attributes']['title'], guid=public_project.id, mimetype='html')
```
alexschiller: Minor; New line at end of file

alexschiller: Also extra space in `... title= fork...`

api_tests/nodes/views/test_node_forks_list.py
```diff
@@ -10,8 +11,11 @@
     ForkFactory
 )
 from rest_framework import exceptions
+from website import mails
 from website.util import permissions
 
+from api.nodes.serializers import NodeForksSerializer
+mails.send_mail = mock.MagicMock()
```
alexschiller: Should this use the decorator?

I've seen cases where using a patcher that is not stopped can linger beyond the test class and I don't know if mocking like this can do the same.

api_tests/nodes/views/test_node_forks_list.py
```diff
@@ -318,3 +322,18 @@ def test_read_only_contributor_can_fork_private_registration(self, app, private_
...
         assert res.json['data']['id'] == private_project.forks.first()._id
         assert res.json['data']['attributes']['title'] == 'Fork of ' + private_project.title
+
+    def test_send_email_success(self, app, user, public_project_url, fork_data_with_title, public_project):
+        res = app.post_json_api(public_project_url, fork_data_with_title, auth=user.auth)
+        assert res.status_code == 201
+        assert res.json['data']['id'] == public_project.forks.first()._id
+        assert mails.send_mail.called_with(user.email, mails.FORK_COMPLETED, title= res.json['data']['attributes']['title'], guid=res.json['data']['id'], mimetype='html')
+
+    def test_send_email_failed(self, app, user, public_project_url, fork_data_with_title, public_project):
```
alexschiller: Can you confirm that this test failed before you fixed the mail? It checks if the mail function is called but since it is mocked it isn't clear that it would catch issues.

api/nodes/views.py
```diff
@@ -1660,7 +1662,14 @@ def get_queryset(self):
...
     # overrides ListCreateAPIView
     def perform_create(self, serializer):
-        serializer.save(node=self.get_node())
+        user = get_user_auth(self.request).user
+        node = self.get_node()
+        try:
+            fork = serializer.save(node=node)
+            mails.send_mail(user.email, mails.FORK_COMPLETED, title=fork.title, guid=fork._id, mimetype='html')
+        except Exception as exc:
+            mails.send_mail(user.email, mails.FORK_FAILED, title=node.title, guid=fork._id, mimetype='html')
```
alexschiller: I think guid needs to be `node._id` since to get here fork would have to fail to save. Did it work locally if you raise an error in serializer?

api/nodes/views.py
```diff
@@ -1660,7 +1662,14 @@ def get_queryset(self):
 
     # overrides ListCreateAPIView
     def perform_create(self, serializer):
-        serializer.save(node=self.get_node())
+        user = get_user_auth(self.request).user
+        node = self.get_node()
+        try:
+            fork = serializer.save(node=node)
+            mails.send_mail(user.email, mails.FORK_COMPLETED, title=fork.title, guid=fork._id, mimetype='html')
```
alexschiller: Also, it might make sense to use the try-except-else so we don't accidentally send a failed fork email if the fork_completed email fails.

api_tests/nodes/views/test_node_forks_list.py
Johnetordoff: This is consistent with how the MagicMock is used on the rest of the site, this isn't really a patch per say.

alexschiller: Talked in person, summary of comment:
1. Mostly concerned about it being top level-- every other MagicMock is nested within a method in a class AFAICT.
2. The other cases of mocking send_mail do use the decorator, should work but if it isn't working I can take a look.

PR #7518, opened 2017-07-26
[Feature] Make `ever_mentioned` a ManyToMany [#OSF-7933]

mfraezz:
## Purpose
Allow for better referential integrity, simpler/optimized querying

## Changes
* Add migration
* Update app / tests to use M2M

## Side effects
None expected

## Ticket
[[OSF-7933]](https://openscience.atlassian.net/browse/OSF-7933)


osf/migrations/0043_ever_mentioned_array_to_m2m.py
```diff
@@ -0,0 +1,48 @@
...
+# Generated by Django 1.11.2 on 2017-07-25 17:16
+from __future__ import unicode_literals
+
+from django.conf import settings
+from django.db import migrations, models
+
+
+def migrate_user_guid_array_to_m2m(state, schema):
+    from osf.models import OSFUser as FindableOSFUser
+    AddableOSFUser = state.get_model('osf', 'osfuser')
```
mfraezz: Note: This is _weird_.

Django seems to expect a "raw" (obtained via `state.get_model`) django model instance for M2M changes in migrations, but these don't have mixed-in things like `load` or `guids`, which makes finding the right one difficult. The imported class has these properties, but is rejected as the wrong type when `.add`'ed to a M2M.

osf/migrations/0044_ever_mentioned_array_to_m2m.py
```diff
@@ -0,0 +1,48 @@
...
+# Generated by Django 1.11.2 on 2017-07-25 17:16
+from __future__ import unicode_literals
+
+from django.conf import settings
+from django.db import migrations, models
+
+
+def migrate_user_guid_array_to_m2m(state, schema):
+    from osf.models import OSFUser as FindableOSFUser
+    AddableOSFUser = state.get_model('osf', 'osfuser')
```
mfraezz: Note: This is _weird_.

Django seems to expect a "raw" (obtained via `state.get_model`) django model instance for M2M changes in migrations, but these don't have mixed-in things like `load` or `guids`, which makes finding the right one difficult. The imported class has these properties, but is rejected as the wrong type when `.add`'ed to a M2M.

PR #7520, opened 2017-07-26
[OSF-8307] Add tests for sitemap generation.

adlius:
## Purpose

This PR aims to add unit tests to the sitemap generation script.

## Changes

scripts/generate_sitemap.py is changed so that the url method for nodes and users are used to generate links.

test_generate_sitemap.py is added as unit tests.

## Ticket

https://openscience.atlassian.net/browse/OSF-8307

### DIFF
--- a/osf_tests/test_generate_sitemap.py
+++ b/osf_tests/test_generate_sitemap.py
@@ -2,10 +2,14 @@ import os
 
 import pytest
 import mock
-import shutil, tempfile, xml, urlparse
+import shutil
+import tempfile
+import xml
+import urlparse
 
 from scripts import generate_sitemap
-from osf_tests import factories
+from osf_tests.factories import (AuthUserFactory, ProjectFactory, RegistrationFactory, CollectionFactory,
+                                 PreprintFactory, PreprintProviderFactory, EmbargoFactory)
 from website import settings
 
 
@@ -15,7 +19,7 @@ def get_all_sitemap_urls():
     generate_sitemap.main()
 
     # Parse the generated XML sitemap file
-    with open(settings.STATIC_FOLDER + '/sitemaps' + '/sitemap_0.xml') as f:
+    with open(os.path.join(settings.STATIC_FOLDER, 'sitemaps/sitemap_0.xml')) as f:
         tree = xml.etree.ElementTree.parse(f)
 
     shutil.rmtree(settings.STATIC_FOLDER)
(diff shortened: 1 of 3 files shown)


adlius:
@caseyrollins Thanks for the notice on optimization! Now it makes perfect sense. I have reverted the changes made, and have added a minor change to the script to remove the trailing slash for user urls.

scripts/tests/test_generate_sitemap.py
```diff
@@ -0,0 +1,113 @@
+import pytest
```
Nesiehr: Follow import style guidelines (http://cosdev.readthedocs.io/en/latest/style_guides/python.html#imports)

scripts/tests/test_generate_sitemap.py
```diff
@@ -0,0 +1,113 @@
...
+
+import shutil, tempfile, xml, os
+from urlparse import urljoin
+from website import settings
+from scripts.generate_sitemap import main
+from osf_tests.factories import AuthUserFactory, PreprintProviderFactory, PreprintFactory, ProjectFactory, \
+    RegistrationFactory, CollectionFactory
+
+@pytest.mark.django_db
+class TestGenerateSitemap:
```
Nesiehr: use pytest.fixture() to create fixtures for reusable components (look through api_tests/nodes/views files for examples)

scripts/tests/test_generate_sitemap.py
```diff
@@ -0,0 +1,113 @@
...
+    def test_all_links_are_included(self):
+        # Create temporary directory for the sitemaps to be generated
+        temp_dir = tempfile.mkdtemp()
+
+        # Set static folder to the path of temp_dir, so that the generator would store the sitemaps under
+        # <path_to_temp_dir>/sitemaps/
+        settings.STATIC_FOLDER = temp_dir
+
+        # Test setup
+        user_1 = AuthUserFactory()
```
Nesiehr: pytest naming guidelines:
user_one > user_1, project_one > project_1, etc. (these are also typically fixtures)

if possible try to avoid using numbers though, if a user is a write contributor a better name than user_two would be user_write_contrib

if the differentiator between projects is whether they are public or private better names would be project_private and project_public instead of project_one and project_two

scripts/tests/test_generate_sitemap.py
```diff
@@ -0,0 +1,113 @@
...
+        preprint_2 = PreprintFactory(provider=provider_2, project=project_2, creator=user_2)
+
+        # Constructed list of urls that should be included
+        list_of_urls = [urljoin(settings.DOMAIN, item['loc']) for item in settings.SITEMAP_STATIC_URLS]
+        list_of_urls.extend([urljoin(settings.DOMAIN, user_1._id), urljoin(settings.DOMAIN, user_2._id)])
+        list_of_urls.extend([urljoin(settings.DOMAIN, project_1._id),
+                             urljoin(settings.DOMAIN, project_2._id),
+                             urljoin(settings.DOMAIN, registration_1._id),
+                             urljoin(settings.DOMAIN, registration_2._id)])
+        list_of_urls.extend([settings.DOMAIN + 'preprints/' + preprint_1._id + '/',
```
Nesiehr: use ".format()" over "+" for string concatenation

scripts/tests/test_generate_sitemap.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # Test setup
+        user_1 = AuthUserFactory()
+        user_2 = AuthUserFactory()
+        project_1 = ProjectFactory(creator=user_1)
+        project_2 = ProjectFactory(creator=user_2)
+        registration_1 = RegistrationFactory(project=project_1, creator=user_1, is_public=True)
+        registration_2 = RegistrationFactory(project=project_2, creator=user_2, is_public=True)
+        provider_1 = PreprintProviderFactory(_id='osf', name='osfprovider')
+        provider_2 = PreprintProviderFactory(_id='adl', name="anotherprovider")
```
Nesiehr: Use ' instead of "

scripts/tests/test_generate_sitemap.py
```diff
@@ -0,0 +1,113 @@
...
+        temp_dir = tempfile.mkdtemp()
+
+        # Set static folder to the path of temp_dir, so that the generator would store the sitemaps under
+        # <path_to_temp_dir>/sitemaps/
+        settings.STATIC_FOLDER = temp_dir
+
+        # Test setup
+        user_1 = AuthUserFactory()
+        user_2 = AuthUserFactory()
+        project_1 = ProjectFactory(creator=user_1)
```
alexschiller: It looks like these implicitly become public when you create a preprint off of them. Better to be explicit.

scripts/tests/test_generate_sitemap.py
```diff
@@ -0,0 +1,113 @@
...
+                            '?action=download'
+                        )
+
+        list_of_urls.extend([url_preprint_file_1, url_preprint_file_2])
+
+        # Run the script
+        main()
+
+        # Parse the generated XML sitemap file
+        with open(temp_dir + '/sitemaps' + '/sitemap_0.xml') as f:
```
alexschiller: second +  is unneeded

scripts/tests/test_generate_sitemap.py
```diff
@@ -0,0 +1,113 @@
+import pytest
+
```
adlius: So it should stay that way right?

alexschiller: No, it should be running with travis since it is this script runs nightly via celery. Per Steve, move it to the `osf_tests` folder.

adlius: Understood.

scripts/generate_sitemap.py
```diff
@@ -177,7 +177,7 @@ def generate(self):
         for obj in objs:
             try:
                 config = settings.SITEMAP_NODE_CONFIG
-                config['loc'] = urlparse.urljoin(settings.DOMAIN, obj['guids___id'])
+                config['loc'] = urlparse.urljoin(settings.DOMAIN, '/{}/'.format(obj['guids___id']))
```
alexschiller: :+1:

osf_tests/test_generate_sitemap.py
```diff
@@ -0,0 +1,164 @@
+import os
+
+import pytest
+import mock
+import shutil, tempfile, xml, urlparse
+
+from scripts import generate_sitemap
+from osf_tests import factories
```
adlius: I am a bit torn here between the conventions shown in osf_tests and the import guide as specified in the dev docs (http://cosdev.readthedocs.io/en/latest/style_guides/python.html#imports), which says we should import the entire module rather than a symbol in a module.

alexschiller: Rules break down when there is a norm in the codebase:
screen shot 2017-07-31 at 10 01 16 am
screen shot 2017-07-31 at 10 01 44 am

adlius: I see. Will fix that.

osf_tests/test_generate_sitemap.py
```diff
@@ -0,0 +1,164 @@
...
+    namespace = '{http://www.sitemaps.org/schemas/sitemap/0.9}'
+    urls = [element.text for element in tree.iter(namespace + 'loc')]
+
+    return urls
+
+
+@pytest.mark.django_db
+class TestGenerateSitemap:
+
+    @pytest.fixture(autouse=True)
```
Nesiehr: I don't usually use the autouse fixture, but if it something that needs to be setup for all of the tests in the file it probably doesn't hurt to have that

osf_tests/test_generate_sitemap.py
```diff
@@ -0,0 +1,164 @@
...
+from website import settings
+
+
+def get_all_sitemap_urls():
+    # Create temporary directory for the sitemaps to be generated
+
+    generate_sitemap.main()
+
+    # Parse the generated XML sitemap file
+    with open(settings.STATIC_FOLDER + '/sitemaps' + '/sitemap_0.xml') as f:
```
sloria: Also, use `os.path.join` to create paths.

osf_tests/test_generate_sitemap.py
```diff
@@ -0,0 +1,164 @@
...
+    urls = [element.text for element in tree.iter(namespace + 'loc')]
+
+    return urls
+
+
+@pytest.mark.django_db
+class TestGenerateSitemap:
+
+    @pytest.fixture(autouse=True)
+    def public_project_owner(self):
```
Nesiehr: user_admin_project_public and user_admin_project_private ideally

osf_tests/test_generate_sitemap.py
```diff
@@ -0,0 +1,168 @@
...
+    @pytest.fixture(autouse=True)
+    def user_admin_project_public(self):
+        return AuthUserFactory()
+
+    @pytest.fixture(autouse=True)
+    def user_admin_project_private(self):
+        return AuthUserFactory()
+
+    @pytest.fixture(autouse=True)
+    def public_project_for_registration(self, user_admin_project_public):
```
alexschiller: I would prefer if all of the fixture were updated to match the naming convention discussed above, but I won't hold this up further.

brianjgeiger: Shouldn't take long to change. Might as well match convention.

adlius:
Fixture names are changed. Waiting for Travis to pass before move to CR again.

alexschiller:
@adlius names look great, thanks 👍

PR #7521, opened 2017-07-27
[feature fix] admin provider import export [OSF-8236]

laurenrevere:
## Purpose

Branded preprint provider taxonomies and licenses were not exporting/importing correctly between the staging and production admin app. 

## Changes

This changes subjects and licenses so they are looked up by name rather than by ID.

## Side effects

Due to custom taxonomies (and how hard it would be to update them), when attempting to update a preprint provider, if subjects for that provider already exist, the subjects from the import json will be ignored.

## Ticket

https://openscience.atlassian.net/browse/OSF-8236

### DIFF
--- a/admin/preprint_providers/views.py
+++ b/admin/preprint_providers/views.py
@@ -179,7 +179,7 @@ class ExportPreprintProvider(PermissionRequiredMixin, View):
         return response
 
     def serialize_subjects(self, provider):
-        if provider._id != "osf":
+        if provider._id != 'osf':
             result = {}
             result['include'] = []
             result['exclude'] = []
@@ -235,7 +235,7 @@ class ImportPreprintProvider(PermissionRequiredMixin, View):
             # make sure not to import an exported access token for SHARE
             cleaned_result = {key: value for key, value in file_json['fields'].iteritems() if key not in FIELDS_TO_NOT_IMPORT_EXPORT}
             preprint_provider = self.create_or_update_provider(cleaned_result)
-            return redirect('preprint_providers:detail', preprint_provider_id = preprint_provider.id)
+            return redirect('preprint_providers:detail', preprint_provider_id=preprint_provider.id)
 
     def parse_file(self, f):
         parsed_file = ''
--- a/admin/preprint_providers/views.py
+++ b/admin/preprint_providers/views.py
@@ -4,11 +4,12 @@ import json
 
 from django.core import serializers
 from django.core.urlresolvers import reverse_lazy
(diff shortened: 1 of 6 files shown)


admin/preprint_providers/views.py
```diff
@@ -216,14 +231,47 @@ def post(self, request, *args, **kwargs):
...
             parsed_file += str(chunk)
         return parsed_file
 
+    def get_page_provider(self):
+        page_provider_id = self.kwargs.get('preprint_provider_id', '')
+        if page_provider_id:
+            return PreprintProvider.objects.get(id=page_provider_id)
+
+    def add_subjects(self, provider, subject_data):
+        from osf.management.commands.populate_custom_taxonomies import migrate
```
brianjgeiger: This concerns me a little. Maybe it's a common Django thing, but I feel like pulling from a management command into web app code is the wrong way around. Ideally, the code would exist inside the web app, and the management command would call that code instead.

That being said, I don't think we'll change it for this, but I will almost certainly be making a refactor follow-up ticket.

admin_tests/preprint_providers/test_views.py
```diff
@@ -219,6 +229,59 @@ def test_certain_fields_not_included(self):
         for field in views.FIELDS_TO_NOT_IMPORT_EXPORT:
             nt.assert_not_in(field, content_dict['fields'].keys())
 
+    def test_export_to_import_new_provider(self):
+        update_taxonomies('test_bepress_taxonomy.json')
```
brianjgeiger: Similarly to the management command concern, this is a little weird. We should be able to send a dictionary to a function to update the taxonomy, and then we could call a test without having to make a test file. Anyways, no change here, just thinking.

osf/management/commands/populate_custom_taxonomies.py
```diff
@@ -1,8 +1,6 @@
 import json
 import logging
 
-import django
```
brianjgeiger: And this doesn't break the management command's normal usage, right?

laurenrevere: Nope!

admin/preprint_providers/views.py
laurenrevere: Valid, I agree, this should be refactored.

admin_tests/preprint_providers/test_views.py
laurenrevere: Makes sense 👍

PR #7522, opened 2017-07-27
[feature] Pytest Conversion: Node Part 5 [OSF-8292]

Nesiehr:
## Purpose

Testing Performance Increase

## Changes

Pytest conversion of far too many large node files
Pytest conversion of two already converted mixins to follow the new pytest mixin format

## Side effects

None known.

## Ticket

https://openscience.atlassian.net/browse/OSF-8292

## QA Considerations

N/A


api_tests/nodes/views/test_node_linked_registrations.py
```diff
@@ -7,421 +7,463 @@
...
-        }
-
-    def make_request(self, node_id=None, auth=None, reg_id=None, expect_errors=False):
-        url = '/{}nodes/{}/relationships/linked_registrations/'.format(API_BASE, node_id)
-        if auth:
-            return self.app.post_json_api(url, self.make_payload(registration_id=reg_id), auth=auth, expect_errors=expect_errors)
-        return self.app.post_json_api(url, self.make_payload(registration_id=reg_id), expect_errors=expect_errors)
-
-    def test_admin_contributor_can_create_linked_registrations_relationship(self):
+    @pytest.fixture()
```
alexschiller: Not sure why these need to be fixtures, does it save any time? Looks like it could just be a payload function not wrapped, though no strong feelings if you decide not to change.

api_tests/registrations/views/test_registration_list.py
```diff
@@ -10,6 +11,7 @@
 from osf.models import AbstractNode as Node, MetaSchema, DraftRegistration
 from website.views import find_bookmark_collection
 from framework.auth.core import Auth, Q
```
alexschiller: Another q?

api_tests/registrations/views/test_registration_list.py
```diff
@@ -10,6 +11,7 @@
...
+from django.db.models import Q
```
alexschiller: confirm used, others weren't

api_tests/registrations/views/test_registration_list.py
```diff
@@ -457,381 +459,375 @@ def test_incorrect_filtering_field_not_logged_in(self):
...
-    def setUp(self):
-        super(TestRegistrationCreate, self).setUp()
 
-        self.schema = MetaSchema.find_one(
-            Q('name', 'eq', 'Replication Recipe (Brandt et al., 2013): Post-Completion') &
-            Q('schema_version', 'eq', LATEST_SCHEMA_VERSION)
-        )
+    @pytest.fixture()
+    def schema(self):
+        return MetaSchema.objects.filter(name='Replication Recipe (Brandt et al., 2013): Post-Completion', schema_version=LATEST_SCHEMA_VERSION).first()
```
alexschiller: get

api_tests/nodes/views/test_node_detail.py
```diff
@@ -25,1032 +18,1151 @@
...
+    @pytest.fixture()
+    def permissions_read(self):
+        return ['read']
+
+    @pytest.fixture()
+    def permissions_write(self):
+        return ['read', 'write']
+
+    @pytest.fixture()
+    def permissions_admin(self):
```
Nesiehr: hmm... not sure? They are only used in that one class, so they could be moved to the top of the file or left where they are, which would you prefer?

alexschiller: I think I prefer top of the file then as defaults, though might let @sloria weigh in before changing.

api_tests/nodes/views/test_node_linked_registrations.py
Nesiehr: I don't believe so, but I think it does make it so that you have to include it as an argument instead of using self.payload? Might be wrong

api_tests/nodes/views/test_node_contributors_list.py
```diff
@@ -1,95 +1,127 @@
...
+    def title(self):
+        return 'Cool Project'
 
-        self.private_project = ProjectFactory(title=self.title,
-                                              description=self.description,
-                                              category=self.category,
-                                              is_public=False,
-                                              creator=self.user)
-        self.private_url = '/{}nodes/{}/'.format(API_BASE, self.private_project._id)
+    @pytest.fixture()
```
sloria: Are these title and description fixtures used in more than one place? Might be overuse of fixtures. Not a blocker, though.

PR #7523, opened 2017-07-27
[Feature] Convert api_tests/collections to pytest[OSF-8328]

fabmiz:
## Purpose

Convert api_tests/collections tests from nose to pytest

## Changes

Convert api collections `views`, `serializers` to pytest

## Ticket

[OSF-8328](https://openscience.atlassian.net/browse/OSF-8328)


api_tests/collections/test_views.py
```diff
@@ -1,174 +1,184 @@
...
+from tests.utils import assert_logs, assert_items_equal
+from website.project.signals import contributor_removed
+from website.util.sanitize import strip_html
+from website.util import disconnected_from_listeners
 from website.views import find_bookmark_collection
 
 
-def node_url_for(node_id):
-    return '/{}nodes/{}/'.format(API_BASE, node_id)
+@pytest.fixture()
```
alexschiller: I don't think this needs to be a fixture if it isn't using a model or complicated to compute

api_tests/collections/test_views.py
```diff
@@ -1,174 +1,184 @@
...
+class TestCollectionCreate:
 
-        self.title = 'Cool Collection'
+    @pytest.fixture()
+    def user_two(self):
+        return AuthUserFactory()
 
-        self.user_two = AuthUserFactory()
+    @pytest.fixture()
+    def title_collection(self):
```
alexschiller: also probably shouldn't be a fixture

api_tests/collections/test_views.py
```diff
@@ -1,174 +1,184 @@
...
+        return 'Cool Collection'
 
-        self.collection = {
+    @pytest.fixture()
+    def data_collection(self, title_collection):
```
alexschiller: just data, probably not a fixture

api_tests/collections/test_views.py
```diff
@@ -222,2261 +236,2635 @@ def test_create_bookmark_collection_with_wrong_title(self):
...
 
+@pytest.mark.django_db
 class TestCollectionUpdate(CollectionCRUDTestCase):
 
-    def test_node_update_invalid_data(self):
-        res = self.app.put_json_api(self.url, "Incorrect data", auth=self.user.auth, expect_errors=True)
-        assert_equal(res.status_code, 400)
-        assert_equal(res.json['errors'][0]['detail'], "Malformed request.")
+    @pytest.fixture()
+    def logs_collection(self, collection):
```
alexschiller: This is a little unusual, again, does this need to be a fixture?

fabmiz: I have not found a `pytest` way to deal with decorators. In this case, the decorator expects a collection as an object not a function object (collection fixture).

api_tests/collections/test_views.py
fabmiz: IMO, and from what I have seen in other tests, it's not desirable to have class attributes; we would need use `self.` in test definitions which is more like nose/unittests way. In addition, for maintenance, IMO, it's more desirable to have all/most fixtures listed as params of the test functions. Since pytest fixtures are cached, performance hit of using simple fixtures vs. strings is minimal.

alexschiller: Alright that's probably fine then. Pass it to Steve when done with changes (if any).

api_tests/collections/test_views.py
sloria: Rhiessen added a context manager, assert_latest_log, that can be used in place of the `assert_logs` decorator. Can you use that instead?

```python
def test_something(...):
     with assert_last_log(NodeLog.EDITED_TITLE, collection):
           # ... some operation
```

api_tests/collections/test_views.py
```diff
@@ -222,2261 +234,2617 @@ def test_create_bookmark_collection_with_wrong_title(self):
...
+        with assert_latest_log(NodeLog.EDITED_TITLE, collection):
+            res = app.put_json_api(url_collection_detail, {
+                'data': {
+                    'id': collection._id,
+                    'type': 'collections',
+                    'attributes': {
+                        'title': new_title_collection,
+                    }
+                }
+            }, auth=user_one.auth)
```
sloria: Minor: Only the above line needs to be in the context manager. Better to make the `with` block as constrained as possible. Please change this in the other usages as well.

PR #7524, opened 2017-07-28
[JUNK][DETRITUS][TIP TOP SECRET: DESTROY BEFORE READING]

TomBaxter:
This is not the PR you are looking for.

## Purpose

Enable ability to create PreprintProvider that pushes it's PreprintService s with SHARE type "thesis" instead of "Preprint"

## Changes

Update osf/models/preprint_provider.py PreprintProvider Add field share_publish_type
Add osf/migrations/0045_preprintprovider_pshare_publish_type.py share_publish_type migration
Update website/preprints/tasks.py 
    1.  on_preprint_updated  Accepts share_type ARG will default to preprint.provider.share_publish_type
    2.  format_preprint  Accepts share_type ARG which on_preprint_updated will pass to it
Updated osf/management/commands/populate_fake_preprint_providers.py to pass share_publish_type
Updated scripts/populate_preprint_providers.py to pass share_publish_typee
Updated tests/test_preprints.py Added test

## Side effects

None known.


tests/test_preprints.py
```diff
@@ -462,13 +467,114 @@ def test_format_preprint(self):
 
         assert nodes == {}
 
+    def test_format_thesis(self):
```
alexschiller: Driveby comment: There is a lot going on in this test. Might consider adding comments explaining what each sub section is testing. E.g., # tests that graph node contains correct thesis information

tests/test_preprints.py
```diff
@@ -462,13 +467,114 @@ def test_format_preprint(self):
...
+        res = format_preprint(self.thesis, self.thesis.provider.push_share_type)
+
+        assert set(gn['@type'] for gn in res) == {'creator', 'contributor', 'throughsubjects', 'subject', 'throughtags', 'tag', 'workidentifier', 'agentidentifier', 'person', 'thesis', 'workrelation', 'creativework'}
```
alexschiller: Minor, personal preference that set tests check lengths as well unless dupes are acceptable.

tests/test_preprints.py
TomBaxter: I'll do my best. The fact is I just copied the test from ```test_format_preprint``` and substituted ```thesis``` where appropriate. It is currently using the same ```Graph``` template. So I'll try to make comments, but I'll have to figure out what it is doing first.

tests/test_preprints.py
TomBaxter: I'll check it out. I believe some/most of these are expected to have duplicates.

tests/test_preprints.py
TomBaxter: After speaking to you and reviewing the test, I am simply adding a description of the test at the top of the test. While there is an overwhelming amount of assertions and list comprehension going on, the big picture is pretty simple.

```format_preprint``` takes a ```PreprintService``` and creates a ```GraphNode``` appropriate for uploading to SHARE. This test, compares the ```PreprintService``` Object and the resulting ```GraphNode``` for accuracy.

Comment (alexschiller) [inline tests/test_preprints.py]: :+1: Not worth the time since it was a copy/paste.

Comment (brianjgeiger) [inline tests/test_preprints.py]: This is a lot just for the one change. If we're super-concerned about the change, it might be worth modifying the original test to run through all of the potential categories. Alternately, for this test we might just verify that the 'thesis' comes through and don't worry so much about all the rest of the things. Having two almost identical complicated tests is generally not ideal, mostly because if we change something (like, say, getting all the metadata from the preprint object instead of half from there and half from node) we don't want to have to change both of those tests.

File: admin_tests/preprint_providers/test_views.py [hunk 3]
```diff
@@ -150,13 +150,14 @@ def test_preprint_provider_form(self):
         formatted_rule = [[[self.parent_1._id], True]]
 
         new_data = {
+            '_id': 'newname',
             'name': 'New Name',
+            'push_share_type': 'Preprint',
```
brianjgeiger: Naming things is hard, but I'm leaning towards `share_publish_type` instead of `push_share_type`.

tests/test_preprints.py
TomBaxter: Complete. Copied test has now been severely pruned.

admin_tests/preprint_providers/test_views.py
TomBaxter: Complete.

brianjgeiger:
@TomBaxter Please update with develop and let me know when tests pass.

PR #7526, opened 2017-07-31
[hotfix] - NodeLog query performance improvements [OSF-8288]

erinspace:
## Purpose

The query for NodeLogs is a little slow - let's make that a little faster!

## Changes

- Update the NodeLog view to use django queries
- Use the `IncludeManager` for `NodeLog` objects and on API queries

## Side effects
None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-8288

### DIFF
--- a/osf/models/node.py
+++ b/osf/models/node.py
@@ -5,6 +5,7 @@ import re
 import urlparse
 import warnings
 
+from django.db.models import Q
 from dirtyfields import DirtyFieldsMixin
 from django.apps import apps
 from django.contrib.contenttypes.fields import GenericRelation
@@ -42,7 +43,7 @@ from osf.models.spam import SpamMixin
 from osf.models.tag import Tag
 from osf.models.user import OSFUser
 from osf.models.validators import validate_doi, validate_title
-from osf.modm_compat import Q
+from osf.modm_compat import Q as MODMQ
 from osf.utils.auth import Auth, get_user
 from osf.utils.datetime_aware_jsonfield import DateTimeAwareJSONField
 from osf.utils.fields import NonNaiveDateTimeField
@@ -687,7 +688,7 @@ class AbstractNode(DirtyFieldsMixin, TypedModel, AddonModelMixin, IdentifierMixi
 
     @classmethod
     def find_by_institutions(cls, inst, query=None):
-        base_query = Q('affiliated_institutions', 'eq', inst)
+        base_query = MODMQ('affiliated_institutions', 'eq', inst)
         if query:
             final_query = base_query & query
         else:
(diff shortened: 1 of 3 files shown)


osf/models/node.py
```diff
@@ -778,15 +779,17 @@ def can_edit(self, auth=None, user=None):
 
     def get_aggregate_logs_query(self, auth):
         return (
-            (   # TODO: remove `list` call during phase 2
-                Q('node_id', 'in', list(Node.objects.get_children(self).can_view(user=auth.user, private_link=auth.private_link).values_list('id', flat=True))) |
-                Q('node_id', 'eq', self.id)
-            ) & Q('should_hide', 'eq', False)
+            (
+                Q(node_id__in=list(Node.objects.get_children(self).can_view(user=auth.user, private_link=auth.private_link).values_list('id', flat=True))) |
```
cwisecarver: `values_list` returns a list so you shouldn't need to call list on it, right? Or am I misreading the parenthesis?

erinspace: sadly `values_list` returns a `ValuesListQuerySet` -- tried many times to avoid this list call but this turned out to be the best in the end given the weird subquery

cwisecarver: It appears we pass a flat `ValuesListQuerySet` directly to an `__in` filter in several places and it works.

```
osf/management/commands/force_archive.py:259:        Q(node__id__in=Node.objects.get_children(reg.registered_from).values_list('id', flat=True)) | Q(node__id=reg.registered_from.id))\
osf/management/commands/populate_custom_taxonomies.py:45:    missing_subjects = Subject.objects.filter(id__in=set([hier[-1].id for ps in PreprintService.objects.filter(provider=custom_provider) for hier in ps.subject_hierarchy])).exclude(id__in=included_subjects.values_list('id', flat=True))
osf/models/node.py:74:        return self.filter(id__in=self.exclude(type='osf.collection').values_list('root_id', flat=True))
osf/models/subject.py:18:        return Subject.objects.filter(Q(id__in=self.values_list('id', flat=True)) | Q(parent__in=self) | Q(parent__parent__in=self))
website/conferences/views.py:183:    nodes = Node.objects.filter(tags__id__in=Tag.objects.filter(name__iexact=meeting, system=False).values_list('id', flat=True), is_public=True, is_deleted=False)

Comment (erinspace) [inline osf/models/node.py]: hmmm.... removing the list call here gives the error:

```
api_1               | ProgrammingError: invalid reference to FROM-clause entry for table "osf_abstractnode"
api_1               | LINE 2:                 "osf_abstractnode".id in (
api_1               |                         ^
api_1               | HINT:  Perhaps you meant to reference the table alias "u0".
```

I worked on this with @sloria for quite awhile and should have done a better job documenting exactly the things I tried and why they didn't work

cwisecarver: It's `include` that's breaking it. @chrisseto should fix that.

PR #7527, opened 2017-07-31
[feature] Use Django's ORM in APIv1 Views and Tests [OSF-8309]

caseyrollins:
#### Purpose
- Removes all occurrences of MODM Q objects in APIv1 codebase views, tests, utils, etc.
  - MODM is deprecated and continuing to use it prevents us from upgrading to Python 3 as well as the next major Django version. 
  - Removing MODM Qs improves code consistency and readability. 

#### Changes
- All MODM queries converted to Django queries.
- This PR also removes some "redirect to twitter" code I stumbled upon.

#### Side Effects
- Probably.

#### QA 
- Entire sweep of OSF functionality.

#### Ticket
- [OSF-8309](https://openscience.atlassian.net/browse/OSF-8309)


website/mails/listeners.py
```diff
@@ -32,9 +31,8 @@ def queue_first_public_project_email(user, node, meeting_creation):
     """
     from osf.models.queued_mail import queue_mail, QueuedMail, NEW_PUBLIC_PROJECT_TYPE, NEW_PUBLIC_PROJECT
     if not meeting_creation:
-        sent_mail = QueuedMail.find(Q('user', 'eq', user) & Q('sent_at', 'ne', None) &
-                                          Q('email_type', 'eq', NEW_PUBLIC_PROJECT_TYPE))
-        if not sent_mail.count():
+        sent_mail = QueuedMail.objects.filter(user=user, sent_at__isnull=False, email_type=NEW_PUBLIC_PROJECT_TYPE)
```
erinspace: in the `QueuedMail` class the `send_at` field has `null=False`, so you might not need that middle query filter param

website/profile/views.py
```diff
@@ -403,9 +402,8 @@ def oauth_application_detail(auth, **kwargs):
 
     # The client ID must be an active and existing record, and the logged-in user must have permission to view it.
     try:
-        #
-        record = ApiOAuth2Application.find_one(Q('client_id', 'eq', client_id))
-    except NoResultsFound:
+        record = ApiOAuth2Application.objects.get(client_id=client_id)
```
erinspace: could use django's `get_object_or_404` here I think!

website/profile/views.py
```diff
@@ -441,8 +439,8 @@ def personal_access_token_detail(auth, **kwargs):
 
     # The ID must be an active and existing record, and the logged-in user must have permission to view it.
     try:
-        record = ApiOAuth2PersonalToken.find_one(Q('_id', 'eq', _id))
-    except NoResultsFound:
+        record = ApiOAuth2PersonalToken.objects.get(_id=_id)
```
erinspace: Same for here?

website/project/utils.py
```diff
@@ -102,7 +101,7 @@ def activity():
                 break
 
     # New and Noteworthy projects are updated manually
-    new_and_noteworthy_projects = list(Node.find_one(Q('_id', 'eq', settings.NEW_AND_NOTEWORTHY_LINKS_NODE)).nodes_pointer)
+    new_and_noteworthy_projects = list(Node.objects.get(_id=settings.NEW_AND_NOTEWORTHY_LINKS_NODE).nodes_pointer)
```
erinspace: Might need `Node.objects.get(guids___id=...)` here

website/project/views/node.py
```diff
@@ -508,27 +507,26 @@ def remove_private_link(*args, **kwargs):
...
-        link.save()
-
-        for node in link.nodes.all():
-            log_dict = {
-                'project': node.parent_id,
-                'node': node._id,
-                'user': kwargs.get('auth').user._id,
-                'anonymous_link': link.anonymous,
-            }
+        link = PrivateLink.objects.get(_id=link_id)
```
erinspace: `get_object_or_404`?

website/project/views/node.py
```diff
@@ -646,7 +644,7 @@ def _view_project(node, auth, primary=False,
             'link': view_only_link,
             'anonymous': anonymous,
             'comment_level': node.comment_level,
-            'has_comments': Comment.objects.filter(node=node).exists(),
+            'has_comments': bool(node.comment_set.count()),
```
erinspace: Could do `node.comment_set.exists()`

website/project/views/register.py
```diff
@@ -253,11 +247,8 @@ def get_referent_by_identifier(category, value):
     if found.
     """
     try:
-        identifier = Identifier.find_one(
-            Q('category', 'eq', category) &
-            Q('value', 'eq', value)
-        )
-    except NoResultsFound:
+        identifier = Identifier.objects.get(category=category, value=value)
```
erinspace: `get_object_or_404`?

website/citations/views.py
```diff
@@ -13,18 +12,16 @@
...
-    query = None
-
-    term = request.args.get('q')
-    if term:
-        query = (
-            Q('_id', 'icontains', term) |
-            Q('title', 'icontains', term) |
-            Q('short_title', 'icontains', term)
+    query = request.args.get('q')
+    citation_styles = CitationStyle.objects.all()
```
caseyrollins: The query won't be executed until `citation_styles` is evaluated, so there's actually no duplication here. 

https://docs.djangoproject.com/en/1.11/ref/models/querysets/#when-querysets-are-evaluated

website/profile/views.py
caseyrollins: I actually tried using it here (and in the other places you've mentioned), but `get_object_or_404` returns an [`Http404`](https://docs.djangoproject.com/en/1.11/topics/http/views/#django.http.Http404) whereas our v1 views are expecting our custom [`HTTPError`](https://github.com/CenterForOpenScience/osf.io/blob/develop/framework/exceptions/__init__.py#L11).

The tl;dr of it is that this works just fine in APIv2, but returning `Http404` broke our beautiful "Page not found" templating in [`error.mako`](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/templates/error.mako). While the shortcut would be nice, it seemed out of scope to change the error handling right now.

website/project/utils.py
caseyrollins: Definitely, nice catch.

osf_tests/test_node.py
```diff
@@ -2,6 +2,7 @@
 
 from django.utils import timezone
 from django.core.exceptions import ValidationError as DjangoValidationError
+from django.db.models import Q as DQ
 from modularodm import Q
```
caseyrollins: This overlaps with your models PR and there's quite a few MODM Qs in this file so I didn't want too many lines (by either django-ifying all of those Qs or importing them as MQ). I figured whichever one of us deals with the conflict last (probably me) can switch the DQ to Q?

erinspace: Makes sense to me!

website/citations/views.py
erinspace: Good call!

website/profile/views.py
erinspace: Aaaaaaaaaaahhhhhhhhhhhh makes sense.  Good to know!

osf_tests/test_node.py
```diff
@@ -1709,12 +1709,17 @@ def test_find_for_user():
     noncontrib = UserFactory()
     Contributor.objects.create(node=node1, user=contrib)
     Contributor.objects.create(node=node2, user=contrib)
-    assert node1 in Node.find_for_user(contrib)
-    assert node2 in Node.find_for_user(contrib)
-    assert node1 not in Node.find_for_user(noncontrib)
 
-    assert node1 in Node.find_for_user(contrib, Q(is_public=False))
-    assert node2 not in Node.find_for_user(contrib, Q(is_public=False))
+    contrib_nodes = Node.find_for_user(contrib)
```
sloria: Eventually we should get rid of `find_for_user`. It doesn't provide anything that isn't already possible with `.nodes`.

website/conferences/views.py
```diff
@@ -93,7 +91,15 @@ def add_poster_by_email(conference, message):
         else:
             set_password_url = None
 
-        node, node_created = utils.get_or_create_node(message.subject, user)
+        node, node_created = Node.objects.get_or_create(
+            title__iexact=message.subject,
+            is_deleted=False,
+            _contributors__guids___id=user._id,
```
sloria: Can this just be `contributors=user` or `_contributors=user`?

website/project/decorators.py
```diff
@@ -223,13 +220,12 @@ def wrapped(*args, **kwargs):
...
 
             kwargs['auth'].private_key = key
-            link_anon = None
             if not include_view_only_anon:
                 from osf.models import PrivateLink
                 try:
-                    link_anon = PrivateLink.find_one(Q('key', 'eq', key)).anonymous
-                except ModularOdmException:
-                    pass
+                    link_anon = PrivateLink.objects.get(key=key).anonymous
```
sloria: Minor: Can be optimized slightly with `PrivateLink.objects.filter(key=key).values_list('anonymous', flat=True).get()`.

website/notifications/tasks.py
```diff
@@ -106,4 +105,4 @@ def remove_notifications(email_notification_ids=None):
     :return:
     """
     for email_id in email_notification_ids:
-        NotificationDigest.remove(Q('_id', 'eq', email_id))
+        NotificationDigest.objects.get(_id=email_id).delete()
```
sloria: Minor: This loops can be replaced with `NotificationDigest.objects.filter(__id__in=email_notification_ids).delete()`.

website/project/__init__.py
```diff
@@ -50,13 +49,13 @@ def new_bookmark_collection(user):
...
-    existing_bookmark_collection = Collection.find(
-        Q('is_bookmark_collection', 'eq', True) &
-        Q('creator', 'eq', user) &
-        Q('is_deleted', 'eq', False)
-    )
+    existing_bookmark_collections = Collection.objects.filter(
+        is_bookmark_collection=True,
+        creator=user,
+        is_deleted=False
+    ).count()
```
sloria: Using `.exists()` is more performant than `.count()`.

website/project/model.py
```diff
@@ -93,8 +90,10 @@ def validate_user(value):
     OSFUser = apps.get_model('osf.OSFUser')
     if value != {}:
         user_id = value.iterkeys().next()
-        if OSFUser.find(Q('_id', 'eq', user_id)).count() != 1:
-            raise ValidationValueError('User does not exist.')
+        try:
+            OSFUser.objects.get(_id=user_id)
```
sloria: Does this work? Wouldn't this need to be `guids___id=user_id`? Also, you can use exists() to make the query slightly more performant:

```python
if not OSFUser.objects.filter(guids___id=user_id).exists():
    raise ValidatoinError(...)
```

website/project/views/node.py
```diff
@@ -1054,18 +1052,16 @@ def search_node(auth, **kwargs):
...
     # Exclude current node from query if provided
     nin = [node.id] + list(node._nodes.values_list('pk', flat=True)) if node else []
 
-    nodes = AbstractNode.find(odm_query).exclude(id__in=nin).exclude(type='osf.collection')
+    nodes = AbstractNode.objects.filter(
+        Q(contributors=auth.user) |
+        Q(is_public=bool(include_public)),
+        title__icontains=query,
+        is_deleted=False
+    ).exclude(id__in=nin, type='osf.collection')
```
sloria: Is this logically equivalent to the old code?

```python
exclude(id__in=nin).exclude(type='osf.collection')
```

should exclude nodes whose `id` is in `nin` and exclude collections.

I think that

```python
.exclude(id__in=nin, type='osf.collection')
```

will only exclude collections whose `id` is in `nin` .

caseyrollins: Definitely not equivalent. Nice catch.

```
WHERE ("osf_abstractnode"."type" = osf.node AND NOT ("osf_abstractnode"."is_public" = True AND "osf_abstractnode"."type" = osf.collection))
```
vs
```
WHERE ("osf_abstractnode"."type" = osf.node AND NOT ("osf_abstractnode"."is_public" = True) AND NOT ("osf_abstractnode"."type" = osf.collection))
```

website/project/model.py
caseyrollins: Lol, nope.

```
FieldError: Cannot resolve keyword '_id' into field.
```

website/conferences/views.py
caseyrollins: I actually think this is necessary, although I don't have a great explanation for why other than: 

`contributors=user` errors with:
```
FieldError: Invalid field name(s) for model Node: 'contributors'.
```
and `_contributors=user` errors with:
```
ValueError: "<AbstractNode: (title='testing', category='') with guid None>" 
needs to have a value for field "id" before this many-to-many relationship 
can be used.
```

osf_tests/test_node.py
caseyrollins: I'm pretty sure I can get rid of this now. I think I didn't originally to avoid overlap with other PRs.

website/project/model.py
```diff
@@ -90,9 +90,7 @@ def validate_user(value):
     OSFUser = apps.get_model('osf.OSFUser')
     if value != {}:
         user_id = value.iterkeys().next()
-        try:
-            OSFUser.objects.get(_id=user_id)
-        except OSFUser.DoesNotExist:
+        if not OSFUser.objects.get(guids___id=user_id).exists():
```
sloria: This should be `filter` instead of `get`. 

It would be great to have a regression test for this function.

caseyrollins: Yeahh, definitely gonna add a test.

caseyrollins: Wait, this function isn't used anywhere. Removing.

caseyrollins: `validate_doi` and `validate_category` are also unused. Removing those as well.

sloria:
Good to merge when Travis is passing.

sloria:
Failing test has been fixed on develop.

PR #7528, opened 2017-07-31
[RFR]Update line that was pointing to the erstwhile ./website/addons now .…

TomBaxter:
No ticket. Found issue while debugging another issue.

## Purpose

Fix Karma tests

## Changes

One line in website/static/js/tests/tests.webpack.js was pointing to ./website/addons which is now no more. However, many/most developers may still have ./website/addons/*/*/*.local.py s in their environments. Which has masked this issue.

## Side effects

None

## Ticket
No ticket. Found issue while debugging another issue.


sloria:
This looks like it fixes https://openscience.atlassian.net/browse/OSF-8278 .

sloria:
However, keeping that ticket open for now until we run JS tests on Travis.

PR #7529, opened 2017-07-31
Use default django queries in addons and tests [OSF-8312]

erinspace:
[#OSF-8312]

## Purpose

Remove usages of modularodm Q and modularodm exceptions from addons!

## Changes

- remove modularodm Q objects
- Remove modularodm exceptions

## Side effects

none anticipated

## Ticket

https://openscience.atlassian.net/browse/OSF-8312

### DIFF
--- a/addons/base/models.py
+++ b/addons/base/models.py
@@ -13,7 +13,6 @@ from osf.models.base import BaseModel, ObjectIDMixin
 from osf.models.external import ExternalAccount
 from osf.models.node import AbstractNode
 from osf.models.user import OSFUser
-from osf.modm_compat import Q
 from osf.utils.datetime_aware_jsonfield import DateTimeAwareJSONField
 from website import settings
 from addons.base import logger, serializer
@@ -319,10 +318,7 @@ class BaseOAuthUserSettings(BaseUserSettings):
         except KeyError:
             pass
         else:
-            connected = Model.find(Q('user_settings', 'eq', user_settings))
-            for node_settings in connected:
-                node_settings.user_settings = self
-                node_settings.save()
+            Model.objects.filter(user_settings=user_settings).update(user_settings=self)
 
         self.save()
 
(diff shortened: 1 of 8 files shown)


addons/osfstorage/listeners.py
```diff
@@ -1,16 +1,14 @@
...
-
 from website.project.signals import contributor_removed
 
 @contributor_removed.connect
 def checkin_files_by_user(node, user):
     ''' Listens to a contributor being removed to check in all of their files
     '''
     from addons.osfstorage.models import OsfStorageFileNode
-    files = OsfStorageFileNode.find(Q('node', 'eq', node) & Q('checkout', 'eq', user))
+    files = OsfStorageFileNode.objects.filter(node=node, checkout=user)
```
caseyrollins: You can use [`update()`](https://docs.djangoproject.com/en/1.11/ref/models/querysets/#update) here and eliminate the loop below. 

Just so you know, `update()` does bypass custom saves, but I checked with Wisecarver that it's perfectly fine to use here.

addons/base/models.py
```diff
@@ -319,7 +318,7 @@ def merge(self, user_settings):
         except KeyError:
             pass
         else:
-            connected = Model.find(Q('user_settings', 'eq', user_settings))
+            connected = Model.objects.filter(user_settings=user_settings)
```
caseyrollins: Similar to other note, I think you can use `update()` here and remove the loop below.

addons/osfstorage/models.py
```diff
@@ -104,10 +104,7 @@ def get_file_guids(cls, materialized_path, provider, node=None):
             for item in children:
                 guids.extend(cls.get_file_guids(item.path, provider, node=node))
         else:
-            try:
-                guid = Guid.find(Q('referent', 'eq', file_obj))[0]
-            except IndexError:
-                guid = None
+            guid = Guid.objects.filter(object_id=file_obj.id, content_type=ContentType.objects.get_for_model(file_obj)).first()
```
caseyrollins: Does `file_obj.referents.first()` work here?

sloria:
Merged into release/next-djangole.

PR #7532, opened 2017-08-02
Feature/shorten contributor logs [OSF-8346]

laurenrevere:
## Purpose

If many contributors are added at one, they make the log entry very long.

## Changes
This changes the contributor added log text is created so that it only shows 3-4 contributors. If there are 4 it shows them all normally if there are 5 or more it shows 3 of the contributor names and then text that states "and 2 others."

## Side effects
I had to change the exports of `logTextParser` from a single export to a dictionary. This means that `LogText` will need to be grabbed by name any time it is used instead of just by importing `logTextParser`. The in most cases this change will look like this `LogText.LogText` instead of just `LogText` (depending on how the import is named).

## Notes
Could easily make it so it always just says "and <#> other(s) added" but this seemed cleaner.

## Ticket
https://openscience.atlassian.net/browse/OSF-8346

19 contributors added
19 added

17 contributors added
17 added

16 contributors added
16 added

6 contributors added (normal log stuff still works)
6 added

### DIFF
--- a/website/static/js/logTextParser.js
+++ b/website/static/js/logTextParser.js
@@ -715,4 +715,7 @@ var LogPieces = {
     }
 };
 
-module.exports = LogText;
+module.exports = {
+    LogText:LogText,
+    getContributorList: getContributorList
+};
--- a/website/static/js/logTextParser.js
+++ b/website/static/js/logTextParser.js
@@ -220,25 +220,37 @@ var LogPieces = {
         view: function (ctrl, logObject) {
             var contributors = logObject.attributes.params.contributors;
             if(paramIsReturned(contributors, logObject)) {
-                return m('span', contributors.map(function(item, index, arr){
-                    var comma = ' ';
-                    if(index !== arr.length - 1){
-                        comma = ', ';
-                    }
-                    if(index === arr.length-2){
-                        comma = ' and ';
-                    }
+                return m('span', (function (){
+                    var contribList = [];
+                    var numContribShown = 15;
 
-                    if (item.active) {
-                        return [ m('a', {href: '/' + item.id + '/'}, item.full_name), comma];
-                    }
-                    else {
-                        if (item.unregistered_name) {
-                            return [item.unregistered_name, comma];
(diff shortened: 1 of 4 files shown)


website/static/js/logTextParser.js
```diff
@@ -220,25 +220,37 @@ var LogPieces = {
...
-                    var comma = ' ';
-                    if(index !== arr.length - 1){
-                        comma = ', ';
-                    }
-                    if(index === arr.length-2){
-                        comma = ' and ';
-                    }
+                return m('span', (function (){
+                    var contribList = [];
+                    var numContribShown = 15;
```
brianjgeiger: Don't put magic numbers inside code; we'll want to make that something easy to change, and easy to find that it needs changing. In this case, probably at the top of the file under the imports. Make a note where it's used where it's defined.

brianjgeiger: Also, though this was vague and unspecified in the ticket, let's make the `numContribShown` 3 rather than 15.

website/static/js/logTextParser.js
```diff
@@ -93,6 +97,44 @@ var returnTextParams = function (param, text, logObject, view_url) {
...
+       var contribList = [];
+       var justOneMore = numContributorsShown === contributors.length -1;
+       for(var i = 0; i < contributors.length; i++){
+           var item = contributors[i];
+           var comma = ' ';
+           if(i !== contributors.length -1 && ((i !== maxShown -1) || justOneMore)){
+               comma = ', ';
+           }
+           if(i === contributors.length -2 || ((i === maxShown -1) && !justOneMore)){
+               comma = ' and ';
```
brianjgeiger: Minor: `', and '`

laurenrevere: As far as I know, the OSF does not use the oxford comma.

brianjgeiger: Mmmmmm. That doesn't make it right. Okay, we'll leave it for now.

PR #7535, opened 2017-08-02
[#OSF-7955]Templating a registration breaks and creates a malformed registration

chennan47:
## Purpose
Fix the problem that templating a registration breaks and creates a malformed registration. Now you can template a registration and it will return a node instead of a malformed registration.

## Changes
1. create a registration
2. create a template of that registration
screen shot 2017-08-02 at 2 52 34 pm

Before change:
It will end in unable to resolve
screen shot 2017-08-02 at 2 53 36 pm
After change:
It will return a correct templated node
screen shot 2017-08-02 at 2 54 01 pm

## Ticket
https://openscience.atlassian.net/browse/OSF-7955


osf_tests/test_node.py
```diff
@@ -3525,6 +3525,11 @@ def test_use_as_template_preserves_license(self, project, auth):
         assert new.license.node_license._id == license.node_license._id
         self._verify_log(new)
 
+    def test_can_template_a_registration(self, auth):
+        registration = RegistrationFactory(creator=user)
```
alexschiller: If you're trying to use the user fixture you need to pass it in:
`def test_can_template_a_registration(self, user, auth):`

Test is failing. The other test that is failing looks unrelated.

osf/models/node.py
```diff
@@ -1991,8 +1992,12 @@ def use_as_template(self, auth, changes=None, top_level=True):
             # TODO: explicitly define attributes which may be changed.
         except (AttributeError, KeyError):
             attributes = dict()
-
+        if self.is_deleted:
```
alexschiller: Can you comment on this change? Not listed in side effects nor does it look necessary. Does it cause problems?

alexschiller: @brianjgeiger Can you add the migrations tag to this (or some other tag) to indicate that Steve should check for any new broken templated registrations and fix them.

chennan47: This line exists in fork_node. I think it is necessary to prevent any user from template a withdrawn pending registration.

alexschiller: That's fine then. `test_cannot_fork_deleted_node` is in `test_node`, can you add a comparable one for nodes/registrations

brianjgeiger:
I think we still need "cannot template deleted registration".

PR #7536, opened 2017-08-02
[OSF-8357] Remove dict as default args

Johnetordoff:
## Purpose

When it comes to Python using dicts as default args there's some weird behavior. The dict used as a default arg is used every time that method is called, potentially growing and retaining old information, here's an example.
```

In [1]: def dict_leaks(label, attrs={}):
    attrs[label] = "alive!"
    print('{}'.format(attrs))

In [2]:  dict_leaks("dog")
{'dog': 'alive!'}

In [3]: dict_leaks("cat")
{'dog': 'alive!', 'cat': 'alive!'}
```

So pretty sneaky and bad.

## Changes

I removed the instances of default dict arguments and changed the code so dicts are freshly instantiated. 

## Side effects

None that I know of.

## Test 

Would be nice if we could test for this but it would have to apply to all python code on the site to be meaningful.

## Ticket

https://openscience.atlassian.net/browse/OSF-8357

### DIFF
--- a/api_tests/preprints/views/test_preprint_list.py
+++ b/api_tests/preprints/views/test_preprint_list.py
@@ -20,7 +20,11 @@ from tests.base import ApiTestCase, capture_signals
 from website.project import signals as project_signals
 from website.util import permissions
 
-def build_preprint_create_payload(node_id=None, provider_id=None, file_id=None, attrs={}):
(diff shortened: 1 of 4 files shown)


brianjgeiger:
I wonder if we could find a linter that checked for that, or configure ours to do so.

Johnetordoff:
yep

PR #7540, opened 2017-08-03
[feature] Check saved_fields before updating institution in search [No ticket]

caseyrollins:
#### Purpose
- From [discussion with @sloria](https://github.com/CenterForOpenScience/osf.io/pull/7477#pullrequestreview-53857522): only update an institution in search if it has changed (i.e. check `saved_fields` first).


sloria:
Actually, I think I led you astray...with this change, new institutions won't get added to ES. Going to revert this one for now.

PR #7544, opened 2017-08-03
[OSF-8339] Fix API 502s when refreshing github metadata

adlius:
## Purpose

This PR complements [this PR](https://github.com/CenterForOpenScience/osf.io/pull/6835) as a fix for #6062 .

## Changes
Suppose a github file name `README.md` exists for a repo connected to an OSF project, the `update()` method for this GithubFile would be called to update the file history when the following api v2 endpoints are called:
1. /nodes/<project_guid>/files/github
2. /nodes/<project_guid>/files/github/README.md

For the first endpoint, waterbutler would update the file history by passing a new entry (a dict whose value for key `modified` is `None`) to the `update()` method. For the second endpoint, however, waterbutler would do so with a dict whose value for `modified` is the actual time of modification.
Therefore, when `insort()` is called to insert the new data from waterbutler as a history entry, comparisons between existing entries and the data would be made to determine where the order of the new entry in the list. If the first api v2 endpoint is called, because the value for `modified` is `None`, a TypeError would be thrown. The [previous PR](https://github.com/CenterForOpenScience/osf.io/pull/6835) fix this problem by only sort in a new entry with `modified` is not None, otherwise, it simply append the new entry to the end of file history. It does not fix the situation when an entry with `modified==None` is already in the file history.

To fix this problem thoroughly, a change is implemented to check for value of `modified` for all file history entires. If one of the entry has an undefined `modified` value, it would also append the new entry to the end of the file history.

Notes for reproducing this bug:
1. Commit a file on a Github Repo linked to a project.
2. Visit the api at `/nodes/<project-guid>/files/github`
3. Commit to the same file again.
4. Visit the api at `/nodes/<project_guid>/files/github/<filename>`

## Side effects

Similar side effects to https://github.com/CenterForOpenScience/osf.io/pull/6835.

## Ticket

https://openscience.atlassian.net/browse/OSF-8339

### DIFF
--- a/osf/models/files.py
+++ b/osf/models/files.py
@@ -445,7 +445,7 @@ class File(models.Model):
                 break
         else:
             # Insert into history if there is no matching etag
-            if data.get('modified'):
+            if data.get('modified') and None not in [ item['modified'] for item in self.history ]:
                 utils.insort(self.history, data, lambda x: x['modified'])
             # If modified not included in the metadata, insert at the end of the history
             else:
--- a/osf/models/files.py
+++ b/osf/models/files.py
@@ -14,6 +14,7 @@ from typedmodels.models import TypedModel, TypedModelManager
 from include import IncludeManager
 
 from framework.analytics import get_basic_counters
+from framework import sentry
 from osf.models.base import BaseModel, OptionalGuidMixin, ObjectIDMixin
 from osf.models.comment import CommentableMixin
 from osf.models.mixins import Taggable
@@ -441,15 +442,12 @@ class File(models.Model):
             version.save()
             self.versions.add(version)
         for entry in self.history:
+            if data['modified'] is not None and data['modified'] < entry['modified']:
+                sentry.log_message('update() receives metatdata older than the newest entry in file history.')
             if ('etag' in entry and 'etag' in data) and (entry['etag'] == data['etag']):
                 break
         else:
-            # Insert into history if there is no matching etag
-            if data.get('modified') and None not in [item['modified'] for item in self.history]:
-                utils.insort(self.history, data, lambda x: x['modified'])
-            # If modified not included in the metadata, insert at the end of the history
-            else:
-                self.history.append(data)
+            self.history.append(data)
 
         # Finally update last touched
         self.last_touched = timezone.now()
--- a/osf/models/files.py
+++ b/osf/models/files.py
@@ -445,7 +445,7 @@ class File(models.Model):
                 break
         else:
             # Insert into history if there is no matching etag
-            if data.get('modified') and None not in [ item['modified'] for item in self.history ]:
+            if data.get('modified') and None not in [item['modified'] for item in self.history]:
                 utils.insort(self.history, data, lambda x: x['modified'])
             # If modified not included in the metadata, insert at the end of the history
             else:
--- a/tests/test_addons.py
+++ b/tests/test_addons.py
@@ -888,6 +888,19 @@ class TestAddonFileViews(OsfTestCase):
         assert_equal(len(file_node.history), 1)
         assert_equal(file_node.history[0], file_data)
 
+    @mock.patch('website.archiver.tasks.archive')
+    def test_missing_modified_date_in_file_history(self, mock_archive):
+        file_node = self.get_test_file()
+        file_node.history.append({'modified': None})
+        file_data = {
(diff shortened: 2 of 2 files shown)


brianjgeiger:
I'm not sure if the failure is probabilistic or has been fixed recently. I know there were test problems last week. Can you make sure you're up-to-date with Develop?

osf/models/files.py
```diff
@@ -445,7 +445,7 @@ def update(self, revision, data, user=None):
                 break
         else:
             # Insert into history if there is no matching etag
-            if data.get('modified'):
+            if data.get('modified') and None not in [item['modified'] for item in self.history]:
```
brianjgeiger: Okay, so if any item anywhere in the history has a modified date of `None` then the new item goes to the bottom of the list? This is definitely better than erroring, but I'm not sure if it's right. Might be worth having a quick discussion with @felliott to see if there are any other potential options, since it's mostly about what can and can't actually be done.

brianjgeiger: After chatting with @felliott, we're going to modify the behavior so that if you have etags, great, if not, then just add it to the end of the list. However, Fitz has a request for a check that will bubble up a sentry error so we can see if we ever hit a particular state. Once he's added the description of the check, we can move this to addl dev.

brianjgeiger: Note from Fitz:

> Okay, this is a weird one. It's possible that we shouldn't be doing the insort at all. It was added 
> early on as part of supporting provenance tracking, but there's no specific message saying 
> why we do it this way. I can't think of a situation that touch() would call update() with anything 
> other than the latest revision data. It would make sense to continue to check the etags, and if 
> not found append the revision to the history.
> 
> One major concern: since this is the update() method for the File class, it might be used in 
> other places in the codebase where the revision isn't the most recent revision (migrations?). 
> Maybe add a debug statement and run the test suite to see who is using this?

adlius: I understand the first part. But if the update() method is for the File class, what kind of "debug statement" should I add to the codebase? Are to log to sentry every time when the update() is called???

brianjgeiger: You can wrap the error in a `dev_mode` if statement to ensure it doesn't go to production. But yeah, probably logging to sentry with a fairly distinctive title and putting in testing notes to check sentry would work well.

adlius: So... just to see if I understand this correctly: I should do something along the lines of
```
If no matching etags;
    append revision to history

if in dev_mode:
  if update() is called by a non GithubFile instance:
    log to sentry with distinctive title
```

brianjgeiger: @felliott Please verify ^^

felliott: @adlius, sorry I'm conflating two different issues.

1. The bit about adding a debug statement would be for local dev only and not a part of the PR.  I'm talking about adding a `logging.error` call to the `update()` method, running the full test suite, then seeing how many times it gets called and from where.  That'll help us figure out where `update()` is being called and what its expected behavior is there.  If all the calls come about as a result of `touch()`, then this fix is probably safe.  If there are other places that call `update()`, we need to see if they require the sorting behavior or if append-only works for them.  If they require the sorting behavior, the fix for this will need to be more complicated.  This issue is answering the question *"will switching to append-only break the non-touch parts of the OSF that call `File.update()`*.

2. The second issue is the sentry logging message.  I'm pretty sure that `update()` called from `touch()` will never receive metadata that is older than the the newest entry already in history.  However, it would be nice to check this to make sure I'm not forgetting some use case.  Therefore, I'd like a sentry message that warns if the new revision information is older than the most recent revision.  This issue is answering the question *"is it correct to assume that metadata from the touch() method will always be the same age or newer than whatever we've stored already"*.

PLMK if I'm talking gibberish.

adlius: @felliott Makes perfect sense! I'll go ahead and do that.

adlius:
Run the full test suite and it seems that two more methods within `comment.py` are calling `update`:
1. def create_new_file(obj, source, destination, destination_node)
2. def find_and_create_file_from_metadata(children, source, destination, destination_node, obj)

It seems that both methods have nothing to do with the sorting of file histories. Therefore it should be safe to assume that appending new entries would not break any "non-touch" functionalities.

I'll update the PR and change to append-only, and will also add a line to log to sentry and see if any new entries from WB is older than the existing ones.

PR #7545, opened 2017-08-03
[Feature] Hide `Force check in` for BitBucket [OSF-8192]

mfraezz:
## Purpose
Fix UI issue

## Changes
* Add special case

## Side effects
None expected

## Ticket
[[#OSF-8192]](https://openscience.atlassian.net/browse/OSF-8192)

### DIFF
--- a/website/static/js/filepage/index.js
+++ b/website/static/js/filepage/index.js
@@ -504,6 +504,7 @@ var FileViewPage = {
             // Special case whether or not to show the delete button for published Dataverse files
             // Special case to not show delete if file is preprint primary file
             // Special case to not show delete for public figshare files
+            // Special case to not show force check-in for read-only providers
             (
                 ctrl.canEdit() &&
                 (ctrl.node.preprintFileId !== ctrl.file.id) &&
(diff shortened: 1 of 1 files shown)


website/static/js/filepage/index.js
```diff
@@ -513,7 +514,7 @@ var FileViewPage = {
             ) ? m('.btn-group.m-l-xs.m-t-xs', [
                         ctrl.isLatestVersion ? m('button.btn.btn-sm.btn-danger.file-delete', {onclick: $(document).trigger.bind($(document), 'fileviewpage:delete') }, 'Delete') : null
             ]) : '',
-            ctrl.context.currentUser.canEdit && (!ctrl.canEdit()) && (ctrl.context.currentUser.isAdmin) ? m('.btn-group.m-l-xs.m-t-xs', [
+            ctrl.context.currentUser.canEdit && (ctrl.file.provider != 'bitbucket') && (!ctrl.canEdit()) && (ctrl.context.currentUser.isAdmin) ? m('.btn-group.m-l-xs.m-t-xs', [
```
alexschiller: I know this is already merged, but this is not particularly maintainable and the same issue will come up if another read-only addon is put in. 

I'm fairly certain this can be resolved by adding a check for a checkoutUser: `ctrl.file.checkoutUser`
@sloria

sloria: Thanks for the suggestion @alexschiller . @mfraezz Can you take a look at this?

PR #7547, opened 2017-08-03
[Feature] Fix api deleted comments [OSF-7290]

fabmiz:
## Purpose

- Return only deleted comments that logged-in user authored (can restore).
- Return only non deleted comments to unauth'd users (public nodes)

## Changes

- Modify api `nodeCommentsList` to filter deleted comments/replies

## Ticket

[OSF-7290](https://openscience.atlassian.net/browse/OSF-7290)

## Before

before

## After

after

## Tests

Add tests for the fix


alexschiller:
Can you close this PR?

PR #7551, opened 2017-08-03
Node taxonomies backend [OSF-8275]

mfraezz:
## Purpose
Add Subjects to nodes and registrations

## Changes
* Add Subjects to nodes and registrations
* Add tests

## Side effects
None expected

## QA Notes
* No UI changes.
* API change - nodes now accept `subjects`, using the same conventions as preprints. This means:
    * If you're an admin, you can set the subjects on the node.
    * If you're not an admin, you can't.
    * If you have view permissions on the node, you can see what subjects it has.

## Ticket
[[OSF-8275]](https://openscience.atlassian.net/browse/OSF-8275)


osf/models/mixins.py
```diff
@@ -459,3 +465,59 @@ def get_extra_log_params(self, comment):
         """Return extra data to pass as `params` to `Node.add_log` when a new comment is
         created, edited, deleted or restored."""
         return {}
+
+class TaxonomizableMixin(models.Model):
```
caseyrollins: I know these functions existed before, but I *really* think `get_subjects()` and `set_subjects` are prime candidates for docstrings. Some bits that I found confusing when reading through:

For `set_subjects`:
- The `preprint_subjects` argument is a list of lists of `Subject` `_ids` (not the objects themselves). 
- Does `preprint_subjects` replace the current object's subjects, or add to them? `self.subjects.clear()` makes me think the former but I would've expected this function to do the latter.

For `get_subjects`:
- What exactly is this function returning? A list of lists of dictionaries containing subject `_ids` and `text`? A documented example return value could be extremely useful so devs don't need to parse through nested for loops that makes calls to a property that contains a list comprehension of another property.

osf/models/mixins.py
```diff
@@ -459,3 +465,59 @@ def get_extra_log_params(self, comment):
...
+
+    def set_subjects(self, preprint_subjects, auth, add_log=True):
+        if self.is_registration:
+            raise PermissionsError('Registrations may not be modified.')
+        if self.is_collection:
+            raise NodeStateError('Collections may not have subjects')
+        if not self.has_permission(auth.user, ADMIN):
+            raise PermissionsError('Only admins can change subjects.')
+
+        old_subjects = list(self.subjects.values_list('id', flat=True))
```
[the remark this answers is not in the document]
mfraezz: Keeping a QS in memory causes the query to be executed each evaluation. When the underlying data changes, so does the result of the evaluation -- meaning it is necessary here.
```
In [8]: olds1 = p.subjects  # .copy isn't a SubjectQuerySet method

In [9]: olds2 = p.subjects.all()

In [10]: p.subjects.count()
Out[10]: 4

In [11]: olds2.count()
Out[11]: 4

In [12]: olds1.count()
Out[12]: 4

In [13]: p.subjects.clear()

In [14]: p.subjects.count()
Out[14]: 0

In [15]: olds1.count()
Out[15]: 0

In [16]: olds2.count()
Out[16]: 0
```
You could work around this with something like 
```
old_sub_qs = Subject.objects.filter(id__in=list(self.subjects.values_list('id', flat=True)))
```
But this doesn't save you from making queries on subsequent evaluations:
```
In [2]: old_sub_qs = Subject.objects.filter(id__in=list(p.subjects.values_list('id', flat=True)))

In [3]: from django.test.utils import CaptureQueriesContext

In [4]: from django.db import connection

In [5]: with CaptureQueriesContext(connection) as ctx0:
   ...:     list(old_sub_qs.all())
   ...:

In [6]: ctx0.final_queries
Out[6]: 4

In [7]: ctx0.final_queries - ctx0.initial_queries
Out[7]: 1

In [8]: with CaptureQueriesContext(connection) as ctx1:
   ...:     list(old_sub_qs.values('id', 'text'))
   ...:

In [9]: ctx1.final_queries
Out[9]: 5

In [10]: ctx1.final_queries - ctx1.initial_queries
Out[10]: 1

```
In order to avoid that, we'd need to evaluate `old_subjects` to a list of objects and pass that around instead, using comprehensions to extract relevant data. Given the (small) extra memory cost, and the fact that it would only save us a couple queries on a small / relatively bounded table, I'm going to say this isn't worth it.

osf/models/mixins.py
mfraezz: Docstring added for `set_subjects`. `get_subjects` was only being used in one place, and unnecessarily. Removed and replaced with a `values_list` call in beceb3e0e26563296a114f49b01e0b1d4a76fdc9

PR #7552, opened 2017-08-04
[feature] Pytest Conversion: api_tests/users [OSF-8329]

Nesiehr:
## Purpose

Testing Performance Increase

## Changes

Pytest conversion of api_tests/users directory (except for test_user_addons.py)
Fixing of two mixins (test_preprint_list_mixin.py mixins)

## Side effects

None known.

## Ticket

https://openscience.atlassian.net/browse/OSF-8329

## QA Considerations

N/A


api_tests/users/views/test_user_list.py
```diff
@@ -1,496 +1,507 @@
 # -*- coding: utf-8 -*-
-from uuid import UUID
-
+from django.db.models import Q
```
alexschiller: Don't think this is used

sloria:
Fixed flake errors when I merged.

PR #7554, opened 2017-08-04
[feature] Add button to remove bad prereg challenge checkouts [OSF-8333]

alexschiller:
## Purpose
Checkins don't seem to always occur when a prereg challenge is approved/rejected. At least one source of these was fixed (deleting a prereg submission while under review). Other sources may still exist that have not been identified. To make it easier to handle them, a button has been added to the prereg view that allows bad checkouts to be identified and handled all at once.

## Changes
In addition to the button mention below, Clicking to view approved and rejected prereg challenge items should no longer check out files. 

Add the below button that removes checkouts by prereg admins from rejected or approved prereg nodes. 
screen shot 2017-08-04 at 4 11 04 pm
screen shot 2017-08-04 at 4 11 14 pm

## Side effects
None known.

## Ticket

https://openscience.atlassian.net/browse/OSF-8333

## QA Notes
1. Three projects: Approved prereg challenge, Unapproved (active prereg submission), Rejected prereg challenge.
2. Add a prereg admin to the project and manually checkout a file.
3. Go to admin app and click on the prereg view and click the "Checkout Checkup" button
4. The count should match the number of projects that have bad prereg checkouts
5. Clicking the button should remove the checkouts from files checked out to the admin from the approved and rejected, but not the unapproved prereg challenge
6. It should not affect any other checkouts but nodes with approved or rejected submissions.
7. Checkouts by other users should not be checked in
8. Viewing approved and rejected preregs on admin should not re-check out the files now
9. Unapproved checkouts should work normally.


alexschiller:
PR doesn't even modify that file 🙄
Updating with dev.

admin_tests/pre_reg/test_views.py
```diff
@@ -369,6 +376,82 @@ def test_checkin_files(self):
         for q, f in self.d_of_qs.iteritems():
             nt.assert_equal(None, f.checkout)
 
+    def test_approved_does_not_checkout_files(self):
```
brianjgeiger: So in the vein of "fewer tests make the testing go faster", would it make sense to make one or two tests that take a file through the various states and checks at each of those states rather than having all the different tests?

brianjgeiger:
@alexschiller Conflicts

PR #7557, opened 2017-08-07
[#OSF-8106]ListNotSubscribedError: d.buccarella@sinapsys.it is not subscribed to list Open Science Framework General

chennan47:
## Purpose
Fix the problem that the unhandled error is preventing user's mailchimp_mailing_lists fields from getting properly updated when a user is already unsubscribed to a mailing list.

## Changes
After this change, you should no longer see those errors preventing user's mailchimp_mailing_lists from updating.

## Ticket
https://openscience.atlassian.net/browse/OSF-8106


website/mailchimp_utils.py
```diff
@@ -76,7 +76,14 @@ def unsubscribe_mailchimp(list_name, user_id, username=None, send_goodbye=True):
...
     list_id = get_list_id_from_name(list_name=list_name)
-    m.lists.unsubscribe(id=list_id, email={'email': username or user.username}, send_goodbye=send_goodbye)
+    
+    try:
+        m.lists.unsubscribe(
+            id=list_id, email={'email': username or user.username},
+            send_goodbye=send_goodbye
+        )
+    except mailchimp.ListNotSubscribedError:
+        pass
```
brianjgeiger: Could you add a comment as to why this is an acceptable pass? Basically, what is the situation around the error, why it's expected, and why we shouldn't be concerned about it? I imagine it is mostly "They were already unsubscribed", but a little bit if you know it about why we don't know they were unsubscribed would be great.

chennan47: The error comes form when trying to do a queue task to unsub a user who has already unsubscribed from the mailchimp. Since user has already unsubscribed and all this function to do is unsubscribe user from mailchimp, it does no harm to silently pass the error and allow to update mailing_list user field.

brianjgeiger: Awesome. Just put a quick comment about it in code, though, so future generations understand.

PR #7563, opened 2017-08-08
[OSF-7927] View-only link returns 403 for add-on files

TomBaxter:
[#OSF-7927]

## Purpose

View only link API requests to api.nodes.views.NodeFilesList are failing because 'view_only' kwarg s not being passed.

Add 'view_only' kwarg to WB request, if present in OSF request.

## Changes

Update api/nodes/views.py class NodeFilesList

Check for 'view_only' in request.query_params and pass to waterbutler_api_url_for if found.

Added test api_tests/nodes/views/test_node_files_list.py::TestNodeFilesList::test_vol_node_files_list

## Side effects

None

## Ticket

[OSF-7927](https://openscience.atlassian.net/browse/OSF-7927)

### DIFF
--- a/api/nodes/utils.py
+++ b/api/nodes/utils.py
@@ -28,7 +28,10 @@ def get_file_object(node, path, provider, request):
     if not node.get_addon(provider) or not node.get_addon(provider).configured:
         raise NotFound('The {} provider is not configured for this project.'.format(provider))
 
-    url = waterbutler_api_url_for(node._id, provider, path, _internal=True, meta=True)
+    view_only = request.query_params.get('view_only', default=None)
+    url = waterbutler_api_url_for(node._id, provider, path, _internal=True,
+                                  meta=True, view_only=view_only)
+
     waterbutler_request = requests.get(
         url,
         cookies=request.COOKIES,
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -623,7 +623,9 @@ class RelationshipField(ser.HyperlinkedIdentityField):
                         else:
                             url = None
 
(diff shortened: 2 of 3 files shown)


api/nodes/utils.py
```diff
@@ -28,7 +28,10 @@ def get_file_object(node, path, provider, request):
     if not node.get_addon(provider) or not node.get_addon(provider).configured:
         raise NotFound('The {} provider is not configured for this project.'.format(provider))
 
-    url = waterbutler_api_url_for(node._id, provider, path, _internal=True, meta=True)
+    kwargs = {}
+    if 'view_only' in request.GET:
+        kwargs['view_only'] = request.query_params.get('view_only', default=None)
+    url = waterbutler_api_url_for(node._id, provider, path, _internal=True, meta=True, **kwargs)
```
brianjgeiger: I lean towards being a little more explicit here and replacing this whole block with: 

```py
view_only = request.query_params.get('view_only', default=None)
url = waterbutler_api_url_for(node._id, provider, path, _internal=True, meta=True, view_only=view_only **kwargs)
```

That should do, essentially, the same thing, I think. You shouldn't need to check to se if `view_only` is in the request while also setting a default of None for getting the query parameter value.

That being said, if this is how we do it everywhere else, and or my way doesn't work, then your way is delightful and we'll go with that.

brianjgeiger:
Oh, also a unit test, please.

api/nodes/utils.py
TomBaxter: Complete.

TomBaxter:
CR response complete.

PR #7564, opened 2017-08-08
[OSF-8007] Cancel changes button on admin preprint config

Johnetordoff:
## Purpose

When modifying preprint info on the admin page there's no way of discarding your changes, even if you restart the page everything is cached and your likely to forget your initial configuration. This fix gives the user a hard refresh in the form of a Discard button.

## Changes

Add a new discard buttom with a js prompt that restarts the page without a cache.

## Side effects

None that I know of.

## Before

screen shot 2017-08-08 at 11 36 52 am

## After 

admin_page_cancel_btn

## Ticket

https://openscience.atlassian.net/browse/OSF-8007

### DIFF
--- a/admin/static/js/preprint_providers/preprintProviders.js
+++ b/admin/static/js/preprint_providers/preprintProviders.js
@@ -100,6 +100,12 @@ $(document).ready(function() {
             });
         }
     });
+    $("#discard").click(function(e) {
+        e.preventDefault();
+        if (window.confirm('Are you sure want to discard all your changes?')) {
+            location.reload(true);
+        }
+    });
 
     $("#show-modify-form").click(function() {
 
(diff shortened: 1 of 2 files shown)


Johnetordoff:
@brianjgeiger did you find any?

brianjgeiger:
Hmmmm. I wonder if that was supposed to say, "_Check_ on potential alternate implementations." If not, it probably should at this point. So: See if you can find a decent way to do that without, you know, writing a whole big thing for it. If this were a normal admin app, we could just enable the cancel button and be done with it. But if there's a quick, "store all the contents of these fields into a variable and, if the cancel button is hit, reset those fields to what they were when the page loaded" type of control, then yay. If not, we can go with this for now.

Johnetordoff:
I checked on typical ways you might think to repopulate things with js nothing seemed remotely appetizing a lot of the data is templated server-side and not exposed to js anyway so rest assured this is the most palatable option for now.

PR #7566, opened 2017-08-09
[feature] Remove junk CSS and mako files

sloria:
## Purpose

Remove junk code and reduce vendor bundle size.

## Changes

- Remove unused CSS and mako files
- Removed a few unused CSS rules
- Removed c3
- Removed two modules from the vendor bundle that were only used in one place

## Side effects

- Faster asset builds
- Slightly smaller vendor bundle
- Less junk in the codebase

## Ticket

Tangentially related to https://openscience.atlassian.net/browse/OSF-8356


sloria:
Travis is passing on my fork, merging.

PR #7567, opened 2017-08-09
[OSF-8278] Fix JS tests and run them on Travis

Johnetordoff:
## Purpose

Currently our Karma tests in for JS are in a state of disrepair, not running in Travis and causing errors locally, this PR aims to restore then to a proper functional state.

## Changes

Shuffles around some the versions of Karma and Sinon to reach something compatible with our tests, also modifies python inv tasks so Karma will run with other python tests.

## Side effects

This opens up a lot of previously unused tests, expect a lot of them to be out of date or fail.

## Ticket

https://openscience.atlassian.net/browse/OSF-8278


tasks/__init__.py
```diff
@@ -520,6 +520,7 @@ def test(ctx, all=False, syntax=False):
         flake(ctx)
         jshint(ctx)
 
+    test_js(ctx)
```
alexschiller: driveby: Should this be a new build or added to our current fasted? Probably can be added to api1 (fastest) and will still be faster than our longest. @sloria @brianjgeiger

brianjgeiger: I believe we are currently maxed out on branches, so added to our fastest, please.

tasks/__init__.py
```diff
@@ -537,6 +537,11 @@ def test_js(ctx):
     jshint(ctx)
     karma(ctx, single=True, browsers='PhantomJS')
 
+@task
+def test_travis_js(ctx, numprocesses=None):
```
brianjgeiger: Since we're putting this in api1, I don't think we need this task for travis. Also, could you rename the branch on travis so it's called `api1_and_js` or something?

alexschiller:
@Johnetordoff Getting the error I expected. Can you (briefly) look into getting karma to work with es6. Failing that, I can convert my other changes to be non-es6.

Johnetordoff:
@alexschiller es6 doesn't seem to work without additional dependencies which themselves seem pretty early in there development.

PR #7568, opened 2017-08-09
feature/fixes to preprint provider import export[OSF-8236]

laurenrevere:
## Purpose

My [PR](https://github.com/CenterForOpenScience/osf.io/pull/7521/) for fixing preprint provider import export did not work correctly on staging. 

## Changes

This PR removes taxonomy import/export and fixes some unicode and licenses problems. 

## Side effects

Importing/Exporting taxonomies will continue to not work.

## Testing Nodes
Subjects should not change on import. A future ticket will fix this. If you are creating a new preprint provider it will assume the default - which is the full bepress taxonomy.

## Ticket
https://openscience.atlassian.net/browse/OSF-8236

### DIFF
--- a/admin_tests/preprint_providers/test_views.py
+++ b/admin_tests/preprint_providers/test_views.py
@@ -246,9 +246,9 @@ class TestPreprintProviderExportImport(AdminTestCase):
 
         nt.assert_equal(res.status_code, 302)
         nt.assert_equal(new_provider._id, 'new_id')
-        nt.assert_equal(new_provider.subjects.all().count(), 1)
+        # nt.assert_equal(new_provider.subjects.all().count(), 1)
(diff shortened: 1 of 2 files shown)


admin_tests/preprint_providers/test_views.py
```diff
@@ -246,9 +246,9 @@ def test_export_to_import_new_provider(self):
 
         nt.assert_equal(res.status_code, 302)
         nt.assert_equal(new_provider._id, 'new_id')
-        nt.assert_equal(new_provider.subjects.all().count(), 1)
+        # nt.assert_equal(new_provider.subjects.all().count(), 1)
```
brianjgeiger: There'll be a follow-up ticket to get these back in, so we'll leave these commented for the moment (not the usual procedure, but shooooould be temporary.

PR #7570, opened 2017-08-09
Send appropriate date_updateds to SHARE

chrisseto:

### DIFF
--- a/website/preprints/tasks.py
+++ b/website/preprints/tasks.py
@@ -62,7 +62,13 @@ def format_preprint(preprint, old_subjects=None):
             preprint.node.tags.filter(name='qatest').exists() or
             preprint.node.is_deleted
         ),
-        'date_updated': preprint.date_modified.isoformat(),
+        # Note: Changing any preprint attribute that is pulled from the node, like title, will NOT bump
+        # the preprint's date modified but will bump the node's date_modified.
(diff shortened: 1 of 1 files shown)


website/preprints/tasks.py
```diff
@@ -62,7 +62,13 @@ def format_preprint(preprint, old_subjects=None):
...
             preprint.node.is_deleted
         ),
-        'date_updated': preprint.date_modified.isoformat(),
+        # Note: Changing any preprint attribute that is pulled from the node, like title, will NOT bump
+        # the preprint's date modified but will bump the node's date_modified.
+        # We have to send the latest date to SHARE to actually get the result to be updated.
+        # If we send a date_updated that is <= the one we previously sent, SHARE will ignore any changes
+        # because it looks like a race condition that arose from preprints being resent to SHARE on
+        # every step of preprint creation.
+        'date_updated': max(preprint.date_modified, preprint.node.date_modified).isoformat(),
```
alexschiller: Driveby: This is really not ideal and should probably wait. There is already a ticket in DjangoLE to separate nodes and preprints and some steps have been made to catch -some- of the cases where nodes do not change preprint dates. Unless you intend on reindexing everything with proper dates when nodes and preprints are separated, this will really should not move forward. There was a fairly lengthy discussion in B/I about this about a week ago @brianjgeiger @laurenrevere

PR #7572, opened 2017-08-10
[OSF-8083] Read-only contributors should be able to unauthorize addons connected to their external accounts

adlius:
## Purpose

If a contributor authorized their external accounts for project addons, it cannot disconnect its external accounts from the project. Currently, if such a contributor wants to disconnect through user settings page, a 403 is returned. The configuration section is also not available on the project settings page for such contributors.

## Changes

To make sure a read-only contributor can disconnect its external account on user settings page, the decorator `@must_have_permission(permissions.WRITE)` is changed to `@must_be_addon_authorizer(addon_short_name)` for the `_deauthorize_node()` method.

For the project settings page, an `is_owner` field is added to the context in `node.py` for every enabled addon. Changes were make to `settings.mako` to make use of the field.

## Side effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/OSF-8083

### DIFF
--- a/addons/base/generic_views.py
+++ b/addons/base/generic_views.py
@@ -113,7 +113,7 @@ def set_config(addon_short_name, addon_full_name, Serializer, set_folder):
 def deauthorize_node(addon_short_name):
     @must_not_be_registration
     @must_have_addon(addon_short_name, 'node')
-    @must_have_permission(permissions.WRITE)
+    @must_be_addon_authorizer(addon_short_name)
     def _deauthorize_node(auth, node_addon, **kwargs):
         node_addon.deauthorize(auth=auth)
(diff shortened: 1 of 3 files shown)


adlius:
@alexschiller It looks like it only changes the way we select the addons but not the way we configure them. But sure this will have to wait until @Johnetordoff 's PR is merged and then I'll see what changes are needed.

Johnetordoff:
Yep, should only effect front end changes, once it merged it will just require a few tweaks.

alexschiller:
@adlius Flake8. Also, since this is waiting on John's PR, I'm going to drop it into on hold and mark it as blocked.

PR #7573, opened 2017-08-10
Hotfix/update share preprint date modified[OSF-8388]

laurenrevere:
## Purpose

SHARE is not actually updated when most preprint fields change. This is because it relies on the `date_modified` attribute of preprints, which currently only updates on preprint exclusive fields (those not on the node). When preprints are separated from nodes, this problem will disappear, but until then, we want to send SHARE a more realistic`date_modified` for preprints. 

## Changes

This sends the greater of the node `date_modified` or the preprint `date_modified` to SHARE whenever a node associated with a preprint is updated.

There is also a management command that resends SHARE data for any node modified later than its associated preprint for all other preprints - this sends SHARE the later node `date_modified`.

Also, because it is hard to check when a primary file is updated, SHARE is updated whenever a preprint node is updated.

## Side effects
SHARE data is now sent any time a preprint node it updated.

The `date_modified` of a node does not link up exactly with what we would want the `date_modified` of a preprint to be, so SHARE might have modified dates that are slightly newer than the preprint data actually is.

## Ticket

https://openscience.atlassian.net/browse/OSF-8388

### DIFF
--- a/osf/management/commands/update_preprint_share_dates.py
+++ b/osf/management/commands/update_preprint_share_dates.py
@@ -1,0 +1,43 @@
+from __future__ import unicode_literals
+import logging
+
+from django.core.management.base import BaseCommand
+
+from scripts import utils as script_utils
+from osf.models import PreprintService
+from website.preprints.tasks import on_preprint_updated
+
+logger = logging.getLogger(__name__)
+
+def update_share_preprint_modified_dates(dry_run=False):
+    dates_updated = 0
+    for preprint in PreprintService.objects.filter():
+        if preprint.node.date_modified > preprint.date_modified:
+            if not dry_run:
+                on_preprint_updated(preprint._id)
+            dates_updated += 1
+    return dates_updated
+
+class Command(BaseCommand):
+    """
+    Send more accurate preprint modified dates to Share (max of node.date_modified and preprint.date_modified)
+    """
+    def add_arguments(self, parser):
+        super(Command, self).add_arguments(parser)
+        parser.add_argument(
+            '--dry',
+            action='store_true',
+            dest='dry_run',
+            help='Say how many preprint updates would be sent to share',
+        )
+
+    def handle(self, *args, **options):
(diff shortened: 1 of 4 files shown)


osf/management/commands/update_preprint_share_dates.py
```diff
@@ -0,0 +1,43 @@
...
+
+from scripts import utils as script_utils
+from osf.models import PreprintService
+from website.preprints.tasks import on_preprint_updated
+
+logger = logging.getLogger(__name__)
+
+def update_share_preprint_modified_dates(dry_run=False):
+    dates_updated = 0
+    for preprint in PreprintService.objects.filter():
```
sloria: Rather than filter in Python, you can do a self-referential query using `django.db.models.F`. Something like

```python
PreprintService.objects.filter(date_modified__lt=F('node__date_modified'))
```

osf/management/commands/update_preprint_share_dates.py
```diff
@@ -0,0 +1,43 @@
...
+        if preprint.node.date_modified > preprint.date_modified:
+            if not dry_run:
```
sloria: If would be useful to log the _id of each preprint to be sent.

osf/management/commands/update_preprint_share_dates.py
```diff
@@ -0,0 +1,43 @@
...
+            dest='dry_run',
+            help='Say how many preprint updates would be sent to share',
+        )
+
+    def handle(self, *args, **options):
+        dry_run = options.get('dry_run', False)
+        if not dry_run:
+            script_utils.add_file_logger(logger, __file__)
+            dates_updated = update_share_preprint_modified_dates()
+            logger.info('Sent %d new preprint modified dates to Share' % dates_updated)
```
sloria: Minor: "SHARE", not "Share".

brianjgeiger:
@sloria Pinging for review of your requested changes.

PR #7588, opened 2017-08-17
QuickFiles Improvements [OSF-8117]

erinspace:
## Purpose
QuickFiles needs a few improvements and a few additions to the file detail serializer when showing a detail page.

## Changes
- Change API routes from `v2/users/files/` to `v2/users/quickfiles/` per product request
- Change user file relationship to quickfiles
- add `BaseFileSerializer` and split into two subclasses:
    - normal `FileSerializer` for a standard file with the node relationship
    - `QuickFilesFileDetailSerializer` that instead has the user relationship
- use `OsfStorageFile` over `BaseFileNode` where appropriate
- use  `QuickFilesNode._typedmodels_type` over `osf.quickfilesnode` where appropriate

## Side effects

none anticipated

## Ticket

### DIFF
--- a/api/files/serializers.py
+++ b/api/files/serializers.py
@@ -295,7 +295,7 @@ class BaseFileSerializer(JSONAPISerializer):
         return instance
 
     def is_valid(self, **kwargs):
-        return super(FileSerializer, self).is_valid(clean_html=False, **kwargs)
+        return super(BaseFileSerializer, self).is_valid(clean_html=False, **kwargs)
 
     def get_file_guid(self, obj):
         if obj:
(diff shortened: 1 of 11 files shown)


api/files/serializers.py
```diff
@@ -336,6 +340,17 @@ class FileDetailSerializer(FileSerializer):
     id = IDField(source='_id', required=True)
 
 
+class QuickFilesSerializer(BaseFileSerializer):
+    user = RelationshipField(related_view='users:user-detail',
+                             related_view_kwargs={'user_id': '<node.creator._id>'},
+                             help_text='The user who updloaded this file'
```
sloria: Minor: Typo here.

api/files/serializers.py
```diff
@@ -336,6 +340,17 @@ class FileDetailSerializer(FileSerializer):
     id = IDField(source='_id', required=True)
 
 
+class QuickFilesSerializer(BaseFileSerializer):
+    user = RelationshipField(related_view='users:user-detail',
+                             related_view_kwargs={'user_id': '<node.creator._id>'},
```
sloria: @erinspace Can you make sure that accessing `node.creator._id` here isn't resulting in a lot of added queries? If so, we might need to do some prefetching or .include-ing.

PR #7595, opened 2017-08-21
[OSF-3622] Disallow user to choose citation styles with no bibliography section.

adlius:
## Purpose
Some citation styles could not be rendered because of the lack of bibliography format. This PR fixes this by limiting users to choose citations styles that have bibliographies.

## Changes
According to @mfraezz 's comment in [the original PR](https://github.com/CenterForOpenScience/osf.io/pull/5697),

> Ideally, we'd be able to prevent styles that we can't necessarily render correctly from being options presented to the user, rather than attempting to parse some json that may or may not be indicative of what the style is supposed to look like.

>Three main points to achieve this:
> * Add flag on the CitationStyle model -- indicates presence of bibliography tag in the xml tree
> * Modify the default query to exclude models without a bib
> * Modify script to set flag on models.

Changes are made accordingly. A migration is also added.

## Side effect

This is in conflict with [this PR](https://github.com/CenterForOpenScience/osf.io/pull/7153) where instead of the three different csl files parsed, another bluebook.cls (that has the bibliography section) is actually served to browser for all three Bluebook citations as a hotfix, resulting in inconsistency between what was parsed and what is actually served. The easiest remedy seems to be manually setting the `has_bibliography` field to `True` for the three Bluebook CitationStyle instances.

## Ticket

https://openscience.atlassian.net/browse/OSF-3622


adlius:
Per @brianjgeiger 's request, some notes are added to `parse_citation_styles.py` as document for future reference.

brianjgeiger:
Please update your migration

adlius:
Migration updated.

brianjgeiger:
@adlius If you have a chance this weekend, could you fix the merge conflict on this? I'm trying to get all the 0.126.0 tickets in before Monday. You'll probably need to update the migration as well.

adlius:
@brianjgeiger Sorry for the delay. Top priority now for fixing these tickets now. Should be done fairly quickly.

brianjgeiger:
Cool, thanks, and no trouble.

adlius:
Merge conflict fixed. Migration is remade.

PR #7599, opened 2017-08-22
[OSF-7519] Make sure sharejs shows disconnect status when it fails.

Johnetordoff:
## Purpose

When the sharejs server is down the status label will persist as "Connecting" even when it's ceased trying to connect. This fix show conclusively when the connection fails.

## Changes

Simple one line js fix.

## Side effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/OSF-7519


Johnetordoff:
@alexschiller The reconnecting websocket we use basically never timesout, even when the setting are set to give a limited number of retries it will continue to try forever if the url can not be found. Not sure how to handle this.

addons/wiki/static/ShareJSDoc.js
```diff
@@ -180,6 +180,8 @@ var ShareJSDoc = function(url, metadata, viewText, editor) {
         onopen(event);
         if (madeConnection) {
             viewModel.status('connected');
+        } else {
+            viewModel.status('disconnect');
```
brianjgeiger: Shouldn't this be 'disconnected' rather than 'disconnect'?

Johnetordoff: Yes, for consistencies sake, but any status other than connecting or connected is going to be read as disconnected

PR #7603, opened 2017-08-22
[OSF-8339]API 502 - Continued

adlius:
## Purpose

The [previous PR](https://github.com/CenterForOpenScience/osf.io/pull/7544) made a change to log to Sentry upon certain condition. However made the same mistake of failing to consider items with undefined 'modified' field. This PR fix this problem and adds a unit test checking for Sentry log.

## Changes

Changes to logging condition to ignore when 'entry['modified']' is None. Added a new unit test.

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-8339

### DIFF
--- a/tests/test_addons.py
+++ b/tests/test_addons.py
@@ -927,14 +927,12 @@ class TestAddonFileViews(OsfTestCase):
                 ignoretz=True,
                 default=timezone.now()  # Just incase nothing can be parsed
             )})
-        print(file_node.history[0])
         data = {
             'name': 'a name',
             'materialized': 'materialized',
             'modified': '2016-08-22T13:54:32.100900'
         }
         file_node.update(revision=None, user=None, data=data)
-        print(file_node.history[1])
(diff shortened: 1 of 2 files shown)


tests/test_addons.py
```diff
@@ -898,9 +914,28 @@ def test_missing_modified_date_in_file_history(self, mock_archive):
...
+    @with_sentry
+    @mock.patch('framework.sentry.sentry.captureMessage')
+    def test_update_logs_to_sentry_when_called_with_disordered_metadata(self, mock_capture):
+        file_node = self.get_test_file()
+        file_node.history.append({'modified': parse_date(
+                '2017-08-22T13:54:32.100900',
+                ignoretz=True,
+                default=timezone.now()  # Just incase nothing can be parsed
+            )})
+        print(file_node.history[0])
```
alexschiller: Driveby: Two lingering print statements in this test

adlius:
Two print statements removed.

PR #7604, opened 2017-08-22
[OSF-8519] Test failure test_get_preprint_count

Johnetordoff:
## Purpose

The test test_get_preprint_count needs Keen which doesn't allow same-day dates to be queried, you can't for example ask them how many users visited the site on Sept 25th 2018 if Sept 25th 2018 hasn't happen or isn't over yet. This means we have to query for dates in the past. On top of that since we aren't querying for a specific time all out dates are rounded to midnight so dates at 11 round to the next day causing an error, this PR fixes that.

## Changes

Pushes date back another hour so we avoid the rounding error.

## Side effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/OSF-8519


Johnetordoff:
I know these are mostly still relative dates, but I can't think of better ones. I still don't know definitively what causes the failure.

osf_tests/test_preprint_summary.py
```diff
@@ -19,11 +18,33 @@ def preprint_provider():
...
+@pytest.fixture()
+def yesterday_right_before_today():
+    return datetime.datetime.utcnow().replace(hour=23,
+                                              minute=59,
+                                              second=59) - datetime.timedelta(days=1)
+
+
+@pytest.fixture()
+def my_birthday_at_midnight():
+    return datetime.datetime(1991, 9, 25, 0)
```
alexschiller: Add timezone info or you may get some lovely naive datetime issues later:
`datetime.datetime(1991, 9, 25, 0, tzinfo=pytz.utc)`

osf_tests/test_preprint_summary.py
```diff
@@ -19,11 +18,33 @@ def preprint_provider():
...
+
 pytestmark = pytest.mark.django_db
 
+
+@pytest.mark.parametrize("date", [yesterday_at_midnight(), yesterday_right_before_today(), my_birthday_at_midnight()])
```
alexschiller: Use single quotes: 'date'

osf_tests/test_preprint_summary.py
```diff
@@ -19,11 +18,33 @@ def preprint_provider():
 def preprint(preprint_provider):
     return PreprintFactory._build(PreprintService, provider=preprint_provider)
 
+
+@pytest.fixture()
+def yesterday_at_midnight():
+    return datetime.datetime.utcnow().replace(hour=0,
```
alexschiller: If you have to use `...utcnow()`, you're probably better off using the timezone aware django util:
```
timezone.now()

```
See https://docs.djangoproject.com/en/1.11/topics/i18n/timezones/

alexschiller: Fun sidenote, `utcnow` does not actually give it tzinfo, but `timezone.now()` does.
screen shot 2017-08-24 at 4 52 25 pm

osf_tests/test_preprint_summary.py
```diff
@@ -19,11 +19,33 @@ def preprint_provider():
 def preprint(preprint_provider):
     return PreprintFactory._build(PreprintService, provider=preprint_provider)
 
+
+@pytest.fixture()
+def yesterday_at_midnight():
+    return timezone.now().replace(hour=0,
```
brianjgeiger: Let's return a dictionary `preprint_date` and `run_date` for each of these fixtures. If you need, you could also do `expected_result`. Instead of basing on now, just use specific dates for each. like 8/25/2017 at 11:59 UTC for the preprint and 8/26/2017 at 12:00 UTC for the run date or whatever. You can replicate these tests but with absolute dates and we'll be set up for more interesting tests later. Then this won't have a way of being probabilistic any more.

osf_tests/test_preprint_summary.py
```diff
@@ -19,11 +19,26 @@ def preprint_provider():
 def preprint(preprint_provider):
     return PreprintFactory._build(PreprintService, provider=preprint_provider)
 
+
+@pytest.fixture()
+def right_before_my_birthday():
+    return {'preprint_dates': datetime.datetime(1991, 9, 25, 0, tzinfo=pytz.utc).replace(hour=23,minute=59,second=59),
```
alexschiller: Minor flake: spaces after commas (`hour=23, minutes=59, `... etc)

osf_tests/test_preprint_summary.py
```diff
@@ -19,11 +19,26 @@ def preprint_provider():
 def preprint(preprint_provider):
     return PreprintFactory._build(PreprintService, provider=preprint_provider)
 
+
+@pytest.fixture()
+def right_before_my_birthday():
+    return {'preprint_dates': datetime.datetime(1991, 9, 25, 0, tzinfo=pytz.utc).replace(hour=23, minute=59, second=59),
+            'run_date': timezone.now()
```
brianjgeiger: Is there a reason that this isn't an absolute date? Does something not work if it's not relative?

Johnetordoff: I thought the idea was the dictionary would keep the current date if it failed again we'd know exactly when it failed. Now I'm unsure why "run_date" is included at all.

brianjgeiger: `preprint_date` is the input for preprint.date_created. `run_date` is the input for `PreprintSummary().get_events`. Does that make sense?

brianjgeiger: If not, drop by and we can chat.

Johnetordoff: Makes sense.

PR #7609, opened 2017-08-23
Add title and contributors to preprint serialzier [OSF-8499]

erinspace:
## Purpose

This is a starting step in the process of separating preprints and nodes -- starting out with the node's title and contributors on the preprint serializer (and making sure the contributor's relationship is embeddable) will make Moderation easier!

## Changes
- add node title and node contributors fields as read only
- add tests to make sure they're there
- add test to make sure node contributors are embeddable on the preprint API view

## Side effects

none anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-8499


sloria:
Merged into develop in #7610

PR #7615, opened 2017-08-23
[OSF-8421] Change "preprint" text to "thesis," etc in emails

adlius:
## Purpose

To change the wording in emails sent by branded preprint sites.

## Changes

Change three email templates to make use of `preprint_word_singular` and `preprint_word_plural` fields. Add a migration file.

## Side effects

The change only works if the `preprint_word_singular` and `preprint_word_plural` fields for each PreprintProvider instances are set on prod/staging.

## Ticket

https://openscience.atlassian.net/browse/OSF-8421

### DIFF
--- a/website/mails/mails.py
+++ b/website/mails/mails.py
@@ -177,7 +177,7 @@ CONFIRM_EMAIL_ERPC = Mail(
 )
 CONFIRM_EMAIL_PREPRINTS = lambda name, provider: Mail(
     'confirm_preprints_{}'.format(name),
-    subject='Open Science Framework Account Verification, {} Preprints Service'.format(provider)
+    subject='Open Science Framework Account Verification, {}'.format(provider)
 )
 CONFIRM_EMAIL_REGISTRIES_OSF = Mail(
     'confirm_registries_osf',
@@ -197,7 +197,7 @@ INVITE_DEFAULT = Mail(
 )
 INVITE_PREPRINT = lambda template, provider: Mail(
     'invite_preprints_{}'.format(template),
-    subject='You have been added as a contributor to {} {} preprint.'.format(get_english_article(provider), provider)
+    subject='You have been added as a contributor to {} {} {}.'.format(get_english_article(provider.name), provider.name, provider.preprint_word)
 )
 CONTRIBUTOR_ADDED_DEFAULT = Mail(
     'contributor_added_default',
@@ -205,7 +205,7 @@ CONTRIBUTOR_ADDED_DEFAULT = Mail(
 )
 CONTRIBUTOR_ADDED_PREPRINT = lambda template, provider: Mail(
     'contributor_added_preprints_{}'.format(template),
-    subject='You have been added as a contributor to {} {} preprint.'.format(get_english_article(provider), provider)
+    subject='You have been added as a contributor to {} {} {}.'.format(get_english_article(provider.name), provider.name, provider.preprint_word)
 )
 CONTRIBUTOR_ADDED_PREPRINT_NODE_FROM_OSF = Mail(
     'contributor_added_preprint_node_from_osf',
--- a/website/templates/emails/confirm_preprints_branded.txt.mako
+++ b/website/templates/emails/confirm_preprints_branded.txt.mako
@@ -1,6 +1,6 @@
 Hello ${user.fullname},
 
-Welcome to ${branded_preprints_provider} Preprints, powered by the Open Science Framework. To continue, please verify your email address by visiting this link:
+Welcome to ${branded_preprints_provider}, powered by the Open Science Framework. To continue, please verify your email address by visiting this link:
 
 ${confirmation_url}
 
--- a/website/templates/emails/invite_preprints_branded.txt.mako
+++ b/website/templates/emails/invite_preprints_branded.txt.mako
@@ -4,11 +4,11 @@
 
 Hello ${fullname},
 
-You have been added by ${referrer.fullname} as a contributor to the preprint "${node.title}" on ${branded_service.name}, powered by the Open Science Framework. To set a password for your account, visit:
+You have been added by ${referrer.fullname} as a contributor to the ${branded_service.preprint_word} "${node.title}" on ${branded_service.name}, powered by the Open Science Framework. To set a password for your account, visit:
 
 ${claim_url}
 
-Once you have set a password, you will be able to make contributions to "${node.title}" and create your own preprints. You will automatically be subscribed to notification emails for this preprint. Each preprint is associated with a project on the Open Science Framework for managing the preprint.  To change your email notification preferences, visit your project or your user settings: ${settings.DOMAIN + "settings/notifications/"}
+Once you have set a password, you will be able to make contributions to "${node.title}" and create your own ${branded_service.preprint_word}. You will automatically be subscribed to notification emails for this ${branded_service.preprint_word}. Each ${branded_service.preprint_word} is associated with a project on the Open Science Framework for managing the ${branded_service.preprint_word}.  To change your email notification preferences, visit your project or your user settings: ${settings.DOMAIN + "settings/notifications/"}
 
 To preview "${node.title}" click the following link: ${node.absolute_url}
 
--- a/website/project/views/contributor.py
+++ b/website/project/views/contributor.py
@@ -454,7 +454,7 @@ def send_claim_email(email, unclaimed_user, node, notify=True, throttle=24 * 360
             email_template, preprint_provider = find_preprint_provider(node)
             if not email_template or not preprint_provider:
                 return
-            mail_tpl = getattr(mails, 'INVITE_PREPRINT')(email_template, preprint_provider.name)
+            mail_tpl = getattr(mails, 'INVITE_PREPRINT')(email_template, preprint_provider)
         else:
             mail_tpl = getattr(mails, 'INVITE_DEFAULT'.format(email_template.upper()))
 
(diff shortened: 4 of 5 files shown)


website/templates/emails/invite_preprints_branded.txt.mako
```diff
@@ -4,11 +4,15 @@
...
 
 ${claim_url}
 
-Once you have set a password, you will be able to make contributions to "${node.title}" and create your own preprints. You will automatically be subscribed to notification emails for this preprint. Each preprint is associated with a project on the Open Science Framework for managing the preprint.  To change your email notification preferences, visit your project or your user settings: ${settings.DOMAIN + "settings/notifications/"}
+%if branded_service.name == 'Thesis Commons':
+Once you have set a password, you will be able to make contributions to "${node.title}" and create your own theses. You will automatically be subscribed to notification emails for this ${branded_service.preprint_word}. Each ${branded_service.preprint_word} is associated with a project on the Open Science Framework for managing the ${branded_service.preprint_word}.  To change your email notification preferences, visit your project or your user settings: ${settings.DOMAIN + "settings/notifications/"}
```
brianjgeiger: No, this won't work. We may one day have another word that won't be modifiable into a plural just by adding `s` to the end. Let's do something like modifying https://github.com/CenterForOpenScience/osf.io/blob/aef8d08d588d54148f4cfa013fc496bbeae93996/osf/models/preprint_provider.py#L43-L48 to have an extra item in the tuple for the plural version. Then you can derive the choice field from that and make a property (which you can use in the emails) that will give you the plural version for any provider. Or if you think of a better way to do the same sort of thing without having two lists that have to be maintained, go for that.

website/mails/mails.py
```diff
@@ -177,7 +177,8 @@ def get_english_article(word):
 )
 CONFIRM_EMAIL_PREPRINTS = lambda name, provider: Mail(
     'confirm_preprints_{}'.format(name),
-    subject='Open Science Framework Account Verification, {} Preprints Service'.format(provider)
+    subject='Open Science Framework Account Verification, {} Preprints Service'.format(provider) if provider != 'Thesis Commons'
```
brianjgeiger: I hate special casing for Thesis Commons, but I don't have a better suggestion right now, so it'll likely stay until we have to rip it out later.

brianjgeiger:
Oh, also could you target this towards Master branch, please. We'll need to make it a hotfix.

website/mails/mails.py
adlius: Maybe another field for the model like a "service_full_name" or "provider_full_name". But yes there's no better way if we want a hotfix and avoid migrations.

adlius:
Updated. Need a migration. Also, after deployed to either staging/production, the `preprint_word_singular` and  `preprint_word_plural` need to be updated.

adlius:
All sorts of conflicts with master......Rebasing does not solve. I'll close this PR and reopen one with a new branch.

osf/models/preprint_provider.py
```diff
@@ -40,13 +40,27 @@ class PreprintProvider(ObjectIDMixin, BaseModel):
...
-        ('thesis', 'Thesis'),
-        ('none', 'None')
+    PREPRINT = 'preprint'
+    PREPRINTS = 'preprints'
+    PAPER = 'paper'
+    PAPERS = 'papers'
+    THESIS = 'thesis'
+    THESES = 'theses'
+
+    PREPRINT_WORD_CHOICES_SINGULAR = (
```
brianjgeiger: This was not what I was hoping for. One of the big parts of the request was that we didn't want multiple lists to maintain, and now we have three instead of one.

How about instead:
```py
PREPRINT_WORD_CHOICS = [
    ['preprint', 'Preprint', 'Preprints'],
    ['thesis', 'Thesis', 'Theses'],
   (and so on)
]

preprint_singulars = tuple iteration of items 0 and 1 from each item in PREPRINT_WORD_CHOICES
preprint_word = models.CharField(max_length=10, choices=preprint_singulars, default='preprint')

@property
preprint_word_plural():
    Here just look up the current preprint word and return the plural version
```

osf/models/preprint_provider.py
```diff
@@ -40,13 +40,27 @@ class PreprintProvider(ObjectIDMixin, BaseModel):
...
+        (PREPRINT, 'Preprint'),
+        (PAPER, 'Paper'),
+        (THESIS, 'Thesis'),
     )
-    preprint_word = models.CharField(max_length=10, choices=PREPRINT_WORD_CHOICES, default='preprint')
```
brianjgeiger: I'm pretty sure this breaks the API because you'd need to change the API serializer to look this up. But rather than fixing the API, let's just refactor how this works to something less disruptive.

osf/models/preprint_provider.py
adlius: Okay.

adlius:
Updated the PR so that a migration is not needed. Change base to master.

website/mails/mails.py
sloria: How about just removing "Preprints Service":

> Open Science Framework Account Verification, Thesis Commons

> Open Science Framework Account Verification, SocArxiv

@nicipfeiffer

osf/models/preprint_provider.py
```diff
@@ -42,13 +42,15 @@ class PreprintProvider(ObjectIDMixin, BaseModel):
     allow_submissions = models.BooleanField(default=True)
     additional_providers = fields.ArrayField(models.CharField(max_length=200), default=list, blank=True)
 
-    PREPRINT_WORD_CHOICES = (
```
sloria: Convention is to have `_CHOICES` variables as 2-tuples.

Rather than breaking this convention, how about adding a separate variable with the plural words:

```python
PREPRINT_WORD_CHOICES = (
    ('preprint', 'Preprint'),
    ('thesis', 'Thesis'),
     # ...
)

PREPRINT_WORD_PLURALS = {
    'preprint': 'preprints',
    'thesis': 'theses',
    # ...
}

@property
def preprint_word_plural(self):
    return self.PREPRINT_WORD_PLURALS[self.preprint_word]
```

website/templates/emails/confirm_preprints_branded.txt.mako
```diff
@@ -1,6 +1,10 @@
 Hello ${user.fullname},
 
+% if branded_preprints_provider == "Thesis Commons":
```
sloria: A more robust check might be:


```mako
% if preprint_word != 'preprint':
```

website/mails/mails.py
sloria: For now, we can use a more robust check:

```python
...Preprints Service'.format(provider) if preprint_word != 'preprint' else ...
```

osf/models/preprint_provider.py
brianjgeiger: Well, I made @adlius jump through hoops so that we didn't have to maintain two lists. Can we do PREPRINT_WORDS as the three item tuples and have `preprint_word_singulars` become `preprint_words_choices`? Or am I just making a mess of things and you would much rather have two lists. 

If we do two lists, please add a test to ensure that both of them contain the same items.

website/mails/mails.py
brianjgeiger: It _could_ be the plural word unless it's Thesis Commons, but yeah, this is going to have to be re-worked soon.

website/templates/emails/confirm_preprints_branded.txt.mako
brianjgeiger: Yes, or this could also use the plural word instead or "Preprints" on L3/6

website/mails/mails.py
sloria: Talked to Nici on Flowdock, it is fine to remove "Preprints Service" in the subject, so the special casing can be removed.

osf/models/preprint_provider.py
sloria: @brianjgeiger I think your suggestion is fine. @adlius , go ahead with that.

website/templates/emails/confirm_preprints_branded.txt.mako
sloria: Discussed with Nici on FD, the special casing can be remove here too:

> Welcome to SoxArxiv, powered by the Open Science Farmework.

> Welcome to Thesis Commons, powered by the Open Science Framework.

website/templates/emails/invite_preprints_branded.txt.mako
```diff
@@ -4,11 +4,11 @@
...
 
 ${claim_url}
 
-Once you have set a password, you will be able to make contributions to "${node.title}" and create your own preprints. You will automatically be subscribed to notification emails for this preprint. Each preprint is associated with a project on the Open Science Framework for managing the preprint.  To change your email notification preferences, visit your project or your user settings: ${settings.DOMAIN + "settings/notifications/"}
+Once you have set a password, you will be able to make contributions to "${node.title}" and create your own ${branded_service.preprint_word_plural}. You will automatically be subscribed to notification emails for this ${branded_service.preprint_word}. Each ${branded_service.preprint_word} is associated with a project on the Open Science Framework for managing the ${branded_service.preprint_word}.  To change your email notification preferences, visit your project or your user settings: ${settings.DOMAIN + "settings/notifications/"}
```
brianjgeiger: Soooo. I apologize about this, but after Steve's fantastic suggestion to just get rid of the troublesome words above, making is word plural isn't really necessary. I chatted with Nici and we can just use the singular form here. Which means that we can get rid of all of the code you made to deal with the plurals and just make this a simple change of templates.

I think we've all, er, learned a lot by this journey through possibility, and, um, grown as people?

adlius: :expressionless: This is literally my face right now.
Yes, I'll go ahead and make the template change. :sweat_smile:

brianjgeiger: I am so sorry.

adlius: All is well. This is in fact a real learning opportunity. As a college student I was used to not question whatever was in the assignment. By habit, I've just never really thought about asking product to change the requirement. I guess if something is minor and it adds potential clutter to our codebase it is always worth asking whether it is actually needed.

adlius:
PR Updated.

PR #7616, opened 2017-08-23
[OSF-6527] Linkify wiki markdown

Johnetordoff:
## Purpose

Our current markdown-it settings don't mark-up URL and leave them of plain text. This fix automatically marks-up URL to clickable links in the wiki edit window.

## Changes

Simple js change enabled linkify and disables the "commonmark" markup which forbids linkification.

## Side effects

None that I know of.

## Before

screen shot 2017-08-23 at 1 11 51 pm

## After
screen shot 2017-08-23 at 1 07 29 pm

## Ticket

https://openscience.atlassian.net/browse/OSF-6527


brianjgeiger:
@Johnetordoff So you can't do commonmark with the exception of linkifying? The only way is to turn it off?

Johnetordoff:
Yes, I'm tested the markup behavior without commonmark and it seems unchanged. Linkify-ing and commonmark are not compatible, as far as I know.

brianjgeiger:
Okay, I'm checking with Nici. If we can't use the commonmark shorthand, then we definitely will need to re-apply all of the individual changes that make up commonmark (other than linkifying) if we go forward with this change.

Johnetordoff:
@brianjgeiger, sorry! false alarm, there was a way to enable linkify with commonmark.

PR #7621, opened 2017-08-24
[OSF-8477] Import contributors in one click

Johnetordoff:
## Purpose

There are a couple problems with the current import contributors feature.

- It doesn't import non-bibliographic contributors
- It doesn't import the contributors permissions
- You can't modify a contributors the bibliographic status before you add them, only their permissions.

This fix will set the permissions and bibliographic status on import so you can import everyone in one click.

## Changes

Changed the get_contributors_from_parent/ endpoint to get more info about the contributors permissions and bibliographic status, then modified the .mako and .js to display the new info without crowding the screen

## Before
before_import_contributors

## After
import_contribs_demo

## Side effects

Modified some of the wrapping in that dialog, so make sure that works in a sane way with long names.

## Testing 
WIP, will see once demo is approved

## Ticket

https://openscience.atlassian.net/browse/OSF-8477


website/templates/project/modal_add_contributor.mako
```diff
@@ -71,7 +71,7 @@
                                             <!-- height and width are explicitly specified for faster rendering -->
                                             <img data-bind="attr: {src: contributor.gravatar_url}" height=35 width=35 />
                                         </td>
-                                        <td width="75%">
+                                        <td width="75%" style="word-break: break-all">
```
alexschiller: No inline style please

website/templates/project/modal_add_contributor.mako
```diff
@@ -180,13 +189,19 @@
                                             <img data-bind="attr: {src: contributor.gravatar_url || '/static/img/unreg_gravatar.png'}" height=35 width=35 />
                                         </td>
 
-                                        <td>
+                                        <td style="word-break: break-all">
```
alexschiller: Another inline style

website/project/views/contributor.py
```diff
@@ -127,8 +127,8 @@ def get_contributors_from_parent(auth, node, **kwargs):
         raise HTTPError(http.FORBIDDEN)
 
     contribs = [
-        profile_utils.add_contributor_json(contrib)
-        for contrib in parent.visible_contributors
+        profile_utils.add_contributor_json(contrib, node=node)
+        for contrib in parent.contributors if contrib not in node.visible_contributors
```
brianjgeiger: Should this be `node.visible_contributors` or `node.contributors`?

website/templates/project/modal_add_contributor.mako
brianjgeiger: Should these be `break-all` or `normal`? Probably should check with @laurenrevere who just went through a journey on breaking things at words. Also, yes, use css for the style on both of these as Alex suggested.

website/project/views/contributor.py
Johnetordoff: The idea of the ticket as I understand it was that you would import non-bibliographic (or "invisible") contributors as well as visible contributors, so importing all contributors not just visible ones was key.

brianjgeiger: Yup, but this looks like, if a contributor is in the parent and in the node, _but_ it's in the node as a non-bibliographic contributor, it will try to add it again.

brianjgeiger: Whereas you'd want to avoid re-adding whether they were bibliographic or no.

brianjgeiger: (You should probably should add a test or two for these cases, I suspect)

Johnetordoff: Right you are, I'll change that.

PR #7623, opened 2017-08-24
[EOSF-824] Allow quickfiles to create guids, and reroute based on fid without no…

hmoco:
## Ticket:
https://openscience.atlassian.net/browse/EOSF-824

## Purpose
For viewing quickfiles: Add new route, like /quickfiles/<path:path>/ that proxies to /project/<quickfiles-guid>/files/osfstorage/<path:path>/
For generating share links: Bypass admin permission check when creating GUIDs for quickfiles when using ?create_guid=1

### DIFF
--- a/api/files/views.py
+++ b/api/files/views.py
@@ -336,13 +336,12 @@ class FileDetail(JSONAPIBaseView, generics.RetrieveUpdateAPIView, FileMixin):
     # overrides RetrieveAPIView
     def get_object(self):
         user = utils.get_user_auth(self.request).user
-
-        if (self.request.GET.get('create_guid', False) and
-                self.get_node().has_permission(user, 'admin') and
-                utils.has_admin_scope(self.request)):
-            self.get_file(check_permissions=True).get_guid(create=True)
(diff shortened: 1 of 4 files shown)


api/files/views.py
```diff
@@ -336,13 +336,12 @@ def get_node(self):
...
     def get_object(self):
         user = utils.get_user_auth(self.request).user
-
-        if (self.request.GET.get('create_guid', False) and
-                self.get_node().has_permission(user, 'admin') and
-                utils.has_admin_scope(self.request)):
-            self.get_file(check_permissions=True).get_guid(create=True)
-
-        return self.get_file()
+        file = self.get_file()
```
alexschiller: very, very minor driveby: `file` is a `type` in python (which is why it gets highlighted funny like that). Better to use a different name (although it is used all over in the OSF so not a big deal if you aren't making other changes).

sloria:
@hmoco Ticket for views test: https://openscience.atlassian.net/browse/EOSF-835

PR #7627, opened 2017-08-28
[Feature] celery for registrations [OSF-7805]

fabmiz:
## Purpose

- Prevent frontend timeouts (504s) when creating a registration for complex and/or large projects
- Offload all of the pre-archive registration steps to celery, return almost immediately.

Registastion flow: 1) user creates draft, 2) user registers draft, 3) pre-archive, 4) archiving

## Changes

- Create a new nodetask to handle pre-archive step of registration.
- Update RegistrationFactory and tests that creates registration in different ways.

## Side effects

- Tests involving registration will take longer as a draft registration is required when using `.register_node` or `mock_archive`. A draft registration is optional when using `RegistrationFactory`, the factory creates if `None`.

- Some UI change related to pre-archive step may now be needed as this step may take long depending on the complexity or size of the registration.

- All registrations have to go through celery regardless of size or complexity.

## Ticket

[OSF-7805](https://openscience.atlassian.net/browse/OSF-7805)


tests/base.py
```diff
@@ -109,6 +109,7 @@ class AppTestCase(unittest.TestCase):
 
     def setUp(self):
         super(AppTestCase, self).setUp()
+        test_app.config['PRESERVE_CONTEXT_ON_EXCEPTION'] = False
```
brianjgeiger: What does this do in context of the PR?

osf_tests/test_registrations.py
```diff
@@ -228,7 +228,7 @@ def test_is_registration(self, registration):
         assert registration.is_registration
 
     def test_registered_date(self, registration):
-        assert_datetime_equal(registration.registered_date, timezone.now(), allowance=3000)
+        assert_datetime_equal(registration.registered_date, timezone.now(), allowance=6000)
```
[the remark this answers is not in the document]
fabmiz: That's what I thought initially but the new step (creating a draft registration) in the RegistrationFactory happens before `.registered_date` is assigned a value, hence should not cause additional delays. I left it back to initial allowance, waiting to see on Travis.

tests/base.py
fabmiz: I ran into issues, where flask was raising two exceptions hence confusing the request context. The fix is suggested here [assertionerror-popped-wrong-request-context](https://stackoverflow.com/questions/26647032/py-test-to-test-flask-register-assertionerror-popped-wrong-request-context/28139033#28139033). Travis log for the error [here](https://travis-ci.org/fabmiz/osf.io/jobs/272465037#L2309). Will investigate further.

brianjgeiger: As long as it doesn't cause problems for other tests. If it's possible to put that into the test that was having problems, rather than in the setup, that would be better, but I can see how it might be too late then.

brianjgeiger:
When you have a moment, could you merge with develop?

sloria:
Closing for now, as this has become stale.

PR #7628, opened 2017-08-28
[#OSF-8154]Institutions pages: registrations missing contrib names

chennan47:
## Purpose
Fix the problem that if you have more than 2 contributors for an institution registration. The contributors will show as "+NaN" instead of the correct number.

## Changes
before change:

image

After change:
The NaN should be replaced with the correct number.

## Ticket
https://openscience.atlassian.net/browse/OSF-8154


api/registrations/views.py
```diff
@@ -380,6 +381,8 @@ class RegistrationContributorsList(BaseContributorList, RegistrationMixin, UserM
     """
     view_category = 'registrations'
     view_name = 'registration-contributors'
+
+    pagination_class = NodeContributorPagination
```
alexschiller: This change looks like it affects the registration endpoint more generally, but the test you added only is for institution registrations. Can you add a comparable one to `test_registration_list` please?

PR #7633, opened 2017-08-29
[hotfix] Fix error handling and mako template data [OSF-8570]

alexschiller:
## Purpose
Files that are not found lead to unable to resolve instead of nicely formatted not found pages (and are leading to a lot of sentry errors).

## Changes
1. Fix mako template data to handle data that was not being formatted properly.
2. Use correct exception catch for oststorage files

## Side effects
NA

## Ticket

https://openscience.atlassian.net/browse/OSF-8570

## QA Notes
1. Test non existent files on various addons (e.g., osf.io/<guid>/files/github/imnotarealfile/)
2. Try deleted addon files via guid
3. Try osfstorage files that don't exist (e.g., osf.io/<guid>/files/imnotarealfile)

In all the above cases, you should get a properly formatted error message and not an "unable to resolve" error

### DIFF
--- a/addons/base/views.py
+++ b/addons/base/views.py
@@ -18,6 +18,8 @@ from modularodm import Q
 from modularodm.exceptions import NoResultsFound
 
 from addons.base.models import BaseStorageAddon
+from addons.osfstorage.models import OsfStorageFileNode
+
 from framework import sentry
 from framework.auth import Auth
 from framework.auth import cas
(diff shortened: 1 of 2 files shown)


brianjgeiger:
Based on this fix, I suspect https://github.com/alexschiller/osf.io/blob/6dab23ed6da5c91927598b1f239f9d35b99e03ed/addons/base/views.py#L471 is a problem

alexschiller:
@brianjgeiger Yes, I discussed this with DjangoLE people yesterday and I decided to only modify the one location that was relevant to this issue (since it is a hotfix) they are aware that parts of that file had not yet been converted and I believe there is a ticket for them.

PR #7634, opened 2017-08-29
Convert addons base to django 

erinspace:
## Purpose

Somehow missed some removals in #7529, here to fix that

## Changes

- Remove Q queries
- replace MODM error checking

## Side effects
none anticipated

### DIFF
--- a/addons/base/views.py
+++ b/addons/base/views.py
@@ -14,11 +14,10 @@ import jwe
 import jwt
 from django.db import transaction
 
-from modularodm import Q
-from modularodm.exceptions import NoResultsFound
-
 from addons.base.models import BaseStorageAddon
 from addons.osfstorage.models import OsfStorageFile
+from addons.osfstorage.models import OsfStorageFileNode
+
 from framework import sentry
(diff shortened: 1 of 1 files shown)


addons/base/views.py
```diff
@@ -199,21 +198,18 @@ def check_access(node, auth, action, cas_resp):
...
-            Q('name', 'eq', 'Prereg Challenge') &
-            Q('schema_version', 'eq', 2)
-        )
+        prereg_schema = MetaSchema.objects.get(name='Prereg Challenge', schema_version=2)
         allowed_nodes = [node] + node.parents
-        prereg_draft_registration = DraftRegistration.find(
-            Q('branched_from', 'in', [n for n in allowed_nodes]) &
-            Q('registration_schema', 'eq', prereg_schema)
+        prereg_draft_registration = DraftRegistration.objects.filter(
+            branched_from__in=[n for n in allowed_nodes],
```
caseyrollins: I think this can be `branched_from__in=allowed_nodes`?

PR #7635, opened 2017-08-29
[OSF-6247] Admins cannot see DraftRegistrations canceled by users.

adlius:
## Purpose

Per @alexschiller 's request, this is an alternative to [this PR](https://github.com/CenterForOpenScience/osf.io/pull/7583) as a solution with much smaller footprint.

## Changes

Change `is_approved` so that it would not return `False` for `DraftRegistration` instances that has `registered_node` defined. Also changed the query to return the correct drafts. Some other changes to make sure doesn't crash.

## Side effects

Nothing I can think of.

## Ticket

https://openscience.atlassian.net/browse/OSF-6247


admin/pre_reg/views.py
```diff
@@ -49,7 +49,9 @@ class DraftListView(PermissionRequiredMixin, ListView):
     def get_queryset(self):
         return DraftRegistration.objects.filter(
             registration_schema=get_prereg_schema(),
-            approval__isnull=False
+        ).exclude(
+            approval__isnull=True,
```
alexschiller: Isn't exclude(approval__isnull=True) the same as filter(approval__isnull=False)? 
I think just adding the exclude(registered_node__isnull=True) is sufficient.

admin/pre_reg/serializers.py
```diff
@@ -66,12 +66,14 @@ def get_url(draft):
...
-    return IMMEDIATE
-
+    if draft.approval is not None:
+        registration_choice = draft.approval.meta.get('registration_choice', None)
+        if registration_choice == EMBARGO:
+            time = parser.parse(draft.approval.meta['embargo_end_date'])
+            return iso8601format(time) if json_safe else time
+        return IMMEDIATE
+    else:
+        return None
```
alexschiller: Does `None` serialize fine? Otherwise you might be explicit and put something like "Canceled Registration" to keep the `type` the same

admin/pre_reg/serializers.py
```diff
@@ -33,7 +33,7 @@ def serialize_draft_registration(draft, json_safe=True):
         'registration_schema': serialize_meta_schema(draft.registration_schema),
         'initiated': iso8601format(draft.datetime_initiated) if json_safe else draft.datetime_initiated,
         'updated': iso8601format(draft.datetime_updated) if json_safe else draft.datetime_updated,
-        'submitted': iso8601format(draft.approval.initiation_date) if json_safe else draft.approval.initiation_date,
+        'submitted': iso8601format(draft.approval.initiation_date) if json_safe else draft.approval.initiation_date if draft.approval is not None else None,
```
alexschiller: I don't know when json_safe is True, but if it is ever `True` and there is no `draft.approval`, this will fail. 

It reads in order:
**1** if True else 2 if False else 3 == 1
**1** if True else 2 if True else 3 == 1
1 if False else **2** if True else 3 == 2
1 if False else 2 if False else **3** == 3

admin/pre_reg/views.py
alexschiller: I think I realize what you are trying to do, but it requires changing your query syntax to something like:
```
prereg_query = Q(registration_schema=get_prereg_schema())
approved_query = Q(approval__isnull=False)
rejected_query = Q(approval__isnull=True, registered_node__isnull=False)
return DraftRegistration.objects.filter(prereg_query & (approved_query | rejected_query))

```

osf/models/registrations.py
```diff
@@ -469,7 +469,10 @@ def is_pending_review(self):
     def is_approved(self):
         if self.requires_approval:
             if not self.approval:
-                return False
+                if self.registered_node is None:
```
alexschiller: `get_approval_status` already has a check for `if draft.registered_node is not None`, but it occurs after `if draft.is_approved` which means it will never hit the 'approved but canceled', right?

If you can just rework the get_approval_status without changing how the model works, I think that would be desirable.

adlius: Um...I don't understand. If `draft.is_approved==True`, then it could mean either `draft.registered_node is not None and draft.approval is None` or `draft.approval.is_approved==True`. If it is the first situation, then it means it would go on and check whether `draft.registered_node.isdeleted==True`. If yes then `approved but canceled` would be hit.
Realistically, I could modify the `get_approval_status` method only (by not using the `is_approved` property at all) without addressing the inherent problem that the `is_approved` method reflects the wrong admin approval status of a draft registration. In that case, there could be weird things happening like `get_approval_status` returns 'Approved but canceled' while `is_approved` returns `False`......

alexschiller: Discussed briefly in person, I *think* that this change won't get hit. Either way, tests will clarify the state 👍

admin_tests/pre_reg/test_serializers.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @pytest.fixture()
+    def draft_approved_and_registered(self, user, project):
+        draft = DraftRegistrationFactory(initiator=user, registration_schema=get_prereg_schema())
+        draft.approval = DraftRegistrationApproval(
+            meta={
+                'registration_choice': 'immediate'
+            }
+        )
+        draft.approval.state = 'approved'
```
alexschiller: While these tests will pass because they all occur within a transaction, if someone reuses these fixtures and tries to filter or refresh from the database, these changes will not persist. Whenever modifying data on a model for a test it is best to `.save()` it.

Makes things more maintainable in the future if someone comes in and adds more tests.

adlius:
Additional changes are made to make tests more maintainable. Also some minor simplifications.

brianjgeiger:
@adlius Conflicts if you have time

adlius:
Conflicts fixed.

PR #7640, opened 2017-08-30
[RFR][OSF-8522] Special characters do not render on registration forms

TomBaxter:
## Purpose
Decode HTML in project -> "View Registration Form"

## Changes

    Registration Form data is HTML encoded before storing in postgres.
    This PR HTML decodes data before presentation. Specifically &amp; , &gt; and, &lt;

## Side effects

Best effort has been made, but this PR should be reviewed with an eye towards javascript injection.

## Ticket

[OSF-8522](https://openscience.atlassian.net/browse/OSF-8522)

### DIFF
--- a/website/static/js/registrationUtils.js
+++ b/website/static/js/registrationUtils.js
@@ -210,6 +210,9 @@ var Question = function(questionSchema, data) {
         value = self.data.value();
     } else {
         value = self.data.value || null;
+        if(value) {
+            value = $osf.decodeText(value);
+        }
     }
 
     if (self.type === 'choose' && self.format === 'multiselect') {


TomBaxter:
@alexschiller per your request https://stackoverflow.com/a/3700369
Also possibly of interest https://www.w3schools.com/JSREF/prop_textarea_value.asp

website/static/js/registrationUtils.js
```diff
@@ -210,6 +210,9 @@ var Question = function(questionSchema, data) {
         value = self.data.value();
     } else {
         value = self.data.value || null;
+        var elem = document.createElement('textarea');
```
brianjgeiger: Can you write a JS test that tries to inject a script and makes sure it doesn't work?

alexschiller: @brianjgeiger If this work as expected, is there any reason we couldn't replace the other amp cleaners and such with this method? E.g., decodeText in osfhelpers

brianjgeiger: Maybe. It would be worth exploring, for sure. I don't know if there are any other consequences to the field, but it would probably be fine.

TomBaxter: I'll give it a go, might take a while. I searched for a current test of registration form data, and couldn't find one.

brianjgeiger: I'm also, especially if we want to generalize this, open to making this a separate function that you can test independently. That's probably the easier way to go.

TomBaxter: I've added textareaElementDecode to osfHelpers as a separate function. However, I have had no luck trying to test for HTML injection. The biggest problem is I have been unable to create a negative result (injection of any kind). However, even if I were to finally contrive an injection, it could only be tested after running though the mako template. Not sure how that would be done.

website/static/js/tests/osfHelpers.test.js
```diff
@@ -264,6 +264,14 @@ describe('osfHelpers', () => {
         });
     });
 
+    describe('textareaElementDecode', () => {
+        it('returns properly decoded text', () => {
+            var text = '&amp; &lt; &gt;';
```
brianjgeiger: What happens if `text = '&amp; &lt; &gt; & <>';`?

TomBaxter: Updated to...
```
      describe('textareaElementDecode', () => {
          it('returns properly decoded text', () => {
~             var text = '&amp; &lt; &gt; & < >';
              var ret = $osf.textareaElementDecode(text);
~             assert.equal(ret, '& < > & < >');
          });
      });
```
Which passes

TomBaxter:
@brianjgeiger Per our conversation, removed new function and test for function. Now decoding &amp; with existing function.

TomBaxter:
@brianjgeiger  Does this need something? Sorry for bugging you, just going through my old PRs.

PR #7641, opened 2017-08-30
[Hotfix] Update BePress version [OSF-8334][OSF-8325]

mfraezz:
## Purpose
Keep our internal taxonomy up-to-date with BePress

## Changes
* Add migration

## Side effects
None expected

## Ticket
[[OSF-8334]](https://openscience.atlassian.net/browse/OSF-8334)
[[OSF-8325]](https://openscience.atlassian.net/browse/OSF-8325)

### DIFF
--- a/osf/management/commands/update_bepress_version.py
+++ b/osf/management/commands/update_bepress_version.py
@@ -1,0 +1,88 @@
+from __future__ import unicode_literals
+import logging
+
+from django.core.management.base import BaseCommand
+from django.db import transaction
+
+from scripts import utils as script_utils
+from osf.models import PreprintProvider, Subject
+from website.preprints.tasks import on_preprint_updated
+
+logger = logging.getLogger(__name__)
+
+BEPRESS_PROVIDER = PreprintProvider.objects.filter(_id='osf').first()
+BEPRESS_CHANGES = {
+    'create': [
+        'Education:Adult and Continuing Education',
+        'Education:Disability and Equity in Education:Gender Equity in Education',
+        'Education:Early Childhood Education',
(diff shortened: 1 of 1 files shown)


osf/management/commands/update_bepress_version.py
```diff
@@ -0,0 +1,88 @@
...
+}
+
+def update(dry_run=False):
+    for hier in BEPRESS_CHANGES['create']:
+        new_text = hier.split(':')[-1]
+        bepress_parent = BEPRESS_PROVIDER.subjects.get(text=hier.split(':')[-2])
+        logger.info('Creating osf - {}'.format(new_text))
+        bepress_subject = Subject.objects.create(parent=bepress_parent, provider=BEPRESS_PROVIDER, text=new_text)
+        for custom_parent in bepress_parent.aliases.all():
+            if custom_parent.children.exists() or not bepress_parent.children.count() > 1:
```
sloria: I'm not sure I understand the rationale for `or not bepress_parent.children.count() > 1`. If the bepress subject has 1 child, then the custom parent gets the new subjects. If the bepress subject has 2 children, then the custom parent would not get the new subjects. Is this intended?

mfraezz: Yes. The `bepress_parent` has a minimum of 1 child at this point, because the `bepress_subject`, which is the child, was just created. If there are 2, it existed before, but was not included in the `custom_parent`'s children when the custom taxonomy was populated.

sloria: OK, thanks for the clarification.

PR #7647, opened 2017-09-01
[OSF-8322] Move sharejs into its own Dockerfile

binoculars:
## Purpose

ShareJS is broken due to node.js upgrade

## Changes

* Created Dockerfile in the addons/wiki directory
* Moved npm dependencies to a separate package.json
* Changed the docker-compose and override files to reflect the new Dockerfile changes

## Notes for QA
The wiki should work again as normal with live editing

## Notes for deployment

We'll have to build and run sharejs from the addon directory, instead of running it from the main OSF image using invoke

## Ticket

https://openscience.atlassian.net/browse/OSF-8322


mattclark:
@icereval, yes I'll coordinate with @sloria and @binoculars

PR #7648, opened 2017-09-01
[feature] Backwards-compatible API changes for preprint/node divorce [OSF-8488]

erinspace:
## Purpose
We soon want to move away from the node - preprint entanglement, now that using django and postgres has made that possibility way more convenient. This is the first step in that process, that will "fake" the change first in the API, now making it possible to `POST`  to `/v2/preprints/` and `PATCH`es to `/v2/preprints/<preprint_id>/ `to create and update *all* formerly node related attributes including `title`, `description`, and `contributors`.

**Current workflow (assuming creating a new node)**
1. `POST` to `/v2/nodes/` to create the node
2. Upload the file to the node's wb upload link via `PUT`
3. `POST` to `/v2/preprints/` to create the preprint with provider and relationship to node
4. `PATCH` to `/v2/preprints/<prerprint_id>/` to add subjects to the preprint
5. `PATCH` to `/v2/nodes/<node_id>/` to add description and license to node
6. `PATCH` to `v2/preprints/<preprint_id>/` to also add license to preprint
7. `POST` to `/v2/nodes/<node_id>/contributors/` to add each contributor 
8. `PATCH` `/v2/preprints/<preprint_id>/` to publish the preprint
9. `PATCH` `/v2/nodes/<node_id>/` to make the node public

**New possible workflow**
1. `POST` to `/v2/preprints/` to create the preprint -- also creates a node if one isn't passed via id
2. Upload the file to the node's wb upload link via `PUT`
3. `PATCH` to `/v2/preprints/<preprint_id>/` to update the preprint with provider and primary file
4. `PATCH` to `/v2/preprints/<prerprint_id>/` to add subjects to the preprint
5. `PATCH` to `v2/preprints/<preprint_id>/` to also add license and description to preprint
6. `PATCH` to `/v2/nodes/<node_id>/` to add license to the node if checkbox is there (can't avoid this separate call to the node because of current frontend implementation)
7. `POST` to `/v2/preprints/<preprint_id>/contributors/` to add each contributor 
8. `PATCH` `/v2/preprints/<preprint_id>/` to publish the preprint (which makes the node public)

## Changes

- Update the preprint serializer to accommodate creating a preprint by `POST`
    - this will create a node if a node isn't passed
    - move setting of the primary file to the `update` method
        - now check for primary file on publish instead of in the preprint `create` method
    - update node properties in the preprint `update` method
    - set node privacy to `public` on publishing a preprint to avoid needing the extra API call
- Add preprint contributors view to handle updating contributors by `POST` to a preprints endpoint
- Add tests for creating a preprint by `POST` to `/v2/preprints/`
- Add tests for updating `title`, `description` and `contributors` by `PATCH` to `/v2/preprints/<preprint_id>/`

## Side effects
None anticipated, as this should be 100% backwards compatible with the way that the preprints ember app currently functions. It just opens the way for a new workflow!

## Ticket
https://openscience.atlassian.net/browse/OSF-8488

### DIFF
--- a/api/preprints/serializers.py
+++ b/api/preprints/serializers.py
@@ -74,12 +74,12 @@ class PreprintSerializer(JSONAPISerializer):
     is_published = ser.BooleanField(required=False)
     is_preprint_orphan = ser.BooleanField(read_only=True)
     license_record = NodeLicenseSerializer(required=False, source='license')
-    title = ser.CharField(source='node.title', read_only=True)
+    title = ser.CharField(source='node.title', required=False)
+    description = ser.CharField(required=False, allow_blank=True, allow_null=True, source='node.description')
 
     contributors = RelationshipField(
         related_view='nodes:node-contributors',
         related_view_kwargs={'node_id': '<node._id>'},
-        read_only=True
     )
 
     citation = RelationshipField(
@@ -173,6 +173,10 @@ class PreprintSerializer(JSONAPISerializer):
             self.set_field(preprint.set_primary_file, primary_file, auth)
             save_node = True
 
+        if 'node' in validated_data:
+            preprint.node.update(fields=validated_data.pop('node'))
+            save_node = True
+
         if 'subjects' in validated_data:
             subjects = validated_data.pop('subjects', None)
             self.set_field(preprint.set_subjects, subjects, auth)
@@ -189,9 +193,13 @@ class PreprintSerializer(JSONAPISerializer):
 
         published = validated_data.pop('is_published', None)
         if published is not None:
+            if not preprint.primary_file:
+                raise exceptions.ValidationError(detail='A valid primary_file must be set before publishing a preprint.')
             self.set_field(preprint.set_published, published, auth)
             save_preprint = True
             recently_published = published
+            preprint.node.set_privacy('public')
+            save_node = True
 
         if save_node:
             try:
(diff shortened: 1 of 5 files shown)


api/preprints/serializers.py
```diff
@@ -227,20 +235,17 @@ class PreprintCreateSerializer(PreprintSerializer):
     id = IDField(source='_id', required=False, allow_null=True)
 
     def create(self, validated_data):
-        node = validated_data.pop('node', None)
-        if not node:
-            raise exceptions.NotFound('Unable to find Node with specified id.')
-        elif node.is_deleted:
+        node = validated_data.pop('node', {})
+        if isinstance(node, dict):
+            node = Node.objects.create(creator=self.context['request'].user, **node)
```
mfraezz: This results in a blank title in the default case. Is there any way to avoid that?

api/preprints/serializers.py
```diff
@@ -250,7 +255,6 @@ def create(self, validated_data):
             raise Conflict('Only one preprint per provider can be submitted for a node. Check `meta[existing_resource_id]`.', meta={'existing_resource_id': node_preprints.first()._id})
 
         preprint = PreprintService(node=node, provider=provider)
-        self.set_field(preprint.set_primary_file, primary_file, auth)
```
mfraezz: Doesn't this removal break backwards-compatibility the front-end flow, which currently passes a file with the preprint `POST`?

erinspace: Nope! The primary preprint file is instead set in the `update()` call -- which happens at the end of the create method. Tested locally on current frontend and works ok. The trade off is that the check for preprint file has to happen on publish instead of `create()`

api/preprints/serializers.py
erinspace: I think it's ok because the node is later updated in the preprint update method [here](https://github.com/CenterForOpenScience/osf.io/pull/7648/files#diff-2d550f722c7b3838d9064d9273e9dba9R177) -- including the title passed by the API POST calls to node in the frontend. Unless I'm misunderstanding what you mean by the default case.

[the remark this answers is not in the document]
erinspace: OSF Storage and Wiki seem to be correctly added when created this way - local preprint node created with "new" workflow is fully in tact!

api_tests/preprints/views/test_preprint_list.py
```diff
@@ -51,6 +51,67 @@ def build_preprint_create_payload(node_id=None, provider_id=None, file_id=None,
...
+        return '/{}preprints/'.format(API_BASE)
+
+    @pytest.fixture()
+    def preprint_payload(self, provider):
+        return {
+            'data': {
+                'type': 'preprints',
+                'attributes': {
+                    'title': 'Greatest Wrestlemania Moment Vol IX',
+                    'description': 'Crush VS Doink the Clown in an epic battle during WrestleMania IX',
```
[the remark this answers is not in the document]
erinspace: doinks

api/preprints/serializers.py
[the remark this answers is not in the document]
sloria: Ah, I see, because the node is saved ~L260.

PR #7649, opened 2017-09-01
[OSF-8319] Add logs to admin app on node detail page

Johnetordoff:
## Purpose

Allows admins to view the logs of private projects.

## Changes

Simple changes to the admin node.html page and and serializers.py to expose log info.

## Side effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/OSF-8319

## Screenshot
new_admin_log

### DIFF
--- a/admin/nodes/views.py
+++ b/admin/nodes/views.py
@@ -25,7 +25,7 @@ from osf.models.admin_log_entry import (
     REINDEX_ELASTIC,
 )
 from admin.nodes.templatetags.node_extras import reverse_node
-from admin.nodes.serializers import serialize_node, serialize_simple_user_and_node_permissions
+from admin.nodes.serializers import serialize_node, serialize_simple_user_and_node_permissions, serialize_log
 from website.project.tasks import update_node_share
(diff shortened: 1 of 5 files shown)


Johnetordoff:
@brianjgeiger I've added pagination as you can see in the gif above, do to the changes I'm going to demo this again. I don't expect many changes as this version is better, but I've changed the look a lot so I need to keep them updated.

admin/nodes/views.py
```diff
@@ -202,6 +202,35 @@ def get_object(self, queryset=None):
...
+
+    def get_object(self, queryset=None):
+        return Node.load(self.kwargs.get('guid')) or Registration.load(self.kwargs.get('guid'))
+
+    def get_queryset(self):
+        return self.get_object().logs.all().order_by(self.ordering)
+
+    def get_context_data(self, **kwargs):
+        query_set = self.get_queryset()
+        print query_set
```
brianjgeiger: Probably don't need this any more

Johnetordoff:
This was verified with product, should be gtg.

PR #7655, opened 2017-09-06
[OSF-8516] Add a modal for deleting projects with components.

adlius:
## Purpose

This PR aims to improve the deletion process of projects with components by adding a helper modal to the front-end, enabling site users to choose delete all the components when deleting a project.

## Changes

There are two parts to this PR. 

The first part is a fix for the bulk deletion API. As of now, bulk deletion of a project and its components would fail even if all the components of the project are specified in postorder in the XHR request. The reason is that a database query to get all node instances does not preserve such order, resulting in error upon deleting nodes with undeleted child nodes.

Changes made:
1. In `api/base/generic_bulk_views.py`, `resource_object_list` is sorted according to the order specified in the request before returning.

The second part is a series of front end assets addition/modification for implementing the helper modal.

1. Added `nodesDelete.js` as a controller for the modal. The KO viewmodel is also defined in this file.
2. Added `nodes_delete.mako` as the template for the modal.
3. Added `NodesDeleteTreebeard.js`, subclassing Treebeard to show the project hierarchy and handles user selection.
4. `project-settings-page.js` is modified to initialize the modal when the current node has a child.
5. `projectSettings.js` is modified to import `nodesDelete.js`.

## Side effects

For the first part, sorting the list after the query may affect performance of the bulk deletion API. However the there seems to be no way to make the query preserve the order. Some thoughts on this are very much welcomed.

## Ticket

https://openscience.atlassian.net/browse/OSF-8516

### DIFF
--- a/api/base/generic_bulk_views.py
+++ b/api/base/generic_bulk_views.py
@@ -76,10 +76,14 @@ class BulkDestroyJSONAPIView(bulk_generics.BulkDestroyAPIView):
         Retrieves resources in request body
         """
         model_cls = request.parser_context['view'].model_class
+
         requested_ids = [data['id'] for data in request_data]
         column_name = 'guids___id' if issubclass(model_cls, GuidMixin) else '_id'
         resource_object_list = model_cls.find(Q(column_name, 'in', requested_ids))
-
+        if column_name == 'guids___id':
+            resource_object_list = [resource_object_list.get(guids___id=id) for id in requested_ids]
+        else:
+            resource_object_list = [resource_object_list.get(_id=id) for id in requested_ids]
         for resource in resource_object_list:
             if getattr(resource, 'is_deleted', None):
                 raise Gone
--- a/api/base/generic_bulk_views.py
+++ b/api/base/generic_bulk_views.py
@@ -80,16 +80,18 @@ class BulkDestroyJSONAPIView(bulk_generics.BulkDestroyAPIView):
         requested_ids = [data['id'] for data in request_data]
         column_name = 'guids___id' if issubclass(model_cls, GuidMixin) else '_id'
         resource_object_list = model_cls.find(Q(column_name, 'in', requested_ids))
-        if column_name == 'guids___id':
-            resource_object_list = [resource_object_list.get(guids___id=id) for id in requested_ids]
-        else:
-            resource_object_list = [resource_object_list.get(_id=id) for id in requested_ids]
+
         for resource in resource_object_list:
             if getattr(resource, 'is_deleted', None):
                 raise Gone
 
         if len(resource_object_list) != len(request_data):
             raise ValidationError({'non_field_errors': 'Could not find all objects to delete.'})
+
+        if column_name == 'guids___id':
+            resource_object_list = [resource_object_list.get(guids___id=id) for id in requested_ids]
+        else:
+            resource_object_list = [resource_object_list.get(_id=id) for id in requested_ids]
 
         return resource_object_list
 
--- a/api_tests/nodes/views/test_node_list.py
+++ b/api_tests/nodes/views/test_node_list.py
@@ -2259,7 +2259,7 @@ class TestNodeBulkDelete:
         assert res.status_code == 400
 
         new_payload = {'data': [{'id': public_component._id, 'type': 'nodes'}, {'id': public_project_parent._id, 'type': 'nodes'}]}
-        res = app.delete_json_api(url, new_payload, auth=user_one.auth, expect_errors=True, bulk=True)
+        res = app.delete_json_api(url, new_payload, auth=user_one.auth, bulk=True)
         assert res.status_code == 204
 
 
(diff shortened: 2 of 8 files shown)


website/static/js/nodesDelete.js
```diff
@@ -0,0 +1,275 @@
...
+ * uses API v2 bulk requests
+ */
+function patchNodesDelete(nodes) {
+    var nodesV2Url = window.contextVars.apiV2Prefix + 'nodes/';
+    var nodesPatch = $.map(nodes, function (node) {
+        return {
+            'type': 'nodes',
+            'id': node.id,
+            'attributes': {
+                'public': node.public
```
brianjgeiger: So I'm guessing you copied this from the bulk privacy changes. A delete request doesn't need this hefty of a payload.

website/static/js/nodesDelete.js
```diff
@@ -0,0 +1,275 @@
...
+        data: JSON.stringify({
+            data: nodesPatch
+        })
+    });
+}
+
+/**
+ * view model which corresponds to nodes_delete.mako (#nodesDelete)
+ *
+ * @type {NodesPrivacyViewModel}
```
brianjgeiger: Change the name here

website/static/js/nodesDelete.js
```diff
@@ -0,0 +1,275 @@
...
+    });
+};
+
+NodesDeleteViewModel.prototype.selectProjects = function() {
+    this.page(this.SELECT);
+};
+
+NodesDeleteViewModel.prototype.confirmWarning =  function() {
+    var nodesState = ko.toJS(this.nodesState);
+    for (var node in nodesState) {
```
brianjgeiger: This looks like a feature that was in the "change to public" that isn't really needed here. If they are using the delete, they will need a confirmation modal, whereas in the original they could try to change everything to public, but if everything was already public, then it wouldn't need a confirmation modal.

website/static/js/nodesDelete.js
adlius: Right.... I simply changed the method to DELETE without checking payload.

website/static/js/nodesDelete.js
adlius: The reason I copied and pasted from nodesPrivacy.js is that the feature specified in the ticket looks a lot like changing nodes privacy in that users have to manually choose and check all the project components shown in the project hierarchy to be able to proceed to next page for confirmation.

adlius: I am not sure I understand. confirmWarning() changes the modal to a warning page where the titles of the project and its components are listed for user reference and asks for further confirmation. This is a specific feature request in the ticket.

brianjgeiger: Gotcha. This one's good; I was thinking it was checking something else.

api/base/generic_bulk_views.py
```diff
@@ -87,6 +88,11 @@ def get_requested_resources(self, request, request_data):
         if len(resource_object_list) != len(request_data):
             raise ValidationError({'non_field_errors': 'Could not find all objects to delete.'})
 
+        if column_name == 'guids___id':
```
brianjgeiger: Since you've made a change to our one of our base views, there should definitely be tests related to the change.

adlius:
Added a test to `test_node_list.py` to make sure bulk deletion for nodes works.
In our codebase, `BulkDestroyJSONAPIView` is subclassed by `NodeList`, `NodeContributorList`, `NodeLinkList` and `CollectionList`. All of these classes have respective tests to make sure bulk deletion works. However, I can't think of any way to test the base class other than testing their subclasses because of the way it is implemented.
Or, if changing the base class should be avoided, we could override the method and make changes only in `NodeList` class.

api_tests/nodes/views/test_node_list.py
```diff
@@ -2245,6 +2253,15 @@ def test_bulk_delete_invalid_payload_one_not_found(self, app, user_one, public_p
         res = app.get(public_project_one_url, auth=user_one.auth)
         assert res.status_code == 200
 
+    def test_bulk_delete_project_with_component(self, app, user_one, public_project_parent, public_component, url):
+        new_payload = {'data': [{'id': public_project_parent._id, 'type': 'nodes'}, {'id': public_component._id, 'type': 'nodes'}]}
+        res = app.delete_json_api(url, new_payload, auth=user_one.auth, expect_errors=True, bulk=True)
+        assert res.status_code == 400
+
+        new_payload = {'data': [{'id': public_component._id, 'type': 'nodes'}, {'id': public_project_parent._id, 'type': 'nodes'}]}
+        res = app.delete_json_api(url, new_payload, auth=user_one.auth, expect_errors=True, bulk=True)
```
brianjgeiger: This looks like it should not expect errors, true?

brianjgeiger:
The changes you made were because you needed to get the results in a particular order, yes? Can you make sure that the new functionality is tested? The existing tests sound like you've verified that you haven't broken anything. The new test you added in https://github.com/CenterForOpenScience/osf.io/pull/7655/commits/7f030827f5618cbf36b31ac01393033a525c805f looks like it handles the base case properly and also shows what happens if they're out of order, so I think that should be fine.

api_tests/nodes/views/test_node_list.py
brianjgeiger: In case I worded that poorly: because this request expects a success, `expect_errors` should probably be set to `False`.

adlius: Ah, right. Will change that.

PR #7664, opened 2017-09-07
[OSF-8358] Ensure admin user search allows spaces, commas and periods.

Johnetordoff:
## Purpose

Currently using periods or commas breaks admin user name search. This fixes that.

## Changes

Added punctuation to the url definition regex. Added a test for that.

## Side effects

None that I know of.

## Test 

Added a unit test.

## Ticket

https://openscience.atlassian.net/browse/OSF-8358


admin_tests/users/test_views.py
```diff
@@ -629,6 +629,16 @@ def test_search_user_by_name(self):
         nt.assert_equal(response.status_code, 302)
         nt.assert_equal(self.view.success_url, '/users/search/Hardy/')
 
+    def test_search_user_by_name_with_punction(self):
+        form_data = {
+            'name': 'Dr. Sportello'
```
brianjgeiger: You should go ahead and put a comma in here as well. 'Dr. Sportello, DDS.' for example.

brianjgeiger:
Does this successfully return appropriate users?

Johnetordoff:
I've tested it in the browser with punctuation  and it returns the correct users.

PR #7667, opened 2017-09-07
[#OSF-7877]Can't disable user in admin on staging

chennan47:
## Purpose
Fix the disable user funtion Api key error.

## Side effects
The ticket is not straight forwards enough. The sentry link provided by the ticket leads to project not found.
screen shot 2017-09-07 at 1 26 49 pm
Based on the information given inside the ticket, I did this fix.

## Ticket
https://openscience.atlassian.net/browse/OSF-7877


brianjgeiger:
Okay, so: this is apparently not really a problem. The sentry error was for staging3 and the ticket said staging, so I checked with QA and staging can totally disable users. So I'm closing this PR and the related ticket.

PR #7668, opened 2017-09-07
Add Quick Files to Profile Page [OSF-8369]

erinspace:
## Purpose
The user’s profile page should display the files the user has uploaded to their public quickfiles project in additional widget along with the public project and public components contributed to by the user.

The widget should only be displayed if the user has uploaded at least one quick file.

A slightly smaller screen:
screen shot 2017-12-01 at 1 06 40 pm

On mobile:

screen shot 2017-12-01 at 1 09 32 pm

## Changes
- Add QuickFiles column to display files that have been added to a user's Quick Files' project
- Add tests to make sure the widget is present, and that it's hidden if a user does not have any files uploaded
- Refactor `PublicNodes.js` to pull out pagination into a separate shared JS module to also be used in `QuickFiles.js`

## Side effects

As it stands, files linked in this widget will lead to the generic file detail page, which will have many things that shouldn't be there (check out button, link to Quick Files node, etc) -- this will be taken care of once the ember side is implemented and QuickFIles files are displayed by an Ember file detail page.

## Ticket
https://openscience.atlassian.net/browse/OSF-8369


sloria:
@erinspace Nici mentioned on the ticket that we should collect analytics for this feature. Would you mind adding that? See the Dashboard and My Projects page for examples on how to send custom events to Google Analytics.

website/templates/profile.mako
```diff
@@ -98,25 +98,52 @@
...
                 </div>
             </div>
         </div>
     </div>
+    % if user['has_quickfiles']:
     <div class="col-sm-6">
         <div class="panel panel-default">
             <div class="panel-heading clearfix">
-                <h3 class="panel-title">Public components</h3>
+                <h3 class="panel-title">Quick Files</h3>
```
sloria: As per osf-style, use title-casing, i.e. "Quick files".

website/static/js/components/profilePagination.js
```diff
@@ -0,0 +1,129 @@
...
+var iconmap = require('js/iconmap');
+var lodashFind = require('lodash.find');
+var mHelpers = require('js/mithrilHelpers');
+var Raven = require('raven-js');
+
+
+var MAX_PAGES_ON_PAGINATOR = 7;
+var MAX_PAGES_ON_PAGINATOR_SIDE = 5;
+
+var ProfilePagination = function(ctrl, buildUrlFunction) {
```
sloria: Is there actually anything profile-specific about this code? If not, then this file and function should be renamed.

website/static/js/components/quickFiles.js
```diff
@@ -0,0 +1,172 @@
...
+                var url = _buildUrl(page, self.user);
+                return self.getQuickFiles(url);
+            }
+        };
+        self.getCurrentQuickFiles();
+    },
+
+    view: function (ctrl) {
+
+        profilePagination.ProfilePagination(ctrl, _buildUrl);
```
sloria: +1 to making pagination re-usable. 

However, I suggest tweaking the approach a bit.

Notice that `ProfilePagination` function can't really be used on its own--it assumes a lot about what's on `ctrl` (i.e. it assumes that `ctrl` has props names `paginators`, `nextPage`, etc.).

Can you encapsulate *all* pagination-related code into one function?

I imagine an API like:

```js
var withPagination = require('js/components/pagination').withPagination;

function buildURL(...) {...}

// withPagination is a higher-order component, i.e. it takes a component as input
// and outputs a new component with added behavior
var QuickFiles = withPagination(buildURL, {
    controller: function() {
         // only quick files stuff here, no pagination stuff
     },
    view: function(ctrl) {
        // only quick files stuff here, no pagination stuff
     },
});
```

website/static/js/components/pagination.js
```diff
@@ -0,0 +1,191 @@
...
+var iconmap = require('js/iconmap');
+var lodashFind = require('lodash.find');
+var mHelpers = require('js/mithrilHelpers');
+var Raven = require('raven-js');
+
+var MAX_PAGES_ON_PAGINATOR = 7;
+var MAX_PAGES_ON_PAGINATOR_SIDE = 5;
+
+
+var WithPagination = function(options) {
```
sloria: For consistency, `withPagination` should be lowercased here.

website/static/js/components/quickFiles.js
```diff
@@ -0,0 +1,150 @@
...
+                    m('span.component-overflow.f-w-lg', {style: 'line-height: 1.5; width: 100%'}, [
+                        m('span.project-statuses-lg', {style: 'width: 50%; float:left'}, [
+                            m('span', {class: ctrl.icon, style: 'padding-right: 5px;'}, ''),
+                            m('a', {'href': viewUrl,
+                                onclick : function () {
+                                    $osf.trackClick('QuickFiles', 'view', 'view-quickfile-from-profile-page');
+                                }
+                            }, ctrl.file.attributes.name),
+                        ]),
+                        m('div', {style: 'width: 50%; float:right; font-size:small; line-height:2;'}, 'downloads: ' + ctrl.file.attributes.extra.downloads)
```
sloria: This is fine, but for future reference, you can pass `style` as an object, e.g.

```javascript
{style: {width: '50%', ..., fontSize: 'small'}}
```

website/static/js/components/quickFiles.js
```diff
@@ -0,0 +1,150 @@
...
+            }
+        };
+        self.getCurrentQuickFiles();
+    },
+
+    view: function (ctrl) {
+
+        return m('ul.list-group m-md', [
+            // Error message if the request fails
+            ctrl.failed ? m('p', [
```
sloria: `ctrl.failed` should initialized in the controller.

website/static/js/components/quickFiles.js
sloria: Instead of making `font-size: small`, does it make sense to use a `<small>` tag?

Also, why is `line-height` being modified? There isn't a lot of text here, so it seems a little strange to change the spacing between lines. Would `margin` make sense?

erinspace: I hadn't really touched the meat of these specific styles and divs and whatnot as this was just moving existing things around, but I'll check on that!

sloria: Oh ok, I didn't realize this code came from somewhere else. In that case, don't worry about this for now.

erinspace: Oh man I was so wrong about that -- it did not come from anywhere else! I'll fix it!

website/static/js/components/quickFiles.js
erinspace: good call! initialized in controller for both this and public nodes!

website/static/js/components/quickFiles.js
```diff
@@ -0,0 +1,151 @@
...
+                    m('span.component-overflow.f-w-lg', {style: 'line-height: 1.5; width: 100%'}, [
+                        m('span.project-statuses-lg', {style: 'width: 50%; float:left'}, [
+                            m('span', {class: ctrl.icon, style: 'padding-right: 5px;'}, ''),
+                            m('a', {'href': viewUrl,
+                                onclick : function () {
+                                    $osf.trackClick('QuickFiles', 'view', 'view-quickfile-from-profile-page');
+                                }
+                            }, ctrl.file.attributes.name),
+                        ]),
+                        m('div', {style: {width: '50%', float: 'right', fontSize: 'small', marginTop: '2%'}}, 'downloads: ' + ctrl.file.attributes.extra.downloads)
```
sloria: Drive-by comment: Might want to use an object consistently above.

erinspace: normalized!

sloria:
Just a couple more items: 

- Responsiveness check: do the widget and its panel respond on different device sizes?
- Should "downloads" be capitalized? I believe osf-style prescribes sentence case

PR #7669, opened 2017-09-07
[#OSF-8608]Templating nodes that are linked to each other hit max recursion

chennan47:
## Purpose
Fix the problem that templating nodes that are linked to each other hit max recursion.

## Changes
Now linked nodes will not be inherited to the templated node

## Ticket
https://openscience.atlassian.net/browse/OSF-8608


osf/models/node.py
```diff
@@ -1984,12 +1984,22 @@ def use_as_template(self, auth, changes=None, top_level=True):
         )
         new.save()
         # deal with the children of the node, if any
-        for node_relation in self.node_relations.select_related('child').filter(child__is_deleted=False):
-            child = node_relation.child
-            if child.can_view(auth):
-                templated_child = child.use_as_template(auth, changes, top_level=False)
-                NodeRelation.objects.get_or_create(parent=new, child=templated_child,
-                                                   is_node_link=node_relation.is_node_link)
+        for node_relation in self.node_relations.filter(child__is_deleted=False):
```
alexschiller: Is there a reason the `select_related` was omitted?

osf/models/node.py
```diff
@@ -1984,12 +1984,22 @@ def use_as_template(self, auth, changes=None, top_level=True):
...
+            node_contained = node_relation.child
+            # tempalte child nodes
```
alexschiller: comment typo `tempalte`

osf/models/node.py
```diff
@@ -1984,12 +1984,22 @@ def use_as_template(self, auth, changes=None, top_level=True):
...
+            if not node_relation.is_node_link:
+                try:  # Catch the potential PermissionsError above
+                    templated_child = node_contained.use_as_template(auth, changes, top_level=False)
+
+                except PermissionsError:
+                    pass  # If this exception is thrown omit the node from the result set
```
alexschiller: Since you set templated_child to None, the `pass` is unneeded

osf/models/node.py
```diff
@@ -1984,12 +1984,22 @@ def use_as_template(self, auth, changes=None, top_level=True):
...
+                    templated_child = None
+                if templated_child is not None:
```
alexschiller: If the only way to hit `templated_child is None` is through a PermissionError, you might consider using try-except-else instead

alexschiller: Actually, this can be a pass and no change to templated_child, if you use the except clause, either way works fine though.,

osf_tests/test_node.py
```diff
@@ -3542,11 +3542,11 @@ def test_complex_template(self, auth, project, pointee, component, subproject):
         new = project.use_as_template(auth=auth)
 
         assert new.title == self._default_title(project)
-        assert len(list(new.nodes)) == len(list(project.nodes))
+        assert len(list(new.nodes)) == len(list(project.nodes)) - 1
```
alexschiller: The changes you are making to this test implicitly exclude linked nodes in a way that is not obvious. 
I'd prefer if this was broken into two tests:
1. remove the pointee from `test_complex_template` and remove the special casing.
2. Add a new test with just a pointee and show that it does not get templated. 

I think that will be much simpler than arbitrarily `-1` and removing linked nodes from the data checks
If you prefer it as it is, adding notes for the special casing would work as well.

tests/test_views.py
```diff
@@ -2810,6 +2810,18 @@ def test_get_pointed_private(self):
         assert_equal(pointed[0]['title'], 'Private Component')
         assert_equal(pointed[0]['authorShort'], 'Private Author(s)')
 
+    def test_can_template_project_linked_to_each_other(self):
+        project2 = ProjectFactory(creator=self.user)
+        self.project.add_pointer(project2, auth=Auth(user=self.user))
+        self.project.save()
```
alexschiller: minor: it looks like `add_pointer` calls `_add_pointers` which has a `node.save` if successful, so I think these saves might be redundant

PR #7670, opened 2017-09-07
[Feature] Fix mithril redraws issues btn logs and tb filebrowser [OSF-8587]

fabmiz:
## Purpose

To solve an issue described in https://github.com/CenterForOpenScience/osf.io/pull/7277,
logs and tb filebrowser mounts where put in same mithril redraw.

Use `m.request` with `background : true` to prevent mithril redraws upon promise return, while
keeping part of solution in https://github.com/CenterForOpenScience/osf.io/pull/7277

## Changes
- Have logFeed use with `m.request` with `background : true`
- Remove `m.startComputation/m.endComputation` pair

## Ticket
[OSF-8587](https://openscience.atlassian.net/browse/OSF-8587)

### DIFF
--- a/website/static/js/components/logFeed.js
+++ b/website/static/js/components/logFeed.js
@@ -59,7 +59,7 @@ var LogFeed = {
                 m.redraw();
             }
             self.logRequestPending(true);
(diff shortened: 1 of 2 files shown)


brianjgeiger:
@fabmiz Could you make a version of this based on Master? Steve is interested in hotfixing it.

PR #7672, opened 2017-09-07
Use es6 shim iff IE [OSF-8381]

mfraezz:
## Purpose
Updates #7643 

## Changes
* Conditionally use shim for IE, on the "My Projects" page

## Side effects
None expected

## Ticket
[[OSF-8381]](https://openscience.atlassian.net/browse/OSF-8381)

### DIFF
--- a/website/templates/base.mako
+++ b/website/templates/base.mako
@@ -279,6 +279,11 @@
 
 <%def name="includes_top()">
 
+    <!-- ES6 shim for IE support for "My Projects" page-->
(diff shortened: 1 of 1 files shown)


sloria:
the script should probably go in the `includes_top` mako def. Also , is the issue only on older versions of IE? If so, you can probably put it with the other shims in `base.mako`.

mfraezz:
> Also , is the issue only on older versions of IE?

All versions.

PR #7674, opened 2017-09-08
[OSF-8569] Mod Meta Cleanup

adlius:
## Purpose

This PR clean up some legacy code as a result of this [PR](https://github.com/CenterForOpenScience/osf.io/pull/7530).

## Changes

1. Remove view method for Dataverse widget, moved to dataverse/utils.py.
2. Remove view method for forward (redirection) widget, moved to forward/utils.py
3. Remove view method for wiki widget, moved to wiki/utils.py
3. Remove api/v1 routes for wiki, zotero and mendeley widgets.
4. Change `nodes.py` to reuse the `widget()` method for citation providers.

## QA Notes

Since this PR is a clean up, the view templates for the widgets are not modified but only how/where we pass data to those widgets. A simple pass over Dataverse, wiki, forward(redirection), zotero and mendeley to make sure they function should be sufficient.

## Side effects

None.

## Ticket

https://openscience.atlassian.net/browse/OSF-8569


adlius:
@alexschiller QA notes updated.

brianjgeiger:
@adlius Conflicts if you have some time.

adlius:
Conflicts fixed.

adlius:
@brianjgeiger @chennan47 Okay. This one is bit of a doozy. I'll try to explain things as clearly as possible but feel free to chat with me for more clarity. 
As you know, this is a follow up ticket for [OSF-8320](https://github.com/CenterForOpenScience/osf.io/pull/7530) to clean the legacy code after removing mod-meta. One of the legacy code to cleanup are the tests in `addons/tests/test_wiki.py`, which should no longer user the `wiki_widget` route to get wiki data but the serialization helper instead.
After OSF-8320 is merged, @chennan47 started working on [OSF-8584](https://github.com/CenterForOpenScience/osf.io/pull/7743) which touches the same functionality. Nan's fix seems to take away 'Read More' for any wiki content because calling `len()` on a BeautifulSoup object would always return `1` (see BS4 code [here](https://github.com/duly/bs4/blob/9def6a42ecf9b5319916c720b2c6057d58b994b1/element.py#L911)), instead of the actual length of the html of the wiki content. However, this change does not break the tests in `addons/tests/tests_wiki.py` since those tests are still getting data from the `wiki_widget` route, which was not removed by OSF-8320.
Therefore, when I work on this ticket, I removed the `wiki_widget` route and change the `addons/tests/test_wiki.py` tests to use the serialization helper. I noticed Nan's fix when I try to resolve the merge conflict so I incorporated the fix in this PR. Because the tests are changed to get the wiki data correctly, they are failing for long content cutoff, as after Nan's fix there would be no content cutoff.
So here is the solution I propose: I will change this PR to not incorporate Nan's fix and we can get this merged in. (No need to revert OSF-8584 as after this PR gets merged it is effectively reverting that change.) Then Nan could probably do some additional dev on OSF-8584, this time in the wiki serialization helper inside `addons/wiki/utils.py` instead of the `view_project` method of `node.py`.
I apologize for the confusion. I should have noticed the need to cleanup the code for the first PR so that this would not have happened in the first place.

PR #7677, opened 2017-09-08
[OSF-6477] Allow the user to cancel the deactivation request for their account.

Johnetordoff:
## Purpose

Currently if a user tries to deactivate your account they send a deactivation request to an admin who will review it and get back to them, so if they click it (and confirm it with a confirmation modal) by accident they can't rescind their request. This PR allows them to cancel their deactivation request.

## Changes

Adds route for canceling request and minor front-end changes to call that route and display reinvent info about account status. 

## Side effects

None that I know of.

## Testing 

I checked for pre-existing relevant tests and there weren't any. It would be beneficial to write some simple tests, but it should probably be done as a separate ticket. It just seems illogical to include tests for canceling a deactivation request when there are none for making a deactivation request.

## Ticket

https://openscience.atlassian.net/browse/OSF-6477

### DIFF
--- a/admin/templates/users/user.html
+++ b/admin/templates/users/user.html
@@ -7,7 +7,7 @@
 {% block content %}
 <div class="container-fluid">
     <div class="row">
-        <div class="col-md-8">
+        <div class="col-md-12">
             <div class="btn-group" role="group">
                 <a href="{% url 'users:search' %}" class="btn btn-default"><i class="fa fa-search"></i></a>
                 {%  if perms.osf.view_spam %}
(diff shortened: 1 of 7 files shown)


admin/templates/users/user.html
```diff
@@ -32,23 +32,39 @@
...
-                <a href="{% url 'users:ham_enable' user.id %}" data-toggle="modal" data-target="#enableHamModal" class="btn btn-success">Re-enable Ham account</a>
-            {% endif %}
-            <a href="{% url 'users:reindex-elastic-user' guid=user.id %}" data-toggle="modal" data-target="#confirmReindexElasticUser" class="btn btn-success">Elastic Reindex</a>
+    </div>
+    <div class="row">
+        <div class="col-md-12">
+            <div class="btn-group padded" role="group">
+                {% if not user.disabled %}
+                    {% if user.requested_deactivation %}
+                        <a href="{% url 'users:disable' user.id %}" data-toggle="modal" data-target="#disableModal" class="btn btn-danger">Requested account deactivation</a>
```
laurenbarker: @Johnetordoff is this right? It's the same as the button if the user hasn't requested deactivation...

laurenbarker: Oh except it's not disabled. Is this how the admin is suppose to cancel the request or did I misunderstand the feature?

PR #7680, opened 2017-09-11
Files list endpoint optimizations [OSF-8523]

erinspace:
## Purpose
There were a large number of database queries (~163) made when retrieving a page of 10 files on the normal provider page and quickfiles page. These optimizations bring that down to about 44 db queries. 

## Changes
- Add `IncludeManager` to `BaseNodeFileManager` to use include for later optimizations
- optimize quickfiles view and provider files view to prefetch related fields
- Add a descending index on `date_created` for `FileVersion`s
- Change default ordering for `FileVersion` and swap calls for `first()` and `last()` to match change. Calls to `last()` cause another reordering to be done, which slows things down. Since we were using more `last()` than `first()`,  they were swapped to speed things up in general.
- Remove `get_current_version` from the serializer in `/files/serializers.py` and instead use a property on the models `current_version_number` to calculate the version -- less expensive and can be special cased for `OSFStorageFile`s
- Remove redundant ordering call for guids in `osf/models/base.py` as guids are already ordered by `-created` and the extra call doesn't allow the optimization to be useful

## Side effects
- Calls to `first()` and `last()` when accessing fileversions will need to be double checked

## Ticket
https://openscience.atlassian.net/browse/OSF-8523


osf/models/files.py
```diff
@@ -615,7 +621,7 @@ class FileVersion(ObjectIDMixin, BaseModel):
     identifier = models.CharField(max_length=100, blank=False, null=False)  # max length on staging was 51
 
     # Date version record was created. This is the date displayed to the user.
-    date_created = NonNaiveDateTimeField(auto_now_add=True)
+    date_created = NonNaiveDateTimeField(auto_now_add=True, db_index=True)
```
sloria: Why was `db_index=True` added? Do we need an ascending index? I assume not, since there isn't a migration for (though running `makemigrations` results in a new migration).

erinspace: Oops, good call, forgot to remove that when we decided to add the negative index instead!

PR #7691, opened 2017-09-12
[feature] Prevent moving and copying quickfiles on the backend [OSF-8117]

erinspace:
[#OSF-8117]

## Purpose
The backend should also prevent moving and copying files in a Quick Files Node, so you can't be sneaky and use the API instead of the frontend. 

## Changes
- Add quickfiles check to move and copy hooks
- add quickfiles exception to make things super clear (and to match checkout and preprint file handling)
- add tests to make sure they're caught properly 

## Side effects

None anticipated

## Ticket

### DIFF
--- a/addons/osfstorage/views.py
+++ b/addons/osfstorage/views.py
@@ -85,7 +85,12 @@ def osfstorage_get_revisions(file_node, node_addon, payload, **kwargs):
 
 @decorators.waterbutler_opt_hook
 def osfstorage_copy_hook(source, destination, name=None, **kwargs):
-    return source.copy_under(destination, name=name).serialize(), httplib.CREATED
+    try:
+        return source.copy_under(destination, name=name).serialize(), httplib.CREATED
+    except exceptions.FileNodeIsQuickFilesNode:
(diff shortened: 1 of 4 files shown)


addons/osfstorage/views.py
```diff
@@ -85,7 +85,12 @@ def osfstorage_get_revisions(file_node, node_addon, payload, **kwargs):
 
 @decorators.waterbutler_opt_hook
 def osfstorage_copy_hook(source, destination, name=None, **kwargs):
-    return source.copy_under(destination, name=name).serialize(), httplib.CREATED
+    try:
+        return source.copy_under(destination, name=name).serialize(), httplib.CREATED
+    except exceptions.FileNodeIsQuickFilesNode:
+        raise HTTPError(httplib.METHOD_NOT_ALLOWED, data={
```
sloria: "405 Method not allowed" is returned when the **HTTP method** is not valid. In this case, the method is supported ("POST"), but the request is semantically invalid, i.e. our business rules prevent copying to quickfiles. Therefore, a 400 is more appropriate here.

PR #7693, opened 2017-09-12
[#OSF-8174]Make file view counts through OSF file storage available through the API

chennan47:
## Purpose
1. Make Pagecounter view counts for OSF file
2. migrate all old files view counts
3.Make file view counts available through OSF file storage available through the API

## Changes
before change:
screen shot 2017-09-12 at 1 13 58 pm
after change:
screen shot 2017-09-12 at 1 14 21 pm

## Ticket
https://openscience.atlassian.net/browse/OSF-8174


osf/models/analytics.py
```diff
@@ -142,3 +147,17 @@ def get_basic_counters(cls, page):
...
         except cls.DoesNotExist:
             return (None, None)
+
+    @classmethod
+    def set_basic_counters(cls, page, count, date=None):
+        cleaned_page = cls.clean_page(page)
+
+        if not date:
+            date = timezone.now()
+        date_string = date.strftime('%Y/%m/%d')
```
brianjgeiger: Is this pulled from somewhere else? I suspect this should really be UTC, and at the very least would need a timezone. Unless there's a really good reason not to, let's make this UTC.

osf/management/commands/file_view_counts.py
```diff
@@ -0,0 +1,64 @@
...
+def set_file_view_counts(state, *args, **kwargs):
+    # get all osfstorage files which is_deleted == False
+    files = BaseFileNode.resolve_class('osfstorage', BaseFileNode.FILE).active.all()
+
+    for file_node in files:
+        # for each file get the file view counts from keen
+        query = [{'property_name': 'page.info.path', 'operator': 'eq', 'property_value': file_node._id}]
+
+        query = urllib.quote(json.dumps(query))
+        url = 'https://api.keen.io/3.0/projects/{}/queries/count' \
```
brianjgeiger: @felliott Given the recent keen changes, is this the approach we should take, or would it be better to try to do some kind of batching? I'm not sure if batching is possible/convenient, so this is probably fine, but wanted you opinion.

osf/models/analytics.py
chennan47: Make it this way just to make it consistent with the date we used to update those counts. https://github.com/CenterForOpenScience/osf.io/pull/7693/files#diff-4519792f13c44cf51a8eec0288fa674aL77

brianjgeiger: I would see if you can update both of those to UTC without breaking anything, or if something breaks, that it's easy to fix. If it's a complicated change, we can re-evaluate, but non-timezoned dates are a bad idea.

sloria: @brianjgeiger Can you clarify? `date` is already timezone-aware in this case.

brianjgeiger: The line I'm specifically pointing to is the date_string which is just yyyy/mm/dd

brianjgeiger: No, wait, it's not a datetime, it's just a date. Sigh. Ignore me.

osf/management/commands/file_view_counts.py
```diff
@@ -0,0 +1,64 @@
...
+        PageCounter.set_basic_counters('view:{0}:{1}'.format(file_node.node._id, file_node._id), file_view_count)
+
+        logger.info('File ID {0}: has inputed "{1}" view counts'.format(file_node._id, file_view_count))
+
+    logger.info('File view counts migration from keen completed.')
+
+
+class Command(BaseCommand):
+    """
+    Backfill Retraction.date_retracted with `RETRACTION_APPROVED` log date.
```
felliott: Minor: this docstring needs to be updated

chennan47: Thanks for remind.

osf/management/commands/file_view_counts.py
felliott: Keen.io Compute Pricing Guide: https://keen.io/docs/compute/compute-pricing-guide/

Compute is charged at $0.01/1M properties scanned.  Number of properties scanned is euqal to E * P, where E is number of events in the timeframe and P is number of properties needed to calculate the query (see link for details).

Private pageviews has about 350k events over the last few days.  P for this query is 2: 1 + one filter property.  That means executing one of these queries involves 350k * 2 or 700k properties scanned, which would cost about $0.007.  But to calculate total cost, we'd have to multiply by the number of time we issued it, so it would end up at `$0.007 * len(files)`.

If instead you did a `group_by` on `page.info.path` that would return the counts for all files id (and non file-id pages) in one query for $0.007.  That might run into payload limits, though.

The third option is to batch, as @brianjgeiger said.  You'd have to futz around a bit to figure out the largest batch size that wouldn't hit payload limits, but that would probably be cheaper than this approach, too.  You'd change the filter parameter to be a `'page.info.path IN ['mst3k', 'ezcas', 'moooo', 'b33fy', ...]` and add a group by.  That would bring the P up to 3, so each query would be $0.012, but you could divide the number of queries you issue by batch size.  See: https://keen.io/docs/compute/query-tuning/#calculating-metrics-for-many-users-at-once

You'll need to decide what the right approach is based of how often you expect to run this and how big `len(files)` is.

osf/management/commands/file_view_counts.py
```diff
@@ -0,0 +1,80 @@
...
+            dest='dry_run',
+            help='Run migration and roll back changes to db',
+        )
+
+    def handle(self, *args, **options):
+        dry_run = options.get('dry_run', False)
+        if not dry_run:
+            script_utils.add_file_logger(logger, __file__)
+        with transaction.atomic():
+            set_file_view_counts
```
sloria: This isn't calling the function

osf/management/commands/file_view_counts.py
```diff
@@ -0,0 +1,80 @@
+from __future__ import unicode_literals
```
sloria: This file should be named something more descriptive. Something lke `migrate_file_view_counts_from_keen.py`

osf/models/analytics.py
```diff
@@ -124,6 +124,11 @@ def update_counter(cls, page, node_info):
                     model_instance.save()
                     return
 
+            if page_type == 'view' and node_info:
```
brianjgeiger: As discussed: For the else blocks here, send to sentry the context for all of the variables that go into the decision, so we can see why it's not going through.

Not discussed: maybe also log if this succeeds, so we can see if maybe just saving isn't doing enough for this.

brianjgeiger: The page_type is the riskiest part of this (although it should be fine because we've used it, there's a lot of code that goes into it and I don't think it's explicitly tested. Before this is done, we may pull out the cleaning and page type determination into its own function so that we can unit test those under different circumstances, but if we get this problem working, that will be a follow-up ticket.

brianjgeiger: Also, if anything fails in https://github.com/CenterForOpenScience/osf.io/pull/7693/files#diff-4519792f13c44cf51a8eec0288fa674aR81 then everything's going to fail, as it's wrapped in an atomic transaction.

brianjgeiger:
@chennan47 I can't do an automatic revert for some reason. Could you make a revert PR?

PR #7699, opened 2017-09-13
Preprint identifier list only return identifiers for that preprint [OSF-8640]

erinspace:
## Purpose
After the modm removal, the identifiers endpoint for preprints was returning too many identifiers!

## Changes
-Update preprint identifier list, remove old override
- change default queryset for IdentifierList to return identifiers for that object
- make preprint identifier list test a bit more robust

## Side effects
none anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-8640

### DIFF
--- a/api_tests/identifiers/views/test_identifier_list.py
+++ b/api_tests/identifiers/views/test_identifier_list.py
@@ -220,10 +220,13 @@ class TestPreprintIdentifierList:
         assert res_preprint_identifier.status_code == 200
         assert res_preprint_identifier.content_type == 'application/vnd.api+json'
 
-    def test_identifier_list_returns_correct_number_and_referent(self, preprint, res_preprint_identifier, data_preprint_identifier, all_identifiers):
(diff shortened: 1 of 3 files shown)


api_tests/identifiers/views/test_identifier_list.py
```diff
@@ -220,10 +220,13 @@ def test_identifier_list_success(self, res_preprint_identifier):
...
 
-    def test_identifier_list_returns_correct_number_and_referent(self, preprint, res_preprint_identifier, data_preprint_identifier, all_identifiers):
+    def test_identifier_list_returns_correct_number_and_referent(self, preprint, res_preprint_identifier, data_preprint_identifier, all_identifiers, user):
+        # add another preprint so there are more identifiers
+        PreprintFactory(creator=user)
+
         # test_identifier_list_returns_correct_number
         total = res_preprint_identifier.json['links']['meta']['total']
-        assert total == all_identifiers.count()
+        assert total == Identifier.objects.filter(object_id=preprint.id).count()
```
sloria: That's a much better way to test it.

PR #7702, opened 2017-09-13
[RFR] [OSF-8498] API: Add id and type to relationships

TomBaxter:
[#OSF-8498]

## Purpose
Return relationship data for those relations of the form...
proto://host:port/v2/<namespace>/<namespace_id>/

## Changes

Update api/base/serializers.py::RelationshipField::to_representation to add "data" to "relationships"

## Side effects

This PR makes assumptions about the form of endpoint URLs and the naming standard of kwargs passed to views. I don't believe that failures to live up to my assumptions will result in returning bad data, it should result in no passed data, but bad data is a possibility.

## Ticket

[OSF-8498](https://openscience.atlassian.net/browse/OSF-8498)

### DIFF
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -1,5 +1,6 @@
 import collections
 import re
+from urlparse import urlparse
 
 import furl
 from django.core.urlresolvers import resolve, reverse, NoReverseMatch
@@ -10,6 +11,7 @@ from rest_framework import exceptions, permissions
 from rest_framework import serializers as ser
 from rest_framework.fields import SkipField
 from rest_framework.fields import get_attribute as get_nested_attributes
+from rest_framework.mixins import RetrieveModelMixin
 
 from api.base import utils
 from api.base.exceptions import InvalidQueryStringError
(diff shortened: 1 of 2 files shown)


api/base/serializers.py
```diff
@@ -713,11 +714,28 @@ def to_representation(self, value):
...
         related_meta = self.get_meta_information(self.related_meta, value)
         self_url = url['self']
         self_meta = self.get_meta_information(self.self_meta, value)
-        return format_relationship_links(related_url, self_url, related_meta, self_meta)
-
+        relationship = format_relationship_links(related_url, self_url, related_meta, self_meta)
+        if related_url and (len(related_path.split('/')) & 1) == 1:
+            resolved_url = resolve(related_path)
+            related_class = resolved_url.func.view_class
+            if not hasattr(related_class, 'LIST_FIELDS'):
```
brianjgeiger: As discussed, better to check if the view derives from `mixins.mixins.RetrieveModelMixin` or doesn't derive from `mixins.ListModelMixin`

TomBaxter: Complete. I'll add a test and set ticket to CR.

TomBaxter:
Merged in https://github.com/CenterForOpenScience/osf.io/pull/7807

PR #7704, opened 2017-09-14
don't run sharejs as root

binoculars:
## Purpose

Run sharejs as non-root user

### DIFF
--- a/addons/wiki/Dockerfile
+++ b/addons/wiki/Dockerfile
@@ -1,6 +1,15 @@
 FROM node:0.12
 
-RUN mkdir -p /code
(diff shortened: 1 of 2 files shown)


addons/wiki/Dockerfile
```diff
@@ -1,6 +1,9 @@
 FROM node:0.12
 
-RUN mkdir -p /code
+RUN mkdir -p /code \
+    && chown node:node /code
+
+USER node
```
icereval: i'd recommend `gosu` since this it hard to ever get root on the container once its running, which can be useful for apt-get installs and tools.

PR #7705, opened 2017-09-14
[hotfix] Update Active Metaschemas [OSF-8641]

caseyrollins:
#### Purpose
- There are several metaschemas in the DB that are incorrectly marked as active. This migration will make inactive all metaschemas that are not version 2 and listed in `ACTIVE_METASCHEMAS`.

#### Ticket
- [OSF-8641](https://openscience.atlassian.net/browse/OSF-8641)


osf/migrations/0055_update_metaschema_active.py
```diff
@@ -0,0 +1,40 @@
...
+    'Replication Recipe (Brandt et al., 2013): Pre-Registration',
+    'Replication Recipe (Brandt et al., 2013): Post-Completion',
+    "Pre-Registration in Social Psychology (van 't Veer & Giner-Sorolla, 2016): Pre-Registration",
+]
+
+def update_metaschema_active(*args, **kwargs):
+    for schema in MetaSchema.objects.all():
+        if schema.name not in ACTIVE_METASCHEMAS or schema.schema_version != LATEST_SCHEMA_VERSION and schema.active:
+            logger.info('{} was active, setting to inactive.'.format(schema))
+            schema.active = False
```
sloria: This is OK but can be improved by using `.update`:

```python
MetaSchema.objects.exclude(name__in=ACTIVE_METASCHEMAS).exclude(schema_version=LATEST_SCHEMA_VERSION).update(active=False)
```

In the near future, all models will have a `modified` field that automatically gets set on `.save()`. Using `.update` prevents that from getting updated.

osf/migrations/0055_update_metaschema_active.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('osf', '0054_add_file_version_indices'),
+    ]
+
+    operations = [
+        migrations.RunPython(update_metaschema_active, ),
```
sloria: Please add a reverse migration.

sloria:
Actually, just thought a case where the migration might not do the right thing: If a new schema is added and a developer sets up a fresh database/dev environment then run the migration, the new schema will be marked as inactive.

Therefore, the migration should probably just be:

```python
MetaSchema.objects.filter(schema_version__lt=2).update(active=False)
```

PR #7709, opened 2017-09-14
[feature] Use select_for_update in APIv1 and APIv2 [OSF-8297]

caseyrollins:
#### Purpose
- Use `select_for_update` throughout the OSF to prevent updates from being overwritten during the course of a request.

#### Changes
- Pass requests to `get_object_or_error` (for APIv2) and `get_or_http_error` (for APIv1). Use `select_for_update` for all non-GET requests.
- Add tests to confirm `FOR UPDATE` exists in non-GET request queries.

#### Side Effects
- Note: `select_for_update` cannot be used with `select_related` (or with `OUTER JOIN`s in general?) See [this](https://stackoverflow.com/questions/36253793/django-select-for-update-and-select-related-on-same-query) SO post and [this](https://www.postgresql.org/message-id/Pine.LNX.4.58.0511142229050.12705@linuxworld.com.au) Postgres thread for more info.

#### Ticket
- [OSF-8297](https://openscience.atlassian.net/browse/OSF-8297)


api/base/utils.py
```diff
@@ -75,8 +75,9 @@ def absolute_reverse(view_name, query_kwargs=None, args=None, kwargs=None):
     return url
 
 
-def get_object_or_error(model_cls, query_or_pk, display_name=None):
+def get_object_or_error(model_cls, query_or_pk, request, display_name=None):
     obj = query = None
+    select_for_update = bool(request.method != 'GET')
```
sloria: This will not account for `HEAD` and `OPTIONS` requests. Rather than comparing to `GET`, you can use `rest_framework.permissions.SAFE_METHODS` which is a list of read-only methods.

PR #7711, opened 2017-09-14
[hotfix] Fix inactive user password reset [OSF-8652]

alexschiller:
## Purpose
inactive users can reset passwords. They shouldn't be able to get in but it ain't great. 

## Changes
Use is_active to determine if a user can receive emails for password resets instead. Do not show error message if attempting to reset passwords for an unconfirmed user. 

## Side effects
N/A

## Ticket

https://openscience.atlassian.net/browse/OSF-8652

## QA Notes
1. confirm disabled users, unconfirmed users, spam users, don't get emails (and don't show the error message when trying to reset passwords). 
2. Confirm that normal password resets work.

### DIFF
--- a/framework/auth/views.py
+++ b/framework/auth/views.py
@@ -165,34 +165,25 @@ def forgot_password_post():
                 status_message = 'You have recently requested to change your password. Please wait a few minutes ' \
                                  'before trying again.'
                 kind = 'error'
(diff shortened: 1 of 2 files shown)


alexschiller:
Accidental dupe of #7710 Closing

PR #7715, opened 2017-09-15
[Feature] Cursor based pagination [#OSF-8565]

mfraezz:
## Purpose
Improve `NodeLogList` performance

## Changes
* Create JSONAPI-compliant subclass for `rest_framework.pagination.CursorPagination`
* Use on `NodeLogList`


sloria:
This is a pretty significant, backwards-incompatible change (worthy of more than a minor version bump). I wonder if we should keep the current pagination scheme to preserve compat but allow changing the pagination style via a query parameter. Looks like there's even a [package for that](https://github.com/tuffnatty/drf-proxy-pagination).

Thoughts, @brianjgeiger ?

sloria:
The loading indicator for the log feed got lost with this change.

That said, the log feed was using the now-deprecated OSF loading spinner. 

Good: Update the log feed to use spinner (`.ball-pulse`).

Better: Use a [skeleton screen](https://www.sitepoint.com/how-to-speed-up-your-ux-with-skeleton-screens/) as a loading indicator. A quick-and-dirty way to achieve this is to reduce the opacity of the items when the widget is in the loading state:

osf___small_project

mfraezz:
According to discussion with @brianjgeiger sometime over the past two weeks, this does not merit an API version bump. The value for linked references change, yes, but the rationale is that the client should only be following the `resp.json['links']`, the references for which do not change. That is, client-side url construction is not a supported use case; they should be using links provided in the response.

I'll fix the loading UI

sloria:
@mfraezz Ah ok, thanks for the clarification. I wasn't sure if the keys of the pagination object were changing here or not. If not, then I agree: no need to bump versions.

PR #7717, opened 2017-09-15
[feature] Fix unregistered contributors registration records [OSF-8624]

alexschiller:
## Purpose
Registrations don't get unclaimed records. Fix that. 

## Changes
1. When nodes are registered, copy over the unclaimed records
2. Script to fix old data

## Side effects
Fix deletion of unclaimed records when contributors are removed. Currently they stick around because of a missing save. Test for that was passing because data wasn't being refreshed from the DB.  Humorously this missing deletion oversight makes the script much more efficient to run and with less data loss.

## Ticket

https://openscience.atlassian.net/browse/OSF-8624

## QA Notes
- Currently unclaimed users on registrations are not claimable via the registration. After the script is run these should work.
- Test creating a new registration with unclaimed contributors allows those contributors to be claimed via the registration.
- Test that creating a registration with unclaimed users, then deleting the originating node allows you to claim the user
- Test that unclaimed users on registration components are also claimable
- Test that unclaimed with and without email addresses work (this can be interspersed with the other tests, no reason it should be different)
- Confirm that citations work on registrations with unregistered contributors

## Deployment Actions
The `scripts/fix_registration_unclaimed_records.py` needs to be run to update the unclaimed records. It can be run in dry mode to confirm everything is working prior to running for real. No special instructions for staging vs prod.

### DIFF
--- a/osf/models/node.py
+++ b/osf/models/node.py
@@ -1328,6 +1328,7 @@ class AbstractNode(DirtyFieldsMixin, TypedModel, AddonModelMixin, IdentifierMixi
         # remove unclaimed record if necessary
         if self._primary_key in contributor.unclaimed_records:
             del contributor.unclaimed_records[self._primary_key]
+            contributor.save()
 
         # If user is the only visible contributor, return False
         if not self.contributor_set.exclude(user=contributor).filter(visible=True).exists():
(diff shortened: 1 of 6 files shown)


brianjgeiger:
@alexschiller Could you add a Deployment Actions section so that, when this is merged to its various environments, the proper folks know what to do?

PR #7718, opened 2017-09-18
Feature/preprint creator email[OSF-7379]

laurenrevere:
## Purpose

Send an email to the creator to confirm that a preprint has been created.

## Changes

Adds text emails that confirm a preprint has been created for both branded and OSF preprints.

## Side effects
None

## Note
Note, these emails are plain text, the plan is to convert them to HTML with a later ticket ([PRODUCT-169](https://openscience.atlassian.net/browse/PRODUCT-169.))

Code review question: Should I have created s signal `preprint_published` to handle this instance?

## Ticket
https://openscience.atlassian.net/browse/OSF-7379

default
branded

### DIFF
--- a/osf/models/preprint_provider.py
+++ b/osf/models/preprint_provider.py
@@ -13,6 +13,7 @@ from osf.utils.fields import EncryptedTextField
 from reviews import permissions as reviews_permissions
 from reviews.models import ReviewProviderMixin
 
+from website import settings
 from website.util import api_v2_url
 
 
@@ -93,6 +94,10 @@ class PreprintProvider(ObjectIDMixin, ReviewProviderMixin, BaseModel):
         else:
             # TODO: Delet this when all PreprintProviders have a mapping
             return rules_to_subjects(self.subjects_acceptable)
+
+    @property
+    def landing_url(self):
+        return self.domain if self.domain else '{}preprints/{}'.format(settings.DOMAIN, self.name.lower())
 
     def get_absolute_url(self):
         return '{}preprint_providers/{}'.format(self.absolute_api_v2_url, self._id)
--- a/osf/models/preprint_service.py
+++ b/osf/models/preprint_service.py
@@ -14,10 +14,9 @@ from osf.models.validators import validate_subject_hierarchy
 from osf.utils.fields import NonNaiveDateTimeField
 from website.preprints.tasks import on_preprint_updated, get_and_set_preprint_identifiers
 from website.project.licenses import set_license
-from website.project import signals as project_signals
 from website.util import api_v2_url
 from website.util.permissions import ADMIN
-from website import settings
+from website import settings, mails
 
 from reviews.models.mixins import ReviewableMixin
 from reviews.workflow import States
@@ -207,13 +206,11 @@ class PreprintService(DirtyFieldsMixin, GuidMixin, IdentifierMixin, ReviewableMi
             # This should be called after all fields for EZID metadta have been set
             enqueue_postcommit_task(get_and_set_preprint_identifiers, (), {'preprint': self}, celery=True)
 
+            self._send_preprint_confirmation(auth)
+
         if save:
             self.node.save()
             self.save()
-
-            # Send creator confirmation email
-            project_signals.contributor_added.send(self.node, contributor=auth.user, auth=auth,
-                                                   email_template='preprint_confirmation')
 
     def set_preprint_license(self, license_detail, auth, save=False):
         license_record, license_changed = set_license(self, license_detail, auth, node_type='preprint')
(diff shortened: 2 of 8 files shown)


laurenrevere:
In relation to moderation and https://openscience.atlassian.net/browse/MOD-53. I talked to @aaxelb and he told me that `set_published` is not called by moderation. Therefore we should not have the case of double emails. If moderation is turned on for a provider, this email request will be bypassed.

osf/models/preprint_service.py
```diff
@@ -199,6 +200,10 @@ def set_published(self, published, auth, save=False):
             self.node.save()
             self.save()
 
+            # Send creator confirmation email
+            project_signals.contributor_added.send(self.node, contributor=auth.user, auth=auth,
```
alexschiller: We chatted about this a bit, but adding notes here so I don't forget:

1. This should probably just be a send_mail. It looks like contributor added hits a couple places, including modifying subscription settings (see [this section](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/notifications/listeners.py#L20))
2. If not a send_mail it can probably just be its own signal, but I'm not 100% clear on when signals should be used so understanding that better is probably necessary
3. See [this pr](https://github.com/CenterForOpenScience/osf.io/pull/7515/files#diff-51eef6ea2a6c35e2b080c9a6fe591e38R1636) for a pretty simple example of a sendmail

website/mails/mails.py
```diff
@@ -211,6 +211,14 @@ def get_english_article(word):
     'contributor_added_preprint_node_from_osf',
     subject='You have been added as a contributor to an OSF project.'
 )
+PREPRINT_CONFIRMATION = lambda template, provider: Mail(
```
alexschiller: Gut feeling says this should be done differently, or at minimum add a comment explaining it.

website/project/views/contributor.py
```diff
@@ -560,6 +564,11 @@ def notify_added_contributor(node, contributor, auth=None, throttle=None, email_
     elif not contributor.is_registered:
         unreg_contributor_added.send(node, contributor=contributor, auth=auth, email_template=email_template)
 
+def get_preprint(node):
+    try:
+        return PreprintService.objects.get(node=node)
```
alexschiller: A single node can have more than one preprint associated with it, which means `get` could cause some problems. 

That's why, for example, node loops over every preprint it is associated with on save ([see here](https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/models/node.py#L1533))

website/templates/emails/preprint_confirmation_branded.txt.mako
```diff
@@ -0,0 +1,37 @@
+<%!
+    from website import settings
+%>
+
+Hello ${user.fullname},
+
+Congratulations on sharing your ${preprint.provider.preprint_word}  "${node.title}" on ${preprint.provider.name}, powered by OSF Preprints: ${preprint.absolute_url}
```
alexschiller: two spaces after 'preprint word}', intentional?

website/templates/emails/preprint_confirmation_branded.txt.mako
```diff
@@ -0,0 +1,37 @@
...
+
+
+
+You will ${'not receive ' if all_global_subscriptions_none else 'be automatically subscribed to '}notification emails for this project. To change your email notification preferences, visit your project or your user settings: ${settings.DOMAIN + "settings/notifications/"}
+
+Sincerely,
+
+Your ${preprint.provider.name} and OSF teams
+
+Want more information? Visit https://osf.io/preprints/${preprint.provider.name.lower()} to learn about ${preprint.provider.name} or https://osf.io/ to learn about the Open Science Framework, or https://cos.io/ for information about its supporting organization, the Center for Open Science.
```
alexschiller: This should probably use preprint.provider.domain if it redirects instead of building the `preprints/provider` route. (though this is probably a product question as well, not sure if custom domains existed when the ticket was created)

Might be worth adding a provider `url` method similar to preprints `url` method [seen here](https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/models/preprint_service.py#L86).

tests/test_preprints.py
```diff
@@ -727,3 +728,27 @@ def test_no_call_async_update_on_400_failure(self, requests, mock_async, mock_ma
...
+        super(TestPreprintConfirmationEmails, self).setUp()
+        self.user = AuthUserFactory()
+        self.write_contrib = AuthUserFactory()
+        self.project = ProjectFactory(creator=self.user)
+        self.project.add_contributor(self.write_contrib, permissions=[permissions.WRITE])
+        self.preprint = PreprintFactory(project=self.project, is_published=False)
+        contributor_added.connect(notify_added_contributor)
+
+    @mock.patch('website.mails.send_mail')
+    def test_creator_gets_email(self, send_mail):
```
alexschiller: This and the below test can be combined-- test failed published first and then real publish.

tests/test_preprints.py
```diff
@@ -727,3 +728,27 @@ def test_no_call_async_update_on_400_failure(self, requests, mock_async, mock_ma
         update_preprint_share(self.preprint)
         assert not mock_async.called
         assert mock_mail.called
+
+class TestPreprintConfirmationEmails(OsfTestCase):
+    def setUp(self):
+        super(TestPreprintConfirmationEmails, self).setUp()
+        self.user = AuthUserFactory()
+        self.write_contrib = AuthUserFactory()
```
alexschiller: My read on this was that only the person creating the preprint was emailed. Should there be a test to confirm emails are not sent to the other users? Or vice-versa if I'm reading it wrong.

osf/models/preprint_provider.py
```diff
@@ -87,6 +88,10 @@ def all_subjects(self):
             # TODO: Delet this when all PreprintProviders have a mapping
             return rules_to_subjects(self.subjects_acceptable)
 
+    @property
+    def landing_url(self):
```
brianjgeiger: I am surprised we didn't have anything like this already.

brianjgeiger:
@laurenrevere Merge conflicts. Please resolve at your earliest convenience.

PR #7719, opened 2017-09-18
Feature/probabilistic test failure - pagination[OSF-8626]

laurenrevere:
## Purpose

`test_search_pagination_default_page_2` was failing probabilistically. 

## Changes

All search tests in `test_search_views` were consolidated except for `test_search_projects_by_title`.

## Side effects

Should speed up tests

## Ticket

https://openscience.atlassian.net/secure/RapidBoard.jspa?rapidView=107&selectedIssue=OSF-8626

### DIFF
--- a/osf_tests/test_search_views.py
+++ b/osf_tests/test_search_views.py
@@ -34,7 +34,8 @@ class TestSearchViews(OsfTestCase):
         import website.search.search as search
         search.delete_all()
 
-    def test_search_contributor(self):
+    def test_search_views(self):
+        #Test search contributor
         url = api_url_for('search_contributor')
         res = self.app.get(url, {'query': self.contrib.fullname})
         assert_equal(res.status_code, 200)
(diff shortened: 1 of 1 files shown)


laurenrevere:
*@brianjgeiger *The whole file, or just the `TestSearchViews` class?

brianjgeiger:
If you can do the whole file, that would be better.

brianjgeiger:
Though I think the test class would be fine if it's not pretty easy to get the whole file in one go.

osf_tests/test_search_views.py
```diff
@@ -34,7 +34,7 @@ def tearDown(self):
         import website.search.search as search
         search.delete_all()
 
-    def test_search_contributor(self):
```
brianjgeiger: Looks like you went juuuust a little too far. This is all in the tear down method. Maybe have `def test_search(self):` here.

laurenrevere:
Just noticed that. And fixed!

PR #7720, opened 2017-09-18
[feature fix] Don't use select_for_update outside of a transaction [OSF-8297]

caseyrollins:
#### Purpose
- Continuation of #7709 

#### Changes
- Check that request is within a transaction before using `select_for_update`

#### Ticket
- [OSF-8297](https://openscience.atlassian.net/browse/OSF-8297)

### DIFF
--- a/framework/database/__init__.py
+++ b/framework/database/__init__.py
@@ -5,6 +5,7 @@ import httplib as http
 import markupsafe
 from django.core.paginator import Paginator
 from django.db.models import Q, QuerySet
+from django.db import transaction
 from flask import request
 from rest_framework.permissions import SAFE_METHODS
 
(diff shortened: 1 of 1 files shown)


framework/database/__init__.py
```diff
@@ -31,7 +32,7 @@ def get_or_http_error(Model, pk_or_query, allow_deleted=False, display_name=None
     display_name = display_name or ''
     # FIXME: Not everything that uses this decorator needs to be markupsafe, but OsfWebRenderer error.mako does...
     safe_name = markupsafe.escape(display_name)
-    select_for_update = bool(request.method not in SAFE_METHODS)
+    select_for_update = bool(request.method not in SAFE_METHODS and not view_has_annotation(NO_AUTO_TRANSACTION_ATTR))
```
sloria: It's still possible for this to be called outside of a transaction, i.e. if a view is managing it's own transactions.

```python
@no_auto_transaction
def myview():
    obj = get_or_http_error(..)
     with transaction.atomic():
            # ...
```

Does Django offer a way to check whether we're in the context of an active transaction?

PR #7732, opened 2017-09-22
Add more metrics to nightly keen runs [OSF-8505]

alexschiller:
## Purpose
There are a lot of metrics we don't store, including stickiness, and other daily change values. Many of which are calculated using multiple queries, etc.

## Changes
- Add a boat-load of daily queries
- Add stickiness as calculated from keen in one query to use in the future
- Add embargo_v2 that uses future embargo dates instead of simply private registrations
- Swap the variable metrics used for embargoes to the proper value since they are currently broken (institutional embargo counts)
- Update "--yesterday" flag to use timezone.now for all summary files
- Update base keen to use timezone.now so that tests don't fail between 8 and midnight
- Add tests to users and nodes (institutions do not have tests yet)
- Consolidate tests to speed up performance (despite more tests, users 42 seconds to 14; nodes 123 seconds to 28)

## Side effects
None known

## Ticket

https://openscience.atlassian.net/browse/OSF-8505

## QA Considerations
Check that institution metrics show up.
Do a quick pass of metrics on admin and make sure they look like they are updating properly: note any odd data, broken charts, etc.


brianjgeiger:
Oh, no! Conflicts!

scripts/analytics/preprint_summary.py
```diff
@@ -81,7 +82,7 @@ def get_class():
     args = preprint_summary.parse_args()
     yesterday = args.yesterday
     if yesterday:
-        date = (datetime.today() - timedelta(1)).date()
+        date = (timezone.now() - timedelta(1)).date()
```
brianjgeiger: It's a little better to use named arguments here, since timedelta includes a lot of optional parameters, and having (day=1) is a tiny bit more readable. Yes, not your code originally, but since we're here.

scripts/analytics/user_summary.py
```diff
@@ -32,6 +42,34 @@ class UserSummary(SummaryAnalytics):
...
+        )
+
+        last_one = client.count_unique(
+            event_collection='pageviews',
+            timeframe={'start': time_one.isoformat(), 'end': time_two_iso},
+            target_property='user.id',
+            timezone='UTC'
+        )
+
+        # avoid unlikely divide by 0 error
```
brianjgeiger: 👍

alexschiller:
Converted the timedeltas to use `days` keywords on all the keen scripts files. Testing running the summaries and reran all the tests:
screen shot 2017-10-03 at 9 56 38 am

brianjgeiger:
Sad to say there are more conflicts.

alexschiller:
Not surprised. Will fix that quick.

alexschiller:
Should be fine now... till more things get merged 😆

brianjgeiger:
@alexschiller Conflicts

scripts/analytics/institution_summary.py
```diff
@@ -29,48 +33,68 @@ def get_events(self, date):
...
+        # `embargoed` used private status to determine embargoes, but old registrations could be private and unapproved registrations can also be private
+        # `embargoed_v2` uses future embargo end dates on root
+        embargo_v2_query = Q(root__embargo__end_date__gt=query_datetime)
 
         for institution in institutions:
-            node_query = (
-                Q(is_deleted=False) &
-                Q(created__lt=query_datetime)
-            )
+            node_qs = institution.nodes.filter(is_deleted=False, created__lt=query_datetime).exclude(type='osf.registration')
```
alexschiller: @brianjgeiger This PR had some fun merge conflicts and in fixing them ended up with some refactoring to what I think is a more readable version. 
Projects and nodes now use a separate base qs that removes some of the need for node_query, etc. 

This turns something like:
```
'private': institution.nodes.filter(project_private_query).exclude(type='osf.registration').get_roots().count(),
```
into 
```
'private': node_qs.filter(private_query).get_roots().count(),
```
If you think this change is undesirable I can swap em back, but I think it makes processing the individual queries easier. 

The node_summary has good test coverage but there never was a institution_summary tests. To validate I compared the data from the same day before and after and all the numbers line up as expected.

Issue #7164, opened 2017-04-28
Allow image size markup in wiki


hafen:
It would be really useful to be able to specify image size for images embedded in a wiki. Often we have no control over the size of the image we are linking to and don't want it to overwhelm the wiki page.

There is a [markdown-it plugin](https://www.npmjs.com/package/markdown-it-imsize) that easily allows this with commonly-used markdown syntax for this scenario, e.g.

```markdown
![test](image.png =100x200)
```

Here are some of my attempts showing it's not currently possible to specify image size.

<img width="1004" alt="screen shot 2017-04-27 at 5 37 22 pm" src="https://cloud.githubusercontent.com/assets/1275592/25509840/9b41d24e-2b70-11e7-86c4-b7c058f376a9.png">

saradbowman:
Hi @hafen, Thanks for the feedback! I'll pass this along to our Product team for triage and will keep you posted if/when this feature becomes available. Don't hesitate to reach out in the future with more feedback/feature requests! (here or email support AT osf DOT io)

saradbowman:
Hi @hafen, this feature is now in production. Try it out!

serbinsh:
Hello.  I am trying to follow the wiki instructions https://help.osf.io/hc/en-us/articles/360019737474-Edit-the-Wiki#embed-images-from-osf-storage and resize an embedded OSF image.  However it doesnt seem, at least from my experimentation, that its possible to resize an osf image like you can with an external, is this accurate?  

This works

```
@[osf](uvr57)
```

This doesnt

```
@[osf](uvr57 =50x50)
```

I was hoping to embed images from my storage space into the wiki but only have them load a small image not full res.  Is this possible?

brianjgeiger:
Hi, @serbinsh,

You might have to use a full URL for this. Have you tried `![osf](https://osf.io/uvr57/download =50x50)`?

serbinsh:
Bingo! thanks!

PR #7733, opened 2017-09-25
[OSF-8011] Markdown image resize

adlius:
## Purpose

This PR allows users to specify image size using markdown syntax for wiki edits. See #7164 

## Changes

Changes are made to `markdown.js` to make markdown-it use the new plugin. Other related changes to add `makrdown-it-imsize` dependency to the project.

Note: Somehow, `{ node: 'empty' }` has to be added to the `module.export` for config files due to a webpack bug, otherwise wepack complains about not being able to resolve `fs`. (see: https://github.com/webpack-contrib/css-loader/issues/447). This could potentially be solved by upgrading to the latest version of webpoack.

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-8011


karma.common.conf.js
```diff
@@ -32,6 +32,9 @@ var webpackTestConfig = {
             // Assume test files are ES6
             {test: /\.test\.js$/, loader: 'babel-loader'},
         ])
+    },
+        node: {
+       fs: 'empty'
```
brianjgeiger: What does this do?

adlius: The added lines serves the same function for our JS tests as it does in `wepack.common.config.js` as a workaround to suppress an error due to a known webpack issue.

package.json
```diff
@@ -44,6 +44,7 @@
     "lodash.includes": "^4.3.0",
     "lodash.set": "^4.3.2",
     "markdown-it": "~4.1.0",
+    "markdown-it-imsize": "^2.0.1",
```
binoculars: You must use `yarn add`/`yarn remove` to update package.json so that the yarn lockfile will be updated.

brianjgeiger: Yeah, this was a PR from September, so I think was from before yarn on the OSF. But we updated the yarn lock in another PR, so unless somethings changed there, this should be better.

PR #7734, opened 2017-09-25
Add Feature-Flipping Capabilities[OSF-8685]

pattisdr:
# Purpose
We often need to disable/enable features for various reasons, e.g. a feature is merged but is not ready to be released on prod due to a dependency on a third party or multiple large features are merged. Instead of holding the release, we incrementally roll out separate features as they complete QA. This also enables new use cases for rolling out features to subsets of users, e.g. promotional material, A/B testing, "beta" features, addons, etc.

# Changes
Add the pieces necessary for **django-waffle** to be used as a feature flag/switch solution.

# Usage

## Configuring Flags/Samples/Switches
- A flag/sample/switch can be created through the Django Admin App or the command line
- Example of adding sample in django admin app:
screen shot 2017-09-25 at 5 32 16 pm

## Migrations
Will have to run migrations to add waffle models to db.  

## Enabling/Disabling API Endpoints
To add a feature flipper to an API endpoint, use the `require_flag`, `require_sample`, or `require_switch` decorator defined in `api/base/waffle_decorators.py` .  Add the decorator to the `get_queryset` method or `get_object` method.  The name of the flag, sample, or switch should be passed in as a parameter to the decorator.  If the flag/sample/switch is inactive for the logged in user, a Not Found will be raised. 

``` python
from api.base.waffle_decorators import require_flag

    @require_flag('test_flag')
    def get_default_queryset(self):
        return default_node_list_permission_queryset(user=self.request.user, model_cls=Node)
```
## Hiding/Showing UI elements in Dashboard
- API endpoint added `localhost:8000/wafflejs`, that when called, exposes a global waffle object that gives access to flags, switches, and samples.
- wafflejs_url added to global context variables
- `<script src="${wafflejs_url}"></script>` added to base.mako, so global waffle object is exposed.
- to use:
```js
if (window.waffle.flag_is_active('test_flag')) {
     Here's the feature that I'm going to expose if the test_flag is active.
}
```

## Enabling/disabling features for only certain users 
You would use a FLAG for this.
In the admin app, there is a place to add specific users to a flag:
screen shot 2017-09-25 at 6 05 37 pm

## Hidden Waffle API Endpoint to use Waffle in Ember Apps
Closed companion PR here: https://github.com/CenterForOpenScience/ember-osf/pull/277 
- Based off of @brianjgeiger's [comments](https://github.com/CenterForOpenScience/osf.io/pull/7734#issuecomment-347988818) and @sloria 's [comments](https://github.com/CenterForOpenScience/osf.io/pull/7734#issuecomment-347682121) below, there is now a nonstandard `_waffle` list endpoint that you can filter against to get flags, switches, and samples.
` /v2/_waffle/?flags=featureA,featureB&samples=sampleA?`

The ember app should have default values FLAG_DEFAULT, SAMPLE_DEFAULT, and SWITCH_DEFAULT for when the flag value in question is not found.
A route should try to load the sample/switch/flag in question. If it loads, use the active value as a conditional for whether or not to show certain pieces in the UI. If it's not found, use the default value listed in the config.

### Example of filtering waffle endpoint:
`
http://localhost:8000/v2/_waffle/?switches=test_switch&flags=test_flag,nonexistant_flag
`

```json
{
    "data": [
        {
            "links": {},
            "attributes": {
                "active": false,
                "note": "Test_flag inactive for everyone",
                "name": "test_flag"
            },
            "type": "waffle",
            "id": "flag_1"
        },
        {
            "links": {},
            "attributes": {
                "active": true,
                "note": "Yes this is active for everyone",
                "name": "test_switch"
            },
            "type": "waffle",
            "id": "switch_1"
        }
    ],
    "meta": {
        "total": 2,
        "per_page": 10,
        "version": "2.5"
    },
    "links": {
        "self": "http://localhost:8000/v2/_waffle/?flags=test_flag%2Cnonexistant_flag&switches=test_switch",
        "first": null,
        "last": null,
        "prev": null,
        "next": null
    }
}

```

Flags have numerous fields on them, and if any of the fields are true for the logged in user, then the flag is active. Instead of exposing all these individual fields through the API, just exposing whether of not the flag is active for the logged in user.  This makes use of the waffle flag_is_active method that takes in the request as an argument. 

## Needed changes to Ember-OSF
Changes started here:
https://github.com/CenterForOpenScience/ember-osf/pull/277

Will need to add a waffle model, waffle serializer, and waffle adapter that overrides the buildUrl method that can point to the hidden `_waffle` endpoint.  

⭐️ NEW: Since `type` is now "flag", "switch", or "serializer", will need to add a Flag, Switch, and Serializer model.

Waffle Example Ember Route:
```js
import Ember from 'ember';

export default Ember.Route.extend({
    model() {
        return this.store.query('waffle', {samples: 'test_sample', flags:'test_flag,second_flag', switches: 'test_switch'});
    }
});
```

Waffle Ember Model
```js
import DS from 'ember-data';
import OsfModel from './osf-model';

export default OsfModel.extend({
    name: DS.attr('fixstring'),
    note: DS.attr('fixstring'),
    active: DS.attr('boolean'),
});

```
Waffle Adapter Example
```
import OsfAdapter from './osf-adapter';

export default OsfAdapter.extend({
    buildURL() {
        return this._super(...arguments).replace('waffles', '_waffle');
    }
});
```

## Ticket
https://openscience.atlassian.net/browse/OSF-8685


sloria:
> Recommend using Django-waffle v 0.11, instead of the latest version. Was getting issues when trying to enable features for certain users.

Are these known issues? If not, it would be good to report these on waffle's issue tracker.

website/templates/home.mako
```diff
@@ -45,6 +45,39 @@
         window.contextVars = $.extend(true, {}, window.contextVars, {
             dashboardInstitutions: ${ dashboard_institutions | sjson, n},
         });
+        var FLAGS = ${ flags | sjson, n }
```
sloria: Rather than writing this code here, could we instead request the JS file, as documented here: https://waffle.readthedocs.io/en/latest/usage/javascript.html ?

The only difference is that instead of using `<script src="{% url 'wafflejs' %}"></script>`, we would need to input the url ourselves. We could put the URL in the global mako context, i.e. in `website.routes.get_globals` so that it can be included in a script tag. Something like

```python
# website/routes
def get_globals():
    #...
    return {
         # ...
          'wafflejs_url': reverse('wafflejs')
    }
```


```mako
## base.mako

<script src="${wafflejs_url}"></script>

File: website/templates/home.mako [hunk 2]
```diff
@@ -45,6 +45,41 @@
         window.contextVars = $.extend(true, {}, window.contextVars, {
             dashboardInstitutions: ${ dashboard_institutions | sjson, n},
         });
+        // Creating global waffle object to give access to flags, switches and samples
```
sloria: Rather than writing this code here, could we instead request the JS file, as documented here: https://waffle.readthedocs.io/en/latest/usage/javascript.html ?

The only difference is that instead of using `<script src="{% url 'wafflejs' %}"></script>`, we would need to input the url ourselves. We could put the URL in the global mako context, i.e. in `website.routes.get_globals` so that it can be included in a script tag. Something like

```python
# website/routes
def get_globals():
    #...
    return {
         # ...
          'wafflejs_url': reverse('wafflejs')
    }
```


```mako
## base.mako

<script src="${wafflejs_url}"></script>

File: api/nodes/views.py [hunk 3]
```diff
@@ -276,8 +277,12 @@ class NodeList(JSONAPIBaseView, bulk_views.BulkUpdateJSONAPIView, bulk_views.Bul
 
     # overrides NodesFilterMixin
     def get_default_queryset(self):
-        user = self.request.user
-        return default_node_list_queryset() & default_node_permission_queryset(user)
+        # this is an example of using a flag inside an API view to enable/disable
+        if waffle.flag_is_active(self.request, 'api_flag_on_user_id'):
```
sloria: It'd be convenient to have a decorator for this use case:

```python
@require_flag('api_flag_on_user_id')
def get_default_queryset(self):
    user = self.request.user
     #....
```

api/users/views.py
```diff
@@ -152,8 +153,11 @@ def get_default_queryset(self):
 
     # overrides ListCreateAPIView
     def get_queryset(self):
-        return self.get_queryset_from_request()
-
+        # This is an example of using a switch to turn an API endpoint on and off within a view
+        if waffle.switch_is_active('api_switch'):
```
sloria: Same here: would be nice to have a decorator, `@requires_switch('api_switch')`.

sloria:
> Strategy is to create three API endpoints, one for samples, switches, and flags. All types have an active field which shows whether that sample/switch/flag is active for the logged in user. (Switches are active for everyone). The ember app should have default values FLAG_DEFAULT, SAMPLE_DEFAULT, and SWITCH_DEFAULT for when the flag value in question is not found.
A route should try to load the sample/switch/flag in question. If it loads, use the active value as a conditional for whether or not to show certain pieces in the UI. If it's not found, use the default value listed in the config.

pinging @brianjgeiger for feedback: Thoughts on this approach?

I'm not sure we want to have three different endpoints. Also, I foresee a use case where we'll need to fetch more than one flag/switch/sample in a request. Ideally, the frontend would make one request for all the flags, etc. it needs to render the current page.

Maybe something like `/v2/_waffle/?flags=featureA,featureB&samples=sampleA`?

This would mean diverging from JSON API, but given that this will be a dark endpoint, maybe that doesn't matter so much? In this case, it's more important for flag-checking to be performant.

sloria:
Just a few drive-by comments for now. Will review this further in depth tomorrow.

brianjgeiger:
Regarding a non-standard private endpoint, I'm not entirely opposed. There are a lot of ways of going about getting the data that would be more or less likely to break spec, including:

- Query parameter/header/etc that puts the data into the meta of whatever response you are currently accessing (we will be doing this soon with the API version, though by default rather than with a flag);
- A non-jsonapi renderer that just generally makes things more compact and that is required or encouraged or default for certain views - could be handy for other things as long as we don't guarantee it like we do the main API, like a CSV renderer or something;
- Returning things as a list in a jsonapi compliant way, but marking that endpoint as being allowed to return 1000, 100000, or infinite items in a single page. We have that functionality used in a couple of cases for things that are guaranteed to not get too long, guaranteed to be performant, or something special for Michael;
- Just make the endpoint work weird, which will either be super-quick or a surprisingly complicated affair.

pattisdr:
@sloria, regarding my recommendation of using Django-waffle v 0.11, instead of the latest version.  Once I upgrade to Django-waffle v 0.12, add a flag to the NodesList endpoint, and navigate to `api/v2/nodes/`, I get an error with the following traceback:
```
api_1               | Traceback (most recent call last):
api_1               |   File "/usr/local/lib/python2.7/wsgiref/handlers.py", line 85, in run
api_1               |     self.result = application(self.environ, self.start_response)
api_1               |   File "/usr/local/lib/python2.7/site-packages/django/contrib/staticfiles/handlers.py", line 63, in __call__
api_1               |     return self.application(environ, start_response)
api_1               |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/wsgi.py", line 157, in __call__
api_1               |     response = self.get_response(request)
api_1               |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 124, in get_response
api_1               |     response = self._middleware_chain(request)
api_1               |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/exception.py", line 43, in inner
api_1               |     response = response_for_exception(request, exc)
api_1               |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/exception.py", line 93, in response_for_exception
api_1               |     response = handle_uncaught_exception(request, get_resolver(get_urlconf()), sys.exc_info())
api_1               |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/exception.py", line 41, in inner
api_1               |     response = get_response(request)
api_1               |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 249, in _legacy_get_response
api_1               |     response = self._get_response(request)
api_1               |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 187, in _get_response
api_1               |     response = self.process_exception_by_middleware(e, request)
api_1               |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 185, in _get_response
api_1               |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
api_1               |   File "/usr/local/lib/python2.7/site-packages/django/utils/decorators.py", line 185, in inner
api_1               |     return func(*args, **kwargs)
api_1               |   File "/usr/local/lib/python2.7/site-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
api_1               |     return view_func(*args, **kwargs)
api_1               |   File "/usr/local/lib/python2.7/site-packages/django/views/generic/base.py", line 68, in view
api_1               |     return self.dispatch(request, *args, **kwargs)
api_1               |   File "/usr/local/lib/python2.7/site-packages/rest_framework/views.py", line 489, in dispatch
api_1               |     response = self.handle_exception(exc)
api_1               |   File "/usr/local/lib/python2.7/site-packages/rest_framework/views.py", line 449, in handle_exception
api_1               |     self.raise_uncaught_exception(exc)
api_1               |   File "/usr/local/lib/python2.7/site-packages/rest_framework/views.py", line 486, in dispatch
api_1               |     response = handler(request, *args, **kwargs)
api_1               |   File "/usr/local/lib/python2.7/site-packages/rest_framework/generics.py", line 201, in get
api_1               |     return self.list(request, *args, **kwargs)
api_1               |   File "/usr/local/lib/python2.7/site-packages/rest_framework/mixins.py", line 45, in list
api_1               |     return self.get_paginated_response(serializer.data)
api_1               |   File "/usr/local/lib/python2.7/site-packages/rest_framework/serializers.py", line 739, in data
api_1               |     ret = super(ListSerializer, self).data
api_1               |   File "/usr/local/lib/python2.7/site-packages/rest_framework/serializers.py", line 263, in data
api_1               |     self._data = self.to_representation(self.instance)
api_1               |   File "/code/api/base/serializers.py", line 988, in to_representation
api_1               |     self.child.to_representation(item, envelope=envelope) for item in data
api_1               |   File "/code/api/base/serializers.py", line 1186, in to_representation
api_1               |     representation = field.to_representation(attribute)
api_1               |   File "/usr/local/lib/python2.7/site-packages/rest_framework/fields.py", line 1754, in to_representation
api_1               |     return method(value)
api_1               |   File "/code/api/waffle_flags/serializers.py", line 20, in get_active
api_1               |     return flag_is_active(request, obj)
api_1               |   File "/usr/local/lib/python2.7/site-packages/waffle/__init__.py", line 17, in flag_is_active
api_1               |     return flag.is_active(request)
api_1               |   File "/usr/local/lib/python2.7/site-packages/waffle/models.py", line 247, in is_active
api_1               |     active_for_user = self._is_active_for_user(request)
api_1               |   File "/usr/local/lib/python2.7/site-packages/waffle/models.py", line 209, in _is_active_for_user
api_1               |     return self.is_active_for_user(request.user)
api_1               |   File "/usr/local/lib/python2.7/site-packages/waffle/models.py", line 187, in is_active_for_user
api_1               |     authed = getattr(user, 'is_authenticated', lambda: False)()
api_1               | TypeError: 'bool' object is not callable
api_1               | [04/Dec/2017 17:11:40] "GET /v2/waffle_flags/ HTTP/1.1" 500 59

```

api/nodes/views.py
pattisdr: good idea, I've added some decorators with example usages.

pattisdr:
@sloria, @brianjgeiger, per your comments, I've begun an implementation of a single waffle list endpoint `/v2/_waffle/?flags=featureA,featureB&samples=sampleA` that can be filtered against.  As this django-waffle PR is still in the exploratory phase, I'll leave it here for now until the review of this PR is finished.

website/templates/home.mako
pattisdr: Getting this wafflejs file in the flask app is the trickiest part.  I spent a long time trying to include it back in September, got stuck, and manually included it myself.

Can you tell me what this `reverse` piece is doing?   I'm not sure where to include the waffle.urls.  They are django url's, how is the flask app supposed to talk to them?

pattisdr:
okay @sloria , ready for another pass of code review!  PR is no longer exploratory, but something that can be merged.

api/base/settings/defaults.py
```diff
@@ -198,28 +199,33 @@
...
 )
 
 TEMPLATES = [
     {
         'BACKEND': 'django.template.backends.django.DjangoTemplates',
         'DIRS': [os.path.join(BASE_DIR, 'templates')],
-        'APP_DIRS': True
+        'APP_DIRS': True,
+        'OPTIONS': {
+            'context_processors': [
```
sloria: Drive-by: Do we actually need these and the above middlewares if we aren't using Django templates?

pattisdr: thanks i'll check it out, I can't remember if I put them in because it was in the docs, or if I needed it.

requirements.txt
```diff
@@ -35,6 +35,9 @@ google-api-python-client==1.6.4
 Babel==2.5.1
 citeproc-py==0.4.0
 boto3==1.4.7
+#Using own fork of django-waffle until https://github.com/jsocol/django-waffle/pull/259 is merged
```
sloria: Looks like that PR has been replaced by https://github.com/jsocol/django-waffle/pull/266 .

No need to change your fork for now, though--it's a simple change, and hopefully the fix will be released upstream soon.

website/routes.py
```diff
@@ -122,6 +123,7 @@ def get_globals():
         'recaptcha_site_key': settings.RECAPTCHA_SITE_KEY,
         'custom_citations': settings.CUSTOM_CITATIONS,
         'osf_support_email': settings.OSF_SUPPORT_EMAIL,
+        'wafflejs_url': '{api_domain}{waffle_url}'.format(api_domain=settings.API_DOMAIN[:-1] if settings.API_DOMAIN[-1] == '/' else settings.API_DOMAIN, waffle_url=reverse('wafflejs'))
```
sloria: Drive-by: You can do `settings.API_DOMAIN.rstrip('/')` to strip the trailing `/` if it exists.

website/templates/base.mako
```diff
@@ -12,6 +12,7 @@
     <meta name="description" content="${self.description()}">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="fragment" content="!">
+    <script src="${wafflejs_url}"></script>
```
sloria: Can this be loaded below the fold, i.e. in `javascript_bottom`?

sloria:
Just some drive-by comments for now. Will take a look at this more closely next week.

pattisdr:
👍 thanks for your comments!

api/base/settings/defaults.py
pattisdr: you're right, i don't think i need this piece.

api/base/settings/defaults.py
```diff
@@ -198,21 +199,20 @@
...
 
-    # 'django.contrib.sessions.middleware.SessionMiddleware',
+    'django.contrib.sessions.middleware.SessionMiddleware',
     'api.base.middleware.CorsMiddleware',
     'django.middleware.common.CommonMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
-    # 'django.contrib.auth.middleware.AuthenticationMiddleware',
-    # 'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
-    # 'django.contrib.messages.middleware.MessageMiddleware',
+    'django.contrib.auth.middleware.AuthenticationMiddleware',
```
sloria: Are these middlewares necessary if we're not using Django templates?

api/waffle/serializers.py
```diff
@@ -0,0 +1,23 @@
...
+        """
+        Use waffle_feature_is_active helper to determine if waffle flag, sample, or switch is active
+        """
+        return waffle_feature_is_active(self.context.get('request'), type(obj).__name__, obj.name)
+
+    def get_id(self, obj):
+        return '{}_{}'.format(type(obj).__name__.lower(), obj.id)
+
+    class Meta:
+        type_ = 'waffle'
```
sloria: Would it make sense to have different types for switches, samples, and flags (e.g. `waffle_flags`). That would allow the frontend to distinguish between the types.

As @brianjgeiger pointed out in https://github.com/CenterForOpenScience/osf.io/pull/7734#issuecomment-347988818 , it's ok to have an endpoint return different types.

api/base/waffle_decorators.py
```diff
@@ -0,0 +1,59 @@
...
+            'waffle_args': (instance_name,)
+        },
+        'Sample': {
+            'waffle_func': waffle.sample_is_active,
+            'waffle_args': (instance_name,)
+        }
+    }[instance_type]
+    return waffle_map['waffle_func'](*waffle_map['waffle_args'])
+
+def require_flag(flag_name):
```
sloria: Nice API! 👍

framework/auth/oauth_scopes.py
```diff
@@ -100,6 +100,8 @@ class CoreScopes(object):
 
     PROVIDERS_WRITE = 'providers_write'
 
+    WAFFLE_READ = 'waffle_read'
```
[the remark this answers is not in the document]
pattisdr: ah, thanks, I missed that

api/waffle/serializers.py
pattisdr: Yes, I can return different types.  (Side note: I am also returning the type in the `IDField`.  Samples, switches, and flags are separate models, so both a sample and a switch can have an ID of 1.  If I am fetching both a sample and a switch in an Ember app with id=1, Ember will not like the duplicate ID's.  So I am returning `flag-1` and `sample-1` as the id's.

Also, Ember-side, we will need to create a flag model, switch model, and sample model, instead of just a single waffle model.

framework/auth/oauth_scopes.py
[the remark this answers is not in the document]
pattisdr: Ember-side, for authenticating with a token (i do this in local development) we will need to be able to create tokens with WAFFLE_READ scope.

requirements.txt
```diff
@@ -35,6 +35,9 @@ google-api-python-client==1.6.4
 Babel==2.5.1
 citeproc-py==0.4.0
 boto3==1.4.7
+#Using own fork of django-waffle until https://github.com/jsocol/django-waffle/pull/266 is merged
+git+https://github.com/pattisdr/django-waffle.git
```
sloria: I think we can get off a fork now; looks like they fixed compatibility with new versions of Django: https://github.com/jsocol/django-waffle/pull/269

website/templates/base.mako
```diff
@@ -338,6 +338,7 @@
 
 <%def name="javascript_bottom()">
     ### Javascript loaded at the bottom of the page ###
+    <script src="${wafflejs_url}"></script>
```
sloria: No need to load JS we're not using. Let's leave this commented out until we have a use case for Waffle.

pattisdr:
👍

pattisdr:
Changes made @sloria

sloria:
Merged into `release/next-next-djangole`!

PR #7736, opened 2017-09-26
Feature/Remove rendundant osf/utils/auth.py [OSF-8362]

pattisdr:
## Purpose

All of the methods in osf/utils/auth.py are duplicated elsewhere in framework/auth/core.py. 

## Changes

The entirety of osf/utils/auth.py can be deleted and the few places where its methods are used in addons and models can be refactored to use the same methods imported from framework.auth instead
## Ticket

https://openscience.atlassian.net/browse/OSF-8362


sloria:

PR #7747, opened 2017-09-28
Remove subjects_acceptable from the preprint_provider serializer (unu…

binoculars:
…sed)

## Purpose

Removes unused `subjects_acceptable` field in preprint_provider serializer to make the API responses smaller and faster. This field is used under /v2/preprint_providers/<provider_id>/taxonomies/ endpoint.

## Changes

Deletes the `subjects_acceptable` field from the serializer and its tests

## Side effects

Faster and smaller API responses from the preprint_providers endpoint

### DIFF
--- a/api/preprint_providers/serializers.py
+++ b/api/preprint_providers/serializers.py
@@ -86,10 +86,6 @@ class PreprintProviderSerializer(JSONAPISerializer):
         ser.CharField(required=False, allow_null=True),
         min_version='2.0', max_version='2.3'
     )
-    subjects_acceptable = ShowIfVersion(
-        ser.ListField(required=False, default=[]),
-        min_version='2.0', max_version='2.4'
-    )
 
     class Meta:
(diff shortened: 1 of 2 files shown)


binoculars:
Negative on the breaking change. This field shouldn't be in there and there is no practical use for it. Should be considered a fix.

brianjgeiger:
It's been in there and has been advertised as a field. Removing a field from the API is a breaking change. That's why we versioned it out. I agree that it's a dumb field that shouldn't have been in, but.

binoculars:
If you all want to coordinate that, be my guest, but it's outside the scope of this PR

brianjgeiger:
Want to coordinate what?

binoculars:
Changing the API version

caseyrollins:
@binoculars, if you look at the line you deleted, it's already been versioned. If you specify v2.5 in your request, `subjects_acceptable` isn't returned in the response.

PR #7748, opened 2017-09-28
Add custom taxonomy creator in admin [OSF-8559]

erinspace:
## Purpose

Add a form in the admin page for editing a preprint provider to add a custom taxonomy. The form's fields have individual "Add" buttons, that when clicked will build the JSON for the custom taxonomy. There is an example box with more information. The form has a validate button as well as a save button, which use the existing management commands to validate and then apply the entered custom taxonomy.

## Changes
- Form, view, template for custom taxonomy form
- When listing subjects acceptable, also include the Bepress text it was mapped from if applicable
- Disable templates panel on django debug toolbar because it was slowing rendering to a halt

## Side effects
None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-8559


admin/preprint_providers/views.py
```diff
@@ -159,6 +169,41 @@ def get_success_url(self, *args, **kwargs):
         return reverse_lazy('preprint_providers:detail', kwargs={'preprint_provider_id': self.kwargs.get('preprint_provider_id')})
 
 
+class ProcessCustomTaxonomy(PermissionRequiredMixin, View):
+
+    permission_required = 'osf.view_preprintprovider'
```
sloria: Shouldn't this be `osf.change_preprintprovider`?

erinspace: Yes, yes it should!

sloria:
Merged into `release/next-djangole`

PR #7757, opened 2017-10-02
[Feature] Make emails configurable and unique [OSF-7803]

alexschiller:
## Purpose

Email addresses are currently generated using factory boy fake emails. These are great because they look like real emails, except that they sometimes can be real (woops). The original ticket was about stoping duplicate key insertions when creating users. In theory this change will also solve that problem for travis run tests, but it will not fix the underlying issue. 

## Changes
Procedurally generate email addresses based on time instead of randomly. 

## Side effects
Changed a BUNCH of places where email addresses are generated. 

## Ticket
https://openscience.atlassian.net/browse/OSF-7803

## QA
No QA needed


osf_tests/factories.py
```diff
@@ -28,6 +30,7 @@
 from addons.osfstorage.models import OsfStorageFile
 
 fake = Factory.create()
+fake_email = lambda: '{}+{}@{}'.format(FAKE_EMAIL_NAME, int(time.time() * 1000000), FAKE_EMAIL_DOMAIN)
```
brianjgeiger: Interestingly, on some systems, time.time() may resolve to the 1s precision. I presume that nothing that we'll do will do that, but it might be worth making a note here in case someone, uh, decides to run the test suite on a… windows 95 computer? A 16 bit processor? A Commodore Vic-20?

brianjgeiger: Based on https://stackoverflow.com/a/18170448 it might be worthwhile to use `time.clock()` instead. We don't really need anything that's unique between runs, we just need it unique during the run, and `time.clock()` sounds like it might be more appropriate in this case.

tests/test_views.py
```diff
@@ -2160,8 +2162,8 @@ def test_claim_user_already_registered_redirects_to_claim_user_registered(self,
     @mock.patch('website.project.views.contributor.send_claim_email')
     def test_claim_user_already_registered_secondary_email_redirects_to_claim_user_registered(self, claim_email):
         name = fake.name()
-        email = fake.email()
-        secondary_email = fake.email()
+        email = fake_email()
+        secondary_email = fake_email()
```
brianjgeiger: It would be worth checking here that the two emails are definitely different and, if not, delaying briefly and trying again. And/or failing with an assert that says that the processor speed is too fast for the clock resolution and that there may be similar problems elsewhere, possibly with a link to this ticket/pr.

tests/test_utils.py
```diff
@@ -496,3 +497,18 @@ def test_generate_csl_given_name_with_given(self):
...
         assert_equal(csl_given_name, 'Cause')
+
+
+@pytest.mark.django_db
+class TestUserFactoryConflict:
+
+    def test_build_create_user_time_conflict(self):
+        # Test that build and create user factories do not create conflicting usernames 
+        # because they occured quickly
+        user_one_build = UserFactory.build()
```
brianjgeiger: I suspect that just calling fake_email twice in a row (as done in https://github.com/CenterForOpenScience/osf.io/pull/7757/files#diff-a4f5c4d64d4811b15e1405573c00f3f4R2165 ) will be a tiny bit faster than even this. Could you add a couple of lines like that with an assert just to make sure?

PR #7763, opened 2017-10-03
[OSF-8691] Allow tags to be added to registrations

adlius:
## Purpose

This PR enables adding tags to registrations.

## Changes

There are two major parts to this PR:
1. API V2 changes to allow edits of tags.
    * Override `add_tag` and `remove_tag` functions in `Registration` to raise `NodeStateError` for withdrawn registrations.
    * Add an abstract method `remove_tag` and method to class `Taggable`.
    * Add a method `update_tags` to class `Taggable` to make use of `remove_tag`, which is implemented in sub classes, and DRY up tags update logic.
    * Change the `update` method in `BaseRegistrationSerializer` to update tags for registrations. `NodeStateError` is caught and raise `Conflict` for 409.
    * Unit tests. Some tests are now consolidated for performance. 
    * Two tests in `TestRegistrationBulkUpdate::test_bulk_update_errors` are deleted. 
        * `400 from attempting to update with empty data` is removed as update with empty data no longer causes error.
        * `400 from attempting to update with bad data` is removed. Attempting to update a registration with an attribute that cannot be updated would return 200, a behavior that is consistent with the `nodes` endpoint. This is in part due to the fact that `update` method only gets `validated_data`, which does not included invalid inputs, as param and therefore knows nothing about invalid inputs included in the request.

2. Front end changes to make use of API V2 for updating tags for nodes.

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-8691


website/static/js/pages/project-dashboard-page.js
```diff
@@ -219,44 +219,85 @@ $(document).ready(function () {
...
+                data: {
+                    type: nodeType,
+                    id: window.contextVars.node.id,
+                    attributes: {
+                        tags: newTagsList
+                    }
+                }
+            };
+
+            var request = $.ajax({
```
alexschiller: driveby: I think this could use the $osf.ajaxJSON helper instead

api_tests/registrations/views/test_registration_detail.py
```diff
@@ -289,3 +292,187 @@ def test_make_public_unapproved_registration_raises_error(self, app, user, unapp
...
+            'data': {
+                'id': registration_withdrawn._id,
+                'type': 'registrations',
+                'attributes': {
+                    'tags': ['new-tag', 'existing-tag'],
+                }
+            }
+        }
+
+    def test_registration_starts_with_no_tags(self, app, url_registration_public):
```
alexschiller: Some of these non-mutating tests could be combined to speed up runtime

api/registrations/serializers.py
```diff
@@ -275,16 +276,35 @@ def get_current_user_permissions(self, obj):
...
+        if 'tags' in validated_data:
+            current_tags = set(validated_data.pop('tags', []))
+        elif self.partial:
+            current_tags = set(old_tags)
         else:
-            raise exceptions.ValidationError('Registrations can only be turned from private to public.')
+            current_tags = set()
+
+        try:
+            for new_tag in (current_tags - old_tags):
```
brianjgeiger: Since this is essentially the same as what's in the node model, it might be good to DRY it up a bit.

brianjgeiger:
I'm not one for consolidating commits, but I'm cool if it happens safely. That being said, you should ensure that your commit message is useful for people looking through the history or the blame of the changes you've made. Generic 'consolidation' messages won't help future code fixers.

PR #7765, opened 2017-10-04
Feature/Add custom taxonomy import/export [OSF-8392]

laurenrevere:
## Purpose

Now that all taxonomies are custom taxonomies, we can easily import taxonomies from staging.

## Changes

I removed taxonomies from the excluded fields when importing a preprint provider.

## Side effects

None anticipated

## Ticket

https://openscience.atlassian.net/browse/OSF-8392

### DIFF
--- a/admin/preprint_providers/views.py
+++ b/admin/preprint_providers/views.py
@@ -18,8 +18,7 @@ from osf.models.preprint_provider import rules_to_subjects
 
 # When preprint_providers exclusively use Subject relations for creation, set this to False
 SHOW_TAXONOMIES_IN_PREPRINT_PROVIDER_CREATE = True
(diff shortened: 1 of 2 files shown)


admin/preprint_providers/views.py
```diff
@@ -249,7 +248,7 @@ def get_page_provider(self):
 
     def add_subjects(self, provider, subject_data):
         from osf.management.commands.populate_custom_taxonomies import migrate
-        migrate(provider._id, subject_data)
+        migrate(provider=provider._id, data=subject_data)
```
brianjgeiger: Yes, github, those were definitely the things that changed and the things that stayed the same.

laurenrevere: lol

PR #7766, opened 2017-10-04
Add appropriate error for accessing old SHARE API v1 search [OSF-8552]

erinspace:
## Purpose
Right now, accessing the old share v1 search route gives a 404 with no more information. Return instead a link to the new SHARE search page!

## Changes

- add new route for /api/v1/share/search/
- add error and message with right URL

## Side effects

None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-8552


sloria:

PR #7767, opened 2017-10-04
[MOD-84][Improvement] Put all published preprints in `accepted` state

aaxelb:
## Purpose
When a preprint provider is set up to use a moderation workflow, any existing published preprints should remain publicly visible.

## Changes
* Add a migration to update all published preprints, setting `reviews_state` to `'accepted'` and `date_last_transitioned` equal to `date_published`
* When a preprint is published via `set_published`, set `reviews_state` to `'accepted'` and `date_last_transitioned` equal to `date_published` 

## Side effects
None

## Ticket
https://openscience.atlassian.net/browse/MOD-84


laurenbarker:
@aaxelb couple tests need to be updated.

aaxelb:
@laurenbarker Sorry about that, should be fixed!

laurenbarker:
@aaxelb, looks like another test issue

PR #7770, opened 2017-10-04
[OSF-8551] Supposedly fix redirection issue for logins from institution landing pages

adlius:
## Purpose

This PR should fix login redirection issues of institution landing pages.

## Changes

Currently, when a user go to an institution landing page using custom url like `osf.ucsd.edu` instead of `osf.io/institutions/ucsd` and click on the `'Sign In'` button, the user would be redirected to the OSF dashboard page instead of the institution landing page.

Before detailing the changes, some brief explanation on how CAS works is needed for better understanding of the issue.

Normally, when a user tries to login through CAS, he should click on a login button and the web app would take him to the cas server with URL like `<cas-server-address>/login?service=<absolute-url-of-the-web-app-page>`. User would then complete the authentication process on that page and, upon success, would be redirected to the web app page previously specified after `service=`. (The details of how authentication is done between the CAS server and the web app server is not relevant.)

However, in this case, probably because CAS session cookies only works for a single domain (osf.io), users trying to login from `osf.ucsd.edu` would first be redirected to `osf.io/login?campaign=institution&redirect_url=/institutions/ucsd` (See [this line](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/templates/nav.mako#L69) in `nav.mako`) before being redirected again to `<cas-server-address>/login?service=<absolute-url-of-the-web-app-page>`. This is where the problem start.

First, the `auth_login` view method needs to catch the redirection url from the request. However, as shown in `auth_login`, the parameter name should be `next` instead of `redirect_url`. Also the value of the parameter should be an absolute url, otherwise the CAS server would not know to which web app the user should be authenticated and redirected to, since CAS is not internal to OSF but serves multiple web apps.

**Therefore, in `nav.mako`, the parameter name is changed from `redirect_url` to `next`. And in `website/views.py`, an absolute url for redirection is constructed instead of a relative one.**

Then, the view method would pass the redirect url to `login_and_register_handler` to construct and redirect user to the correct CAS server url for authentication. But the handler would automatically reset the redirect url to the dashboard page if `campaign=institution` is specified in the request.

**Hence, in `auth/views.py`, a conditional is added to check if the redirection url is `None` before resetting.**

## Side effects

I can't verify that the entire login process for institution landing pages works as expected after the change as I can only view the landing pages locally using an OSF url, not a custom one. However I can verify that the `auth_login` view would now redirect to the CAS server url with the correct service url. Everything should work well after that since the CAS server works properly.

## Ticket

https://openscience.atlassian.net/browse/OSF-8551


brianjgeiger:
@adlius Like the other ticket, if you have a chance this weekend, please merge conflicts.

adlius:
Merge conflict fixed.

PR #7344, opened 2017-06-15
[WIP] [Feature] OSF/API change for the new CAS and Account Management [OSF-8289][CAS-47]

cslzchen:
Note: the description will be updated in the correct format later.

Here is the [document](https://docs.google.com/document/d/1YPXHTLjNMQijKF8lyyhQh89f5UktovorI3S_uyLeD_k/edit#).

Here is the [ticket](https://openscience.atlassian.net/browse/OSF-8289).

Here is the CAS [ticket](https://openscience.atlassian.net/browse/CAS-47) and [PR](https://github.com/CenterForOpenScience/cas-overlay/pull/64).

PR #7773, opened 2017-10-04
Feature/Port Move and Copy Hooks to New API [OSF-8632]

pattisdr:
## Purpose
The code for the osfstorage_copy_hook and osfstorage_view_hook views are messy, returning different status codes depending on whether the target file is checked out, is a preprint file, or is a quick file. Also, v1 is in the past. Let's move forward to v2.

## Changes
- Added new endpoints, nested under `_`:
```python
/_/wb/hooks/<node_id>/move/
/_/wb/hooks/<node_id>/copy/
```
- Now if there's an issue with moving/copying checked out files, preprint files or files where filename already exist, a 400 error is returned
- Added new API views `MoveFileMetadataView` (Returns 200 OK) and `MoveFileMetadataCopy` (Returns 201 Created)
  - View extends generics.CreateAPIView
  - View is undocumented. Only waterbutler should be calling this view, this view shouldn't be called directly
  - WB should send a signed payload, identical to payload currently being sent to v1
- Added new serializer `WaterbutlerMetadataSerializer ` which extends the Serializer class
  - Two write only fields - source/destination expected.
  - Response is very similar to existing response returned in[ v1 osfstorage_move_hook](https://github.com/CenterForOpenScience/osf.io/blob/d2beedc810c56a4ae7be16c8fc0b4be79a555a60/addons/osfstorage/views.py#L94)
- Added a lot of tests similar to the[ tests written for the original hooks ](https://github.com/CenterForOpenScience/osf.io/blob/develop/addons/osfstorage/tests/test_views.py#L661)
- DID NOT remove the old hooks/tests.  Leaving them until the WB ticket is complete and then the old code can be removed.

## Side effects

None - old views/routes have not been deleted.

## Ticket

https://openscience.atlassian.net/browse/OSF-8632


api_tests/nodes/views/test_node_provider_file_metadata_create.py
```diff
@@ -0,0 +1,350 @@
...
+@pytest.fixture()
+def user():
+    return AuthUserFactory()
+
+
+@pytest.mark.django_db
+class TestNodeProviderFileMetadataCreateView:
+
+    @pytest.fixture()
+    def node(self, user):
```
alexschiller: Prefer `<thing it is>_<descriptor/number>` syntax. Most of the newer fixtures use written numbers instead of digits 

If there is more than one fixture, best to put a number on both:
`node_one`
`node_two`

api_tests/nodes/views/test_node_provider_file_metadata_create.py
```diff
@@ -0,0 +1,350 @@
...
+    @pytest.fixture()
+    def node2(self, user):
+        return ProjectFactory(creator=user)
+
+    @pytest.fixture()
+    def file(self, user, node):
+        return api_utils.create_test_file(node, user, create_guid=False)
+
+    @pytest.fixture()
+    def root(self, node):
```
alexschiller: `root_node_one`

api_tests/nodes/views/test_node_provider_file_metadata_create.py
```diff
@@ -0,0 +1,350 @@
...
+
+    @pytest.fixture()
+    def create_metadata_url(self, node):
+        return '/{}nodes/{}/files/providers/osfstorage/file_metadata/'.format(API_BASE, node._id)
+
+    @pytest.fixture()
+    def move_payload(self, file, folder):
+        return {
+            "data": {
+                "type": "file_metadata",
```
alexschiller: prefer single over double quotes unless the string contains single quotes

api_tests/nodes/views/test_node_provider_file_metadata_create.py
```diff
@@ -0,0 +1,350 @@
...
+    @pytest.fixture()
+    def node(self, user):
+        return ProjectFactory(creator=user)
+
+    @pytest.fixture()
+    def node2(self, user):
+        return ProjectFactory(creator=user)
+
+    @pytest.fixture()
+    def file(self, user, node):
```
alexschiller: very, very minor, especially since I know this came from elsewhere: `file` is a `type` in python.

api_tests/nodes/views/test_node_provider_file_metadata_create.py
```diff
@@ -0,0 +1,350 @@
...
+        return api_utils.create_test_file(node, user, create_guid=False)
+
+    @pytest.fixture()
+    def root(self, node):
+        return node.get_addon('osfstorage').get_root()
+
+    @pytest.fixture()
+    def node2_root(self, node2):
```
alexschiller: `root_node_two`

api_tests/nodes/views/test_node_provider_file_metadata_create.py
```diff
@@ -0,0 +1,350 @@
...
+
+@pytest.mark.django_db
+class TestNodeProviderFileMetadataCreateView:
+
+    @pytest.fixture()
+    def node(self, user):
+        return ProjectFactory(creator=user)
+
+    @pytest.fixture()
+    def node2(self, user):
```
alexschiller: `node_two`

api_tests/nodes/views/test_node_provider_file_metadata_create.py
alexschiller: `file_one`

api_tests/nodes/views/test_node_provider_file_metadata_create.py
```diff
@@ -0,0 +1,350 @@
...
+        assert len(root.children) == 2
+        res = app.post_json_api(create_metadata_url, copy_payload, auth=user.auth)
+        assert res.status_code == 201
+        assert len(folder.children) == 1
+        assert len(root.children) == 2
+        assert res.json['data']['id'] == folder.children[0]._id
+        assert res.json['data']['attributes']['name'] == file.name
+
+    def test_copy_file_from_subfolder_to_root(self, app, user, create_metadata_url, copy_payload, folder, root):
+        file2 = folder.append_file("Second file")
```
alexschiller: `file_two`

api_tests/nodes/views/test_node_provider_file_metadata_create.py
```diff
@@ -0,0 +1,350 @@
...
+    AuthUserFactory,
+    ProjectFactory,
+    UserFactory,
+)
+from osf.models import QuickFilesNode
+from website.util import permissions
+
+
+@pytest.fixture()
+def user():
```
alexschiller: Fine to be general, but with only one class it can reasonable go in that class

pattisdr:
thanks @alexschiller, i'll make these changes shortly

icereval:
@pattisdr, probably don't want to mix the special WB endpoints in with the existing API that is end user facing, @cslzchen 's CAS PR might provide some insight + it has a middleware for JWE+JWT @ https://github.com/CenterForOpenScience/osf.io/pull/7344

pattisdr:
@sloria, latest round of changes made, wb can now send same signed request to new API endpoint.  Failing tests indicate a stalled build - i don't think it's related.

api/base/parsers.py
```diff
@@ -208,3 +210,27 @@ def flatten_relationships(self, relationships):
         if ret.get('target_type') and ret.get('id'):
             return {related_resource: ret['id']}
         return ret
+
+
+class HMACSignedParser(JSONParser):
+
+    def parse(self, stream, media_type=None, parser_context=None):
+        """
+        Parses the incoming bytestream as JSON and returns the resulting data.
```
sloria: Minor: this docstring could be more precise. How about

```
Parses the incoming bytestream as JSON. Validates the 'signature' in the payload then
returns the resulting data.
```

api/wb/views.py
```diff
@@ -0,0 +1,89 @@
...
+            AbstractNode,
+            node_id,
+            self.request,
+            display_name='node'
+        )
+        if node.is_registration:
+            raise NotFound
+        return node
+
+    def perform_create(self, serializer, action):
```
sloria: It's an antipattern to change a method's signature in a subclass if you can avoid it (because you're breaking an API contract).

Better to use a new method name. Maybe `perform_file_action`?

api_tests/base/test_views.py
```diff
@@ -36,6 +37,8 @@
     for patt in urlpatterns:
         VIEW_CLASSES.append(patt.callback.cls)
 
+EXCLUDED_VIEWS = [MoveFileMetadataView, CopyFileMetadataView]
```
sloria: Minor: since this only applies to `TestApiBaseViews`, it might make sense for this to be a class variable instead of a top-level variable.

api_tests/wb/views/test_wb_hooks.py
```diff
@@ -0,0 +1,608 @@
...
+    AuthUserFactory,
+    ProjectFactory,
+    UserFactory,
+)
+from api_tests.utils import create_test_file
+from osf.models import QuickFilesNode
+
+
+@pytest.mark.django_db
+class HookTestCase():
```
sloria: Now that we're using pytest fixtures, we can usually avoid inheritance in our tests (which can make tests more complicated than they need to be).

So instead of having `HookTestCase`, you can turn the method fixtures below into module-level function fixtures. `sign_payload` can also be a function instead of a method. Then `HookTestCase` can go away!

api/wb/views.py
```diff
@@ -0,0 +1,89 @@
...
+
+from osf.models import AbstractNode
+from addons.osfstorage.models import OsfStorageFileNode, OsfStorageFolder
+from api.base.utils import get_object_or_error
+from api.base.parsers import HMACSignedParser
+from api.wb.serializers import (
+    WaterbutlerMetadataSerializer
+)
+
+class FileMetadataMixin(generics.CreateAPIView):
```
sloria: I'm not sure `CreateAPIView` is the right generic class. It's also non-intuitive that `WaterbutlerMetadataSerializer.create` doesn't actually create anything and contains logic for both copying and moving.

We can improve the design by putting the different actions in the different view classes.

Ideally it would look something like

```python
class MoveFileMetadataView(FileMetadataView):
    def perform_file_action(self, source, destination, name):
        return source.move_under(destination, name)

class CopyFileMetadataView(FileMetadataView):
    def perform_file_action(self, source, destination, name):
        return source.copy_under(destination, name)
```

The hypothetical `FileMetadataView` would do the work of parsing the payload returning the response, etc.

pattisdr:
Thanks for the review @sloria!

pattisdr:
Changes made @sloria

api/base/parsers.py
```diff
@@ -208,3 +210,27 @@ def flatten_relationships(self, relationships):
...
+
+class HMACSignedParser(JSONParser):
+
+    def parse(self, stream, media_type=None, parser_context=None):
+        """
+        Parses the incoming bytestream as JSON. Validates the 'signature' in the payload then returns the resulting data.
+        """
+        data = super(HMACSignedParser, self).parse(stream, media_type=media_type, parser_context=parser_context)
+
+        try:
```
sloria: Minor: it would be useful to know that this code was copied from `must_be_signed`. Maybe add a code comment above `must_be_signed` to remove it once Waterbutler is sending requests to the v2 endpoints.

api/wb/views.py
```diff
@@ -0,0 +1,96 @@
...
+class FileMetadataView(APIView):
+    """
+    Mixin with common code for WB move/copy hooks
+    """
+    parser_classes = (HMACSignedParser,)
+    serializer_class = WaterbutlerMetadataSerializer
+    view_category = 'wb'
+    node_lookup_url_kwarg = 'node_id'
+
+    # Overrides CreateAPIView
```
sloria: Minor: This comment can be removed.

api/wb/views.py
```diff
@@ -0,0 +1,95 @@
...
+        return self.get_node(self.kwargs[self.node_lookup_url_kwarg])
+
+    def get_node(self, node_id):
+        node = get_object_or_error(
+            AbstractNode,
+            node_id,
+            self.request,
+            display_name='node'
+        )
+        if node.is_registration:
```
sloria: Why was this check added? I don't see it in the original v1 endpoints.

api/wb/views.py
```diff
@@ -0,0 +1,95 @@
...
+    def get_serializer_context(self):
+        """
+        Extra context provided to the serializer class.
+        """
+        return {
+            'view': self
+        }
+
+    def post(self, request, *args, **kwargs):
+        serializer = WaterbutlerMetadataSerializer(data=request.data, context=self.get_serializer_context())
```
sloria: Should this use `self.serializer_class`?

api/wb/views.py
pattisdr: [osfstorage_move_hook](https://github.com/CenterForOpenScience/osf.io/blob/develop/addons/osfstorage/views.py#L128) has a waterbutler_opt_hook decorator on it - [there's a must_not_be_registration check there](https://github.com/CenterForOpenScience/osf.io/blob/develop/addons/osfstorage/decorators.py#L66)

sloria: Ah ok, I missed that. Thanks.

sloria: Looks like the decorator checks archiving status:

```python
        if node.is_registration and not node.archiving:
            raise HTTPError(
                http.BAD_REQUEST,
                data={
                    'message_short': 'Registrations cannot be changed',
                    'message_long': "The operation you're trying to do cannot be applied to registered projects, which are not allowed to be changed",
                }
            )
```

Also, a 400 response is returned instead of a 404. It may make sense to keep it as a 400. That would allow the client (WB) too differentiate between request for an invalid node GUID and a request for an archived registration.

api_tests/wb/views/test_wb_hooks.py
```diff
@@ -0,0 +1,598 @@
...
+            'user': user._id,
+            'destination': {
+                'parent': folder_two._id,
+                'node': folder_two.node._id,
+                'name': folder_two.name,
+            }
+        })
+        res = app.post_json(move_url, signed_payload, expect_errors=True)
+        assert res.status_code == 400
+        assert 'Cannot move file as it is the primary file of preprint.' in res._app_iter[0]
```
sloria: Can this be `res.json['errors'][0]['detail']`?

api/wb/views.py
```diff
@@ -0,0 +1,95 @@
...
+    def get_serializer_context(self):
+        """
+        Extra context provided to the serializer class.
+        """
+        return {
+            'view': self
+        }
+
+    def post(self, request, *args, **kwargs):
+        serializer = self.serializer_class
```
sloria: You'll still need to instantiate the serializer like you were doing before, except without the hardcoded class:

```python
serializer = self.serializer_class(....)
```

sloria: Made this change in deaa1a4

pattisdr: thank you for fixing for me @sloria!

PR #7776, opened 2017-10-05
[hotfix] Add bulk registration update to allow embargos to be terminated early [OSF-8751]

alexschiller:
## Purpose
Embargos are trying to terminate using then /nodes/ route which previously worked because of AbstractNode. Since it was changed to just Node, registrations can no longer bulk update privacy. 

## Changes
1. Add Bulk updating to the registration api
2. Modify the privacy function to accept a `type` of node or registration
3. Add a boat load of tests for bulk updates

## Side effects
None known

## Ticket

https://openscience.atlassian.net/browse/OSF-8751

## QA Notes
1. Confirm that you can still make projects (and their components) public/private (both by making the parent public and the nested components)
2. Ensure that making an embargo'd registration public works as expected
end_embargo_early

### DIFF
--- a/api/registrations/views.py
+++ b/api/registrations/views.py
@@ -1,9 +1,10 @@
 from rest_framework import generics, permissions as drf_permissions
-from rest_framework.exceptions import ValidationError, NotFound
+from rest_framework.exceptions import ValidationError, NotFound, PermissionDenied
 from framework.auth.oauth_scopes import CoreScopes
 
 from osf.models import AbstractNode, Registration
 from api.base import permissions as base_permissions
+from api.base import generic_bulk_views as bulk_views
 from api.base.filters import ListFilterMixin
 from api.base.views import JSONAPIBaseView, BaseContributorDetail, BaseContributorList, BaseNodeLinksDetail, BaseNodeLinksList, WaterButlerMixin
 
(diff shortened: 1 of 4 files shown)


api_tests/registrations/views/test_registration_list.py
```diff
@@ -837,6 +840,217 @@ def test_cannot_register_draft_that_has_already_been_approved(self, app, user, p
...
+
+    @pytest.fixture()
+    def bad_payload(self, registration_one, registration_two):
+        return {
+            'data': [
+                {
+                    'id': registration_one._id,
+                    'type': 'registrations',
+                    'attributes': {
+                        'title': 'Nerds',
```
brianjgeiger: I would say do one valid first and an invalid second, so make this set the public while the next one sets the title. This should verify that a) the title didn't get set, and b) that the privacy setting did not change.

api/registrations/views.py
```diff
@@ -155,6 +156,16 @@ class RegistrationList(JSONAPIBaseView, generics.ListAPIView, NodesFilterMixin):
     ordering = ('-date_modified',)
     model_class = Registration
 
+    # overrides BulkUpdateJSONAPIView
+    def get_serializer_class(self):
+        """
+        Use RegistrationDetailSerializer which requires 'id'
+        """
+        if self.request.method in ('PUT', 'PATCH'):
```
sloria: This is fine for now, but maybe change to `self.request.method not in permissions.SAFE_METHODS` in the future.

alexschiller: Good to know :+1:. Was basically copypasta from nodelist https://github.com/CenterForOpenScience/osf.io/blob/develop/api/nodes/views.py#L307

brianjgeiger: I don’t think safe_methods is accurate in this case. Post, get, or delete would go to one branch, and put and patch would go to another.

PR #7778, opened 2017-10-05
[Feature][MOD-128] Update email notifications for new submissions

sheriefvt:
## Purpose

Don't add the confirmation email to the Notification Digest, just send it. Keep the status updates and comments the same.
**Why**:
Users shouldn't be able to unsubscribe from confirmation emails and they should receive the email instantly.

## Changes

- Add new separate signal with handler for new preprint submissions
- Update existing signal/code
- Update the test

## Ticket

https://openscience.atlassian.net/browse/MOD-128


aaxelb:
@sheriefvt Could you resolve the conflicts?

sheriefvt:
@aaxelb done 🔧

reviews/models/mixins.py
```diff
@@ -265,3 +261,16 @@ def reviews_notification(self, context, notify_submit=False):
                     node_lineage=node_lineage_ids
                 )
                 digest.save()
+
+# Handle email notifications for a new submission.
+@reviews_signals.reviews_email_submit.connect
+def reviews_submit_notification(self, context):
+    template = ''.join(context.get('template'))
```
aaxelb: Why the `''.join()` instead of just `template = context['template']`?

reviews/models/mixins.py
```diff
@@ -265,3 +261,16 @@ def reviews_notification(self, context, notify_submit=False):
...
+    for user_id in context.get('email_recipients'):
+        user = OSFUser.load(user_id)
+        if user == context.get('reviewable').node.creator:
```
aaxelb: Simplified: `context['is_creator'] = user == context['reviewable'].node.creator`

reviews/models/mixins.py
```diff
@@ -265,3 +261,16 @@ def reviews_notification(self, context, notify_submit=False):
                     node_lineage=node_lineage_ids
                 )
                 digest.save()
+
+# Handle email notifications for a new submission.
+@reviews_signals.reviews_email_submit.connect
+def reviews_submit_notification(self, context):
+    template = ''.join(context.get('template'))
+    for user_id in context.get('email_recipients'):
```
aaxelb: If a key will always be in a dictionary, if the key missing would be a programming error, I'd prefer to use `[]` instead of `get` -- the exception raised is much easier to understand/debug than when it tries to do something with `None` later on.

If the key could reasonably be missing without error, then use `get` and handle the possible `None`.

PR #7779, opened 2017-10-06
[OSF-8237] citation preview available even if individual fields are not filled

adlius:
## Purpose

Enable citation preview on user profile setting page when only full name is specified.

## Changes

Originally, the "impute" api v1 endpoint would only be called whenever the user hits "Auto-fill" button to auto-fill the specific names. However, if we want to show citation preview using only the full name information, we have to impute the full name upon page load and on changes and then get the corresponding MLA/APA citation previews.

Therefore, in `profile.js`, four more knockout observables are added. The `impute` function is changed to be invoked on page load and subsequent changes and store the imputed names to the added observables. An `autoFill` function is added to set the specific name fields from imputed name observables when user click the button. `citeMla` and `citeApa` is changed to user imputed names when specific name fields are not filled.

## QA Notes

The citation preview would always show that of the full name unless both Family name and Given Name are specified. This is to match the actual citation as shown in project page. Additional tests should be taken to make sure under all circumstances the preview would always match the actual citation of the project page.

## Ticket

https://openscience.atlassian.net/browse/OSF-8237

### DIFF
--- a/website/static/js/profile.js
+++ b/website/static/js/profile.js
@@ -478,7 +478,7 @@ var NameViewModel = function(urls, modes, preventUnsaved, fetchCallback) {
     };
 
     self.hasDetail = ko.computed(function() {
-        return !! self.given();
+        return !! (self.given() && self.family());
     });
 
     self.citeApa = ko.computed(function() {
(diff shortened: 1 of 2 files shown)


adlius:
Then I could change `hasDetail` to check whether both Given and Family names are provided. If not, then it would default to show preview of full name. This should be an easy fix.

alexschiller:
Cool, sounds good. Might add QA notes to confirm that in all iterations the citation name matches an actual citation on a project. :+1:

PR #7785, opened 2017-10-09
[OSF-8710] File title is made editable in file detail page

adlius:
## Purpose

This PR makes file title editable in file detail page.

## Changes

In `file-page.js`, the file title element is made editable by using the bootstrap-editable library. 

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-8710

### DIFF
--- a/website/static/js/pages/file-page.js
+++ b/website/static/js/pages/file-page.js
@@ -58,4 +58,42 @@ $(function() {
         });
     }
 
+    if(window.contextVars.currentUser.canEdit && !window.contextVars.node.isRegistration) {
+        $('#fileTitleEditable').editable({
+            type: 'text',
+            mode: 'inline',
+            send: 'always',
+            url: window.contextVars.file.urls.delete,
+            ajaxOptions: {
+                type: 'post',
+                contentType: 'application/json',
+                dataType: 'json',
+                beforeSend: $osf.setXHRAuthorization
+            },
+            validate: function(value) {
+                if($.trim(value) === ''){
+                    return 'The file title cannot be empty.';
+                } else if(value.length > 100){
+                    return 'The file title cannot be more than 100 characters.';
+                }
+            },
+            params: function(params) {
+                var payload = {
+                    action: 'rename',
+                    rename: params.value,
+                };
+                return JSON.stringify(payload);
+            },
+            success: function(response) {
+                window.location.reload();
+            },
+            error: function (response) {
+                if (response.status === 409){
+                    $osf.growl('Cannot rename file.', 'An item with the same name already exists in this location.')
+                } else {
+                    return $osf.handleEditableError
+                }
+            }
+        });
+    }
 });
--- a/website/static/js/pages/file-page.js
+++ b/website/static/js/pages/file-page.js
@@ -58,7 +58,16 @@ $(function() {
         });
     }
 
-    if(window.contextVars.currentUser.canEdit && !window.contextVars.node.isRegistration) {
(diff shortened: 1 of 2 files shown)


website/static/js/pages/file-page.js
```diff
@@ -58,4 +58,42 @@ $(function() {
...
+                return JSON.stringify(payload);
+            },
+            success: function(response) {
+                window.location.reload();
+            },
+            error: function (response) {
+                if (response.status === 409){
+                    $osf.growl('Cannot rename file.', 'An item with the same name already exists in this location.');
+                } else {
+                    return $osf.handleEditableError;
```
alexschiller: Pretty sure this isn't going to render properly-- handleEditableError doesn't growl.
 
Probably needs something like: `$osf.growl($osf.handleEditableError(response));` (no guarantees that will work)

website/static/js/pages/file-page.js
```diff
@@ -58,4 +58,42 @@ $(function() {
         });
     }
 
+    if(window.contextVars.currentUser.canEdit && !window.contextVars.node.isRegistration) {
```
alexschiller: At a minimum this should handle checkouts-- iirc disabling the button if there is a `window.contextVars.file.checkoutUser` would stop bad things from happening.

Also, multiple addons do not allow renames (iirc: dataverse, maybe published figshares, bitbucket is read-only). You should be able to access context vars `window.contextVars.file.provider` to disable the button for those providers easily enough.

brianjgeiger: For this ticket, we're going to let the API fail, but there's a follow-up ticket to more gracefully handle these things in the future. https://openscience.atlassian.net/browse/OSF-8776

website/static/js/pages/file-page.js
brianjgeiger: It growled in the demo.

adlius:
Disallow title of checkout files and files of dataverse/bitbucket/figshare to be edited. Changed error rendering to show error message directly underneath the editable field.

website/static/js/pages/file-page.js
```diff
@@ -58,4 +58,52 @@ $(function() {
         });
     }
 
+    var titleEditable = function () {
+        var readOnlyProviders = ['bitbucket', 'figshare', 'dataverse'];
+        var ctx = window.contextVars;
+        if (readOnlyProviders.includes(ctx.file.provider) || ctx.file.checkoutUser || !ctx.currentUser.canEdit || ctx.node.isRegistration)
```
brianjgeiger: @alexschiller Is this all correct? What if the checkoutUser is the current user? Can they then check out the file? Also, does the primary preprint file also check out the file, or is that considered a bit different?

brianjgeiger: On the plus side, it'll still error even if we aren't catching all of the conditions, but I think that the current checkout user may worth checking as possibly overly aggressive. I believe the user who has currently checked out the file can edit it.

alexschiller: @brianjgeiger Looks right to me-- Via fangorn, the only two actions allowed for a checkout user are uploading a new version and checking in the file. [The rename button does not show up if the file is checked out.](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/static/js/fangorn.js#L1918)

The configs for bitbucket, figshare, and dataverse don't write a rename button at all. 

So all in all, AFAICT this matches fangorn functionality. Whether the fangorn behavior is desirable or not is questionable.

PR #7787, opened 2017-10-09
[Improvement][MOD-132] Update email html templates

sheriefvt:
## Purpose

Update HTML templates for update comments and status change notifications. 
## Changes

Remove inheritance from notify_base.mako which has the base template (contains header and footer).  

### Reason 
Those two types of notifications are stored into the notification digest and later use the digest.html.mako, which already has both the header and footer, causing duplication in content. 
New submission notification is not stored in the notification digest (not using the digest.html.mako) template, therefore the inheritance from notify_base.mako is kept.

## Demo
https://jsfiddle.net/sherifvt/yvhpx9d3/
https://jsfiddle.net/sherifvt/3b4rap7y/

## Ticket
https://openscience.atlassian.net/browse/MOD-132

### DIFF
--- a/website/templates/emails/reviews_update_comment.html.mako
+++ b/website/templates/emails/reviews_update_comment.html.mako
@@ -1,26 +1,23 @@
 ## -*- coding: utf-8 -*-
-<%inherit file="notify_base.mako"/>
-<%def name="content()">
-    <div style="margin: 40px;">
-        Hello ${user.fullname},
-        <br><br>
-        Your ${reviewable.provider.preprint_word} "${reviewable.node.title}" has an updated comment by the moderator. To view the comment, go to your <a href="${reviewable.absolute_url}">${reviewable.provider.preprint_word}</a>.
-        <br><br>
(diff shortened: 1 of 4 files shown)


laurenbarker:
@sheriefvt, would you be willing to change the double `<br>`s to `<p>`? I originally put them there because I was copying other templates but `<br>` isn't the appropriate element to use for spacing (h/t @aaxelb from the OG PR). I also forgot to add updating the `<br>`s to the ticket...

sheriefvt:
@laurenbarker yes sure will do

website/templates/emails/reviews_update_comment.html.mako
```diff
@@ -1,26 +1,18 @@
...
+    <p>You will ${'not receive ' if no_future_emails else 'be automatically subscribed to '}future notification emails for this ${reviewable.provider.preprint_word}. Each ${reviewable.provider.preprint_word} is associated with a project on the Open Science Framework for managing the ${reviewable.provider.preprint_word}. To change your email notification preferences, visit your <a href="${domain + 'settings/notifications/'}">project user settings</a>.</p>
+    <p>If you have been erroneously associated with "${reviewable.node.title}", then you may visit the project's "Contributors" page and remove yourself as a contributor.</p>
+    <p>For more information about ${reviewable.provider.name}, visit <a href="${provider_url}">${provider_url}</a> to learn more. To learn about the Open Science Framework, visit <a href="https://osf.io/">https://osf.io/</a>.</p>
+    <p>For questions regarding submission criteria, please email ${provider_contact_email}</p>
+    <br>
+    Sincerely,
+    <br>
+    Your ${reviewable.provider.name} and OSF teams
+    <br><br>
```
aaxelb: Could you wrap the address in `<p>` too, instead of this double `<br>`?

PR #7793, opened 2017-10-10
[MOD-110][Fix] Allow filtering preprints by the node's is_public

aaxelb:
## Purpose
Allow excluding preprints with private nodes from the reviews UI.

## Changes
Allow filtering with `filter[node_is_public]=true` on all preprint list endpoints.

Exclude actions on preprints with private nodes from `/users/me/actions/`

## Side effects
None

## Ticket
https://openscience.atlassian.net/browse/MOD-110

### DIFF
--- a/api_tests/reviews/mixins/filter_mixins.py
+++ b/api_tests/reviews/mixins/filter_mixins.py
@@ -55,8 +55,8 @@ class ActionFilterMixin(object):
     @pytest.fixture()
     def all_actions(self, providers):
         actions = []
-        for p in providers:
-            preprint = PreprintFactory(provider=p, project=ProjectFactory(is_public=True))
+        for provider in providers:
+            preprint = PreprintFactory(provider=provider, project=ProjectFactory(is_public=True))
             for _ in range(5):
                 actions.append(ActionFactory(target=preprint))
         return actions
--- a/api_tests/reviews/mixins/filter_mixins.py
+++ b/api_tests/reviews/mixins/filter_mixins.py
@@ -8,6 +8,7 @@ from osf_tests.factories import (
     AuthUserFactory,
     PreprintFactory,
     PreprintProviderFactory,
+    ProjectFactory,
 )
 from reviews.permissions import GroupHelper
 
@@ -55,7 +56,7 @@ class ActionFilterMixin(object):
     def all_actions(self, providers):
         actions = []
         for p in providers:
-            preprint = PreprintFactory(provider=p)
+            preprint = PreprintFactory(provider=p, project=ProjectFactory(is_public=True))
             for _ in range(5):
                 actions.append(ActionFactory(target=preprint))
         return actions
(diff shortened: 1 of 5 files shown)


api_tests/reviews/mixins/filter_mixins.py
```diff
@@ -67,7 +68,7 @@ def allowed_providers(self, providers):
     @pytest.fixture()
     def expected_actions(self, all_actions, allowed_providers):
         provider_ids = set([p.id for p in allowed_providers])
-        return [l for l in all_actions if l.target.provider_id in provider_ids]
+        return [a for a in all_actions if a.target.provider_id in provider_ids]
```
laurenbarker: Would this be clearer if `a` --> `action`?

api_tests/users/views/test_user_actions.py
```diff
@@ -21,6 +21,14 @@ class TestActionFilters(ActionFilterMixin):
     def url(self):
         return '/{}users/me/actions/'.format(API_BASE)
 
+    @pytest.fixture()
+    def expected_actions(self, all_actions, allowed_providers):
+        expected = super(TestActionFilters, self).expected_actions(all_actions, allowed_providers)
+        node = expected[0].target.node
+        node.is_public = False
+        node.save()
+        return [a for a in expected if a.target.node.is_public]
```
laurenbarker: Same here, `a` --> `action`

api_tests/reviews/mixins/filter_mixins.py
```diff
@@ -55,7 +56,7 @@ def providers(self):
     def all_actions(self, providers):
         actions = []
         for p in providers:
```
laurenbarker: I'm not a huge fan of one letter variables but ultimately it's up to you.

api_tests/users/views/test_user_actions.py
```diff
@@ -21,6 +21,14 @@ class TestActionFilters(ActionFilterMixin):
...
+
     def test_no_permission(self, app, url, expected_actions):
```
laurenbarker: What was `expected_actions` before? (I can't seem to figure out where is came from)

laurenbarker: Where did `expected_actions` come from before if it wasn't defined in this file or imported? I feel like I'm missing something...

api_tests/reviews/mixins/filter_mixins.py
aaxelb: I think I'm good with one-letter names in list comprehensions, but I agree with you everywhere else.

api_tests/users/views/test_user_actions.py
aaxelb: It's a pytest fixture [defined on the parent class, `ActionFilterMixin`](https://github.com/aaxelb/osf.io/blob/f549e490357e0b5052e75408acbf7e03eb851aed/api_tests/reviews/mixins/filter_mixins.py#L69). It's the subset of actions that should be visible to the user in the api. The overriding method calls `super()` and filters it further to exclude actions on preprints with private nodes.

api_tests/users/views/test_user_actions.py
aaxelb: I changed `expected` to `actions` to hopefully make it clearer?

api_tests/reviews/mixins/filter_mixins.py
aaxelb: I agree here.

PR #7797, opened 2017-10-11
[MOD-138][Improvement] Allow contributors to see unpublished preprints

laurenbarker:
## Purpose

Allow contributors to see unpublished preprints in the review process.

## Changes

* Contributors can see the preprint detail if provider uses reviews and the preprint is not in the initial state.
* Contributors can see preprint in preprint list if provider uses reviews and the preprint is not in the initial state. h/t @alexschiller
* Add tests

## Ticket

https://openscience.atlassian.net/browse/MOD-138

### DIFF
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -505,10 +505,8 @@ class PreprintFilterMixin(ListFilterMixin):
             admin_user_query = Q(node__contributor__user_id=auth_user.id, node__contributor__admin=True)
             reviews_user_query = Q(node__is_public=True, provider__in=get_objects_for_user(auth_user, 'view_submissions', PreprintProvider))
             if allow_contribs:
-                contrib_query_base = Q(node__contributor__user_id=auth_user.id, node__contributor__read=True)
-                not_reviews_contrib_query = contrib_query_base & Q(is_published=True)
-                reviews_contrib_query = contrib_query_base & (Q(provider__reviews_workflow__isnull=False) & ~Q(reviews_state=States.INITIAL.value))
-                query = default_query & (no_user_query | not_reviews_contrib_query | reviews_contrib_query | admin_user_query | reviews_user_query)
+                contrib_user_query = ~Q(reviews_state=States.INITIAL.value) & Q(node__contributor__user_id=auth_user.id, node__contributor__read=True)
+                query = default_query & (no_user_query | contrib_user_query | admin_user_query | reviews_user_query)
             else:
                 query = default_query & (no_user_query | admin_user_query | reviews_user_query)
         else:
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -20,6 +20,7 @@ from rest_framework import serializers as ser
 from rest_framework.filters import OrderingFilter
 from osf.models import Subject, PreprintProvider
 from osf.models.base import GuidMixin
+from reviews.workflow import States
 
 
 def lowercase(lower):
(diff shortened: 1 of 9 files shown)


aaxelb:
The failing test is called `test_unpublished_invisible_to_write_contribs`, which makes me think this is deliberate behavior we should maybe not change? Maybe for unreviewed providers, preserve the existing behavior, but allow contributors to see preprints on reviewed providers as long as it's not in `initial` state?

api/preprints/permissions.py
```diff
@@ -17,6 +17,8 @@ def has_object_permission(self, request, view, obj):
             if auth.user is None:
                 return obj.verified_publishable
             else:
+                if obj.provider.reviews_workflow:
+                    return obj.verified_publishable or (node.is_public and auth.user.has_perm('view_submissions', obj.provider)) or node.has_permission(auth.user, osf_permissions.ADMIN) or (node.is_contributor(auth.user) and obj.reviews_state != 'initial')
```
aaxelb: Minor: I don't like how much is duplicated between this and the line below. Could you do something like assign the part in common to a variable and `return foo or (node.is_contributor(auth.user) and obj.reviews_state != 'initial')` here?

api/preprints/permissions.py
```diff
@@ -17,7 +17,10 @@ def has_object_permission(self, request, view, obj):
             if auth.user is None:
                 return obj.verified_publishable
             else:
-                return obj.verified_publishable or (node.is_public and auth.user.has_perm('view_submissions', obj.provider)) or node.has_permission(auth.user, osf_permissions.ADMIN)
+                user_has_permissions = obj.verified_publishable or (node.is_public and auth.user.has_perm('view_submissions', obj.provider)) or node.has_permission(auth.user, osf_permissions.ADMIN)
+                if obj.provider.reviews_workflow:
+                    return user_has_permissions or (node.is_contributor(auth.user) and obj.reviews_state != 'initial')
```
aaxelb: Instead of `'initial'`, use `States.INITIAL.value` from `reviews.workflow`

api/base/filters.py
```diff
@@ -504,8 +505,10 @@ def preprints_queryset(self, base_queryset, auth_user, allow_contribs=True):
             admin_user_query = Q(node__contributor__user_id=auth_user.id, node__contributor__admin=True)
             reviews_user_query = Q(node__is_public=True, provider__in=get_objects_for_user(auth_user, 'view_submissions', PreprintProvider))
             if allow_contribs:
-                contrib_user_query = Q(is_published=True, node__contributor__user_id=auth_user.id, node__contributor__read=True)
-                query = default_query & (no_user_query | contrib_user_query | admin_user_query | reviews_user_query)
+                contrib_query_base = Q(node__contributor__user_id=auth_user.id, node__contributor__read=True)
+                not_reviews_contrib_query = contrib_query_base & Q(is_published=True)
+                reviews_contrib_query = contrib_query_base & (Q(provider__reviews_workflow__isnull=False) & ~Q(reviews_state=States.INITIAL.value))
+                query = default_query & (no_user_query | not_reviews_contrib_query | reviews_contrib_query | admin_user_query | reviews_user_query)
```
aaxelb: This is a fun pile of logic that keeps on growing.

laurenbarker: Yeah, once there are multiple workflows this isn't going to work.

aaxelb: Hmm, while waiting for travis it occurred to me that with #7767, all preprints for unreviewed providers are set to `accepted` when they're published, so I think we could do:
```python
contrib_user_query = ~Q(reviews_state=States.INITIAL.value) & Q(node__contributor__user_id=auth_user.id, node__contributor__read=True)
```
regardless of workflow. Does that seem reasonable?

aaxelb: And it *might* reasonably hold for future workflows that a preprint should be visible to contributors once it's transitioned away from `initial`.

aaxelb:
@laurenbarker Travis is sad :(

api/base/filters.py
laurenbarker: Yeah, that would work. For some reason I was thinking `reviews_state` would be `null` for provider who don't use moderation but the default is `initial` and it's not a nullable field.

PR #7801, opened 2017-10-11
Finalize standardization of query API [OSF-8295]

erinspace:
## Purpose

Most of modularodm has been removed from the codebase. Now is the time to make the final push to remove the compatibility methods and switch their uses, all imports of modm, and change tests, all to make way for python 3.

## Changes

- Remove `find()` and `find_one()` compatibility methods, switch where they were used to the appropriate django queries
- Remove modm imports, exceptions, queries
- Rename  `ODMOrderingFilter ` to `OSFOrderingFilter` to continue to preserve existing ordering functionality.
- Remove all of `modm_compat.py` along with uses of its methods
- Update scripts

## Side effects
This touches a lot of files - while the tests hopefully cover most of the edge cases, it might be useful to double check most functionality. Maybe not on the level of a full site regression, but it's hard to pinpoint exactly where

## Ticket
https://openscience.atlassian.net/browse/OSF-8295


scripts/populate_conferences.py
```diff
@@ -1,1034 +0,0 @@
-#!/usr/bin/env python
```
caseyrollins: Do you know if there is an equivalent Django management command to replace this script? I know we can make meetings locally using the admin app but this could be kinda nice for devs to quickly populate a bunch of meetings? But also if this script is super broken/out of date it's probably not worth the trouble.

tests/test_registrations/test_embargoes.py
```diff
@@ -903,7 +903,7 @@ def test_POST_register_embargo_is_not_public(self, mock_enqueue):
 
         assert_equal(res.status_code, 202)
 
-        registration = Registration.find().order_by('-registered_date').first()
+        registration = Registration.objects.all().order_by('-registered_date').first()
```
caseyrollins: Similar to the FYI note above, `.all()` isn't needed here (but doesn't hurt anything either).

website/conferences/message.py
```diff
@@ -177,7 +177,7 @@ def attachments(self):
     def allowed_types(self):
         Conference = apps.get_model('osf.Conference')
         allowed_types = []
-        for conf in Conference.find():
+        for conf in Conference.objects.all():
```
caseyrollins: `.values('field_names')`?

osf/modm_compat.py
```diff
@@ -1,174 +0,0 @@
-# -*- coding: utf-8 -*-
```
caseyrollins: 🎉

api/base/settings/__init__.py
```diff
@@ -31,7 +31,7 @@ def load_origins_whitelist():
 
     institution_origins = tuple(domain.lower() for domain in itertools.chain(*[
         institution.domains
-        for institution in Institution.find()
+        for institution in Institution.objects.all()
```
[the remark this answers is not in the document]
sloria: Agreed: this should be able to be simplified to

```python
institution_origins = tuple(domain.lower() for domain in itertools.chain(*Institution.objects.values_list('domains', flat=True)))
```

framework/auth/utils.py
```diff
@@ -89,8 +89,9 @@ def privacy_info_handle(info, anonymous, name=False):
 
 def ensure_external_identity_uniqueness(provider, identity, user=None):
     from osf.models import OSFUser
-
-    users_with_identity = OSFUser.find(Q('external_identity.{}.{}'.format(provider, identity), 'ne', None))
+    users_with_identity = OSFUser.objects.filter(
+        Q(**{'external_identity__{}__{}__isnull'.format(provider, identity): False})
```
sloria: Minor: no need to use `Q` here.

osf/models/base.py
```diff
@@ -85,26 +79,8 @@ def load(cls, data, select_for_update=False):
...
-    def find(cls, query=None):
-        if not query:
-            return cls.objects.all()
-        else:
-            return cls.objects.filter(to_django_query(query, model_cls=cls))
-
-    @classmethod
     def remove(cls, query=None):
-        return cls.find(query).delete()
+        return cls.objects.filter(query).delete() if query else cls.objects.all().delete()
```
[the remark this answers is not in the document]
sloria: I agree that we don't need `.remove` and `.remove_one`, but those can be remove in a follow-up PR

osf/models/preprint_service.py
```diff
@@ -47,7 +47,7 @@ class Meta:
         )
 
     def __unicode__(self):
-        return '{} preprint (guid={}) of {}'.format('published' if self.is_published else 'unpublished', self._id, self.node.__unicode__())
+        return '{} preprint (guid={}) of {}'.format('published' if self.is_published else 'unpublished', self._id, self.node.__unicode__() if self.node else None)
```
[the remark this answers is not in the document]
erinspace: This was added for [this particular test](https://github.com/CenterForOpenScience/osf.io/blob/develop/api_tests/preprints/views/test_preprint_list_mixin.py#L194) in one of those bizzare cases of "how did this ever work??"

scripts/populate_conferences.py
erinspace: I tried fixing it first! It was though so super broken (and had clearly not been updated in quite awhile) that it would probably be way easier to just write a new one if we needed it than to try to fix this one.

framework/sessions/utils.py
```diff
@@ -11,7 +10,8 @@ def remove_sessions_for_user(user):
     """
     from osf.models import Session
 
-    Session.remove(Q('data.auth_user_id', 'eq', user._id))
+    if user._id:
```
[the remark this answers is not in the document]
erinspace: Right! So there are several tests that create a user and then [set the password before saving it](https://github.com/CenterForOpenScience/osf.io/blob/develop/osf_tests/test_user.py#L765), resulting in user without an _id. Setting a password calls `remove_sessions_for_user` which then uses this query. Since `Q` was OK with querying on `None` and Django is not, I added the check for the `._id` before querying for a session to remove instead of changing every test that does this (there are 11 that do this) thinking that if a user isn't actually saved yet, there are no sessions to remove for realz. I could instead change the tests tho so lemmie know!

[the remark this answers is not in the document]
sloria: I was wondering this too. Thanks for the explanation. I think this is fine as is.

sloria:
Merged into release/next-djangole

PR #7803, opened 2017-10-12
[Feature][Migration Required] Zotero Group Library Integration [OSF-2546]

pattisdr:
## Purpose
Currently, only personal Zotero libraries are supported in OSF, but a great feature of Zotero over Mendeley is that their group collaboration libraries are free. Add support for Zotero group libraries. 

## Changes

- When creating  your API key for zotero,  default permissions changed so the key has read permissions to groups
- Add library_id to node_settings model to store selected group library. Migration required.
- Create zotero_node_settings.mako template to build custom UI for selecting groups first, and then folders.
- New v1 route added: all group libraries that your API key has access to are returned as part of the request to `api/v1/project/<pid>/zotero/groups/'`.
- UI loops through all groups the key has access to, as well as still providing your Personal Library as an option
- Folder picker not shown until library(group) has been selected.
- Selecting library and hitting 'Save' sends a PUT request to `api/v1/project/{node_id}/zotero/settings/`, same endpoint that adds folder.  Except here, library_id is added to node_settings and library key is added to user_settings oauth_grants
- Once library selected, then you can load folders from that library.  This process is same as before.
- Citations on node project page are loaded either from the group or the user's account, depending on which library was selected on the node settings page.
- New log action: `zotero_library_selected`
- New logic for whether or not zotero addon is completed - need both library_id and folder_id to both be specified before citations loaded. 

## Importing Library (options are personal library or your group libraries), then Folder
giphy

## Importing numerous libraries (more than a couple).  Five libraries loaded by default, additional libraries loaded on scroll.  Max height of library picker kept constant.

giphy 3

## New Log:

screen shot 2017-10-18 at 12 21 44 pm

## UI displays Library first, then folder
screen shot 2017-10-18 at 12 22 10 pm

## Library picker has same UI as folder picker for consistency
screen shot 2017-10-18 at 12 24 16 pm

## Side effects

Make sure I didn't break Mendeley. I shouldn't have, separated all the new parts.
## Ticket

https://openscience.atlassian.net/browse/OSF-2546


Johnetordoff:
Minor bug here where clicking "Close" button actually makes the file picker open.

zotero

pattisdr:
ahh, good catch @Johnetordoff!

Johnetordoff:
@pattisdr 
I'm sorry I probably could have caught this before, but there's a problem logging this. When switching between libraries I get a message "Unable to to retrieve log details."
screen shot 2018-02-02 at 3 30 12 pm

Otherwise the code content is really excellent and can't find a thing I'd find questionable. 🎉 👍

pattisdr:
Hey @Johnetordoff, I think you need to recreate your assets.  The logs for libraries are new and you will get this message until those are recreated.

Johnetordoff:
@pattisdr  Right you are it now works, passed ✅ !

pattisdr:
Thanks for taking the time to review @Johnetordoff !

sloria:
Merged into `release/next-platform` 🚢 🇮🇹

PR #7807, opened 2017-10-12
[MOD-109][Feature] Reviews, Volume I: Moderation

aaxelb:
Replaces #7708 

https://openscience.atlassian.net/browse/MOD-109

## Purpose
All the backend changes required for OSF Reviews. Adds workflows for moderators to control which preprints are publicly available in their branded interface.

### New changes since #7708 was approved
- Allow including a top-level `meta` object in API responses
  - override `get_renderer_context` in the view, something like:
    ```python
    def get_renderer_context(self):
        context = super(...).get_renderer_context()
        context['meta'] = { ... }
        return context
    ```
  - At `/preprint_providers/<id>/preprints/`, add `meta[reviews_state_counts]=true` in the query string to get counts of that provider's preprints grouped by `reviews_state` to top-level `meta`
- Make all a user's preprints visible on their "My Projects -> All my preprints" page, regardless whether they've been approved/rejected by a moderator
  - Set `_has_abandoned_preprint=False` on submission to providers using a reviews workflow
- Send email notifications when a preprint transitions states
- Mint DOIs with EZID when a preprint first transitions into a public state (should match existing behavior on unreviewed providers)
- Create a "preprint initiated" NodeLog when a preprint is submitted for review
- Keep previously published preprints public when a provider first sets up a reviews workflow
  - Migration to set all preprints for unreviewed providers to "accepted" state
  - When a preprint is published to an unreviewed provider, set "accepted" state
- Allow filtering preprint lists with `filter[node_is_public]`, so we can exclude preprints with private nodes from the reviews interface
- Various bugfixes in the new reviews code

### Original changes
- Add `Action` top-level object
    - Each action represents a transition, a state change on an object
    - e.g. "Abram `submit`ted preprint X for review at time Y, changing its state from `initial` to `pending`"
    - Once created, actions are meant to be immutable
    - New API endpoints:
        - `/actions/`
            - Write-only endpoint for creating actions
            - GET returns an empty list for the sake of the browsable API
        - `/actions/<id>/`
            - Read-only action detail
        - `/preprints/<guid>/actions/`
            - Read-only list of actions which target the given preprint
        - `/users/me/actions/`
            - Read-only list of actions from providers for which the user is a moderator
- Add `ReviewableMixin` for objects which can be reviewed
    - Right now just `PreprintService`
    - Fields:
        - `reviews_state`: Current state of the preprint
    - Includes methods to trigger transitions and create corresponding `Action`s
- Add `ReviewProviderMixin` for objects which own Reviewables and can choose a workflow for them
    - Right now just `PreprintProvider`
    - Fields:
        - `reviews_workflow`: Provider's chosen workflow (see below)
        - `reviews_comments_private`: Whether comments made by moderators are visible to the submitter/contributors
        - `reviews_comments_anonymous`: Whether the name/ID of the moderators who accept/reject/act on a preprint are visible to the submitter/contributors
    - Modified API endpoints:
        - `/preprint_providers/<id>/` is once-writable
            - Provider admins can set the three reviews fields exactly once to set up reviews
            - If `reviews_workflow` is non-null, the endpoint is read-only
- Add workflows that define valid transitions
    - Workflows:
        - `pre-moderation`: Preprints are publicly visible as soon as they're submitted, but can later be removed by moderators
        - `post-moderation`: Preprints are visible only after approval by a moderator
    - Valid states:
        - `initial`: Default state, created but not yet submitted
        - `pending`: Submitted for and awaiting review, may be publicly visible depending on the workflow
        - `accepted`: Has explicit moderator approval, publicly visible under the provider
        - `rejected`: Has explicit moderator disapproval, only visible to contributors and moderators
    - Valid action triggers
        - `submit`
            - Move from `inital` to `pending`, or from `rejected` to `pending` if resubmission is allowed by the workflow
            - Node admins and moderators can `submit`
        - `accept`
            - Move from `pending` or `rejected` to `accepted`
            - Only moderators can `accept`
        - `reject`
            - Move from `pending` or `accepted` to `rejected`
            - Only moderators can `reject`
        - `edit_comment`
            - Modify the comment visible to moderators (and possibly contributors) without changing the preprint's state
            - Only moderators can `edit_comment`
- Use django guardian for permissions
    - Create `admin` and `moderator` groups for each provider, assign group various object permissions on that provider
    - Add users to groups to make them admin and/or moderator

### Preprints workflow
If a provider has non-null `reviews_workflow`, authors can no longer publish their preprints directly (by setting `is_published`). Instead, they must create a `submit` action at `/actions/`. The corresponding changes to ember-preprints are here: https://github.com/CenterForOpenScience/ember-preprints/pull/465

If a provider has null `reviews_workflow`, the existing behavior is unchanged. Workflows are entirely opt-in per provider, for now.

### Ember OSF
ember-osf changes: https://github.com/CenterForOpenScience/ember-osf/pull/272


website/static/js/myProjects.js
```diff
@@ -121,7 +121,7 @@ function NodeFetcher(type, link, handleOrphans, regType, regLink) {
         link = link ? link : $osf.apiV2Url('users/me/nodes/', { query : { 'filter[preprint]': true, 'related_counts' : 'children', 'embed' : ['contributors', 'preprints'], 'fields[users]' : sparseUserFields, 'fields[contributors]' : sparseContributorFields}});
     }
 
-    this.nextLink = link ? 
+    this.nextLink = link ?
         link + '&version=2.2' :
```
brianjgeiger: It wouldn't be a bad idea to upgrade to 2.6 for this, but I don't think it would hold anything up leaving it at 2.2.

reviews/models/mixins.py
```diff
@@ -177,6 +177,7 @@ def update_last_transitioned(self, ev):
     def save_changes(self, ev):
         now = self.action.date_created if self.action is not None else timezone.now()
         should_publish = self.reviewable.in_public_reviews_state
+        self.reviewable.node._has_abandoned_preprint = False
```
sloria: Minor: to avoid potentially repeating queries, save `self.reviewable.node` to a variable.

```python
node = self.reviewable.node
node._has_abandones_preprint = False

....
node.save()
```

api/preprints/serializers.py
```diff
@@ -79,6 +80,7 @@ class PreprintSerializer(JSONAPISerializer):
     title = ser.CharField(source='node.title', required=False)
     description = ser.CharField(required=False, allow_blank=True, allow_null=True, source='node.description')
     tags = JSONAPIListField(child=NodeTagField(), required=False, source='node.tags')
+    node_is_public = ser.BooleanField(read_only=True, source='node__is_public')
```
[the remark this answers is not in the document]
aaxelb: If there's a way to make that work, that'd be lovely, but I get a 400:
```json
{
    "errors": [
        {
            "source": {
                "parameter": "filter"
            },
            "meta": {},
            "detail": "'node__is_public' is not a valid field for this endpoint."
        }
    ]
}
```
even if I add `node__is_public` to `filterable_fields`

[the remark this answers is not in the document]
brianjgeiger: I was about to be very surprised and pleased and possibly a little concerned if that worked.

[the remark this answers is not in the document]
brianjgeiger: I don't think it would take a *lot* to update the filtering to handle this; it's probably just the verification that a field is named what you think it is, but it would be a bit to do.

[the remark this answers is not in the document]
sloria: Ah, it looks like we check if the passed filter name is one of the serializer fields here:

```python
    def _get_field_or_error(self, field_name):
        """
        Check that the attempted filter field is valid

        :raises InvalidFilterError: If the filter field is not valid
        """
        serializer_class = self.serializer_class
        if field_name not in serializer_class._declared_fields:
            raise InvalidFilterError(detail="'{0}' is not a valid field for this endpoint.".format(field_name))
        if field_name not in getattr(serializer_class, 'filterable_fields', set()):
            raise InvalidFilterFieldError(parameter='filter', value=field_name)
        return serializer_class._declared_fields[field_name]
```

Does Django support hidden fields? Maybe if `read_only=True` and `write_only=True`?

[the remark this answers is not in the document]
sloria: No need for this to block merging this PR, but it's worth a few minutes to see if we can avoid adding the extra field.

website/templates/emails/reviews_submission_confirmation.txt.mako
```diff
@@ -0,0 +1,31 @@
...
+For more information about ${reviewable.provider.name}, visit ${provider_url} to learn more. To learn about the Open Science Framework, visit https://osf.io/
+
+For questions regarding submission criteria, please email ${provider_contact_email}
+
+
+Sincerely,
+
+Your ${reviewable.provider.name} and OSF teams
+
+Center for Open Science
```
chennan47: You need to put one empty line between 'Center for Open Science' and '210 Ridge McIntire Road, Suite 500, Charlottesville, VA 22903-5083' or gmail render will just think they are at the same line.

website/templates/emails/reviews_submission_status.txt.mako
```diff
@@ -0,0 +1,28 @@
...
+For more information about ${reviewable.provider.name}, visit ${provider_url} to learn more. To learn about the Open Science Framework, visit https://osf.io/
+
+For questions regarding submission criteria, please email ${provider_contact_email}
+
+
+Sincerely,
+
+Your ${reviewable.provider.name} and OSF teams
+
+Center for Open Science
```
chennan47: One line space.

website/templates/emails/reviews_submission_status.txt.mako
```diff
@@ -0,0 +1,28 @@
...
+210 Ridge McIntire Road, Suite 500, Charlottesville, VA 22903-5083
+
```
chennan47: A link to privacy policy.

website/templates/emails/reviews_update_comment.txt.mako
```diff
@@ -0,0 +1,21 @@
...
+For more information about ${reviewable.provider.name}, visit ${provider_url} to learn more. To learn about the Open Science Framework, visit https://osf.io/
+
+For questions regarding submission criteria, please email ${provider_contact_email}
+
+
+Sincerely,
+
+Your ${reviewable.provider.name} and OSF teams
+
+Center for Open Science
```
chennan47: One line space.

website/templates/emails/reviews_update_comment.txt.mako
```diff
@@ -0,0 +1,21 @@
...
+210 Ridge McIntire Road, Suite 500, Charlottesville, VA 22903-5083
+
```
chennan47: A link to privacy policy

website/templates/emails/reviews_submission_confirmation.html.mako
```diff
@@ -0,0 +1,30 @@
...
+        <p>If you have been erroneously associated with "${reviewable.node.title}", then you may visit the project's "Contributors" page and remove yourself as a contributor.</p>
+        <p>For more information about ${reviewable.provider.name}, visit <a href="${provider_url}">${provider_url}</a> to learn more. To learn about the Open Science Framework, visit <a href="https://osf.io/">https://osf.io/</a>.</p>
+        <p>For questions regarding submission criteria, please email ${provider_contact_email}</p>
+        <br>
+        Sincerely,
+        <br>
+        Your ${reviewable.provider.name} and OSF teams
+        <p>
+        Center for Open Science<br>
+        210 Ridge McIntire Road, Suite 500, Charlottesville, VA 22903-5083
```
chennan47: This is fine for now. But after https://github.com/CenterForOpenScience/osf.io/pull/7613 got merged in. There will be duplicate in html email templates for COS physical address and privacy policy. Create a ticket for the later on clean up. https://openscience.atlassian.net/browse/OSF-8826

website/templates/emails/reviews_submission_confirmation.txt.mako
```diff
@@ -0,0 +1,31 @@
+Hello ${user.fullname},
```
chennan47: Product team has specified all emails to be in html email template instead of txt. There is a ticket to change all current existing txt email to html, which is https://openscience.atlassian.net/browse/OSF-8770. This is fine for now. But we should stop writing any more txt email template.

reviews/models/mixins.py
```diff
@@ -0,0 +1,288 @@
...
+            'reviewable': self.reviewable,
+            'workflow': self.reviewable.provider.reviews_workflow,
+            'provider_url': self.reviewable.provider.domain or '{domain}preprints/{provider_id}'.format(domain=settings.DOMAIN, provider_id=self.reviewable.provider._id),
+            'provider_contact_email': self.reviewable.provider.email_contact or 'contact@osf.io',
+            'provider_support_email': self.reviewable.provider.email_support or 'support@osf.io',
+        }
+
+# Handle email notifications including: update comment, accept, and reject of submission.
+@reviews_signals.reviews_email.connect
+def reviews_notification(self, context):
```
aaxelb: Note: This will be refactored in the 0.1.0 release to use a `mails.notify`-like function instead of checking subscriptions and creating `NotificationDigest`s manually.
https://openscience.atlassian.net/browse/MOD-150

website/mails/mails.py
```diff
@@ -372,3 +372,8 @@ def get_english_article(word):
     'send_data_share_preprint_error_desk',
     subject='Share Error'
 )
+
+REVIEWS_SUBMISSION_CONFIRMATION = lambda provider_name: Mail(
+    'reviews_submission_confirmation',
+    subject='Confirmation of your submission to {}'.format(provider_name)
```
sloria: @laurenbarker Did `subject='Confirmation of your submission to ${provider_name}'` not work?

reviews/models/mixins.py
```diff
@@ -266,12 +266,16 @@ def reviews_notification(self, context):
...
     for user_id in context['email_recipients']:
         user = OSFUser.load(user_id)
         user_subscriptions = get_user_subscriptions(user, event_type)
         context['no_future_emails'] = user_subscriptions['none']
         context['is_creator'] = user == context.get('reviewable').node.creator
-        email = mails.Mail(template, subject='Confirmation of your submission to {provider}'.format(provider=context.get('reviewable').provider.name))
-        mails.send_mail(user.username, email, mimetype='html', user=user, **context)
+        mails.send_mail(
+            user.username,
+            getattr(mails, 'REVIEWS_SUBMISSION_CONFIRMATION')(context.get('reviewable').provider.name),
```
sloria: Why is `getattr` necessary?

website/mails/mails.py
sloria: Woops, disregard. Didn't see this one done in a future commit.

sloria:
🚢 🇮🇹 !

PR #7809, opened 2017-10-13
[OSF-8695] Prevent sendgrid from spamming when used locally.

adlius:
## Purpose

This PR hopefully prevents use of sendgrid to send emails to unintended recipients by adding a whitelist for local development.

## Changes

In `defaults.py`, `SENDGRID_WHITELIST_MODE` is added and defaults to False to make sure this does not affect Sendgrid on prod/staging. Also added `SENDGRID_EMAIL_WHITELIST`.
In `local-dist.py`, `SENDGRID_WHITELIST_MODE` is added and defaults to True to make sure no spamming could happen locally. Also added `SENDGRID_EMAIL_WHITELIST`.
In `tasks.py`, the `_send_with_sendgrid` is modified to check whether the app in `SENDGRID_WHITELIST_MODE`. If so, it would further check to see if the recipient is in the whitelist specified in settings file. Added sentry logs to report failures to send emails in case whitelist mode is accidentally turned on on staging/prod.

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-8695


framework/email/tasks.py
```diff
@@ -85,19 +85,22 @@ def _send_with_smtp(from_addr, to_addr, subject, message, mimetype='html', ttls=
...
-    mail.set_subject(subject)
-    if mimetype == 'html':
-        mail.set_html(message)
-    else:
-        mail.set_text(message)
-    if categories:
-        mail.set_categories(categories)
-    if attachment_name and attachment_content:
-        mail.add_attachment_stream(attachment_name, attachment_content)
+    if settings.DEV_MODE is False or (settings.SENDGRID_EMAIL_WHITELIST and to_addr in settings.SENDGRID_EMAIL_WHITELIST):
```
alexschiller: IIRC `DEV_MODE` is True on staging, which would prevent emails from working properly. `DEBUG_MODE` might work, but it is worth considering adding a new variable that has a big warning and note about using the SENDGRID_EMAIL_WHITELIST

framework/email/tasks.py
```diff
@@ -85,19 +85,22 @@ def _send_with_smtp(from_addr, to_addr, subject, message, mimetype='html', ttls=
...
+            mail.set_categories(categories)
+        if attachment_name and attachment_content:
+            mail.add_attachment_stream(attachment_name, attachment_content)
 
-    status, msg = client.send(mail)
-    return status < 400
+        status, msg = client.send(mail)
+        return status < 400
+    else:
+        pass
```
alexschiller: Might drop a logger warning or even a sentry throw error here in case this setting accidentally gets triggered on prod or various staging servers, not sure which is appropriate.

website/settings/local-dist.py
```diff
@@ -80,5 +80,9 @@
 
 USE_CDN_FOR_CLIENT_LIBS = False
 
+# When DEV_MODE == True and sendgrid is configured locally.
+# `send_mail` would only send emails to addresses included in this whitelist to prevent unintentional spamming
+SENDGRID_EMAIL_WHITELIST= ['']
```
alexschiller: This should probably be in defaults as empty as well so it doesn't fail badly.

brianjgeiger:
@adlius Based on the current setup, will there need to be devops changes in order to keep emails running, or will existing environments work properly as near as you can tell?

adlius:
@brianjgeiger In `defaults.py`, `SENDGRID_WHITELIST_MODE` is set to False. As long as settings file on prod/staging does not override that variable (which I can be sure), then email should work as usual on staging without any devops changes.

PR #7812, opened 2017-10-16
[#OSF-8700]Make support email address configurable

chennan47:
## Purpose
Make support email address configurable, prevent local and staging send emails to help desk and provide full url links to help desk for failed registration.

## Ticket
https://openscience.atlassian.net/browse/OSF-8700

### DIFF
--- a/website/settings/defaults.py
+++ b/website/settings/defaults.py
@@ -123,11 +123,23 @@ USE_CDN_FOR_CLIENT_LIBS = True
 USE_EMAIL = True
 FROM_EMAIL = 'openscienceframework-noreply@osf.io'
 
-if not DEV_MODE and not DEBUG_MODE:
-    SUPPORT_EMAIL = 'support@osf.io'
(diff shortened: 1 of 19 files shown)


website/settings/defaults.py
```diff
@@ -122,7 +122,12 @@ def parent_dir(path):
 
 USE_EMAIL = True
 FROM_EMAIL = 'openscienceframework-noreply@osf.io'
-SUPPORT_EMAIL = 'support@osf.io'
+
+if not DEV_MODE and not DEBUG_MODE:
```
alexschiller: One reason this ticket surfaced was that other people have been hosting their own versions of the OSF (albeit locally for the most part). I think the preferred route is to not configure a support email automatically at all and to have devops add it to their local config.

chennan47:
closed in favor of https://github.com/CenterForOpenScience/osf.io/pull/7828

PR #7814, opened 2017-10-16
[#OSF-7297]Preprints PATCH requests do not require ID and Type

chennan47:
## Purpose
Add id and type check for patching requests.

## Changes
Add id and type check for patching requests.

## QA notes
nothing to be tested by QA
## Ticket
https://openscience.atlassian.net/browse/OSF-7297


brianjgeiger:
It occurs to me that, since this is literally breaking things, this will be a breaking change, and so needs to be versioned.

api/base/parsers.py
```diff
@@ -74,6 +75,14 @@ def flatten_data(self, resource_object, parser_context, is_list):
             if object_type is None:
                 raise JSONAPIException(source={'pointer': '/data/type'}, detail=NO_TYPE_ERROR)
 
+        if parser_context['request'].version >= 2.6:
```
brianjgeiger: Has to be the next available version because it will break existing 2.6 code if we add it. We may see about releasing this with any other upcoming breaking changes when the time comes.

api_tests/preprints/views/test_preprint_detail.py
```diff
@@ -22,10 +22,11 @@
 from reviews.workflow import States
 
 
-def build_preprint_update_payload(node_id, attributes=None, relationships=None):
+def build_preprint_update_payload(node_id, attributes=None, relationships=None, type='POST'):
```
brianjgeiger: The types here should be the content type (`nodes`, `users`, `preprints`, etc). If these tests don't fail with the current data, then please ensure that the type that you're using matches the type of the object as used by the API.

api/base/parsers.py
```diff
@@ -74,6 +75,14 @@ def flatten_data(self, resource_object, parser_context, is_list):
...
+            if 'preprint' in parser_context['request'].path and request_method == 'PATCH':
+                if object_id is None:
+                    raise JSONAPIException(source={'pointer': '/data/id'}, detail=NO_ID_ERROR)
+
+                if object_type is None:
+                    raise JSONAPIException(source={'pointer': '/data/type'}, detail=NO_TYPE_ERROR)
```
brianjgeiger: Please also check if the object type exists and matches one of the accepted types for the current view (see comment below).

api_tests/preprints/views/test_preprint_detail.py
brianjgeiger: And also please make an explicit test where you put in no data it returns your error, and one where you put the wrong type for the endpoint and it returns an error (`409 Conflict` in the case of the wrong type)

api/preprints/serializers.py
```diff
@@ -71,6 +71,7 @@ class PreprintSerializer(JSONAPISerializer):
     ])
 
     id = IDField(source='_id', read_only=True)
+    type = TypeField()
```
brianjgeiger: Well, that's a fairly simple solution; very nice. Could you put that field into all of the rest of the serializers?

api_tests/preprints/views/test_preprint_detail.py
```diff
@@ -242,6 +243,23 @@ def test_update_primary_file(self, app, user, preprint, url):
         assert log.action == 'preprint_file_updated'
         assert log.params.get('preprint') == preprint._id
 
+    def test_update_preprints_with_no_type(self, app, user, preprint, url):
+        payload = {
```
brianjgeiger: Why are you building a payload by hand rather than using `build_preprint_update_payload()`?

Johnetordoff: Seconded

api/preprints/serializers.py
Johnetordoff: I assume @brianjgeiger meant all JSONAPISerializer based sterilizers, if that makes more tests fail might be worth spinning that as it's own ticket. Not sure.

api_tests/preprints/views/test_preprint_detail.py
chennan47: By assigning the content_type to be None by using the build_preprint_update_payload(), it will raise python type check error since content_type expect type 'str' but got 'None' instead. So in this case, build a small dict just for this test is more convenient.

api/preprints/serializers.py
chennan47: JSONAPISerializer based sterilizers is inherited by 36 Serializer in the code base and a lot of them don't have type field currently and not sure if they necessarily need one, for example, there is a BaseActionSerializer inheriting JSONAPISerializer. This will definitely break a lot of tests and if that's what we want. It's better to make another ticket for this case.

api/base/parsers.py
```diff
@@ -206,10 +203,100 @@ def flatten_relationships(self, relationships):
         return rel
 
 
+class PreprintDetailJSONAPIMultipleRelationshipsParser(JSONAPIMultipleRelationshipsParser):
```
sloria: Since these are preprint-specific, they should go in `api.preprints.parsers`.

api/base/parsers.py
```diff
@@ -206,10 +203,100 @@ def flatten_relationships(self, relationships):
...
+
+
+class PreprintDetailJSONAPIMultipleRelationshipsParserForRegularJSON(JSONAPIMultipleRelationshipsParserForRegularJSON):
+
+    def flatten_data(self, resource_object, parser_context, is_list):
+        """
+               Flattens data objects, making attributes and relationships fields the same level as id and type.
+               """
+
+        relationships = resource_object.get('relationships')
```
sloria: Is this code exactly the same as above? If so, you can DRY it up with a mixin. Something like

```python
class PreprintDetailMixin(...):

    def flatten_data(...):


class PreprintDetailJSONAPIMultipleRelationshipsParser(PreprintDetailMixin, JSONAPIMultipleRelationshipsParser):


```

api/preprints/parsers.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class PreprintDetailMixin(JSONAPIParser):
+
+    def flatten_data(self, resource_object, parser_context, is_list):
+        """
+               Flattens data objects, making attributes and relationships fields the same level as id and type.
+               """
+
+        relationships = resource_object.get('relationships')
```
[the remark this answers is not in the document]
chennan47: There is only one difference https://github.com/CenterForOpenScience/osf.io/pull/7814/files#diff-32ed851425c582df165eef7a30585ec0R41

Doing a version check for the type requirement.

api/preprints/parsers.py
```diff
@@ -0,0 +1,68 @@
...
+    JSONAPIMultipleRelationshipsParserForRegularJSON,
+    JSONAPIMultipleRelationshipsParser,
+    NO_RELATIONSHIPS_ERROR,
+    NO_ATTRIBUTES_ERROR,
+    NO_ID_ERROR,
+    NO_TYPE_ERROR
+)
+
+
+class PreprintDetailMixin(JSONAPIParser):
```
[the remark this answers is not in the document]
chennan47: Need to inherit from JSONAPIParser or need to overwrite flatten_relationships https://github.com/CenterForOpenScience/osf.io/pull/7814/files#diff-32ed851425c582df165eef7a30585ec0R52

api/preprints/parsers.py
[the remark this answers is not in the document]
sloria: Ah, in that case, how about adding a flag to the parser context, similar to `attributes_required`:

```python
    def get_parser_context(self, http_request):
        res = super(...)....
        res['type_required'] = True
        return res
```

sloria:
Ah, my previous comment got blown away in the diff, so reposting:

How about adding a flag to the parser context, similar to `attributes_required`:

```python
    def get_parser_context(self, http_request):
        res = super(...)....
        res['type_required'] = True
        return res
```

That way, we don't need special parsers just for PreprintDetail, and the code is more consistent with the previous code for `attributes_required`.

api/base/parsers.py
```diff
@@ -182,7 +182,7 @@ def parse(self, stream, media_type=None, parser_context=None):
             if id_ is None:
                 raise JSONAPIException(source={'pointer': '/data/id'}, detail=NO_ID_ERROR)
 
-            if type_required and parser_context['request'].version >= 2.7 and request_method == 'PATCH':
+            if type_required and request_method == 'PATCH':
```
sloria: `type` is also required for `POST` requests:

> A resource can be created by sending a POST request to a URL that represents a collection of resources. The request MUST include a single resource object as primary data. The resource object MUST contain at least a type member.

http://jsonapi.org/format/#crud-creating

chennan47: solved in 2ab5eed

api_tests/preprints/views/test_preprint_detail.py
```diff
@@ -19,12 +19,12 @@
 
 
 def build_preprint_update_payload(
-        node_id,
-        attributes=None,
-        relationships=None):
+        node_id, attributes=None, relationships=None,
+        content_type='preprints'):
```
sloria: `content_type` is easily confused with `ContentType` in Django or `Content-Type` in HTTP. How about `jsonapi_type`?

api_tests/preprints/views/test_preprint_detail.py
```diff
@@ -293,6 +293,23 @@ def test_update_primary_file(self, app, user, preprint, url):
         assert log.action == 'preprint_file_updated'
         assert log.params.get('preprint') == preprint._id
 
+    def test_update_preprints_with_no_type(self, app, user, preprint, url):
+        payload = {
+            'data': {
+                'id': preprint._id,
+                'type': None,
```
sloria: Can you also add test where `type` is missing from the payload?

api_tests/preprints/views/test_preprint_detail.py
```diff
@@ -617,7 +634,8 @@ def url(self, preprint):
     def make_payload(self):
         def payload(
                 node_id, license_id=None, license_year=None,
-                copyright_holders=None):
+                copyright_holders=None, content_type='preprints'
```
sloria: See previous comment re: renaming this.

api_tests/preprints/views/test_preprint_detail.py
```diff
@@ -304,8 +304,18 @@ def test_update_preprints_with_no_type(self, app, user, preprint, url):
...
 
+    def test_update_preprints_with_no_type(self, app, user, preprint, url):
+        payload = {
+            'data': {
+                'id': preprint._id,
+            }
+        }
+
+        res = app.patch_json_api(url, payload, auth=user.auth, expect_errors=True)
+        assert res.status_code == 400
```
sloria: Can you add an assertion that the error object is correct? It's possible for this to pass if an unrelated error is returned (e.g. missing attributes).

api/base/parsers.py
```diff
@@ -66,12 +69,16 @@ def flatten_data(self, resource_object, parser_context, is_list):
         object_id = resource_object.get('id')
         object_type = resource_object.get('type')
 
+        type_required = not (
+            legacy_type_allowed and parser_context['request'].version < 2.7 and request_method == 'PATCH'
```
sloria: `type` is also required for `POST` requests.

sloria: Talked in person: this is correct as is.

PR #7816, opened 2017-10-16
[MOD-147][Improvement] Set abandoned to False on submission

laurenbarker:
## Purpose

Don't display abandoned preprints on "All my preprints" page.

## Changes

Set abandoned to False on submission.

## Side effects

None

## Ticket
https://openscience.atlassian.net/browse/OMOD-147


osf/migrations/0065_auto_20171016_1201.py
```diff
@@ -0,0 +1,29 @@
...
+from django.db import migrations
+
+from reviews.workflow import States
+
+
+# Make sure all submitted preprints have _has_abandoned_preprint=False.
+def set_abandoned_false_for_all_submitted_preprints(apps, schema_editor):
+    Preprint = apps.get_model('osf', 'PreprintService')
+    abandoned_submitted_preprints = Preprint.objects.filter(node___has_abandoned_preprint=True).exclude(reviews_state=States.INITIAL.value)
+    for preprint in abandoned_submitted_preprints.iterator():
```
aaxelb: Could do `abandoned_submitted_preprints.update(_has_abandoned_preprint=False)` -- one query instead of instead of 1 + n. There shouldn't be very many, so it doesn't matter much, but it's good practice.

laurenbarker: Maybe I missed something, but the reason I didn't use `update()` was that I needed to update a related model.
> it can only update columns in the model’s main table, not on related models.
-[Django Docs](https://docs.djangoproject.com/en/1.11/ref/models/querysets/#django.db.models.query.QuerySet.update)

aaxelb: Oh whoops, I mixed that up. You can still do it in one query, though:
```
Node.objects.filter(
    id__in=abandoned_submitted_preprints.values('id')
).update(_has_abandoned_preprint=False)
```
or
```
Node.objects.filter(
    _has_abandoned_preprint=True,
    preprints__isnull=False
).exclude(
    preprints__reviews_state=States.INITIAL.value
).update(_has_abandoned_preprint=False)
```

laurenbarker: Oh cool, I could not figure out the inverse relation. I was expecting `node.preprint`.

laurenbarker:
@aaxelb, tested in the shell. If you could run the migration locally that would be great. If not, let me know so I can finagle my db and test it out.

osf/migrations/0065_auto_20171016_1201.py
aaxelb: That would be reasonable, wouldn't it?

PR #7817, opened 2017-10-16
[MOD-153][Fix] Update status change email so notify comment is always checked

laurenbarker:
## Purpose
Include language about comment for pre-moderation too.

## Changes
Check for comment for both pre-moderation and post-moderation.

## Side effects
None

## Ticket

https://openscience.atlassian.net/browse/MOD-153

### DIFF
--- a/website/templates/emails/reviews_submission_status.html.mako
+++ b/website/templates/emails/reviews_submission_status.html.mako
@@ -1,18 +1,31 @@
 ## -*- coding: utf-8 -*-
 <div style="margin: 40px;">
     <p>Hello ${user.fullname},</p>
-    % if workflow == 'pre-moderation':
-        <p>Your submission "${reviewable.node.title}", submitted to ${reviewable.provider.name} has ${'not been accepted. You may edit the '+ reviewable.provider.preprint_word+ ' and resubmit, at which time it will becoming pending moderation.' if is_rejected else 'been accepted by the moderator and is now discoverable to others.'}</p>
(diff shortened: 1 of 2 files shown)


website/templates/emails/reviews_submission_status.html.mako
```diff
@@ -1,18 +1,31 @@
...
-    <p>If you have been erroneously associated with "${reviewable.node.title}", then you may visit the project's "Contributors" page and remove yourself as a contributor.</p>
-    <p>For more information about ${reviewable.provider.name}, visit <a href="${provider_url}">${provider_url}</a> to learn more. To learn about the Open Science Framework, visit <a href="https://osf.io/">https://osf.io/</a>.</p>
-    <p>For questions regarding submission criteria, please email ${provider_contact_email}</p>
-    <br>
-    Sincerely,
+    <p>
+        % if workflow == 'pre-moderation':
+            Your submission "${reviewable.node.title}", submitted to ${reviewable.provider.name} has ${'not been accepted. You may edit the ' + reviewable.provider.preprint_word + ' and resubmit, at which time it will becoming pending moderation.' if is_rejected else 'been accepted by the moderator and is now discoverable to others. '}
```
aaxelb: Would you mind moving the `is_rejected` checks to a block `% if` too?

laurenbarker: I was worried that one would mess up the spacing between sentences. I'll try it though.

PR #7818, opened 2017-10-17
[OSF-8163] Remove QA items from elastic search and SHARE.

adlius:
## Purpose

Currently, nodes for QA purposes (marked with tag "qatest") are indexed by both elastic search and SHARE. This PR should remove such items from search results and SHARE index and should prevent future items to be indexed.

## Changes

There are two parts to this PR:
1. Remove existing items from Elastic Search results and SHARE index
    * A script named `remove_qa_test_items_from_elastic_and_share.py` is added. In the script, `update_node()` and `update_node_share()` are called for node instances whose tags include 'qatest' or 'qa test'. `update_node()` and `update_node_share()` are methods that will be invoked upon node updates. However, this script uses it to force update the target nodes to Elastic and SHARE with new data, which is explained further in the second part of this PR.
    * DB pagination is added to the script with increment of 20.

2. Prevent future QA test items with 'qatest' tags to be indexed by Elastic Search and SHARE.
    * In `website/settings/defaults.py`, a `DO_NOT_INDEX_LIST` is added as a configurable list for QA tags and titles.
    * In `tasks.py::format_registartion()`, a conditional is added to set the 'is_deleted' field for SHARE payload to True for target nodes. The same thing is done for `tasks.py::format_node()`. This prevents SHARE from indexing future qa items.
    * In `elastic_search.py::update_node()`, the same conditional is added. This prevents search from indexing future QA items.
    * The previous script would force trigger update to SHARE and Elastic without actually updating the target nodes themselves, therefore removing the target nodes from search results.

3. Added some tests.

## Side effects

Tested locally and it works for Elastic search. However, I cannot verity that it works for SHARE because local SHARE actually uses data of staging.

## Ticket

https://openscience.atlassian.net/browse/OSF-8163


website/project/tasks.py
```diff
@@ -102,15 +102,15 @@ def format_node(node):
         }, {
             '@id': '_:789',
             '@type': 'project',
-            'is_deleted': not node.is_public or node.is_deleted or node.is_spammy,
+            'is_deleted': not node.is_public or node.is_deleted or node.is_spammy or 'qatest' in node.tags.all().values_list('name', flat=True)
```
brianjgeiger: This only has qatest while you script also has 'qa test'.

brianjgeiger:
That makes sense, only updating the nodes that you find. Good thinking.

brianjgeiger:
@adlius Please fix the flake8 error

brianjgeiger:
Tests failing. At least some of them look relevant.

website/project/tasks.py
```diff
@@ -102,15 +104,18 @@ def format_node(node):
...
 
 def format_registration(node):
+    is_qa_node = bool(set(settings.DO_NOT_INDEX_LIST['tags']).intersection(node.tags.all().values_list('name', flat=True))) \
+        or any(substring in node.title for substring in settings.DO_NOT_INDEX_LIST['titles'])
+
     registration_graph = GraphNode('registration', **{
         'title': node.title,
         'description': node.description or '',
-        'is_deleted': not node.is_public or 'qatest' in (node.tags.all() or []) or node.is_deleted,
+        'is_deleted': not node.is_public or is_qa_node,
```
brianjgeiger: You seem to have gotten rid of the `node.is_deleted` case in this statement, which I _suspect_ is important, unless it's happening implicitly somehow in `is_qa_node`, but if so, it's pretty subtle and might need to be made more explicit.

adlius: You are definitely right....... I somehow removed the `is_deleted` when I added the `is_qa_node`.

brianjgeiger: I'm a little concerned that it didn't break anything, and I expect it's because it's in a task. We'll probably need to fix that at some point, but not for this as long as you run some manual tests and do your best to make sure it's okay before it goes to QA.

adlius: Since this is formatting the payload of request to SHARE for updating node information on SHARE side, we can't really test it with OSF unit tests. I also can't do manual tests that confirm whether the change would remove the item from SHARE results. However I did confirm with Abram that if we we want to remove a node from SHARE we only need to set the `is_deleted` to True in the payload. In this case, I am simply adding another condition to set the `is_deleted` field so this should be fine.

adlius: Umm... On second thought, we could at least write some tests to make sure the payload it formatted properly upon node updates. It seems currently we do have some tests doing some trivial test on the payload (e.g. whether the payload is formatted correctly) but nothing substantial.

adlius:
Turns out that there were tests checking the payload sent to SHARE. But since the tests were on a `Project` instance thus did not trigger the `format_registration` method. I've added tests to check for both projects and registrations. I've also added tests to test the payload for QA items.

osf_tests/test_node.py
```diff
@@ -3371,6 +3375,75 @@ def test_update_share_correctly(self, requests, node, user, request_context):
             graph = kwargs['json']['data']['attributes']['data']['@graph']
             assert graph[1]['is_deleted'] == case['is_deleted']
 
+    @mock.patch('website.project.tasks.settings.SHARE_URL', 'https://share.osf.io')
+    @mock.patch('website.project.tasks.settings.SHARE_API_TOKEN', 'Token')
+    @mock.patch('website.project.tasks.requests')
+    @mock.patch('osf.models.registrations.Registration.archiving', mock.PropertyMock(return_value=False))
+    def test_update_share_correctly_for_registrations(self, requests, registration, user, request_context):
+        cases = [{
```
brianjgeiger: Maybe a test for the is_qa_node functionality?

brianjgeiger:
@adlius If you ping me I can restart whichever branches may be failing rather than closing and reopening the whole PR.

adlius:
@brianjgeiger I didn't know you could do that...... Good to know!

PR #7823, opened 2017-10-18
[MOD-151][Fix] Include no_future_emails in submission confirmation email

laurenbarker:
## Purpose

Don't say that the user will receive notifications for future emails if their default settings is "never".

## Changes

Include no_future_emails in submission confirmation email

## Side effects

None

## Ticket
https://openscience.atlassian.net/browse/MOD-151

### DIFF
--- a/reviews/models/mixins.py
+++ b/reviews/models/mixins.py
@@ -267,8 +267,11 @@ def reviews_notification(self, context):
 @reviews_signals.reviews_email_submit.connect
 def reviews_submit_notification(self, context):
     template = context['template']
+    event_type = utils.find_subscription_type('global_reviews')
     for user_id in context['email_recipients']:
         user = OSFUser.load(user_id)
+        user_subscriptions = get_user_subscriptions(user, event_type)
+        context['no_future_emails'] = user_subscriptions['none']
         context['is_creator'] = user == context.get('reviewable').node.creator
         email = mails.Mail(template, subject='Confirmation of your submission to {provider}'.format(provider=context.get('reviewable').provider.name))
         mails.send_mail(user.username, email, mimetype='html', user=user, **context)


reviews/models/mixins.py
```diff
@@ -267,8 +267,11 @@ def reviews_notification(self, context):
 @reviews_signals.reviews_email_submit.connect
 def reviews_submit_notification(self, context):
     template = context['template']
+    event_type = utils.find_subscription_type('global_reviews')
     for user_id in context['email_recipients']:
         user = OSFUser.load(user_id)
+        user_subscriptions = get_user_subscriptions(user, event_type)
```
aaxelb: I think we should maybe use `compile_subscriptions` here, to allow for node subscription settings? And then check `user_id in subscriptions['none']` to determine whether they'll be getting future emails. @sloria, does that make sense?

laurenbarker: Hm, yeah this is a little weird. If the user is creating a preprint from an existing project this could be inaccurate.

sloria: I'm not sure if `compile_subscriptions` is correct here. Are review emails configurable on nodes? If not, then I think this is correct.

aaxelb: They will be configurable on nodes [shortly](https://openscience.atlassian.net/browse/MOD-150).

sloria:
osf notifications 2017-10-18 14-26-39

should be "Preprint submissions updated"

reviews/models/mixins.py
laurenbarker: They are configurable on nodes (or will be soon). Ideally they would be configured on preprints instead of nodes though. Once the node/preprint divorce happens this setting won't make any sense on projects.

laurenbarker:
@sloria Nici chose "Preprints submission updated" it's not a typo. "Preprints" because the service is plural. I think "Preprints submissions updated" would be most correct but does sound weird.

sloria:
@laurenbarker But the name of the service is "OSFPreprints", not "Preprints". So how about "OSFPreprints submissions updated"?

@nicipfeiffer

laurenbarker:
@sloria @aaxelb I talked with Nici and she agreed to "Preprint submissions updated" and not having project specific settings for now.

PR #7824, opened 2017-10-18
[feature] Files Page Optimizations [OSF-8677]

caseyrollins:
#### Purpose
- The files page is suuuuuper slow to load for large projects. This PR improves performance (slightly) by reducing the number of queries being made and not loading all children beyond first level unless they are manually expanded.
- Note: This PR only makes a *slight* improvement because we're still automatically expanding top-level children on page load. For an even faster page load, we should NOT automatically expand children, but that's a larger UI/UX decision that needs to be discussed amongst Product. 

#### Changes
- Removes the osf-logo spinner in favor of ball-scale. 
- Rewrote a bit of rubeus.py to make fewer queries (uses subqueries, annotations, other fun stuff)
  - Additionally, only children of the parent project and first-level children are serialized in the files grid response. 
- Adds functionality to the treebeard grid to only request data about second level children when manually expanded by the user.

#### Side Effects
- None expected.

#### QA Notes 
- The files page UI should appear exactly the same when the page loads (i.e. addons and first-level children are automatically expanded). When expanding a second-level (or greater) child, a request is made to retrieve the children and you'll see the loading icon until the component expands. 
- This PR *should* only affect the way the file tree loads and how folders/addons expand and collapse, although a basic sweep of functionality (moving files, creating folders, etc) wouldn't be a bad idea. Pay special attention to permissions (i.e. that component names aren't shown to non-contributors). Also, please check different node types (forks, links, registrations) in the file tree. 

#### Stats 🐱 
```
/api/v1/project/<node>/files/grid/

Project with 42 top-level children, 93 totol children
    BEFORE
	num queries: 4131
	time (avg): 64.2 s
    AFTER 
	num queries: 1148
	time: 25.03 s	

Project with 1 top-level child 
    BEFORE 
	num queries: 47
	time: 5.1 s
    AFTER 
	num queries: 41
	time: 4.76 s
```

#### Ticket
- [OSF-8677](https://openscience.atlassian.net/browse/OSF-8677)


website/static/js/fangorn.js
```diff
@@ -1609,14 +1613,37 @@ function expandStateLoad(item) {
...
             if (item.children[i].data.isAddonRoot || item.children[i].data.addonFullName === 'OSF Storage' ) {
                 tb.updateFolder(null, item.children[i]);
             }
         }
     }
-        $('.fangorn-toolbar-icon').tooltip();
+
+    if (item.depth > 2 && !item.data.isAddonRoot && item.children.length === 0 && item.open) {
+        if (icon.get(0)) {
+            m.render(icon.get(0), tbOptions.resolveRefreshIcon());
```
sloria: Calling `m.render` is generally a bad idea. Can you explain why it's necessary here?

website/templates/project/files.mako
```diff
@@ -9,13 +9,9 @@
 %endif
 
 <div id="treeGrid">
-	<div class="spinner-loading-wrapper">
-		<div class="logo-spin logo-lg"></div>
-		<p class="m-t-sm fg-load-message"> Loading files...  </p>
-	</div>
+    <div class="ball-scale ball-scale-blue text-center m-v-xl"><div></div></div>
```
sloria: ❤️  🔵

website/util/rubeus.py
```diff
@@ -231,10 +207,29 @@ def _serialize_node(self, node, visited=None, parent=None):
...
 
+    def _get_nodes(self, node):
+        AbstractNode = apps.get_model('osf.AbstractNode')
+        Contributor = apps.get_model('osf.Contributor')
+        NodeRelation = apps.get_model('osf.NodeRelation')
+
+        data = []
+        if node.can_view(auth=self.auth):
+            serialized_addons = self._collect_addons(node)
+            linked_node_sqs = NodeRelation.objects.filter(parent=node, is_node_link=True)
```
sloria: Minor: you can use the `related_name` here : `node.node_relations.filter(is_node_link=True)`.

website/util/rubeus.py
```diff
@@ -231,10 +207,29 @@ def _serialize_node(self, node, visited=None, parent=None):
...
+            has_write_perm_sqs = Contributor.objects.filter(node=OuterRef('pk'), write=True, user=self.auth.user)
+            children = (AbstractNode.objects
+                        .filter(is_deleted=False, _parents__parent=node)
+                        .annotate(linked_node=Exists(linked_node_sqs))
```
sloria: Minor: `is_linked_node` might be a more appropriate name.

website/static/js/fangorn.js
```diff
@@ -1609,14 +1613,37 @@ function expandStateLoad(item) {
...
+            '/api/v1/project/' + item.data.nodeID + '/files/grid/'
+        ).done(function(xhr) {
+            var data = xhr.data[0].children;
+            tb.updateFolder(data, item);
+            tb.redraw();
+            if (icon.get(0)) {
+                m.render(icon.get(0), tbOptions.resolveToggle(item));
+            }
+        }).fail(function(xhr) {
+            item.notify.update('Unable to retrieve components.', 'danger', undefined, 3000);
```
sloria: We don't expect an error here, so better log this to sentry using Raven.

website/static/js/fangorn.js
```diff
@@ -1609,14 +1613,37 @@ function expandStateLoad(item) {
...
-        $('.fangorn-toolbar-icon').tooltip();
+
+    if (item.depth > 2 && !item.data.isAddonRoot && item.children.length === 0 && item.open) {
+        if (icon.get(0)) {
+            m.render(icon.get(0), tbOptions.resolveRefreshIcon());
+        }
+        $osf.ajaxJSON(
+            'GET',
+            '/api/v1/project/' + item.data.nodeID + '/files/grid/'
+        ).done(function(xhr) {
```
sloria: "xhr" is misleading. The first argument is not an `XHR` object, it's the parsed response data. Maybe "response" or "resp"?

website/static/js/fangorn.js
caseyrollins: This renders the loading indicator until the request a few lines below finishes. It's exactly what treebeard does ([see here](https://github.com/CenterForOpenScience/treebeard/blob/75eca74fc66009e42bb5629481858939524701e0/scripts/grid.js#L1044)), I'm just avoiding the rest of the `toggleFolder` functionality. 

However, the second `m.render` (in the `.done()`) isn't necessary so I'm removing that.

sloria: Ah, ok, thanks. It would be worth adding a comment that this is taken from treebeard.

caseyrollins: Good point. I'll add that real quick.

website/static/js/fangorn.js
```diff
@@ -1622,23 +1622,27 @@ function expandStateLoad(item) {
...
+            var data = response.data[0].children;
             tb.updateFolder(data, item);
             tb.redraw();
-            if (icon.get(0)) {
-                m.render(icon.get(0), tbOptions.resolveToggle(item));
-            }
         }).fail(function(xhr) {
             item.notify.update('Unable to retrieve components.', 'danger', undefined, 3000);
             item.open = false;
+            Raven.captureMessage('Unable to retrieve components for node {}'.format(item.data.nodeID), {
```
sloria: This is JS, not Python =)

caseyrollins: Lol, this is excellent.

PR #7825, opened 2017-10-18
Respond to Reviews CR

laurenbarker:
Before this is merged let Steve know so he can rollback the staging3 database to `0060_reviews` to prevent the name changes from breaking things.

Staging3 doesn't have very many users but if needed, `0061_add_reviews_notification_subscription` can be skipped/faked and subscriptions can be added using `python manage.py add_notification_subscription --notification global_reviews`.

## Purpose

Respond to CR on https://github.com/CenterForOpenScience/osf.io/pull/7807.

Note that this PR doesn't update the myProjects page to use 2.6. I think that would need a regression test from QA which is out of scope of reviews changes. 

## Changes

* Update global_reviews notification display text
* Rename migrations
* Use variable for related node
* Use constant in mails.py
* Change `message` field on `NotificationDigest` to `TextField`
* Add management command for adding subscriptions to users `add_notification_subscription --notification global_reviews`

## Side effects

Probably messes up your migration history. I just deleted the relevant migration history from `django_migrations` table using DataGrip and reran `migrate`.*

**Probably not the recommended way to handle inconsistent history.*


website/mails/mails.py
```diff
@@ -372,3 +372,8 @@ def get_english_article(word):
     'send_data_share_preprint_error_desk',
     subject='Share Error'
 )
+
+REVIEWS_SUBMISSION_CONFIRMATION = lambda provider_name: Mail(
```
sloria: The `lambda` shouldn't be necessary. The subject can be a string template that receives the same context as the email body.

For example

```python
FILE_OPERATION_SUCCESS = Mail(
    'file_operation_success',
    subject='Your ${action} has finished',
)
```

reviews/models/mixins.py
```diff
@@ -272,9 +272,10 @@ def reviews_submit_notification(self, context):
         user_subscriptions = get_user_subscriptions(user, event_type)
         context['no_future_emails'] = user_subscriptions['none']
         context['is_creator'] = user == context.get('reviewable').node.creator
+        context['provider_name']= context.get('reviewable').provider.name
```
sloria: Is `reviewable` guaranteed to be in the context?

If so, then context['reviewable'] is preferred, to make it clear that it's guaranteed.
If not, then this code will need to be changed to handle the case when it's not there.

sloria:
OK, stage3 is now rolled back to 0060. Let's move on..

PR #7828, opened 2017-10-19
[#OSF-8700]Make support email address configurable

chennan47:
## Purpose
Make support email address configurable, prevent local and staging send emails to help desk and provide full url links to help desk for failed registration.

## Changes
All support@cos.io and support@osf.io usage has been removed from database. And all will use the variable from settings.py to set those two across the code base.

## Side effects

After this change, 'support@osf.io' is hard coded only in osfHelpers.js since osfLanguage.js cannot have window imported before the function start. 

## Ticket
https://openscience.atlassian.net/browse/OSF-8700


website/settings/defaults.py
```diff
@@ -122,7 +122,11 @@ def parent_dir(path):
 
 USE_EMAIL = True
 FROM_EMAIL = 'openscienceframework-noreply@osf.io'
-SUPPORT_EMAIL = 'support@osf.io'
+
+# support email
+OSF_SUPPORT_EMAIL = 'support@osf.io'
```
brianjgeiger: After this is launched, I think we'll make sure all of the environments are up-to-date with an override for this, then do a follow-up ticket that makes the default something else. In the meantime, let's add to the local-dist.py fake-support@osf/cos.io for these two variables.

website/settings/defaults.py
```diff
@@ -129,7 +129,11 @@ def parent_dir(path):
...
+COS_SUPPORT_EMAIL = 'support@cos.io'
```
alexschiller: So I wasn't 100% clear in the ticket, but the intention was for support@cos.io and support@osf.io to be replaced with the single variable. support@cos.io goes to the same place as support@osf.io and there is not a consistent pattern to where they occur so support@cos.io can be removed.

brianjgeiger:
@chennan47 Conflicts

brianjgeiger:
@chennan47 The error on branch 3 is "Uncaught ReferenceError: SUPPORT_LINK is not defined
  at webpack:///website/static/js/osfLanguage.js:69:0 <-website/static/js/tests/tests.webpack.js:113816" which sounds like it could be related to your PR.

chennan47:
fixed in 1b906ab

PR #7831, opened 2017-10-19
Feature/Add Admins from Parent to Child [OSF-8594] [PLAT-75]

pattisdr:
~~❗️ Is dependent on bug fix https://github.com/CenterForOpenScience/osf.io/pull/8097. If that work is merged, I need to remove this commit: https://github.com/CenterForOpenScience/osf.io/pull/7831/commits/22b8803ba4be726b72f852dc4d5b500bae1a7b11~~

## Purpose

 Add admins from the parent project to a component in one-click.

## Changes

- Put an Add button beside Admins on Parent Projects.  Clicking on the Add button adds the contributor directly to the project with *read* permissions and visible = *True*.
- Created new knockout computed observable for whether or not this "Add" button is visible.  To add an admin on the parent project to a component, the current user must be an admin on the component, the component cannot be a registration, and the user being added must be an admin on the parent project.
- Fixed bug in logic on whether "Remove" button is visible beside a contributor.  Previously, if the logged in user was also the admin on the parent project, and navigated directly to that component's page, they would see a Remove button by their name.  This user has implicit read and should not see that read button. (The read button didn't actually do anything, but it shouldn't be there). 
- Added `parent_admin_contributors` field to NodeModel that only returns contributors that are on admins on a parent component, excluding admins of current node.  Fixes error introduced here: https://github.com/CenterForOpenScience/osf.io/commit/bc85642f54b9a7c226d30f810fce5cbb23b3b5fd
## Demo
giphy 6

## Demo - scenario where request to add contributor fails
giphy 2

## Bug fixed
When logged in as Test User, there is a remove button by my name.  This remove button should not be here and is fixed in this PR.  Admin Parent contributors list currently includes all admins from current node, not just parent admins. 

screen shot 2017-10-19 at 4 51 28 pm

## Side effects
Should be none.

## Ticket

https://openscience.atlassian.net/browse/OSF-8594
https://openscience.atlassian.net/browse/PLAT-75


website/static/js/contribManager.js
```diff
@@ -113,6 +116,37 @@ var ContributorModel = function(contributor, currentUserCanEdit, pageOwner, isRe
             id:self.id});
     };
 
+    self.addParentAdmin = function () {
+        // Immediately adds parent admin to the component with permissions=read and visible=True
+        var self = this;
```
caseyrollins: I don't think this declaration is necessary. 

`self = this` is set for the entire ViewModel [here](https://github.com/CenterForOpenScience/osf.io/pull/7831/files#diff-f4df2c5c528f874302fff1ee05a65e8aR42).

website/static/js/contribManager.js
```diff
@@ -113,6 +116,37 @@ var ContributorModel = function(contributor, currentUserCanEdit, pageOwner, isRe
...
+        $osf.block();
+        self.nodeId = window.contextVars.node.id;
```
caseyrollins: I haven't looked at Javascript in *quite* sometime, but why
```
self.nodeId = ...
self.nodeApiUrl = ...
``` 
instead of 
``` 
var nodeId = ...
var nodeApiUrl = ...
```
?

If these variables need to be defined on the `ContributorModel`, can they be placed outside of this function?

website/static/js/contribManager.js
pattisdr: ah, thanks!

website/static/js/contribManager.js
```diff
@@ -125,7 +159,11 @@ var ContributorModel = function(contributor, currentUserCanEdit, pageOwner, isRe
     self.profileUrl = ko.observable(contributor.url);
 
     self.canRemove = ko.computed(function(){
-        return (self.id === pageOwner.id) && !isRegistration;
+        return (self.id === pageOwner.id) && !isRegistration && !self.isAdmin;
+    });
+
+    self.canAddAdminContrib = ko.computed(function() {
+        return self.currentUserCanEdit && self.isAdmin;
```
caseyrollins: Are both of these conditions necessary? I would've thought that `isAdmin` encapsulates `currentUserCanEdit` (assuming `isAdmin` is referring to the current user as well).

website/static/js/contribManager.js
```diff
@@ -113,6 +116,37 @@ var ContributorModel = function(contributor, currentUserCanEdit, pageOwner, isRe
...
     };
 
+    self.addParentAdmin = function () {
+        // Immediately adds parent admin to the component with permissions=read and visible=True
+        var self = this;
+        $osf.block();
+        self.nodeId = window.contextVars.node.id;
+        self.nodeApiUrl = '/api/v1/project/' + self.nodeId + '/';
+        var url = self.nodeApiUrl + 'contributors/';
+        var userData = JSON.parse(ko.toJSON(self));
```
caseyrollins: It looks like [`self.serialize()`](https://github.com/CenterForOpenScience/osf.io/pull/7831/files#diff-f4df2c5c528f874302fff1ee05a65e8aR105) is a helper for `ko.toJSON(self)`and could be used instead.

caseyrollins: Also, I'm not sure that I'm following all of the logic correctly, but is it necessary to serialize the entire view model or would something like `ko.toJSON(self.contributor)` work?

pattisdr:
Thanks for the review @caseyrollins! you brought up some stuff I hadn't thought about!

website/static/js/contribManager.js
pattisdr: thank you, they don't need to be defined on the ContributorModel, I don't really need to define them at all, since they are only used once.  I'll move this to where it's used.

website/static/js/contribManager.js
pattisdr: There's no self.contributor, but the 'contributor' here I think is a ContributorModel, which is the same thing as self. So probably just `self.serialize()`, like you suggested would work.

website/static/js/contribManager.js
pattisdr: isAdmin refers to the user you are intending to add, not the current user.  I will add some comments, maybe change variable names to clarify this.

osf/models/node.py
```diff
@@ -902,12 +902,34 @@ def get_admin_contributor_ids(node):
             admin_ids.update(set(admins).difference(contributor_ids))
         return admin_ids
 
+    @property
+    def parent_admin_contributor_ids(self):
```
caseyrollins: I think these properties could be a bit DRYer. You could add a helper function with the duped code and each property could just call that function with the appropriate params. Something like this (extremely rough pseudocode where my logic is probably backwards, but you get the idea):
```
def _get_admin_contributor_ids(node, include_self=False):
    def get_admin_contributor_ids(node):
        return ...
    contributor_ids = ...
    admin_ids = set() if not include_self else set(get_admin_contributor_ids(node)
    ...

@property
def admin_contributor_ids(self):
    return _get_admin_contributor_ids(self, include_self=True)

@property
def parent_admin_contributor_ids(self):
    return _get_admin_contributor_ids(self, include_self=True)
```

pattisdr:
Just waiting for tests to pass @caseyrollins

PR #7832, opened 2017-10-19
[Improvement][MOD-150] Use existing notify method for reviews

sheriefvt:
## Purpose

Create new method for notifications that serve both reviews and mentions notifications. 

## Changes

- Update reviews model mixin
- Create notify_global_event method to handle global_reviews/mention event.
- Update existing methods (store, notify_mention)
- Rename mentions email template files
- Update tests

## Ticket
https://openscience.atlassian.net/browse/MOD-150


reviews/models/mixins.py
```diff
@@ -244,30 +245,15 @@ def get_context(self):
...
-                    timestamp=timestamp,
-                    send_type=send_type,
-                    event='global_reviews',
-                    message=message,
-                    node_lineage=node_lineage_ids
-                )
-                digest.save()
+def reviews_notification(self, context, node):
+    user = context['email_sender']
+    auth = Auth(user)
```
aaxelb: It looks like `auth` is unused except for `auth.user` -- could we just use `user`?

reviews/models/mixins.py
```diff
@@ -244,30 +245,15 @@ def get_context(self):
...
+    time_now = timezone.now()
+    email_recipients = context['email_recipients']
+    email_recipients.remove(user._id)  # remove email sender
+    for user_id in email_recipients:
+        context['target_user'] = OSFUser.load(user_id)
```
aaxelb: In `notify` it says `target_user` is "The user who the current comment is a reply to", which doesn't seem right for our use. Seems like we want to pass as `user` the user who triggered this transition (the moderator who made a decision, or the user who resubmitted a preprint), which is available in the transition callback as `ev.kwargs['user']`

reviews/models/mixins.py
```diff
@@ -244,30 +245,15 @@ def get_context(self):
...
-                    node_lineage=node_lineage_ids
-                )
-                digest.save()
+def reviews_notification(self, context, node):
+    user = context['email_sender']
+    auth = Auth(user)
+    time_now = timezone.now()
+    email_recipients = context['email_recipients']
+    email_recipients.remove(user._id)  # remove email sender
+    for user_id in email_recipients:
```
aaxelb: `notify` finds users who are subscribed to an event on a node. It seems like it's meant to be called just once for the event, not once per email recipient.

website/notifications/emails.py
```diff
@@ -87,7 +88,8 @@ def store_emails(recipient_ids, notification_type, event, user, node, timestamp,
     if notification_type == 'none':
         return
 
-    template = event + '.html.mako'
+    # Reviews has three email templates for the global_reviews event.
+    template = context['template'] + '.html.mako' if event == 'global_reviews' else event + '.html.mako'
```
aaxelb: I don't think we should have logic that depends on specific event names. Instead, maybe do something like `context.get('template', event)`, allowing anyone to pass in a custom template if they wish. Or, to be safer in case anyone ever actually has a `'template'` key in their context, add `template=None` as a parameter to `notify`.

website/notifications/emails.py
```diff
@@ -43,7 +43,8 @@ def notify(event, user, node, timestamp, **context):
         # If target, they get a reply email and are removed from the general email
         if target_user and target_user._id in subscriptions[notification_type]:
             subscriptions[notification_type].remove(target_user._id)
-            store_emails([target_user._id], notification_type, 'comment_replies', user, node, timestamp, **context)
+            event_name = 'global_reviews' if event_type == 'global_reviews' else 'comment_replies'
```
aaxelb: Same as below, avoid logic that depends on event names.

website/templates/emails/reviews_resubmission_confirmation.html.mako
```diff
@@ -1,11 +1,30 @@
...
-        <p>${reviewable.provider.name} has chosen to moderate their submissions using a pre-moderation workflow, which means your submission is pending until accepted by a moderator. You will receive a separate notification informing you of any status changes.</p>
-        <p>You will ${'not receive ' if no_future_emails else 'be automatically subscribed to '}future notification emails for this ${reviewable.provider.preprint_word}. Each ${reviewable.provider.preprint_word} is associated with a project on the Open Science Framework for managing the ${reviewable.provider.preprint_word}. To change your email notification preferences, visit your <a href="${domain + 'settings/notifications/'}">project user settings</a>.</p>
-        <p>If you have been erroneously associated with "${reviewable.node.title}", then you may visit the project's "Contributors" page and remove yourself as a contributor.</p>
-        <p>For more information about ${reviewable.provider.name}, visit <a href="${provider_url}">${provider_url}</a> to learn more. To learn about the Open Science Framework, visit <a href="https://osf.io/">https://osf.io/</a>.</p>
+        <p>
```
aaxelb: ✨ nice ✨

reviews/models/mixins.py
sheriefvt: This condition was confusing me in notify method:
    if target_user and event_type in constants.USER_SUBSCRIPTIONS_AVAILABLE:
        # global user
        subscriptions = get_user_subscriptions(target_user, event_type)
    else:
        # local project user
        subscriptions = compile_subscriptions(node, event_type, event)
it seems that nofiy get subscriptions global if target_user has value, since we use now using global subscriptions not local project one, I pass the recepient as a target user in each loop.

website/notifications/emails.py
```diff
@@ -52,21 +52,22 @@ def notify(event, user, node, timestamp, **context):
...
-            if (
-                notification_type != 'none' and
-                subscriptions[notification_type] and
-                m in subscriptions[notification_type]
-            ):
-                store_emails([m], notification_type, 'mentions', user, node,
-                                 timestamp, **context)
-                sent_users.extend([m])
+            if (notification_type != 'none' and subscriptions[notification_type] and target_id in subscriptions[notification_type]):
+                store_emails([target_id], notification_type, event, sender_user, node, timestamp, **context)
```
aaxelb: Minor (feel free to ignore): Since `store_emails` can take a list of recipients, it seems like it would be nicer to first collect the user subscriptions together, then call `store_emails` just once per `notification_type`

reviews/models/mixins.py
```diff
@@ -244,30 +244,17 @@ def get_context(self):
...
-                )
-                digest.save()
+def reviews_notification(self, creator, context, node):
+    time_now = timezone.now()
+    email_recipients = context['email_recipients']
+    emails.notify_global_event(
+        event='global_reviews',
+        sender_user=creator,
+        node=node,
+        timestamp=time_now,
```
aaxelb: It might make more sense to use the action's `date_created` as the timestamp, since this code might run somewhat later.

website/notifications/emails.py
```diff
@@ -52,25 +52,40 @@ def notify(event, user, node, timestamp, **context):
...
     sent_users = []
-    new_mentions = context.get('new_mentions', [])
-    for m in new_mentions:
-        mentioned_user = OSFUser.load(m)
-        subscriptions = get_user_subscriptions(mentioned_user, event_type)
+
+    # Initialize the subscriptions dict
+    users_subscriptions = {}
+    for key in constants.NOTIFICATION_TYPES:
+        users_subscriptions[key] = []
```
aaxelb: In case there are duplicates, this might be a good time to use a `set()`

website/notifications/emails.py
```diff
@@ -52,25 +52,40 @@ def notify(event, user, node, timestamp, **context):
...
+def notify_global_event(event, sender_user, node, timestamp, recipients, template=None, context=None):
     event_type = utils.find_subscription_type(event)
     sent_users = []
-    new_mentions = context.get('new_mentions', [])
-    for m in new_mentions:
-        mentioned_user = OSFUser.load(m)
-        subscriptions = get_user_subscriptions(mentioned_user, event_type)
+
+    # Initialize the subscriptions dict
+    users_subscriptions = {}
```
aaxelb: FYI, dictionary comprehensions are neat. This could be:
```py
{key: [] for key in constants.NOTIFICATION_TYPES}
```

website/notifications/emails.py
```diff
@@ -52,25 +52,40 @@ def notify(event, user, node, timestamp, **context):
...
         for notification_type in subscriptions:
-            if (
-                notification_type != 'none' and
-                subscriptions[notification_type] and
-                m in subscriptions[notification_type]
-            ):
-                store_emails([m], notification_type, 'mentions', user, node,
-                                 timestamp, **context)
-                sent_users.extend([m])
+            if (notification_type != 'none' and subscriptions[notification_type] and recipient._id in subscriptions[notification_type]):
```
aaxelb: If you do use `set`s, it might be cleaner to get all the guids together first:
```py
for notification_type, user_guids in subscriptions.items():
    users_subscriptions[notification_type].update(user_guids)
```
then limit to the recipients below:
```py
for notification_type, user_ids in users_subscriptions.items():
    send_to = user_ids.intersection(recipient_ids)
    sent_users.extend(send_to)
    store_emails(...)
```
or something like that. Just a matter of taste, but I want to encourage you to rethink the logic that was already there, and always try to leave code cleaner than you found it.

website/notifications/emails.py
```diff
@@ -52,25 +52,40 @@ def notify(event, user, node, timestamp, **context):
...
+                users_subscriptions[notification_type].append(recipient._id)
+                if (sender_user._id != recipient):
+                    sent_users.append(recipient._id)
+
+    # For each notification type store the list of users
+    for type in users_subscriptions:
```
aaxelb: Avoid using built-in names like `type`, it can be confusing.

website/notifications/emails.py
```diff
@@ -87,7 +102,9 @@ def store_emails(recipient_ids, notification_type, event, user, node, timestamp,
     if notification_type == 'none':
         return
 
-    template = event + '.html.mako'
+    # Reviews has multiple email templates for the global_reviews event.
+    template = '{template}.html.mako'.format(template=template) if template else '{event}.html.mako'.format(event=event)
```
aaxelb: Try not to repeat yourself, this could be condensed:
```py
'{template}.html.mako'.format(template=template or event)
```

website/notifications/emails.py
```diff
@@ -87,7 +102,9 @@ def store_emails(recipient_ids, notification_type, event, user, node, timestamp,
     if notification_type == 'none':
         return
 
-    template = event + '.html.mako'
+    # Reviews has multiple email templates for the global_reviews event.
```
aaxelb: Rather than a reviews-specific comment, explain how the arguments are used. Maybe something like "If \`template\` is not specified, default to using a template with name \`event\`"

website/notifications/emails.py
```diff
@@ -52,25 +52,37 @@ def notify(event, user, node, timestamp, **context):
...
+        users_subscriptions.append(get_user_subscriptions(recipient, event_type))
+
+    # Merge the dictionaries using dict comprehension and exclude 'none' notification type.
+    # Output e.g. {'email_transactional': [u'vsu7t', u'evz43'], 'email_digest': [u'39dbp']}
+    merged_user_subscriptions = {k: [d[k][0] for d in users_subscriptions if d[k]] for k in users_subscriptions[0] if k != 'none'}
+
+    # For each notification type store the list of users
+    for notify_type in merged_user_subscriptions:
+        # Group of users assigned this notification type
+        user_group_type = merged_user_subscriptions[notify_type]
```
aaxelb: Iterating through a dictionary, using both keys and values, I think it's better to do:
```py
for notify_type, user_group_type in merged_user_subscriptions.items():
```
instead of looking up each key separately.

website/notifications/emails.py
```diff
@@ -52,25 +52,37 @@ def notify(event, user, node, timestamp, **context):
...
-                sent_users.extend([m])
+    users_subscriptions = []
+
+    # Create a list of users' subscription dictionaries
+    for recipient in recipients:
+        users_subscriptions.append(get_user_subscriptions(recipient, event_type))
+
+    # Merge the dictionaries using dict comprehension and exclude 'none' notification type.
+    # Output e.g. {'email_transactional': [u'vsu7t', u'evz43'], 'email_digest': [u'39dbp']}
+    merged_user_subscriptions = {k: [d[k][0] for d in users_subscriptions if d[k]] for k in users_subscriptions[0] if k != 'none'}
```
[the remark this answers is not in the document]
sheriefvt: yes for each user, we will have one element per notification type

reviews/models/mixins.py
```diff
@@ -244,45 +251,30 @@ def get_context(self):
...
-                digest = NotificationDigest(
-                    user=user,
-                    timestamp=timestamp,
-                    send_type=send_type,
-                    event='global_reviews',
-                    message=message,
-                    node_lineage=node_lineage_ids
-                )
-                digest.save()
+def reviews_notification(self, creator, recipients, time_now, template, context, node):
```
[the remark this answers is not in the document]
sheriefvt: how can we get recipients from action, currently i do            
 recipients = list(self.reviewable.node.contributors)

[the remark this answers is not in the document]
aaxelb: `action.target.node.contributors`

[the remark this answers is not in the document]
sheriefvt: thanks

PR #7837, opened 2017-10-23
Feature/Always Return Preprint_DOI link in Published Preprints [OSF-8819]

pattisdr:
## Purpose

Preprint DOIs should be displayed immediately after they are published, regardless if doi has been sent to datacite. Change the API to always return preprint_doi_link in the response of published preprints. Create nightly script to find any published preprints that are missing their DOI and retry the EZID request

## Changes

- Modify `preprint_doi` SerializerMethodField on PreprintSerializer to return a preprint_doi link on published preprints.  If preprint doi has not already been sent to datacite, build the link ourselves and return it.
- Add `preprint_doi_created` field to PreprintService model with default value of None.  When the DOI is created, set this field to the current datetime
- Add date migration to go through all published preprints with doi identifiers and set preprint_doi_created field to date_published.
- Add 3 PreprintDetail api tests for testing 1) unpublished preprints do not return a preprint_doi  link 2) preprint_doi link exists on published preprints when preprint doi has not actually been created (`preprint_doi_created` is null)  3) preprint_doi link also exists on published preprints when `preprint_doi_created` field is populated
- Add script for remaking requests to datacite to retry EZID request on preprints that don't have doi.  Scheduled to run nightly. 

## Screenshot: preprint_doi link being returned on published preprint even though `preprint_date_created` field is null.
screen shot 2017-11-27 at 12 52 28 pm

## Screenshot: preprint_doi link being returned on published preprint where doi has been set in data migration. `preprint_doi_created` is equal to `date_published`
screen shot 2017-11-27 at 12 52 10 pm

## Side effects

API is returning `preprint_doi` link under `PreprintSerializer > Links` on published preprints, regardless of whether the DOI has actually been created.  This means that link will not always work.  The presence of the `preprint_doi_created` field can be used to determine if the link should be shown or not.  If `preprint_doi_created` is null, DOI has not actually been created yet.

## Ticket
https://openscience.atlassian.net/browse/OSF-8819


api/preprints/serializers.py
```diff
@@ -81,6 +82,7 @@ class PreprintSerializer(JSONAPISerializer):
     description = ser.CharField(required=False, allow_blank=True, allow_null=True, source='node.description')
     tags = JSONAPIListField(child=NodeTagField(), required=False, source='node.tags')
     node_is_public = ser.BooleanField(read_only=True, source='node__is_public')
+    preprint_doi_on_datacite = ser.SerializerMethodField(read_only=True)
```
sloria: For performance, I think this should only be returned in the preprint detail endpoint. Otherwise, we'd be adding at least one query per preprint on the preprint list endpoint.

I'll confirm with @jamescdavis that it's only needed on the detail endpoint.

api_tests/preprints/views/test_preprint_detail.py
```diff
@@ -94,6 +103,39 @@ def test_embed_contributors(self, app, user, preprint):
         for contrib in embeds['contributors']['data']:
             assert contrib['id'] in ids
 
+    def test_preprint_doi_link_absent_in_unpublished_preprints(self, app, user, unpublished_preprint, unpublished_url):
+        res = app.get(unpublished_url, auth=user.auth)
+        assert res.json['data']['id'] == unpublished_preprint._id
+        assert res.json['data']['attributes']['is_published'] == False
```
sloria: Minor: Use `is` to compare with `False`.

api_tests/preprints/views/test_preprint_detail.py
```diff
@@ -94,6 +103,39 @@ def test_embed_contributors(self, app, user, preprint):
...
+        assert 'self' in res.json['data']['links'].keys()
+        assert 'html' in res.json['data']['links'].keys()
```
sloria: No need to duplicate these asserts across tests. Better to have the tests be focused on a few things so that when they fail, it's more obvious where to look for the bug.

api/preprints/serializers.py
sloria: OK, discussed with James. He pointed that for consistency, it would be better to return the flag in both the list and the detail endpoint.

To avoid having to do n queries in the list endpoint, I recommend

- Store the datetime the DOI was created in a new field `preprint_doi_created`. Make the value `null` if the DOI hasn't been created
- Expose that datetime in this serializer

pattisdr:
Thanks @sloria, will address later this afternoon

pattisdr:
Have been sidetracked with Lookit, but next steps would be:

- Bring PR up to date
- Adding preprint_doi_created to preprint model w/ default value of null and creating a schema migration
- Data migration to go through preprints and populate preprint_doi_created with preprint's published datetime
- Where preprint doi is actually created, preprint_doi_created needs to be populated at that time.  The nightly script that runs to catch missed doi's needs to do this as well.

pattisdr:
Update: working on this today, should have changes up by this afternoon 11/27

pattisdr:
Ready for another pass @sloria

osf/migrations/0069_auto_20171127_1119.py
```diff
@@ -0,0 +1,51 @@
...
+def add_preprint_doi_created(apps, schema_editor):
+    """
+    Sets preprint_doi_created equal to date_published for existing published preprints.
+    """
+    null_preprint_doi_created = PreprintService.objects.filter(preprint_doi_created__isnull=True, date_published__isnull=False)
+    preprints_count = null_preprint_doi_created.count()
+    current_preprint = 0
+    logger.info('{} published preprints found with preprint_doi_created is null.'.format(preprints_count))
+
+    for preprint in null_preprint_doi_created:
```
sloria: Rather than looping and doing n queries, you should be able to do one query:

```python
PreprintService.objects.filter(...).update(preprint_doi_created=F('date_published')
```

osf/migrations/0069_auto_20171127_1119.py
```diff
@@ -0,0 +1,51 @@
...
+def reverse_func(apps, schema_editor):
+    """
+    Reverses data migration. Sets preprint_doi_created field back to null.
+    """
+    preprint_doi_created_not_null = PreprintService.objects.filter(preprint_doi_created__isnull=False)
+    preprints_count = preprint_doi_created_not_null.count()
+    current_preprint = 0
+    logger.info('Reversing preprint_doi_created migration.')
+
+    for preprint in preprint_doi_created_not_null:
```
sloria: Same here:

```python
PreprintService.objects.filter(...).update(preprint_doi_created=None)
```

osf/migrations/0069_auto_20171127_1119.py
pattisdr: @sloria, I wanted to check whether the DOI had actually been created before populating the preprint_doi_created field, which was the main reason for the loop.

sloria: Ah ok.

PR #7838, opened 2017-10-24
Feature/add preprint original publication date [OSF-8782]

laurenrevere:
## Purpose

Preprints need a field to list a publication date for if the "preprint" was first posted or published somewhere else. This allows the original publication date to be used for the citations we generate

## Changes

I added a field called `original_publication_date` was added to the `preprint_service` model. This field can be changed when creating or updating a preprint through the api.

In citations, if a preprint has an `original_publication_date` that is used instead of the latest update date to the node.

This PR also adds filtering by `original_publication_date`.

## Side effects
None!

## Ticket

https://openscience.atlassian.net/browse/OSF-8782

screen shot 2017-10-24 at 1 51 25 pm


api_tests/preprints/views/test_preprint_citations.py
```diff
@@ -49,9 +50,23 @@ def test_citation_url_is_preprint_url_not_project(self):
...
 		self.published_preprint_url = '/{}preprints/{}/citation/apa/'.format(API_BASE, self.published_preprint._id)
 		self.unpublished_preprint_url = '/{}preprints/{}/citation/apa/'.format(API_BASE, self.unpublished_preprint._id)
+
+	def test_citation_contains_correct_date(self):
+		res = self.app.get(self.published_preprint_url)
+		assert_equal(res.status_code, 200)
+		expected_date = self.published_preprint.date_modified.strftime('%Y, %B %-d')
+		assert_true(expected_date in res.json['data']['attributes']['citation'])
+
+		self.published_preprint.original_publication_date = datetime(2017, 12, 24)
```
brianjgeiger: If this is a datetime rather than just a date, be sure to use a timezone for this, just to ensure that there won't be any future test problems. It's not risky in this case, but just in case.  If it definitely doesn't matter because we are naively just chopping off the time and tz info or something, then feel free to ignore.

laurenrevere: No time info is returned for these citations. They only return date info.

osf/models/node.py
```diff
@@ -656,7 +656,12 @@ def csl(self):  # formats node information into CSL format for citation parsing
             csl['DOI'] = doi
 
         if self.logs.exists():
-            csl['issued'] = datetime_to_csl(self.logs.latest().date)
+            citation_date = self.logs.latest().date
+
+            if self.is_preprint and self.preprints.get_queryset()[0].original_publication_date:
```
alexschiller: Pretty sure I'm the reason `self.preprints.get_queryset()[0]` exists  -- You should be able to just do `self.preprints.first()` instead. 

Also take a look at [is_preprint](https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/models/node.py#L437) -- it is a bit misleading since it has some particular cases in which it is is false (e.g., the node is made private).

You might be better off doing:

```
if self.preprints.exists():
    citation_date = self.preprints.first().original_publication_date
```

osf/models/node.py
```diff
@@ -656,10 +656,14 @@ def csl(self):  # formats node information into CSL format for citation parsing
             csl['DOI'] = doi
 
         if self.logs.exists():
-            citation_date = self.logs.latest().date
+            citation_date = self.date_modified
```
brianjgeiger: Could you really quick check with @mfraezz and see if there's a way to do this that will work better with his upcoming change? There might be an alternate field in the near future that you could use instead.

mfraezz: Currently, the future-proof way is to keep it as `logs.latest().date`. 

In the hopefully near future, there will be a `last_logged` field that prevents having to do a related lookup, but this will break when `date_modified` becomes `modified`. Then again, so will `preprint.date_modified` below. Please ensure that conditional branch is covered by a test

PR #7841, opened 2017-10-25
[OSF-8848] Update docker and docker sync settings for ember

adlius:
## Purpose

Update the yml files for local dev env setups.

## Changes

1. In `docker-compose.override.yml`, command for `emberosf`, `preprints` and `registries` are changed to make sure `node_modules` and `bower_components` folders are deleted upon restarting the container.

2. In `docker-sync.yml`, the name for local repo of preprints container is changed to reflect recent changes to the remote repo name.

## Side effects

None.

## Ticket

https://openscience.atlassian.net/browse/OSF-8848

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -69,9 +69,10 @@ services:
 ##        yarn link &&
 ##        cd /code &&
 ##        (rm -r node_modules || true) &&
-##        ./node_modules/.bin/bower install --allow-root --config.interactive=false &&
 ##        yarn --frozen-lockfile &&
 ##        yarn link @centerforopenscience/ember-osf &&
+##        (rm -r bower_components || true) &&
+##        ./node_modules/.bin/bower install --allow-root --config.interactive=false &&
 ##        yarn start --host 0.0.0.0 --port 4201 --live-reload-port 41954
 
 #  registries:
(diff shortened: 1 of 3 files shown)


docker-compose.override.yml
```diff
@@ -120,6 +123,8 @@ services:
 #      - /bin/bash
 #      - -c 
 #      - yarn --frozen-lockfile --ignore-engines &&
+#       (rm -r node_modules || true) &&
```
jamescdavis: Need to do this before `yarn --frozen-lockfile --ignore-engines`

docker-compose.override.yml
```diff
@@ -88,9 +89,11 @@ services:
 ##      - cd /ember-osf &&
 ##        yarn link &&
 ##        cd /code &&
-##        yarn link @centerforopenscience/ember-osf &&
+##        (rm -r node_modules || true) &&
+##        (rm -r bower_components || true) &&
 ##        ./node_modules/.bin/bower install --allow-root --config.interactive=false &&
 ##        yarn --frozen-lockfile &&
```
jamescdavis: This needs to run before `bower install` because yarn installs bower.

docker-compose.override.yml
```diff
@@ -68,9 +68,10 @@ services:
 ##      - cd /ember-osf &&
 ##        yarn link &&
 ##        cd /code &&
-##        yarn link @centerforopenscience/ember-osf &&
+##        (rm -r node_modules || true) &&
 ##        ./node_modules/.bin/bower install --allow-root --config.interactive=false &&
 ##        yarn --frozen-lockfile &&
```
jamescdavis: This needs to run before `bower install` because yarn installs bower.

docker-compose.override.yml
```diff
@@ -118,7 +121,9 @@ services:
 #    build: ../ember-osf
 #    command:
 #      - /bin/bash
-#      - -c 
+#      - -c
+#       (rm -r node_modules || true) &&
```
jamescdavis: This needs to be:
```
- (rm -r node_modules || true) &&
```

docker-compose.override.yml
```diff
@@ -118,7 +121,9 @@ services:
...
+#       (rm -r bower_components || true) &&
 #      - yarn --frozen-lockfile --ignore-engines &&
```
jamescdavis: Remove dash here.

PR #7842, opened 2017-10-26
[hotfix] External storage speed fix [OSF-8794]

alexschiller:
## Purpose
External providers are super duper slow when there are lots of files. It scales linearly with the number of files because we update ALL files every time we hit a list endpoint. This PR makes it at least faster than it was before.

## Changes
screen shot 2017-10-26 at 11 41 30 am

1. Use bulk_update to save everything at one time
2. Use bulk_create to create all new objects at one time
3. Remove permission check on every file (permissions are already checked earlier)

## Side effects
None Known

## Ticket

https://openscience.atlassian.net/browse/OSF-8794

## QA Notes
1. It is important to test files that have never been viewed and those that have been viewed once. I'll detail an example of how to test below. I'm most concerned about dataverse and figshare, since they had a small change special casing that I did not test locally.

2. Testing the speed improvements is not 100% necessary, mostly making sure that no views on the API fail is the most important part. Testing at least one folder with 200+ files may be worth doing. I can help you make a bunch of empty files very quickly if helpful.

3. Confirm that permissions are still handled properly (since a permissions check was removed)

So for each storage provider (not including osfstorage), here is a recommended workflow:
1. Connect an addon to a new private project and view a folder that contains multiple files via the API
2. Go to a list view for that provider
GitHub example: https://staging-api.osf.io/v2/nodes/f27hn/files/github/ten/

3. Confirm that the "last_touched" date is now
4. Refresh the page and confirm that it doesn't break and the "last_touched" date changes too
5. Open the same list view link in an incognito window and confirm permission error
6. Navigate to a file detail page
GitHub example: https://staging-api.osf.io/v2/nodes/f27hn/files/github/ten/1.txt

7. Confirm that "last_touched" date is now
8. Open the same detail page link in an incognito window and confirm permission error

### DIFF
--- a/api/base/views.py
+++ b/api/base/views.py
@@ -908,14 +908,17 @@ class WaterButlerMixin(object):
                 # create method on BaseFileNode appends provider, bulk_create bypasses this step so it is added here
                 file_obj = base_class(node=node, _path='/' + attrs['path'].lstrip('/'), provider=base_class._provider)
                 objs_to_create[base_class].append(file_obj)
+            else:
+                file_objs.append(file_obj)
 
             file_obj.update(None, attrs, user=self.request.user, save=False)
-            file_objs.append(file_obj)
+
+        bulk_update(file_objs)
 
         for base_class in objs_to_create:
             base_class.objects.bulk_create(objs_to_create[base_class])
+            file_objs += objs_to_create[base_class] 
 
-        bulk_update(file_objs)
         return file_objs
 
     def get_file_node_from_wb_resp(self, item):
--- a/api/base/views.py
+++ b/api/base/views.py
@@ -1,3 +1,4 @@
+from bulk_update.helper import bulk_update
 from django.conf import settings as django_settings
 from django.db import transaction
 from django.http import JsonResponse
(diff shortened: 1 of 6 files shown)


api/nodes/views.py
```diff
@@ -1925,9 +1925,8 @@ def get_default_queryset(self):
             # Resolve to a provider-specific subclass, so that
             # trashed file nodes are filtered out automatically
             ConcreteFileNode = BaseFileNode.resolve_class(provider, BaseFileNode.ANY)
-            return ConcreteFileNode.objects.filter(
-                id__in=[self.get_file_item(file).id for file in files_list],
-            )
+            file_ids = [f.id for f in self.bulk_get_file_nodes_from_wb_resp(files_list)]
```
brianjgeiger: Just verifying that there's not a better way to do this. Since this seems to be the only thing currently using the function, it could potentially output the ids. One way is, since you're going through all the items in the bulk_get_etc function, you could just construct and return the ids as a list rather than the objects.  If we're expecting a use for the actual file objects at some point, then this would be an excessive optimization.

api/base/views.py
```diff
@@ -879,7 +882,44 @@ class WaterButlerMixin(object):
...
-    def get_file_item(self, item):
+    def bulk_get_file_nodes_from_wb_resp(self, files_list):
+        """Takes a list of file data from wb response, touches/updates metadata for each, and returns list of file objects.
+        This function mirrors all the actions of get_file_node_from_wb_resp except the create and updates are done in bulk.
+        The bulk_update and bulk_create do not call the base class update and create so the actions of those functions are
+        done here where needed
+        """
+        node = self.get_node(check_object_permissions=False)
+
+        objs_to_create = defaultdict(lambda: [])
```
sloria: I think this needs to be defaultdict(list) to prevent modifying the same list.

api/base/views.py
```diff
@@ -879,7 +882,44 @@ class WaterButlerMixin(object):
...
+
+            # mirrors BaseFileNode get_or_create
+            try:
+                file_obj = base_class.objects.get(node=node, _path='/' + attrs['path'].lstrip('/'))
+            except base_class.DoesNotExist:
+                # create method on BaseFileNode appends provider, bulk_create bypasses this step so it is added here
+                file_obj = base_class(node=node, _path='/' + attrs['path'].lstrip('/'), provider=base_class._provider)
+                objs_to_create[base_class].append(file_obj)
+
+            file_obj.update(None, attrs, user=self.request.user, save=False)
```
sloria: When do these filenodes get saved?

sloria: Oh, nvm; I didn't see the `bulk_update` call; disregard.

api/base/views.py
```diff
@@ -879,7 +882,44 @@ class WaterButlerMixin(object):
...
+            file_objs.append(file_obj)
+
+        for base_class in objs_to_create:
+            base_class.objects.bulk_create(objs_to_create[base_class])
+
+        bulk_update(file_objs)
```
sloria: You're updating the files that have just been created here; is that necessary?

api/base/views.py
sloria: Discussed on FD: this is fine.

PR #7851, opened 2017-10-31
[MOD-172][Hotfix] Allow contributors to see private preprints for unmoderated services

laurenbarker:
## Purpose
Allow contributors to see private preprints for unmoderated services.

## Changes
Use the same permissions for moderated and unmoderated providers.

## Side effects
Contributors can now see preprints that were submitted to an unmoderated service and then made private.

## Ticket
https://openscience.atlassian.net/browse/MOD-172

### DIFF
--- a/api_tests/preprints/views/test_preprint_detail.py
+++ b/api_tests/preprints/views/test_preprint_detail.py
@@ -922,7 +922,7 @@ class TestPreprintDetailPermissions:
         assert res.json['data']['id'] == private_preprint._id
 
     #   test_private_visible_to_write_contribs
-        res = app.get(private_url, auth=write_contrib.auth, expect_errors=True)
+        res = app.get(private_url, auth=write_contrib.auth)
         assert res.status_code == 200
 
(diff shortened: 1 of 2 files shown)


api_tests/preprints/views/test_preprint_detail.py
```diff
@@ -905,9 +921,9 @@ def test_preprint_is_public_detail(self, app, admin, write_contrib, non_contrib,
         res = app.get(private_url, auth=admin.auth)
         assert res.json['data']['id'] == private_preprint._id
 
-    #   test_private_invisible_to_write_contribs
+    #   test_private_visible_to_write_contribs
         res = app.get(private_url, auth=write_contrib.auth, expect_errors=True)
```
aaxelb: This has `expect_errors=True` but the status is 200. How does this pass?

laurenbarker: Oh whoops. I don't know how that passes.

PR #7852, opened 2017-11-01
[hotfix] embedding users on quickfile detail page [OSF-8874]

erinspace:
## Purpose

Allow embedding of the quickfile's node's owner's information, aka user in the serializer, in a quickfiles detail page

## Changes

- Use the `get_serializer` method when checking for fields that might have embeds instead of calling the property directly
- add tests

## Side effects

None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-8874

### DIFF
--- a/api/base/views.py
+++ b/api/base/views.py
@@ -145,7 +145,7 @@ class JSONAPIBaseView(generics.GenericAPIView):
         else:
             embeds = self.request.query_params.getlist('embed') or self.request.query_params.getlist('embed[]')
 
-        fields_check = self.serializer_class._declared_fields.copy()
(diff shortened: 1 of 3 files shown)


api/base/views.py
```diff
@@ -145,7 +145,7 @@ def get_serializer_context(self):
         else:
             embeds = self.request.query_params.getlist('embed') or self.request.query_params.getlist('embed[]')
 
-        fields_check = self.serializer_class._declared_fields.copy()
+        fields_check = self.get_serializer_class()._declared_fields.copy()
```
jamescdavis: getter > direct prop access :+1:

sloria:
Build failure unrelated. This is g2g

PR #7857, opened 2017-11-02
[OSF-8516] Fix link project recognized as child components

adlius:
## Purpose

Currently, project treats its linked nodes as child, therefore preventing correct deletion behavior. 

## Changes

In "node.py", a new ctx variable `child_exists` is added to use the `get_children()` method of the Node manager. This enables deletion of child components with linked nodes on project overview page.

In "views.py", the existing ctx variable `childExists` is modified to use the same method. This corrects the deletion behavior on the project settings page.

Tests are added.

## Side effects

None.

## Ticket

https://openscience.atlassian.net/browse/OSF-8516


adlius:
Tests are added. @caseyrollins Thanks for the advice.

PR #7858, opened 2017-11-03
[feature] Address Django Deprecation Warnings [OSF-8655]

caseyrollins:
#### Purpose
- Various Django features that we use will be removed or otherwise made backwards incompatible when v2.0 is released. Addressing these deprecations now will make the upgrade to Django 2.0  much smoother in the future, and will remove _most of_ the logged deprecation warnings until we ugrade. 

#### Upgrades
See the [Django Deprecation Timeline](https://docs.djangoproject.com/en/dev/internals/deprecation/#deprecation-removed-in-2-0) for more info.
- Add the `on_delete` argument to all `ForeignKey` and `OnetoOneField` fields.
    > on_delete will become a required argument in Django 2.0. In older versions it defaults to CASCADE.

- Pass a tuple containing `pattern_list` and `app_name` to the `.include()` URL utility.
    > Support for the app_name argument is deprecated and will be removed in Django 2.0. Specify the app_name as explained in URL namespaces and included URLconfs instead.
    >
    > Support for passing a 3-tuple is also deprecated and will be removed in Django 2.0. Pass a 2-tuple containing the pattern list and application namespace, and use the namespace argument instead.
    >
    > Lastly, support for an instance namespace without an application namespace has been deprecated and will be removed in Django 2.0. Specify the application namespace or remove the instance namespace.

- Use `_meta.private_fields` instead of `_meta.virtual_fields`
    > The private attribute virtual_fields of Model._meta is deprecated in favor of private_fields.

#### Upgrades NOT Included (noted and re-ticketed)
- Middleware upgrades. See https://docs.djangoproject.com/en/dev/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware
    > Support for old-style middleware using settings.MIDDLEWARE_CLASSES will be removed in Django 2.0.

- Upgrading dependencies that have deprecation warnings. 

#### Side Effects
- None expected.

#### Ticket
- [OSF-8655](https://openscience.atlassian.net/browse/OSF-8655)


api/base/urls.py
```diff
@@ -16,31 +16,31 @@
             [
                 url(r'^$', views.root, name='root'),
                 url(r'^status/', views.status_check, name='status_check'),
-                url(r'^actions/', include('api.actions.urls', namespace='actions')),
-                url(r'^addons/', include('api.addons.urls', namespace='addons')),
-                url(r'^applications/', include('api.applications.urls', namespace='applications')),
-                url(r'^citations/', include('api.citations.urls', namespace='citations')),
-                url(r'^collections/', include('api.collections.urls', namespace='collections')),
-                url(r'^comments/', include('api.comments.urls', namespace='comments')),
+                url(r'^actions/', include(('api.actions.urls', 'actions'), namespace='actions')),
```
sloria: I'm a bit confused by this change. 

According to the docs, `include(module, namespace=None, app_name=None)` is valid usage.

Passing a 3-tuple is deprecated, but we weren't doing that in the first place.

Passing a 2-tuple is valid, i.e. `include((pattern_list, app_namespace), namespace=None)`, but if I'm understanding correctly, the 2nd item of the tuple is supposed to be the app name. `actions`, `addons`, `applications`, etc. aren't proper Django apps (they're not in `INSTALLED_APPS`), so I'm not sure it's correct to pass them as app names.

caseyrollins: `include(module, namespace=None, app_name=None)` is a valid signature in Django 1.11, but won't be in 2.0 (see the 2.0 docs [here](https://docs.djangoproject.com/en/2.0/ref/urls/#include)). 

I do agree with your last point re: `actions`, `addons`, etc. being improper arguments for `app_namespace` and I'm looking into that now.

caseyrollins: After reading the Django 2.0 docs further, I think the recommended approach is to specify `app_name` (which should be `osf`) in the included modules and continue using the `include(module, namespace=None)` signature we have now. 

Note that the difference in Django 2.0 is that `app_name` is required (i.e. it needs to be set in the included module and can no longer be `None`).

sloria: Ah ok, thanks for the clarification.

sloria:
Merged into release/next-djangole

PR #7862, opened 2017-11-06
[feature/fix] Stop script from updating deleted registrations [OSF-8624]

alexschiller:
## Purpose
Deleted registrations break the fix registration unclaimed records script. When a registration is deleted the connection back to the originating node is severed, which breaks the script. This PR removes deleted reg from the queryset of registrations to update. Note this is different than a retraction, deletions occur when a registration is canceled.

## Changes
Don't update deleted registrations.

## Side effects
any registration that is un-deleted may need to be updated, but unlikely.

## Ticket

https://openscience.atlassian.net/browse/OSF-8624


alexschiller:
The script test that isn't run on travis passed just fine.

PR #7867, opened 2017-11-06
[#OSF-8849]Make institutional login more prominent

chennan47:
## Purpose
Change the style of login through institution and fix the title for sign up page.

## Changes
before change:
screen shot 2017-11-06 at 3 25 29 pm
After change:
screen shot 2017-11-06 at 3 25 01 pm

## Ticket
https://openscience.atlassian.net/browse/OSF-8849
https://openscience.atlassian.net/browse/OSF-8226

### DIFF
--- a/website/templates/public/register.mako
+++ b/website/templates/public/register.mako
@@ -147,7 +147,7 @@
                         %endif
                             <div class="col-xs-12">
                                 <span class="pull-right p-t-sm">
-                                    <a class="btn btn-info" href="${institution_login_url}">Sign in with your institution</a>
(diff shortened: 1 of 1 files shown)


laurenbarker:
We want to avoid having two buttons that are "primary".

A good explanation by Nathan Curtis in his [blog post about buttons](https://medium.com/eightshapes-llc/buttons-in-design-systems-eac3acf7e23).
> But you may find yourself needing to pair a secondary option with that inviting, saturated primary. You avoid a second saturated color, because that results in two saturated buttons next to one another, like green for Save and blue for Submit. Not even you, let alone your users, know which is more important.

PR #7870, opened 2017-11-06
[feature] Upgrade dependencies [OSF-6108]

caseyrollins:
#### Purpose
- Keeps us up to date and gets us one step closer to Python 3 (!) 

#### Ticket
- [OSF-6108](https://openscience.atlassian.net/browse/OSF-6108)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,6 @@ Flask==0.10.1
 gevent==1.2.2
 Mako==1.0.7
 Markdown==2.6.9
-Pygments==1.6
 WTForms==1.0.4
 beautifulsoup4==4.6.0
 celery==3.1.25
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -91,7 +90,7 @@ psycopg2==2.6.2
 ujson==1.35
 sqlparse==0.2.4
 psycogreen==1.0
-django-bulk-update==1.1.10
+django-bulk-update==2.2.0
```
sloria: The package was renamed from `bulk_update` to `django_bulk_update`: https://github.com/aykut/django-bulk-update/blob/master/CHANGELOG#L13 . Doesn't this break?

sloria:
Merged into `release/next-djangole`

PR #7871, opened 2017-11-06
[feature] vol url param [OSF-8820]

fabmiz:
## Purpose

Add view_only param to node files list api endpoint

## Changes

Modify api.base.serializers to check for vol in params 
and add vol param to wb action links. 

## Ticket

[OSF-8820](https://openscience.atlassian.net/browse/OSF-8820)

## After

screen shot 2017-11-10 at 11 15 15 am

screen shot 2017-11-10 at 11 14 38 am

### DIFF
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -980,7 +980,8 @@ class WaterbutlerLink(Link):
 
         if 'view_only' not in self.kwargs:
             view_only = request.query_params.get('view_only', False)
-            self.kwargs['view_only'] = view_only
+            if view_only:
+                self.kwargs['view_only'] = view_only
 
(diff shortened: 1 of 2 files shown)


api/base/serializers.py
```diff
@@ -977,6 +977,11 @@ def resolve_url(self, obj, request):
             raise SkipField
         if self.must_be_file is True and obj.path.endswith('/'):
             raise SkipField
+
+        if 'view_only' not in self.kwargs:
+            view_only = request.query_params.get('view_only', False)
+            self.kwargs['view_only'] = view_only
```
alexschiller: This looks like it is going to add `view_only=False` to all wb links that don't have vols, is that intentional?

fabmiz: Oh no, it's not intentional. I will update the PR to fix that.

PR #7873, opened 2017-11-07
[OSF-8477] Import contributors in one click - fix pagination

Johnetordoff:
## Purpose

There are a couple problems with the current import contributors feature.

- It doesn't import non-bibliographic contributors
- It doesn't import the contributors permissions
- You can't modify a contributors the bibliographic status before you add them, only their permissions.

This fix will set the permissions and bibliographic status on import so you can import everyone in one click.

## Changes

Changed the get_contributors_from_parent/ endpoint to get more info about the contributors permissions and bibliographic status, then modified the .mako and .js to display the new info without crowding the screen

## Before
before_import_contributors

## After
import_contribs_demo

## Side effects

Modified some of the wrapping in that dialog, so make sure that works in a sane way with long names.

## Testing 

Nothing special

## Ticket

https://openscience.atlassian.net/browse/OSF-8477

### DIFF
--- a/website/static/js/contribAdder.js
+++ b/website/static/js/contribAdder.js
@@ -111,10 +111,6 @@ AddContributorViewModel = oop.extend(Paginator, {
             return self.query() && self.results().length && !self.parentImport();
         });
 
-        self.parentPagination = ko.pureComputed(function () {
-            return self.doneSearching() && self.parentImport();
-        });
-
         self.noResults = ko.pureComputed(function () {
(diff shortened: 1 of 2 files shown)


brianjgeiger:
Not going to hold up this ticket, but in the future the PR description should be for the specific change, not for the original problem. That way I could see, for example, the paginated way in the old ticket and the unpaginated way in the new ticket.

PR #7876, opened 2017-11-07
[feature] Allow opening same file in multiple tabs [OSF-8780]

fabmiz:
## Purpose

Allow (back) open same file in multiple tabs.

## Changes

Removes front-end hotfix that prevents users from opening
same file in multiple tabs as the main issue was fixed in the backend.

## Ticket

[OSF-8780](https://openscience.atlassian.net/browse/OSF-8780)

### DIFF
--- a/website/static/js/fangorn.js
+++ b/website/static/js/fangorn.js
@@ -1348,7 +1348,7 @@ function gotoFileEvent (item, toUrl) {
     }
 
     if (COMMAND_KEYS.indexOf(tb.pressedKey) !== -1) {
-        window.open(fileurl, '_blank');
+        window.open(fileurl, item.data.materialized);
     } else {
         window.open(fileurl, '_self');
     }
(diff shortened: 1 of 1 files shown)


website/static/js/fangorn.js
```diff
@@ -1348,7 +1348,7 @@ function gotoFileEvent (item, toUrl) {
     }
 
     if (COMMAND_KEYS.indexOf(tb.pressedKey) !== -1) {
-        window.open(fileurl, '_blank');
+        window.open(fileurl, item.data.materialized);
```
alexschiller: If product agrees with this behavior, you'll want to check the fangornconfigs for bitbucket, figshare, github, dataverse since some of them overwrite this behavior. I know for sure github does.

fabmiz: Yes, when it's accepted I will take the time to test similar fix for other addons.

alexschiller: Cool, I moved it to demo then 👍

PR #7883, opened 2017-11-09
[feature] Add highlighted subjects bool flag [OSF-8881]

fabmiz:
## Purpose

Add a flag to indicate a preprint service is using a custom taxonomy

## Changes

- Add `has_custom_subjects` property to `preprint_provider` model
- Add flag check to existing tests

## Ticket

[OSF-8881](https://openscience.atlassian.net/browse/OSF-8881)

## After
screen shot 2017-11-13 at 5 09 42 pm


osf/models/preprint_provider.py
```diff
@@ -70,6 +70,10 @@ def __unicode__(self):
         return '{} with id {}'.format(self.name, self.id)
 
     @property
+    def has_custom_subjects(self):
+        return self.subjects.exists()
```
brianjgeiger: Based on the linked EOSF ticket, this looks like it should be self.subjects.filter(highlighted=True).exists() and change https://github.com/CenterForOpenScience/osf.io/pull/7883/files#diff-03133408b78d570d4fe79642d38b06c8L74 to use this new property.

api/preprint_providers/serializers.py
```diff
@@ -108,6 +109,9 @@ class PreprintProviderSerializer(JSONAPISerializer):
     class Meta:
         type_ = 'preprint_providers'
 
+    def get_highlighted_subjects_exists(self, obj):
```
brianjgeiger: Minor, and should have thought of this yesterday: change this to `get_has_highlighted_subjects()` so that it's consistently named.

PR #7886, opened 2017-11-10
[Feature][MOD-180] Show the preprint banner regardless of preprint status

sheriefvt:
## Purpose

Show the preprint banner on projects to contributors regardless of preprint status.
Display the preprint status on the banner. 

This is because contributors are used to getting to their preprint from the projects page but it doesn't show up if the preprint is pending or rejected.

## Changes

- Add new properties to node model
- Update node view
- Update project.mako template

1. If preprint provider supports moderation
image

2. If preprint provider doesn't support moderation
image

## Ticket

https://openscience.atlassian.net/browse/MOD-180


osf/models/node.py
```diff
@@ -441,12 +441,25 @@ def is_preprint(self):
         if not self.preprint_file_id or not self.is_public:
             return False
         if self.preprint_file.node_id == self.id:
-            return self.has_published_preprint
+            return self.has_published_preprint or self.has_pending_rejected_preprint
```
aaxelb: It might make more sense to have `has_submitted_preprint` that checks for preprints with non-`initial` state. Since submitted preprints are a superset of published preprints, you could use that instead of both of these.

website/templates/project/project.mako
```diff
@@ -276,7 +276,11 @@
 <div class="row">
     <div class="col-xs-12">
         <div class="pp-notice m-b-md p-md clearfix">
-            This project represents a preprint. <a href="http://help.osf.io/m/preprints">Learn more</a> about how to work with preprint files.
+            This project represents a preprint.
+            % if node['has_preprint_moderated']:
+                The preprint is currently in ${node['preprint_state']} state.
```
aaxelb: This might be more confusing than helpful. If we're displaying the state, we should maybe include a description of what it means, like we do on the preprint status banner.

osf/models/node.py
```diff
@@ -441,12 +441,25 @@ def is_preprint(self):
...
             self._is_preprint_orphan = True
             return False
 
     @property
+    def has_pending_rejected_preprint(self):
+        return self.preprints.filter(Q(reviews_state='pending') | Q(reviews_state='rejected')).exists()
+
+    @property
+    def has_preprint_moderated(self):
+        return self.preprints.filter(provider__reviews_workflow__isnull=False).exists()
```
aaxelb: Do we want to show the banner for unsubmitted (abandoned) preprints, or should this filter out `initial` state?

osf/models/node.py
```diff
@@ -441,12 +441,25 @@ def is_preprint(self):
...
         else:
             self._is_preprint_orphan = True
             return False
 
     @property
+    def has_pending_rejected_preprint(self):
+        return self.preprints.filter(Q(reviews_state='pending') | Q(reviews_state='rejected')).exists()
+
+    @property
+    def has_preprint_moderated(self):
```
aaxelb: I think `has_moderated_preprint` would be clearer and more consistent with other names

website/templates/project/project.mako
sheriefvt: do you think a popover message might be a good idea or just add to message text

laurenbarker:
## Wording from product:
This project represents a [STATE] [PREPRINT-WORD] submitted to [PROVIDER] [question mark icon thing].

Tooltip on hover over the question mark icon (used elsewhere on the OSF) with the wording shown on moderator detail page. 
screen shot 2017-11-10 at 11 30 46 am

laurenbarker:
@sheriefvt, can you upgrade this to a hotfix?

aaxelb:
@laurenbarker should the banner display for `initial` preprints?

sheriefvt:
@laurenbarker @aaxelb I will create a new branch based on master with another PR and close this one.

laurenbarker:
@aaxelb @sheriefvt the banner should not display for initial preprints.

PR #7887, opened 2017-11-10
Feature/remind abandoned prereg [OSF-8546]

laurenrevere:
## Purpose

People sometimes start and do not finish Preregistration Challenge drafts. This will send them emails to remind them to finish said drafts.

## Changes

- Add a cron job that runs a script to check daily for neglected preregistration drafts and adds reminder emails to the email queue.

**Note:** The specifications want there to be a possible 3 emails in 2 weeks, but the email queue we currently use sends 1 email a week. This feature uses the existing structure.

## Side effects

- I deleted `tests/test_mails.py` because had the same tests as`osf_tests/test_queued_mail.py`

## Ticket
https://openscience.atlassian.net/browse/OSF-8546

screen shot 2017-11-16 at 2 45 48 pm


website/templates/emails/prereg_reminder.html.mako
```diff
@@ -0,0 +1,16 @@
...
+        <br>
+        Dear ${fullname},
+        <br><br>
+        You have an unsubmitted preregistration on the Open Science Framework that could be eligible for a $1,000 prize as part of the <a href="https://cos.io/prereg">Prereg Challenge</a>. If you would like this to be considered for a prize, please complete <a href="${prereg_url}">your preregistration</a> and submit it for review, available on the last page of the preregistration form. This review process is required for a prize and simply checks to make sure the preregistration includes a complete analysis plan.
+        If you have questions about a particular field in the preregistration, you may review the <a href="https://cos.io/prereg">FAQ on the website</a>, <a href="mailto:prereg@cos.io">email us with a question</a>, or use our <a href="https://cos.io/stats_consulting">free statistical consulting services</a>.
+        Thank you for using the OSF!
+        <br><br>
+        Sincerely,
+        <br>
+        The team at the Center for Open Science
```
brianjgeiger: Does this have the address somewhere in it? It might be in the full email, but if so, the footer was cut off on the screen shot.

laurenrevere: Ya, sorry. The address is in fact there.

scripts/remind_draft_preregistrations.py
```diff
@@ -0,0 +1,59 @@
...
+                    fullname=prereg.initiator.fullname,
+                    prereg_url=prereg.absolute_url,
+                    draft_id=prereg._id,
+                )
+
+def find_neglected_prereg_within_reminder_limit():
+
+    queue_data = QueuedMail.objects.filter(
+        email_type=PREREG_REMINDER_TYPE,
+    ).values('data')
```
alexschiller: Can this queue_data be limited based on time as well? Otherwise the size of this query (which is then modified in python) may get undesirably large.
Potential code below (untested!!!!!!!):

```
queue_data = QueuedMail.objects.filter(
    email_type=PREREG_REMINDER_TYPE,
    data__draft_id__isnull=False, # should remove entries with no draft_id
    send_at__gte=timezone.now() - settings.PREREG_AGE_LIMIT # I think this works, because it should only returned queued mails that could potentially be queued twice, but don't trust me.
).values_list('data', flat=True)

already_queued = [entry['draft_id'] for entry in queue_data]
```

I tried in vain to get the already_queued directly through the values_list but couldn't find a way to directly extract the data in the jsonfield :(

website/mails/presends.py
```diff
@@ -32,6 +32,23 @@ def new_public_project(email):
...
+    :param email: QueuedMail object, with the 'draft_id' in its data field
+    :return: boolean based on whether the email should be sent
+    """
+    # In line import to prevent circular importing
+    from osf.models.queued_mail import QueuedMail
+    from osf.models import DraftRegistration
+
+    draft_id = email.data['draft_id']
+    draft = DraftRegistration.load(draft_id)
+    reminders_sent = QueuedMail.objects.filter(data__draft_id=draft_id).exclude(sent_at=None)
```
alexschiller: .exists() is generally more efficient when you don't need to do anything with the queryset
`QueuedMail.objects.filter(data__draft_id=draft_id).exclude(sent_at=None).exists()`

From django queryset docs on exists:
[In many cases it will perform better than a subquery since the database is able to stop evaluation of the subquery when a first matching row is found](https://docs.djangoproject.com/en/1.11/ref/models/expressions/#exists-subqueries)

and then...

website/mails/presends.py
```diff
@@ -32,6 +32,23 @@ def new_public_project(email):
...
+
+    return draft and not len(reminders_sent) and not draft.approval
```
alexschiller: ... you can remove `len` here

osf_tests/test_queued_mail.py
```diff
@@ -149,3 +152,41 @@ def test_user_is_not_active_is_not_confirmed(self, mock_mail):
             mail=NO_ADDON,
         )
         assert mail.send_mail() is False
+
+    @mock.patch('osf.models.queued_mail.send_mail')
+    def test_remind_prereg_presend(self, mock_mail, user):
+        prereg = DraftRegistrationFactory(registration_schema=MetaSchema.objects.get(name='Prereg Challenge'))
```
alexschiller: `MetaSchema.objects.get(name='Prereg Challenge')` and/or your prereg draft should probably be a fixture

osf_tests/test_queued_mail.py
```diff
@@ -149,3 +152,41 @@ def test_user_is_not_active_is_not_confirmed(self, mock_mail):
...
+        mail = self.queue_mail(mail=PREREG_REMINDER, user=user, draft_id=prereg._id)
+        prereg.delete()
+        assert not mail.send_mail()
+
+    @mock.patch('osf.models.queued_mail.send_mail')
+    def test_remind_prereg_presend_already_sent(self, mock_mail, user):
+        prereg = DraftRegistrationFactory(registration_schema=MetaSchema.objects.get(name='Prereg Challenge'))
+        mail = self.queue_mail(mail=PREREG_REMINDER, user=user, draft_id=prereg._id)
+        mail.send_mail()
```
alexschiller: If I'm reading this correctly, it looks like this test repeats all the steps of test 1-- this second half could probably be tacked onto the first test with a note saying `# test remind prereg presend already sent` to save some time.

osf_tests/test_remind_draft_preregistrations.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+        draft.datetime_initiated = timezone.now() - settings.PREREG_WAIT_TIME
+        draft.save()
+        return draft
+
+
+    def test_trigger_prereg_reminder(self, draft):
+        main(dry_run=False)
+
+        assert len(QueuedMail.objects.filter(email_type=PREREG_REMINDER_TYPE)) == 1
```
alexschiller: Prefer to use count on querysets `QueuedMail.objects.filter(email_type=PREREG_REMINDER_TYPE).count() == 1`

> len(). A QuerySet is evaluated when you call len() on it. This, as you might expect, returns the length of the result list.
> 
> Note: If you only need to determine the number of records in the set (and don’t need the actual objects), it’s much more efficient to handle a count at the database level using SQL’s SELECT COUNT(*). Django provides a count() method for precisely this reason.

scripts/remind_draft_preregistrations.py
```diff
@@ -0,0 +1,59 @@
...
+
+    already_queued = [entry['data'].get('draft_id') for entry in queue_data if entry['data'].get('draft_id')]
+
+    return DraftRegistration.objects.filter(
```
alexschiller: Oh and Casey has a PR to add the ability mark draft registrations as deleted #7896, which might mean you'll need to filter those out if her PR goes in first.

scripts/remind_draft_preregistrations.py
laurenrevere: @alexschiller, I originally limited it on time, and then changed it. I guess I thought it was weird that more things would queue that would never be sent. If that's not a problem, then I agree with you. Do you have a suggestion for a time limit. I had it just as two weeks at first (because that's the it's the PREREG_WAIT_TIME)... but that's kind of arbitrary.

laurenrevere: Update: you were right all along I just didn't pay enough attention reading your original comment.

brianjgeiger:
Conflicts 😞

PR #7888, opened 2017-11-10
[Fix][OSF-8925] Show the preprint banner regardless of preprint status

sheriefvt:
## Purpose

Show the preprint banner on projects to contributors regardless of preprint status.
Display the preprint status on the banner. 

This is because contributors are used to getting to their preprint from the projects page but it doesn't show up if the preprint is pending or rejected.

## Changes

- Add new properties to node model
- Update node view
- Update project.mako template

1. If preprint provider supports moderation

image

image

2. If preprint provider doesn't support moderation

image

## Note
This PR against master branch replaces closed PR #7886 which were against develop branch.

## Ticket

https://openscience.atlassian.net/browse/MOD-180


laurenbarker:
Sorry, I should have included more information. The help icon should be the same as the one on the "My Projects" page -- including the color (see below). 
screen shot 2017-11-10 at 2 56 50 pm

The second sentence -- "Learn more about how to work with preprint files." -- should be kept in both situations. Just the first sentence should be dependent on moderation.

Other then those small changes this is looking good! The users will be much happier once it's in.

sheriefvt:
@laurenbarker updates 👍  🐼

osf/models/node.py
```diff
@@ -441,12 +441,39 @@ def is_preprint(self):
...
             return False
         if self.preprint_file.node_id == self.id:
-            return self.has_published_preprint
+            return self.has_preprint_non_initial_state
         else:
             self._is_preprint_orphan = True
             return False
 
     @property
+    def has_preprint_non_initial_state(self):
```
aaxelb: I think it'd be clearer and more consistent to call this `has_submitted_preprint`, or some other `has_<adjective>_preprint`

osf/models/node.py
```diff
@@ -441,12 +441,39 @@ def is_preprint(self):
...
+        return self.preprints.filter(~Q(reviews_state='initial')).exists()
```
aaxelb: Using `exclude` instead of `~Q` would be more understandable

osf/models/node.py
```diff
@@ -441,12 +441,39 @@ def is_preprint(self):
...
+
+    @property
+    def has_moderated_preprint(self):
```
aaxelb: These properties seem like they should be computed in the view, they're not useful on the model. Maybe a method here to get the preprint (prefer a published one, if it exists), then fill out the view data based on that?

osf/models/node.py
```diff
@@ -441,12 +441,39 @@ def is_preprint(self):
...
+        return self.preprints.filter(provider__reviews_workflow__isnull=False).exists()
+
+    @property
+    def preprint_state(self):
```
aaxelb: Compute in the view

osf/models/node.py
```diff
@@ -441,12 +441,39 @@ def is_preprint(self):
...
+        if self.has_moderated_preprint:
+            return self.preprints.get_queryset()[0].reviews_state
+
+    @property
+    def preprint_word(self):
```
aaxelb: Compute in the view

osf/models/node.py
```diff
@@ -441,12 +441,39 @@ def is_preprint(self):
...
+        if self.is_preprint:
+            return self.preprints.get_queryset()[0].provider.preprint_word
+
+    @property
+    def preprint_provider(self):
```
aaxelb: Compute in the view

aaxelb:
One more thing

aaxelb:
One more thing (sorry, closed it by misclicking): Instead of the built-in browser hover text, the tooltip should be whatever is used elsewhere on the OSF, so it pops up right away and is more clearly visible, like on the "My Projects" page:
screen shot 2017-11-14 at 2 39 44 pm

sheriefvt:
@aaxelb OSF is using default bootstrap style. I just added data-toggle="tooltip" data-placement="bottom" to look the same as OSF

osf/models/node.py
```diff
@@ -441,12 +441,16 @@ def is_preprint(self):
...
         if self.preprint_file.node_id == self.id:
-            return self.has_published_preprint
+            return self.has_submitted_preprint
         else:
             self._is_preprint_orphan = True
             return False
 
     @property
+    def has_submitted_preprint(self):
+        return self.preprints.exclude(reviews_state='initial').exists()
```
aaxelb: Use `States.INITIAL.value` instead of a naked string.

website/project/views/node.py
```diff
@@ -735,6 +735,14 @@ def _view_project(node, auth, primary=False,
             'institutions': get_affiliated_institutions(node) if node else [],
             'has_draft_registrations': node.has_active_draft_registrations,
             'is_preprint': node.is_preprint,
+            'has_moderated_preprint': node.preprints.filter(provider__reviews_workflow__isnull=False).exists(),
+            'preprint_state': node.preprints.get_queryset()[0].reviews_state
+            if node.preprints.filter(provider__reviews_workflow__isnull=False).exists() else '',
+            'preprint_word': node.preprints.get_queryset()[0].provider.preprint_word if node.is_preprint else '',
+            'preprint_provider': {
+                'name': node.preprints.get_queryset()[0].provider.name,
+                'workflow': node.preprints.get_queryset()[0].provider.reviews_workflow
```
aaxelb: Instead of repeating this query over and over, it might be better to add a method/property to the node to get the node's preprint. You could call it once and use the result several times.

It might make sense to move the logic you added to `preprint_url` there -- first try to get a published preprint, then fall back to any preprint -- and use the new method in `preprint_url`

website/project/views/node.py
```diff
@@ -735,6 +735,13 @@ def _view_project(node, auth, primary=False,
             'institutions': get_affiliated_institutions(node) if node else [],
             'has_draft_registrations': node.has_active_draft_registrations,
             'is_preprint': node.is_preprint,
+            'has_moderated_preprint': node.has_moderated_preprint,
+            'preprint_state': node.linked_preprint.reviews_state if node.has_moderated_preprint else '',
+            'preprint_word': node.linked_preprint.provider.preprint_word if node.is_preprint else '',
+            'preprint_provider': {
+                'name': node.linked_preprint.provider.name,
+                'workflow': node.linked_preprint.provider.reviews_workflow
```
aaxelb: Since properties aren't cached, this is basically the same as before (doing `get_queryset()[0]` over and over). It'd be nicer to get the preprint higher up in this function, then use the variable.

osf/models/node.py
```diff
@@ -459,12 +464,23 @@ def is_preprint_orphan(self):
     def has_published_preprint(self):
         return self.preprints.filter(is_published=True).exists()
 
+    @property
+    def linked_preprint(self):
+        return self.preprints.get_queryset()[0]
+
+    @property
+    def has_moderated_preprint(self):
+        return self.preprints.filter(provider__reviews_workflow__isnull=False).exists()
```
aaxelb: Instead of `has_moderated_preprint`, couldn't the view just check `node.linked_preprint.provider.reviews_workflow`? This seems like an unnecessary query.

osf/models/node.py
```diff
@@ -459,12 +464,23 @@ def is_preprint_orphan(self):
     def has_published_preprint(self):
         return self.preprints.filter(is_published=True).exists()
 
+    @property
+    def linked_preprint(self):
+        return self.preprints.get_queryset()[0]
```
aaxelb: It might make sense to first try for a published preprint, then fall back to any preprint. Basically the logic in `preprint_url` (which could then just `return self.linked_preprint.url`)

sheriefvt: i used linked_preprint as a way to get any preprint and used it in preprint_url

osf/models/node.py
```diff
@@ -461,10 +466,23 @@ def has_published_preprint(self):
 
     @property
     def preprint_url(self):
+        if self.linked_preprint:
+            return self.linked_preprint.url
```
aaxelb: `linked_preprint` makes several queries, I'd rather not call it twice in a row

website/project/views/node.py
```diff
@@ -735,6 +736,13 @@ def _view_project(node, auth, primary=False,
             'institutions': get_affiliated_institutions(node) if node else [],
             'has_draft_registrations': node.has_active_draft_registrations,
             'is_preprint': node.is_preprint,
+            'has_moderated_preprint': node.has_moderated_preprint,
```
aaxelb: Instead of `node.has_moderated_preprint`, could this check `node_linked_preprint.provider.reviews_workflow`?

website/project/views/node.py
```diff
@@ -735,6 +736,13 @@ def _view_project(node, auth, primary=False,
...
+            'preprint_state': node_linked_preprint.reviews_state if node.has_moderated_preprint else '',
+            'preprint_word': node_linked_preprint.provider.preprint_word if node.is_preprint else '',
+            'preprint_provider': {
+                'name': node_linked_preprint.provider.name,
+                'workflow': node_linked_preprint.provider.reviews_workflow
+            } if node.is_preprint else {},
```
aaxelb: Could these three check whether `node_linked_preprint` is null, instead of `node.has_moderated_preprint` and `node.is_preprint`?

osf/models/node.py
```diff
@@ -461,10 +466,23 @@ def has_published_preprint(self):
...
+            return self.linked_preprint.provider.reviews_workflow
+
+    @property
+    def linked_preprint(self):
         if self.is_preprint:
             try:
                 # if multiple preprints per project are supported on the front end this needs to change.
-                return self.preprints.filter(is_published=True)[0].url
+                if self.has_published_preprint:
+                    return self.preprints.filter(is_published=True)[0]
```
aaxelb: `has_published_preprint` already does this exact query (with `exists()`), it seems wasteful to do it again.

PR #7893, opened 2017-11-13
[proof of concept] Get ready for uploads-to-anything with some migrations  [OSF-8884]

erinspace:
## Purpose
This is a necessary first step to first store the target_object_id and target_content_type for the
still-to-come `target` GenericRelation() that will eventually replace the node ForeignKey. 

## Changes
- Add `target_content_type` and `target_object_id` to `BaseFileNode` model
- Migration to add `target_content_type` and `target_object_id` fields

## Side effects
Needs 2 migrations, the second of which will take awhile -- for my staging dump took about 2 minutes. Staging has around 450K `BaseFileNode`s, while as prod has over 3 million, so by that extension should take about 15 mins of downtime on prod.

## Ticket
part 1 of
https://openscience.atlassian.net/browse/OSF-8884


erinspace:
I don't actually need this to be a separate PR! Can all go in at once.

PR #7894, opened 2017-11-13
[Hotfix] Stop quickfiles from moving with merges [OSF-8911]

alexschiller:
## Purpose
Merging users moves the quickfiles node. The files are moved properly away from the old quickfiles node, but the node itself also is moved. 

## Changes
Stop moving quickfiles nodes when transferring node contributor-ship. Add script to run to fix merged users.

## Side effects
Stop potential bad things/weird things in the future.

## Deployment Instructions
Run the `fix_merged_user_quickfiles` script in --dry mode first to confirm everything works properly.

## Ticket

https://openscience.atlassian.net/browse/OSF-8911

## QA Notes
Just merge a user and make sure everything works properly. I'll confirm that things are working properly on the backend.

### DIFF
--- a/scripts/fix_merged_user_quickfiles.py
+++ b/scripts/fix_merged_user_quickfiles.py
@@ -6,7 +6,7 @@ from django.db.models import F
 
 from website.app import setup_django
 setup_django()
-from osf.models import AbstractNode
+from osf.models import QuickFilesNode
 from scripts import utils as script_utils
 
 
@@ -18,7 +18,7 @@ def main():
         # If we're not running in dry mode log everything to a file
         script_utils.add_file_logger(logger, __file__)
     with transaction.atomic():
-        qs = AbstractNode.objects.filter(type='osf.quickfilesnode').exclude(_contributors=F('creator'))
+        qs = QuickFilesNode.objects.exclude(_contributors=F('creator'))
         logger.info('Found {} quickfiles nodes with mismatched creator and _contributors'.format(qs.count()))
 
         for node in qs:
--- a/scripts/fix_merged_user_quickfiles.py
+++ b/scripts/fix_merged_user_quickfiles.py
@@ -1,0 +1,35 @@
+import logging
+import sys
+
+from django.db import transaction
+from django.db.models import F
+
+from website.app import setup_django
+setup_django()
+from osf.models import AbstractNode
+from scripts import utils as script_utils
+
+
+logger = logging.getLogger(__name__)
+
+def main():
+    dry = '--dry' in sys.argv
+    if not dry:
+        # If we're not running in dry mode log everything to a file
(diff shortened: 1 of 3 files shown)


brianjgeiger:
There's a little more going on than shows in the diff. I'm guessing that `merge_dupe()` gets called somewhere along the save chain so that, once the users are fixed, everything gets merged together?

scripts/fix_merged_user_quickfiles.py
```diff
@@ -0,0 +1,35 @@
...
+
+logger = logging.getLogger(__name__)
+
+def main():
+    dry = '--dry' in sys.argv
+    if not dry:
+        # If we're not running in dry mode log everything to a file
+        script_utils.add_file_logger(logger, __file__)
+    with transaction.atomic():
+        qs = AbstractNode.objects.filter(type='osf.quickfilesnode').exclude(_contributors=F('creator'))
```
sloria: Drive-by: Instead of filtering on type, you should be able to use the concrete class.

```python
QuickFilesNode.exclude(...)
```

alexschiller: @sloria Neat, fixed 👍

sloria:
When running the script on a prod dump locally, I got an error:

```
  File "/usr/local/Cellar/python/2.7.14/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 174, in _run_module_as_main
  File "/usr/local/Cellar/python/2.7.14/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/Users/sloria/projects/cos-projects/osf/scripts/fix_merged_user_quickfiles.py", line 35, in <module>
    main()
  File "/Users/sloria/projects/cos-projects/osf/scripts/fix_merged_user_quickfiles.py", line 26, in main
    bad_contrib = node._contributors.get()
  File "/Users/sloria/.local/share/virtualenvs/osf/lib/python2.7/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/Users/sloria/.local/share/virtualenvs/osf/lib/python2.7/site-packages/django/db/models/query.py", line 380, in get
    self.model._meta.object_name
osf.models.user.DoesNotExist: OSFUser matching query does not exist.
```

`node` is a `QuickFilesNode` that has no contributors. The creator was merged into another user who has only one `QuickFilesNode`. So I think the fix is to just skip over any QuickFilesNodes with no contributors.

```python
qs = QuickFilesNode.objects.exclude(_contributors=F('creator')).annotate(contrib_count=Count('_contributors')).exclude(contrib_count=0)
```

sloria:
I can make the above fix when I merge.

PR #7896, opened 2017-11-13
[feature] Mark DraftRegistrations as `deleted` [OSF-8907]

caseyrollins:
#### Purpose
- Currently, draft registrations are removed from the database when they are deleted. This goes against our general policy of marking things as deleted instead of actually deleting data. Additionally, it prevents us from restoring a draft registration is a user accidentally deletes it. 

#### Changes
- Add a `deleted` datetime field to the `DraftRegistration` model. 
- Set `deleted = timezone.now()` when "deleting" draft registrations. 
- Add/update tests (v1 and v2) to ensure that deleted draft registrations `410` if they're accessed, and don't appear in draft registration list views.

#### Side Effects
- None expected.

#### QA
- Make sure deleted draft registrations aren't accessible/listed on the OSF or in the API

#### Ticket
- [OSF-8907](https://openscience.atlassian.net/browse/OSF-8907)


osf/models/registrations.py
```diff
@@ -386,9 +386,10 @@ class DraftRegistration(ObjectIDMixin, BaseModel):
 
     datetime_initiated = NonNaiveDateTimeField(auto_now_add=True)
     datetime_updated = NonNaiveDateTimeField(auto_now=True)
+    deleted = NonNaiveDateTimeField(null=True, blank=True)
```
alexschiller: Random thought: was there any discussion of making this `is_deleted` and adding a `deleted_date` to match node? Not that there is much consistency as files have a `deleted_on` and a `deleted_by_id`... 😐

caseyrollins: @alexschiller, consistency coming soon! See [OSF-8916](https://openscience.atlassian.net/browse/OSF-8916) (and Steve's comment on [OSF-8907](https://openscience.atlassian.net/browse/OSF-8907))

PR #7912, opened 2017-11-16
[#OSF-8904]When top level project is forked, logs from grandchildren do not show on top-level fork

chennan47:
## Purpose
Fix the problem forked_component doesn't have the correct noderelationship.

## Changes
before change:
The grandchild log won't show up on top level forked project.
after change: 
The log will show up on top level forked project.

## Side effects
This pr needs to run a migration after merged in.

## Deployment notes
You will need to run fix_roots after this release

## Ticket
https://openscience.atlassian.net/browse/OSF-8904


brianjgeiger:
Minor thing for future PRs. The effects of your changes should be in the Purpose. "Logs of grandchild nodes are not showing up in root nodes, this fixes that so they will" is a good purpose. Changes would be more like, "Refactored the fork_node so that it would expect <blah> and removed unnecessary permissions change. Also added a manual migration so root would <blah>." This will give the code reviewer some context about what you were intending to change and why, and then it'll be easier to evaluate the code that you did change for potential side effects or whatever.

osf/models/node.py
```diff
@@ -1695,7 +1695,7 @@ def register_node(self, schema, auth, data, parent=None):
             node_contained = node_relation.child
             # Register child nodes
             if not node_relation.is_node_link:
-                registered_child = node_contained.register_node(  # noqa
+                node_contained.register_node(  # noqa
```
brianjgeiger: Why did this get changed in this PR? Also, is the noqa comment in there because flaking was failing due to the unused variable? If so, can it be removed (if we keep this change)?

osf/models/node.py
```diff
@@ -1860,28 +1860,24 @@ def fork_node(self, auth, title=None, parent=None):
         # Need to save here in order to access m2m fields
         forked.save()
 
+        forked.tags.add(*self.all_tags.values_list('pk', flat=True))
```
brianjgeiger: Does this move have a functional effect?

scripts/fix_fork_component_root.py
```diff
@@ -0,0 +1,35 @@
...
+
+from website.app import setup_django
+setup_django()
+from osf.models import AbstractNode
+from scripts import utils as script_utils
+
+
+logger = logging.getLogger(__name__)
+
+def main():
```
brianjgeiger: This script looks like it could just be made into an automatic_migration rather than requiring someone to run it manually. It should only have to be done once when this change is in. I don't expect it would need a reverse because there's no good reason to have the wrong root. @sloria do you agree, or do you think this would be too much overhead for day-to-day developers?

osf/models/node.py
chennan47: Yes, this is an unused variable which will cause flake 8 error. Not sure why it doesn't raise that error before. Since that variable isn't used anywhere, better to remove it.

osf/models/node.py
chennan47: Nope, it is just being moved by solving merge conflict. Doesn't really change anything.

osf/models/node.py
brianjgeiger: I think you should be able to remove the `#noqa` then

scripts/fix_fork_component_root.py
brianjgeiger: An example of us doing this was in https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/migrations/0055_update_metaschema_active.py

osf/models/node.py
chennan47: fixed in 991d290

tests/test_views.py
```diff
@@ -821,6 +821,18 @@ def test_view_project_returns_whether_to_show_wiki_widget(self):
         assert_equal(res.status_code, http.OK)
         assert_in('show_wiki_widget', res.json['user'])
 
+    def test_fork_grandcomponents_has_correct_root(self):
+        user = AuthUserFactory()
+        project = ProjectFactory(creator=user)
+        child = NodeFactory(parent=project, creator=self.user)
```
brianjgeiger:
```
_________ TestProjectViews.test_fork_grandcomponents_has_correct_root __________
[gw1] linux2 -- Python 2.7.14 /home/travis/virtualenv/python2.7.14/bin/python
tests/test_views.py:827: in test_fork_grandcomponents_has_correct_root
    child = NodeFactory(parent=project, creator=self.user)
E   AttributeError: 'TestProjectViews' object has no attribute 'user'
============================ pytest-warning summary ============================
```

scripts/fix_fork_component_root.py
sloria: I'm not sure we actually need this. We already have `scripts.fix_root`. I can just run that after this fix is deployed.

brianjgeiger: Okay, we'll take this out and make a note in the deployment.

sloria:
Merged into `release/next-platform`. 🚢 🇮🇹

PR #7915, opened 2017-11-17
[hotfix] handle search error [OSF-8944]

alexschiller:
## Purpose
Query syntax errors from elasticsearch are raising an exception with 'search_phase_execution_exception' that is not caught properly. This is a hack-ish fix to get our builds passing and present a better error message via the api. Currently it is pasting the raw exception data on prod (which is enormous).

## Changes
Catch `search_phase_execution_exception` and reraise as a MalformedQueryException.

## Side effects
Slightly less detail on error since the failed query is not passed back when this error is caught.

## Ticket

https://openscience.atlassian.net/browse/OSF-8944

### DIFF
--- a/website/search/elastic_search.py
+++ b/website/search/elastic_search.py
@@ -112,6 +112,8 @@ def requires_search(func):
             except NotFoundError as e:
                 raise exceptions.IndexNotFoundError(e.error)
             except RequestError as e:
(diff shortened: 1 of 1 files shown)


alexschiller:
Changed my mind, making this a feature, I don't know the current state of prod to be sure this will not cause issues.

PR #7919, opened 2017-11-20
[Feature] Access Requests [#OSF-8576]

mfraezz:
## Purpose
Allow users to request access to a project

## Changes
* New models
* New API views
* Tests

## Note
Blocked by ~~#7843~~ #7924  

## Side effects
None expected

## Ticket
[[#OSF-8576]](https://openscience.atlassian.net/browse/OSF-8576)


api/requests/urls.py
```diff
@@ -0,0 +1,8 @@
+from django.conf.urls import url
+
+from . import views
+
+urlpatterns = [
+    url(r'^(?P<request_id>\w+)/$', views.NodeRequestDetail.as_view(), name=views.NodeRequestDetail.view_name),
+    url(r'^(?P<request_id>\w+)/actions/$', views.CreateNodeRequestAction.as_view(), name=views.CreateNodeRequestAction.view_name),
```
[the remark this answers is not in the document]
caseyrollins: Related: If the above was the case, I don't think the `/requests/` endpoints would be needed (all operations would be handled within `/actions/` or `nodes/<id>/requests/`).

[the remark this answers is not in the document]
caseyrollins: Also, it seems like you meant to make requests a general model (`AbstractRequest`), but these endpoints seem to only handle a specific type of request (`NodeRequest`). 

Do requests need to be this abstracted? What are the future use cases? If so, I think these endpoints should be generalized a bit more.

osf/models/node.py
```diff
@@ -302,6 +302,7 @@ def contributors(self):
     is_fork = models.BooleanField(default=False, db_index=True)
     is_public = models.BooleanField(default=False, db_index=True)
     is_deleted = models.BooleanField(default=False, db_index=True)
+    access_requests_enabled = models.BooleanField(default=True, db_index=True)
```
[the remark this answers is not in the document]
mfraezz: From epic ticket:
>After discussion with [Product], "allow users to request access to a project" functionality will be opt-out via the project settings page, which should alleviate the above concerns.

There's also a user-stories doc somewhere that specified this as opt-out. ¯\\\_(ツ)_/¯

api/requests/urls.py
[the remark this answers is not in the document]
mfraezz: >It seems that most `BaseAction` operations should be handled within the `/actions/` endpoint. [...]

When this was written, the `Generalize Actions` ticket had received cross-team discussion, and it was agreed that related views (e.g. `/requests/<id>/actions`) were preferable to sub-routes (`actions/requests/`). Some time later, this decision was reversed, and this PR hadn't been updated from that. It would make sense to do that now.

>  I don't think the `/requests/` endpoints would be needed

You're probably right about that... at least until other types of `AbstractRequest` are added.

>Do requests need to be this abstracted? 

According to all the discussion on this (from months ago), yes.

>What are the future use cases?

Other things users might "request"? -- "Unlock this file", "Let me update this file without having full-write", "Share this private (non-node) thing with me" ¯\\\_(ツ)_/¯ 
All of these are guesses, and I'm sure there are things no one has thought of yet. To fully answer this question, you'd need a 🔮 

>If so, I think these endpoints should be generalized a bit more.

The `/actions/` sub-route and related view through `/nodes/` should accomplish that.

api/actions/views.py
```diff
@@ -65,7 +73,7 @@ class ActionDetail(JSONAPIBaseView, generics.RetrieveAPIView):
     permission_classes = (
         permissions.IsAuthenticatedOrReadOnly,
         base_permissions.TokenHasScope,
-        ReviewActionPermission,
```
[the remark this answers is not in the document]
mfraezz: It broke `RequestActions` going through the same endpoint. IIRC, I did some object permission magic to avoid breaking any of the tests for `ReviewActions`, but I'll double check that

osf/utils/workflows.py
```diff
@@ -62,3 +62,7 @@ class DefaultTriggers(ChoiceEnum):
         'after': ['save_action', 'save_changes', 'notify_edit_comment'],
     },
 ]
+
+@unique
+class RequestTypes(ChoiceEnum):
```
erinspace: Just a question - does this belong here, or perhaps in `request.py` as it seems to be very `AbstractRequest` specific and maybe not needed by other models?

osf/utils/machines.py
```diff
@@ -143,3 +144,64 @@ def get_context(self):
...
+        return False
+
+    def notify_submit(self, ev):
+        """ Notify admins that someone is requesting access
+        """
+        context = self.get_context()
+        context['contributors_url'] = '{}contributors/'.format(self.machineable.target.absolute_url)
+        context['project_settings_url'] = '{}settings/'.format(self.machineable.target.absolute_url)
+        for admin in self.machineable.target.admin_contributors:
+            mails.send_mail(
```
erinspace: just curious - seems like reviews uses signals to send emails, is there an advantage to sending the email directly here instead of sending a signal for consistency?

osf/utils/machines.py
```diff
@@ -143,3 +144,64 @@ def get_context(self):
...
+        """ Notify requester that admins have approved/denied
+        """
+        if ev.event.name == DefaultTriggers.REJECT.value:
+            context = self.get_context()
+            mails.send_mail(
+                self.machineable.creator.username,
+                mails.ACCESS_REQUEST_DENIED,
+                **context
+            )
+        else:
```
erinspace: The comment is nice, but if this isn't gonna handle other potential values of `DefaultTriggers` dunno if you need a whole empty `except` for it at the moment

osf/utils/machines.py
```diff
@@ -143,3 +144,64 @@ def get_context(self):
...
         }
+
+class RequestMachine(BaseMachine):
+    ActionClass = NodeRequestAction
+
+    def save_changes(self, ev):
+        """ Handles contributorship changes and state transitions
+        """
+        if ev.event.name == DefaultTriggers.ACCEPT.value:
+            self.machineable.target.add_contributor(self.machineable.creator, auth=Auth(ev.kwargs['user']), permissions=permissions.READ, send_email='{}_request'.format(self.machineable.request_type))
```
erinspace: A few more line breaks would go a long way with readability here

api/actions/views.py
```diff
@@ -75,10 +82,21 @@ class ActionDetail(JSONAPIBaseView, generics.RetrieveAPIView):
     view_category = 'actions'
     view_name = 'action-detail'
 
+    def get_serializer_class(self):
+        # Not allowed to view NodeRequestActions yet, making extra logic unnecessary
+        # obj = self.get_object()
```
erinspace: Assuming these commented out bits are for later on? Since they're simple enough, may make sense to remove them for now for realz and put them back when it's needed?

api/requests/permissions.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class NodeRequestPermission(drf_permissions.BasePermission):
+    def has_object_permission(self, request, view, obj):
+        auth = get_user_auth(request)
+        if auth.user is None:
+            return False
+
+        target = None
+        trigger = None
```
erinspace: I don't think you need to define this here - if trigger isn't set in any of the below if/elif statements it'll raise a `ValueError` anyhow

erinspace:
Just a couple questions from me! I realize I probs should have done a real formal review so they show up better but oh well, lesson learned... 

Pass complete!

osf/utils/workflows.py
mfraezz: Could've gone either way with this. I opted to put it here for three main reasons:
* It seemed to fit more with `utils` than with `models`, as it's not stored in the database
* `AbstractRequest` objects are machineable, and machines have workflow-related metadata that governs their behavior. This seemed to qualify as "workflow-related metadata," the rest of which resides in this file.
* This may need to be checked in certain places where `AbstractRequests` might not need to be imported. At present, it's not, but there could be cases where doing an import from the model file would cause a circular import

Given that, if you think I should put it with the model, I will.

osf/utils/machines.py
mfraezz: Advantages include simplicity, readability, and transparency. There was a moderation bug where emails were not being sent -- this was because the signal listeners weren't imported before app initialization. The actual mail is sent asynchronously, so it shouldn't be any less efficient than using signals. 

The only case signals are really worth the hassle is when you need to trigger the same behavior from multiple locations where a method import would cause problems (e.g. circular imports). That isn't true here (and may or may not have been true for moderation), so I didn't use signals.

osf/utils/machines.py
mfraezz: There are only two cases for possible event triggers here. One is covered by the above case, and the comment in this passed conditional branch is present to serve as an explanation for why the other case is not explicitly covered. The `if` is necessary, and `else` should add approximately 0.0ms to execution time. Also, it seemed to make sense to visually order it this way, as a comment at the top would need more words.

api/actions/views.py
mfraezz: True. I had them in for real, realized that view permissions for these objects was "none yet", and just commented it out with a brief explanation. Since we'll be moving to `django-guardian` (hopefully) soon, this might be obviated by the time it becomes necessary anyway. I'll remove.

mfraezz:
@erinspace, review responded. 👾

osf/utils/workflows.py
erinspace: Naw, makes sense!

osf/utils/machines.py
erinspace: Yup, makes sense as well!

osf/utils/machines.py
erinspace: all good!

api/actions/urls.py
```diff
@@ -6,5 +6,6 @@
 
 urlpatterns = [
     url(r'^reviews/$', views.ReviewActionListCreate.as_view(), name=views.ReviewActionListCreate.view_name),
+    url(r'^requests/$', views.NodeRequestActionCreate.as_view(), name=views.NodeRequestActionCreate.view_name),
```
erinspace: The only other question I had (this really is just a one off, not a whole review) is this requests actions url  
 is `NodeRequestAction` specific -- will there ever be a case where the `/requests/` endpoint will ever be used to create anything but NodeActions? Would it make sense to make this generic now or when the time comes?

mfraezz: See [this comment thread](https://github.com/CenterForOpenScience/osf.io/pull/7919#discussion_r161871794). With `django-guardian` on the horizon, we'll be able to grant permissions in a more granular way, with `*Request` objects driving their allocation. It's somewhat ephemeral at this time, I admit, but it doesn't make sense to not do a little future-proofing now. At worst, it gets rewritten (as it would need to be anyway), and at best, it's extensible in a way that saves us a lot of headaches down the road 🐼

api/nodes/serializers.py
```diff
@@ -166,6 +166,7 @@ class NodeSerializer(JSONAPISerializer):
     preprint = ser.BooleanField(read_only=True, source='is_preprint')
     fork = ser.BooleanField(read_only=True, source='is_fork')
     collection = ser.BooleanField(read_only=True, source='is_collection')
+    access_requests_enabled = ser.BooleanField(read_only=False, required=False)
```
brianjgeiger: Drive-by: I think we need a relationship on the node for this, and that relationship should be able to give meta on how many requests currently exist with the related_counts query parameter (see https://osf.io/73dq8/wiki/home/ )

sloria:
Merged into `release/next-platform` 🚢 🇮🇹

PR #7925, opened 2017-11-21
[EOSF-933] Raise GONE for deleted quick files view

jamescdavis:
## Purpose

When a Quick File has been deleted, the OSF should return a page not found rather than routing to Ember.

## Changes

Add a condition to check if a Quick File is deleted and raise GONE if it is.

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/EOSF-933

### DIFF
--- a/website/views.py
+++ b/website/views.py
@@ -306,7 +306,7 @@ def resolve_guid(guid, suffix=None):
 
         if isinstance(referent, BaseFileNode) and referent.is_file and referent.node.is_quickfiles:
             if referent.is_deleted:
-                raise HTTPError(http.NOT_FOUND)
+                raise HTTPError(http.GONE)
             if PROXY_EMBER_APPS:
                 resp = requests.get(EXTERNAL_EMBER_APPS['ember_osf_web']['server'], stream=True)
                 return Response(stream_with_context(resp.iter_content()), resp.status_code)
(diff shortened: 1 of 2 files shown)


website/views.py
```diff
@@ -305,6 +305,8 @@ def resolve_guid(guid, suffix=None):
             return send_from_directory(preprints_dir, 'index.html')
 
         if isinstance(referent, BaseFileNode) and referent.is_file and referent.node.is_quickfiles:
+            if referent.is_deleted:
+                raise HTTPError(http.NOT_FOUND)
```
alexschiller: Two thoughts:
1. Should this be 410 (GONE)? No strong feelings, but IIRC the api was supposed to 410.
2. A test would be desirable as well to make sure this doesn't happen again. Here's some potential code (no promises and will likely need to modify imports (`from osf.models import AbstractNode, NodeLog, QuickFilesNode`):

https://github.com/CenterForOpenScience/osf.io/blob/develop/tests/test_views.py#L4546

```
    def test_deleted_quickfiles_404(self):
        user = AuthUserFactory()
        quickfiles = QuickFilesNode.objects.get(creator=user)
        osfstorage = quickfiles.get_addon('osfstorage')
        root = osfstorage.get_root()
        test_file = root.append_file('pole_position_was_ok.txt')
        guid = test_file.get_guid(create=True)._id
        test_file.is_deleted = True
        test_file.save()
        
        url = web_url_for('resolve_guid', _guid=True, guid=guid)
        res = self.app.get(url, expect_errors=True)

        assert_equal(res.status_code, 404)
        assert_equal(
            res.request.path,
            '/{}/'.format(guid)
        )
```

PR #7926, opened 2017-11-21
[IN-241][OSF] Add whitelist functionality to admin site for SHARE providers.

adlius:
## Purpose

Add whitelist functionality to admin site for SHARE providers so that preprint discover page could get the info from `/preprint_providers` API endpoint instead of hardcoding them in the codebase.

## Changes

Add a new `SharePreprintProviderWhitelisted` model. New views are created for the admin site to add and delete SHARE providers.

## Side effects

None.

## Ticket

https://openscience.atlassian.net/browse/OSF-8595


api_tests/reviews/mixins/filter_mixins.py
```diff
@@ -259,5 +259,10 @@ def test_review_provider_filters(self, app, url, moderator_pair, admin_pair, exp
...
         # filter by permissions requires auth
-        res = get_actual(app, url, expect_errors=True, permissions='set_up_moderation')
-        assert res.status_code == 401
+        from django.db import transaction
+        try:
+            with transaction.atomic():
+                res = get_actual(app, url, expect_errors=True, permissions='set_up_moderation')
+                assert res.status_code == 401
+        except:
+            pass
```
aaxelb: This looks like an actual problem that shouldn't be swept under the rug... If there's an error while handling `/v2/preprint_providers/`, `get_renderer_context` will cause a 500 (instead of a friendly 4xx) by trying to make an additional query.

Maybe it could check whether the response has an error code and, if so, skip the `meta`? Or is there a way to directly check if the current transaction/connection has errored?

```
  File "/usr/local/lib/python2.7/wsgiref/handlers.py", line 85, in run
  File "/usr/local/lib/python2.7/site-packages/django/contrib/staticfiles/handlers.py", line 63, in __call__
  File "/usr/local/lib/python2.7/site-packages/django/core/handlers/wsgi.py", line 157, in __call__
  File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 124, in get_response
    response = self._middleware_chain(request)
  File "/usr/local/lib/python2.7/site-packages/django/core/handlers/exception.py", line 43, in inner
    response = response_for_exception(request, exc)
  File "/usr/local/lib/python2.7/site-packages/django/core/handlers/exception.py", line 93, in response_for_exception
    response = handle_uncaught_exception(request, get_resolver(get_urlconf()), sys.exc_info())
  File "/usr/local/lib/python2.7/site-packages/django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 249, in _legacy_get_response
    response = self._get_response(request)
  File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 187, in _get_response
  File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 185, in _get_response
  File "/usr/local/lib/python2.7/site-packages/django/utils/decorators.py", line 185, in inner
    return func(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/django/views/generic/base.py", line 68, in view
  File "/usr/local/lib/python2.7/site-packages/rest_framework/views.py", line 491, in dispatch
    self.response = self.finalize_response(request, response, *args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/rest_framework/views.py", line 416, in finalize_response
    response.renderer_context = self.get_renderer_context()
  File "/code/api/preprint_providers/views.py", line 85, in get_renderer_context
    'whitelisted_providers': [unicode(provider) for provider in SharePreprintProviderWhitelisted.objects.all()]
  File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 250, in __iter__
    self._fetch_all()
  File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 1118, in _fetch_all
    self._result_cache = list(self._iterable_class(self))
  File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 53, in __iter__
    results = compiler.execute_sql(chunked_fetch=self.chunked_fetch)
  File "/usr/local/lib/python2.7/site-packages/django/db/models/sql/compiler.py", line 894, in execute_sql
    raise original_exception
TransactionManagementError: An error occurred in the current transaction. You can't execute queries until the end of the 'atomic' block.
[09/Jan/2018 14:10:14] "GET /v2/preprint_providers/?filter[permissions]=set_up_moderation HTTP/1.1" 500 59
```

adlius:
PR updated. The `with ... except:` in the test was added to avoid the error and is removed as the error seems to be no longer occurring.

osf/models/preprint_provider.py
```diff
@@ -128,8 +128,22 @@ def rules_to_subjects(rules):
...
+
+class SharePreprintProviderWhitelisted(models.Model):
+    id = models.AutoField(primary_key=True)
+    provider_name = models.CharField(default=None, blank=True, unique=True, max_length=200)
+
+    @classmethod
+    def create(cls, provider_name):
+        provider, created = SharePreprintProviderWhitelisted.objects.get_or_create(provider_name=provider_name)
+        if created:
+            provider.save()
```
aaxelb: If `provider` was created, it's already saved.

I think it would make more sense to delete `SharePreprintProviderWhitelisted.create` and instead use `SharePreprintProviderWhitelisted.objects.get_or_create`, since they now do exactly the same thing, and the latter is more djangoey/consistent.

osf/models/preprint_provider.py
```diff
@@ -128,8 +128,22 @@ def rules_to_subjects(rules):
...
-
 @receiver(post_save, sender=PreprintProvider)
 def create_provider_auth_groups(sender, instance, created, **kwargs):
     if created:
         GroupHelper(instance).update_provider_auth_groups()
+
+
+class SharePreprintProviderWhitelisted(models.Model):
+    id = models.AutoField(primary_key=True)
+    provider_name = models.CharField(default=None, blank=True, unique=True, max_length=200)
```
aaxelb: If it's not nullable, `default=None` will error. Probably don't want a default at all?

aaxelb: Also not sure we want `blank=True`, this seems like the sort of thing that should always have a value.

admin/static/js/pages/whitelist-add-page.js
```diff
@@ -0,0 +1,62 @@
...
+    var self = this;
+    self.preprint_providers = ko.observableArray([]);
+    self.providers_added = ko.observableArray([]);
+    self.share_api_url = window.templateVars.share_api_url + "/api/v2/search/creativeworks/_search";
+    self.api_v2_url = window.templateVars.api_v2_url + "preprint_providers/"
+};
+
+AddWhitelistPreprintProvidersViewModel.prototype.updatePreprintProviders = function() {
+    var self = this;
+    var query_obj = {
```
sloria: Minor: as per our style guidelines, use `lowerCamelCase` for variable names.

osf/models/preprint_provider.py
```diff
@@ -128,8 +128,15 @@ def rules_to_subjects(rules):
...
     return Subject.objects.filter(reduce(lambda x, y: x | y, q)) if len(q) > 1 else (Subject.objects.filter(q[0]) if len(q) else Subject.objects.all())
 
-
 @receiver(post_save, sender=PreprintProvider)
 def create_provider_auth_groups(sender, instance, created, **kwargs):
     if created:
         GroupHelper(instance).update_provider_auth_groups()
+
+
+class SharePreprintProviderWhitelisted(models.Model):
```
sloria: Minor: `WhitelistedSHAREPreprintProvider` would be more syntactically correct.

admin/static/js/pages/whitelist-add-page.js
```diff
@@ -0,0 +1,62 @@
+'use strict';
+
+var $ = require('jquery');
+var ko = require('knockout');
+
+var AddWhitelistPreprintProvidersViewModel = function() {
+    var self = this;
+    self.preprint_providers = ko.observableArray([]);
+    self.providers_added = ko.observableArray([]);
+    self.share_api_url = window.templateVars.share_api_url + "/api/v2/search/creativeworks/_search";
```
sloria: Minor: attributes should also use `lowerCamelCase`.

admin/preprint_providers/views.py
```diff
@@ -367,3 +367,38 @@ def get_context_data(self, *args, **kwargs):
         kwargs['show_taxonomies'] = SHOW_TAXONOMIES_IN_PREPRINT_PROVIDER_CREATE
         kwargs['tinymce_apikey'] = settings.TINYMCE_APIKEY
         return super(CreatePreprintProvider, self).get_context_data(*args, **kwargs)
+
+
+class SharePreprintProviderWhitelistAdd(PermissionRequiredMixin, View):
+    permission_required = 'osf.change_preprintprovider'
```
sloria: You'll need to add a custom migration to add this permission to the `osf_admin` Group. See this for an example: https://github.com/CenterForOpenScience/osf.io/blob/f575caf19c2b3b08d5f0b5636d03bae992a13f18/osf/migrations/0078_add_banner_permissions.py

adlius: `change_preprintprovider` is an existing Permission that I reuse. I checked through osfshell locally and `osf_admin` group has this permission. It seems that a previous migration (https://github.com/CenterForOpenScience/osf.io/blob/f575caf19c2b3b08d5f0b5636d03bae992a13f18/osf/migrations/0030_preprint_provider_institution_permissions.py#L44) already adds this permission to the group.

sloria: Ah, sure enough. Sorry, I thought this was a new permission. No action necessary then

admin/preprint_providers/urls.py
```diff
@@ -17,4 +17,6 @@
     url(r'^(?P<preprint_provider_id>[a-z0-9]+)/delete/$', views.DeletePreprintProvider.as_view(), name='delete'),
     url(r'^(?P<preprint_provider_id>[a-z0-9]+)/export/$', views.ExportPreprintProvider.as_view(), name='export'),
     url(r'^(?P<preprint_provider_id>[a-z0-9]+)/share_source/$', views.ShareSourcePreprintProvider.as_view(), name='share_source'),
+    url(r'^whitelist_add/$', views.SharePreprintProviderWhitelistAdd.as_view(), name='whitelist_add'),
```
sloria: You should be able to have one endpoint rather than 2 if you make these URLs more RESTful:

```python
POST .../whitelist/
DELETE .../whitelist/
```

adlius:
Per @sloria 's comment, I refactored the entire front-end implementation and do away with Django forms. Two different urls and their views are consolidated into one `/whitelist` view. However, because HTTP form submission does not allow `DELETE` (I did not know this until today), the `/whitelist` endpoint allows both `GET` and `POST` and some simple logic is implemented in the `post` method to distinguish addition from deletion. The resulting PR is much more organized and less cluttered.

admin/package.json
```diff
@@ -9,10 +9,10 @@
     "bower": "^1.3.12",
     "font-awesome": "^4.5.0",
     "font-awesome-webpack": "0.0.5-beta.2",
-    "json-loader": "^0.5.1",
     "jquery-datetimepicker": "^2.5.16",
+    "json-loader": "^0.5.1",
     "keen-js": "^3.0.0",
-    "knockout": "~3.2.0",
+    "knockout": "3.4.0",
```
sloria: Why is this necessary?

admin/static/js/pages/whitelist-page.js
```diff
@@ -0,0 +1,100 @@
...
+                "name": item,
+            });
+        });
+        self.isLoading(false);
+    });
+};
+
+WhitelistPreprintProvidersViewModel.prototype.switchPage = function() {
+    var self = this;
+    self.preprintProviders = ko.observableArray([]);
```
sloria: No need to initialize these again. You can just reset them to the empty values:

```js
self.preprintProviders([]);
```

adlius:
Changed to submit AJAX request.

admin/package.json
adlius: Somehow, Knockout@~3.2.0 does not seem to support CommonJS style `require('knockout')` imports. I believe https://github.com/knockout/knockout/pull/1526 is relevant. Bumping to 3.4.0 solves the problem.

laurenbarker:
@adlius you have a failing test -- is it related to your PR? (I re-ran the job and it still failed)

adlius:
@laurenbarker the latest failure is a migration conflict. It should work now. On the side note, I've updated the migration for about the 10th time now. It would be nice if we can get this merged ASAP.

PR #7929, opened 2017-11-24
[hotfix] Return 410 for a deleted file [OSF-8963]

fabmiz:
## Purpose
Currently, looking up a deleted file via api file detail endpoint with **guid** returns 200 and the file object. 

## Changes

Add check for `is_deleted` on the returned file object

## Ticket

[OSF-8963](https://openscience.atlassian.net/browse/OSF-8963)

## After

screen shot 2017-11-24 at 9 21 47 am

### DIFF
--- a/api_tests/files/views/test_file_detail.py
+++ b/api_tests/files/views/test_file_detail.py
@@ -60,11 +60,24 @@ class TestFileView:
         res = app.get(file_url, auth=non_contributor.auth, expect_errors=True)
         assert res.status_code == 403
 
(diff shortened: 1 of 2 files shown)


api_tests/files/views/test_file_detail.py
```diff
@@ -393,7 +406,7 @@ def test_write_contrib_can_checkin(self, app, node, file, file_url):
         )
         file.reload()
         assert res.status_code == 200
-        assert file.checkout == None
+        assert file.checkout is None
```
sloria: Thanks for changing these.

PR #7934, opened 2017-11-28
Feature/Dependency Upgrades [OSF-8936]

pattisdr:
https://openscience.atlassian.net/browse/OSF-8936


pattisdr:
More breaking changes than I originally thought. Need to work on this more.

PR #7940, opened 2017-11-29
[Hotfix] Fix roots for Forks/Templates [OSF-8993]

mfraezz:
## Purpose
Be better.

## Changes
* Ensure `NodeRelation`s exist when they need to

## Side effects
None expected

## Ticket
[OSF-8993](https://openscience.atlassian.net/browse/OSF-8993)

### DIFF
--- a/osf/models/node.py
+++ b/osf/models/node.py
@@ -1828,11 +1828,12 @@ class AbstractNode(DirtyFieldsMixin, TypedModel, AddonModelMixin, IdentifierMixi
         return False
 
     # TODO: Optimize me (e.g. use bulk create)
-    def fork_node(self, auth, title=None):
+    def fork_node(self, auth, title=None, parent=None):
         """Recursively fork a node.
 
         :param Auth auth: Consolidated authorization
         :param str title: Optional text to prepend to forked title
+        :param Node parent: Sets parent, should only be non-null when recursing
         :return: Forked node
         """
         Registration = apps.get_model('osf.Registration')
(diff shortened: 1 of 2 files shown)


osf/models/node.py
```diff
@@ -1869,13 +1870,17 @@ def fork_node(self, auth, title=None):
...
         for node_relation in original.node_relations.filter(child__is_deleted=False):
             node_contained = node_relation.child
             # Fork child nodes
             if not node_relation.is_node_link:
                 try:  # Catch the potential PermissionsError above
-                    forked_node = node_contained.fork_node(auth=auth, title='')
+                    forked_node = node_contained.fork_node(auth=auth, title='', parent=forked)
                 except PermissionsError:
                     pass  # If this exception is thrown omit the node from the result set
                     forked_node = None
```
mfraezz: There's a good chance that this line to the next `else` is superfluous, but didn't want to risk it in a hotfix, as it's not hurting anything (other than performance by perhaps a couple milliseconds).

osf/models/node.py
```diff
@@ -2048,20 +2055,26 @@ def use_as_template(self, auth, changes=None, top_level=True):
...
             node_contained = node_relation.child
             # template child nodes
             if not node_relation.is_node_link:
                 try:  # Catch the potential PermissionsError above
-                    templated_child = node_contained.use_as_template(auth, changes, top_level=False)
+                    templated_child = node_contained.use_as_template(auth, changes, top_level=False, parent=new)
                 except PermissionsError:
                     pass
                 else:
                     NodeRelation.objects.get_or_create(
```
mfraezz: Ditto this call, as above

PR #7962, opened 2017-12-04
Feature/Requirements Upgrade [OSF-8936]

pattisdr:
## Ticket
https://openscience.atlassian.net/browse/OSF-8936

(Please see ticket for notes)

## QA
All user-entered fields for preprint providers need to be thoroughly checked - preprint footers, for example. Make sure the formatting doesn't look weird.
(Note to self: add more detail to this section)


admin_tests/preprint_providers/test_views.py
```diff
@@ -188,7 +188,7 @@ def test_html_fields_are_stripped(self):
 
         stripped_advisory_board = '<div><ul><li>Bill Nye</li></ul></div>'
         stripped_description = '<span>Open Preprints Open Science</span>'
-        stripped_footer_links = '<p>Xiv: Support | Contact | <a href=""><span class="fa fa-facebook"></span></a></p>'
+        stripped_footer_links = '<p>Xiv: Support | </p>Contact | <a href=""><span class="fa fa-facebook"></span></a><p></p>'
```
sloria: Drive-by: Will we need to update `footer_links` and other similar fields for existing preprint providers? If so, what will need to be changed?

pattisdr: @sloria, it will depend on the specifics of a provider's footer_link.  From my investigation if certain tags are <em>nested inside other tags</em>, you get this weird closing tag behavior.  


screen shot 2017-12-21 at 2 49 22 pm


For this particular test, the footer_link that is being stripped looks like this: There is a script tag around "Support" and a pre tag around "Contact".  So if providers are trying to input text like that, then it might break oddly for them.  But if they're not wrapping Support and Contact with certain tags, this wouldn't happen.

```
'footer_links': '<p>Xiv: <script>Support</script> | <pre>Contact<pre> | <a href=""><span class="fa fa-facebook"></span></a></p>',"
```

sloria: Just looked in the prod DB: there are no footer_links that have `<pre>` or `<h*>` tags. That said, we'll just need to let QA know to thoroughly check all the HTML fields for each preprint provider.

sloria:
The conflict turned out to small, so I fixed it myself and merged this into `release/next-djangole`.

pattisdr:
thanks @sloria!

admin_tests/preprint_providers/test_views.py
pattisdr: Reason for strange nesting behavior: https://github.com/mozilla/bleach/issues/280#issuecomment-353466274

PR #7963, opened 2017-12-05
Feature/Remove .remove() and .remove_one() methods [OSF-8921]

pattisdr:
## Purpose
These are unnecessary artifacts of MODM, and it's dangerous to have shortcuts for deleting things from the database (we don't want to do this any more).

## Changes

Removed the BaseModel.remove and BaseModel.remove_one methods and their usages.

## Ticket

https://openscience.atlassian.net/browse/OSF-8921

## QA Considerations
Shouldn't need QA testing, there shouldn't be any obvious changes on the front end


sloria:
Merged in to release/next-djangole

PR #7981, opened 2017-12-06
[OSF-8056] Fix for addon asset loading

Johnetordoff:
## Purpose

One staging and production the osf will try to grab js assets for directories which don't exist causing a key error. This happens on addons because default addons, like osfstorage are listed on the addons page, but don't actually need any js assets because they are always enabled and configured the same way. This fix ensures we never try to load non-existent  assets for default addons. 

Currently osfstorage is the only default addon.

## Changes

Cleans up the code that loads js assets into the mako file by adding comments and unittests.

## QA Notes

Check to make sure there are no gaps in the lower level panels, the ones that say "connect account" on them.

## Side Effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/OSF-8056


tests/test_addons.py
```diff
@@ -1099,3 +1102,56 @@ def test_other_addon_redirect_download(self):
...
+        assert len(default_addons) == 1
+        assert default_addons[0]['short_name'] == 'osfstorage'
+
+    def test_collect_node_config_js(self):
+
+        addon_dicts = serialize_addons(self.node)
+
+        asset_paths = collect_node_config_js(addon_dicts)
+
+        # Default addons should be in in addon dicts, but they have no js assets because you can't
```
alexschiller: very minor: 'should be in **in** addon'

website/project/views/node.py
```diff
@@ -281,6 +281,24 @@ def node_addons(auth, node, **kwargs):
 
     ret = _view_project(node, auth, primary=True)
 
+    addon_settings = serialize_addons(node)
+
+    ret['addon_capabilities'] = settings.ADDON_CAPABILITIES
+
+    # If an an is default you cannot connect/disconnect so we don't have to load it.
```
alexschiller: very minor: 'If an **addon** is'

PR #7998, opened 2017-12-07
[Feature] Add vol param to 'info' link [OSF-8820]

fabmiz:
## Purpose

Add view_only url parameter to 'info' link for files endpoint

Follow up PR https://github.com/CenterForOpenScience/osf.io/pull/7871

## Changes

- Add vol url param to info link
- Update tests

## QA Notes

This is an API change. When looking up file detail say v2/nodes/jc3vf/files/github/?view_only=82b06ecfb70c448b95449c437a68ca0e Make sure the info has the view-only parameter as well.

## Ticket

[OSF-8820](https://openscience.atlassian.net/browse/OSF-8820)

### DIFF
--- a/api/base/serializers.py
+++ b/api/base/serializers.py
@@ -887,6 +887,9 @@ class LinksField(ser.Field):
         # not just the field attribute.
         return obj
 
+    def extend_absolute_info_url(self, obj):
+        return extend_querystring_if_key_exists(obj.get_absolute_info_url(), self.context['request'], 'view_only')
+
     def extend_absolute_url(self, obj):
         return extend_querystring_if_key_exists(obj.get_absolute_url(), self.context['request'], 'view_only')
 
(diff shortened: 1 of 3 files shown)


api/base/serializers.py
```diff
@@ -900,7 +900,10 @@ def to_representation(self, obj):
             else:
                 ret[name] = url
         if hasattr(obj, 'get_absolute_url') and 'self' not in self.links:
-            ret['self'] = self.extend_absolute_url(obj)
+            extended_url = self.extend_absolute_url(obj)
+            ret['self'] = extended_url
+            if 'info' in ret:
+                ret['info'] = extended_url
```
brianjgeiger: @fabmiz Is there a good way to extend the info url rather than replacing it with the self url? If not, verify that the previous self url matches the info url before replacing it, in case someone changes how the info urls work in the future.

PR #7999, opened 2017-12-07
[feature] Add category icons

fabmiz:
## Purpose

Add category icons for editable node category (project overview page)

## Changes

Add new logic to make dropdown menu behave
nearly similar to the category editable field.

## QA Notes

Please making sure the categories get updated on submit.
Try different browsers to make sure the dropdown works as expected.

## Side Effects

NA

## Ticket

[OSF-8596](https://openscience.atlassian.net/browse/OSF-8596)

## After

after-icons

### DIFF
--- a/website/static/css/style.css
+++ b/website/static/css/style.css
@@ -849,3 +849,11 @@ table.add-links {
 .comment-box {
     padding-bottom: 25px;
 }
+
+.node-category {
+    display: inline-block;
+}
+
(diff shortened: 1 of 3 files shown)


brianjgeiger:
@fabmiz Could you pull from Develop? I'm seeing an API failure that I'm hoping was fixed, otherwise we'll have a new ticket for it. Could also be probabilistic.

PR #8001, opened 2017-12-07
[OSF-8247] Fix Fork completion failure email language 

Johnetordoff:
## Purpose

Forking emails and error messages have bad, casing, grammar and language. This fixes that.

## Changes

Language and textual changes.
- Fork completed email should have the node title rather than the fork title;
- The line that says that they can opt-out of this type of email;
- Changed to sentence case from title case on the subject of the email;
- Used suggested wording for forking notification;
- Adjusted wording of body of success email based on suggestion and made it more compact;
- On failure email, re-worded so user sees that they can visit the original project, not the failed fork.

## QA Notes

This should now match the language described in the original ticket.

Please verify that at least one other HTML email still gives the unsubscribe option.

## Side Effects

None that I know of.

## Ticket

### DIFF
--- a/api/nodes/views.py
+++ b/api/nodes/views.py
@@ -1623,10 +1623,10 @@ class NodeForksList(JSONAPIBaseView, generics.ListCreateAPIView, NodeMixin, Node
         try:
             fork = serializer.save(node=node)
         except Exception as exc:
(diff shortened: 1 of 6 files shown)


brianjgeiger:
@Johnetordoff This looks all good, but please take a look at the changes I made to your PR description. This will be much more useful to testers to understand the changes and the consequences of those changes. If you could do PRs more like this in the future, you would save a lot of QA time and effort. Thanks!

PR #8016, opened 2017-12-08
[Hotfix] Distinct preprints [OSF-9058]

mfraezz:
## Purpose
Patch regression

## Changes
* Use distinct

## QA Notes
Check `/v2/preprints/` and ensure there are no dupes

## Side Effects
None expected.

## Ticket
[[OSF-9058]](https://openscience.atlassian.net/browse/OSF-9058)

### DIFF
--- a/api/base/filters.py
+++ b/api/base/filters.py
@@ -511,4 +511,4 @@ class PreprintFilterMixin(ListFilterMixin):
         else:
             query = no_user_query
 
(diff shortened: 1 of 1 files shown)


mfraezz:
>Can the dupes issue be reproduced in a test?

Paired to discuss. Short answer: not without knowing the exact conditions that are causing dupes, and we don't at present. Tests should be created when we solve the underlying problem

PR #8018, opened 2017-12-09
[OSF-8319] Fix admin app logs to show information about components

Johnetordoff:
## Purpose

Currently the logs in the admin app only show information strictly dealing with the node being observed and not it's children, this is a bit different then the public node logs which show some info like component creation which only is associated with the child, but not the parent.

This fix makes the queries for the normal osf.io node logs and the private admin app logs match.

## Changes

 - Changes query for the admin app private logs to match the logs on the node overview page.
 - Adds unit tests for the admin node log view. 

## QA Notes

- Create some components and see if they're logged in the admin logs.
- The node overview logs should now match the admin logs entry-for-entry.

## Side Effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/OSF-8319

### DIFF
--- a/admin_tests/nodes/test_views.py
+++ b/admin_tests/nodes/test_views.py
@@ -354,7 +354,6 @@ class TestAdminNodeLogView(AdminTestCase):
         self.user = AuthUserFactory()
         self.auth = Auth(self.user)
         self.node = ProjectFactory(creator=self.user)
-        self.node.save()
 
     def test_get_object(self):
 
@@ -372,7 +371,7 @@ class TestAdminNodeLogView(AdminTestCase):
 
         logs = view.get_queryset()
 
-        log_entry = logs.all()[0]
+        log_entry = logs.first()
         nt.assert_true(log_entry.action == 'edit_title')
         nt.assert_true(log_entry.params['title_new'] == u'New Title')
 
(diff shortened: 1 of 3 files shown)


admin_tests/nodes/test_views.py
```diff
@@ -341,3 +343,70 @@ def test_confirm_registration_as_ham(self):
...
+        log_entry = logs[0][0]
+        log_params = logs[0][1]
+        nt.assert_true(log_entry.action == NodeLog.EDITED_TITLE)
+        nt.assert_true((u'title_new', u'New Title') in log_params)
+        nt.assert_true((u'node', self.node._id) in log_params)
+
+    def test_get_logs_for_children(self):
+        """ The "create component" action is actually logged as a create_project action
+        for it's child with a log parameter 'node' having it's guid as a value. We have to ensure
+        that all the components a parent has created appear in it's admin app logs, so we can't just
```
alexschiller: very minor non-blocking grammar: 
it's = it is
its = possessive

the three it's above should be its

admin_tests/nodes/test_views.py
```diff
@@ -341,3 +343,70 @@ def test_confirm_registration_as_ham(self):
...
+class TestAdminNodeLogView(AdminTestCase):
+
+    def setUp(self):
+        super(TestAdminNodeLogView, self).setUp()
+
+        self.request = RequestFactory().post('/fake_path')
+        self.user = AuthUserFactory()
+        self.auth = Auth(self.user)
+        self.node = ProjectFactory(creator=self.user)
+        self.node.save()
```
alexschiller: unneeded save

admin_tests/nodes/test_views.py
```diff
@@ -341,3 +343,70 @@ def test_confirm_registration_as_ham(self):
...
+    def test_get_queryset(self):
+
+        self.node.set_title('New Title', auth=self.auth, save=True)
+
+        view = AdminNodeLogView()
+        view = setup_log_view(view, self.request, guid=self.node._id)
+
+        logs = view.get_queryset()
+
+        log_entry = logs.all()[0]
```
alexschiller: logs.first() is slightly more efficient by making the db do the work.

PR #8019, opened 2017-12-09
[OSF-8247] Change fork emails to use raw link instead of text

Johnetordoff:
## Purpose

Currently the fork sucess/failure emails use text labels on links instead of raw links, this fix makes things more consistent other emails by using the raw links.

## Changes

- Fork success email now uses a raw link (osf.io/...) instead of label text ("here.")
- Fork failure email now uses a raw link (osf.io/...) instead of label text ("here.")

## QA Notes

- You should know be able to see the full text of a link instead of the label, "here."

## Side Effects

None that I know of.

## Ticket

### DIFF
--- a/website/templates/emails/fork_completed.html.mako
+++ b/website/templates/emails/fork_completed.html.mako
@@ -6,7 +6,7 @@
 <tr>
   <td style="border-collapse: collapse;">
 
-    <h3 class="text-center" style="padding: 0;margin: 30px 0 0 0;border: none;list-style: none;font-weight: 300;text-align: center;">The fork of <b>${title}</b> has been successfully created: <a href="${settings.DOMAIN + guid}">here.</a></h3>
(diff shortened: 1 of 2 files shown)


website/templates/emails/fork_completed.html.mako
```diff
@@ -6,7 +6,7 @@
 <tr>
   <td style="border-collapse: collapse;">
 
-    <h3 class="text-center" style="padding: 0;margin: 30px 0 0 0;border: none;list-style: none;font-weight: 300;text-align: center;">The fork of <b>${title}</b> has been successfully created: <a href="${settings.DOMAIN + guid}">here.</a></h3>
+    <h3 class="text-center" style="padding: 0;margin: 30px 0 0 0;border: none;list-style: none;font-weight: 300;text-align: center;">The fork of <b>${title}</b> has been successfully created here: ${settings.DOMAIN + guid}</h3>
```
brianjgeiger: Do email clients and sites automatically parse urls even on html emails, or should these still be links?

Johnetordoff: They seem to for plain text, but I found a counter example in prereg_challenge email so I'll change this.

PR #7457, opened 2017-07-12
[Feature] Show all projects in 'link projects' [OSF-5360]

fabmiz:
## Purpose

Show all user projects and registrations in 'link projects' or 'add node link' modal.

Currently, the user has to know the name of the project/registrations to which they wish
to link a project. With name, they can search but registrations and projects results are not distinguishable.

This PR update adds now available  API support for distinguishing node links vs registration links,
presumably (add/remove registrations links).

## Changes

Build from PR https://github.com/CenterForOpenScience/osf.io/pull/6137 from [mluke93](https://github.com/mluke93) which was closed due to lack of API support for /registration_links/

This PR update:
- Refactor code 
- Reduce requests needed to tell a node's node-links given myprojects or all public projects.
- Fix issues with loaders and use new css loaders
- Add clear/cleanup function at modal exit.
- Only reload if user adds/removes a node link
 
## Ticket

[OSF-5360](https://openscience.atlassian.net/browse/OSF-5360)

## Before

before

## After
after-responsiveness-fix-2

**latest**
latest-fix

## QA considerations
Since the ticket has been sitting for a while, please make sure the behavior is as shown in the latest gif. Mostly regular usage.
- Check if guid searching works
- See if the the modal behaves as expected for smaller screens and mobile
- See if users with only one name display correctly
- See if you leave the query in the search box, and try pressing the different options (myprojects, all projects, registrations, etc.) they display results related to that query.
- Make sure the current project or its current descendants are not returned as nodes you can link to.

PR #8028, opened 2017-12-11
[feature] Fix multiple filtering, add tests, modal ui [OSF-5360]

fabmiz:
## Purpose

Follow up for https://github.com/CenterForOpenScience/osf.io/pull/7457
to address issues found in QA testing 

## Changes

- Fix filtering when you have multiple filters
- Update tests
- Clear on modal exit or when user clicks `done`
- Clear errors and warnings upon successful linking

## Ticket

[OSF-5360](https://openscience.atlassian.net/browse/OSF-5360)

### DIFF
--- a/api_tests/nodes/filters/test_filters.py
+++ b/api_tests/nodes/filters/test_filters.py
@@ -24,8 +24,19 @@ class NodesListFilteringMixin(object):
         return AuthUserFactory()
 
     @pytest.fixture()
-    def project(self, user):
-        return ProjectFactory(creator=user, title='Lait Cafe et Sucre', is_public=True)
+    def parent_project_one(self, user):
+        parent_project_one = ProjectFactory(creator=user)
+        parent_project_one.title = parent_project_one._id
+        parent_project_one.save()
+        return parent_project_one
+
+    @pytest.fixture()
+    def child_project_one(self, user, parent_project_one):
+        return ProjectFactory(parent=parent_project_one, title="Child of {}".format(parent_project_one._id), creator=user)
+
+    @pytest.fixture()
+    def project(self, user, parent_project_one):
+        return ProjectFactory(creator=user, title='Neighbor of {}'.format(parent_project_one._id))
 
     @pytest.fixture()
     def parent_project(self, user, contrib):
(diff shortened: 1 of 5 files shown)


api_tests/nodes/filters/test_filters.py
```diff
@@ -142,6 +142,13 @@ def test_non_mutating_list_filtering_tests(
         assert child_node_two._id not in ids
         assert project._id in ids
 
+    #   test_root_ne_with_title_id_excludes_children_with_query_in_title_or_id
```
brianjgeiger: So to verify, this gets all nodes that have a title or id equal to the `child_node_two` id as long as they aren't a child of the `root` node, yes?

api_tests/nodes/filters/test_filters.py
```diff
@@ -142,6 +142,13 @@ def test_non_mutating_list_filtering_tests(
...
+        id_or_title = 'filter[title,id]={}'.format(child_node_two._id)
+        url = '{}{}&{}'.format(root_ne_url, parent_project._id, id_or_title)
+        res = app.get(url, auth=user.auth)
+        assert res.status_code == 200
+        assert len(res.json['data']) == 0
```
brianjgeiger: This would be a better test if it searched for exactly something. Preferably it should turn up two or three nodes, one that matches on the title condition and one that matches on the id condition and possibly one that matches on both at the same time, but definitely excludes something.

So, for example, if you had two node structures:

- id: AAAA name: Root 1
    - id: AAAB name: Friend of CCCC (this is a child of AAAA)
- id: CCCC name: CCCC
    - id: CCCD name: Child of CCCC (this is a child of CCCC)
- id: EEEE name: Not related to CCCC

So if you were filtering title,id = CCCC and the ne root = AAAA, You should get back 
- CCCC
- CCCD
- EEEE

But not AAAB and definitely not AAAA. Does that make sense?

api_tests/nodes/filters/test_filters.py
fabmiz: Correct!!

api_tests/nodes/filters/test_filters.py
fabmiz: Nice!! I have addressed this in the PR update.

PR #8031, opened 2017-12-12
[#OSF-8584]Trigger for wiki widget "Read More" does not match display field length

chennan47:
## Purpose
Fix the case when character.length == 400. Now doesn't truncate the text until character length hit 401.

## Changes
Before change:
when your wiki has 400 character,
screen shot 2017-12-12 at 1 32 05 pm
After change:
when your wiki has 400 character,
screen shot 2017-12-12 at 1 32 43 pm

## Ticket
https://openscience.atlassian.net/browse/OSF-8584#add-comment


brianjgeiger:
@chennan47 Did you test this for values around the edge case (398 through 402 or whatever)?

chennan47:
Yes. The backend actually works fine. The reason for this 400 character truncate happening is the truncate method we called in the javascript. The number now is set to 401 for the text. truncate. I demo the result to @Amanda with 399 , 400 and 401 characters. All work as expected.

PR #8040, opened 2017-12-13
[OSF-7619] Make download as zip button visible to non-contribs

Johnetordoff:
## Purpose

Currently Github's Download-as-zip button only appears when the user is a contributor to that project. This fix makes the button visible and usable for all users.

## Changes

Changed js condition to apply more broadly to all folders.

## QA Notes

Look at a public project which has a Github addon, all folders in that addon should be downloadable as zip. Try as a contributor and non-contributor. 

## Side Effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/OSF-7619

### DIFF
--- a/addons/github/static/githubFangornConfig.js
+++ b/addons/github/static/githubFangornConfig.js
@@ -238,7 +238,7 @@ var _githubItemButtons = {
                             },
                             icon: 'fa fa-download',
                             className: 'text-primary'
-                        }, 'Download'),
+                        }, 'Download as zip'),
                         m.component(Fangorn.Components.button, {
                             onclick: function (event) {
                                 window.open(item.data.urls.repo, '_blank');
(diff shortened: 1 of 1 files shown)


addons/github/static/githubFangornConfig.js
```diff
@@ -192,6 +192,14 @@ var _githubItemButtons = {
                         className: 'text-info'
                     }, branchArray)
                 );
+            } else {
```
brianjgeiger: @Johnetordoff So if I'm reading this correctly, you can't download the entire repo as a zip. Is that a) what this else block means; and b) correct?

Johnetordoff: So the user can download the entire repo as a zip. They have always been able to to this via the normal download button on the root level directory. The "Download" button there actually functions as a download as zip button.

Here: https://github.com/CenterForOpenScience/osf.io/blob/develop/addons/github/static/githubFangornConfig.js#L236

To download the entire repo you presumably need to the branch as a parameter which is why this is a little strange. Didn't really want to change existing behavior, but upon reflection we should probably change the root level download button to "Download as Zip" to bring it in line.

PR #8043, opened 2017-12-13
[feature] Add leading slash only if none exists [OSF-8696]

alexschiller:
## Purpose
Uploads are getting an extra leading slash which is causing problems.

## Changes
Only add a slash if there isn't one already

## QA Notes
Sufficient test plan already in place.

## Side Effects
NA

## Ticket

https://openscience.atlassian.net/browse/OSF-8696

### DIFF
--- a/website/conferences/utils.py
+++ b/website/conferences/utils.py
@@ -56,7 +56,9 @@ def prepare_contributors(admins):
 
 def upload_attachment(user, node, attachment):
     attachment.seek(0)
-    name = '/' + (attachment.filename or settings.MISSING_FILE_NAME)
+    name = (attachment.filename or settings.MISSING_FILE_NAME)
+    if name[0] != '/':
+        name = '/' + name
     content = attachment.read()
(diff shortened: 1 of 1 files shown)


website/conferences/utils.py
```diff
@@ -56,7 +56,9 @@ def prepare_contributors(admins):
 
 def upload_attachment(user, node, attachment):
     attachment.seek(0)
-    name = '/' + (attachment.filename or settings.MISSING_FILE_NAME)
+    name = (attachment.filename or settings.MISSING_FILE_NAME)
+    if name[0] != '/':
+        name = '/' + name
```
brianjgeiger: There's a python method to do this. I can't remember which, and I have a meeting, but I found it via @laurenrevere so she can probably tell you what it is.

brianjgeiger: `urllib.parse.urljoin`

laurenrevere: In python 2.7 it's `urlparse.urljoin`, though.

PR #8049, opened 2017-12-14
Feature/Clean Up Identifiers [OSF-8932]

pattisdr:
## Purpose

Identifier methods that are being used for multiple types of objects - preprints and nodes for example, should be in a more central location.   

## Changes
- Move update_ezid_metadata_on_change from website/preprints/tasks.py to website/identifiers/tasks.py
- Move the method for creating identifiers out of website/project/views/register.py

## QA Notes
Underlying code organization, no need for QA Testing
## Side Effects

None

## Ticket

https://openscience.atlassian.net/browse/OSF-8932


pattisdr:
Failures seem unrelated to my changes, moving to CR.

caseyrollins:
@erinspace, does it make sense for all of the identifiers code to be moved into `api/` or `osf/` in this PR (considering #8069/[OSF-9001](https://openscience.atlassian.net/browse/OSF-9001))?

erinspace:
@caseyrollins ah that's a good question! Technically that PR is just for website utils, not tasks, but indeed this move might make sense considering that is a precursor to getting rid of the `website` directory entirely. @sloria what d'you think?

sloria:
Yeah, we definitely want to move all the `tasks` modules to `osf` in the future, but I think it'd be better to do it all at once to retain consistency. So I'm good with the organization as is.

caseyrollins:
@sloria @erinspace, sounds good, thank you for clarifying! 

In that case, everything looks flawless to me @pattisdr.

sloria:
Merged into `release/next-next-djangole`.

PR #8069, opened 2017-12-18
Remove usages of website utilities outside of the website directory [OSF-9001]

erinspace:
## Purpose

This is in anticipation of a much larger refactor where the website directory is removed entirely. 

## Changes

Remove website utils that are being used outside of the website directory to a more central place that is using them

## QA Notes

This should not change any user-facing things, but there is a lot of potential for error as this touches so much code. This and #8095 are both essentially "move a bunch of things around and make sure the new places are referenced correctly." It'd be good to test this and #8095 at the same time, as they're very similar in scope.

## Side Effects

none anticipated, though this touches a LOT of imports, so nervous about that...

## Ticket
https://openscience.atlassian.net/browse/OSF-9001

PR #8070, opened 2017-12-19
[EOSF-878] Add 'move' link to quickfiles

baylee-d:
## Purpose

When files are uploaded, they are automatically given a move link through waterbutler.  However, when files are loaded they are missing this link.  This adds a move link to the quick file serializer.

## Changes

- Add move link to the quick files serializer

## QA Notes

Before this fix, users were only able to move files that had just been uploaded.  With these changes, users should be able to have pre-uploaded files that can now be moved.  To test:

1. Upload a file
2. Move to project or component
3. Test that it works

4. Upload another file
5. Refresh the page
6. Move file to project or component
7. Test that it works 

## Ticket

https://openscience.atlassian.net/browse/EOSF-878


jamescdavis:
@baylee-d Your test is going to need some revision when #8069 is merged in.

PR #8071, opened 2017-12-19
Hotfix/Exclude preregs not submitted for approval from admin prereg list

laurenrevere:
## Purpose

Preregistrations that were that were registered without being submitted were showing up on the admin preregistration challenge list. 

## Changes

Previous query was slightly wrong - it was excluding things that were both still drafts AND not submitted for the preregistration challenge. This changes the query so that it is excluding all registrations that have not been submitted for the prereg challenge.

## QA Notes

None

## Side Effects
Don't think so.

## Ticket

None


laurenrevere:
@brianjgeiger 👍

brianjgeiger:
This branch is failing in the way that the fix from Alex handled. There are two or three kinds of timeouts. Taking too long to run the whole branch is just a Travis thing. Taking 10 minutes to generate output is what Alex fixed. And there might be another for devops-y problems. I'm wondering if this isn't all the way up to date with Develop?

PR #8074, opened 2017-12-19
[hotfix] sitemap permissions and progress

icereval:


scripts/generate_sitemap.py
```diff
@@ -44,6 +47,10 @@ def __init__(self):
                 region_name='us-east-1'
             )
 
+    def __del__(self):
```
sloria: Using `__del__` is a bit counterintuitive. `del` is typically used to remove an object in memory, so doing IO is unexpected. Maybe rename to `clean()` or `cleanup()`.

That said, is this actually called anywhere?

PR #8077, opened 2017-12-20
[#OSF-9101]Error when subscribing to mailchimp: "Due to restrictions on your account, double opt-in is required for this list"

chennan47:
## Purpose
Add a script to sub users to OSF general mailchimp for whose users who created during the double-opt-in mailchimp error.

## Ticket
https://openscience.atlassian.net/browse/OSF-9101


scripts/fix_user_mailchimp.py
```diff
@@ -0,0 +1,43 @@
...
+        # If we're not running in dry mode log everything to a file
+        script_utils.add_file_logger(logger, __file__)
+    with transaction.atomic():
+        start_time = datetime.strptime('2017-12-20 08:25:25', '%Y-%m-%d %H:%M:%S')
+        start_time = start_time.replace(tzinfo=timezone.now().tzinfo)
+
+        end_time = datetime.strptime('2017-12-20 18:05:00', '%Y-%m-%d %H:%M:%S')
+        end_time = start_time.replace(tzinfo=timezone.now().tzinfo)
+
+        users = OSFUser.objects.filter(is_registered=True, date_disabled__isnull=True, date_registered__range=[start_time, end_time])
```
alexschiller: Do we know if any of the affected users were merges/changed primary emails?

scripts/fix_user_mailchimp.py
```diff
@@ -0,0 +1,43 @@
...
+
+        for user in users:
+            subscribe_mailchimp(settings.MAILCHIMP_GENERAL_LIST, user._id)
```
sloria: This part should be skipped in dry mode.

alexschiller: Should this have a check to make sure they are listed as subscribed `False` e.g., `user.mailchimp_mailing_lists.get(settings.MAILCHIMP_GENERAL_LIST, False) is False:` 

Not 100% it is needed, but we don't want to send to anyone accidentally.

scripts/fix_user_mailchimp.py
```diff
@@ -0,0 +1,43 @@
...
+            logger.info('User {} has been subscribed to OSF general mailing list'.format(user._id))
+
+        logger.info('{} users have been subscribed to OSF general mailing list'.format(users.count()))
+
+        if dry:
```
alexschiller: Since this is hitting an external endpoint dry will just make us not save the data, but the requests will still go out. Prefer if there is no dry option to prevent bad things.

scripts/fix_user_mailchimp.py
```diff
@@ -0,0 +1,43 @@
...
+        end_time = datetime.strptime('2017-12-20 18:05:00', '%Y-%m-%d %H:%M:%S')
+        end_time = start_time.replace(tzinfo=timezone.now().tzinfo)
+
+        users = OSFUser.objects.filter(is_registered=True, date_disabled__isnull=True, date_registered__range=[start_time, end_time])
+
+        for user in users:
+            subscribe_mailchimp(settings.MAILCHIMP_GENERAL_LIST, user._id)
+            logger.info('User {} has been subscribed to OSF general mailing list'.format(user._id))
+
+        logger.info('{} users have been subscribed to OSF general mailing list'.format(users.count()))
```
alexschiller: Should this count match the number of sentry errors? Do we need to do further investigations if they don't match?

scripts/fix_user_mailchimp.py
```diff
@@ -0,0 +1,43 @@
...
+    dry = '--dry' in sys.argv
+    if not dry:
+        # If we're not running in dry mode log everything to a file
+        script_utils.add_file_logger(logger, __file__)
+    with transaction.atomic():
+        start_time = datetime.strptime('2017-12-20 08:25:25', '%Y-%m-%d %H:%M:%S')
+        start_time = start_time.replace(tzinfo=timezone.now().tzinfo)
+
+        end_time = datetime.strptime('2017-12-20 18:05:00', '%Y-%m-%d %H:%M:%S')
+        end_time = start_time.replace(tzinfo=timezone.now().tzinfo)
```
alexschiller: end_time is redefined as start_time.replace. Looks like it should be `end_time.replace`

chennan47:
closed in favor of https://github.com/CenterForOpenScience/osf.io/pull/8078

PR #8078, opened 2017-12-20
[#OSF-9101]Error when subscribing to mailchimp: "Due to restrictions on your account, double opt-in is required for this list"

chennan47:
## Purpose
Add a script to sub users to OSF general mailchimp for whose users who created during the double-opt-in mailchimp error.

## Ticket
https://openscience.atlassian.net/browse/OSF-9101

### DIFF
--- a/scripts/fix_user_mailchimp.py
+++ b/scripts/fix_user_mailchimp.py
@@ -1,0 +1,45 @@
+import logging
+import sys
+from datetime import datetime
+
+from django.db import transaction
+from django.utils import timezone
+
+from website.app import setup_django
+setup_django()
+from osf.models import OSFUser
+from scripts import utils as script_utils
+from website.mailchimp_utils import subscribe_mailchimp
+from website import settings
+
+logger = logging.getLogger(__name__)
+
+
+def main():
+    dry = '--dry' in sys.argv
+    if not dry:
+        # If we're not running in dry mode log everything to a file
+        script_utils.add_file_logger(logger, __file__)
+
+    with transaction.atomic():
+        start_time = datetime.strptime('2017-12-20 08:25:25', '%Y-%m-%d %H:%M:%S')
+        start_time = start_time.replace(tzinfo=timezone.now().tzinfo)
+
+        end_time = datetime.strptime('2017-12-20 18:05:00', '%Y-%m-%d %H:%M:%S')
+        end_time = end_time.replace(tzinfo=timezone.now().tzinfo)
+
+        users = OSFUser.objects.filter(is_registered=True, date_disabled__isnull=True, date_registered__range=[start_time, end_time])
+
+        if not dry:
+            for user in users:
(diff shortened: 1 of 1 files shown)


scripts/fix_user_mailchimp.py
```diff
@@ -0,0 +1,45 @@
...
+        end_time = end_time.replace(tzinfo=timezone.now().tzinfo)
+
+        users = OSFUser.objects.filter(is_registered=True, date_disabled__isnull=True, date_registered__range=[start_time, end_time])
+
+        if not dry:
+            for user in users:
+                subscribe_mailchimp(settings.MAILCHIMP_GENERAL_LIST, user._id)
+                logger.info('User {} has been subscribed to OSF general mailing list'.format(user._id))
+
+            logger.info('{} users have been subscribed to OSF general mailing list'.format(users.count()))
```
alexschiller: Worth moving this out of the `if not dry` block so that you can actually see the count of users that would be subscribed even if dry.

scripts/fix_user_mailchimp.py
```diff
@@ -0,0 +1,45 @@
...
+        start_time = datetime.strptime('2017-12-20 08:25:25', '%Y-%m-%d %H:%M:%S')
+        start_time = start_time.replace(tzinfo=timezone.now().tzinfo)
+
+        end_time = datetime.strptime('2017-12-20 18:05:00', '%Y-%m-%d %H:%M:%S')
+        end_time = end_time.replace(tzinfo=timezone.now().tzinfo)
+
+        users = OSFUser.objects.filter(is_registered=True, date_disabled__isnull=True, date_registered__range=[start_time, end_time])
+
+        if not dry:
+            for user in users:
```
sloria: This whole block shouldn't be in this `if` block; we still want to see the logging when running in dry mode. Only the `subscribe_mailchimp` call should be skipped in dry mode.

chennan47: fixed in f122d8a

scripts/fix_user_mailchimp.py
chennan47: fixed in f122d8a

scripts/fix_user_mailchimp.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def main():
+    dry = '--dry' in sys.argv
+    if not dry:
+        # If we're not running in dry mode log everything to a file
+        script_utils.add_file_logger(logger, __file__)
+
+    with transaction.atomic():
+        start_time = datetime.strptime('2017-12-20 08:25:25', '%Y-%m-%d %H:%M:%S')
```
sloria: Minor: It's preferable to build a datetime directly,

```
import pytz
from datetime import datetime

start_time = datetime(2017, 12, 20, 8, 25, 25, tzinfo=pytz.UTC)
```

chennan47: fixed in 5b11b1f

PR #8086, opened 2018-01-03
Hotfix/fix copyright date[OSF-9024]

laurenrevere:
## Purpose

It's 2018 - let's have our copyright date show that. While we're at it let's make copyright dates change automatically with the years (like we do on the ember side).

## Changes
Generate the copyright date based on the actual time for the footer.

## QA Notes
Does the copyright date on the bottom of the OSF say 2018?

## Tests
Because this is a change in the mako it doesn't need a test.

## Side Effects

None expected

## Ticket

https://openscience.atlassian.net/browse/OSF-9024

  
screen shot 2018-01-03 at 1 37 20 pm

### DIFF
--- a/website/templates/copyright.mako
+++ b/website/templates/copyright.mako
@@ -1,7 +1,9 @@
+<% from datetime import datetime %>
+
 <div class="container copyright">
     <div class="row">
         <div class="col-md-12">
-            <p>Copyright &copy; 2011-2017 <a href="https://cos.io/">Center for Open Science</a> |
+            <p>Copyright &copy; 2011-${datetime.utcnow().year} <a href="https://cos.io/">Center for Open Science</a> |
                 <a href="https://github.com/CenterForOpenScience/centerforopenscience.org/blob/master/TERMS_OF_USE.md">Terms of Use</a> |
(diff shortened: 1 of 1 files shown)


website/templates/copyright.mako
```diff
@@ -1,7 +1,9 @@
+<% from datetime import datetime %>
+
 <div class="container copyright">
     <div class="row">
         <div class="col-md-12">
-            <p>Copyright &copy; 2011-2017 <a href="https://cos.io/">Center for Open Science</a> |
+            <p>Copyright &copy; 2011-${datetime.now().year} <a href="https://cos.io/">Center for Open Science</a> |
```
sloria: We should use `utcnow` so that we're not relying on the server's timezone. I can change this when I merge.

laurenrevere:
Ah okay. Thanks @sloria.

website/templates/copyright.mako
laurenrevere: Ah okay. Thanks, @sloria.

sloria:
Woops; looks like we made the same change. I already changed to utcnow() when I merged. 🚢 🇮🇹

PR #8088, opened 2018-01-03
Hotfix/remove ARKs from user-facing locations[OSF-9118]

laurenrevere:
## Purpose

The things we say are ARKs are not really ARKs and we should no longer show them to users. 

## Changes

This removes showing ARKs and references to ARKs in user-facing locations. References to ARKs were removed from the project detail page and the `/identifiers/`, `/node/<node-id>/identifiers/`, `/registrations/<registration-id>/identifiers/` and `/preprints/<preprint-id/identifiers/` api endpoints.

## QA Notes
ARKs should no longer show up on the project detail pages or at the `nodes/<nid>/identifiers/` endpoint. References that used to be "DOI/ARK" in these locations should just say DOI now. Mentions of 'identifiers' in these cases should be singular now.

## Side Effects

Note: This only changes the front end visibility of ARKs the fake ones we have are still being stored and used to check if identifiers exist (which is not ideal). This should be removed when we change how we get DOIs. 

## Ticket

https://openscience.atlassian.net/browse/OSF-9118

screen shot 2018-01-03 at 3 32 49 pm
screen shot 2018-01-03 at 3 47 33 pm
screen shot 2018-01-03 at 3 47 50 pm
screen shot 2018-01-03 at 4 01 31 pm

### DIFF
--- a/api/identifiers/views.py
+++ b/api/identifiers/views.py
@@ -26,7 +26,7 @@ class IdentifierList(JSONAPIBaseView, generics.ListAPIView, ListFilterMixin):
 
         name           type                   description
         ----------------------------------------------------------------------------
-        category       string                 e.g. 'ark', 'doi'
+        category       string                 e.g. 'doi'
         value          string                 the identifier value itself
 
     ##Links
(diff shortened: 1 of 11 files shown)


api/identifiers/views.py
```diff
@@ -71,7 +71,7 @@ def get_object(self, *args, **kwargs):
 
     def get_default_queryset(self):
         obj = self.get_object()
-        return obj.identifiers.all()
+        return obj.identifiers.all().exclude(category='ark')
```
alexschiller: Driveby: Pretty sure the `.all()` doesn't do anything if you're excluding

laurenrevere: ooh good to know. Thanks!

PR #8089, opened 2018-01-04
[feature] Allow updates of all auto_now fields on all models to be disabled [OSF-9081]

caseyrollins:
#### Purpose
- Adds a precaution to ensure auto_now fields aren't updated in migrations. A dev may forget to pass all models that would get automatically updated by auto_now.

#### Changes
- Allows `disable_auto_now_fields` to take no arguments. If no arguments are passed, **all** auto_now fields on **all** models in the `osf` and `addons` apps will be disabled.

#### QA Notes
- None required/possible.

#### Side Effects
- None expected.

#### Ticket
- [OSF-9081](https://openscience.atlassian.net/browse/OSF-9081)

### DIFF
--- a/osf/utils/migrations.py
+++ b/osf/utils/migrations.py
@@ -13,15 +13,26 @@ from website.project.metadata.schemas import OSF_META_SCHEMAS
 logger = logging.getLogger(__file__)
 
 
-@contextmanager
-def disable_auto_now_fields(model=None):
+def get_osf_models():
     """
-    Context manager to disable updates of all auto_now fields for a given model.
-    If model=None, updates for all auto_now fields on all models will be disabled.
+    Helper function to retrieve all osf related models.
 
+    Example usage:
+        with disable_auto_now_fields(models=get_osf_models()):
+            ...
     """
-    all_app_models = list(itertools.chain(*[app.get_models() for app in apps.get_app_configs() if 'addons' in app.label or 'osf' in app.label]))
-    models = [model] if model else all_app_models
+    return list(itertools.chain(*[app.get_models() for app in apps.get_app_configs() if app.label.startswith('addons_') or app.label.startswith('osf')]))
+
+@contextmanager
+def disable_auto_now_fields(models=None):
+    """
+    Context manager to disable auto_now field updates.
+    If models=None, updates for all auto_now fields on *all* models will be disabled.
+
+    :param list models: Optional list of models for which auto_now field updates should be disabled.
+    """
+    if not models:
+        models = apps.get_models()
 
     changed = []
     for model in models:
--- a/osf/utils/migrations.py
+++ b/osf/utils/migrations.py
@@ -1,8 +1,10 @@
 import os
+import itertools
 import json
 import logging
 
 from contextlib import contextmanager
+from django.apps import apps
 
 from website import settings
 from osf.models import NodeLicense, MetaSchema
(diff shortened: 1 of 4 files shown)


osf/utils/migrations.py
```diff
@@ -12,16 +14,21 @@
...
 
 @contextmanager
-def disable_auto_now_fields(model):
+def disable_auto_now_fields(model=None):
     """
     Context manager to disable updates of all auto_now fields for a given model.
+    If model=None, updates for all auto_now fields on all models will be disabled.
 
     """
+    all_app_models = list(itertools.chain(*[app.get_models() for app in apps.get_app_configs() if 'addons' in app.label or 'osf' in app.label]))
```
sloria: Is there a reason to filter out non-osf models here?

caseyrollins: Not necessarily, I was just uncertain about if non-osf models should be included here. The models excluded here are:
```
 django.contrib.auth.models.Permission,
 django.contrib.auth.models.Group,
 django.contrib.contenttypes.models.ContentType,
 django.contrib.sessions.models.Session,
 django.contrib.admin.models.LogEntry,
 django_celery_beat.models.SolarSchedule,
 django_celery_beat.models.IntervalSchedule,
 django_celery_beat.models.CrontabSchedule,
 django_celery_beat.models.PeriodicTasks,
 django_celery_beat.models.PeriodicTask,
 corsheaders.models.CorsModel,
 guardian.models.UserObjectPermission,
 guardian.models.GroupObjectPermission,
```

caseyrollins: If those should be included, then this line could just be
```
all_models = apps.get_models()
```

sloria: I'm just considering how we can make this code decoupled from osf code.

Additionally, it might make sense to pass a list of models rather than a single model. That way, we could always do

```python
with disable_auto_now_fields(models=get_osf_models()):
```

if we ever wanted that behavior.

caseyrollins: +1 to a list of models -- I thought about that while working on this ticket. 

I can change the `model` kwarg to `models`, add an `get_osf_models` helper, and the context manager can disable *all* model auto_now fields if `models` isn't specified. 

```
def get_osf_models():
    return list(itertools.chain(*[app.get_models() for app in apps.get_app_configs() if 'addons' in app.label or 'osf' in app.label]))
```
```
def disable_auto_now_fields(models=[]):
    if not models:
        models = apps.get_models()
    ... 
```
```
# most common OSF usage, probably 
with disable_auto_now_fields(models=get_osf_models()):
```
Thoughts?

sloria: Yep, sounds good to me. Just make sure to do

```python
def disable_auto_now_fields(models=None):
```

instead of using a list as the default.

osf/utils/migrations.py
```diff
@@ -11,17 +13,33 @@
...
 
+def get_osf_models():
+    """
+    Helper function to retrieve all osf related models.
+
+    Example usage:
+        with disable_auto_now_fields(models=get_osf_models()):
+            ...
+    """
+    return list(itertools.chain(*[app.get_models() for app in apps.get_app_configs() if 'addons' in app.label or 'osf' in app.label]))
```
sloria: The check could be more precise, e.g, `app.label.startswith('addons_')`, so we don't get bit by something like `django_model_addons`.

sloria:

PR #8093, opened 2018-01-05
Feature/Ignore download_file and download_zip log actions [OSF-8785]

laurenrevere:
## Purpose
We want Waterbutler to be able to send download information to `create_waterbutler_log` without raising an error or actually creating log data.

## Changes
This allows `create_waterbutler_log` to accept `download_file` and `download_zip` actions. Currently it just ignores them (This allows waterbutler to send us the data without returning 400s. We will use this data in OSF-8786 to tally downloads).

## QA Notes
When a preprint is downloaded, nothing should be added to its node log.

## Side Effects
None expected

## Ticket
https://openscience.atlassian.net/browse/OSF-8785

### DIFF
--- a/addons/base/views.py
+++ b/addons/base/views.py
@@ -314,6 +314,9 @@ def create_waterbutler_log(payload, **kwargs):
     with transaction.atomic():
         try:
             auth = payload['auth']
+            # Don't log download actions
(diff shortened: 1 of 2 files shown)


felliott:
LGTM!  :+1:

PR #8095, opened 2018-01-08
Move website settings used outside of website [OSF-9000]

erinspace:
## Purpose
In anticipation of deleting the website directory entirely, clean up the code a little by moving settings used in places outside of the website directory to the place where it's used, or a central location.

## Changes

- Move settings out of website settings where appropriate
- Change imports of those settings

does not include:
- any path-related settings that rely on relative path such as `BASE_PATH` and `ROOT`

## QA Notes

Very similar to #8069, this should not change any user-facing things, but there is a lot of potential for error as this touches so much code. Would be really useful to test both of these at the same time, as they're both essentially "move a bunch of things around and make sure the new places are referenced correctly." If there is a full site regression happening at some point, this and #8069 will be best tested by that.

## Side Effects

None anticipated

## Ticket
https://openscience.atlassian.net/browse/OSF-9000


api/base/settings/defaults.py
```diff
@@ -39,33 +46,49 @@
...
+# TODO: Override SECRET_KEY in local.py in production
+SECRET_KEY = 'CHANGEME'
+SESSION_COOKIE_SECURE = SECURE_MODE
+SESSION_COOKIE_HTTPONLY = True
+
 # session:
 SESSION_COOKIE_NAME = 'api'
-SESSION_COOKIE_SECURE = osf_settings.SECURE_MODE
-SESSION_COOKIE_HTTPONLY = osf_settings.SESSION_COOKIE_HTTPONLY
+SESSION_COOKIE_SECURE = SECURE_MODE
```
chennan47: duplicate as line https://github.com/CenterForOpenScience/osf.io/pull/8095/files#diff-604cc8ea21ab8c06c9850dbe4d3f6d96R73

website/settings/__init__.py
```diff
@@ -17,9 +17,9 @@
...
-    from . import local
-    from . import defaults
-    for setting in ('WATERBUTLER_JWE_SECRET', 'WATERBUTLER_JWE_SALT', 'WATERBUTLER_JWT_SECRET', 'JWT_SECRET', 'DEFAULT_HMAC_SECRET', 'POPULAR_LINKS_NODE', 'NEW_AND_NOTEWORTHY_LINKS_NODE', 'SENSITIVE_DATA_SALT', 'SENSITIVE_DATA_SECRET'):
-        assert getattr(local, setting, None) and getattr(local, setting, None) != getattr(defaults, setting, None), '{} must be specified in local.py when DEV_MODE is False'.format(setting)
+#
+# if not DEV_MODE:
+#     from . import local
+#     from . import defaults
+#     for setting in ('WATERBUTLER_JWE_SECRET', 'WATERBUTLER_JWE_SALT', 'WATERBUTLER_JWT_SECRET', 'JWT_SECRET', 'DEFAULT_HMAC_SECRET', 'POPULAR_LINKS_NODE', 'NEW_AND_NOTEWORTHY_LINKS_NODE', 'SENSITIVE_DATA_SALT', 'SENSITIVE_DATA_SECRET'):
+#         assert getattr(local, setting, None) and getattr(local, setting, None) != getattr(defaults, setting, None), '{} must be specified in local.py when DEV_MODE is False'.format(setting)
```
chennan47: Are those commented-out codes supposed to be removed?

api/base/settings/defaults.py
```diff
@@ -269,3 +325,1663 @@
...
+# akismet spam check
+AKISMET_APIKEY = None
+SPAM_CHECK_ENABLED = False
+SPAM_CHECK_PUBLIC_ONLY = True
+SPAM_ACCOUNT_SUSPENSION_ENABLED = False
+SPAM_ACCOUNT_SUSPENSION_THRESHOLD = datetime.timedelta(hours=24)
+SPAM_FLAGGED_MAKE_NODE_PRIVATE = False
+SPAM_FLAGGED_REMOVE_FROM_SEARCH = False
+
+ENABLE_ARCHIVER = True
```
chennan47: It will be better to move this into the ARCHIVER blog in line 1906

chennan47:
Code review done. Only some minor problems.

chennan47:
Pass done. Look good.

addons/bitbucket/models.py
```diff
@@ -15,9 +16,10 @@
 from framework.auth import Auth
 from osf.models.external import ExternalProvider
 from osf.models.files import File, Folder, BaseFileNode
-from website import settings
 from website.util import web_url_for
 
+settings = importlib.import_module('api.base.settings')
```
sloria: Why is this necessary?

erinspace: I don't remember the exact error, but this was to get around a circular import when importing node addon settings I believe. I'll get more details!

sloria: Discussed on FD: using a normal import does cause a circular import.

We haven't figured out why using `import_module` fixes it, but it's a fine workaround for the time being.

website/settings/defaults.py
```diff
@@ -129,64 +68,18 @@ def parent_dir(path):
...
 MAILGUN_API_KEY = None
 
-# Use Celery for file rendering
-USE_CELERY = True
-
 # File rendering timeout (in ms)
 MFR_TIMEOUT = 30000
 
-# TODO: Override in local.py in production
-DB_HOST = 'localhost'
```
sloria: To avoid confusion with the app's Postgres settings, we should rename these to something like `SHAREJS_DB_HOST`, etc.

I've made a follow-up ticket: https://openscience.atlassian.net/browse/PLAT-676

website/settings/defaults.py
```diff
@@ -283,31 +132,8 @@ def parent_dir(path):
...
-    },
-    'private': {
-        'project_id': '',
-        'write_key': '',
-        'read_key': '',
-    },
-}
-
-SENTRY_DSN = None
-SENTRY_DSN_JS = None
```
sloria: I don't see this in the api settings. @erinspace  Should this have been moved, too?

website/settings/defaults.py
```diff
@@ -328,1587 +154,15 @@ def parent_dir(path):
...
-ARK_NAMESPACE = 'ark:99999/fk4'
-
-# For creating DOIs and ARKs through the EZID service
-EZID_USERNAME = None
-EZID_PASSWORD = None
-# Format for DOIs and ARKs
-EZID_FORMAT = '{namespace}osf.io/{guid}'
-
-# Leave as `None` for production, test/staging/local envs must set
-SHARE_PREPRINT_PROVIDER_PREPEND = None
```
sloria: This also doesn't appear to be defined. Should this go into admin settings? @erinspace

erinspace: Moved to admin settings!

website/settings/defaults.py
erinspace: Moved to api base defaults!

sloria:
This got automatically closed because I removed `release/next-platform`. @erinspace Can you resend this to `develop` please?

PR #8097, opened 2018-01-09
[#OSF-8990]Top level project have an 'Admins on Parent Projects' section

chennan47:
## Purpose
Fix the 'admin on parents' table on contributor tab of projects

## Changes
1. Change the serialize_contributor function to handle the case when it is serialized for admin_contributor. Remove the return of the contributor on both the current project and its parent. 2. Improve the way to calculate the visibility of contributor on parent project in serialize_user function.
Before the change:
screen shot 2018-01-09 at 3 48 11 pm
After the change:
screen shot 2018-01-09 at 3 48 36 pm

## QA Notes
Test setup:
1. Create a project A with one child node B and one grandchild node C
2. add a new admin permission visible contributor X to node A
3. add a new admin permission invisible contributor Y to node B
Check:
1. go to the contributor tab of project A
expected: no 'admin on parents' table show up in that page
2. go to the contributor tab of project B
expected: 'Admin on parents' table shows up in that page with only one contributor X in it and X's 'Bibliographic Contributor' checkbox is checked
3. go to the contributor tab of project C
expected:  'Admin on parents' table shows up in that page with two contributors in it. Contributor X's 'Bibliographic Contributor' checkbox is checked and contributor Y's 'Bibliographic Contributor' checkbox is  NOT checked.

## Ticket
https://openscience.atlassian.net/browse/OSF-8990

### DIFF
--- a/osf/models/node.py
+++ b/osf/models/node.py
@@ -903,9 +903,31 @@ class AbstractNode(DirtyFieldsMixin, TypedModel, AddonModelMixin, IdentifierMixi
         return admin_ids
 
     @property
+    def parent_admin_contributor_ids(self):
+        def get_admin_contributor_ids(node):
+            return Contributor.objects.select_related('user').filter(
+                node=node,
+                user__is_active=True,
+                admin=True
+            ).values_list('user__guids___id', flat=True)
+
+        contributor_ids = set(self.contributors.values_list('guids___id', flat=True))
+        admin_ids = set()
+        for parent in self.parents:
+            admins = get_admin_contributor_ids(parent)
+            admin_ids.update(set(admins).difference(contributor_ids))
+        return admin_ids
+
+    @property
     def admin_contributors(self):
         return OSFUser.objects.filter(
             guids___id__in=self.admin_contributor_ids
+        ).order_by('family_name')
+
+    @property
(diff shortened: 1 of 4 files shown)


pattisdr:
@chennan47 and whoever reviews this PR, I added a [similar commit](https://github.com/CenterForOpenScience/osf.io/pull/7831/commits/22b8803ba4be726b72f852dc4d5b500bae1a7b11), in https://github.com/CenterForOpenScience/osf.io/pull/7831, which is the *Add Admins from Parent to Child* feature.   I will probably end up removing my commit in favor of the work here, because this ticket is specifically for that bug, and mine is for a different feature that just needs this bug fixed.

One comment though, there was some [discussion about the confusing names for things](https://www.flowdock.com/app/cos/django-le/threads/rzC9ow21TnkvxkzBDrl_lWbgYrr).  When we talk about an admin contributor, sometimes we are referring to admins of the current project, and other times we are referring to admins on the parent project.  We often use similar variable names to describe both which probably caused this bug in the first place.  My recommendation is to call admins on parent project something like `parent_admin_contributor`.

caseyrollins:
Closing in favor of #7831

PR #8098, opened 2018-01-10
Feature/banner creator[OSF-8957]

laurenrevere:
## Purpose

Banners are time consuming and error-prone to add to the OSF manually. It communication from the community, ember, and and osf teams. Let's add this process to the admin app!

## Changes

- Create a `ScheduledBanner` model (so we can store banner information in the database)
- Add a banner creator in the admin app
- Add the api endpoint `v2/banners/current` to surface the current banner information
- Use `banners/current` to display banners on the osf home page and meetings landing page (it should not be painful to add banners to the preprints landing page)

## QA Notes

There are a lot of things. For now: do banners work?

## Side Effects

Currently, the images are stored in the database, this shouldn't be too bad, but is something to note. There is also a separate model for `BannerImage`s, there could be an implementation where the binary data is just in a row in the `ScheduledBanner` table. It is also little strange because it is actually defined within `storage.py` instead of in the models folder. 

## Ticket

https://openscience.atlassian.net/browse/OSF-8957


admin/static/css/banners.css
```diff
@@ -0,0 +1,5 @@
+.banner-background-color {
+    margin-top: 10px;
+    padding-left: 5px;
+    padding-right: 5px;
+}
```
erinspace: So when I was trying this out, I grabbed a random SVG and popped it in there and it was waaaaaay too big - perhaps adding a `max-height` style property to the image component would help prevent people from making silly mistakes like I did?

screen shot 2018-01-23 at 12 19 55 pm

admin_tests/banners/test_views.py
```diff
@@ -0,0 +1,120 @@
...
+        assert today_banner in banners
+        assert tomorrow_banner in banners
+
+    def test_context_data(self, view, today_banner, tomorrow_banner):
+        view.object_list = view.get_queryset()
+        data = view.get_context_data()
+        assert type(data) is dict
+        assert (len(data['banners']), 2)
+        assert type(data['banners'][0]) is ScheduledBanner
+
```
erinspace: Might wanna add tests to make sure only users who have the right permissions can view banners

admin_tests/banners/test_views.py
```diff
@@ -0,0 +1,120 @@
...
+    def test_delete(self, view, req, app):
+        res = view.delete(req)
+        assert res.url == '/banners/'
+        assert res.status_code == 302
+
+    def test_get(self, view, req):
+        res = view.get(req)
+        assert res.status_code == 200
+
```
erinspace: Same as above - add tests to make sure only users with right permissions can create/delete a banner

api/banners/serializers.py
```diff
@@ -0,0 +1,49 @@
...
+    def get_default_text(self, banner):
+        return self.add_license(banner, banner.default_text)
+
+    def get_mobile_text(self, banner):
+        if banner.mobile_text:
+            return self.add_license(banner, banner.mobile_text)
+        return self.get_default_text(banner)
+
+    def add_license(self, banner, text):
+        if banner.license and not banner.license.lower() == 'none':
```
erinspace: Having the string 'none' be a special case seems tricky - can you just make the license field on the form not required instead of having the user typing 'none' result in no license text? If this was specifically requested all good!

api_tests/banners/views/test_current_banner.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @pytest.fixture()
+    def data(self, res):
+        return res.json['data']
+
+    def test_banner_detail(self, banner, res, data):
+        assert res.status_code == 200
+        assert res.content_type == 'application/vnd.api+json'
+        assert data['type'] == 'banners'
+        # assert data['id'] == banner._id
```
erinspace: np: remove comment please!

api_tests/banners/views/test_current_banner.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # When there's no current banner endpoint returns dummy ScheduledBanner with no data
+    def test_no_current_banner(self, tomorrow_banner, yesterday_banner, data):
+        assert not data.get('start_date')
+
+    def test_no_banners(self, data):
```
erinspace: this seems like a duplicate of the above `test_no_current_banner` - can probably be removed

osf/migrations/0077_bannerimage_scheduledbanner.py
```diff
@@ -0,0 +1,43 @@
...
+                ('default_photo', models.FileField(storage=osf.utils.storage.BannerImageStorage(), upload_to=b'')),
+                ('default_text', models.TextField()),
+                ('mobile_photo', models.FileField(storage=osf.utils.storage.BannerImageStorage(), upload_to=b'')),
+                ('mobile_text', models.TextField(blank=True, null=True)),
+            ],
+            options={
+                'permissions': (('view_banner', 'Can view banner details'), ('change_banner', 'Can change banner details'), ('delete_banner', 'Can delete banner')),
+            },
+        ),
+    ]
```
erinspace: In addition to this migration, can you add a migration to add the new permissions to the existing permission groups? I think the appropriate group names would be `read_only` and `osf_admin`

api/banners/serializers.py
laurenrevere: You can leave it empty as it was not required, just as I was making banners I found myself accidentally typing 'none' in the license box. I added this as a nicety in case someone else accidentally did that, but it can totally be removed.

osf_tests/factories.py
```diff
@@ -815,3 +815,18 @@ class Meta:
...
+    name = factory.Faker('name')
+    default_text = factory.Faker('text')
+    mobile_text = factory.Faker('text')
+    default_photo = factory.Faker('file_name')
+    mobile_photo = factory.Faker('file_name')
+    license = factory.Faker('name')
+    color = 'white'
+    start_date = timezone.now()
+    end_date = start_date
+    #TODO: Make setting dates smarter
```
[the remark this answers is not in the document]
laurenrevere: This should work that way in general because start_date starts at midnight and end_date ends at 11:59 if they are the same it should be a default 1 day banner.

osf/utils/storage.py
```diff
@@ -0,0 +1,33 @@
...
+        return ContentFile(icon.image)
+
+    def _save(self, name, content):
+        BannerImage.objects.update_or_create(filename=name, defaults={'image': content.read()})
+        return name
+
+    def delete(self, name):
+        BannerImage.objects.get(filename=name).delete()
+
+    def get_available_name(self, name, max_length=None):
```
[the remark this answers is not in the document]
laurenrevere: Get the default `get_available_name` looks for available names on the os, but these files are being stored in the database.

 A thing to note here is that if you name two images the same thing the older one will be overwritten. It could be a good move to make this more like the original `get_available_name` which appends random numbers to the end of filenames if there are collisions. (Should I add this?) One reason not to would be that if we happen to use the same mobile banner for a whole campaign we wouldn't want it stored 5x (but that is unlikely).

admin/static/css/banners.css
laurenrevere: That's a good point. I think I'm going to add validation on the uploader with max height/width so that we aren't storing images that are bigger than we need.

api/banners/serializers.py
erinspace: Naw makes sense! Woo!

admin/static/css/banners.css
erinspace: Ah, but you can still restrict its height using CSS -- I think this is actually pretty important just to prevent any accidental things -- something like setting the `max-height` property in `banners.css` would do the trick

screen shot 2018-01-29 at 12 13 38 pm

admin_tests/banners/test_views.py
erinspace: 🎉

erinspace:
Looks great! It's so big and involves a lot of changes, so another person should take a look too, but it's solid 🚀

admin/templates/banners/confirm_delete.html
```diff
@@ -0,0 +1,24 @@
+{% extends "base.html" %}
+{% load static %}
```
sloria: Nit: I don't think these template helpers need to be loaded.

website/static/js/home-page/donateBannerPlugin.js
```diff
@@ -1,128 +1,66 @@
...
-        },
-    }
-];
+        // Load banner
+        var bannerUrl = $osf.apiV2Url('banners/current/');
+        var bannerPromise = m.request({method: 'GET', url: bannerUrl}, background=true);
+        bannerPromise.then(function(result){
+            self.banner(result.data);
+            self.bannerLoaded(true);
+        });
```
sloria: Error responses need to be handled here.

sloria:
Conflicts resolved and updates made. Just waiting for travis now: https://travis-ci.org/sloria/osf.io/builds/335811692 . Once green, will merge.

sloria:
Merged into `release/next-platform`.

PR #8103, opened 2018-01-11
[feature] bapi 502s on deleted files [OSF-8963]

fabmiz:
## Purpose

Fix issues where browsable API 502s on deleted files.

For files, we are determining the serializer_class dynamically (can be either `FileDetailSerializer` or `QuickFilesDetailSerializer`) by looking up the _id or guid from the URL. Sometimes, the file has already been deleted in which case, NotFound (_id) or Gone(guid) is raised. With the current version of DRF, it seems, any exceptions raised in `get_serializer_class` results in 502s. One of the solutions; we should not raise any errors when `get_file` is called from the serializer_class in case the object is found (does not matter if `is_deleted` is `True`. In case NotFound is raised just default to FileDetailSerializer. https://github.com/CenterForOpenScience/osf.io/blob/2881a189684f4e1389fe30633b59a36ee2f31695/api/files/views.py#L37-L50

## Changes
Add `ser` bool flag to tell when get_file is called from `get_serializer_class` or `get_object`. in the first case, we only expect an object and we handle NotFound by defaulting to `FileDetailSerializer`.

## QA Notes
- This fix pertains to the browsable API, hence API testing is required.
- Make sure for deleted file: url of the sorts https://staging-api.osf.io/v2/files/4r523/ and https://staging-api.osf.io/v2/files/5a56835b08a8d80206c787e5/ do not 502.
- Try the same urls in postman or any other similar API testing tool.

## Side Effects

NA

## Ticket
[OSF-8963](https://openscience.atlassian.net/browse/OSF-8963)


api/files/views.py
```diff
@@ -34,11 +34,13 @@ class FileMixin(object):
     serializer_class = FileSerializer
     file_lookup_url_kwarg = 'file_id'
 
-    def get_file(self, check_permissions=True):
+    def get_file(self, check_permissions=True, ser=False):
```
caseyrollins: A more descriptive argument name here (not an abbreviation) might be better. At first glance I would assume `ser` is short for `serializer`, but that makes me think that a serializer class would be passed, not a boolean value.

api_tests/files/views/test_file_detail.py
```diff
@@ -60,19 +60,25 @@ def test_must_have_auth_and_be_contributor(self, app, file_url):
...
         res = app.get(url, auth=user.auth)
         assert res.status_code == 200
 
         deleted_file.delete(user=user, save=True)
         res = app.get(url, auth=user.auth, expect_errors=True)
         assert res.status_code == 410
 
+        url = '/{}files/{}/'.format(API_BASE, file_id)
+        res = app.get(url, auth=user.auth, expect_errors=True)
+        assert res.status_code == 404
```
caseyrollins: Should this 410 instead?

api/files/views.py
```diff
@@ -34,15 +34,23 @@ class FileMixin(object):
     serializer_class = FileSerializer
     file_lookup_url_kwarg = 'file_id'
 
-    def get_file(self, check_permissions=True):
+    def get_file_obj(self, only_404=True):
         try:
-            obj = utils.get_object_or_error(BaseFileNode, self.kwargs[self.file_lookup_url_kwarg], self.request)
-        except (NotFound, Gone):
```
caseyrollins: @fabmiz, I think this entire fix might be as simple as not excepting `Gone` here so that it's actually raised. I really don't think `get_object_or_error` or `get_serializer_class` should need to be changed for this fix. 

Can you reevaluate and see if you agree?

fabmiz: @caseyrollins, the reason why we need these changes is `get_serializer_class`. It needs the node object to determine if it returns `QuickFilesDetailSerializer` or `FileDetailSerializer`. Now, if we raise an error, be it `404` or `410`, in any call from within `get_serializer_class`, it translates to `502` (seems to be an issue with bapi/drf), which was the original issue. In the case of `404`, it does not really matter since there is the object does not exist, so we can return whatever serializer or the default. However in the case of `410`, `get_serializer_class` can get the object and return the right serializer. This is my current understanding. 

However, we could not except `410` in `get_file` as you suggest, except both errors and return a serializer we choose, since it does really matter.

PR #8107, opened 2018-01-17
Feature/persistent file download links[OSF-8982]

laurenrevere:
## Purpose

Currently the API surfaces waterbutler links for file downloads, we want to add persistent download links.

## Changes

- File download link `www.osf.io/download/<file_path_or_guid>/` added. 
(It currently redirects to the waterbutler link)
- Replaced the waterbutler links in the file serializer with the above url.
- Added the above link with a query parameter for version to the `/v2/files/<file_path_or_guid>/versions/` endpoint.

## QA Notes

- Add files to both public/private projects in order to check if the download links maintain permissions.
- Have files that have both have a guid and some that do not.
- Create multiple versions of a file and check the version links and default link work as expected. (The plain download link should download the latest version).

## Side Effects

None expected

## Ticket

https://openscience.atlassian.net/browse/OSF-8982

screen shot 2018-02-09 at 6 20 54 am

screen shot 2018-02-09 at 6 21 33 am


addons/base/views.py
```diff
@@ -693,6 +694,30 @@ def addon_view_or_download_file(auth, path, provider, **kwargs):
...
+    if not file:
+        guid = Guid.load(id_or_guid)
+        if guid:
+            file = guid.referent
+        else:
+            raise HTTPError(httplib.NOT_FOUND, data={
+                'message_short': 'File Not Found',
+                'message_long': 'The requested file could not be found.'
+            })
+
```
chennan47: What if the download link points to a folder rather than a file for osfstorage?
Below is the error I got when I have a folder inside a project and try to click the download link in the api
screen shot 2018-01-24 at 11 49 00 am

website/routes.py
```diff
@@ -1168,6 +1168,12 @@ def make_url_map(app):
             addon_views.addon_view_or_download_file,
             OsfWebRenderer('project/view_file.mako', trust=False)
         ),
+        Rule(
+            '/<pid>/download/<fid_or_guid>/',
+            'get',
+            addon_views.persistent_file_download,
+            json_renderer,
+        ),
```
chennan47: Ideally, the persistent download link should be addressed in V2 API instead of building up another V1 API.

laurenrevere: Talked with @brianjgeiger about this as that's what I first did. He said that these base osf routes would be guaranteed where as API routes would not be.

addons/base/views.py
laurenrevere: Ah thanks for catching this. I shouldn't have links for folders at all.

Johnetordoff:
LGTM 👍  We might want to do the same thing for zipped folder downloads, this setup would make that super easy.

laurenrevere:
@Johnetordoff, I agree that would be a good future improvement!

api/files/serializers.py
```diff
@@ -380,3 +385,24 @@ def absolute_url(self, obj):
 
     def get_absolute_url(self, obj):
         return self.self_url(obj)
+
+    def get_download_link(self, obj):
+        return get_file_download_link(
+            self.context['view'].get_file(), version=obj.identifier,
```
sloria: I'm a little wary of how this might affect performance. Won't this result in an additional query for each version in the response? @laurenrevere Would you mind seeing how response times and # queries are affected by this in BAPI locally?

sloria:
Discussed on Flowdock: we can save dupe queries from `self.get_file()` using something like:

```python
def get_queryset(self):
    file = self.get_file()
    # Assign file to view so that we can use it in the serializer context
    self.file = file
    ...

def get_serializer_context(self):
    ...
    context['file'] = self.file
    return context
```

website/routes.py
```diff
@@ -1170,6 +1170,12 @@ def make_url_map(app):
             addon_views.addon_view_or_download_file,
             OsfWebRenderer('project/view_file.mako', trust=False)
         ),
+        Rule(
+            '/<pid>/download/<fid_or_guid>/',
```
sloria: @laurenrevere Why is the `<pid`> necessary?

laurenrevere: It's used for permissions `@must_be_contributor_or_public` uses it.

sloria: discussed in person: Remove `/pid/` from the path because it's extra and these URLs will be exposed to users.

sloria:
Once the path is updated, this is good to merge.

website/project/decorators.py
```diff
@@ -421,3 +397,31 @@ def wrapper(*args, **kwargs):
             )
         return func(*args, **kwargs)
     return wrapper
+
+def check_contributor_auth(node, auth, include_public, include_view_only_anon):
```
sloria: 👍

addons/base/views.py
```diff
@@ -694,6 +695,38 @@ def addon_view_or_download_file(auth, path, provider, **kwargs):
...
+            raise HTTPError(httplib.NOT_FOUND, data={
+                'message_short': 'File Not Found',
+                'message_long': 'The requested file could not be found.'
+            })
+    if not file.is_file:
+        raise HTTPError(httplib.BAD_REQUEST, data={
+            'message_long': 'Downloading folders is not permitted.'
+        })
+
+    auth_redirect = check_contributor_auth(file.node, auth, True, True)
```
sloria: Nit: Would be more readable to pass the last two arguments as keyword arguments, e.g., `include_public=True`. I can do this when I merge.

sloria:
Merged into `release/next-next-platform`.

PR #8108, opened 2018-01-18
Add deleted datetime field to models with is_deleted [PLAT-512]

erinspace:
## Purpose

Add a `deleted` field to all models that have `is_deleted`, change other models that have similar fields with different names. Also add a `deleted` field to preprints.

## Changes

- Add `deleted` fields of type `NonNaiveDateTimeField` to models with `is_deleted` Boolean field
- Also add `deleted` field to the `PreprintService` model
- Change naming for models that had fields like `deleted_on` or `deleted_date`
- Set deleted datetime when object is deleted and unset when undeleted
- Migrations to add fields and back populate where possible
- If a node was deleted and doesn't have a 
- Update preprint detail's `perform_destroy` method to just set the `deleted` field to the current datetime instead of actually deleting it from the database
- Update `get_object` for PreprintMixin filter for preprints that have `null` in the `deleted` field

## QA Notes

This change most likely doesn't need a lot of QA, as it's mostly backend related. The new deleted field isn't yet for the most part used for anything at the moment.

## Side Effects
Preprints used to be actually deleted when sending a `delete` request through the API, but no longer are and simply marked `deleted` - this is the only significant possibly breaking change.

For preprints, the `deleted` field will largely remain unpopulated - the migration here only adds a deleted datetime from the log when the preprint's node has been marked as deleted. This was done, instead of relying on all the other various ways a preprint could be considered "deleted" (node made private, is an orphan, not verified_publishable, etc) as changes in the upcoming Node Preprint Divorce will hopefully make this a lot clearer and more straightforward. 

## Ticket
https://openscience.atlassian.net/browse/PLAT-512


Johnetordoff:
@erinspace Unless there's some sort of president can we be a little more explicit and called `deleted` `deleted_date` or something? It's just to me `deleted` and `is_deleted` both sound like bools.

erinspace:
@Johnetordoff The idea was to be as consistent with the newly consistent `created` and `modified` datetime fields. Also, `is_deleted` is going away in the near future and this was partially to prepare for that

Johnetordoff:
@erinspace makes sense 👍

brianjgeiger:
Replaced by #9133

PR #8110, opened 2018-01-19
[PLAT-413] Make folder list `?info` query parameter that returns the folder info

Johnetordoff:
## Purpose

Current hitting the file list view v2 endpoint for a folder give us the metadata  that folder’s children. It would be really convenient if we could get the metadata for the folder itself without have the folder’s id and seeing it on the file detail endpoint.

## Changes 

- The get_query_from_request method on the file list view to take a new query param ‘info’ so metadata for the desired folder will be returned.
- Add line to the docs about new param
- Adds test.

## QA Notes

Hit an api endpoint for a folder using a path, so something like `/v2/nodes/fmsaw/files/googledrive/test/` this should show you what's inside folder 'test' on your googledrive addon. The try `/v2/nodes/fmsaw/files/googledrive/test/?info=`, this should show you info about folder 'test' itself.

## Side Effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/PLAT-413

## developer.osf.io update PR

https://github.com/CenterForOpenScience/developer.osf.io/pull/6


api/nodes/views.py
```diff
@@ -1952,6 +1954,23 @@ def get_default_queryset(self):
     def get_queryset(self):
         return self.get_queryset_from_request().distinct()
 
+    def get_queryset_from_request(self):
```
caseyrollins: I don't think `get_queryset_from_request` needs to be overridden here. All of this logic could be placed in `get_queryset`.

api/nodes/views.py
```diff
@@ -1873,6 +1873,8 @@ class NodeFilesList(JSONAPIBaseView, generics.ListAPIView, WaterButlerMixin, Lis
 
     + `filter[<fieldname>]=<Str>` -- fields and values to filter the search results on.
 
+    + `info=` -- when present this param returns the metadata for the folder it's pointed to, instead of that folder's children. This has no effect on files.
```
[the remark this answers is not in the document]
Johnetordoff: Done @ CenterForOpenScience/developer.osf.io#6

api/nodes/views.py
```diff
@@ -1950,7 +1952,21 @@ def get_default_queryset(self):
...
+            if isinstance(fobj, list):
+                node = self.get_node(check_object_permissions=False)
+                base_class = BaseFileNode.resolve_class(self.kwargs[self.provider_lookup_url_kwarg], BaseFileNode.FOLDER)
+                return base_class.objects.filter(node=node, _path=path)
+            elif isinstance(fobj, OsfStorageFolder):
+                return BaseFileNode.objects.filter(id__in=[fobj.id])
+            else:
+                raise NotFound
+        else:
+            return super(NodeFilesList, self).get_queryset_from_request().distinct()
```
caseyrollins: Minor: should be `self.get_queryset_from_request().distinct()`

api/nodes/views.py
```diff
@@ -956,7 +956,21 @@ def get_default_queryset(self):
 
     # overrides ListAPIView
     def get_queryset(self):
-        return self.get_queryset_from_request().distinct()
+        path = self.kwargs[self.path_lookup_url_kwarg]
+        # query param info when used on a folder gives that folder's metadata instead of the metadata of it's children
+        if 'info' in self.request.query_params and path.endswith('/'):
```
sloria: It might be more user-friendly to raise an error if `info` is passed on a non-folder.

Johnetordoff: I don't really think so. Imagine a someone wants metadata on a path and they don't know if it's a file or a folder, with a 'info' parameter they are guaranteed that that item's metadata. If it throws an error they'll have to put in logic to retry the request without the parameter.

sloria: Ah, I didn't realize that the client receives the same metadata for files as folders with `?info=true`. In that case, I agree with you, and this is fine as is.

PR #8113, opened 2018-01-19
[#PLAT-542]Update Identifier creation to no longer create ARKs

chennan47:
## Purpose
Update the identifier process, where we allow users to click 'Create DOI/ARK' to only create DOIs going forward.
Continue to display ARKs that were previously created
Update the UI so that users no longer see the option to Create ARKs, and only have the option to get a DOI for their content.

## Changes
Now you should only see 'Create DOI' for public projects we are admin on. For all new created DOI, it will only have DOI. For all existing identifier, it will show DOI and ARK.

## QA Notes
1. go to a public project you have already created DOI/ ARK, everything should be the same.
2. create a new public project
3. You should see ' Create DOI' instead of 'Create DOI/ ARK'
4. After clicking that link and DOI is created, you should only see 'DOI: ........' without any 'ARK:.......'

## Ticket
https://openscience.atlassian.net/browse/PLAT-542

### DIFF
--- a/osf/models/preprint_service.py
+++ b/osf/models/preprint_service.py
@@ -228,10 +228,8 @@ class PreprintService(DirtyFieldsMixin, GuidMixin, IdentifierMixin, ReviewableMi
         if save:
             self.save()
 
-    def set_identifier_values(self, doi, ark=None, save=False):
+    def set_identifier_values(self, doi, save=False):
         self.set_identifier_value('doi', doi)
-        if ark:
-            self.set_identifier_value('ark', ark)
         self.preprint_doi_created = timezone.now()
 
         if save:
--- a/osf/models/preprint_service.py
+++ b/osf/models/preprint_service.py
@@ -228,9 +228,10 @@ class PreprintService(DirtyFieldsMixin, GuidMixin, IdentifierMixin, ReviewableMi
         if save:
             self.save()
 
-    def set_identifier_values(self, doi, ark, save=False):
+    def set_identifier_values(self, doi, ark=None, save=False):
         self.set_identifier_value('doi', doi)
-        self.set_identifier_value('ark', ark)
+        if ark:
+            self.set_identifier_value('ark', ark)
         self.preprint_doi_created = timezone.now()
 
         if save:
--- a/tests/test_identifiers.py
+++ b/tests/test_identifiers.py
@@ -234,10 +234,7 @@ class TestIdentifierViews(OsfTestCase):
             res.json['doi'],
             self.node.get_identifier_value('doi')
         )
-
-        with assert_raises(KeyError):
-            ark = res.json['ark']
-
+        assert_not_in(res.json.keys(), 'ark')
         assert_equal(res.status_code, 201)
 
     @httpretty.activate
--- a/tests/test_identifiers.py
+++ b/tests/test_identifiers.py
@@ -234,12 +234,11 @@ class TestIdentifierViews(OsfTestCase):
             res.json['doi'],
             self.node.get_identifier_value('doi')
         )
-        assert_equal(
-            res.json['ark'],
-            self.node.get_identifier_value('ark')
-        )
-        assert_equal(res.status_code, 201)
 
+        with assert_raises(KeyError):
+            ark = res.json['ark']
+
+        assert_equal(res.status_code, 201)
 
     @httpretty.activate
     def test_create_identifiers_exists(self):
--- a/website/identifiers/utils.py
+++ b/website/identifiers/utils.py
@@ -130,7 +130,11 @@ def parse_identifiers(ezid_response):
             'ark': '{0}{1}'.format(settings.ARK_NAMESPACE.replace('ark:', ''), suffix),
         }
     else:
-        return {'doi': resp['success'].strip().split(':')[1]}
+        identifiers = dict(
+            [each.strip('/') for each in pair.strip().split(':')]
+            for pair in resp['success'].split('|')
+        )
+        return identifiers['doi']
 
 
 def get_or_create_identifiers(target_object):
--- a/website/identifiers/utils.py
+++ b/website/identifiers/utils.py
@@ -103,7 +103,8 @@ def request_identifiers_from_ezid(target_object):
                 raise
             resp = client.get_identifier(doi)
             new = False
-
+        if new:
+            resp = {'success': resp['success'].split('|')[0]}
         return {
             'response': resp,
             'already_exists': already_exists,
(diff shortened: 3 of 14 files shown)


website/identifiers/utils.py
```diff
@@ -94,19 +94,24 @@ def request_identifiers_from_ezid(target_object):
 
         client = get_ezid_client()
         already_exists = False
+        new = True
```
erinspace: It seems as though `new` is just the opposite of `already_exists` -- since `already_exists` is being set and used already, can you just use this variable instead of setting and keeping track of a new one?

website/identifiers/utils.py
```diff
@@ -94,19 +94,24 @@ def request_identifiers_from_ezid(target_object):
...
             resp = client.create_identifier(doi, metadata)
         except HTTPError as error:
             already_exists = True
             if 'identifier already exists' not in error.message.lower():
                 raise
             resp = client.get_identifier(doi)
-
+            new = False
+        if new:
+            resp = {'success': resp['success'].split('|')[0]}
```
erinspace: Instead of parsing the response manually twice, it might be better to leave the response alone here in the `request_identifiers_from_ezid` method, and do the parsing later in `parse_identifiers` below.  Parsing only once leaves less room for error and would be easier to catch if something changes with the response in the future

website/identifiers/utils.py
```diff
@@ -125,10 +130,8 @@ def parse_identifiers(ezid_response):
             'ark': '{0}{1}'.format(settings.ARK_NAMESPACE.replace('ark:', ''), suffix),
         }
     else:
-        return dict(
-            [each.strip('/') for each in pair.strip().split(':')]
-            for pair in resp['success'].split('|')
-        )
+        return {'doi': resp['success'].strip().split(':')[1]}
```
erinspace: Instead of relying on an index for pulling out the DOI (I'm just nervous about the service ever changing the order of their response) what about leaving the first response as a dict and explicitly taking the `doi` key:

```
identifiers = dict([each.strip('/') for each in pair.strip().split(':')] for pair in resp['success'].split('|'))
return identifiers['doi']
```

website/identifiers/utils.py
```diff
@@ -142,16 +145,16 @@ def get_or_create_identifiers(target_object):
 
         resp = response_dict['response']
         exists = response_dict['already_exists']
-
+        new = response_dict['new']
```
erinspace: same as above, can you just use the `exists` parameter here instead of using `new`?

website/identifiers/utils.py
```diff
@@ -142,16 +145,16 @@ def get_or_create_identifiers(target_object):
...
+                    'ark': '{0}{1}'.format(settings.ARK_NAMESPACE.replace('ark:', ''), suffix),
+                }
+            else:
+                return {'doi': doi.replace('doi:', '')}
         else:
-            return dict(
-                [each.strip('/') for each in pair.strip().split(':')]
-                for pair in resp['success'].split('|')
-            )
+            return {'doi': resp['success'].strip().split(':')[1]}
```
erinspace: I think this if statement can be simplified to not need a nested `if/else` if you use `exists` instead of `new`. Also see above comment about using a specific dict key of `doi` instead of relying on an index

website/static/js/nodeControl.js
```diff
@@ -196,9 +200,9 @@ var ProjectViewModel = function(data, options) {
         bootbox.confirm({
             title: 'Create identifiers',
             message: '<p class="overflow">' +
-                'Are you sure you want to create a DOI and ARK for this ' +
-                $osf.htmlEscape(self.nodeType) + '? DOI and ARK identifiers' +
-                ' are persistent and will always resolve to this page.',
+                'Are you sure you want to create a DOI for this ' +
+                $osf.htmlEscape(self.nodeType) + '? DOI  identifier' +
```
erinspace: Change to "A DOI" -- DOI stands for "Digital Object Identifier" so the extra identifier is redundant, but adding an "A" will make it a complete sentence

website/identifiers/utils.py
chennan47: this 'new' is not just opposite of already_exist, 
already_exist applies to any projects that DOI is already generated, which means that project can have valid ARK
new only applies to any project that has DOI after this change, which means that project only have DOI and no ARK.

website/identifiers/utils.py
chennan47: Ideally yes, but this is used by two different functions for preprints and nodes.

website/identifiers/utils.py
chennan47: It should return {'doi': '.......'}, but for identifiers['doi'], the return value is a str instead of a dict. Rebuild a dict like return {'doi': identifiers['doi']} is not worthy.

website/static/js/nodeControl.js
```diff
@@ -196,9 +200,9 @@ var ProjectViewModel = function(data, options) {
         bootbox.confirm({
             title: 'Create identifiers',
             message: '<p class="overflow">' +
-                'Are you sure you want to create a DOI and ARK for this ' +
-                $osf.htmlEscape(self.nodeType) + '? DOI and ARK identifiers' +
-                ' are persistent and will always resolve to this page.',
+                'Are you sure you want to create a DOI for this ' +
+                $osf.htmlEscape(self.nodeType) + '? A DOI identifier' +
```
erinspace: Remove the word "identifier"

tests/test_identifiers.py
```diff
@@ -234,12 +234,11 @@ def test_create_identifiers_not_exists(self):
             res.json['doi'],
             self.node.get_identifier_value('doi')
         )
-        assert_equal(
-            res.json['ark'],
-            self.node.get_identifier_value('ark')
-        )
-        assert_equal(res.status_code, 201)
 
+        with assert_raises(KeyError):
```
erinspace: This would be slightly a clearer assert if it was something like `assert_not_in(res.json.keys(), 'ark')`, as you don't really need to save the `ark` variable

osf/models/preprint_service.py
```diff
@@ -228,9 +228,10 @@ def set_preprint_license(self, license_detail, auth, save=False):
         if save:
             self.save()
 
-    def set_identifier_values(self, doi, ark, save=False):
+    def set_identifier_values(self, doi, ark=None, save=False):
```
erinspace: This is more a question, does `ark` even still need to be in the signature for `set_identifier_values`? How often outside of tests will we ever be setting the values?

erinspace: You should also update the useage of `set_identifier_values` in the `get_and_set_preprint_identifiers` method of `website/preprints/tasks.py` --

        preprint.set_identifier_values(doi=id_dict['doi'], ark=id_dict['ark'], save=True)

 if `parse_identifiers` returns an `id_dict` here with no `ark` key, this would raise a key error

erinspace:
Looks great! 🍨

sloria:
Merged into `release/next-platform`. 🚢 🇮🇹

